PR #917, opened 2020-01-09
Make RuleSampler the default sampler

marcotc:
This PR changes the default sampler to the new `RuleSampler`.

No behaviour should change for existing applications, as we fall back to the current `RateByServiceSampler` when no custom rule is configured.

Having the `RuleSampler` as the default allows for simpler configuration for the global sampling rate, rate limit and custom sampling rules.

PR #2128, opened 2022-07-05
Single Span Sampling

marcotc:
This PR adds support for single span sampling to the tracer.

Single Span Sampling allows you to:
```md
You can configure sampling rule that allow you keep spans despite their
respective traces being dropped by trace-level sampling.
```

It is configured through the documented environment variables: `DD_SPAN_SAMPLING_RULES`,`ENV_SPAN_SAMPLING_RULES_FILE`

All changes in this feature branch have been individually reviewed.

PR #2287, opened 2022-09-21
Fix:Agent sampling rates not applied when environment is set

marcotc:
The agent returns sampling rates to the tracer whenever traces are flushed.
These agent rates are applied based on a trace's `service` and `env` properties: the agent generates unique sampling rates for each combination of `service` and `env` that it receives.

When `env` is set in `ddtrace` (through `DD_ENV` or `Datadog.configure { |c| c.env = ... }`), we should have tried to match rates with that same `env`, but were matching against `'env:'` because of a classic Symbol vs String Hash access mistake:
```ruby
# We used
tags[:env]
# When we should have been using
tags['env']
```

This PR now uses the correct environment for matching agent rates.

This is not a new bug, introduced in https://github.com/DataDog/dd-trace-rb/pull/917 in the https://github.com/DataDog/dd-trace-rb/releases/tag/v0.32.0 release.

This issue was only caught because I was adding Feature Tests for Sampling Decision Mechanism, backfilling all sampling methods.

This PR backports he Feature Tests for agent-based sampling rates.

**I believe this is a strong indicator that we need more feature-level testing when delivering any sizeable new features.**


spec/datadog/tracing/integration_spec.rb
```diff
@@ -24,9 +24,31 @@
   shared_context 'agent-based test' do
```
marcotc, line 24: I'm overhauling this file in the Sampling-Propagation feature branch.
A few of the `shared_context`s have been backported here to ease testing and reduce merging conflicts.
They seem like overkill now, but will simplify and increase correctness of agent-based testing quite a bit.

spec/datadog/tracing/integration_spec.rb
```diff
@@ -312,13 +338,6 @@ def agent_receives_span_step3(previous_success)
         c.tracing.priority_sampling = priority_sampling if priority_sampling
       end
 
-      # Capture trace segments as they are about to be serialized
-      allow_any_instance_of(Datadog::Transport::Traces::Transport)
-        .to receive(:send_traces).and_wrap_original do |function, traces|
-          trace_segments.concat(traces)
-          function.call(traces)
-        end
-
```
marcotc, line 321: This approach was used by Single Span Sampling testing, but it's incomplete: it captures trace segments before they are formatted. Unformatted spans can look quite different from formatted ones.

This is specific snippet is now obsolete and actually conflicting with the new `trace_segments` capture at the beginning of this file above:
```ruby
      # Capture trace segments as they are about to be serialized
      segments = trace_segments
      allow_any_instance_of(Datadog::Transport::TraceFormatter)
        .to receive(:format!).and_wrap_original do |original|
          segments << original.call
        end
```

spec/datadog/tracing/integration_spec.rb
```diff
@@ -312,12 +338,7 @@ def agent_receives_span_step3(previous_success)
         c.tracing.priority_sampling = priority_sampling if priority_sampling
       end
 
-      # Capture trace segments as they are about to be serialized
-      allow_any_instance_of(Datadog::Transport::Traces::Transport)
-        .to receive(:send_traces).and_wrap_original do |function, traces|
-          trace_segments.concat(traces)
-          function.call(traces)
-        end
```
marcotc, line 320: This approach was used by Single Span Sampling testing, but it's incomplete: it captures trace segments before they are formatted. Unformatted spans can look quite different from formatted ones.

This is specific snippet is now obsolete and actually conflicting with the new `trace_segments` capture at the beginning of this file above:
```ruby
      # Capture trace segments as they are about to be serialized
      segments = trace_segments
      allow_any_instance_of(Datadog::Transport::TraceFormatter)
        .to receive(:format!).and_wrap_original do |original|
          segments << original.call
        end
```

spec/datadog/tracing/integration_spec.rb
```diff
@@ -312,12 +338,7 @@ def agent_receives_span_step3(previous_success)
...
+      WebMock.enable!
```
marcotc, line 341: These `WebMock` changes are due to a flaky test, `sets the Datadog-Client-Computed-Top-Level header to a non-empty value`, that can fail if `WebMock` was disabled by a previous test case.

This is unrelated to this PR, but became very prominent when I added more WebMock related tests to this file.

The fix is in this PR because CI broke in this very branch due to this flakiness.

spec/datadog/tracing/integration_spec.rb
```diff
@@ -619,24 +716,19 @@ def sample!(trace)
...
-        expect(stats[:transport].server_error).to eq(0)
-        expect(stats[:transport].internal_error).to eq(0)
+      expect(stats[:traces_flushed]).to eq(1)
+      expect(stats[:transport].client_error).to eq(0)
+      expect(stats[:transport].server_error).to eq(0)
+      expect(stats[:transport].internal_error).to eq(0)
 
-        expect(out).to have_received(:puts)
-      end
+      expect(out).to have_received(:puts)
```
marcotc, line 731: This test created 3 traces but didn't wait enough time for the Writer thread to flush them all.

Then, it waited for exactly 1 of the traces to be flushed and asserted 3 times that only 1 span was flushed.
Adding a sleep in this iteration shows that all 3 spans will be flushed.

I simplified the test, as the 3-time assertions don't add value.
The only difference between the spans is the `trace.sampling_priority = i`, but sampling priority won't affect this code path (all spans will be flushed at this point, regardless of sampling priority).

This is also a backport from the Sampling Decision Mechanism feature.

It's backported here because it broke CI in this branch :(

spec/datadog/tracing/integration_spec.rb
```diff
@@ -619,24 +716,19 @@ def sample!(trace)
...
```
marcotc, line 731: This test created 3 traces but didn't wait enough time for the Writer thread to flush them all.

Then, it waited for exactly 1 of the traces to be flushed and asserted 3 times, in quick succession, that only 1 span was flushed.
Modifying this test by adding a `sleep` in the iteration shows that all 3 spans will be flushed normally.

I simplified the test, as the 3-time assertions don't add value.
The only difference between the spans is the `trace.sampling_priority = i`, but sampling priority won't affect this code path (all spans will be flushed at this point, regardless of sampling priority).

This is also a backport from the Sampling Decision Mechanism feature.

It's backported here because it broke CI in this branch 😢

spec/datadog/tracing/integration_spec.rb
```diff
@@ -577,6 +599,81 @@ def sample!(trace)
...
+
+    context 'with agent rates' do
+      before do
+        WebMock.enable!
+        stub_request(:post, %r{/v0.4/traces}).to_return(status: 200, body: service_rates.to_json)
+      end
+
+      after { WebMock.disable! }
+
+      let(:service_rates) { { rate_by_service: { 'service:kept,env:' => 1.0, 'service:dropped,env:' => Float::MIN } } }
```
ivoanjo, line 611: This `Float::MIN` made me curious, and I did some grepping in the codebase. I see that we reference it in `rate_by_key_sampler_spec.rb` as being "for testing purposes".

My reading from that is that this value would never be returned by the agent. Is that really the case? If so, can I convince you to add a bit of context here about what this value means as well?

**Edit**: A second pass left me wondering: is `Float::MIN` relevant at all to what we're testing here? Could we instead remove it?

spec/datadog/tracing/integration_spec.rb
```diff
@@ -577,6 +599,81 @@ def sample!(trace)
...
+        end
+
+        context 'with a span not matching the environment rates' do
+          before do
+            tracer.trace('kept.span', service: 'kept', tags: { 'env' => 'not-right' }) {}
+            try_wait_until { tracer.writer.stats[:traces_flushed] >= 1 }
+          end
+
+          it_behaves_like 'priority sampled', 1.0
+        end
```
ivoanjo, line 674: I... don't think we're testing the right thing here? In particular, I reverted the change to `rule_sampler.rb` and the test still passed?

spec/datadog/tracing/integration_spec.rb
ivoanjo: May I suggest moving the `enable!`/`disable!` to the `context` that actually uses it, rather than enabling it for the whole file?

spec/datadog/tracing/integration_spec.rb
```diff
@@ -24,9 +24,31 @@
...
+
+      # Ensure background Writer worker doesn't wait, making tests faster.
+      stub_const('Datadog::Tracing::Workers::AsyncTransport::DEFAULT_FLUSH_INTERVAL', 0)
+
+      # Capture trace segments as they are about to be serialized
+      segments = trace_segments
+      allow_any_instance_of(Datadog::Transport::TraceFormatter)
+        .to receive(:format!).and_wrap_original do |original|
+          segments << original.call
+        end
```
ivoanjo, line 36: I know this was already like this but... is it just me or is this super wonky? Could we use webmock or something similar to look at what comes out of the pipe without interceding in it?

spec/datadog/tracing/integration_spec.rb
```diff
@@ -24,9 +24,31 @@
...
+    end
+
+    let(:trace_segments) { [] }
+    let(:span) do
+      expect(trace_segments).to have(1).item
+      expect(trace_segments[0].spans).to have(1).item
+
+      trace_segments[0].spans[0]
+    end
+    let(:sampling_priority) { span.get_tag('_sampling_priority_v1') }
```
ivoanjo, line 51: Minor: If I'm looking at this correctly, I think this can be moved closer to the tests that actually need it. I think it's especially important to try to do that in a file like `tracing/integration_spec.rb` because it naturally will accumulate more and more complexity over time and so anything we can do to make it easier to understand and maintain I think is worth it.

spec/datadog/tracing/integration_spec.rb
ivoanjo: 👍  I very much like this change :)

ivoanjo (commented):
The change itself looks reasonable, but the tests don't seem to be working correctly 😅😛

spec/datadog/tracing/integration_spec.rb
marcotc: Webmock has it serialized as msgpack. Maybe I can try to deserialize it from there. 🤔

marcotc:
This whole PR was pretty much done in #2128

PR #2289, opened 2022-09-22
Fix OpenTracing context propagation

### LINKED ISSUE #2191: Unusual behavior using manual instrumentation of distributed traces using OpenTracer in v1.x
Hi,

We’re encountering an issue when trying to doing the major version upgrade of our codebase from ddtrace v0.54.2 to v1.2.0

We’re using OpenTracing and doing manual instrumentation between different types of services: apis —> jobs. On the jobs side, it doesn’t seem like we’re able to successfully resume a trace like we were in the 0.54.x version.

**Current behaviour**
When attempting to upgrade to v1.2.0, we are missing many spans from our downstream job execution (activesupport, memcached, etc.) and two of our custom traces are out of order
image

**Expected behaviour**
On the v0.54.2 version of ddtrace, we see the expected trace hierarchy and spans from many integrations
image

**Steps to reproduce**

I’ve written a small spec that fails that might be related to the problem - it seems like there’s an issue using the extracted trace context from `OpenTracing.extract(…)`. https://github.com/okhwaja/ddtrace-bug is the repo and the test code is here: https://github.com/okhwaja/ddtrace-bug/blob/main/spec/v1_repro_spec.rb. The setup instructions are in the README

One odd thing I noticed when looking in code, [the RackPropagator always returns a nil value for datadog_context](https://cs.github.com/DataDog/dd-trace-rb/blob/f1a8956baa5c1f552c89f8a6c72cc04e075e5d5a/lib/datadog/opentracer/rack_propagator.rb#L53-L57) since it seems a migration was made to use trace_digests instead. Then in `Datadog::OpenTracer::Tracer.start_active_span`, [we use that always nil value](https://cs.github.com/DataDog/dd-trace-rb/blob/f1a8956baa5c1f552c89f8a6c72cc04e075e5d5a/lib/datadog/opentracer/tracer.rb#L95-L99). Note that if `child_of` were not passed in, [the code would generate a span with a not-null datadog_context](https://cs.github.com/DataDog/dd-trace-rb/blob/f1a8956baa5c1f552c89f8a6c72cc04e075e5d5a/lib/datadog/opentracer/tracer.rb#L81) and use that instead.

Happy to provide more information if you need it! Appreciate any guidance or tips you’re able to provide - thanks!

**Environment**

* **ddtrace version:** v1.2.0
* **Configuration block (`Datadog.configure ...`):** I can provide a full version if needed - summarized version below
```
# Important: OpenTracing.global_tracer must be assigned before the Datadog.configure block.
datadog_opentracer = Datadog::OpenTracer::Tracer.new(
  enabled: true, ...
)
OpenTracing.global_tracer = datadog_opentracer

Datadog.configure do |c|
  c.tracing.instrument :rack, request_queuing: true,
    service_name: "#{@service_name}-rack",
    web_service_name: "#{@service_name}-proxy"
  c.tracing.instrument :rails, request_queuing: true

  c.tracing.instrument :active_support, cache_service: "#{@service_name}-dbcache"

  c.tracing.instrument :active_record,
    analytics_enabled: false,
    service_name: "#{@service_name}-activerecord"
  c.tracing.instrument :dalli,
    analytics_enabled: false
  c.tracing.instrument :elasticsearch,
    analytics_enabled: false
  c.tracing.instrument :faraday,
    analytics_enabled: false
  c.tracing.instrument :http,
    analytics_enabled: false,
    split_by_domain: _to_bool(ENV["DATADOG_APM_HTTP_SPLIT_BY_DOMAIN"])
  c.tracing.instrument :redis,
    analytics_enabled: false
  c.tracing.instrument :sidekiq,
    analytics_enabled: false,
    service_name: "#{@service_name}"
  c.tracing.instrument :aws
end
```
* **Ruby version:** v2.7.4
* **Operating system:**
* **Relevant library versions:** [see repro repo](https://github.com/okhwaja/ddtrace-bug/blob/main/example.gemspec#L34-L38)

delner:
Fixes #2191 

Context propagation in OpenTracing was not working properly. Ordinarily, each new `Span` should get its own `SpanContext`, containing the relevant `Datadog::Tracing::Context` state. Instead, of building this `SpanContext` with the current state, it would build it with the previous state in inherited from its parent. This would case the child to inherit the grandparent's ID as its parent ID.

This change makes sure we build a new `SpanContext` per `Span`, and only propagate the relevant `Datadog::Tracing::Context`.

### DIFF
--- a/lib/datadog/opentracer/tracer.rb
+++ b/lib/datadog/opentracer/tracer.rb
@@ -151,12 +151,10 @@ module Datadog
           tags: tags || {}
         )
 
-        # Build or extend the OpenTracer::SpanContext
-        span_context = if parent_span_context
-                         SpanContextFactory.clone(span_context: parent_span_context)
-                       else
-                         SpanContextFactory.build(datadog_context: datadog_context)
-                       end
+        span_context =  SpanContextFactory.build(
+                          datadog_context: datadog_context || datadog_tracer.send(:call_context),
+                          baggage: parent_span_context ? parent_span_context.baggage.dup : {}
+                        )
 
         # Wrap the Datadog span and OpenTracer::Span context in a OpenTracer::Span
         Span.new(datadog_span: datadog_span, span_context: span_context)
--- a/lib/datadog/opentracer/tracer.rb
+++ b/lib/datadog/opentracer/tracer.rb
@@ -151,10 +151,10 @@ module Datadog
           tags: tags || {}
         )
 
-        span_context =  SpanContextFactory.build(
-                          datadog_context: datadog_context || datadog_tracer.send(:call_context),
-                          baggage: parent_span_context ? parent_span_context.baggage.dup : {}
-                        )
+        span_context = SpanContextFactory.build(
+          datadog_context: datadog_context || datadog_tracer.send(:call_context),
+          baggage: parent_span_context ? parent_span_context.baggage.dup : {}
+        )
 
         # Wrap the Datadog span and OpenTracer::Span context in a OpenTracer::Span
         Span.new(datadog_span: datadog_span, span_context: span_context)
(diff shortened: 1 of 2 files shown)


marcotc (approved):
Confirmed to work in a production setting.

PR #2291, opened 2022-09-27
Action:Automatic pull request labeler

marcotc:
This PR automatically adds GitHub labels to PRs where it's safe to assume their intent.

Some examples are PRs that:
* Only change documentation files. 
* Only change test files.
* Only change release-related files.
* Changes files in a specific product directory (e.g. ci, profiling).

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,38 @@
+# Configuration for GitHub Action `.github/workflows/pull-request-labeler.yml`.
+# See https://github.com/marketplace/actions/labeler for more information.
+
+# Only documentation changes
+docs:
+  - all: [ '**/*.md' ]
+
+# Only test changes
+dev/testing:
+  - all: [ '{spec/**,integration/**}' ]
+
+# Changes to Profiling
+profiling:
+  - any: [ 'lib/datadog/profiling/**' ]
+
+# Changes to CI-App
+ci-app:
+  - any: [ 'lib/datadog/ci/**' ]
+
+# Changes to ASM
+appsec:
+  - any: [ 'lib/datadog/appsec/**' ]
+
+# Changes to Tracing integrations
+integrations:
+  - any: [ 'lib/datadog/tracing/contrib/**' ]
+
+# Only repository GitHub changes
+dev/github:
+  - all: [ '.github/**' ]
+
+# Only repository CI changes
+dev/ci:
+  - all: [ '{.circleci/**,.gitlab-ci.yml}' ]
+
+# Version bump pull request
+release:
+  - all: [ '{CHANGELOG.md,lib/ddtrace/version.rb,gemfiles/**}' ]
--- a/.github/workflows/pull-request-labeler.yml
+++ b/.github/workflows/pull-request-labeler.yml
@@ -1,0 +1,16 @@
+name: "Pull Request Labeler"
+on:
+  - pull_request_target
+
+jobs:
+  triage:
+    permissions:
+      contents: read
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - uses: kachkaev/labeler@d89797c51d07680aec17049cc6790e9d323d9a93 # actions/labeler@v4 + https://github.com/actions/labeler/pull/316
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
+          configuration-path: .github/labeler.yml
+          dot: true # From https://github.com/actions/labeler/pull/316


.github/labeler.yml
```diff
@@ -0,0 +1,38 @@
+# Configuration for GitHub Action `.github/workflows/pull-request-labeler.yml`.
+# See https://github.com/marketplace/actions/labeler for more information.
+
+# Only documentation changes
+docs:
+  - all: [ '**/*.md' ]
+
+# Only test changes
+dev/testing:
+  - all: [ '{spec/**,integration/**}' ]
```
ivoanjo, line 10: Minor: Small additions 😄 

```suggestion
dev/testing:
  - all: [ '{spec/**,integration/**,benchmarks/**,sorbet/**}' ]
```

.github/labeler.yml
```diff
@@ -0,0 +1,38 @@
...
+
+# Changes to Profiling
+profiling:
+  - any: [ 'lib/datadog/profiling/**' ]
```
ivoanjo, line 14: Minor: Add the C stuff in as well

```suggestion
profiling:
  - any: [ '{lib/datadog/profiling/**,ext/ddtrace_profiling_loader/**,ext/ddtrace_profiling_native_extension/**}' ]
```

.github/labeler.yml
```diff
@@ -0,0 +1,38 @@
...
+dev/github:
+  - all: [ '.github/**' ]
+
+# Only repository CI changes
+dev/ci:
+  - all: [ '{.circleci/**,.gitlab-ci.yml}' ]
+
+# Version bump pull request
+release:
+  - all: [ '{CHANGELOG.md,lib/ddtrace/version.rb,gemfiles/**}' ]
```
ivoanjo, line 38: Really minor: Since we've done a few releases where we didn't bump all the gemfiles, I wonder if it would make sense to relax this rule to take that into account.

ivoanjo (approved):
👍 Thanks for being our GitHub wizard and pushing our automation forward :)

.github/labeler.yml
lloeki: Is that so? When a release is done at least `ddtrace (x.y.z)` should get an update in the lockfiles, correct?

.github/labeler.yml
```diff
@@ -0,0 +1,38 @@
...
+ci-app:
+  - any: [ 'lib/datadog/ci/**' ]
+
+# Changes to ASM
+appsec:
+  - any: [ 'lib/datadog/appsec/**' ]
+
+# Changes to Tracing integrations
+integrations:
+  - any: [ 'lib/datadog/tracing/contrib/**' ]
```
lloeki, line 26: Should there be `'lib/datadog/appsec/contrib/**'` here too? Maybe `'lib/datadog/ci/contrib/**' as well?

lloeki: Also should there be a general tracing one as well?

```
tracing:
  - any: [ 'lib/datadog/tracing/**' ]
```

lloeki: Not sure what the behaviour is when multiple criteria apply (e.g resulting in `AppSec`+`Integrations` tags, which sounds good to me if someone changes AppSec integration code).

lloeki: Also this non-product one sounds OK to me as well:

```
core:
  - any: [ 'lib/datadog/core/**' ]
```

.github/labeler.yml
```diff
@@ -0,0 +1,38 @@
...
+dev/testing:
+  - all: [ '{spec/**,integration/**}' ]
+
+# Changes to Profiling
+profiling:
+  - any: [ 'lib/datadog/profiling/**' ]
+
+# Changes to CI-App
+ci-app:
+  - any: [ 'lib/datadog/ci/**' ]
```
kachkaev, line 18: Thanks for giving my fork a go folks! Does this glob work the same as `lib/datadog/ci/**/*`? I can update docs in https://github.com/actions/labeler/pull/316 if so.

.github/labeler.yml
ivoanjo: Yes and no. The way we've set up CI is that the CI runners will do the update locally anyway, thus we can actually not update it and things will still be fine.

But we do want the lockfiles to be updated from time to time, so this is why we have it as a step of the release process.

So my question/suggestion was more of a: should we consider here the exceptional cases where we do a release and skip the lockfile update?

marcotc: I'm removing gemfile bump from this list, to ensure we have no false positives on the `release` tag.

.github/labeler.yml
marcotc: Applied all 3 suggested changes.

.github/labeler.yml
marcotc: Thank you for the changes, @kachkaev! Couldn't use the GH Action without it! 🙇

Yes, it works the same.
```js
minimatch("root/a/b/c", "root/**/*")
// true
minimatch("root/a/b/c", "root/**")
// true 
```

PR #2292, opened 2022-09-27
Allow DD_TAGS values to have the colon character

marcotc:
Colons are allowed in Datadog tags: they are considered part of the key's "value".

From: https://docs.datadoghq.com/getting_started/tagging/#define-tags

> <p> The key always precedes the first colon of the global tag definition, for example:</p>
> 
> Tag | Key | Value
> -- | -- | --
> env:staging:east | env | staging:east

This PR allows for DD_TAGS to contain colons in the tag "value" component.


ivoanjo (approved):
👍 Looks great, thanks for the fix!

PR #2293, opened 2022-09-27
fix: AppSec Rails should check for railties instead of rails meta gem

seuros:
As discussed in slack. 
rails gem is a meta gem and should not be used to detect the presence of the rails framework.

If an application is running with stripped gem-set , the AppSec Extension fail to load with a comparison  error `nil >= MINIMUM_VERSION`


marcotc (approved):
Thank you so much, @seuros!

lloeki:
Hey @seuros, this change was just released in v1.5.0!

PR #2296, opened 2022-09-28
Fix `service_nam` typo to `service_name`

jackweinbender:
Just a simple typo fix in the configuration file of the Graphql integration. I don't know if this is the source of some of my recent issues, but if the option is going to be there... it should be right.


lloeki:
Hey @jackweinbender, this change was just released in v1.5.0!

PR #2282, opened 2022-09-16
Sampling-Propagation:Set sampling decision mechanism tag

marcotc:
Follow up from #2266

This PR adds a trace-level tag to propagate the sampling mechanism used to make a sampling decision: `_dd.p.dm`.

The currently defined mechanisms are (many unused):
```ruby
        module Mechanism
          # Used before the tracer receives any rates from agent and there are no rules configured.
          DEFAULT = 0
          # The sampling rate received in the agent's http response.
          AGENT_RATE = 1
          # Auto. Reserved for future use.
          REMOTE_RATE_AUTO = 2
          # Sampling rule or sampling rate based on tracer config.
          TRACE_SAMPLING_RULE = 3
          # User directly sets sampling priority via {Tracing.reject!} or {Tracing.keep!},
          # or by a custom sampler implementation.
          MANUAL = 4
          # Formerly AppSec.
          ASM = 5
          # User-defined target. Reserved for future use.
          REMOTE_RATE_USER = 6
          # Reserved for future use.
          REMOTE_RATE_BY_DATADOG = 7
          # Single Span Sampled.
          SPAN_SAMPLING_RATE = 8
        end
```

The sampling mechanism can either come from an upstream service, if the sampling decision was performed upstream, or added by one of `ddtrace`s sampling classes.

This PR also adds many feature test around different sampling mechanism (e.g. manual sampling, agent-based rates) in order to test for the newly added sampling mechanism tag. This unveiled a few bugs that have been fixed in separated PRs.

## `_dd.p.dm` serialization format

One quirk of the implementation is due to the historical format of `_dd.p.dm` on other tracers: `_dd.p.dm` used to be comprised of two pieces of information, separated by a dash (`-`), e.g. `info1-sampling_mechanism`.

`info1` is now removed but `sampling_mechanism` stays. This means that `_dd.p.dm` will look like this: `-sampling_mechanism`.

This ensures users using old versions of Datadog products that support the `-sampling_mechanism` format can still participate in the sampling propagation chain.

PR #2298, opened 2022-09-28
Sampling-Propagation:Export/import sampling decision through distributed trace tags

marcotc:
Follows up from #2282.

~This PR ensures the `TraceOperation#sampling_mechanism` populated in #2282 is injected and extracted during distributed tracing.~

This PR became trivial because the sampling decision is now considered a simple, opaque tag.

Only tests were added to cover the propagation specific tags, but not code changes are needed.


delner (approved):
:+1:

PR #2299, opened 2022-09-28
Sampling-Propagation:Exclude problematic existing tag

marcotc:
Follow up from #2298

The distributed tracing tag `_dd.p.upstream_services` was at one point populated by other language traces, but we do not need this information anymore.

Although the Ruby tracer has never populated this tag, we must make sure we don't propagate this tag downstream if we ever receive it in a distributed tracing request context.


ivoanjo (approved):
👍 Looks reasonable!

delner (approved):
:+1:

PR #2240, opened 2022-08-24
Upgrade to libdatadog 0.8.0.1.0

ivoanjo:
**What does this PR do?**:

Upgrade dd-trace-rb to use the latest libdatadog release.

This latest version brings a lot of renames to clean up the last
remnants of the library being called libddprof instead of libdatadog.

This PR only updates our code to use the new names, and otherwise
behavior is unchanged (and all specs are still passing).

**Motivation**:

The 0.8.0 includes at least one new feature that will be needed in the new Ruby profiler.

**Additional Notes**:

I am opening this as a draft PR until libdatadog 0.8.0 gets released on Rubygems.org.
This can already be tested locally as described in https://github.com/DataDog/libdatadog/pull/44.

**Also, we may want to wait until after the dd-trace-rb 1.4.0 release goes out to merge this PR.**

**How to test the change?**:

Because this PR doesn't really change any behavior, our existing test suite should be enough to validate that things remain working (and it's green for me locally -- it should become green in CI once libdatadog gets released on rubygems.org).

PR #2062, opened 2022-06-01
[PROF-6070] Remove legacy profiling transport

ivoanjo:
In #2059 we changed the profiler to use the new `HttpTransport` to report profiling data, but we left around the legacy transport code:

> Afterwards, I got to thinking that if customers do report any issue that we suspect may be related to HttpTransport, it would be really great if there was an easy way to test that out by simply toggling a configuration option, rather than customers needing to compare different versions of dd-trace-rb. So, I've revived the old code, but changed it to comply to the new interfaces and structures. This makes it trivial to remove it again: just revert the Retrofit legacy profiling transport implementation as a backup to `HttpTransport` class and Allow enabling legacy profiling transport codepath via setting/env variable commits.

The 1.2.0->1.5.0 releases have defaulted to use the new `Profiling::HttpTransport`, and other than #2151/#2152 which was solved in 1.3.0, no other issues have been reported so I think we're ready to remove the old code.

PR #2302, opened 2022-10-03
Upgrade to libdatadog 0.9.0.1.0

ivoanjo:
**What does this PR do?**

Upgrade dd-trace-rb to use the latest libdatadog release (0.9.0.1.0).

This PR also incorporates the upgrade to version 0.8.0.1.0 initially worked on in #2240 but that ended up being blocked by https://github.com/DataDog/libdatadog/issues/47 and thus never merged.

**I strongly recommend reviewing this PR commit-by-commit 😄**

Expect the following changes:

* A bunch of libdatadog APIs were renamed (the API names were remnants of the library being called libddprof in the past).
* Removed compression from `Profiling::Exporter`, since libdatadog handles compression now
* Adapted to use the new libdatadog 0.9 API in `Profiling::HttpTransport`
* The `Profiling::HttpTransport` integration tests now expect libdatadog to report in the intake v2.4 format, including automatic lz4 compression of data.
* Marked `Datadog::Core::Utils::Compression` as deprecated for removal, since profiling was its only user

**Motivation**

Keep up with latest improvements and features being added to libdatadog.

**Additional Notes**

I'm opening this PR:
* ~~as a draft because I'm still waiting for the libdatadog 0.9 stable release~~
* on top of #2062 since that PR removes a few files that we otherwise would need to update

**How to test the change?**

The changes in this PR either include test coverage or are already covered by existing tests.


ivoanjo:
Libdatadog 0.9 has now been released and is up on rubygems.org. Marking this PR as ready for review :)

lib/datadog/profiling/exporter.rb
```diff
@@ -60,10 +59,9 @@ def flush
           start: start,
           finish: finish,
           pprof_file_name: Datadog::Profiling::Ext::Transport::HTTP::PPROF_DEFAULT_FILENAME,
-          pprof_data: Datadog::Core::Utils::Compression.gzip(uncompressed_pprof),
+          pprof_data: uncompressed_pprof.to_s,
```
marcotc, line 62: Is the data just left uncompressed here now? I assume it's going to be compressed later down the line, before it's flushed out of the process?

lib/datadog/core/utils/compression.rb
```diff
@@ -6,10 +6,13 @@
 module Datadog
   module Core
     module Utils
-      # Common database-related utility functions.
+      # Compression/decompression utility functions.
+      #
+      # @deprecated This is no longer used by ddtrace and will be removed in 2.0.
```
marcotc, line 11: Does it make sense to point to the new Compression replacement, or is there none?

spec/datadog/profiling/http_transport_spec.rb
```diff
@@ -291,23 +292,28 @@
...
+          'family' => 'ruby',
+          'version' => '4',
         )
 
-        tags = body['tags[]'].list
-        expect(tags).to contain_exactly('tag_a:value_a', 'tag_b:value_b')
+        require 'extlz4' # Lazily required, to avoid trying to load it on JRuby
+
+        expect(LZ4.decode(body.fetch(pprof_file_name))).to eq pprof_data
+        expect(LZ4.decode(body.fetch(code_provenance_file_name))).to eq code_provenance_data
```
marcotc, line 316: I can't find to understand where this LZ4 compression is happening now.
I can only assume it now lives in libdatadog, but I can't find any hints in non-tests that

ivoanjo: Yes. The compression now entirely lives inside libdatadog, and is transparent. When calling `ddog_ProfileExporter_send` the byte arrays it receives with the data as compressed and sent to the backend.

> I can only assume it now lives in libdatadog, but I can't find any hints outside of this file of any LZ4 compression (or any form of compression) being done

You're right. And in hindsight maybe it's a bit too cryptic on the test.

I'll push a commit to move these tests to a specific `it 'reports the payload as lz4-compressed files, that get compressed by libdatadog'` so that this fact is clearly called out and documented in the code.

lib/datadog/core/utils/compression.rb
ivoanjo: There isn't any, since libdatadog ships the lz4 compression inside itself and does not (currently) expose it to callers. (Which is why I needed to get the `extlz4` gem for the rspec testing)

lib/datadog/profiling/exporter.rb
ivoanjo: Exactly; it goes from the `Profiling::Exporter` to the `Profiling::HttpTransport` uncompressed. The lifetime for the pprof data is quite short (until the report is successful), so I don't expect this to have a noticeable impact on application memory.

spec/datadog/profiling/http_transport_spec.rb
ivoanjo: Fixed in c8920081cb84180a8f824d34cb7532467353ce01

marcotc (approved):

PR #2265, opened 2022-09-09
Use full URL in `http.url` tag

lloeki:
**What does this PR do?**

*Adjust HTTP URL tags to comply with span tag unification RFC*

- align http.url with tag naming RFC (and Rack's #url) to include the
  complete request URL
- drop http.base_url, now unneeded
- preserve REQUEST_URI logic that differs from Rack's logic
- add SCRIPT_NAME which was previously unhandled for PATH_INFO
- append QUERY_STRING when using PATH_INFO to match REQUEST_URI

*Enable query string quantization by default*
    
This change is mandated by the unifying tag RFC.

**Motivation**

Compliance with internal RFC document.

PR #2311, opened 2022-10-12
Handle URLs with invalid characters

### LINKED ISSUE #2309: dd-trace causes `bad URI(is not URI?)` with invalid URIs
**Current behaviour**
If the URL contains a `|`, then rails actually handles the request (it might return route not found / 404 etc), but dd-trace seems to explode. If I remove dd-trace from our system, it works as expected.

```
curl "http://localhost:3000/|"

Puma caught this error: bad URI(is not URI?): "http://localhost:3000/|" (URI::InvalidURIError)
/usr/local/lib/ruby/3.0.0/uri/rfc3986_parser.rb:67:in `split'
/usr/local/lib/ruby/3.0.0/uri/rfc3986_parser.rb:72:in `parse'
/usr/local/lib/ruby/3.0.0/uri/common.rb:171:in `parse'
/usr/local/bundle/gems/ddtrace-1.5.0/lib/datadog/tracing/contrib/utils/quantization/http.rb:26:in `base_url'
/usr/local/bundle/gems/ddtrace-1.5.0/lib/datadog/tracing/contrib/rack/middlewares.rb:185:in `set_request_tags!'
/usr/local/bundle/gems/ddtrace-1.5.0/lib/datadog/tracing/contrib/rack/middlewares.rb:110:in `call'
/usr/local/bundle/gems/railties-7.0.4/lib/rails/engine.rb:530:in `call'
/usr/local/bundle/gems/puma-5.6.5/lib/puma/configuration.rb:252:in `call'
/usr/local/bundle/gems/puma-5.6.5/lib/puma/request.rb:77:in `block in handle_request'
/usr/local/bundle/gems/puma-5.6.5/lib/puma/thread_pool.rb:340:in `with_force_shutdown'
/usr/local/bundle/gems/puma-5.6.5/lib/puma/request.rb:76:in `handle_request'
/usr/local/bundle/gems/puma-5.6.5/lib/puma/server.rb:443:in `process_client'
/usr/local/bundle/gems/puma-5.6.5/lib/puma/thread_pool.rb:147:in `block in spawn_thread
```

**Expected behaviour**
It should handle invalid URIs safely.

**Steps to reproduce**
Start a rails app with dd-trace ; Go to `http://localhost:3000/|` or run `curl "http://localhost:3000/|"`

**Environment**

* **ddtrace version:**  1.5.0
```ruby
Datadog.configure do |c|
  c.tracing.enabled = false
  c.env = Rails.env
  c.agent.port = 8127
  c.service = "app_name"
  c.tracing.instrument :rails,
                       :controller_service => "rails-controller",
                       :cache_service => "rails-cache"

  # Normally these are not commented out, but I can reproduce even with those commented-out

  # c.tracing.instrument :redis
  # c.tracing.instrument :sidekiq
  # c.tracing.instrument :http
  # c.tracing.instrument :rack,
  #                      :quantize => {
  #                        :base => :exclude,
  #                        :query => :show,
  #                        :fragment => :show
  #                      }
  c.telemetry.enabled = false
end
```

* **Ruby version:** 3
* **Operating system:** Ubuntu 20.04
* **Relevant library versions:** Rails 7.0.4

lloeki:
**What does this PR do?**

Handle additional URL corner cases.

**Motivation**

Fixes #2309

**Additional Notes**

Not entirely sure about the expected `placeholder` behaviour but it's an undocumented feature.

**How to test the change?**

Specs.

### DIFF
--- a/lib/datadog/tracing/contrib/utils/quantization/http.rb
+++ b/lib/datadog/tracing/contrib/utils/quantization/http.rb
@@ -19,7 +19,9 @@ module Datadog
             def url(url, options = {})
               url!(url, options)
             rescue StandardError
-              options[:placeholder] || PLACEHOLDER
+              placeholder = options[:placeholder] || PLACEHOLDER
+
+              options[:base] == :exclude ? placeholder : "#{base_url(url)}/#{placeholder}"
             end
 
             def base_url(url, options = {})
@@ -28,6 +30,12 @@ module Datadog
                 uri.query = nil
                 uri.fragment = nil
               end.to_s
+            rescue StandardError
+              if (m = %r{([a-z]+://[^/]+)(?:/|$)}.match(url))
+                m[1]
+              else
+                ''
+              end
             end
 
             def url!(url, options = {})
--- a/lib/datadog/tracing/contrib/utils/quantization/http.rb
+++ b/lib/datadog/tracing/contrib/utils/quantization/http.rb
@@ -13,6 +13,8 @@ module Datadog
             include Kernel # Ensure that kernel methods are always available (https://sorbet.org/docs/error-reference#7003)
 
             PLACEHOLDER = '?'.freeze
+
+            # taken from Ruby 3.1.2 lib/uri/rfc3986_parser.rb and adjusted to stop at path
             RFC3986_URL_BASE = /\A(?<URI>(?<scheme>[A-Za-z][+\-.0-9A-Za-z]*):(?<hier-part>\/\/(?<authority>(?:(?<userinfo>(?:%\h\h|[!$&-.0-;=A-Z_a-z~])*)@)?(?<host>(?<IP-literal>\[(?:(?<IPv6address>(?:\h{1,4}:){6}(?<ls32>\h{1,4}:\h{1,4}|(?<IPv4address>(?<dec-octet>[1-9]\d|1\d{2}|2[0-4]\d|25[0-5]|\d)\.\g<dec-octet>\.\g<dec-octet>\.\g<dec-octet>))|::(?:\h{1,4}:){5}\g<ls32>|\h{1,4}?::(?:\h{1,4}:){4}\g<ls32>|(?:(?:\h{1,4}:)?\h{1,4})?::(?:\h{1,4}:){3}\g<ls32>|(?:(?:\h{1,4}:){,2}\h{1,4})?::(?:\h{1,4}:){2}\g<ls32>|(?:(?:\h{1,4}:){,3}\h{1,4})?::\h{1,4}:\g<ls32>|(?:(?:\h{1,4}:){,4}\h{1,4})?::\g<ls32>|(?:(?:\h{1,4}:){,5}\h{1,4})?::\h{1,4}|(?:(?:\h{1,4}:){,6}\h{1,4})?::)|(?<IPvFuture>v\h+\.[!$&-.0-;=A-Z_a-z~]+))\])|\g<IPv4address>|(?<reg-name>(?:%\h\h|[!$&-.0-9;=A-Z_a-z~])*))(?::(?<port>\d*))?)))(?:\/|\z)/.freeze # rubocop:disable Style/RegexpLiteral, Layout/LineLength
 
             module_function
--- a/lib/datadog/tracing/contrib/utils/quantization/http.rb
+++ b/lib/datadog/tracing/contrib/utils/quantization/http.rb
@@ -14,7 +14,10 @@ module Datadog
 
             PLACEHOLDER = '?'.freeze
 
-            # taken from Ruby 3.1.2 lib/uri/rfc3986_parser.rb and adjusted to stop at path
+            # taken from Ruby https://github.com/ruby/uri/blob/ffbab83de6d8748c9454414e02db5317609166eb/lib/uri/rfc3986_parser.rb
+            # but adjusted to parse only <scheme>://<host>:<port>/ components
+            # and stop there, since we don't care about the path, query string,
+            # and fragment components
             RFC3986_URL_BASE = /\A(?<URI>(?<scheme>[A-Za-z][+\-.0-9A-Za-z]*):(?<hier-part>\/\/(?<authority>(?:(?<userinfo>(?:%\h\h|[!$&-.0-;=A-Z_a-z~])*)@)?(?<host>(?<IP-literal>\[(?:(?<IPv6address>(?:\h{1,4}:){6}(?<ls32>\h{1,4}:\h{1,4}|(?<IPv4address>(?<dec-octet>[1-9]\d|1\d{2}|2[0-4]\d|25[0-5]|\d)\.\g<dec-octet>\.\g<dec-octet>\.\g<dec-octet>))|::(?:\h{1,4}:){5}\g<ls32>|\h{1,4}?::(?:\h{1,4}:){4}\g<ls32>|(?:(?:\h{1,4}:)?\h{1,4})?::(?:\h{1,4}:){3}\g<ls32>|(?:(?:\h{1,4}:){,2}\h{1,4})?::(?:\h{1,4}:){2}\g<ls32>|(?:(?:\h{1,4}:){,3}\h{1,4})?::\h{1,4}:\g<ls32>|(?:(?:\h{1,4}:){,4}\h{1,4})?::\g<ls32>|(?:(?:\h{1,4}:){,5}\h{1,4})?::\h{1,4}|(?:(?:\h{1,4}:){,6}\h{1,4})?::)|(?<IPvFuture>v\h+\.[!$&-.0-;=A-Z_a-z~]+))\])|\g<IPv4address>|(?<reg-name>(?:%\h\h|[!$&-.0-9;=A-Z_a-z~])*))(?::(?<port>\d*))?)))(?:\/|\z)/.freeze # rubocop:disable Style/RegexpLiteral, Layout/LineLength
 
             module_function
--- a/lib/datadog/tracing/contrib/utils/quantization/http.rb
+++ b/lib/datadog/tracing/contrib/utils/quantization/http.rb
@@ -13,6 +13,7 @@ module Datadog
             include Kernel # Ensure that kernel methods are always available (https://sorbet.org/docs/error-reference#7003)
 
             PLACEHOLDER = '?'.freeze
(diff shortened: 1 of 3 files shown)


lib/datadog/tracing/contrib/utils/quantization/http.rb
```diff
@@ -19,7 +19,9 @@ module HTTP
             def url(url, options = {})
               url!(url, options)
             rescue StandardError
-              options[:placeholder] || PLACEHOLDER
+              placeholder = options[:placeholder] || PLACEHOLDER
+
+              options[:base] == :exclude ? placeholder : "#{base_url(url)}/#{placeholder}"
             end
 
             def base_url(url, options = {})
```
sailor, line 27: Maybe another way of doing it would be to escape the url before trying to parse it with something like this:
```suggestion
            def base_url(url, options = {})
              parser = URI::RFC2396_Parser.new
              URI.parse(parser.escape_url(url)).tap to |uri|
```
With this url `http://example.com/|`, you get this parsable escaped_url `http://example.com/%7C`, and you don't have use the regexp in the rescue block.

lloeki: This would URL escape the whole URL, only for the non-base part to be thrown away, which seems a bit wasteful especially for longer `url`.

That said, your note makes me think about just using the regex in all cases instead of going through "full parsing" (which is a regex too), as it would save a couple of allocs and string walking.

Also, while unlikely, the regex should probably look more like [the beginning of this one](https://github.com/ruby/ruby/blob/4b1504ae0a5118153f8b47d7bd0fff7e61553d17/lib/uri/rfc3986_parser.rb#L5).

sailor: That's very true! Also yes the regexp from the stdlib is probably more precise. Thank's for the answer

lloeki: Thanks for your feedback! I made the changes as described.

lib/datadog/tracing/contrib/utils/quantization/http.rb
```diff
@@ -13,21 +13,24 @@
             include Kernel # Ensure that kernel methods are always available (https://sorbet.org/docs/error-reference#7003)
 
             PLACEHOLDER = '?'.freeze
+            RFC3986_URL_BASE = /\A(?<URI>(?<scheme>[A-Za-z][+\-.0-9A-Za-z]*):(?<hier-part>\/\/(?<authority>(?:(?<userinfo>(?:%\h\h|[!$&-.0-;=A-Z_a-z~])*)@)?(?<host>(?<IP-literal>\[(?:(?<IPv6address>(?:\h{1,4}:){6}(?<ls32>\h{1,4}:\h{1,4}|(?<IPv4address>(?<dec-octet>[1-9]\d|1\d{2}|2[0-4]\d|25[0-5]|\d)\.\g<dec-octet>\.\g<dec-octet>\.\g<dec-octet>))|::(?:\h{1,4}:){5}\g<ls32>|\h{1,4}?::(?:\h{1,4}:){4}\g<ls32>|(?:(?:\h{1,4}:)?\h{1,4})?::(?:\h{1,4}:){3}\g<ls32>|(?:(?:\h{1,4}:){,2}\h{1,4})?::(?:\h{1,4}:){2}\g<ls32>|(?:(?:\h{1,4}:){,3}\h{1,4})?::\h{1,4}:\g<ls32>|(?:(?:\h{1,4}:){,4}\h{1,4})?::\g<ls32>|(?:(?:\h{1,4}:){,5}\h{1,4})?::\h{1,4}|(?:(?:\h{1,4}:){,6}\h{1,4})?::)|(?<IPvFuture>v\h+\.[!$&-.0-;=A-Z_a-z~]+))\])|\g<IPv4address>|(?<reg-name>(?:%\h\h|[!$&-.0-9;=A-Z_a-z~])*))(?::(?<port>\d*))?)))(?:\/|\z)/
```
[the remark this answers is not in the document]
lloeki: taken from Ruby 3.1.2 lib/uri/rfc3986_parser.rb and adjusted to stop at path

lib/datadog/tracing/contrib/utils/quantization/http.rb
```diff
@@ -14,20 +14,25 @@ module HTTP
...
             module_function
 
             def url(url, options = {})
               url!(url, options)
             rescue StandardError
-              options[:placeholder] || PLACEHOLDER
+              placeholder = options[:placeholder] || PLACEHOLDER
+
+              options[:base] == :exclude ? placeholder : "#{base_url(url)}/#{placeholder}"
             end
```
ivoanjo, line 28: I don't quite understand this change... Previously, if there was any error, we returned the placeholder (either the default one, or a provided one).

But now, we add the `base_url` unless `options[:base]` is `:exclude`. Which is not entirely unreasonable, but this does not seem a fix, but rather a behavior change, and a default behavior change at that.

Can you doublecheck if this is indeed what is intended?

lib/datadog/tracing/contrib/utils/quantization/http.rb
```diff
@@ -14,20 +14,25 @@ module HTTP
 
             PLACEHOLDER = '?'.freeze
 
+            # taken from Ruby 3.1.2 lib/uri/rfc3986_parser.rb and adjusted to stop at path
+            RFC3986_URL_BASE = /\A(?<URI>(?<scheme>[A-Za-z][+\-.0-9A-Za-z]*):(?<hier-part>\/\/(?<authority>(?:(?<userinfo>(?:%\h\h|[!$&-.0-;=A-Z_a-z~])*)@)?(?<host>(?<IP-literal>\[(?:(?<IPv6address>(?:\h{1,4}:){6}(?<ls32>\h{1,4}:\h{1,4}|(?<IPv4address>(?<dec-octet>[1-9]\d|1\d{2}|2[0-4]\d|25[0-5]|\d)\.\g<dec-octet>\.\g<dec-octet>\.\g<dec-octet>))|::(?:\h{1,4}:){5}\g<ls32>|\h{1,4}?::(?:\h{1,4}:){4}\g<ls32>|(?:(?:\h{1,4}:)?\h{1,4})?::(?:\h{1,4}:){3}\g<ls32>|(?:(?:\h{1,4}:){,2}\h{1,4})?::(?:\h{1,4}:){2}\g<ls32>|(?:(?:\h{1,4}:){,3}\h{1,4})?::\h{1,4}:\g<ls32>|(?:(?:\h{1,4}:){,4}\h{1,4})?::\g<ls32>|(?:(?:\h{1,4}:){,5}\h{1,4})?::\h{1,4}|(?:(?:\h{1,4}:){,6}\h{1,4})?::)|(?<IPvFuture>v\h+\.[!$&-.0-;=A-Z_a-z~]+))\])|\g<IPv4address>|(?<reg-name>(?:%\h\h|[!$&-.0-9;=A-Z_a-z~])*))(?::(?<port>\d*))?)))(?:\/|\z)/.freeze # rubocop:disable Style/RegexpLiteral, Layout/LineLength
```
ivoanjo, line 18: I suggest:

1. pointing at the latest canonical version (since this gem has been extracted into a separate repo that then gets syncd into Ruby) https://github.com/ruby/uri/blob/master/lib/uri/rfc3986_parser.rb
2. Including the copyright header for the library here, as well as mentioning this in `LICENSE-3rdparty.csv`
3. Adding a comment being more specific about what "adjusted to stop at path" means (e.g. to make maintenance easier in the future)

ivoanjo (approved):
This seems reasonable; I've left a couple of comments, I'm especially curious about the behavior change around exclude.

lib/datadog/tracing/contrib/utils/quantization/http.rb
lloeki: Glad you mention that!

> but rather a behavior change

Indeed https://github.com/DataDog/dd-trace-rb/pull/2265 changed the behaviour of `http.url` form being a path to being a proper URL.

So my reasoning was:

- `http.url` tag was path only
- `options[:placeholder]` replaced `http.url` tag on failure to parse the URI path, so 
- `http.base_url` tag was present still even on failure to parse the URI path
- there was no placeholder handling for `http.base_url`
- when `options[:base]` is `:show`, `http.url` tag goes from path to true URL
- when `options[:base]` is `:show`, `http.base_url` tag disappears and is folded into `http.url`
- To preserve the "keep url base and replace the failing path part" behaviour while having `http.url` be an actual URL, combine base with placeholder where the path would be.

Alternatives that made sense to me too are: 

- have the `options[:placeholder]` replace "whole" of `http.url` when a single part of it fails to be obtained (e.g even if we could obtain `http.base_url`), e.g `http://example.com/|` => `?`
- have the `options[:placeholder]` replace "failing parts" of `http.url`, combining it with the one we managed to have, e.g:
  - failing path: `http://example.com/|` => `http://example.com/?`
  - failing base url:
    - `http://<invalid>/path` => `?/path`
    - `<invalid>://example.com/path` => `?/path`
  - failing base url (alternative):
    - `http://<invalid>/path` => `http://?/path`
    - `<invalid>://example.com/path` => `?://example.com/path`

Note: `options[:placeholder]` does not seem to be documented, so I assumed it was some non-public configurable. Maybe it was an oversight?

WDYT?

lib/datadog/tracing/contrib/utils/quantization/http.rb
lloeki: > since this gem has been extracted into a separate repo that then gets syncd into Ruby

Good point.

> pointing at the latest canonical version (since this gem has been extracted into a separate repo that then gets syncd into Ruby)

Then I would rather point at the commit from which it was taken to make it a permalink.

lib/datadog/tracing/contrib/utils/quantization/http.rb
ivoanjo: > Note: options[:placeholder] does not seem to be documented, so I assumed it was some non-public configurable. Maybe it was an oversight?

Sigh. I somewhat wish we were more strict with these kinds of things -- e.g. allowing customers to just provide a hash that then gets fed to an internal class directly without any filter. This means that `:placeholder` became a public config setting, even though we may not have even intended it, because you can provide it via `c.tracing.instrument :rack, quantize: ...`

I think your approach here seems reasonable. Thanks for sharing your thoughts! :)

Issue #2312, opened 2022-10-12
`environment_logger.rb` Failed to collect environment information: undefined method `client'.


duncanista:
**Current behaviour**

`dd-trace@1.5.0` warns an error for the environment logger when using the writer `SyncWriter` in the configuration. `TraceWriter` doesn't trigger this warning, but I was recommended not to use it yet.

```sh
W, [2022-10-11T18:17:25.593797 #9]  WARN be6ea2a2-7353-42a1-8fae-7af333be7b68 -- ddtrace: [ddtrace] Failed to collect environment information: undefined method `client' for #<Datadog::Transport::IO::Client:0x0000000002a478e0> Location: /opt/ruby/gems/2.7.0/gems/ddtrace-1.5.0/lib/datadog/core/diagnostics/environment_logger.rb:113:in `agent_url'
```

Configuration
```rb
# under Datadog.configure do |c|
c.tracing.writer = Datadog::Tracing::SyncWriter.new(
  transport: Datadog::Transport::IO.default
)
```

**Expected behaviour**

Not to error, same behavior as `TraceWriter`.

```rb
c.tracing.writer = Datadog::Tracing::Workers::TraceWriter.new(
  transport: Datadog::Transport::IO.default
)
```

**Steps to reproduce**

Configure a Ruby application with the tracing writer `Datadog::Tracing::SyncWriter` with the transport `Datadog::Transport::IO.default`.

```rb
Datadog.configure do |c|
  c.tracing.writer = Datadog::Tracing::SyncWriter.new(
    transport: Datadog::Transport::IO.default
  )
  c.tags = { "_dd.origin": 'lambda' }
  yield(c) if block_given?
end
```

**How does `ddtrace` help you?**

To support `datadog-lambda-rb` layer for AWS Lambda.

**Environment**

* **ddtrace version: `1.5.0`
* **Configuration block (`Datadog.configure ...`): (described above)
* **Ruby version: `2.7.0`
* **Operating system: `Lambda Runtime` | `macOS M1`
* **Relevant library versions: `n/a`

PR #2313, opened 2022-10-13
Fix agent_url in environment logger when io transport 

TonyCTHsu:
**What does this PR do?**

Mitigate the [error](https://github.com/DataDog/dd-trace-rb/issues/2312) from environment logger when IO transport is configured.

The environment logger, seems to be only working for HTTP transport, the object has client attribute, and api and adapter so on…, but not for IO transport

### DIFF
--- a/lib/datadog/core/diagnostics/environment_logger.rb
+++ b/lib/datadog/core/diagnostics/environment_logger.rb
@@ -106,13 +106,13 @@ module Datadog
           Datadog.configuration.version
         end
 
-        # @return [String] target agent URL for trace flushing
+        # @return [String, nil] target agent URL for trace flushing
         def agent_url
           # Retrieve the effect agent URL, regardless of how it was configured
           transport = Tracing.send(:tracer).writer.transport
 
-          # return '' with IO transport
-          return '' unless transport.respond_to?(:client)
+          # return `nil` with IO transport
+          return unless transport.respond_to?(:client)
 
           adapter = transport.client.api.adapter
           adapter.url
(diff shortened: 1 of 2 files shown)


lib/datadog/core/diagnostics/environment_logger.rb
```diff
@@ -110,6 +110,10 @@ def dd_version
         def agent_url
           # Retrieve the effect agent URL, regardless of how it was configured
           transport = Tracing.send(:tracer).writer.transport
+
+          # return '' with IO transport
+          return '' unless transport.respond_to?(:client)
+
           adapter = transport.client.api.adapter
           adapter.url
         end
```
ivoanjo, line 119: I'm... not the biggest of fans of this approach. The issue showed up because we reach quite deep inside the transport object to get some information, and our "reaching" ended up being problematic. Let me inline it all so we can see how deep it is:

`Datadog::Tracing.send(:tracer).writer.transport.client.api.adapter.url`

This code is assuming a very very specific transport object, and will break with all others.

I feel that by probing for `respond_to?(:client)` we're solving the symptom (oops we need to be slightly more careful when reaching) and not the cause (we're reaching quite deep).

I won't block the PR on this (so I'll approve it anyway) BUT my suggestion would be to add an `#agent_url` method directly on the `Datadog::Transport::Traces::Transport` class.

I would also suggest having a useful fallback for others. E.g., if there's no agent url, perhaps we could instead use `"(custom transport: #{transport.class})"` or something like that.

Since the whole objective of this block is having something useful when reporting issues, I think having a clear indication that there's a non-standard config being used is quite valuable, instead of having it blank and whoever is looking at the issue will need to hunt around for a while to understand why.

ivoanjo (approved):
Left a few suggestions. I don't consider them blocking, but would really like to see some or all adressed.

Issue #2315, opened 2022-10-13
trace: incorrect request method when Rails renders error page (exceptions_app)


wilsonehusin:
**Current behaviour**

Given a Rails application setup, when a non-GET request is sent and the app raises Exception, [the recorded trace's HTTP method turns into GET](https://github.com/rails/rails/blob/04972d9b9ef60796dc8f0917817b5392d61fcf09/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L45).

**Expected behaviour**

Original client's HTTP request method is recorded.

Possible solution is to make Rack middleware to be ActionDispatch-aware, by fetching `env['action_dispatch.original_request_method']` first, then fallback to `env['REQUEST_METHOD']` if necessary.

https://github.com/DataDog/dd-trace-rb/blob/4e7ecdaa4b23cc8e6c5e38cb63fd54b90bfcb08a/lib/datadog/tracing/contrib/rack/middlewares.rb#L163-L165

**Steps to reproduce**

1. Rails configuration:
    - `config.exceptions_app = self.routes`
    - `config.consider_all_requests_local = false`
    - `config.action_dispatch.show_exceptions = true`
    - Add `ddtrace`
1. Code configuration:
    - In `config/routes.rb`:
        - Add non-GET path, e.g. `patch "/foo", to: "foo#patch"`
        - Add errors path, i.e. `get "/500", to: "errors#internal_server_error"`
    - Create `FooController`
      ```ruby
      class FooController < ApplicationController
        def patch
          raise Exception
        end
      end
      ```
    - Create `ErrorsController`
      ```ruby
      class ErrorsController < ApplicationController
        def internal_server_error
          format.any { head :internal_server_error, content-type: "text/plain" }
        end
      end
      ```
1. Run `curl -X PATCH localhost:3000/foo`
1. Trace object shows `meta['http.method']` as `GET`

**Environment**

* **ddtrace version:** 1.5.0
* **Configuration block (`Datadog.configure ...`):**
    ```ruby
    Datadog.configure do |c|
      ...
      c.tracing.enabled = enabled
      c.tracing.analytics.enabled = enabled
      c.tracing.log_injection = enabled
      c.runtime_metrics.enabled = enabled
      c.logger.instance = SemanticLogger["Datadog"]
      c.diagnostics.debug = Rails.env.development?
    
      c.tracing.instrument :rails, service_name: "myapp", exception_controller: ErrorsController
[...]

PR #2317, opened 2022-10-17
Fix Non-GET request method with rails exception controller

TonyCTHsu:
**What does this PR do?**

https://github.com/DataDog/dd-trace-rb/issues/2315

When instrumenting a non-GET request, a Rails app with [exceptions_app](https://guides.rubyonrails.org/configuring.html#config-exceptions-app) configured would mutated the request method to `GET` before dispatching to the exception handling app. 

https://github.com/rails/rails/blob/8015c2c2cf5c8718449677570f372ceb01318a32/actionpack/lib/action_dispatch/middleware/show_exceptions.rb#L45

Hence, it would be more accurate to fetch the value from the original rack env(prior to the mutation).


delner (approved):
Looks good! One failing system test (I think it's flaky). Try to re-run and get that green, otherwise, good to merge.

PR #2304, opened 2022-10-05
Support placeholder Garbage Collection frames in Collectors::Stack

ivoanjo:
**What does this PR do?**:

This PR adds a new feature to the `Collectors::Stack`: being able to include, at the top of the sampled stack, a placeholder frame saying "Garbage Collection".

**Motivation**:

The "Garbage Collection" placeholder is part of a new feature to show on the flamegraph the CPU and wall-clock time spent in garbage collection.

Right now we are not yet collecting this information, but I will follow up with a later PR to add this feature to `Collectors::CpuAndWallTime`.

**Additional Notes**:

~~I'm opening this PR as a draft because:~~
* ~~it's on top of #2302 to avoid a later rebase~~
* ~~libdatadog 0.9 has not been released yet, so CI fails~~

~~Thus, once libdatadog 0.9 is released and #2302 is merged, I'll re-trigger CI and mark this PR as non-draft when CI is back to green. I don't expect to need any code changes for that.~~

**How to test the change?**:

The changes to `Collectors::Stack` include code coverage. I plan to add later integration tests for the whole "gather GC time" feature.

PR #2323, opened 2022-10-20
[PROF-6260] Add support for tracking time spent in GC to CpuAndWallTime collector, second try

ivoanjo:
**What does this PR do?**:

This PR builds on top of https://github.com/DataDog/dd-trace-rb/pull/2304 to deliver the new "showing time spent in GC in the flamegraph feature". It does not yet offer the full solution: it is still missing the part where we integrate with the Ruby VM to automatically track the in-gc/not-in-gc state for threads. This final bit will come in a follow-up PR.

Specifically, this PR modifies the cpu and wall-time tracking in the `CpuAndWallTime` collector to introduce a new implicit state that a thread can be in: doing garbage collection.

This works by introducing three new functions:

* `cpu_and_wall_time_collector_on_gc_start`
* `cpu_and_wall_time_collector_on_gc_finish`
* `cpu_and_wall_time_collector_sample_after_gc`

When `cpu_and_wall_time_collector_on_gc_start` gets called, the current cpu and wall-time get recorded to the thread context: `cpu_time_at_gc_start_ns` and `wall_time_at_gc_start_ns`.

While these fields are set, regular samples (if any) do not account for any time that passes after these two timestamps.

(Regular samples can still account for the time between the previous sample and the start of GC.)

When `cpu_and_wall_time_collector_on_gc_finish` gets called, the current cpu and wall-time again get recorded to the thread context: `cpu_time_at_gc_finish_ns` and `wall_time_at_gc_finish_ns`.

Finally, when `cpu_and_wall_time_collector_sample_after_gc` gets called, the following happens:

1. A sample gets taken, using the special `SAMPLE_IN_GC` sample type, which produces a stack with a placeholder `Garbage Collection` frame as the latest frame. This sample gets assigned the cpu-time and wall-time period that was recorded between calls to `on_gc_start` and `on_gc_finish`.

2. The thread is no longer marked as being in gc (all gc tracking fields get reset back to `INVALID_TIME`).

3. The `cpu_time_at_previous_sample_ns` and `wall_time_at_previous_sample_ns` get updated with the elapsed time in GC, so that all time is accounted for -- e.g. the next sample will not get "blamed" by time spent in GC.

Looking at the history and at the name of this PR, you may notice that this is the "second try". The first try was never merged in https://github.com/DataDog/dd-trace-rb/pull/2308, and it unfortunately suffered from a flaw when used in the context of Ruby's garbage collection hooks -- it could trigger memory allocation in rare occasions (usually exceptions), which is actually not allowed to happen during Ruby's garbage collection start/finish hooks.

So I needed to rewrite the approach to make sure that `on_gc_start` and `on_gc_finish` can never allocate memory, and to move the sample generation to `sample_after_gc`, which the `Collectors::CpuAndWallTimeWorker` will make sure to call as soon as "safe" after GC ends.

**Motivation**:

This work will enable us to show the time spent in garbage collection in the flamegraph.

This time would previously be invisible, and accounted for as time for the unlucky Ruby method that was running when GC happened.

**Additional Notes**:

As discussed above, this PR is only a piece of the puzzle, and is still missing changes to the `CpuAndWallTimeWorker´ to integrate with the Ruby VM.

**How to test the change?**:

The change includes code coverage.


ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
```diff
@@ -147,7 +189,6 @@ static VALUE _native_new(VALUE klass) {
    // "numtable" is an awful name, but TL;DR it's what should be used when keys are `VALUE`s.
     st_init_numtable();
   state->recorder_instance = Qnil;
-  state->sample_count = 0;
```
marcotc, line 150: Is this initialized somewhere else now?

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
```diff
@@ -238,6 +292,190 @@ VALUE cpu_and_wall_time_collector_sample(VALUE self_instance) {
...
+  if (!rb_typeddata_is_kind_of(self_instance, &cpu_and_wall_time_collector_typed_data)) return;
+  // This should never fail the the above check passes
+  TypedData_Get_Struct(self_instance, struct cpu_and_wall_time_collector_state, &cpu_and_wall_time_collector_typed_data, state);
+
+  struct per_thread_context *thread_context = get_context_for(rb_thread_current(), state);
+
+  // If there was no previously-existing context for this thread, we won't allocate one (see safety). For now we just drop
+  // the GC sample, under the assumption that "a thread that is so new that we never sampled it even once before it triggers
+  // GC" is a rare enough case that we can just ignore it.
+  // We can always improve this later if we find that this happens often (and we have the counter to help us figure that out)!
```
marcotc, line 315: This seems like a reasonable scenario what we can't record the GC sample.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
```diff
@@ -238,6 +292,190 @@ VALUE cpu_and_wall_time_collector_sample(VALUE self_instance) {
...
+  if (thread_context == NULL) {
+    state->stats.gc_samples_missed_due_to_missing_context++;
+    return;
+  }
+
+  // If these fields are set, there's an existing GC sample that still needs to go out. As a simplification, we just drop
+  // the new GC sample, under the assumption that this won't happen often.
```
marcotc, line 322: Did you see this happening, or this more like "we can't fully guarantee this won't happen, but it would be weird if it did".

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
```diff
@@ -336,24 +609,45 @@ static VALUE _native_per_thread_context(DDTRACE_UNUSED VALUE _self, VALUE collec
...
 
-static long wall_time_now_ns() {
+// Safety: This function is assumed never to raise exceptions by callers when raise_on_failure == false
+static long wall_time_now_ns(bool raise_on_failure) {
   struct timespec current_monotonic;
 
-  if (clock_gettime(CLOCK_MONOTONIC, &current_monotonic) != 0) rb_sys_fail("Failed to read CLOCK_MONOTONIC");
+  if (clock_gettime(CLOCK_MONOTONIC, &current_monotonic) != 0) {
+    if (raise_on_failure) rb_sys_fail("Failed to read CLOCK_MONOTONIC");
+    else return 0;
```
marcotc, line 644: Does zero have a special meaning? I ask because returning zero on failure seems odd in C.

marcotc: I'd think that `INVALID_TIME` would be the return here, based on the rest of the code base.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
marcotc: As a follow up, does it make sense for us to reset the values of `cpu_time_at_finish_ns` and `wall_time_at_finish_ns` if we find ourselves in an invalid state?

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
```diff
@@ -238,6 +292,190 @@ VALUE cpu_and_wall_time_collector_sample(VALUE self_instance) {
...
+// This function gets called shortly after Ruby has finished running the Garbage Collector.
+// It creates a new sample including the cpu and wall-time spent by the garbage collector work, and resets any
+// GC-related tracking.
+//
+// Specifically, it will search for thread(s) which have gone through a cycle of on_gc_start/on_gc_finish
+// and thus have cpu_time_at_start_ns, cpu_time_at_finish_ns, wall_time_at_start_ns, wall_time_at_finish_ns
+// set on their context.
+//
+// Assumption 1: This function is called in a thread that is holding the Global VM Lock. Caller is responsible for enforcing this.
+// Assumption 2: This function is allowed to raise exceptions. Caller is responsible for handling them, if needed.
```
marcotc, line 377: I think it's worth explicitly mentioning that this method is allowed to allocate memory, in contrasts with `cpu_and_wall_time_collector_on_gc_finish` and `cpu_and_wall_time_collector_on_gc_start`.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
```diff
@@ -238,6 +292,190 @@ VALUE cpu_and_wall_time_collector_sample(VALUE self_instance) {
...
+  ddog_Slice_i64 metric_values_slice,
+  sample_type type
+) {
+  VALUE thread_name = thread_name_for(thread);
+  bool have_thread_name = thread_name != Qnil;
+
+  int label_count = 1 + (have_thread_name ? 1 : 0);
+  ddog_Label labels[label_count];
+
+  labels[0] = (ddog_Label) {.key = DDOG_CHARSLICE_C("thread id"), .str = thread_context->thread_id_char_slice};
```
marcotc, line 461: A bit unrelated, but where's the definition of `DDOG_CHARSLICE_C`? I can only find the definition of `VALUE_STRING`.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
ivoanjo: Good question!

Since the memory is allocated with `calloc` (more specifically, `ruby_xcalloc`, on line 184), it's guaranteed to be set to zero.

I was torn between adding initialization to all the new stats (`gc_samples`, `gc_samples_missed_due_to_missing_context`, `gc_samples_missed_due_to_missing_sample_after_gc`) as well, and not having it at all, since it's not needed.

I decided to not have the explicit initialization. to hopefully reduce some of the boilerplate, but let me know if this is confusing.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
ivoanjo: Indeed, that was my thinking!

Furthermore, in the future we'll want to integrate with the GVL tracing events, one of which is `RUBY_INTERNAL_THREAD_EVENT_STARTED` which is fired just after the thread is initialized.

Thus, if we do context initialization when that event gets triggered, this situation of "GC on a thread with no context" should be completely eliminated on Ruby 3.2+.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
ivoanjo: > Did you see this happening, or this more like "we can't fully guarantee this won't happen, but it would be weird if it did".

It's more of the latter, yes. In my older attempt at this (#2304) this couldn't happen, since the plan was to flush the sample immediately when `on_gc_finish` was called.

Because that didn't work out (as explained in the PR description), there is no way to guarantee that Ruby will not trigger more than one GC cycle before we can actually run the `sample_after_gc` method (which we'll queue for execution using `rb_postponed_job_register_one`, which is the same approach we use for regular samples).

But yeah I don't expect that to happen in a normal situation.

> As a follow up, does it make sense for us to reset the values of cpu_time_at_finish_ns and wall_time_at_finish_ns if we find ourselves in an invalid state?

I... don't think so, because I wouldn't consider this case an invalid state -- just a really rare one.

E.g., if we do find out that this is not as rare as we thought, I think the solution is to accumulate multiple samples --- e.g. maybe have extra fields that can collect the total cpu time / wall time that was not flushed yet across multiple GC executions ---  rather than just ignoring the sample that we did not process yet.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
ivoanjo: You raise a good point. Let me try to address it in parts.

It's not only `wall_time_now_ns` that returns 0 on failure -- this was already that `cpu_time_now_ns` did.

My intention in doing so is to simplify the callers -- both sources of time are always used to compute deltas: how much cpu/wall elapsed in this period, e.g. between two calls of `cpu/wall_time_now_ns`.

Returning `INVALID_TIME` would mean that the callers would need to keep checking -- is either of the values (period started, period finished) not valid? If so, they need to be handled separately, and usually the "handling" would be setting the period to 0.

Thus, by setting it to 0, the caller can just do `finish - start` and not care about the result.

For GC tracking in particular it also simplifies a bit the error handling. We know that `wall_time_at_start_ns`/`cpu_time_at_start_ns` will be something other than `INVALID_TIME` during a GC (e.g. until `sample_after_gc` gets called), which seems a bit simpler than handling the case where we got two errors and the values are set to `INVALID_TIME` because we failed to resolve them. We would probably need to introduce a `MISSING_TIME` or similar to avoid it.

TL;DR: Using 0 seemed slightly simpler to the caller functions, so I went with it, but if you feel this is confusing I can take a stab at using something else (`INVALID_TIME` or `MISSING_TIME`?).

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
ivoanjo: Agreed! Added in d991cf72f34f7a34db23f6a387256ded1d4966a5

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
ivoanjo: It's in `libdatadog/common.h`.

It gets set up here https://github.com/DataDog/libdatadog/blob/81b1753db3454c39f96446cae0c7397ded032a79/ddcommon-ffi/cbindgen.toml#L17-L23 and you can see the generated file by runnning `bundle open libdatadog` and finding the `common.h` file.

ivoanjo:
Thanks for the feedback :)

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
marcotc: Makes sense thank you!

Could you capture this in a comment:
> there is no way to guarantee that Ruby will not trigger more than one GC cycle before we can actually run the sample_after_gc method

I didn't think of this case of multiple GCs before `sample_after_gc`, but it's something important to keep in mind in the future.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
marcotc: `finish - start`: don't you have to still check for error if `finish` is zero but `start` is not? I see a few checks like this were introduced (`"BUG: Unexpected negative"`).

If the code actually contains fewer assertions like this by returning zero then sure. But if it's the same amount, then I suggest returning an invalid value marker. Whatever looks simpler.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
marcotc: Thank you!

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
ivoanjo: Added in 9477bba5fcdd5b5812ff7ea626814a4e01eaa4a3 (also added a bit more docs)

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
ivoanjo: I actually did see a potential bug relating to this in a follow-up PR I'm preparing.

Let me get back to you on this one...

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
ivoanjo: Actually... I found a bit of a bug around this, but decided that it was a feature instead and revised my approach yet again in 872f5af3550a2215631cfdba29f6b91fbbaf42d2.

The bug was when the following call order happened: `on_gc_start` (time=0) -> `on_gc_finish` (time=1) -> `on_gc_start` (time=2) -> `on_gc_finish` (time=3) -> `sample_after_gc`

As we've discussed here, the second `on_gc_start` would be a no-op (since there was a sample to be flushed), but the second `on_gc_finish` actually was not a no-op, because it forgot to test for the same situation.

But because the second `on_gc_finish` clobbered the existing finish timestamps with later ones, the final effect is that the two GCs were "collapsed" into one bigger one in our accounting, e.g. the `sample_after_gc` would use time=0 for start and time=3 for finish.

Which... is actually not bad semantics. If GCs are so back-to-back that we can't even get to `sample_after_gc`, it's kinda irrelevant if in between two GCs we actually still executed a handful Ruby bytecodes or not.

Thus, I've added specs to clarify that this behavior is intended, and updated the comments.

marcotc (approved):
🙇‍♂️

ivoanjo:
Merged in master to fix unrelated CI issues.

PR #2324, opened 2022-10-21
SQL comment propagation for `mysql2`

TonyCTHsu:
**What does this PR do?**

Inject sql comment to enable database monitoring. 

By default, it is `disabled`. 

With `service` mode, our tracer propagates service context to database, which enable Datadog to build service map.


lib/datadog/tracing/contrib/propagation/sql_comment/ext.rb
```diff
@@ -0,0 +1,19 @@
...
+
+module Datadog
+  module Tracing
+    module Contrib
+      module Propagation
+        module SqlComment
+          module Ext
+            ENV_SQL_COMMENT_PROPAGATION_MODE = 'DD_TRACE_SQL_COMMENT_PROPAGATION_MODE'.freeze
+
+            DISABLED = 'disabled'.freeze
```
marcotc, line 11: So ugly, I assume it's part of the RFC. What happened to `false` 😢?

lib/datadog/tracing/contrib/propagation/sql_comment/comment.rb
```diff
@@ -0,0 +1,37 @@
...
+            end
+
+            def to_s
+              @string ||= begin
+                ret = String.new
+
+                @hash.each do |key, value|
+                  next if value.nil?
+
+                  value = ERB::Util.url_encode(value) # url encode
```
marcotc, line 21: I recommend using the built-in [CGI.escape](https://rubyapi.org/2.7/o/cgi/util#method-i-escape) method.

lib/datadog/tracing/contrib/propagation/sql_comment/comment.rb
```diff
@@ -0,0 +1,37 @@
...
+                  value.gsub!("'", "/'")              # escaping single quote
```
marcotc, line 22: Don't forget the `# frozen_string_literal: true` at the top of this file, to ensure these strings are frozen and not reallocated every time.

lib/datadog/tracing/contrib/propagation/sql_comment/ext.rb
```diff
@@ -0,0 +1,21 @@
...
+            SERVICE = 'service'.freeze
+            FULL = 'full'.freeze
```
marcotc, line 13: Can you document, at high level, what SERVICE and FULL represent?

lib/datadog/tracing/contrib/propagation/sql_comment/ext.rb
```diff
@@ -0,0 +1,21 @@
...
+
+            TAG_DBM_TRACE_INJECTED = '_dd.dbm_trace_injected'.freeze
```
marcotc, line 15: I recommend documenting when this tag is applied, and what value it can be set to.

lib/datadog/tracing/contrib/propagation/sql_comment/comment.rb
```diff
@@ -0,0 +1,41 @@
...
+
+            def to_s
+              @string ||= begin
+                ret = String.new
+
+                @hash.each do |key, value|
+                  next if value.nil?
+
+                  value = ERB::Util.url_encode(value) # url encode
+                  value.tr!("'", "\'") # escaping single quote
```
marcotc, line 26: Given this:
```ruby
ERB::Util.url_encode("'")
# => "%27"
```

I don't think we can have `'` in `value` anymore at this point. I think we can remove this `tr!` call.

Also, the `tr!` arguments seem incorrect, given they are the same: 
```ruby
"'" == "\'"
# => true
```

spec/datadog/tracing/contrib/propagation/sql_comment/comment_spec.rb
```diff
@@ -0,0 +1,32 @@
...
+        "/*url_encode='DROP%20TABLE%20FOO'*/"
+      ],
+      [
+        { route: '/polls 1000' },
+        "/*route='%2Fpolls%201000'*/"
+      ],
+      [
+        { escape_single_quote: "Dunkin' Donuts" },
+        "/*escape_single_quote='Dunkin%27%20Donuts'*/"
+      ]
```
marcotc, line 27: Also, I recommend a test that ensures all characters we insert manually cannot come from user input.
We insert these three characters: `/`, `*`, and `=`. (e.g. `/*key=value*/`).

One test with these three characters would suffice.

spec/datadog/tracing/contrib/mysql2/patcher_spec.rb
```diff
@@ -103,6 +104,71 @@
             .to eq("Unknown column 'INVALID' in 'field list'")
         end
       end
+
+      context 'when sql comment propagation' do
+        let(:sql_statement) { 'SELECT 1' }
+
+        subject { client.query(sql_statement) }
+
+        shared_examples_for 'propagated with sql comment propagation' do |mode, span_op_name|
```
marcotc, line 113: These shared examples are probably also compatible with testing `pg` and other integrations that will be affected by this SqlComment propagation.

We don't need to do this for this PR, but I recommend sharing a few `shared_examples_for` when we add `pg`, for example.

lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -0,0 +1,49 @@
...
+          def self.annotate!(span_op, mode)
+            return unless mode.enabled?
+
+            span_op.set_tag(Ext::TAG_DBM_TRACE_INJECTED, true) if mode.full?
+          end
+
+          def self.prepend_comment(sql, span_op, mode, tags: {})
+            return sql unless mode.enabled?
+
+            tags.merge!(service_context)
```
marcotc, line 21: `service_context` is always overriding the values passed in `tags`. This means it is not possible, with this implementation, to override the service name.

I actually recommend reducing the API a bit here, and only accepting `service_name: nil`, instead of `tags: {}`. I don't see a scenario where we want to tag with a custom `env` or custom `version`, I only see `service` as mutable here. And we should make sure if service is provided, we don't override it with `datadog_configuration.env`.

lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -0,0 +1,49 @@
...
+            tags.merge!(trace_context(span_op)) if mode.full?
```
marcotc, line 22: I recommend commenting out this line for now, so that this can never run until we implement it, even if we mistakenly set it to `full`.

spec/datadog/tracing/contrib/propagation/sql_comment/comment_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+        { escape_single_quote: "Dunkin' Donuts" },
+        "/*escape_single_quote='Dunkin%27%20Donuts'*/"
+      ],
+      [
+        { star: '*' },
+        "/*star='%2A'*/"
+      ],
+      [
+        { back_slash: '/' },
+        "/*back_slash='%2F'*/"
```
marcotc, line 34: `/` is a forward slash :)

marcotc (approved):

marcotc:
Anything keeping this PR in draft state?

TonyCTHsu:
@marcotc I am still working on the documentation for this 😄

marcotc (approved):
🔥

PR #2325, opened 2022-10-24
Improve faraday test

TonyCTHsu:
**What does this PR do?**

Replace unnecessary variable `request_span` with `span`, since there is only one `span` and assertion also apply on the name of span


ivoanjo (approved):
LGTM 👍
Thanks for making stuff simpler :)

PR #2329, opened 2022-10-26
Forward port 1.5-stable

lloeki:
**What does this PR do?**

Sync version update and changelog back to master


lloeki:
Duh, seems like merging re-applied changes that should not.

TonyCTHsu (approved):

PR #2209, opened 2022-08-05
[PROF-5860] Allow new CPU Profiling 2.0 **alpha** profiler to be enabled

ivoanjo:
# The new Ruby profiler, aka "CPU Profiling 2.0", is considered to be alpha state. We do not recommend turning it on.

But! We actually can turn it on now -- by using `DD_PROFILING_FORCE_ENABLE_NEW=true`.

The rest of the pieces have been put into place in previous PRs.

**What does this PR do?**:

Add a setting that allows choosing between the "old" profiler codepath, and the new "CPU Profiling 2.0" codepath.

**Motivation**:

Making it possible to test the new profiler.

**Additional Notes**

This PR sits atop #2208 because without the component in #2208 the profiler would not be able to be turned on. It is otherwise independent from that change.

**How to test the change?**:

Here's a simple run of the new profiler:

```
$ DD_TRACE_DEBUG=true DD_PROFILING_ENABLED=true DD_SERVICE=ivoanjo-testing DD_ENV=staging DD_PROFILING_FORCE_ENABLE_NEW=true bundle exec ddtracerb exec ruby -e sleep
I, [2022-08-05T14:40:38.715368 #24783]  INFO -- ddtrace: [ddtrace] (dd-trace-rb/lib/datadog/profiling/tasks/setup.rb:47:in `check_if_cpu_time_profiling_is_supported') CPU time profiling skipped because native CPU time is not supported: Feature requires Linux; macOS is not supported. Profiles containing 'Wall time' data will still be reported.
D, [2022-08-05T14:40:38.727312 #24783] DEBUG -- ddtrace: [ddtrace] (dd-trace-rb/lib/datadog/core/workers/async.rb:133:in `start_worker') Starting thread for: #<Datadog::Core::Telemetry::Heartbeat:0x00007f825b982080>
D, [2022-08-05T14:40:38.727510 #24783] DEBUG -- ddtrace: [ddtrace] (dd-trace-rb/lib/datadog/core/configuration/components.rb:371:in `startup!') Profiling started
D, [2022-08-05T14:40:38.727647 #24783] DEBUG -- ddtrace: [ddtrace] (dd-trace-rb/lib/datadog/profiling/collectors/cpu_and_wall_time_worker.rb:33:in `block in start') Starting thread for: #<Datadog::Profiling::Collectors::CpuAndWallTimeWorker:0x00007f827b91b950>
[...]

PR #2181, opened 2022-07-25
[PROF-5858] Add synchronization mechanism to StackRecorder

ivoanjo:
**What does this PR do?**:

Adds a concurrency-safe synchronization mechanism to `StackRecorder`, because this component will be used from threads that can be accessing it in parallel.

**Motivation**:

The `StackRecorder` component will be accessed from two threads that can be operating in parallel:
* The sampler thread (a thread that is recording stack samples)
* The serializer thread (a thread that wants to report profiling data)

Thus, we need a concurrency mechanism to ensure that such access is safe.

**Additional Notes**:

Below I've included an extensive description of how the concurrency mechanism works.

I suggest reviewing this PR commit-by-commit.

**How to test the change?**:

This PR already includes test coverage. This mechanism will also be indirectly tested by our other integration tests and test applications, since it sits at the core of the profiler.

---

## Concurrency-safe mechanism design notes

The state of the StackRecorder is managed using a set of locks to avoid concurrency issues.

This is needed because the state is expected to be accessed, in parallel, by two different threads.

1. The thread that is taking a stack sample and that called `record_sample`, let's call it the **sampler thread**.
In the current implementation of the profiler, there can only exist one **sampler thread** at a time; if this
constraint changes, we should revise the design of the StackRecorder.

2. The thread that serializes and reports profiles, let's call it the **serializer thread**. We enforce that there
cannot be more than one thread attempting to serialize profiles at a time.

If both the sampler and serializer threads are trying to access the same `ddprof_ffi_Profile` in parallel, we will
have a concurrency issue. Thus, the StackRecorder has an added mechanism to avoid this.

As an additional constraint, the **sampler thread** has absolute priority and must never block while
recording a sample.

### The solution: Keep two profiles at the same time

To solve for the constraints above, the StackRecorder keeps two `ddprof_ffi_Profile` profile instances inside itself.
They are called the `slot_one_profile` and `slot_two_profile`.

Each profile is paired with its own mutex. `slot_one_profile` is protected by `slot_one_mutex` and `slot_two_profile`
is protected by `slot_two_mutex`.

We additionally introduce the concept of **active** and **inactive** profile slots. At any point, the sampler thread
can probe the mutexes to discover which of the profiles corresponds to the active slot, and then records samples in it.
When the serializer thread is ready to serialize data, it flips the active and inactive slots; it reports the data
on the previously-active profile slot, and the sampler thread can continue to record in the previously-inactive
profile slot.

[...]

PR #2333, opened 2022-10-26
Fix incorrect start time for profiles

ivoanjo:
**What does this PR do?**:

(Important note: This fix only affects the new CPU Profiling 2.0 profiler which is still in **alpha**, see #2209)

In #2181 we added a concurrency-safe mechanism for `StackRecorder`. Key to how it works, is that we keep two `ddog_Profile` instances at every point in time, and we alternate between using them.

But this alternation introduced a bug with the `start_time` of profiles. Previously, the `start_time` got set whenever a profile got created or reset, and we used the current time for it.

That caused the following effect:

```
t=0
  created slot_one_profile (start_time=0)
  created slow_two_profile (start_time=0)
  make_active slot_one_profile

t=60
  serialized slot_one_profile (start_time=0, finish_time=60)
  reset slot_one_profile (start_time=60)
  make_active slot_two_profile

t=120
  serialized slot_two_profile (start_time=0, finish_time=120)
  reset slot_two_profile (start_time=120)
  make_active slot_one_profile

t=180
  serialized slot_one_profile (start_time=60, finish_time=180)
  reset slow_one_profile (start_time=180)
  make_active slot_two_profile
```

That is, other than the first profile (which is why we previously missed this bug), every profile got double the duration as intended, because we reset it after serialization, but that profile would not be used for the next period.

To fix this issue, we additionally change the "make_active" step above (actually implemented in `serializer_flip_active_and_inactive_slots`) to set the correct `start_time` on the profile that becomes active.

Thus, we get the correct behavior:

```
t=0
  created slot_one_profile (start_time=0)
  created slow_two_profile (start_time=0) # Ignored, will be changed later
  make_active slot_one_profile (start_time=0)

t=60
  serialized slot_one_profile (start_time=0, finish_time=60)
  reset slot_one_profile (start_time=60) # Ignored, will be changed later
  make_active slot_two_profile (start_time=60) # Correct start_time

t=120
  serialized slot_two_profile (start_time=60, finish_time=120)
  reset slot_two_profile (start_time=120) # Ignored, will be changed later
  make_active slot_one_profile (start_time=120) # Correct start_time

t=180
  serialized slot_one_profile (start_time=120, finish_time=180)
  reset slow_one_profile (start_time=180)
  make_active slot_two_profile (start_time=180)
```

**Motivation**:

Having profiles with the wrong duration breaks profile aggregation.

**Additional notes**:

As I mentioned above, this only affects the new CPU Profiling 2.0 profiler codepath, so I don't expect any customers to have ever ran into this issue.

**How to test the change?**:

Change includes coverage. Furthermore, running the profiler with `DD_TRACE_DEBUG` prints the start/finish timestamps after  serialization, which can be used to confirm they are correct.

Finally, comparing profiles before/after in the profiler UX will also show the difference.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/stack_recorder.c
+++ b/ext/ddtrace_profiling_native_extension/stack_recorder.c
@@ -168,11 +168,12 @@ static VALUE ruby_time_from(ddog_Timespec ddprof_time);
 static void *call_serialize_without_gvl(void *call_args);
 static struct active_slot_pair sampler_lock_active_profile();
 static void sampler_unlock_active_profile(struct active_slot_pair active_slot);
-static ddog_Profile *serializer_flip_active_and_inactive_slots(struct stack_recorder_state *state);
+static ddog_Profile *serializer_flip_active_and_inactive_slots(struct stack_recorder_state *state, ddog_Timespec start_timestamp_for_next_profile);
 static VALUE _native_active_slot(DDTRACE_UNUSED VALUE _self, VALUE recorder_instance);
 static VALUE _native_is_slot_one_mutex_locked(DDTRACE_UNUSED VALUE _self, VALUE recorder_instance);
 static VALUE _native_is_slot_two_mutex_locked(DDTRACE_UNUSED VALUE _self, VALUE recorder_instance);
 static VALUE test_slot_mutex_state(VALUE recorder_instance, int slot);
+static ddog_Timespec time_now();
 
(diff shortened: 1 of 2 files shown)


marcotc (approved):

ivoanjo:
Thanks Marco for the review as usual! :)

I've force pushed a rebase on top of master so I could get a clean CI again. Going to merge now!

Issue #2334, opened 2022-10-26
Cause: undefined method `map' for #<Datadog::Tracing::TraceSegment:0x00007fcec02bf728>


DougEdey:
**Current behaviour**

When traces are too large, we're logging errors but missing the traces without any indiciation

```
E, [2022-10-26T12:41:01.889163 #1] ERROR -- ddtrace: [ddtrace] (<snip>/vendor/ruby/2.7.0/gems/ddtrace-1.5.1/lib/datadog/tracing/workers.rb:60:in `rescue in callback_traces') Error during traces flush: dropped 365 items. Cause: undefined method `map' for #<Datadog::Tracing::TraceSegment:0x00007fcec02bf728>
Did you mean?  tap Location: /<snip>/vendor/ruby/2.7.0/gems/ddtrace-1.5.1/lib/ddtrace/transport/traces.rb:83:in `block in encode_one'
```

**Expected behaviour**

Traces should drop correctly with logs indicating what the trace was.

**Steps to reproduce**

I'm not exactly sure how to reproduce this in isolation, I've created a test in a new PR that proposes a potential fix to this issue

https://github.com/DataDog/dd-trace-rb/pull/2335

**Environment**

* **ddtrace version: 1.5.1**
* **Configuration block (`Datadog.configure ...`):**
```
Datadog.configure do |c|
    # Global settings
    c.service = EnvironmentHelper.read_env("DD_SERVICE", default: "opslevel")

    # Tracing settings
    c.tracing.enabled = OpslevelCom::Application.config.dd_trace_enabled

    # Instrumentation
    c.tracing.instrument(:sidekiq, service_name: "opslevel-sidekiq", tag_args: true)
    c.tracing.instrument(:active_record, service_name: "opslevel-db")
    c.tracing.instrument(:elasticsearch, service_name: "opslevel-elasticsearch")
  end

  Datadog::Tracing.before_flush(
    Datadog::Tracing::Pipeline::SpanFilter.new { |span| span.resource == "Api::HealthController#ping" }
  )
```
* **Ruby version: 2.7.6**
* **Operating system: Alpine**
* **Relevant library versions:**

marcotc:
Fixed in #2335, thank you again @DougEdey!

PR #2335, opened 2022-10-26
Ensure that Datadog::Tracing::TraceSegment can be reported correctly when they are dropped

DougEdey:
**What does this PR do?**

This proposes a fix for https://github.com/DataDog/dd-trace-rb/issues/2334

TraceSegment doesn't respond to `map` but the spans inside it do, so it should provide usable debug information going forward.

I'm not confident that the test is the best it can be, I'm just focused on the issue I've encountered, I had to change the logger statement to actually render the log line to ensure it works appropriately.

**Motivation**

I found the bug

**Additional Notes**

**How to test the change?**

### DIFF
--- a/lib/ddtrace/transport/traces.rb
+++ b/lib/ddtrace/transport/traces.rb
@@ -80,14 +80,7 @@ module Datadog
 
           if encoded.size > max_size
             # This single trace is too large, we can't flush it
-            span_output = if trace.respond_to?(:map)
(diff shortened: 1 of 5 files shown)


marcotc (approved):
Thank you so much, @DougEdey!

I've made a few changes, but the scenario your reported is indeed problematic. 🙇

DougEdey:
Thanks @marcotc , I'm glad to see this fixed

PR #2336, opened 2022-10-26
[PROF-6260] Enable tracking of time spent in GC

ivoanjo:
**What does this PR do?**:

(Important note: This feature is only available on the new CPU Profiling 2.0 profiler which is still in **alpha**, see https://github.com/DataDog/dd-trace-rb/pull/2209)

This PR is the last piece of the puzzle started in https://github.com/DataDog/dd-trace-rb/pull/2304 and https://github.com/DataDog/dd-trace-rb/pull/2323.
With this change, time (both cpu-time and wall-time) spent by threads doing garbage collection is now accounted for, and shows up in the flamegraph.

This works by creating a new Ruby VM TracePoint to handle the `RUBY_INTERNAL_EVENT_GC_ENTER` and `RUBY_INTERNAL_EVENT_GC_EXIT` events. (These events are only available at the C-level;
Ruby-level Tracepoints cannot use them).

Then, whenever Ruby calls the TracePoint, we call the previously-added `cpu_and_wall_time_collector_on_gc_start` and `cpu_and_wall_time_collector_on_gc_finish` to track the time spent in GC, and then insert it as a sample in the profiling output (using `cpu_and_wall_time_collector_sample_after_gc`).

**Motivation**:

Without this work, time spent doing garbage collection is invisible and blamed on methods directly. By making it visible, we enable customers to make better informed decisions on what needs to be optimized (or fixed!).

**Additional Notes**:

As I mentioned above, this only affects the new CPU Profiling 2.0 profiler codepath.

During development, I initially attempted to compare the time spent in GC gathered via the TracePoint to the one that Ruby exposes via [`GC::Profiler#total_time`](https://rubyapi.org/3.1/o/gc/profiler). The results were off and when I looked into why, I discovered by looking at Ruby's `gc_start` function in `gc.c` that the time tracking for `GC::Profiler` only covers a subset of the GC process, not the entire time.

Furthermore, from what I was able to observe, it accounts for cpu-time process-wide on Linux, and in some cases (macos), only cpu-time in usermode, so that's another reason for discrepancies.

Thus, when comparing values, expect the `GC::Profiler` to be way less than the values exposed by the profiler.

I belive the tracepoint-based approach and values recorded by the profiler to be more accurate since Ruby calls the `ENTER` event quite early in the GC process, and `EXIT` quite late. See also <https://github.com/ivoanjo/gvl-tracing/pull/6> for a similar approach proposed by a Ruby core developer.

**Note**: On macOS, there's a situation where time goes backwards even when there's no obvious weird system-wide change going on. I'm still investigating it, but I don't think that needs to block this PR -- the time difference is so small that it'll not change the output (usually ~0.1s).

**How to test the change?**:

Beyond the included code coverage, you should see `Garbage Collection` frames representing time spent in GC in flamegraphs (when using the new CPU Profiling 2.0 codepath, see above).

Here's a tiny Ruby script that triggers a lot of allocation:

```ruby
def do_alloc(n)
  alloc(n)
end

def alloc(n)
  n.times do
    Object.new
  end
end

while true
  do_alloc(100_000)
end
```

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
@@ -62,6 +62,8 @@
 #define THREAD_ID_LIMIT_CHARS 20
 #define RAISE_ON_FAILURE true
 #define DO_NOT_RAISE_ON_FAILURE false
+#define IS_WALL_TIME true
+#define IS_NOT_WALL_TIME false
 
 // Contains state for a single CpuAndWallTime instance
 struct cpu_and_wall_time_collector_state {
@@ -136,7 +138,7 @@ static VALUE stats_as_ruby_hash(struct cpu_and_wall_time_collector_state *state)
 static void remove_context_for_dead_threads(struct cpu_and_wall_time_collector_state *state);
 static int remove_if_dead_thread(st_data_t key_thread, st_data_t value_context, st_data_t _argument);
 static VALUE _native_per_thread_context(VALUE self, VALUE collector_instance);
-static long update_time_since_previous_sample(long *time_at_previous_sample_ns, long current_time_ns, long gc_start_time_ns);
+static long update_time_since_previous_sample(long *time_at_previous_sample_ns, long current_time_ns, long gc_start_time_ns, bool is_wall_time);
 static long cpu_time_now_ns(struct per_thread_context *thread_context);
 static long wall_time_now_ns(bool raise_on_failure);
 static long thread_id_for(VALUE thread);
@@ -302,12 +304,14 @@ VALUE cpu_and_wall_time_collector_sample(VALUE self_instance) {
     long cpu_time_elapsed_ns = update_time_since_previous_sample(
       &thread_context->cpu_time_at_previous_sample_ns,
       current_cpu_time_ns,
-      thread_context->gc_tracking.cpu_time_at_start_ns
+      thread_context->gc_tracking.cpu_time_at_start_ns,
+      IS_NOT_WALL_TIME
     );
     long wall_time_elapsed_ns = update_time_since_previous_sample(
       &thread_context->wall_time_at_previous_sample_ns,
       current_wall_time_ns,
-      thread_context->gc_tracking.wall_time_at_start_ns
+      thread_context->gc_tracking.wall_time_at_start_ns,
+      IS_WALL_TIME
     );
 
     int64_t metric_values[ENABLED_VALUE_TYPES_COUNT] = {0};
@@ -450,8 +454,11 @@ VALUE cpu_and_wall_time_collector_sample_after_gc(VALUE self_instance) {
     long gc_wall_time_elapsed_ns =
       thread_context->gc_tracking.wall_time_at_finish_ns - thread_context->gc_tracking.wall_time_at_start_ns;
 
+    // We don't expect non-wall time to go backwards, so let's flag this as a bug
     if (gc_cpu_time_elapsed_ns < 0) rb_raise(rb_eRuntimeError, "BUG: Unexpected negative gc_cpu_time_elapsed_ns between samples");
-    if (gc_wall_time_elapsed_ns < 0) rb_raise(rb_eRuntimeError, "BUG: Unexpected negative gc_wall_time_elapsed_ns between samples");
+    // Wall-time can actually go backwards (e.g. when the system clock gets set) so we can't assume time going backwards
+    // was a bug
+    if (gc_wall_time_elapsed_ns < 0) gc_wall_time_elapsed_ns = 0;
 
     if (thread_context->gc_tracking.wall_time_at_start_ns == 0 && thread_context->gc_tracking.wall_time_at_finish_ns != 0) {
       // Avoid using wall-clock if we got 0 for a start (meaning there was an error) but not 0 for finish so we don't
@@ -657,7 +664,7 @@ static VALUE _native_per_thread_context(DDTRACE_UNUSED VALUE _self, VALUE collec
   return per_thread_context_st_table_as_ruby_hash(state);
 }
 
-static long update_time_since_previous_sample(long *time_at_previous_sample_ns, long current_time_ns, long gc_start_time_ns) {
+static long update_time_since_previous_sample(long *time_at_previous_sample_ns, long current_time_ns, long gc_start_time_ns, bool is_wall_time) {
   // If we didn't have a time for the previous sample, we use the current one
   if (*time_at_previous_sample_ns == INVALID_TIME) *time_at_previous_sample_ns = current_time_ns;
 
@@ -680,7 +687,16 @@ static long update_time_since_previous_sample(long *time_at_previous_sample_ns, 
     *time_at_previous_sample_ns = current_time_ns;
   }
 
-  if (elapsed_time_ns < 0) rb_raise(rb_eRuntimeError, "BUG: Unexpected negative elapsed_time_ns between samples");
+  if (elapsed_time_ns < 0) {
+    if (is_wall_time) {
+      // Wall-time can actually go backwards (e.g. when the system clock gets set) so we can't assume time going backwards
+      // was a bug
+      elapsed_time_ns = 0;
+    } else {
+      // We don't expect non-wall time to go backwards, so let's flag this as a bug
+      rb_raise(rb_eRuntimeError, "BUG: Unexpected negative elapsed_time_ns between samples");
+    }
+  }
 
   return elapsed_time_ns;
 }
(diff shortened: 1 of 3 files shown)


marcotc (approved):
🔥 !!!

ivoanjo:
I just force-pushed a rebase on top of master to get the latest CI fixes. Merging away!

ivoanjo:
> Note: On macOS, there's a situation where time goes backwards even when there's no obvious weird system-wide change going on. I'm still investigating it, but I don't think that needs to block this PR -- the time difference is so small that it'll not change the output (usually ~0.1s).

I was able to reproduce this even outside of Ruby. It takes minutes/hours, but with this example it eventually shows up for me (macOS 11.7.1 / i7-1068NG7):

```c
#include <stdio.h>
#include <time.h>
#include <stdbool.h>
#include <unistd.h>
#include <pthread.h>
#include <limits.h>

static _Thread_local long previous_read = INT_MIN;

static long wall_time_now_ns() {
  struct timespec current_monotonic;

  if (clock_gettime(CLOCK_MONOTONIC, &current_monotonic) != 0) {
    fprintf(stderr, "Error reading clock_gettime\n");
    return 0;
  }

  return current_monotonic.tv_nsec + (current_monotonic.tv_sec * 1000 * 1000 * 1000);
}

static long latest_global = 0;

void *run_in_thread(void *arg) {
  int i = (int) arg;
  fprintf(stderr, "Thread %d started\n", i);

  long loops = 0;

  while (true) {
    loops++;
    if (loops % 10000000 == 0) {
      fprintf(stderr, "i=%d loops=%ld\n", i, loops);
    }

    long read = wall_time_now_ns();

    if (read < previous_read) {
      fprintf(stderr, "Time has gone backwards! read=%ld previous_read=%ld\n", read, previous_read);
    }

    previous_read = read;
  }
}

#define NUM_THREADS 2

int main(int argc, char **argv) {
  long latest_global = wall_time_now_ns();

  pthread_t thread_id[NUM_THREADS];

  for (int i = 0; i < NUM_THREADS; i++) {
    pthread_create(&thread_id[i], NULL, run_in_thread, (void *) (long) i);
  }

  sleep(720000);

  return 0;
}
```

(I am aware the above example can be simplified, but I went through a bunch of variations to experiment different ways of teasing it out, and this is the weird final version which also showed the issue, so I leave making it better to the reader)

eregon:
I've tried that program on macOS 26.4.1 (M4 Max) and I have seen it (after ~3h30, might have happened multiple times):
```
...
i=0 loops=91730000000
i=1 loops=91740000000
i=0 loops=91740000000
i=1 loops=91750000000
i=0 loops=91750000000
Time has gone backwards! read=2846785780942000 previous_read=2846785780952000
i=1 loops=91760000000
i=0 loops=91760000000
i=1 loops=91770000000
i=0 loops=91770000000
i=1 loops=91780000000
...
```

The `return current_monotonic.tv_nsec + (current_monotonic.tv_sec * 1000 * 1000 * 1000);` is a bit fishy but should be fine on 64-bit according to Claude.
Worth noting the program doesn't stop when it detects `Time has gone backwards!`.
I made a slightly cleaner version:
```c
#include <stdio.h>
#include <time.h>
#include <stdbool.h>
#include <stdlib.h>
#include <unistd.h>
#include <pthread.h>
#include <limits.h>

static _Thread_local long previous_read = INT_MIN;

static long wall_time_now_ns() {
  struct timespec current_monotonic;

  if (clock_gettime(CLOCK_MONOTONIC, &current_monotonic) != 0) {
    fprintf(stderr, "Error reading clock_gettime\n");
    exit(2);
  }

  return current_monotonic.tv_nsec + (current_monotonic.tv_sec * 1000000000L);
}

static long latest_global = 0;

void *run_in_thread(void *arg) {
  int i = (int) (long) arg;
  fprintf(stderr, "Thread %d started\n", i);

  long loops = 0;

  while (true) {
    loops++;
    if (loops % 10000000 == 0) {
      fprintf(stderr, "i=%d loops=%ld\n", i, loops);
    }

    long read = wall_time_now_ns();

    if (read < previous_read) {
      fprintf(stderr, "Time has gone backwards! read=%ld previous_read=%ld\n", read, previous_read);
      exit(1);
    }

    previous_read = read;
  }
}

#define NUM_THREADS 2

int main(int argc, char **argv) {
  long latest_global = wall_time_now_ns();

  pthread_t thread_id[NUM_THREADS];

  for (int i = 0; i < NUM_THREADS; i++) {
    pthread_create(&thread_id[i], NULL, run_in_thread, (void *) (long) i);
  }

  sleep(720000);

  return 0;
}
```

eregon:
With the new version:
```
i=0 loops=83830000000
i=1 loops=83870000000
i=0 loops=83840000000
i=0 loops=83850000000
i=1 loops=83880000000
i=0 loops=83860000000
i=1 loops=83890000000
i=0 loops=83870000000
i=1 loops=83900000000
i=0 loops=83880000000
i=1 loops=83910000000
i=0 loops=83890000000
i=1 loops=83920000000
i=0 loops=83900000000
i=1 loops=83930000000
i=0 loops=83910000000
i=1 loops=83940000000
i=0 loops=83920000000
i=1 loops=83950000000
i=0 loops=83930000000
i=1 loops=83960000000
i=0 loops=83940000000
i=1 loops=83970000000
i=0 loops=83950000000
i=1 loops=83980000000
i=0 loops=83960000000
i=1 loops=83990000000
i=0 loops=83970000000
i=1 loops=84000000000
i=0 loops=83980000000
i=1 loops=84010000000
i=0 loops=83990000000
i=1 loops=84020000000
i=0 loops=84000000000
i=1 loops=84030000000
i=0 loops=84010000000
i=1 loops=84040000000
i=0 loops=84020000000
i=1 loops=84050000000
i=0 loops=84030000000
i=1 loops=84060000000
Time has gone backwards! read=2852849262857000 previous_read=2852849262987000
zsh: exit 1     ./a.out
./a.out  10142.01s user 43.61s system 199% cpu 1:25:03.03 total
```

Will retry with CLOCK_MONOTONIC_RAW (https://github.com/DataDog/dd-trace-rb/pull/5994)

eregon:
0 failures with CLOCK_MONOTONIC_RAW:
It ran for
```
38737.59s user 170.84s system 60% cpu 17:50:00.46 total
```
Which is 10h45min of user time.

IOW, on macOS CLOCK_MONOTONIC isn't monotonic, but CLOCK_MONOTONIC_RAW is.
Since we use CLOCK_MONOTONIC_RAW since https://github.com/DataDog/dd-trace-rb/pull/5994 we could error if we see `elapsed_time_ns < 0` for wall time too, WDYT @ivoanjo ?

eregon:
BTW, not the first time macOS is violating POSIX:

> clock_gettime(CLOCK_MONOTONIC) can actually go backwards on macOS, I have seen it with time_backwards.c, expected? No   problem with CLOCK_MONOTONIC_RAW. Isn't that a violation of POSIX? What about clock_gettime(CLOCK_MONOTONIC) on Linux?

### `clock_gettime(CLOCK_MONOTONIC)` can go backwards on macOS

### POSIX spec

`CLOCK_MONOTONIC` represents "the monotonic clock for the system" whose value "cannot be set via `clock_settime()`." The name "monotonic" means non-decreasing. POSIX does say it "may be affected by incremental adjustments performed by `adjtime()` and NTP" — but those adjustments are frequency slewing (speeding up or slowing down), not backward jumps. Going backwards violates the monotonicity guarantee.

  ### macOS

  Apple's `CLOCK_MONOTONIC` implementation applies NTP frequency adjustments (slewing) on top of the hardware counter. The implementation can produce small backward jumps, likely due to how these adjustments interact with cross-core TSC reads or rounding. This is a bug in Apple's implementation.

  `CLOCK_MONOTONIC_RAW` reads the raw hardware counter with no NTP adjustments, which is why it doesn't exhibit the problem.

  ### Linux

  The kernel explicitly guarantees `CLOCK_MONOTONIC` never goes backwards. Even with NTP slewing applied, the kernel's timekeeping path has safeguards to ensure monotonicity. `CLOCK_MONOTONIC_RAW` also exists on Linux (no NTP adjustments) but `CLOCK_MONOTONIC` is already safe.

  ### Summary

  | | macOS | Linux |
  |---|---|---|
  | `CLOCK_MONOTONIC` | Can go backwards (bug) | Guaranteed monotonic |
  | `CLOCK_MONOTONIC_RAW` | Guaranteed monotonic | Guaranteed monotonic |

ivoanjo:
Agreed -- seems like in current master it's reasonable to say "we don't expect time going backwards, and if it does, we want to know about it".

Since that touches some of the same code from https://github.com/DataDog/dd-trace-rb/pull/6006 I've made a note to circle back to this next week once I fix the flaky CI issues as otherwise both PRs would conflict. LMK if you disagree!

PR #2342, opened 2022-10-31
[PROF-4905] Implement support for code hotspots in CpuAndWallTime collector

ivoanjo:
**What does this PR do?**:

This PR reimplements support for the code hotspots feature (connecting traces to profiles) in the `CpuAndWallTime` collector.

Although the implementation is different, the approach is conceptually the same as provided by `Datadog::Profiling::TraceIdentifiers::Ddtrace` for the old profiler codepaths.

Key contraints in this implementation are:
* No objects are created to get this information
* No Ruby code is called to get this information

These constraints are key to:
* Maintain performance on the `CpuAndWallTime` collector (we never get out of C code).
* Keep precise control of the Global VM Lock. Since we never get out of C, we take the sample in one pass. Calling into Ruby code may mean being "interrupted" in the middle of sampling, which would break the internal consistency of the profiler. (Assuming that we never lose   the GVL makes the profiler simpler and more performant).
* Correctness of customer code. Sampling happens from customer threads directly (using `rb_postponed_job_register_one`) so we don't want to call Ruby methods and potentially trigger side effects during a sample.
* Memory profiling. We don't yet implement allocation or heap profiling, but it's coming soon and we don't want to allocate memory to track memory allocations.

Thus, I chose to implement this feature by accessing the tracer context directly from the sampling process. Because all the information needed is available as instance variables of the objects, they can be read safely and within the needed constraints.

This implementation also includes extensive checks to make sure that if tracing evolves or is not available, profiling safely skips getting the information.

Furthermore, it includes an integration test with a real tracer, so that we'll immediately know when an incompatibility is introduced, so that the profiler can be updated.

**Motivation**:

Code hotspots is a key feature of our APM, allowing customers to see the profile for an individual trace.

**Additional Notes**:

This first implementation does not include support for endpoint aggregation; this will come in a separate PR.

**How to test the change?**:

This change includes code coverage. It can also be tested by enabling profiling and tracing on a Ruby app, and checking that the "Code Hotspots" tab on a generated trace is showing the
profile for that trace.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
@@ -457,7 +457,9 @@ VALUE cpu_and_wall_time_collector_sample_after_gc(VALUE self_instance) {
     // We don't expect non-wall time to go backwards, so let's flag this as a bug
     if (gc_cpu_time_elapsed_ns < 0) rb_raise(rb_eRuntimeError, "BUG: Unexpected negative gc_cpu_time_elapsed_ns between samples");
     // Wall-time can actually go backwards (e.g. when the system clock gets set) so we can't assume time going backwards
-    // was a bug
+    // was a bug.
+    // @ivoanjo: I've also observed time going backwards spuriously on macOS, see discussion on
+    // https://github.com/DataDog/dd-trace-rb/pull/2336.
     if (gc_wall_time_elapsed_ns < 0) gc_wall_time_elapsed_ns = 0;
 
     if (thread_context->gc_tracking.wall_time_at_start_ns == 0 && thread_context->gc_tracking.wall_time_at_finish_ns != 0) {
@@ -699,7 +701,9 @@ static long update_time_since_previous_sample(long *time_at_previous_sample_ns, 
   if (elapsed_time_ns < 0) {
     if (is_wall_time) {
       // Wall-time can actually go backwards (e.g. when the system clock gets set) so we can't assume time going backwards
-      // was a bug
+      // was a bug.
+      // @ivoanjo: I've also observed time going backwards spuriously on macOS, see discussion on
+      // https://github.com/DataDog/dd-trace-rb/pull/2336.
       elapsed_time_ns = 0;
     } else {
       // We don't expect non-wall time to go backwards, so let's flag this as a bug
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
@@ -64,6 +64,12 @@
 #define DO_NOT_RAISE_ON_FAILURE false
 #define IS_WALL_TIME true
 #define IS_NOT_WALL_TIME false
+#define MISSING_TRACER_CONTEXT_KEY 0
+
+static ID at_active_trace_id; // id of :@active_trace in Ruby
+static ID at_root_span_id;    // id of :@root_span in Ruby
+static ID at_active_span_id;  // id of :@active_span in Ruby
+static ID at_id_id;           // id of :@id in Ruby
 
 // Contains state for a single CpuAndWallTime instance
 struct cpu_and_wall_time_collector_state {
@@ -76,6 +82,10 @@ struct cpu_and_wall_time_collector_state {
   st_table *hash_map_per_thread_context;
   // Datadog::Profiling::StackRecorder instance
   VALUE recorder_instance;
+  // If the tracer is available and enabled, this will be the fiber-local symbol for accessing its running context,
+  // to enable code hotspots and endpoint aggregation.
+  // When not available, this is set to MISSING_TRACER_CONTEXT_KEY.
+  ID tracer_context_key;
   // Track how many regular samples we've taken. Does not include garbage collection samples.
   // Currently **outside** of stats struct because we also use it to decide when to clean the contexts, and thus this
   // is not (just) a stat.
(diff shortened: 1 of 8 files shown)


ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
```diff
@@ -508,26 +546,44 @@ static void trigger_sample_for_thread(
...
+  trace_identifiers_for(state, thread, &trace_identifiers_result);
+
+  if (trace_identifiers_result.valid) {
+    labels[label_pos++] = (ddog_Label) {.key = DDOG_CHARSLICE_C("local root span id"), .str = trace_identifiers_result.local_root_span_id};
+    labels[label_pos++] = (ddog_Label) {.key = DDOG_CHARSLICE_C("span id"), .str = trace_identifiers_result.span_id};
+  }
+
+  if (label_pos > max_label_count) {
+    rb_raise(rb_eRuntimeError, "BUG: Unexpected label_pos (%d) > max_label_count (%d)", label_pos, max_label_count);
+  }
```
marcotc, line 579: Can this ever happen?

marcotc: It looks like someone has to modify the function for this to happen.
Does it make sense to have one of those "conditional assertion that does not exist in the production release"?

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
```diff
@@ -759,3 +819,38 @@ static VALUE _native_stats(DDTRACE_UNUSED VALUE _self, VALUE collector_instance)
 
   return stats_as_ruby_hash(state);
 }
+
+static void trace_identifiers_for(struct cpu_and_wall_time_collector_state *state, VALUE thread, struct trace_identifiers *trace_identifiers_result) {
```
marcotc, line 823: Do we have the GVL when calling this function?

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
```diff
@@ -759,3 +819,38 @@ static VALUE _native_stats(DDTRACE_UNUSED VALUE _self, VALUE collector_instance)
...
+  if (root_span == Qnil || active_span == Qnil) return;
+
+  VALUE numeric_local_root_span_id = rb_ivar_get(root_span, at_id_id /* @id */);
+  VALUE numeric_span_id = rb_ivar_get(active_span, at_id_id /* @id */);
+  if (numeric_local_root_span_id == Qnil || numeric_span_id == Qnil) return;
+
+  unsigned long long local_root_span_id = NUM2ULL(numeric_local_root_span_id);
+  unsigned long long span_id = NUM2ULL(numeric_span_id);
+
+  snprintf(trace_identifiers_result->local_root_span_id_buffer, MAXIMUM_LENGTH_64_BIT_IDENTIFIER, "%llu", local_root_span_id);
```
marcotc, line 843:
```c
"%llu"
```
TIL
[206273_LLU](https://www.abbreviations.com/LLU)

marcotc (commented):
The implementation of `trace_identifiers_for` looks pretty short and easy to follow!

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
ivoanjo: > Can this ever happen?
> It looks like someone has to modify the function for this to happen.

Yes, exactly! I've pushed bb95c1c737f237a94d2c76e1856d94a142e358ca to better explain why that's there so you and others don't have to "reverse engineer" my intent ;)

> Does it make sense to have one of those "conditional assertion that does not exist in the production release"?

In this particular case, I don't think so, although I have been thinking about some mechanism to do that. But before I explain why, I'm curious why you ask:

* Is it because you're concerned about performance? If so, since this is a comparison between a constant and a local that recently got used, I'd expect this to be really cheap. In fact, I wonder if the compiler can be smart enough to optimize it out, by looking at the flow of `label_pos` in the function and proving it never can get bigger than `max_label_count`.

* Is it because you're concerned about correctness? The profiler is correctly built to handle exceptions. When they happen, we log them out and then stop profiling. I think for these "this should never happen" situations, stopping profiling entirely seems like a reasonable thing to do to avoid any impact on the user application.

---

That said, here's why "In this particular case, I don't think so": Because I fear that contributors and customers may not realize such a mode exists.

In fact, there's a known fork of dd-trace-rb that changes exactly this function -- 
https://github.com/zendesk/dd-trace-rb/commit/192e5c88437987276b395a3ecb8fcdda32cfe89d -- and when that fork gets updated with this change, I wanted to make it easier to update this code without issues.

In its current approach, any incorrect change to this function would appear quite soon in a quite dramatic way (the profiler stops), which I think would allow us/customers to catch any bugs sooner.

---

Finally, I have been thinking about having some mode that we could use in CI/testing to enable extra checks, **especially** extra checks that may be expensive/break customer setups and we wouldn't want to have on.

A good example is compiler warnings -- we don't know which version of the C compiler our customers are using, so using `-Werror` to turn compiler warnings into errors on the release version of the gem is a bad idea -- suddenly a newer version of the compiler gets released and the profiler stops working because there's a new warning.

BUUUT we could enable this mode for just for CI/development, and it would allow us to catch new warnings as we introduce them.

So yeah, I haven't yet looked into such a setup, but suggestions welcome. (But in this case, even if such a setup was already in place, I would still not move this check to be done only on CI/dev mode, for the reasons I explained above).

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
ivoanjo: Yes! Added additional documentation in c7f6401bdb066fa37c188509cf000c7ebbb90be2, but currently the whole sampling approach is done while holding the GVL.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
ivoanjo: `unsigned long long long long long long long long long long long long`... ;)

marcotc (approved):

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
marcotc: Thanks for the detailed response.

I see this check as a compiler shortcoming: the compiler should catch this. So my thought was "let's augment the compilation step then".
It seems like leaving the check as is is the best option today, so I'm happy with that.

PR #2327, opened 2022-10-25
Avoid using problematic rspec-mocks 3.11.2

ivoanjo:
**What does this PR do?**

Adds a restriction on our `Gemfile` (only used for testing) to not use `rspec-mocks` 3.11.2.

**Motivation**

This version of rspec-mocks breaks CI for Ruby 2.7, 3.0 and 3.1 on the following tests:

* ./spec/datadog/core/configuration_spec.rb:377 (Ruby 2.7)
* ./spec/datadog/core_spec.rb:35 (Ruby 2.7)
* ./spec/ddtrace/transport/http/builder_spec.rb:250 (Ruby 3.0, 3.1)

References:

* <https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/7560/workflows/24205733-d9fb-4781-962c-87499bad162d/jobs/280394>
* <https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/7560/workflows/24205733-d9fb-4781-962c-87499bad162d/jobs/280387>
* <https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/7560/workflows/24205733-d9fb-4781-962c-87499bad162d/jobs/280396>

**Additional Notes**

I have not yet reported this upstream, but wanted to first unblock our CI.

**How to test the change?**

Validate that CI is green.

PR #2346, opened 2022-11-02
Remove workaround for rspec-mocks issue and bump RSpec to 3.12

ivoanjo:
**What does this PR do?**:

This PR removes the workaround added in #2327 and #2337 to lock us out of using the latest version of rspec-mocks.

It then bumps the minimum version of RSpec used to the latest 3.12 series (which still works on all Ruby versions we support).

**Motivation**:

I was planning on reporting the RSpec bug we saw in #2327 and #2337 upstream, but then could no longer reproduce the issue in the latest version.

When I pushed the upgrade to 3.12 to CI, I still saw an incompatibility with Ruby 2.7, but  that seems to have already been reported upstream https://github.com/rspec/rspec-mocks/issues/1492 and because it only happened in a couple of places I just tweaked the code to avoid triggering the issue.

**Additional Notes**:

(Nothing to add)

**How to test the change?**:

Validate that CI is still green.


spec/datadog/tracing/contrib/active_support/notifications/subscription_spec.rb
```diff
@@ -11,7 +11,7 @@
     subject(:subscription) { described_class.new(span_name, options, &block) }
 
     let(:span_name) { double('span_name') }
-    let(:options) { {} }
+    let(:options) { { resource: 'dummy_resource' } }
```
TonyCTHsu, line 14: Is this change also necessary?

TonyCTHsu (approved):
Thanks for looking into this!

spec/datadog/tracing/contrib/active_support/notifications/subscription_spec.rb
ivoanjo: Yes -- see https://github.com/DataDog/dd-trace-rb/pull/2346/commits/158f165cc21966fa46b191d4c3f187f51801da4a for why :)

PR #2349, opened 2022-11-02
Add required hostname value to Datadog agent

marcotc:
The Datadog agent now requires a meaningful hostname in order to start (https://github.com/DataDog/datadog-agent/issues/14152)

The CircleCI environment is not really a fully fledged production environment, thus the agent fails to find a suitable hostname.

This PR adds a hostname explicitly to the agent.


ivoanjo (approved):
Thanks Marco. This seems like a weird oversight (e.g. I wonder if it will be reverted), but the fix/workaround is small enough that I think it makes sense to just add regardless.

ivoanjo:
Since this is a tiny "unblock-everyone" PR, I'll go ahead and merge it now!

PR #2350, opened 2022-11-03
Add profiler helper for detecting if we're on the main Ractor

ivoanjo:
**What does this PR do?**:

Add a new helper to `private_vm_access.c` that allows the caller thread to ask if it's running as part of the main Ractor.

On Rubies without Ractor support, as a simplification, we always return `true` so that callers don't need to probe for Ractor support.

**Motivation**:

This helper will be used to make sure the profiler code does not run on the wrong Ractor, since the profiler code is currently not built to be called in parallel from multiple Ractors.

**Additional Notes**:

Hey, I added the first spec that uses Ractors to our test suite!

**How to test the change?**:

Test coverage included.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/extconf.rb
+++ b/ext/ddtrace_profiling_native_extension/extconf.rb
@@ -125,6 +125,9 @@ $defs << '-DNO_RB_NATIVE_THREAD' if RUBY_VERSION < '3.2'
 # On older Rubies, we need to use a backported version of this function. See private_vm_api_access.h for details.
 $defs << '-DUSE_BACKPORTED_RB_PROFILE_FRAME_METHOD_NAME' if RUBY_VERSION < '3'
 
+# On older Rubies, there are no Ractors
+$defs << '-DNO_RACTORS' if RUBY_VERSION < '3'
+
 # On older Rubies, we need to use rb_thread_t instead of rb_execution_context_t
 $defs << '-DUSE_THREAD_INSTEAD_OF_EXECUTION_CONTEXT' if RUBY_VERSION < '2.5'
 
--- a/ext/ddtrace_profiling_native_extension/private_vm_api_access.c
+++ b/ext/ddtrace_profiling_native_extension/private_vm_api_access.c
@@ -690,3 +690,28 @@ int ruby_thread_has_gvl_p(void) {
   return 0;
 }
 #endif // NO_THREAD_HAS_GVL
+
+#ifndef NO_RACTORS
+  // This API and definition are exported as a public symbol by the VM BUT the function header is not defined in any public header, so we
+  // repeat it here to be able to use in our code.
+  bool rb_ractor_main_p_(void);
+  extern struct rb_ractor_struct *ruby_single_main_ractor;
+
+  // Taken from upstream ractor_core.h at commit d9cf0388599a3234b9f3c06ddd006cd59a58ab8b (November 2022, Ruby 3.2 trunk)
+  // to allow us to ensure that we're always operating on the main ractor (if Ruby has ractors)
+  // Modifications:
+  // * None
+  bool ddtrace_rb_ractor_main_p(void)
+  {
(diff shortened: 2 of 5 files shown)


ivoanjo:
Note: I may have found another Ruby bug (or at least something strange) that happens after the first Ractor gets created in a Ruby app. I suspect this PR will cause flaky tests because of that, so I'm moving it back to draft until I can ensure this will not be the case.

ivoanjo:
Update: I was able to track down the flake issue to a Ruby VM bug, which I've reported upstream https://bugs.ruby-lang.org/issues/19112 .

I've also deployed a workaround in https://github.com/DataDog/dd-trace-rb/pull/2354, and changed this PR to be on top of that one, to underline the implicit dependency -- the workaround needs to go in first so as to avoid introducing a flaky test.

Marking this PR as "ready for review" once more.

ext/ddtrace_profiling_native_extension/extconf.rb
```diff
@@ -125,6 +125,9 @@ def add_compiler_flag(flag)
 # On older Rubies, we need to use a backported version of this function. See private_vm_api_access.h for details.
 $defs << '-DUSE_BACKPORTED_RB_PROFILE_FRAME_METHOD_NAME' if RUBY_VERSION < '3'
 
+# On older Rubies, there are no Ractors
+$defs << '-DNO_RACTORS' if RUBY_VERSION < '3'
```
marcotc, line 129: There's no built in `define` around this? Interesting...

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
```diff
@@ -690,3 +690,28 @@ int ruby_thread_has_gvl_p(void) {
   return 0;
 }
 #endif // NO_THREAD_HAS_GVL
+
+#ifndef NO_RACTORS
+  // This API and definition are exported as a public symbol by the VM BUT the function header is not defined in any public header, so we
+  // repeat it here to be able to use in our code.
```
marcotc, line 696: Does it make sense to request/perform this change upstream?

marcotc (approved):

ext/ddtrace_profiling_native_extension/extconf.rb
ivoanjo: I did a quick search and there isn't any -- the closest I found has `HAVE_RUBY_RACTOR_H` which can maybe thought of as a way of doing that but... I decided to keep this one for now ;) (Unless you think it's a really bad idea)

I think the reason why there is no define is because you can't disable Ractors (that I know of), and thus you can always probe them by looking at the Ruby version, so that's probably why they don't have a more specific `#define`.

On our side, as you can kinda tell by this file, I've been somewhat avoiding doing something like `$defs << "-DRUBY_#{RUBY_VERSION}"` and then doing `#ifdef RUBY_3.1` in the code, in favor of being a bit more semantic -- e.g. rather than saying "this is the codepath for Ruby 3.1", say "this is the codepath for when this structure has a member called `foo`".

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
ivoanjo: I think this "public-ish" is actually on purpose. That is, the missing header is not a "oops we forgot", it's a "this isn't really a public API, but some things outside of Ruby make good use of it, and we don't mind exposing it for now but you're on your own".

For instance in https://github.com/ruby/ruby/commit/15476c695d708796f2271d99deb3cd5288d016d2 you see a similar API that was published so that `fiddle` could use it:

https://github.com/ruby/ruby/blob/c3de7a3c58bf9a138ff8720ed56c0045d2b8e01d/internal/thread.h#L48

(Out of curiosity, there was some openness during a few chats with core team members recently to perhaps promote a few more APIs that would be really useful for profilers to this "public-ish" state).

PR #2352, opened 2022-11-04
Refactor:Move distributed propagation to Contrib

marcotc:
The `Datadog::Tracing::Distributed` namespace lives in the core `Datadog::Tracing` namespace, but a lot of its logic is propagation style-specific. We have two propagation styles, based on different protocols: HTTP Headers and gRPC metadata.

The current implementation contains a lot of complex, duplicated logic between these two styles. Also, some features from the HTTP headers implementation are missing from the gRPC version because it's easy to forget to keep them in sync.

This PR moves style-specific log from the `Datadog::Tracing::Distributed` namespace to `Datadog::Tracing::Contrib`.

**This is simply a large refactor**: there are no changes to externally visible behavior.
Tests remain the same, and passing, with coverage increased in a few cases.

Affected modules that are part of the public API, `Datadog::Tracing::Distributed::Headers::Ext` and `Datadog::Tracing::Propagation::HTTP`, have a compatibility layer added to this PR to ensure there are no breaking changes.


lib/datadog/tracing/contrib/grpc/patcher.rb
```diff
@@ -19,8 +19,6 @@ def target_version
           end
 
           def patch
-            require_relative '../../propagation/grpc'
-            require_relative 'datadog_interceptor'
```
marcotc, line 23: These were just unnecessary and actually the responsibility of `intercept_with_datadog.rb`.

spec/datadog/opentracer/rack_propagator_spec.rb
```diff
@@ -27,15 +27,15 @@
     before do
       # Expect carrier to be set with Datadog trace properties
       expect(carrier).to receive(:[]=)
-        .with(Datadog::Tracing::Distributed::Headers::Ext::HTTP_HEADER_TRACE_ID, trace_id.to_s)
+        .with('x-datadog-trace-id', trace_id.to_s)
```
marcotc, line 30: Instead of fixing the namespace changes, I chose to inline the string values here: it makes tests more robust and easier to read.

ivoanjo (commented):
I... ran out of time for today so here's 1/3 of a review :)

lib/datadog/tracing/distributed/datadog.rb
```diff
@@ -0,0 +1,150 @@
...
+  module Tracing
+    module Distributed
+      # Datadog-style trace propagation.
+      class Datadog
+        def initialize(
+          trace_id: Ext::HTTP_HEADER_TRACE_ID,
+          parent_id: Ext::HTTP_HEADER_PARENT_ID,
+          sampling_priority: Ext::HTTP_HEADER_SAMPLING_PRIORITY,
+          origin: Ext::HTTP_HEADER_ORIGIN,
+          tags: Ext::HTTP_HEADER_TAGS,
```
ivoanjo, line 18: See my notes on the b3 class about calling these headers :)

lib/datadog/tracing/distributed/datadog.rb
```diff
@@ -0,0 +1,150 @@
...
+          encoded_tags = DatadogTagsCodec.encode(digest.trace_distributed_tags)
+
+          if encoded_tags.size > ::Datadog.configuration.tracing.x_datadog_tags_max_length
+            active_trace = Tracing.active_trace
+            active_trace.set_tag('_dd.propagation_error', 'inject_max_size') if active_trace
+
+            ::Datadog.logger.warn(
+              "Failed to inject x-datadog-tags: tags are too large (size:#{encoded_tags.size} " \
+                "limit:#{::Datadog.configuration.tracing.x_datadog_tags_max_length}). This limit can be configured " \
+                'through the DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH environment variable.'
```
ivoanjo, line 92: Minor: Slight rewording

```suggestion
              "Failed to inject x-datadog distributed tracing tags: too many tags for configured limit (size:#{encoded_tags.size} " \
                "limit:#{::Datadog.configuration.tracing.x_datadog_tags_max_length}). This limit can be configured " \
                'through the DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH environment variable.'
```

lib/datadog/tracing/distributed/datadog.rb
```diff
@@ -0,0 +1,150 @@
...
+          ::Datadog.logger.warn(
+            "Failed to inject x-datadog-tags: #{e.class.name} #{e.message} at #{Array(e.backtrace).first}"
+          )
+        end
+
+        # Import `x-datadog-tags` tags as trace distributed tags.
+        # Only tags that have the `_dd.p.` prefix are processed.
+        def extract_tags(data)
+          tags = data[@tags]
+          return if !tags || tags.empty?
```
ivoanjo, line 110: Minor: This seems to actually get the `fetcher` as an input

```suggestion
        def extract_tags(fetcher)
          tags = fetcher[@tags]
          return if !tags || tags.empty?
```

lib/datadog/tracing/distributed/datadog.rb
```diff
@@ -0,0 +1,150 @@
...
+        end
+
+        private
+
+        # Export trace distributed tags through the `x-datadog-tags` key.
+        #
+        # DEV: This method accesses global state (the active trace) to record its error state as a trace tag.
+        # DEV: This means errors cannot be reported if there's not active span.
+        # DEV: Ideally, we'd have a dedicated error reporting stream for all of ddtrace.
+        # DEV: The same comment applies to the {.extract_tags}.
```
ivoanjo, line 73: To be honest, this warning is important enough that I would suggest is actually worth copy-pasting into `#extract_tags`, rather than only referring to it here.

lib/datadog/tracing/distributed/datadog.rb
```diff
@@ -0,0 +1,150 @@
...
+            active_trace = Tracing.active_trace
+            active_trace.set_tag('_dd.propagation_error', 'extract_max_size') if active_trace
+
+            ::Datadog.logger.warn(
+              "Failed to extract x-datadog-tags: tags are too large (size:#{tags.size} " \
+                "limit:#{::Datadog.configuration.tracing.x_datadog_tags_max_length}). This limit can be configured " \
+                'through the DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH environment variable.'
+            )
+            return
+          end
```
ivoanjo, line 128: This seems to be almost exactly the same as for `#inject_tags` (I believe the only change is `inject_max_size` vs `extract_max_size` for the kind of error). Perhaps consider extracting to a shared method?

lib/datadog/tracing/distributed/datadog.rb
```diff
@@ -0,0 +1,150 @@
...
+
+          tags_hash = DatadogTagsCodec.decode(tags)
+          # Only extract keys with the expected Datadog prefix
+          tags_hash.select! do |key, _|
+            key.start_with?(Tracing::Metadata::Ext::Distributed::TAGS_PREFIX) && key != EXCLUDED_TAG
+          end
```
ivoanjo, line 134: Minor: Consider moving this inside the `DatadogTagsCodec`? It's kinda weird that the `DatadogTagsCodec` returns things that are not Datadog tags and makes the caller do the work of actually filtering out non-datadog stuff.

lib/datadog/tracing/distributed/ext.rb
```diff
@@ -0,0 +1,31 @@
...
+
+        # Prefix used by all Datadog-specific distributed tags
+        DATADOG_PREFIX = 'x-datadog-'
+
+        # B3 keys used for distributed tracing.
+        # @see https://github.com/openzipkin/b3-propagation
+        B3_HEADER_TRACE_ID = 'x-b3-traceid'
+        B3_HEADER_SPAN_ID = 'x-b3-spanid'
+        B3_HEADER_SAMPLED = 'x-b3-sampled'
+        B3_HEADER_SINGLE = 'b3'
```
ivoanjo, line 27: Should these be moved to the B3 classes? They only seem to be used there.

lib/datadog/tracing/distributed/metadata/b3.rb
```diff
@@ -1,55 +0,0 @@
...
-require_relative 'parser'
-require_relative '../headers/ext'
-require_relative '../../trace_digest'
-
-module Datadog
-  module Tracing
-    module Distributed
-      module Metadata
-        # B3 provides helpers to inject or extract metadata for B3 style headers
-        module B3
```
ivoanjo, line 13: Whaaaaat this entire class was copy-pasted before mostly to change headers to metadata :O :O :O 

Big thanks for improving this! 🙇

spec/datadog/tracing/contrib/grpc/distributed/propagation_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    describe '.extract' do
+      subject(:extract) { propagation.extract(data) }
+      let(:trace_digest) { extract }
+
+      # Metadata values can also be arrays
+      # https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md
+      context 'given populated data in array format' do
+        let(:data) do
+          { 'x-datadog-trace-id' => %w[12345 67890],
+            'x-datadog-parent-id' => %w[98765 43210],
```
ivoanjo, line 21: This looks weird to me, as it somehow doesn't seem realistic. Other than a bug, would we expect this to ever happen in for these specific headers?

It may be worth commenting here (or in the original class) why we include support for this, as I myself can't really figure it out. Is it just future-proofing? In case we actually add a new header that uses arrays?

spec/datadog/tracing/distributed/b3_single_spec.rb
```diff
@@ -29,7 +27,7 @@
         )
       end
 
-      it { is_expected.to eq(Datadog::Tracing::Distributed::Headers::Ext::B3_HEADER_SINGLE => '2710-4e20') }
+      it { is_expected.to eq('b3' => '2710-4e20') }
```
ivoanjo, line 30: In other cases in this PR, I've very much agreed with inlining the actual header names, because `x-datadog-potato` is reasonably obvious what it is.

Here.... I'm way less sure. Can I convince you to add a `let(:b3_single_header) { 'b3' }` and use that instead? It seems to be me that it would help with readibility :)

spec/datadog/tracing/distributed/fetcher_spec.rb
```diff
@@ -2,69 +2,47 @@
...
-        context "fetched as #{header}" do
-          let(:env) { { env_header('request_id') => 'rid' } }
-
-          it { expect(headers.header(header)).to eq('rid') }
-        end
-      end
+    context 'with a string value associated' do
+      let(:data) { { key => 'value' } }
+      it { is_expected.to eq('value') }
     end
```
ivoanjo, line 25: This test description is somewhat misleading, because it suggests that the fetcher is doing something based on a string value being associated or not, where actually that's not the case -- it works with any type.

```suggestion
    context 'with a value associated' do
      let(:data) { { key => 'value' } }
      it { is_expected.to eq('value') }
    end
```

lib/datadog/tracing/distributed/b3_single.rb
```diff
@@ -0,0 +1,67 @@
...
+        def extract(env)
+          fetcher = @fetcher.new(env)
+          value = fetcher[@key]
+
+          return unless value
+
+          parts = value.split('-')
+          trace_id = Helpers.value_to_id(parts[0], 16) unless parts.empty?
+          span_id = Helpers.value_to_id(parts[1], 16) if parts.length > 1
+          sampling_priority = Helpers.value_to_number(parts[2]) if parts.length > 2
```
ivoanjo, line 53: I believe these should go through the `fetcher`, right? Or otherwise, should we just drop the `id` and `number` methods from the fetcher and have callers use the `Helpers` directly?

spec/datadog/tracing/distributed/fetcher_spec.rb
```diff
@@ -2,69 +2,47 @@
...
   end
 
   describe '#id' do
```
ivoanjo, line 28: This seems somewhat out of place -- should it instead be in a `helpers_spec` perhaps?

Same for `#number` below

ivoanjo (commented):
Here's my full review :) 

First, let me thank you for getting rid of all those duplicate copy-paste classes and specs. Goodbye!

Second, I couldn't comment on the rename of `spec/datadog/tracing/metadata_spec.rb → spec/datadog/tracing/grpc_spec.rb` but it doesn't seem correct -- can you doublecheck?

Third and finally, I know it can be hard in these cases, but I would suggest trying to either refactor OR move stuff in the same PR, and not do both at once. Doing so makes it really hard to follow what code was actually changed -- 200 lines may be copy pasted and exactly the same, and then in the middle there's a 1-line change that may be relevant to consider.

TonyCTHsu (approved):
I like changes but left some thoughts about naming after reading through.

lib/datadog/tracing/contrib/http/distributed/fetcher.rb
```diff
@@ -0,0 +1,34 @@
...
+          class Fetcher < Tracing::Distributed::Fetcher
+            using Core::Utils::Refinement::String unless ::String.method_defined?(:+@)
+
+            # TODO: Don't assume Rack format.
+            #       Make distributed tracing headers apathetic.
+            # DEV: Should we try to parse both verbatim an Rack-formatted headers,
+            # DEV: given Rack-formatted is the most common format in Ruby?
+            def [](name)
+              # DEV: `String#+@` is not needed in Ruby >= 3.0, as interpolated strings are not frozen.
+              rack_header = (+"http-#{name}").upcase!.tr('-', '_')
```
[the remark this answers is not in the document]
marcotc: > Furthermore, as this is an optimization, should we use tr! instead of tr?

I missed this one, thank you!

lib/datadog/tracing/distributed/headers/ext.rb
```diff
@@ -1,31 +1,17 @@
...
-          B3_HEADER_SAMPLED = 'x-b3-sampled'.freeze
-          B3_HEADER_SINGLE = 'b3'.freeze
-
-          # gRPC metadata keys for distributed tracing. https://github.com/grpc/grpc-go/blob/v1.10.x/Documentation/grpc-metadata.md
-          GRPC_METADATA_TRACE_ID = 'x-datadog-trace-id'.freeze
-          GRPC_METADATA_PARENT_ID = 'x-datadog-parent-id'.freeze
-          GRPC_METADATA_SAMPLING_PRIORITY = 'x-datadog-sampling-priority'.freeze
-          GRPC_METADATA_ORIGIN = 'x-datadog-origin'.freeze
+          include Distributed::Ext
         end
```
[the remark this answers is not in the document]
marcotc: I moved them now because:
* It was possible.
* It allows us to starting using the correct new namespace for new files that reference these constants.

It was quite a bit of work that I'd rather accept as done now, given it's already done in the PR, than to delay to a poor soul to do it again during 2.0.

lib/datadog/tracing/distributed/propagation.rb
```diff
@@ -0,0 +1,110 @@
...
+        # DEV-2.0: argument is the common use case, but also allows us to set error tags in the `trace`
+        # DEV-2.0: if needed.
+        # DEV-2.0: Ideally, we'd have a separate stream to report tracer errors and never
+        # DEV-2.0: touch the active span.
+        #
+        # @param digest [TraceDigest]
+        # @param data [Hash]
+        def inject!(digest, data)
+          if digest.nil?
+            ::Datadog.logger.debug('Cannot inject distributed trace data: digest is nil.')
```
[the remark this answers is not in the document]
marcotc: This method is part of our public API, and its contract is to never fail. It normally appears in the middle of a user's application code, thus rising the error is not great, unless we truly want to interrupt the host application's code path.

I added a comment that this method should never raise errors, but instead log them.

lib/datadog/tracing/distributed/propagation.rb
```diff
@@ -0,0 +1,110 @@
...
+            return
+          end
+
+          digest = digest.to_digest if digest.is_a?(TraceOperation)
```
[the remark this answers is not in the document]
marcotc: 🦆

lib/datadog/tracing/distributed/propagation.rb
```diff
@@ -0,0 +1,110 @@
...
+        # @param data [Hash]
+        def extract(data)
+          trace_digest = nil
+          dd_trace_digest = nil
+
+          ::Datadog.configuration.tracing.distributed_tracing.propagation_extract_style.each do |style|
+            propagator = @propagation_styles[style]
+            next if propagator.nil?
+
+            # DEV: `propagator.extract` will return `nil`, where `{STYLE}#extract` will not
```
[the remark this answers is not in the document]
marcotc: This seems incorrect now, it was referring to both this class, `Propagation`, and the propagation styles (e.g. `B3`, `Datadog`). But they can both return `nil`, so this doesn't apply.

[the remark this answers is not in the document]
marcotc: I removed the comment.

lib/datadog/tracing/distributed/propagation.rb
```diff
@@ -0,0 +1,110 @@
...
+              trace_digest = extracted_trace_digest
+            else
+              unless trace_digest.trace_id == extracted_trace_digest.trace_id \
+                    && trace_digest.span_id == extracted_trace_digest.span_id
+                # Return an empty/new trace digest if we have a mismatch in values extracted
+                msg = "#{trace_digest.trace_id} != #{extracted_trace_digest.trace_id} && " \
+                    "#{trace_digest.span_id} != #{extracted_trace_digest.span_id}"
+                ::Datadog.logger.debug("Cannot extract distributed trace data: extracted styles differ, #{msg}")
+                # DEV: This will return from `self.extract` not this `each` block
+                return TraceDigest.new
```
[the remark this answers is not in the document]
marcotc: I refactored this and documented it: basically, we don't know what context we want to be part of as there's conflicting information. We chose to play it safe and create a branch new context here (through `TraceDigest.new`).

lib/datadog/tracing/distributed/propagation.rb
```diff
@@ -0,0 +1,110 @@
...
+            # No previously extracted trace data, use the one we just extracted
+            if trace_digest.nil?
+              trace_digest = extracted_trace_digest
+            else
+              unless trace_digest.trace_id == extracted_trace_digest.trace_id \
+                    && trace_digest.span_id == extracted_trace_digest.span_id
+                # Return an empty/new trace digest if we have a mismatch in values extracted
+                msg = "#{trace_digest.trace_id} != #{extracted_trace_digest.trace_id} && " \
+                    "#{trace_digest.span_id} != #{extracted_trace_digest.span_id}"
+                ::Datadog.logger.debug("Cannot extract distributed trace data: extracted styles differ, #{msg}")
```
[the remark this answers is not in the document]
marcotc: 👍, we need a cop for this! 😅

lib/datadog/tracing/contrib/http/distributed/propagation.rb
```diff
@@ -0,0 +1,29 @@
...
+require_relative 'datadog'
+
+module Datadog
+  module Tracing
+    module Contrib
+      module HTTP
+        module Distributed
+          # Extracts and injects propagation through HTTP headers.
+          class Propagation < Tracing::Distributed::Propagation
+            def initialize
```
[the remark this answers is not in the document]
marcotc: I didn't have time to do the whole thing, which involves touching `components.rb` and managing the lifecycle of `Propagation`.

The changes that are needed in the future, to perform that correct dependency injection, will work directly with these classes in this PR: they are now classes, not singletons, and are in their correct namespace.

lib/datadog/tracing/distributed/b3.rb
```diff
@@ -0,0 +1,63 @@
...
+  module Tracing
+    module Distributed
+      # B3-style trace propagation.
+      # @see https://github.com/openzipkin/b3-propagation#multiple-headers
+      class B3
+        def initialize(
+          trace_id: Ext::B3_HEADER_TRACE_ID,
+          span_id: Ext::B3_HEADER_SPAN_ID,
+          sampled: Ext::B3_HEADER_SAMPLED,
+          fetcher: Fetcher
```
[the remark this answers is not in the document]
marcotc: I can remove the default, no real preference here.
The default is a nice "I'm just accessing it like a hash", which is what HTTP should be doing, but then there's Rack...

I'll remove it.

lib/datadog/tracing/distributed/b3.rb
```diff
@@ -0,0 +1,63 @@
...
+module Datadog
+  module Tracing
+    module Distributed
+      # B3-style trace propagation.
+      # @see https://github.com/openzipkin/b3-propagation#multiple-headers
+      class B3
+        def initialize(
+          trace_id: Ext::B3_HEADER_TRACE_ID,
+          span_id: Ext::B3_HEADER_SPAN_ID,
+          sampled: Ext::B3_HEADER_SAMPLED,
```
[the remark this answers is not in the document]
marcotc: Ha, these are not headers, they are keys. When propagation happens in non-http cases, header becomes a misnomer, like in gPRC. But I can see `trace_id_key` as more clear than `trace_id` here. Will change it.

lib/datadog/tracing/contrib/http/distributed/propagation.rb
```diff
@@ -0,0 +1,29 @@
...
+              super(
+                propagation_styles: {
+                  Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_B3 => B3.new,
+                  Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_B3_SINGLE_HEADER => B3Single.new,
+                  Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_DATADOG => Datadog.new,
+                })
```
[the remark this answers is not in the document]
marcotc: 🤔 , let me try this!

spec/datadog/tracing/contrib/grpc/distributed/propagation_spec.rb
marcotc: gPRC metadata is a list of `key,value` tuples, and keys can be repeated: https://github.com/grpc/grpc-go/blob/master/Documentation/grpc-metadata.md#constructing-metadata

The Ruby gRPC gem "conveniently" uses a Hash to represent this data structure. When more than one value is present, the key becomes an array. I tested this and indeed it can be both an integral or list value.

There's a comment about this in the gPRC fetcher: `# Metadata values can be arrays (multiple values for the same key)`

marcotc: I expanded that comment in the Fetcher as well.

spec/datadog/tracing/distributed/b3_single_spec.rb
marcotc: Sure, that seems fine!

lib/datadog/tracing/contrib/http/distributed/propagation.rb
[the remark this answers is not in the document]
marcotc: It worked! It looks great, thank you for the suggestion!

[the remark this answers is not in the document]
marcotc: -82 lines!

lib/datadog/tracing/distributed/b3.rb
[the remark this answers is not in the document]
marcotc: Done.

lib/datadog/tracing/distributed/b3.rb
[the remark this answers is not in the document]
marcotc: Done.

lib/datadog/tracing/distributed/datadog.rb
marcotc: Done!

lib/datadog/tracing/distributed/datadog.rb
marcotc: Applied manually, due to a conflict.

lib/datadog/tracing/distributed/datadog.rb
marcotc: Done.

lib/datadog/tracing/distributed/datadog.rb
marcotc: Regarding sharing code between `extract_tags` and `inject_tags`, I don't think I can refactor it to share code.
Every other line has a different error message or string.

Unless you are referring to a specific line range, I don't have good ideas to refactor this.

lib/datadog/tracing/distributed/datadog.rb
marcotc: I think this is a hairy point: `Distributed::TAGS_PREFIX` only referrers to `_dd.p.*` tags, which is a subset of all `_dd.*` tags. The `DatadogTagsCodec` is meant to parse any tags in our custom Datadog-created format, but doesn't care if they are `_dd.p.*` or not.

lib/datadog/tracing/distributed/ext.rb
marcotc: Yes, I have this comment at the top of this file, and in my opinion we don't need these to be part of the public API:
```
      # DEV-2.0: These constants are part of the public API through {Datadog::Tracing::Distributed::Headers::Ext}.
      # DEV-2.0: We should not expose these constants, as we might not move them during refactor, and they are easily
      # DEV-2.0: and publicly documented in Datadog's and B3's documentation.
```
I'll add that we should move B3 and other close to their usage location.

lib/datadog/tracing/distributed/metadata/b3.rb
marcotc: That's the main reason I'm doing this: I have to add a new propagator, and didn't want to write 2.

spec/datadog/tracing/distributed/b3_single_spec.rb
marcotc: Done for b3 single header.
I was going to do b3 multi header, but those actually read nicely like:
`x-b3-traceid`
`x-b3-spanid`
`x-b3-sampled`

Do you think the b3 multi header would benefit from it as well?

lib/datadog/tracing/distributed/b3_single.rb
marcotc: This is debatable: the fetcher knows to get data from the provided input `data`. It's currently a `key/value` accessor adapter.

This code her is a subdivision of a `value`, returned from a `key/value` pair.
Because this subdivision has nothing to do with retrieving the value from the `key/value` `data` input, I'm inclined to leave this as is. But as I said, this is debatable.

The fetcher is also protocol-specific (HTTP, gRPC), not style-specific (B3, Datadog) today, so _currently_ I think it makes sense to leave this as is. But this classification of the fetcher I just mentioned is arbitrary so 🤷

lib/datadog/tracing/distributed/b3.rb
```diff
@@ -0,0 +1,63 @@
...
+require_relative 'helpers'
+require_relative '../trace_digest'
+
+module Datadog
+  module Tracing
+    module Distributed
+      # B3-style trace propagation.
+      # @see https://github.com/openzipkin/b3-propagation#multiple-headers
+      class B3
+        def initialize(
```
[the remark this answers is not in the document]
marcotc: Ha, read my mind! This was done!

lib/datadog/tracing/distributed/fetcher.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+# typed: true
+
+require_relative 'helpers'
+
+module Datadog
+  module Tracing
+    module Distributed
+      # Common fetcher that retrieves fields from a Hash data input
+      class Fetcher
```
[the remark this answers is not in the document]
marcotc: It could! Any suggestions? 😄
It doesn't know how to set data btw, only read. Setting data is always done into tracer-internal classes (`TraceDigest.new(your,data,here)`), thus doesn't need a custom "Settter" per protocol.

Also, this is completely internal. I'm more than happy to change this in the future.

lib/datadog/tracing/distributed/propagation.rb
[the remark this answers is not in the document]
ivoanjo: I see. Should we perhaps return `false` or `:error` so the caller can figure out something is off even if they don't have logging enabled?

lib/datadog/tracing/distributed/propagation.rb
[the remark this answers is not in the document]
ivoanjo: Thanks for improving this, I really like how the new version make very explicit the decision we're making :)

lib/datadog/tracing/contrib/http/distributed/propagation.rb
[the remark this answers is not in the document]
ivoanjo: I'll admit that I'm not convinced these "midway points" are useful, but I'll trust you this one :)

lib/datadog/tracing/contrib/http/distributed/propagation.rb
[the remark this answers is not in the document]
ivoanjo: Yay, great victory for the code review cabal! :)

spec/datadog/tracing/contrib/grpc/distributed/propagation_spec.rb
ivoanjo: Indeed, I can understand that in the general case, gprc metadata can be arrays.

But my question was more in the particular case of these keys -- are they expected in regular operation to ever be arrays?

E.g., is an array for any of these keys an indication that something is wrong, or something that is normal and happens every day? That's the part that puzzled me a bit.

(And if it's an error, should we perhaps log or mark the occurrence in some way?)

spec/datadog/tracing/distributed/b3_single_spec.rb
ivoanjo: Nah, I think single header is the only one that's confusing. I think the others are specific enough to be obvious in context (and one can even quickly grep/google them to find out what they are)

lib/datadog/tracing/distributed/datadog.rb
ivoanjo: Aaaaahhhh I see. Thanks for clarifying, I had indeed missed that.

lib/datadog/tracing/distributed/ext.rb
ivoanjo: If the only reason we're not moving them to the places we'd like to is backwards compatibility, have you considered making a copy instead?

E.g. leaving the original definitions in `Datadog::Tracing::Distributed::Headers::Ext` with a note saying "these should not be used and will be removed", and then add a local copy to the B3/etc classes that is used by those classes instead.

That way, we wouldn't need to do any further changes for 2.0 other than deleting `lib/datadog/tracing/distributed/headers/ext.rb`

lib/datadog/tracing/distributed/datadog.rb
ivoanjo: I was thinking along the lines of:

```diff
diff --git a/lib/datadog/tracing/distributed/datadog.rb b/lib/datadog/tracing/distributed/datadog.rb
index ffaebbf83..fa1d21b39 100644
--- a/lib/datadog/tracing/distributed/datadog.rb
+++ b/lib/datadog/tracing/distributed/datadog.rb
@@ -72,25 +72,12 @@ module Datadog
         # DEV: Ideally, we'd have a dedicated error reporting stream for all of ddtrace.
         def inject_tags(digest, data)
           return if digest.trace_distributed_tags.nil? || digest.trace_distributed_tags.empty?
-
-          if ::Datadog.configuration.tracing.x_datadog_tags_max_length <= 0
-            active_trace = Tracing.active_trace
-            active_trace.set_tag('_dd.propagation_error', 'disabled') if active_trace
-            return
-          end
+          return try_setting_propagation_error_tag_on_active_trace(error_kind: 'disabled') if datadog_tags_disabled?

           encoded_tags = DatadogTagsCodec.encode(digest.trace_distributed_tags)

-          if encoded_tags.size > ::Datadog.configuration.tracing.x_datadog_tags_max_length
-            active_trace = Tracing.active_trace
-            active_trace.set_tag('_dd.propagation_error', 'inject_max_size') if active_trace
-
-            ::Datadog.logger.warn(
-              'Failed to inject x-datadog distributed tracing tags: too many tags for configured limit ' \
-              "(size:#{encoded_tags.size} >= limit:#{::Datadog.configuration.tracing.x_datadog_tags_max_length}). This " \
-              'limit can be configured with the DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH environment variable.'
-            )
-            return
+          if too_many_tags?(encoded_tags.size, situation: 'inject')
+            return try_setting_propagation_error_tag_on_active_trace(error_kind: 'inject_max_size')
           end

           data[@tags_key] = encoded_tags
@@ -111,23 +98,10 @@ module Datadog
         def extract_tags(fetcher)
           tags = fetcher[@tags_key]
           return if !tags || tags.empty?
+          return try_setting_propagation_error_tag_on_active_trace(error_kind: 'disabled') if datadog_tags_disabled?

-          if ::Datadog.configuration.tracing.x_datadog_tags_max_length <= 0
-            active_trace = Tracing.active_trace
-            active_trace.set_tag('_dd.propagation_error', 'disabled') if active_trace
-            return
-          end
-
-          if tags.size > ::Datadog.configuration.tracing.x_datadog_tags_max_length
-            active_trace = Tracing.active_trace
-            active_trace.set_tag('_dd.propagation_error', 'extract_max_size') if active_trace
-
-            ::Datadog.logger.warn(
-              "Failed to extract x-datadog-tags: tags are too large (size:#{tags.size} >= " \
-                "limit:#{::Datadog.configuration.tracing.x_datadog_tags_max_length}). This limit can be configured " \
-                'through the DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH environment variable.'
-            )
-            return
+          if too_many_tags?(tags.size, situation: 'extract')
+            return try_setting_propagation_error_tag_on_active_trace(error_kind: 'inject_max_size')
           end

           tags_hash = DatadogTagsCodec.decode(tags)
@@ -144,6 +118,28 @@ module Datadog
           )
         end

+        def try_setting_propagation_error_tag_on_active_trace(error_kind:)
+          active_trace = Tracing.active_trace
+          active_trace.set_tag('_dd.propagation_error', error_kind) if active_trace
+          nil
+        end
+
+        def datadog_tags_disabled?
+          ::Datadog.configuration.tracing.x_datadog_tags_max_length <= 0
+        end
+
+        def too_many_tags?(tags_size, situation:)
+          return false if tags_size <= ::Datadog.configuration.tracing.x_datadog_tags_max_length
+
+          ::Datadog.logger.warn(
+            "Failed to #{situation} x-datadog-tags: too many tags for configured limit (size:#{tags_size} >= " \
+              "limit:#{::Datadog.configuration.tracing.x_datadog_tags_max_length}). This limit can be configured " \
+              'through the DD_TRACE_X_DATADOG_TAGS_MAX_LENGTH environment variable.'
+          )
+
+          true
+        end
+
         # We want to exclude tags that we don't want to propagate downstream.
         EXCLUDED_TAG = '_dd.p.upstream_services'
         private_constant :EXCLUDED_TAG
```

lib/datadog/tracing/propagation/http.rb
```diff
@@ -1,117 +1,14 @@
...
-          end
-
-          # Return the extracted trace headers if we found one or else a new empty trace headers
-          # Always return the Datadog trace headers if one exists since it has more
-          #   information than the B3 headers e.g. origin, expanded priority
-          #   sampling values, etc
-          dd_trace_digest || trace_digest || nil
-        end
-      end
+      HTTP = Tracing::Contrib::HTTP::Distributed::Propagation.new
```
ivoanjo, line 11: I was searching the codebase for `Propagation.new` and this seems to be the only place where we instance the `HTTP` propagation class. Is this expected? Should we match what we do with grpc (e.g. have an `INSTANCE` in the "new place"?)

lib/datadog/tracing/contrib/grpc/distributed/propagation.rb
```diff
@@ -0,0 +1,38 @@
...
+                  Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_B3_SINGLE_HEADER =>
+                    Tracing::Distributed::B3Single.new(fetcher: Fetcher),
+                  Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_DATADOG =>
+                    Tracing::Distributed::Datadog.new(fetcher: Fetcher)
+                })
+            end
+
+            # DEV: Singleton kept until a larger refactor is performed.
+            # DEV: See {Datadog::Tracing::Distributed::Propagation#initialize} for more information.
+            INSTANCE = Propagation.new
```
ivoanjo, line 32: Minor: I slightly wonder if we should take the next step and also kill the `GRPC::Distributed::Propagation` subclass, since it's only a shortcut for calling `Tracing::Distributed::Propagation` with the defaults :)

lib/datadog/tracing/contrib/http/distributed/b3.rb
```diff
@@ -0,0 +1,23 @@
...
+      module HTTP
+        module Distributed
+          # B3-style trace propagation through HTTP headers.
+          # @see https://github.com/openzipkin/b3-propagation#multiple-headers
+          class B3 < Tracing::Distributed::B3
+            def initialize
+              super(fetcher: Fetcher)
+            end
+          end
+        end
```
ivoanjo, line 19: Should this/b3_single/datadog be removed after the refactor where propagation directly instances the superclass with the fetcher?

lib/datadog/tracing/distributed/b3.rb
```diff
@@ -0,0 +1,63 @@
...
+          fetcher:,
+          trace_id_key: Ext::B3_HEADER_TRACE_ID,
+          span_id_key: Ext::B3_HEADER_SPAN_ID,
+          sampled: Ext::B3_HEADER_SAMPLED
```
ivoanjo, line 18: Minor: Suggest calling it `sampled_key`, to align with the two others :)

ivoanjo (commented):
Left a few more notes! Also see my note on the rename of `spec/datadog/tracing/grpc_spec.rb` seeming incomplete (the spec looks like it needs updating?).

lib/datadog/tracing/propagation/http.rb
marcotc: This one is public, the gRPC is not.
I'll see if they can match, or if it makes sense to.

lib/datadog/tracing/distributed/b3.rb
marcotc: I missed this one, good catch!

lib/datadog/tracing/contrib/http/distributed/b3.rb
marcotc: 👍

lib/datadog/tracing/contrib/grpc/distributed/propagation.rb
marcotc: I think this is complex enough that it should be encapsulated:
```ruby
                propagation_styles: {
                  Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_B3 =>
                    Tracing::Distributed::B3.new(fetcher: Fetcher),
                  Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_B3_SINGLE_HEADER =>
                    Tracing::Distributed::B3Single.new(fetcher: Fetcher),
                  Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_DATADOG =>
                    Tracing::Distributed::Datadog.new(fetcher: Fetcher)
                })
```
It's the one place where we wire the protocol-specific logic.
I'd like to request permission to leave it as it.

lib/datadog/tracing/propagation/http.rb
marcotc: Like you said, this is the same as `INSTANCE`, but both should be removed in a future breaking change release.
This one is called `HTTP` because it's a documented public API, so I don't want to touch it, and I don't think it helps too much to have `INSTANCE` here as an alias.

Given none of them should remain for too long, I'd like to leave it as is.

lib/datadog/tracing/contrib/http/distributed/propagation.rb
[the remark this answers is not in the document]
marcotc: > I'll admit that I'm not convinced these "midway points" are useful, but I'll trust you this one :)

I had to break down the work somehow, given I already put more time that I budgeted for for this refactor, so I chose a stopping point that allowed me move forward with adding new propagators without pulling hairs.
In an ideal scenario, the whole thing would be refactored, but I'd probably need another week to do it.

lib/datadog/tracing/distributed/propagation.rb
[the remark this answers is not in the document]
marcotc: I added and documented 3 return values:
`true` successfully injected; `false` no error, but no injection was found; `nil` error, please check logs.

marcotc:
`spec/datadog/tracing/metadata_spec.rb` is now reverted, it was indeed a mistake! 🙇

spec/datadog/tracing/distributed/fetcher_spec.rb
marcotc: Done, refactored these and reconciled both fetcher and helpers specs.

lib/datadog/tracing/distributed/ext.rb
marcotc: I can do that!

lib/datadog/tracing/distributed/datadog.rb
marcotc: Sweet patch, applied! Looks great after, thank you so much!

lib/datadog/tracing/contrib/http/distributed/fetcher.rb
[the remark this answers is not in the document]
marcotc: I find that this pattern of "is this string mutable, if not dup" is common in other parts of the tracer, and `+@` is super helpful. Because I'm removing the use case here, I'll remove the refinement, but I think it's a great tool to have in other scenarios.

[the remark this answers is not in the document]
marcotc: Removed refinement changes and made suggested changes.

lib/datadog/tracing/distributed/ext.rb
marcotc: Done, it looks much nicer, thank you!

lib/datadog/tracing/contrib/http/distributed/propagation.rb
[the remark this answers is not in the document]
ivoanjo: Mega fair! Hope you didn't take the above as criticism of you, it's more of a "I suspect this detour may need to be slightly redone and so I would prefer omitting it, but I don't have any data/proof to match my suspicion and this is reasonable as well so since we need to make a decision, I trust yours."

lib/datadog/tracing/contrib/http/distributed/fetcher.rb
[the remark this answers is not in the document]
ivoanjo: > I find that this pattern of "is this string mutable, if not dup" is common in other parts of the tracer, and +@ is super helpful. Because I'm removing the use case here, I'll remove the refinement, but I think it's a great tool to have in other scenarios.

I guess it's just me not being used to either `+` or refinements, so I may be putting the bar to use them a bit higher, e.g. I think that indeed we may end up using them, but I think in this case the added complexity was not completely worth it.

spec/datadog/tracing/contrib/grpc/distributed/propagation_spec.rb
ivoanjo: Let's leave it as-is then, until new info on this shows up :)

lib/datadog/tracing/distributed/propagation.rb
```diff
@@ -0,0 +1,129 @@
...
+                # It's safer to create a new context than to attach ourselves to the wrong context.
+                trace_digest = TraceDigest.new
+
+                ::Datadog.logger.debug do
+                  'Cannot extract distributed trace data: extracted styles differ, ' \
+                  "#{trace_digest.trace_id} != #{extracted_trace_digest.trace_id} && " \
+                  "#{trace_digest.span_id} != #{extracted_trace_digest.span_id}"
+                end
+
+                return trace_digest # Early return from the whole method
```
ivoanjo, line 116: This `trace_digest = TraceDigest.new` seems to be a bit too early, making this code incorrect. E.g. it because we replace the `trace_digest` before logging, the logged message will always be nonsense.

Suggested fix:

```suggestion
                # We have two mismatched propagation contexts
                ::Datadog.logger.debug do
                  'Cannot extract distributed trace data: extracted styles differ, ' \
                  "#{trace_digest.trace_id} != #{extracted_trace_digest.trace_id} && " \
                  "#{trace_digest.span_id} != #{extracted_trace_digest.span_id}"
                end

                # It's safer to create a new context than to attach ourselves to the wrong context
                return TraceDigest.new # Early return from the whole method
```

lib/datadog/tracing/distributed/propagation.rb
```diff
@@ -0,0 +1,129 @@
...
+            if extracted_trace_digest && style == Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_DATADOG
+              dd_trace_digest = extracted_trace_digest
+            end
+
+            # No previously extracted trace data, use the one we just extracted
+            if trace_digest.nil?
+              trace_digest = extracted_trace_digest
+            else
+              unless trace_digest.trace_id == extracted_trace_digest.trace_id \
+                    && trace_digest.span_id == extracted_trace_digest.span_id
```
ivoanjo, line 105: Should we also look to see if non-datadog digests match the datadog digest? This was not in the code prior to the refactoring either, but I'm curious if it would make sense.

spec/datadog/opentracer/text_map_propagator_spec.rb
```diff
@@ -51,13 +51,13 @@
...
+          .with('x-datadog-trace-id', trace_id)
         expect(carrier).to have_received(:[]=)
-          .with(Datadog::OpenTracer::DistributedHeaders::HTTP_HEADER_PARENT_ID, span_id)
+          .with('x-datadog-parent-id', span_id)
         expect(carrier).to have_received(:[]=)
-          .with(Datadog::OpenTracer::DistributedHeaders::HTTP_HEADER_SAMPLING_PRIORITY, sampling_priority)
+          .with('x-datadog-sampling-priority', sampling_priority)
         expect(carrier).to have_received(:[]=)
-          .with(Datadog::OpenTracer::DistributedHeaders::HTTP_HEADER_ORIGIN, origin)
+          .with(Datadog::OpenTracer::DistributedHeaders::ORIGIN_KEY, origin)
```
ivoanjo, line 60: I... don't think `ORIGIN_KEY` is in this class anymore? (also applies to a few other opentracer specs as well)

spec/datadog/tracing/distributed/helpers_spec.rb
```diff
@@ -73,12 +73,9 @@
     context 'when value is' do
       [
         [nil, nil],
+        ['not a number', nil],
         ['0', nil],
-        ['value', nil],
-        ['867-5309', nil],
```
ivoanjo, line 78: This test in particular I think was useful because:

```
[4] pry(main)> '867-5309'.to_i
=> 867
```

and actually we have extra code to make it `nil`

(Same note applies below)

lib/datadog/tracing/distributed/datadog.rb
```diff
@@ -0,0 +1,153 @@
...
+          @fetcher = fetcher
+        end
+
+        def inject!(digest, data)
+          return if digest.nil?
+
+          data[@trace_id_key] = digest.trace_id.to_s
+          data[@parent_id_key] = digest.span_id.to_s
+          data[@sampling_priority_key] = digest.trace_sampling_priority.to_s if digest.trace_sampling_priority
+          data[@origin_key] = digest.trace_origin.to_s unless digest.trace_origin.nil?
```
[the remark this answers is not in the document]
marcotc: > I'm looking forward to that safe navigation operator...

They should add that to the language :trollface:

lib/datadog/tracing/distributed/headers/ext.rb
```diff
@@ -1,31 +1,33 @@
...
 
 module Datadog
   module Tracing
     module Distributed
       module Headers
-        # HTTP headers one should set for distributed tracing.
-        # These are cross-language (eg: Python, Go and other implementations should honor these)
+        # DEV-2.0: This module only exists for backwards compatibility with the public API. It should be removed.
+        # @deprecated use [Datadog::Tracing::Distributed::Ext]
         # @public_api
```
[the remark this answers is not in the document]
marcotc: We can't, as it will completely hide it from our [public API docs](https://s3.amazonaws.com/gems.datadoghq.com/trace/docs/Datadog/Tracing/Distributed/Headers/Ext.html) and that's technically a breaking change.

Best I can do is a stern comment for now :)

lib/datadog/tracing/distributed/propagation.rb
marcotc: Great catch!

lib/datadog/tracing/distributed/propagation.rb
marcotc: I'm making larger changes to this logic, including how we prioritize the Datadog digest over others in a follow up PR. I'd rather touch this part there. I believe this logic will change quite a bit.

spec/datadog/opentracer/text_map_propagator_spec.rb
marcotc: Forgot to inline this one 😓 .
Done!

spec/datadog/tracing/distributed/helpers_spec.rb
marcotc: Restored as `'1.to_i returns 1'`, thank you!

spec/datadog/tracing/distributed/helpers_spec.rb
```diff
@@ -73,12 +73,10 @@
     context 'when value is' do
       [
         [nil, nil],
+        ['not a number', nil],
+        ['1.to_i returns 1', nil],
```
ivoanjo, line 77: Minor: I'm not entirely sure it's clear what we're testing here (and to be fair, I think this was part of the problem of the old test -- it was actually not clear what it was testing).

I suggest perhaps `['1 2 3', nil], # '1 2 3'.to_i => 1 and we have extra logic to avoid this`.

(Also applies below)

ivoanjo (approved):
Great refactoring! LGTM 👍

lib/datadog/tracing/distributed/headers/ext.rb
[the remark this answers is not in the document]
Judahmeek: @marcotc @ivoanjo

I'm pretty sure y'all implemented a breaking change anyway.

I've got a bot PR that broke trying to `Bump [ddtrace](https://github.com/DataDog/dd-trace-rb) from 1.6.1 to 1.7.0.` because `NameError: uninitialized constant Datadog::Tracing::Distributed::Headers`

Also, I could be wrong, but I don't think `Datadog::Tracing::Distributed::Ext` exists.

If it does, I'd appreciate a reference to it. ^^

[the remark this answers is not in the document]
ivoanjo: Apologies for the delay in answering @Judahmeek.

Most modules under `Datadog::Tracing::Distributed::Headers` are currently not considered part of the public API, as they are not tagged with `@public_api`.

Nevertheless, we have broken your setup, which is definitely not what we intended! 

Let us know if we can help you move to 1.7.0+, and also, if you can, can you share how you're using these modules so we can hopefully take that into consideration for future changes? That will help us evaluate the current API and if there's something we're currently not exposing and that we should.

PR #2354, opened 2022-11-08
Disable GC profiling on Ruby 3 by default due to Ruby VM Ractor-related bug

ivoanjo:
**What does this PR do?**:

This PR:

* Adds a toggle that can be used to enable/disable the GC profiling feature added in #2336
* Enables the GC profiling feature by default on Ruby 2.x and disables the feature by default on Ruby 3.x
* Exposes a new `DD_PROFILING_FORCE_ENABLE_GC` / `settings.profiling.advanced.force_enable_gc_profiling` setting to allow customers that don't use Ractors to manually re-enable the GC profiling feature.

**Motivation**:

While validating that the new profiler correctly works with Ractors, I discovered that adding specs that invoked Ractors caused the previously-built GC profiling feature (and attached specs) to start failing.

I was able to track this down to an actual VM bug, which I've reported upstream as <https://bugs.ruby-lang.org/issues/19112>.

To avoid impacting any customers (specifically, they would potentially observe profiles with incomplete data), I decided to disable GC profiling for Ruby 3.x customers for now.

I am hoping that the Ruby core team can be convinced to fix this issue and backport it to existing Ruby 3.x stable releases, so that we can again re-enable this feature for these users.

As an alternative, if this issue sticks around for a long time and we really want this feature to be on by default for Ruby 3.x customers, there are probably are ways for us to automatically detect if Ractors are in use or not. I chose not to pursue such a mechanism for now.

**Additional Notes**:

(None)

**How to test the change?**:

The change includes test coverage. You can also observe this feature by profiling a Ruby application:

* On Ruby 2.x you will see garbage collection frames
* On Ruby 3.x you will not see garbage collection frames unless you force enable the feature.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -64,7 +64,7 @@ struct cpu_and_wall_time_worker_state {
   // telling the sampling trigger loop to stop, but if we ever need to communicate more, we should move to actual
   // atomic operations. stdatomic.h seems a nice thing to reach out for.
   volatile bool should_run;
-
+  bool gc_profiling_enabled;
   VALUE cpu_and_wall_time_collector_instance;
 
   // When something goes wrong during sampling, we record the Ruby exception here, so that it can be "re-raised" on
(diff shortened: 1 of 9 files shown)


TonyCTHsu (approved):

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -128,7 +133,7 @@ void collectors_cpu_and_wall_time_worker_init(VALUE profiling_module) {
   // https://bugs.ruby-lang.org/issues/18007 for a discussion around this.
   rb_define_alloc_func(collectors_cpu_and_wall_time_worker_class, _native_new);
 
-  rb_define_singleton_method(collectors_cpu_and_wall_time_worker_class, "_native_initialize", _native_initialize, 2);
+  rb_define_singleton_method(collectors_cpu_and_wall_time_worker_class, "_native_initialize", _native_initialize, 3);
```
ivoanjo, line 136: For context, the `2` to `3` change here is the number of positional arguments to the method. This is documented [here](https://github.com/ruby/ruby/blob/master/doc/extension.rdoc#method-and-singleton-method-definition-).

Effectively there’s three ways to “wire up” a C function to be called as a Ruby method:

* if you pass in a number >= 0 then that’s the number of positional arguments to the method. E.g. `def foo(a, b, c)` => `rb_define_method(..., foo_implementation, 3)` and `foo_implementation` is expected to have a signature that receives self + 3 other arguments, e.g. `VALUE foo_implementation(VALUE self, VALUE a, VALUE b, VALUE c)`.
* if you pass in `-1` then Ruby will pass the arguments as a C array
* if you pas in `-2` then Ruby will pass in the arguments as a Ruby array (e.g. it’s the C equivalent of `def foo(*args)`)

I moved it from `2` to `3` since I was passing the new `gc_profiling_enabled` value

PR #2357, opened 2022-11-09
[PROF-5942] Disallow sampling ractors in profiler

ivoanjo:
**What does this PR do?**:

This PR builds on top of #2350 and changes the `Collectors::CpuAndWallTimeWorker` to never trigger sampling if any of its methods get called from outside the main ractor.

This means that the profiler will not have visibility beyond the main ractor. Building support for that will come later, as we'll need to deal with the complexity of the profiler potentially being called from multiple ractors.

**Motivation**:

We don't yet support Ractors, but on an application using Ractors the profiler should not fail or misbehave. The expected behavior is that it just won't profile Ractors beyond the main one.

**Additional Notes**:

(Nothing)

**How to test the change?**:

I've included partial coverage, but testing the GC-related behavior was harder than I expected (also because of #2354), so for now I didn't add it.

To test this feature, you can enable profiling on a Ruby app using ractors and confirm that no samples appear from Ractors, but otherwise profiling works correctly (for threads in the main Ractor).

Here's a test app I've been using:

```ruby
puts RUBY_DESCRIPTION

def fib(n)
  return n if n <= 1
  fib(n-1) + fib(n-2)
end

Thread.current.name = "#{Thread.current.object_id} main"

ractor1 = Ractor.new do
  Thread.current.name = "#{Thread.current.object_id} ractor1 main"

  Thread.new do
    Thread.current.name = "#{Thread.current.object_id} ractor1 background"

    loop { fib(40) }
  end

  sleep 1
  puts "ractor1 => #{Thread.list}"

  loop { fib(40) }
end

ractor2 = Ractor.new do
  Thread.current.name = "#{Thread.current.object_id} ractor2 main"

  Thread.new do
    Thread.current.name = "#{Thread.current.object_id} ractor2 background"

    loop { fib(40) }
  end

  sleep 1
  puts "ractor2 => #{Thread.list}"

  loop { fib(40) }
end

puts "main => #{Thread.list}"

sleep
```

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
@@ -323,6 +323,7 @@ static VALUE _native_sample_after_gc(DDTRACE_UNUSED VALUE self, VALUE collector_
 // Assumption 2: This function is allowed to raise exceptions. Caller is responsible for handling them, if needed.
 // Assumption 3: This function IS NOT called from a signal handler. This function is not async-signal-safe.
 // Assumption 4: This function IS NOT called in a reentrant way.
+// Assumption 5: This function is called from the main Ractor (if Ruby has support for Ractors).
 VALUE cpu_and_wall_time_collector_sample(VALUE self_instance) {
   struct cpu_and_wall_time_collector_state *state;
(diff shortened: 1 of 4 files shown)


lloeki (approved):
This all seems very reasonable!

PR #2358, opened 2022-11-09
[PROF-6288] Enable `-Werror` to turn compiler warnings into errors in CI

ivoanjo:
**What does this PR do?**:

This PR enables the `-Werror` compiler flag, which turns any compiler warnings into errors, when compiling profiler C code in CI.

I also refreshed the `extconf.rb` for the `ddtrace_profiling_loader` to match the compiler flags from `ddtrace_profiling_native_extension` as well.

**Motivation**:

Most compiler warnings are quite useful at point out potential bugs, and we want to make sure our code is warning-free on all Ruby versions we support.

**Additional Notes**:

We don't enable `-Werror` always because we can't control what compiler our customers use. E.g. they may use a new compiler which introduces a new warning which we haven't seen and it doesn't make sense to "break" profiling for them just because of that.

**This PR is on top of `ivoanjo/prof-5942-ractor-support` just out of convenience and does not really depend on anything on that PR**.

**How to test the change?**:

Check that CI is still green.

### DIFF
--- a/ext/ddtrace_profiling_loader/extconf.rb
+++ b/ext/ddtrace_profiling_loader/extconf.rb
@@ -24,6 +24,16 @@ def add_compiler_flag(flag)
   end
 end
 
+# Because we can't control what compiler versions our customers use, shipping with -Werror by default is a no-go.
+# But we can enable it in CI, so that we quickly spot any new warnings that just got introduced.
+add_compiler_flag '-Werror' if ENV['CI'] == 'true'
+
+# Older gcc releases may not default to C99 and we need to ask for this. This is also used:
+# * by upstream Ruby -- search for gnu99 in the codebase
+# * by msgpack, another ddtrace dependency
+#   (https://github.com/msgpack/msgpack-ruby/blob/18ce08f6d612fe973843c366ac9a0b74c4e50599/ext/msgpack/extconf.rb#L8)
+add_compiler_flag '-std=gnu99'
+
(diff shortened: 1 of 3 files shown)


ext/ddtrace_profiling_loader/extconf.rb
```diff
@@ -24,6 +24,16 @@ def add_compiler_flag(flag)
...
+add_compiler_flag '-Werror' if ENV['CI'] == 'true'
+
+# Older gcc releases may not default to C99 and we need to ask for this. This is also used:
+# * by upstream Ruby -- search for gnu99 in the codebase
+# * by msgpack, another ddtrace dependency
+#   (https://github.com/msgpack/msgpack-ruby/blob/18ce08f6d612fe973843c366ac9a0b74c4e50599/ext/msgpack/extconf.rb#L8)
+add_compiler_flag '-std=gnu99'
+
 # Gets really noisy when we include the MJIT header, let's omit it
 add_compiler_flag '-Wno-unused-function'
```
ivoanjo, line 38: This flag, as well as others below were already in `ext/ddtrace_profiling_native_extension/extconf.rb` but I had forgot to enable them for `ext/ddtrace_profiling_loader/extconf.rb` so I took the opportunity to copy them over as well.

lloeki (approved):
> We don't enable -Werror always because we can't control what compiler our customers use.

That was my fear upon jumping on this PR, but you addressed it immediately.

LGTM!

PR #2359, opened 2022-11-09
[PROF-5953] Ruby VM forking fixes for new profiler

ivoanjo:
**What does this PR do?**:

This PR collects fixes for two issues related to the new profiler and Ruby apps that employ fork. See individual commits for details.

**Motivation**:

We fully support Ruby applications that use fork, so I've been spending time making sure the new profiler is correct for these applications as well.

**Additional Notes**:

I suspect there may be a few more fixes needed before the new profiler is solid on Ruby apps that use fork, so expect more PRs on this subject.

**This PR is on top of #2358 as I wanted to avoid any conflicts, but is independent of it**

**How to test the change?**:

Both changes include test coverage. Additionally, the first change can be easily observed by checking that the profiler restarts successfully after a fork in a Ruby app that uses forking.


lloeki (approved):
LGTM

PR #2360, opened 2022-11-09
Support Sinatra 3

TonyCTHsu:
**What does this PR do?**

Create Sinatra's appraisal group avoids the version update be blocked. Hence we could include Sinatra 3 into our test suite.


spec/datadog/tracing/contrib/sinatra/activerecord_spec.rb
```diff
@@ -1,122 +0,0 @@
-# typed: ignore
```
marcotc, line 1: Why was this test file removed? Does it not provide value? Why was it added in first place?

marcotc (approved):

spec/datadog/tracing/contrib/sinatra/activerecord_spec.rb
TonyCTHsu: I think this test is duplicated with our ActiveRecord test, and does not provide much value.

Rakefile
```diff
@@ -376,6 +376,9 @@ task :ci do
   # explicitly test Hanami compatability
   declare '❌ 2.1 / ❌ 2.2 / ✅ 2.3 / ✅ 2.4 / ✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ❌ 3.0 / ❌ 3.1 / ❌ 3.2 / ❌ jruby' => 'bundle exec appraisal hanami-1 rake spec:hanami'
 
+  # explicitly test Sinatra compatability
+  declare '✅ 2.1 / ✅ 2.2 / ✅ 2.3 / ✅ 2.4 / ✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ jruby' => 'bundle exec appraisal sinatra rake spec:sinatra'
```
ivoanjo, line 380: I think something is off here. Is it on purpose that we repeat `'bundle exec appraisal sinatra rake spec:sinatra'` here and above? I suspect that will make the sinatra tests run twice in a row.

Appraisals
```diff
@@ -1556,6 +1587,11 @@ elsif ruby_version?('3.2')
     gem 'nokogiri', platform: :ruby # TODO: binary gem has max ruby version constraint excluding previews, switch to using minimum version constraint once a non-3.2-excluding binary gem is released
   end
 
+  appraise 'sinatra' do
+    gem 'sinatra'
+    gem 'rack-test'
+  end
```
ivoanjo, line 1593: Minor: Should we add a `>= 3` to this block on the supported Rubies, so that we ensure that we get the intended Sinatra version? (It's not a problem now, but I'm thinking this could be a just-in-case other dependency starts preventing the intended version from being picked up)

ivoanjo (commented):
Out of curiosity, does the Sinatra 3 integration need no changes? I somehow got the impression that we'd need to fix something to get it working.

Appraisals
marcotc: Adding a minimum version doesn't harm, and makes gem resolution faster, so I agree here.

Rakefile
```diff
@@ -314,7 +314,7 @@ task :ci do
   declare '✅ 2.1 / ✅ 2.2 / ✅ 2.3 / ✅ 2.4 / ✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ jruby' => 'bundle exec appraisal contrib rake spec:sequel'
   declare '❌ 2.1 / ✅ 2.2 / ✅ 2.3 / ✅ 2.4 / ✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ jruby' => 'bundle exec appraisal contrib rake spec:shoryuken'
   declare '✅ 2.1 / ✅ 2.2 / ✅ 2.3 / ✅ 2.4 / ✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ jruby' => 'bundle exec appraisal contrib rake spec:sidekiq'
-  declare '✅ 2.1 / ✅ 2.2 / ✅ 2.3 / ✅ 2.4 / ✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ jruby' => 'bundle exec appraisal contrib rake spec:sinatra'
+  declare '✅ 2.1 / ✅ 2.2 / ✅ 2.3 / ✅ 2.4 / ✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ jruby' => 'bundle exec appraisal sinatra rake spec:sinatra'
```
marcotc, line 317: Follow up on https://github.com/DataDog/dd-trace-rb/pull/2360/files#r1018806426, we should remove this line, and keep the one on line 380.

TonyCTHsu:
@ivoanjo I setup an integration application some weeks ago when Sinatra 3 is released. It just worked!

This PR is mainly addressing the fact that we are using Sinatra 3 in our test suite instead of 2.

ivoanjo:
> @ivoanjo I setup an integration application some weeks ago when Sinatra 3 is released. It just worked!
>
> This PR is mainly addressing the fact that we are using Sinatra 3 in our test suite instead of 2.

Awesome, glad to hear it! :)

PR #2362, opened 2022-11-10
Improve clearing of profile data after Ruby app forks

ivoanjo:
**What does this PR do?**:

When a Ruby app forks, we automatically restart the profiler in the child process as well. This ensures that in apps such as forking webservers we still provide full profiling coverage.

As part of automatically restarting the profiler in the child process, we want to clear any profiling data that has accumulated from the parent, as otherwise both the parent as well as the child process would report that data.

This "clearing of profiling data" happens in `Scheduler#after_fork`. Previously, this method would "clear data" by asking the exporter for the latest profile and then throwing this away.

This worked but it was rather heavy handed -- it did some useless work, as well as generated a few slightly-confusing debug logs:

```
DEBUG (dd-trace-rb/lib/datadog/profiling/scheduler.rb:73:in `after_fork') Flushing exporter in child process #after_fork and discarding data
DEBUG (dd-trace-rb/lib/datadog/profiling/stack_recorder.rb:25:in `serialize') Encoded profile covering 2022-11-10T09:42:22Z to 2022-11-10T09:42:22Z
DEBUG (dd-trace-rb/lib/datadog/profiling/exporter.rb:52:in `flush') Skipped exporting profiling events as profile duration is below minimum
```

To improve this for the new Ruby profiler, I've decided to implement a `StackRecorder#clear`, and then refactor the `Scheduler` and `Exporter` to use it instead.

For the old Ruby profiler (still the default, in-use by customers), I chose not to implement a `#clear`, and thus the old approach of serializing a profile and throwing it away is still used.

**Motivation**:

I've wanted to improve this for some time, but chose not to because I didn't want to invest on improving the old profiler just to then throw it away. Now that the new profiler is getting in better shape, I finally decided to implement this.

**Additional Notes**:

This PR is on top of #2359 just for my development convenience; it has no explicit dependencies on it.

**How to test the change?**:

Validate that profiles from the child process of a Ruby app that forks do not contain any profiling data from the parent process by looking at the profiler UX.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/stack_recorder.c
+++ b/ext/ddtrace_profiling_native_extension/stack_recorder.c
@@ -445,6 +445,10 @@ static VALUE _native_clear(DDTRACE_UNUSED VALUE _self, VALUE recorder_instance) 
 
   ddog_Timespec finish_timestamp = time_now();
 
+  // Why flip the slots as part of clearing? This makes clear behave exactly like serialize in terms of concurrency
+  // properties, even though right now there probably won't be any concurrency between calling clear and adding samples
+  // to the profile because we do not release the global VM lock while clearing.
+  // See also https://github.com/DataDog/dd-trace-rb/pull/2362/files#r1019659427 for more details.
   ddog_Profile *profile = serializer_flip_active_and_inactive_slots(state, finish_timestamp);
   if (!ddog_Profile_reset(profile, NULL /* start_time is optional */ )) {
     return rb_ary_new_from_args(2, error_symbol, rb_str_new_cstr("Failed to reset profile"));
--- a/ext/ddtrace_profiling_native_extension/stack_recorder.c
+++ b/ext/ddtrace_profiling_native_extension/stack_recorder.c
@@ -174,6 +174,7 @@ static VALUE _native_is_slot_one_mutex_locked(DDTRACE_UNUSED VALUE _self, VALUE 
 static VALUE _native_is_slot_two_mutex_locked(DDTRACE_UNUSED VALUE _self, VALUE recorder_instance);
 static VALUE test_slot_mutex_state(VALUE recorder_instance, int slot);
 static ddog_Timespec time_now();
+static VALUE _native_clear(DDTRACE_UNUSED VALUE _self, VALUE recorder_instance);
 
 void stack_recorder_init(VALUE profiling_module) {
   stack_recorder_class = rb_define_class_under(profiling_module, "StackRecorder", rb_cObject);
@@ -191,6 +192,7 @@ void stack_recorder_init(VALUE profiling_module) {
   rb_define_alloc_func(stack_recorder_class, _native_new);
 
   rb_define_singleton_method(stack_recorder_class, "_native_serialize",  _native_serialize, 1);
+  rb_define_singleton_method(stack_recorder_class, "_native_clear", _native_clear, 1);
   rb_define_singleton_method(testing_module, "_native_active_slot", _native_active_slot, 1);
   rb_define_singleton_method(testing_module, "_native_slot_one_mutex_locked?", _native_is_slot_one_mutex_locked, 1);
   rb_define_singleton_method(testing_module, "_native_slot_two_mutex_locked?", _native_is_slot_two_mutex_locked, 1);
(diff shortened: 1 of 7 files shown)


spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -39,6 +39,41 @@ def slot_two_mutex_locked?
     end
   end
 
+  shared_examples_for 'locking behavior' do |operation|
+    context 'when slot one was the active slot' do
+      it 'sets slot two as the active slot' do
+        expect { stack_recorder.public_send(operation) }.to change { active_slot }.from(1).to(2)
+      end
+
+      it 'locks the slot one mutex and keeps it locked' do
```
marcotc, line 48: Isn't the part "and keeps it locked" redundant?
```suggestion
      it 'locks the slot one mutex' do
```

marcotc: Actually a better question.
Aren't these 3 conditions `change { active_slot }.from(1).to(2)`, `change { slot_one_mutex_locked? }.from(false).to(true)`, and `change { slot_two_mutex_locked? }.from(true).to(false)` actually representing one single thing: "the slot was flipped"?

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -436,3 +438,17 @@ static ddog_Timespec time_now() {
...
   return (ddog_Timespec) {.seconds = current_time.tv_sec, .nanoseconds = (uint32_t) current_time.tv_nsec};
 }
+
+static VALUE _native_clear(DDTRACE_UNUSED VALUE _self, VALUE recorder_instance) {
+  struct stack_recorder_state *state;
+  TypedData_Get_Struct(recorder_instance, struct stack_recorder_state, &stack_recorder_typed_data, state);
+
+  ddog_Timespec finish_timestamp = time_now();
+
+  ddog_Profile *profile = serializer_flip_active_and_inactive_slots(state, finish_timestamp);
```
marcotc, line 448: Why does the slot need to be flipped here?

ivoanjo: That is a good question. The answer is... they don't strictly need to be right now, but doing so leaves us with some flexibility for the future.

The more specific answer is: the `StackRecorder` is built in such a way that it can serialize (and now clear) a profile even while another thread from the profiler is writing profiling data to the profile.

Thus, "theoretically" the slot flipping is needed to retain this property -- the other slot by definition is kept empty so the other thread can start writing to it even while we're cleaning the previously-active profile (which is the one that will have leftover data after the fork).

BUT in practice right now writing data to the profile requires the global VM lock, and `_native_clear` is only executed with the global VM lock acquired, so right now there can be no concurrency between writing and clearing. So we could just clear the currently-active profile and keep that as the active, rather than doing the flipping. But that would create potential problems when (if?), in the future, we move more of the "writing to profile" process to be executed outside of the GVL.

spec/datadog/profiling/stack_recorder_spec.rb
ivoanjo: > Isn't the part "and keeps it locked" redundant?

Good point! Will simplify.

> Actually a better question.
> Aren't these 3 conditions change { active_slot }.from(1).to(2), change { slot_one_mutex_locked? }.from(false).to(true), and change { slot_two_mutex_locked? }.from(true).to(false) actually representing one single thing: "the slot was flipped"?

Yes; "slot being flipped" means 3 different observable effects:

* `state->active_slot` gets changed
* `slot_one_mutex` gets locked/unlocked
* `slot_two_mutext` gets locked/unlocked

I chose to test the three effects here in a more "white-box way", but indeed this wasn't the only option. For instance, I could add a RSpec helper to represent the three checks, or even move that to the `StackRecorder` class itself (using Ruby or C code).

This spec is effectively asserting on private state of the class, rather than the class abstracting this away and the spec asserting on the higher-leve  abstraction of flipping. BUT, since the class itself is already quite complex (especially being in C), I thought having more of it documented and tested on the Ruby side makes it more understandable, and so worthy of the exception to the usual "don't test private stuff".

ivoanjo: Descriptions improved in 1f2fdf5b8d45fb43dfbd1fff21a66923036a6255

ext/ddtrace_profiling_native_extension/stack_recorder.c
ivoanjo: Added a note about this in 96acfb253f84fc78778f6ea62a26a5a591a6dc17

marcotc: Makes sense, I think the explanation you added in https://github.com/DataDog/dd-trace-rb/commit/96acfb253f84fc78778f6ea62a26a5a591a6dc17 clarifies it a lot! 🙇

marcotc (approved):

PR #2363, opened 2022-11-10
Fix Redis configuration pin

TonyCTHsu:
**What does this PR do?**

Current redis instrumentation is looking for pin on client object instead of the redis instance object. Hence it is different from the behavior the document claimed.

```
require 'redis'
require 'ddtrace'

Datadog.configure do |c|
  c.tracing.instrument :redis # Enabling integration instrumentation is still required
end

customer_cache = Redis.new
invoice_cache = Redis.new

Datadog.configure_onto(customer_cache, service_name: 'customer-cache')
Datadog.configure_onto(invoice_cache, service_name: 'invoice-cache')

# Traced call will belong to `customer-cache` service
customer_cache.get(...)
# Traced call will belong to `invoice-cache` service
invoice_cache.get(...)
```

The PR implements patches provide mean to delegate the pin the be accessible to the client object.


marcotc (approved):

lib/datadog/tracing/contrib/redis/patcher.rb
```diff
@@ -12,6 +12,42 @@ module Redis
...
+            def self.included(base)
+              base.prepend(InstanceMethods)
+            end
+
+            # Instance method patch for redis instance
+            module InstanceMethods
+              def initialize(options = {})
+                options[:redis_instance] = self
+
+                super(options)
```
zenspider, line 26: should prolly be

```ruby
super(options.merge(redis_instance: self))
```

PR #2364, opened 2022-11-10
Redis 6.2 image for CI

TonyCTHsu:
**What does this PR do?**

Use Redis 6.2 image for CI. This would allow us to gradually testing newer gem version of redis and sidekiq 

1. Align the images used for CI and development. 6.2 for new versions of redis and 3.0 for older version.
2. Since `qless` is having issues with newer versions of redis. Establish a new old redis container for it. We should move `qless` to `contrib-old` in future. Because the gem is not actively maintained and has a couple of dependencies that are blocking other gems in the same appraisal group.
3. Fixing failing spec, due to slightly different error message from DB redis.


.env
```diff
@@ -28,3 +28,5 @@ TEST_PRESTO_HOST=localhost
 TEST_PRESTO_PORT=8080
 TEST_REDIS_HOST=127.0.0.1
 TEST_REDIS_PORT=6379
+TEST_REDIS_OLD_HOST=127.0.0.1
+TEST_REDIS_OLD_PORT=6380
```
ivoanjo, line 32: Nice! I was not entirely sure if this was enough to get docker-compose to assign an alternative port when starting the containers but it is, good find! :)

spec/datadog/tracing/contrib/qless/support/job.rb
```diff
@@ -25,8 +25,9 @@ def string
 RSpec.shared_context 'Qless job' do
   include Qless::WorkerHelpers
 
-  let(:host) { ENV.fetch('TEST_REDIS_HOST', '127.0.0.1') }
-  let(:port) { ENV.fetch('TEST_REDIS_PORT', 6379) }
+  let(:host) { ENV.fetch('TEST_REDIS_OLD_HOST', '127.0.0.1') }
+  let(:port) { ENV.fetch('TEST_REDIS_OLD_PORT', 6379) }
```
ivoanjo, line 29: Minor: For consistency, I suggest always making this a string

```suggestion
  let(:port) { ENV.fetch('TEST_REDIS_OLD_PORT', '6379') }
```

docker-compose.yml
```diff
@@ -616,6 +667,12 @@ services:
       - "6379"
     ports:
       - "${TEST_REDIS_PORT}:6379"
+  redis_old:
+    image: redis:3.0
+    expose:
+      - "6379"
+    ports:
+      - "${TEST_REDIS_OLD_PORT}:6379"
```
ivoanjo, line 675: Since we're adding this just for qless, I suggest leaving a comment here stating so -- why this is needed, and that we can remove it if/when we drop support for qless.

ivoanjo (approved):
# Looks good but please remember to squash the history before merging so that a bunch of WIP commits and similar don't end up on master

PR #2367, opened 2022-11-15
[PROF-5943] Clear leftover state on new profiler after Ruby VM forks

ivoanjo:
**What does this PR do?**:

This PR is a follow-up from work started in #2359 and continued in #2362 for fixing and improving support for Ruby VM forks in the new profiler.

When a Ruby application forks (such as the puma webserver in multiprocess mode), we automatically restart the profiler in the child processes.

But, as part of restarting we should also make sure to reset any state that is leftover from the parent process.

This PR makes sure that, for the new profiler:
1. Resetting the state is done in a safe way (e.g. without concurrency issues). This required a refactoring of how we do it.
2. There is no leftover state from the parent.

While the old profiler codepath was also affected by the refactoring in this PR, it was already doing the two steps above correctly; the changes in this PR are just to make it easier to support the feature in the new profiler.

**Motivation**:

It's common for Ruby apps to make use of fork, and this is a configuration we expect to continue to support in the new Ruby profiler.

**Additional Notes**:

This PR is easier to review commit-by-commit.

**How to test the change?**:

Besides the included test coverage, this can be manually seen in action by validating that the profiler (both on the old codepath and [the new codepath](https://github.com/DataDog/dd-trace-rb/pull/2209)) restarts correctly on child processes after a fork and that profiles reported by those child processes are correct and do not contain left-over state from the parent.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
@@ -91,7 +91,7 @@ struct cpu_and_wall_time_collector_state {
   // is not (just) a stat.
   unsigned int sample_count;
 
-  struct {
+  struct stats {
     // Track how many garbage collection samples we've taken.
     unsigned int gc_samples;
     // See cpu_and_wall_time_collector_on_gc_start for details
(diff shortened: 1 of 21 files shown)


ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
```diff
@@ -91,7 +91,7 @@ struct cpu_and_wall_time_collector_state {
   // is not (just) a stat.
   unsigned int sample_count;
 
-  struct {
+  struct stats {
```
ivoanjo, line 94: This struct was previously anonymous, but I needed to name it so I could reference it from `_native_reset_after_fork`

marcotc (approved):

PR #718, opened 2019-03-13
Initial integration of Roda.

randy-girard:
I have a need to instrument Roda framework, so I took at stab at it locally and just ported it into the Gem. Feel free to rip it apart and fix it up for release :)

PR #2368, opened 2022-11-15
Roda Integration

wantsui:
**What does this PR do?**

Continues on the great work started by @randy-girard  in https://github.com/DataDog/dd-trace-rb/pull/718 by working on a Roda Integration that works with **dd-trace-rb 1.x .**

**Code to trigger automatic instrumentation**

```
require "roda"
require "ddtrace"
class SampleApp < Roda
  use Datadog::Tracing::Contrib::Rack::TraceMiddleware
  Datadog.configure do |c|
    c.tracing.instrument :roda, **options
  end
  route do |r|
    r.root do
      r.get do
        'Hello World!'
      end
    end
  end
end
```

**Available Configuration Options**
- enabled
- service_name

**Example trace:**

RodaTrace

**Motivation**

Add support for Roda

**Additional Notes**

Addressed Notes:
1) About error statuses:
- While testing this, it was found that if an application leads to a 500 internal server error due to a thrown error, Roda doesn't have access to the status code. To account for this, I updated this to default to a 500 status code.
2) What version(s) should be compatible? 
- To be safe, support is up until 3.x. Whenever 4.x comes out, this integration will need to be revisited.
3) About Rack:
It is highly recommended to load in Rack through `use Datadog::Tracing::Contrib::Rack::TraceMiddleware` because Rack handles the context propagation for distributed tracing.

**How to test the change?**

See Checks ran.


lib/datadog/tracing/contrib/roda/instrumentation.rb
```diff
@@ -0,0 +1,75 @@
...
+              Contrib::Analytics.set_measured(span)
+
+              ensure
+                response = yield
+              end
+              
+              status_code = response[0]
+
+              # Adds status code to the resource name once the resource comes back
+              span.resource = "#{request_method} #{status_code}"
```
marcotc, line 54: For unexpected errors (e.g. `raise 'oops'`), this part of the code here is likely unreachable.
We should try to match the response code that Roda returns to the HTTP client in that scenario.

lib/datadog/tracing/contrib/roda/instrumentation.rb
```diff
@@ -0,0 +1,84 @@
...
+      module Roda
+        # Instrumentation for Roda
+        module Instrumentation
+
+          def _roda_handle_main_route
+            instrument(Ext::SPAN_REQUEST) { super }
+          end
+
+          def call
+            instrument(Ext::SPAN_REQUEST) { super }
```
TonyCTHsu, line 17: Can move the [error handling logic](https://github.com/DataDog/dd-trace-rb/pull/2368/files#diff-882be2b4ec172fe41860d166e42fd865e3b6610a7ee43d0e4b382a03033ea0cbR50-R57) to the caller

```
          def call
            instrument(Ext::SPAN_REQUEST) do |span|
              super
            rescue => e
              # ... Handle error
            end
          end
```
So you could be confident yielding in the block without nested begin/rescue.
```
response = yield span
```

marcotc (commented):
Looks pretty good! 👍

lib/datadog/tracing/contrib/roda/instrumentation.rb
wantsui: It looks like in the Roda code, dispatch can land on one of these methods ([seen in this conditional](https://github.com/jeremyevans/roda/blob/04f7e70215b1ec1151286095b8825626335d5604/lib/roda.rb#L456))
- [call](https://github.com/jeremyevans/roda/blob/04f7e70215b1ec1151286095b8825626335d5604/lib/roda.rb#L522)
- [_roda_handle_main_route ](https://github.com/jeremyevans/roda/blob/04f7e70215b1ec1151286095b8825626335d5604/lib/roda.rb#L493)

Would I need to duplicate the rescue block and handle the error in both since either one of these can handle the dispatch?

lib/datadog/tracing/contrib/roda/instrumentation.rb
```diff
@@ -0,0 +1,79 @@
...
+    module Contrib
+      module Roda
+        # Instrumentation for Roda
+        module Instrumentation
+
+          def _roda_handle_main_route
+            instrument do
+              super
+            rescue => e
+              ['500'] # [status, headers, body]
```
wantsui, line 15: After testing this iteration out on an endpoint that should have thrown `undefined local variable or method`, it looks like server errors now throw an `undefined method `each' for nil:NilClass` for all the 5xx type errors that I tried out.

If I comment out `ddtrace`, the web server returns back `undefined local variable or method`.

Any concerns if I revert the rescue block back into the `instrument` method so it continues pass the status/headers/body without being affected by ddtrace? 

https://github.com/DataDog/dd-trace-rb/blob/aa1db9727d2b048865e47f505db306c5872c9ead/lib/datadog/tracing/contrib/roda/instrumentation.rb#L50-L66

marcotc: The new changes are incorrect for error handling: the error is now swallowed by `lib/datadog/tracing/contrib/roda/instrumentation.rb`.

Please revert back to the previous iteration.

wantsui: Thanks! This definitely makes sense, I'll revert it back!

Appraisals
```diff
@@ -131,6 +131,7 @@ if ruby_version?('2.1')
     gem 'rake', '< 12.3'
     gem 'rest-client'
     gem 'resque', '< 2.0'
+    gem 'roda', '>= 2.0.0', '< 4.0.0'
```
marcotc, line 134: Any particular reason to lock this to `< 4.0.0`.
We don't automatically upgrade gem versions: it has to be manually done. This means someone will supervise any upgrades.
We normally leave gems unlocked, so we can upgrade the easily in the future.

wantsui: I wasn't sure if this instrumentation would work in 4.x so I set a maximum version for now, to avoid a scenario where we say we support 2.0.0+, only for 4.x to have a breaking change.

If this isn't needed at the Appraisal level, I can definitely remove it!

marcotc:
Let also run `bundle exec rake install_appraisal_gemfiles` after the `Appraisals` changes, and commit the results (there will be changes to `gemfiles/*` files). This way we make sure we are always running with the same version of the Roda gem.

Appraisals
wantsui: I removed the "maximum version" from the latest Appraisal file and ran `bundle exec rake install_appraisal_gemfiles`, but it looks like it changed a lot of files in the gemfiles folder.

lib/datadog/tracing/contrib/roda/configuration/settings.rb
```diff
@@ -0,0 +1,35 @@
...
+              o.default { env_to_bool(Ext::ENV_ANALYTICS_ENABLED, false) }
+              o.lazy
+            end
+
+            option :analytics_sample_rate do |o|
+              o.default { env_to_float(Ext::ENV_ANALYTICS_SAMPLE_RATE, 1.0) }
+              o.lazy
+            end
+
+            option :distributed_tracing, default: true
```
delner, line 28: Can Roda be the root span in a trace? Or would it always be under Rack?

lib/datadog/tracing/contrib/roda/instrumentation.rb
```diff
@@ -0,0 +1,84 @@
...
+                if Contrib::Analytics.enabled?(configuration[:analytics_enabled])
+                  Contrib::Analytics.set_sample_rate(span, configuration[:analytics_sample_rate])
+                end
+
+                # Measure service stats
+                Contrib::Analytics.set_measured(span)
+              ensure
+                begin
+                  response = yield
+                rescue
```
delner, line 50: Generally recommend classifying what kind of error is caught: `StandardError` if you don't want to trap kill signals, `Exception` if you want all signals (have be to very careful when catching `Exception`). `StandardError` is likely correct here.

To clarify, the purpose here is to add additional tags in case of an error? I think it might be better to provide an `on_error` argument to the `#trace` call, as it will handle the exception and what not for you. this way you can eliminate the `begin/rescue` in the `ensure`. Bonus is that it will handle `Exception` for you.

lib/datadog/tracing/contrib/roda/instrumentation.rb
```diff
@@ -0,0 +1,84 @@
...
+                  # The status code is unknown to Roda and decided by the upstream web runner.
+                  # In this case, spans default to status code 500 rather than a blank status code.
+                  default_error_status = '500'
+                  span.resource = "#{request_method} #{default_error_status}"
```
delner, line 54: Possibly strange edge case here where `request.request_method.to_s.upcase` raised an error, hit the `ensure` block, request raised an error, caught by this `rescue` block, then `request_method` will likely be undefined, possibly raising another error.

lib/datadog/tracing/contrib/roda/integration.rb
```diff
@@ -0,0 +1,45 @@
...
+module Datadog
+  module Tracing
+    module Contrib
+      module Roda
+        # Description of Roda integration
+        class Integration
+          include Contrib::Integration
+
+          MINIMUM_VERSION = Gem::Version.new('2.0.0')
+          MAXIMUM_VERSION = Gem::Version.new('4.0.0')
```
delner, line 18: Good to know this maximum. Do we know we don't support Roda 4.0.0? If so, why don't we?

Also "maximum" version as a name is slightly misleading in that its exclusive (we don't actually support 4.0.0.)

docs/GettingStarted.md
```diff
@@ -511,6 +512,7 @@ For a list of available integrations, and their configuration options, please re
 | Redis                      | `redis`                    | `>= 3.2`                 | `>= 3.2`                 | *[Link](#redis)*                    | *[Link](https://github.com/redis/redis-rb)*                                    |
 | Resque                     | `resque`                   | `>= 1.0`                 | `>= 1.0`                  | *[Link](#resque)*                   | *[Link](https://github.com/resque/resque)*                                     |
 | Rest Client                | `rest-client`              | `>= 1.8`                 | `>= 1.8`                  | *[Link](#rest-client)*              | *[Link](https://github.com/rest-client/rest-client)*                           |
+| Roda                       | `roda`                     | `>= 2.1`                 | `>= 2.1`                  | *[Link](#roda)*                     | *[Link](https://github.com/jeremyevans/roda)*                                  |
```
delner, line 515: If there's an upper limit on versions we support, we should add that here as well.

delner (commented):
Looking clean, well organized as I'd expect a good integration to be!

Some minor tweaks on edge cases involving errors/versions worth examining. Please see comments for more details.

Where relevant or useful, I would suggest also adding code comment documentation as it might help an unfamiliar developer wrap their head around instrumentation behavior.

lib/datadog/tracing/contrib/roda/configuration/settings.rb
delner: I think I answered my own question looking at the example trace you linked in the PR description. Seems like it can be a root span.

Different question, could it NOT be a root span?

lib/datadog/tracing/contrib/roda/integration.rb
wantsui: At the time of this comment: Roda 4.x isn't out. The latest version is 3.65.0. 

Given the changes that occurred between 1.x and 2.x, I'm just being cautious by setting a maximum of 4.x for now. That way, in our public docs, we can be clear that this logic is intended for 2.x and 3.x. 

Whenever 4.x comes out, it'll require visiting to make sure it still functions as expected.

docs/GettingStarted.md
wantsui: Great catch! I was being cautious by hardcoding a limit of 4.x, so I should update this Getting Started doc too!

lib/datadog/tracing/contrib/roda/configuration/settings.rb
wantsui: Based on how Rack can be used with Roda, it certainly seems possible that Rack could be considered a parent of Roda in this case. I tested this locally with this PR and ended up with two separate traces: `rack.request` and another one `roda.request`. Would the expectation that instead of two distinct requests, that there just be one trace?

wantsui: After going over this with @marcotc , the `set_distributed_tracing_context!` that I currently have led to a scenario where if the `rack` integration and `roda` integration gets activated, it led to separate traces.

TODO: I'm going to remove this handling of http headers at the `roda` level and write a note to activate the `rack` integration if distributed tracing is expected.

docs/GettingStarted.md
wantsui: In the latest commit, this now gives a range ending <4 (since it works with 3.x):
```
| Roda                       | `roda`                     | `>= 2.1, <4`             | `>= 2.1, <4`              | *[Link](#roda)*                     | *[Link](https://github.com/jeremyevans/roda)*   
```

lib/datadog/tracing/contrib/roda/instrumentation.rb
wantsui: I'm not sure how this error can be thrown because all routes should be defined with a type from the start. Do you have a suggestion for handling this scenario?

marcotc: I paired with Wan and it's not possible for `request.request_method.to_s.upcase` to fail, as it comes straight from Rack and it's part of the Rack contract. This is the same behaviour of our other Rack-ralated integrations (Sinatra, Rails, Hanami, Grace)

lib/datadog/tracing/contrib/roda/instrumentation.rb
marcotc: I paired with Wan on this one as well, and the `on_error` would not fit well for this case because we need the local closure variable `request_method` in the error handler.

If we extract the error handling to `on_error: method(:handle_error)`, then `def handle_error` does not have access to `request_method`.
Alternatively, we could declare it inlince `on_error: ->{...}` but the method is pretty large.

I think we can keep it as in this file, as the alternatives are less legible.

marcotc: I paired with Wan on this one as well, and the `on_error` would not fit well for this case because we need the local closure variable `request_method` in the error handler.

If we extract the error handling to `on_error: method(:handle_error)`, then `def handle_error` does not have access to `request_method`.
Alternatively, we could declare it inlince `on_error: ->{...}` but the method is pretty large.

I think we can keep it as in this file, as the alternatives are less legible.

lib/datadog/tracing/contrib/roda/patcher.rb
```diff
@@ -0,0 +1,31 @@
...
+require_relative 'instrumentation'
+
+module Datadog
+  module Tracing
+    module Contrib
+      module Roda
+        # Patcher enables patching of 'roda'
+        module Patcher
+          include Contrib::Patcher
+          ::Roda.use Datadog::Tracing::Contrib::Rack::TraceMiddleware if defined? ::Roda
```
marcotc, line 16: We should remove this top-level call and instead, due to lack of a good way to automatically instrument Roda+Rack, we should add this manual step to the readme.

lib/datadog/tracing/contrib/roda/instrumentation.rb
wantsui: As an update to this, I wasn't able to move the error handling logic anywhere else because it interfered with the errors produced by the underlying app.

Appraisals
wantsui: I just ran `bundle exec rake install_appraisal_gemfiles` today and it didn't change any of the gemfiles so it looks like the gemfiles are already up to date.

marcotc (approved):
Thank you so much, @wantsui! 🙇

Issue #2369, opened 2022-11-16
FrozenError from Redis patcher #initialize


KJTsanaktsidis:
**Current behaviour**
When instantiating a Redis client using the current master branch of dd-trace-rb, and passing in a frozen options hash, an exception is raised:

```
FrozenError: can't modify frozen Hash: {:url=>"redis://localhost:6379", :timeout=>1.0}
    /home/runner/vendor/bundle-52/ruby/2.7.0/bundler/gems/dd-trace-rb-f387e1747a44/lib/datadog/tracing/contrib/redis/patcher.rb:24:in `initialize'
```

**Expected behaviour**
It should not raise an exception :) This was broken a few days ago in https://github.com/DataDog/dd-trace-rb/pull/2363 I believe

**Steps to reproduce**

`Redis.new({}.freeze)` should do it I think?

**How does `ddtrace` help you?**

It rocks! Really looking forward to rolling out the new native profiler implementation.

**Environment**

* **ddtrace version: Git master, ref `f387e1747a44` **
* **Configuration block doubt this is relevant?:**
* **Ruby version: 2.7.6**
* **Operating system: Ubuntu 18.04**
* **Relevant library versions: N/A**

TonyCTHsu:
@KJTsanaktsidis 👋  I just released `1.6.1`, this would fix the frozen input. Will be working on redis 5 instrumentation next!

PR #2370, opened 2022-11-16
Fix frozen input for Redis 

TonyCTHsu:
https://github.com/DataDog/dd-trace-rb/issues/2369

**What does this PR do?**

FrozenError, when user provide a frozen hash to Redis.

Instead of mutating the input hash, use `Hash#merge` to return a new hash.


lib/datadog/tracing/contrib/redis/patcher.rb
```diff
@@ -21,9 +21,7 @@ def self.included(base)
             # Instance method patch for redis instance
             module InstanceMethods
               def initialize(options = {})
-                options[:redis_instance] = self
-
-                super(options)
+                super(options.merge(redis_instance: self))
               end
```
ivoanjo, line 25: Minor: It may be worth calling out specifically in a comment that `options` can be frozen and that this has been an issue for customers in the past. (So that the next refactoring we do here doesn't accidentally reintroduce the issue)

ivoanjo (approved):
👍 LGTM

marcotc (approved):

PR #2378, opened 2022-11-17
[PROF-6566] Fix ddtrace installation issue when users have CI=true

### LINKED ISSUE #2377: Installation failing on 1.6.0 and 1.6.1 in CI/GitHub Actions due to "error: missing initializer for field" compiler error
**Current behaviour**

Reported internally by the https://github.com/DataDog/datadog-api-client-ruby team. Their CI started failing with compilation errors:

* `collectors_stack.c:151:7: error: missing initializer for field ‘start_line’ of   ‘ddog_Function’ {aka ‘struct ddog_Function’}`
* `collectors_stack.c:307:5: error: missing initializer for field ‘start_line’ of
  ‘ddog_Function’ {aka ‘struct ddog_Function’}`
* `collectors_stack.c:344:5: error: missing initializer for field ‘start_line’ of
  ‘ddog_Function’ {aka ‘struct ddog_Function’}`
* `collectors_stack.c:376:5: error: missing initializer for field ‘is_folded’ of
  ‘ddog_Location’ {aka ‘struct ddog_Location’}`

This is caused by a combination of https://github.com/DataDog/dd-trace-rb/pull/2358 triggering erroneously (e.g. outside of ddtrace CI) and compiler warnings which we hadn't previously seen.

Specifically, installing ddtrace with the `CI=true` environment variable set will fail if the C compiler emits a warning. This is the case with `gcc version 9.4.0` (Ubuntu 20.04 default).

Issue can be worked around by defining `DD_PROFILING_NO_EXTENSION` environment variable to `true` prior to running `bundle install`. See https://github.com/DataDog/datadog-api-client-ruby/pull/1137 for an example.

**Expected behaviour**

A compiler warning should not cause a user of ddtrace to see an installation error. No workaround should be needed.

**Steps to reproduce**

(Still working on this)

**Environment**

* **ddtrace version:** 1.6.0 and 1.6.1
* **Configuration block (`Datadog.configure ...`):** (Not relevant)
* **Ruby version:** ruby 2.7.6p219 (2022-04-12 revision c9c2245c0a) [x86_64-linux]
* **Operating system:** Linux (GitHub Actions)
* **Relevant library versions:** (None)

ivoanjo:
**What does this PR do?**:

In #2358 we changed how the profiling native extension is built in CI by adding the `-Werror` compiler option to turn warnings into errors.

We did this automatically by checking if the `CI` environment variable was set to `true`.

Unfortunately, this broke at least one user, see <https://github.com/DataDog/datadog-api-client-ruby/pull/1137>. The warning that was triggered in that case was actually not relevant, see my investigation in <https://github.com/DataDog/dd-trace-rb/issues/2377>.

The `CI=true` was never meant to trigger outside of our CI, but clearly it's too generic of a name for us to rely on that.

To fix this, I've gone ahead and changed the configuration so that `-Werror` only gets added when `DD_PROFILING_CI=true`. I also changed our docker and CircleCI configurations to set this by default.

**Motivation**:

This issue affected one Datadog internal customer -- <https://github.com/DataDog/datadog-api-client-ruby/pull/1137>.

This should never happen -- we don't want any issues when compiling the profiling bits to affect ddtrace installation.

**Additional Notes**:

(Nothing)

**How to test the change?**:

Validate that `-Werror` only gets added when `DD_PROFILING_CI=true`.

To trigger a warning for testing, try removing `DDTRACE_UNUSED` from function argument declarations, and you should see that the compiler starts failing due to that.

---

Fixes #2377

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,7 +16,7 @@ services:
       - redis
       - redis_old
     env_file: ./.env
-    environment:
+    environment: &common-environment
       - BUNDLE_GEMFILE=/app/Gemfile
       - DD_AGENT_HOST=ddagent
       - TEST_DATADOG_INTEGRATION=1
@@ -52,20 +52,7 @@ services:
       - redis
       - redis_old
     env_file: ./.env
-    environment:
-      - BUNDLE_GEMFILE=/app/Gemfile
-      - DD_AGENT_HOST=ddagent
-      - TEST_DATADOG_INTEGRATION=1
-      - TEST_ELASTICSEARCH_HOST=elasticsearch
-      - TEST_MEMCACHED_HOST=memcached
-      - TEST_MONGODB_HOST=mongodb
-      - TEST_MYSQL_HOST=mysql
-      - TEST_POSTGRES_HOST=postgres
-      - TEST_PRESTO_HOST=presto
-      - TEST_PRESTO_PORT=8080
-      - TEST_REDIS_HOST=redis
-      - TEST_REDIS_OLD_HOST=redis_old
-      - TEST_REDIS_OLD_PORT=6379
+    environment: *common-environment
     stdin_open: true
     tty: true
     volumes:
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -16,7 +16,7 @@ services:
       - redis
       - redis_old
     env_file: ./.env
-    environment:
+    environment: &common-environment
```
marcotc, line 19: 🧹

marcotc (approved):
Minor: I think a gem-global, `DD_CI` would probably suffice. Maybe `DD_CI` is too short and prone to being used by clients as well (e.g. Dunkin' Donuts' CI pipeline? 🍩), so maybe `DDTRACE_CI` (but how about Discount Drugmart's own tracing CI tests 🙃)?

Regardless, I'm happy to keep `DD_PROFILING_CI` as well, no biggie.

marcotc:
OK I HAVE IT: "DD_⚙️" or "📀🐶🧪"

marcotc (approved):
Sad approval without an emoji variable.

ben:
I'll add myself to the ☝🏼 list of people who ended up here because their CI can't do a `bundle install` because of this. Looks like 1.6.1 shipped 12 days ago, and this merged a day later, but there's no 1.6.2 yet. Do you know an ETA when a release will be cut with this fix?

ivoanjo:
@ben thanks for the ping! The fix will be released this week, and sorry again for the issue.

ivoanjo:
Thanks for the patience, everyone! The fix has now been released as part of [version 1.7.0](https://github.com/DataDog/dd-trace-rb/releases/tag/v1.7.0).


As always, let us know how we can help and keep the feedback coming! 😄

PR #2383, opened 2022-11-17
Move `qless` to `contrib-old`

TonyCTHsu:
**What does this PR do?**

Move `qless` to `contrib-old`, preparing for deprecation


Rakefile
```diff
@@ -301,7 +301,6 @@ task :ci do
   declare '✅ 2.1 / ✅ 2.2 / ✅ 2.3 / ✅ 2.4 / ✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ❌ jruby' => 'bundle exec appraisal contrib rake spec:mysql2' # disabled on JRuby, built-in jdbc is used instead
   declare '✅ 2.1 / ✅ 2.2 / ✅ 2.3 / ✅ 2.4 / ✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ❌ jruby' => 'bundle exec appraisal contrib rake spec:pg'
   declare '✅ 2.1 / ✅ 2.2 / ❌ 2.3 / ❌ 2.4 / ❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ❌ 3.0 / ❌ 3.1 / ❌ 3.2 / ❌ jruby' => 'bundle exec appraisal contrib rake spec:presto'
-  declare '❌ 2.1 / ✅ 2.2 / ❌ 2.3 / ❌ 2.4 / ✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ jruby' => 'bundle exec appraisal contrib rake spec:qless'
```
marcotc, line 304: I see support for Ruby 2.2 has been removed, which seems ok given there was no support for 2.3 and 2.4, so 2.2 support was probably a "lucky fluke".

One thing I noticed is that `gemfiles/*` for Ruby 2.2 haven't changed: wouldn't we want to remove `qless` from there, given we don't need it anymore?

TonyCTHsu: @marcotc Yes, so, I currently could not find a way to remove it from the lock file without updating all other dependencies. This could cause a lot more failing tests. I think I will address this when fixing our stale edge build. What do you think?

marcotc (approved):

PR #2384, opened 2022-11-18
[PROF-6463] Implement support for endpoint profiling in CpuAndWallTime collector

ivoanjo:
**What does this PR do?**:

This PR reimplements support for the profiling endpoint feature in the `CpuAndWallTime` collector.

Although the implementation is different, the approach is conceptually the same as provided by `Datadog::Profiling::TraceIdentifiers::Ddtrace` for the old profiler codepaths.

The constraints of this implementation are the same as the ones in #2342 and effectively this is only a small extension of the work done in that PR.

**Motivation**:

Endpoint profiling allows customers to filter down their profiles to individual endpoints (for web apps) that they may be interested in investigating.

**Additional Notes**:

Comparing a Ruby string with a `char *` is really annoying.

**How to test the change?**:

Beyond the code coverage included by the change, it can also be tested by enabling profiling and tracing for a Ruby web app, and then checking that the application's endpoints show up in the right sidebar in the profiling UX.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
@@ -66,10 +66,12 @@
 #define IS_NOT_WALL_TIME false
 #define MISSING_TRACER_CONTEXT_KEY 0
 
-static ID at_active_trace_id; // id of :@active_trace in Ruby
-static ID at_root_span_id;    // id of :@root_span in Ruby
 static ID at_active_span_id;  // id of :@active_span in Ruby
+static ID at_active_trace_id; // id of :@active_trace in Ruby
 static ID at_id_id;           // id of :@id in Ruby
+static ID at_resource_id;     // id of :@resource in Ruby
+static ID at_root_span_id;    // id of :@root_span in Ruby
+static ID at_type_id;         // id of :@type in Ruby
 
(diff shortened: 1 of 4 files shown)


ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
```diff
@@ -572,12 +578,29 @@ static void trigger_sample_for_thread(
...
     labels[label_pos++] = (ddog_Label) {.key = DDOG_CHARSLICE_C("local root span id"), .str = trace_identifiers_result.local_root_span_id};
     labels[label_pos++] = (ddog_Label) {.key = DDOG_CHARSLICE_C("span id"), .str = trace_identifiers_result.span_id};
+
+    if (trace_identifiers_result.trace_endpoint != Qnil) {
+      // The endpoint gets recorded in a different way because it is mutable in the tracer and can change during a
+      // trace.
+      //
+      // Instead of each sample for the same local_root_span_id getting a potentially-different endpoint,
+      // `record_endpoint` (via libdatadog) keeps a list of local_root_span_id values and their most-recently-seen
+      // endpoint values, and at serialization time the most-recently-seen endpoint is applied to all relevant samples.
```
marcotc, line 594: > most-recently-seen endpoint is applied to all relevant samples

This is a very reasonable approach! 👍 👍 👍

marcotc (approved):

PR #2387, opened 2022-11-18
Support sidekiq 7

danhodge:
**What does this PR do?**
* Updates the `contrib` appraisal group to use the latest version of the `sidekiq` gem (v7)
* Updates specs to work with the `sidekiq` version 7 API
* Updates the tracer code to work with `sidekiq` version 7 while continuing to work with prior supported versions

### DIFF
--- a/lib/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat.rb
+++ b/lib/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat.rb
@@ -44,6 +44,11 @@ module Datadog
                 super
               end
             end
+
+            # Sidekiq v7.0.1 changes the functionality of the Sidekiq::Launcher#heartbeat method
+            # and introduces the Sidekiq::Launcher#beat method, which does the same thing as the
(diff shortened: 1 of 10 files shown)


lib/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat.rb
```diff
@@ -44,6 +44,11 @@ def heartbeat
                 super
               end
             end
+
+            # Sidekiq v7.0.1 changes the functionality of the Sidekiq::Launcher#heartbeat method
+            # and introduces the Sidekiq::Launcher#beat method, which does the same thing as the
+            # Sidekiq::Launcher#heartbeat method in versions of Sidekiq prior to v7.0.1
+            alias beat heartbeat
```
marcotc, line 51: I recommend keeping the main code closer to the latest library version, and add an alias for the "legacy" versions.
In this case, rename the `def heartbeat` method above to `beat`, and change the alias to `alias heartbeat beat`.

TonyCTHsu:
Closing this and keep track on https://github.com/DataDog/dd-trace-rb/pull/2810

PR #2391, opened 2022-11-22
[PROF-6556] Profiler signal handler setup improvements

ivoanjo:
**What does this PR do?**:

This PR:
* Refactors the signal handler installation/removal/blocking code from the `CpuAndWallTimeWorker` implementation to a separate file
* Modifies the `install_sigprof_signal_handler` to receive the name of the handler being setup, so that exception messages can also include this information.

**Motivation**:

I'm working on two other signal-related changes, and extracted these commits from my branch to enable smaller reviews :)

**Additional Notes**:

The signal handler code has test coverage via the `CpuAndWallTimeWorker` specs. I decided to leave the tests there for now, rather than separating them out.

This is because some of these low-level behaviors are hard to test in isolation.

**How to test the change?**:

Existing tests already cover this change.


ivoanjo:
(Note: The broken system tests are unrelated to this PR, they're broken due their base images having issues)

marcotc (approved):

PR #2395, opened 2022-11-22
Fix CircleCI JUnit test result uploading

marcotc:
Two distinct rspec configuration sources `SPEC_OPTS` and `.rspec-local` were in conflict when configuring the test output format.

This PR merged them into the `.rspec-local` file.
Not JUnit format results are correctly written to disk and uploaded to CircleCI. [The `Tests` tab is back](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/7958/workflows/161b2770-9cce-4617-95e0-eb327ce957cd/jobs/295447/tests):
Screen Shot 2022-11-22 at 3 09 16 PM

Also, our [CircleCI Tests Insights](https://app.circleci.com/insights/github/DataDog/dd-trace-rb/workflows/build-and-test/tests?branch=master) will populate again in a few days.


.circleci/config.yml
```diff
@@ -168,11 +166,15 @@ step_run_all_tests: &step_run_all_tests
   run:
     name: Run tests
     command: |
+        # Ensures it's possible to debug hung tests in CI
+        echo "--format=documentation" > .rspec-local
```
ivoanjo, line 170: Minor: Maybe use `>>` here as well, so that it doesn't accidentally clobber other configurations in the future?

```suggestion
        # Ensures it's possible to debug hung tests in CI
        echo "--format=documentation" >> .rspec-local
```

ivoanjo (approved):
👍 Thanks for always being our make-CI-better wizard :)

.circleci/config.yml
marcotc: I left it as `>` on purpose, to ensure it was a clean file, but I see the opposite use case as well.

ivoanjo: That's a reasonable point.

I guess we could add a failure if the file exists (e.g. something like `cat .rspec-local && echo "File not expected to exist" && exit 1`) and then we'd get a best of both worlds, although in this tiny case it may not be worth the trouble :)

PR #2409, opened 2022-11-23
Use Ruby 2.6.10 in CircleCI

lloeki:
Depends on https://github.com/DataDog/dd-trace-rb/pull/2402


marcotc (commented):
Do we need to update the docket-compose file was well?

lloeki:
@marcotc no, it has already been updated in https://github.com/DataDog/dd-trace-rb/pull/2402

marcotc (approved):

PR #2413, opened 2022-11-23
Adds `span.kind` tag for `racecar`

zhamza99:
**What does this PR do?**
Adds `span.kind` tag for integrations `racecar`

Refer to [Opentelemetry](https://github.com/open-telemetry/opentelemetry-specification/blob/9fa7c656b26647b27e485a6af7e38dc716eba98a/specification/trace/api.md#spankind) for definitions of consumer

Racecar is only a kafka consumer library, so not possible to tag any spans created with `producer` from this integration

Note that there are changes made to the Appraisals file because the racecar installation for certain ruby versions was disabled. However, it seems the [issue](https://github.com/appsignal/rdkafka-ruby/pull/144) that was causing them to be disabled has been closed.


Appraisals
```diff
@@ -1452,7 +1452,7 @@ elsif ruby_version?('3.0') || ruby_version?('3.1')
     gem 'mongo', '>= 2.8.0', '< 2.15.0' # TODO: FIX TEST BREAKAGES ON >= 2.15 https://github.com/DataDog/dd-trace-rb/issues/1596
     gem 'mysql2', '>= 0.5.3', platform: :ruby
     gem 'pg', '>= 1.1', platform: :ruby
-    # gem 'racecar', '>= 0.3.5' # Pending release of our fix: https://github.com/appsignal/rdkafka-ruby/pull/144
+    gem 'racecar', '>= 0.3.5'
```
TonyCTHsu, line 1455: 1. Installing `racecar` in appraisal group should eventually ends up changes at the corresponding lock file.  I suggested this to be done after our [current progress of changing images](https://github.com/DataDog/dd-trace-rb/pull/2407). 
2. Should also enable CI with changes in `Rakefile`

TonyCTHsu (changes requested):

Appraisals
zhamza99: 1. Merged main branch which already has 3.0.4 changes
2. Rakefile changes implemented to include all version 3.0 and above

marcotc (approved):
Thank you for restoring `racecar` tests as well 🙏

PR #2140, opened 2022-07-08
[PROF-5774] Remove support for profiling Ruby 2.1

ivoanjo:
**What does this PR do?**

This PR remove support for profiling Ruby 2.1 using the Continuous Profiling product. Other products shipped in ddtrace (tracing, application security monitoring, CI, etc) are unaffected.

This dropping of support does not impact our customer's ability to install the latest dd-trace-rb on Ruby 2.1. What it does is

1. Skip compilation of the profiling native extension on Ruby 2.1
2. Print a warning message when customers try to enable profiling on Ruby 2.1 (but does not otherwise block or break their application)

The warning message shown to customers is:

> W, [2022-07-08T14:08:51.295193 #51899]  WARN -- ddtrace: [ddtrace] Profiling was requested but is not supported, profiling disabled: Your ddtrace installation is missing support for the Continuous Profiler because
> ~~your Ruby version (2.1) is too old to be supported~~ the profiler only supports Ruby 2.2 or newer.
> Upgrade to a modern Ruby to enable profiling for your app.

**Motivation**

There is little customer interest on profiling Ruby 2.1, and supporting old Rubies also comes with an extra tax, especially as we build towards the new native profiler.

**Additional Notes**

**I'm opening this PR as a draft not because it's missing anything, but because it MUST only be merged after dd-trace-rb 1.2.0 is released**.

**How to test the change?**

* Run test suite on Ruby 2.1
* Try enabling profiling on a Ruby 2.1 application

PR #2415, opened 2022-11-23
[PROF-6556] Fix profiler incorrectly detecting if a thread is holding the Global VM Lock

ivoanjo:
**What does this PR do?**:

While working on improving the way we trigger profiling samples on the new profiler, I discovered that the Ruby VM API `ruby_thread_has_gvl_p()` that we were using to detect if our signal handler had "landed" on the right thread was actually giving incorrect results.

Specifically, this API can still return `true` even when a thread DOES NOT HAVE the global VM lock. 

In particular, looking at the implementation, that API assumes that if a thread is not in a "blocking region" then it will have the GVL which is probably true for the situations that API was designed to be called from BUT this assumption does not hold true when calling `ruby_thread_has_gvl_p` from a signal handler. (Because the thread may have lost the GVL due to a scheduler decision, not because it decided to block.)

To fix this, this PR introduces our own implementation -- `is_current_thread_holding_the_gvl`, and switches over to using it.

Unfortunately, implementing `is_current_thread_holding_the_gvl` is non-trivial, including making it work for all Ruby versions we want to support. So this is why the diff on this PR is bigger than I would've liked.

**Motivation**:

There's a bigger comment explaining why we need to know if our signal handler is executing on the thread with the GVL inside the `collectors_cpu_and_wall_time_worker.c` but the TL;DR is that we need to call a Ruby API -- `rb_postponed_job_register_one` which from my analysis is not safe to call from any thread other than the one holding the GVL.

**Additional Notes**:

While implementing `is_current_thread_holding_the_gvl` for Ruby < 2.6 I ran into an issue: we need to information from two different memory locations and there's no guarantee that we don't observe an intermediate inconsistent state.

I've documented in the code the issue, as well as added a warning that will be printed to users when they try to use the new Ruby profiler on Ruby < 2.6.

I also plan to add this documentation to a more visible place once we are preparing to release the new profiler out as beta (and then as GA).

One benefit that `is_current_thread_holding_the_gvl` provides is that previously the new profiler could not be used on Ruby 2.2 because `ruby_thread_has_gvl_p` is not available to be called from extension on that version.

The new function does not have that limitation and I was successfully able to end-to-end profile a Ruby 2.2 application.

This PR is on top of `ivoanjo/prof-6556-leave-sigprof-signal-handler-v2` just to avoid any conflicts, but it otherwise indepentent.

Finally, I noticed there's still some CI issues on 2.3 and 2.4. I believe the 2.4 issue is related to the new images and should be fixed with a rebase; the 2.3 issue seems legit, I'll investigate it tomorrow.

**How to test the change?**:

This change includes code coverage! You can tell it was non-trivial to add, but hopefully it's doable to follow along.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -299,18 +299,16 @@ static VALUE stop(VALUE self_instance, VALUE optional_exception) {
 // We need to be careful not to change any state that may be observed OR to restore it if we do. For instance, if anything
 // we do here can set `errno`, then we must be careful to restore the old `errno` after the fact.
 static void handle_sampling_signal(DDTRACE_UNUSED int _signal, DDTRACE_UNUSED siginfo_t *_info, DDTRACE_UNUSED void *_ucontext) {
-  if (!ruby_thread_has_gvl_p()) {
-    return; // Not safe to enqueue a sample from this thread
-  }
-  if (!ddtrace_rb_ractor_main_p()) {
-    return; // We're not on the main Ractor; we currently don't support profiling non-main Ractors
-  }
+  if (
+    !ruby_native_thread_p() || // Not a Ruby thread
+    !is_current_thread_holding_the_gvl() || // Not safe to enqueue a sample from this thread
+    !ddtrace_rb_ractor_main_p() // We're not on the main Ractor; we currently don't support profiling non-main Ractors
+  ) return;
 
   // We implicitly assume there can be no concurrent nor nested calls to handle_sampling_signal because
   // a) we get triggered using SIGPROF, and the docs state second SIGPROF will not interrupt an existing one
   // b) we validate we are in the thread that has the global VM lock; if a different thread gets a signal, it will return early
   //    because it will not have the global VM lock
-  // TODO: Validate that this does not impact Ractors
 
   // Note: rb_postponed_job_register_one ensures that if there's a previous sample_from_postponed_job queued for execution
   // then we will not queue a second one. It does this by doing a linear scan on the existing jobs; in the future we
--- a/ext/ddtrace_profiling_native_extension/private_vm_api_access.c
+++ b/ext/ddtrace_profiling_native_extension/private_vm_api_access.c
@@ -65,29 +65,71 @@ bool is_current_thread_holding_the_gvl(void) {
   return owner.valid && pthread_equal(pthread_self(), owner.owner);
 }
 
-current_gvl_owner gvl_owner(void) {
-  // TODO: Reading owner/running is a racy read. Should we perhaps add a read barrier here?
-  const rb_thread_t *current_owner =
-    #ifndef NO_RB_THREAD_SCHED // Introduced in Ruby 3.2 as a replacement for struct rb_global_vm_lock_struct
-      GET_RACTOR()->threads.sched.running;
-    #elif HAVE_RUBY_RACTOR_H
-      GET_RACTOR()->threads.gvl.owner;
-    #else
-      GET_VM()->gvl.owner;
-    #endif
-
-  if (current_owner == NULL) return (current_gvl_owner) {.valid = false};
-
-  return (current_gvl_owner) {
-    .valid = true,
-    .owner =
-      #ifndef NO_RB_NATIVE_THREAD
-        current_owner->nt->thread_id
+#ifndef NO_GVL_OWNER // Ruby < 2.6 doesn't have the owner/running field
+  current_gvl_owner gvl_owner(void) {
+    // TODO: Reading owner/running is a racy read. Should we perhaps add a read barrier here?
+    const rb_thread_t *current_owner =
+      #ifndef NO_RB_THREAD_SCHED // Introduced in Ruby 3.2 as a replacement for struct rb_global_vm_lock_struct
+        GET_RACTOR()->threads.sched.running;
+      #elif HAVE_RUBY_RACTOR_H
+        GET_RACTOR()->threads.gvl.owner;
       #else
-        current_owner->thread_id
+        GET_VM()->gvl.owner;
       #endif
-  };
-}
+
+    if (current_owner == NULL) return (current_gvl_owner) {.valid = false};
+
+    return (current_gvl_owner) {
+      .valid = true,
+      .owner =
+        #ifndef NO_RB_NATIVE_THREAD
+          current_owner->nt->thread_id
+        #else
+          current_owner->thread_id
+        #endif
+    };
+  }
+#else
+  current_gvl_owner gvl_owner(void) {
+    rb_vm_t *vm = GET_VM();
+
+    // BIG Issue: Ruby < 2.6 did not have the owner field. The really nice thing about the owner field is that it's
+    // "atomic" -- when a thread sets it, it "declares" two things in a single step
+    // * Declaration 1: Someone has the GVL
+    // * Declaration 2: That someone is the specific thread
+    //
+    // Observation 1: On older versions of Ruby, this ownership concept is actually split. Specifically, `gvl.acquired`
+    // is a boolean that represents declaration 1 above, and `vm->running_thread` (or `ruby_current_thread`/
+    // `ruby_current_execution_context_ptr`) represents declaration 2.
+    //
+    // Observation 2: In addition, when a thread releases the GVL, it only sets `gvl.acquired` back to 0 **BUT CRUCIALLY
+    // DOES NOT CHANGE THE OTHER global variables**.
+    //
+    // Observation 1+2 above lead to the following possible race:
+    // * Thread A grabs the GVL (`gvl.acquired == 1`)
+    // * Thread A sets `running_thread` (`gvl.acquired == 1` + `running_thread == Thread A`)
+    // * Thread A releases the GVL (`gvl.acquired == 0` + `running_thread == Thread A`)
+    // * Thread B grabs the GVL (`gvl.acquired == 1` + `running_thread == Thread A`)
+    // * Thread A calls gvl_owner. Due to the current state (`gvl.acquired == 1` + `running_thread == Thread A`), this
+    //   function returns an incorrect result.
+    // * Thread B finally sets `running_thread` (`gvl.acquired == 1` + `running_thread == Thread B`)
+    //
+    // This is especially problematic because we use `gvl_owner` to implement `is_current_thread_holding_the_gvl` which
+    // is called in a signal handler to decide "is it safe for me to call `rb_postponed_job_register_one` or not".
+    // (See constraints in `collectors_cpu_and_wall_time_worker.c` comments for why).
+    //
+    // Thus an incorrect `is_current_thread_holding_the_gvl` result may lead to issues inside `rb_postponed_job_register_one`.
+    //
+    // For this reason we currently do not enable the new Ruby profiler on Ruby 2.5 and below by default, and we print a
+    // warning when customers force-enable it.
+    bool gvl_acquired = vm->gvl.acquired != 0;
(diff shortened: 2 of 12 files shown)


ext/ddtrace_profiling_native_extension/private_vm_api_access.c
```diff
@@ -49,6 +50,92 @@ rb_nativethread_id_t pthread_id_for(VALUE thread) {
...
+//
+// Thus we need our own gvl-checking method which actually looks at the gvl structure to determine if it is the owner.
+bool is_current_thread_holding_the_gvl(void) {
+  current_gvl_owner owner = gvl_owner();
+  return owner.valid && pthread_equal(pthread_self(), owner.owner);
+}
+
+#ifndef NO_GVL_OWNER // Ruby < 2.6 doesn't have the owner/running field
+  current_gvl_owner gvl_owner(void) {
+    // TODO: Reading owner/running is a racy read. Should we perhaps add a read barrier here?
```
marcotc, line 70: Will a barrier help here, given the data source (owner/running) won't have any synchronization with `is_current_thread_holding_the_gvl`?

marcotc: It might ok to just say something like "we can't synchronize with the data source, so this read is racy"

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
```diff
@@ -49,6 +50,92 @@ rb_nativethread_id_t pthread_id_for(VALUE thread) {
...
 }
 
+// Queries if the current thread is the owner of the global VM lock.
+//
+// @ivoanjo: Ruby has a similarly-named `ruby_thread_has_gvl_p` but that API is insufficient for our needs because it can
+// still return `true` even when a thread DOES NOT HAVE the global VM lock.
+// In particular, looking at the implementation, that API assumes that if a thread is not in a "blocking region" then it
+// will have the GVL which is probably true for the situations that API was designed to be called from BUT this assumption
+// does not hold true when calling `ruby_thread_has_gvl_p` from a signal handler. (Because the thread may have lost the
+// GVL due to a scheduler decision, not because it decided to block.)
```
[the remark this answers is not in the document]
ivoanjo: This is a good question... I think probably not. For a regular C extension, the "blocking region" approach is completely correct -- if you are outside a blocking region then you will have the GVL and if you aren't then you won't.

But because we're interrupting Ruby VM threads half-way through their regular work and calling that API, we see the "extra" situation.

We could try it and see if upstream is interested BUT I have been trying to be very selective, because there's not a lot of resources, so there's a few things I'll try to get fixed/improved (such as https://bugs.ruby-lang.org/issues/19112, https://bugs.ruby-lang.org/issues/19114, https://bugs.ruby-lang.org/issues/18464, ...) before I think it's worth giving this a shot.

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
ivoanjo: You're right; this is always racy so unless we do decide to acquire the lock that Ruby uses to protect the `owner`, it will always be possible to have a racy read.

To be honest, the racy read is actually not a problem in this case:
* Only a thread sets and unset itself as the owner, so if a thread sees itself as the owner, that will be correct
* In a future PR I'll use `gvl_owner` to target the SIGPROF signals in the profiler, but in that case sending a signal to another thread is always a race so if we read something outdated here it's no problem, that's why the signal handler checks if it landed on the right thread.

I'll update the comment to reflect this discussion 👍

ivoanjo: Comment added in d9b420c3e6024b9fa6a7a6f7c4acc6197dc901fc

ivoanjo:
Update: I've rebased this PR on top of master + fixed the flaky spec on Ruby 2.3 (8af18d0894493272c4daa4c761bc36e1ee6d0862).

ivoanjo:
This PR exposes a bug in the profiler in the form of a flaky test (the root cause was already there -- we just didn't have any test that triggered it). The fix is in #2429 .

I don't want to join both PRs (I think it's easier to evaluate them separately) but I'll make sure to merge them both together, so nobody gets hit by the flaky test.

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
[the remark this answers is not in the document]
lloeki: I was going to comment on that as well. I think it's still worth bringing it up.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -299,18 +299,16 @@ static VALUE stop(VALUE self_instance, VALUE optional_exception) {
...
-  if (!ruby_thread_has_gvl_p()) {
-    return; // Not safe to enqueue a sample from this thread
-  }
-  if (!ddtrace_rb_ractor_main_p()) {
-    return; // We're not on the main Ractor; we currently don't support profiling non-main Ractors
-  }
+  if (
+    !ruby_native_thread_p() || // Not a Ruby thread
+    !is_current_thread_holding_the_gvl() || // Not safe to enqueue a sample from this thread
+    !ddtrace_rb_ractor_main_p() // We're not on the main Ractor; we currently don't support profiling non-main Ractors
```
lloeki, line 305: Shoudl that comment be a `// TODO:`?

I was wondering if we should have a categorization system to help focusing on some work, eg: `TODO: [Ractor] blah blah`, so that we can `grep -e 'TODO: [Ractor]' lib` and have all the TODOs on the `Ractor` support topic.

lib/datadog/core/configuration/components.rb
```diff
@@ -345,6 +347,23 @@ def should_enable_gc_profiling?(settings)
               false
             end
           end
+
+          def print_new_profiler_warnings
+            if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.6')
+              Datadog.logger.warn(
+                'New Ruby profiler has been force-enabled. This feature is in alpha state. Please report any issues ' \
+                'you run into!'
```
lloeki, line 355: Should the message include a conveninence link to `https://github.com/DataDog/dd-trace-rb/issues/new`?

lib/datadog/core/configuration/components.rb
```diff
@@ -345,6 +347,23 @@ def should_enable_gc_profiling?(settings)
...
+              )
+            else
+              # For more details on the issue, see the "BIG Issue" comment on `gvl_owner` function in
+              # `private_vm_api_access.c`.
+              Datadog.logger.warn(
+                'New Ruby profiler has been force-enabled on a legacy Ruby version (< 2.6). This is not recommended in ' \
+                'production environments, as due to limitations in Ruby APIs, we suspect it may lead to crashes in very ' \
+                'rare situations. Please report any issues you run into!'
```
lloeki, line 363: Should the message include a conveninence link to `https://github.com/DataDog/dd-trace-rb/issues/new`?

spec/datadog/core/configuration/components_spec.rb
```diff
@@ -1037,8 +1038,32 @@
...
+              /New Ruby profiler has been force-enabled on a legacy Ruby version \(< 2.6\). This is not recommended/
+            )
+
+            build_profiler
+          end
+        end
+
         context 'on Ruby 2.x' do
-          before { skip 'Behavior does not apply to current Ruby version' if RUBY_VERSION >= '3' }
+          before { skip 'Behavior does not apply to current Ruby version' if RUBY_VERSION >= '3.' }
```
lloeki, line 1066: Why not `3.0`?

spec/datadog/core/configuration/components_spec.rb
```diff
@@ -1050,7 +1075,7 @@
         end
 
         context 'on Ruby 3.x' do
-          before { skip 'Behavior does not apply to current Ruby version' if RUBY_VERSION < '3' }
+          before { skip 'Behavior does not apply to current Ruby version' if RUBY_VERSION < '3.' }
```
lloeki, line 1078: Why not `3.0`?

spec/datadog/profiling/collectors/interesting_backtrace_helper.rb
```diff
@@ -135,6 +135,7 @@ def self.hello
 end
 
 def ibh_method_with_complex_parameters(a, b = nil, *c, (d), f:, g: nil, **h, &i)
+  d.to_s
```
lloeki, line 138: What is the rationale for this call to `#to_s` into void?

spec/datadog/profiling/native_extension_spec.rb
```diff
@@ -197,4 +197,64 @@ def wait_for_thread_to_die
...
+      subject(:is_current_thread_holding_the_gvl) do
+        Datadog::Profiling::NativeExtension::Testing._native_release_gvl_and_call_is_current_thread_holding_the_gvl
+      end
+
+      it { is_expected.to be false }
+    end
+
+    describe 'correctness' do
+      before do
+        skip 'Ruby 2.2 does not expose ruby_thread_has_gvl_p so nothing to compare to' if RUBY_VERSION.start_with?('2.2.')
```
lloeki, line 220: What of Ruby 2.1?

lloeki (approved):
Gnarly issue! LGTM, with a few comments.

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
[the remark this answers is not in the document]
ivoanjo: Fair enough, y'all convinced me 😄. Opened upstream: https://bugs.ruby-lang.org/issues/19172

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
ivoanjo: Supporting Ractors will need A LOT of changes to the components keeping state to make them safe to be called concurrently. 

Furthermore, Ractors themselves are still quite buggy so I don't think they will be a production thing for several years.

So because of that, the future seems so nebulous that I don't expect any TODOs I leave for now will be useful for later -- we're always going to need to comb through these components and re-evaluate them closely if and when we decide to move to Ractors.

lib/datadog/core/configuration/components.rb
ivoanjo: Fixed in 960c4fa128a6dbf67b076baaa5849c17a0e77395

spec/datadog/core/configuration/components_spec.rb
ivoanjo: Either is fine! Changed in 482d94643fc45c023482d38938380d85d3107e99

spec/datadog/core/configuration/components_spec.rb
ivoanjo: Either is fine! Changed in 482d94643fc45c023482d38938380d85d3107e99

spec/datadog/profiling/native_extension_spec.rb
ivoanjo: The profiler no longer supports 2.1 -- we spit out a warning instead ( https://github.com/DataDog/dd-trace-rb/pull/2140 ) and don't even try to compile the native extension.

ivoanjo:
Update: As I documented above, this change exposes some issues for the `CpuAndWallTimeWorker` component that were already there but were previously not caught in our testing and that manifest themselves as flaky tests. I am, and will address these issues in the 4+ PRs will sit atop this one.

To enable incremental delivery of the PRs, as well as help me not go mad with rebases and whatnot, I've just pushed a commit (6cfff38439f6539e1f169c98999934ce59f640df) which disables the `cpu_and_wall_time_worker_spec.rb`. This will be reverted at the end of the full set of PRs.

PR #2417, opened 2022-11-23
Fix ruby 2.4 image

lloeki:
**What does this PR do?**

Ruby 2.4 image update to Debian Buster should have used a new cache, but a CI race condition on the `binary_version` file update may have caused cache pollution with Stretch-built gems, in turn making any later attempt to use this image fail.

**Motivation**

CircleCI `test-2.4` job failing all around.

**How to test the change?**

CircleCI `test-2.4` job should turn green.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -494,7 +494,7 @@ job_configuration:
     resource_class_to_use: medium+
   - &config-2_4
     <<: *filters_all_branches_and_tags
-    ruby_version: '2.4'
+    ruby_version: '2.4.10'
     image: ghcr.io/datadog/dd-trace-rb/ruby:2.4.10-dd
     resource_class_to_use: medium+
   - &config-2_5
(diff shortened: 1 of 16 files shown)


marcotc (approved):

lloeki:
What happened:

* In [the 2.4 PR](https://github.com/DataDog/dd-trace-rb/pull/2404) I updated the base image from Stretch to Buster
* To use a new CI cache I bumped the `binary_version` file contents from `1` to `2` (this allows using a new cache key without affecting branches before that commit, which the `CIRCLECI_CACHE_VERSION` does since it is an absolute global)
* But the image is not yet published, so the PR ran on CircleCI with `2` using the _old_ image, producing a cache with the key intended for the new image
* THEN the new 2.4 image was published by a manual `workflow_dispatch` on GHA
* And ONLY THEN subsequent CircleCI jobs can pick that image up, use the proper key for the image but with mismatched content, and fail...

.circleci/config.yml
```diff
@@ -494,7 +494,7 @@ job_configuration:
     resource_class_to_use: medium+
   - &config-2_4
     <<: *filters_all_branches_and_tags
-    ruby_version: '2.4'
+    ruby_version: '2.4.10'
```
ivoanjo, line 497: Consider perhaps documenting here why this is different from the others? Otherwise it may be a bit confusing next time someone needs to update the images :)

PR #2412, opened 2022-11-23
Update to JRuby 9.2.21.0 and 9.3.9.0

lloeki:
Built 9.3 from a buster base, vendored in to attempt to avoid any issues with bullseye.

Also, a bit of alignment on 9.2.8.0.

And removed the `-latest` suffixes as they don't mean much.

PR #2418, opened 2022-11-23
Use updated JRubies in CircleCI

lloeki:
**What does this PR do?**

Use JRubies built from https://github.com/DataDog/dd-trace-rb/pull/2412

**How to test the change?**

CI shoudl be green


marcotc (approved):

lloeki:
Bah, a commit went missing during rebase,

lloeki:
@marcotc looks like we might have legit failures on JRuby 9.3.9.0, I threw a retry at it and the same expectations failed.

ivoanjo:
The spec that failed is a weird one because different Rubies have different behaviors:

https://github.com/DataDog/dd-trace-rb/blob/29c245e26b4fc3f0bc907032f27bc054d212fed0/spec/datadog/core/error_spec.rb#L133-L164

I guess JRuby has now adopted pre 2.6/TruffleRuby behavior?

lloeki:
Changed base to depend on https://github.com/DataDog/dd-trace-rb/pull/2422

lloeki:
> I guess JRuby has now adopted pre 2.6/TruffleRuby behavior?

In the [JRuby 9.3.7.0 release notes](https://www.jruby.org/2022/08/16/jruby-9-3-7-0.html) there's a reference to https://github.com/jruby/jruby/issues/7267. 

> Circular exception handling can cause infinite loop since 9.3.4.0

spec/datadog/core/error_spec.rb
```diff
@@ -130,7 +130,7 @@ def call
             end
           end
 
-          it 'reports errors only once', if: (RUBY_VERSION < '2.6.0' || PlatformHelpers.truffleruby?) do
+          it 'reports errors only once', if: (RUBY_VERSION < '2.6.0' || PlatformHelpers.truffleruby? || PlatformHelpers.jruby? && RUBY_ENGINE_VERSION >= '9.3.7.0') do
```
ivoanjo, line 133: I suggest adding some parenthesis so people don't have to wonder about precedence ;)

```suggestion
          it 'reports errors only once', if: (RUBY_VERSION < '2.6.0' || PlatformHelpers.truffleruby? || (PlatformHelpers.jruby? && RUBY_ENGINE_VERSION >= '9.3.7.0')) do
```

ivoanjo (approved):
👍 LGTM. Rubocop has a line-too-long complaint. Sigh, I hate this Rubocop rule -- you add one more character to a PR that worked and Rubocop breaks everything :(

spec/datadog/core/error_spec.rb
lloeki: Add too much parentheses and Rubocop will complain :D

I would otherwise agree in general but I'm pretty sure most everyone knows the precedence between `||` and `&&` so it's quire readable.

PR #2421, opened 2022-11-24
RFC:Remove devtools from main gem set

marcotc:
It would be best to have the test matrix run with only gems relevant to the test at hand.

We have many development tools in our Gemfile, which get included in all gem set combinations, in some occasions causing conflicts.

It seems more practical to me, at this point, to separate everything that **we know** is not necessary to run `ddtrace` or its test suite into separate "groups", and keep what's left as "essential to ddtrace or testing ddtrace".

This PR, at this point, is just a sketch for possible approaches we can take.


Gemfile
```diff
@@ -84,12 +73,31 @@ if RUBY_PLATFORM != 'java'
...
-# Sorbet releases almost daily, with new checks introduced that can make a
-# previously-passing codebase start failing. Thus, we need to lock to a specific
-# version and bump it from time to time.
-# Also, there's no support for windows
-if RUBY_VERSION >= '2.4.0' && (RUBY_PLATFORM =~ /^x86_64-(darwin|linux)/)
-  gem 'sorbet', '= 0.5.9672'
-  gem 'spoom', '~> 1.1'
+group :appraisal do
+  gem 'appraisal', '~> 2.2'
+end
```
marcotc, line 78: ### Approach 1

Very specific groups, pretty much matching the tool name: `appraisal`, `rubocop`, `yard`, `sorbet`, etc.

Gemfile
```diff
@@ -84,12 +73,31 @@ if RUBY_PLATFORM != 'java'
...
+
+group :release do
+  gem 'pimpmychangelog', '>= 0.1.2' # Formats the CHANGELOG.md file
+  gem 'redcarpet', '~> 3.4' if RUBY_PLATFORM != 'java' # Used by YARD to generate docs
+  gem 'yard', '~> 0.9'
+end
```
marcotc, line 84: ### Approach 2

Locally grouped gems, matching the high-level activity they are part of: e.g. `release`, `lint`, `benchmark`, `test`, etc.

marcotc: ### Approach 2a

Locally grouped gems, but more granular then `2`: e.g. `changelog` (has `pimpmychangelog`), `docs` (has `redcarpet` & `yard`), `lint` (has `rubocop*`), `type_check` (has `sorbet` & `spoom`) etc.

Gemfile
```diff
@@ -84,12 +73,31 @@ if RUBY_PLATFORM != 'java'
...
+
+group :static_check do
```
lloeki, line 86: I usually call that just `:check` or `:lint`

Gemfile
lloeki: This would be extremely granural, not too keen on that solution.

lloeki:
I'm very much liking this, if only because it should resolve the `aarch64` vs `x86_64` inconsistency of Sorbet in appraisal gemfiles and lockfiles (which I planned to solve in this way but this goes further)

Gemfile
ivoanjo: I don't have strong thoughts either way, but I think coarser-grained groups seem like a good enough start, and we could always break them up as needed later.

ivoanjo:
This looks like a neat idea; nevertheless, I'm thinking this needs a lot of rework to make useful again -- we've moved A LOT of stuff around since this PR was opened.

So... I'm going to go ahead and close this one until we have bandwidth to restart it >_>

PR #2422, opened 2022-11-24
Align Ruby CI cache keys

lloeki:
**What does this PR do?**

Making sure to use a dedicated cache per patch version

**Motivation**

Changing versions (even patch) sometimes comes with userland changes as well (e.g glibc gets bumped) as the base image gets updated.

This has caused some issues in the past, either directly or via cache pollution from an updated branch saving a new cache then an earlier non-updated branch picking that cache up.

So making sure to use a dedicated cache per patch version reduces potential for breakage.

**Additional Notes**

This also adds a `ruby-` prefix, materialising the ruby engine in use, making it consistent with other non-MRI engines. It shoudl probably apply to job names as well, but I punted on that because of the GitHub PR merge enforcement on job names.

**How to test the change?**

CI should be green.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -481,37 +481,37 @@ job_configuration:
   # MRI
   - &config-2_1
     <<: *filters_all_branches_and_tags
-    ruby_version: '2.1'
+    ruby_version: 'ruby-2.1.10'
     image: ghcr.io/datadog/dd-trace-rb/ruby:2.1.10-dd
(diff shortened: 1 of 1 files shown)


ivoanjo (approved):
👍 LGTM

To be honest the commented-out TruffleRuby stuff on `config.yml` has been there for quite a while, so feel free to remove it, we can always re-add it later.

PR #2424, opened 2022-11-24
Add jruby 9.4.0.0

lloeki:
**What does this PR do?**

Add JRuby 9.4.0.0 for local development

**Motivation**

Challenge running on 9.4 so we can either fix things our side or report any issue upstream as early as possible.

**Additional Notes**

There's no official Docker image yet, so we build our own.

This is based on buster + OpenJDK 11 to avoid muddying the waters with additional dimensions along which things may break. Maybe it could be worthwhile to move that to a later OpenJDK at some point down the road.

**How to test the change?**

```
docker volume rm dd-trace-rb_bundle-jruby-9.4
docker buildx build . --platform linux/$(arch) -f Dockerfile-jruby-9.4.0.0 -t ghcr.io/datadog/dd-trace-rb/jruby:9.4.0.0-dd
docker-compose run --rm tracer-jruby-9.4 /bin/bash -c 'rm -vf Gemfile.lock && bundle install && bundle exec appraisal generate && bundle exec appraisal install'
```


lloeki:
Based on https://github.com/DataDog/dd-trace-rb/pull/2418

ivoanjo (approved):
👍 LGTM. I was not able to test the image locally, since it's not uploaded yet, but the Dockerfile looks good, basically the same as for other Rubies other than the version bump, so I see no reason to not merge this in :)

PR #2408, opened 2022-11-23
Use Ruby 2.7.6 in CircleCI

lloeki:
Depends on https://github.com/DataDog/dd-trace-rb/pull/2398

PR #2425, opened 2022-11-24
Tweak profiler code to avoid triggering bug in compiler warnings

ivoanjo:
**What does this PR do?**:

This PR slightly tweaks some of the profiler C code to avoid triggering an incorrect compiler warning, and thus breaking CI (because we turn all warnings into errors in CI).

**Motivation**:

In <https://github.com/DataDog/dd-trace-rb/issues/2377> I documented a set of compiler warnings that are incorrect. Usually having incorrect warnings is not an issue, but in CI we compile with `-Werror` to turn all warnings into errors, so we can't just ignore them.

In <https://github.com/DataDog/dd-trace-rb/pull/2408> while upgrading our Ruby images, we ran into this issue again on our own CI.

**Additional Notes**:

Can I go back to coding Ruby?

**How to test the change?**:

CI should remain green, even when using the 2.7 docker image from <https://github.com/DataDog/dd-trace-rb/pull/2408>.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_stack.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_stack.c
@@ -144,11 +144,10 @@ void sample_thread(
 
   // Samples thread into recorder, including as a top frame in the stack a frame named "Garbage Collection"
   if (type == SAMPLE_IN_GC) {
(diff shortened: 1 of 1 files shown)


lloeki (commented):
> Can I go back to coding Ruby?

No.

🤪

lloeki (approved):

PR #2426, opened 2022-11-24
Update actions/checkout and actions/*-artifact to Node 16

lloeki:
**What does this PR do?**

Update actions to use Node 16-based ones.

**Motivation**

Node 12 actions are deprecated.

**How to test the change?**

GHA CI should stay green, no deprecation notice should be shown in summaries.


ivoanjo (approved):
👍 LGTM

PR #2429, opened 2022-11-25
[PROF-6556] Fix profiler issue due to Ruby native thread reuse

ivoanjo:
[PROF-6556] Fix profiler issue due to Ruby native thread reuse

**What does this PR do?**:

This PR fixes an issue that caused Ruby threads to inherit SIGPROF blocking from the `CpuAndWallTimeWorker` worker thread.

The issue is caused by an interaction with Ruby's native thread cache. When a Ruby `Thread` dies, Ruby does not immediately kill the underlying OS native thread.

Instead, it keeps it around for some time, and if another Ruby `Thread` starts in the meanwhile, it reuses the OS native thread, rather than getting a new one.

This can be observed easily starting on Ruby 3.1+ which actually added a method to observe the native thread id:

```ruby
puts RUBY_DESCRIPTION

puts Thread.new { Thread.current.native_thread_id }.value
puts Thread.new { Thread.current.native_thread_id }.value
puts Thread.new { Thread.current.native_thread_id }.value

puts "Sleeping for a bit..."
sleep 6

puts Thread.new { Thread.current.native_thread_id }.value
```

result:

```
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-darwin20]
293997
293997
293997
Sleeping for a bit...
294151
```

(Note that this behavior is much older than Ruby 3.1, but  `#native_thread_id` was only introduced on that version).

The issue fixed by this PR was that the profiler blocked all SIGPROF signal delivery on its worker thread, because SIGPROF signals should never land on it.

After the profiler was stopped, its worker `Thread` died, but Ruby could decide to reuse the OS native thread, which meant that a random `Thread` instance created shortly after stopping the profiler could "inherit" the SIGPROF blocking.

This is definitely not intended and actually caused a recently-introduced spec to become flaky, because the test created a new `Thread` and that thread was inheriting the SIGPROF blocking.

```
Run options: include {:focus=>true, :ids=>{"./spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb"=>["1:2:7:1"], "./spec/datadog/profiling/native_extension_spec.rb"=>["1:5:3:1"]}}

Randomized with seed 19584

Datadog::Profiling::Collectors::CpuAndWallTimeWorker
  #start
    when a previous signal handler existed
      does not start the sampling loop

Datadog::Profiling::NativeExtension
  is_current_thread_holding_the_gvl
    correctness
      returns accurate results when compared to ruby_thread_has_gvl_p (FAILED - 1)

Failures:

  1) Datadog::Profiling::NativeExtension is_current_thread_holding_the_gvl correctness returns accurate results when compared to ruby_thread_has_gvl_p
     Failure/Error:
       result = Datadog::Profiling::NativeExtension::Testing
         ._native_trigger_holding_the_gvl_signal_handler_on(background_thread)

     RuntimeError:
       Could not signal background_thread
     # ./spec/datadog/profiling/native_extension_spec.rb:255:in `_native_trigger_holding_the_gvl_signal_handler_on'
     # ./spec/datadog/profiling/native_extension_spec.rb:255:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:216:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:108:in `block (2 levels) in <top (required)>'
     # /Users/ivo.anjo/.rvm/gems/ruby-2.7.6/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'

```

To fix this issue, I've refactored the setup and teardown of the `CpuAndWallTimeWorker` to make sure all clean up code runs correctly AND added a new clean up step to re-enable SIGPROF signal delivery on the worker `Thread` before it dies.

**Motivation**:

Beyond fixing the flaky test, this issue could cause the profiler to not be able to take samples, if e.g. a customer app was doing all its work on a `Thread` that happened to get SIGPROF signal delivery blocked.

**Additional Notes**:

The native thread reuse mechanism is interesting, but I wonder how many native gems have bugs hidden where they store things per-native-thread and forget that this can happen.

This PR is on top of #2415 (and the flaky test is introduced on that PR).

GitHub seems to be having issues with their container hosting, so CircleCI is not able to run CI. I'll re-trigger CI later.

**How to test the change?**:

[PROF-6556]: https://datadoghq.atlassian.net/browse/PROF-6556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ext/ddtrace_profiling_native_extension/profiling.c
+++ b/ext/ddtrace_profiling_native_extension/profiling.c
@@ -112,6 +112,7 @@ static VALUE _native_trigger_holding_the_gvl_signal_handler_on(DDTRACE_UNUSED VA
 
   pthread_mutex_lock(&holding_the_gvl_signal_handler_mutex);
 
+  // We keep trying for ~5 seconds (500 x 10ms) to try to avoid any flakiness if the test machine is a bit slow
   for (int tries = 0; holding_the_gvl_signal_handler_result[0] == Qfalse && tries < 500; tries++) {
     pthread_kill(thread, SIGPROF);
 
--- a/ext/ddtrace_profiling_native_extension/profiling.c
+++ b/ext/ddtrace_profiling_native_extension/profiling.c
@@ -1,5 +1,6 @@
 #include <ruby.h>
 #include <ruby/thread.h>
+#include <errno.h>
 
 #include "clock_id.h"
 #include "helpers.h"
@@ -111,14 +112,28 @@ static VALUE _native_trigger_holding_the_gvl_signal_handler_on(DDTRACE_UNUSED VA
 
   pthread_mutex_lock(&holding_the_gvl_signal_handler_mutex);
 
-  for (int tries = 0; holding_the_gvl_signal_handler_result[0] == Qfalse && tries < 100; tries++) {
+  for (int tries = 0; holding_the_gvl_signal_handler_result[0] == Qfalse && tries < 500; tries++) {
     pthread_kill(thread, SIGPROF);
 
+    // pthread_cond_timedwait is simply awful -- the deadline is based on wall-clock using a struct timespec, so we need
+    // all of the below complexity just to tell it "timeout is 10ms". The % limit dance below is needed because the
+    // `tv_nsec` part of a timespec can't go over the limit.
     struct timespec deadline;
     clock_gettime(CLOCK_REALTIME, &deadline);
-    deadline.tv_nsec += 10 * 1000 * 1000 /* 10ms */;
 
-    pthread_cond_timedwait(&holding_the_gvl_signal_handler_executed, &holding_the_gvl_signal_handler_mutex, &deadline);
+    unsigned int timeout_ns = 10 * 1000 * 1000 /* 10ms */;
+    unsigned int tv_nsec_limit = 1000 * 1000 * 1000 /* 1s */;
+    if ((deadline.tv_nsec + timeout_ns) < tv_nsec_limit) {
+      deadline.tv_nsec += timeout_ns;
+    } else {
+      deadline.tv_nsec = (deadline.tv_nsec + timeout_ns) % tv_nsec_limit;
+      deadline.tv_sec++;
+    }
+
+    int error = pthread_cond_timedwait(&holding_the_gvl_signal_handler_executed, &holding_the_gvl_signal_handler_mutex, &deadline);
+    if (error && error != ETIMEDOUT) {
+      rb_exc_raise(rb_syserr_new_str(error, rb_sprintf("Unexpected failure in _native_trigger_holding_the_gvl_signal_handler_on")));
+    }
   }
 
   pthread_mutex_unlock(&holding_the_gvl_signal_handler_mutex);
--- a/lib/datadog/profiling/collectors/cpu_and_wall_time_worker.rb
+++ b/lib/datadog/profiling/collectors/cpu_and_wall_time_worker.rb
@@ -62,7 +62,6 @@ module Datadog
 
             return unless @worker_thread
 
-            @worker_thread.kill
             self.class._native_stop(self)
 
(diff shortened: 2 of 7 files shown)


lib/datadog/profiling/collectors/cpu_and_wall_time_worker.rb
```diff
@@ -62,7 +62,6 @@ def stop(*_)
 
             return unless @worker_thread
 
-            @worker_thread.kill
```
lloeki, line 65: So we're not actively killing `@worker_thread` anymore, merely relying on the `join` below to wait for it to disappear?

What is stopping `@worker_thread` then? Is that part of a side effect of `self.class._native_stop(self)`?

lloeki (approved):
Changes look reasonable, I do have a few Qs though for parts that I can't quite wrap my head around.

ext/ddtrace_profiling_native_extension/profiling.c
```diff
@@ -111,14 +112,28 @@ static VALUE _native_trigger_holding_the_gvl_signal_handler_on(DDTRACE_UNUSED VA
 
   pthread_mutex_lock(&holding_the_gvl_signal_handler_mutex);
 
-  for (int tries = 0; holding_the_gvl_signal_handler_result[0] == Qfalse && tries < 100; tries++) {
+  for (int tries = 0; holding_the_gvl_signal_handler_result[0] == Qfalse && tries < 500; tries++) {
```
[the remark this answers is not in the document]
ivoanjo: Yes! This spec is inherently a bit fragile since it relies on multiple threads and some communication between them. I've increased the tries from 100 (that is 100 x 10 => 1 second) to 500 (500 x 10 => 5 seconds) to avoid any potential flakiness caused by the test machine being slow.

I've documented this in f66bc7d3af26bcc0805e1eb0eaceade13967caf8

lib/datadog/profiling/collectors/cpu_and_wall_time_worker.rb
ivoanjo: The worker thread is "stuck" inside `_native_sampling_loop`, which **releases the Global VM Lock** and calls out to `run_sampling_trigger_loop` which is doing

```c
  while (state->should_run) {

    // TODO: This is still a placeholder for a more complex mechanism. In particular:
    // * We want to signal a particular thread or threads, not the process in general
    // * We want to track if a signal landed on the thread holding the global VM lock and do something about it
    // * We want to do more than having a fixed sampling rate

    kill(getpid(), SIGPROF);
    nanosleep(&time_between_signals, NULL);
  }
```

The `_native_stop` sets `state->should_run = false` and so makes it come out of this loop, and thus return and finish cleanly.

Calling `@worker_thread.kill` effectively did the same -- because the worker thread runs that loop without the global VM lock, and we give the VM an "interruption function", which it calls when it wants to force us to stop what we're doing:

```c
// This is called by the Ruby VM when it wants to shut down the background thread
static void interrupt_sampling_trigger_loop(void *state_ptr) {
  struct cpu_and_wall_time_worker_state *state = (struct cpu_and_wall_time_worker_state *) state_ptr;

  state->should_run = false;
}
```

The difference between `_native_stop` and `Thread#kill` is only what happens when the thread re-acquires the global VM lock before coming back to Ruby land.

With `_native_stop`, the thread acquires the GVL normally and returns from `_native_sampling_loop` back to the Ruby code you see in `cpu_and_wall_time_worker.rb`. With `Thread#kill`, the VM will behave as if an exception was raised, and `_native_sampling_loop` will never return back to Ruby code.

PR #2436, opened 2022-11-28
[PROF-6556] Add stat tracking to `CpuAndWallTimeWorker`

ivoanjo:
**What does this PR do?**:

This PR adds an inner `stats` struct to the `CpuAndWallTimeWorker` state. This is similar to what we already did for the `CpuAndWallTime` state.

Additionally, it adds 3 stats:
* `trigger_sample_attempts`
* `signal_handler_enqueued_sample`
* `signal_handler_wrong_thread`

and starts keeping track of them.

**Motivation**:

I'm currently working on changing the profiler to send a SIGPROF signal not to the whole process as we do currently, but to the specific thread that owns the Global VM Lock. (See #2415 for some of the groundwork for this).

But, testing such a change is kinda hard! How do you test that you're now sending signals to a specific thread instead of the whole process?

I decided to approach the testing by looking at stats: for an example that is known to cause issues (this will come on a later PR), how often are we able to land on the correct thread and sample?

I've extracted this PR from my working branch, as it can be reviewed in isolation.

**Additional Notes**:

This PR is on top of #2429 for my development convenience, but does not otherwise depend on it.

**How to test the change?**:

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time.c
@@ -693,8 +693,6 @@ static VALUE per_thread_context_st_table_as_ruby_hash(struct cpu_and_wall_time_c
   return result;
 }
 
-#define VALUE_COUNT(array) (sizeof(array) / sizeof(VALUE))
-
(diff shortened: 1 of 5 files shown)


lloeki (approved):
Looks good!

PR #2437, opened 2022-11-28
[PROF-6556] Refactor profiling global variables

ivoanjo:
**What does this PR do?**:

This PR adds a new global variable, `active_sampler_instance_state`, that is easier and cheaper to access during profiling to look up and record information. It then changes previous usages of `active_sampler_instance` to use this new one directly.

Finally, the `active_sampler_owner_thread` global variable was removed, and it was moved inside the state.

**Motivation**:

When the profiler is enabled, it keeps a reference to its state on a global variable. This is needed because in some contexts (e.g. signal handling) we don't have a good way to pass in the current state as an argument.

Up until now, we were only storing the `active_sampler_instance`, e.g. the Ruby `VALUE` from which we can extract the `CpuAndWallTimeWorker` state.

But... this got really awkward, and potentially more expensive, as in a bunch of places on the `CpuAndWallTimeWorker` we keep needing to do `rb_typeddata_is_kind_of` and `TypedData_Get_struct`.

To simplify this, we now store a direct pointer to the `CpuAndWallTimeWorker` state.

**Additional Notes**:

This PR also includes a few unrelated spec tweaks that I've extracted from my branch. See individual commits for details.

This PR is on top of #2436 to avoid merge conflicts. It is otherwise independent.

**How to test the change?**:

This refactor did not change any behavior, so the existing tests being green signal it's working correctly.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -76,6 +76,7 @@ struct cpu_and_wall_time_worker_state {
   // atomic operations. stdatomic.h seems a nice thing to reach out for.
   volatile bool should_run;
   bool gc_profiling_enabled;
+  VALUE self_instance;
   VALUE cpu_and_wall_time_collector_instance;
 
   // When something goes wrong during sampling, we record the Ruby exception here, so that it can be "re-raised" on
@@ -131,8 +132,10 @@ static VALUE _native_stats(DDTRACE_UNUSED VALUE self, VALUE instance);
 // Global state -- be very careful when accessing or modifying it
 
 // Note: Global state must only be mutated while holding the global VM lock (we piggy back on it to ensure correctness).
-// The active_sampler_instance needs to be global because we access it from the signal handler.
+// The active_sampler_instance_state needs to be global because we access it from a few functions (e.g. signal handler)
+// where we can't easily pass in the context as an argument.
 static VALUE active_sampler_instance = Qnil;
+struct cpu_and_wall_time_worker_state *active_sampler_instance_state = NULL;
 // ...We also store active_sampler_owner_thread to be able to tell who the active_sampler_instance belongs to (and also
 // to detect when it is outdated)
 static VALUE active_sampler_owner_thread = Qnil;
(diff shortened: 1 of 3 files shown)


ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -442,9 +438,9 @@ static VALUE release_gvl_and_run_sampling_trigger_loop(VALUE instance) {
 // This method exists only to enable testing Datadog::Profiling::Collectors::CpuAndWallTimeWorker behavior using RSpec.
 // It SHOULD NOT be used for other purposes.
 static VALUE _native_is_running(DDTRACE_UNUSED VALUE self, VALUE instance) {
-  return \
-    (active_sampler_owner_thread != Qnil && is_thread_alive(active_sampler_owner_thread) && active_sampler_instance == instance) ?
-    Qtrue : Qfalse;
+  struct cpu_and_wall_time_worker_state *state = active_sampler_instance_state; // Read from global variable
+
+  return (state != NULL && is_thread_alive(state->owner_thread) && state->self_instance == instance) ? Qtrue : Qfalse;
```
lloeki, line 443: I was wondering about the thread safety of this check. Line 255 has a comment and it seems to be the only area where mutation happens.

I would suggest annotating this with a comment though.

spec/datadog/profiling/integration_spec.rb
```diff
@@ -120,8 +120,9 @@ def stack_two
           @current_span = span
           @current_root_span = trace.send(:root_span)
           example.run
+
+          Datadog::Tracing.shutdown!
```
lloeki, line 124: Is tracing shutdown happening within a trace do .. end block a legit move? It feels odd.

lloeki (approved):
Looks reasonable. A couple notes/Qs.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
ivoanjo: I've clarified it in c7f25c3f1e2ecf8a3bcd5a90f0557e7582eee683; let me know if that's still not clear.

spec/datadog/profiling/integration_spec.rb
ivoanjo: I've dropped a note to the tracing folks to ask them if this is OK or not. I think it can easily be reverted if needed, so I'll go ahead and merge the PR with this change.

delner: My concern is that this will cause side-effects in other tests (as it might mutate global state that isn't reset.)

My suggestions would be:

 - Add a mock to black hole the traces
 - Activate the "test" mode which introduces an adapter that intercepts outgoing payloads and puts them in an array instead.

ivoanjo: Opened https://github.com/DataDog/dd-trace-rb/pull/2456 to improve this.

PR #2439, opened 2022-11-29
Fix using legacy C definitions

ivoanjo:
**What does this PR do?**:

This PR fixes one case where we were using a legacy style for defining a C function + explicitly enables compiler warnings for these kinds of legacy definitions.

**Motivation**:

We want to avoid the use of legacy definitions, to avoid issues moving to future compilers and/or future C language versions (C11 I'm looking at you).

**Additional Notes**:

For some reason `-Wall -Wextra` did not enable these compiler warnings. I'll be on the look out for more warnings that we can also enable.

**How to test the change?**:

Validate that CI is still green.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -194,6 +194,7 @@ services:
       - TEST_REDIS_HOST=redis
       - TEST_REDIS_OLD_HOST=redis_old
       - TEST_REDIS_OLD_PORT=6379
+      - DDTRACE_CI=true
     stdin_open: true
     tty: true
     volumes:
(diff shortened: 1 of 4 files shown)


ivoanjo:
> Wondering if we should add CI testing that just attempts to build the profiling extension against a matrix of compilers (clang + gcc + versions).

It's an interesting idea... The hard part would be building docker images for all the variants + Ruby VM builds, because the extension needs a built with Ruby as well.

Maybe in the future once we get a bit more resources; for now I think it's more important to focus our testing improvements on Ruby app coverage (e.g. relenv).

PR #2440, opened 2022-11-29
Fix sidekiq spec

TonyCTHsu:
**What does this PR do?**

* The filename suffix should be `_spec.rb` to be picked up by test suite.


ivoanjo (approved):
👍 Nice catch!

PR #2441, opened 2022-11-29
`exit` during sidekiq test

TonyCTHsu:
**What does this PR do?**

The method `run_sidekiq_server` in this example would prematurely exit the program, because of the ordering of the spec/examples causing randomness of coverage.

I found this out by running `bundle exec appraisal ruby-3.0.4-contrib rake spec:sidekiq` multiple times, different number of examples being tested each time.

```
  def run_sidekiq_server
    app_tempfile = Tempfile.new(['sidekiq-server-app', '.rb'])

    require 'sidekiq/cli'

    configure_sidekiq

    cli = Sidekiq::CLI.instance
    cli.parse(['--require', app_tempfile.path]) # boot the "app"
    launcher = Sidekiq::Launcher.new(cli.send(:options))
    launcher.stop
    exit
  ensure
    app_tempfile.close
    app_tempfile.unlink
  end
  ```

### DIFF
--- a/spec/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat_spec.rb
+++ b/spec/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat_spec.rb
@@ -15,6 +15,8 @@ RSpec.describe 'Server internal tracer heartbeat' do
   end
 
   it 'traces the looping heartbeat' do
+    # TODO: Fix `expect_in_sidekiq_server`
+    skip('`expect_in_sidekiq_server` is giving false positive which is always green')
     expect_in_sidekiq_server do
       span = spans.find { |s| s.service == tracer.default_service && s.name == 'sidekiq.heartbeat' }
 
--- a/spec/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat_spec.rb
+++ b/spec/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat_spec.rb
@@ -15,8 +15,6 @@ RSpec.describe 'Server internal tracer heartbeat' do
   end
 
   it 'traces the looping heartbeat' do
-    # TODO: Fix `expect_in_sidekiq_server`
-    skip('`expect_in_sidekiq_server` is giving false positive which is always green')
     expect_in_sidekiq_server do
       span = spans.find { |s| s.service == tracer.default_service && s.name == 'sidekiq.heartbeat' }
 
--- a/spec/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat_spec.rb
+++ b/spec/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat_spec.rb
@@ -30,6 +30,8 @@ RSpec.describe 'Server internal tracer heartbeat' do
 
   context 'traces the stop command' do
     before do
+      skip('`exit` from `run_sidekiq_server` would terminate the program prematurely')
+
       run_sidekiq_server
     end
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Review by marcotc — APPROVED** [review]
The approach looks good! I think the CI failures are because we are not waiting for the tracer to flush spans before asserting on them.

**Comment by TonyCTHsu** [comment]
@marcotc I was not able to fix that. I also found that `expect_in_sidekiq_server` is giving false positive(always green). 

I am skipping those for now.

File: spec/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat_spec.rb [hunk 1]
```diff
@@ -15,6 +15,8 @@
   end
 
   it 'traces the looping heartbeat' do
+    # TODO: Fix `expect_in_sidekiq_server`
+    skip('`expect_in_sidekiq_server` is giving false positive which is always green')
     expect_in_sidekiq_server do
       span = spans.find { |s| s.service == tracer.default_service && s.name == 'sidekiq.heartbeat' }
```
ivoanjo, line 21: I'm curious -- was this always incorrect, or does it just not work on modern versions of sidekiq?

TonyCTHsu: @ivoanjo Nice catch!

I did a more detailed investigation against different sidekiq versions. And found that the false positive results are only detected with  `6.4.1`, the version locked in our `contrib`, but `6.5.8` and `5.2.10` are working fine. 

I somehow noticed that the one spec was failing for `6.5.8` when I put a cap on [this](https://github.com/DataDog/dd-trace-rb/blob/master/Appraisals#L1129), but this seems to be a test setup issue, instead of an issue of our instrumentation code .

```
  2) Server internal tracer heartbeat traces the stop command
     Failure/Error: launcher = Sidekiq::Launcher.new(cli.send(:options))

     NoMethodError:
       undefined method `options' for #<Sidekiq::CLI:0x0000aaaace9c8dc0>
       Did you mean?  option_parser
```
Since this error occurs in the `run_sidekiq_server` method which prematurely exit the program. I believe we should be upgrading the supported version to `6.5.8` and then work on supporting `7.x`

PS. Sidekiq refactor internal on `6.5.0`. https://github.com/mperham/sidekiq/blob/main/Changes.md#650

ivoanjo: Just to clarify, do the changes in 6.5 affect only our tests or also our production support of sidekiq?

marcotc (approved):
We still have work do on our Sidekiq testing, but the PR as is ensures the whole Sidekiq contrib CI runs to completion.

spec/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat_spec.rb
TonyCTHsu: I think it only affect our test suite

spec/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat_spec.rb
```diff
@@ -30,6 +30,8 @@
 
   context 'traces the stop command' do
     before do
+      skip('`exit` from `run_sidekiq_server` would terminate the program prematurely')
```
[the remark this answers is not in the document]
ivoanjo: On a second look, I wonder if it's worth adding a `raise "Don't use this helper it's broken"` to `#run_sidekiq_server` in `sidekiq/support/helper.rb` just so it doesn't get used elsewhere accidentally.

PR #2445, opened 2022-11-29
Remove ruby platform check for sorbet dependency

TonyCTHsu:
**What does this PR do?**

With the ruby platform check, the appraisal lockfile would always end up with difference. However, currently we are using a compatible version that could be installed on `x86` or `arm64-darwin21` from M1 Mac. So we could remove this conditional 

**How to test this**

With M1 Mac, `bundle install && bundle exec rake typecheck`

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -88,8 +88,8 @@ end
 # Sorbet releases almost daily, with new checks introduced that can make a
 # previously-passing codebase start failing. Thus, we need to lock to a specific
 # version and bump it from time to time.
-# Also, there's no support for windows
-if RUBY_VERSION >= '2.4.0' && (RUBY_PLATFORM =~ /^x86_64-(darwin|linux)/)
+# Also, there's no support for windows, but M1 Mac with `arm64-darwin21` architecture is doing fine
+if RUBY_VERSION >= '2.4.0'
   gem 'sorbet', '= 0.5.9672'
   gem 'spoom', '~> 1.1'
 end


marcotc:
Waiting on https://github.com/sorbet/sorbet/issues/4119 to be addressed. There's already a PR up for it, currently under review.

Gemfile
```diff
@@ -88,8 +88,8 @@ end
 # Sorbet releases almost daily, with new checks introduced that can make a
 # previously-passing codebase start failing. Thus, we need to lock to a specific
 # version and bump it from time to time.
-# Also, there's no support for windows
-if RUBY_VERSION >= '2.4.0' && (RUBY_PLATFORM =~ /^x86_64-(darwin|linux)/)
+# Also, there's no support for windows, but M1 Mac with `arm64-darwin21` architecture is doing fine
+if RUBY_VERSION >= '2.4.0'
```
ivoanjo, line 92: Note that there's two changes being made here: enabling it on arm64 for macOS **and** for Linux. Afaik it doesn't work for Linux /yet/, so I think we should leave that part of the check for now.

ivoanjo (commented):
> Waiting on https://github.com/sorbet/sorbet/issues/4119 to be addressed. There's already a PR up for it, currently under review.

Note that upstream seems very slow on this issue, so I don't think we can expect a fix anytime soon 😰

lloeki:
> With the ruby platform check, the appraisal lockfile would always end up with difference

Important note: the locks that needs to be commited to work under aarch64-linux MUST NOT have sorbet, as this prevents `bundle lock --lockfile=ruby_x.y.z_foo.lock --add-platform aarch64-linux` from working.

As I said to @marcotc the other day, conditionals in `Gemfile`s are bad as they are evaluated on the currently run `RUBY_PLATFORM` + `RUBY_VERSION`, which has terrible effects on the descriptiveness of the DSL preventing proper cross platform + cross version usage of bundler.

I have a tentative change locally to address the `Gemfile` conditionals, making it entirely descriptive but didn't push a PR yet because I have a path issue to deal with when using the `appraisal` command.

lloeki:
> Note that upstream seems very slow on this issue, so I don't think we can expect a fix anytime soon

It's been working on a fork for something like a year. It's clearly just not their priority.

PR #2449, opened 2022-11-30
Add static type checking via RBS + Steep

lloeki:
**What does this PR do?**

Add static type checking via RBS + Steep.

**Motivation**

Adding typing to dd-trace-rb has been discussed within the Ruby Guild, and initially attempted with Sorbet.

Unfortunately due to its focus coming from its Stripe ownership, Sorbet's scope and design is to typecheck applications, which doesn't lend itself well to libraries, and has technical limitations such as requiring possibly insecure side-effectful source evaluation or not being compatible with aarch64.

Sorbet's value is also limited without manual type annotations.

Static typing libddwaf-rb with RBS and Steep has found a dozen of edge-case issues, two of which would have had dramatic impact, proving the usage is promising at a larger scale.

This is therefore attempted with RBS syntax and Steep type checker, both projects being backed by the Ruby core team.

**Additional Notes**

This was prototyped during initial AppSec development, then removed by 8d05385654daa6b572fa180bd9fb538f9382c614. Lots has changed since then though.

**How to test the change?**

```
bundle exec steep check
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -84,18 +84,20 @@ if RUBY_PLATFORM != 'java'
   end
 end
 
-# For type checking
-# Sorbet releases almost daily, with new checks introduced that can make a
-# previously-passing codebase start failing. Thus, we need to lock to a specific
-# version and bump it from time to time.
-# Also, there's no support for windows
-if RUBY_VERSION >= '2.4.0' && (RUBY_PLATFORM =~ /^x86_64-(darwin|linux)/)
-  gem 'sorbet', '= 0.5.9672'
-  gem 'spoom', '~> 1.1'
-end
+group :check do
+  # For type checking
+  # Sorbet releases almost daily, with new checks introduced that can make a
+  # previously-passing codebase start failing. Thus, we need to lock to a specific
+  # version and bump it from time to time.
+  # Also, there's no support for windows
+  if RUBY_VERSION >= '2.4.0' && (RUBY_PLATFORM =~ /^x86_64-(darwin|linux)/)
+    gem 'sorbet', '= 0.5.9672'
+    gem 'spoom', '~> 1.1'
+  end
 
-# type checking with steep
-if RUBY_VERSION >= '2.6.0' && RUBY_PLATFORM != 'java'
-  gem 'rbs', '~> 2.8.1', require: false
-  gem 'steep', '~> 1.3.0', require: false
+  # type checking with steep
+  if RUBY_VERSION >= '2.6.0' && RUBY_PLATFORM != 'java'
+    gem 'rbs', '~> 2.8.1', require: false
+    gem 'steep', '~> 1.3.0', require: false
+  end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -93,3 +93,9 @@ if RUBY_VERSION >= '2.4.0' && (RUBY_PLATFORM =~ /^x86_64-(darwin|linux)/)
   gem 'sorbet', '= 0.5.9672'
   gem 'spoom', '~> 1.1'
 end
+
+# type checking with steep
+if RUBY_VERSION >= '2.6.0' && RUBY_PLATFORM != 'java'
+  gem 'rbs', '~> 2.8.1', require: false
+  gem 'steep', '~> 1.3.0', require: false
+end
--- a/docs/StaticTypingGuide.md
+++ b/docs/StaticTypingGuide.md
@@ -1,0 +1,193 @@
+# Stating Typing Guide
+
+Static typing description is achieved via Ruby core [RBS](https://github.com/ruby/rbs).
+
+Static type checking is achieved via [Steep](https://github.com/soutaro/steep).
+
+## Quick start
+
+
+### Run the type check
+
+```
+bundle exec steep check [sources]
+```
+
+The `sources` arguments are optional and used to scope type checking to a smaller set of files or directories.
+
+### Generate type signature skeleton
+
+```
+bundle exec rbs prototype rb [source files]
+```
+
+Outputs `.rbs` content on stdout. The `source files` arguments lists the files from which the skeleton will be statically built from parsing. No evaluation occurs, therefore this has limited typing analysis capability, typically resulting in a lot of `untyped`.
+
+Note: Comments are reproduced as is which is useful for a visual check but should be manually removed when moving the skeleton to a `.rbs` file, because of the duplication and risk of comments getting desynced.
+
+## Layout
+
+Ruby code in `.rb` files are, as is customary for a gem, stored in `lib`.
+
+While RBS type annotations could be put inline, this creates a lot of noise, hampering "pure Ruby" readability. While the closeness with the code itself may look like an advantage, such annotations live in comments, which are harder to read and mix with other comments.
+
+RBS types can be described in any number of `.rbs` files, stored in `sig`. These files can be generated, syntax highlighted, checked, linted, and more. This is therefore the chosen approach.
+
+While the presence of `.rb` and `.rbs` files is entirely decoupled, here we choose to have one `.rbs` file per `.rb` file, mirroring the `lib` structure in `sig`. This has a number of advantages such as tracking typing progress, noticing stale files, generating new files without messing with existing type information, configuring IDEs and editors to jump from source to signature and back...
+
+Tools such as `rbs prototype` output comments. These should be removed, and only comments relevant to typing should end up in `.rbs` files.
+
+## Progressive typing
+
+Similar to many other Ruby tools, Steep reads project configuration from a DSL in `Steepfile`. We will use that to allow progressive typing.
+
+### Type checking vs signature loading
+
+Steep distinguishes between loading signatures and actually checking code for signatures. This is extremely useful to progressively type code, limiting check scope while still being able to provide signatures to code that can't be fully checked yet.
+
+```
+target :default do
+  signature "sig"           # ALL signatures from this directory will be loaded
+
+  check "lib/foo/bar"       # ONLY this source code folder will be checked against, using ALL signatures above
+  ignore "lib/foo/bar/baz"  # EXCEPT this subfolder
+end
+```
+
+### Dependency signatures
+
+Steep starts with a [minimal core](https://github.com/ruby/rbs/tree/master/core) loaded type signatures. Adding more [types from the Ruby stdlib](https://github.com/ruby/rbs/tree/master/stdlib) should be done progressively as required:
+
+```
+  library "set" # adds typing for Ruby stdlib's Set
+```
+
+Note: These signatures are part of [`rbs`](), which is included in Ruby releases since Ruby 3.0.
+
+Gems can embed a `sig` directory, which can be used directly:
+
+```
+  library "some_gem_with_a_sig_dir"
+```
+
+Some gems don't have typing information.
+
+In addition, a [vast collection of gems](https://github.com/ruby/gem_rbs_collection) have been typed. These can be fetched via a Rubygems/Bundler-like feature of RBS called [collections](https://github.com/ruby/rbs/blob/e91be7275f4005b1aeac8eadc2faa2b4ad5fdfef/docs/collection.md)
+
+```
+  collection_config "rbs_collection.steep.yaml"
+```
+
+This yaml file is akin to a Gemfile, describes the sources and gem signatures to fetch, and also has a lockfile mechanism. It can also integrate with `bundler` to match the signatures with the gem versions in use.
+
+Otherwise signatures can be vendored:
+
+```
+  repo_path "vendor/rbs"
+  library "subdir"
+```
+
+Typically these are be written as needed for gems entirely missing signatures, and ideally contributed back either upstream to the gem project itself or to the gem rbs collection project.
+
+### Measuring progress
+
+With the described layout and 1:1 match, it becomes easy to track coarse-grained coverage, additions, removals, changes through refactorings, in a similar way as is usually done with unit tests or specs.
+
+In addition, to output typing detailed coverage statistics:
+
+```
+bundle exec steep stats
+```
+
+## Typing a file
+
+### Basics
+
+To type a `.rb` file without a matching `.rbs` file, start with the skeleton:
+
+```
+mkdir -p sig/foo
+bundle exec rbs prototype rb lib/foo/bar.rb > sig/foo/bar.rbs
+```
+
+One can then proceed to [adjusting the signatures](https://github.com/ruby/rbs/blob/e91be7275f4005b1aeac8eadc2faa2b4ad5fdfef/docs/syntax.md) ([by example](https://github.com/ruby/rbs/blob/e91be7275f4005b1aeac8eadc2faa2b4ad5fdfef/docs/rbs_by_example.md)), removing as much `untyped` as possible.
+
+### Type profiling
+
+To discover types, one can leverage [`typeprof`](https://github.com/ruby/typeprof). Contrary to `rbs prototype rb` which relies solely on static parsing, `typeprof` is a Ruby interpreter, except it doesn't *execute* Ruby code, merely evaluates it to track types. Entry point calls to explore the various codepaths are required.
+
+With this file:
+
+```
+# test.rb
+def foo(x)
+  p x        # reveal type of x
+
+  if x > 10
+    x.to_s
+  else
+    nil
+  end
+end
+
+foo(42)      # this call is needed otherwise there's nothing evaluated!
+foo(3)       # make sure to explore as many codepaths as possible to get best coverage
+```
+
+The following is evaluated:
+
+```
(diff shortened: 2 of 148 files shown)


ivoanjo:
This looks nice, thanks for taking a stab at it!

I see our spec related to "things we're shipping" needs to be updated -- is your plan to ship the sigs to customers with the first PR, or do that later?

If we're shipping this to customers, we should probably integrate steep into the usual CI test cycle, to make sure we don't regress.

If, on the other hand, we plan on working on this a bit before that, are there any downsides in merging this PR almost as-is?

marcotc:
I'm more than happy to support merging this as well, and helping with any tasks that ease our adoption of this static typing.

Echoing what @ivoanjo said, if we want to break it down, merging the foundation first then slowly working towards a fully fledged typing suite (for both ourselves and users), I'm more than happy to support incremental work even if there's no immediate value in each individual PR until the very end.

lloeki:
Fixed some more typing. My first goal will be to have `lib/datadog/appsec` fully typed. Still some work to do though, some of which will require refactoring.

Also, added a simple CI thing.

lloeki:
Ha, and Sorbet gets hopelessly confused by some of these syntax things xD

```
sorbet/rbi/hidden-definitions/hidden.rbi:3414: Redefining constant Datadog::AppSec::Configuration::DSL::Instrument https://srb.help/4012
    3414 |class Datadog::AppSec::Configuration::DSL::Instrument
    3415 |  def self.[](*arg); end
    3416 |
    3417 |  def self.members(); end
    3418 |end
    lib/datadog/appsec/configuration.rb:18: Previous definition
    18 |        Instrument = _ = Struct.new(:name, :options)
                ^^^^^^^^^^

sorbet/rbi/hidden-definitions/hidden.rbi:3420: Redefining constant Datadog::AppSec::Configuration::Settings::Integration https://srb.help/4012
    3420 |class Datadog::AppSec::Configuration::Settings::Integration
    3421 |  def self.[](*arg); end
    3422 |
    3423 |  def self.members(); end
    3424 |end
    lib/datadog/appsec/configuration/settings.rb:117: Previous definition
     117 |        Integration = _ = Struct.new(:integration, :options)
                  ^^^^^^^^^^^

lib/datadog/appsec/configuration.rb:28: Wrong number of arguments for constructor. Expected: 0, got: 2 https://srb.help/7004
    28 |          @instruments << Instrument.new(name, options)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Errors: 3
```

lib/datadog/appsec/configuration.rb
```diff
@@ -14,7 +14,9 @@ def self.included(base)
 
       # Configuration DSL implementation
       class DSL
-        Instrument = Struct.new(:name, :options)
+        # Struct constant whisker cast for Steep
```
marcotc, line 17: What's is this `whisker` pattern here for Steep?

marcotc:
@lloeki, so we are aware how to work with this new tool, could you add a bit of documentation under https://github.com/DataDog/dd-trace-rb/tree/master/docs for how to use RBS/Steep in `ddtrace`?
Basic things like "how to type my new .rb file" would be nice, so we know how to expand type coverage.

lloeki:
Sure @marcotc! Let me fix the remaining non-typing issues so that it's not technically blocked and I will add something.

lib/datadog/appsec/configuration.rb
lloeki: Assigning to `_` swallows typing information so that `_ = foo` always works. As a consequence `bar = _ = foo` acts as a forceful cast, ignoring the type of `foo` when assigning to `bar`.

This pattern should not be used lightly but is sometimes necessary. In this case it's because `Struct.new` returns a `Struct[untyped]` but here we want the `Instrument` constant to be of type `Instrument`. [See here](https://github.com/ruby/rbs/issues/627#issuecomment-1335085271) for details.

(The name comes from the looks of this assignment: `= _ =` looks like cat whiskers)

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -126,37 +126,45 @@ def initialize
         end
 
         def enabled
-          @options[:enabled]
+          # Cast for Steep
+          _ = @options[:enabled]
```
marcotc, line 130: @lloeki is there no cleaner option for `_ = ...`? I ask because this assignment is not actually needed, and quite distracting.
I dream of a `@options[:enabled]` # steep:cast` or similar comment.

lloeki: Sadly no: types appear to be assigned to each language construct + each value in a way that "pass through" many things, except `_`. So `@options` (and `@options[:enable]`) have consistent types throughout the codebase: there's no form of locality to typing.

Therefore the only way to cast is to go through an assignment to have a second type in there, which can normally go on the left hand side only to a type encompassing the one on the right hand side. To move from a non-encompassing type it needs to go through the special-typed `_` on the left hand side of the assignment, which "swallows" the type information. The new type is then "forced back" by either assignment or return type.

Even the following annotation does not work:

```
        def enabled
          # @type var _cast: untyped
          _cast = @options[:enabled]
        end
```

Indeed the `::Datadog::AppSec::Configuration::DSL::option` type still carries through:

```
lib/datadog/appsec/configuration/settings.rb:128:12: [error] Cannot allow method body have type `::Datadog::AppSec::Configuration::DSL::option` because declared as type `bool`
│   ::Datadog::AppSec::Configuration::DSL::option <: bool
│     ::Datadog::AppSec::Configuration::DSL::option <: (true | false)
│       (::Integer | bool | ::Symbol | ::String | ::StringIO | ::File | ::Regexp | ::Hash[(::Symbol | ::String), ::Datadog::AppSec::Configuration::DSL::option] | ::Array[::Datadog::AppSec::Configuration::DSL::option]) <: (true | false)
│         ::Integer <: (true | false)
│           ::Integer <: true
│
│ Diagnostic ID: Ruby::MethodBodyTypeMismatch
│
└         def enabled
              ~~~~~~~
```

As seen above, `@options[:enabled]` is a value of the `@options` hash, which has a very wide type to be able to store all sorts of types for all possible option keys, but we want the return type of each option to be stricter than the general case.

The only way around that is to have a fully typed, non-generic store, e.g a struct or an actual class instead of a dumb hash with dumb values.

lloeki:
@marcotc @ivoanjo I just added a guide! LMKWYT

I think we should be all good now.

lloeki:
> is your plan to ship the sigs to customers with the first PR, or do that later?

Later. Signatures will start being directly useful for customers when we type the public API as it will e.g provide autocompletion as well as argument and return value type checking.

There are two complementary ways to approach progressive typing:

- start with precise typing of leafs (types that have little to no type dependencies) and go up from there
- start with precise typing of public API endpoints and go down from there

marcotc (approved):
The `docs/StaticTypingGuide.md` is very good! Thank you!
Let's give this a shot!

Gemfile
```diff
@@ -84,12 +84,20 @@ if RUBY_PLATFORM != 'java'
...
+  # previously-passing codebase start failing. Thus, we need to lock to a specific
+  # version and bump it from time to time.
+  # Also, there's no support for windows
+  if RUBY_VERSION >= '2.4.0' && (RUBY_PLATFORM =~ /^x86_64-(darwin|linux)/)
+    gem 'sorbet', '= 0.5.9672'
+    gem 'spoom', '~> 1.1'
+  end
+
+  # type checking with steep
+  if RUBY_VERSION >= '2.6.0' && RUBY_PLATFORM != 'java'
```
ivoanjo, line 99: Just out of curiosity -- does steep not run on JRuby?

docs/StaticTypingGuide.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Generate type signature skeleton
+
+```
+bundle exec rbs prototype rb [source files]
+```
+
+Outputs `.rbs` content on stdout. The `source files` arguments lists the files from which the skeleton will be statically built from parsing. No evaluation occurs, therefore this has limited typing analysis capability, typically resulting in a lot of `untyped`.
+
+Note: Comments are reproduced as is which is useful for a visual check but should be manually removed when moving the skeleton to a `.rbs` file, because of the duplication and risk of comments getting desynced.
```
ivoanjo, line 26: :+1: Thanks for calling this out, it's pretty important!

docs/StaticTypingGuide.md
```diff
@@ -0,0 +1,193 @@
...
+bundle exec rbs prototype rb lib/foo/bar.rb > sig/foo/bar.rbs
+```
+
+One can then proceed to [adjusting the signatures](https://github.com/ruby/rbs/blob/e91be7275f4005b1aeac8eadc2faa2b4ad5fdfef/docs/syntax.md) ([by example](https://github.com/ruby/rbs/blob/e91be7275f4005b1aeac8eadc2faa2b4ad5fdfef/docs/rbs_by_example.md)), removing as much `untyped` as possible.
+
+### Type profiling
+
+To discover types, one can leverage [`typeprof`](https://github.com/ruby/typeprof). Contrary to `rbs prototype rb` which relies solely on static parsing, `typeprof` is a Ruby interpreter, except it doesn't *execute* Ruby code, merely evaluates it to track types. Entry point calls to explore the various codepaths are required.
+
+With this file:
```
ivoanjo, line 119: Do you know if there is any tool that gets types based on actual runs, not just abstract runs?

I'm thinking it'd be cool to have something like that to help me type the profiler native stuff. Otherwise, I'll just do it manually -- it's a bit of work but I don't have particularly complex types going in/out of native anyway.

lib/datadog/appsec/configuration.rb
ivoanjo: To be honest, this file is so big that I would just suggest just making it a regular `class` and not need any particular weird typing things :)

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -8,7 +8,8 @@ module Configuration
       class Settings
         class << self
           def boolean
-            lambda do |v|
+            # @type ^(::String) -> bool
```
ivoanjo, line 11: I'm curious -- why add the types inline here and not on an rbs?

lib/datadog/appsec/configuration/settings.rb
ivoanjo: Oh wow, I'm kinda with Marco, this is kinda distracting.

I don't think it's a blocker, but here's two suggestions:
* Replace the options actual instance variables. We already need quite a lot of boilerplate to access them inside options, so perhaps making them instance variables would reduce the boilerplate and make steep happy as a side-effect
* Remove the `# Cast for steep` (or maybe just leave the first one). It shows up a bit too often, so it really adds a lot of noise. And if someone thinks this is not needed and removes is, they'll quickly get reminded once they push to CI so it's not like they'll go far without learning it anyway :)

lib/datadog/appsec/extensions.rb
```diff
@@ -112,6 +114,14 @@ def trace_rate_limit
           @settings.trace_rate_limit
         end
 
+        def obfuscator_key_regex
+          @settings.obfuscator_key_regex
+        end
+
+        def obfuscator_value_regex
+          @settings.obfuscator_key_regex
+        end
```
ivoanjo, line 123: Wait, is this change (and a few others on this file) a new feature? It didn't quite look like this existed before.

sig/datadog/appsec/contrib/auto_instrument.rbs
```diff
@@ -0,0 +1,11 @@
+module Datadog
+  module AppSec
+    module Contrib
+      # Auto-instrumentation for security integrations
+      # TODO: this implementation is trivial, check for shareable code with tracer
```
ivoanjo, line 5: Leftover comment ;)

Should we maybe have it part of the CI check that there are no comments here, or is that a bit too agressive?

ivoanjo: (There's a few other .rbs files with comments as well)

sig/datadog/appsec/contrib/rails/ext.rbs
```diff
@@ -0,0 +1,14 @@
+module Datadog
+  module AppSec
+    module Contrib
+      module Rails
+        # Rack integration constants
+        module Ext
+          APP: "rails"
+
+          ENV_ENABLED: "DD_TRACE_RAILS_ENABLED"
+        end
```
ivoanjo, line 10: Does steep complain if we change these constants (especially given we're not checking this folder), or do we need to remember to keep them up-to-date?

sig/datadog/appsec/waf.rbs
```diff
@@ -0,0 +1,230 @@
+module Datadog
+  module AppSec
+    module WAF
+      module LibDDWAF
```
ivoanjo, line 4: Should this perhaps go inside libddwaf directly?

lib/datadog/appsec/utils/http/media_range.rb
```diff
@@ -0,0 +1,201 @@
...
+require_relative 'media_type'
+
+module Datadog
+  module AppSec
+    module Utils
+      module HTTP
+        # Implementation of media range for content negotiation
+        class MediaRange
+          class ParseError < ::StandardError
+          end
```
ivoanjo, line 12: Wow this is a big addition just to support `Response#format`. I hesistate to ask but... is there any way we could we keep some of the simplicity of the old code?

ivoanjo (commented):
Github doesn't let me comment on empty files but there's a few:

* `sig/datadog/appsec/contrib/rails/integration.rbs`
* `sig/datadog/appsec/contrib/sinatra/integration.rbs`

-- is that expected?

Other than that, I gave it a quick pass and it looks reasonable.

Sorry for not having time to do a full pass, hopefully this still helps :)
On my side I see no reason not to merge this :+1:

Gemfile
lloeki: I'm honestly not sure. I don't think it matters either as it's static analysis, like Rubocop.

docs/StaticTypingGuide.md
lloeki: There's `rbs prototype runtime` but turns out I never actually used it so I'm not entirely sure.

It's important to note that these are helpful in understanding typing and bootstrapping types, but ultimately it has to be reviewed by a human.

lib/datadog/appsec/configuration.rb
lloeki: Yeah, I thought the same. I'll try to improve that on future quality iterations.

Overall I'll be trying to refactor all these things requiring syntax or comment directive oddities.

lib/datadog/appsec/extensions.rb
lloeki: It was actually an overlook, which typing pointed me to! So it's a consistency fix for an existing feature.

lib/datadog/appsec/utils/http/media_range.rb
lloeki: Yeah, I tried. The thing is, the old code was not correct WRT what real-life UAs might send.

So ultimately my choice was either to pile up some more hacks that may still be incomplete in face of whatever UAs do, or implement the real thing.

This is also on a feature that is implemented but currently gated by an undocumented, experimental setting, so the actual risk to the change here is low.

sig/datadog/appsec/contrib/auto_instrument.rbs
lloeki: I think we do want comments in `.rbs`, only related to typing, e.g to explain why this type or alias is that way, or TODOs, or... you get the point!

lloeki: (Good catch though)

sig/datadog/appsec/contrib/rails/ext.rbs
lloeki: I have to check but I think it does. This can also happen on method arguments and return values that have a known constant value or list of values possible.

When I wrote some of these by hand I put the type in instead of the value though. This is one of the things about which I'm not sure yet what would be best. e.g does Steep check for such static values at call sites and for method body returns, kind of like typed C/Go enums? I can see that being useful balanced against the value duplication.

sig/datadog/appsec/waf.rbs
lloeki: It actually exists in libddwaf, but its `sig` isn't currently in the gem. I went the simple way first, and plan to address that ultimately.

sig/datadog/appsec/contrib/auto_instrument.rbs
ivoanjo: Good point. I guess we can keep an eye out for "oops this comment should not be in" during reviews and at some point it should get pretty rare because we'll not be creating new `.rbs` files very often.

PR #2450, opened 2022-11-30
Add benchmark to test new profiler

ivoanjo:
**What does this PR do?**:

This PR adds a new stand alone benchmark `benchmarks/profiler_sample_loop_v2.rb` to test the new profiler.

The benchmark itself is a copy-pasted version of the original `benchmarks/profiler_sample_loop.rb` but adjusted to use the new profiler.

**Motivation**:

This benchmark is useful to test performance improvements on the sampling part of the profiler (as well as libdatadog -- see for instance <https://github.com/DataDog/libdatadog/pull/72>).

**Additional Notes**:

(None)

**How to test the change?**:

Run the benchmark! Also, I've updated the rspec test that runs it once during the regular test suite to ensure that it doesn't bitrot.

### DIFF
--- a/benchmarks/profiler_sample_loop_v2.rb
+++ b/benchmarks/profiler_sample_loop_v2.rb
@@ -1,0 +1,68 @@
+# typed: ignore
+
+# Used to quickly run benchmark under RSpec as part of the usual test suite, to validate it didn't bitrot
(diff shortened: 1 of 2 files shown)


marcotc (approved):
🏎️

PR #2451, opened 2022-11-30
Add W3C Trace Context propagator

marcotc:
This PR adds support trace propagation through the the "W3C Trace Context": https://www.w3.org/TR/trace-context/.

This standard is used by OpenTelemetry to propagate traces by default.

This PR adds the propagator class **but does not wire it yet**, as to make this PR smaller to review.

### DIFF
--- a/lib/datadog/tracing/sampling/priority_sampler.rb
+++ b/lib/datadog/tracing/sampling/priority_sampler.rb
@@ -85,6 +85,17 @@ module Datadog
           @priority_sampler.update(rate_by_service, decision: decision)
         end
 
+        # Check if the Priority Sampling decision is to keep or drop the trace.
+        # Other factors can influence the sampling decision; this method is only
+        # responsible for interpreting the Sampling Priority decision.
+        #
+        # @param priority_sampling [Integer] priority sampling number
+        # @return [Boolean] true if trace is "kept" by priority sampling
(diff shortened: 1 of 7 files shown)


lib/datadog/tracing/sampling/priority_sampler.rb
```diff
@@ -85,6 +85,17 @@ def update(rate_by_service, decision: nil)
...
+        # Check if the Priority Sampling decision is to keep or drop the trace.
+        # Other factors can influence the sampling decision; this method is only
+        # responsible for interpreting the Sampling Priority decision.
+        #
+        # @param priority_sampling [Integer] priority sampling number
+        # @return [Boolean] true if trace is "kept" by priority sampling
+        # @return [Boolean] false if trace is "dropped" by priority sampling
+        def self.sampled?(priority_sampling)
+          priority_sampling >= Ext::Priority::AUTO_KEEP
+        end
```
marcotc, line 97: I got tired of obscure sampling comparisons all over `ddtrace`, like `sampling_priority > AUTO_KEEP` or `sampling_priority <= AUTO_KEEP`. It's never clear what they mean.
I added this so we can start cleaning such checks all over `ddtrace` now.

TonyCTHsu (approved):

PR #2452, opened 2022-12-01
Fix unsafely raising exceptions without global VM lock in new profiler

ivoanjo:
**What does this PR do?**:

While working on other parts of the profiler, I observed that if the `serializer_flip_active_and_inactive_slots` function in `stack_recorder.c` tried to raise an exception, the Ruby VM could crash.

**Fortunately, the `StackRecorder` is only used on the new profiler codepath, which is still considered in alpha state and off by default**. Furthermore, the exceptions would only happen in cases that are not expected to ever happen in production.

The crash was caused by the fact that `serializer_flip_active_and_inactive_slots` is called without the Global VM Lock being held, and it's not valid to try to raise exceptions in this state.

To fix this, and because I still find raising exceptions to be useful in the native parts of the profiler, I've introduced two new helpers

* `grab_gvl_and_raise`
* `grab_gvl_and_raise_syserr`

that can be used to safely raise exceptions by making sure they grab the GVL before calling the Ruby exception-raising APIs.

**Motivation**:

The added helpers may seem slightly overkill as they're currently only used by the `StackRecorder`, but I plan to use them right away in the implementation of a new component that's living in a different branch.

**Additional Notes**:

(None)

**How to test the change?**:

Additionally, if you're curious to see the VM crashing, run the `stack_recorder_spec.rb` without this fix and add

```ruby
Thread.new do
  Object.new while true
end
```

to the very top of the spec file. This will usually trigger the crash because there will be two threads trying to allocate objects in the Ruby heap -- the added one, as well as the thread running `serializer_flip_active_and_inactive_slots`, which should set off the fireworks.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/profiling.c
+++ b/ext/ddtrace_profiling_native_extension/profiling.c
@@ -16,9 +16,9 @@ void http_transport_init(VALUE profiling_module);
 void stack_recorder_init(VALUE profiling_module);
 
 static VALUE native_working_p(VALUE self);
-static VALUE _native_grab_gvl_and_raise(DDTRACE_UNUSED VALUE _self, VALUE exception_class, VALUE test_message, VALUE test_message_arg);
+static VALUE _native_grab_gvl_and_raise(DDTRACE_UNUSED VALUE _self, VALUE exception_class, VALUE test_message, VALUE test_message_arg, VALUE release_gvl);
 static void *trigger_grab_gvl_and_raise(void *trigger_args);
-static VALUE _native_grab_gvl_and_raise_syserr(DDTRACE_UNUSED VALUE _self, VALUE syserr_errno, VALUE test_message, VALUE test_message_arg);
+static VALUE _native_grab_gvl_and_raise_syserr(DDTRACE_UNUSED VALUE _self, VALUE syserr_errno, VALUE test_message, VALUE test_message_arg, VALUE release_gvl);
 static void *trigger_grab_gvl_and_raise_syserr(void *trigger_args);
 static VALUE _native_ddtrace_rb_ractor_main_p(DDTRACE_UNUSED VALUE _self);
 static VALUE _native_is_current_thread_holding_the_gvl(DDTRACE_UNUSED VALUE _self);
@@ -46,8 +46,8 @@ void DDTRACE_EXPORT Init_ddtrace_profiling_native_extension(void) {
 
   // Hosts methods used for testing the native code using RSpec
   VALUE testing_module = rb_define_module_under(native_extension_module, "Testing");
-  rb_define_singleton_method(testing_module, "_native_grab_gvl_and_raise", _native_grab_gvl_and_raise, 3);
-  rb_define_singleton_method(testing_module, "_native_grab_gvl_and_raise_syserr", _native_grab_gvl_and_raise_syserr, 3);
+  rb_define_singleton_method(testing_module, "_native_grab_gvl_and_raise", _native_grab_gvl_and_raise, 4);
+  rb_define_singleton_method(testing_module, "_native_grab_gvl_and_raise_syserr", _native_grab_gvl_and_raise_syserr, 4);
   rb_define_singleton_method(testing_module, "_native_ddtrace_rb_ractor_main_p", _native_ddtrace_rb_ractor_main_p, 0);
   rb_define_singleton_method(testing_module, "_native_is_current_thread_holding_the_gvl", _native_is_current_thread_holding_the_gvl, 0);
   rb_define_singleton_method(
@@ -72,7 +72,7 @@ struct trigger_grab_gvl_and_raise_arguments {
   int test_message_arg;
 };
 
-static VALUE _native_grab_gvl_and_raise(DDTRACE_UNUSED VALUE _self, VALUE exception_class, VALUE test_message, VALUE test_message_arg) {
+static VALUE _native_grab_gvl_and_raise(DDTRACE_UNUSED VALUE _self, VALUE exception_class, VALUE test_message, VALUE test_message_arg, VALUE release_gvl) {
   ENFORCE_TYPE(test_message, T_STRING);
 
   struct trigger_grab_gvl_and_raise_arguments args;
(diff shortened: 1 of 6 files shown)


ivoanjo:
Update: Pushed rebase to fix conflict with current master

ext/ddtrace_profiling_native_extension/profiling.c
```diff
@@ -59,6 +66,70 @@ static VALUE native_working_p(DDTRACE_UNUSED VALUE _self) {
...
+
+  struct trigger_grab_gvl_and_raise_arguments args;
+
+  args.exception_class = exception_class;
+  args.test_message = StringValueCStr(test_message);
+  args.test_message_arg = test_message_arg != Qnil ? NUM2INT(test_message_arg) : -1;
+
+  rb_thread_call_without_gvl(trigger_grab_gvl_and_raise, &args, NULL, NULL);
+
+  rb_raise(rb_eRuntimeError, "This should never happen");
```
marcotc, line 86: If I'm reading this right and this should be unreachable, unless we messed up code in `trigger_grab_gvl_and_raise`,  I think the error could be more descriptive, similar to `  rb_bug("[DDTRACE] Unexpected: Reached the end of grab_gvl_and_raise while raising '%s'\n", args.exception_message);`.

ext/ddtrace_profiling_native_extension/profiling.c
```diff
@@ -59,6 +66,70 @@ static VALUE native_working_p(DDTRACE_UNUSED VALUE _self) {
...
+
+  struct trigger_grab_gvl_and_raise_syserr_arguments args;
+
+  args.syserr_errno = NUM2INT(syserr_errno);
+  args.test_message = StringValueCStr(test_message);
+  args.test_message_arg = test_message_arg != Qnil ? NUM2INT(test_message_arg) : -1;
+
+  rb_thread_call_without_gvl(trigger_grab_gvl_and_raise_syserr, &args, NULL, NULL);
+
+  rb_raise(rb_eRuntimeError, "This should never happen");
```
marcotc, line 118: Same as comment above for `_native_grab_gvl_and_raise` function.

ext/ddtrace_profiling_native_extension/ruby_helpers.c
```diff
@@ -22,3 +25,53 @@ void raise_unexpected_type(
...
+void grab_gvl_and_raise(VALUE exception_class, const char *format_string, ...) {
+  struct raise_arguments args;
+
+  args.exception_class = exception_class;
+
+  va_list format_string_arguments;
+  va_start(format_string_arguments, format_string);
+  vsnprintf(args.exception_message, MAX_RAISE_MESSAGE_SIZE, format_string, format_string_arguments);
+
+  rb_thread_call_with_gvl(trigger_raise, &args);
```
marcotc, line 50: A general question: is it possible that it is not valid to acquire the GVL here? Is there a failure scenario like "cannot acquired GVL because ..." or worse a crash (that's caused by `rb_thread_call_with_gvl` itself, not what's in `trigger_raise`).

marcotc (commented):
So annoying that you have to do everything twice because of `rb_raise` vs `rb_syserr_fail`.

ext/ddtrace_profiling_native_extension/profiling.c
ivoanjo: Since this is only for testing the code, do you think it's worth it?

Because this `_native_grab_gvl_and_raise` is directly called from Ruby code, the stack trace seems pretty self-explanatory as well:

```
  3) Datadog::Profiling::NativeExtension grab_gvl_and_raise_syserr raises an exception with the passed in message and errno
     Failure/Error:
       expect do
         described_class::Testing._native_grab_gvl_and_raise_syserr(Errno::EINTR::Errno, 'this is a test', nil)
       end.to raise_exception(Errno::EINTR, "#{Errno::EINTR.exception.message} - this is a test")

       expected Errno::EINTR with "Interrupted system call - this is a test", got #<RuntimeError: This should never happen> with backtrace:
         # ./spec/datadog/profiling/native_extension_spec.rb:186:in `_native_grab_gvl_and_raise_syserr'
         # ./spec/datadog/profiling/native_extension_spec.rb:186:in `block (4 levels) in <top (required)>'
         # ./spec/datadog/profiling/native_extension_spec.rb:185:in `block (3 levels) in <top (required)>'
         # ./spec/spec_helper.rb:216:in `block (2 levels) in <top (required)>'

#...

  6) Datadog::Profiling::NativeExtension grab_gvl_and_raise raises the requested exception with the passed in message
     Failure/Error:
       expect { described_class::Testing._native_grab_gvl_and_raise(ZeroDivisionError, 'this is a test', nil) }
         .to raise_exception(ZeroDivisionError, 'this is a test')

       expected ZeroDivisionError with "this is a test", got #<RuntimeError: This should never happen> with backtrace:
         # ./spec/datadog/profiling/native_extension_spec.rb:166:in `_native_grab_gvl_and_raise'
         # ./spec/datadog/profiling/native_extension_spec.rb:166:in `block (4 levels) in <top (required)>'
         # ./spec/datadog/profiling/native_extension_spec.rb:166:in `block (3 levels) in <top (required)>'
         # ./spec/spec_helper.rb:216:in `block (2 levels) in <top (required)>'
```

(Here I simulated the issue by commenting out the `rb_thread_call_without_gvl` lines on both functions).

ext/ddtrace_profiling_native_extension/ruby_helpers.c
ivoanjo: `rb_thread_call_with_gvl` I believe does the correct validations: https://github.com/ruby/ruby/blob/d0888f7784d2eec575d342e757f8ad29deb39b70/thread.c#L1756-L1779

...although because this is a really awkward error to have, the VM will chose to die in those situations; looking at the comments the Ruby developers seemed to be a bit unsure of what to do with a failure in that situation.

But you got me thinking that it's harmless for us to add an additional test to check if the helper is being misused and handle that situation without letting Ruby crash the VM. Added in a45ba5ded51799d994dd6bd72952f6a1a8db7353 .

The "not a Ruby thread" situation I don't think it's worth guarding against explicitly because a) we never create non-Ruby threads, b) we guard against running any code on the signal handlers and c) it's really hard to do something sane in that situation (because the caller is expecting to raise an exception and stop execution -- we can't just return to the caller and keep executing whatever else comes after the code that wanted to raise an exception...).

ivoanjo:
> So annoying that you have to do everything twice because of rb_raise vs rb_syserr_fail.

So much this! I've been finding it really hard to build abstractions and try to avoid repeating code in C :/

ext/ddtrace_profiling_native_extension/ruby_helpers.c
marcotc: I like the changes in https://github.com/DataDog/dd-trace-rb/commit/a45ba5ded51799d994dd6bd72952f6a1a8db7353 👍 👍 👍 .
I agree with not guarding against `rb_thread_call_with_gvl` being called by non-Ruby C threads.

ext/ddtrace_profiling_native_extension/profiling.c
ivoanjo: Ah, I see your point -- indeed the exception is written from the POV of someone reading the source, but if you just get it somehow, it's not clear what "this" is. Tweaked the messages in 831bf46de688f2853c636469957ab5c1b40ef044 .

PR #2457, opened 2022-12-02
Remove unused LanguageHelpers

marcotc:
I was browsing our code coverage report and noticed a low coverage at some of our test files: https://app.codecov.io/gh/DataDog/dd-trace-rb/tree/master/spec/support

Turns out a few of the are never used and don't seem very useful either.
There are quite a few more seemingly unused methods that are actually useful, so I didn't touch those.

This PR removes files and methods that I believe are not helpful to our test suite and are currently unused.


spec/support/language_helpers.rb
```diff
@@ -1,24 +0,0 @@
...
-      transform_keys(&:to_s)
-    end
-
-    def symbolize_keys
-      transform_keys(&:to_sym)
-    end
-  end
-end
-
-Hash.prepend(LanguageHelpers::HashHelpers)
```
marcotc, line 24: This is actually quite dangerous, as it masks the fact that Ruby < 2.5 does not have these methods.

spec/support/synchronization_helpers.rb
```diff
@@ -40,17 +40,6 @@ def expect_in_fork(fork_expectations: nil)
...
-    # Start in thread
-    t = Thread.new(&block)
-
-    # Wait for thread to finish, retrieve its return value.
-    status = t.value
-
-    # Expect thread and assertions to have completed successfully.
-    expect(status).to be true
-  end
-
```
marcotc, line 53: I added this while revamping the `synchronization_helpers`, but I was mistaken thinking that one would find it useful to assert on a Thread's return value like this, especially a simple `status == true` assertion.

The idea here was borrowed from `expect_in_fork`, but that one has a much clearer exit code to check on, and the separate memory space of forks makes that variant much more useful.

ivoanjo (approved):
👍 LGTM I love PRs removing unused stuff!

> There are quite a few more seemingly unused methods that are actually useful, so I didn't touch those.

To be honest, I'd probably still remove those. We can always resurrect them if needed, and having less code is quite helpful :)

PR #2458, opened 2022-12-02
Wire W3C propagator to HTTP & gRPC propagation

marcotc:
This PR adds support for the newly introduced [W3C Trace Context propagator](https://github.com/DataDog/dd-trace-rb/pull/2451).

This means that it is now possible to use the propagation style `tracecontext`, introduced in #2454, as a valid propagation style.


ivoanjo (approved):
This seems reasonable, although I haven't been following the overall change closely.

I grepped the codebase, and noticed that we usually also have the `PROPAGATION_STYLE_...` constants in `settings.rb` (+ `settings_spec.rb`) + all propagation styles in `spec/datadog/tracing/distributed/propagation_spec.rb`.

Do check if it makes sense to change those as well.

Otherwise, it LGTM 👍

marcotc:
@ivoanjo thanks for bringing out these two points.
The PR as it stands is correct `settings.rb` & `settings_spec.rb` would only change if W3C propagation because a default value, which is not the case yet.

And even though `propagation_spec.rb` currently has all other propagation styles, the testing there is only relevant for Datadog vs non-Datadog propagation styles, so there's no reason to introduce W3C there despite B3 being that file.

PR #2459, opened 2022-12-03
Move Tracing config settings from Core to Tracing

delner:
**What does this PR do?**

Moves Tracing specific configuration out of Core and to Tracing.

**Motivation**

The larger goal is to reduce Core's knowledge of Tracing components, thereby reducing coupling.

This particular pull request does not fundamentally change the dependency of Core upon Tracing, but it does at least move the code itself into the appropriate folder. The resulting changes exposes the dependency a bit more explicitly, and should make it easier to invert later.

**Additional Notes**

The settings themselves did not change, despite the large line count. `settings :tracing` merely moved into `tracing/configuration/settings.rb` instead. Consequently, so did the specs for these settings.

It's worth noting that there's at least one lingering configuration setting `agent.tracer.port`. This one likely constitutes a breaking change, but its ownership should also likely be moved to `Tracing`. Left this alone for now.

Follow up from #2453


spec/datadog/tracing/configuration/settings_spec.rb
```diff
@@ -0,0 +1,656 @@
...
+
+RSpec.describe Datadog::Tracing::Configuration::Settings do
+  # TODO: Core::Configuration::Settings directly extends Tracing::Configuration::Settings
+  #       In the future, have tracing add this behavior itself. For now,
+  #       just use the core metrics class to drive the tests.
+  subject(:settings) { Datadog::Core::Configuration::Settings.new(options) }
+
+  let(:options) { {} }
+
+  describe '#tracing' do
```
marcotc, line 22: These specs are now duplicated: one copy here, another in `spec/datadog/core/configuration/settings_spec.rb`.
They are pretty long hard to keep in sync.
I was recently making changes to `propagation_extract_style` & `propagation_inject_style` and would have to make duplicate version of the same test in both places after this PR.

Is the goal to remove these specs from `spec/datadog/core/configuration/settings_spec.rb` after the refactoring is done? If, should we do this now, as to avoid any possible divergence between this file and `spec/datadog/core/configuration/settings_spec.rb`?

delner: I messed up. Meant to delete them from `core/configuration/settings_spec.rb`. Should be fixed now.

marcotc (approved):

PR #2460, opened 2022-12-03
Extra testing from W3C spec

marcotc:
I looked at our internal RFC and found a few cases where we could process invalid `traceparent` headers:
1. Hex integers that are too large or too small are not allowed.
2. Trailing whitespace is allowed.

### DIFF
--- a/spec/datadog/tracing/distributed/trace_context_spec.rb
+++ b/spec/datadog/tracing/distributed/trace_context_spec.rb
@@ -268,6 +268,33 @@ RSpec.shared_examples 'Trace Context distributed format' do
 
     let(:digest) { extract }
 
+    context 'with traceparent fields with incorrect sizes' do
+      context 'version with incorrect size' do
+        let(:traceparent) { '0-00000000000000000000000000c0ffee-0000000000000bee-01' }
+        it { is_expected.to be_nil }
+      end
+
+      context 'trace_id with incorrect size' do
+        let(:traceparent) { '00-fee-0000000000000bee-01' }
+        it { is_expected.to be_nil }
+      end
+
+      context 'parent_id with incorrect size' do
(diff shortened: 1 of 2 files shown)


spec/datadog/tracing/distributed/trace_context_spec.rb
```diff
@@ -268,6 +268,33 @@
...
+      context 'trace flags with incorrect size' do
+        let(:traceparent) { '00-00000000000000000000000000c0ffee-0000000000000bee-0' }
+        it { is_expected.to be_nil }
+      end
+
+      context 'more fields than expected' do
+        let(:traceparent) { '00-00000000000000000000000000c0ffee-0000000000000bee-01-FFFFF' }
+        it { is_expected.to be_nil }
+      end
+    end
```
ivoanjo, line 296: Minor: More of a test style thing, but I would suggest perhaps structuring these tests as something like

```ruby
let(:traceparent) { "#{version}-#{trace_id}-#{parent_id}-#{trace_flags}" }

context 'when version has incorrect size' do
  let(:version) { "#{super()}123" }
  it { is_expected.to be_nil }
end

context 'when trace_id has incorrect size' do
  let(:trace_id) { "#{super()}123" }
```

Why? Because right now you kinda need to kinda need to look closely at the "correct" vs "incorrect" strings to figure out what changed, where with this approach it becomes (hopefully) more obvious which segment is being modified and what's being done to it.

ivoanjo (approved):
👍 LGTM

spec/datadog/tracing/distributed/trace_context_spec.rb
marcotc: Done, it looks better, thank you!

PR #2461, opened 2022-12-03
Semicolons not allwed in 'origin'

marcotc:
Recent spec change: `;` are not allowed characters in the `origin` propagation field. They are replaced with `_`, as are all other disallowed characters.


ivoanjo (approved):
👍 LGTM

PR #2464, opened 2022-12-05
SQL comment propagation full mode with traceparent

TonyCTHsu:
**What does this PR do?**

Implement SQL comment propagation with `full` mode, which includes `traceparent`

### DIFF
--- a/lib/datadog/tracing/contrib/propagation/sql_comment.rb
+++ b/lib/datadog/tracing/contrib/propagation/sql_comment.rb
@@ -17,11 +17,13 @@ module Datadog
             span_op.set_tag(Ext::TAG_DBM_TRACE_INJECTED, true) if mode.full?
           end
 
-          def self.prepend_comment(sql, digest, mode)
+          # Inject span_op and trace_op instead of TraceDigest to improve memory usage
+          # for `disabled` and `service` mode
+          def self.prepend_comment(sql, span_op, trace_op, mode)
             return sql unless mode.enabled?
 
             tags = {
-              Ext::KEY_DATABASE_SERVICE => digest.span_service,
+              Ext::KEY_DATABASE_SERVICE => span_op.service,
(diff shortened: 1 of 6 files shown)


lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -12,36 +14,30 @@ module SqlComment
           def self.annotate!(span_op, mode)
             return unless mode.enabled?
 
-            # PENDING: Until `traceparent`` implementation in `full` mode
-            # span_op.set_tag(Ext::TAG_DBM_TRACE_INJECTED, true) if mode.full?
+            span_op.set_tag(Ext::TAG_DBM_TRACE_INJECTED, true) if mode.full?
```
delner, line 17: What is this tag?

lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -12,36 +14,30 @@ module SqlComment
...
-              Ext::KEY_DATABASE_SERVICE => span_op.service,
+              Ext::KEY_DATABASE_SERVICE => digest.span_service,
               Ext::KEY_ENVIRONMENT => datadog_configuration.env,
               Ext::KEY_PARENT_SERVICE => datadog_configuration.service,
               Ext::KEY_VERSION => datadog_configuration.version
             }
 
-            # PENDING: Until `traceparent`` implementation in `full` mode
-            # tags.merge!(trace_context(span_op)) if mode.full?
+            if mode.full?
```
marcotc, line 30: The `trace_op.to_digest` is expensive'ish to create.
It's totally ok to call it when propagation is necessary, but I recommend not creating a digest if we are not going to use it for propagation. In other words, if `mode.full?` is false, I think we shouldn't call `trace_op.to_digest`, as this will make all traces slower, even the ones that don't have SQL comment propagation enabled.

marcotc (approved):

PR #2466, opened 2022-12-06
Add DD_TRACE_PROPAGATION_STYLE option

marcotc:
This PR as a new propagation style configuration: `DD_TRACE_PROPAGATION_STYLE`.

This setting sets the value of both `tracing.distributed_tracing.propagation_extract_style` & `tracing.distributed_tracing.propagation_inject_style` to the provided value. This is meant to ease custom configuration where both injection and extraction have the same format, which is common.

The existing options `DD_TRACE_PROPAGATION_STYLE_EXTRACT` & `DD_TRACE_PROPAGATION_STYLE_INJECT` have higher priority than the newly introduced `DD_TRACE_PROPAGATION_STYLE`, allowing clients be more specific when needed without changing the general `DD_TRACE_PROPAGATION_STYLE` setting.


lib/datadog/tracing/configuration/ext.rb
```diff
@@ -37,7 +37,7 @@ module Distributed
           # Sets both extract and inject propagation style tho the provided value.
           # Has lower precedence than `DD_TRACE_PROPAGATION_STYLE_INJECT` or
           # `DD_TRACE_PROPAGATION_STYLE_EXTRACT`.
-          ENV_PROPAGATION_STYLE_EXTRACT = 'DD_TRACE_PROPAGATION_STYLE'.freeze
+          ENV_PROPAGATION_STYLE = 'DD_TRACE_PROPAGATION_STYLE'.freeze
```
marcotc, line 40: I introduced this constant recently (a change still not released), but copy/pasted the constant name. This fixes that.

TonyCTHsu (approved):

PR #2467, opened 2022-12-06
Simplify system API error handling in profiler

ivoanjo:
**What does this PR do?**:

This PR introduces two new helpers in the profiler native code -- `ENFORCE_SUCCESS_GVL` and `ENFORCE_SUCCESS_NO_GVL` and updates existing code to use it.

**Motivation**:

Constantly checking for errors on system APIs such as `pthread_mutex_lock` is a lot of boiler plate that constantly breaks the flow of the code.

I decided to take a stab at improving this by introducing the new helpers.

**Additional Notes**:

The new helpers usually contain all of the information needed to tie the error back to where it came from in the native code. Here is one such example:

```
Errno::EINTR: Interrupted system call - Failure returned by 'syserr_errno' at
../../../../ext/ddtrace_profiling_native_extension/profiling.c:246:in `trigger_enforce_success'
```

Note that it includes:
* The errno string and description (these come from Ruby)
* The expression that caused the error (or variable containing it)
* The exact file, line, and function where it came from.

**How to test the change?**:

### DIFF
--- a/spec/datadog/profiling/native_extension_spec.rb
+++ b/spec/datadog/profiling/native_extension_spec.rb
@@ -317,8 +317,7 @@ RSpec.describe Datadog::Profiling::NativeExtension do
   describe 'enforce_success' do
     context 'when there is no error' do
       it 'does nothing' do
-        described_class::Testing._native_enforce_success(0, true)
-        # Nothing is raised
+        expect { described_class::Testing._native_enforce_success(0, true) }.to_not raise_error
       end
     end
 
--- a/spec/datadog/profiling/native_extension_spec.rb
+++ b/spec/datadog/profiling/native_extension_spec.rb
@@ -313,4 +313,30 @@ RSpec.describe Datadog::Profiling::NativeExtension do
       end
     end
   end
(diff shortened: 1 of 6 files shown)


spec/datadog/profiling/native_extension_spec.rb
```diff
@@ -313,4 +313,30 @@ def wait_for_thread_to_die
       end
     end
   end
+
+  describe 'enforce_success' do
+    context 'when there is no error' do
+      it 'does nothing' do
+        described_class::Testing._native_enforce_success(0, true)
+        # Nothing is raised
```
marcotc, line 321:
```suggestion
        expect { described_class::Testing._native_enforce_success(0, true) }.to_not raise_error
```

What do you think?

marcotc (approved):

spec/datadog/profiling/native_extension_spec.rb
ivoanjo: I like it, thanks for the suggestion! 👍

PR #2468, opened 2022-12-06
[PROF-6556] Fix new alpha profiler not being able to sample in periods of idleness

ivoanjo:
**What does this PR do?**:

This PR finishes the work started in #2415 and #2438.

In particular, #2415 ("[PROF-6556] Fix profiler incorrectly detecting if a thread is holding the Global VM Lock") opened a bit of a can of worms, because we learned (and fixed) that we were asking the VM to sample in situations where we don't think it's safe.

Furthermore, we also learned that we could not sample when the application was idle. This caused, among other things, several `CpuAndWallTimeWorker` specs to be flaky, because sometimes we got unlucky and could not sample/sample enough during the spec.

This PR actually makes us able to take samples in periods of idleness in the Ruby VM, and also solves a bunch of flakiness in specs that missing this feature introduced.

In a first approach, I thought having the `CpuAndWallTimeWorker` grab the global VM lock when it was free would be enough, but actually it wasn't. In some corner cases, because there is no "trylock" GVL, we could observe the GVL to be free, then decide to grab it, then once we go to grab it other thread already grabbed it and the `CpuAndWallTimeWorker` would then be at the mercy of the Ruby scheduler.

Instead, this PR introduces a new component -- the `IdleSamplingHelper` which keeps a background thread that fulfills requests to grab the GVL and sample coming from the `CpuAndWallTimeWorker`. This way, the `CpuAndWallTimeWorker` never blocks and is always free to set the pace for samples.

**Motivation**:

This PR makes sure that the new profiler is able to sample correctly applications that are fully or partially idle.

**Additional Notes**:

This PR is atop #2467 as makes use of the new helpers introduced in that PR.

Furthermore, this PR includes a few more small changes for issues that I ran into while developing it. **I recommend reviewing commit-by-commit**, and as always let me know if this PR is getting a bit too hard to follow and I can try breaking it down further.

**How to test the change?**:

This PR includes test coverage. Furthermore, applications that have idle periods (or are entirely idle) are now correctly and accurately profiled by the new profiler.

[PROF-6556]: https://datadoghq.atlassian.net/browse/PROF-6556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_idle_sampling_helper.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_idle_sampling_helper.c
@@ -1,0 +1,241 @@
+#include <ruby.h>
+#include <ruby/thread.h>
+#include <pthread.h>
+#include <stdbool.h>
+
+#include "helpers.h"
+#include "ruby_helpers.h"
+
+// Used by the Collectors::CpuAndWallTimeWorker to gather samples when the Ruby process is idle.
+//
+// Specifically, the IdleSamplingHelper is expected to be triggered by the CpuAndWallTimeWorker whenever it needs to
+// trigger a sample, but the VM is otherwise idle. See implementation of CpuAndWallTimeWorker for details.
+//
+// The IdleSamplingHelper keeps a background thread that waits for functions to run on a single-element "queue".
+// Other threads communicate with it by asking it to ACTION_RUN a `requested_action` or ACTION_STOP to terminate.
+//
+// The state is protected by the `wakeup_mutex`, and the background thread is woken up after changes using the
+// `wakeup` condition variable.
+
+typedef enum { ACTION_WAIT, ACTION_RUN, ACTION_STOP } action;
+
+// Contains state for a single CpuAndWallTimeWorker instance
+struct idle_sampling_loop_state {
+  pthread_mutex_t wakeup_mutex;
+  pthread_cond_t wakeup;
+  action requested_action;
+  void (*run_action_function)(void);
+};
+
+static VALUE _native_new(VALUE klass);
+static void reset_state(struct idle_sampling_loop_state *state);
+static VALUE _native_idle_sampling_loop(DDTRACE_UNUSED VALUE self, VALUE self_instance);
+static VALUE _native_stop(DDTRACE_UNUSED VALUE self, VALUE self_instance);
+static void *run_idle_sampling_loop(void *state_ptr);
+static void interrupt_idle_sampling_loop(void *state_ptr);
+static VALUE _native_reset(DDTRACE_UNUSED VALUE self, VALUE self_instance);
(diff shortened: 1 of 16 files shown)


ext/ddtrace_profiling_native_extension/collectors_idle_sampling_helper.c
```diff
@@ -0,0 +1,241 @@
...
+  int error = 0;
+
+  while (true) {
+    ENFORCE_SUCCESS_NO_GVL(pthread_mutex_lock(&state->wakeup_mutex));
+
+    action next_action;
+    void (*run_action_function)(void);
+
+    // Await for an action
+    while ((next_action = state->requested_action) == ACTION_WAIT) {
```
marcotc, line 126: I'm assuming it's correct to check for an `ACTION_WAIT` immediately on start? (while vs do-while kind of question.)

ext/ddtrace_profiling_native_extension/collectors_idle_sampling_helper.c
```diff
@@ -0,0 +1,241 @@
...
+  // handle because we get called by the VM in a situation where we can't really raise exceptions, and the VM really really
+  // just wants us to stop what we're doing and return control of the thread to it.
+  //
+  // So if we return immediately on error, we may leave the VM hanging because we didn't actually interrupt the thread.
+  // We're also not at a great location to flag errors.
+  // That's why: a) I chose to log to stderr, as a last-ditch effort; b) even if something goes wrong we still try to
+  // ask the thread to stop, instead of exiting early.
+
+  error = pthread_mutex_lock(&state->wakeup_mutex);
+  if (error) { fprintf(stderr, "[DDTRACE] Error during pthread_mutex_lock in interrupt_idle_sampling_loop (%s)\n", strerror(error)); }
```
marcotc, line 173: Are you using this `fprintf(stderr` error pattern frequently?
If you foresee this usage in the future, we can add another error checking macro.

marcotc: Nah, it's very specific to this case.

marcotc (approved):
Looks good overall! 👍

ext/ddtrace_profiling_native_extension/collectors_idle_sampling_helper.c
ivoanjo: Yup, I really don't want to make it a common thing, was a bit of a last-ditch effort here.

If it does become a common thing, I definitely will extract it and perhaps even make it slightly more configurable; this case was really the "I think it's important to flag when this happens, but there's no good way right now".

PR #2469, opened 2022-12-06
Changes `error.msg` to `error.message` for UNC

zhamza99:
What does this PR do?

Renames the `error.msg` tag to `error.message` tag in order to meet the requirements of naming across all tracers.


delner:
Seems simple enough! Just to be sure, does Datadog ingest this tag properly for errors? Just want to make sure its supported downstream before we change behavior.

marcotc (approved):

PR #2470, opened 2022-12-06
Improve redis integration patching

TonyCTHsu:
**What does this PR do?**

Until `redis` 4, all instrumentation happened in one gem: `redis`. Since `redis` 5, instrumentation happens in a separate gem: `redis-client`. Because `redis` 4 does not depend on `redis-client`, it's possible for both gems to be installed at the same time. 

For example, if `sidekiq` 7 and `redis` 4 are installed: both `redis` and `redis-client` will be installed.

This implementation improves the patching condition to apply to whenever it is possible.

PS. The minimun `redis-client` supported would be `0.11.0`, where `Config#custom` is implemented.

### DIFF
--- a/lib/datadog/tracing/contrib/redis/integration.rb
+++ b/lib/datadog/tracing/contrib/redis/integration.rb
@@ -14,19 +14,52 @@ module Datadog
 
           MINIMUM_VERSION = Gem::Version.new('3.2')
 
+          # Support `Config#custom`
+          # https://github.com/redis-rb/redis-client/blob/master/CHANGELOG.md#0110
+          REDISCLIENT_MINIMUM_VERSION = Gem::Version.new('0.11.0')
+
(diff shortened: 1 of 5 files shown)


lib/datadog/tracing/contrib/redis/integration.rb
```diff
@@ -14,19 +14,52 @@ class Integration
...
             !defined?(::Redis).nil?
           end
 
+          def self.redis_client_loaded?
+            !defined?(::RedisClient).nil?
+          end
+
           def self.compatible?
-            super && version >= MINIMUM_VERSION
+            redis_compatible? || redis_client_compatible?
```
marcotc, line 54:
```suggestion
            super && (redis_compatible? || redis_client_compatible?)
```
Missing the `super` call.

marcotc (approved):

PR #2471, opened 2022-12-06
Test Propagation class instead of styles directly

marcotc:
I noticed, while adding a new propagator style, that `spec/datadog/tracing/contrib/grpc/distributed/propagation_spec.rb` & `spec/datadog/tracing/contrib/http/distributed/propagation_spec.rb` were not failing tests when I accidentally removed existing propagator styles from their configuration.

This happened because those tests were testing the style classes directly, `Distributed::B3Multi.new(fetcher: Contrib::GRPC::Distributed::Fetcher)`, instead of through the Propagation class, `Contrib::GRPC::Distributed::Propagation.new`.

This PR makes those test cover the same test surface, but by using the respective `Propagation` class instead.

Now, there should be no coverage lacking in distributed tracing propagation.

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -2349,26 +2349,21 @@ Service C:
 
 Tracing supports the following distributed trace formats:
 
- - `Datadog::Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_DATADOG` (Default)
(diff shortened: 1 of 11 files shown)


ivoanjo (approved):
👍 LGTM

PR #2477, opened 2022-12-07
Fix sidekiq flaky spec

TonyCTHsu:
**What does this PR do?**

https://github.com/DataDog/dd-trace-rb/issues/2474

PR #2478, opened 2022-12-07
Allow try_wait_until to be configured with seconds

marcotc:
While adding a `try_wait_until` of 10 seconds to https://github.com/DataDog/dd-trace-rb/pull/2477 I noticed how opaque the configuration options are for `try_wait_until`.

This PR adds a hopefully simpler to reason about `seconds` configuration option.

Instead of configuring `try_wait_until(attempts: 100, backoff: 0.1)` we can now do `try_wait_until(seconds: 10)`: these two scenarios yield the same configuration to `try_wait_until`.

### DIFF
--- a/spec/datadog/core/workers/runtime_metrics_spec.rb
+++ b/spec/datadog/core/workers/runtime_metrics_spec.rb
@@ -344,7 +344,7 @@ RSpec.describe Datadog::Core::Workers::RuntimeMetrics do
 
               # Restart worker & wait
               worker.perform
-              try_wait_until(attempts: 30) { @flushed }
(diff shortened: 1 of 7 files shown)


ivoanjo (approved):
Hah, great catch! I swear I scrolled through the sidekiq PR and thought that I didn't quite like

```
      # 10 second wait time
      try_wait_until(attempts: 100) { wait_until.call } if wait_until
```

...because it was encoding knowledge from the helper which we may want to change at some point and could break the assumption there.

And then I keep scrolling on my inbox and found that you had fixed it already! 🙇 👍

PR #2479, opened 2022-12-08
Add debug information for flaky grape test

marcotc:
This is an old flaky test that doesn't happen super often but does happen still: https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/8259/workflows/4ad2adbf-5954-471a-bf16-6584d79e913d/jobs/307083/parallel-runs/21?filterBy=FAILED

I tried my best to debug it, but couldn't find out why. And I couldn't reproduce it either.

I'm instead then adding additional debug information in the event of a CI failure.


ivoanjo (approved):
👍 LGTM

PR #2485, opened 2022-12-08
W3C: Propagate unknown values as-is

marcotc:
Unknown Datadog-specific fields should be propagated as-is, as they might belong to future versions Datadog Tracers.

`ddtrace`'s job is to simply pass the them downstream with the same incoming values.

### DIFF
--- a/lib/datadog/tracing/distributed/trace_context.rb
+++ b/lib/datadog/tracing/distributed/trace_context.rb
@@ -43,7 +43,7 @@ module Datadog
 
           return unless trace_id # Could not parse traceparent
 
-          tracestate, sampling_priority, origin, tags = extract_tracestate(fetcher[@tracestate_key])
+          tracestate, sampling_priority, origin, tags, unknown_fields = extract_tracestate(fetcher[@tracestate_key])
 
           sampling_priority = parse_priority_sampling(sampled, sampling_priority)
 
@@ -54,8 +54,9 @@ module Datadog
             trace_sampling_priority: sampling_priority,
             trace_distributed_tags: tags,
             trace_distributed_id: trace_id,
+            trace_flags: trace_flags,
             trace_state: tracestate,
-            trace_flags: trace_flags
+            trace_state_unknown_fields: unknown_fields,
           )
         end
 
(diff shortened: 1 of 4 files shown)


spec/datadog/tracing/distributed/trace_context_spec.rb
```diff
@@ -242,7 +242,7 @@
             let(:upstream_tracestate) { 'dd=old_value,other=vendor,dd=oops_forgot_to_remove_this' }
 
             it 'removes existing `dd=` values, prepending new `dd=` value' do
-              expect(tracestate).to eq('dd=o:origin,other=vendor')
+              expect(tracestate).to eq('dd=o:origin;future=field,other=vendor')
             end
```
ivoanjo, line 246: I think you either forgot to update the `upstream_tracestate` or you changed the wrong `expect` here. (Tests are breaking)

ivoanjo (commented):
The change looks reasonable, although this is one of those PRs that it's probably best someone from tracing give a second pass as I understand the code changes, but I don't understand if they are the correct code changes for the outcome we want ;)

lib/datadog/tracing/distributed/trace_context.rb
```diff
@@ -299,10 +304,14 @@ def extract_datadog_fields(dd_tracestate)
...
               tags["#{Tracing::Metadata::Ext::Distributed::TAGS_PREFIX}#{key}"] = value
+            else
+              unknown_fields ||= String.new
+              unknown_fields << pair
+              unknown_fields << ';'
             end
           end
 
-          [origin, sampling_priority, tags]
+          [origin, sampling_priority, tags, unknown_fields]
```
TonyCTHsu, line 314: Perhaps we should extract this to its own object, if the array keeps on growing. It feels weird to me that the caller need to understand the index of the array to get the corresponding value.

TonyCTHsu (approved):
Looks great!

lib/datadog/tracing/distributed/trace_context.rb
marcotc: I'll do it if we get to 5! 🦆

spec/datadog/tracing/distributed/trace_context_spec.rb
marcotc: Thank you!

PR #2488, opened 2022-12-12
Update banner on 1.0 upgrade guide

marcotc:
As the 1.x ddtrace series has been released for quite some time, thus the phrasing `We've recently released the 1.x version series` is no longer accurate.

Also the banner was placed at the very top of our documentation pages, which I believe is not longer needed.

This PR moves the banner to the first actual documentation section and rephrases the "recently released" language.


ivoanjo (approved):
👍 LGTM

PR #2491, opened 2022-12-15
Bump to version 1.8.0

ivoanjo:
**What does this PR do?**:

This PR sets up the 1.8.0 release. It includes the version bump, the changelog update, and the usual gemfiles refresh.

**Motivation**:

Release 1.8.0 :)

**Additional Notes**:

(N/A)

**How to test the change?**:

No code changes in this PR! Check that CI is still green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ You can enable it:
 
 ```ruby
 Datadog.configure do |c|
-  # … existing configuration …
+  # ... existing configuration ...
 
   c.profiling.advanced.force_enable_new_profiler = true
 end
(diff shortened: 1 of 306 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,106 @@
...
 ## [Unreleased]
 
+## [1.8.0] - 2022-12-14
+
+Happy holidays! This release includes two big items we're pretty excited about and want to call out explicitly:
+
+* CPU Profiling 2.0 is now in beta!
+* `redis` 5 instrumentation upgrade notes
+
+### CPU Profiling 2.0 is now in beta!
```
lloeki, line 12: Maybe that's me but I kind of expected `CHANGELOG.md` to be very synthetic, and if a significant amount of text is required it would be present in other relevant documents and linked to from here. Maybe a file in `docs`? WDYT?

lloeki (approved):
Approved so as not to block the release, although I feel like the amount of text here crosses over the "summariness" threshold of a changelog.

CHANGELOG.md
ivoanjo: I hear your concern:

* I hesitate a bit about adding it to a doc since this is somewhat tied to 1.8.0 (e.g. we may want to update it for 1.9.0).
* So instead I just removed the detailed text and linked back to the release notes, which will also contain the same text.

Hopefully this is reasonable :)

lloeki: Makes sense!

PR #2495, opened 2022-12-19
128 bits trace ID configuration

TonyCTHsu:
**What does this PR do?**

To extend the current tracing system to support both the existing 64-bit and proposed 128-bit TraceIds in the tracer.

Three different environment variables to govern the configuration:

| Function | configuration | Environment variable | default |
| -------- | --------------| -------------------- | ------- |
| Generation | trace_id_128_bit_generation_enabled | `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED` | `false` |
| Logging | trace_id_128_bit_logging_enabled | `DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED` | `false` |
| Propagation | trace_id_128_bit_propagation_enabled | `DD_TRACE_128_BIT_TRACEID_PROPAGATION_ENABLED` | `true` |

Flag Sanity Checking
If `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED`=true and `DD_TRACE_128_BIT_TRACEID_PROPAGATION_ENABLED`=false, a warning should be logged as this configuration likely doesn't do what the application intends.


lib/datadog/tracing/configuration/ext.rb
```diff
@@ -7,6 +7,7 @@ module Configuration
       # e.g. Env vars, default values, enums, etc...
       module Ext
         ENV_ENABLED = 'DD_TRACE_ENABLED'.freeze
+        ENV_TRACE_ID_128_BIT_GENERATION_ENABLED = 'DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED'.freeze
```
marcotc, line 10: Are these very long environment variable names part of the RFC?

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -175,6 +176,44 @@ def self.extended(base)
...
+                o.default { env_to_bool(Tracing::Configuration::Ext::Distributed::ENV_TRACE_ID_128_BIT_PROPAGATION_ENABLED, false) }
+                o.lazy
+                o.depends_on [:trace_id_128_bit_generation_enabled]
+
+                o.on_set do |v|
+                  if get_option(:trace_id_128_bit_generation_enabled) && v != true
+                    Datadog.logger.warn(
+                      '128 bit trace_id is generated but not propagated. '\
+                      "Enable environment variable `#{Tracing::Configuration::Ext::Distributed::ENV_TRACE_ID_128_BIT_PROPAGATION_ENABLED}` with `true` "\
+                      "to enable option `#{o.instance_variable_get(:@name)}`"
```
marcotc, line 202:
```suggestion
                      "to enable propagation."
```

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -175,6 +176,44 @@ def self.extended(base)
...
+                    )
+                  end
+                end
+              end
+
+              # Enable 128 bit trace id injected for logging.
+              #
+              # @default `DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED` environment variable, otherwise `false`
+              # @return [Boolean]
+              option :trace_id_128_bit_logging_enabled do |o|
```
marcotc, line 212: Should we also emit a warning here when 128-bit is enabled but this is not?

TonyCTHsu: I think that make sense!

PR #2497, opened 2022-12-20
Set maximum supported Ruby version

marcotc:
Bump maximum Ruby version to the next, unreleased version 3.3 (`< 3.4` in practice).

This PR also changes the comment around the `MAXIMUM_RUBY_VERSION` constant in order to document why we limit it, and what version we limit it to.


gemfiles/ruby_3.2.0_contrib.gemfile.lock
```diff
@@ -8,6 +8,16 @@ GIT
       simplecov-html (~> 0.11)
       simplecov_json_formatter (~> 0.1)
 
+GIT
+  remote: https://github.com/toy..raisal
```
marcotc, line 12: Why is the remote abbreviated? I think this is causing the current test errors.
```
  remote: https://github.com/toy..raisal
```

lloeki:
From the [changelog](https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/):

Notable changes that may be impactful, sometimes silently (as in they may appear to work but displace/lose data):

# Delegation of args and kwargs

> `ruby2_keywords` is still there, but has a breaking change:
>
> Methods taking a rest parameter (like `*args`) and wishing to delegate keyword arguments through `foo(*args)` must now be marked with `ruby2_keywords` (if not already the case). In other words, all methods wishing to delegate keyword arguments through `*args` must now be marked with `ruby2_keywords`, with no exception. This will make it easier to transition to other ways of delegation once a library can require Ruby 3+. Previously, the `ruby2_keywords` flag was kept if the receiving method took `*args`, but this was a bug and an inconsistency. A good technique to find potentially missing `ruby2_keywords` is to run the test suite, find the last method which must receive keyword arguments for each place where the test suite fails, and use `puts nil, caller, nil` there. Then check that each method/block on the call chain which must delegate keywords is correctly marked with `ruby2_keywords`.
>
> ```
>   def target(**kw)
>   end
> 
>   # Accidentally worked without ruby2_keywords in Ruby 2.7-3.1, ruby2_keywords
>   # needed in 3.2+. Just like (*args, **kwargs) or (...) would be needed on
>   # both #foo and #bar when migrating away from ruby2_keywords.
>   ruby2_keywords def bar(*args)
>     target(*args)
>   end
> 
>   ruby2_keywords def foo(*args)
>     bar(*args)
>   end
> 
>   foo(k: 1)
> ```

We must make sure to have `ruby2_keywords` everywhere we expect `*args` to delegate both `*args` and `**kwargs`. Depending on code/specs things may appear to silently work but produce incorrect results.

Long term, in order to support both ruby <= 2.6, ruby 2.7, ruby 3.0+, and the future ruby version where `ruby2_keywords` is removed, we have to split such cases in two:

```
if RUBY_VERSION < '3.0'
  def foo(*args)
     bar(*args)
  end

  def bar(*args)
     target(*args)
  end
else
  def foo(*args, **kwargs)
     bar(*args, **kwargs)
  end

  def bar(*args, **kwargs)
     target(*args, **kwargs)
  end
end
```

# Proc arguments don't autosplat anymore

> A proc that accepts a single positional argument and keywords will no longer autosplat.
>
> ```
> proc{|a, **k| a}.call([1, 2])
> # Ruby 3.1 and before
> # => 1
> # Ruby 3.2 and after
> # => [1, 2]
> ```

We should check whether we have such cases. Depending on code/specs things may appear to silently work but produce incorrect results.

# Constant assignment evaluation order

> Constant assignment evaluation order for constants set on explicit objects has been made consistent with single attribute assignment evaluation order. With this code:
> 
> ```
> foo::BAR = baz
> ```
>
> `foo` is now called before `baz`

This doesn't appear to impact us: `rg '\W[^A-Z][a-z]+::' lib` returns no such assignment.

# `Hash#shift` behaviour on empty hash

> `Hash#shift` now always returns nil if the hash is empty, instead of returning the default value or calling the default proc

We don't appear to be using this method: `rg shift lib` returns few results, all on arrays.

# Removed methods

> The following deprecated methods are removed.
>
>    Dir.exists?
>    File.exists?
>    Kernel#=~
>    Kernel#taint, Kernel#untaint, Kernel#tainted?
>    Kernel#trust, Kernel#untrust, Kernel#untrusted?

We don't appear to be using any of these.

unicornzero:
I'm adding dd-trace to an app already on Ruby 3.2, running into issue #2534 would love to know when Ruby 3.2 support will be available.

ivoanjo:
There's quite a few things here -- I plan to separate this PR out a bit, and make sure we have Ruby 3.2 support for the next release of dd-trace-rb.

ivoanjo:
This PR is mostly replaced by https://github.com/DataDog/dd-trace-rb/pull/2601 . I'm leaving it open as we may still want to import the tweak to the ruby maximum version.

marcotc:
I've refreshed this PR, and now it tackles only the maximum version change.

ivoanjo:
This looks good to me, although I think there was some ongoing discussion internally on < 3.3 vs < 3.4 so I guess we should probably arrive to a conclusion.

Also -- can I annoy you to remove the old commits from the PR and keep only the latest change? ;)

marcotc:
> Also -- can I annoy you to remove the old commits from the PR and keep only the latest change? ;)

But then I can't inflate my commit stats in the repo!

ivoanjo:
> But then I can't inflate my commit stats in the repo!

Ha! Only I am allowed to do that xD

ivoanjo (approved):
:+1: LGTM and thanks for capturing our reasoning.

I agree it makes sense to double-check with Loic as you mentioned on slack, since he's also been very interested in this discussion so far :)

lloeki:
I feel like we should still keep a note about `ruby2_keywords` in the comments, but otherwise I'm OK with the change.

lloeki (approved):

PR #2498, opened 2022-12-20
Move Metrics ENV constants to Core::Configuration::Ext

delner:
**What does this PR do?**

Moves `ENV_DEFAULT_HOST` and `ENV_DEFAULT_PORT` in `Metrics` to `Configuration::Ext`

**Motivation**

I'd like configuration settings to be self-contained: they shouldn't need to load the code for the features they configure. This reduces some complexity (in particular, makes it more plausible to load "just" the configuration.)

**Additional Notes**

This does move some constants around. Kinda sorta a breaking change? No one should be references these constants directly; their values are unchanged, as is their behavior.


lib/datadog/core/metrics/ext.rb
```diff
@@ -7,8 +7,11 @@ module Metrics
       module Ext
         DEFAULT_HOST = '127.0.0.1'.freeze
         DEFAULT_PORT = 8125
-        ENV_DEFAULT_HOST = 'DD_AGENT_HOST'.freeze
-        ENV_DEFAULT_PORT = 'DD_METRIC_AGENT_PORT'.freeze
+
+        #
+        # TODO: REMOVE THIS COMMENT.
+        #       !!! DO NOT MERGE ME !!!
```
marcotc, line 13: 🎗️

marcotc (approved):

marcotc (approved):

PR #2500, opened 2022-12-20
128 bit trace id generation

TonyCTHsu:
**What does this PR do?**

The generation of new TraceIds will be controlled by the feature flag:
`DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED`, default value `false`

* When `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED` is false, new TraceIds should be generated using the current approach for generating 64-bit TraceIds, with the higher-order bits set to 0.
* When `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED` is true, new TraceIds should be generated using the current approach for generating 64-bit TraceIds to generate 2 64-bit TraceIds - one used for the lower-order 64 bits and the other for the higher-order 64 bits. Neither the lower-order or higher-order 64 bits should be zero. In addition, if `DD_TRACE_128_BIT_TRACEID_PROPAGATION_ENABLED` is true, set the trace tag `_dd.p.tid` to the hex encoding of the higher-order 64 bits.


spec/datadog/tracing/trace_operation_spec.rb
```diff
@@ -88,6 +88,68 @@
       it do
         expect(trace_op.send(:metrics)).to eq({})
       end
+
+      context 'when 128 bit trace id generation enabled' do
+        around do |example|
+          ClimateControl.modify(
+            'DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED' => 'true'
```
marcotc, line 95: You should mock `allow(Datadog.configuration.tracing).to receive(:trace_id_128_bit_generation_enabled).and_return(true)` here, given we don't actually depend on the environment variable, and the environment variable test is already done by the original PR that introduced `Datadog.configuration.tracing.trace_id_128_bit_generation_enabled`.

lib/datadog/tracing/trace_id_generation.rb
```diff
@@ -0,0 +1,46 @@
...
+
+# typed: ignore
+
+require_relative 'utils'
+require_relative 'metadata/tagging'
+require_relative 'metadata/ext'
+
+module Datadog
+  module Tracing
+    module TraceIdGeneration
```
marcotc, line 11: Is there any reason for this to not be directly in `TraceOperation` instead on its own module?

spec/datadog/tracing/trace_operation_spec.rb
```diff
@@ -88,6 +88,68 @@
...
+            700846779308066302, # high_order
+            230828285073546607, # low_order
+            729068622763299748, # Prevent rspec mock to return previous values
+          )
+
+          # high_order + low_order
+          expect(trace_op.id).to eq(12928341172779498062566790137492790639)
+
+          # 700846779308066302.to_s(16) => '9b9e86ebaa245fe'
+          expect(trace_op.get_tag('_dd.p.tid')).to eq('9b9e86ebaa245fe')
```
marcotc, line 116: Writing these numbers in hex or binary might make this easier to read, like:
```suggestion
          allow(Datadog::Tracing::Utils).to receive(:next_id).and_return(
            0x9b9e86ebaa245fe, # high_order
            0x334111a2febdd6f, # low_order
            0xa1e2c0cb4b093a4, # Prevent rspec mock to return previous values
          )

          # high_order + low_order
          expect(trace_op.id).to eq(0x9b9e86ebaa245fe0334111a2febdd6f)

          # 700846779308066302.to_s(16) => '9b9e86ebaa245fe'
          expect(trace_op.get_tag('_dd.p.tid')).to eq('9b9e86ebaa245fe')
```

marcotc: Or even, change the numbers to make the test even easier to read:
```suggestion
          allow(Datadog::Tracing::Utils).to receive(:next_id).and_return(
            0xfffffffffffffff, # high_order
            0xaaaaaaaaaaaaaaa, # low_order
            0xa1e2c0cb4b093a4, # Prevent rspec mock to return previous values
          )

          # high_order + low_order
          expect(trace_op.id).to eq(0xfffffffffffffffaaaaaaaaaaaaaaa)

          expect(trace_op.get_tag('_dd.p.tid')).to eq('fffffffffffffff')
```

spec/datadog/tracing/trace_operation_spec.rb
```diff
@@ -88,6 +88,68 @@
...
+          around do |example|
+            ClimateControl.modify(
+              'DD_TRACE_128_BIT_TRACEID_PROPAGATION_ENABLED' => 'false'
+            ) do
+              example.run
+            end
+          end
+
+          it do
+            expect(trace_op.has_tag?('_dd.p.tid')).to eq false
```
marcotc, line 129: Because the method is nicely named `has_tag?`, I think you can use the fancy RSpec syntax:
```suggestion
            expect(trace_op).to_not have_tag('_dd.p.tid')
```

spec/datadog/tracing/trace_operation_spec.rb
```diff
@@ -88,6 +88,68 @@
...
+        around do |example|
+          ClimateControl.modify(
+            'DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED' => 'false'
+          ) do
+            example.run
+          end
+        end
+
+        it '64 bits trace_id' do
+          expect(trace_op.id.to_s(2).length).to be <= 64
```
marcotc, line 144: I think [bit_length](https://ruby-doc.org/core-2.5.0/Integer.html#method-i-bit_length) can work here:
```suggestion
          expect(trace_op.id.bit_length).to be <= 64
```

spec/datadog/tracing/trace_operation_spec.rb
```diff
@@ -88,6 +88,68 @@
...
+        end
+
+        it do
+          exepcted_trace_id = 700846779308066302
```
marcotc, line 148:
```suggestion
          expected_trace_id = 700846779308066302
```

lib/datadog/tracing/trace_id_generation.rb
TonyCTHsu: I recognize that trace id is generated from `SpanOperation` and `Span`, I thought extracting `TraceIdGeneration` to be included for `SpanOperation` and `Span` would be consistent.

Issue #2492, opened 2022-12-16
Add the ability to define a custom error handler for Net/HTTP


caramcc:
**Is your feature request related to a problem? Please describe.**
The applications I'm instrumenting make a lot of API calls to external services using Net/HTTP. Since the services are user-configured, we expect a relatively large volume of these requests to fail with 4xx or 5xx status codes. As a result, we end up ingesting a lot of traces due to `ingestion_reason:error` that we don't consider to be an error and don't otherwise care about.

**Describe the goal of the feature**
Having the ability to configure a custom `error_handler` for the Net/HTTP service similar to the one configured for Faraday would allow us to configure which responses we want to treat as an error. The status quo is [using the hard-coded `response code in 400...599`](https://github.com/DataDog/dd-trace-rb/blob/e5bb0eb9892ac9474f0887d703052eb55d3b1e6b/lib/datadog/tracing/contrib/http/instrumentation.rb#L98-L101) configuration for errors.

**Describe alternatives you've considered**
Even being able to configure which net/http response codes we want to treat as an error would be a big improvement over the status quo.

delner:
Thanks for sharing!

Although I like the idea of custom handlers, there are some challenges with designing a feature and committing to supporting it as a public API. If our hope is to support Ractors in the future, custom blocks like this won't work (they will raise isolation errors). I think there are ways around this, but we just need to be a bit careful.

In the interim, I think configuring error code ranges should be a good stop gap measure, and generally more easy to control from a deployment perspective. We already do this for the web frameworks (like Rails), so there's a well worn path for implementation and support there. This is the fastest way to get some improvement.

 We don't have capacity right now for taking this on in the short term, but we will certainly track it on our own TODO list. If anyone from the community would like to make a contribution to add this, we'd be happy to guide & support that effort as well!

caramcc:
Thank you for the advice! Totally makes sense about Ractor support.

I added a PR to make the HTTP status codes configurable, since that would be a huge help for us as a stop-gap measure.

TonyCTHsu:
👋 @caramcc  , `1.9.0` has been released, give it a try!

caramcc:
Awesome, thanks for the update!

PR #2501, opened 2022-12-20
Configure which response codes are considered errors for Net/HTTP

caramcc:
**What does this PR do?**
Adds the ability to configure which HTTP response codes are considered errors for the Net/HTTP module

**Motivation**
The applications I'm instrumenting make a lot of API calls to external services using Net/HTTP. Since the services are user-configured, we expect a relatively large volume of these requests to fail with 4xx or 5xx status codes. As a result, we end up ingesting a lot of traces due to `ingestion_reason:error` that we don't consider to be an error and don't otherwise care about.

Adding the ability to configure which HTTP status codes are considered errors will allow us to reduce our ingestion rates without entirely ignoring all Net/HTTP spans that return a 4xx or 5xx status code.

(See my [feature request for this issue](https://github.com/DataDog/dd-trace-rb/issues/2492) for more context.)

**How to test the change?**
I tested the change with `bundle exec appraisal ruby-3.0.4-contrib rake spec:http` from the Docker container described in the development guide. All specs, including the new one that I've added in this PR, passed locally.


TonyCTHsu (approved):
👋 @caramcc , Thanks for the contribution and great work!

lib/datadog/tracing/contrib/http/configuration/settings.rb
```diff
@@ -33,6 +33,8 @@ class Settings < Contrib::Configuration::Settings
               o.lazy
             end
 
+            option :response_code_errors, default: 400...599
```
delner, line 36: Thinking maybe this should be configurable by env var, too? Not all users can practically configure using code in their deployments.

@marcotc thoughts?

spec/datadog/tracing/contrib/http/request_spec.rb
```diff
@@ -101,6 +101,17 @@
         expect(span.get_tag('error.message')).to be nil
       end
 
+      context 'with custom response error codes' do
```
delner, line 104: I think it'd be good to see some positive and bad input examples as well (e.g. `nil`, in range, etc)

delner (commented):
Pretty straightforward; looking good! Just a few small suggestions for accessibility, quality.

spec/datadog/tracing/contrib/http/request_spec.rb
caramcc: I've added additional test cases for in-range errors and the default case, but I'm a bit stuck on testing bad input (e.g. explicitly setting the value to `nil` or another wrong type).

Is the convention for this project to do type-checking or type-casting in the settings, in the application code where the setting is invoked, something that the configuration module is expected to handle, or is it the end-user's responsibility to use the correct types for configuration?

Happy to make any changes as appropriate — if you're aware of prior art around checking settings like that, could you share an example or two?

lib/datadog/tracing/contrib/http/configuration/settings.rb
caramcc: I went ahead and added this as well so it could be reviewed, happy to revert that change if it would be better not to add it

docs/GettingStarted.md
```diff
@@ -1367,6 +1367,7 @@ content = Net::HTTP.get(URI('http://127.0.0.1/index.html'))
 | `distributed_tracing` | Enables [distributed tracing](#distributed-tracing) | `true` |
 | `service_name` | Service name used for `http` instrumentation | `'net/http'` |
 | `split_by_domain` | Uses the request domain as the service name when set to `true`. | `false` |
+| `response_code_errors` | Range or Array of HTTP status codes that should be traced as errors. | `400...599` |
```
marcotc, line 1370: I suggest changing this to `error_status_codes`, given "status code" here can only refer to the response (the HTTP request does not have a status code):
```suggestion
| `error_status_codes` | Range or Array of HTTP status codes that should be traced as errors. | `400...599` |
```

lib/datadog/tracing/contrib/http/configuration/settings.rb
```diff
@@ -33,6 +33,11 @@ class Settings < Contrib::Configuration::Settings
               o.lazy
             end
 
+            option :response_code_errors do |o|
+              o.default { env_to_list(Ext::ENV_RESPONSE_CODE_ERRORS, 400...599, comma_separated_only: false)}
```
marcotc, line 37:
```ruby
(400...599).include?(599)
# => false
```

The endless range does not include `599`. I suggest changing the end value to 600, keeping the range endless:

```suggestion
              o.default { env_to_list(Ext::ENV_RESPONSE_CODE_ERRORS, 400...600, comma_separated_only: false)}
```

spec/datadog/tracing/contrib/http/request_spec.rb
marcotc: > something that the configuration module is expected to handle

For our global configuration settings, those that appear in the `Datadog.configure do |c| c.this_option_here end`, we have testing covering different variations of configured inputs, [for example](https://github.com/DataDog/dd-trace-rb/blob/feb11095c88cc6672c33a734357d42f8a7d0325d/spec/datadog/core/configuration/settings_spec.rb#L616-L662).

I noticed that configuration inside `contrib/` doesn't have this strict testing today, because most settings under `contrib/` influence a few key behaviours that are tested in the integration tests of the instrumented gem at hand.

For the case of `response_code_errors`, because it's such a good option that we probably want to also implement it for all other HTTP client gems that we support, I suggest we move the tests you have here, inside `context 'with custom response error codes' do`, into its own group of `shared_examples`, and then include that shared_example here.

This way, these tests can be used by all other HTTP client gems when someone goes to implement them.

And because the tests will be shared, it's a good idea to be very through with them, as each test in that group will be executed multiple times, across different `contrib/` gems.

I suggest adding tests for:
* "with a status code lesser than the default range" (399)
* "with a status code at the beginning of the default range" (400)
* "with a status code at end beginning of the default range" (599)
* "with a status code greater than the default range" (600)
* "with an empty array"

lib/datadog/tracing/contrib/http/configuration/settings.rb
caramcc: That makes sense 👍 

FWIW This default value of `400...599` was taken from the current behavior in order to keep this change backwards-compatible: https://github.com/DataDog/dd-trace-rb/blob/e5bb0eb9892ac9474f0887d703052eb55d3b1e6b/lib/datadog/tracing/contrib/http/instrumentation.rb#L98-L101

Do you think it's worth making a backwards-incompatible change to update this range?

marcotc: Given the exclusion of 599 is likely an off-by-one bug, it should be fixed.
Bug fixes are allowed to perform behaviour changes.

spec/datadog/tracing/contrib/http/request_spec.rb
marcotc: >* Which file specifically should I be adding the `shared_examples` definition to? Most of the `settings_spec.rb` files that I looked at seem to be validating that configuration options get set, and nothing referencing spans, so I'm guessing that isn't right.

For HTTP status testing, because it will be shared with other `contrib/` HTTP clients, I suggest adding it to a new file: 
`spec/datadog/tracing/contrib/http_examples.rb`. For reference, this is related to this comment from earlier:
> I suggest adding tests for:
   > "with a status code lesser than the default range" (399)
   > "with a status code at the beginning of the default range" (400)
  >  "with a status code at end beginning of the default range" (599)
 >   "with a status code greater than the default range" (600)
>    "with an empty array"


>* I noticed some `shared_examples` defined in the other HTTP library specs (e.g. [HttpClient](https://github.com/DataDog/dd-trace-rb/blob/279edc9/spec/datadog/tracing/contrib/httpclient/instrumentation_spec.rb?rgh-link-date=2023-01-10T19%3A32%3A02Z#L81), [HttpRb](https://github.com/DataDog/dd-trace-rb/blob/279edc9/spec/datadog/tracing/contrib/httprb/instrumentation_spec.rb?rgh-link-date=2023-01-10T19%3A32%3A02Z#L85)) but afaict both are just referenced in the same file (with `it_behaves_like 'instrumented request'`) — is that what I should be doing here?

You will be creating a `shared_examples` in `spec/datadog/tracing/contrib/http_examples.rb` and referencing it in `spec/datadog/tracing/contrib/http/request_spec.rb` with a `include_examples`. You'll have to require `http_examples.rb` for it to be visible to `request_spec.rb`.

marcotc: Here's a POC snippet:
```ruby
RSpec.shared_examples 'with error status code configuration' do
  context 'with a custom range' do
    context "with an Range object" do
      let(:configuration_options) { { response_code_errors: 500..502 } }

      context 'with a status code within the range' do
        let(:status_code) { 399 }

        it 'marks the span as an error' do
          expect(span).to have_error
        end
      end

      context 'with a status code outside of the range' do
        let(:status_code) { 503 }

        it 'does not mark the span as an error' do
          expect(span).to_not have_error
        end
      end
    end

    context "with an Array object" do
      let(:status_code) { 500 }

      context "with an empty array" do
        let(:configuration_options) { { response_code_errors: [] } }

        it 'does not mark the span as an error' do
          expect(span).to_not have_error
        end
      end

      context 'with a status code in the array' do
        let(:configuration_options) { { response_code_errors: [400,500] } }
        let(:status_code) { 400 }

        it 'marks the span as an error' do
          expect(span).to have_error
        end
      end

      context 'with a status code not in the array' do
        let(:configuration_options) { { response_code_errors: [400,500] } }
        let(:status_code) { 399 }

        it 'does not mark the span as an error' do
          expect(span).to_not have_error
        end
      end
    end
  end

  context 'with the default range' do
    context "with a status code lesser than the range" do
      let(:status_code) { 399 }

      it 'does not mark the span as an error' do
        expect(span).to_not have_error
      end
    end

    context "with a status code at the beginning of the range" do
      let(:status_code) { 400 }

      it 'marks the span as an error' do
        expect(span).to have_error
      end
    end

    context "with a status code at the end of the range" do
      let(:status_code) { 599 }

      it 'marks the span as an error' do
        expect(span).to have_error
      end
    end

    context "with a status code greater than the range" do
      let(:status_code) { 600 }

      it 'does not mark the span as an error' do
        expect(span).to_not have_error
      end
    end
  end
end
```

Notice these `shared_examples` depend on the test that includes it to use the value for `configuration_options`, `status_code`, and `span`.
You tests already use `configuration_options` and `span`, so that part should just work, but you'll have to use `status_code` to configure your test to return the status code in the `status_code` local method, like so: `before { stub_request(...).to_return(status: status_code, ...) }`. This will, hopefully, make tests pass.

caramcc: That was incredibly helpful guidance, thank you very much! I've updated the specs to use this pattern (using your POC for the `http_examples`) — they're all passing locally for me

marcotc (approved):
Thank you, @caramcc! It all looks great! ✅ 

We'll address the CI failures, which are unrelated to your changes.

caramcc:
Awesome, thank you! I should have a PR to add this feature to `httpclient` and `httprb` ready to go within the next couple days as well.

PR #2502, opened 2022-12-21
128 bit trace id logging

TonyCTHsu:
**What does this PR do?**

TBD


lib/datadog/tracing/correlation.rb
```diff
@@ -71,6 +71,15 @@ def to_log_format
...
           end
         end
+
+        def trace_id
+          if Datadog.configuration.tracing.trace_id_128_bit_logging_enabled &&
+              Tracing::TraceIdConversion.to_high_order(@trace_id) != 0
+            return @trace_id
+          end
+
+          Tracing::TraceIdConversion.to_low_order(@trace_id)
```
marcotc, line 81: Is it possible to "cache" the trace_id at `initialize` time, so we don't have to recalculate it on every call to `#trace_id`. If I understand correctly, the trace id is immutable after class initialization.

marcotc: It's ok if we have to create or rename instance variables (they are internal anyway).

PR #2508, opened 2022-12-27
Action:Upgrade setup-ruby to fix bundler version matching

marcotc:
Upgrade to release https://github.com/ruby/setup-ruby/releases/tag/v1.131.0, titled "Use the correct Bundler version depending on the target Ruby version".

This should fix our GitHub Action CI failures due to bundler version mismatches.


marcotc:
Merging to unblock master build.

PR #2510, opened 2022-12-28
Move to Ruby 3.2.0 final

lloeki:
**What does this PR do?**

Move from the preview ruby image to the final one.

**Motivation**

So we can base ourselves on the latest thing to make things pass.

**Additional Notes**

CircleCI test is still disabled as tests don't pass yet, so we can change that right away as well without caring about the image being published.

**How to test the change?**

This should build the image in CI. As always publishing is done later.

Image can also be built locally as always, and then tested using docker compose:

```
docker buildx build . --platform linux/$(uname -m | sed 's/arm64/aarch64/') -f Dockerfile-3.1.1 -t ghcr.io/datadog/dd-trace-rb/ruby:3.1.1-dd
docker-compose run --rm tracer-3.2 /bin/bash
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -529,8 +529,8 @@ job_configuration:
     resource_class_to_use: medium+
   - &config-3_2
     <<: *filters_all_branches_and_tags
-    ruby_version: 'ruby-3.2.0-preview3'
-    image: ghcr.io/datadog/dd-trace-rb/ruby:3.2.0-preview3-dd
+    ruby_version: 'ruby-3.2.0'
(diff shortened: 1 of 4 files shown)


marcotc:
Similar to https://github.com/DataDog/dd-trace-rb/pull/2497/files.

I found an issue with Appraisal not being compatible with Ruby 3.2, which is kind of a deal breaker :(

lloeki:
Yeah @marcotc, although this PR is merely about bumping the version from 3.2.0-preview3 to 3.2.0 throughouth, not about actually fixing and enabling 3.2 (which IIUC is the goal of #2497)

marcotc (approved):

PR #2512, opened 2022-12-30
Add event kit API

lloeki:
**What does this PR do?**

Implement event sugar API for users to emit either login events or custom business events to be tracked.

**Motivation**

These events are a key feature that Sqreen had and enabled higher level analysis and partially to fully automated security response.

**Additional Notes**

Just as with `Datadog::Kit::Identity#set_user` this looks like convenience syntactic sugar. Nonetheless, just as with `set_user`, wrapping the tag keys allows for these keys to evolve, or even the transport mechanism to change.

In addition, blocking may be handled and triggered, which is probably a feature we don't want in `set_tag`!

**How to test the change?**

There are a bunch of specs added.

### DIFF
--- a/lib/datadog/kit/events.rb
+++ b/lib/datadog/kit/events.rb
@@ -38,11 +38,14 @@ module Datadog
       # @param event [String] Mandatory. Event code.
       # @param trace [TraceOperation] Trace to attach data to.
       # @param others [Hash<Symbol, String>] Additional free-form
-      #   event information to attach to the trace.
+      #   event information to attach to the trace. Key must not
+      #   be :track.
       def self.track(event, trace, **others)
         trace.set_tag("appsec.events.#{event}.track", 'true')
 
         others.each do |k, v|
+          raise ArgumentError, 'key cannot be :track' if k.to_sym == :track
+
           trace.set_tag("appsec.events.#{event}.#{k}", v) unless v.nil?
         end
 
--- a/lib/datadog/kit/events.rb
+++ b/lib/datadog/kit/events.rb
@@ -1,4 +1,5 @@
 # typed: false
+# frozen_string_literal: true
 
 module Datadog
   module Kit
--- a/lib/datadog/kit/events.rb
+++ b/lib/datadog/kit/events.rb
@@ -1,0 +1,31 @@
+# typed: false
+
+module Datadog
+  module Kit
+    # Tracking events
+    module Events
+      def self.track_login(trace, id:, success:, **others)
+        subtag = success ? 'success' : 'failure'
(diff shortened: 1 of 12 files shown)


lib/datadog/kit/events.rb
```diff
@@ -0,0 +1,31 @@
+# typed: false
+
+module Datadog
+  module Kit
+    # Tracking events
+    module Events
+      def self.track_login(trace, id:, success:, **others)
+        subtag = success ? 'success' : 'failure'
```
marcotc, line 8: I recommend adding the frozen_string_literal comment to all new production files.

lib/datadog/kit/events.rb
```diff
@@ -0,0 +1,31 @@
...
+
+        if success
+          trace.set_tag('usr.id', id)
+        else
+          trace.set_tag("appsec.events.users.login.#{subtag}.usr.id", id)
+        end
+      end
+
+      def self.track(event, trace, **others)
+        trace.set_tag("appsec.events.#{event}.track", true)
```
marcotc, line 21:
```suggestion
        trace.set_tag("appsec.events.#{event}.track", 'true')
```

Because `#set_tag` will always call `to_s`, this making it a string will save a bit on processing.

lib/datadog/kit/events.rb
```diff
@@ -0,0 +1,31 @@
...
+
+        others.each do |k, v|
+          trace.set_tag("appsec.events.#{event}.#{k}", v) unless v.nil?
```
marcotc, line 24: I'm assuming `k == 'track'` is unlikely, right?

lib/datadog/kit/events.rb
lloeki: Good point. I wonder why Rubocop didn't yell at me (I recall it usually does for this).

lib/datadog/kit/events.rb
lloeki: Very, although I could add a check for that though, to be thorough.

There also appear to be an update to the RFC where `k` would be nested under a `#{event}.metadata` level.

lib/datadog/kit/events.rb
lloeki: Ah that's because of this:

```
# TODO: Disabling until we categorize which file are safe to do so.
Style/FrozenStringLiteralComment:
  Enabled: false
```

lib/datadog/kit/events.rb
lloeki: Check added.

marcotc (approved):

PR #2513, opened 2023-01-02
Distributed tracing for Sidekiq

TonyCTHsu:
**What does this PR do?**

Implement distributed tracing for sidekiq


lib/datadog/tracing/contrib/sidekiq/client_tracer.rb
```diff
@@ -22,7 +23,9 @@ def initialize(options = {})
           def call(worker_class, job, queue, redis_pool)
             resource = job_resource(job)
 
-            Datadog::Tracing.trace(Ext::SPAN_PUSH, service: @sidekiq_service) do |span|
+            Datadog::Tracing.trace(Ext::SPAN_PUSH, service: @sidekiq_service) do |span, trace_op|
+              propagation.inject!(trace_op, job) if configuration[:distributed_tracing]
```
marcotc, line 27: Are we setting values string into the `job` object (e.g. `job['trace-id']`)? Does this work? Where is this data stored by Sidekiq?

TonyCTHsu: The `job` object is the payload, which is the data being serialized into the queue and picked up by the worker.

An [example](https://github.com/DataDog/dd-trace-rb/pull/2513/files#diff-40b161d4d52adb26d78db98c5c3ce9d1bc8c5c22c3149e2d676d3c05a89c7d80R76-R84) of it could be found in the test. We could also consider having a namespace for those keys instead of first level keys.

marcotc: They are all `x-datadog-...` or something similar, so I think that's fine.

lib/datadog/tracing/contrib/sidekiq/distributed/propagation.rb
```diff
@@ -0,0 +1,40 @@
...
+    module Contrib
+      module Sidekiq
+        module Distributed
+          # Extracts and injects propagation through HTTP headers.
+          class Propagation < Tracing::Distributed::Propagation
+            def initialize
+              super(
+                propagation_styles: {
+                  Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_B3_MULTI_HEADER =>
+                    Tracing::Distributed::B3Multi.new(fetcher: Fetcher),
```
marcotc, line 23: You can use `Tracing::Distributed::Fetcher` directly here, so you can delete the file `lib/datadog/tracing/contrib/sidekiq/distributed/fetcher.rb`:
```suggestion
                    Tracing::Distributed::B3Multi.new(fetcher: Tracing::Distributed::Fetcher),
```

lib/datadog/tracing/contrib/sidekiq/distributed/propagation.rb
```diff
@@ -0,0 +1,40 @@
...
+                    Tracing::Distributed::B3Single.new(fetcher: Fetcher),
+                  Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_DATADOG =>
+                    Tracing::Distributed::Datadog.new(fetcher: Fetcher),
+                  Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_TRACE_CONTEXT =>
+                    Tracing::Distributed::TraceContext.new(fetcher: Fetcher),
+                  Tracing::Configuration::Ext::Distributed::PROPAGATION_STYLE_NONE => Tracing::Distributed::None.new
+                })
+            end
+
+            INSTANCE = Propagation.new
```
marcotc, line 34: Don't add this singleton, as it's a bad practice from HTTP and gRPC that only exist for backwards compatibility.

lib/datadog/tracing/contrib/sidekiq/server_tracer.rb
```diff
@@ -68,10 +76,16 @@ def call(worker, job, queue)
...
               yield
             end
+            # rubocop:enable Metrics/MethodLength
+            # rubocop:enable Metrics/AbcSize
           end
 
           private
 
+          def propagation
+            Sidekiq::Distributed::Propagation::INSTANCE
```
marcotc, line 86: Use `@propagation = Sidekiq::Distributed::Propagation.new` in the `initializer` instead.

spec/datadog/tracing/contrib/sidekiq/distributed_tracing_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      expect(span.get_tag('sidekiq.job.queue')).to eq('default')
+      expect(span.status).to eq(0)
+      expect(span.get_tag('component')).to eq('sidekiq')
+      expect(span.get_tag('operation')).to eq('push')
+      expect(span.get_tag('span.kind')).to eq('producer')
+
+      expect(job['x-datadog-trace-id']).to eq(span.trace_id.to_s)
+      expect(job['x-datadog-parent-id']).to eq(span.id.to_s)
+      expect(job['x-datadog-sampling-priority']).to eq('1')
+      # expect(job["x-datadog-tags"]).to eq("_dd.p.dm=-0")
```
marcotc, line 63: You can delete this:
```suggestion
```

spec/datadog/tracing/contrib/sidekiq/distributed_tracing_spec.rb
```diff
@@ -0,0 +1,119 @@
+# typed: ignore
```
marcotc, line 1: I think you can write just two tests: one with `distributed_tracing: true` and another with it `false`.

Then the test setup is like:
1. Start a span in the test run. Set a non-default sampling priority, origin, and a dummy `_dd.p.*` tag in the active trace.
1. Run `EmptyWorker.perform_async` with the previous span still active.
1. Check that you have two spans, where the `perform_async` span is a child of the span in `1`.
2. Ensure both spans have the same sampling priority, origin, and `_dd.p.*` tags.

sled:
Is there any way I could help getting this PR merged? @TonyCTHsu

TonyCTHsu:
👋 @sled , thanks for offering to help. 

I hesitated to continue this due to the fact that it is difficult to make sense of distributed tracing for asynchronous process in UI. The mechanism for distributed tracing is by context propagation, however, asynchronous process leaves a huge gap between the time when a job being pushed into a queue and before picked up by workers (a typical http request/response cycle would be milliseconds, while this gap could take seconds or longer, depends on how long it has been sitting and waiting inside of the queue). This gap takes too much space for the entire trace graph and makes the spans (actual code execution) tiny and hard to read. 

Furthermore, how does the entire trace expected to look like when a job fail and push back into the queue, then retry several times before it is considered dead? The duration of the trace could easily extended beyond days and weeks for asynchronous process. 

If you are interested in this feature, perhaps we could release it in a opt-in configuration, how does that sound?

sled:
@TonyCTHsu if this is just a display issue, maybe the Datadog UI team could have a look at it? The big gaps between enqueueing and execution of the job could be squished for example. 

I think distributed tracing is an excellent fit for asynchronous systems like background jobs or event processing because it allows you to stitch together the whole picture starting from the initiator. Retries should also continue the original trace,  a common scenario is a faulty application (v1) which enqueued jobs with wrong parameters. This gets fixed in v2, but you might still see job retries. With distributed tracing, you can easily trace those retried jobs to the faulty version which originally enqueued the job.

sled:
I've implemented a Sidekiq middleware based on your PR, here's how the distributed traces look:

Screenshot 2023-03-21 at 18 24 49

TonyCTHsu:
👋 @sled , thanks for sharing! Other languages tracers have synthetic spans to fill the gap which did not fully address the UI issue, but Datadog is working on alternative solution for asynchronous tracing instead of implementing distributed tracing.

Since I don't know when the alternative solution would be available and your graph looks fine, if this make sense to you, I believe we should move forward!

Appraisals
```diff
@@ -1622,8 +1622,8 @@ elsif ruby_version?('3.0') || ruby_version?('3.1') || ruby_version?('3.2')
   appraise 'core-old' do
     gem 'dogstatsd-ruby', '~> 4'
   end
-# ----------------------------------------------------------------------------------------------------------------------
-# ADD NEW RUBIES HERE
+  # ----------------------------------------------------------------------------------------------------------------------
+  # ADD NEW RUBIES HERE
```
marcotc, line 1626: Oops

marcotc (approved):

PR #2514, opened 2023-01-03
Lint: Check for presence of frozen_string_literal

marcotc:
**What does this PR do?**

This PR enables the `Style/FrozenStringLiteralComment` cop, which checks for the presence of the magic comment `frozen_string_literal` at the top of a Ruby file.
The cop does not care if `frozen_string_literal` is set to `true` or `false`, only that an explicit decision was made.

_**This PR adds an exception for all existing files, enforcing this cop only for new files.**_

**Motivation**

The magic `frozen_string_literal: true` ensure that literal strings in a file are not recreated every time their instruction is executed. This is normally the desired behavior, and saves CPU and memory.

It's possible to set `frozen_string_literal: false` in case a needs its strings to be mostly duplicated on execution.

**Additional Notes**

**How to test the change?**


.rubocop_todo.yml
```diff
@@ -1,82 +1,60 @@
...
-    - 'test/contrib/rails/test_helper.rb'
-
-# Offense count: 3
+# Offense count: 1
+# This cop supports unsafe autocorrection (--autocorrect-all).
 Lint/DuplicateRequire:
   Exclude:
-    - 'lib/datadog/ci/contrib/cucumber/integration.rb'
-    - 'lib/datadog/ci/contrib/rspec/integration.rb'
     - 'spec/datadog/tracing/contrib/presto/client_spec.rb'
```
ivoanjo, line 22: (Minor: This seems to be a low-hanging fruit fix :) )

ivoanjo (approved):
:+1: LGTM

PR #2515, opened 2023-01-03
Lint: Rubocop running out of memory in CI

marcotc:
The CI Rubocop job (`lint`) has been running at [99% memory recently](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/8450/workflows/552247af-bfb0-4a9d-9a96-d2856fcb0ac8/jobs/314215/resources):
Screen Shot 2023-01-03 at 12 44 31 PM

Sometimes, it fails with `Parallel::DeadWorker`, likely due to out of memory errors.

This PR bumps up the machine size from 2Gb, to 6Gb (our default). I skipped the 4Gb runner because it would add combination of runners to our `config.yml`, and the 6Gb runner is also faster (more CPU) so there are still benefits to it.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,7 +133,11 @@ step_bundle_install: &step_bundle_install
 step_rubocop: &step_rubocop
   run:
     name: Delint with Rubocop
-    command: bundle exec rake rubocop
+    # There's no straightforward way to get the number of available processors & CPU threads in CircleCI.
+    # Currently it always return 18 physical processors and 36 threads, regardless of executor size.
+    # The workaround is to use `cpu.shares / 1024`:
+    # https://discuss.circleci.com/t/environment-variable-set-to-the-number-of-available-cpus/32670/4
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -133,7 +133,11 @@ step_bundle_install: &step_bundle_install
 step_rubocop: &step_rubocop
   run:
     name: Delint with Rubocop
-    command: bundle exec rake rubocop
+    # There's no straightforward way to get the number of available processors & CPU threads in CircleCI.
+    # Currently it always return 18 physical processors and 36 threads, regardless of executor size.
+    # The workaround is to use `cpu.shares / 1024`:
+    # https://discuss.circleci.com/t/environment-variable-set-to-the-number-of-available-cpus/32670/4
+    command: PARALLEL_PROCESSOR_COUNT=$((`cat /sys/fs/cgroup/cpu/cpu.shares` / 1024)) bundle exec rake rubocop
```
ivoanjo, line 140: :+1: Nice!

PR #2516, opened 2023-01-04
Add hostname for agent container

TonyCTHsu:
**Why**

Since version 7.40, the Agent properly recognizes failed hostname resolution in containerized environments. Without a resolved hostname, the Agent exits with an error shortly after it starts.

```
Error: Error while getting hostname, exiting: unable to reliably determine the host name. You can define one in the agent config file or in your hosts file
AGENT EXITED WITH CODE 255, SIGNAL 0, KILLING CONTAINER
```

https://docs.datadoghq.com/agent/troubleshooting/hostname_containers/?tab=helm

**What does this PR do?**

Configure agent containers with `DD_HOSTNAME`, on my machine when running the integration app, it would shows

```
2023-01-04 10:23:09 UTC | CORE | INFO | (cmd/agent/app/run.go:257 in StartAgent) | Starting Datadog Agent v7.40.1
2023-01-04 10:23:09 UTC | CORE | INFO | (cmd/agent/app/run.go:315 in StartAgent) | Hostname is: COMP-JKGQTW6NN4
```


ivoanjo:
Do you know why this didn't fail in CI previously? Does the CI docker environment already somehow have a properly-configured hostname?

The change itself looks reasonable, but I'm really curious why we hadn't spotted this before.

TonyCTHsu:
https://github.com/DataDog/dd-trace-rb/blob/master/.circleci/config.yml#L97

We have this configured in our CI, the changes are made for our integration apps

ivoanjo:
> It seems like we have this configured in our CI, the changes are made for our integration apps

Indeed! My question was directed at when we run the integration apps in CI -- why were they still working even without the missing configuration? I'm asking because the `docker-compose.yml` for the integration apps also gets used in CI... I think?

marcotc:
This was fixed for CI and docker-compose.yml: https://github.com/DataDog/dd-trace-rb/pull/2349/files

Setting the environment to `DD_HOSTNAME=dd-trace-rb-integrations` is probably ok here.
I see you set it to `DD_HOSTNAME=$(hostname)`, and that looks ok as well, probably can keep it as `$(hostname)`.

ivoanjo (approved):
:+1: Tested and works great!

PR #2517, opened 2023-01-04
Remove agent container for integration test on CI

TonyCTHsu:
**What does this PR do?**

https://github.com/DataDog/dd-trace-rb/pull/2516#issuecomment-1370904457

Agent container is redundant for running integration test on CI. Hence, I commented out the configuration.


marcotc:
Isn't this the agent that allows us to run `integrations/*` tests locally, with a real API key, and see the live results in the Datadog UI?
Are we depending on the agent provided by the main `docker-compose.yml` instead for local testing?

marcotc (approved):

marcotc:
We can completely delete the agent lines, as they will grow stale otherwise.

ivoanjo (approved):
Thanks Tony for the investigation and the overall discussion around this.

Like Marco said, we could remove the `ddagent:` entirely, but please do leave the comment explaining the context, I think that's helpful for future colleagues and changes :)

PR #2522, opened 2023-01-04
Support PG calls with a block

### LINKED ISSUE #2505: `NoMethodError: undefined method 'ntuples' for` with `PG#exec` when used with a block parameter
**Current behaviour**

When using `PG#exec` with a block parameters, `ddtrace` PG auto instrumentation is crashing with:

```
NoMethodError: undefined method `ntuples' for 1:Integer

              span.set_tag(Contrib::Ext::DB::TAG_ROW_COUNT, result.ntuples)
                                                                  ^^^^^^^^
from /Users/beauraf/.asdf/installs/ruby/3.1.3/lib/ruby/gems/3.1.0/gems/ddtrace-1.7.0/lib/datadog/tracing/contrib/pg/instrumentation.rb:163:in `annotate_span_with_result!'
```

**Expected behaviour**

No crash. Auto instrumentation working well, and reporting metrics.

**Steps to reproduce**

```ruby
PG::Connection.open(host: 'localhost', port: '5432', dbname: 'dbname').exec("SELECT count(*) FROM pg_stat_activity where state not like \'%%idle%%\';") { 1 }
```

**Analysis**

Issue happen here https://github.dev/DataDog/dd-trace-rb/blob/ac49c557df3fbdc372b64176c65d9083a6132122/lib/datadog/tracing/contrib/pg/instrumentation.rb#L104, because it expect that `result` is a `PG::Result` but, when used with a block parameters `PG#exec` return the result of the block.

**Environment**

* **ddtrace version:** 1.7.0
* **Ruby version:** 3.1.3
* **pg version:** 1.4.5

marcotc:
Fixes #2505

TODO: Tests

### DIFF
--- a/lib/datadog/tracing/contrib/pg/instrumentation.rb
+++ b/lib/datadog/tracing/contrib/pg/instrumentation.rb
@@ -20,63 +20,66 @@ module Datadog
 
           # PG::Connection patch methods
           module InstanceMethods
-            def exec(sql, *args)
-              trace(Ext::SPAN_EXEC, sql: sql) do |sql_statement|
-                super(sql_statement, *args)
+            def exec(sql, *args, &block)
+              trace(Ext::SPAN_EXEC, sql: sql, block: block) do |sql_statement, wrapped_block|
+                super(sql_statement, *args, &wrapped_block)
               end
             end
 
-            def exec_params(sql, params, *args)
-              trace(Ext::SPAN_EXEC_PARAMS, sql: sql) do |sql_statement|
-                super(sql_statement, params, *args)
+            def exec_params(sql, params, *args, &block)
+              trace(Ext::SPAN_EXEC_PARAMS, sql: sql, block: block) do |sql_statement, wrapped_block|
+                super(sql_statement, params, *args, &wrapped_block)
               end
             end
 
-            def exec_prepared(statement_name, params, *args)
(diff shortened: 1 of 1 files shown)


lib/datadog/tracing/contrib/pg/instrumentation.rb
```diff
@@ -101,9 +104,18 @@ def trace(name, sql: nil, statement_name: nil)
...
                 end
 
-                result = yield(propagated_sql_statement)
-                annotate_span_with_result!(span, result)
-                result
+                # Read metadata from PG::Result
+                if block
+                  yield(propagated_sql_statement, proc do |result|
+                    ret = block.call(result)
+                    annotate_span_with_result!(span, result)
```
TonyCTHsu, line 111: > If the optional code block is given, it will be passed result as an argument, and the [PG::Result](https://deveiate.org/code/pg/PG/Result.html) object will automatically be cleared when the block terminates. In this instance, conn.exec returns the value of the block.

According to the documentation, the `PG::Result` would be garbage collected once the block ended. I think this implementation is very nice to keep the same promise.

Since the `result` is already available for annotation, would it make sense to annotate before invoking the given block?

```
annotate_span_with_result!(span, result)
block.call(result)
```

delner (approved):

PR #2529, opened 2023-01-05
Pin Appraisal to a specific commit to support Ruby 3.2

marcotc:
Until a fix for https://github.com/thoughtbot/appraisal/issues/199 is released, we can't run appraisal with Ruby 3.2.

There are two proposed fixes upstream: https://github.com/thoughtbot/appraisal/pull/201 and https://github.com/thoughtbot/appraisal/pull/200.

This PR pins the appraisal gem version to the fix in https://github.com/thoughtbot/appraisal/pull/201. Both seems equally correct.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -4,7 +4,9 @@ gemspec
 
 # Development dependencies
 gem 'addressable', '~> 2.4.0' # locking transitive dependency of webmock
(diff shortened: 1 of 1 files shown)


marcotc:
Cherry-picked into #2497

PR #2533, opened 2023-01-10
Support runtime metrics without tracing

marcotc:
This PR ensures the RuntimeMetrics worker, when `c.runtime_metrics.enabled = true`, starts its background worker thread on `ddtrace` startup.

Before this PR, the RuntimeMetrics worker only starts reporting metrics when the first trace is flushed by the `Datadog::Tracing::Writer`, and a service association is performed: https://github.com/DataDog/dd-trace-rb/blob/cf56509d29ef6f1163477242e30256e431f8222a/lib/datadog/tracing/writer.rb#L139-L142

This means no runtime metrics will be reported if tracing is disabled.

After this PR, the RuntimeMetrics worker will start immediately. It starts up being associated with `DD_SERVICE`, so metrics are reported to the correct service: https://github.com/DataDog/dd-trace-rb/blob/cf56509d29ef6f1163477242e30256e431f8222a/lib/datadog/core/configuration/components.rb#L41

### DIFF
--- a/lib/datadog/core/configuration/components.rb
+++ b/lib/datadog/core/configuration/components.rb
@@ -412,6 +412,8 @@ module Datadog
           else
             @logger.debug('Profiling is disabled')
           end
+
+          @runtime_metrics.perform # Starts worker. No-op if disabled.
         end
 
         # Shuts down all the components in use.
--- a/spec/datadog/core/configuration/components_spec.rb
+++ b/spec/datadog/core/configuration/components_spec.rb
@@ -1283,6 +1283,13 @@ RSpec.describe Datadog::Core::Configuration::Components do
         end
       end
     end
+
+    context 'for runtime metrics' do
+      it 'initializes the worker' do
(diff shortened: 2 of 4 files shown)


ivoanjo (approved):
:+1: Seems quite reasonable to me. I had already hit this issue before, when trying to use metrics+profiling but no tracing.

TBH I would've submitted this PR before if I thought it was this easy :smile: 

I have two additional suggestions:

1. (small) It seems it probably makes sense to remove the `perform` from `register_service`, no?

https://github.com/DataDog/dd-trace-rb/blob/239a7af5aed7625af553d6451d0cb12c68ffec75/lib/datadog/core/workers/runtime_metrics.rb#L42-L45

2. Is there any reason to keep `register_service` at all? Could it perhaps be deprecated? Do we still want this "we report the same metrics under different services" behavior?

delner (changes requested):
Hey @marcotc can we hold off on this for just a bit? Need to check with the team on how this is supposed to function.

PR #2536, opened 2023-01-11
Pin bundler and rubygems fro Rubies < 2.6

lloeki:
**What does this PR do?**

Pin rubygems and bundler to last known compatible versions for old rubies

**Motivation**

Newer rubygems and bundler versions require 2.6

**Additional Notes**

Removed a duplicate install of bundler from jruby.

**How to test the change?**

Images should build in CI.

### DIFF
--- a/.circleci/images/primary/Dockerfile-2.1.10
+++ b/.circleci/images/primary/Dockerfile-2.1.10
@@ -131,7 +131,10 @@ RUN echo 'deb http://ftp.debian.org/debian stretch-backports main\ndeb-src http:
   && rm -rf /var/lib/apt/lists/*
 
 # Install Dockerize
-RUN curl -sfL $(curl -s https://api.github.com/repos/powerman/dockerize/releases/latest | grep -i /dockerize-$(uname -s)-$(dpkg --print-architecture)\" | cut -d\" -f4) | install /dev/stdin /usr/local/bin/dockerize; dockerize --version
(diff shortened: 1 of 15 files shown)


ivoanjo (approved):
:+1: LGTM Thanks for the quick fix!

lloeki:
Noticed some weird output for `dockerize`, so I'm adding a check:

```
#13 [ 8/13] RUN curl -sfL $(curl -s https://api.github.com/repos/powerman/dockerize/releases/latest | grep -i /dockerize-$(uname -s)-$(dpkg --print-architecture)" | cut -d" -f4) | install /dev/stdin /usr/local/bin/dockerize
#13 0.177 curl: no URL specified!
#13 0.177 curl: try 'curl --help' or 'curl --manual' for more information
#13 DONE 0.2s
```

lloeki:
Weirdest thing ever, I could not reproduce the `no URL specified` error. Instead, locally curl would somehow succeed but result in a 0 byte download, which creates an empty but executable `dockerize`, which would still pass the test.

lloeki:
Now that's pure nonsense, the same dockerfile results in a different behaviour in CI, still failing with `no URL specified` and an empty `dockerize`.

Attempting to change the install process.

lloeki:
Nearly there, turns out the URL was wrong for `x86_64` (`dpkg --print-architecture` returns `amd64`). I don't know how it worked before 🤔

Anyway `dockerize` is now properly fetched and will also properly fail if there is any issue.

ivoanjo (approved):
:+1: Thanks for the patience

lloeki:
Now running the manual workflow to publish the images.

PR #2537, opened 2023-01-11
Expose development docker container ports to localhost only

ivoanjo:
**What does this PR do?**:

This PR modifies our `docker-compose.yml` files to bind any ports on the development docker containers only for access by the host machine.

This should not impact any development use-cases.

**Motivation**:

If the ports are not bound to localhost, they could potentially be accessed by other machines in the same network, which is not a great idea.

**Additional Notes**:

(N/A)

**How to test the change?**:

Check that running tests using `docker-compose` still works, as well as running them from your development machine directly.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -365,8 +365,8 @@ services:
       - "8125/udp"
       - "8126"
     ports:
-      - "${DD_METRIC_AGENT_PORT}:8125/udp"
-      - "${DD_TRACE_AGENT_PORT}:8126"
+      - "127.0.0.1:${DD_METRIC_AGENT_PORT}:8125/udp"
+      - "127.0.0.1:${DD_TRACE_AGENT_PORT}:8126"
     volumes:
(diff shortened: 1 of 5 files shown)


lloeki:
Hmm, this may break some development setups on macos.

Picture this: macos <-> NAT'd vm <-> containerized process; to be reachable by macos, depending how port forwarding is done, the containerized process may need to expose to `0.0.0.0`. In that situation, since the VM is NAT'd, it is unreachable on the local network.

TonyCTHsu (approved):

ivoanjo:
@lloeki good point, I'm not quite sure. Would you be able to test on your mac dev machine? Mine is someout out-of-order since I've moved to a Linux box and it has been kicked from the Datadog garden.

marcotc (approved):

ivoanjo:
I've doublechecked with Loic and he agreed with merging this and evaluating if there were any reports of broken development use-cases. Thanks Loic! :bow:

PR #2541, opened 2023-01-11
[PROF-6660] Add `profiling_enabled` state to environment logger output

ivoanjo:
**What does this PR do?**:

This PR adds a `profiling_enabled` entry in the environment logger output, which reflects the current configuration for profiling.

**Motivation**:

This was requested internally as being useful to support, and at least Node and Java already provide this key.

We also had a TODO for it, so one less TODO.

**Additional Notes**:

I hesitated between adding the configuration of the profiler being on, or the profiler actually being on (e.g. it's possible that the profiler has been requested on, but can't be enabled because for instance google-protobuf is missing).

My thinking is that the information about if the profiler really is on could be instead relayed via telemetry, so it may not be worth duplicating it here.

**How to test the change?**:

### DIFF
--- a/lib/datadog/core/diagnostics/environment_logger.rb
+++ b/lib/datadog/core/diagnostics/environment_logger.rb
@@ -221,9 +221,9 @@ module Datadog
           !!Datadog.configuration.diagnostics.health_metrics.enabled
         end
 
-        # TODO: Populate when profiling is implemented
-        # def profiling_enabled
-        # end
+        def profiling_enabled
+          !!Datadog.configuration.profiling.enabled
+        end
 
         # TODO: Populate when automatic log correlation is implemented
         # def logs_correlation_enabled
@@ -254,6 +254,7 @@ module Datadog
             partial_flushing_enabled: partial_flushing_enabled,
             priority_sampling_enabled: priority_sampling_enabled,
             health_metrics_enabled: health_metrics_enabled,
+            profiling_enabled: profiling_enabled,
             **instrumented_integrations_settings
           }
         end
--- a/spec/datadog/core/diagnostics/environment_logger_spec.rb
+++ b/spec/datadog/core/diagnostics/environment_logger_spec.rb
@@ -54,7 +54,8 @@ RSpec.describe Datadog::Core::Diagnostics::EnvironmentLogger do
           'runtime_metrics_enabled' => false,
           'version' => DDTrace::VERSION::STRING,
           'vm' => be_a(String),
-          'service' => be_a(String)
+          'service' => be_a(String),
+          'profiling_enabled' => false,
         )
       end
     end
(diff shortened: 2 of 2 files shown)


lloeki (approved):

lib/datadog/core/diagnostics/environment_logger.rb
```diff
@@ -221,9 +221,9 @@ def health_metrics_enabled
           !!Datadog.configuration.diagnostics.health_metrics.enabled
         end
 
-        # TODO: Populate when profiling is implemented
-        # def profiling_enabled
-        # end
+        def profiling_enabled
+          !!Datadog.configuration.profiling.enabled
```
TonyCTHsu, line 225: We don't need `!!`, since the option is already coerced with `env_bool`.

```
         option :enabled do |o|
            o.default { env_to_bool(Profiling::Ext::ENV_ENABLED, false) }
            o.lazy
          end
```

TonyCTHsu (approved):

lib/datadog/core/diagnostics/environment_logger.rb
lloeki: Can someone do `c.profiling.enabled = nil`?

TonyCTHsu: Oh yeah, technically it can be set to any value. but I think our [document](https://docs.datadoghq.com/profiler/enabling/ruby/?tab=incode) make it clear this is a boolean and [Yard doc](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/core/configuration/settings.rb#L193) is being consistent with that.

ivoanjo: Good point. It turns out you can use whatever you want:

```
$ DD_TRACE_DEBUG=true bundle exec pry
[1] pry(main)> require 'ddtrace'
=> true
[2] pry(main)> Datadog.configure { |c| c.profiling.enabled = 'i like pie' };
D, [2023-01-12T13:52:08.108222 #628982] DEBUG -- ddtrace: [ddtrace] (dd-trace-rb/lib/datadog/core/workers/async.rb:132:in `start_worker') Starting thread for: #<Datadog::Core::Telemetry::Heartbeat:0x000062ba9b97a3e8>
D, [2023-01-12T13:52:08.108383 #628982] DEBUG -- ddtrace: [ddtrace] (dd-trace-rb/lib/datadog/core/configuration/components.rb:405:in `startup!') Profiling started
```

So I think the `!!` is probably best to keep for now, although we should probably think about making the configuration system itself less permissive.

marcotc (approved):

ivoanjo:
CI is broken for JRuby 9.2.8.0 due to an unrelated issue, going ahead and merging this.

PR #2543, opened 2023-01-11
128 bits trace id 

TonyCTHsu:
**What does this PR do?**

Support 128-bit trace id. With this change, our trace is backward compatible with 128-bit trace id, which is used for W3C trace context and b3 propagation.

1. Implement backward compatible distributed tracing propagation with 128-bit trace id
2. Implement backward compatible means for sending spans to agent endpoint
3. Enable 128-bit trace id generation with environment variable `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED`
4. Option to log 128-bit trace id with environment variable `DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED`(Not yet supported)


lib/datadog/tracing/utils.rb
```diff
@@ -43,6 +43,34 @@ def self.reset!
...
+      module TraceId
+        MAX = (1 << 128) - 1
+
+        module_function
+
+        def next_id
+          return Utils.next_id unless Datadog.configuration.tracing.trace_id_128_bit_generation_enabled
+
+          concatenate(
+            Time.now.to_i << 32,
```
marcotc, line 57: We have a time provider, that allows users to mock the `ddtrace` time provider for testing, we should use that:
```suggestion
            Core::Utils::Time.now.to_i << 32,
```

lib/datadog/tracing/utils.rb
```diff
@@ -43,6 +43,34 @@ def self.reset!
...
+
+      # The module handles bitwise operation for trace id
+      module TraceId
+        MAX = (1 << 128) - 1
+
+        module_function
+
+        def next_id
+          return Utils.next_id unless Datadog.configuration.tracing.trace_id_128_bit_generation_enabled
+
```
marcotc, line 55: Document the 128 format, very rough draft: "32 bit timestamp in seconds seconds in epoch; 32-bits set to zero, currently reserved; random 64-bits"

lib/ddtrace/transport/trace_formatter.rb
```diff
@@ -47,6 +47,7 @@ def format!
         tag_rate_limiter_rate!
         tag_sample_rate!
         tag_sampling_decision_maker!
+        tag_high_order_trace_id!
```
marcotc, line 50: We should also adjust the the `span.trace_id` on all spans in the `TraceFormatter`, so the `TraceFormatter` always makes sure the `TraceSegment` provided is ready to be serialized without any changes.

This simplifies the serializer's logic.

lib/datadog/tracing/utils.rb
```diff
@@ -43,6 +43,34 @@ def self.reset!
...
+        def to_high_order(trace_id)
+          trace_id.to_i >> 64
+        end
+
+        def to_low_order(trace_id)
+          trace_id.to_i & 0xFFFFFFFFFFFFFFFF
+        end
+
+        def concatenate(high_order, low_order)
+          high_order.to_i << 64 | low_order.to_i
```
marcotc, line 71: Let's remove `#to_i` calls from here, given that we only expect `Integer`s or `nil` as arguments.
`#to_i` is a very lenient method that easily allows for unintended objects to be converted to numbers, and that's not the indented use for these methods.

lib/datadog/tracing/metadata/ext.rb
```diff
@@ -53,6 +53,9 @@ module Distributed
           # Trace tags with this prefix will propagate from a trace through distributed tracing.
           # Distributed headers tags with this prefix will be injected into the active trace.
           TAGS_PREFIX = '_dd.p.'
+
+          TID = 'tid'
```
marcotc, line 57: Could you add a small documentation on what `tid` represents?

lib/datadog/tracing/distributed/helpers.rb
```diff
@@ -20,54 +20,27 @@ def self.clamp_sampling_priority(sampling_priority)
...
-          # Be sure we have a string
-          value = value.to_s
+        def self.parse_hex_id(value, length: nil)
+          return unless value
 
-          # If we are parsing base16 number then truncate to 64-bit
-          value = Helpers.truncate_base16_number(value) if base == 16
+          value = value.to_s.downcase
+          value = value[value.length - length, length] if length && value.length > length
+          value = value.sub(/^0*(?=(0$)|[^0])/, '')
```
marcotc, line 39: Let's add a comment here that we'd like to remove these extra operations on hex ids, now that we support 128-bit a lot of validation Just Works™️.

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -173,6 +173,24 @@ def self.extended(base)
                 o.lazy
               end
 
+              # Enable 128 bit trace id generation.
+              #
+              # @default `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED` environment variable, otherwise `false`
+              # @return [Boolean]
+              option :trace_id_128_bit_generation_enabled do |o|
```
marcotc, line 180: Let's add it these options `GettingStarted.md`.

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -173,6 +173,24 @@ def self.extended(base)
...
+                o.default { env_to_bool(Tracing::Configuration::Ext::ENV_TRACE_ID_128_BIT_GENERATION_ENABLED, false) }
+                o.lazy
+              end
+
+              # Enable 128 bit trace id injected for logging.
```
marcotc, line 185: Let's add a comment here saying it's not supported by the backend yet. Also, we shouldn't surface it in our `GettingStarted.md` yet.

lib/ddtrace/transport/trace_formatter.rb
TonyCTHsu: I actually hesitate to commit to this change since 

`TraceFormatter` is performing at the trace level attributes which is a constraint of agent's API. I would prefer the current approach of decorating `SerailizableSpan` than mutating `Span` from `TraceFormatter`.

marcotc (approved):

PR #2546, opened 2023-01-12
Fix hanami integration app dependencies

TonyCTHsu:
**What does this PR do?**

Since we do not check lock files into version control, the dependencies are stale without version constraint. Causing error

```
app_1          | /usr/local/bundle/gems/dry-configurable-0.16.1/lib/dry/configurable/dsl.rb:31:in `setting': wrong number of arguments (given 2, expected 1) (ArgumentError)
hanami_app_1 exited with code 1
```

1. Limit `dry-configurable`
2. Remove constant for Puma 6

### DIFF
--- a/integration/apps/hanami/Gemfile
+++ b/integration/apps/hanami/Gemfile
@@ -6,6 +6,7 @@ gem 'rake'
 gem 'hanami',       '~> 1.3'
 gem 'hanami-model', '~> 1.3'
 gem 'dry-container', '~> 0.8.0'
+gem 'dry-configurable', '~> 0.12.0'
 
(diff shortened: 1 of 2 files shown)


marcotc (approved):

ivoanjo:
Would it be possible to add hanami to the set of apps that get run in CI on every PR? That's usually the best way to ensure that things don't bitrot ;)

ivoanjo:
@TonyCTHsu is this still relevant?

p-datadog:
This was merged as part of https://github.com/DataDog/dd-trace-rb/pull/3639.

PR #2555, opened 2023-01-13
Test:Fix logger.debug test failures in spec:main

marcotc:
A few tests started failing due to them catching background class to `Datadog.logger.debug` as part of their own test case.

This is a case for having class/module-specific loggers, similar to [how Java does](https://github.com/apache/logging-log4j2/blob/f22731a889e89882ff039331eb3365a7f9805a78/src/site/xdoc/manual/configuration.xml.vm#LL307). I've discussed this with Ivo in the past. This way one could assert on `MyClassUnderTest.logger`, instead of `Datadog.logger`.


ivoanjo (approved):
:+1: Indeed this PR is a bit of a band-aid, but it's a reasonable one until we put a bigger fix in place.

ivoanjo:
CI is broken for JRuby 9.2.8.0 due to an unrelated issue, going ahead and merging this.

ivoanjo:
To add a bit of context, these tests started failing when instrumentation telemetry is on:

```
DD_INSTRUMENTATION_TELEMETRY_ENABLED=true bundle exec rspec spec/datadog/core/diagnostics/environment_logger_spec.rb
```

lloeki (approved):

ivoanjo:
I discussed this with @TonyCTHsu and he had a great point that perhaps we should look into the instrumentation telemetry failure as something that needs to be fixed before we enable it by default, rather than patching over it in this PR.

So we agreed to keep this open for now, and revisit it once we're ready to enable the instrumentation telemetry feature again.

lloeki:
Should we mark this `do not merge` then?

ivoanjo:
Yes, added label!

PR #2557, opened 2023-01-14
Add Stripe instrumentation

agrobbin:
**What does this PR do?**

This adds instrumentation of the [stripe-ruby](https://github.com/stripe/stripe-ruby) gem.

**Motivation**

While using the Stripe gem, I thought it would be really helpful to have spans show up in DataDog APM whenever we interact with Stripe!


lib/datadog/tracing/contrib/stripe/request.rb
```diff
@@ -0,0 +1,60 @@
...
+
+              tag_span(span, event)
+
+              # Finish the span
+              span.finish
+            end
+          end
+
+          def tag_span(span, event)
+            span.resource = event.object_name if event.respond_to?(:object_name) && event.object_name
```
agrobbin, line 35: This particular resource configuration is dependent upon https://github.com/stripe/stripe-ruby/pull/1168, which exposes `object_name`. I'm happy to remove this while that PR is considered if it's preferred!

TonyCTHsu:
👋 @agrobbin , Thanks for your awesome contribution! 

I was not particularly familiar with this gem, but at first glance, it seems fine! Let me spend some time to dig around with the gem to provide proper review.

One thing I would suggest is that, since it is supporting `5.15`, we should setup our test matrix like

| Ruby version | Stripe version |
|-----------|------------|
| 2.3  | 5.x |
| 2.4  | 6.x |
| 2.5  | 7.x |
| 2.6+  | 8.x |

agrobbin:
@TonyCTHsu I just updated this to test against various `stripe` versions like you asked. I set it up so that on Ruby 2.7+, it tests on the latest version of `stripe`. Let me know if there's anything else I can do to help with review here!

lib/datadog/tracing/contrib/stripe/request.rb
```diff
@@ -0,0 +1,64 @@
...
+
+          def finish_span(event)
+            event.user_data[:datadog_span].tap do |span|
+              # If no active span, return.
+              return nil if span.nil?
+
+              tag_span(span, event)
+
+              # Finish the span
+              span.finish
```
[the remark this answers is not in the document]
agrobbin: `tag_span` has a `rescue StandardError` @delner, but I see the pattern you're referring to elsewhere in the library, so I've updated this to match!

lib/datadog/tracing/contrib/stripe/request.rb
TonyCTHsu: I think it would be great if we could derive `resource` from Stripe's api resource (with your Stripe's PR). 

However, I feel like having `stripe.customer` instead of `customer` would be an improvement (currently, without an object name, it is `stripe.request`). What do you think?

TonyCTHsu (approved):
👋 @agrobbin, Great job!

lib/datadog/tracing/contrib/stripe/request.rb
agrobbin: Yep that makes a lot of sense, @TonyCTHsu. I've updated this to be `"stripe.#{event.object_name}"`.

I have not yet heard back from Stripe on that PR, so I'm curious what you suggest doing while we wait. I'm happy to remove this to unblock the rest of this instrumentation. Then I can open a new PR here once stripe/stripe-ruby#1168 (hopefully) gets merged.

TonyCTHsu: I think we can merge this with a code comment referencing the Stripe's PR. What bothers me more would be a missing spec to test this condition.

agrobbin: @TonyCTHsu makes sense! I've just pushed up a test that I think best exercises this. I can keep an eye on stripe/stripe-ruby#1168, and update as appropriate based on what happens with that.

PR #2560, opened 2023-01-16
Extract components out of Core

lloeki:
**What does this PR do?**

Move component-specific code out of `Core::Configuration::Components`

**Motivation**

Reduce coupling. Core should be unaware of component details.

**Additional Notes**

Using the [noop noop noop interface maneuver](https://www.dotconferences.com/2016/10/katrina-owen-the-scandalous-story-of-the-dreadful-code-written-by-the-best-of-us).

As a first step I simply moved the most obvious code out of `lib/datadog/core/configuration/components` into `lib/datadog/{tracing,profiling}`, inside their own module, and then `include`'d these modules right back where the methods used to be.

Second step should be:

- changing these modules to classes that create the tracer and profiler instances and their dependencies
- swapping the tracing and profile instances in `Core::Configuration::Components` with instances of the respective Component classes
- delegating relevant methods from each `<foo>::Component` to each respective `<foo>` instance
- `Core::Configuration::Components` would still reference

Third step should be:

- having each component call a register method on `Core::Configuration::Components` upon the `lib/datadog/<foo>/component.rb` file being required
- having `Core::Configuration::Components` use the registered list by removing direct references to `<foo>` from it
- any dependency between components would be handled by having a `require '../<bar>/component'` at the top of `<foo>/component`. This would make each component registered in order.

**How to test the change?**

Specs should still pass.

### DIFF
--- a/lib/datadog/core/configuration/components.rb
+++ b/lib/datadog/core/configuration/components.rb
@@ -8,11 +8,8 @@ require_relative '../runtime/metrics'
 require_relative '../telemetry/client'
 require_relative '../workers/runtime_metrics'
 
-require_relative '../../tracing/tracer'
-require_relative '../../tracing/flush'
-require_relative '../../tracing/sync_writer'
-require_relative '../../tracing/sampling/span/rule_parser'
-require_relative '../../tracing/sampling/span/sampler'
+require_relative '../../tracing/component'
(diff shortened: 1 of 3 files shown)


lloeki:
Closed because of a GitHub issue. Superseded by #2560

PR #2561, opened 2023-01-16
Extract tracing and profiling components out of the core namespace

lloeki:
**What does this PR do?**

Move component-specific code out of `Core::Configuration::Components`

**Motivation**

Reduce coupling. Core should be unaware of component details.

**Additional Notes**

Using the [noop noop noop interface maneuver](https://www.dotconferences.com/2016/10/katrina-owen-the-scandalous-story-of-the-dreadful-code-written-by-the-best-of-us).

As a first step I simply moved the most obvious code out of `lib/datadog/core/configuration/components` into `lib/datadog/{tracing,profiling}`, inside their own module, and then `include`'d these modules right back where the methods used to be.

Second step should be:

- changing these modules to classes that create the tracer and profiler instances and their dependencies
- swapping the tracing and profile instances in `Core::Configuration::Components` with instances of the respective Component classes
- delegating relevant methods from each `<foo>::Component` to each respective `<foo>` instance
- `Core::Configuration::Components` would still reference

Third step should be:

- having each component call a register method on `Core::Configuration::Components` upon the `lib/datadog/<foo>/component.rb` file being required
- having `Core::Configuration::Components` use the registered list by removing direct references to `<foo>` from it
- any dependency between components would be handled by having a `require '../<bar>/component'` at the top of `<foo>/component`. This would make each component registered in order.

**How to test the change?**

Specs should still pass.

### DIFF
--- a/lib/datadog/core/configuration/components.rb
+++ b/lib/datadog/core/configuration/components.rb
@@ -56,35 +56,6 @@ module Datadog
           def build_telemetry(settings)
             Telemetry::Client.new(enabled: settings.telemetry.enabled)
           end
-
-          private
-
-          def build_tracer_tags(settings)
-            settings.tags.dup.tap do |tags|
-              tags[Core::Environment::Ext::TAG_ENV] = settings.env unless settings.env.nil?
-              tags[Core::Environment::Ext::TAG_VERSION] = settings.version unless settings.version.nil?
-            end
-          end
-
-          def build_test_mode_trace_flush(settings)
-            # If context flush behavior is provided, use it instead.
-            settings.tracing.test_mode.trace_flush || build_trace_flush(settings)
-          end
-
-          def build_test_mode_sampler
-            # Do not sample any spans for tests; all must be preserved.
-            # Set priority sampler to ensure the agent doesn't drop any traces.
-            Tracing::Sampling::PrioritySampler.new(
-              base_sampler: Tracing::Sampling::AllSampler.new,
-              post_sampler: Tracing::Sampling::AllSampler.new
-            )
-          end
-
-          def build_test_mode_writer(settings, agent_settings)
-            # Flush traces synchronously, to guarantee they are written.
-            writer_options = settings.tracing.test_mode.writer_options || {}
-            Tracing::SyncWriter.new(agent_settings: agent_settings, **writer_options)
-          end
         end
 
         attr_reader \
--- a/lib/datadog/core/configuration/components.rb
+++ b/lib/datadog/core/configuration/components.rb
@@ -1,4 +1,4 @@
-# typed: false
+# typed: ignore
 
 require_relative 'agent_settings_resolver'
 require_relative '../diagnostics/environment_logger'
--- a/lib/datadog/core/configuration/components.rb
+++ b/lib/datadog/core/configuration/components.rb
@@ -8,11 +8,8 @@ require_relative '../runtime/metrics'
 require_relative '../telemetry/client'
 require_relative '../workers/runtime_metrics'
 
-require_relative '../../tracing/tracer'
-require_relative '../../tracing/flush'
-require_relative '../../tracing/sync_writer'
-require_relative '../../tracing/sampling/span/rule_parser'
-require_relative '../../tracing/sampling/span/sampler'
+require_relative '../../tracing/component'
+require_relative '../../profiling/component'
 
 module Datadog
   module Core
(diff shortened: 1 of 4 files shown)


lib/datadog/profiling/component.rb
```diff
@@ -0,0 +1,157 @@
...
+          collector = build_profiler_oldstack_collector(settings, recorder, trace_identifiers_helper)
+        end
+
+        exporter = build_profiler_exporter(settings, recorder)
+        transport = build_profiler_transport(settings, agent_settings)
+        scheduler = Profiling::Scheduler.new(exporter: exporter, transport: transport)
+
+        Profiling::Profiler.new([collector], scheduler)
+      end
+
```
ivoanjo, line 91: Minor: All other methods other than `build_profiler` are supposed to be `private` :)

lib/datadog/core/configuration/components.rb
```diff
@@ -57,223 +57,6 @@ def build_telemetry(settings)
...
-            exporter = build_profiler_exporter(settings, recorder)
-            transport = build_profiler_transport(settings, agent_settings)
-            scheduler = Profiling::Scheduler.new(exporter: exporter, transport: transport)
-
-            Profiling::Profiler.new([collector], scheduler)
-          end
-
           private
 
           def build_tracer_tags(settings)
```
ivoanjo, line 62: Should some of these methods also be moved to `tracing/components.rb`? I think most (all?) are only needed by the code that builds the tracer.

lib/datadog/core/configuration/components.rb
```diff
@@ -8,18 +8,18 @@
...
+require_relative '../../profiling/component'
 
 module Datadog
   module Core
     module Configuration
       # Global components for the trace library.
       class Components
         class << self
+          include Datadog::Tracing::Component
+          include Datadog::Profiling::Component
```
ivoanjo, line 21: So, I know you're going for the smallest possible change here, and I can live with this one, buuuuuut I'm going to try to convince you to go a tiny small step further.

Can I convince you to, instead of re-injecting the original behaviors back using a module, to modify the `initialize` method below to instead call directly into the modules?

E.g. instead of 
```ruby
        def initialize(settings)
          # Logger
          @logger = self.class.build_logger(settings)

          agent_settings = AgentSettingsResolver.call(settings, logger: @logger)

          # Tracer
          @tracer = self.class.build_tracer(settings, agent_settings)

          # Profiler
          @profiler = self.class.build_profiler(settings, agent_settings, @tracer)

          # ... etc...
```

Do:
 
```ruby
        def initialize(settings)
          # Logger
          @logger = self.class.build_logger(settings)

          agent_settings = AgentSettingsResolver.call(settings, logger: @logger)

          # Tracer
          @tracer = Datadog::Tracing::Component.build_tracer(settings, agent_settings)

          # Profiler
          @profiler = Datadog::Profiling::Component.build_profiler(settings, agent_settings, @tracer)

          # ... etc...
```

I like the new version better because it makes the dependency explicit, rather than implicit, and I think it moves us closer to a future where the components would register themselves somewhere and provide these functions to be called.

ivoanjo (approved):
:+1: I've left a few suggestions, but overall I'm happy with this change, even in its current state.

I think it makes sense to get another :+1: from someone in tracing, but from my side, you have my ~axe~ ~sword~ LGTM.

lib/datadog/core/configuration/components.rb
lloeki: It looks like so. Moved.

lib/datadog/profiling/component.rb
lloeki: Correct, done.

lib/datadog/core/configuration/components.rb
lloeki: > it makes the dependency explicit, rather than implicit

I agree, and this explicit dependencing will also fail (which is a good thing!) if there's a reference in these methods to a ivar or method of `Core::Configuration::Components`... sadly this needs further validation than I can afford right now, hence this first noop move whose goal is to tackle source code ownership.

I'll attempt to tackle live code dependency in a second PR.

lloeki:
Ha, [eff4e3d](https://github.com/DataDog/dd-trace-rb/pull/2561/commits/eff4e3df9c36531027a946965d2406ed21964016) did it. I thought I tried that before and it failed. I rescind my comment above (Still a pain to work with without local execution).

ivoanjo:
> Ha, [eff4e3d](https://github.com/DataDog/dd-trace-rb/pull/2561/commits/eff4e3df9c36531027a946965d2406ed21964016) did it. I thought I tried that before and it failed. I rescind my comment above (Still a pain to work with without local execution).

So this is one thing where it's hard for me to evaluate "are we at the pain threshold yet?" because it usually works for me (:tm:) but I'm definitely up to revisit if we should keep Sorbet (I would suggest in the next guild meeting?) and I offer to do the work of removing Sorbet if that's the outcome we decide on.

PR #2562, opened 2023-01-16
[ci visibility] Fix copy-paste error in teamcity env vars

albertvaka:
**What does this PR do?**
Fixes a copy-paste error where the wrong env var was being used to compute the pipeline URL for TeamCity.


ivoanjo (approved):
:+1: LGTM. Is this checked in system-tests or elsewhere? It looks like it'd be good to have test coverage for it :)

ivoanjo:
P.s.: We're working on fixing the unrelated CI hiccup, we should have it back to green soon and then I'll merge in the PR :)

albertvaka:
We have a repo (not linking it here because it is private) with tests that run against all tracers. We don't have tests for TeamCity yet but I will add them today.

ivoanjo:
I've merged in the master branch to fix CI, making this good to go -- pressing the magic button now!

PR #2563, opened 2023-01-17
Pin bundler on 3.0

lloeki:
**What does this PR do?**

**Motivation**
This somehow went missing from the previous pinning commit in #2559 

**Additional Notes**

I should get a brain. The current one appears to be defective.

**How to test the change?**

Same as #2559


ivoanjo (approved):
:+1: LGTM

Issue #2443, opened 2022-11-29
Testing appraisal `install_appraisal_gemfiles` step breaks JRuby 9.2.8.0 docker image


ivoanjo:
**Current behaviour**

While working on releasing `ddtrace` 1.7.0 (https://github.com/DataDog/dd-trace-rb/pull/2442) I saw bundler failing in CI (https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/8146/workflows/3d64ab35-5320-4402-8f66-9b36afee1ab3/jobs/302585) with the following issue:

````
>> bundle check --gemfile='/app/gemfiles/jruby_9.2.8.0_cucumber4.gemfile' || bundle install --gemfile='/app/gemfiles/jruby_9.2.8.0_cucumber4.gemfile' --retry 1
/opt/jruby/lib/ruby/stdlib/rubygems/defaults/jruby.rb:34: warning: constant Gem::ConfigMap is deprecated
The following gems are missing
 * cucumber (4.1.0)
 * cucumber-core (7.1.0)
 * cucumber-create-meta (1.0.0)
 * cucumber-cucumber-expressions (10.3.0)
 * cucumber-gherkin (14.2.0)
 * cucumber-html-formatter (7.2.0)
 * cucumber-messages (12.4.0)
 * cucumber-wire (3.1.0)
 * diff-lcs (1.3)
 * sys-uname (1.2.2)
 * cucumber-tag-expressions (2.0.4)
 * protobuf-cucumber (3.10.8)
 * middleware (0.1.0)
 * i18n (1.10.0)
Install missing gems with `bundle install`
/opt/jruby/lib/ruby/stdlib/rubygems/defaults/jruby.rb:34: warning: constant Gem::ConfigMap is deprecated
Fetching gem metadata from https://rubygems.org/.........
Using rake 13.0.6
Using concurrent-ruby 1.1.10
Using minitest 5.15.0
Using bundler 2.3.26
Using addressable 2.4.0
Using zeitwerk 2.5.4
Using thor 1.2.1
Using benchmark-ips 2.10.0
Using ast 2.4.2
Using builder 3.2.4
Using memory_profiler 0.9.14
Using climate_control 0.2.0
Using coderay 1.1.3
Using colorize 0.8.1
Using rexml 3.2.5
Using cri 2.15.11
Using thread_safe 0.3.6 (java)
Fetching middleware 0.1.0
Using ffi 1.15.5 (java)
[...]

PR #2566, opened 2023-01-18
Isolate docker-compose main lockfiles

lloeki:
**What does this PR do?**

Make each container version have its own lockfile.

**Motivation**

- This enables working with multiple versions opened in parallel (e.g several appraisal install, comparing version behaviours...)
- This also enables working independently outside and inside a container (e.g back and forth running specs inside containers and editing outside while having linters still work in the editor, which requires a proper lockfile to pick the correct version)

**Additional Notes**

- working through a VM file sharing (e.g Docker for Mac, colima, etc...) is subject to delays in syncing, which occasionally results in inconsistencies, especially when changes are run in quick succession both inside and outside of containers (e.g rake appraisal tasks)
- this also enables getting rid of `rm -vf Gemfile.lock` every time we want to switch version context

Yes, it's ugly, but except for `bundle lock` it is not possible to specify the path to a lockfile, as it is always derived from the gemfile filename.

**How to test the change?**

Run `bundle install` in the various `docker-compose` versions, and also outside.


lloeki:
> this also enables getting rid of rm -vf Gemfile.lock every time we want to switch version context

Probably I should add such a change to this PR as well.

lloeki:
Damn, #2443 happened again with bundler 2.3.6.

lloeki:
CI failure is because of #2568

TonyCTHsu (approved):

PR #2569, opened 2023-01-18
Fix appraisal specs

TonyCTHsu:
**What does this PR do?**

After regenerating gemfile and lockfiles with appraisal, lots of specs failed with updated dependencies. 

This PR address those failing specs by fixing dependencies and other means.


Appraisals
```diff
@@ -57,10 +57,9 @@ def self.gem_cucumber(version)
     #
     # DEV: Ideally, the profiler would not be loaded when running cucumber tests as it is unrelated.
     if Gem::Version.new(version) >= Gem::Version.new('4.0.0') &&
-        Gem::Version.new(version) < Gem::Version.new('7.0.0') &&
-        RUBY_PLATFORM != 'java' &&
-        Bundler::VERSION > '2.0.0'
-      gem 'google-protobuf', force_ruby_platform: true
+      Gem::Version.new(version) < Gem::Version.new('7.0.0')
```
lloeki, line 60: Indentation has taken me aback here, I thought this line was returning in void context!

gemfiles/ruby_2.2.10_contrib.gemfile.lock
```diff
@@ -1338,7 +1338,7 @@ GEM
     rake-compiler (1.2.1)
       rake
     redcarpet (3.5.1)
-    redis (4.1.1)
+    redis (3.3.5)
```
lloeki, line 1341: Any reason not to put `redis` `< 4` as a constraint in the 2.2 section of `Appraisals`? Otherwise it risks breaking if someone runs an update.

Appraisals
```diff
@@ -949,6 +949,7 @@ elsif ruby_version?('2.5')
     gem 'shoryuken'
     gem 'sidekiq'
     gem 'sneakers', '>= 2.12.0'
+    gem 'bunny', '~> 2.19.0' # uninitialized constant OpenSSL::SSL::TLS1_3_VERSION for jruby, https://github.com/ruby-amqp/bunny/issues/645
```
lloeki, line 952: Interesting. This updated Ruby 2.5.9 as well since we only have one section for 2.5, which covers both Ruby and JRuby. I wonder if we should split Ruby and JRuby at some point.

In fact I've been wondering for a while if we should extract all these per-version appraisal sections into their own files, something like:

```
# in Appraisals
eval(File.read("appraisals/#{RUBY_ENGINE}_#{RUBY_ENGINE_VERSION.gsub(/\.\d+$/, '')}.appraisals"))
```

This would mirror the prefixes we have in `gemfiles`, the ruby image names, the docker-compose ruby services...

lloeki (approved):
LGTM, a few Qs and thoughts, and maybe a constraint to adjust in Appraisals instead of in a lockfile, but I won't bark at that too much since this finally makes things go GREEN 💚 

Thanks a lot for that last mile!

lloeki:
Duh, CI blocks because of an endless stream of `log writing failed. closed stream`:

```
Finished in 3.42 seconds (files took 4.45 seconds to load)
102 examples, 0 failures

Randomized with seed 61951

log writing failed. closed stream
log writing failed. closed stream
log writing failed. closed stream
log writing failed. closed stream
log writing failed. closed stream
log writing failed. closed stream
log writing failed. closed stream
--------8<-------
```

gemfiles/ruby_2.2.10_contrib.gemfile.lock
TonyCTHsu: I got this error, it seems to me resque is using a wrong version of redis. I guess these versions of gems are so old that probably not maintained hence the dependencies could be prone the error.

```
Failure/Error: ::Resque.redis = url

NoMethodError:
  undefined method `connect' for Redis:Class
Shared Example Group: "job execution tracing" called from ./spec/datadog/tracing/contrib/resque/instrumentation_spec.rb:200
/usr/local/bundle/gems/resque-1.27.4/lib/resque.rb:116:in `redis='
./spec/datadog/tracing/contrib/resque/instrumentation_spec.rb:44:in `block (2 levels) in <top (required)>'
./spec/datadog/tracing/contrib/resque/instrumentation_spec.rb:56:in `block (2 levels) in <top (required)>'
./spec/datadog/tracing/contrib/support/tracer_helpers.rb:96:in `block (2 levels) in <module:TracerHelpers>'
./spec/spec_helper.rb:220:in `block (2 levels) in <top (required)>'
./spec/spec_helper.rb:112:in `block (2 levels) in <top (required)>'
```

Appraisals
TonyCTHsu: I like the idea of seperation. Currently with this setup, any changes with 2.5 will have to run

```
bundle exec rake appraisal:install[jruby-9.2.8.0,jruby-9.2,2.5]
```

PR #2570, opened 2023-01-19
Post #2558 and #2569 merge cleanups

lloeki:
**What does this PR do?**

- address review comments form #2558 
- resolve a merge conflict that restored a deleted file when merging master into either #2558 or #2569
- stop removing `Gemfile.lock` since that's not needed anymore

**Motivation**

There were a few dangling bits, and it appears a merging conflict was misresovled.

**Additional Notes**

None

**How to test the change?**

```
bundle exec rake appraisal:install
```


tasks/update_appraisal_gemfiles.rake
```diff
@@ -1,68 +0,0 @@
...
-         "bundle #{forced_bundler_version} exec appraisal install --without force-appraisal-to-always-run-bundle-install'"
-       ].join
-  end
-end
-
-desc 'Update ALL gems based on Appraisals and Gemfile changes, ' \
-     'accepts list of tracer versions as task argument, defaults to all versions.'
-task :update_appraisal_gemfiles do |_task, args|
-  tracer_version_arg = args.to_a
-  versions = tracer_version_arg.empty? ? TRACER_VERSIONS : tracer_version_arg
```
ivoanjo, line 54: Since we've trained some of our occasional contributors to use these tasks, what do you think of perhaps leaving behind these tasks, but having them only print a message saying "You should use XXX instead. We'll remove this message for dd-trace-rb 2.0"?

That way they don't need to wonder why they've rebased their branch and suddenly this is missing and what they need to do next :)

ivoanjo (approved):
Change looks good! I tested `bundle exec rake appraisal:install` locally and it worked as well.

tasks/update_appraisal_gemfiles.rake
lloeki: Hmm, the old tasks are kind of broken, maybe having these be shims and either call the new tasks or just fail and print out a message directing them to the new ones?

Re: 2.0, do we consider such dev tasks to be part of the public API?

ivoanjo: > Hmm, the old tasks are kind of broken, maybe having these be shims and either call the new tasks (with a message, but I fear the message would be lost in the output) or just fail and print out a message directing them to the new ones?

I'm thinking just printing a message saying "we removed this, please use XX instead" -- no need to get fancy.

> Re: 2.0, do we consider such dev tasks to be part of the public API?

Nope! My thinking was that we already have a bunch of things marked as "# remove later when we go 2.0", so if we leave such a note here we'll remember to remove it (and by that time, nobody will be trying to use this).

lloeki: Done

tasks/appraisal.rake
```diff
@@ -166,7 +162,7 @@ TRACER_VERSIONS = [
   '3.0',
   '3.1',
   '3.2',
-  'jruby-9.2.8.0', # TODO: disabled for possible removal
+  # 'jruby-9.2.8.0', # TODO: disabled for possible removal
```
ivoanjo, line 165: Ooops! Messed that one up when I moved to an array, thanks for catching it!

ivoanjo (approved):
:+1: LGTM

PR #2565, opened 2023-01-18
Pin bundler to 2.3.6 for jruby 9.2.8.0

lloeki:
**What does this PR do?**

Restore a fix for https://github.com/DataDog/dd-trace-rb/issues/2443

**Motivation**

The above issue is flaky and went under the radar for a while.

**Additional Notes**

Nope

**How to test the change?**

This should call bundler 2.3.6, always:

```
rm -vf Gemfile.lock && bundle -v && bundle install && bundle exec rake 'install_appraisal_gemfiles[jruby-9.2.8.0]'
```

and *not* fail with:

```
>> BUNDLE_GEMFILE=/app/gemfiles/jruby_9.2.8.0_hanami_1.gemfile bundle exec _2.3.6_ install --without force-appraisal-to-always-run-bundle-install
/opt/jruby/lib/ruby/stdlib/rubygems/defaults/jruby.rb:34: warning: constant Gem::ConfigMap is deprecated
..
bundler: command not found: _2.3.6_
Install missing gem executables with `bundle install`
ERROR: 1
rake aborted!
```

PR #2448, opened 2022-11-30
Override appraisal bundler versions on legacy rubies

ivoanjo:
**What does this PR do?**:

This PR fixes #2443 and #2444 by forcing the affected Rubies to run on older versions of bundler.

With this PR I could run `bundler exec rake install_appraisal_gemfiles` cleanly on my machine.

**Motivation**:

Make sure the `install_appraisal_gemfiles` works correctly.

**Additional Notes**:

(None)

**How to test the change?**:

Check that the affected rake tasks work correctly.

---

Fixes #2443
Fixes #2444

PR #2572, opened 2023-01-20
Test only with JRuby 9.2.21.0 in CI (disable 9.2.8.0)

ivoanjo:
**What does this PR do?**:

This PR disables testing of JRuby 9.2.8.0 in CI, as well as appraisal generation for it (as well as removes the existing appraisals).

**Motivation**:

Maintaining JRuby 9.2.8.0 working has been quite a burden: this version has many, many, many known issues -- you will notice that there have been **13** other bugfix releases on top of it; latest JRuby is 9.2.21.0.

Some examples of PRs where this JRuby version caused toil are
 #2565, #2558, #2448, #2412, #1869, etc

We've decided to stop running it in CI for now, and plan to evaluate if we should remove it entirely from our test harness.

We recommend that all our customers use JRuby 9.2.21.0, and also recommend evaluating the move to JRuby 9.3, as the JRuby developers have marked JRuby 9.2 as end-of-life:
<https://twitter.com/jruby/status/1611063274459566106>.

**Additional Notes**:

I didn't have to remove the appraisal gemfiles from the tree, but since I disabled the tasks that automatically keep them up-to-date, they wouldn't be maintained anyway.

If/when we do want to revert this change, we should regenerate them using the usual rake tasks.

**How to test the change?**:

Validate that CI is green and that the JRuby 9.2.8.0 configuration is not shown there.


tasks/appraisal.rake
```diff
@@ -166,11 +166,11 @@ TRACER_VERSIONS = %w[
   3.0
   3.1
   3.2
-  jruby-9.2.8.0
```
lloeki, line 169: Maybe we should convert this `%w[` to an actual array, so that we can add the comment inline, e.g:

```
3.1
3.2
#jruby-9.2.8.0 # TODO: jruby-9.2.8.0 disabled for possible removal
jruby-9.3
jruby-9.4
```

Minor sidenote: I'm not quite a fan of this Rubocop rule of forcing people to use `%w[]` % the likes. As an old Perl user I get where it comes from and there are definitely legit uses where it really adds to readability, but I feel like most of the time it doesn't really add to readability, or forces inconsistency and breaks rhythm which hampers readability.

The above is an example of that, where the jruby-9.2.8.0 specific comment gets to be hoisted out because of Rubocop.

lloeki (approved):
LGTM, although since we remove image generation from GHA, we should probably remove its use from `docker-compose.yml` as well.

tasks/appraisal.rake
ivoanjo: Yeah, I really dislike this Rubocop rule as well. I can never remember what's the correct one to use...

Fixed in 25534a2425a70c78d4aeee507bcf47196e25fdb6 .

TonyCTHsu: Don't like `%w[]` +1

ivoanjo: Thanks for the feedback about the `%w[]` y'all. I've jotted a note to open a separate PR to tweak our Rubocop rules about this :)

PR #2577, opened 2023-01-23
Update integrations documentation

GustavoCaso:
**What does this PR do?**
While testing some of the [integration](https://github.com/DataDog/dd-trace-rb/tree/master/integration) apps I noticed a couple things:

- We do not respect the `-v` when building the ruby images. So we have to manually remove the ruby images from the list.
- When running `docker-compose up` from inside the different apps the ruby version is hardcoded. I made it so we can changed  the version using the `RUBY_VERSION` env var from each app folder.

### DIFF
--- a/integration/apps/hanami/.envrc.sample
+++ b/integration/apps/hanami/.envrc.sample
@@ -1,2 +1,3 @@
 export DD_API_KEY=<Your Datadog API key here>
 export DD_HOSTNAME=$(hostname)
+export RUBY_VERSION=<Ruby version you are testing>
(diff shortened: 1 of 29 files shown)


integration/apps/hanami/.envrc.sample
```diff
@@ -1,2 +1,3 @@
 export DD_API_KEY=<Your Datadog API key here>
 export DD_HOSTNAME=$(hostname)
+export RUBY_VERSION=<Ruby version you are testing>
```
TonyCTHsu, line 3: I suggested provided a benign default value for the file. 

```suggestion
export RUBY_VERSION=2.7
```

TonyCTHsu (approved):

integration/apps/hanami/.envrc.sample
GustavoCaso: 👍

PR #2578, opened 2023-01-23
Update Event SDK with latest RFC changes

lloeki:
**What does this PR do?**

- Separate login and failure events
- Add a namespace (pegged to `:appsec`) for the generic event generator

**Motivation**

RFC for event SDK was updated following some additional feedback.

**Additional Notes**

Nope

**How to test the change?**

```
bundle exec rspec spec/datadog/kit/events_spec.rb
```

### DIFF
--- a/lib/datadog/kit/events.rb
+++ b/lib/datadog/kit/events.rb
@@ -7,41 +7,57 @@ module Datadog
   module Kit
     # Tracking events
     module Events
-      # Attach login event information to the trace
+      # Attach login success event information to the trace
       #
       # @param trace [TraceOperation] Trace to attach data to.
-      # @param success [bool] Whether login operation was successful or not.
       # @param user [Hash<Symbol, String>] User information to pass to
       #   Datadog::Kit::Identity.set_user. Must contain at least :id as key.
       # @param others [Hash<String || Symbol, String>] Additional free-form
       #   event information to attach to the trace.
-      def self.track_login(trace, success:, user:, **others)
-        subtag = success ? 'success' : 'failure'
-        event = "users.login.#{subtag}"
+      def self.track_login_success(trace, user:, **others)
+        event = 'users.login.success'
 
-        track(event, trace, **others)
+        track(:appsec, event, trace, **others)
 
         user_options = user.dup
         user_id = user.delete(:id)
 
         raise ArgumentError, 'missing required key: :user => { :id }' if user_id.nil?
 
-        if success
-          Kit::Identity.set_user(trace, id: user_id, **user_options)
-        else
-          trace.set_tag("appsec.events.users.login.#{subtag}.usr.id", user_id)
-        end
+        Kit::Identity.set_user(trace, id: user_id, **user_options)
+      end
+
+      # Attach login failure event information to the trace
+      #
+      # @param trace [TraceOperation] Trace to attach data to.
+      # @param user_id [String] User id that attempted login
+      #   Datadog::Kit::Identity.set_user. Must contain at least :id as key.
+      # @param user_exists [bool] Whether the user id that did a login attempt exists.
+      # @param others [Hash<String || Symbol, String>] Additional free-form
+      #   event information to attach to the trace.
+      def self.track_login_failure(trace, user_id:, user_exists:, **others)
+        event = 'users.login.failure'
+
+        track(:appsec, event, trace, **others)
+
+        raise ArgumentError, 'missing required key: :user => { :id }' if user_id.nil?
+
+        trace.set_tag('appsec.events.users.login.failure.usr.id', user_id)
+        trace.set_tag('appsec.events.users.login.failure.usr.exists', user_exists)
       end
 
       # Attach custom event information to the trace
       #
+      # @param namespace [Symbol] Mandatory. Event namespace. Only :appsec is supported.
       # @param event [String] Mandatory. Event code.
       # @param trace [TraceOperation] Trace to attach data to.
       # @param others [Hash<Symbol, String>] Additional free-form
       #   event information to attach to the trace. Key must not
       #   be :track.
-      def self.track(event, trace, **others)
-        trace.set_tag("appsec.events.#{event}.track", 'true')
+      def self.track(namespace, event, trace, **others)
+        raise ArgumentError, "namespace cannot be #{namespace.inspect}" if namespace.to_sym != :appsec
+
+        trace.set_tag("#{namespace}.events.#{event}.track", 'true')
 
         others.each do |k, v|
           raise ArgumentError, 'key cannot be :track' if k.to_sym == :track
--- a/spec/datadog/kit/events_spec.rb
+++ b/spec/datadog/kit/events_spec.rb
@@ -10,64 +10,87 @@ require 'datadog/kit/events'
 RSpec.describe Datadog::Kit::Events do
   subject(:trace_op) { Datadog::Tracing::TraceOperation.new }
 
-  describe '#track_login' do
-    context 'login success' do
-      it 'sets event tracking key on trace' do
-        trace_op.measure('root') do
-          described_class.track_login(trace_op, user: { id: '42' }, success: true)
-        end
-        trace = trace_op.flush!
-        expect(trace.send(:meta)).to include('appsec.events.users.login.success.track' => 'true')
+  describe '#track_login_success' do
+    it 'sets event tracking key on trace' do
+      trace_op.measure('root') do
+        described_class.track_login_success(trace_op, user: { id: '42' })
       end
-
-      it 'sets user id on trace' do
-        trace_op.measure('root') do
-          described_class.track_login(trace_op, user: { id: '42' }, success: true)
-        end
-        trace = trace_op.flush!
-        expect(trace.send(:meta)).to include('usr.id' => '42')
-      end
-
-      it 'sets other keys on trace' do
-        trace_op.measure('root') do
-          described_class.track_login(trace_op, user: { id: '42' }, success: true, foo: 'bar')
-        end
-        trace = trace_op.flush!
-        expect(trace.send(:meta)).to include('usr.id' => '42', 'appsec.events.users.login.success.foo' => 'bar')
-      end
+      trace = trace_op.flush!
+      expect(trace.send(:meta)).to include('appsec.events.users.login.success.track' => 'true')
     end
 
-    context 'login failure' do
-      it 'sets event tracking key on trace' do
-        trace_op.measure('root') do
-          described_class.track_login(trace_op, user: { id: '42' }, success: false)
-        end
-        trace = trace_op.flush!
-        expect(trace.send(:meta)).to include('appsec.events.users.login.failure.track' => 'true')
+    it 'sets user id on trace' do
+      trace_op.measure('root') do
+        described_class.track_login_success(trace_op, user: { id: '42' })
       end
+      trace = trace_op.flush!
+      expect(trace.send(:meta)).to include('usr.id' => '42')
+    end
 
-      it 'sets user id on trace' do
-        trace_op.measure('root') do
-          described_class.track_login(trace_op, user: { id: '42' }, success: false)
-        end
-        trace = trace_op.flush!
-        expect(trace.send(:meta)).to include('appsec.events.users.login.failure.usr.id' => '42')
+    it 'sets other keys on trace' do
+      trace_op.measure('root') do
+        described_class.track_login_success(trace_op, user: { id: '42' }, foo: 'bar')
       end
+      trace = trace_op.flush!
+      expect(trace.send(:meta)).to include('usr.id' => '42', 'appsec.events.users.login.success.foo' => 'bar')
+    end
+  end
 
-      it 'sets other keys on trace' do
-        trace_op.measure('root') do
-          described_class.track_login(trace_op, user: { id: '42' }, success: false, foo: 'bar')
-        end
-        trace = trace_op.flush!
-        expect(trace.send(:meta)).to include('appsec.events.users.login.failure.foo' => 'bar')
+  describe '#track_login_failure' do
+    it 'sets event tracking key on trace' do
+      trace_op.measure('root') do
+        described_class.track_login_failure(trace_op, user_id: '42', user_exists: true)
       end
+      trace = trace_op.flush!
+      expect(trace.send(:meta)).to include('appsec.events.users.login.failure.track' => 'true')
+    end
+
+    it 'sets user id on trace' do
+      trace_op.measure('root') do
+        described_class.track_login_failure(trace_op, user_id: '42', user_exists: true)
+      end
+      trace = trace_op.flush!
+      expect(trace.send(:meta)).to include('appsec.events.users.login.failure.usr.id' => '42')
+    end
+
+    it 'sets user existence on trace' do
+      trace_op.measure('root') do
+        described_class.track_login_failure(trace_op, user_id: '42', user_exists: true)
+      end
+      trace = trace_op.flush!
+      expect(trace.send(:meta)).to include('appsec.events.users.login.failure.usr.exists' => 'true')
+    end
+
+    it 'sets user non-existence  on trace' do
+      trace_op.measure('root') do
+        described_class.track_login_failure(trace_op, user_id: '42', user_exists: false)
+      end
+      trace = trace_op.flush!
+      expect(trace.send(:meta)).to include('appsec.events.users.login.failure.usr.exists' => 'false')
+    end
+
+    it 'sets other keys on trace' do
+      trace_op.measure('root') do
+        described_class.track_login_failure(trace_op, user_id: '42', user_exists: true, foo: 'bar')
+      end
+      trace = trace_op.flush!
+      expect(trace.send(:meta)).to include('appsec.events.users.login.failure.foo' => 'bar')
     end
   end
 
   describe '#track' do
+    it 'rejects unexpected namespaces' do
+      trace_op.measure('root') do
+        expect { described_class.track(:foo, 'bar', trace_op) }.to raise_error ArgumentError
+      end
+
+      trace = trace_op.flush!
+      expect(trace.send(:meta)).to_not include('foo.events.bar.track' => 'true')
+    end
+
     it 'sets event tracking key on trace' do
       trace_op.measure('root') do
-        described_class.track('foo', trace_op)
+        described_class.track(:appsec, 'foo', trace_op)
       end
       trace = trace_op.flush!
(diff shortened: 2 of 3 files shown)


lib/datadog/kit/events.rb
```diff
@@ -7,41 +7,63 @@ module Datadog
...
+      # @param namespace [Symbol] Mandatory. Event namespace. Only :appsec is supported.
       # @param event [String] Mandatory. Event code.
       # @param trace [TraceOperation] Trace to attach data to.
       # @param others [Hash<Symbol, String>] Additional free-form
       #   event information to attach to the trace. Key must not
       #   be :track.
-      def self.track(event, trace, **others)
-        trace.set_tag("appsec.events.#{event}.track", 'true')
+      def self.track(namespace, event, trace, **others)
+        raise ArgumentError, "namespace cannot be #{namespace.inspect}" if namespace.to_sym != :appsec
```
GustavoCaso, line 64: By reading the RFC for the track API https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2755793809/Application+Security+Events+Tracking+API+SDK

I can not help to wonder if we need to raise an exception if the namespace is not `appsec`. Do we think other namespaces would use this API outside of `appsec`?

lib/datadog/kit/events.rb
```diff
@@ -7,41 +7,63 @@ module Datadog
...
-      # @param success [bool] Whether login operation was successful or not.
       # @param user [Hash<Symbol, String>] User information to pass to
       #   Datadog::Kit::Identity.set_user. Must contain at least :id as key.
       # @param others [Hash<String || Symbol, String>] Additional free-form
       #   event information to attach to the trace.
-      def self.track_login(trace, success:, user:, **others)
-        subtag = success ? 'success' : 'failure'
-        event = "users.login.#{subtag}"
+      def self.track_login_success(trace, user:, **others)
+        event = 'users.login.success'
```
GustavoCaso, line 20: Should we extract this into a constant? Same for the `'users.login.failure'` string

lloeki: Good idea

lib/datadog/kit/events.rb
lloeki: There's a use case where customers used that for business events and complained that disabling AppSec also disabled these events, which they still wanted to be reported even though they may not be processed by AppSec backend-side, and that ended up being troublesome for them.

The reverse was true too, as some events must be disabled when AppSec is disabled.

These use cases don't appear to be quite defined yet, so I wanted to keep room for that while being explicit, with the following consequence:

- Anything appsec will be disabled when appsec is disabled
- Anything not appsec will stay enabled when appsec is disabled, whether it ends up being used or not (it sets user-visible tags, and users can do things backend side with user visible tags outside of appsec as they see fit, like filters and triggers and whatnot)

Granted, the current code prevents the second case, but it does so explicitly while leaving room for that to change.

GustavoCaso: I understand. I like the idea of leaving things open for the second use case to be introduce.

Thanks for the detailed explanation

GustavoCaso (approved):

lib/datadog/kit/events.rb
```diff
@@ -7,41 +7,62 @@ module Datadog
   module Kit
     # Tracking events
     module Events
-      # Attach login event information to the trace
+      APPSEC_LOGIN_SUCCESS_EVENT = 'users.login.success'
+      APPSEC_LOGIN_FAILURE_EVENT = 'users.login.failure'
```
ivoanjo, line 11: On naming, it seems a bit weird to call these things `APPSEC_..._EVENT`s since we're inside `datadog/kit` and I thought the reason to have this inside kit rather than inside appsec is because this is intended to be generic.

(E.g. should we at least hide these constants?)

lib/datadog/kit/events.rb
```diff
@@ -7,41 +7,62 @@ module Datadog
...
+
+      # Attach login success event information to the trace
+      #
+      # This method is experimental and may change in the future.
```
ivoanjo, line 15: I think this approach of having something and declaring it experimental is fine BUT should we be more specific on what users can expect?

E.g. does this mean that an API added on 1.9 can be removed on 1.10? Or would we just deprecate and no-op it?

lib/datadog/kit/events.rb
```diff
@@ -7,41 +7,62 @@ module Datadog
...
-        event = "users.login.#{subtag}"
-
-        track(event, trace, **others)
+      def self.track_login_success(trace, user:, **others)
+        track(:appsec, APPSEC_LOGIN_SUCCESS_EVENT, trace, **others)
 
         user_options = user.dup
         user_id = user.delete(:id)
 
         raise ArgumentError, 'missing required key: :user => { :id }' if user_id.nil?
```
ivoanjo, line 28: Minor: We could probably simplify this by doing `if user.key?(:id)` -- we're splatting the `user` in the last line so this dup/mutate ends up being kinda useless.

lib/datadog/kit/events.rb
```diff
@@ -7,41 +7,62 @@ module Datadog
...
+        Kit::Identity.set_user(trace, id: user_id, **user_options)
+      end
+
+      # Attach login failure event information to the trace
+      #
+      # This method is experimental and may change in the future.
+      #
+      # @param trace [TraceOperation] Trace to attach data to.
+      # @param user_id [String] User id that attempted login
+      #   Datadog::Kit::Identity.set_user. Must contain at least :id as key.
```
ivoanjo, line 39: This description doesn't seem correct -- it mentions that this is supposed to be the `user_id`, but then mentions it needs to contain `:id` as the key. Can you take a second look?

lib/datadog/kit/events.rb
```diff
@@ -7,41 +7,62 @@ module Datadog
...
+      # @param user_exists [bool] Whether the user id that did a login attempt exists.
+      # @param others [Hash<String || Symbol, String>] Additional free-form
+      #   event information to attach to the trace.
+      def self.track_login_failure(trace, user_id:, user_exists:, **others)
+        track(:appsec, APPSEC_LOGIN_FAILURE_EVENT, trace, **others)
+
+        raise ArgumentError, 'missing required key: :user => { :id }' if user_id.nil?
```
ivoanjo, line 46: Continuing from my comment on the description of `user_id` above, this error description seems odd as well -- it mentions both a `:user` and an `:id` and then that's not what's being checked.

lib/datadog/kit/events.rb
```diff
@@ -7,41 +7,62 @@ module Datadog
...
+      #
+      # @param namespace [Symbol] Mandatory. Event namespace. Only :appsec is supported.
       # @param event [String] Mandatory. Event code.
       # @param trace [TraceOperation] Trace to attach data to.
       # @param others [Hash<Symbol, String>] Additional free-form
       #   event information to attach to the trace. Key must not
       #   be :track.
-      def self.track(event, trace, **others)
-        trace.set_tag("appsec.events.#{event}.track", 'true')
+      def self.track(namespace, event, trace, **others)
```
ivoanjo, line 62: Since it's not entirely clear how this is going to be used going forward (or even beyond appsec), have you considered making this method private for now, e.g. making it so it can only be called by other methods in this file?

lib/datadog/kit/events.rb
```diff
@@ -7,41 +7,62 @@ module Datadog
...
+        raise ArgumentError, "namespace cannot be #{namespace.inspect}" if namespace.to_sym != :appsec
```
ivoanjo, line 63: Minor: I suggest extending the error to make it a bit more self-describable.

E.g. right now if I see `ArgumentError: namespace cannot be :potato` I've learned that I did something wrong, but I'll need to go read the code or the API docs to understand what I did wrong exactly and how to fix it; whereas with `ArgumentError: namespace cannot be :potato, only :appsec is allowed` then I immediately know that I need to either switch to the allowed namespace or reconsider using this API.

spec/datadog/kit/events_spec.rb
```diff
@@ -10,72 +10,95 @@
...
-          described_class.track_login(trace_op, user: { id: '42' }, success: true, foo: 'bar')
-        end
-        trace = trace_op.flush!
-        expect(trace.send(:meta)).to include('usr.id' => '42', 'appsec.events.users.login.success.foo' => 'bar')
+    it 'sets other keys on trace' do
+      trace_op.measure('root') do
+        described_class.track_login_success(trace_op, user: { id: '42' }, foo: 'bar')
       end
+      trace = trace_op.flush!
+      expect(trace.send(:meta)).to include('usr.id' => '42', 'appsec.events.users.login.success.foo' => 'bar')
```
ivoanjo, line 35: Minor: This bit seems to repeat quite a lot

```ruby
trace = trace_op.flush!
trace.send(:meta)
```

consider maybe making it a let?

spec/datadog/kit/events_spec.rb
```diff
@@ -10,72 +10,95 @@
...
+      end
+      trace = trace_op.flush!
+      expect(trace.send(:meta)).to include('appsec.events.users.login.failure.foo' => 'bar')
     end
   end
 
   describe '#track' do
+    it 'rejects unexpected namespaces' do
+      trace_op.measure('root') do
+        expect { described_class.track(:foo, 'bar', trace_op) }.to raise_error ArgumentError
```
ivoanjo, line 84: Minor: While I don't think it's a good idea to match on the exact string, I suggest also not being too vague to avoid a test passing due to other kinds of argument errors.

Aka something like `.to raise_error(ArgumentError, /namespace cannot be/)`

spec/datadog/kit/events_spec.rb
```diff
@@ -10,72 +10,95 @@
...
+      expect(trace.send(:meta)).to include('appsec.events.users.login.failure.track' => 'true')
+    end
 
-      it 'sets user id on trace' do
-        trace_op.measure('root') do
-          described_class.track_login(trace_op, user: { id: '42' }, success: false)
-        end
-        trace = trace_op.flush!
-        expect(trace.send(:meta)).to include('appsec.events.users.login.failure.usr.id' => '42')
+    it 'sets user id on trace' do
```
ivoanjo, line 48: Minor: The test description here is exactly the same as in track_login_success ("sets user id on trace") and yet we set different tags on both. I would suggest maybe changing the test description on this one ("sets the appsec failure user id on the trace" or something similar) to highlight this fact :)

ivoanjo (commented):
Left a few nitpicks, but overall this looks quite reasonable! :)

lib/datadog/kit/events.rb
GustavoCaso: That is a very interesting point. If this API will be available for users and we later want to change it it would be hard to do. 

I agree with the sentiment of making it private until we have more knowledge of how `track` is going to be use

lib/datadog/kit/events.rb
lloeki: It won't be removed but the signature may change, and maybe the method names or namespace, but hopefully not, I'm trying to nail that down before merging.

lib/datadog/kit/events.rb
lloeki: Good catch

lib/datadog/kit/events.rb
lloeki: It's part of the public API of the event RFC, so we can't go private here. I'm thinking about moving these public things either to the `Datadog::AppSec` namespace (that would mean introducing something like `Datadog::AppSec::Kit`) or to a `Datadog::Kit::AppSec` namespace.

spec/datadog/kit/events_spec.rb
lloeki: Hmm, I have to think about it, I'm not to keen on `let` that do "actions", as it creates magic that gets unwieldy.

I've suffered a lot through that in some specs (IIRC Rack or Sinatra ones). If I can figure something sane quickly I'll do it, otherwise I'd favour explicitness in tests because it can definitely enhance readability even though it's repetitive.

lib/datadog/kit/events.rb
lloeki: Good catch

lib/datadog/kit/events.rb
lloeki: Good catch. The dreaded comment that doesn't get updated in sync!

lib/datadog/kit/events.rb
lloeki: Yeah I'm not too happy with the whole thing, but I don't feel like it belongs to `Datadog::AppSec` because of the level of abstraction: this is kind of sugar above Tracing (like `Datadog::Kit::Identity.set_user`) and glue between AppSec and Tracing.

lib/datadog/kit/events.rb
ivoanjo: I think any of the ones you mentioned are fine, as long as we document that (if I'm understanding what you're saying) it can break even between dd-trace-rb minor versions -- just so customers that pick it up are aware of this possibility.

lib/datadog/kit/events.rb
lloeki: Ah I seem to recall why I did that: static typing. `id:` being a mandatory kwarg for `set_user`, it can't be known statically that it's in the option hash, and even if it is, of which type (only the hash generic value type is known), thus typing failed.

I'll add that as an upstream issue because it *could* at least be known with a `if user.key?(:id)` check but `steep` doesn't seem to support that (maybe intentional, on grounds of concurrent mutability) and if the hash value type matches then it should be able to pass.

lib/datadog/kit/events.rb
lloeki: I will try a namespace move in a subsequent PR so that we can merge this one.

spec/datadog/kit/events_spec.rb
lloeki: OK I've done that.

lib/datadog/kit/events.rb
delner: I agree with @ivoanjo here: I think this concrete implementation should be isolated to `Security`.

Is `Kit` supposed to be product aware? If so, when and why would someone use `Kit`? I think I need clarity about the purpose of this namespace.

lib/datadog/kit/events.rb
delner: Hmmm, maybe we could add a YARD doc annotation like we did for `@public_api`? `@experimental`?

Alternatively, if the user wouldn't call this method directly, it may be worth privatizing until it's stable.

lib/datadog/kit/events.rb
```diff
@@ -7,41 +7,63 @@ module Datadog
...
+      # @param user_id [String] User id that attempted login
+      # @param user_exists [bool] Whether the user id that did a login attempt exists.
+      # @param others [Hash<String || Symbol, String>] Additional free-form
+      #   event information to attach to the trace.
+      def self.track_login_failure(trace, user_id:, user_exists:, **others)
+        track(:appsec, APPSEC_LOGIN_FAILURE_EVENT, trace, **others)
+
+        raise ArgumentError, 'user_id cannot be nil' if user_id.nil?
+
+        trace.set_tag('appsec.events.users.login.failure.usr.id', user_id)
```
delner, line 47: Are these simple, unidirectional tags? A.k.a. they get set once, never read, and forwarded along to the Datadog backend?

(As opposed to using tags to facilitate communication between different parts of our instrumentation/products in the Ruby app...)

delner (commented):
Only hangup on my end is what we're doing with `Kit`. I think we could discuss this more, figure out what its supposed to be, and how this fits in.

lib/datadog/kit/events.rb
lloeki: > should we at least hide these constants

What would be the ddtrace-idiomatic way to "hide" constants?

ivoanjo: I usually use `private_constant` for that:

https://github.com/DataDog/dd-trace-rb/blob/1dac1c1281e3fc3672037a2b18957f54491060dd/lib/datadog/profiling.rb#L11

lib/datadog/kit/events.rb
lloeki: Like `Identity::set_user`, this is a sugary public API to be used by customers to help them annotate traces with a higher level of abstraction. It can't be made private.

Regarding `@experimental`, I would love that, although it sounds a bit scary: since we are going to release that as a public beta feature I would love to have a `@beta` annotation as well.

To me the difference would be:

- experimental: "we're attempting something that you can try if you find it useful, but we're exploring something, so all bets are off and it might change in any shape or form, or may very well vanish in the future"
- beta: "this is something that's not quite stable yet, it will probably not change much but it still may, and will most likely _not_ vanish from under your feet"

lib/datadog/kit/events.rb
GustavoCaso: Can we use `private_constant`? 

I grep the codebase and it seems we use it in some places

lib/datadog/kit/events.rb
lloeki: > Are these simple, unidirectional tags?

Unequivocally yes.

lib/datadog/kit/events.rb
GustavoCaso: One interesting thing we could consider is having private modules within the scope ok `Datadog::Kit`

Example: 
```ruby
module Datadog
  module Kit
    # Tracking events
    module EventSDK
      class << self
        def track(namespace, event, trace, **rest)
          puts namespace
          puts event
          puts trace
          puts rest
        end
      end
    end

    private_constant :EventSDK

    module Events

      module Appsec
        LOGIN_SUCCESS_EVENT = 'users.login.success'
        LOGIN_FAILURE_EVENT = 'users.login.failure'
      end

      def self.track_login_success(trace, user:, **others)
        EventSDK.track(:appsec, Appsec::LOGIN_SUCCESS_EVENT, nil, **others)
      end

      def self.track_login_failure(trace, user_id:, user_exists:, **others)
        EventSDK.track(:appsec, Appsec::LOGIN_FAILURE_EVENT, nil, **others)
      end

      def self.track(namespace, event_name, trace, user:, **others)
        EventSDK.track(namespace, event_name, nil, **others)
      end
    end
  end
end


Datadog::Kit::Events.track_login_success(nil, user: "me")

DataDog::Kit::EventSDK.track(:appsec, Appsec::LOGIN_FAILURE_EVENT, nil)
# test.rb:42:in `<main>': uninitialized constant DataDog (NameError)

# DataDog::Kit::EventSDK.track(:appsec, Appsec::LOGIN_FAILURE_EVENT, nil)
# ^^^^^^^^^^
# Did you mean?  Datadog
```

PR #2581, opened 2023-01-24
Skip simplecov for development

TonyCTHsu:
**What does this PR do?**

Skip simplecov executino for develop environment by default from environment variable

**Motivation**

I am tired of typing `export SKIP_SIMPLECOV=1` every time I boot a container for development.


ivoanjo (approved):
:+1: LGTM

lloeki:
Ha I used to either prefix commands with `env SKIP_SIMPLECOV=1 ...` or start docker with `-e SKIP_SIMPLECOV=1` so thanks for that.

BTW it feels like the condition is reversed somehow, like it should be `ENABLE_SIMPLECOV=1` set where it matters and that the default should be without it.

PR #2584, opened 2023-01-25
Add hanami getting started doc

TonyCTHsu:
**What does this PR do?**

Add hanami instrumentation is the Getting started doc

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -55,6 +55,7 @@ To contribute, check out the [contribution guidelines][contribution docs] and [d
      - [Grape](#grape)
      - [GraphQL](#graphql)
      - [gRPC](#grpc)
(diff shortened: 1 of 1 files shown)


docs/GettingStarted.md
```diff
@@ -1169,6 +1171,29 @@ alternate_client = Demo::Echo::Service.rpc_stub_class.new(
...
+```
+
+and create an initializer file in your `config/initializers` folder:
+
+```ruby
+# config/initializers/datadog.rb
+Datadog.configure do |c|
+  c.tracing.instrument :hanami, **options
+end
+```
```
ivoanjo, line 1189: Should we clarify that this is optional? (Since the defaults are applied by auto_instrument?)

ivoanjo (approved):
:+1: LGTM

PR #2587, opened 2023-01-26
Move events to appsec specific namespace

lloeki:
**What does this PR do?**

Move events to AppSec specific namespace.

**Motivation**

The general use case is unclear. Let's stick with AppSec.

**Additional Notes**

Why `Kit::AppSec` and not `AppSec::Kit`?

- These are actually high-level sugary trace annotation methods, the same way `Identity#set_user` is
- This depends more on Tracing than AppSec, these are "glue" methods
- There is discussion as to whether some of these cases should still push events even when AppSec is blanket disabled, which would presumably disable the whole of AppSec namespace, possibly not even `require`-ing it

Follow-up to #2578 

**How to test the change?**

Specs should cover it.


ivoanjo (approved):
I'm gonna say it -- I still don't fully understand why `Kit` exists, and I understand less `Kit::AppSec`.

(My reasons are -- `Kit` is already not product-independent, it's for traces, and I even more don't understand what's the meaning of having an `AppSec` namespace under this supposedly product-independent namespace)

Having said that, I also feel we have bigger fish to fry, so I'll trust you and others that chimed in on #2578 that we're on the right path, and well, if it turns out we aren't, like everything else, we'll correct it later :)

So here's my approval.

lloeki:
I hear you @ivoanjo and am not happy at the situation either. I do have some ideas to improve on that but they would be somewhat impactful.

I will do my best to explain internally why I initially created Kit, what I think its scope is, and why I feel it doesn't fold into other namespaces.

PR #2588, opened 2023-01-27
Reduce the number of literal string and use constant in the appsec library

GustavoCaso:
**What does this PR do?**
While going over the `appsec` library code to familiarise myself  with it. I noticed we have a lot os string literals over the code base. I created a bunch of constant in the `Ext` namespace for each the integrations. I replace the string literals with the constants. 

Also, I added the `# frozen_string_literal: true` on the files I touched. 

Additional I fixed the `LOAD_PATH` for the binary that we ship with the gem `bin/ddtracerb`

**Motivation**
Code cleanup 

**Additional Notes**
Am I using the Ext namespace correctly? 

**How to test the change?**

CI should be green.

### DIFF
--- a/bin/ddtracerb
+++ b/bin/ddtracerb
@@ -1,4 +1,6 @@
 #!/usr/bin/env ruby
+$LOAD_PATH.push File.expand_path('../lib', __dir__)
+
 require 'datadog/profiling/tasks/exec'
 require 'datadog/profiling/tasks/help'
 
--- a/bin/ddtracerb
+++ b/bin/ddtracerb
@@ -1,5 +1,5 @@
 #!/usr/bin/env ruby
-$LOAD_PATH.push File.expand_path('../lib', __dir__)
+$LOAD_PATH.unshift File.expand_path('../lib', __dir__)
 
 require 'datadog/profiling/tasks/exec'
 require 'datadog/profiling/tasks/help'
--- a/lib/datadog/appsec/contrib/rack/ext.rb
+++ b/lib/datadog/appsec/contrib/rack/ext.rb
@@ -11,7 +11,7 @@ module Datadog
           ENV_ENABLED = 'DD_TRACE_RACK_ENABLED' # TODO: DD_APPSEC?
 
           RESPONSE_STATUS = 'response.status'
-          RESQUEST_BODY = 'request.body'
+          REQUEST_BODY = 'request.body'
           REQUEST_HEADERS = 'request.headers'
           REQUEST_URI_RAW = 'request.uri.raw'
           REQUEST_QUERY = 'request.query'
--- a/lib/datadog/appsec/contrib/rack/ext.rb
+++ b/lib/datadog/appsec/contrib/rack/ext.rb
@@ -1,4 +1,5 @@
 # typed: ignore
+# frozen_string_literal: true
 
 module Datadog
   module AppSec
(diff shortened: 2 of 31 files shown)


bin/ddtracerb
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env ruby
+$LOAD_PATH.push File.expand_path('../lib', __dir__)
```
GustavoCaso, line 2: Without this the `require` below do not work

lib/datadog/appsec/contrib/rack/reactive/request.rb
```diff
@@ -9,29 +11,29 @@ module Rack
         module Reactive
           # Dispatch data from a Rack request to the WAF context
           module Request
+            WAF_ADDRESSES = [
```
GustavoCaso, line 14: No need to create a new array for every time we call `self.subscribe`

lib/datadog/appsec/contrib/rails/reactive/action.rb
```diff
@@ -9,22 +11,24 @@ module Rails
         module Reactive
           # Dispatch data from a Rails request to the WAF context
           module Action
+            WAF_ADDRESSES = [
```
GustavoCaso, line 14: No need to create a new array for every time we call `self.subscribe`

TonyCTHsu (approved):

lloeki (changes requested):
While `lib/datadog/appsec/contrib` is currently excluded from RBS static type checking (because of references to dependencies such as Rack, Rails, Sinatra), it would be wise to update `.rbs` in `sig` to match the changes.

lib/datadog/appsec/contrib/rack/ext.rb
```diff
@@ -1,13 +1,26 @@
...
 module Datadog
   module AppSec
     module Contrib
       module Rack
         # Rack integration constants
         module Ext
-          APP = 'rack'.freeze
-          ENV_ENABLED = 'DD_TRACE_RACK_ENABLED'.freeze # TODO: DD_APPSEC?
+          APP = 'rack'
+          ENV_ENABLED = 'DD_TRACE_RACK_ENABLED' # TODO: DD_APPSEC?
```
lloeki, line 11: Is this actually used? If not, unused constants should be removed.

lib/datadog/appsec/contrib/rack/ext.rb
```diff
@@ -1,13 +1,26 @@
...
+
+          RESPONSE_STATUS = 'response.status'
```
lloeki, line 13: These changes should be mirrored in `sig`

lib/datadog/appsec/contrib/rack/reactive/request.rb
lloeki: I would drop the `WAF_` prefix here. Indeed currently only the WAF is called, but it may not be so in the future as some processing parts that would react to these addresses may be implemented outside of libddwaf.

In fact, the names are different from what you can see in `waf_args` (for which the keys are the WAF addresses) _on purpose_.

lib/datadog/appsec/contrib/rack/integration.rb
```diff
@@ -20,7 +22,7 @@ class Integration
           register_as :rack, auto_patch: false
 
           def self.version
-            Gem.loaded_specs['rack'] && Gem.loaded_specs['rack'].version
+            Gem.loaded_specs[Ext::APP] && Gem.loaded_specs[Ext::APP].version
```
lloeki, line 25: I experience cognitive dissonance here: the values may be the same (currently) but the semantics is different: this is not the app name but the actual gem name, so the factorisation is incorrect here.

lloeki (changes requested):

lib/datadog/appsec/contrib/rack/ext.rb
GustavoCaso: They are use in `lib/datadog/appsec/contrib/rack/configuration/settings.rb`

lib/datadog/appsec/contrib/rack/ext.rb
GustavoCaso: Done

lib/datadog/appsec/contrib/rack/reactive/request.rb
GustavoCaso: Done

ivoanjo (commented):
Left a few notes! :)

GustavoCaso:
Closing the PR. 

The PR tries to tackle multiple things. Going to break each change on separate PRs

lloeki:
This is one: https://github.com/DataDog/dd-trace-rb/pull/2597

PR #2590, opened 2023-01-31
Add messaging.system tag to all messaging integrations

rarguelloF:
**What does this PR do?**

Adds messaging.system tag to all messaging integrations.

**Motivation**

This is part of the Unified Naming Convention initiative.

**Additional Notes**

**How to test the change?**


spec/datadog/tracing/contrib/kafka/patcher_spec.rb
```diff
@@ -56,6 +56,7 @@
           expect(span).to_not have_error
           expect(span.get_tag(Datadog::Tracing::Metadata::Ext::TAG_COMPONENT)).to eq('kafka')
           expect(span.get_tag(Datadog::Tracing::Metadata::Ext::TAG_OPERATION)).to eq('connection.request')
+          expect(span.get_tag(Datadog::Tracing::Contrib::Ext::Messaging::TAG_SYSTEM)).to eq('kafka')
```
TonyCTHsu, line 59: Prefer plain text to indirection in test

```suggestion
          expect(span.get_tag('messaging.system)).to eq('kafka')
```

lib/datadog/tracing/contrib/sidekiq/server_internal_tracer/redis_info.rb
```diff
@@ -13,6 +13,8 @@ def redis_info
               Datadog::Tracing.trace(Ext::SPAN_REDIS_INFO, service: configuration[:service_name]) do |span|
                 span.span_type = Datadog::Tracing::Metadata::Ext::AppTypes::TYPE_WORKER
 
+                span.set_tag(Contrib::Ext::Messaging::TAG_SYSTEM, Ext::TAG_MESSAGING_SYSTEM)
```
TonyCTHsu, line 16: Does it make sense to apply the tag when the instrumented operation isn't really related to push/pop a job from a queue?

TonyCTHsu (commented):
Generally fine, but test assertions also should be applied to those failed cases instead of those happy cases.

lib/datadog/tracing/contrib/sidekiq/server_internal_tracer/redis_info.rb
rarguelloF: Hmm in this case I'm not sure what's the most correct way to go, since for example if you are using `sqs` and run the operation `ListQueues`, which is not a push/pop operation but could be considered as type `messaging` since you are doing an operation on a messaging system.

So basically this is the logic I followed to tag these operations as messaging (similar logic could be applied for `db` tags) - also I think making the decision at the integration level makes things easier for future contributors (only need to make these decisions when adding new integrations).

Not a super strong opinion though, so let me know if this makes sense to you.

lib/datadog/tracing/contrib/delayed_job/ext.rb
```diff
@@ -21,6 +21,8 @@ module Ext
           TAG_OPERATION_ENQUEUE = 'enqueue'.freeze
           TAG_OPERATION_JOB = 'job'.freeze
           TAG_OPERATION_RESERVE_JOB = 'reserve_job'.freeze
+
+          TAG_MESSAGING_SYSTEM = 'delayed_job'.freeze
```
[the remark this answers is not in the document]
rarguelloF: Sounds good! I'll use the `TAG_COMPONENT` value for integrations where we are not using the message broker itself like kafka, amazonsqs, rabbitmq, etc.

Also I was wondering if there's a good place to put these whitelisted values as mentioned above, that can appear in several integrations, instead of repeating it in the `ext.rb` file for each one?

TonyCTHsu (approved):

zhamza99 (approved):
LGTM

Issue #2064, opened 2022-06-02
HTTP request queuing duration


ib-drct-dev:
I want to measure how much a request waits in a queue until it is served by an app.

There is a way to generate traces that include the time spent in a queue. The instrumentation produces http_server.queue.duration metric. Which is not exactly I need and expect.

1) Is it possible to compute queue waiting time some how from the information we have? I thought about subtracting `http_server.queue.duration` from `rack.request.duration` but `http_server.queue.duration` filters out some traces so aggregated duration values could be skewed and we won't get accurate values. Maybe there are other options?
2) Maybe it's possible to add another metric which subtracts `X-Request-Start` from the time a request has arrived in an app?
3) Why the produced metric is called `http_server.queue.duration` even it does not compute queue waiting time? Wouldn't `http_server.request.duration` be more accurate name?

https://docs.datadoghq.com/tracing/setup_overview/setup/ruby/#http-request-queuing
https://github.com/DataDog/dd-trace-rb/blob/3cbcea6f1fd240de00200b7135c907f3fd194c9a/lib/datadog/tracing/contrib/rack/middlewares.rb#L116

Thanks!

wantsui:
Regarding metrics such as `trace.http_server.queue.duration` and `trace.rack.request.duration`:

These metrics follow the pattern `trace.<SPAN_NAME>.duration`. According to our [documentation on this metric ](https://docs.datadoghq.com/tracing/metrics/metrics_namespace/#duration), this metric records the collection/sum of all spans in a period of time:

> [LEGACY] Measure the total time for a collection of spans within a time interval, including child spans seen in the collecting service.

What this means is that the values it gives back is an aggregated number which is why the numbers can be large. As a result, when you divide this by `hits`, you can get `average`, seen in the docs as well:

> When trace.<SPAN_NAME>.duration is divided by trace.<SPAN_NAME>.hits, the result can yield an average latency, but this is not the recommended approach for calculating the average latency. Instead, refer to the [Latency Distribution](https://docs.datadoghq.com/tracing/metrics/metrics_namespace/#latency-distribution) section for average latency calculations.

As a result if you tried to do a computation via `trace.http_server.queue.duration - trace.rack.request.duration`, the results wouldn't be useful or accurate for any given request since it's aggregated information. 


[...]

PR #2591, opened 2023-01-31
Only count time spent in queue as part of the `http_server.queue` span

agrobbin:
**What does this PR do?**

This is an attempt at resolving the inconsistency raised in #2064 between `http_server.queue`'s duration and real world server queuing time by introducing an outer `http_server.request` span, which contains `http_server.queue` and `rack.request` as sibling spans, rather than the current parent/child relationship that exists between the 2 of them.

**Motivation**

If you have a request that spends 10ms in queue, and 5s being executed, you currently end up with these 2 spans in APM:

* `http_server.queue` (duration: 5.01s)
  * `rack.request` (duration: 5s)

However, that isn't really semantically correct. The queuing itself did not take 5.01s, the entire lifecycle of the request took 5.01s, with queueing making up 10ms of that time.

After this change, you should end up with something like this:

* `http_server.request` (duration: 5.01s)
  * `http_server.queue` (duration: 10ms)
  * `rack.request` (duration: 5s)

I'd love to get some general feedback on whether this kind of change is reasonable, as it *is* a breaking change (the definition of `http_server.queue` would change for anyone who depends on it). I feel that it is correcting a misleading and confusing APM span, but I'm not the one who maintains this library, or fields inbound customer support requests!

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -1562,7 +1562,7 @@ run app
 | `quantize.query.obfuscate.with` | Defines the string to replace obfuscated matches with. May be a String. Option must be nested inside the `query.obfuscate` option. | `'<redacted>'` |
 | `quantize.query.obfuscate.regex` | Defines the regex with which the query string will be redacted. May be a Regexp, or `:internal` to use the default internal Regexp, which redacts well-known sensitive data. Each match is redacted entirely by replacing it with `query.obfuscate.with`. Option must be nested inside the `query.obfuscate` option. | `:internal` |
 | `quantize.fragment` | Defines behavior for URL fragments. May be `:show` to show URL fragments, or `nil` to remove fragments. Option must be nested inside the `quantize` option. | `nil` |
-| `request_queuing` | Track HTTP request time spent in the queue of the frontend server. See [HTTP request queuing](#http-request-queuing) for setup details. Set to `true` to enable. | `false` |
+| `request_queuing` | Track HTTP request time spent in the queue of the frontend server. See [HTTP request queuing](#http-request-queuing) for setup details. | `false` |
 | `web_service_name` | Service name for frontend server request queuing spans. (e.g. `'nginx'`) | `'web-server'` |
 
 Deprecation notice:
@@ -1639,7 +1639,7 @@ end
 | Key | Description | Default |
 | --- | ----------- | ------- |
 | `distributed_tracing` | Enables [distributed tracing](#distributed-tracing) so that this service trace is connected with a trace of another service if tracing headers are received | `true` |
-| `request_queuing` | Track HTTP request time spent in the queue of the frontend server. See [HTTP request queuing](#http-request-queuing) for setup details. Set to `true` to enable. | `false` |
+| `request_queuing` | Track HTTP request time spent in the queue of the frontend server. See [HTTP request queuing](#http-request-queuing) for setup details. | `false` |
 | `exception_controller` | Class or Module which identifies a custom exception controller class. Tracer provides improved error behavior when it can identify custom exception controllers. By default, without this option, it 'guesses' what a custom exception controller looks like. Providing this option aids this identification. | `nil` |
 | `middleware` | Add the trace middleware to the Rails application. Set to `false` if you don't want the middleware to load. | `true` |
 | `middleware_names` | Enables any short-circuited middleware requests to display the middleware name as a resource for the trace. | `false` |
@@ -2479,7 +2479,14 @@ server {
 }
 ```
 
-Then you must enable the request queuing feature, by setting `request_queuing: true`, in the integration handling the request. For Rack-based applications, see the [documentation](#rack) for details.
+Then you must enable the request queuing feature. The following options are available for the `:request_queuing` configuration:
+
+| Option             | Description |
+| ------------------ | ----------- |
+| `:include_request` | A `http_server.queue` span will be the root span of a trace, including the total time spent processing the request *in addition* to the time spent waiting for the request to begin being processed. This is the behavior when the configuration is set to `true`. This is the selected configuration when set to `true`. |
+| `:exclude_request` | A `http.proxy.request` span will be the root span of a trace, with the `http.proxy.queue` child span duration representing only the time spent waiting for the request to begin being processed. *This is an experimental feature!* |
+
+For Rack-based applications, see the [documentation](#rack) for details.
 
 ### Processing Pipeline
 
--- a/integration/apps/rails-seven/config/initializers/datadog.rb
+++ b/integration/apps/rails-seven/config/initializers/datadog.rb
@@ -11,7 +11,7 @@ Datadog.configure do |c|
   if Datadog::DemoEnv.feature?('tracing')
     c.tracing.analytics.enabled = true if Datadog::DemoEnv.feature?('analytics')
 
-    c.tracing.instrument :rails, request_queuing: true
+    c.tracing.instrument :rails, request_queuing: :exclude_request
     c.tracing.instrument :redis, service_name: 'acme-redis'
     c.tracing.instrument :resque
   end
--- a/lib/datadog/tracing/contrib/rack/middlewares.rb
+++ b/lib/datadog/tracing/contrib/rack/middlewares.rb
@@ -35,22 +35,34 @@ module Datadog
             request_start = Contrib::Rack::QueueTime.get_request_start(env)
             return if request_start.nil?
 
-            frontend_span = Tracing.trace(
-              Ext::SPAN_HTTP_SERVER_QUEUE,
-              span_type: Tracing::Metadata::Ext::HTTP::TYPE_PROXY,
-              start_time: request_start,
-              service: configuration[:web_service_name]
-            )
+            case configuration[:request_queuing]
+            when true, :include_request # DEV: Switch `true` to `:exclude_request` in v2.0
+              queue_span = trace_http_server(Ext::SPAN_HTTP_SERVER_QUEUE, start_time: request_start)
+              # Tag this span as belonging to Rack
+              queue_span.set_tag(Tracing::Metadata::Ext::TAG_COMPONENT, Ext::TAG_COMPONENT)
+              queue_span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_HTTP_SERVER_QUEUE)
+              queue_span.set_tag(Tracing::Metadata::Ext::TAG_KIND, Tracing::Metadata::Ext::SpanKind::TAG_SERVER)
+              queue_span
+            when :exclude_request
+              request_span = trace_http_server(Ext::SPAN_HTTP_PROXY_REQUEST, start_time: request_start)
+              request_span.set_tag(Tracing::Metadata::Ext::TAG_COMPONENT, Ext::TAG_COMPONENT_HTTP_PROXY)
+              request_span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_HTTP_PROXY_REQUEST)
+              request_span.set_tag(Tracing::Metadata::Ext::TAG_KIND, Tracing::Metadata::Ext::SpanKind::TAG_PROXY)
 
-            # Tag this span as belonging to Rack
-            frontend_span.set_tag(Tracing::Metadata::Ext::TAG_COMPONENT, Ext::TAG_COMPONENT)
-            frontend_span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_HTTP_SERVER_QUEUE)
-            frontend_span.set_tag(Tracing::Metadata::Ext::TAG_KIND, Tracing::Metadata::Ext::SpanKind::TAG_SERVER)
+              queue_span = trace_http_server(Ext::SPAN_HTTP_PROXY_QUEUE, start_time: request_start)
 
-            # Set peer service (so its not believed to belong to this app)
-            frontend_span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE, configuration[:web_service_name])
+              # Measure service stats
+              Contrib::Analytics.set_measured(queue_span)
 
-            frontend_span
+              queue_span.set_tag(Tracing::Metadata::Ext::TAG_COMPONENT, Ext::TAG_COMPONENT_HTTP_PROXY)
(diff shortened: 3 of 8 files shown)


delner:
Hey @agrobbin! Thanks for sharing this!

Generally speaking I really like the idea here. The trace it produces seems more realistic, and it addresses a practical issue.

The only a few (but not insurmountable) reservations that I have are:

1. The artificiality of the new `http_server.request` span. What `service` does it belong to? It bridges two actual processes (a proxy server and the Rack process). It seems like the span is basically analogous to "the trace", as it's just a container for all the spans generated within the trace. (Or maybe it's more than that, but that's just my perception).
2. Changing root spans is more or less a breaking change. It will affect the monitors/metrics/dashboards of users who use this queuing feature. Additionally, on the Datadog side (where we tend to treat root spans as special) we will want to double check this won't have side effects that need to be handled elsewhere, and that it fits with the general pattern for how we want instrumentation to work.
3. Following from (2), 2.0 isn't on the docket yet. How do we incorporate such a breaking change, if we were to merge it tomorrow? Feature flags are an option, but if there's a better idea, I'd like to consider that before adding in another configuration option that would eventually be deprecated/removed.

My overall disposition is that I'm inclined to adopt this new behavior; it's just a matter of addressing the logistics stated above. We'll do our part to start a discussion internally, but I'm interested in hearing your thoughts to those points.

agrobbin:
Thanks for the thoughts @delner! I'm excited at the prospect of getting this included! I've responded to each of your questions below.

> The artificiality of the new `http_server.request` span. What service does it belong to? It bridges two actual processes (a proxy server and the Rack process). It seems like the span is basically analogous to "the trace", as it's just a container for all the spans generated within the trace. (Or maybe it's more than that, but that's just my perception).

I definitely had a similar concern originally, but what led me here is that we need *some* root span, and conceptually, the overarching HTTP request is that "root thing". I didn't consider it totally surprising to be within the same service as the existing `http_server.queue`, as it mapped in my head to the same high-level system (even if *technically* the queuing time can be spent at the HTTP server-level and also potentially at the load balancer-level). I'm open to an alternative, though!

> Changing root spans is more or less a breaking change. It will affect the monitors/metrics/dashboards of users who use this queuing feature.
> 
> Following from (2), 2.0 isn't on the docket yet. How do we incorporate such a breaking change, if we were to merge it tomorrow? Feature flags are an option, but if there's a better idea, I'd like to consider that before adding in another configuration option that would eventually be deprecated/removed.

I completely agree this is breaking, and I imagine this is something currently depended on by a bunch of users.

I do lean toward a configuration option, but maybe `request_queuing` switches from a boolean to an enumerated config of `:inclusive` and `:exclusive`), and the current `true` maps to `:inclusive` until 2.0? What are your thoughts? Maintaining backwards compatibility here is not a massive lift, even past 2.0 (where just the meaning of `true` could change, but `inclusive` is still a valid option). Let me know what you're thinking!

> Additionally, on the Datadog side (where we tend to treat root spans as special) we will want to double check this won't have side effects that need to be handled elsewhere, and that it fits with the general pattern for how we want instrumentation to work.

I'll defer to you on this for sure, but if there is anything I can do in terms of experimentation or testing here, let me know and I can set that up within my company's non-production systems.

delner:
I think the model proposed is reasonable. I just need to reconcile with our plans for consistency with other Datadog instrumentation in other Datadog tracers.

> I do lean toward a configuration option, but maybe request_queuing switches from a boolean to an enumerated config of `:inclusive` and `:exclusive`, and the current `true` maps to `:inclusive` until 2.0?

Maybe slightly different names, but I think this is a clever solution. Satisfies my concerns. Probably the way to go!

Give me a little time to try to shake the tree for some internal feedback, but if I don't get back to you quickly, please feel free to check back in and give me a nudge. In the case of non-decision, I think we adopt this under feature flag (non-default behavior). In the interim, if you can update the PR with the above configuration behavior (and tests/documentation to match) then that will help expedite things.

agrobbin:
@delner I think I've got tests passing (excluding a seemingly unrelated failure on Ruby 2.7) with the new configuration option functionality.

I went with `:include_request` and `:exclude_request` as the potential configurations, though I'm not at all attached to those if you prefer something else.  I've made sure that `true` keeps the existing behavior (the same as `:include_request`). The documentation has also been updated. It seems reasonable to me that in the next major release, `:exclude_request` becomes the default, but that's definitely not my call!

I look forward to hearing what you find on your internal-to-DataDog side of things!

integration/apps/rails-seven/config/initializers/datadog.rb
```diff
@@ -11,7 +11,7 @@
   if Datadog::DemoEnv.feature?('tracing')
     c.tracing.analytics.enabled = true if Datadog::DemoEnv.feature?('analytics')
 
-    c.tracing.instrument :rails, request_queuing: true
+    c.tracing.instrument :rails, request_queuing: :exclude_request
```
delner, line 14: I wonder if this should be the default for the Rails integration app? Somewhat indifferent.

Need to think more about what these settings mean to an integration app: its not really meant to test all the different flags, as much as it is meant to test holistic stability of Rails 7 running typical `ddtrace` deployments.

docs/GettingStarted.md
```diff
@@ -2479,7 +2479,7 @@ server {
 }
 ```
 
-Then you must enable the request queuing feature, by setting `request_queuing: true`, in the integration handling the request. For Rack-based applications, see the [documentation](#rack) for details.
+Then you must enable the request queuing feature, by setting `:request_queuing` to either `:include_request` or `:exclude_request`, in the integration handling the request. For Rack-based applications, see the [documentation](#rack) for details.
```
delner, line 2482: My suggestion is that we clarify what options are available and what they mean. Perhaps in this section we could include a table that maps each value to a description of the behavior/span structure generated, just so its clear what "exclude request" means.

lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -35,22 +35,31 @@ def compute_queue_time(env)
...
-              Ext::SPAN_HTTP_SERVER_QUEUE,
-              span_type: Tracing::Metadata::Ext::HTTP::TYPE_PROXY,
-              start_time: request_start,
-              service: configuration[:web_service_name]
-            )
-
-            # Tag this span as belonging to Rack
-            frontend_span.set_tag(Tracing::Metadata::Ext::TAG_COMPONENT, Ext::TAG_COMPONENT)
-            frontend_span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_HTTP_SERVER_QUEUE)
-            frontend_span.set_tag(Tracing::Metadata::Ext::TAG_KIND, Tracing::Metadata::Ext::SpanKind::TAG_SERVER)
```
delner, line 48: It's worth noting that these tags aren't arbitrary display, but are conventional and used to drive certain behaviors within Datadog. I think it's a good idea if we preserve `kind`, `component` and similar tags on the queue span, in case Datadog needs these for some queue-specific feature in the UI.

 - `service` reflects the application name of the process that performs the operation. Because "http server request" and "queue" represent time spent in the proxy (not in Rack), I think they should reflect `configuration[:web_service_name]`. This will help make the statistics in the UI clear, too.
 - `component` is supposed to describe which library a span is associated with. The "rack request" would be `rack`, whereas the "http server request" and "queue" would ideally reflect the library responsible (e.g. `nginx`). If we're unable to determine an obviously "correct" value from the request, I think we could attribute it to `http_server` (or something similar) to those latter two.
 
   Long term, Ruby shouldn't produce these "http server request" and "queue" spans at all; the proxy server should be instrumented and determine this name for itself.
 - `kind` reflects the role of the thing performing the action (e.g. `server` or `client`). All these spans should be tagged as `server` I think.

delner (commented):
Looking really good thus far. Added some suggestions for documentation and tagging these spans.

delner:
So far it doesn't look like we have a convention for HTTP queue spans. The existing convention defines `{protocol}.{client|server}.request`. This to me looks like it's `{protocol}.{kind}.{operation}`. Based on this, my inclination is naming should be:

 - "proxy http server request" --> `http.proxy.request`
 - "proxy http server queue" --> `http.proxy.queue`
 - "application http server request" --> `http.server.request`
 
 Will double check internally to see if there's any objections.

integration/apps/rails-seven/config/initializers/datadog.rb
agrobbin: I'll happily defer to you here @delner, I did a grep for `request_queuing` to make sure I didn't miss anything, but will adjust as you see fit.

agrobbin:
@delner just to be clear, renaming existing `http_server.queue` span to `http.proxy.queue` will likely have similar breaking effects for existing users who depend on this in monitors, metrics, and other places. Are you suggesting we make that change pre-2.0, or are you suggesting that when `request_queuing` is set to `:include_request`, we continue to use `http_server.queue`, and only use `http.proxy.queue` when configured as `:exclude_request`?

delner:
Breaking change renaming span duly noted. If it doesn't complicate things too much, then ideally `:exclude_request` reflects the desired span names, whereas default behavior reflects the existing span names. That way if you enable the new behavior, you just make all the breaking change updates accordingly, if you don't you don't need to make any.

If possible, I would add comments (such as a `// DEV`) that notes that the existing naming/style (`:include_request`) is deprecated and should be removed in 2.0, whenever that comes in the future.

agrobbin:
@delner I just updated this based on the above discussion!

1 last question... would you like there to a user-facing deprecation warning (via `Datadog.logger.warn`) when this is set to `true`, to highlight this for end-users? I included a code comment, but right now don't have any actual log output.

delner:
> would you like there to be a user-facing deprecation warning

Not something that we've really put into practice but seems reasonable. Only case where I see this problematic is when an application is reconfigured more than once... might spam more than once then. Still, worth trying I think.

ivoanjo:
> Not something that we've really put into practice but seems reasonable. Only case where I see this problematic is when an application is reconfigured more than once... might spam more than once then. Still, worth trying I think.

We actually have a helper to solve exactly that issue! Example of it being used in such a manner:

https://github.com/DataDog/dd-trace-rb/blob/eadd5e126221694e0805578a636cf879d144763b/lib/datadog/core/metrics/client.rb#L185-L194

agrobbin:
@delner @ivoanjo I just added a deprecation warning for `true` and `:include_request`, and noted it in `docs/GettingStarted.md`! I did it in the `Patcher`, which I don't think can be called multiple times, so I didn't include the `Utils::OnlyOnce` logic, but if I am missing something, let me know and I can add that.

delner:
@ivoanjo Right, forgot about `OnlyOnce`. Thanks for that.

Yeah, `Patcher` has its own `OnlyOnce` so you shouldn't have to worry about that. I just want to wrap up some discussion on my end, and do a final pass, but otherwise should be good to go.

delner:
@agrobbin I finished my internal discussion; sounds like we're good to go on this.

Only caveat is that we want to mark this as "experimental" for the time being. Internally, we're interested in creating consistency across HTTP instrumentation as much as possible, and we may need to rename these spans in the future to reflect that consistent naming. Might happen, might never happen. But the trace structure itself is something we intend to keep for the foreseeable future, so nothing to worry about there.

Can you make a note in the documentation (under the "request queuing" section) that this is "experimental"? Once that's in place, I'd be very happy to submit my approval on this. :)

agrobbin:
@delner if this is experimental, should the user-facing deprecation be removed? It seems a little confusing to end-users to introduce an experimental feature that simultaneously deprecates the old behavior.

delner:
@agrobbin Yeah, I think that makes sense, given this change in direction. We can always re-add the deprecation later if we know its what we will break in a future major version. Sorry for the confusion!

agrobbin:
No need to apologize! I've just made the updates. Let me know if there is anything else you see that you'd like to change.

lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -35,22 +35,30 @@ def compute_queue_time(env)
...
+              request_span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_HTTP_PROXY_REQUEST)
+              request_span.set_tag(Tracing::Metadata::Ext::TAG_KIND, Tracing::Metadata::Ext::SpanKind::TAG_PROXY)
+
+              queue_span = trace_http_server(Ext::SPAN_HTTP_PROXY_QUEUE, start_time: request_start)
+              queue_span.set_tag(Tracing::Metadata::Ext::TAG_COMPONENT, Ext::TAG_COMPONENT_HTTP_PROXY)
+              queue_span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_HTTP_PROXY_QUEUE)
+              queue_span.set_tag(Tracing::Metadata::Ext::TAG_KIND, Tracing::Metadata::Ext::SpanKind::TAG_PROXY)
+              # finish the `queue` span now to record only the time spent *in queue*,
+              # excluding the time spent processing the request itself
+              queue_span.finish
```
marcotc, line 58: @agrobbin, currently the `http.proxy.queue` span, as implemented in this PR, request users to individually inspect each flamegraph in order to gather the wait time measured.

Would you say that it makes sense to provide a metric for the duration of the `http.proxy.queue`?

If you think that's valuable in your experience, I suggest setting the `http.proxy.queue` to "measured":
```ruby
Contrib::Analytics.set_measured(queue_span)
```

This will generated a statsd metric for its duration.
(as a side note, root spans are "measured" by default)

"Measured" are not free for our backend, but it's important to make useful spans as "measured".

marcotc:
@agrobbin, it all looks great, thank you so much!
I'm just back from an extended leave, so I'm pick up where our team left off here.

I only left one small comment; save that this PR is ready to go.

agrobbin:
@marcotc updated to mark the queue span as measured!

marcotc:
@agrobbin, that looks good, could I trouble you to add a simple test asserting that the span is marked as "measured"? I promise that's the last of it! :)

agrobbin:
@marcotc whoops, sorry about that, I could've sworn I included that. Should be all set now.

marcotc (approved):
Thank you so much, @agrobbin!

TonyCTHsu:
👋 @agrobbin , we are on the verge of release 2.0 ! 

This release we have made this as the default behavior when enabling `request_queueing`.  Would you like to give it a try by upgrading to `2.0.0.rc1` and give us feedback?

```ruby
gem 'datadog', '>= 2.0.0.rc1'
```

agrobbin:
That's great! We'll try and test it before you get to the official v2.0 release.

PR #2593, opened 2023-02-02
Change base to keyword argument

TonyCTHsu:
**What does this PR do?**

Change method signature to keyword arguments with default, a bit more explicit during invokation and easier to be extended with more arguments


ivoanjo (approved):
:+1: LGTM.

I hesitated a bit at first as I was unsure if this module is part of our public API, and I doublechecked it's not (it's not marked with `@public_api`) so I see no reason not to make this nice improvement :)

PR #2598, opened 2023-02-02
update waf rules to 1.5.0 version

GustavoCaso:
**What does this PR do?**

Update appsec waf rules to [1.5.0 version](https://github.com/DataDog/appsec-event-rules/releases/tag/1.5.0) 

- The risky rules have been remove. I remove the file as well and all the mentions of risky in the code base
-  If a customer was using that level, we will automatically switch to `:recommended` and show a warning message


lib/datadog/appsec/assets/waf_rules/recommended.json
```diff
@@ -1,7 +1,7 @@
 {
```
ivoanjo, line 1: :+1: I've checked the sha256 of the new files match the ones in https://github.com/datadog/appsec-event-rules/tree/1.5.0:

0be5309e8c0ea542b3b2aee4b367144573c1d0921a9e78e021a9ba93b3418db2  recommended.json
d302af9ee296b28742e38015077a76e0786fa82537683f4d6b18ad8993d1ab4d  strict.json

lib/datadog/appsec/processor.rb
```diff
@@ -103,7 +103,7 @@ def load_ruleset
 
         begin
           @ruleset = case ruleset_setting
-                     when :recommended, :risky, :strict
+                     when :recommended, :strict
                        JSON.parse(Datadog::AppSec::Assets.waf_rules(ruleset_setting))
                      when String
                        JSON.parse(File.read(ruleset_setting))
```
ivoanjo, line 109: Done in this way, this change seems like it can cause problems when upgrading -- a customer that previously used the `risky` ruleset will start getting an error when they upgrade between dd-trace-rb 1.9.0 and a later version.

Should we perhaps do something along the lines of: when customers pick `risky`, we instead choose the closest one for them (is that `recommended` or `strict`?), and also print a warning (maybe similar [to this](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/core/metrics/client.rb#L188)) saying that `risky` is now a no-op.

This way the upgrade experience is always smooth.

GustavoCaso: That is a great point @ivoanjo 

I'll ask the security folks for best closest alternative to `risky` and print a warning 😄 

Thank for the feedback

GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/2598/commits/b3465c9bc477f5704e84fad76abac5effebee183

lib/datadog/appsec/processor.rb
```diff
@@ -103,8 +103,14 @@ def load_ruleset
...
           @ruleset = case ruleset_setting
-                     when :recommended, :risky, :strict
+                     when :recommended, :strict
                        JSON.parse(Datadog::AppSec::Assets.waf_rules(ruleset_setting))
+                     when :risky
+                       JSON.parse(Datadog::AppSec::Assets.waf_rules(:recommended))
+                       Datadog.logger.warn(
+                         'Ignoring user-supplied ruleset `:risky`. That configuration value has been deprecated,'\
+                         'selecting `:recommended`.'
+                       )
```
ivoanjo, line 113: Putting my user hat on, I think this message could be improved.

---

Imagine working on a rails app with a dozen other people. When you boot the app, there's a bunch of output scrolling: puma this, database that.

And then in between is this message:

> WARN -- ddtrace: [ddtrace] Ignoring user-supplied ruleset `:risky`. That configuration value has been deprecated,selecting `:recommended`.

Would you know what to do, other than go read the ddtrace sources? Or maybe grepping by `risky`?

Here's a suggestion:

> The `:risky` Application Security Management ruleset has been deprecated and no longer available. The `:recommended` ruleset will be used instead. Please remove the `appsec.ruleset = :risky` setting from your `Datadog.configure` block.

I think this message improves on the current one by:
1. Identifying which component this is coming from. Helps googling, talking to support, etc.
2. Documenting the next steps that customers should take to make this warning go away.

ivoanjo (approved):
Left a suggestion, but otherwise it LGTM :+1:

lib/datadog/appsec/processor.rb
lloeki: That is a concern of me too.

@Taiki-San I think the removal of `risky` is a breaking change, which appears to run against the ruleset release SemVer 2.0 contract, WDYT?

lib/datadog/appsec/processor.rb
lloeki: > has been deprecated

I feel this is incorrect, as usually "deprecation" would mean that it went through a grace period where it was still available.

An option would be to actually deprecate `risky` while keeping things at 1.4.2 for at least one ddtrace release. before we remove `risky` and move to `1.5.0` in a subsequent version.

This way there is no risk of breakage and users of `risky` can move towards either `recommended` or `strict` at their own discretion with due warning.

lloeki (commented):
I'm of two minds here, but I think I'd rather have a release with the deprecation warning about `risky` being removed in the next version, then update to 1.5.0 + remove `risky` and have that released.

So essentially:

- open another PR that adds the deprecation warning
- keep this PR up as "do not merge" until a release with the warning has been published

WDYT @ivoanjo @GustavoCaso ?

ivoanjo:
I think this is one case where the final decision is on your side.

My input is that I think these kinds of decisions should be done on a case-by-case basis, and on this specific case I think it's fine to go ahead with this PR. I think it's a plus that this PR gets customers on our current recommended configuration faster, vs having them stay for longer on an outdated version of the ruleset.

GustavoCaso:
@ivoanjo Thanks so much for the feedback on the error message. Now it is much better and actionable from the point of view of a customer 👍 

@lloeki I'm on the same side as Ivo. I think we should be good to ship this PR. Is less work for us to maintain and the fact that customer might be using `risky` and is a list of rules unmaintained might lead to problems in the future.

lloeki (approved):
Sounds fair enough. Approving.

PR #2600, opened 2023-02-07
Fix Ruby VM statistics broken and causing errors on Ruby 3.2

### LINKED ISSUE #2534: Ruby 3.2.0 gauge error for the RubyVM.stat API changes
**Current behavior**

The gem started logging the following error in our app after we upgraded to Ruby 3.2.0:

```ruby
E, [2023-01-11T09:02:59.815928 #1] ERROR -- ddtrace: [ddtrace] (/usr/local/bundle/ruby/3.2.0/gems/ddtrace-1.5.0/lib/datadog/core/metrics/client.rb:137:in `rescue in gauge') Failed to send gauge stat. Cause: ArgumentError ArgumentError Source: /usr/local/bundle/ruby/3.2.0/gems/ddtrace-1.5.0/lib/datadog/core/metrics/client.rb:133:in `gauge'
```

**Expected behaviour**

The error shouldn't occur.

**Steps to reproduce**

Use any version of the `ddtrace` library (even the latest) on `Ruby 3.2.0`. When a metric flush occurs the above error is logged.

The library should handle correctly the new [Ruby 3.2.0 RubyVM.stat changes](https://github.com/ruby/ruby/pull/5433) and rely on the new key `:constant_cache_invalidations` instead of `:global_constant_state` for `Ruby >= 3.2.0`. It might also gauge the newly added key `:constant_cache_misses` when available.

See https://github.com/ruby/ruby/pull/5433, https://github.com/DataDog/dd-trace-rb/blob/v1.5.0/lib/datadog/core/environment/vm_cache.rb#L23-L25, https://github.com/newrelic/newrelic-ruby-agent/pull/1436

**Environment**

* **ddtrace version:** I've tested with `1.5.0` and `1.8.0` but looks like all the versions are affected
* **Ruby version:** 3.2.0
* **Operating system:** Ubuntu Focal

ivoanjo:
**What does this PR do?**:

This PR fixes #2534 by:

a. Making sure we don't try to report non-existing statistics on Ruby 3.2
b. Adding reporting of the new Ruby 3.2 statistics

It also includes a few cleanups to get us in better shape to support Ruby 3.2. 

**Motivation**:

dd-trace-rb should definitely not get in the way of our awesome customers using Ruby 3.2.

**Additional Notes**

The newly-reported metrics will not show up in the Datadog UX yet, I'm preparing a separate PR to the Datadog UX and default Ruby Runtime Metrics dashboard to show them.

**How to test the change?**:

We are not yet running Ruby 3.2 in CI (I'm preparing a follow-up PR for this), so you'll need to run the test suite locally to validate this.

---

Fixes #2534


spec/datadog/core/environment/vm_cache_spec.rb
```diff
@@ -22,21 +22,67 @@
...
+        it { is_expected.to be nil }
+      end
     end
 
     describe '.global_method_state' do
       subject(:global_method_state) { described_class.global_method_state }
 
-      context 'with Ruby < 3', if: RUBY_VERSION < '3.0.0' do
+      context 'on Ruby 2.x' do
+        before { skip('Test only runs on Ruby 2.x') unless RUBY_VERSION.start_with?('2.') }
```
ivoanjo, line 42: Why replace `if: RUBY_VERSION < '3.0.0'` with the explicit skip?

I like the explicit skip when some test is not running, because it shows up in the rspec output that the testcase was skipped.

Otherwise, it's easy to do a change, run RSpec, it says all tests passed and you just broke some other Ruby version which is being skipped. Yes CI would catch it, but I still favor the more explicit style.

lloeki: I like this approach as well.

spec/datadog/tracing/contrib/support/tracer_helpers.rb
```diff
@@ -68,8 +68,8 @@ def clear_traces!
       config.before do
         # DEV `*_any_instance_of` has concurrency issues when running with parallelism (e.g. JRuby).
         # DEV Single object `allow` and `expect` work as intended with parallelism.
-        allow(Datadog::Tracing::Tracer).to receive(:new).and_wrap_original do |method, *args, &block|
-          instance = method.call(*args, &block)
+        allow(Datadog::Tracing::Tracer).to receive(:new).and_wrap_original do |method, **args, &block|
+          instance = method.call(**args, &block)
```
lloeki, line 72: Minor: make it explicit that this is keyword arguments now.

```suggestion
        allow(Datadog::Tracing::Tracer).to receive(:new).and_wrap_original do |method, **kwargs, &block|
          instance = method.call(**kwargs, &block)
```

Also, wondering what happened to `*args`: is is not needed anymore?

My spider-sense is tingling here WRT the argument changes that happened between 2.6, 2.7, 3.0+3.1, and 3.2.

lloeki (approved):
Approved, but left a few open questions that may be resolved later.

spec/datadog/tracing/contrib/support/tracer_helpers.rb
ivoanjo: Note that this is a specific simpler fix for the `Tracer` class only -- if this was something on some outside gem we'd need to sprinkle our friend `ruby2_keywords` instead ;)

What happened here is that on dd-trace-rb 0.x, this used to be the classic `options = {}`:

https://github.com/DataDog/dd-trace-rb/blob/v0.54.2/lib/ddtrace/tracer.rb#L76

...but on 1.0 it was moved to use real keyword arguments:

https://github.com/DataDog/dd-trace-rb/blob/v1.0.0/lib/datadog/tracing/tracer.rb#L50

Up to 3.2, the forwarding we were doing here still worked, but it stopped working in 3.2. And for all the others, this change works because the class gets keyword arguments.

lib/datadog/core/runtime/metrics.rb
```diff
@@ -57,15 +57,27 @@ def flush
 
           try_flush do
             if Core::Environment::VMCache.available?
-              gauge(
+              # Only on Ruby < 3.2
+              gauge_if_not_nil(
```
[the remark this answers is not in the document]
ivoanjo: > WDYT about an explicit version check? (if or case RUBY_VERSION when ...). That would make the code self-explanatory, and materialise the else case.

I chose not to do the explicit version check so that in the future we wouldn't get bitten by the same issue: let's say Ruby 3.5 removes some of the new metrics, and again we would break if we checked the Ruby version and not the underlying data.

It's true that our comments could become inaccurate in that situation, but since our tests validate the expected results, we would have an indication that there's something we need to fix -- while still not breaking released dd-trace-rb versions.

> WDYT about a feature check? something like if Core::Utils::Features::Ruby.vmcache_global_constant_state?

A feature check per metric would mean more code + more tests to replace the current `nil` check. I wouldn't oppose such a change, but I'm not particularly convinced it's worth the extra complexity either.

PR #2603, opened 2023-02-07
add tests for Appsec reactive engine

GustavoCaso:
**What does this PR do?**

Adding test for `appsec/intrumention/*` and `appsec/reactive/*` components.
Also: 
- adds frozen string comment to updated files
- small refactors

**Motivation**

Started learning about how appsec works underneath. Noticed that some components were missing unit tests and I used that oppportunity to learn more about how they work and write some tests

**Additional Notes**

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/appsec/reactive/engine.rb
+++ b/lib/datadog/appsec/reactive/engine.rb
@@ -1,4 +1,5 @@
 # typed: ignore
+# frozen_string_literal: true
 
 require_relative 'address_hash'
 require_relative 'subscriber'
@@ -8,6 +9,8 @@ module Datadog
     module Reactive
       # Reactive Engine
       class Engine
+        attr_reader :subscribers, :data
+
         def initialize
           @data = {}
           @subscribers = AddressHash.new { |h, k| h[k] = [] } # TODO: move to AddressHash initializer
--- a/lib/datadog/appsec/reactive/engine.rb
+++ b/lib/datadog/appsec/reactive/engine.rb
@@ -13,12 +13,11 @@ module Datadog
 
         def initialize
           @data = {}
-          @subscribers = AddressHash.new { |h, k| h[k] = [] } # TODO: move to AddressHash initializer
-          @children = []
+          @subscribers = AddressHash.new
         end
 
         def subscribe(*addresses, &block)
-          @subscribers[addresses.freeze] << Subscriber.new(&block).freeze # TODO: move freeze to Subscriber
+          @subscribers[addresses.freeze] << Subscriber.new(&block)
         end
 
         def publish(address, data)
--- a/lib/datadog/appsec/reactive/engine.rb
+++ b/lib/datadog/appsec/reactive/engine.rb
@@ -9,8 +9,6 @@ module Datadog
     module Reactive
       # Reactive Engine
       class Engine
-        attr_reader :subscribers, :data
-
         def initialize
           @data = {}
           @subscribers = AddressHash.new
@@ -20,12 +18,12 @@ module Datadog
           @subscribers[addresses.freeze] << Subscriber.new(&block)
         end
 
-        def publish(address, data)
+        def publish(address, value)
           # check if someone has address subscribed
           if @subscribers.addresses.include?(address)
 
-            # someone will be interested, set data
-            @data[address] = data
+            # someone will be interested, set value
+            @data[address] = value
 
             # find candidates i.e address groups that contain the just posted address
             @subscribers.with(address).each do |addresses|
@@ -40,6 +38,10 @@ module Datadog
             end
           end
         end
+
+        private
+
+        attr_reader :subscribers, :data
       end
     end
   end
--- a/lib/datadog/appsec/reactive/operation.rb
+++ b/lib/datadog/appsec/reactive/operation.rb
@@ -1,4 +1,5 @@
 # typed: ignore
+# frozen_string_literal: true
 
 require_relative 'engine'
 
--- a/lib/datadog/appsec/reactive/operation.rb
+++ b/lib/datadog/appsec/reactive/operation.rb
@@ -55,6 +55,13 @@ module Datadog
           def active
             Thread.current[:datadog_security_active_operation]
           end
+
+          private
+
+          # For testing only.
+          def reset!
+            Thread.current[:datadog_security_active_operation] = nil
+          end
         end
       end
     end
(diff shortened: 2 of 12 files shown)


lib/datadog/appsec/reactive/operation.rb
```diff
@@ -40,6 +41,16 @@ def finalize
...
 
+        private
+
+        def select_reactive_engine(reactive, parent)
+          return reactive if reactive
+
+          return parent.reactive unless parent.nil?
+
+          Reactive::Engine.new
+        end
```
GustavoCaso, line 52: I believe this block is more readable than the previous one: 
`reactive_engine || (parent.reactive unless parent.nil?) || Reactive::Engine.new`

lib/datadog/appsec/reactive/engine.rb
```diff
@@ -8,14 +9,15 @@ module AppSec
     module Reactive
       # Reactive Engine
       class Engine
+        attr_reader :subscribers, :data
+
         def initialize
           @data = {}
-          @subscribers = AddressHash.new { |h, k| h[k] = [] } # TODO: move to AddressHash initializer
-          @children = []
```
GustavoCaso, line 14: This variable is not us at all 🔥

spec/datadog/appsec/reactive/operation_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+require 'datadog/appsec/spec_helper'
+require 'datadog/appsec/reactive/operation'
+
+RSpec.describe Datadog::AppSec::Reactive::Operation do
+  after(:each) do
+    Thread.current[:datadog_security_active_operation] = nil
+  end
+
+  context 'initialize' do
+    it 'set current datadog_security_active_operation thread variable for the duration of the block' do
```
lloeki, line 13: `sets`

spec/datadog/appsec/reactive/operation_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      active_operation = Thread.current[:datadog_security_active_operation]
+      expect(active_operation).to be_nil
+      described_class.new('test') do |op|
+        expect(Thread.current[:datadog_security_active_operation]).to eq(op)
+      end
+      expect(active_operation).to be_nil
+    end
+
+    it 'set parent attrribute and set datadog_security_active_operation to parent at the end of the block' do
```
lloeki, line 22: `sets`

spec/datadog/appsec/reactive/operation_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+require 'datadog/appsec/reactive/operation'
+
+RSpec.describe Datadog::AppSec::Reactive::Operation do
+  after(:each) do
+    Thread.current[:datadog_security_active_operation] = nil
+  end
+
+  context 'initialize' do
+    it 'set current datadog_security_active_operation thread variable for the duration of the block' do
+      active_operation = Thread.current[:datadog_security_active_operation]
```
lloeki, line 14: The thread storage is an implementation detail, so I'm thinking this should be a method, to be defined in the scope above.

```
def active_operation
  Thread.current[:datadog_security_active_operation]
end
```

This would turn the example into something higher level, like:

```
it 'sets the active operation for the duration of the block' do
  expect(active_operation).to be_nil
    described_class.new('test') do |op|
      expect(active_operation).to eq(op)
    end
  expect(active_operation).to be_nil
end
```

And of course something similar for the one below.

lloeki (changes requested):

spec/datadog/appsec/reactive/operation_spec.rb
```diff
@@ -66,22 +66,22 @@
...
       operation = described_class.new('test')
       expect(operation.reactive).to receive(:publish).with(:a, 'hello world')
       operation.publish(:a, 'hello world')
     end
   end
 
-  context 'finalize' do
-    it 'set datadog_security_active_operation to parent' do
+  describe '#finalize' do
+    it 'set active to parent' do
```
lloeki, line 78: `sets`

spec/datadog/appsec/reactive/operation_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+  end
+
+  context 'initialize' do
+    it 'set current datadog_security_active_operation thread variable for the duration of the block' do
+      active_operation = Thread.current[:datadog_security_active_operation]
+      expect(active_operation).to be_nil
+      described_class.new('test') do |op|
+        expect(Thread.current[:datadog_security_active_operation]).to eq(op)
+      end
+      expect(active_operation).to be_nil
```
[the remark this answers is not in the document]
lloeki: This still does not test properly as lines 15 and 19 are equivalent:

```
foo = nil
bar = foo
bar.nil? # true
foo = 42
bar.nil? # still true
```

I think this should be `expect(described_class.active).to be_nil`.

spec/datadog/appsec/instrumentation/gateway_spec.rb
```diff
@@ -55,7 +56,7 @@
       expect(opts).to be nil
     end
 
-    it 'wrap provided middlewares on top provided block' do
+    it 'wraps of provided middlewares on top of provided block' do
```
lloeki, line 59: There's an extra `of`: `wraps of provided middlewares on top of provided block`

spec/datadog/appsec/reactive/operation_spec.rb
```diff
@@ -9,22 +9,22 @@
     Thread.current[:datadog_security_active_operation] = nil
```
lloeki, line 9: I would suggest hiding the implementation detail behind a private class-level `reset!` method. This way the implementation detail is self-contained in the class's implementation and the spec is oblivious to it.

Here's an example: https://github.com/DataDog/dd-trace-rb/blob/20616adb41018c0ab947ce3327845652f5db9da5/lib/datadog/appsec/configuration/settings.rb#L204-L207

lloeki (changes requested):

lloeki (approved):
LGTM, with a few comments

spec/datadog/appsec/reactive/operation_spec.rb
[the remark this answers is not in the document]
GustavoCaso: You are right @lloeki I missed this comment 😢 

I changed to `expect(described_class.active).to be_nil`

PR #2605, opened 2023-02-08
[APPSEC-7801] Appsec block request base on response addresses matches

GustavoCaso:
**What does this PR do?**

When calling the waf with the response result if a rule matches and
the action is to block we record the appsec event but, the customer
will see a 403 on the result.

**Additional Notes**

I rename some variables to better reflect what the stored information is. 

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/appsec/contrib/rack/request_middleware.rb
+++ b/lib/datadog/appsec/contrib/rack/request_middleware.rb
@@ -48,7 +48,7 @@ module Datadog
               @waf_context = context
             end
 
-            _response_return, _response_response = Instrumentation.gateway.push('rack.response', response)
+            _response_return, response_response = Instrumentation.gateway.push('rack.response', response)
 
             context.events.each do |e|
               e[:response] ||= response
@@ -56,6 +56,10 @@ module Datadog
             end
 
             AppSec::Event.record(*context.events)
+
+            if response_response && response_response.any? { |action, _event| action == :block }
+              request_return = AppSec::Response.negotiate(env).to_rack
+            end
 
             request_return
           ensure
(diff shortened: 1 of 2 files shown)


lib/datadog/appsec/contrib/rack/request_middleware.rb
```diff
@@ -57,6 +57,10 @@ def call(env)
 
             AppSec::Event.record(*context.events)
 
+            if response_action && response_action.any? { |action, _event| action == :block }
```
TonyCTHsu, line 60: Looks good to me, I think this is the behavior change

TonyCTHsu (approved):

lib/datadog/appsec/contrib/rack/request_middleware.rb
GustavoCaso: Yes that is 😄

lib/datadog/appsec/contrib/rack/request_body_middleware.rb
```diff
@@ -24,11 +24,11 @@ def call(env)
 
             request = ::Rack::Request.new(env)
 
-            request_return, request_response = Instrumentation.gateway.push('rack.request.body', request) do
+            request_return, request_action = Instrumentation.gateway.push('rack.request.body', request) do
```
lloeki, line 27: I disagree with this change, the gateway `response` does contain an action but is a richer and different concept. In the future this return value should be an `Instrumentation::Gateway::Response` Struct (or struct-like object) instead of a basic array of `[action, event]`.

In addition this refactoring makes this file inconsistent with other places that use the instrumentation gateway.

lloeki (approved):
Approved, but I think we should remove e03f43603557477b4f1f65b19790eaa1dbe7b772

PR #772, opened 2019-06-26
[documentation] Ruby documentation audit

apigirl:
This is my audit to make sure that everything that exists currently in https://docs.datadoghq.com/tracing/setup/ruby/ is added here. Also did some small grammar and style updates to make sure this doc reflects well into the documentation site. 

Staging link: https://docs-staging.datadoghq.com/gus/ruby-doc/tracing/setup/ruby/, created in [this PR](https://github.com/DataDog/documentation/pull/4882).

The staging link reflects what is currently in this file and will update once this PR is merged in.

PR #2606, opened 2023-02-08
Create init image for lib-injection

TonyCTHsu:
TBD


lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,19 @@
+if ENV['DDTRACE_AUTOINJECT'] && !ENV['skip_autoinject']
+  if system 'skip_autoinject=true bundle show ddtrace'
+    puts 'ddtrace already installed...'
+  else
+    puts 'ddtrace is not installed... Perform auto-injection...'
+
+    if system 'skip_autoinject=true bundle add ddtrace'
+      puts 'ddtrace added to bundle...'
+    else
+      puts 'Something went wrong when adding ddtrace to bundle...'
```
marcotc, line 10: One common reason for failure of `bundle add` will be when the application uses Bundler in "deployment" mode: https://bundler.io/guides/deploying.html

"deployment" mode means the application maintainer wants to make sure the gems their application needs in order run are not tampered with.

We'll get an error like this in such scenarios when we run `bundle add`:
```
You are trying to install in deployment mode after changing your Gemfile.
Run `bundle install` elsewhere and add the updated Gemfile.lock to version control.

If this is a development machine, remove the /Gemfile freeze by running
`bundle config unset deployment`.
```

It's possible to circumvent this, as exemplified in the error message itself (e.g. `bundle config unset deployment`), but should we?

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,19 @@
+if ENV['DDTRACE_AUTOINJECT'] && !ENV['skip_autoinject']
+  if system 'skip_autoinject=true bundle show ddtrace'
+    puts 'ddtrace already installed...'
+  else
+    puts 'ddtrace is not installed... Perform auto-injection...'
+
+    if system 'skip_autoinject=true bundle add ddtrace'
+      puts 'ddtrace added to bundle...'
```
marcotc, line 8: When `bundle add` fails, it can leave the Gemfile modified but the Gemfile.lock untouched. (e.g. `bundle install && bundle install --deployment && bundle add nokogiri ; bundle exec irb` will fail to execute `bundle exec irb`)

This can cause errors when the application tries to run `bundle exec` in order to start.

We should either completely succeed with the auto-injection, or not leave  any changes in the system.

ivoanjo:
Any thoughts on how to test this? There's quite a bit of non-trivial logic in `auto_inject.rb`, so it would be great to have some tests that give us confidence that we don't accidentally break it as we change things.

lib-injection/Dockerfile
```diff
@@ -0,0 +1,23 @@
+# This image provides the files needed to install the dd-trace-rb
+# and auto instrument Ruby applications in containerized environments.
+FROM busybox
+
+ARG UID=10000
```
marcotc, line 5: Why is the user id set to 10000? Any reason for this specific value?

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,56 @@
+if !ENV['skip_autoinject']
+  if system 'skip_autoinject=true bundle show ddtrace'
```
marcotc, line 2: `system` will print it's output to STDOUT and STDERR, meaning the user will see the output of all commands in their terminal:
```
$ ruby -e 'puts system "bundle show pry"'
~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/pry-0.14.2
true
```

I recommend using [`open3`](https://docs.ruby-lang.org/en/2.1.0/Open3.html) to hide or capture STDOUT and STDERR from the commands we run here.
Sometime we might care about them, other times not: it's up to each case.

Here's an example of using it in `ddtrace`: https://github.com/DataDog/dd-trace-rb/blob/2b163dd91c8d349ceb92370c16f4a7be4a691c35/lib/datadog/ci/ext/environment.rb#L501-L511

There are many combinations of methods in `open3`, I recommend using the simplest one for you use case.

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,56 @@
...
+    puts 'ddtrace already installed... skipping auto-injection'
+  else
+    version = "<DD_TRACE_VERSION_TO_BE_REPLACED>"
+    sha = "<DD_TRACE_SHA_TO_BE_REPLACED>"
+
+    condition =
+      if !version.empty?
+        # For public release
+        "--version '#{version.gsub(/^v/, '')}'"
```
marcotc, line 11: It's good to [escape strings before shell execution](https://ruby-doc.org/stdlib-2.5.1/libdoc/shellwords/rdoc/Shellwords.html#method-c-shellescape) to avoid [Command Injection](https://owasp.org/www-community/attacks/Command_Injection) attacks:

```suggestion
        "--version '#{version.gsub(/^v/, '').shellescape}'"
```

marcotc: The same should be done for all strings interpolated into shell commands.

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,56 @@
...
+    lockfile = Bundler::SharedHelpers.default_lockfile
+
+    datadog_gemfile = gemfile.dirname + ("datadog-#{gemfile.basename}")
+    datadog_lockfile = lockfile.dirname + ("datadog-#{lockfile.basename}")
+
+    # Copies for trial
+    FileUtils.cp gemfile, datadog_gemfile
+    FileUtils.cp lockfile, datadog_lockfile
+
+    if system "skip_autoinject=true BUNDLE_GEMFILE=#{datadog_gemfile} bundle add ddtrace #{condition} --require ddtrace/auto_instrument"
```
marcotc, line 34: I recommend checking for usual cases where this command will fail, like when the bundle is in `deployment` mode or `frozen` (using the return value from `Bundler.frozen_bundle?`):
https://github.com/rubygems/rubygems/blob/451ac56078a3deaacef640906070d3d8f7502570/bundler/lib/bundler.rb#L220-L224

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,56 @@
...
+      # Trial success, replace the original
+      FileUtils.cp datadog_gemfile, gemfile
+      FileUtils.cp datadog_lockfile, lockfile
+    else
+      puts 'Something went wrong when adding ddtrace to bundle...'
+    end
+
+    # Remove the copies
+    FileUtils.rm datadog_gemfile
+    FileUtils.rm datadog_lockfile
```
marcotc, line 46: Make sure to remove these, even if an exception is raised. I recommend using an `ensure` block.

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,56 @@
...
+      else
+        puts "NO VERSION"
+      end
+
+    puts "ddtrace is not installed... Perform auto-injection... for dd-trace-rb"
+
+    require 'bundler'
+    require 'fileutils'
+
+    gemfile  = Bundler::SharedHelpers.default_gemfile
```
marcotc, line 24: It's possible that there's no gemfile present.

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,56 @@
...
+    FileUtils.rm datadog_lockfile
+  end
+
+  begin
+    require 'ddtrace'
+  rescue LoadError => e
+    puts e
+  else
+    puts 'ddtrace loaded...'
+  end
```
marcotc, line 55: If we have just installed `ddtrace` with `bundle add`, it is not possible that `ddtrace` is available in this process. I recommend skipping this for the case where we install ddtrace.

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,56 @@
...
+        # For public release
+        "--version '#{version.gsub(/^v/, '')}'"
+      elsif !sha.empty?
+        # For internal testing
+        "--github 'datadog/dd-trace-rb' --ref '#{sha}'"
+      else
+        puts "NO VERSION"
+      end
+
+    puts "ddtrace is not installed... Perform auto-injection... for dd-trace-rb"
```
GustavoCaso, line 19:
```suggestion
    puts "[AUTO INJECT] ddtrace is not installed. Auto-injection dd-trace-rb"
```

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+  rescue => e
+    puts "[DATADOG LIB INJECTION] #{e}"
+  ensure
+    # Remove the copies
+    FileUtils.rm datadog_gemfile
+    FileUtils.rm datadog_lockfile
+  end
+rescue LoadError, Bundler::BundlerError => e
+  puts "[DATADOG LIB INJECTION] #{e}"
```
marcotc, line 79: I suggest differentiating between these 3 lines `puts "[DATADOG LIB INJECTION]`, so we know which exact 	`puts` was triggered, something like `[DATADOG LIB INJECTION] bundle add error: #{e}`.

lib-injection/Dockerfile
marcotc: Found it: http://www.linfo.org/uid.html

marcotc (approved):
Looks solid! Let's get this out and hear the feedback!

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,84 @@
...
+  if Bundler.frozen_bundle?
+    puts "[DATADOG LIB INJECTION] Cannot inject with frozen Bundler"
+    return
+  end
+
+  version = "<DD_TRACE_VERSION_TO_BE_REPLACED>"
+  sha = "<DD_TRACE_SHA_TO_BE_REPLACED>"
+
+  # Stronger restrict
+  condition = if !version.empty?
```
GustavoCaso, line 25: I don't think `condition` express well what is storing. Maybe we could rename it to something, like `ddtrace_installation_path` 

```ruby
ddtrace_installation_path = if !version.empty?
  # For public release
  "bundle add ddtrace --version #{version.gsub(/^v/, '').shellescape} --require ddtrace/auto_instrument"
elsif !sha.empty?
  # For internal testing
  "bundle add ddtrace--github datadog/dd-trace-rb --ref #{sha.shellescape} --require ddtrace/auto_instrument"
end
```

Then on line 58, we could replace it with:

```ruby
output, status = Open3.capture2e(
  { 'skip_autoinject' => 'true', 'BUNDLE_GEMFILE' => datadog_gemfile.to_s },
  ddtrace_installation_path
)
```

What do you think?

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,84 @@
...
+      "bundle add ddtrace #{condition} --require ddtrace/auto_instrument"
+    )
+
+    if status.success?
+      puts '[DATADOG LIB INJECTION] ddtrace added to bundle...'
+
+      FileUtils.cp datadog_gemfile, gemfile
+      FileUtils.cp datadog_lockfile, lockfile
+    else
+      puts "[DATADOG LIB INJECTION] #{output}"
```
GustavoCaso, line 69: This condition happens when something wrong happens, should we mention something as part of the log line? `[DATADOG LIB INJECTION][ERROR]`?

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,84 @@
...
+    end
+  rescue => e
+    puts "[DATADOG LIB INJECTION] trial error: #{e}"
```
GustavoCaso, line 72: Trial error? What does it mean?

lib-injection/auto_inject.rb
GustavoCaso: Or [DATADOG LIB INJECTION] Error: #{output}

lib-injection/auto_inject.rb
TonyCTHsu: The ruby script here is conducting a trial for adding ddtrace into the bundle. The trial could be success or not and then  cleanup the filesystem. I added this to be safe if there is an exception thrown but I don't see it coming.

GustavoCaso (approved):
Great work

.github/workflows/lib-injection.yml
```diff
@@ -0,0 +1,59 @@
...
+          build-args: DDTRACE_RUBY_SHA=${{ github.sha }}
+          context: ./lib-injection
+
+  test:
+    needs:
+      - build-and-publish-test-image
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
```
ivoanjo, line 33: I'm curious about this -- does the current `test` action still require write permissions? Is this perhaps a left-over from an example/another action we started from?

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,84 @@
+return if ENV['skip_autoinject']
```
ivoanjo, line 1: Usually environment variables are set in `ALL_CAPS`, so I'm curious why deviate from this standard?

Also, since this is a ddtrace-wide configuration, perhaps use `DD_TRACE_` the prefix as well?

Finally, consider actually testing the value. Doing only `if ENV['skip_autoinject']` can be confusing because if I set it to `skip_autoinject=no` or `=false`, the auto-inject is still skipped.

All together, my suggestion is to instead use:

```ruby
return if ENV['DD_TRACE_SKIP_AUTOINJECT'] == 'true'
```

(or `['true', 'y', 'yes', '1'].include?(...)` if you'd like to support a few other variants)

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,84 @@
...
+
+begin
+  require 'open3'
+
+  _, status = Open3.capture2e({'skip_autoinject' => 'true'}, 'bundle show ddtrace')
+
+  if status.success?
+    puts '[DATADOG LIB INJECTION] ddtrace already installed... skipping auto-injection'
```
ivoanjo, line 9: In this file, there's a lot of `puts` for debugging. Should there be a way of disabling them? E.g. by setting `DD_TRACE_DEBUG` perhaps?

Also, consider writing to standard error, e.g. `STDERR.puts`, rather than to standard out, so as to not affect applications that are run with pipes or where the output is important.

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,84 @@
...
+    return
+  end
+
+  require 'bundler'
+  require 'shellwords'
+
+  if Bundler.frozen_bundle?
+    puts "[DATADOG LIB INJECTION] Cannot inject with frozen Bundler"
```
ivoanjo, line 17: As documented in https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#custom-logging , we usually use `[ddtrace]` (or `[DDTRACE]` in a few cases), when logging this kind of information.

Unless `[DATADOG LIB INJECTION]` is a kind of "standard" of all lib injection scripts, I would suggest still using `[ddtrace]` here, so it's easy for customers to locate all ddtrace-related logs (e.g. when support asks for them), as otherwise they may miss these.

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,84 @@
...
+    return
+  end
+
+  version = "<DD_TRACE_VERSION_TO_BE_REPLACED>"
+  sha = "<DD_TRACE_SHA_TO_BE_REPLACED>"
```
ivoanjo, line 22: I suggest leaving here a comment explaining that these are templated as part of the docker image creation process.

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,84 @@
...
+      "bundle add ddtrace --require ddtrace/auto_instrument --version #{version.gsub(/^v/, '').shellescape}"
+    elsif !sha.empty?
+      # For internal testing
+      "bundle add ddtrace --require ddtrace/auto_instrument --github datadog/dd-trace-rb --ref #{sha.shellescape}"
+    end
+
+  unless bundle_add_ddtrace_cmd
+    puts "[DATADOG LIB INJECTION] NO VERSION"
+    return
+  end
```
ivoanjo, line 36: I see a number of error messages in this file. They seem like messages that we will show to customers, so I would suggest approaching them from the POV of a customer that knows nothing about dd-trace-rb.

For instance, compare getting

> [DATADOG LIB INJECTION] NO VERSION

and this example ([randomly picked from](https://github.com/DataDog/dd-trace-rb/pull/2646))

> W, [2023-02-24T10:33:59.671680 #772]  WARN -- ddtrace: [ddtrace] Profiling was requested but is not
supported, profiling disabled: Your ddtrace installation is missing support for the Continuous
Profiler because the `pkg-config` system tool is missing. This issue can usually be fixed by installing
one of the following: the `pkg-config` package on Homebrew and Debian/Ubuntu-based Linux; the `pkgconf`
package on Arch and Alpine-based Linux; the `pkgconf-pkg-config` package on Fedora/Red Hat-based
Linux. (Tip: When fixing this, ensure `pkg-config` is installed **before** running `bundle install`,
and remember to clear any installed gem caches). For help solving this issue, please contact Datadog
support at <https://docs.datadoghq.com/help/>. You can also check out the Continuous Profiler
troubleshooting page at <https://dtdg.co/ruby-profiler-troubleshooting>.

E.g. when writing this message, I think we should think about the customer that will see it. As an example, on this one in particular, I think this issue would probably be a packaging problem, that maybe the customer can't do anything about. So perhaps:

> [ddtrace] Datadog lib-injection failed: missing version configuration. We suggest trying to upgrade to the latest lib-injection version. For help solving this issue, please contact Datadog support at <https://docs.datadoghq.com/help/>.

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  puts "[DATADOG LIB INJECTION] ddtrace is not installed... Perform lib injection for dd-trace-rb."
```
ivoanjo, line 38: Again, since this is a user-visible message, consider making it a bit nicer; perhaps tell customers what's going on (and why), and even include helpful debug information, like the version.

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,84 @@
...
+  gemfile   = Bundler::SharedHelpers.default_gemfile
+  lockfile  = Bundler::SharedHelpers.default_lockfile
+
+  if gemfile.basename.to_s == 'gems.rb'
+    datadog_gemfile = gemfile.dirname + "datadog-Gemfile"
+    datadog_lockfile = lockfile.dirname + "datadog-Gemfile.lock"
+  else
+    datadog_gemfile = gemfile.dirname + ("datadog-#{gemfile.basename}")
+    datadog_lockfile = lockfile.dirname + ("datadog-#{lockfile.basename}")
+  end
```
ivoanjo, line 49: Minor: Consider using the same version of the datadog gemfile for both cases -- after all, we're going to tell bundler to use a specific file, so I don't think there's any reason to name it `datadog-Gemfile` or `datadog-gems.rb` depending on the case.

Consider always calling it `datadog-gems.rb` (or `datadog-Gemfile`).

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  require 'fileutils'
+
+  begin
+    # Copies for trial
```
ivoanjo, line 54: Consider explaining here in a bit more detail why we're doing this

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,84 @@
...
+    FileUtils.cp gemfile, datadog_gemfile
+    FileUtils.cp lockfile, datadog_lockfile
+
+    output, status = Open3.capture2e(
+      { 'skip_autoinject' => 'true', 'BUNDLE_GEMFILE' => datadog_gemfile.to_s },
```
ivoanjo, line 59: Minor: When would `datadog_gemfile` not be a string?

lib-injection/auto_inject.rb
ivoanjo: I think customers may have the same question, and I suggest that the answer is part of the error message, not just a comment here :)

lib-injection/auto_inject.rb
```diff
@@ -0,0 +1,84 @@
...
+    FileUtils.rm datadog_gemfile
+    FileUtils.rm datadog_lockfile
+  end
+rescue Bundler::BundlerError => e
+  puts "[DATADOG LIB INJECTION] BundlerError: #{e}"
+rescue LoadError => e
+  puts "[DATADOG LIB INJECTION] LoadError: #{e}"
+rescue Exception => e
+  puts "[DATADOG LIB INJECTION] Exception: #{e}"
+end
```
ivoanjo, line 84: I had left an earlier comment on this PR, that perhaps you missed but -- I think we're missing some specs for this. We do have some coverage via the system tests, but consider testing some or all of the behaviors here using RSpec as well, to give us more confidence when changing this script.

ivoanjo (commented):
Left a few notes. I wouldn't say my feedback is blocking, but please do consider looking into it before we release this to customers. 

Especially the current log messages I think need a bit of love.

PR #2607, opened 2023-02-08
[PROF-6558] Show profiler overhead in flamegraph for CPU Profiling 2.0

ivoanjo:
**What does this PR do?**:

This PR changes the cpu-time/wall-time sampling in the CPU Profiling 2.0 profiler to expose the profiler overhead as part of the captured data.

Specifically, during sampling, the profiler now measures how much cpu-time/wall-time is spent by the profiler itself, and adds that to the profiler as a separate stack (that shows up as belonging to dd-trace-rb).

**Motivation**:

In the old Ruby profiler, the profiler always worked from a background thread. Because we also sampled that background thread, the flamegraph included the overhead of the profiler -- it was the overhead of the background thread.

With CPU Profiling 2.0, sampling is done in whatever thread is holding the Global VM Lock, which means that this part of the profiler overhead was not being exposed. With this change, this impact on code is now accounted for.

**Additional Notes**:

I was torn between a few options, and decided to use the stack of the `CpuAndWallTimeWorker` thread to represent the overhead of the profiler.

Thus, when this stack shows up in another thread, it represents the overhead of the profiler on that thread.

I could have gone with a placeholder (e.g. just a frame that says "Profiler Sampling" or something similar), but such placeholders break the "Only My Code" feature.

Using a stack from a different thread than the one we're supposedly "sampling" adds a bit of complexity; but we can always undo this and go with the placeholder in this becomes problematic.

**How to test the change?**:


marcotc:
> decided to use the stack of the CpuAndWallTimeWorker thread to represent the overhead of the profiler.

This seems ok. I agree that there's no straightforward lightweight solution here.
Any precise solution I can think of involves adding an extra dimension to all samples.
I think the chosen solution is a good compromise.

marcotc (approved):

ivoanjo:
Thanks Marco for the review! :)

Merging away!

PR #2610, opened 2023-02-09
Minor: Don't actually start profiling in core specs

ivoanjo:
**What does this PR do?**:

This PR tweaks two specs in the `spec/datadog/core` folder (for the environment logger and the telemetry collector) so that they don't actually start a real profiler during the test.

**Motivation**:

The changed tests are validating that the environment logger and the telemetry collector report if profiling was enabled or not, which is why they enable profiling.

But we don't need to actually profile during these specs, and actually doing so slows them down and can generate logging noise (profiler starting, trying to report, etc.).

**Additional Notes**:

In the future, when we separate out the components, we'll probably have a better way of doing this than `allow_any_instance_of`, but for now I couldn't come up with a simpler solution.

**How to test the change?**:

Validate that specs still pass.

### DIFF
--- a/spec/datadog/core/diagnostics/environment_logger_spec.rb
+++ b/spec/datadog/core/diagnostics/environment_logger_spec.rb
@@ -316,7 +316,10 @@ RSpec.describe Datadog::Core::Diagnostics::EnvironmentLogger do
       end
 
       context 'with profiling enabled' do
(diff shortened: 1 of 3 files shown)


ivoanjo:
Ooops, just noticed this makes JRuby unhappy; will push an extra tweak :)

marcotc (approved):

PR #2612, opened 2023-02-10
[APPSEC-8112] Appsec allow to set user id denylist

GustavoCaso:
**What does this PR do?**
Introduce a new appsec configuration option `user_id_denylist`

That list would be use to populate the waf rules data at the initialization of a the `AppSec::Processor`. 

Changes in `AppSec::Processor`:
Previously we had a public method called `update_ip_denylist` which populated the waf rule data just for the IP denylist. Now that we have multiple configuration points for deny lists we have to change the code to account for that. 

I created a new private method `update_rules_data_with_static_configured_values` that creates all the necessary waf rule data for the different deny list configurations and call the waf handle `update_rule_data` function once with all the data. I made that change to reduce the number of calls to the handle. 

Also, I remove the public function `update_ip_denylist` from the processor interface. I thought about leaving it public and adding an extra public function `update_user_id_denylist`, but it seems that those are not needed. Ultimately in the future once Remote Configure is implemented we would have to compute the rules data separate and simply call `processor.update_rule_data(data)`

### DIFF
--- a/lib/datadog/appsec/configuration.rb
+++ b/lib/datadog/appsec/configuration.rb
@@ -40,6 +40,10 @@ module Datadog
           options[:ip_denylist] = value
         end
 
+        def user_id_denylist=(value)
+          options[:user_id_denylist] = value
+        end
+
         # in microseconds
         def waf_timeout=(value)
           options[:waf_timeout] = value
--- a/lib/datadog/appsec/configuration.rb
+++ b/lib/datadog/appsec/configuration.rb
@@ -37,11 +37,11 @@ module Datadog
         end
 
         def ip_denylist=(value)
-          options[:ip_denylist] = value
+          options[:ip_denylist] = value || []
         end
 
         def user_id_denylist=(value)
-          options[:user_id_denylist] = value
+          options[:user_id_denylist] = value || []
         end
 
         # in microseconds
--- a/lib/datadog/appsec/configuration.rb
+++ b/lib/datadog/appsec/configuration.rb
@@ -37,11 +37,11 @@ module Datadog
         end
 
         def ip_denylist=(value)
-          options[:ip_denylist] = value || []
+          options[:ip_denylist] = value
         end
 
         def user_id_denylist=(value)
-          options[:user_id_denylist] = value || []
+          options[:user_id_denylist] = value
         end
 
         # in microseconds
--- a/lib/datadog/appsec/extensions.rb
+++ b/lib/datadog/appsec/extensions.rb
@@ -54,6 +54,12 @@ module Datadog
           @settings.merge(dsl)
         end
 
+        def user_id_denylist=(arg)
+          dsl = AppSec::Configuration::DSL.new
+          dsl.user_id_denylist = arg
+          @settings.merge(dsl)
+        end
+
         def waf_timeout=(arg)
           dsl = AppSec::Configuration::DSL.new
           dsl.waf_timeout = arg
@@ -100,6 +106,10 @@ module Datadog
 
         def ip_denylist
           @settings.ip_denylist
+        end
+
+        def user_id_denylist
+          @settings.user_id_denylist
         end
 
         def waf_timeout
--- a/lib/datadog/appsec/processor.rb
+++ b/lib/datadog/appsec/processor.rb
@@ -51,7 +51,7 @@ module Datadog
           return
         end
 
-        update_ip_denylist
+        update_rules_data_with_static_configured_values
       end
 
       def ready?
@@ -70,20 +70,6 @@ module Datadog
         @handle.toggle_rules(map)
       end
 
-      def update_ip_denylist(denylist = Datadog::AppSec.settings.ip_denylist, id: 'blocked_ips')
-        denylist ||= []
-
-        ruledata_setting = [
-          {
-            'id' => id,
-            'type' => 'data_with_expiration',
-            'data' => denylist.map { |ip| { 'value' => ip.to_s, 'expiration' => 2**63 } }
-          }
-        ]
-
-        update_rule_data(ruledata_setting)
-      end
-
       def finalize
         @handle.finalize
       end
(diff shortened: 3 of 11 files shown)


lib/datadog/appsec/processor.rb
```diff
@@ -94,6 +80,30 @@ def finalize
...
+          'type' => 'data_with_expiration',
+          'data' => denylist.map { |ip| { 'value' => ip.to_s, 'expiration' => 2**63 } }
+        }
+      end
+
+      def user_id_denylist_data(denylist, id: 'blocked_users')
+        {
+          'id' => id,
+          'type' => 'data_with_expiration',
+          'data' => denylist.map { |ip| { 'value' => ip.to_s, 'expiration' => 2**63 } }
```
GustavoCaso, line 103: Should we allow to configure the expiration time for user_id? I thought of having an expiration time of `0` since it seems that if some manually configure to exclude an `user_id` from the application it would be indefinitely until other wise, in which case they would have to change the configuration.

Thoughts @lloeki ?

lib/datadog/appsec/extensions.rb
```diff
@@ -54,6 +54,12 @@ def ip_denylist=(arg)
           @settings.merge(dsl)
         end
 
+        def user_id_denylist=(arg)
+          dsl = AppSec::Configuration::DSL.new
+          dsl.user_id_denylist = arg
+          @settings.merge(dsl)
+        end
```
GustavoCaso, line 61: We do not provide any kind of validation on the appsec settings. Is that intended? 

Thoughts @lloeki ?

lib/datadog/appsec/processor.rb
lloeki: Since when configured this way it cannot be changed afterwards at runtime thus the configuration is essentially static, the only useful value of expiration is approximating infinity, otherwise the denylist entry would expire at some point and would not be able to be changed afterwards.

lib/datadog/appsec/extensions.rb
lloeki: We kind of do, but [only for env vars](https://github.com/DataDog/dd-trace-rb/blob/65eabdbccace4e1de67aaf74032d902bd42dacf6/lib/datadog/appsec/configuration/settings.rb#L106-L114) because they need to be parsed from string to whatever value.

Note that converting (for env vars) is a different operation than validating setting types.

This is due for improvement but that will be part of the configuration revamp project (effort here would be wasted).

lib/datadog/appsec/processor.rb
```diff
@@ -94,6 +80,30 @@ def finalize
...
+        }
+      end
+
+      def user_id_denylist_data(denylist, id: 'blocked_users')
+        {
+          'id' => id,
```
lloeki, line 101: With the current implementation, the `id:` kwarg must always be the same value, so it should be folded in (and probably turned into a constant):

```suggestion
      def user_id_denylist_data(denylist)
        {
          'id' =>  'blocked_users',
```

But in truth I'm thinking both this and the note below pointing at becoming a single `denylist_data(denylist, id)` method. WDYT?

lib/datadog/appsec/processor.rb
```diff
@@ -94,6 +80,30 @@ def finalize
 
       private
 
+      def update_rules_data_with_static_configured_values
```
lloeki, line 83: Thinking we might want that to be more general, something like:

```
merge_to_rule_data('blocked_ips' => Datadog::AppSec.settings.ip_denylist, 'blocked_users' => Datadog::AppSec.settings.user_id_denylist)
```

And a companion:

```
def apply_denylist(settings)
  update_rule_data(merge_to_rule_data('blocked_ips' => settings.ip_denylist, 'blocked_users' => settings.user_id_denylist))
end

# usage: apply_denylist(Datadog::AppSec.settings)
```

WDYT?

lib/datadog/appsec/configuration.rb
```diff
@@ -40,6 +40,10 @@ def ip_denylist=(value)
           options[:ip_denylist] = value
         end
 
+        def user_id_denylist=(value)
+          options[:user_id_denylist] = value
```
lloeki, line 44: Wondering if we should rather resolve to a default here:

```suggestion
          options[:user_id_denylist] = value || []
```

lib/datadog/appsec/configuration.rb
```diff
@@ -40,6 +40,10 @@ def ip_denylist=(value)
           options[:ip_denylist] = value
```
lloeki, line 40: Wondering if we should rather resolve to a default here:
```suggestion
          options[:ip_denylist] = value || []
```

lloeki (approved):
Approving because the public facing API looks correct, nonetheless I feel like changes to the private methods are in order.

lib/datadog/appsec/configuration.rb
lloeki: Internally conflicted about allowing `nil` to mean "unset, use default" vs having a consistent type and eschew downstream `nil` checks.

This ties in with the whole configuration redesign, and I feel like the suggestion here is safer for now instead of littering the code with default values and nil checks at every call site.

lib/datadog/appsec/processor.rb
GustavoCaso: I like the suggestion. 

One thing that I do not think is necessary right now is the `merge_to_rule_data` method. If we see that the mreing logic gets complicated or that we use it in other places of the codebase we could consider creating the method and exposing it as part of the Processor API. 

But for now I think this is readable enough:

```ruby
def apply_denylist_data(settings)
  ruledata_setting = []
  ruledata_setting << denylist_data('blocked_ips', settings.ip_denylist)
  ruledata_setting << denylist_data('blocked_users', settings.user_id_denylist)

  update_rule_data(ruledata_setting)
end
```

lloeki: Fair enough

lib/datadog/appsec/configuration.rb
GustavoCaso: @lloeki I actually change the location in which we return the default empty array. Rather than on the setter I returned an empty array at the time of reading the value of `client_ip_denylist` or `user_id_denylist` if nothing has been configured. 

Check https://github.com/DataDog/dd-trace-rb/pull/2612/commits/2c89745ed0b7250b18f3a6013303b4b4b61891ec

PR #2615, opened 2023-02-13
update ruby 3.2 appraisal gemfile lock files

GustavoCaso:
**What does this PR do?**
Update appraisal gemfile looks for ruby 3.2

**Additional Notes**
Should we update all the other ruby version as well in the same PR?


marcotc:
>  Should we update all the other ruby version as well in the same PR?

Not necessarily: any amount of updates are good updates!
I understand it takes quite a bit to time to upgrade all the combinations, so it's ok to piecemeal them.

marcotc (approved):

ivoanjo (approved):
:+1: Thanks for this, and +1 on what Marco said!

PR #2619, opened 2023-02-14
Adds span.kind:client for dalli

zhamza99:
**What does this PR do?**
Adds `span.kind:client` for dalli

**Motivation**
Unify span tags across tracers


GustavoCaso (approved):
🎉

PR #2620, opened 2023-02-14
Add mandatory rpc and grpc tags for grpc integration

rarguelloF:
**What does this PR do?**

Adds the following tags to the `grpc` integration (both to client and server side interceptors):

- `rpc.grpc.full_method` =>  equivalent to `/$package.$service/$method` from the original protobuf definitions (not from Ruby class / method names)
- `rpc.grpc.status_code` => equivalent to the numeric value of the grpc response status code (https://grpc.github.io/grpc/core/md_doc_statuscodes.html)

Other changes are:

- Includes `.proto` files used for `grpc` tests and update them to use actual protoc generated code.
- Adds a script and docs for generating protoc Ruby files locally.
- Adds `Contrib::GRPC::Formatting` module for extracting grpc resource attributes from the full method string that comes in the client interceptor, and from the ruby grpc method object received on the server side one.  

**Motivation**

Unify span tags across tracers.

**Additional Notes**

**How to test the change?**


lib/datadog/tracing/contrib/ext.rb
```diff
@@ -18,6 +18,10 @@ module RPC
           TAG_SYSTEM = 'rpc.system'
           TAG_SERVICE = 'rpc.service'
           TAG_METHOD = 'rpc.method'
+
+          TAG_GRPC_STATUS_CODE = 'rpc.grpc.status_code'
+          TAG_GRPC_PACKAGE = 'rpc.grpc.package'
+          TAG_GRPC_FULL_METHOD = 'rpc.grpc.full_method'
```
zhamza99, line 24: `rpc.grpc.full_method` isn't an agreed upon tag. I think you mean rpc.grpc.path which is the entire method path

zhamza99 (changes requested):
Looks mostly good, but just ensure the tag names are correct

marcotc:
For `lint` failures, you can run `bundle exec rubocop -a` locally to fix most of them.

lib/datadog/tracing/contrib/grpc/formatting.rb
```diff
@@ -0,0 +1,115 @@
+# typed: ignore
```
ivoanjo, line 1: Minor: This comment was used by the Sorbet type checker, which we removed in #2647 (we only rely on Steep now, which does not need this comment).

TL;DR: You can remove this now :)

rarguelloF: Thanks for letting me know! I'll update the PR 👍

lib/datadog/tracing/contrib/ext.rb
rarguelloF: as discussed offline, this tag will be named `rpc.grpc.full_method`

zarirhamza (approved):

ivoanjo:
Since this PR is about a tracing integration, I'm humbly going to remove myself from the list of reviewers, since the folks working on tracing have a lot more context here than me :)

lib/datadog/tracing/contrib/grpc/datadog_interceptor/client.rb
```diff
@@ -17,32 +18,45 @@ module DatadogInterceptor
...
+                end
               end
             end
 
             private
 
-            def annotate!(trace, span, metadata, call)
-              span.set_tags(metadata)
+            def annotate!(trace, span, keywords, formatter)
+              metadata = keywords[:metadata]
```
TonyCTHsu, line 48: I think we can remove `keywords[:metadata] ||= {}` and change this line to 
```
metadata = keywords[:metadata] || {}
```

spec/datadog/tracing/contrib/grpc/support/grpc_helper.rb
```diff
@@ -1,12 +1,21 @@
 require 'grpc'
-
 require 'spec/support/thread_helpers'
 
+if RUBY_VERSION < '2.3'
+  require_relative './gen/grpc-1.19.0/test_service_services_pb'
+else
+  require_relative './gen/test_service_services_pb'
+end
```
TonyCTHsu, line 8: Can this be replaced by requiring [this file?](https://github.com/DataDog/dd-trace-rb/pull/2620/files#diff-f7651db66dc15d9c30b482d82173a8403282611f14e041cb61f9359e8da28e5d)

lib/datadog/tracing/contrib/grpc/formatting.rb
```diff
@@ -0,0 +1,114 @@
...
+              # protobuf definition
+              owner.service_name.to_s
+            end
+
+            # extract_grpc_method attempts to find the original method name from the Protobuf file definition,
+            # since grpc gem forces the implementation method name to be in snake_case.
+            def extract_grpc_method(grpc_method_object)
+              owner = grpc_method_object.owner
+
+              return VALUE_UNKNOWN unless owner.instance_variable_defined?(:@rpc_descs)
```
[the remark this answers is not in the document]
rarguelloF: I added this check because for the case of the previous grpc Ruby class (manual) implementation this was not defined, and this is added by the protoc Ruby generated one.

Even though I just checked the [grpc gem](https://github.com/grpc/grpc/blob/2d4f3c56001cd1e1f85734b2f7c5ce5f2797c38a/src/ruby/lib/grpc/generic/rpc_server.rb#L351-L360) fails to start if you provide a class without this or this having no items.

Would you prefer to remove this check? (in that case I think we probably need to update some unit tests that are relying on a non-autogenerated grpc test service class if I recall correctly)

spec/datadog/tracing/contrib/grpc/support/grpc_helper.rb
rarguelloF: I think so, thanks for catching this up!

lib/datadog/tracing/contrib/grpc/datadog_interceptor/client.rb
```diff
@@ -16,33 +17,45 @@ module DatadogInterceptor
...
             private
 
-            def annotate!(trace, span, metadata, call)
-              span.set_tags(metadata)
+            def annotate!(trace, span, keywords, formatter)
+              metadata = keywords[:metadata] || {}
+              call = keywords[:call]
 
-              span.set_tag(Contrib::Ext::RPC::TAG_SYSTEM, Ext::TAG_SYSTEM)
+              span.set_tags(metadata)
```
marcotc, line 50: Is this setting all gRPC `metadata` key/value pairs as span tags?
Isn't `metadata` arbitrary user data, meaning it could have PII?
If so, we can't simply tag it like this. Is this a requirement from some internal doc?

Also, the keys will become Datadog tag names, we should make sure they don't conflict our tag namespace, likely by namespacing them to `metadata.METADATA_KEY = METADATA_VALUE`.

rarguelloF: Hey @marcotc 👋 !
This code was already here before my changes (I just moved the lines around)
I think the concerns that you raise make sense but I think they would be out of scope of this PR. If you wanna discuss this offline please let me know!

marcotc (approved):

marcotc:
@rarguelloF, can you rebase this PR?

I'll merge it right after.

PR #2623, opened 2023-02-15
flaky:Datadog::Core::Workers::RuntimeMetrics integration test

marcotc:
Fixes this flaky test (to the best of my abilities): https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/8829/workflows/d051131f-b5fe-4967-aa7d-e99b8773996b/jobs/328470/tests

```
Failure/Error: expect(metrics).to have_received(:flush).twice

  (InstanceDouble(Datadog::Core::Runtime::Metrics) (anonymous)).flush(*(any args))
      expected: 2 times with any arguments
      received: 1 time with any arguments
./spec/datadog/core/workers/runtime_metrics_spec.rb:289:in `block in <main>'
./spec/spec_helper.rb:220:in `block in <main>'
./spec/spec_helper.rb:112:in `block in <main>'
/usr/local/bundle/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block in <main>'
```

### DIFF
--- a/spec/datadog/core/workers/runtime_metrics_spec.rb
+++ b/spec/datadog/core/workers/runtime_metrics_spec.rb
@@ -282,11 +282,9 @@ RSpec.describe Datadog::Core::Workers::RuntimeMetrics do
       it 'produces metrics every interval' do
         worker.perform
 
-        sleep(default_flush_interval * 2)
-
         # Metrics are produced once right away
         # and again after an interval.
-        expect(metrics).to have_received(:flush).at_least(2).times
+        wait(default_flush_interval * 2).for(metrics).to have_received(:flush).at_least(2).times
       end
     end
 
--- a/spec/datadog/core/workers/runtime_metrics_spec.rb
+++ b/spec/datadog/core/workers/runtime_metrics_spec.rb
@@ -268,7 +268,7 @@ RSpec.describe Datadog::Core::Workers::RuntimeMetrics do
 
   describe 'integration tests', :integration do
     describe 'interval' do
-      let(:default_flush_interval) { 0.1 }
+      let(:default_flush_interval) { 0.01 }
 
       before do
         stub_const(
@@ -284,7 +284,7 @@ RSpec.describe Datadog::Core::Workers::RuntimeMetrics do
 
         # Metrics are produced once right away
         # and again after an interval.
-        wait(default_flush_interval * 2).for(metrics).to have_received(:flush).at_least(2).times
+        wait_for(metrics).to have_received(:flush).at_least(2).times
       end
     end
 
--- a/spec/datadog/core/workers/runtime_metrics_spec.rb
+++ b/spec/datadog/core/workers/runtime_metrics_spec.rb
@@ -282,11 +282,11 @@ RSpec.describe Datadog::Core::Workers::RuntimeMetrics do
       it 'produces metrics every interval' do
         worker.perform
 
-        sleep(default_flush_interval * 1.1)
+        sleep(default_flush_interval * 2)
 
         # Metrics are produced once right away
         # and again after an interval.
-        expect(metrics).to have_received(:flush).twice
+        expect(metrics).to have_received(:flush).at_least(2).times
       end
     end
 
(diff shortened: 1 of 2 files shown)


spec/datadog/core/workers/runtime_metrics_spec.rb
```diff
@@ -282,11 +282,11 @@
       it 'produces metrics every interval' do
         worker.perform
 
-        sleep(default_flush_interval * 1.1)
+        sleep(default_flush_interval * 2)
```
GustavoCaso, line 285: Could we try not using `sleep`

```ruby
describe 'interval' do
  let(:default_flush_interval) { 0.1 }

  before do
    stub_const(
      'Datadog::Core::Workers::RuntimeMetrics::DEFAULT_FLUSH_INTERVAL',
      default_flush_interval
    )
  end

  after { worker.stop }

  let(:metrics) do
    class StubMetrics
      attr_reader :flush_count

      def initialize
        @flush_count = 0
      end

      def flush
        @flush_count+=1
      end
    end.new
  end

  it 'produces metrics every interval' do
    worker.perform

    begin
      Timeout::timeout(5) {
        loop do
          break if metrics.flush_count >= 2
        end
      }
      expect(true).to eq(true)
    rescue Timeout::Error
      fail 'metrics are not flush every interval'
    end
  end
end
```

ivoanjo (approved):
:+1: Agreed that tests which rely on sleep are not great, but I think we can live with this for now

spec/datadog/core/workers/runtime_metrics_spec.rb
GustavoCaso: I know timeout is a sort of sleep in a way, but I feel that it less magical 😄

ivoanjo: And yeah, timeout is indeed basically a sleep (although they use wait on a conditional variable): https://github.com/ruby/ruby/blob/1a4b4cd7f8b53fc52191c1ddcf3f624a5e687c24/lib/timeout.rb#L113 .

GustavoCaso: If we have a dedicated method for doing this sort of things I'm all in favour of using it. Even if under the hood it uses sleep 😉 

I suggested `Timeout` because as you stated it uses a conditional variable which in my opinion is better use case for this kind of scenarios

GustavoCaso:
@ivoanjo what about: 

```ruby
describe 'interval' do
  let(:default_flush_interval) { 0.1 }

  before do
    stub_const(
      'Datadog::Core::Workers::RuntimeMetrics::DEFAULT_FLUSH_INTERVAL',
      default_flush_interval
    )
  end

  after { worker.stop }

  let(:metrics) do
    class StubMetrics
      attr_reader :flush_count

      def initialize
        @flush_count = 0
      end

      def flush
        @flush_count+=1
      end
    end.new
  end

  it 'produces metrics every interval' do
    worker.perform

    try_wait_until { metrics.flush_count >= 2 }
  end
end
```

ivoanjo:
I would maybe use an `instance_double` for the metrics, and then add a stub on `flush` to increment `@flush_count` instead of using the `StubMetrics` class kinda similarly to how I suggested in https://github.com/DataDog/dd-trace-rb/pull/2624#discussion_r1108229751 but that's just me nitpicking, I think your solution is quite reasonable as well :)

GustavoCaso:
>  would maybe use an instance_double for the metrics, and then add a stub on flush to increment @flush_count instead of using the StubMetrics

Totally fair. I'm in favour of using `instance_double` over a mock class

spec/datadog/core/workers/runtime_metrics_spec.rb
lloeki: I would consider any usage of `Timeout` _very carefully_: https://jvns.ca/blog/2015/11/27/why-rubys-timeout-is-dangerous-and-thread-dot-raise-is-terrifying/

marcotc: I agree that waiting for the actual condition is more correct here. I'll change the condition.

ivoanjo (approved):
:+1: LGTM

PR #2624, opened 2023-02-15
flaky:Datadog::Core::Workers::Polling

marcotc:
Fixes flaky test ["Datadog::Core::Workers::Polling when included into a worker #perform when #enabled? is true is expected to have received perform(*(any args)) 1 time"](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/8907/workflows/0f3a873c-3af7-427c-864e-c80b4add5adc/jobs/331375) with the use of a new gadget:

```
Failure/Error: expect(worker_spy).to have_received(:perform).at_least(:once)

  (Double "worker spy").perform(*(any args))
      expected: at least 1 time with any arguments
      received: 0 times with any arguments
./spec/datadog/core/workers/polling_spec.rb:43:in `block (5 levels) in <top (required)>'
./spec/spec_helper.rb:220:in `block (2 levels) in <top (required)>'
./spec/spec_helper.rb:112:in `block (2 levels) in <top (required)>'
/usr/local/bundle/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
```

## New gadget

The most straightforward wait to write the offending test is: "Wait until the `worker_spy` has received `#perform`. If it takes too long, fail the test.".

We were using a proxy for the wait condition before, `worker.running? && worker.run_loop?`, but that's not a perfect proxy.

The best thing would be to wait for the real assertion itself. The issue is that the assertion condition is private to the RSpec spy `worker_spy`. The only way to inspect it is to trigger the expectation `expect(worker_spy).to have_received(:perform)`, but this exits the test immediately.

It is possible to assert `expect(...)` in a loop, but you have to rescue RSpec exceptions and it gets messy.

Thankfully `@laserlemon` did the hard work here: https://github.com/laserlemon/rspec-wait

This allows us to wait for the assertion itself, regardless of the complexity of the wait condition: `wait_for(worker_spy).to have_received(:perform)`

This is a generic approach that works for any expectation and has a timeout (I configured the default to 5 seconds, see PR).

### DIFF
--- a/spec/datadog/core/workers/polling_spec.rb
+++ b/spec/datadog/core/workers/polling_spec.rb
@@ -22,25 +22,14 @@ RSpec.describe Datadog::Core::Workers::Polling do
       let(:task) { proc { |*args| worker_spy.perform(*args) } }
       let(:worker_spy) { double('worker spy') }
 
-      before { allow(worker_spy).to receive(:perform) { sleep 5 } }
-
-      context 'by default' do
-        before { skip('TODO: This test sometimes hangs forever and other times flakes. Requires further investigation.') }
-
-        it do
-          perform
-          try_wait_until { worker.running? && worker.run_loop? }
-          expect(worker_spy).to have_received(:perform).at_least(:once)
-        end
-      end
+      before { allow(worker_spy).to receive(:perform) }
 
       context 'when #enabled? is true' do
         before { allow(worker).to receive(:enabled?).and_return(true) }
 
         it do
           perform
-          try_wait_until { worker.running? && worker.run_loop? }
-          expect(worker_spy).to have_received(:perform).at_least(:once)
+          wait_for(worker_spy).to have_received(:perform)
         end
       end
 
--- a/Gemfile
+++ b/Gemfile
@@ -49,6 +49,7 @@ gem 'rake-compiler', '~> 1.1', '>= 1.1.1' # To compile native extensions
 gem 'redcarpet', '~> 3.4' if RUBY_PLATFORM != 'java'
 gem 'rspec', '~> 3.12'
 gem 'rspec-collection_matchers', '~> 1.1'
+gem 'rspec-wait', '~> 0'
 if RUBY_VERSION >= '2.3.0'
   gem 'rspec_junit_formatter', '>= 0.5.1'
 else
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -4,6 +4,7 @@ $LOAD_PATH.unshift File.expand_path('..', __dir__)
 $LOAD_PATH.unshift File.expand_path('../lib', __dir__)
 require 'pry'
 require 'rspec/collection_matchers'
+require 'rspec/wait'
 require 'webmock/rspec'
 require 'climate_control'
 
@@ -90,6 +91,7 @@ RSpec.configure do |config|
   config.order = :random
   config.filter_run focus: true
   config.run_all_when_everything_filtered = true
+  config.wait_timeout = 5 # default timeout for `wait_for(...)`, in seconds
 
   if config.files_to_run.one?
     # Use the documentation formatter for detailed output,


spec/datadog/core/workers/polling_spec.rb
```diff
@@ -22,25 +22,14 @@
...
+      before { allow(worker_spy).to receive(:perform) }
 
       context 'when #enabled? is true' do
         before { allow(worker).to receive(:enabled?).and_return(true) }
 
         it do
           perform
-          try_wait_until { worker.running? && worker.run_loop? }
-          expect(worker_spy).to have_received(:perform).at_least(:once)
+          wait_for(worker_spy).to have_received(:perform)
```
ivoanjo, line 32: I like rspec-wait and I've used it in the past, but... it does seem to be a bit heavy to add a new dependency only for this single use-case.

Here's an alternative solution without 

```diff
diff --git a/spec/datadog/core/workers/polling_spec.rb b/spec/datadog/core/workers/polling_spec.rb
index 4b72daf1a..b4c839ac2 100644
--- a/spec/datadog/core/workers/polling_spec.rb
+++ b/spec/datadog/core/workers/polling_spec.rb
@@ -22,14 +22,17 @@ RSpec.describe Datadog::Core::Workers::Polling do
       let(:task) { proc { |*args| worker_spy.perform(*args) } }
       let(:worker_spy) { double('worker spy') }
 
-      before { allow(worker_spy).to receive(:perform) }
+      before do
+        @perform_called = false
+        allow(worker_spy).to receive(:perform) { @perform_called = true }
+      end
 
       context 'when #enabled? is true' do
         before { allow(worker).to receive(:enabled?).and_return(true) }
 
         it do
           perform
-          wait_for(worker_spy).to have_received(:perform)
+          try_wait_until { @perform_called }
         end
       end
```

GustavoCaso:
I feel like adding an extra gem to solve the problem of flaky test is a little overkill. 

Looking at the code that the gems adds is using the ruby `Timeout` class https://github.com/laserlemon/rspec-wait/blob/master/lib/rspec/wait/handler.rb#L9

I think it would be best for use to improve the code of our `try_wait_until` method to better suit our needs

marcotc:
> I feel like adding an extra gem to solve the problem of flaky test is a little overkill.
> 
> Looking at the code that the gems adds is using the ruby `Timeout` class [laserlemon/rspec-wait@`master`/lib/rspec/wait/handler.rb#L9](https://github.com/laserlemon/rspec-wait/blob/master/lib/rspec/wait/handler.rb?rgh-link-date=2023-02-16T09%3A38%3A20Z#L9)
> 
> I think it would be best for use to improve the code of our `try_wait_until` method to better suit our needs, especially since `try_wait_until` does a similar job

`try_wait_until` cannot be easily made to work with RSpec expecations.
The way to do it, which I started implementing, is how `rspec-wait` does it: https://github.com/laserlemon/rspec-wait/blob/8258760da1e94df1f1b979b5cf6ee52caa1ea304/lib/rspec/wait/handler.rb#L13
```ruby
rescue RSpec::Expectations::ExpectationNotMetError => failure
```

But it's already done for us.

I can copy their approach, into our code base if the issue is "one more development dependency gem": we'll basically inline this.

This pattern done in this PR is required over and over again in our test suite btw, this is not a one-off. A lot of our tests have flaky wait conditions today.

ivoanjo (approved):
Fair enough! If you're looking at a few more issues and see that this pattern is coming up more often, I agree with you that manually implementing it is rather sub-optimal vs just using something that actually integrates with rspec.

Here sir, take my :+1:

GustavoCaso:
@marcotc Fairpoint regarding copying someone else code into our codebase. 

Could we use this PR to replace all usage of `try_wait_until` with `wait_for` and remove the custom flaky custom helper `try_wait_until` within this PR?

marcotc:
Thank you for the reviews and patience @ivoanjo & @GustavoCaso.

> Could we use this PR to replace all usage of try_wait_until with wait_for and remove the custom flaky custom helper try_wait_until within this PR?

We can't remove `try_wait_until` because it works for simple conditions, and is used across a lot of the test base.
It's actually more performance than `rspec-wait`, but doesn't do the RSpec error handling for you.

Overall, I'm not 100% super happy with it (it's retry timing is not perfect, for example), but it's pretty much as idiomatic as it can be when used to assert RSpec conditions.

I have a chain of PRs with flaky tests and we can discuss further how well `rspec-wait` is applied to other test scenarios.
I wouldn't mind having our own "reliably wait for this condition" helper, which we kind of have today. If `rspec-wait` turns out to not be as adaptable as needed, I'll refactor it as we go along this flaky test fixing journey.

PR #2626, opened 2023-02-16
Minor: Remove unused AgentSettings#merge method

ivoanjo:
**What does this PR do?**:

This PR removes a `merge` method that is no longer in use. This is safe to remove because it is not part of ddtrace's public API.

This was only used by the legacy profiling transport (as hinted in the comment attached to the code), and the legacy profiling transport was removed in https://github.com/DataDog/dd-trace-rb/pull/2062 (search for `configure_for_agent` inside `lib/datadog/profiling/transport/http.rb`).

**Motivation**:

Remove unused code :)

**Additional Notes**:

N/A

**How to test the change?**:

Check that CI is still green.

### DIFF
--- a/lib/datadog/core/configuration/agent_settings_resolver.rb
+++ b/lib/datadog/core/configuration/agent_settings_resolver.rb
@@ -1,4 +1,4 @@
-# typed: true
+# typed: false
 
 require 'uri'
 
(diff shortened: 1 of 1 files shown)


lloeki (approved):
All good!

PR #2629, opened 2023-02-16
Bump sorbet version to one compatible with Mac Ventura

marcotc:
When running bundle install in the new Mac Ventura OS, sorbet fails to install:
```
Could not find compatible versions

Because sorbet >= 0.5.9672, < 0.5.9680 depends on sorbet-static = 0.5.9672
and sorbet-static = 0.5.9672 could not be found in rubygems repository
https://rubygems.org/ or installed locally for any resolution platforms
(x86_64-darwin-22),
  sorbet >= 0.5.9672, < 0.5.9680 is forbidden.
So, because Gemfile depends on sorbet = 0.5.9672,
  version solving has failed.

The source contains the following gems matching 'sorbet-static (= 0.5.9672)':
  * sorbet-static-0.5.9672-java
  * sorbet-static-0.5.9672-universal-darwin-14
  * sorbet-static-0.5.9672-universal-darwin-15
  * sorbet-static-0.5.9672-universal-darwin-16
  * sorbet-static-0.5.9672-universal-darwin-17
  * sorbet-static-0.5.9672-universal-darwin-18
  * sorbet-static-0.5.9672-universal-darwin-19
  * sorbet-static-0.5.9672-universal-darwin-20
  * sorbet-static-0.5.9672-universal-darwin-21
  * sorbet-static-0.5.9672-x86_64-linux
```

This happens because there's no `darwin-22` for the version of sorbet we have locked to `0.5.9672`.

This PR bumps up our version of sorbet to the first one that supports `darwin-22`, `0.5.10201`.

This only happens for versions of Ruby compiled after my upgrade to Ventura: older versions of Ruby I already had compiled for `darwin-21` still work locally.

One sorbet failure was raised with the upgrade: I updated the hidden definitions to address the issue.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -117,7 +117,7 @@ group :check do
   # version and bump it from time to time.
   # Also, there's no support for windows
   if RUBY_VERSION >= '2.4.0' && (RUBY_PLATFORM =~ /^x86_64-(darwin|linux)/)
-    gem 'sorbet', '= 0.5.9672'
+    gem 'sorbet', '= 0.5.10201'
     gem 'spoom', '~> 1.1'
   end
 
--- a/sorbet/rbi/hidden-definitions/hidden.rbi
+++ b/sorbet/rbi/hidden-definitions/hidden.rbi
@@ -149,8 +149,6 @@ class Array
 
   def deconstruct(); end
 
-  def quote(); end
-
   def shelljoin(); end
 
   def to_h(); end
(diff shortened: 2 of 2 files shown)


Gemfile
```diff
@@ -117,7 +117,7 @@ group :check do
   # version and bump it from time to time.
   # Also, there's no support for windows
   if RUBY_VERSION >= '2.4.0' && (RUBY_PLATFORM =~ /^x86_64-(darwin|linux)/)
-    gem 'sorbet', '= 0.5.9672'
+    gem 'sorbet', '= 0.5.10201'
```
ivoanjo, line 120: Nothing that needs to be changed in this PR, but I'm curious if there's any reason to pick 0.5.10201 specifically, since it's a version from last August.

ivoanjo (approved):
:+1: LGTM

Gemfile
marcotc: It's the oldest version that has `darwin-22` packages published.
I did a conservative upgrade, just because I didn't know how much risk it would be to fully upgrade it.

marcotc:
Merging this to unblock local development.

Gemfile
ivoanjo: Ack, thanks for satisfying my curiosity ;)

PR #2630, opened 2023-02-17
Add better error information to flaky thread-leak detection test

marcotc:
I can't reproduce the flaky test [Datadog integration graceful shutdown for threads closes tracer threads](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/8871/workflows/78282204-4709-4f46-8a2e-2a864c911a23/jobs/330031/tests).

For now, I'm adding additional debug information: the stack trace of all running threads before and after the test runs.

### DIFF
--- a/spec/datadog/integration_spec.rb
+++ b/spec/datadog/integration_spec.rb
@@ -25,15 +25,12 @@ RSpec.describe 'Datadog integration' do
     end
 
     context 'for threads' do
-      let!(:original_thread_count) { thread_count }
-      let!(:original_inspect_threads) { inspect_threads }
+      let(:original_threads) { Thread.list }
+      let!(:original_threads_inspect) { inspect_threads(original_threads) } # Store result as stack trace will change
+      let(:threads) { Thread.list }
 
-      def thread_count
-        Thread.list.count
-      end
-
-      def inspect_threads
-        Thread.list.map.with_index { |t, idx| "#{idx}=#{t.backtrace}" }.join(';')
+      def inspect_threads(threads)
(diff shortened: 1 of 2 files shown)


spec/datadog/integration_spec.rb
```diff
@@ -26,18 +26,25 @@ def wait_for_tracer_sent
...
       it 'closes tracer threads' do
         start_tracer
         wait_for_tracer_sent
 
         shutdown
 
-        expect(thread_count).to eq(original_thread_count)
+        expect(thread_count).to eq(original_thread_count),
+          "Expected #{original_thread_count} threads (#{original_inspect_threads}), "\
+                  "got #{thread_count} threads (#{inspect_threads})"
```
ivoanjo, line 47: This may be slightly flaky/misleading, since `Thread.list` gets called 4 times in this spec, and there's no guarantee that the results are consistent (which may be hella confusing).

What do you think of introducing `let`s for the `Thread.list` before and after, and then using that to obtain the counts and prints? That would ensure that the counts used in the assertion always matched the information we have for debugging.

marcotc: I didn't think that would be an issue, given the test is not time sensitive but state sensitive.
I'll make the changes.

ivoanjo (approved):
:+1: LGTM

spec/datadog/integration_spec.rb
ivoanjo: I see! My thinking is that the background thread may be some leftover from something else that may still finish at some point, and thus why I think even if the test itself was not sensitive to time, the thread may be.

PR #2633, opened 2023-02-20
Add AppSec basic transport

lloeki:
**What does this PR do?**

Implement HTTP transport through the agent.

**Motivation**

Remote configuration needs to communicate with the agent. This goes through the agent API.

**Additional Notes**

The `ddtrace/transport` code is almost independent of `Tracing`, but is coupled on several key points that makes it hardly usable by another product.

In addition, it appears that the agent API version scope cannot be negotiated independently of endpoints: remote configuration needs `v0.7/config` but tracing implements `v0.3/traces` and `v0.4/traces`, while `v0.7/traces` may or may not be compatible.

Therefore duplication of several files is required, either because they need to be modified or their generic class ends up including a module that implements a generic method override that makes it specialized for `Tracing` (notably `send_payload` without room for it being usable by other non-tracing APIs.

**How to test the change?**

This should call `/info` on the agent, which makes the agent report its configuration and features it supports, whihc is very useful for feature negotiation.

```
ObjectSpace.each_object(Datadog::AppSec::Component).to_a.first.transport.send_info
```

### DIFF
--- a/lib/datadog/appsec/component.rb
+++ b/lib/datadog/appsec/component.rb
@@ -25,14 +25,15 @@ module Datadog
         end
       end
 
-      attr_reader :processor, :transport
+      attr_reader :processor, :transport_root, :transport_v7
 
       def initialize(agent_settings, processor:)
         @processor = processor
         transport_options = {}
         transport_options[:agent_settings] = agent_settings if agent_settings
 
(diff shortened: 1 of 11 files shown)


lloeki:
Superseded by other PRs.

PR #1891, opened 2022-02-08
Remove SpanOperation.allocations

marcotc:
The allocations metric on Span has been around for a long time, sitting unused, but only recently exposed as a metric.

Because we only collect global statistics, the number of allocations reported on each span can be grossly exaggerated. Besides, it has the highest cost of any operation in the `SpanOperation` class, affecting the critical path performance.

### Breaking change

* Removed: span `allocations` metric.

PR #1805, opened 2021-12-06
RFC: Send number of allocations in span as metric

ivoanjo:
I recently noticed that while we usually collect the number of memory allocations on a given span, we send it as an "allocations" field to the agent and this data never seems to make it to the backend.

Instead, this refactor reports it as a span metric. Does this make sense?

This is still missing tests and whatnot, but I wanted to get a feel if I'm doing this correctly before I did the full changes.

Also, as requested by @delner I'm targeting this to the 1.0 branch.

PR #597, opened 2018-10-24
Track the number of allocations made in each span

dasch:
Inspired by https://github.com/rails/rails/commit/42fec4b8de8c40d7778f936e200081c0dded1ed0, this change starts capturing the number of allocations made in each span.

I'm not sure how you'd like to integrate this data – do you just want to use a tag, or would you prefer a first-class attribute on Span?

Issue #468, opened 2018-06-20
Support for memory allocation metrics and statistics


w32-blaster:
In some cases it would be useful to see memory related metrics. It could help to find endpoints which are having higher memory footprint.

delner:
Thanks for the suggestion @w32-blaster . We currently have an issue #312 open that tracks these memory stats. Can we add any additional considerations there and close this one?

w32-blaster:
It's similar but that request is specific for garbage collection. Other tools support complete memory stats for requests, so you can see which are the more problematic requests/jobs in the app

marcotc:
👋 I captured your feature request, did some research and added implementation details in #2164.
#2164 will serve as our implementation plan and actionable discussion around how to collect and report allocation metrics.
I'll close this issue, but this feature development can be followed #2164.

PR #2635, opened 2023-02-20
[PROF-7252] Add `Profiling.allocation_count` API for new profiler

### LINKED ISSUE #2164: Collect Ruby VM internal event metrics (object allocation)
## Motivation

We would like more visibility over the internal overhead of Ruby into our application execution.
It can be useful to know if a web request has created many more objects that all other requests in the same application, for example.

## How
Ruby has internal events related to object lifecycle (`NEWOBJ`, `FREEOBJ`) that are broadcast to any registered internal tracepoint event listener.

[Here are all the events available in Ruby 3.1.2](https://github.com/ruby/ruby/blob/cb4e2cb55a59833fc4f1f6db2f3082d1ffcafc80/include/ruby/internal/event.h#L88-L94):
```c
#define RUBY_INTERNAL_EVENT_NEWOBJ          0x100000 /**< Object allocated. */
#define RUBY_INTERNAL_EVENT_FREEOBJ         0x200000 /**< Object swept. */
#define RUBY_INTERNAL_EVENT_GC_START        0x400000 /**< GC started. */
#define RUBY_INTERNAL_EVENT_GC_END_MARK     0x800000 /**< GC ended mark phase. */
#define RUBY_INTERNAL_EVENT_GC_END_SWEEP   0x1000000 /**< GC ended sweep phase. */
#define RUBY_INTERNAL_EVENT_GC_ENTER       0x2000000 /**< `gc_enter()` is called. */
#define RUBY_INTERNAL_EVENT_GC_EXIT        0x4000000 /**< `gc_exit()` is called. */
```

[Here's how to register an internal tracepoint event listener](https://github.com/ruby/ruby/blob/744d17ff6c33b09334508e8110007ea2a82252f5/ext/-test-/tracepoint/tracepoint.c#L60-L63):
```c
    VALUE tpval = rb_tracepoint_new(0, RUBY_INTERNAL_EVENT_NEWOBJ | RUBY_INTERNAL_EVENT_FREEOBJ |
				    RUBY_INTERNAL_EVENT_GC_START | RUBY_INTERNAL_EVENT_GC_END_MARK |
				    RUBY_INTERNAL_EVENT_GC_END_SWEEP,
				    tracepoint_track_objspace_events_i, &track);
```

One detail to keep in mind is that despite [`TracePoint`](https://ruby-doc.org/core-3.1.2/TracePoint.html) being a public Ruby class, it is not allowed to trace `RUBY_INTERNAL_EVENT_*` events. Some of these events are too high frequency (e.g. `NEWOBJ` would be too slow to trace with Ruby code) or happen at sensitive times (GC events don't play well with Ruby objects being allocated during GC), thus not available for Ruby callback tracing, only native tracing.

## What should be done

We should create native tracepoints to collect object allocation (`RUBY_INTERNAL_EVENT_NEWOBJ`). These numbers should be collected on a per-thread basis, in order to reduce the noise and allow for measuring allocations on a single trace.

The number of allocations should be stored on the top-level span.
Another option is to store this number for every span: if the overhead is not significant, this provides more value and makes it easier to pinpoint which span is creating an excessive number of objects, for example.

## Other considerations

As you can see above, we have GC internal events as well.
As tempting as they may seem to collect, because GC runs are global, we have enough information in the public [`GC.stat`](https://ruby-doc.org/core-3.1.2/GC.html#method-c-stat) method to gather GC metrics. `GC.stat` should be used instead, unless we see value in collecting internal GC details like Mark and Sweep events.

ivoanjo:
**What does this PR do?**:

This PR adds a new profiler public API: `Datadog::Profiling.allocation_count`.

The public documentation for this API is as follows:

> Returns an ever-increasing counter of the number of allocations observed by the profiler in this thread.
>
> Note 1: This counter may not start from zero on new threads. It should only be used to measure how many allocations have happened between two calls to this API:
> ```ruby
> allocations_before = Datadog::Profiling.allocation_count
> do_some_work()
> allocations_after = Datadog::Profiling.allocation_count
> puts "Allocations during do_some_work: #{allocations_after - allocations_before}"
> ```
> (This is similar to some OS-based time representations.)
>
> Note 2: All fibers in the same thread will share the same counter values.
>
> Only available when the profiler is running, the new CPU Profiling 2.0 profiler is in use, and allocation-related features are not disabled via configuration.
> For instructions on enabling CPU Profiling 2.0 see the ddtrace release notes.

On Ruby 2.x, as long as CPU Profiling 2.0 is in use, this API is enabled by default.

On Ruby 3.x, this API is disabled by default [because of a Ractor bug](https://bugs.ruby-lang.org/issues/18464). On Ruby 3 apps that don't use Ractors, this is still safe to enable (via configuration).

To manually enable or disable this feature, this PR adds a new setting:

```ruby
Datadog.configure do |c|
  c.profiling.advanced.allocation_counting_enabled = # ...
end
```

**Motivation**:

This feature has long been something we want to provide with ddtrace, see issues #2164 and #468, as well as PRs #1891, #1805, #597.

As part of the ongoing work of enabling allocation profiling, counting the number of allocations comes at a very cheap cost since the profiler needs to have a `RUBY_INTERNAL_EVENT_NEWOBJ` tracepoint anyway -- it's just a matter of also incrementing a counter inside it.

**Additional Notes**:

Note that this does not yet change any user-visible feature for ddtrace. I'm counting on @marcotc to pick up the task of using this API to make some tracing magic :)

**This PR is on top of #2634 just for my convenience, but both changes are independent.**

**How to test the change?**:

This change includes code coverage.

---

Fixes #2164

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -153,6 +153,7 @@ void *simulate_sampling_signal_delivery(DDTRACE_UNUSED void *_unused);
 static void grab_gvl_and_sample(void);
 static void reset_stats(struct cpu_and_wall_time_worker_state *state);
 static void sleep_for(uint64_t time_ns);
+static VALUE _native_allocation_count(DDTRACE_UNUSED VALUE self);
 
 // Note on sampler global state safety:
 //
@@ -188,6 +189,7 @@ void collectors_cpu_and_wall_time_worker_init(VALUE profiling_module) {
   rb_define_singleton_method(collectors_cpu_and_wall_time_worker_class, "_native_stop", _native_stop, 2);
   rb_define_singleton_method(collectors_cpu_and_wall_time_worker_class, "_native_reset_after_fork", _native_reset_after_fork, 1);
   rb_define_singleton_method(collectors_cpu_and_wall_time_worker_class, "_native_stats", _native_stats, 1);
+  rb_define_singleton_method(collectors_cpu_and_wall_time_worker_class, "_native_allocation_count", _native_allocation_count, 0);
   rb_define_singleton_method(testing_module, "_native_current_sigprof_signal_handler", _native_current_sigprof_signal_handler, 0);
   rb_define_singleton_method(testing_module, "_native_is_running?", _native_is_running, 1);
   rb_define_singleton_method(testing_module, "_native_install_testing_signal_handler", _native_install_testing_signal_handler, 0);
@@ -774,4 +776,9 @@ static void sleep_for(uint64_t time_ns) {
       ENFORCE_SUCCESS_NO_GVL(errno);
     }
   }
+}
+
+static VALUE _native_allocation_count(DDTRACE_UNUSED VALUE self) {
+  // TODO
+  return Qnil;
 }
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -80,6 +80,7 @@ struct cpu_and_wall_time_worker_state {
   atomic_bool should_run;
 
   bool gc_profiling_enabled;
+  bool allocation_counting_enabled;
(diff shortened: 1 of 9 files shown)


marcotc (approved):
Awesome work! 🙇

PR #2637, opened 2023-02-22
make sure to assign a valid processor to the appsec component

GustavoCaso:
**What does this PR do?**

After merging https://github.com/DataDog/dd-trace-rb/pull/2632, we got alerted that our system tests were failing. 
The scenario that is failing is when the appsec processor fails to parse the WAF rules. 
The test fails when the app wants tries to shout down and we try to finalize the appsec component. The problem is that the processor was never ready, so finalizing a non-ready processor caused an error. 

This PR fix that case by making sure to check if the processor is ready to assign to the appsec component. 

**Additional Notes**

Should we make the `AppSec::Component.new` method private and only allow to use `build_appsec_component`? 

**How to test the change?**

### DIFF
--- a/lib/datadog/appsec/component.rb
+++ b/lib/datadog/appsec/component.rb
@@ -32,7 +32,7 @@ module Datadog
       end
 
       def shutdown!
-        if processor
+        if processor && processor.ready?
           processor.finalize
           @processor = nil
         end
--- a/lib/datadog/appsec/component.rb
+++ b/lib/datadog/appsec/component.rb
@@ -11,19 +11,31 @@ module Datadog
         def build_appsec_component(settings)
           return unless settings.enabled
 
-          new
+          processor = create_processor
+          new(processor: processor)
+        end
+
+        private
+
+        def create_processor
+          processor = Processor.new
+          return nil unless processor.ready?
+
+          processor
         end
       end
 
       attr_reader :processor
 
-      def initialize(processor: Processor.new)
+      def initialize(processor:)
         @processor = processor
       end
 
       def shutdown!
-        processor.finalize if processor
-        @processor = nil
+        if processor
+          processor.finalize
+          @processor = nil
+        end
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/datadog/appsec/component.rb
```diff
@@ -11,19 +11,31 @@ class << self
...
         @processor = processor
       end
 
       def shutdown!
-        processor.finalize if processor
-        @processor = nil
+        if processor
+          processor.finalize
+          @processor = nil
+        end
```
lloeki, line 38: To improve thread safety, `Processor` might be finalized already but not yet `nil`.

Also, prefer a guard clause:

```suggestion
        return if processor.nil? || !processor.ready?
        
        processor.finalize
        @processor = nil
```

lib/datadog/appsec/contrib/rack/request_middleware.rb
```diff
@@ -27,7 +27,7 @@ def call(env)
 
             processor = Datadog::AppSec.processor
 
-            return @app.call(env) unless processor.ready?
+            return @app.call(env) unless processor
```
lloeki, line 30: To improve thread safety, Processor might be finalized already but not yet nil:

```suggestion
            return @app.call(env) if processor.nil? || !processor.ready?
```

lloeki (approved):
LGTM, with a few suggestions.

We should improve thread safety further when we get to remote configuration.

lib/datadog/appsec/component.rb
GustavoCaso: I prefer a guard clause too. It looks like steep likes:

```ruby
def shutdown!
  if processor && processor.ready?
    processor.finalize
    @processor = nil
  end
end
```

With this form, steep is able to resolve the type checking issue we discussed earlier in which steep does not find the `finalize` method 

```
lib/datadog/appsec/component.rb:28:18: [error] Type `::Datadog::AppSec::Processor` does not have method `finalize`
│
└         processor.finalize if processor
                    ~~~~~~~~

```

PR #2638, opened 2023-02-22
Link to central Single Span Sampling docs

marcotc:
This PR addresses a TODO in our documentation, to link to the now public Single Span Sampling documentation.

It also clarifies that Single Span Sampling can only be used to keep spans, not drop spans.

The bulk of documentation is language-agnostic, thus located in the public link added to this PR. That documentation will also be expanded in the near future.


ivoanjo (approved):
:+1: LGTM

dussault-antoine (approved):

PR #2639, opened 2023-02-22
Flaky:Increase worker shutdown timing tolerance

marcotc:
This PR changes the timing constraints for the test for the timely shutdown of `Datadog::Tracing::Workers::AsyncTransport`. 
This is the worker used by the tracer's transport.

[This test has flaked recently in CI](https://github.com/DataDog/dd-trace-rb/actions/runs/4245838791/jobs/7381877786):
```
  1) Datadog::Workers::AsyncTransport integration tests when terminating the worker which overruns the timeout is expected to be within 0.5 of 1
     Failure/Error:
       expect(@shutdown_end - @shutdown_beg)
         .to be_within(0.5).of(
           Datadog::Tracing::Workers::AsyncTransport::SHUTDOWN_TIMEOUT
         )

       expected 1.762315 to be within 0.5 of 1
     # ./spec/datadog/tracing/workers_integration_spec.rb:251:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:225:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:117:in `block (2 levels) in <top (required)>'
     # ./vendor/bundle/ruby/3.2.0/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
     # ./vendor/bundle/ruby/3.2.0/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block (2 levels) in <top (required)>'
```

This PR uses our test default timeout of 5 seconds for this.

### DIFF
--- a/spec/datadog/tracing/workers_integration_spec.rb
+++ b/spec/datadog/tracing/workers_integration_spec.rb
@@ -243,13 +243,13 @@ RSpec.describe 'Datadog::Workers::AsyncTransport integration tests' do
     end
 
     context 'which overruns the timeout' do
-      let(:task) { proc { sleep(interval) } }
+      let(:task) { proc { sleep(10) } }
       let(:trace_task) { task }
       let(:service_task) { task }
 
-      it do
+      it 'interrupts the worker to speed up shutdown' do
         expect(@shutdown_end - @shutdown_beg)
-          .to be_within(0.5).of(
+          .to be_within(5).of(
             Datadog::Tracing::Workers::AsyncTransport::SHUTDOWN_TIMEOUT
           )
       end


spec/datadog/tracing/workers_integration_spec.rb
```diff
@@ -243,13 +243,13 @@ def wait_for_flush(num = 1, period = 0.1)
     end
 
     context 'which overruns the timeout' do
-      let(:task) { proc { sleep(interval) } }
+      let(:task) { proc { sleep(10) } }
```
marcotc, line 246: Because the now has a fixed lenience of 5 seconds (`be_within(5)`), I changed the background thread sleep time to match that fixed constraint: the job will try to sleep for 10 seconds, but should shut down earlier than that to pass.

ivoanjo: I wonder if it would make sense to even convert this to a regular `sleep()` with no timeout, since this test is explicitly testing the interruption, and thus the  `sleep` is never expected to finish by itself.

ivoanjo (approved):
:+1: LGTM

spec/datadog/tracing/workers_integration_spec.rb
marcotc: I wary of infinite sleeps because the test might get hung up forever in the future for some reason.
I set this timeout much larger than it should be, but still not too large to hand CI.

ivoanjo: Hmmm, fair enough, although I somewhat worry about the potential for "accidentally passes for the wrong reason".

PR #2640, opened 2023-02-23
Flaky:Remove race condition from integration test

marcotc:
This [flaky test](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/9045/workflows/6841a910-893c-4922-93af-6e4948a13d04/jobs/336502/tests) happens because the wait condition for a forked process is time-dependent, and short:
```
Failure/Error: it { expect(terminated_process).to eq(graceful_signal) }

  expected: "graceful"
       got: ""

  (compared using ==)
./spec/datadog/tracing/integration_spec.rb:572:in `block (4 levels) in <top (required)>'
./spec/spec_helper.rb:225:in `block (2 levels) in <top (required)>'
./spec/spec_helper.rb:117:in `block (2 levels) in <top (required)>'
/usr/local/bundle/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
/usr/local/bundle/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block (2 levels) in <top (required)>'
```

This PR waits for a message from the child process instead.

### DIFF
--- a/spec/datadog/tracing/integration_spec.rb
+++ b/spec/datadog/tracing/integration_spec.rb
@@ -548,11 +548,13 @@ RSpec.describe 'Tracer integration tests' do
             span.service = 'my.service'
           end
 
-          sleep(1)
+          write.write('!') # Signals that this fork is ready
+          sleep(5)
         end
 
-        # Give the fork a chance to setup and sleep
-        sleep(0.2)
+        # Wait for fork to get ready
+        IO.select([read], [], [], 5) # 5 second timeout
+        read.getc # Reads '!'
 
         # Kill the process
         write.close
--- a/spec/datadog/tracing/integration_spec.rb
+++ b/spec/datadog/tracing/integration_spec.rb
@@ -549,12 +549,13 @@ RSpec.describe 'Tracer integration tests' do
           end
 
           write.write('!') # Signals that this fork is ready
-          sleep(5)
+          sleep(5) # Should be interrupted
+          exit! # Should not be reached, will skip shutdown hooks
         end
 
         # Wait for fork to get ready
         IO.select([read], [], [], 5) # 5 second timeout
-        read.getc # Reads '!'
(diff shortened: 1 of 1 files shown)


spec/datadog/tracing/integration_spec.rb
```diff
@@ -548,11 +548,13 @@ def sample!(trace)
...
-          sleep(1)
+          write.write('!') # Signals that this fork is ready
+          sleep(5)
         end
 
-        # Give the fork a chance to setup and sleep
-        sleep(0.2)
+        # Wait for fork to get ready
+        IO.select([read], [], [], 5) # 5 second timeout
+        read.getc # Reads '!'
```
ivoanjo, line 557: Minor: Consider checking that we read the expected character, just in case something is wrong maybe?

spec/datadog/tracing/integration_spec.rb
```diff
@@ -548,11 +548,13 @@ def sample!(trace)
             span.service = 'my.service'
           end
 
-          sleep(1)
+          write.write('!') # Signals that this fork is ready
+          sleep(5)
```
ivoanjo, line 552: If I understand it correctly, the intention of this spec is to check that when the process receives a SIGTERM, the ddtrace shutdown runs as expected.

So, like in another PR I just reviewed, this `sleep` is always supposed to be interrupted, it's never supposed to "run out" -- so perhaps it should be a `sleep` without a timeout?

ivoanjo (approved):
:+1: LGTM

spec/datadog/tracing/integration_spec.rb
marcotc: I changed this so that it exists "badly" if the sleep is not interrupted. This will ensure the test fails if not interrupted.

I don't like the idea of a non-timeout wait in our test suite, given it can always go wrong and hand the test suite.

PR #1607, opened 2021-07-22
RFC: Add Sorbet typechecker to dd-trace-rb

ivoanjo:
## TL;DR

* Minimal bootstrapping for now.
* Type checking is optional. Default is no type checking for files   without a magic comment. Should have minor impact on  people that don't want to touch the type checking.
* `bundle exec srb tc` runs the type checking.
* I've enabled type checking during CI runs

## Notes for reviewing

* This PR has two commits -- one that has all of the setup and relevant things, and another which is 100% the result of running `bundle exec srb init`. I strongly suggest reviewing them separately, for added sanity. (If you want to quickly glance through the second commit -- your terminal won't choke on the diff like github does 😉 )
* ~~This PR depends on #1605, and CI is failing for unrelated reasons. Thus, I've opened it as a draft until both of them are cleared and the runway is clear. It otherwise is ready for review/discussion.~~ Now updated.

 ## What?

While not a replacement for all our tests and other tooling, a type checker is a nice extra line of defense for catching issues that we otherwise may have missed, such as #1602.

The Ruby ecosystem around type checking is still rather immature. You may have heard that Ruby 3 shipped "with type checking" but that's an oversimplification: it actually only shipped with a "language for describing types" and a "type analysis tool" (<https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/>).
You may have noticed that nowhere in the previous sentence the words "type checker" were included.

So, Ruby 3 itself provides no type checker. In practice, there are two typecheckers in "wide" use for Ruby:

* [Steep](https://github.com/soutaro/steep), created by a Ruby core team member that also worked on manny of the other type-related bits that made it to Ruby 3 (but steep itself was not made part of Ruby)
* [Sorbet](https://sorbet.org/), a type checker in wide use at Stripe and Shopify

This RFC PR does a first pass at onboarding Sorbet to our codebase. There's a number of limitations on this "first pass", which I document below.

 ## Why not Steep?

I've tried onboarding our codebase to steep, and gave up due to two big issues:

1. Steep relies on the Ruby 3 rbs format, but unlike Sorbet, there doesn't seem to be any way of starting small on a mostly-untyped codebase without any `.rbs` files, and iterate from there. There are tools to generate skeleton `.rbs` files (the `rbs` gem has two different modes for it + the `typeprof` gem), but they had a hard time with ddtrace.

[...]

PR #2647, opened 2023-02-24
Remove Sorbet typechecker

### LINKED ISSUE #2641: Remove sorbet
When we added Sorbet in #1607, I documented a number of limitations that we had with its integration (see PR).

That PR also ends with

> Because our usage of Sorbet is rather simple, we can at any point switch to Steep, or any other tool, or just disable it, if it's not providing value. So we are not locked at all into its usage.

Recently the team met and decided that indeed Sorbet, with the limitations we face on our codebase (cannot include type annotations, not built for libraries, hard to run on arm64), was causing more trouble than the value we were getting from it. Furthermore, in #2449 we've added Steep and that approach seems to be working well for the modules we're using it with.

So the outcome of our decision was to remove sorbet. I'll be opening a few PRs to remove all traces of it from our repository for now.

We can always go back in the future by referencing PRs that will be linked to this issue and restoring the changes.

ivoanjo:
**What does this PR do?**:

This PR is spiritually a revert of #1607, when we added the Sorbet typechecker to dd-trace-rb.

It includes three commits: One where we remove all configuration and scaffolding surrounding Sorbet, and others where we remove all of the `# typed: ...` magic comments and `include Kernel` definitions added to make Sorbet happy.

**Note**: You **definitely** want to review this PR commit-by-commit.

**Motivation**:

As documented in #2641, the team has decided that the value vs pain equation for Sorbet has shifted in the past months, and thus that it was time to remove Sorbet.

**Additional Notes**:

Sorbet type checking in CI was actually removed earlier this week in #2617.

**How to test the change?**:

CI should still be green.

---

Fixes #2641


lloeki (approved):
LGTM

For the fun of it (and also how I reviewed the last one):

```
$ git show origin/ivoanjo/drop-sorbet | grep -e '^-' | grep -v -e '---' | sort |uniq -c
   1107 -
    544 -# typed: false
    292 -# typed: ignore
     17 -# typed: strict
    305 -# typed: true
```

I was wondering about not seeing `.circleci/config.yml` being modified but it's been removed already!

PR #2648, opened 2023-02-24
Introduce AppSec::Instrumentation::Gateway::Argument

GustavoCaso:
This PR introduces `AppSec::Instrumentation::Gateway::Argument` class. 

This class is the expected object to be provided when passing information to `Gateway#push`

Each contrib integration can define their own `Argument` objects. The idea is to encapsulate all the logic regarding argument handling inside those classes.  This helps with encapsulation and ease of testing each argument individually.

For now, the parent class `AppSec::Instrumentation::Gateway::Argument` has no method defined, as there isn't any shared logic among the types of arguments. This could change in the future. 

The different classes added in this PR are:

`Datadog::AppSec::Instrumentation::Gateway::User`
`Datadog::AppSec::Contrib::Rack::Gateway::Request`
`Datadog::AppSec::Contrib::Rack::Gateway::Response`
`Datadog::AppSec::Contrib::Sinatra::Gateway::Request`
`Datadog::AppSec::Contrib::Sinatra::Gateway::RouteParams`
`Datadog::AppSec::Contrib::Rails::Gateway::Request`

`AppSec::Instrumentation::Gateway::User` is a particular case since it is framework agnostic and currently is only used in [`kit/identity.rb`](https://github.com/DataDog/dd-trace-rb/blob/9a64baa7ec1e21c857bd900f6e16e10895c0bcdb/lib/datadog/kit/identity.rb#L62)

For that reason, the definition of the class is inside `lib/datadog/appsec/instrumentation/gateway/argument.rb` and not in any `contrib` folder.


lib/datadog/appsec/instrumentation/gateway/object.rb
```diff
@@ -0,0 +1,24 @@
...
+module Datadog
+  module AppSec
+    module Instrumentation
+      class Gateway
+        class Object; end
+
+        class User < Object
+          attr_reader :id
+
+          def initialize(id)
```
ivoanjo, line 17: I somewhat shudder at re-using the word `Object`, since it's something really core to Ruby as well. Of course this doesn't clash for the programming language itself but... it may be really confusing from time to time.

Maybe consider `Payload`? `Argument`? `Value`? Something like that?

GustavoCaso: Absolutely. We are experimenting with the naming and implementation. Still, totally agree that the object might be too generic and also lead to confusion with the ruby `Object` class. 

I like `Argument`

lib/datadog/appsec/contrib/sinatra/patcher.rb
```diff
@@ -55,7 +57,9 @@ def dispatch!
 
             # TODO: handle exceptions, except for super
 
-            request_return, request_response = Instrumentation.gateway.push('sinatra.request.dispatch', request) do
+            gateway_request = Datadog::AppSec::Contrib::Rack::Gateway::Request.new(env)
```
lloeki, line 60: WDYT of having a `Contrib::Sinatra::Gateway::Request` inheriting from `Contrib::Rack::Gateway::Request`? instead of depending on the latter directly?

lib/datadog/appsec/contrib/rails/gateway/request.rb
```diff
@@ -0,0 +1,67 @@
...
+
+require_relative '../../../instrumentation/gateway/argument'
+
+module Datadog
+  module AppSec
+    module Contrib
+      module Rails
+        module Gateway
+          # Gateway Request argument. Normalized extration of data from ActionDispatch::Request
+          class Request < Instrumentation::Gateway::Argument
```
lloeki, line 11: Would it make sense to inherit from `Contrib::Rack::Gateway::Request`?

lib/datadog/appsec/contrib/rack/gateway/watcher.rb
```diff
@@ -65,10 +65,10 @@ def watch_request(gateway = Instrumentation.gateway)
               end
 
               def watch_response(gateway = Instrumentation.gateway)
-                gateway.watch('rack.response', :appsec) do |stack, response|
+                gateway.watch('rack.response', :appsec) do |stack, gateway_response|
                   block = false
                   event = nil
-                  waf_context = response.instance_eval { @waf_context }
+                  waf_context = AppSec::Processor.active_context
```
lloeki, line 71: This can be problematic, for instance because of [stuff like this](https://github.com/rails/rails/blob/1512cf2ba578282c898b8eb68ac401ea5243b7b5/actionpack/lib/action_controller/metal/live.rb#L261-L263).

That's why I favoured accessing the context from a known source. Now that we have objects in our control we can have the active context accessible without a `instance_eval` hack.

lloeki (approved):
LGTM, except usage of `AppSec::Processor.active_context` (which uses a thread local) should be made minimal when it can be accessed reliably otherwise.

lib/datadog/appsec/contrib/rails/gateway/request.rb
GustavoCaso: At first, I thought about, but since I saw we use the `Rack::RequestBody` directly in the [Sinatra contrib folder](https://github.com/DataDog/dd-trace-rb/pull/2648/files#diff-b20f56877521f6a704cab6694c422e0bb66bad6eff3d0d9b7b05188aa976a0dbR49) I thought it would be ok. 

I'm happy to change it.

PR #2649, opened 2023-02-27
Fix 'uninitialized constant GRPC::Interceptor' when with the 'gapic-common' gem

marcotc:
This PR fixes the error 
```
Failed to apply Datadog::Tracing::Contrib::GRPC::Patcher patch. Cause: uninitialized constant GRPC::Interceptor
```
when gRPC was partially loaded.

This has been reported to us by users of the Google [Calendar User Availability API](https://developers.google.com/calendar/api/user-availability/reference/rest) gem. That gem is not publicly available yet, but its dependency on the public [`gapic-common` gem](https://rubygems.org/gems/gapic-common) is where the issue lies, allowing us to address this issue.

`gapic-common` only loads [a couple file from gPRC](https://github.com/googleapis/gapic-generator-ruby/blob/f1c2e73219453e497b6ec2dc807a907e939e1342/gapic-common/lib/gapic/common.rb#L15-L16), which only contain constants and error classes.

There's nothing of value to instrument if only such subset of gPRC is loaded.

This PR ensures we check for the presence of the relevant gPRC client and server files that we actually have an instrumentation point.

### DIFF
--- a/lib/datadog/tracing/contrib/grpc/integration.rb
+++ b/lib/datadog/tracing/contrib/grpc/integration.rb
@@ -20,7 +20,14 @@ module Datadog
           end
 
           def self.loaded?
-            !defined?(::GRPC).nil?
+            !defined?(::GRPC).nil? &&
+              # When using the Google "Calendar User Availability API"
+              # (https://developers.google.com/calendar/api/user-availability/reference/rest), though the gem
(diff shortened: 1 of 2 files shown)


ivoanjo (approved):
:+1: LGTM

TonyCTHsu (approved):

PR #2650, opened 2023-02-28
Update the appraisal gemfiles for ruby versions

GustavoCaso:
Executed `bundle exec rake appraisal:update` for ruby versions 😄


TonyCTHsu (approved):
👍 with Green CI

PR #2654, opened 2023-03-01
[PROF-7307] Rename `CpuAndWallTime` collector to `ThreadContext` collector

ivoanjo:
**What does this PR do?**:

This PR renames the `Collectors::CpuAndWallTime` to `Collectors::ThreadContext`, including all references to the old name.

It also expands the documentation on the new profiler.

**Motivation**:

As I'm working on implementing allocation profiling, I realized that it was going to heavily reuse the `CpuAndWall` time collector behavior, but just with a slightly different metric (allocation counts vs cpu-time).

The emerging pattern seems to be that if we're going to take a stack sample for some profiling type, we'll always need to need to include the labels and other info gathered by this component.

Thus, re-examining it, I decided to rename it to make it closer to that expectation.

**Additional Notes**:

See the changes to `ProfilingDevelopment.md` for more details.

**How to test the change?**:

We already have test coverage for this component.


TonyCTHsu (approved):
Renaming is fine without behavior changes!

PR #2656, opened 2023-03-01
Fix system test runner commit sha

TonyCTHsu:
**What does this PR do?**

I figured out that the commit SHA must be on existing master branch, since the previous sha fails after PR merged/branch deleted.

### DIFF
--- a/.github/workflows/lib-injection.yml
+++ b/.github/workflows/lib-injection.yml
@@ -51,7 +51,7 @@ jobs:
     steps:
       - name: lib-injection test runner
         id: lib-injection-test-runner
-        uses: DataDog/system-tests/lib-injection/runner@1af3241d5b6a928199528a8cbfc5698564f5d260
+        uses: DataDog/system-tests/lib-injection/runner@4eb1ec01b61c1c7e84f219127a91d07f03b0f85e
         with:
           docker-registry: ghcr.io
           docker-registry-username: ${{ github.repository_owner }}


.github/workflows/lib-injection.yml
```diff
@@ -51,7 +51,7 @@ jobs:
     steps:
       - name: lib-injection test runner
         id: lib-injection-test-runner
-        uses: DataDog/system-tests/lib-injection/runner@1af3241d5b6a928199528a8cbfc5698564f5d260
+        uses: DataDog/system-tests/lib-injection/runner@4eb1ec01b61c1c7e84f219127a91d07f03b0f85e
```
ivoanjo, line 54: Is there any way to get the "latest", or "master"? Or would we want to explicitly always use an older version here?

(Also, would it make sense to leave a comment here explaining how to update to the latest version?)

ivoanjo (approved):
:+1: Thanks for looking into this!

.github/workflows/lib-injection.yml
GustavoCaso: I was going to suggest always pointing to `master|main`. Could we do that?

GustavoCaso (approved):

.github/workflows/lib-injection.yml
TonyCTHsu: Nope. It is a SHA.

ivoanjo: Maybe we could use https://github.com/marketplace/actions/latest-tag ? That seems exactly what we're missing here.

PR #2657, opened 2023-03-01
[PROF-7307] Add support for allocation samples to `ThreadContext`

ivoanjo:
**What does this PR do?**:

This PR adds the `thread_context_collector_sample_allocation` API to the `ThreadContext` collector.

This will be used to support allocation profiling (but right now it's not being called automatically when allocations happen, yet).

**Motivation**:

This is an incremental step towards supporting allocation profiling.

I'm still working on the scaling/weighting strategy, and may end up adding that logic directly on the `ThreadContext`, haven't decided yet.

**Additional Notes**:

I'm opening this PR is on top of #2654 to avoid merge conflicts, but this PR is otherwise completely independent from it.

**How to test the change?**:

Change includes test coverage. End-to-end testing will come later once this is wired up in the `CpuAndWallTimeWorker`.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,5 +1,8 @@
 $LOAD_PATH.unshift File.expand_path('..', __dir__)
 $LOAD_PATH.unshift File.expand_path('../lib', __dir__)
+
+Thread.main.name = 'Thread.main' unless Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.3')
+
 require 'pry'
 require 'rspec/collection_matchers'
 require 'rspec/wait'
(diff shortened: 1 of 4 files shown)


spec/spec_helper.rb
```diff
@@ -1,5 +1,8 @@
 $LOAD_PATH.unshift File.expand_path('..', __dir__)
 $LOAD_PATH.unshift File.expand_path('../lib', __dir__)
+
+Thread.main.name = 'Thread.main' unless Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.3')
```
GustavoCaso, line 4: Why are we adding this change?

ivoanjo: It's to make it easier to debug when we have multiple threads -- if you print/list the threads it's sometimes not obvious which thread is the main one (and main is the one driving the RSpec framework). It's unrelated to the other changes (which is why I put it on a separate commit -- 67b7fa84e4d77e91b414fa7992f00d7762475bef ).

ivoanjo: Here's one example where this is useful. [This spec failed](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/9211/workflows/8431a38e-fa1d-4a93-a937-b8488b960a89/jobs/342577/tests#failed-test-0) and printed the list of threads:

```
[#<Thread:0x000055df58fcafc0@Thread.main run>, #<Thread:0x000055df5b3d1f90 /app/spec/spec_helper.rb:241 sleep>, #<Thread:0x000055df5d235c48 /usr/local/bundle/gems/webrick-1.7.0/lib/webrick/utils.rb:158 sleep_forever>]
```

GustavoCaso (approved):

PR #2658, opened 2023-03-01
Negative test cases for lib injection

TonyCTHsu:
**What does this PR do?**

I have prepared some apps to demonstrate the constraints/limitation for Ruby lib injection. However, our system test has not provided the framework to conduct test for that. So I wrote a workflow for those apps to make sure the apps keep running even with lib injection failure.

### DIFF
--- a/.github/workflows/lib-injection.yml
+++ b/.github/workflows/lib-injection.yml
@@ -57,3 +57,43 @@ jobs:
           docker-registry-username: ${{ github.repository_owner }}
           docker-registry-password: ${{ secrets.GITHUB_TOKEN }}
           test-script: ./lib-injection/run-manual-lib-injection.sh
+
+  test-negative:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        variant:
+          - dd-lib-ruby-init-test-rails-conflict
+          - dd-lib-ruby-init-test-rails-bundle-deploy
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


Kyle-Verhoog (approved):
lgtm!

.github/workflows/lib-injection.yml
```diff
@@ -57,3 +57,43 @@ jobs:
           docker-registry-username: ${{ github.repository_owner }}
           docker-registry-password: ${{ secrets.GITHUB_TOKEN }}
           test-script: ./lib-injection/run-manual-lib-injection.sh
+
+  test-negative:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        variant:
+          - dd-lib-ruby-init-test-rails-conflict
```
ivoanjo, line 66: I'm curious -- do you know why there's no error message in https://github.com/DataDog/dd-trace-rb/actions/runs/4307107275/jobs/7511764104 ?

ivoanjo (approved):
I just pressed the approve button but... I have some concerns about where we're going with this approach.

Specifically, having a new CI check per single test, that includes a full image build and whatnot seems... very heavy. This doesn't seem like it will scale if we add N new singular CI checks for every feature.

The GitHub UI certainly doesn't scale. It's already hard to find which test failed with the current ~100 checks, I'm not looking forward to it when there's 200...

So... Consider when building more tests, if they can be reasonably packaged together, so we can curb the growth a bit.

PR #2663, opened 2023-03-03
fix: prevent unintentional modification of arguments

coneill-enhance:
**What does this PR do?**
Fixes what appears to be a bug

**Motivation**
1. The original code looks as though this was the intention[1]
2. It's bad practice to unintentionally modify the parameters of a function

*1: In the original code:
```ruby
          user_options = user.dup
          user_id = user.delete(:id)

          raise ArgumentError, 'missing required key: :user => { :id }' if user_id.nil?

          Kit::Identity.set_user(trace, id: user_id, **user_options)
```

`Kit::Identity.set_user(trace, id: user_id, **user_options)` would be redundant as `:id` exists in `user_options` and would replace the previously specified argument

Additionally, if this was intended, then the following is equivalent and better:
```ruby
          user_id = user.delete(:id)

          raise ArgumentError, 'missing required key: :user => { :id }' if user_id.nil?

          Kit::Identity.set_user(trace, id: user_id, **user)
```

or, not equivalent, but better still:

```ruby
          raise ArgumentError, 'missing required key: :user => { :id }' if user[:id].nil?

          Kit::Identity.set_user(trace, **user)
```

in all cases you still get a non-descriptive error if `user` is `nil`

Also checking for `user.id` here is redundant, and somewhat irrelevant as this is done again by:
https://github.com/DataDog/dd-trace-rb/blob/1d20aa8bd58ccd60265fd0d3c7e0de58456242cc/lib/datadog/kit/identity.rb#L33

So probably just go with:

```ruby
          Kit::Identity.set_user(trace, **user)
```

### DIFF
--- a/lib/datadog/kit/appsec/events.rb
+++ b/lib/datadog/kit/appsec/events.rb
@@ -23,7 +23,7 @@ module Datadog
           track(LOGIN_SUCCESS_EVENT, trace, **others)
 
           user_options = user.dup
-          user_id = user.delete(:id)
+          user_id = user_options.delete(:id)
 
           raise ArgumentError, 'missing required key: :user => { :id }' if user_id.nil?
 
--- a/spec/datadog/kit/appsec/events_spec.rb
+++ b/spec/datadog/kit/appsec/events_spec.rb
@@ -32,6 +32,15 @@ RSpec.describe Datadog::Kit::AppSec::Events do
       end
       expect(meta).to include('usr.id' => '42', 'appsec.events.users.login.success.foo' => 'bar')
     end
+
+    it 'maintains integrity of user argument' do
+      user_argument = { id: '42' }
+      user_argument_dup = user_argument.dup
+      trace_op.measure('root') do
+        described_class.track_login_success(trace_op, user: user_argument, foo: 'bar')
+      end
+      expect(user_argument).to eql(user_argument_dup)
+    end
   end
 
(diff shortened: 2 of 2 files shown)


GustavoCaso:
@coneill-enhance Thank you so much for the code changes. 

It makes total sense not to modify the original user object. As for the suggestions to change the internal code to make the API easier to work with are also legit. Going to check some internal documents regarding the internals of the API, and if there is room to make those adjustments I will make sure to add them. 

In the meantime, if you don't mind, I will make sure to merge your changes.

GustavoCaso:
A suggestion. It would be good to modify the test in `spec/datadog/kit/appsec/events_spec.rb` to ensure the user object does not get modified when calling `track_login_success`.

If you do not have the bandwidth to tackle it, not a problem. Let me know. I would happily add them myself. 

Again, thanks for the contributions @coneill-enhance

GustavoCaso:
Amazing work @coneill-enhance 

Yeah, the test is not related to your change. We are aware that we have some flaky tests on our end. Not worry I triggered another run. 

Once the CI is green I will merge the PR. 

Thanks for your contribution.

GustavoCaso (approved):
Thank you so much for your contribution.

GustavoCaso:
@coneill-enhance. We have released 1.10.1 with the fix you summited. 

Thanks for your contribution.

PR #2664, opened 2023-03-03
Clarify non-framework instrumentation use case

marcotc:
This PR clarifies that the `Ruby applications` category of apps does not refer to all Ruby applications, but instead to applications with frameworks we don't have special integration for.

Also, this fix fixes all the broken internal markdown links in `docs/GettingStarted.md`.

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -28,8 +28,8 @@ To contribute, check out the [contribution guidelines][contribution docs] and [d
  - [Installation](#installation)
      - [Setup the Datadog Agent for tracing](#setup-the-datadog-agent-for-tracing)
      - [Instrument your application](#instrument-your-application)
-        - [Rails applications](#rails-applications)
-        - [Ruby applications](#ruby-applications)
+        - [Rails or Hanami applications](#rails-or-hanami-applications)
+        - [Other Ruby applications](#other-ruby-applications)
         - [Configuring OpenTracing](#configuring-opentracing)
         - [Configuring OpenTelemetry](#configuring-opentelemetry)
      - [Connect your application to the Datadog Agent](#connect-your-application-to-the-datadog-agent)
@@ -219,7 +219,7 @@ OR
 
 ### Instrument your application
 
-#### Rails/Hanami applications
+#### Rails or Hanami applications
 
 1. Add the `ddtrace` gem to your Gemfile:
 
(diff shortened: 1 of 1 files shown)


docs/GettingStarted.md
```diff
@@ -244,7 +244,9 @@ OR
       - [Add additional configuration settings](#additional-configuration)
       - [Activate or reconfigure instrumentation](#integration-instrumentation)
 
-#### Ruby applications
+#### Other Ruby applications
+
+If you application does not use of the supported gems above, you can set it up as follows:
```
wantsui, line 249:
```suggest
If your application does not use the supported gems (Rails or Hanami) above, you can set it up as follows:
```

wantsui (commented):
Looks good to me! I made a small comment about one of the sentences in the "Other Ruby applications".

docs/GettingStarted.md
```diff
@@ -335,7 +337,7 @@ If your Datadog Agent is listening at any of these locations, no further configu
 If your agent runs on a different host or container than your application, or you would like to send traces via a different protocol, you will need to configure your application accordingly.
 
   - [How to send trace data via HTTP over TCP to agent](#changing-default-agent-hostname-and-port)
-  - [How to send trace data via Unix Domain Socket (UDS) to agent](#using-the-unix-domain-socket-uds-adapter)
+  - [How to send trace data via Unix Domain Socket (UDS) to agent](#using-the-unix-domain-socket--uds--adapter)
```
ivoanjo, line 340: Hmm at least for me both GitHub as well as docs.datadoghq.com are still rendering this anchor as `using-the-unix-domain-socket-uds-adapter` (the old version), and not the new one.

ivoanjo (approved):
LGTM :+1:

GustavoCaso (approved):

docs/GettingStarted.md
```diff
@@ -28,8 +28,8 @@ To contribute, check out the [contribution guidelines][contribution docs] and [d
  - [Installation](#installation)
      - [Setup the Datadog Agent for tracing](#setup-the-datadog-agent-for-tracing)
      - [Instrument your application](#instrument-your-application)
-        - [Rails applications](#rails-applications)
-        - [Ruby applications](#ruby-applications)
+        - [Rails or Hanami applications](#rails-or-hanami-applications)
```
GustavoCaso, line 31: Should we include Sinatra as well?

marcotc: Sinatra does not have the same framework support because we do not have a reliable instrumentation point that happens "after all gems have been required by the host application".
Both Rails and Hanami have such callback.

docs/GettingStarted.md
marcotc: Fixed it, thank you!
My local markdown rendered expected something different from GitHub and our public docs. The updated version works for both GitHub and public docs.

docs/GettingStarted.md
GustavoCaso: Thanks for the explanation

PR #2669, opened 2023-03-06
Capture user-relevant implementation details for the Processing Pipeline

marcotc:
This PR adds a few important details about how the Processing Pipeline behaves when activate by the user.

These concerns have been brought up in the past and can cause confusion when not known by the application maintainer.

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -2549,8 +2549,9 @@ In both cases, the processor method *must* return the `trace` object; this retur
 
 #### Caveats
 
-1. The [debug mode logs](#enabling-debug-mode) reports the state of spans *before* the Processing Pipeline is executed.
-2. Removing a span in the Processing Pipeline also removes all children spans from the removed span. This prevents orphan spans in the trace graph.
+1. Removed spans will not generate trace metrics, affecting monitors and dashboards.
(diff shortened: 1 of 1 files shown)


ivoanjo (approved):
:+1: LGTM

TonyCTHsu (approved):

docs/GettingStarted.md
```diff
@@ -2547,6 +2547,11 @@ Datadog::Tracing.before_flush(MyCustomProcessor.new)
 
 In both cases, the processor method *must* return the `trace` object; this return value will be passed to the next processor in the pipeline.
 
+#### Caveats
+
+1. The [debug mode logs](#enabling-debug-mode) reports the state of spans *before* the Processing Pipeline is executed.
+2. Removing a span in the Processing Pipeline also removes all children spans from the removed span. This prevents orphan spans in the trace graph.
```
wantsui, line 2553: A third item to add here is also that because the spans get removed, trace metrics will exclude these spans.
ie:
3.  Removing a span in the Processing Pipeline will also remove the span from being included in trace metrics.

marcotc:
@ivoanjo @wantsui, can you take a look again?
I'm trying to make it succinct but accurate and clear.

wantsui (approved):
Looks good to me now, thanks for adding this!

PR #2672, opened 2023-03-07
RFC: Run steep typechecker on new files by default

ivoanjo:
**What does this PR do?**:

This PR takes a page from my previous integration of Sorbet (#1607) by changing the type checking from opt-in to opt-out.

AKA by default all files will be type checked, unless they are ignored via the `Steepfile`. My intention here is to make this a bit like the `.rubocop_todo.yml` file -- whatever is in, is in, but over time we can add more and more type signatures.

I've also tweaked the type checker CI task to go through rake, and to print a nice error message to guide people that haven't used Steep before.

I'm opening this as a an RFC as I'd like to know the team's thoughts on this approach.

**Motivation**:

Gently prod us to adopt more typechecking :)

**Additional Notes**:

Steep actually spotted a bug (#2671) while I was testing this change out.

**How to test the change?**:

Check that CI is still green!


lloeki:
Interesting approach! I like that :D

lloeki (approved):

GustavoCaso (approved):
Fantastic 🎉

ivoanjo:
Cool! I'm going to merge this in, let's see how it feels. If it gets too annoying, we can back out of it again :)

PR #2673, opened 2023-03-07
Rename conflicting `send_payload` method

lloeki:
**What does this PR do?**

Move a generically named method out of the way.

**Motivation**

The transport infrastructure is mostly generic, except for a few choke points. This addresses one of them.

The `send_payload` method, while defined in a trace-specific module, is included at the class level.

When attempting to reuse transport for other use cases, the method conflicts with other modules extending the generic client class.

Luckily its use is very limited and self-contained to a trace transport specific scope. By renaming it additional modules are able to be included and can follow a consistent pattern without conflict.

**Additional Notes**

This has been discovered while implementing additional transport facilities for remote configuration, from an AppSec impetus.

The whole transport code is fairly involved and due for a rewrite, nonetheless it handles many corner cases. which would make a drop-in rewrite carry much risk to tracing.

Therefore a more segregated approach is being taken to implement the required additional transports, aiming:
- to be a second use case, highlighting the limitations and hopefully providing insights for improvements.
- to be of little invasiveness to the current transport code; still, some are required.

This is one of those changes.

**How to test the change?**

CI should be green.

### DIFF
--- a/lib/ddtrace/transport/http/traces.rb
+++ b/lib/ddtrace/transport/http/traces.rb
@@ -25,7 +25,7 @@ module Datadog
 
         # Extensions for HTTP client
         module Client
-          def send_payload(request)
+          def send_traces_payload(request)
             send_request(request) do |api, env|
               api.send_traces(env)
             end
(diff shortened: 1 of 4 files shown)


GustavoCaso (approved):
The change LGTM. 

Regarding your comment: 
> When attempting to reuse transport for other use cases, the method conflicts with other modules extending the generic client class.

Would be interesting to provide an example 😄

PR #2675, opened 2023-03-08
Fix profiling/validate_benchmarks_spec failing with benchmark-ips 2.12 + fix CI failures due to steep 1.3.1

ivoanjo:
**What does this PR do?**:

This PR fixes the `spec/datadog/profiling/validate_benchmarks_spec.rb` failing with benchmark-ips 2.12 with
`uninitialized constant Benchmark::IPS::NoopSuite (NameError)`.

The specific issue is caused by a refactor introduced in https://github.com/evanphx/benchmark-ips/pull/111 which removed the `NoopSuite` class.

Previously, we were subclassing it to implement our own custom "suite" which allowed reporting of benchmark-ips results via dogstatsd.

The only usefulness of this class was providing no-op implementations for all methods that are required by benchmark-ips, so I've now just directly added them to our class.

Furthermore, while working on this PR, I've also ran into an issue with the also-recently-released steep 1.3.1 failing in github actions CI when running on Ruby 3.0 and 3.1 due to:

> ``/Users/runner/hostedtoolcache/Ruby/3.1.3/x64/lib/ruby/gems/3.1.0/gems/bundler-2.4.7/lib/bundler/runtime.rb:304:in `check_for_activated_spec!': You have already activated pathname 0.2.0, but your Gemfile requires pathname 0.2.1. Since pathname is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports pathname as a default gem. (Gem::LoadError)``

For the life of me, I tried updating to the latest bundler, different ways of invoking rake (binstubs, ...) and could not fix this issue. As a final workaround, I decided to opt for not installing steep on the offending setups. Hopefully, if you want to run steep on macos, you can either manually remove this, or use a different Ruby (like 3.2)

**Motivation**:

Fix broken CI! :)

**Additional Notes**:

N/A

**How to test the change?**:

Validate that CI is back to green.


ivoanjo:
Converted to draft while I try to also fix the steep Ruby 3.0/3.1 CI incompatibility

GustavoCaso (approved):

PR #2686, opened 2023-03-13
[APPSEC-8287] Add AppSec::Processor::RuleMerger

GustavoCaso:
In Appsec, there has to be some transformation when the remote configuration client is going to merge existing information with the provided by the agent. 

The agent can provide multiple pieces of information regarding new rules data, rules override, or entirely new rules set. 

Before being able to instantiate a new `AppSec::Processor` we need to make sure the rules payload has the correct format. 

The correct format is:

```
{
    version: ...,
    metadata: ...,
    rules: [...],
    exclusions: [...],
    rules_override: [...],
    rules_data: [...]
}
```

There is also some merging logic we need to account for when merging `rules_data`

When combining rules data with the same id, type and value, we need to keep the one with the highest expiration or if no expiration is provided.

### DIFF
--- a/lib/datadog/appsec/processor/rule_merger.rb
+++ b/lib/datadog/appsec/processor/rule_merger.rb
@@ -6,19 +6,54 @@ module Datadog
       # RuleMerger merge different sources of information
       # into the rules payload
       module RuleMerger
+        # RuleVersionMismatchError
+        class RuleVersionMismatchError < StandardError
+          def initialize(version1, version2)
+            msg = 'Merging rule files with different version could lead to unkown behaviour. '\
+              "We have receieve two rule files with versions: #{version1}, #{version2}. "\
+              'Please validate the configuration is correct and try again.'
+            super(msg)
+          end
+        end
+
         class << self
           def merge(rules:, data: nil, overrides: nil)
-            rules_dup = rules.dup
+            combined_rules = combine_rules(rules)
 
             rules_data = combine_data(data) if data
             rules_overrides_and_exclusions = combine_overrides(overrides) if overrides
 
-            rules_dup.merge!(rules_data) if rules_data
-            rules_dup.merge!(rules_overrides_and_exclusions) if rules_overrides_and_exclusions
-            rules_dup
+            combined_rules.merge!(rules_data) if rules_data
+            combined_rules.merge!(rules_overrides_and_exclusions) if rules_overrides_and_exclusions
+            combined_rules
           end
 
           private
+
+          def combine_rules(rules)
+            return rules[0] if rules.size == 1
+
+            final_rules = []
+            # @type var final_version: ::String
+            final_version = (_ = nil)
+
+            rules.each do |rule_file|
+              version = rule_file['version']
+
+              if version && !final_version
+                final_version = version
+              elsif final_version != version
+                raise RuleVersionMismatchError.new(final_version, version)
+              end
+
+              final_rules.concat(rule_file['rules'])
+            end
+
+            {
+              'version' => final_version,
+              'rules' => final_rules
+            }
+          end
 
           def combine_data(data)
             result = []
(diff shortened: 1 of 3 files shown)


lib/datadog/appsec/processor/rule_merger.rb
```diff
@@ -0,0 +1,107 @@
...
+            end
+
+            data2.each do |data|
+              if result.key?(data['value'])
+                # The value is duplicated so we need to keep
+                # the one with the highest expiration value
+                # We replace it if the expiration is higher than the current one
+                # or if no experiration
+                expiration = result[data['value']]
+                result[data['value']] = data['expiration'] if data['expiration'].nil? || data['expiration'] > expiration
```
lloeki, line 66: Could that last `expiration` be `nil`?

lib/datadog/appsec/processor/rule_merger.rb
```diff
@@ -0,0 +1,107 @@
...
+
+module Datadog
+  module AppSec
+    class Processor
+      # RuleMerger merge different sources of information
+      # into the rules payload
+      module RuleMerger
+        class << self
+          def merge(rules:, data: nil, overrides: nil)
+            rules_dup = rules.dup
```
lloeki, line 11: Looking at the logic below it's a bit hard to track whether no nested mutation happens.

WDYT of adding some deep `freeze` to the `let`s in specs so that we ensure we never mutate the originals?

lib/datadog/appsec/processor/rule_merger.rb
```diff
@@ -0,0 +1,107 @@
...
+                    result << value
+                  end
+                else
+                  # First entry for that id
+                  result << value
+                end
+              end
+            end
+
+            return unless result.any?
```
lloeki, line 49: I guess this is because empty rulesets are disallowed by `libddwaf`

lib/datadog/appsec/processor/rule_merger.rb
```diff
@@ -0,0 +1,107 @@
+# frozen_string_literal: true
+
+module Datadog
+  module AppSec
+    class Processor
+      # RuleMerger merge different sources of information
+      # into the rules payload
+      module RuleMerger
```
lloeki, line 8: This is not merging rules but collections of rules, expirations, and overrides. I'm not happy with the name but I can't figure another one right now so let's go with that.

lloeki (approved):
LGTM, but I think it needs a few adjustments.

Also, I feel like it would make sense to rebase this PR onto the chain of PRs, the latest `add-remote-config-component` one would be a good candidate as a rebase target + base branch for the PR. This would make integration of the merger usable immediately.

ivoanjo:
> Also, I feel like it would make sense to rebase this PR onto the chain of PRs, the latest add-remote-config-component one would be a good candidate as a rebase target + base branch for the PR. This would make integration of the merger usable immediately.

Consider maybe doing it the other way -- merging this to master (as long as it has no impact on existing users) and merging master into the working branch. This way allows incremental changes to land in master, rather than a big "land-everything" PR that is very hard to review.

lloeki:
> Consider maybe doing it the other way -- merging this to master (as long as it has no impact on existing users) and merging master into the working branch. This way allows incremental changes to land in master, rather than a big "land-everything" PR that is very hard to review.

I'm confused, there is no "big PR", there is a chain of (currently) seven PRs, each with a specific and progressive subset of change that should be easier to review in order and in isolation than a big PR, and each allowing incremental changes to land. Whether this PR here is incrementally merged now or downstream of the sequence of PRs is immaterial.

Each PR having their own branch based on the previous one, merging this to master and then merging master to the (currently) last branch would effectively freeze all the branches since they would not be rebase-able anymore†. So any change on a specific "top" branch would mean merging that branch down, once in each downstream branch. This would make history a terrible mess. The alternative would be to merge only on the last one, effectively turning the chained but independent branches into one big PR behind the scenes.

† I can see some ways to work around that, like merge-preserving rebasing only the last one, then `branch -f` on each intermediate branches to the newly rebased commits.

PR #2690, opened 2023-03-13
[PROF-7307] Avoid triggering allocation sampling during sampling

ivoanjo:
**What does this PR do?**:

In rare cases (such as when throwing exceptions), the profiler may need to allocate Ruby objects during sampling.

This object allocation will trigger the `object_allocation_tracepoint` but we don't want to try to sample in the middle of sampling.

To avoid any issues altogether, I've introduced a `during_sample` flag that can be used to detect this and avoid trying to do sampling during sampling.

Note that the functions modified all run while holding on to the global VM lock, which is why this flag does not need any special concerns re: concurrency.

I've also chosen to keep a counter for how often this happens that may come in handy when debugging any issues around this (and should be cheap enough to keep).

**Motivation**:

This PR is one more piece on the path to providing allocation profiling. The key missing part is the "// TODO: Sampling goes here", which is what will actually trigger recording of stack traces when objects get allocated.

In the spirit of small, incremental changes, I'm opening a PR with just this, and will later follow up with the rest of the implementation.

**Additional Notes**:

I also slightly moved around the declarations of fields inside `struct cpu_and_wall_time_worker_state` and in the constructor. **I've made this change in a separate commit, so I recommend reviewing this PR commit-by-commit as that makes a lot more sense.**

**How to test the change?**:

This change is a bit hard to test, since it affects only the C-level internal state of the profiler. I can't think of a great way of testing this that doesn't involve adding a lot of complexity to the profiler C code just for the purpose of enabling testing.

Instead, I plan to test this using integration testing, e.g. when making sure that the allocation profiling can be run at the same time as other kinds of profiling.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -77,7 +77,7 @@
 
 // Contains state for a single CpuAndWallTimeWorker instance
 struct cpu_and_wall_time_worker_state {
-  atomic_bool should_run;
+  // These are immutable after initialization
 
   bool gc_profiling_enabled;
   bool allocation_counting_enabled;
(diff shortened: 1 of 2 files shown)


ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -813,6 +834,29 @@ static void on_newobj_event(DDTRACE_UNUSED VALUE tracepoint_data, DDTRACE_UNUSED
...
+  struct cpu_and_wall_time_worker_state *state = active_sampler_instance_state; // Read from global variable, see "sampler global state safety" note above
+
+  // This should not happen in a normal situation because the tracepoint is always enabled after the instance is set
+  // and disabled before it is cleared, but just in case...
+  if (state == NULL) return;
+
+  // In a few cases, we may actually be allocating an object as part of profiler sampling. We don't want to recursively
+  // sample, so we just return early
+  if (state->during_sample) {
+    state->stats.allocations_during_sample++;
```
marcotc, line 847: Are we tracking it "just in case"?

ivoanjo: This is a good question. No, that wasn't my intention; I vaguely intended to use it for validation, but hadn't thought very far after that.

In fact, I initially implemented this on my working branch before I did the change in #2644, so I've been sitting on this for quite some time.

TL;DR, now that #2644 is in, I think rather than vague intentions, I can actually add specs asserting that common operations in the profiler actually don't cause allocations.

So, let me revisit this PR and add a few tests where I actually assert on this value. That seems like a nice low-hanging win that can already go in with this change.

ivoanjo: Added a spec that makes use of this in 1d165359fdc3b280a5f408e8f0cd87f4c161a9a7

PR #2692, opened 2023-03-14
Add remote configuration component

lloeki:
**What does this PR do?**

- Add a new component class for remote configuration
- Set up and tear down the component via `Datadog::Core::Configuration::Components`

**Motivation**

Remote configuration

**Additional Notes**

- The component is instantiated but not started until its `sync` method is called, which should be done on demand.
- Its creation is not yet controlled by configuration
- Its settings such as polling rate is not yet controlled by configuration

**How to test the change?**

```
  require 'ddtrace'

  Datadog.configure do |c|
    c.agent.host = '192.168.135.133'
    #c.tracing.enabled = true
    #c.diagnostics.debug = true
  end

  Datadog::Core::Remote.active_remote.sync
  repository = Datadog::Core::Remote.active_remote.client.instance_eval { @repository }

  loop do
    pp repository
    sleep 1
  end
```

### DIFF
--- a/lib/datadog/core/remote/component.rb
+++ b/lib/datadog/core/remote/component.rb
@@ -1,0 +1,42 @@
+# frozen_string_literal: true
+
+require_relative 'worker'
+require_relative 'client'
+require_relative '../transport/http'
+require_relative '../remote'
+
+module Datadog
+  module Core
+    module Remote
+      class Component
+        attr_reader :client
+
+        def initialize(settings, agent_settings)
+          transport_options = {}
+          transport_options[:agent_settings] = agent_settings if agent_settings
+
+          @transport_v7 = Datadog::Core::Transport::HTTP.v7(**transport_options.dup)
+
+          @client = Client.new(@transport_v7)
+          @worker = Worker.new(interval: 1) { @client.sync }
+        end
+
+        def sync
+          # TODO: start elsewere, block smartly. this way makes it start on demand for now
+          @worker.start
+        end
+
+        def shutdown!
+          @worker.stop
+        end
+
+        class << self
+          def build(settings, agent_settings)
+            # TODO: condition with configuration
+            new(settings, agent_settings)
+          end
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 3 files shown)


lloeki:
Rebased using `git rebase --onto add-remote-config-worker 33d6c8767^ add-remote-config-component`

lib/datadog/core/remote/component.rb
```diff
@@ -0,0 +1,42 @@
...
+        attr_reader :client
+
+        def initialize(settings, agent_settings)
+          transport_options = {}
+          transport_options[:agent_settings] = agent_settings if agent_settings
+
+          @transport_v7 = Datadog::Core::Transport::HTTP.v7(**transport_options.dup)
+
+          @client = Client.new(@transport_v7)
+          @worker = Worker.new(interval: 1) { @client.sync }
```
GustavoCaso, line 21: Will the interval be something that could be configured via configuration?

marcotc: At minimum, extract it to a constant locally, something like `DEFAULT_INTERVAL =  1`.

GustavoCaso (approved):

lib/datadog/core/remote/component.rb
ivoanjo: (Also, consider including the unit -- I'm guessing seconds?)

lib/datadog/core/remote/component.rb
```diff
@@ -0,0 +1,44 @@
...
+      # Configures the HTTP transport to communicate with the agent
+      # to fetch and sync the remote configuration
+      class Component
+        attr_reader :client
+
+        def initialize(settings, agent_settings)
+          transport_options = {}
+          transport_options[:agent_settings] = agent_settings if agent_settings
+
+          transport_v7 = Datadog::Core::Transport::HTTP.v7(**transport_options.dup)
```
ivoanjo, line 20: Consider passing in the actual `agent_settings` object to the transport, rather than "peeling it" -- this object exists already to contain all info that's needed to reach the agent.

Also consider not making `agent_settings` optional -- in the `components.rb` there's always an `AgentSettings`, so I'm not sure it's worth additionally supporting cases where there isn't one (I'm guessing this may follow some of the patterns from tracing, but I do think they just reflect legacy and not something we should keep on doing).

ivoanjo (commented):
Looks reasonable; left a few extra notes!

PR #2694, opened 2023-03-15
Use workaround for steep issue only in GitHub Actions

ivoanjo:
**What does this PR do?**:

This PR tweaks the workaround added in #2675 for the failing combination of steep 1.3.1 with Ruby 3.0 and 3.1 on GitHub Actions (GHA).

In the prior PR, I just excluded `steep` from all macOS configurations running Ruby 3.0 and 3.1, which was a lot more than was needed. In this PR I tweak this to only exclude `steep` when running on GHA, since we have not seen this issue elsewhere.

**Motivation**:

The previous workaround broke the local development setup for one of our colleagues :(

**Additional Notes**:

I also tried bumping to the latest version of the `ruby/setup-ruby` prebuilt action, but that didn't help (but it doesn't hurt either, so I left it in).

**How to test the change?**:

Validate that the "test macOS" GitHub Action, along with all others, are green. Furthermore, validate that `steep` is installed by default for development on macOS.

### DIFF
--- a/.github/workflows/test-macos.yaml
+++ b/.github/workflows/test-macos.yaml
@@ -31,7 +31,7 @@ jobs:
       # head rubies
       - if: ${{ matrix.ruby == 'head' }}
         run: sed -i~ -e '/spec\.required_ruby_version/d' ddtrace.gemspec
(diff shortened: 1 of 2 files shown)


lloeki (approved):
LGTM, thanks!

PR #2695, opened 2023-03-15
Fix Rails < 3 conditional check in Utils#railtie_supported?

nightpool:
**What does this PR do?**
The `Rails::VERSION::MAJOR >= 3` condition evalutes to `false`, but `false` is not `nil?`, so railtie_supported? was returning true, causing the code to attempt to load the railties on Rails version prior to 3. To fix this, I've removed the explicit `.nil?` check and replaced it with a boolean coercion (which will return false for both `nil` and `false`, rather then returning false for `nil` and true for `false`.

**Motivation**
We're integrating ddtrace into an application that uses Rails 2 in production while also supporting Rails 3 in development. 

**Additional Notes**
This change lets the gem at least boot fully enough to instrument some of our ancillary gems (postgres, redis) and run the continuous profiler, which was our biggest goal. There may be more PRs in the future if desired as we enable more Rails 2 functionality, but given the level of customization involved in our Rails 2 setup, I can't necessarily guarantee it.

**How to test the change?**

My server boots ^.^ 

Probably the easiest way to test this is to just mock out the Rails 2 constants in the console to confirm to yourself that the old logic doesn't work and the new logic does work.

### DIFF
--- a/spec/datadog/tracing/contrib/rails/utils_spec.rb
+++ b/spec/datadog/tracing/contrib/rails/utils_spec.rb
@@ -5,10 +5,6 @@ RSpec.describe Datadog::Tracing::Contrib::Rails::Utils do
   describe 'railtie_supported?' do
     subject(:railtie_supported?) { described_class.railtie_supported? }
 
-    before do
-      stub_const('::Rails::Railtie', instance_double('::Rails::Railtie'))
-    end
-
     context 'on rails 2 and below' do
       before { stub_const('::Rails::VERSION::MAJOR', 2) }
 
@@ -18,6 +14,7 @@ RSpec.describe Datadog::Tracing::Contrib::Rails::Utils do
     context 'on rails 3 and above' do
       before do
         stub_const('::Rails::VERSION::MAJOR', 3)
+        stub_const('::Rails::Railtie', instance_double('::Rails::Railtie'))
       end
 
       it { is_expected.to be true }
--- a/spec/datadog/tracing/contrib/rails/utils_spec.rb
+++ b/spec/datadog/tracing/contrib/rails/utils_spec.rb
@@ -1,0 +1,26 @@
+require 'lib/datadog/tracing/contrib/rails/utils'
+require 'rails/version'
+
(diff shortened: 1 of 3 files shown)


spec/datadog/tracing/contrib/rails/utils_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'lib/datadog/tracing/contrib/rails/utils'
+require 'rails/version'
+
+RSpec.describe Datadog::Tracing::Contrib::Rails::Utils do
+  describe 'railtie_supported?' do
+    subject(:railtie_supported?) { described_class.railtie_supported? }
+
+    before do
+      stub_const('::Rails::Railtie', instance_double('::Rails::Railtie'))
+    end
```
nightpool, line 10: This constant isn't defined on Rails 2, so it should be moved to the Rails 3 block

spec/datadog/tracing/contrib/rails/utils_spec.rb
nightpool, line 1: It seems like the main place this code was previously used was as a "Is Rails installed at all?" check so it might be good to add a test with no constants defined.

nightpool (commented):
Thanks for adding a test! Left a few comments

ivoanjo (approved):
Rails 2! Ouch... I feel your pain and I wish you good luck on your endeavor.

I've pushed a quick spec and will merge as soon as CI is green.

Finally, let us know if you run into any issues with profiling! And consider trying out [CPU Profiling 2.0](https://github.com/DataDog/dd-trace-rb/releases/tag/v1.10.0) -- it will give you even finer-grained data.

nightpool:
Thanks! The profiling experience has been great so far. But I'd love the ability to break down memory allocations like I can with the stackprof or memory_profiler gem! That would be a huge gamechanger for us in production.

ivoanjo:
> Thanks! The profiling experience has been great so far. But I'd love the ability to break down memory allocations like I can with the stackprof or memory_profiler gem! That would be a huge gamechanger for us in production.

Cool! I'm working on this, indeed looking forward to having it as well

PR #2697, opened 2023-03-17
[PROF-6555] Decouple profiling component initialization from core

ivoanjo:
**What does this PR do?**:

Similar to the approach taken by the AppSec component, this PR changes the Profiling component initialization to be self method on `Datadog::Profiling::Component` that is completely separate from the core components.

This also included moving over the specs for this component to a separate file.

**Motivation**:

Finish the migration started in #2561 when the Profiling component was moved away from the core.

**Additional Notes**:

I've also sprinkled a few documentation improvements throughout.

**How to test the change?**:

This change is covered by the included and existing specs.


GustavoCaso:
Should we update the RBS files 😄 ?

GustavoCaso:
Are we planning on moving the `profiling` settings into the profile package, similar to the appsec and CI? It doesn't have to be done on this PR. Just asking 😄

ivoanjo:
>  Should we update the RBS files :smile: ?

Yes -- let me take a stab at enabling type checking for them!

> Are we planning on moving the `profiling` settings into the profile package, similar to the appsec and CI? It doesn't have to be done on this PR. Just asking 😄 

No plans to do this change until we make a decision on what exactly the "next-gen" settings are going to look like.

marcotc (approved):

ivoanjo:
I've opened a separate PR adding type signatures to `Datadog::Profiling::Component` and a bunch of other profiling classes (#2700), so I'm going to go ahead and merge this one as-is.

PR #2698, opened 2023-03-17
[PROF-6555] Support disabling endpoint names collection in new profiler

ivoanjo:
**What does this PR do?**:

This PR updates the new CPU Profiling 2.0 profiler to honor the existing `profiling.advanced.endpoint.collection.enabled` setting.

This setting defaults to enabled. When disabled, the profiler will omit endpoint names in the collected data.

**Motivation**:

Not supporting this was an oversight; the new profiler (still in beta) should honor this setting.

**Additional Notes**:

I'm opening this PR on top of #2697 to avoid conflicts, but both PRs are conceptually independent.

**How to test the change?**:

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
@@ -97,6 +97,8 @@ struct thread_context_collector_state {
   unsigned int sample_count;
   // Reusable array to get list of threads
   VALUE thread_list_buffer;
+  // Used to omit endpoint names (retrieved from tracer) from collected data
+  bool endpoint_collection_enabled;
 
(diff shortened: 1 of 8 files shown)


ext/ddtrace_profiling_native_extension/collectors_thread_context.c
```diff
@@ -911,6 +932,8 @@ static void trace_identifiers_for(struct thread_context_collector_state *state,
 
   trace_identifiers_result->valid = true;
 
+  if (!state->endpoint_collection_enabled) return;
```
ivoanjo, line 935: All this setup and passing the setting around just to do this ;)

marcotc (approved):

PR #2700, opened 2023-03-17
Add type signatures for a bunch of profiling classes

ivoanjo:
**What does this PR do?**:

This PR adds type signatures and enables type checking for a number of profiling classes (see `Steepfile` for list of files that are no longer ignored).

**Motivation**:

This was discussed/requested in #2697.

**Additional Notes**:

I wish the steep errors were a bit more user-friendly.

The `StackRecorder#clear` method actually should not exist anymore, and would actually break because it called `_native_clear` which was deleted in #2367. (Ooops)

Other than that, there's a few minor changes to code to make code more type-checkable, but nothing of notice.

**How to test the change?**:

Validate that CI is still green and typechecking passes.


sig/datadog/profiling/stack_recorder.rbs
```diff
@@ -0,0 +1,27 @@
...
+
+      def initialize: (cpu_time_enabled: bool, alloc_samples_enabled: bool) -> void
+
+      def self._native_initialize: (
+        Datadog::Profiling::StackRecorder recorder_instance,
+        bool cpu_time_enabled,
+        bool alloc_samples_enabled,
+      ) -> true
+
+      def serialize: () -> untyped
```
lloeki, line 14: Looking at `serialize!`, what can this return except for a `String`? Is that tied to the unhappy path?

sig/datadog/profiling/scheduler.rbs
```diff
@@ -0,0 +1,19 @@
+module Datadog
+  module Profiling
+    class Scheduler < Core::Worker
+      include Core::Workers::Polling
+
+      def initialize: (
+        exporter: Datadog::Profiling::Exporter,
+        transport: Datadog::Profiling::HttpTransport,
+        ?fork_policy: untyped,
```
lloeki, line 9: Sometimes for `untyped` things I create a type alias, e.g:

```
type fork_policy = untyped

def initialize:
  ...
  ?fork_policy: fork_policy
```

Allowing me to (and Steep) to differentiate between a blanket `uptyped` and two "different-but-known" kinds of `untyped`, and then if/when I get to actually type it, I can replace it in one place.

WDYT?

sig/datadog/profiling/component.rbs
```diff
@@ -0,0 +1,33 @@
+module Datadog
+  module Profiling
+    module Component
+      def self.build_profiler_component: (
+        settings: untyped,
```
lloeki, line 5: IIRC this one may be typable as `Core::Configuration::Settings`

lloeki (approved):
Very cool to see some moves on that front! Left a few minor notes, feel free to address if you feel it makes sense for this PR.

sig/datadog/profiling/stack_recorder.rbs
ivoanjo: The signature of this method should be either a `[::Time, ::Time, ::String]` or a `nil`. But when I used `[::Time, ::Time, ::String]?`, steep started complaining that `exporter.rb` that uses this output (but checks if it is `nil` first) may have stuff as `nil`.

I'm not entirely sure what's wrong (and I'm guessing I could introduce a `Struct` to contain stuff instead of an array), but since I was timeboxing this effort, I left this for a later attempt :)

sig/datadog/profiling/stack_recorder.rbs
```diff
@@ -0,0 +1,27 @@
...
+
+      def self._native_serialize: (Datadog::Profiling::StackRecorder recorder_instance) -> [:ok | :error, untyped]
```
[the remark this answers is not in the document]
ivoanjo: Steep seems to be confused by this one. The type is 

`([:ok, [::Time, ::Time, ::String]] | [:error, ::String])` 

e.g. either you have an `:ok` and an array with the output, or an `:error` with an error string.

But steep doesn't like the way I use it in `#serialize`:

```
lib/datadog/profiling/stack_recorder.rb:25:67: [error] Type `([::Time, ::Time, ::String] | ::String)` does not have method `iso8601`
│
└           Datadog.logger.debug { "Encoded profile covering #{start.iso8601} to #{finish.iso8601}" }
                                                                     ~~~~~~~

lib/datadog/profiling/stack_recorder.rb:25:88: [error] Type `nil` does not have method `iso8601`
│
└           Datadog.logger.debug { "Encoded profile covering #{start.iso8601} to #{finish.iso8601}" }                                                             

```

As I said for `serialize!` I could probably introduce a `Struct` here to contain the Time+Time+String and make steep happy but I didn't have time yet :)

PR #2703, opened 2023-03-20
[DOCS-5086] Update link text in Ruby tracing setup

kayayarai:
**What does this PR do?**
Docs only - updates link text in getting started

**Motivation**
DOCS-5086

**Additional Notes**

**How to test the change?**

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -9,7 +9,7 @@ databases and microservices so that developers have high visibility into bottlen
 
 For the general APM documentation, see our [setup documentation][setup docs].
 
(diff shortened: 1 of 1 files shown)


ivoanjo (approved):
The change itself makes sense, since we're redirecting to the new url anyway.

---

As an addendum to be honest I kinda got confused by the docs page itself:

image

When I arrived on it, the page seems like it will have a big list of terms, and customers may miss that actually if you scroll down below there's a lot of screenshots, videos, animations, etc.

(I actually searched around other pages before coming back to this page and realizing this)

I wonder if would make sense to start the page on a lighter note, rather than on a huge scary table that hints at a very lengthy and not very visual page?

PR #2710, opened 2023-03-21
Allow `1` as true value in environment variables

marcotc:
We currently only support the string `true` as a truthy boolean configuration value (case insensitive).

Our standard for APM Tracing configuration is actually to allow either `true` or `1`.

This PR changes the logic to accept `1` as a valid true value for environment variable configuration.

There are no other functional changes to configuration parsing.

### DIFF
--- a/spec/datadog/core/environment/variable_helpers_spec.rb
+++ b/spec/datadog/core/environment/variable_helpers_spec.rb
@@ -72,9 +72,10 @@ RSpec.describe Datadog::Core::Environment::VariableHelpers do
       %w[
         true
         TRUE
+        1
       ].each do |value|
         context value.to_s do
-          let(:env_value) { value.to_s }
+          let(:env_value) { value }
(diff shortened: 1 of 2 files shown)


spec/datadog/core/environment/variable_helpers_spec.rb
```diff
@@ -85,11 +86,11 @@
         '',
         'false',
         'FALSE',
-        0,
-        1
+        '0',
```
TonyCTHsu, line 89: Can we add a couple of example with leading and trailing spaces? such as `'   1'`  and `'1   '`

TonyCTHsu (approved):

spec/datadog/core/environment/variable_helpers_spec.rb
marcotc: Certainly!

PR #2711, opened 2023-03-22
Gauge primitive RubyVM::YJIT.runtime_stats, if YJIT is enabled

HeyNonster:
**What does this PR do?**

Adds YJIT runtime stats to the Ruby Runtime Metrics.

**Motivation**
YJIT was [deemed production ready in Ruby 3.2](https://www.ruby-lang.org/en/news/2022/12/25/ruby-3-2-0-released/). YJIT includes its own [runtime stats](https://github.com/ruby/ruby/blob/master/doc/yjit/yjit.md#other-statistics), which would be useful to gauge. 

Of particular use is the `OBJECT_SHAPE_COUNT` metric. Ruby has a ceiling on the total number of unique shapes but we currently don't have a convenient way of keeping track of those figures to determine if optimizations need to be made.

**Additional Notes**

The comments in `lib/datadog/core/environment/yjit.rb` I mostly copied directly from the [YJIT source code](https://github.com/ruby/ruby/blob/2759cf46f28b1f5ba52f6b7686fd23cb8d792ede/yjit/src/stats.rs#L453).

#### Primitive Stats vs Extended Stats
This PR tracks YJIT stats when running Ruby with the `--yjit` flag, but currently only the primitive stats (not the additional stats that are available when the `--yjit-stats` flag is also set) are gauged.

It's limited to the primitive stats because YJIT, by default, has the following runtime stats:

```
$ RUBYOPT='--yjit' ruby -e 'pp RubyVM::YJIT.runtime_stats' =>
{
  :inline_code_size=>0,
  :outlined_code_size=>116,
  :freed_page_count=>0,
  :freed_code_size=>0,
  :live_page_count=>1,
  :code_gc_count=>0,
  :code_region_size=>16384,
  :object_shape_count=>236
}
```

However, with the `--yjit-stats` flag, it returns 325 separate stats:

```
$ RUBYOPT="--yjit --yjit-stats" ruby -e 'pp RubyVM::YJIT.runtime_stats.keys.length' => 325
```

And not all of them are numeric:

```
$ RUBYOPT="--yjit --yjit-stats" ruby -e 'pp RubyVM::YJIT.runtime_stats.values.select { |value| !value.kind_of?(Numeric) }.any?' => true
```

**How to test the change?**

```
# Using Ruby 3.2+
RUBYOPT="--yjit" bundle exec rake spec:yjit

bundle exec rake spec:yjit # shows the YJIT specs get skipped
```

### DIFF
--- a/lib/datadog/core/runtime/metrics.rb
+++ b/lib/datadog/core/runtime/metrics.rb
@@ -5,6 +5,7 @@ require_relative '../environment/class_count'
 require_relative '../environment/gc'
 require_relative '../environment/thread_count'
 require_relative '../environment/vm_cache'
+require_relative '../environment/yjit'
 
 module Datadog
   module Core
@@ -78,6 +79,8 @@ module Datadog
               )
             end
           end
+
+          flush_yjit_stats
         end
 
         def gc_metrics
@@ -133,6 +136,46 @@ module Datadog
 
         def gauge_if_not_nil(metric_name, metric_value)
           gauge(metric_name, metric_value) if metric_value
+        end
+
+        def flush_yjit_stats
+          # Only on Ruby >= 3.2
+          try_flush do
+            if Core::Environment::YJIT.available?
+              gauge_if_not_nil(
+                Core::Runtime::Ext::Metrics::METRIC_YJIT_CODE_GC_COUNT,
+                Core::Environment::YJIT.code_gc_count
+              )
+              gauge_if_not_nil(
+                Core::Runtime::Ext::Metrics::METRIC_YJIT_CODE_REGION_SIZE,
+                Core::Environment::YJIT.code_region_size
+              )
+              gauge_if_not_nil(
+                Core::Runtime::Ext::Metrics::METRIC_YJIT_FREED_CODE_SIZE,
+                Core::Environment::YJIT.freed_code_size
+              )
+              gauge_if_not_nil(
+                Core::Runtime::Ext::Metrics::METRIC_YJIT_FREED_PAGE_COUNT,
+                Core::Environment::YJIT.freed_page_count
+              )
+              gauge_if_not_nil(
+                Core::Runtime::Ext::Metrics::METRIC_YJIT_INLINE_CODE_SIZE,
+                Core::Environment::YJIT.inline_code_size
+              )
+              gauge_if_not_nil(
+                Core::Runtime::Ext::Metrics::METRIC_YJIT_LIVE_PAGE_COUNT,
+                Core::Environment::YJIT.live_page_count
+              )
+              gauge_if_not_nil(
+                Core::Runtime::Ext::Metrics::METRIC_YJIT_OBJECT_SHAPE_COUNT,
+                Core::Environment::YJIT.object_shape_count
+              )
+              gauge_if_not_nil(
+                Core::Runtime::Ext::Metrics::METRIC_YJIT_OUTLINED_CODE_SIZE,
+                Core::Environment::YJIT.outlined_code_size
+              )
+            end
+          end
         end
       end
     end
--- a/.github/workflows/test-yjit.yaml
+++ b/.github/workflows/test-yjit.yaml
@@ -1,0 +1,32 @@
+name: Test YJIT
+on: [push]
+jobs:
+  test-yjit:
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+        ruby:
+          - '3.2'
+          # ADD NEW RUBIES HERE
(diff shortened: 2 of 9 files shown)


lib/datadog/core/runtime/metrics.rb
```diff
@@ -78,6 +79,9 @@ def flush
               )
             end
           end
+
+          # Only on Ruby >= 3.2
+          try_flush { yjit_metrics.each { |metric, value| gauge(metric, value) } if Core::Environment::YJIT.available? }
```
marcotc, line 84: I think this explicitly declare all the values that it reports, which today would be:
```ruby
[:inline_code_size,
 :outlined_code_size,
 :freed_page_count,
 :freed_code_size,
 :live_page_count,
 :code_gc_count,
 :code_region_size,
 :object_shape_count]
```

marcotc (changes requested):
This is a great contribution, thank you!

For background information, for each runtime metric, we have to provide a human readable title and a short description, as well as choosing if they belong to a the same graph or different graphs:

Screenshot 2023-03-22 at 2 30 11 PM

For example, the "Slots" graph above contains all these 4 metrics:
```
runtime.ruby.gc.heap_available_slots
runtime.ruby.gc.heap_live_slots
runtime.ruby.gc.heap_marked_slots
runtime.ruby.gc.heap_free_slots
```

This presentation work is done by us internally after we merge the Ruby Tracer changes.

In order to provide this information in the UI, we have to know what runtime metrics we are collecting, and be able to articulate what they are and if they should be directly correlated with other runtime metrics.

For this PR, I suggest declaring explicitly the new runtime metrics that are being created, instead of iterating over the results of `RubyVM::YJIT.runtime_stats`.

lib/datadog/core/runtime/metrics.rb
HeyNonster: I've changed the iteration to explicit methods. The comments for each method I lifted directly from the Y[JIT source code](https://github.com/ruby/ruby/blob/2759cf46f28b1f5ba52f6b7686fd23cb8d792ede/yjit/src/stats.rs#L453).

lib/datadog/core/runtime/metrics.rb
```diff
@@ -78,6 +79,45 @@ def flush
               )
             end
           end
+
+          # Only on Ruby >= 3.2
+          try_flush do
+            if Core::Environment::YJIT.available?
```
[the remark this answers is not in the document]
HeyNonster: @GustavoCaso  I removed line 120 because it was mistakenly left over from a previous attempt at this.

I opted not to move the conditional check inline with the `end` to keep the style consistent with the rest of this file e.g. lines 57-81:

```ruby
          try_flush do
            if Core::Environment::VMCache.available?
              # Only on Ruby < 3.2
              gauge_if_not_nil(
                Core::Runtime::Ext::Metrics::METRIC_GLOBAL_CONSTANT_STATE,
                Core::Environment::VMCache.global_constant_state
              )
              ...
```

GustavoCaso:
@HeyNonster 

Thanks for the changes. I noticed that `rubocop` is failing. I usually run `bundle exec rubocop -a` locally to try fixing those. Sometimes it has to be fixed manually since the tool doesn't know how to fix it.

Feel free to let me know if you need any help with that.

Another thing, if you don't mind rebasing master into your branch? That would fix all the broken CI steps 🔨 

cc @marcotc if you want to have one last look before merging

HeyNonster:
@gustavo

I've rebased onto the latest master! :)

As for Rubocop, the one complaint was that the `#flush` method is now too long. I'm not sue how y'all would prefer to handle that. As one possible solution, I pushed a `fixup!` commit that moves the YJIT stats flushing into a separate `#dump_yjit_stats` private method that `#flush` calls.

If you like that approach, I can squash the `fixup!` commit; if not, I can drop it and try something else 😄

GustavoCaso:
@HeyNonster I like the approach from the last commit 😄 

Also, sorry for the delay with the reply. 

@marcotc are we good to go and merge it?

HeyNonster:
> @HeyNonster I like the approach from the last commit 😄
> 
> Also, sorry for the delay with the reply.
> 
> @marcotc are we good to go and merge it?

Thanks, @GustavoCaso , I'll keep an eye on this and squash the fixup when y'all give me the final greenlight! 🙂

marcotc (approved):
Thank you so much, @HeyNonster! 🙇

HeyNonster:
> Thank you so much, @HeyNonster! 🙇

Thank you, @marcotc and @GustavoCaso! I've squashed the `fixup!` and rebased the branch onto main. :)

PR #2712, opened 2023-03-22
Telemetry:Remove nil values from payload

marcotc:
Our V1 telemetry implementation returns `nil` values in key/value mappings, when telemetry dictates that `nil` values should be omitted from the flushed payload.

This behaviour is enforced by system-tests, which started failing recently:
```
tests/test_schemas.py::Test_Library::test_full - Exception: Schema is invalid in logs/interfaces/library/000__...
tests/test_schemas.py::Test_Library::test_non_regression - Exception: Schema is invalid in logs/interfaces/lib...
tests/test_schemas.py::Test_Agent::test_full - Exception: Schema is invalid in logs/interfaces/agent/003__api_...
tests/test_schemas.py::Test_Agent::test_non_regression - Exception: Schema is invalid in logs/interfaces/agent...
```

This PR addresses these system-tests failures by removing empty values from the returned payload.

Testing compliance for the full Telemetry schema is the responsibility of the system-tests repository: there's a strict JSON Schema validator in that repository. The tests present in this PR are simple unit tests.

### DIFF
--- a/Steepfile
+++ b/Steepfile
@@ -118,6 +118,7 @@ target :ddtrace do
   ignore 'lib/datadog/core/utils.rb'
   ignore 'lib/datadog/core/utils/compression.rb'
   ignore 'lib/datadog/core/utils/forking.rb'
+  ignore 'lib/datadog/core/utils/hash.rb' # Refinement module
   ignore 'lib/datadog/core/utils/network.rb'
   ignore 'lib/datadog/core/utils/object_set.rb'
   ignore 'lib/datadog/core/utils/only_once.rb'
--- a/lib/datadog/core/telemetry/collector.rb
+++ b/lib/datadog/core/telemetry/collector.rb
@@ -5,6 +5,7 @@ require 'etc'
 require_relative '../configuration/agent_settings_resolver'
 require_relative '../environment/ext'
 require_relative '../environment/platform'
+require_relative '../utils/hash'
 require_relative 'v1/application'
 require_relative 'v1/dependency'
 require_relative 'v1/host'
@@ -18,6 +19,7 @@ module Datadog
       # Module defining methods for collecting metadata for telemetry
       module Collector
         include Datadog::Core::Configuration
+        using Core::Utils::Hash::Refinement
 
         # Forms a hash of configuration key value pairs to be sent in the additional payload
         def additional_payload
@@ -41,12 +43,13 @@ module Datadog
 
         # Forms a hash of standard key value pairs to be sent in the app-started event configuration
         def configurations
-          configurations = {
+          hash = {
             DD_AGENT_HOST: Datadog.configuration.agent.host,
             DD_AGENT_TRANSPORT: agent_transport,
             DD_TRACE_SAMPLE_RATE: format_configuration_value(Datadog.configuration.tracing.sampling.default_rate),
           }
-          compact_hash(configurations)
+          hash.compact!
+          hash
         end
 
         # Forms a telemetry app-started dependencies object
@@ -134,7 +137,8 @@ module Datadog
           options['logger.instance'] = configuration.logger.instance.class.to_s
           options['appsec.enabled'] = configuration.dig('appsec', 'enabled') if configuration.respond_to?('appsec')
           options['tracing.opentelemetry.enabled'] = !defined?(Datadog::OpenTelemetry::LOADED).nil?
-          compact_hash(options)
+          options.compact!
+          options
         end
 
         def format_configuration_value(value)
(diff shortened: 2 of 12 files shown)


GustavoCaso:
@marcotc Do we need to use refinement for this? 

My general knowledge is that refinement as well as monkey patching, is not recommended. 

Since the code changes `require_relative` per file that want to remove nil values from the hash, couldn't we have an utility module that would remove the nil values and make sure to call that at the return? 


Something like:

```ruby
require_relative '../../utils/hash'

module Datadog
  module Core
    module Telemetry
      module V1
        class Application
          .
          .
          .
          def to_h
            Core::Utils::Hash.remove_nil_values({
              env: @env,
              language_name: @language_name,
              language_version: @language_version,
              products: @products.to_h,
              runtime_name: @runtime_name,
              runtime_patches: @runtime_patches,
              runtime_version: @runtime_version,
              service_name: @service_name,
              service_version: @service_version,
              tracer_version: @tracer_version
            })
          end
        end
      end
    end
  end
end
```

GustavoCaso:
Also, since we are adding a new file, we should not ignore it from the point of view of Steep, but rather add the necessary rbs signatures.

ivoanjo (approved):
:+1 LGTM. I'm also not the biggest fan of using refinements, but I think I can live with this one.

> Also, since we are adding a new file, we should not ignore it from the point of view of Steep, but rather add the necessary rbs signatures.

I also think it'd be nice to increase the type checking coverage, but I don't consider it blocking.

marcotc:
@ivoanjo @GustavoCaso I did try to add the rbs file, but as far as my research goes I can't do it for this refinement. I was excited to add my first rbs! I tried to ignore only the module `Datadog::Core::Utils::Hash::Refinement` inside the `hash.rb` file, but also couldn't find a way.

marcotc:
@GustavoCaso refinements solve a very annoying problem we have: we can't write modern Ruby because we support such old versions of Ruby.
But we do drop old versions as time goes on, so one day we will be able to do `array.to_set` without any workarounds.
Until that day comes we have to satisfy the early 2010s gods.

To make our code easier to transition and easier to a modern rubyist to read, it would be great if future methods could be called today.
Turns out that's not an original idea: https://github.com/marcandre/backports. This library does that for you, but with on caveat: it hard-monkey-patches your classes, so for example `Hash` is forever modified in the host application when you load an array backport.

What we want is for `Hash` to have nice modern features only inside `ddtrace`, but not leak to the host application.

[Refinements](https://docs.ruby-lang.org/en/2.1.0/syntax/refinements_rdoc.html) we created exactly for that: "Refinements are designed to reduce the impact of monkey patching on other users of the monkey-patched class. Refinements provide a way to extend a class locally.".

Also, they don't add performance overhead:
```ruby
require 'benchmark/ips'

class Native
  def self.compact(h)
    h.compact
  end
end

class Refinement
  module HashRefinement
    refine Hash do
      def compact
        reject { |_k, v| v.nil? }
      end
    end unless {}.respond_to?(:compact)
  end

  using HashRefinement

  def self.compact(h)
    h.compact
  end
end

class Method
  def self.compact(h)
    compact_util(h)
  end

  def self.compact_util(hash)
    hash.reject { |_k, v| v.nil? }
  end

  def self.compact_optimized(hash)
    hash.respond_to?(:compact) ? hash.compact : hash.reject { |_k, v| v.nil? }
  end

  if {}.respond_to?(:compact)
    def self.compact_optimized_2(hash)
      hash.compact
    end
  else
    def self.compact_optimized_2(hash)
      hash.reject { |_k, v| v.nil? }
    end
  end
end

Benchmark.ips do |x|
  h = {a:1, b:nil, c:nil, d:2}

  x.time = 5
  x.warmup = 1

  x.report('native') { Native.compact(h) } if {}.respond_to?(:compact)
  x.report('refinement') { Refinement.compact(h) }
  x.report('method') { Method.compact(h) }
  x.report('method optimized') { Method.compact_optimized(h) }
  x.report('method optimized 2') { Method.compact_optimized_2(h) }

  x.compare!
end
```
```
# Ruby 3.1
# Comparison:
#               native:  5071303.2 i/s
#   method optimized 2:  5011656.3 i/s - same-ish: difference falls within error
#           refinement:  4991040.0 i/s - same-ish: difference falls within error
#     method optimized:  4107541.7 i/s - 1.23x  slower
#               method:  2644705.9 i/s - 1.92x  slower

# Ruby 2.1
# Comparison:
#   method optimized 2:  1488325.7 i/s
#           refinement:  1445207.0 i/s - same-ish: difference falls within error
#               method:  1436844.1 i/s - same-ish: difference falls within error
#     method optimized:  1331451.5 i/s - 1.12x  slower
```

TonyCTHsu (approved):

PR #2713, opened 2023-03-23
Setup python 3.9 for system tests

cbeauchesne:
**What does this PR do?**

This PR installs python 3.9 for system tests jobs.

**Motivation**

System test executor will run on the host rather than inside a container (see https://github.com/DataDog/system-tests/pull/958). It will allow users to do step-by-step debugging, and will allow to unify architecture with parametric tests.


ivoanjo (approved):
:+1: Seems reasonable!

PR #2714, opened 2023-03-23
Glue remote config together

lloeki:
**What does this PR do?**

Ties all the things together to Make It Work.

See commit messages for the details.

**Motivation**

Remote configuration

**Additional Notes**

- This mostly handles only the happy path of things (e.g agent with endpoint, no agent disappearing, etc...)
- There is AppSec stuff in Core still. This will be moved away shortly.

**How to test the change?**

Interactively:

```
docker run --rm -it -e DD_APM_ENABLED=true -e DD_ENV=${your_env} -e DD_HOSTNAME=lloeki.cassini -e DD_BIND_HOST=0.0.0.0 -e DD_APM_NON_LOCAL_TRAFFIC=true -e DD_API_KEY=${your_api_key} -e DD_REMOTE_CONFIGURATION_ENABLED=true -p 8126:8126 datadog/agent:7.42.0
```

Then:

```
env DD_AGENT_HOST=${agent_ip} DD_APPSEC_ENABLED=true DD_REMOTE_CONFIGURATION_ENABLED=true DD_REMOTE_CONFIGURATION_POLL_INTERVAL_SECONDS=10 DD_DIAGNOSTICS=true bundle exec puma -w 0 -p 9292
```

And:

```
# Note that the first request is the one that kicks off the worker, so it is not covered by the RC update.

# should block: rules override
curl -vvv -H 'User-Agent: Arachni/v2' 'http://user:pass@127.0.0.1:9292/?q=1%20OR%201;'
# should not block: exclusion filter
curl -vvv -H 'User-Agent: Nessus SOAP' 'http://user:pass@127.0.0.1:9292/?q=1%20OR%201;'

# these could block or not block depending on exclusion filters and denylist item expiration
curl -vvv -H 'X-Cluster-Client-IP: 80.80.80.80' 'http://user:pass@127.0.0.1:9292/'
curl -vvv -H 'X-Cluster-Client-IP: 1.0.0.1' 'http://user:pass@127.0.0.1:9292/'
curl -vvv -H 'X-Cluster-Client-IP: 1.2.3.4' 'http://user:pass@127.0.0.1:9292/'
```

Toggle stuff in the UI and see things being updated live in debug logs.

### DIFF
--- a/lib/datadog/appsec/component.rb
+++ b/lib/datadog/appsec/component.rb
@@ -28,12 +28,28 @@ module Datadog
 
       def initialize(processor:)
         @processor = processor
+        @mutex = Mutex.new
+      end
+
+      def reconfigure(ruleset:)
+        @mutex.synchronize do
+          old = @processor
+          @processor = Processor.new(ruleset: ruleset)
+
+          old.finalize if processor && processor.ready?
+        end
+      end
+
+      def reconfigure_lock(&block)
+        @mutex.synchronize(&block)
       end
 
       def shutdown!
-        if processor && processor.ready?
-          processor.finalize
-          @processor = nil
+        @mutex.synchronize do
+          if processor && processor.ready?
+            processor.finalize
+            @processor = nil
+          end
         end
       end
     end
--- a/lib/datadog/appsec/component.rb
+++ b/lib/datadog/appsec/component.rb
@@ -33,10 +33,13 @@ module Datadog
 
       def reconfigure(ruleset:)
         @mutex.synchronize do
-          old = @processor
-          @processor = Processor.new(ruleset: ruleset)
+          new = Processor.new(ruleset: ruleset)
 
-          old.finalize if processor && processor.ready?
+          if new && new.ready?
+            old = @processor
+            @processor = new
+            old.finalize
+          end
         end
       end
 
--- a/lib/datadog/core/remote/client.rb
+++ b/lib/datadog/core/remote/client.rb
@@ -191,6 +191,24 @@ module Datadog
           cap_to_hexs.each_with_object([]) { |hex, acc| acc << hex }.map { |e| e.to_i(16) }.pack('C*')
         end
 
+        def select_content(repository, product, config_ids = [])
+          repository.contents.select do |content|
+            content.path.product == 'ASM_DATA' && (config_ids.empty? || config_ids.include?(content.path.config_id))
+          end
+        end
+
+        def parse_content(content)
+          data = content.data.read
+
+          content.data.rewind
+
+          raise ReadError, 'EOF reached' if data.nil?
+
+          JSON.parse(data)
+        end
+
+        class ReadError < StandardError; end
+
         def register_receivers
           matcher = Dispatcher::Matcher::Product.new(products)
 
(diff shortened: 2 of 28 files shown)


lib/datadog/core/remote/client.rb
```diff
@@ -27,7 +28,11 @@ def sync
 
           if response.ok?
             # when response is completely empty, do nothing as in: leave as is
-            return if response.empty?
+            if response.empty?
+              Datadog.logger.debug { 'remote: empty response => NOOP' }
```
GustavoCaso, line 32: Do we need to include the `=> NOOP` part?

lib/datadog/core/remote/client.rb
```diff
@@ -185,8 +194,35 @@ def capabilities_binary
...
+            overrides = repository.contents.select do |content|
+              content.path.product == 'ASM' && ['blocking', 'disabled_rules'].include?(content.path.config_id)
+            end.map! { |content| JSON.parse(content.data.read.tap { content.data.rewind }) }
+
+            exclusions = repository.contents.select do |content|
+              content.path.product == 'ASM' && content.path.config_id == 'exclusion_filters'
+            end.map! { |content| JSON.parse(content.data.read.tap { content.data.rewind }) }
+
+            ruleset = AppSec::Processor::RuleMerger.merge(
+              rules: rules,
```
GustavoCaso, line 219: Is. Is it a warranty that rules will always be populated here? 

If the remote config does not send the `ASM_DD` product information, the `RuleMerger` might break.

lib/datadog/appsec/component.rb
```diff
@@ -28,12 +28,28 @@ def create_processor
...
         @processor = processor
+        @mutex = Mutex.new
+      end
+
+      def reconfigure(ruleset:)
+        @mutex.synchronize do
+          old = @processor
+          @processor = Processor.new(ruleset: ruleset)
+
+          old.finalize if processor && processor.ready?
```
GustavoCaso, line 39: Should we handle the case that the new processor is not ready and re-assign the old processor as the main processor?

lib/datadog/core/remote/client.rb
```diff
@@ -185,8 +194,35 @@ def capabilities_binary
...
+              content.path.product == 'ASM_DATA' && ['blocked_ips', 'blocked_users' ].include?(content.path.config_id)
+            end.map! { |content| JSON.parse(content.data.read.tap { content.data.rewind }) }
+
+            overrides = repository.contents.select do |content|
+              content.path.product == 'ASM' && ['blocking', 'disabled_rules'].include?(content.path.config_id)
+            end.map! { |content| JSON.parse(content.data.read.tap { content.data.rewind }) }
+
+            exclusions = repository.contents.select do |content|
+              content.path.product == 'ASM' && content.path.config_id == 'exclusion_filters'
+            end.map! { |content| JSON.parse(content.data.read.tap { content.data.rewind }) }
```
GustavoCaso, line 216: All of the calls to `.map!` could break if the sync does not include those products. Should we hoist the result into an intermediate variable and check for it before?

lib/datadog/appsec/component.rb
lloeki: That would make sense.

lib/datadog/core/remote/client.rb
lloeki: The calls to `map!` themselves won't break: `contents.select` would just return an empty array.

What should be done here is handling the corner case of having these variables be empty, in which case we should fill them with the default local values (e.g vendored ruleset).

GustavoCaso: > The calls to map! themselves won't break: contents.select would just return an empty array.

You are right 😄

lib/datadog/appsec/component.rb
lloeki: Done.

spec/datadog/appsec/component_spec.rb
```diff
@@ -59,7 +59,112 @@
     end
   end
 
+  describe '#reconfigure' do
+    context 'lock' do
+      it 'makes sure to synchronize' do
+        mutex = Mutex.new
+        processor = instance_double(Datadog::AppSec::Processor)
+        component = described_class.new(processor: processor, mutex: mutex)
+        expect(mutex).to receive(:synchronize)
```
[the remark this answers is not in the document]
GustavoCaso: I had to tweak your suggestion to avoid messing up with Rspec internal `Mutex`

```ruby
it 'makes sure to synchronize' do
  mutex = Mutex.new
  processor = instance_double(Datadog::AppSec::Processor)
  component = described_class.new(processor: processor)
  component.instance_variable_set(:@mutex, mutex)
  expect(mutex).to receive(:synchronize)
  component.reconfigure(ruleset: {})
end
```

GustavoCaso (approved):

PR #2715, opened 2023-03-23
Omit `_dd.span_sampling.max_per_second` when not configured

marcotc:
For Single Span Sampling, when no rate limit is provided, we should not include the `_dd.span_sampling.max_per_second`.

Before this PR, `_dd.span_sampling.max_per_second` was included with a default value of `-1`, meaning unlimited.

After this PR, `_dd.span_sampling.max_per_second` will be omitted when not configured. The backend will assume an unlimited rate for such cases.

This fixes an inconsistency we had with the Single Span Sampling specification.

### DIFF
--- a/lib/datadog/tracing/sampling/span/rule.rb
+++ b/lib/datadog/tracing/sampling/span/rule.rb
@@ -22,7 +22,7 @@ module Datadog
           def initialize(
             matcher,
             sample_rate: Span::Ext::DEFAULT_SAMPLE_RATE,
-            rate_limit: Span::Ext::DEFAULT_MAX_PER_SECOND
+            rate_limit: nil
           )
 
             @matcher = matcher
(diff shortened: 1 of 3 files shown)


lib/datadog/tracing/sampling/span/rule.rb
```diff
@@ -22,7 +22,7 @@ class Rule
           def initialize(
             matcher,
             sample_rate: Span::Ext::DEFAULT_SAMPLE_RATE,
-            rate_limit: Span::Ext::DEFAULT_MAX_PER_SECOND
+            rate_limit: nil
```
delner, line 25: Interesting. Why demote this from constructor args to constructor body?

marcotc: I need to know when `rate_limit` was not set (thus `nil`).

Before, it would fall back to the default unlimited value `-1`, but when I can't tell if the user has explicitly set `-1` or not set anything at all.

We only want to remove the `span_op.set_metric(Span::Ext::TAG_MAX_PER_SECOND, ...` tag if the user did not provide a `rate_limit`.

ivoanjo:
This has been stale for quite a long time so I'm going to go ahead and close it.

PR #2082, opened 2022-06-12
Allow passing request_queuing option to Rack through Rails tracer

KieranP:
Supporting this means I don't need to add and hacky overrides

PR #2716, opened 2023-03-24
Allow passing web_service_name option to Rack through Rails tracer

Surgo:
Allow passing `web_service_name` as same as #2082


marcotc (commented):
Thank you for this PR, @Surgo! 🙇 
Would you mind adding a test to cover this new option?

docs/GettingStarted.md
```diff
@@ -1732,6 +1732,7 @@ end
 | --- | ----------- | ------- |
 | `distributed_tracing` | Enables [distributed tracing](#distributed-tracing) so that this service trace is connected with a trace of another service if tracing headers are received | `true` |
 | `request_queuing` | Track HTTP request time spent in the queue of the frontend server. See [HTTP request queuing](#http-request-queuing) for setup details. | `false` |
+| `web_service_name` | Service name for frontend server request queuing spans. (e.g. `'nginx'`) | `'web-server'` |
```
buraizu, line 1735:
```suggestion
| `web_service_name` | Service name for frontend server request queuing spans. (such as `'nginx'`) | `'web-server'` |
```

buraizu (approved):
Thanks, approving with a minor update for accessibility

esalter:
bump! Would love to see this land.

tylerwillingham:
Just voicing support - my team would love to see this land too :)

lib/datadog/tracing/contrib/rails/configuration/settings.rb
```diff
@@ -49,6 +49,8 @@ def initialize(options = {})
             end
 
             option :distributed_tracing, default: true, type: :bool
+            option :request_queuing, default: false, type: :bool
```
anmarchenko, line 52: this line is probably not needed here as this option is still defined on line 55

Surgo: Removed at cb9ccbe

marcotc:
@Surgo is there an issue with this being done at the Rack level, `c.tracing.instrument :rack, web_service_name: ...`?

tylerwillingham:
@Surgo speaking from another team at another company that was also surprised to learn that we couldn't pass `web_service_name` to the configuration for the `rails` instrument (like we do with `request_queuing`): we were similarly surprised when this option didn't pass through.

We spent a decent bit of time debugging and reading source code to realize "request_queuing" is getting special treatment (and it's pretty nice) but its sibling argument of "web_service_name" wasn't.

It feels like a precedent has been set to allow these args to pass through and it makes the configuration a little cleaner

langsharpe:
> is there an issue with this being done at the Rack level, `c.tracing.instrument :rack, web_service_name: ...`?

I'm finding this isn't working for us (gem version 1.23.2). Any ideas? Otherwise, it would be nice for this PR to be merged.

```ruby
Datadog.configure do |config|
    service_name = ENV.fetch("HEROKU_APP_NAME")

    config.tracing.instrument :rails, service_name: service_name, request_queuing: true
...
    config.tracing.instrument :rack, service_name: "#{service_name}-rack", web_service_name: "#{service_name}-puma"
end
```

PR #2717, opened 2023-03-24
[APPSEC-8458] introduce AppSec::Utils::ReadWriteLock

GustavoCaso:
**What does this PR do?**

Introduce a native ruby ReadWrite Lock implementation. This RWLock would be used in subsequent work for the Remote Client Configuration.  

You can check this [PR](https://github.com/DataDog/dd-trace-rb/pull/2714) to understand where we will use the read-write lock.

The implementation provides a fair reader write lock. 

It allows for as many concurrent readers but only one concurrent writer. To avoid the writer thread being exhausted, once a writer thread has signalled that it wants to acquire the lock, no more reader threads can acquire the read lock. The new reader threads will wait for the writer thread to finish the writing. 

**Motivation**

The motivation is to reduce the contention when reading or writing the `AppSec.processor` singleton variable.

There will be a background thread that might modify the processor instance. During that process, we want any request to wait until the background thread has finished updating the processor instance. 

In this case, we do not have any background thread modifying the processor instance. We want as many concurrent read operations as possible.

**Additional Notes**

I provided a simple unit test for the implementation, and I also tested against this script: 

```ruby
require 'benchmark'

TOTAL_THREADS = 100 # set as high as practicable, for a better test

def test(lock)
  puts "READ INTENSIVE (80% read, 20% write):"
  single_test(lock, (TOTAL_THREADS * 0.8).floor, (TOTAL_THREADS * 0.2).floor)
  puts "WRITE INTENSIVE (80% write, 20% read):"
  single_test(lock, (TOTAL_THREADS * 0.2).floor, (TOTAL_THREADS * 0.8).floor)
  puts "BALANCED (50% read, 50% write):"
  single_test(lock, (TOTAL_THREADS * 0.5).floor, (TOTAL_THREADS * 0.5).floor)
end

def single_test(lock, n_readers, n_writers, reader_iterations=50, writer_iterations=50, reader_sleep=0.001, writer_sleep=0.001)
  puts "Testing #{lock.class} with #{n_readers} readers and #{n_writers} writers. Readers iterate #{reader_iterations} times, sleeping #{reader_sleep}s each time, writers iterate #{writer_iterations} times, sleeping #{writer_sleep}s each time"
  mutex = Mutex.new
  bad   = false
  data  = 0

  result = Benchmark.measure do
    readers = n_readers.times.collect do |i|
                Thread.new do
                  reader_iterations.times do
                    lock.with_read_lock do
                      mutex.synchronize { bad = true } if (data % 2) != 0
                      sleep(reader_sleep)
                      mutex.synchronize { bad = true } if (data % 2) != 0
                    end
                  end
                end
              end
    writers = n_writers.times.collect do |i|
                Thread.new do
                  writer_iterations.times do
                    lock.with_write_lock do
                      value = (data += 1)
                      sleep(writer_sleep)
                      data  = value+1
                    end
                  end
                end
              end

    readers.each { |t| t.join }
    writers.each { |t| t.join }
    puts "BAD!!! Readers+writers overlapped!" if mutex.synchronize { bad }
    puts "BAD!!! Writers overlapped! Expected data #{(n_writers * writer_iterations * 2)} got #{data}" if data != (n_writers * writer_iterations * 2)
    puts "Expected data #{(n_writers * writer_iterations * 2)} got #{data}"
  end
  puts result
end

test(Datadog::AppSec::Utils::ReadWriteLock.new)
```

This is the results from running it:
```
READ INTENSIVE (80% read, 20% write):
Testing Locks::ReadWriteMutex with 80 readers and 20 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
Expected data 2000 got 2000
  0.028927   0.095037   0.123964 (  1.388397)

WRITE INTENSIVE (80% write, 20% read):
Testing Locks::ReadWriteMutex with 20 readers and 80 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
Expected data 8000 got 8000
  0.089277   0.143242   0.232519 (  5.363435)

BALANCED (50% read, 50% write):
Testing Locks::ReadWriteMutex with 50 readers and 50 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
Expected data 5000 got 5000
  0.066041   0.151607   0.217648 (  3.345871)
```

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/appsec/utils/read_write_lock.rb
+++ b/lib/datadog/appsec/utils/read_write_lock.rb
@@ -1,0 +1,78 @@
+# frozen_string_literal: true
+
+module Datadog
+  module AppSec
+    module Utils
+      # Simple implentation for a fair read write lock
+      # It allows for as many concurrent readers but only one concurrent writer.
+      # To avoid the writer thread being exhausted, once a writer thread has signaled that it wants to acquire the lock,
+      # no more reader threads can acquire the lock.
+      # The new reader threads will wait for the writer thread to finish.
+      class ReadWriteLock
+        def initialize
+          @reader_mutex    = Mutex.new
+          @reader_q        = ConditionVariable.new
+          @reader_count    = 0
+          @reader_releases = 0
+
+          @writer_mutex    = Mutex.new
+          @writer_q        = ConditionVariable.new
+          @writter         = false
+        end
+
+        def with_rlock
+          rlock
+          yield
+        ensure
+          runlock
+        end
+
+        def with_lock
+          lock
+          yield
+        ensure
+          unlock
+        end
+
+        def rlock
+          @reader_mutex.synchronize do
+            @reader_count += 1
+
+            writter = false
+
+            @writer_mutex.synchronize do
+              writter = @writter
+            end
+
+            @reader_q.wait(@reader_mutex) if writter
+          end
+        end
+
+        def runlock
+          @reader_mutex.synchronize do
+            @reader_releases += 1
+
+            @writer_mutex.synchronize { @writer_q.signal } if @reader_releases == @reader_count
+          end
+        end
+
+        def lock
+          @writer_mutex.synchronize do
+            @writer_q.wait(@writer_mutex) while (@reader_releases != @reader_count) || @writter
+
+            @writter = true
+          end
+        end
+
+        def unlock
+          @writer_mutex.synchronize do
+            @writter = false
+            @writer_q.signal
+          end
+
+          @reader_mutex.synchronize { @reader_q.broadcast }
+        end
+      end
+    end
+  end
+end
--- a/spec/datadog/appsec/utils/read_write_lock_spec.rb
+++ b/spec/datadog/appsec/utils/read_write_lock_spec.rb
@@ -1,0 +1,129 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+require 'datadog/appsec/utils/read_write_lock'
+
+RSpec.describe Datadog::AppSec::Utils::ReadWriteLock do
+  subject(:read_write_lock) { described_class.new }
+
+  context 'read locks' do
+    it 'allows for many readers' do
+      rw_lock = read_write_lock
+      state = 0
+      read_threads = Array.new(3) do
+        Thread.new do
+          rw_lock.with_rlock do
+            expect(state).to eq(0)
+          end
+        end
+      end
+
+      read_threads.each(&:join)
+    end
+
+    it 'release the read lock when an exception occurs' do
+      rw_lock = read_write_lock
+      expect(rw_lock).to receive(:runlock)
+      begin
+        rw_lock.with_rlock do
+          raise StandardError
+        end
+      rescue
+        # do nothing
+      end
+    end
+
+    it 'writter waits while readers holds the lock' do
+      rw_lock = read_write_lock
+      state = 0
+      queue = Queue.new
+      queue2 = Queue.new
+      queue3 = Queue.new
+
+      read_thread = Thread.new do
+        rw_lock.with_rlock do
+          queue << 1
+          queue3 << 1
+          queue2.pop
+        end
+      end
+
+      write_thread = Thread.new do
+        queue3.pop
+        rw_lock.with_lock do
+          state = 1
+        end
+      end
+
+      # The read thead holds the lock and the writter tries to acquire the lock
+      queue.pop
+      expect(state).to eq(0)
+
+      # we relase the reader lock ans the write locks acquires it and modify state
+      queue2 << 1
+      read_thread.join
+      write_thread.join
+      expect(state).to eq(1)
+    end
+  end
+
+  context 'writer locks' do
+    it 'release the write lock when an exception occurs' do
+      rw_lock = read_write_lock
+      expect(rw_lock).to receive(:unlock)
+      begin
+        rw_lock.with_lock do
+          raise StandardError
+        end
+      rescue
+        # do nothing
+      end
+    end
+
+    it 'read threads wait for writter lock to be released' do
+      rw_lock = read_write_lock
+      state = 0
+      queue = Queue.new
+      queue2 = Queue.new
+      queue3 = Queue.new
+      queue4 = Queue.new
+
+      read_thread = Thread.new do
+        queue2.pop
+        expect(state).to eq(0)
+        queue3 << 1
+
+        # Wait for writer thread to release the lock. That happens at line 119
+        rw_lock.with_rlock do
+          expect(state).to eq(1)
+        end
+      end
+
+      write_thread = Thread.new do
+        rw_lock.with_lock do
+          queue.pop
+          # Allow the reader thread to start running. Check the vale of state before the writer changes it
+          queue2 << 1
+
+          # The reader thread has signal that state is 0 and procced to acquire the reader lock.
+          # It has to wait because the writer thread holds the lock
+          queue3.pop
+
+          # Modify state
+          state = 1
+
+          queue4.pop
+        end
+      end
+
+      # The writter thread holds the lock and modify state
+      queue << 1
+
+      # We release the writter lock and assert that state has change inside the reader thread
+      queue4 << 1
+      read_thread.join
+      write_thread.join
+    end
+  end
+end
--- a/sig/datadog/appsec/utils/read_write_lock.rbs
+++ b/sig/datadog/appsec/utils/read_write_lock.rbs
@@ -1,0 +1,29 @@
+module Datadog
+  module AppSec
+    module Utils
+      class ReadWriteLock
+        attr_reader reader_mutex: Thread::Mutex
+        attr_reader writer_mutex: Thread::Mutex
+        attr_reader writer_q: Thread::ConditionVariable
+        attr_reader reader_q: Thread::ConditionVariable
+        attr_reader reader_count: Integer
+        attr_reader reader_releases: Integer
+        attr_reader writter: bool
+
+        def initialize: () -> void
+
+        def with_rlock: () { () -> untyped } -> void
+
+        def with_lock: () { () -> untyped } -> void
+
+        def rlock: () -> void
+
+        def runlock: () -> void
+
+        def lock: () -> void
+
+        def unlock: () -> void
+      end
+    end
+  end
+end


lib/datadog/appsec/utils/read_write_lock.rb
```diff
@@ -0,0 +1,78 @@
...
+module Datadog
+  module AppSec
+    module Utils
+      # Simple implentation for a fair read write lock
+      # It allows for as many concurrent readers but only one concurrent writer.
+      # To avoid the writer thread being exhausted, once a writer thread has signaled that it wants to acquire the lock,
+      # no more reader threads can acquire the lock.
+      # The new reader threads will wait for the writer thread to finish.
+      class ReadWriteLock
+        def initialize
```
GustavoCaso, line 12: Should we allow to set the maximum number of readers?

lib/datadog/appsec/utils/read_write_lock.rb
```diff
@@ -0,0 +1,78 @@
...
+            @writer_q.wait(@writer_mutex) while (@reader_releases != @reader_count) || @writter
+
+            @writter = true
+          end
+        end
+
+        def unlock
+          @writer_mutex.synchronize do
+            @writter = false
+            @writer_q.signal
```
GustavoCaso, line 70: We signal the waiting writers that they can acquire the lock

lib/datadog/appsec/utils/read_write_lock.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+module Datadog
+  module AppSec
+    module Utils
+      # Simple implentation for a fair read write lock
+      # It allows for as many concurrent readers but only one concurrent writer.
+      # To avoid the writer thread being exhausted, once a writer thread has signaled that it wants to acquire the lock,
+      # no more reader threads can acquire the lock.
+      # The new reader threads will wait for the writer thread to finish.
```
ivoanjo, line 10: I'm still a bit curious if this is the best policy for our situation; would it be better to perhaps have some kind of a softer limit where the writer tries to wait for a period without any readers for a while without "forcing itself" in, and after that period we fall back to the "writer gets priority" policy?

lib/datadog/appsec/utils/read_write_lock.rb
```diff
@@ -0,0 +1,78 @@
...
+      class ReadWriteLock
+        def initialize
+          @reader_mutex    = Mutex.new
+          @reader_q        = ConditionVariable.new
+          @reader_count    = 0
+          @reader_releases = 0
+
+          @writer_mutex    = Mutex.new
+          @writer_q        = ConditionVariable.new
+          @writter         = false
```
ivoanjo, line 20: Typo: `writter`

lib/datadog/appsec/utils/read_write_lock.rb
```diff
@@ -0,0 +1,78 @@
...
+          end
+        end
+
+        def runlock
+          @reader_mutex.synchronize do
+            @reader_releases += 1
+
+            @writer_mutex.synchronize { @writer_q.signal } if @reader_releases == @reader_count
+          end
+        end
```
ivoanjo, line 57: I'm assuming the reason for having `@reader_mutex` and `@writer_mutex` be separate is performance. But... are we really getting better performance out of this design?

It appears to me that the reader lock/unlock code paths are the most performance-sensitive: they're the ones that will get executed the most often, and that impact application code.

But these same paths have to acquire **both** locks! So I think we're actually penalizing them by having them acquire two locks most of the times.

My suggestion is going with a simpler 1-mutex implementation or at least change the 2-mutex implementation so the extra work is on the writers side, not on the readers.

lib/datadog/appsec/utils/read_write_lock.rb
```diff
@@ -0,0 +1,78 @@
...
+
+        def lock
```
ivoanjo, line 59: Consider checking if the writer lock is already taken -- right now if a writer tries to acquire a lock twice, it will either succeed (if there are no readers), and then incorrectly release the lock after the first unlock (oops), or it will deadlock (if there are active readers waiting -- I leave why that is as an exercise for the reader).

spec/datadog/appsec/utils/read_write_lock_spec.rb
```diff
@@ -0,0 +1,129 @@
...
+    it 'allows for many readers' do
+      rw_lock = read_write_lock
+      state = 0
+      read_threads = Array.new(3) do
+        Thread.new do
+          rw_lock.with_rlock do
+            expect(state).to eq(0)
+          end
+        end
+      end
```
ivoanjo, line 20: This expectation is a bit weak, since it's trivially realized even if the code never runs.

One possible idea to make this test a bit stronger is perhaps having each thread have an id (e.g. `(1..3).map` ...) and then each thread could put is own id on a `Queue`. At the end of the test, you should be able to pop all 3 distinct ids out of the queue.

ivoanjo (commented):
Left a few notes! I still recommend not making our design more complex than it needs to be, as it's easy to get things wrong, and the more complexity, the harder it is to reason about the implementation and figure out it is wrong.

lib/datadog/appsec/utils/read_write_lock.rb
```diff
@@ -0,0 +1,78 @@
...
+          @reader_mutex.synchronize do
+            @reader_count += 1
+
+            writter = false
+
+            @writer_mutex.synchronize do
+              writter = @writter
+            end
+
+            @reader_q.wait(@reader_mutex) if writter
```
lloeki, line 47: IIUC the `@writer_mutex.synchronize` makes `rlock` wait once a write lock attempt has started (signalled by the same `@writer_mutex.synchronize` in `#lock`) in order to implement this:

```
      # To avoid the writer thread being exhausted, once a writer thread has signaled that it wants to acquire the lock,
      # no more reader threads can acquire the lock.
```

Then `@writer_q.wait(@writer_mutex) while (@reader_releases != @reader_count)` makes `lock` wait until all readers have released their locks.

The ` || @writter` part is what makes it wait if another writer is in progress, so it's what implements the lock for writers.

Right after that `lock` sets `@writter = true` to actually acquire the (write) lock.

Since `@writter` is a boolean this is what makes it have only a single writer as described:

```
      # It allows for ... only one concurrent writer.
```

Now, back to this highlight, the intent of `@reader_q.wait(@reader_mutex) if writter` seems to be that if the writer lock is held by a writer then we wait for it to be unlocked to proceed with the remainder of `rlock` (which happens to be nothing), which effectively blocks the read lock to be acquired until the write lock is released.

`writter` is obtained from `@writter` within the `@writer_mutex.synchronize` block, and used after the end of that block. This gives the opportunity for the Ruby VM to switch threads, so we must ensure that in all cases `writter` and `@writter` keep having the same value, otherwise `if writter` will not check reality.

After the entirety of either `lock` or `unlock` there are two cases.

If it is after `unlock` then `writter` will be `false`. By that point a writer cannot alter `@writter` to `true` (which is only done via `lock`) because of a) `@reader_count += 1` above and b) no other reader can `runlock` in a way that would compensate that `+= 1` (because of the surrounding `@reader_mutex.synchronize`). So it seems like we're good here.

If it is after `lock` then `writter` will be `true`. By that point a writer has locked the lock, and in between `synchronize do .. end` and `@reader_q.wait(@reader_mutex) if writter` that writer could alter `@writter` to `false` (which, by the way, is only done via `unlock`).

Even though `@writter` is now `false`, `writter is still `true` and `@reader_q.wait(@reader_mutex)` will be called. After having set `@writter`, `unlock` will attempt  `@reader_mutex.synchronize { @reader_q.broadcast }` but be unable to proceed since `@reader_mutex` is held by `rlock`.

So on each side we have:

- `@reader_mutex.synchronize do ... @reader_q.wait(@reader_mutex) ... end` waiting for `@reader_q.broadcast` to be unblocked
- `@reader_mutex.synchronize { @reader_q.broadcast }` waiting for `@reader_mutex.synchronize do ... end` to be unblocked

Which if I'm not mistaken would be a deadlock.

lloeki (commented):
I like how small it looks like. There are a few flaws though.

lib/datadog/appsec/utils/read_write_lock.rb
lloeki: I concur, ideally `rlock` and `runlock` should be lockless but that's not possible in pure Ruby. The nested locks are scary.

As for performance, ultimately I would be only convinced with benchmarks of various kinds to see if it is performing better, and in which cases, so that we can understand the tradeoffs.

lib/datadog/appsec/utils/read_write_lock.rb
```diff
@@ -0,0 +1,78 @@
...
+        ensure
+          runlock
+        end
+
+        def with_lock
+          lock
+          yield
+        ensure
+          unlock
+        end
```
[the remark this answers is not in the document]
lloeki: This API is inspired by:

- the [Go API's RWMutex](https://pkg.go.dev/sync#RWMutex)
- the Ruby API's `Mutex`

Which seemed like very close relatives.

Both `concurrent_ruby` and the Java API veer off significantly from Ruby's `Mutex` interface.

Regarding that one:

```
lock.read_lock.synchronize { puts "Hey I'm reading!" }
lock.write_lock.synchronize { puts "Hey, I'm writing!" }
```

This method chaining feels very Rails-y and would need two method receiving intermediate facade instances for the sole purpose of delegating to the parent one, especially since internal state is by nature shared. Probably if you want to keep a single `synchronize` instead of `with_*` then `lock.synchronize(:read) { }` would suffice, with the default argument value being `:write`, thus making that `RWMutex` a drop-in replacement for `Mutex`.

lib/datadog/appsec/utils/read_write_lock.rb
```diff
@@ -0,0 +1,78 @@
...
+            @writer_mutex.synchronize { @writer_q.signal } if @reader_releases == @reader_count
+          end
+        end
+
+        def lock
+          @writer_mutex.synchronize do
+            @writer_q.wait(@writer_mutex) while (@reader_releases != @reader_count) || @writter
+
+            @writter = true
+          end
```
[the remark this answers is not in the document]
lloeki: > Otherwise readers will keep arriving

Indeed it looks like `@reader_count += 1` could happen while a writer is waiting.

> you probably want the writer to set the flag to true before waiting on the condition variable

- It would mean the write lock is taken before waiting?
- The `(@reader_releases != @reader_count) || @writter` condition would always be `true` (since no `unlock` would reset it to `false` because of `@writer_mutex.synchronize`, which would not make sense anyway since that `unlock` would presumably apply to _another_ write lock)

Instead, symmetrically to how the writer lock is taken _after_ waiting, it seems like the read lock should be taken only _after_ waiting, so when no writer is in progress, so it looks like `@reader_count += 1`  should be inside the `@writer_mutex.synchronize do ... end` block of `rlock`.

lib/datadog/appsec/utils/read_write_lock.rb
```diff
@@ -0,0 +1,78 @@
...
+              writter = @writter
+            end
+
+            @reader_q.wait(@reader_mutex) if writter
+          end
+        end
+
+        def runlock
+          @reader_mutex.synchronize do
+            @reader_releases += 1
```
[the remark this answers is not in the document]
GustavoCaso: The implementation never reset the reader and release count to 0. So answering your question, yes, it will go towards infinite. 

That could be an issue, as the value of reader count and release will keep growing and might lead to problems if the customer application is not deployed, releasing the values. 

We might be able to reset the reader and release count to 0 once the values are the same in the `rlock` method.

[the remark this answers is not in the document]
GustavoCaso: Or as you suggested in the comment below, we could have a single variable `reader_count `

lib/datadog/appsec/utils/read_write_lock.rb
GustavoCaso: @ivoanjo @lloeki 

I updated the benchmark so we can compare the actual implementation with a more straightforward implementation using a single mutex. 

The code for a single `Mutex`, is just that a single mutex, no counter no conditional variables. 

Here is the code of the benchmark 

```ruby
# frozen_string_literal: true

require_relative 'read_write_mutex'

require 'benchmark'

TOTAL_THREADS = 100 # set as high as practicable, for a better test

def test(lock)
  puts "READ INTENSIVE (80% read, 20% write):"
  single_test(lock, (TOTAL_THREADS * 0.8).floor, (TOTAL_THREADS * 0.2).floor)
  puts "WRITE INTENSIVE (80% write, 20% read):"
  single_test(lock, (TOTAL_THREADS * 0.2).floor, (TOTAL_THREADS * 0.8).floor)
  puts "BALANCED (50% read, 50% write):"
  single_test(lock, (TOTAL_THREADS * 0.5).floor, (TOTAL_THREADS * 0.5).floor)
end

def single_test(lock, n_readers, n_writers, reader_iterations=50, writer_iterations=50, reader_sleep=0.001, writer_sleep=0.001)
  puts "Testing #{lock.class} with #{n_readers} readers and #{n_writers} writers. Readers iterate #{reader_iterations} times, sleeping #{reader_sleep}s each time, writers iterate #{writer_iterations} times, sleeping #{writer_sleep}s each time"
  mutex = Mutex.new
  bad   = false
  data  = 0

  result = Benchmark.measure do
    readers = n_readers.times.collect do |i|
                Thread.new do
                  reader_iterations.times do
                    lock.with_rlock do
                      mutex.synchronize { bad = true } if (data % 2) != 0
                      sleep(reader_sleep)
                      mutex.synchronize { bad = true } if (data % 2) != 0
                    end
                  end
                end
              end
    writers = n_writers.times.collect do |i|
                Thread.new do
                  writer_iterations.times do
                    lock.with_lock do
                      value = (data += 1)
                      sleep(writer_sleep)
                      data  = value+1
                    end
                  end
                end
              end

    readers.each { |t| t.join }
    writers.each { |t| t.join }
    puts "BAD!!! Readers+writers overlapped!" if mutex.synchronize { bad }
    puts "BAD!!! Writers overlapped! Expected data #{(n_writers * writer_iterations * 2)} got #{data}" if data != (n_writers * writer_iterations * 2)
    puts "Expected data #{(n_writers * writer_iterations * 2)} got #{data}"
  end
  puts result
end

# for performance comparison with ReadWriteLock
class SimpleMutex
  def initialize; @mutex = Mutex.new; end
  def with_rlock
    @mutex.synchronize { yield }
  end
  alias :with_lock :with_rlock
end

test(SimpleMutex.new)
test(Locks::ReadWriteMutex.new)
```

The result of running it is: 

```
READ INTENSIVE (80% read, 20% write):
Testing SimpleMutex with 80 readers and 20 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
Expected data 2000 got 2000
  0.062491   0.126706   0.189197 (  6.403470)
WRITE INTENSIVE (80% write, 20% read):
Testing SimpleMutex with 20 readers and 80 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
Expected data 8000 got 8000
  0.059176   0.115606   0.174782 (  6.465615)
BALANCED (50% read, 50% write):
Testing SimpleMutex with 50 readers and 50 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
Expected data 5000 got 5000
  0.066263   0.132126   0.198389 (  6.446812)


READ INTENSIVE (80% read, 20% write):
Testing Locks::ReadWriteMutex with 80 readers and 20 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
Expected data 2000 got 2000
  0.040352   0.142648   0.183000 (  1.363321)
WRITE INTENSIVE (80% write, 20% read):
Testing Locks::ReadWriteMutex with 20 readers and 80 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
Expected data 8000 got 8000
  0.076468   0.122011   0.198479 (  5.293635)
BALANCED (50% read, 50% write):
Testing Locks::ReadWriteMutex with 50 readers and 50 writers. Readers iterate 50 times, sleeping 0.001s each time, writers iterate 50 times, sleeping 0.001s each time
Expected data 5000 got 5000
  0.052424   0.122417   0.174841 (  3.277331)
```

lib/datadog/appsec/utils/read_write_lock.rb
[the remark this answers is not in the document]
GustavoCaso: > For that to happen, you probably want the writer to set the flag to true before waiting on the condition variable.

In that case the writter thread will always wait for the condition variable. 

> Instead, symmetrically to how the writer lock is taken after waiting, it seems like the read lock should be taken only after waiting, so when no writer is in progress, so it looks like @reader_count += 1 should be inside the @writer_mutex.synchronize do ... end block of rlock.

Wouldn't make more sense to increment `reader_count` after the wait outside of the `@write_mutex.syncronize`  than inside of it? 

Something like: 

```ruby
def rlock
  @reader_mutex.synchronize do
    writter = false

    @writer_mutex.synchronize do
      writter = @writter
    end

    @reader_q.wait(@reader_mutex) if writter

    @reader_count += 1
  end
end
```

cc @lloeki @ivoanjo

lib/datadog/appsec/utils/read_write_lock.rb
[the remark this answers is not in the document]
GustavoCaso: I like the `lock.synchronize(:read) { }` API

lib/datadog/appsec/utils/read_write_lock.rb
```diff
@@ -0,0 +1,78 @@
...
+            @reader_count += 1
+
+            writter = false
+
+            @writer_mutex.synchronize do
+              writter = @writter
+            end
+
+            @reader_q.wait(@reader_mutex) if writter
+          end
```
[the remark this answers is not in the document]
GustavoCaso: That is correct. That is the intent of this implementation. 

We can see the comparison result between a single Mutex implementation and this implementation that uses two Mutexes. 

https://github.com/DataDog/dd-trace-rb/pull/2717#discussion_r1149176712

lib/datadog/appsec/utils/read_write_lock.rb
GustavoCaso: @lloeki, your explanation makes sense; the code could lead to a deadlock situation.

GustavoCaso:
@lloeki and I created a simple test app to get some results of using a single mutex or the RWLock implementation from this PR. 

The results show improved throughput and wait time when using RWLock over a single mutex. Still, the improvement is different from orders of magnitude that wouldn't force us to start using the RWLock implementation in production. We will begin using a single mutex, see the results on our reliability environment, and if there is a lot of degradation, we will revisit this topic. 

Here are the code and results: 

App: 

```ruby 
# config.ru

require 'pry'
require 'logger'
require 'digest'
require 'securerandom'
require 'json'
load 'rwmutex.rb'

logger = Logger.new(STDOUT, Logger::DEBUG)

mode = ENV['MODE']
read_work = (ENV['READ_WORK'] || 0.001).to_f
write_work = (ENV['WRITE_WORK'] || 0.01).to_f
req_work = (ENV['REQ_WORK'] || 0.10).to_f
calibration_filename = ENV['CALIBRATION'] || 'calibration.json'
interval = (ENV['INTERVAL'] || 1).to_f


mutex = case mode
        when 'mutex'
          Mutex.new
        when 'rwmutex'
          ReadWriteLock.new
        end


def work(times)
  times.times do
    Digest::SHA256.hexdigest(SecureRandom.uuid)
  end
end

if File.exist?(calibration_filename)
  data = JSON.parse(File.read(calibration_filename))
  $rate = data['rate']

  logger.debug { "#{Thread.current}: calibrated: #{$rate}/s" }
else
  logger.debug { "#{Thread.current}: calibrating" }
  times = 100_000

  start = Time.now
  work(times)
  end_ = Time.now

  duration = end_ - start
  $rate = times.to_f / duration

  File.write(calibration_filename, JSON.dump({ rate: $rate }))

  logger.debug { "#{Thread.current}: calibrated: #{$rate}/s" }
end

def work_for(duration)
  work(($rate * duration).to_i)
end

app = proc do |env|
  case mode
  when 'mutex'
    mutex.synchronize do
      work_for(read_work)
    end
  when 'rwmutex'
    mutex.with_rlock do
      work_for(read_work)
    end
  else
    work_for(read_work)
  end

  request = Rack::Request.new(env)

  if request.path == '/'
    work_for(req_work)

    [ 200, {'Content-Type' => 'text/plain'}, ['Hello, wat is love?'] ]
  else
    [ 404, {'Content-Type' => 'text/plain'}, ['not found'] ]
  end
end

Thread.new do
  loop do
    logger.debug { "#{Thread.current}: working" }
    # work
    case mode
    when 'mutex'
      mutex.synchronize do
        work_for(write_work)
      end
    when 'rwmutex'
      mutex.with_lock do
        work_for(write_work)
      end
    else
      work_for(write_work)
    end

    # wait
    logger.debug { "#{Thread.current}: sleeping" }
    sleep interval
  end
end

run app
```

Here are the results using [vegeta](https://github.com/tsenart/vegeta) 

```
env MODE=none READ_WORK=0.001 WRITE_WORK=0.01 REQ_WORK=0.1 bundle exec puma -w 0 -t 10 -p 9292

echo "GET http://localhost:9292/" | vegeta attack -duration=60s -rate=0 -max-workers=20 | tee results.bin | vegeta report
Requests      [total, rate, throughput]         621, 10.35, 10.05
Duration      [total, attack, wait]             1m2s, 1m0s, 1.767s
Latencies     [min, mean, 50, 90, 95, 99, max]  97.991ms, 1.961s, 913.402ms, 2.663s, 12.822s, 14.512s, 17.544s
Bytes In      [total, mean]                     11799, 19.00
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:621  
Error Set:
```

```
env MODE=mutex READ_WORK=0.001 WRITE_WORK=0.01 REQ_WORK=0.1 bundle exec puma -w 0 -t 10 -p 9292

echo "GET http://localhost:9292/" | vegeta attack -duration=60s -rate=0 -max-workers=20 | tee results.bin | vegeta report
Requests      [total, rate, throughput]         623, 10.37, 10.03
Duration      [total, attack, wait]             1m2s, 1m0s, 2.057s
Latencies     [min, mean, 50, 90, 95, 99, max]  100.14ms, 1.963s, 930.49ms, 3.547s, 12.218s, 14.262s, 19.494s
Bytes In      [total, mean]                     11837, 19.00
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:623  
Error Set:
```

```
env MODE=rwmutex READ_WORK=0.001 WRITE_WORK=0.01 REQ_WORK=0.1 bundle exec puma -w 0 -t 10 -p 9292

echo "GET http://localhost:9292/" | vegeta attack -duration=60s -rate=0 -max-workers=20 | tee results.bin | vegeta report
Requests      [total, rate, throughput]         607, 10.11, 9.80
Duration      [total, attack, wait]             1m2s, 1m0s, 1.908s
Latencies     [min, mean, 50, 90, 95, 99, max]  98.565ms, 2.011s, 936.244ms, 2.822s, 12.596s, 14.772s, 19.719s
Bytes In      [total, mean]                     11533, 19.00
Bytes Out     [total, mean]                     0, 0.00
Success       [ratio]                           100.00%
Status Codes  [code:count]                      200:607  
Error Set:
```

We are going. to leave the PR open until we fully agree on the implementation we are going to use in production

lib/datadog/appsec/utils/read_write_lock.rb
ivoanjo: Note that I'm not advocating here to use a single `Mutex` instead of a `ReadWriteMutex` (although that may be an interesting option to explore). I'm advocating to use a single `Mutex` in the underlying implementation of the `ReadWriteMutex`, instead of two different `Mutexes`, which makes implementation and validation more complex.

ivoanjo:
Given that we ended up not using it, shall we go ahead and close this PR for now?

ivoanjo:
I'm going to go ahead and close this one, since we ended up not using it for now :)

PR #2719, opened 2023-03-24
Test:Fix start_with typo

marcotc:
Minor typo on `String#start_with?` method. Does not affect test correctness, but breaks deadline thread.


ivoanjo (approved):
Urgghhh thanks Marco for fixing my derps

PR #2720, opened 2023-03-24
Fix w3c propagation special character handling

marcotc:
This PR fixes three w3c `tracecontext` propagation issues:
1. If the provided `TraceDigest#trace_distributed_tags` contained Hash values that needed to be escaped during distributed trace injection, those value were being modified leaving the original `TraceDigest#trace_distributed_tags` changed. The provided `TraceDigest#trace_distributed_tags` does not change anymore on injection. This never affected distributed header extraction.
2. `:` is not considered an invalid `trace_distributed_tags` value anymore, thus it is not going to be escaped.
3. `=` is now serialized as `~`, instead of `:`.

These issues were caught by our tracestate headers system tests: https://github.com/DataDog/system-tests/blob/92064c6af29562f2962176767c417fc2b0eda867/parametric/test_headers_tracestate_dd.py

### DIFF
--- a/lib/datadog/tracing/distributed/trace_context.rb
+++ b/lib/datadog/tracing/distributed/trace_context.rb
@@ -194,13 +194,13 @@ module Datadog
         # Serialize `_dd.p.{key}` by first removing the `_dd.p.` prefix.
         # Then replacing invalid characters with `_`.
         #
-        # Returns a new object if modified given `name` is a Hash key, thus always frozen.
+        # Returns a new object, given `name` is a Hash key thus always frozen.
         def serialize_tag_key(name)
           key = name.delete_prefix(Tracing::Metadata::Ext::Distributed::TAGS_PREFIX)
 
           # DEV: It's unlikely that characters will be out of range, as they mostly
           # DEV: come from Datadog-controlled sources.
-          # DEV: Trying to `match?` is measurably faster than a `gsub` that does not match.
+          # DEV: Trying to `match?` is measurably faster than a `gsub!` that does not match.
           key.gsub!(INVALID_TAG_KEY_CHARS, '_') if INVALID_TAG_KEY_CHARS.match?(key)
 
           key
--- a/lib/datadog/tracing/distributed/trace_context.rb
+++ b/lib/datadog/tracing/distributed/trace_context.rb
@@ -193,20 +193,22 @@ module Datadog
 
         # Serialize `_dd.p.{key}` by first removing the `_dd.p.` prefix.
         # Then replacing invalid characters with `_`.
+        #
+        # Returns a new object if modified given `name` is a Hash key, thus always frozen.
         def serialize_tag_key(name)
           key = name.delete_prefix(Tracing::Metadata::Ext::Distributed::TAGS_PREFIX)
 
           # DEV: It's unlikely that characters will be out of range, as they mostly
           # DEV: come from Datadog-controlled sources.
           # DEV: Trying to `match?` is measurably faster than a `gsub` that does not match.
-          if INVALID_TAG_KEY_CHARS.match?(key)
-            key.gsub(INVALID_TAG_KEY_CHARS, '_')
-          else
-            key
-          end
+          key.gsub!(INVALID_TAG_KEY_CHARS, '_') if INVALID_TAG_KEY_CHARS.match?(key)
+
+          key
         end
 
-        # Replaces invalid characters with `_`, then replaces `=` with `:`.
+        # Replaces invalid characters with `_`, then replaces `=` with `~`.
+        #
+        # Returns a new object if modified given `value` belongs to the {TraceDigest} being serialized.
         def serialize_tag_value(value)
           # DEV: It's unlikely that characters will be out of range, as they mostly
           # DEV: come from Datadog-controlled sources.
@@ -217,9 +219,12 @@ module Datadog
                   value
                 end
 
-          # DEV: Checking for an unlikely '=' is faster than a no-op `tr!`.
-          ret.tr!('=', ':') if ret.include?('=')
-          ret
+          # DEV: Checking for an unlikely '=' is faster than a no-op `tr`.
+          if ret.include?('=')
+            ret.tr('=', '~')
+          else
+            ret
+          end
         end
 
         def extract_traceparent(traceparent)
@@ -309,9 +314,9 @@ module Datadog
           [origin, sampling_priority, tags, unknown_fields]
         end
 
-        # Restore `:` back to `=`.
+        # Restore `~` back to `=`.
         def deserialize_tag_value(value)
-          value.tr!(':', '=')
+          value.tr!('~', '=')
           value
         end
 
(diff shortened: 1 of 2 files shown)


lib/datadog/tracing/distributed/trace_context.rb
```diff
@@ -193,20 +193,22 @@ def serialize_origin(value)
...
           # DEV: Trying to `match?` is measurably faster than a `gsub` that does not match.
-          if INVALID_TAG_KEY_CHARS.match?(key)
-            key.gsub(INVALID_TAG_KEY_CHARS, '_')
-          else
-            key
-          end
+          key.gsub!(INVALID_TAG_KEY_CHARS, '_') if INVALID_TAG_KEY_CHARS.match?(key)
+
+          key
         end
```
ivoanjo, line 207: I'm confused by this change and the new comment.

Comment:

1. The comment states "Returns a new object if [...]". But I think we always return a new object -- because `String#delete_prefix` always returns a new object.
2. The comment mentions things being frozen, but the return of `String#delete_prefix` and `gsub!` is not frozen. Should we freeze the result?
3. I didn't understand the part about `name` being a Hash key -- can you clarify?

The change itself:

This seems to be an optimization -- mutate the string inline instead of creating a new one.

Have you considered going a bit further and replacing it with `key.gsub!(INVALID_TAG_KEY_CHARS, '_') || key`? Currently we seem to be "paying" for the regular expression matching twice -- once in `.match?` and again in `.gsub!`.

(Something similar seems to happen in `#serialize_origin` above)

lib/datadog/tracing/distributed/trace_context.rb
```diff
@@ -309,9 +314,9 @@ def extract_datadog_fields(dd_tracestate)
           [origin, sampling_priority, tags, unknown_fields]
         end
 
-        # Restore `:` back to `=`.
+        # Restore `~` back to `=`.
         def deserialize_tag_value(value)
-          value.tr!(':', '=')
+          value.tr!('~', '=')
           value
```
ivoanjo, line 320: I'm curious -- why use `tr!` instead of `gsub!`? Is it a performance optimization?

spec/datadog/tracing/distributed/trace_context_spec.rb
```diff
@@ -209,15 +213,18 @@
             "\u0000", # First unicode character
             "\u001F", # Last lower invalid character
             ',',
-            ':',
             ';',
-            "\u007F", # First upper invalid character
+            '~', # First upper invalid character
```
ivoanjo, line 217: Minor: May be worth noting here that `~` is `\u007E` (I had to check the manpage ;)

lib/datadog/tracing/distributed/trace_context.rb
```diff
@@ -309,9 +314,9 @@ def extract_datadog_fields(dd_tracestate)
           [origin, sampling_priority, tags, unknown_fields]
         end
 
-        # Restore `:` back to `=`.
+        # Restore `~` back to `=`.
         def deserialize_tag_value(value)
-          value.tr!(':', '=')
+          value.tr!('~', '=')
```
ivoanjo, line 319: (Just to doublecheck: Is this one ok to mutate, unlike the others that were changed to copy instead?)

ivoanjo (approved):
Left a few notes but overall LGTM :+1:

spec/datadog/tracing/distributed/trace_context_spec.rb
marcotc: You didn't memorize the Ascii table? 😅 
Adding a comment here.

lib/datadog/tracing/distributed/trace_context.rb
marcotc: `tr!` is much faster for single character replacement:
```ruby
require 'benchmark/ips'

Benchmark.ips do |x|
  x.report('gsub!') { 'a'.gsub!('a', 'b') }
  x.report('tr!') { 'a'.tr!('a', 'b') }
  x.compare!
end

# ruby 3.2.1
# Comparison:
#                  tr!:  3905918.3 i/s
#                gsub!:  1702250.5 i/s - 2.29x  slower
```

lib/datadog/tracing/distributed/trace_context.rb
marcotc: Yes, `value` always comes from the result of a `String#split` in this case.

spec/datadog/tracing/distributed/trace_context_spec.rb
ivoanjo: I should memorize it yeah, may come in handy if I ever get stranded on Mars.

lib/datadog/tracing/distributed/trace_context.rb
marcotc: >     1. The comment states "Returns a new object if [...]". But I think we always return a new object -- because `String#delete_prefix` always returns a new object.

I think I copy/pasted from `serialize_tag_value`, which **conditionally** creates a new object. I changed the comment here given it's always a new object.

>     2. The comment mentions things being frozen, but the return of `String#delete_prefix` and `gsub!` is not frozen. Should we freeze the result?

We don't care if the result is frozen: it's only for our own consumption so leaving it either frozen or not is fine.

>     3. I didn't understand the part about `name` being a Hash key -- can you clarify?

String Hash keys are always frozen: 
```ruby
{'a' => 1}.keys[0].frozen?
=> true
```

And the parameter to this function (`name`) comes directly from a String Hash key.
Because we know it's frozen, we know we have to dup it to modify it.

> Have you considered going a bit further and replacing it with `key.gsub!(INVALID_TAG_KEY_CHARS, '_') || key`? Currently we seem to be "paying" for the regular expression matching twice -- once in `.match?` and again in `.gsub!`.

`key.gsub!(INVALID_TAG_KEY_CHARS, '_')` is unlikely to make changes to the String, given `INVALID_TAG_KEY_CHARS` are rare as the argument to this method is created by Datadog itself.
A `INVALID_TAG_KEY_CHARS.match?(key)` that returns false is much faster than a `key.gsub!(INVALID_TAG_KEY_CHARS, '_')` that doesn't change anything:

```ruby
require 'benchmark/ips'

REGEXP = /x/

Benchmark.ips do |x|
  x.report('gsub!') { 'a'.gsub!(REGEXP, 'b') }
  x.report('match?') { 'a'.match?(REGEXP) }
  x.compare!
end

# ruby 3.2.1
# Comparison:
#               match?:  7659457.3 i/s
#                gsub!:  3016072.8 i/s - 2.54x  slower
```

ivoanjo: Sounds good! Thanks for sharing the benchmarks -- as usual, it's much better to measure than to rely on intuition when talking about performance :)

ivoanjo (approved):
New revision LGTM :+1: 

I wonder if it'd be worth explicitly documenting which methods can (and should) modify their inputs vs the ones that mustn't.

E.g. something like

```
# The input string is assumed to be OK to mutate
# The input string MUST NOT be mutated
```

(Although I wonder if in the cases there the string must not be mutated we should instead `freeze` them at parsing time -- thus preventing bugs like the one in `#serialize_tag_value`)

marcotc:
@ivoanjo I documented the argument sources.

spec/datadog/tracing/distributed/trace_context_spec.rb
```diff
@@ -429,7 +436,7 @@
 
         context "{ 'key' => 'value=with=equals' }" do
           let(:tags) { 't.key:value:with:equals' }
-          it { is_expected.to eq('_dd.p.key' => 'value=with=equals') }
+          it { is_expected.to eq('_dd.p.key' => 'value:with:equals') }
         end
```
ivoanjo, line 440: I think the description for this test is now outdated -- it previously tested that we deserialized `value:with:equals` back to `value=with=equals`, but now we don't do that (and while the test has been corrected, the description and strings still mention the equals which is kinda confusing).

marcotc: Nice catch! I updated the test to reflect what it actually wants to test.

Issue #2709, opened 2023-03-21
Autoloading in Rails initializer


oskarpearson:
**Is your feature request related to a problem? Please describe.**

In Rails 7 autoloading is disabled when an application boots. So things in the `config/initalizers` folder don't support autoloading.

However, the Datadog dd-trace-rb gem Rails integration takes a `exception_controller` parameter which needs to point at the class. Something like this:

```
Datadog.configure do |c|
  c.tracing.instrument :rails, exception_controller: ErrorsController
```

To use this, we need to jump through some hoops - like wrapping the whole of the datadog initialisation in `Rails.application.reloader.to_prepare do` (which I expect won't work anyway?), or the `config.autoload_once_paths` Rails feature. Using the latter (as I read it) means we can't autoload code when ErrorsController loads.

**Describe the goal of the feature**

I'd like to be able to pass a Symbol or string as the `exception_controller` parameter, rather than the actual class.

- In sidekiq they refer to passing a symbol https://github.com/sidekiq/sidekiq/issues/4192#issuecomment-501413464 
- In ActiveAdmin they pass a string https://activeadmin.info/13-authorization-adapter.html

So the Datadog config would change to something like `c.tracing.instrument :rails, exception_controller: :ErrorsController` or potentially `c.tracing.instrument :rails, exception_controller: "ErrorsController"`

**Describe alternatives you've considered**

None

**Additional context**

https://github.com/DataDog/dd-trace-rb/blob/cb6f3a4ed44d41719570ed30ae8da19dd1f73c51/lib/datadog/tracing/contrib/action_pack/action_controller/instrumentation.rb would need to try and use `constantize` of the supplied parameter (currently it just returns false if it's not a class or module)

**How does `ddtrace` help you?**

N/A

TonyCTHsu:
👋 @oskarpearson , Thanks for sharing this. I think it make sense `exception_controller` should be a string instead of a constant. I guess the changes will not be difficult with `constantize`.

However, I would like learn more about the trouble of our current implementation with Rails 7 autoloading. 

Is it broken in some ways we overlooked? or it is working as expected but having some kind of bad developer experience?

Could you provide me a bit more description about the current behaviour and what is causing trouble? So that I can search through the project to see with there are issues with similar patterns.

jens:
For us the application fails to boot with a NameError uninitialized constant ErrorsController.
So using the exception_controller with Rails 7 doesn't work.

oskarpearson:
[...]

Issue #2611, opened 2023-02-10
Custom exception controller overriding root trace resource


timonv:
**Current behaviour**
In Rails it's possible to configure a custom exception controller using `exceptions_app` in the application configuration. When an exception is raised elsewhere, the final handler is this Rack application. Often and in our case this is a Rails controller.

In datadog the root resource name of the trace is set to the exception controller. Like this:
![image](https://user-images.githubusercontent.com/49373/218048267-9faa56a0-c941-4f55-bcda-f85ad448fa8c.png)

**Expected behaviour**
This behaviour is problematic, as it obfuscates monitors and SLOs with the wrong resource. I'd expect the original controller to be the resource instead.

![image](https://user-images.githubusercontent.com/49373/218048591-ae944bd7-c803-4947-b511-be27eda6cd00.png)

In this case that would be `StatusController#error`.

Unfortunately not with any success.

**Steps to reproduce**
1. Create an ExceptionController like:

```
class ExceptionController < ActionController::Base
...
def server_error
  render text: 'Woopsie'
end
```

2. Configure Rails to use it
```
# In application.rb
config.exceptions_app = self.routes

# In routes
get '/500', to: 'exception#server_error'

```

3. Trigger an exception elsewhere

**How does `ddtrace` help you?**
I'd expect the exception_controller config in the rails instrumentation (frankly also the autodetection) to pick this up. But it doesn't seem to work. It does set `exception_controller?` in the payload in the rack instrumentation, but I'm not sure if that's used.

I also tried some hacking around in the custom error controller like:

```
        trace = Datadog::Tracing.active_trace
        original_resource = trace.instance_variable_get('@spans').find { |s| s.resource =~ /Controller/ }&.resource
        # Does set it but doesn't seem to affect dd
        trace.resource = original_resource

        # Set it on the root span directly
        # Also doesn't seem to work
        trace.send(:root_span).resource = original_resource
```

**Environment**

* **ddtrace version:**  1.9.0
* **Configuration block (`Datadog.configure ...`):**
```
if defined?(Datadog)
  Datadog.configure do |c|
    # If we want to trace a service separately, treat it as a separate root
    # span with it's own errors we name it different from the deployed service.
    # Otherwise inherit from deployed. Additionally, if it's an interface to
    # a remote service (database, cache, api), useful to have a separate service name
    # *unless* this puts the error on the wrong root span
    #
    # This will activate auto-instrumentation for Rails
[...]

Issue #2567, opened 2023-01-18
All "unrescued" exceptions show up with the resource set to our errors_controller (rails)


danielmklein:
**Current behaviour**
When a controller action trace from an uncaught error hits Datadog, its resource will be `ErrorsController#internal_server_error`, _not_ the controller action in which it actually occurred.

**Expected behaviour**
I expect the resource in Datadog to be the _original_ controller action, _not_ the ErrorsController action -- this makes it impossible to determine where errors originally came from.

In full transparency, I'm not super confident this is a _bug_ but maybe a misconfiguration on our part or a weird interaction with some other gem or middleware.

**Steps to reproduce**
Suppose we have the following simplistic example:
```ruby
# in a controller:
def my_controller_action
  raise StandardError("I've raised an error that won't be rescued by anything")
end
```

Furthermore, we have `ErrorsController` defined to handle `:internal_server_error` via `config.exceptions_app = routes` in `application.rb` and `match "/500", to: "errors#internal_server_error", via: :all` in our routes file.

**How does `ddtrace` help you?**
We use `ddtrace` in all of our Rails apps, and it's generally pretty great!

**Environment**

* **ddtrace version:** 1.8
* **Configuration block (`Datadog.configure ...`):**
```ruby
Datadog.configure do |c|
  c.tracing.enabled &= Rails.env.in? %w[production staging]

  # log injection is true by default as of ddtrace v1.0.0
  c.tracing.instrument :rails
  c.tracing.instrument :active_record, service_name: "core-postgres"

  # ... snip unrelated stuff
end
```
* **Ruby version:** 3.1.3
* **Operating system:** Alpine linux 3.16
* **Relevant library versions:** we also have version 3.4.0 of the `rollbar` gem installed, which adds its own error reporting middleware. This _could_ be interacting with ddtrace somehow, but I'm not sure.

TonyCTHsu:
👋 @danielmklein , Thanks for reporting. Could you provide your version for Rails and Rack? I would also be interested with your middleware stack looks like.

danielmklein:
@TonyCTHsu thanks for the quick response!

We are on rails 6.1.7.1 and rack 2.2.6.2.

Here's our middleware stack from `rails middleware`:
```
use Rack::Cors
use Datadog::Tracing::Contrib::Rack::TraceMiddleware
use Rack::MiniProfiler
use ActionDispatch::HostAuthorization
use ActionDispatch::SSL
use Rack::Sendfile
[...]

PR #2726, opened 2023-03-28
Improve controller instrumentation and deprecate option `exception_controller`

TonyCTHsu:
### Why?

This PR addresses issues (https://github.com/DataDog/dd-trace-rb/issues/2709 , https://github.com/DataDog/dd-trace-rb/issues/2611 , https://github.com/DataDog/dd-trace-rb/issues/2567) we've had for instrumenting an exception handling controller.

1. The trace resource being overwritten by the exception handling controller, make it hard understand the origin of error.
2. Configuration for `exception_controller` with constant does not work properly with zeitwerk autoloading.

----

### What? 

After careful examination of the problem, I have concluded that the current behaviour of option `exception_controller` is bringing more harm than benefit, because of the fundamental problem to answer the question about what is the strategy to decide whether to overwrite trace resource?

🚫  It should not be made by based on whether the current controller is handling the request being defined as an exceptional controller, since a request could be sent to the path directly (such case: overwrite trace resource).
✅ **The decision should be made by whether an exception being redirected by `ActionDispatch::ShowException`.** 

The changes are made to depend on the default behaviour for `ActionDispatch::ShowException` and make the option `exception_controller` obsolete. We decided to deprecate the option and no longer required to solve the problem with zeitwerk autoloading.

**We recommend our users to remove `exception_controller`  option from the configuration.**

----

### Snapshot of changes

Screenshot 2023-03-28 at 11 42 21

### DIFF
--- a/lib/datadog/tracing/contrib/action_pack/configuration/settings.rb
+++ b/lib/datadog/tracing/contrib/action_pack/configuration/settings.rb
@@ -1,6 +1,8 @@
 require_relative '../../configuration/settings'
 require_relative '../ext'
 
+require_relative '../../../../core'
+
 module Datadog
   module Tracing
     module Contrib
@@ -24,7 +26,18 @@ module Datadog
               o.lazy
             end
 
-            option :exception_controller
+            # DEV-2.0: Breaking changes for removal.
+            option :exception_controller do |o|
+              o.on_set do |value|
+                if value
+                  Datadog::Core.log_deprecation do
+                    "Option `#{o.instance_variable_get(:@name)}` has been deprecated and will be removed, "\
+                    'the value provided is no longer effective.'
+                  end
+                end
+              end
+            end
+
             option :service_name
           end
         end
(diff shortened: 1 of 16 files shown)


lib/datadog/tracing/contrib/action_pack/configuration/settings.rb
```diff
@@ -24,7 +26,18 @@ class Settings < Contrib::Configuration::Settings
...
             end
 
-            option :exception_controller
+            # DEV-2.0: Breaking changes for removal.
+            option :exception_controller do |o|
+              o.on_set do |value|
+                if value
+                  Datadog::Core.log_deprecation do
+                    "Option `#{o.instance_variable_get(:@name)}` has been deprecated and will be removed, "\
+                    'the value provided is no longer effective.'
```
marcotc, line 35: It's good to mention why it's being deprecated ("what will happen if I remove `:exception_controller`?"):
```suggestion
                    "Option `#{o.instance_variable_get(:@name)}` is no longer required to will be removed."
```

marcotc (approved):

PR #2727, opened 2023-03-28
Implement Span Attribute Schema Environment Variable

zarirhamza:
**What does this PR do?**

For future modifications to the repo, we will depend on the usage of a schema version to enable certain features. This includes enabling the default `service_name` to `DD_SERVICE`. This repo adds the following:

- `DD_TRACE_SPAN_ATTRIBUTE_SCHEMA` env variable to enable future service name changes
- Adds new v1 service naming functionality in the AWS integration as a first integration test by changing service name defaults
- Adds unit testing for all new changes


lib/datadog/tracing/configuration/settings.rb
```diff
@@ -437,6 +437,15 @@ def self.extended(base)
                 o.default { env_to_int(Tracing::Configuration::Ext::Distributed::ENV_X_DATADOG_TAGS_MAX_LENGTH, 512) }
                 o.lazy
               end
+
+              # Schema version for span attributes that enables various features
+              #
+              # @default `DD_TRACE_SPAN_ATTRIBUTE_SCHEMA` environment variable, otherwise `v0`
+              # @return [String,nil]
```
marcotc, line 444: I don't think we need to allow empty values here:
```suggestion
              # @return [String]
```

lib/datadog/tracing/contrib/aws/instrumentation.rb
```diff
@@ -37,8 +37,10 @@ def annotate!(span, context)
             span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_COMMAND)
 
             # Tag as an external peer service
-            span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE, span.service)
-            span.set_tag(Tracing::Metadata::Ext::TAG_PEER_HOSTNAME, context.safely(:host))
+            if Datadog.configuration.tracing.span_attribute_schema == 'v0'
```
marcotc, line 40: The string `'v0'` is being repeated many times, I recommend putting it into a constant.

lib/datadog/tracing/contrib/aws/configuration/settings.rb
```diff
@@ -25,7 +25,15 @@ class Settings < Contrib::Configuration::Settings
...
             option :service_name do |o|
-              o.default { ENV.fetch(Ext::ENV_SERVICE_NAME, Ext::DEFAULT_PEER_SERVICE_NAME) }
+              o.default do
+                ENV.fetch(Ext::ENV_SERVICE_NAME) do
+                  if Datadog.configuration.tracing.span_attribute_schema == 'v1'
+                    Datadog.configuration.service
+                  else
+                    Ext::DEFAULT_PEER_SERVICE_NAME
+                  end
+                end
```
marcotc, line 35: Because this is going to be repeated over and over again, I recommend abstracting this behaviour, something like:
```ruby
  ENV.fetch(env) do
    if Datadog.configuration.tracing.span_attribute_schema == 'v1'
      Datadog.configuration.service
    else
      default
    end
  end
end
```

TonyCTHsu: > Because this is going to be repeated over and over again, I recommend abstracting this behaviour, something like:
> 
> ```ruby
> def fetch_service_name(env, default)
>   ENV.fetch(env) do
>     if Datadog.configuration.tracing.span_attribute_schema == 'v1'
>       Datadog.configuration.service
>     else
>       default
>     end
>   end
> end
> ```

I am expecting this pattern would be repeated in multiple integrations, I recommend extract and abstract the stateless function to be utilities being reusable.

lib/datadog/tracing/contrib/aws/instrumentation.rb
TonyCTHsu: The conditional for checking schema is also expected to be repeated by multiple integrations, I recommend the extract this.

lib/datadog/tracing/contrib/aws/instrumentation.rb
```diff
@@ -37,8 +38,10 @@ def annotate!(span, context)
             span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_COMMAND)
 
             # Tag as an external peer service
-            span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE, span.service)
-            span.set_tag(Tracing::Metadata::Ext::TAG_PEER_HOSTNAME, context.safely(:host))
+            if Contrib::SpanAttributeSchema.default_span_attribute_schema?
+              span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE, span.service)
+              span.set_tag(Tracing::Metadata::Ext::TAG_PEER_HOSTNAME, context.safely(:host))
+            end
```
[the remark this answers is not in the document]
zarirhamza: No for v1, we will definitely set `peer.service`, it's just that the details aren't flushed out yet completely, so I didn't implement them yet. The goal is for the first pass through all integrations is to change service_name while setting up for the changes on the second pass with `peer.service`. Anything that isn't finalized yet (like `peer.service`, operation-name) will be done in the second pass

spec/datadog/tracing/contrib/aws/instrumentation_spec.rb
```diff
@@ -62,6 +64,8 @@
         expect(span.get_tag(Datadog::Tracing::Metadata::Ext::TAG_COMPONENT)).to eq('aws')
         expect(span.get_tag(Datadog::Tracing::Metadata::Ext::TAG_OPERATION))
           .to eq('command')
+
+        # if enabled
```
TonyCTHsu, line 68: The comment is ambiguous, consider remove it.

spec/datadog/tracing/contrib/v1_schema_examples.rb
```diff
@@ -0,0 +1,22 @@
+RSpec.shared_examples 'v1 schema test' do
```
TonyCTHsu, line 1: The shared_examples is fine right now, might need to tweak a bit in the future for wider adoption.

lib/datadog/tracing/contrib/aws/configuration/settings.rb
```diff
@@ -25,7 +26,12 @@ class Settings < Contrib::Configuration::Settings
             end
 
             option :service_name do |o|
-              o.default { ENV.fetch(Ext::ENV_SERVICE_NAME, Ext::DEFAULT_PEER_SERVICE_NAME) }
+              o.default do
```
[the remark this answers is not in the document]
zarirhamza: Done

lib/datadog/tracing/contrib/span_attribute_schema.rb
```diff
@@ -0,0 +1,25 @@
+module Datadog
+  module Tracing
+    module Contrib
+      # Contains methods for fetching values according to span attributes schema
+      module SpanAttributeSchema
+        module_function
+
+        def fetch_service_name(env, default)
+          ENV.fetch(env) do
+            if Datadog.configuration.tracing.span_attribute_schema == 'v1'
```
marcotc, line 10: I recommend moving `v1` to a constant, alongside `SpanAttributeSchema::ENV_DEFAULT_SCHEMA` given it represents a value the is exposed to the outside world.

marcotc (approved):

.rubocop_todo.yml
```diff
@@ -680,6 +680,7 @@ Style/FrozenStringLiteralComment:
     - 'lib/datadog/tracing/contrib/sinatra/patcher.rb'
     - 'lib/datadog/tracing/contrib/sinatra/tracer.rb'
     - 'lib/datadog/tracing/contrib/sinatra/tracer_middleware.rb'
+    - 'lib/datadog/tracing/contrib/span_attribute_schema.rb'
```
marcotc, line 683: We should address the https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/FrozenStringLiteralComment issue and remove this line from `.rubocop_todo.yml`.

Removing this line and running `bundle exec rubocop -A` will address this specific issue.

TonyCTHsu (approved):

marcotc (approved):

PR #2170, opened 2022-07-20
Add internal tracing to Sidekiq / DelayedJob

jennchenn:
**What does this PR do?**

This PR adds additional instrumentation to Sidekiq and DelayedJob to wraps spans from Redis/ActiveRecord to prevent them from appearing as top-level spans.

image
image
image
image

**Motivation**

Addresses #2056 

**Additional Notes**

**How to test the change?**

Additional unit tests were added to verify the changes.

PR #2728, opened 2023-03-29
Patch the ❤ method for Sidekiq 7 compatibility

madleech:
**What does this PR do?**

Fixes `sidekiq.heartbeat` traces in Sidekiq 7.

In Sidekiq 6.5.8, the [heartbeat code](https://github.com/sidekiq/sidekiq/blob/v6.5.8/lib/sidekiq/launcher.rb#L73-L79) looks like this:
```ruby
    def start_heartbeat
      loop do
        heartbeat
        sleep BEAT_PAUSE
      end
      logger.info("Heartbeat stopping...")
    end
```

While in Sidekiq 7.0.7 [it is this](https://github.com/sidekiq/sidekiq/blob/v6.5.8/lib/sidekiq/launcher.rb#L73-L79):
```ruby
    def start_heartbeat
      loop do
        beat
        sleep BEAT_PAUSE
      end
      logger.info("Heartbeat stopping...")
    end
```

So the change is that it now calls `beat` rather than `heartbeat`, and `heartbeat` is only called at sidekiq startup.

Both these methods internally call `Sidekiq::Launcher::❤` (isn't unicode great) which does the actual beating logic.

Previously, ddtrace was wrapping the `❤` method so it would work on Sidekiq 6 or 7, in PR #2170 this was changed to wrap the `heartbeat`.

Here you can see the `sidekiq.heartbeat` traces stop after upgrading to Sidekiq 7:
image

**Motivation**
We have monitors that watch for `sidekiq.heartbeat` and now they are failing due to the missing traces.

**Additional Notes**

**How to test the change?**

When running on Sidekiq 7, the heartbeat event gets sent through.

### DIFF
--- a/lib/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat.rb
+++ b/lib/datadog/tracing/contrib/sidekiq/server_internal_tracer/heartbeat.rb
@@ -27,7 +27,7 @@ module Datadog
 
             private
 
-            def heartbeat
+            def ❤ # rubocop:disable Naming/AsciiIdentifiers, Naming/MethodName
               configuration = Datadog.configuration.tracing[:sidekiq]
 
               Datadog::Tracing.trace(Ext::SPAN_HEARTBEAT, service: configuration[:service_name]) do |span|


ioquatix:
@ivoanjo any chance we can get someone to take a look at merging this sooner rather than later?

ioquatix:
cc @mperham

ivoanjo:
> @ivoanjo any chance we can get someone to take a look at merging this sooner rather than later?

Yeap, I'll make sure this gets picked up, thanks for the ping!

ioquatix:
Thanks @ivoanjo you are awesome!

TonyCTHsu:
👋 @madleech , Thanks for PR! 
 
 I have been working on supporting Sidekiq 7 on another branch and [this branch](https://github.com/DataDog/dd-trace-rb/pull/2387), but I have not find the time to complete it. 
 
 Yes, I am aware of that `heartbeat` change to `beat` in Sidekiq 7. However, the changes do not seem to be backward compatible with the previous versions of Sidekiq.

TonyCTHsu:
Closing this and keep track on https://github.com/DataDog/dd-trace-rb/pull/2810

PR #2730, opened 2023-03-29
Consistent APM Span tags for AWS SDK Requests

zARODz11z:
Adds 
```
`aws_service` **(added to all sdk request spans)**
`region`  **(added to all sdk request spans)**
<resourcename> such as `queuename`
`aws_account` **(where possible i.e. if I was able to extract something meaningful like an arn of the targeted resource then I could string manipulate the account id out)**
```

### Motivation

**attempted to add `aws-account` but no luck with all services as the aws-sdk params change with each type of request. I'd be interested in hearing thoughts on this.**
Serverless inferred related resources, pivot from traces to aws integration metrics ootb, see this doc `Consistent APM span tags for AWS requests` (search this document in Google Cloud Search)

### Testing
 Unit tests with stubbed aws-sdk requests


spec/datadog/tracing/contrib/aws/instrumentation_spec.rb
```diff
@@ -74,7 +74,7 @@
     end
   end
 
-  context 'with an S3 client' do
+  context 'with an S3 client' if: RUBY_VERSION >= '2.2.0' do
```
TonyCTHsu, line 77: Unfortunately, there are flaky specs for aws 2.x need to be addressed. 

I was able to reproduce it after changing dependency version with

`bundle exec appraisal ruby-3.1.2-contrib rspec spec/datadog/tracing/contrib/aws/instrumentation_spec.rb spec/datadog/tracing/contrib/aws/parsed_context_spec.rb --seed 1331`

zARODz11z: @TonyCTHsu 
Im not sure I understand, yes the tests were failing on ruby 3.1.2 but according to this[ aws doc](https://aws.amazon.com/blogs/developer/announcing-the-end-of-support-for-ruby-runtimes-1-9-2-0-2-1-and-2-2-for-the-aws-sdk-for-ruby/#:~:text=Ruby%201.9.3%20%E2%80%93%20EOL,2018%2D03%2D31)
, ruby 2.1 and 2.2 are not supported in the aws sdk anymore thus the reason why I added `if: RUBY_VERSION >= '2.2.0'`. These tests use the aws sdk thus I added the check to only run them if the ruby version is 2.3.0 or greater. 

All checks passed in CI now

zARODz11z: please correct me if i'm wrong but I believe this does have to do with older rubies not being compatible with the aws-sdk thus for the unit tests we should accept 2.2.0 and greater

TonyCTHsu: >I believe this does have to do with older rubies not being compatible with the aws-sdk

No, it is due to the fact that other ruby versions appraisal groups have aws 3.x installed as dependency. You can verified the behaviour by installing aws 2.x

zARODz11z:
updated gem files as I believe that may have been why the tests were failing

TonyCTHsu (changes requested):
As I mentioned, we would like to address the test failures for aws-sdk 2.x

lib/datadog/tracing/contrib/aws/instrumentation.rb
```diff
@@ -60,6 +64,96 @@ def annotate!(span, context)
...
+            span.set_tag(Ext::TAG_TABLE_NAME, table_name)
+          end
+
+          def add_kinesis_tags(span, params)
+            stream_arn = params.fetch(:stream_arn, nil)
+            stream_name = params.fetch(:stream_name, '')
+            if stream_arn
+              # example stream_arn: arn:aws:kinesis:us-east-1:123456789012:stream/my-stream
+              stream_name = stream_arn.split('/')[-1] rescue ''
+              aws_account = stream_arn.split(':')[-2] rescue ''
```
marcotc, line 121: When do you expect these to fail? I ask because of the `rescue ''` here.

lib/datadog/tracing/contrib/aws/instrumentation.rb
```diff
@@ -60,6 +64,96 @@ def annotate!(span, context)
...
+            when 'states'
+              add_states_tags(span, params)
+            when 's3'
+              add_s3_tags(span, params)
+            end
+          end
+
+          def add_sqs_tags(span, params)
+            queue_url = params.fetch(:queue_url, nil)
+            queue_name = params.fetch(:queue_name, nil)
```
marcotc, line 88: `anything.fetch(key, nil)` can be replaced with `anything[key]`, which returns `nil` when not found in the hash.
Give it a try, it should work 🤞

lib/datadog/tracing/contrib/aws/instrumentation.rb
```diff
@@ -60,6 +64,96 @@ def annotate!(span, context)
...
+            if queue_url
+              # example queue_url: https://sqs.sa-east-1.amazonaws.com/12345678/MyQueueName
+              queue_name = queue_url.split('/')[-1]
+              aws_account = queue_url.split('/')[-2]
+              span.set_tag(Ext::TAG_AWS_ACCOUNT, aws_account)
+            end
+            span.set_tag(Ext::TAG_QUEUE_NAME, queue_name)
```
marcotc, line 95: String operations are expensive.
We are splitting `queue_url` twice here. We should only split it once then collect the result.

Also, I recommend being more strict with this splitting, given invalid URLs are unlikely to happen here. And if they do, we probably want to skip tagging.

```suggestion
            if queue_url
              _, _, _, aws_account, queue_name = queue_url.split('/')
              if aws_account && queue_name
                span.set_tag(Ext::TAG_AWS_ACCOUNT, aws_account)
                span.set_tag(Ext::TAG_QUEUE_NAME, queue_name)
              end
            end
```

zARODz11z: I tried adding your suggested change but it would incorrectly add the queue_name and aws_account when given an invalid queue_url. I wrote a unit test for this case. I ended up just doing a regex match

lib/datadog/tracing/contrib/aws/instrumentation.rb
zARODz11z: I didn't have a specific case in mind where these would fail however I tried to write a test where I inputed an invalid stream_arn and the aws-sdk would fail the test.

zARODz11z: I think the rescues are not needed

zARODz11z:
> As I mentioned, we would like to address the test failures for aws-sdk 2.x, since it is aligned with our support policy

@TonyCTHsu I cherry picked your changes. Thank you again!

zARODz11z:
@marcotc I addressed your 1st round of comments

lib/datadog/tracing/contrib/aws/instrumentation.rb
```diff
@@ -60,6 +64,98 @@ def annotate!(span, context)
...
+
+          def add_sqs_tags(span, params)
+            queue_url = params[:queue_url]
+            queue_name = params[:queue_name]
+            # Regular expression to match the SQS queue URL pattern
+            # https://sqs.sa-east-1.amazonaws.com/12345678/MyQueueName
+            pattern = %r{https://sqs\.[^/]+\.amazonaws\.com/(\d+)/([^/]+)}
+
+            if queue_url && (match = queue_url.match(pattern))
+              aws_account, queue_name = match.captures
```
marcotc, line 94: 👋 @zARODz11z, it seems like regular expression matching is quite a bit slower String#split.

You can run this benchmark locally by dropping this code snippet into a file and running `ruby that_file.rb`:

```ruby
# Run `gem install benchmark-ips` before running this file.
require 'benchmark/ips'

queue_url = "https://sqs.us-stubbed-1.amazonaws.com/123456789012/MyQueueName"

Benchmark.ips do |x|
  x.time = 5
  x.warmup = 0.2
  
  x.report('regex') do
    if queue_url && (match = queue_url.match(%r{https://sqs\.[^/]+\.amazonaws\.com/(\d+)/([^/]+)}))
      aws_account, queue_name = match.captures
    end
  end
  x.report('split') do
    if queue_url
      _,_,_, aws_account, queue_name = queue_url.split('/')
    end
  end

  x.compare!
end
```

Result:
```
# ruby 3.2.1
Comparison:
               split:  1915116.4 i/s
               regex:  1110268.4 i/s - 1.72x  slower
```

lib/datadog/tracing/contrib/aws/instrumentation.rb
```diff
@@ -60,6 +64,98 @@ def annotate!(span, context)
...
+            span.set_tag(Ext::TAG_QUEUE_NAME, queue_name)
+          end
+
+          def add_sns_tags(span, params)
+            topic_arn = params[:topic_arn]
+            topic_name = params[:name]
+            if topic_arn
+              # example topic_arn: arn:aws:sns:us-west-2:123456789012:my-topic-name
+              topic_name = topic_arn.split(':')[-1]
+              aws_account = topic_arn.split(':')[-2]
```
marcotc, line 106: Similar benchmark can be made here to show that it's better to split the string only once:
```ruby
require 'benchmark/ips'

topic_arn = 'arn:aws:sns:us-west-2:123456789012:my-topic-name'

Benchmark.ips do |x|
  x.time = 5
  x.warmup = 0.2

  x.report('two splits') do
    topic_name = topic_arn.split(':')[-1]
    aws_account = topic_arn.split(':')[-2]
  end

  x.report('one splits') do
    _,_,_,_,aws_account, topic_name = topic_arn.split(':')
  end

  x.compare!
end
```

```
Comparison:
          one splits:  1657116.8 i/s
          two splits:   873666.9 i/s - 1.90x  slower
```

TonyCTHsu: Yeah, double split could be avoided. Perhaps

```
irb(main):015:0> *_, a, b = [1,2,3,4]
=> [1, 2, 3, 4]
irb(main):016:0> a
=> 3
irb(main):017:0> b
=> 4
```

lib/datadog/tracing/contrib/aws/instrumentation.rb
```diff
@@ -60,6 +64,98 @@ def annotate!(span, context)
...
+            state_machine_arn = params[:state_machine_arn]
+            execution_arn = params[:execution_arn]
+
+            if execution_arn
+              # 'arn:aws:states:us-east-1:123456789012:execution:example-state-machine:example-execution'
+              state_machine_name = execution_arn.split(':')[-2]
+            end
+
+            if state_machine_arn
+              # example statemachinearn: arn:aws:states:us-east-1:123456789012:stateMachine:MyStateMachine
```
TonyCTHsu, line 145: Should we be leveraging `ArnParser` from the gem?

https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sdk-core/lib/aws-sdk-core/arn_parser.rb

lib/datadog/tracing/contrib/aws/instrumentation.rb
```diff
@@ -60,6 +64,98 @@ def annotate!(span, context)
...
+            end
+            span.set_tag(Ext::TAG_TOPIC_NAME, topic_name)
+          end
+
+          def add_dynamodb_tags(span, params)
+            table_name = params[:table_name]
+            span.set_tag(Ext::TAG_TABLE_NAME, table_name)
+          end
+
+          def add_kinesis_tags(span, params)
```
TonyCTHsu, line 117: There are conditionals for each aws service, I would recommend to extract them, instead of put them all in this single file. This would give us more clarity about each one of them, especially we want to have tests with each one  of them. Particularly, verifying those conditions, for instance, when some values are `nil` or given a bad data.

zARODz11z:
@TonyCTHsu @marcotc Addressed your previous round of comments, thank you!

lib/datadog/tracing/contrib/aws/instrumentation.rb
zARODz11z: I put them in their own files and added some tests.

lib/datadog/tracing/contrib/aws/instrumentation.rb
zARODz11z: I tried using arn parser but it just made things more complex as it assumes all arns have 6 parts or at least splits that way thus causing `resourcename:stateMachine:MyStateMachine` meaning we would need to split again (not to mention the splitting they already do in the arn parser parse function. Additionally arn parser was not available on the aws sdk v2 (at least the one we were using for tests) so I would of had to check if the gem was defined dynamically and use a custom one if not. 
TLDR: Its seems to be for the better to use our own string parser regardless

lib/datadog/tracing/contrib/aws/services/dynamodb.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+require_relative '../ext'
+
+def add_dynamodb_tags(span, params)
```
marcotc, line 5: Each of these service-specific files would look like:
```ruby
module Datadog
  module Tracing
    module Contrib
      module Aws
        module Service
          class DynamoDB < Base
            def add_tags(span, params)
              table_name = params[:table_name]
              span.set_tag(Aws::Ext::TAG_TABLE_NAME, table_name)
            end
          end
        end 
      end 
    end 
  end 
end
```

You'd also create a base implemented, just in case case we want a fallback behaviour (in the `base.rb` file):
```ruby
module Datadog
  module Tracing
    module Contrib
      module Aws
        module Service
          class Base
            def add_tags(span, params)
            end
          end
        end
      end
    end
  end
end
```

lib/datadog/tracing/contrib/aws/instrumentation.rb
```diff
@@ -33,6 +34,10 @@ def annotate!(span, context)
             span.span_type = Tracing::Metadata::Ext::HTTP::TYPE_OUTBOUND
             span.name = Ext::SPAN_COMMAND
             span.resource = context.safely(:resource)
+            aws_service = span.resource.split('.')[0]
+            span.set_tag(Ext::TAG_AWS_SERVICE, aws_service)
+            params = context.safely(:params)
+            add_service_specific_tags(span, aws_service, params)
```
marcotc, line 40: Here it would look similar to:
```ruby
if (handler = Aws::SERVICE_HANDLER[aws_service])
  handler.add_tags(span, params)
end
```

lib/datadog/tracing/contrib/aws/services/helpers.rb
```diff
@@ -0,0 +1,28 @@
...
+    add_dynamodb_tags(span, params)
+  when 'kinesis'
+    add_kinesis_tags(span, params)
+  when 'eventbridge'
+    add_eventbridge_tags(span, params)
+  when 'states'
+    add_states_tags(span, params)
+  when 's3'
+    add_s3_tags(span, params)
+  end
```
marcotc, line 27: This block would be removed, and effectively replaced by a new table `SERVICE_HANDLERS`, added to the existing `services.rb` file:
```ruby
module Datadog
  module Tracing
    module Contrib
      module Aws
        SERVICES = %w[
          # ...
        ].freeze

        SERVICE_HANDLERS = {
          'sqs' => Service::SQS.new,
          'sns' => Service::SNS.new,
          # ...
        }
      end
    end
  end
end
```

lib/datadog/tracing/contrib/aws/services/dynamodb.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
```
marcotc, line 1: For namespacing, in Ruby, it's standard practice to use singular names for folders, so I recommend renaming this directory from `aws/services` to `aws/service`.

lib/datadog/tracing/contrib/aws/services/kinesis.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+require_relative '../ext'
+
+def add_kinesis_tags(span, params)
```
marcotc, line 5: As a side-point, immaterial after the proposed changes: a method declaration like this, at the top of the file, not inside any `class` or `module` creates a global function in the current process.
This means, for example, `add_kinesis_tags` is globally available for the client to invoke, from pretty much anywhere in their code base. This not desirable for a library like ours to do because it pollutes the application naming space.
In the worst case scenario, the user has a `add_kinesis_tags` method already and we just overwrote it here.

marcotc (changes requested):
The change to individual files looks good, and also the testing being split on a per-aws service basis.

I left a suggestion about how to structure this new code refactor so it more seamlessly integrate with our library code base.

marcotc (approved):

PR #2731, opened 2023-03-30
Harden remote worker against internal exceptions

lloeki:
**What does this PR do?**

Add a pokémon exception handler so that the worker does not die.

**Motivation**

We don't want the remote worker to die on error.

**Additional Notes**

More exceptions are to be handled.

**How to test the change?**

CI


lib/datadog/core/remote/component.rb
```diff
@@ -20,7 +20,20 @@ def initialize(settings, agent_settings)
...
+          @worker = Worker.new(interval: settings.remote.poll_interval_seconds) do
+            begin
+              @client.sync
+            rescue StandardError => e
+              Datadog.logger.error do
+                "remote worker error: #{e.class.name} #{e.message} location: #{Array(e.backtrace).first}"
+              end
+
+              # client state is unknown, state might be corrupted
+              @client = Client.new(transport_v7)
```
GustavoCaso, line 32: Why don't we add a method on the client that resets the internal state? 

`@client.reset` will make sure repository is a branch new after calling `reset`

spec/datadog/core/remote/component_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      end
+
+      it 'logs an error' do
+        allow(Datadog::Core::Remote::Client).to receive(:new).and_return(client)
+
+        expect(Datadog.logger).to receive(:error).and_return(nil)
+
+        Thread.pass # allow worker thread to work
+        sleep 0.1
+      end
```
GustavoCaso, line 60: Doing expectations on `Datadog.logger` is not very valuable, in my opinion. They cause the test suite to be more brittle since if we remove the logline, we have to remove the test. I will focus on testing the actual behaviours. Ex:  `creates a new client`

spec/datadog/core/remote/component_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+
+      it 'catches exceptions' do
+        allow(Datadog::Core::Remote::Client).to receive(:new).and_return(client)
+
+        # if the error is uncaught it will crash the test, so a mere passing is good
+
+        Thread.pass # allow worker thread to work
+        sleep 0.1
```
GustavoCaso, line 68: Do we still need the sleep after thread pass if we are not doing any expectations?

spec/datadog/core/remote/component_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      end
+
+      it 'creates a new client' do
+        expect(Datadog::Core::Remote::Client).to receive(:new).and_return(second_client)
+
+        expect(component.client.object_id).to eql(client.object_id)
+
+        Thread.pass # allow worker thread to work
+        sleep 0.1
```
GustavoCaso, line 77: We could do better and remove the thread pass and sleep. We could extract the code that passes the block to the worker on the component to a private method. In the test, we can override that method to still calls client sync, but use a queue to make sure the tests  waits until we know we are in the correct state.


```ruby
# remote/component.rb

Worker.new(interval: settings.remote.poll_interval_seconds) { sync_client }

private 

def sync_client
  @client.sync
end
```

```ruby
it 'creates a new client' do
  queue = Queue.new
  expect(Datadog::Core::Remote::Client).to receive(:new).and_return(second_client)
  expect(component.client.object_id).to eql(client.object_id)

  expect(component).to receive(:sync_client).and_wrap_original do |method, arg|
    method.call(arg)
    queue << 1
  end

  queue.pop
  expect(component.client.object_id).to eql(second_client.object_id)
end
```

spec/datadog/core/remote/component_spec.rb
lloeki: The goal of this test is precisely to test that an error-level log is emitted on exception catching.

spec/datadog/core/remote/component_spec.rb
lloeki: My plan is to replace the `Thread.pass` and `sleep` with an effectively working `barrier(:once)`, which will effectively wait until the worker has executed once.

lib/datadog/core/remote/component.rb
lloeki: Such resets should be avoided at all costs as they provide a dangerous public API, especially in concurrent contexts.

In this case it would mean:
- changing the client id, which conceptually makes no sense: the instance immediately has an id for its whole lifetime
- it would require resetting the repository and any other internal state

There is already a facility for that: `new` creates an exactly new client.

IOW it would change the class from encapsulating the notion of being a specific client to being a global singleton in disguise.

lloeki: Sidenote: it also points to the client's dispatcher having receivers be registered to not be encapsulated in the client but done by the component, something I anticipated to be the case even before this change.

spec/datadog/core/remote/component_spec.rb
lloeki: Sadly it seems like so, `Thread.pass` does not guarantee the other thread will execute, nor that it won't go back to this thread. So sometimes the test would flakily fail.

As noted in the comment, the no-exception expectation is implicit and relies on `sync` being actually called, which is asserted by the `expect(client).to receive(:sync)` part just above.

The proper fix is to use a working `barrier(:once)`

spec/datadog/core/remote/component_spec.rb
GustavoCaso: I still think that testing this kind of behaviour doesn't yield much benefits

spec/datadog/core/remote/component_spec.rb
GustavoCaso: `The proper fix is to use a working barrier(:once)` What is stopping us from doing that work on this PR?

spec/datadog/core/remote/component_spec.rb
GustavoCaso: If you plan on replacing it with a `barrier(:once)` I would suggest to work on that on this PR

lloeki:
Rebased!

lib/datadog/core/remote/component.rb
```diff
@@ -23,27 +23,71 @@ def initialize(settings, agent_settings)
...
+          end
+
+          def lift
+            @mutex.synchronize do
+              @once ||= true
+
+              @condition.broadcast
+            end
+          end
+        end
```
ivoanjo, line 89: Hey! I'm always curious when concurrency stuff shows up. May I suggest adding a few comments explaining what this is and how it's used?

In particular, I'm curious about the intended semantics, with e.g. calling the block outside of the lock, and checking `@once` outside as well, but having the condition variable and setting `@once` inside the lock.

lloeki: I will! And also figure out some better naming.

For context: I was trying to get the implementation right + a bunch of passing specs before diving into comments.

spec/datadog/core/remote/component_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+      end
+    end
+  end
+end
+
+RSpec.describe Datadog::Core::Remote::Component::Barrier do
+  let(:delay) { 0.5 }
+  let(:block) { -> { true } }
+
+  subject(:barrier) { described_class.new(&block) }
```
GustavoCaso, line 85: Do we still need the `block` portion of the initialise?

spec/datadog/core/remote/component_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+      Thread.new do
+        loop do
+          record << :wait
+          barrier.wait_next
+        end
+      end
+    end
+
+    before do
+      record
```
GustavoCaso, line 104: What is `record` doing in this cases?

spec/datadog/core/remote/component_spec.rb
```diff
@@ -0,0 +1,214 @@
...
+
+        expect(record).to eq [:one, :two]
+      end
+    end
+
+    context 'with waiters' do
+      include_context 'waiter thread'
+
+      it 'unblocks waiters' do
+        sleep delay
```
GustavoCaso, line 155: @lloeki, we talked about it. Just leaving it here for reference.  It would be good to try using `Queue` rather than sleep. Since the intention of the sleep if to force the VM to switch to another thread, we could use `queue.pop` to wait until the other thread unblock us using `queue << 1`

spec/datadog/core/remote/component_spec.rb
lloeki: Huh, correct. I'm surprised it does not crash the spec.

spec/datadog/core/remote/component_spec.rb
lloeki: Just creating the array. I guess it's unneeded really.

lib/datadog/core/remote/component.rb
lloeki: > Thread 1 finally calls wait_next

You mean, because it it scheduled back in, resuming `wait_once` after bullet point 1?

> Can it get stuck there waiting for a lift that already happened?

Well if there's only one `lift`, yes. But the expectation is that `lift` is happening in a loop and gets called regularly. So worst case it gets next `lift`, which means `wait_once` waited for a `lift` _at least_ once (not _exactly_ once).

Practically, we're using `wait_once` so that we wait for a RC sync to happen at least once before proceeding.

Same goes for `wait_next`, it is meant to wait for a subsequent `lift`, not necessarily the _immediate_ next one (which is a fuzzy thing to define without additional synchronization in the Barrier caller).

That said, I'm fine with doing this instead:

```
          def wait_once
            @mutex.synchronize do
              return if @once

              @condition.wait(@mutex)
            end
          end
```

Which should cover your concern at that level.

spec/datadog/core/remote/component_spec.rb
lloeki: Done

spec/datadog/core/remote/component_spec.rb
lloeki: Done

lib/datadog/core/remote/component.rb
lloeki: Done.

spec/datadog/core/remote/component_spec.rb
lloeki: Done

spec/datadog/core/remote/component_spec.rb
lloeki: Removed

lib/datadog/core/remote/component.rb
```diff
@@ -22,26 +22,80 @@ def initialize(settings, agent_settings)
...
+                "remote worker error: #{e.class.name} #{e.message} location: #{Array(e.backtrace).first}"
+              end
+
+              # client state is unknown, state might be corrupted
+              @client = Client.new(transport_v7, capabilities)
+
+              # TODO: bail out if too many errors?
+            end
+
+            @barrier.lift
```
GustavoCaso, line 42: Just thinking out loud. Should the lift be just right after `@client.sync`? 


Imagine this scenario. In the rack middleware, we want to wait for the first sync using `Datadog::Core::Remote.active_remote.barrier(:once)`

Imagine we hit `@client.sync` and something fails. We are going to rescue the exception and call `@barrier.lift` 
We are going to continue without having sync against remote configuration. 

I can understand we don't want that because if the error is not transient and is not recovered, we would hold that request indefinitely. 

Just want to leave this train of thought here

spec/datadog/core/remote/component_spec.rb
GustavoCaso: That is a very fair point. 

Out of curiosity, I tried to make it work with `Queue`

Here is the diff, the tests are passing 
image

GustavoCaso (approved):
LGTM. Left a couple comments but they are not blockers by any means

lib/datadog/core/remote/component.rb
ivoanjo: > Well if there's only one lift, yes. But the expectation is that lift is happening in a loop and gets called regularly. So worst case it gets next lift, which means wait_once waited for a lift at least once (not exactly once).

I think we're really missing some higher-level documentation on how this is intended to get used and all of the assumptions being made. This expectation you mentioned was not clear to me at all; and even after you mentioned it, it's still not very clear to me what "gets called regularly" means.

Looking at the code, it looks like this `barrier` gets called from a regular request thread, so does missing a wake-up mean that it can hang until the next remote configuration upgrade comes in?

Writing comments can come across as really dry so I just want to clarify that I'm trying to help :smile: Getting this mechanism wrong I think can both have high impact (e.g. disrupting user requests) as well as be very hard to debug, so I'm asking all of these things to help review this code.

spec/datadog/core/remote/component_spec.rb
```diff
@@ -0,0 +1,212 @@
...
+      end
+
+      it 'creates a new client' do
+        expect(Datadog::Core::Remote::Client).to receive(:new).and_return(second_client)
+
+        expect(component.client.object_id).to eql(client.object_id)
+
+        component.barrier(:once)
+
+        expect(component.client.object_id).to eql(second_client.object_id)
```
ivoanjo, line 75: Minor: I think this can be simplified as

```suggestion
        expect(component.client).to be client

        component.barrier(:once)

        expect(component.client).to be second_client
```

spec/datadog/core/remote/component_spec.rb
```diff
@@ -0,0 +1,212 @@
...
+RSpec.describe Datadog::Core::Remote::Component do
+  let(:settings) { Datadog::Core::Configuration::Settings.new }
+  let(:agent_settings) { Datadog::Core::Configuration::AgentSettingsResolver.call(settings, logger: nil) }
+
+  describe '#initialize' do
+    subject(:component) { described_class.new(settings, agent_settings) }
+
+    let(:transport_v7) { double }
+    let(:client) { double }
+    let(:worker) { component.instance_eval { @worker } }
```
ivoanjo, line 15: Minor: You may want to consider making the `Worker` a dependency that can be injected. This way it would make the class more testable (because you can inject a custom implementation, instead of having to "grab" the internal state of the class to get the one that one created), as well as more flexible -- in the future we could reuse the class with a different worker with zero changes!

lib/datadog/core/remote/component.rb
lloeki: > I can understand we don't want that because if the error is not transient and is not recovered, we would hold that request indefinitely.

Exactly, that's what we absolutely do _not_ want.

It's a best effort to ensure that in most cases at least one sync call has happened. Even when sync goes through the backend reply may or may not include the full expected configuration, but we _can't_ know if what we receive is what the user expects (e.g it might be legitimately empty or it might be empty because it's simply delayed by rate limiting).

spec/datadog/core/remote/component_spec.rb
ivoanjo: I am slightly worried that test correctness here is dependent on racy sleeps as well.

ivoanjo (commented):
Left a few notes on the spec.

lib/datadog/core/remote/component.rb
lloeki: > Looking at the code, it looks like this barrier gets called from a regular request thread, so does missing a wake-up mean that it can hang until the next remote configuration upgrade comes in?

With `wait_next`, yes, but the request barrier is using `wait_once`, so the blocking is only happening until there's at least one sync attempt, which happens "right away" because the RC worker is lazily-started immediately (or rather, as immediately as thread scheduling allows it)

lloeki: >  I just want to clarify that I'm trying to help

I did not take it any other way! ;)

ivoanjo: So when does `wait_next` get used instead? Is it only for testing?

ivoanjo (approved):
:+1: LGTM

PR #2732, opened 2023-03-30
[APPSEC-8867] Appsec reuse rules when ASM_DD is empty

GustavoCaso:
**What does this PR do?**

During Remote Client sync, there could be occasions when the rules information from `ASM_DD` product is not present or not available. 

In those situations, we should default to the ones defined on the appsec settings. By default is `:recommended`

**Additional Notes**

I had to extract away the parsing of the ruleset from the `AppSec::Processor`, so I could reuse it in other places in the code base. I like it because it removes that functionality from `Processor`, which is only responsible for ensuring a valid processor and context are created, with the `ruleset` provided.

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/core/remote/client.rb
+++ b/lib/datadog/core/remote/client.rb
@@ -5,6 +5,7 @@ require 'securerandom'
 require_relative 'configuration'
 require_relative 'dispatcher'
 require_relative '../../appsec/processor/rule_merger'
+require_relative '../../appsec/processor/rule_loader'
 
 module Datadog
   module Core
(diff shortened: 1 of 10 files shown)


lib/datadog/core/remote/client.rb
```diff
@@ -215,17 +199,33 @@ def register_receivers
...
+              case content.path.product
+              when 'ASM_DD'
+                rules << parse_content(content)
+              when 'ASM_DATA'
+                data << parse_content(content) if asm_data_config_types.include?(content.path.config_id)
+              when 'ASM'
+                overrides << parse_content(content) if asm_overrides_config_types.include?(content.path.config_id)
+                exclusions << parse_content(content) if content.path.config_id == 'exclusion_filters'
+              end
+            end
```
GustavoCaso, line 220: This reduces the time we iterate over `repository.contents` from 4 to 1.

lloeki (approved):
LGTM

PR #2733, opened 2023-03-31
[APPSEC-6713] Extract Remote capabilities, products and receivers to their own components

GustavoCaso:
**What does this PR do?**

Move specific logic outside of the `Core::Remote::Client` to their respective component. Right now, the only component using remote configuration is AppSec.

Created a new dependency `Core::Remote::Client::Capabilities` for the `Core::Remote::Client`.  This new dependency checks which component has to register any specific configuration for remote configuration. 

**Motivation**

In the future, when other components use remote configuration they would be able to register capabilities, products and receivers independently without polluting the  `Core::Remote::Client` code.

**Additional Notes**

I tested these changes by running locally and modifying the remote configuration using the UI

image

image

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/appsec/remote.rb
+++ b/lib/datadog/appsec/remote.rb
@@ -39,18 +39,18 @@ module Datadog
           'ASM_DATA',     # config files (IP addresses or users for blocking)
         ].freeze
 
-        def remote_capabilities
+        def capabilities
           remote_features_enabled? ? ASM_CAPABILITIES : []
         end
 
-        def remote_products
+        def products
           remote_features_enabled? ? ASM_PRODUCTS : []
         end
 
-        def remote_receivers
+        def receivers
           return [] unless remote_features_enabled?
 
-          receiver = Dispatcher::Receiver.new(Dispatcher::Matcher::Product.new(ASM_PRODUCTS)) do |repository, changes|
+          receiver = Core::Remote::Dispatcher::Receiver.new(Core::Remote::Dispatcher::Matcher::Product.new(ASM_PRODUCTS)) do |repository, changes|
             changes.each do |change|
               Datadog.logger.debug { "remote config change: '#{change.path}'" }
             end
--- a/lib/datadog/appsec/remote.rb
+++ b/lib/datadog/appsec/remote.rb
@@ -1,0 +1,116 @@
+# frozen_string_literal: true
+
+require_relative '../core/remote/dispatcher'
+require_relative 'processor/rule_merger'
+require_relative 'processor/rule_loader'
+
+module Datadog
+  module AppSec
+    module Remote
+      class ReadError < StandardError; end
+      class NoRulesError < StandardError; end
+
+      class << self
+        CAP_ASM_ACTIVATION                = 1 << 1 # Remote activation via ASM_FEATURES product
+        CAP_ASM_IP_BLOCKING               = 1 << 2 # accept IP blocking data from ASM_DATA product
+        CAP_ASM_DD_RULES                  = 1 << 3 # read ASM rules from ASM_DD product
+        CAP_ASM_EXCLUSIONS                = 1 << 4 # exclusion filters (passlist) via ASM product
+        CAP_ASM_REQUEST_BLOCKING          = 1 << 5 # can block on request info
+        CAP_ASM_RESPONSE_BLOCKING         = 1 << 6 # can block on response info
+        CAP_ASM_USER_BLOCKING             = 1 << 7 # accept user blocking data from ASM_DATA product
+        CAP_ASM_CUSTOM_RULES              = 1 << 8 # accept custom rules
+        CAP_ASM_CUSTOM_BLOCKING_RESPONSE  = 1 << 9 # supports custom http code or redirect sa blocking response
+
+        # TODO: we need to dynamically add CAP_ASM_ACTIVATION once we support it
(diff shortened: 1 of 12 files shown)


lib/datadog/core/remote/client.rb
```diff
@@ -4,14 +4,45 @@
...
+          def register_receivers(receivers)
+            @receivers.concat(receivers)
+          end
+
+          def register_products(products)
+            @products.concat(products)
+          end
+
+          def binary_capabilities
+            return @binary_capabilities if defined?(@binary_capabilities)
```
lloeki, line 37: I read recently that `defined?` can be slow.

Better use `@binary_capabilities.nil?`, and set `@binary_capabilities = nil` in `initialize`.

lib/datadog/core/remote/client.rb
```diff
@@ -4,14 +4,45 @@
...
+        class SyncError < StandardError; end
+
+        class << self
+          attr_reader :products, :capabilities, :receivers
+
+          def init
+            @capabilities = []
+            @products = []
+            @receivers = []
+          end
```
lloeki, line 22: This looks like it should be an actual class, and have an instance of it in the client.

lib/datadog/core/configuration/components.rb
```diff
@@ -149,6 +151,18 @@ def shutdown!(replacement = nil)
...
+
+        def register_remote_functionality
+          return unless remote
+
+          if appsec
+            Datadog::Core::Remote.register_capabilities(Datadog::AppSec::Remote.capabilities)
+            Datadog::Core::Remote.register_receivers(Datadog::AppSec::Remote.receivers)
+            Datadog::Core::Remote.register_products(Datadog::AppSec::Remote.products)
+          end
+        end
```
lloeki, line 165: This should be able to move to the `Client::Capabilities` instance I mentioned below, and test against the `settings` instance that gets passed to its initializer.

Rationale: we very much want to avoid static globals (and pseudo-globals such as class variables) as much as possible.

spec/datadog/core/remote/client/capabilities_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  context 'when no component enabled' do
+    let(:settings) do
+      appsec_settings = Datadog::AppSec::Configuration::Settings.new
+      dsl = Datadog::AppSec::Configuration::DSL.new
+      dsl.enabled = false
+      appsec_settings.merge(dsl)
+
+      settings = Datadog::Core::Configuration::Settings.new
+      expect(settings).to receive(:appsec).at_least(:once).and_return(appsec_settings)
```
GustavoCaso, line 22: @lloeki I believe there is some leak going on on our testing suite for `dd-trace-rb`

I added a note for me to take a look after we release the latest version, but for now, the quickest way to fix it was making sure this test state stays correct, rather than fixing the original leak

lib/datadog/core/remote/client/capabilities.rb
```diff
@@ -0,0 +1,57 @@
...
+
+          def register_receivers(receivers)
+            @receivers.concat(receivers)
+          end
+
+          def register_products(products)
+            @products.concat(products)
+          end
+
+          def compute_capabilities
```
lloeki, line 45: `to_base64` should be a better name now

lib/datadog/core/remote/client.rb
```diff
@@ -4,27 +4,32 @@
...
   module Core
     module Remote
       # Client communicates with the agent and sync remote configuration
       class Client
+        class SyncError < StandardError; end
+
         attr_reader :transport, :repository, :id, :dispatcher
 
-        def initialize(transport, repository: Configuration::Repository.new)
+        def initialize(transport, capabilities, repository: Configuration::Repository.new)
```
lloeki, line 17: Reminder that I don't quite like that repository argument: a new client should always start anew with an empty repository.

lib/datadog/appsec/remote.rb
```diff
@@ -0,0 +1,126 @@
...
+
+        def parse_content(content)
+          data = content.data.read
+
+          content.data.rewind
+
+          raise ReadError, 'EOF reached' if data.nil?
+
+          JSON.parse(data)
+        end
```
lloeki, line 122: Not to be fixed now, but I feel like this should be in `Core`, which hints me at something being amiss in terms of abstraction.

lloeki (approved):
LGTM. I feel like we'll need to do a cleanup path sometime, I feel there are leaky abstractions. But first, let's make this work.

lib/datadog/appsec/remote.rb
GustavoCaso: Agree 100%. This method was the only one that didn't feel right when extracting everything, but I couldn't figure out where to place it. 

But as you mention, let's focus getting released and then refactor the code 😄

lib/datadog/core/remote/client.rb
GustavoCaso: In production, that is the case.

But for testing is easer to provide a repository object, so when doing assertion of the repository object is more accessible. 

I think we could get rid of the `attr_reader : repository ` as I introduced for steep, but it looks like it complains about not having the accessor, but we can simply define it in the `*.rbs` file

Once we have everything in production running we can do a lot of cleanup

Issue #2721, opened 2023-03-24
New profiler breaks libgit2 ssh transport


tjk:
**Current behaviour**

Setting `profiling.advanced.force_enable_new_profiler` to `true` makes using rugged (libgit2) ssh transport break.

**Steps to reproduce**

Make sure to compile rugged with :ssh feature, and the do the following:

```
creds = Rugged::Credentials::SshKey.new(username: 'git', publickey: '/path/to/key.pub', privatekey: '/path/to/key')
Rugged::Repository.clone_at('ssh://git@github.com/<username>/<repo>.git', '/tmp/some-directory', credentials: creds)
```

You will see an error like `Failed to open SSH channel: Error waiting on socket (Rugged::SshError)`

**Environment**

* **ddtrace version:** 1.9.0
* **Configuration block (`Datadog.configure ...`):** see above
* **Ruby version:** 2.7.7
* **Operating system:** Debian Buster
* **Relevant library versions:** rugged 1.5.0.1 (means libgit2 1.5.0)

ivoanjo:
Thanks a lot @tjk for the report. This definitely looks similar to the mysql2 issue an other customer reported [(see known issues in the release notes)](https://github.com/DataDog/dd-trace-rb/releases/tag/v1.10.0).

You mentioned that 

> EDIT: While disabling the new profiler seems to fix the issue, there might be an underlying problem related to different libssl due to different libssh2 that is just heightened with this feature on... will continue investigating and report back.

Please do let us know about you find out! I'll work to reproduce this as well... For now indeed the best solution is to not use the new profiler for this application.

ivoanjo:
I was able to reproduce this easily -- thanks for the clear notes! It's indeed the same issue as [called out in the release notes for the mysql2 gem](https://github.com/DataDog/dd-trace-rb/releases/tag/v1.10.0).

I can confirm the observation that the issue seems to be triggered in the ssh transport, and when cloning via http there seems to be no issue. It's unclear to me if the issue right now is in libssh2 (used by the ssh transport) or libgit2 (I don't think it comes from `rugged` directly).

@tjk: Actually getting a fix in may take a while, since I'll need to report this to the `rugged` maintainers (and possibly the `libgit2` and `libssh2` maintainers as well).

Are you OK with not enabling the new profiler for this specific app while we work on this?

[...]

PR #2741, opened 2023-04-04
[PROF-7409] Do not auto-enable new profiler when rugged gem is detected

ivoanjo:
**What does this PR do?**:

This PR adds the `rugged` gem to the list of gems that prevent the new CPU Profiling 2.0 profiler from being auto-enabled.

This is needed because the `rugged` gem (more specifically, its C-level dependencies) do not correctly handle the unix signals that the new profiler uses.

We plan to do more than just add this workaround, but until we do so, this change will prevent customers from being bitten by the issue.

**Motivation**:

Make sure that customers being moved to the new CPU Profiling 2.0 profiler have a smooth experience.

**Additional Notes**:

N/A

**How to test the change?**:

The snippet shared by the customer in <https://github.com/DataDog/dd-trace-rb/issues/2721> reproduces the issue for me every time.

With this change, as expected, the issue no longer appears.

### DIFF
--- a/lib/datadog/profiling/component.rb
+++ b/lib/datadog/profiling/component.rb
@@ -172,8 +172,19 @@ module Datadog
             'library used by the mysql2 gem) have a bug in their signal handling code that the new profiler can trigger. ' \
             'This bug (https://bugs.mysql.com/bug.php?id=83109) is fixed in libmysqlclient versions 8.0.0 and above. ' \
             'If your Linux distribution provides a modern libmysqlclient, you can force-enable the new CPU Profiling 2.0 ' \
-            'profiler by using the `DD_PROFILING_FORCE_ENABLE_NEW` or `c.profiling.advanced.force_enable_new_profiler` ' \
-            'settings.'
+            'profiler by using the `DD_PROFILING_FORCE_ENABLE_NEW` environment variable or the ' \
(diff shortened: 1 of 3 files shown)


TonyCTHsu (approved):

lib/datadog/profiling/component.rb
```diff
@@ -172,8 +172,19 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
...
+            'profiler by using the `DD_PROFILING_FORCE_ENABLE_NEW` environment variable or the ' \
+            '`c.profiling.advanced.force_enable_new_profiler` setting.' \
+          )
+          return false
+        end
+
+        if Gem.loaded_specs['rugged']
+          Datadog.logger.warn(
+            'Falling back to legacy profiler because rugged gem is installed. Some operations on this gem are ' \
+            'currently incompatible with the new CPU Profiling 2.0 profiler, as detailed in ' \
```
marcotc, line 184: Is there a subset of the rugged gem that's safe, like an specific version? I'm asking so users know when it makes sense to try to enable profiling 2.0 anyway.

ivoanjo: Currently, there isn't. It's on my next-things-to-do list to report this upstream and even seeing if I can submit a PR with a fix or workaround, which would allow us to indeed make this finer-grained than what this PR does.

PR #2742, opened 2023-04-04
Tag known leaky Timeout internal thread

marcotc:
This PR tags `Timeout.ensure_timeout_thread_created` as a known leaky thread: this one is created internally by the `Timeout` class.

This ensures this `Timeout` thread is not reported as a leaky thread from `ddtrace`.


GustavoCaso (approved):

GustavoCaso:
Thanks 🎉

PR #2745, opened 2023-04-04
Lint:Add frozen_string_literal to files with no strings

marcotc:
This PR adds the `# frozen_string_literal: true` magic header to the top of files that do to contain any strings, thus having no side effects. (I found the list of files using [`Fast.search_file('(str)', path).empty?`](https://github.com/jonatas/fast#fastsearch_file))

This PR is part of an effort to have `Style/FrozenStringLiteralComment` enabled by default on all `ddtrace` files.

It also removes paths that have been moved or deleted from the `Style/FrozenStringLiteralComment: Exclude:` list.


ivoanjo (approved):
LGTM :+1: 

I've double checked that none of the files modified have strings using the quick command 

```
git diff-tree --no-commit-id --name-only no-strings -r | xargs grep "[%\"'<]"
```

and indeed the only hits I find are comments and things other than strings.

PR #2746, opened 2023-04-04
Lint:frozen_string_literal when only string usage is for require

marcotc:
Similar to #2745, this PR find easily fixable files by parsing source Ruby files into AST, and in this case finding all string usages that are only occurring in `require` and `require_relative` statements.

This means that there's no functional impact to freezing all strings in the effected files.

The gist used for find all files changed in this PR: https://gist.github.com/marcotc/7237b8a4cc1c602a7e166677f0469597


ivoanjo (approved):
:+1: LGTM

I've checked with `git diff-tree --no-commit-id --name-only simple-strings -r | xargs grep "[%\"'<]" | grep -v "require '" | grep -v "require_relative '"` (that indeed there's no strings other than require/require_relative in these files)

PR #2747, opened 2023-04-04
Lint:frozen_string_literal when only string usage is for Hash[]

marcotc:
Following up from #2746, this PR finds files that only use strings as keys to `Hash#[]`.

I manually inspected each file, given the `:[]` method can be overridden. All files changed are cases where the `:[]` method in question is indeed from `Hash`.

The script used to find these files is: https://gist.github.com/marcotc/7237b8a4cc1c602a7e166677f0469597


ivoanjo (approved):
:+1:

Given Ruby's duck typing it's hard to be entirely sure that the things being indexed with a string are actually instances of `Hash` or just things that implement a similar API, but after inspecting the individual files and how they're used I think the risk is very very low so here's my approval :)

PR #2754, opened 2023-04-05
Internal Schema Tag

zarirhamza:
**What does this PR do?**
WIP


lib/datadog/tracing/span_attribute_schema.rb
```diff
@@ -0,0 +1,37 @@
...
+            default
+          end
+        end
+      end
+
+      def default_span_attribute_schema?
+        Datadog.configuration.tracing.span_attribute_schema == Ext::DEFAULT_VERSION
+      end
+
+      def get_schema_version_numeric(version = Ext::DEFAULT_VERSION)
```
marcotc, line 28: The pattern in Ruby is to not have `get_` as a prefix, but instead simply have the method name `schema_version_numeric`.
The equivalent `set_` would be `def schema_version_numeric=()`.

lib/datadog/tracing/component.rb
```diff
@@ -29,6 +29,9 @@ def build_tracer(settings, agent_settings)
 
         subscribe_to_writer_events!(writer, sampler, settings.tracing.test_mode.enabled)
 
+        current_schema_version = SpanAttributeSchema
+                                   .get_schema_version_numeric(Datadog.configuration.tracing.span_attribute_schema)
```
marcotc, line 33: It's not safe to call `Datadog.configuration` inside this method, as the configuration itself is being modified when `Component`s are being build.

To address that, this method receives the configuration in the `settings` argument:
```suggestion
                                   .get_schema_version_numeric(settings.tracing.span_attribute_schema)
```

lib/datadog/tracing/contrib/aws/instrumentation.rb
```diff
@@ -38,7 +38,7 @@ def annotate!(span, context)
             span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_COMMAND)
 
             # Tag as an external peer service
-            if Contrib::SpanAttributeSchema.default_span_attribute_schema?
+            if SpanAttributeSchema.default_span_attribute_schema?
               span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE, span.service)
               span.set_tag(Tracing::Metadata::Ext::TAG_PEER_HOSTNAME, context.safely(:host))
```
amarziali, line 43: Why PEER_HOSTNAME is not sent on v1?

zarirhamza: This is the old `peer.service` tag which is being contained solely in v0. In the future, we will set the new tag in v1

PR #2755, opened 2023-04-05
Set libddwaf timeout to a high value for AppSec specs

lloeki:
**What does this PR do?**

Set libddwaf timeout to a high value for AppSec specs

**Motivation**

This removes a frequent cause of flakiness

**Additional Notes**

For these test cases we want libddwaf to always run to completion. 10s per request should be enough!

**How to test the change?**

CI should be green. Hopefully it removes most of the flakiness, which can only be observed over time.


ivoanjo (approved):
:+1: LGTM

TonyCTHsu (approved):

PR #2756, opened 2023-04-05
Check info endpoint for remote configuration availability in the agent

lloeki:
**What does this PR do?**

Check info endpoint for remote configuration availability in the agent

**Motivation**

RC should not even start if the agent is unreachable or not configured properly.

**Additional Notes**

This should also warn the user so that corrective action can be taken self-served.

**How to test the change?**

Either one of:

- Do not start agent
- Start agent without `DD_REMOTE_CONFIGURATION_ENABLED=false` env var

And then start an app with `DD_REMOTE_CONFIGURATION_ENABLED=true`

### DIFF
--- a/lib/datadog/core/remote/component.rb
+++ b/lib/datadog/core/remote/component.rb
@@ -20,28 +20,6 @@ module Datadog
           transport_options = {}
           transport_options[:agent_settings] = agent_settings if agent_settings
 
-          transport_root = Datadog::Core::Transport::HTTP.root(**transport_options.dup)
-
-          res = transport_root.send_info
-          if res.ok?
-            if res.endpoints.include?('/v0.7/config')
-              Datadog.logger.debug { 'agent reachable and reports remote configuration endpoint' }
-            else
-              Datadog.logger.error do
-                'agent reachable but does not report remote configuration endpoint: ' \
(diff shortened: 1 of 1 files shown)


lib/datadog/core/remote/component.rb
```diff
@@ -18,6 +18,28 @@ def initialize(settings, agent_settings)
...
+
+          res = transport_root.send_info
+          if res.ok?
+            if res.endpoints.include?('/v0.7/config')
+              Datadog.logger.debug { 'agent reachable and reports remote configuration endpoint' }
+            else
+              Datadog.logger.error do
+                'agent reachable but does not report remote configuration endpoint: ' \
+                  'disabling remote configuration for this process.'
+              end
```
ivoanjo, line 31: Since this is a user-visible message, consider perhaps suggesting that users upgrade their agent to get rid of this message?

(Same for the unreachable message below -- perhaps tell them that something seems wrong and they'll probably not be able to report data?)

GustavoCaso: Great point @ivoanjo 

Since we are targeting this change beta release. We can improve the error message for the official release 😄

GustavoCaso (approved):
LGTM

Issue #2045, opened 2022-05-25
Memory leak and memory bloat in resque processes when profiling is enabled (suspected to impact at least 0.54.2 and above)


jorgemanrubia:
When we upgraded to version 1.0 our resque processes started to leak memory and, sporadically, we registered some sudden multi-GB memory bloats:

<img width="1362" alt="CleanShot 2022-05-25 at 12 53 20@2x" src="https://user-images.githubusercontent.com/129938/170246253-3c4b0735-1938-44bc-8311-c5e3122a3e51.png">

I disabled the `resque` and `active_job` instrumentation integrations but it was still leaking, so not really sure it's related to those specifically. The processes leaking memory were the resque workers for sure. Once reverted to the previous gem, things went back to normal.

![CleanShot 2022-05-25 at 12 55 37@2x](https://user-images.githubusercontent.com/129938/170246685-0c5a8d17-24f2-48e8-936a-b8c081129e01.png)

This was in a beta server that had barely any activity.

delner:
Thanks for the report. Is there anything you can share to help us replicate this? (Config, Gemfile etc?)

jorgemanrubia:
@delner 👍 I’ll prepare something I can share here.

jorgemanrubia:
I found something interesting when trying to come up with a minimal setup to share: what causes the leak is to enable profiling. So I can reproduce with this setup:

```ruby
datadog_enabled = true

Datadog.configure do |config|
  config.service = "bc3"
  config.env = Rails.env

  config.tracing.instrument :rails
  config.tracing.instrument :resque

  unless datadog_enabled
    config.tracing.transport_options = proc do |transport|
      transport.adapter :test
    end
  end

  config.profiling.enabled = datadog_enabled
  config.tracing.enabled = datadog_enabled
end
```

But the problem gets solved if I remove this line:

```ruby
config.profiling.enabled = datadog_enabled
```

I can't share our full `Gemfile` but [here are some relevant gems and versions we use](https://gist.github.com/jorgemanrubia/0a103c2728eb28c6bcdfefb2b7e31079) related to Rails, resque, elasticsearch and AWS.

ivoanjo:
Hey 👋 . I work on the profiler bits so it seemed relevant to step in.

Between 0.54.2 and 1.0.0 there were very little changes to the profiler, so perhaps that will make it easier to track it down.

[...]

PR #2757, opened 2023-04-06
fix: Correctly set mutex

ixti:
**What does this PR do?**

Fixes Mutex assignment to be thread-safe.

**Motivation**

Tried to find what might be causing https://github.com/DataDog/dd-trace-rb/issues/2045 and found (unfortunately unrelated) but somewhat problematic code. We've tried this patch in production, unfortunately as I said, it's not fixing #2045, but still worth fixing.

**Additional Notes**

This is a simple demo, why `@mutex ||= Mutex.new` is a bad idea. Save the below ruby script as `demo.rb`:

``` ruby
    module BeforePatch
      def mutex
        @mutex ||= Mutex.new
      end
    end

    module AfterPatch
      def self.included(base)
        base.prepend(PrependMethods)
      end

      module PrependMethods
        def initialize(...)
          @mutex = Mutex.new
          super(...)
        end

        attr_reader :mutex
      end
    end

    class Demo
      include(ENV["FIXED"] ? AfterPatch : BeforePatch)
    end

    demo    = Demo.new
    results = []
    workers = Array.new(32) do
      Thread.new do
        results << demo.mutex.object_id
      end
    end

    workers.each(&:join)
    puts results.tally
```

Run:

``` bash
    while true; do ruby demo.rb; done
```

At some point you will see that there are more than one mutex. Now, run:

``` bash
    while true; do FIXED=1 ruby demo.rb; done
```

There should be no such issue.

**How to test the change?**

It's not very feasible to automatically test this. But the change is small enough.

### DIFF
--- a/lib/datadog/core/workers/async.rb
+++ b/lib/datadog/core/workers/async.rb
@@ -11,6 +11,10 @@ module Datadog
           FORK_POLICY_RESTART = :restart
           SHUTDOWN_TIMEOUT = 1
 
+          # This single shared mutex is used to avoid concurrency issues during the
+          # initialization of per-instance lazy-initialized mutexes.
+          MUTEX_INIT = Mutex.new
+
           def self.included(base)
             base.prepend(PrependedMethods)
           end
@@ -86,7 +90,7 @@ module Datadog
             :result
 
           def mutex
-            @mutex ||= Mutex.new
+            @mutex || MUTEX_INIT.synchronize { @mutex ||= Mutex.new }
           end
 
           def after_fork
@@ -99,7 +103,7 @@ module Datadog
             :pid
 
           def mutex_after_fork
-            @mutex_after_fork ||= Mutex.new
+            @mutex_after_fork || MUTEX_INIT.synchronize { @mutex_after_fork ||= Mutex.new }
           end
 
           def worker
--- a/lib/datadog/core/workers/interval_loop.rb
+++ b/lib/datadog/core/workers/interval_loop.rb
@@ -10,6 +10,10 @@ module Datadog
         BACK_OFF_MAX = 5
         BASE_INTERVAL = 1
 
+        # This single shared mutex is used to avoid concurrency issues during the
+        # initialization of per-instance lazy-initialized mutexes.
+        MUTEX_INIT = Mutex.new
+
         def self.included(base)
           base.prepend(PrependedMethods)
         end
@@ -82,7 +86,7 @@ module Datadog
           :loop_base_interval
 
         def mutex
-          @mutex ||= Mutex.new
+          @mutex || MUTEX_INIT.synchronize { @mutex ||= Mutex.new }
         end
 
         private


lib/datadog/core/workers/async.rb
```diff
@@ -17,9 +17,24 @@ def self.included(base)
 
           # Methods that must be prepended
           module PrependedMethods
+            def initialize(*args, **kwargs, &block)
```
delner, line 20: Is there a reason this needs to be done in the constructor?

Generally speaking its unsafe to prepend onto `initialize`, because not all objects have the same constructor signature. It introduces a kind of unspoken coupling (expects implementing objects to have a very specific constructor.)

delner: Attempting to answer my own question, I'm going to guess this has something to do with lazy initialization not being sufficient if that object is accessed concurrently prior to initialization...

delner: FWIW the breakage I'm seeing in CI right now is related to the constructor definition:

```
68.2) Failure/Error:
              def initialize(&block)
                @task = block
              end
            
            ArgumentError:
              wrong number of arguments (given 1, expected 0)
            # ./lib/datadog/core/worker.rb:8:in `initialize'
```

But it's only on Ruby 2.1-2.4. Probably the `kwargs`, so we may need to introduce some compatibility checks there.

ixti: > Attempting to answer my own question, I'm going to guess this has something to do with lazy initialization not being sufficient if that field is accessed concurrently prior to initialization...

Yeah, there's a chance that one thread will use one mutex instance, and then another will use another one. If we can move mutex into a constant — that would solve the issue entirely :D but can cause an issue with performance (haven't looked into details how those classes are used).

ixti:
@delner I have an alternative idea how to ensure all threads are using correct mutex, not sure how will you like the idea though:

``` ruby
module Async
  GLOBAL_MUTEX = Mutex.new

  def mutex
    @mutex || GLOBAL_MUTEX.synchronize { @mutex ||= Mutex.new }
  end
end
```

That will definitely work on all Ruby versions and will avoid modifying initializer :D

lib/datadog/core/workers/async.rb
```diff
@@ -10,6 +10,7 @@ module Thread
           FORK_POLICY_STOP = :stop
           FORK_POLICY_RESTART = :restart
           SHUTDOWN_TIMEOUT = 1
+          GLOBAL_MUTEX = Mutex.new
```
ivoanjo, line 13:
```suggestion
          # This single shared mutex is used to avoid concurrency issues during the
          # per-instance lazy-initialized mutexes.
          MUTEX_INIT = Mutex.new
```

lib/datadog/core/workers/async.rb
```diff
@@ -86,7 +87,7 @@ def fork_policy
             :result
 
           def mutex
-            @mutex ||= Mutex.new
+            @mutex || GLOBAL_MUTEX.synchronize { @mutex ||= Mutex.new }
```
ivoanjo, line 90:
```suggestion
            @mutex || MUTEX_INIT.synchronize { @mutex ||= Mutex.new }
```

lib/datadog/core/workers/async.rb
```diff
@@ -99,7 +100,7 @@ def after_fork
             :pid
 
           def mutex_after_fork
-            @mutex_after_fork ||= Mutex.new
+            @mutex_after_fork || GLOBAL_MUTEX.synchronize { @mutex_after_fork ||= Mutex.new }
```
ivoanjo, line 103:
```suggestion
            @mutex_after_fork || MUTEX_INIT.synchronize { @mutex_after_fork ||= Mutex.new }
```

lib/datadog/core/workers/interval_loop.rb
```diff
@@ -80,7 +81,7 @@ def loop_wait_before_first_iteration?
           :loop_base_interval
 
         def mutex
-          @mutex ||= Mutex.new
+          @mutex || GLOBAL_MUTEX.synchronize { @mutex ||= Mutex.new }
```
ivoanjo, line 84:
```suggestion
          @mutex || MUTEX_INIT.synchronize { @mutex ||= Mutex.new }
```

lib/datadog/core/workers/interval_loop.rb
```diff
@@ -7,6 +7,7 @@ module IntervalLoop
         BACK_OFF_RATIO = 1.2
         BACK_OFF_MAX = 5
         BASE_INTERVAL = 1
+        GLOBAL_MUTEX = Mutex.new
```
ivoanjo, line 10:
```suggestion
          # This single shared mutex is used to avoid concurrency issues during the
          # per-instance lazy-initialized mutexes.
          MUTEX_INIT = Mutex.new
```

ivoanjo (approved):
Hey @ixti :wave:

I spoke with @delner and we're very happy to accept this version -- it's a great solution and contribution.

Indeed creating a mutex is so fast and specific that having a single shared lock for it does not seem like it will ever be a scalability limit.

I've left a couple of suggestions, but otherwise happy to merge this in :)

ixti:
@ivoanjo Thank you! Applied the suggestions and cleaned-up commits :D

ivoanjo:
Thanks again, merging away!

delner:
@ixti thanks for the contribution! We really appreciate it. :)

PR #2760, opened 2023-04-06
Change default `service_name` values

zarirhamza:
Process to change all default values for `service_names` per integration to `DD_SERVICE`

- Change default `service_name` values based on span attribute schema version. If version is v0, the default stays as is, otherwise if it is version v1, the default changes to `DD_SERVICE`
- Adds and updates all tests for impacted integrations. This includes a new `settings_spec.rb` per integration that tests the switching of defaults between v0 and v1.
- Encapsulates old `peer.service` values behind `v0` schema version only as there will be changes to the value based on version in the future.

Affected integrations: `aws`, `active_record`, `active_support`, `dalli`, `elasticsearch`, `ethon`, `excon`, `faraday`, `grpc`, `http`


spec/datadog/tracing/contrib/span_attribute_schema_examples.rb
```diff
@@ -1,4 +1,4 @@
-RSpec.shared_examples 'span attributes schema' do
+RSpec.shared_examples 'schema version span' do |_default_service_name|
```
TonyCTHsu, line 1: Should we be using this block argument?

spec/datadog/tracing/contrib/http/configuration/configuration_spec.rb
```diff
@@ -0,0 +1,31 @@
+require 'datadog/tracing/contrib/http/configuration/settings'
+
+RSpec.describe Datadog::Tracing::Contrib::HTTP::Configuration::Settings do
```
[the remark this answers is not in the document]
zarirhamza: Done

lib/datadog/tracing/contrib/ethon/easy_patch.rb
```diff
@@ -139,7 +142,9 @@ def datadog_tag_request
               span.set_tag(Tracing::Metadata::Ext::NET::TAG_TARGET_HOST, uri.host)
               span.set_tag(Tracing::Metadata::Ext::NET::TAG_TARGET_PORT, uri.port)
 
-              span.set_tag(Tracing::Metadata::Ext::TAG_PEER_HOSTNAME, uri.host)
+              if Contrib::SpanAttributeSchema.default_span_attribute_schema?
+                span.set_tag(Tracing::Metadata::Ext::TAG_PEER_HOSTNAME, uri.host)
```
TonyCTHsu, line 146: The other integrations are not guarded by the conditions, any reason for this?

zarirhamza: Fixed

TonyCTHsu (approved):

PR #2765, opened 2023-04-07
Change default `service_name` values Part 2

zarirhamza:
Process to change all default values for `service_names` per integration to `DD_SERVICE`

- Change default `service_name` values based on span attribute schema version. If version is v0, the default stays as is, otherwise if it is version v1, the default changes to `DD_SERVICE`
- Adds and updates all tests for impacted integrations. This includes a new `settings_spec.rb` per integration that tests the switching of defaults between v0 and v1.
- Encapsulates old `peer.service` values behind `v0` schema version only as there will be changes to the value based on version in the future.

Affected integrations: `httpclient`, `httprb`, `mongodb`, `mysql2`, `pg`, `presto`, `racecar` , `rack`, `redis`, `rest_client`, `sequel`


marcotc (approved):
Looks great, @zarirhamza!

Are there still missing integrations to come in a future PR, or this is basically it?

zarirhamza:
> Looks great, @zarirhamza!
> 
> Are there still missing integrations to come in a future PR, or this is basically it?

This is the final PR! If it could be merged, that would be perfect, thank you!

PR #2767, opened 2023-04-10
[APPSEC-7791] Handle exceptions when fetching remote config:

GustavoCaso:
When fetching remote configuration, we could face errors. 

- The connection fails to open to communicate with the agent 
- The agent is up and running, but the `/v0.7/config` endpoint is not present
- The agent and the config endpoint are up and running, but the response from the config endpoint is invalid

We want to guard ourselves against those cases.

The code on this PR handles those cases. 

In the case of the agent, the endpoint being down or an invalid response payload, we return `nil` rather than `Datadog::Core::Transport::Config::Response`

In the case that the agent returns a 200 with an empty response, we are going to return an empty                                       `Datadog::Core::Transport::Config::Response`

### DIFF
--- a/Steepfile
+++ b/Steepfile
@@ -643,6 +643,7 @@ target :ddtrace do
   library 'tsort'
   library 'json'
   library 'ipaddr'
+  library 'socket'
   library 'net-http'
   library 'securerandom'
   library 'base64'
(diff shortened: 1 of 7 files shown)


GustavoCaso:
These changes are no longer needed. Closing it.

PR #2768, opened 2023-04-11
Enable Style/FrozenStringLiteralComment cop

GustavoCaso:
**What does this PR do?**

Enable Style/FrozenStringLiteralComment on all files. 

I remove the exclusions from the `.rubocop_todo.yml` and run `bundle exec rubocop -A`. That added the missing comments on the files and removed the unnecessary `.freeze` from strings. 

I had to change some files to make sure the test continued to pass 😄 

I made the changes on separate commits so it is easier to review

- Remove exclusions from rubocop_todo
- Run `bundle exec rubocop -A`
- Fix cases where we manipulate frozen strings

**Motivation**

Make sure we have consistent codebase and enforce new files to add the `# frozen_string_literal: true` comment 

**Additional Notes**

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/appsec/contrib/rails/patcher.rb
+++ b/lib/datadog/appsec/contrib/rails/patcher.rb
@@ -133,7 +133,7 @@ module Datadog
           end
 
           def inspect_middlewares(app)
-            Datadog.logger.debug { 'Rails middlewares: ' << app.middleware.map(&:inspect).inspect }
+            Datadog.logger.debug { "Rails middlewares: #{app.middleware.map(&:inspect).inspect}" }
           end
 
           def patch_after_intialize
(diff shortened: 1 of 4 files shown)


TonyCTHsu (approved):

ivoanjo:
So this change is something I quite like but I believe @marcotc was hesitating on doing because (if I remember correctly) he had concerns that customers may be mutating some of the strings our APIs provide, and thus freezing those strings would cause problems upgrading.

(I believe this is why @marcotc did the work of finding out only safe uses of strings in #2745, #2746 and #2747).

I suggest double checking with @marcotc before releasing this.

marcotc:
I don't think this is safe, unless all strings have been inspected in the affected files.
If the were, then all good!
I wasn't able to inspect all strings myself because there are just so many, so I was trying to do it slowly on safe sunset of files.

PR #2769, opened 2023-04-11
Fix remote configuration path parsing

lloeki:
**What does this PR do?**

Fix remote configuration path parsing

**Motivation**

Turns out `config` hardcodedness was kind of ambiguous in the spec and I misunderstood it as being general. This now fails as `blocked_ips` has the config name be some kind of hash.

**Additional Notes**

RFC has been clarified.

**How to test the change?**

RC should now work with `blocked_ips` from `ASM_DATA` as sent from the backend.

### DIFF
--- a/lib/datadog/core/remote/configuration/path.rb
+++ b/lib/datadog/core/remote/configuration/path.rb
@@ -21,7 +21,7 @@ module Datadog
               /
               (?<config_id>[^/]+)
               /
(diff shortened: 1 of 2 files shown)


lloeki:
I redid the whole spec to increase coverage.

GustavoCaso (approved):
Thanks for improving the spec coverage

PR #2770, opened 2023-04-11
[PROF-6447] Improve detection of mysql2 gem incompatibilities with profiler

ivoanjo:
**What does this PR do?**:

As part of #2702 (more specifically, in a7d3b73321) we added a check for the `mysql2` gem which avoided turning on the new profiler when we detected it.

This check was a bit heavy-handed because the known incompatibility between the `mysql2` gem and the profiler only happens if the `mysql2` gem is using a legacy version of the `libmysqlclient` C library. (Specifically, a version prior to 8.0.0) 

In this PR, we improve the check to _actually_ check the `libmysqlclient` version. Thus, we will only fall back to using the legacy profiler as needed, not always as before.

**Motivation**:

The `mysql2` gem is a commonly-used database driver for Ruby (and Ruby on Rails) applications, and we don't want these customers to get stuck using the old profiler.

**Additional Notes**:

As part of the check, we need to `require 'mysql2'` during profiler initialization (since we need to call a method provided by the gem).

In the rare just-in-case situation that this causes issues and customers don't want this to happen, this PR also introduces a new option to disable this behavior:

* `DD_PROFILING_SKIP_MYSQL2_CHECK` via environment variable
* `c.profiling.advanced.skip_mysql2_check` via code

**How to test the change?**:

Test coverage is included.

To test this with the actual `mysql2` gem, our CI images have a modern version of `libmysqlclient` (specifically, the mariadb variant):

```
$ DD_PROFILING_ENABLED=true DD_TRACE_DEBUG=true bundle exec ddtracerb exec ruby -e "sleep 1"
DEBUG -- ddtrace: [ddtrace] (/app/lib/datadog/profiling/component.rb:205:in `compatible_libmysqlclient_version?') Requiring `mysql2` to check if the `libmysqlclient` version it uses is compatible with profiling
DEBUG -- ddtrace: [ddtrace] (/app/lib/datadog/profiling/component.rb:218:in `compatible_libmysqlclient_version?') The `mysql2` gem is using a compatible version of the `libmysqlclient` library (10.5.15)
DEBUG -- ddtrace: [ddtrace] (/app/lib/datadog/core/configuration/components.rb:92:in `startup!') Profiling started
DEBUG -- ddtrace: [ddtrace] (/app/lib/datadog/profiling/collectors/cpu_and_wall_time_worker.rb:58:in `block in start') Starting thread for: #<Datadog::Profiling::Collectors::CpuAndWallTimeWorker:0x00005e979dd5e498>
```

...and you can use an older Ubuntu 18.04 image to see the failing case:

```
$ DD_PROFILING_ENABLED=true DD_TRACE_DEBUG=true bundle exec ddtracerb exec ruby -e "sleep 1"
DEBUG -- ddtrace: [ddtrace] (/working/lib/datadog/profiling/component.rb:205:in `compatible_libmysqlclient_version?') Requiring `mysql2` to check if the `libmysqlclient` version it uses is compatible with profiling
DEBUG -- ddtrace: [ddtrace] (/working/lib/datadog/profiling/component.rb:218:in `compatible_libmysqlclient_version?') The `mysql2` gem is using an incompatible version of the `libmysqlclient` library (5.7.41)
 WARN -- ddtrace: [ddtrace] (/working/lib/datadog/profiling/component.rb:170:in `enable_new_profiler?') Falling back to legacy profiler because an incompatible version of the mysql2 gem is installed. Older versions of libmysqlclient (the C library used by the mysql2 gem) have a bug in their signal handling code that the new profiler can trigger. This bug (https://bugs.mysql.com/bug.php?id=83109) is fixed in libmysqlclient versions 8.0.0 and above.
DEBUG -- ddtrace: [ddtrace] (/working/lib/datadog/core/configuration/components.rb:92:in `startup!') Profiling started
DEBUG -- ddtrace: [ddtrace] (/working/lib/datadog/core/workers/async.rb:130:in `start_worker') Starting thread for: #<Datadog::Profiling::Collectors::OldStack:0x00005604f7ac46b0>
```


lib/datadog/profiling/component.rb
```diff
@@ -166,14 +166,12 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
 
         return false if RUBY_VERSION.start_with?('2.3.', '2.4.', '2.5.')
 
-        if Gem.loaded_specs['mysql2']
+        if Gem.loaded_specs['mysql2'] && !compatible_libmysqlclient_version?(settings)
```
GustavoCaso, line 169: Should we encapsulate the `Gem.loaded_specs['mysql2']` inside the `compatible_libmysqlclient_version?(settings)` method?

lib/datadog/profiling/component.rb
```diff
@@ -191,6 +189,47 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
...
+      # Versions of libmysqlclient prior to 8.0.0 are known to have buggy handling of system call interruptions.
+      # The profiler can sometimes cause system call interruptions, and so this combination can cause queries to fail.
+      #
+      # See https://bugs.mysql.com/bug.php?id=83109 and
+      # https://docs.datadoghq.com/profiler/profiler_troubleshooting/ruby/#unexpected-run-time-failures-and-errors-from-ruby-gems-that-use-native-extensions-in-dd-trace-rb-1110
+      # for details.
+      #
+      # The `mysql2` gem's `info` method can be used to determine which `libmysqlclient` version is in use, and thus to
+      # detect if it's safe for the profiler to use signals or if we need to employ a fallback.
+      private_class_method def self.compatible_libmysqlclient_version?(settings)
```
GustavoCaso, line 202: We could rename the method to `incompatible_libmysqlclient_version?` that way in the check above we do not have to negate the result

GustavoCaso (approved):

lib/datadog/profiling/component.rb
ivoanjo: Changed in c44e9a7991a32fb1e5a0e264047999ffffd58ab5

PR #2773, opened 2023-04-11
Protect Processor::Context#run with a mutex

lloeki:
**What does this PR do?**

Protect Processor::Context#run with a mutex

**Motivation**

`ddwaf_run` mutates `ddwaf_context` in a non-thread-safe way.

**Additional Notes**

Sicne there is a 1:1:1 context - request - web worker thread mapping it should very rarely happen in typical apps so the lock will just fly through.

This aims to protect from code that spawns threads within a request and that would theoretically trigger a context run (e.g [ActionController::Live](https://github.com/rails/rails/blob/2675c906b1cc4cc877ddee4d62dc33f544adc748/actionpack/lib/action_controller/metal/live.rb)). There is currently none but there could be in the future.

**How to test the change?**

Difficult since there's no such real case currently. Even synthetically (e.g in a spec) it would be hard as it would require producing a reliably failing case of libddwaf stepping onto itself.

### DIFF
--- a/lib/datadog/appsec/processor.rb
+++ b/lib/datadog/appsec/processor.rb
@@ -22,11 +22,13 @@ module Datadog
 
           start_ns = Core::Utils::Time.get_time(:nanosecond)
 
+          # this WAF::Context#run call is not thread safe as it mutates the context
           # TODO: remove multiple assignment
           _code, res = @context.run(input, timeout)
 
           stop_ns = Core::Utils::Time.get_time(:nanosecond)
 
+          # these updates are not thread safe and should be protected
           @time_ns += res.total_runtime
           @time_ext_ns += (stop_ns - start_ns)
           @timeouts += 1 if res.timeout
--- a/lib/datadog/appsec/processor.rb
+++ b/lib/datadog/appsec/processor.rb
@@ -14,9 +14,12 @@ module Datadog
           @time_ext_ns = 0.0
           @timeouts = 0
           @events = []
+          @run_mutex = Mutex.new
         end
 
         def run(input, timeout = WAF::LibDDWAF::DDWAF_RUN_TIMEOUT)
+          @run_mutex.lock
+
           start_ns = Core::Utils::Time.get_time(:nanosecond)
 
           # TODO: remove multiple assignment
@@ -30,6 +33,8 @@ module Datadog
           @timeouts += 1 if res.timeout
 
           res
+        ensure
+          @run_mutex.unlock
         end
 
         def finalize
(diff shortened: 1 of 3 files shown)


GustavoCaso (approved):
Makes sense. Feel free to 🚢 once Steep is happy

lib/datadog/appsec/processor.rb
```diff
@@ -14,9 +14,12 @@ def initialize(processor)
           @time_ext_ns = 0.0
           @timeouts = 0
           @events = []
+          @run_mutex = Mutex.new
         end
 
         def run(input, timeout = WAF::LibDDWAF::DDWAF_RUN_TIMEOUT)
+          @run_mutex.lock
+
           start_ns = Core::Utils::Time.get_time(:nanosecond)
```
ivoanjo, line 23: Minor: Consider using `@run_mutex.synchronize` -- I think that one's a bit easier to maintain and "break" with future changes.

ivoanjo (commented):
This seems reasonable, but it's a very small/subtle code change for a potential big issue. Thus, I would suggest perhaps leaving a few comments in the code explaining why this tiny addition is important, otherwise this may get hard to maintain in the future (since people would need to partially "reverse engineer" why this was added here and what it's protecting exactly).

lib/datadog/appsec/processor.rb
lloeki: Rationale:

- blocks carry a cost: def/ensure/end is faster
- it makes the function return value less explicit (you have to konw/remember/lookup that synchronize does return the block value)
- it creates whitespace churn on the code change
- maybe I'm doing too much Go but def/ensure/end is a common pattern that matches Go's `defer` ;)

lloeki: Added a comment to clarify the intent!

lib/datadog/appsec/processor.rb
```diff
@@ -14,22 +14,28 @@ def initialize(processor)
...
           @time_ns += res.total_runtime
           @time_ext_ns += (stop_ns - start_ns)
           @timeouts += 1 if res.timeout
 
           res
+        ensure
+          @run_mutex.unlock
         end
 
         def finalize
```
ivoanjo, line 41: Just a final question -- does `@context.finalize` not need protecting too?

lloeki: Nope, that one is fine.

Context initialization and finalization happen on the same thread and should outlive any transient thread that would theoretically use the same context (you'd get much bigger problems otherwise, that a simple synchronize would not solve). It's all theoretical anyway because no code is currently allowing getting the context from another thread, so this is mostly bulletproofing following a discussion around `run`.

ivoanjo (approved):
:+1: LGTM

Issue #2724, opened 2023-03-28
NameError: uninitialized constant Rack::Request


orien:
There is an incompatibility with non-Rack projects that make use of the `rack/utils` library.

In this scenario, the `Rack` module is loaded/defined, but `Rack::Request` is not.

https://github.com/DataDog/dd-trace-rb/blob/25686d40586c992e8950113b8e80483633341dc2/lib/datadog/appsec/contrib/rack/gateway/request.rb#L39

**Current behaviour**
Error raised (NameError: uninitialized constant Rack::Request)

**Expected behaviour**

Error should not be raised.

**Steps to reproduce**

Run this script:

```ruby
#!/usr/bin/env ruby

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rack', require: 'rack/utils'
  gem 'ddtrace', '1.10.1', require: 'ddtrace/auto_instrument'
end
```

See error:

```
/ruby/3.2.1/lib/ruby/gems/3.2.0/gems/ddtrace-1.10.1/lib/datadog/appsec/contrib/rack/gateway/request.rb:39:in `<class:Request>': uninitialized constant Rack::Request (NameError)

            if defined?(::Rack) && ::Rack::Request.instance_methods.include?(:each_header)
                                         ^^^^^^^^^
Did you mean?  Rack::REQUEST_PATH
	from /ruby/3.2.1/lib/ruby/gems/3.2.0/gems/ddtrace-1.10.1/lib/datadog/appsec/contrib/rack/gateway/request.rb:13:in `<module:Gateway>'
	from /ruby/3.2.1/lib/ruby/gems/3.2.0/gems/ddtrace-1.10.1/lib/datadog/appsec/contrib/rack/gateway/request.rb:11:in `<module:Rack>'
	from /ruby/3.2.1/lib/ruby/gems/3.2.0/gems/ddtrace-1.10.1/lib/datadog/appsec/contrib/rack/gateway/request.rb:10:in `<module:Contrib>'
	from /ruby/3.2.1/lib/ruby/gems/3.2.0/gems/ddtrace-1.10.1/lib/datadog/appsec/contrib/rack/gateway/request.rb:9:in `<module:AppSec>'
	from /ruby/3.2.1/lib/ruby/gems/3.2.0/gems/ddtrace-1.10.1/lib/datadog/appsec/contrib/rack/gateway/request.rb:8:in `<module:Datadog>'
[...]

PR #2775, opened 2023-04-12
require appsec instrumentation files only when instrumentation has been activated.

GustavoCaso:
The main reason to do this is to make sure we are not loading files that depend on customer dependencies
unless we are certain the customer wants to instrument such a library.

The customer complained that when trying to load `dd-trace-rb` for non-rack projects, some required files that depend on `Rack::Request` being loaded. 

The test script to validate that things are working as expected is:

```ruby
#!/usr/bin/env ruby

require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'rack', require: 'rack/utils'
  gem 'ddtrace', '1.10.1', require: 'ddtrace/auto_instrument'
end
```

As we can see, the `rack` is loaded, but only a subset of the rack gem.

The changes on this PR fix it with two changes:
- Make sure only to require dedicated files for the instrumentation of rack, sinatra and rails when they get activated
- Making sure the rack integration condition `loaded?` does not simply check for the existence of `Rack` class, but `Rack::Request`, which signals that certain rack components have been loaded.

### DIFF
--- a/lib/datadog/appsec/contrib/rack/integration.rb
+++ b/lib/datadog/appsec/contrib/rack/integration.rb
@@ -4,8 +4,6 @@ require_relative '../integration'
 
 require_relative 'configuration/settings'
 require_relative 'patcher'
-require_relative 'request_middleware'
-require_relative 'request_body_middleware'
 
 module Datadog
   module AppSec
@@ -24,7 +22,7 @@ module Datadog
           end
 
           def self.loaded?
-            !defined?(::Rack).nil?
+            !defined?(::Rack::Request).nil?
           end
 
           def self.compatible?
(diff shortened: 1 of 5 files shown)


TonyCTHsu (approved):

ivoanjo (commented):
Don't forget to mention the customer issue -- #2724 :)

To be honest, I'm not the biggest fan of moving the requires inside the methods unless we really can't avoid it, because it can be hard to test/validate, and can even lead to flaky tests (e.g. some test runs before/after the require and has a different result).

But I don't have a lot of concrete examples of things that can go wrong in moving it inside the `patch` methods...

On the other hand, there's a potential advantage to loading less code into the Ruby VM if we can avoid it, which speeds up booting the app.

So you know my opinion now, I'll leave the final decision to you :)

PR #2776, opened 2023-04-12
Make `racecar` traces use the default service name

dasch:
Right now, if no `service_name` is specified for a `racecar` trace, the default will not come from the `DD_SERVICE` env var, but will rather be the static string `"racecar"`. This seems like a poor default, assuming an organization runs more than one consumer.

**What does this PR do?**
It removes the override for the `service_name` option, allowing the default behavior to take place.

**Motivation**
We want to configure our Datadog service names with Kubernetes annotations, but some integrations do not respect those, and we have to force the use of the env var on a case by case basis.

**How to test the change?**
Test that it uses the env var `DD_SERVICE` rather than just `"racecar"`.

### DIFF
--- a/lib/datadog/tracing/contrib/racecar/configuration/settings.rb
+++ b/lib/datadog/tracing/contrib/racecar/configuration/settings.rb
@@ -25,8 +25,6 @@ module Datadog
               o.default { env_to_float(Ext::ENV_ANALYTICS_SAMPLE_RATE, 1.0) }
               o.lazy
             end
-
-            option :service_name, default: Ext::DEFAULT_PEER_SERVICE_NAME
           end
         end
       end


delner:
Hey @dasch ! Happy to see you again in the repo. :)

I think this is a great change; it's moving in the general direction of where we want to go with service naming: where instrumentation reflects `DD_SERVICE` when measuring operations internal to the Ruby application.

Only consideration is that it could be potentially considered a breaking change. If users depend upon the default (for some reason), then changing the default may affect their services/service map/monitors unexpectedly. We generally reserve breaking changes for major version releases.

Is there some kind of intermediate step we can take here?

dasch:
Good to be back! 😄 

I'm not sure if there's an intermediary, non-breaking change – you could push this to the next breaking change if you'd like. Perhaps do a search for all cases where `service_name` gets defaulted on an integration level and do all of them in one go?

delner:
Also, slightly off topic, but just to pick your brain @dasch...

Users often (mis)use the `service_name` on these integrations to make distinct layers (middleware, controllers, presentation) visible within their application. This way they get a coloration on the trace graph, and a time spent % table on their traces.

Do you employ this practice at all?

The downside here is that it mixes up applications with layers of applications, which can make a mess of the service map. We'd like users to have both a nice service map and have introspection into the layers of their application. The latter is the tricky part. If you could envision what it would be like to label and visualize these application layers in Datadog, what would that process look like?

We're looking for some inspiration from user cases here, if you have any to offer :)

dasch:
Sounds great! I'm happy to wait until late 2023.

> Also, slightly off topic, but just to pick your brain @dasch...
> 
> Users often (mis)use the `service_name` on these integrations to make distinct layers (middleware, controllers, presentation) visible within their application. This way they get a coloration on the trace graph, and a time spent % table on their traces.
> 
> Do you employ this practice at all?

Guilty as charged 😆 part of this is path dependence: we adopted APM when it was brand new, and a lot of the later features, like the dependency map, didn't exist back then. Crucially, I don't think there was any UI for selecting which operation to use for the resource breakdown, and we wanted exactly what you mention: distinct views for different layers/aspects of our application, such as errors rates and latency by controller/view/middleware/etc.

> The downside here is that it mixes up applications with layers of applications, which can make a mess of the service map. We'd like users to have both a nice service map and have introspection into the layers of their application. The latter is the tricky part. If you could envision what it would be like to label and visualize these application layers in Datadog, what would that process look like?
> 
> We're looking for some inspiration from user cases here, if you have any to offer :)

I think improved UX around switching between operations in the APM views would help here, e.g. a way to "zoom out" of a specific operation and view an overview of operations, then zoom back in to a specific operation. Still keep the "primary operation" concept, but make multi-operation services easier to navigate.

It would also help to have a multi-tiered visualization in trace flamegraphs – one primary color per service, but some variation per operation name.

delner:
It's available on most of our integrations, but not actively used to create coloration/groupings. Seems like a good candidate. Are there any other such tags that you think would be particularly useful to group on?

dasch:
Not that I can think of. What's component used for currently, btw?

dasch:
As in, how is it used in the UI?

delner:
It's currently not really "in use" other than that its populated with the library name of the integration that produced it. This value is a constant e.g. `active_record`). It's a tag we introduced in our out-of-the-box Ruby integrations; I wouldn't expect to find it on all instrumentation from all languages.

I suppose the question is more: "Is this conceptual descriptor useful to group on? Are there other conceptually useful descriptors that may not exist that are also useful to group on?"

If that seems like the most obvious one, and no others, that's okay too.

nightpool:
> The downside here is that it mixes up applications with layers of applications, which can make a mess of the service map. We'd like users to have both a nice service map and have introspection into the layers of their application. The latter is the tricky part. If you could envision what it would be like to label and visualize these application layers in Datadog, what would that process look like?

This is the biggest part for me that prevents me from enabling this today for our current integrations. The UI just doesn't support surfacing "sub-service" values in a very good way if at all. Ideally everything you can do to a "service" today in the UI (group by, sort by, get a table for) you should be able to do to a "component" or similar.

dasch:
I would prefer flexibility for "component". It's less important for there to be a standard about what it means than it is for it to support the use cases that are valuable – specifically, being able to associate parts of applications with specific owners.

PR #2778, opened 2023-04-12
Remove check for `::Rack::Request.instance_methods.include?(:each_header)` at load time

GustavoCaso:
Fixes https://github.com/DataDog/dd-trace-rb/issues/2724

The easiest solution rather than checking for the existence of `each_header` method for a rack request is to simply use `request.env.each_with_object`.

Removing that code fixes the load time issue when using `dd-trace-rb` on a non rack base application


lloeki (approved):
LGTM!

PR #2781, opened 2023-04-12
Scope the benchmarking service account to benchmarking jobs only

randomanderson:
**What does this PR do?**
Scope the benchmarking service account to benchmarking jobs only

**Motivation**
Ruby deployments to the reliability environment have been broken for a week. The cause was traced to https://github.com/DataDog/dd-trace-rb/commit/65459510a66ef8ce11550b3df0e720dc4ad42065

**How to test the change?**
* Validate deploys to the reliability env are fixed
* Validate running benchmarks still works

### DIFF
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -3,7 +3,6 @@
 # -----------------------------------------------------
 
 variables:
-  KUBERNETES_SERVICE_ACCOUNT_OVERWRITE: dd-trace-rb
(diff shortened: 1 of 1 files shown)


marcotc (approved):

ivoanjo:
Nice, thanks for chasing this one down!

ddyurchenko:
This is an interesting case, since we already overwrite service account in RelEnv - https://github.com/DataDog/datadog-reliability-env/blob/master/.gitlab-ci.yml#L70. Nice catch Laplie! :tada:

PR #2782, opened 2023-04-12
OpenTracing:Use Datadog configuration for the OpenTracer instance

### LINKED ISSUE #1563: Agent configuration settings ignored when using OpenTracing
Following the [Quickstart for OpenTracing](https://docs.datadoghq.com/tracing/setup_overview/setup/ruby/#quickstart-for-opentracing) documentation, I've created the following sample:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'ddtrace', path: '.'
  gem 'opentracing'
end

require 'opentracing'
require 'ddtrace'
require 'ddtrace/opentracer'

# # Activate the Datadog tracer for OpenTracing
OpenTracing.global_tracer = Datadog::OpenTracer::Tracer.new

Datadog.configure do |c|
  c.diagnostics.debug = true
  c.tracer.hostname = 'foo'
end

OpenTracing.start_active_span('operation_name') do |scope|
  # do nothing
end
```

Contrary to what may be expected, the `hostname` specified in the `Datadog.configure` block is completely ignored:

```
D, [2021-06-21T10:33:12.268754 #34028] DEBUG -- ddtrace: [ddtrace] (/Users/ivo.anjo/datadog/dd-trace-rb/lib/ddtrace/tracer.rb:363:in `write') Writing 1 spans (enabled: true)
D, [2021-06-21T10:33:12.268954 #34028] DEBUG -- ddtrace: [ddtrace] (/Users/ivo.anjo/datadog/dd-trace-rb/lib/ddtrace/tracer.rb:366:in `write') [
 Name: operation_name
Span ID: 2361826171763954817
Parent ID: 0
Trace ID: 3477994216835371584
Type:
Service: opentracing-testcase
Resource: operation_name
Error: 0
Start: 1624267992268434944
End: 1624267992268709888
Duration: 0.000275
Allocations: 48
Tags: [
   runtime-id => 46a72af0-fd9a-41c1-abb1-f7309fe96819]
Metrics: [
   _dd.agent_psr => 1.0,
   system.pid => 34028.0,
   _sampling_priority_v1 => 1.0]]

D, [2021-06-21T10:33:12.269029 #34028] DEBUG -- ddtrace: [ddtrace] (/Users/ivo.anjo/datadog/dd-trace-rb/lib/ddtrace/workers.rb:72:in `block in start') Starting thread for: #<Datadog::Workers::AsyncTransport:0x00007f8d2b985a20>
E, [2021-06-21T10:33:12.270126 #34028] ERROR -- ddtrace: [ddtrace] (/Users/ivo.anjo/datadog/dd-trace-rb/lib/ddtrace/transport/http/client.rb:35:in `rescue in send_request') Internal error during HTTP transport request. Cause: Failed to open TCP connection to 127.0.0.1:8126 (Connection refused - connect(2) for "127.0.0.1" port 8126) Location: /Users/ivo.anjo/.rvm/rubies/ruby-2.7.3/lib/ruby/2.7.0/net/http.rb:960:in `initialize'
I, [2021-06-21T10:33:12.270512 #34028]  INFO -- ddtrace: [ddtrace] (/Users/ivo.anjo/datadog/dd-trace-rb/lib/ddtrace/diagnostics/environment_logger.rb:31:in `log_environment!') DATADOG TRACER CONFIGURATION - {"date":"2021-06-21T10:33:12+01:00","os_name":"x86_64-apple-darwin19","version":"0.50.0","lang":"ruby","lang_version":"2.7.3","enabled":true,"service":"opentracing-testcase","agent_url":"http://127.0.0.1:8126?timeout=1","agent_error":"Datadog::Transport::InternalErrorResponse ok?: unsupported?:, not_found?:, client_error?:, server_error?:, internal_error?:true, payload:, error_type:Errno::ECONNREFUSED error:Failed to open TCP connection to 127.0.0.1:8126 (Connection refused - connect(2) for \"127.0.0.1\" port 8126)","debug":true,"analytics_enabled":false,"runtime_metrics_enabled":false,"vm":"ruby-2.7.3","partial_flushing_enabled":false,"priority_sampling_enabled":false,"health_metrics_enabled":false}
W, [2021-06-21T10:33:12.270545 #34028]  WARN -- ddtrace: [ddtrace] (/Users/ivo.anjo/datadog/dd-trace-rb/lib/ddtrace/diagnostics/environment_logger.rb:35:in `log_error!') DATADOG TRACER DIAGNOSTIC - Agent Error: Datadog::Transport::InternalErrorResponse ok?: unsupported?:, not_found?:, client_error?:, server_error?:, internal_error?:true, payload:, error_type:Errno::ECONNREFUSED error:Failed to open TCP connection to 127.0.0.1:8126 (Connection refused - connect(2) for "127.0.0.1" port 8126)
```

It seems configuration via code is broken for this usecase. Using environment variables (e.g. `DD_AGENT_HOST`) still works.

marcotc:
Fixes #1563

We currently create a new instance of `Datadog::Tracing::Tracer` for the OpenTracing integration.

This a byproduct of previous implementation details, where creating a new instance of the tracer for distinct use cases was appropriate.
Today, using the global Tracer instance is the correct way to integrate OpenTracing with Datadog tracing: this ensures the OpenTracer respects any configured Datadog configuration. Currently, this configuration is wiped clean when the OpenTracer is created.

There are no documentation changes because the current documentation actually matches the behaviour of this PR.


lib/datadog/opentracer/tracer.rb
```diff
@@ -15,10 +15,14 @@ class Tracer < ::OpenTracing::Tracer
...
-      # (see Datadog::Tracer#initialize)
-      def initialize(**options)
+      # Create an OpenTracing::Trace instance that is bound to the active
+      # Datadog tracer instance.
+      # To configure this tracer, use the global tracing configuration.
+      #
+      # DEV-2.0: Remove options as configuration is done through the global tracing configuration.
+      def initialize(**_legacy_options)
         super()
-        @datadog_tracer = Datadog::Tracing::Tracer.new(**options)
```
ivoanjo, line 21: Minor: I wonder if it's worth adding a warning when `_legacy_options` is not empty, to warn people they're being ignored and they should stop passing them in.

marcotc: Done!

lib/datadog/opentracer/tracer.rb
```diff
@@ -13,10 +14,21 @@ class Tracer < ::OpenTracing::Tracer
...
-      def initialize(**options)
+      # Create an OpenTracing::Trace instance that is bound to the active
+      # Datadog tracer instance.
+      # To configure this tracer, use the global tracing configuration.
+      #
+      # DEV-2.0: Remove options as configuration is done through the global tracing configuration.
+      def initialize(**ignored_options)
         super()
-        @datadog_tracer = Datadog::Tracing::Tracer.new(**options)
+        @datadog_tracer = Datadog.send(:components).tracer
```
ivoanjo, line 24: Since the `global_tracer` instance is a bit of a singleton that is supposed to live for the whole lifetime of the process, I'm not sure it's correct to try to store a reference to the `Datadog::Tracing::Tracer` that was available when it gets initialized.

In fact, I don't think this PR would fix my example from #1563, which I've updated to work with dd-trace-rb 1.x:

```ruby
require 'opentracing'
require 'ddtrace'
require 'datadog/opentracer'

# # Activate the Datadog tracer for OpenTracing
OpenTracing.global_tracer = Datadog::OpenTracer::Tracer.new

Datadog.configure do |c|
  c.diagnostics.debug = true
  c.agent.host = 'foo'
end

OpenTracing.start_active_span('operation_name') do |scope|
  # do nothing
end
```

This is because the `Datadog::OpenTracer::Tracer` has a reference to the tracer that existed before the configuration block run.

I'm thinking that perhaps the right way to go is to replace the `attr_reader :datadog_tracer` with a

```ruby
def datadog_tracer
  Datadog::Tracing.send(:tracer)
end
```

which would almost ensure that the latest tracer is always the one that gets used...

I was actually testing it out and it _still_ wasn't picking up the configuration because of [this monkey patch](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/opentracer/global_tracer.rb#L11) which I claim also doesn't seem correct with the way that dd-trace-rb 1.x expects components to be used.

After disabling that monkey patch, my example seems to work but -- I may be missing something so take my suggestions with a big grain of salt :)

TonyCTHsu: >I was actually testing it out and it still wasn't picking up the configuration because of [this monkey patch](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/opentracer/global_tracer.rb#L11) which I claim also doesn't seem correct with the way that dd-trace-rb 1.x expects components to be used.

I think it make sense that if we delegate `datadog_tracer` to the current tracer, we should avoid injecting tracer instance for datadog tracing configuration like the monkey patch. This way, the dependency relationships would be more streamlined.

PR #2784, opened 2023-04-13
Remove unused remote configuration payload from specs

GustavoCaso:
Reduce the amount of unnecessary information from the remote specs


lloeki:
Hmm 🤔 I wanted to keep them around as is because that's what the backend sends.

This ensures that the data schema is not crashing anything nor being unexpectedly processed.

PR #2785, opened 2023-04-13
Add appsec configuration benchmarks

lloeki:
**What does this PR do?**

Add appsec configuration benchmarks

**Motivation**

Do it early so that we can get actual data

**Additional Notes**

[Variants are already defined](https://github.com/DataDog/benchmarking-platform/tree/132c96cd0d331ddd090e36142273fab2bae2c74c) so hopefully this works right away!

**How to test the change?**

benchmark env should run and report data.

### DIFF
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -56,3 +56,13 @@ profiling-and-tracing-benchmarks:
   extends: .benchmarks
   variables:
     DD_RELENV_CONFIGURATION: profiling-and-tracing
+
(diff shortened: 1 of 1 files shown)


GustavoCaso (approved):

marcotc:
Sorry for the basic question but where do these run? And warn system is running them?

lloeki:
@marcotc On the benchmarking platform, triggered manually or scheduled (currently0 on private GitLab instance, on fixed-performance AWS hardware with no neighbour noise.

PR #2790, opened 2023-04-14
Migrate ddprof microbenchmarks from RelEnv

ddyurchenko:
**What does this PR do?**

Migrate ddprof microbenchmarks from RelEnv into Benchmarking Platform.

**Motivation**

These microbenchmarks are broken in RelEnv, hopefully they will have some more love here.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,6 +1,7 @@
 stages:
   - deploy
-  - benchmarks
+  - macrobenchmarks
+  - microbenchmarks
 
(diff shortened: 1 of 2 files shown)


ivoanjo (approved):
Thanks for this!

Just a nitpick: they're actually Ruby profiler benchmarks, not "ddprof" (ddprof is the Datadog native profiler -- it was present in some of the Ruby benchmarks as an additional tool to investigate performance, but it's not what's being measured :D)

PR #2794, opened 2023-04-17
[APPSEC-6712] Remote configuration `config_state`

GustavoCaso:
**What does this PR do?**
Stores and set remote configuration content version. Is that information to populate the client `config_state`

@lloeki will work on the remaining portion of `config_state`, which is adding the information if any error occurred during the remote configuration sync 

**Motivation**

**Additional Notes**

Here are the [system test assertions](https://github.com/DataDog/system-tests/blob/main/tests/remote_config/rc_expected_requests_asm_dd.json) on the expected values for `config_states`

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/core/remote/configuration/target.rb
+++ b/lib/datadog/core/remote/configuration/target.rb
@@ -48,16 +48,18 @@ module Datadog
             def parse(hash)
               length = Integer(hash['length'])
               digests = Configuration::DigestList.parse(hash['hashes'])
+              version = Integer(hash['custom']['v'])
 
-              new(digests: digests, length: length)
+              new(digests: digests, length: length, version: version)
             end
           end
 
-          attr_reader :length, :digests
+          attr_reader :length, :digests, :version
 
-          def initialize(digests:, length:)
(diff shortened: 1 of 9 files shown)


lloeki (approved):
LGTM (once the failure is fixed!)

lib/datadog/core/remote/configuration/target.rb
```diff
@@ -48,16 +48,18 @@ class << self
             def parse(hash)
               length = Integer(hash['length'])
               digests = Configuration::DigestList.parse(hash['hashes'])
+              version = Integer(hash['custom']['v'])
```
lloeki, line 51: This is failing:

```
NoMethodError:
       undefined method `[]' for nil:NilClass
     # ./lib/datadog/core/remote/configuration/target.rb:51:in `parse'
     # ./spec/datadog/core/remote/configuration/target_spec.rb:178:in `block (3 levels) in <top (required)>'
     # ./spec/datadog/core/remote/configuration/target_spec.rb:199:in `block (5 levels) in <top (required)>'
```

PR #2795, opened 2023-04-17
Fix version format in remote configuration payload

lloeki:
**What does this PR do?**

Fix version format in remote configuration payload.

**Motivation**

RFC mandates Semantic Versioning 2.0.0. Without this format, the payload gets silently rejected and RC response is always lacking targets, thus inoperative.

A workaround specifically for the already published `1.11.0.beta1` version which is affected has been deployed backend-side.

**Additional Notes**

The implementation hinges on the version pattern being enforced by `[MAJOR, MINOR, PATH, PRE].compact.join('.')`

Side note: I'm wondering why we have so many constants regarding versions, all computed from one another. I feel it makes it hard to test and refactor.

**How to test the change?**

A bit tough since a fix has been deployed:

- alter version to have PRE be e.g `'beta2'`
- start an app with RC and debug level logger
- witness configuration paths being updated

This should ultimately be covered by CI, but attempts at calling to RC from integration specs has proven unfruitful for setup+infrastructure reasons.

### DIFF
--- a/lib/datadog/core/environment/identity.rb
+++ b/lib/datadog/core/environment/identity.rb
@@ -49,8 +49,17 @@ module Datadog
           Core::Environment::Ext::LANG_VERSION
         end
 
+        # Returns tracer version, rubygems-style
         def tracer_version
           Core::Environment::Ext::TRACER_VERSION
+        end
+
+        # Returns tracer version, comforming to https://semver.org/spec/v2.0.0.html
+        def tracer_version_semver2
+          # from ddtrace/version.rb, we have MAJOR.MINOR.PATCH plus an optional .PRE
+          # - transform .PRE to -PRE if present
+          # - keep triplet before that
+          tracer_version.sub(/\.([a-zA-Z0-9]*[a-zA-Z][a-zA-Z0-9]*.*)$/, '-\1')
         end
       end
     end
--- a/lib/datadog/core/environment/identity.rb
+++ b/lib/datadog/core/environment/identity.rb
@@ -56,11 +56,58 @@ module Datadog
 
         # Returns tracer version, comforming to https://semver.org/spec/v2.0.0.html
         def tracer_version_semver2
-          # from ddtrace/version.rb, we have MAJOR.MINOR.PATCH plus an optional .PRE
+          # from ddtrace/version.rb, we have MAJOR.MINOR.PATCH plus optional .PRE and .BUILD
           # - transform .PRE to -PRE if present
-          # - keep triplet before that
-          tracer_version.sub(/\.([a-zA-Z0-9]*[a-zA-Z][a-zA-Z0-9]*.*)$/, '-\1')
+          # - transform .BUILD to +BUILD if present
+          # - keep triplet segments before that
+
+          m = SEMVER2_RE.match(tracer_version)
+
+          pre = "-#{m[:pre]}" if m[:pre]
+          build = "+gha#{m[:gha_run_id]}.g#{m[:git_sha]}.#{m[:branch].tr('.', '-')}" if m[:build]
+
+          "#{m[:major]}.#{m[:minor]}.#{m[:patch]}#{pre}#{build}"
         end
+
+        SEMVER2_RE = /
+          ^
+          # mandatory segments
+          (?<major>\d+)
+          \.
+          (?<minor>\d+)
+          \.
+          (?<patch>\d+)
+
+          # pre segments start with a value
+          # - containing at least one alpha
+          # - that is not part of our build segments expected values
+          # and stop with a value that is not part of our build segments expected values
+          (?:
+            \.
+            (?<pre>
+              (?!gha)
+              [a-zA-Z0-9]*[a-zA-Z][a-zA-Z0-9]*
+              (?:
+                \.
+                (?!gha)
+                [a-zA-Z0-9]+
+              )*
+            )
+          )?
+
+          # build segments: ours include CI info (`gha`), then git (`g`), then branch name
+          (?:
+            \.
+            (?<build>
+              gha(?<gha_run_id>\d+)
+              \.
+              g(?<git_sha>[a-f0-9]+)
+              \.
+              (?<branch>(?:[a-zA-Z0-9.])+)
+            )
+          )?
+          $
+        /xm.freeze
       end
     end
   end
--- a/lib/datadog/core/remote/client.rb
+++ b/lib/datadog/core/remote/client.rb
@@ -119,7 +119,7 @@ module Datadog
           client_tracer = {
             runtime_id: Core::Environment::Identity.id,
             language: Core::Environment::Identity.lang,
-            tracer_version: Core::Environment::Identity.tracer_version,
+            tracer_version: Core::Environment::Identity.tracer_version_semver2,
             service: Datadog.configuration.service,
             env: Datadog.configuration.env,
             tags: [], # TODO: add nice tags!
--- a/spec/datadog/core/environment/identity_spec.rb
+++ b/spec/datadog/core/environment/identity_spec.rb
@@ -69,4 +69,24 @@ RSpec.describe Datadog::Core::Environment::Identity do
 
     it { is_expected.to eq(Datadog::Core::Environment::Ext::TRACER_VERSION) }
   end
+
+  describe '::tracer_version_semver2' do
+    subject(:tracer_version) { described_class.tracer_version_semver2 }
+
+    context 'when not prerelease' do
+      before do
+        expect(described_class).to receive(:tracer_version).and_return('10.20.30')
+      end
+
+      it { is_expected.to eq('10.20.30') }
+    end
+
+    context 'when prerelease' do
+      before do
+        expect(described_class).to receive(:tracer_version).and_return('10.20.30.beta40')
+      end
+
+      it { is_expected.to eq('10.20.30-beta40') }
+    end
+  end
 end
--- a/spec/datadog/core/environment/identity_spec.rb
+++ b/spec/datadog/core/environment/identity_spec.rb
@@ -88,5 +88,21 @@ RSpec.describe Datadog::Core::Environment::Identity do
 
       it { is_expected.to eq('10.20.30-beta40') }
     end
+
+    context 'when development' do
+      before do
+        expect(described_class).to receive(:tracer_version).and_return('10.20.30.gha12345.ga1b2c3d4.a.branch.name')
+      end
+
+      it { is_expected.to eq('10.20.30+gha12345.ga1b2c3d4.a-branch-name') }
+    end
+
+    context 'when prerelease and development' do
+      before do
+        expect(described_class).to receive(:tracer_version).and_return('10.20.30.beta40.gha12345.ga1b2c3d4.a.branch.name')
+      end
+
+      it { is_expected.to eq('10.20.30-beta40+gha12345.ga1b2c3d4.a-branch-name') }
+    end
   end
 end
--- a/sig/datadog/core/environment/identity.rbs
+++ b/sig/datadog/core/environment/identity.rbs
@@ -4,19 +4,21 @@ module Datadog
       module Identity
         extend Datadog::Core::Utils::Forking
 
-        def self?.id: () -> untyped
+        def self?.id: () -> ::String
 
-        def self?.lang: () -> untyped
+        def self?.lang: () -> ::String
 
-        def self?.lang_engine: () -> untyped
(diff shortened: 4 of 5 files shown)


lib/datadog/core/environment/identity.rb
```diff
@@ -49,9 +49,18 @@ def lang_version
...
         def tracer_version
           Core::Environment::Ext::TRACER_VERSION
         end
+
+        # Returns tracer version, comforming to https://semver.org/spec/v2.0.0.html
+        def tracer_version_semver2
+          # from ddtrace/version.rb, we have MAJOR.MINOR.PATCH plus an optional .PRE
+          # - transform .PRE to -PRE if present
+          # - keep triplet before that
+          tracer_version.sub(/\.([a-zA-Z0-9]*[a-zA-Z][a-zA-Z0-9]*.*)$/, '-\1')
```
GustavoCaso, line 62: I find the regex a little grok (Probably because I never took the time to master it)

Wondering if having a hardcoded constant in `version.rb` would do the trick:

```ruby
  STRING = [MAJOR, MINOR, PATCH, PRE].compact.join('.')
  SEMVER2 = PRE ? "#{MAJOR}.#{MINOR}.#{PATCH}-#{PRE}" : STRING
```

GustavoCaso (approved):
Left comment suggesting an alternative approach. Still, not a blocker, but at least a way to open a discussion 😄

ivoanjo:
Wait -- I'd like to suggest an alternative.

ivoanjo:
This change seems a bit weird -- have a separate version string for reporting to telemetry that is different from the one we use.

Here's my suggestion -- since the cat is "out of the bag" anyway for `1.11.0.beta1` and this won't retroactively fix it, my suggestion is -- let's add a _spec_  that enforces that the current version must be semver 2 compatible.

This way, we'll remember to not use `.betaX` again, and instead use `-betaX` (because the test will break during a release) and we avoid creating a sharp edge of a different version in different parts of the product.

What do you think?

marcotc (changes requested):
I like @ivoanjo's suggestion.

For tests to pass immediately, I suggest bumping `version.rb` to `1.11.0` or `1.11.0-beta2` (even though we don't normally do that at this time).

lloeki:
Problem is, the Semantic Version 2.0.0 (thereafter shortened as semver2) pattern is at odds with the RubyGems version pattern. Indeed, the former can be processed this way, with `+` > `-` > `.` as parsing precedence order:

- split once on first `+`, RHS is a freeform "build information" string that is _not_ considered for ordering
- split the LHS of the above once on first `-`, RHS is pseudo-freeform "prerelease version" string that is considered for ordering
- split the LHS of the above on `.`, this is MAJOR.MINOR.PATCH

Comparatively, RubyGems expects only one separator, which is `.`:

```
irb(main):001:0> v = Gem::Version.new('1.11.0.beta1')
=> Gem::Version.new("1.11.0.beta1")
irb(main):002:0> v.segments
=> [1, 11, 0, "beta", 1]
irb(main):003:0> v.to_s
=> "1.11.0.beta1"
```

If you stuff a `-` in there, `Gem::Version` will helpfully parse it, but it will end up as something else:

```
irb(main):004:0> v = Gem::Version.new('1.11.0-beta1')
=> Gem::Version.new("1.11.0.pre.beta1")
irb(main):005:0> v.segments
=> [1, 11, 0, "pre", "beta", 1]
irb(main):006:0> v.to_s
=> "1.11.0.pre.beta1"
```

So, really `1.11.0-beta1` _never_ exists besides being accepted as input. Only the dotted `.pre.` version ever exists, and it would (presumably) show up as that on some or all of `rubygems.org`, `gem install`, `bundle`, `Gemfile.lock`, with a very inconsistent experience.

It also gives way to surprising results:

```
irb(main):009:0> Gem::Requirement.new('< 1.11.0.c').satisfied_by?(Gem::Version.new('1.11.0.beta1'))
=> true
irb(main):010:0> Gem::Requirement.new('< 1.11.0.c').satisfied_by?(Gem::Version.new('1.11.0-beta1'))
=> false
```

This is already awkward to have a non-identity version string, but it gets further.

> a separate version string for reporting to telemetry

FYI this is not for telemetry, this is for backend RC to be able to apply predicates on versions to decide to do "stuff", hence why respecting semver 2.0 strictly is important, including the part about `+<build info>` being unordered.

And as it turns out, `+` is entirely unprocessed:

```
irb(main):007:0> Gem::Version.new('1.11.0-beta1+g1a2b3c4d')
/nix/store/pjm178yi50xx14nb67wrh1sq5dzj3f1d-ruby-3.1.2/lib/ruby/3.1.0/rubygems/version.rb:223:in `initialize': Malformed version number string 1.11.0-beta1+g123456 (ArgumentError)                   
        from /nix/store/pjm178yi50xx14nb67wrh1sq5dzj3f1d-ruby-3.1.2/lib/ruby/3.1.0/rubygems/version.rb:206:in `new'
        from /nix/store/pjm178yi50xx14nb67wrh1sq5dzj3f1d-ruby-3.1.2/lib/ruby/3.1.0/rubygems/version.rb:206:in `new'
        from (irb):7:in `<main>'                   
        from /nix/store/pjm178yi50xx14nb67wrh1sq5dzj3f1d-ruby-3.1.2/lib/ruby/gems/3.1.0/gems/irb-1.4.1/exe/irb:11:in `<top (required)>'
        from /Users/loic.nageleisen/.nix-profile/bin/irb:25:in `load'
	from /Users/loic.nageleisen/.nix-profile/bin/irb:25:in `<main>'
```

Where it matters is for our prerelease builds, we use branches, and we'd have to conform to semver2 with the + part, so we'd have the build id + sha after a `+` for RC to be able to process it correctly, but we can't use that Ruby-side.

To sum this up:

- `1.11.0-beta1` as a string works but is:
  - only an input string
  -  not idiomatic to Ruby (neither technically nor community-wise)
  -  fraught with surprises and perils
- `1.11.0+foo` as a string flat out does not work, so we'd have to have two versions anyway

Therefore I veered on the side of having:

- Ruby things be Ruby idioms on Ruby land
- things being semver only when they are a) known compatible and b) absolutely required (so, RC)

Therefore, even though I appreciate and understand the intent to keep things simple with a unique string here, I believe† it is ultimately a source of many issues: we should keep the dotted version, with a known semantic pattern, and extract and convert to other version formats at the edge as required.

† A.K.A "I make nightmares about my journeys into bundler's dependency resolution code"

lloeki:
I hesitated making that semver2 function a private method inside of `Core::Remote::Client`, maybe I should have to make the intent clearer

ivoanjo:
If I understand your analysis correctly, @lloeki, what you're stating is that in any future betas the correct thing to do will still be to use `.betaX`.

If that's the case, I think your great analysis points at something: that the current parsing restrictions on the backend are not applicable for Ruby. And thus, I think it's time to raise this as a concern for potentially modifying in the backend before we commit to "lying" to it just to make it happy.

It seems to me this is a case where the backend should be the one that "bends" to accommodate the idiomatic-Ruby practices of the library.

Yes, it would be great if we could be strictly follow semver2, but since we can't...

lloeki:
> If I understand your analysis correctly, @lloeki, what you're stating is that in any future betas the correct thing to do will still be to use .betaX.

Correct.

It goes further, it turns out we enforce MAJOR.MINOR.PATCH through constants, but Ruby-side we're at liberty of having as many components as we'd like, e.g what if we want to introduce a versioning scheme for hotfixes such as MAJOR.MINOR.PATCH.HOTFIX, e.g 1.11.0.1 that would be applied even for someone pinning at PATCH level using the squiggly operator `~> 1.11.0.0`. The backend would be unable to distinguish between 1.11.0 and 1.11.0.1 since we can't send the HOTFIX value.

lloeki:
Ultimately we can view this as a normalizing stage, which happens to be lossy with semver2. For other platforms similar problems arise:

- .Net has [assemblies versioned](https://learn.microsoft.com/en-us/dotnet/standard/assembly/versioning) as MAJOR.MINOR.BUILD.REVISION, which can become MAJOR.MINOR.REVISION+BUILD, maybe?
- Python has [PEP-440](https://peps.python.org/pep-0440/), which defines final releases (with arbitrary count segments), prereleases, dev releases, and post releases. Arbitrary count segments and post releases are not encodeable using semver2 without loss.

marcotc:
It reads to me that the backend has to be able to faithfully match the complete Ruby version string (look like it's the same case for Python and .Net). Without complete version matching, we open ourselves to possible version matching issues.

But because the backend doesn't support that today, we'll temporarily implement this `tracer_version_semver2` converter. Is that correct?

lib/datadog/core/environment/identity.rb
marcotc, line 62: We probably need support for `BUILD`, introduced in https://github.com/DataDog/dd-trace-rb/pull/2800.

marcotc: I find this pattern more legible as well.

lloeki:
> It reads to me that the backend has to be able to faithfully match the complete Ruby version string (looks like it's the same case for Python and .Net, where more complex matching than semver is needed). Without complete version matching, we open ourselves to possible version matching issues.

Correct. We met internally with the backend team. Decision was to:

- have our versioning constrained to be compatible with semver 2.0 _semantics_ so that no loss occurs when converting from ruby version to semver 2.0 _syntax_.
- implement exact gem version string reporting via tags https://github.com/DataDog/dd-trace-rb/pull/2811 which is what should be used should one want to _display_ the version, the semver one being only used internally for version constraint evaluation.

spec/datadog/core/environment/identity_spec.rb
```diff
@@ -69,4 +69,40 @@
...
+
+      it { is_expected.to eq('10.20.30-beta40') }
+    end
+
+    context 'when development' do
+      before do
+        expect(described_class).to receive(:tracer_version).and_return('10.20.30.gha12345.ga1b2c3d4.a.branch.name')
+      end
+
+      it { is_expected.to eq('10.20.30+gha12345.ga1b2c3d4.a-branch-name') }
```
GustavoCaso, line 97: Validated with https://jubianchi.github.io/semver-check/#/version/10.20.30%2Bgha12345.ga1b2c3d4.a-branch-name

spec/datadog/core/environment/identity_spec.rb
```diff
@@ -69,4 +69,40 @@
...
+    end
+
+    context 'when prerelease and development' do
+      before do
+        expect(described_class).to receive(:tracer_version).and_return('10.20.30.beta40.gha12345.ga1b2c3d4.a.branch.name')
+      end
+
+      it { is_expected.to eq('10.20.30-beta40+gha12345.ga1b2c3d4.a-branch-name') }
```
GustavoCaso, line 105: Validated with https://jubianchi.github.io/semver-check/#/version/10.20.30-beta40%2Bgha12345.ga1b2c3d4.a-branch-name

spec/datadog/core/environment/identity_spec.rb
```diff
@@ -69,4 +69,40 @@
...
+
+      it { is_expected.to eq('10.20.30') }
+    end
+
+    context 'when prerelease' do
+      before do
+        expect(described_class).to receive(:tracer_version).and_return('10.20.30.beta40')
+      end
+
+      it { is_expected.to eq('10.20.30-beta40') }
```
GustavoCaso, line 89: Validated with https://jubianchi.github.io/semver-check/#/version/10.20.30-beta40

lib/datadog/core/remote/client.rb
```diff
@@ -119,7 +119,7 @@ def payload
           client_tracer = {
             runtime_id: Core::Environment::Identity.id,
             language: Core::Environment::Identity.lang,
-            tracer_version: Core::Environment::Identity.tracer_version,
+            tracer_version: Core::Environment::Identity.tracer_version_semver2,
```
GustavoCaso, line 122: I believe we should hoist the result value from it since running the regex for every request for remote configuration is going to be expensive

GustavoCaso (approved):
LGTM. I would make sure to avoid generating the semver version for each request

marcotc (approved):

lib/datadog/core/remote/client.rb
lloeki: Given the CI hell we're currently in I'll merge it as is and fix in a separate PR.

lloeki: PR: https://github.com/DataDog/dd-trace-rb/pull/2819

PR #2797, opened 2023-04-18
[APPSEC-9143] Extract expires information from target map payload.

GustavoCaso:
**What does this PR do?**

Propagate expires information to content instances. 
Make sure to remove content information from `config_state` when content has expired.

**Motivation**

The `config_states` information have to change based on expiration information 

**Additional Notes**

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/core/remote/configuration/repository.rb
+++ b/lib/datadog/core/remote/configuration/repository.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require 'date'
 require_relative 'content'
 
 module Datadog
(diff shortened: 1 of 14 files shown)


lib/datadog/core/remote/configuration/repository.rb
```diff
@@ -90,8 +91,12 @@ def initialize(repository)
             def contents_to_config_states(contents)
               return [] if contents.empty?
 
-              contents.map do |content|
-                {
+              contents.each_with_object([]) do |content, acc|
```
GustavoCaso, line 94: I had to change it to `each_with_object` because steep was complaining about using `map` and `next`

lib/datadog/core/remote/configuration/target.rb
```diff
@@ -9,34 +10,43 @@ module Remote
...
+              opaque_backend_state = signed['custom']['opaque_backend_state']
+              version = signed['version']
+              expires = signed['expires']
 
               map = new
 
               map.instance_eval do
                 @opaque_backend_state = opaque_backend_state
                 @version = version
+                @expires = DateTime.iso8601(expires) if expires
```
ivoanjo, line 28: Note that according to the [Ruby documentation](https://rubyapi.org/3.2/o/datetime#method-c-iso8601):

> [DateTime](https://rubyapi.org/3.2/o/datetime) class is considered deprecated. Use [Time](https://rubyapi.org/3.2/o/time) class.

:sweat_smile:

ivoanjo:
Is this PR still relevant?

PR #2800, opened 2023-04-19
Add workflow to publish gem on GitHub Packages

lloeki:
**What does this PR do?**

- Add workflow to publish gem on GitHub Packages
- Introduce machine-parsable semantics in the version, compatible with [Semantic Versioning 2.0.0](https://semver.org/spec/v2.0.0.html). This follows from the root cause leading to https://github.com/DataDog/dd-trace-rb/pull/2795, without which Remote Configuration payloads would be silently rejected by the backend should the reported version string not respect Semantic Versioning 2.0.0.

To this end, the prerelease version naming scheme was enhanced and refactored for CI usage:

  - Introduce a new `BUILD` component to the versioning scheme, that can be translated to `+<build>` in Semantic Versioning 2.0.0, as well as not leaking orderless build information in Semantic Versioning 2.0.0 `-<pre>`.
  - PRE is `alpha`, in order to be ordered before `beta` and `rc`; alternatives considered include `dev` and `ci`; in any case, a specific segment denoting this is both extremely useful for sorting and also required by Semantic Versioning 2.0.0 to fill in the `-<pre>` part and properly mark this as a prerelase.
  - BUILD starts with CI run id for convenience of ordering.
  - BUILD has CI run id for traceability, prefixed by `gha` for identification.
  - BUILD has commit next for traceability, prefixed git-describe style by `g` for identification.
  - BUILD has branch name last since it has to be separated by dots and thus has variable version segment size and unpredictable ordering; it can thus be reliably extracted and does not impair readability in lists

**Motivation**

- Our AWS S3 thing is not operational for CircleCI reasons
- It unconditionally publishes on every CircleCI run and builds accrete quickly, even when most of those are not useful
- The "dumb" S3 rubygems format makes index syncs slow and updates vulnerable to concurrency issues (download-add-upload cycle is racey)

**Additional Notes**

Using GitHub Packages and Actions offer many advantages:

- Secure integration with `GITHUB_TOKEN` for CI
- Manual trigger with UI via `workflow_dispatch` 
- Easy artifact archival and download per CI run

Regarding the conversion to Semantic Versioning 2.0.0, it is unlikely the backend will move to implement per-language version semantics and views Semantic Versioning 2.0.0 as a normalization tool. Whether the conversion happens in the backend or in the library, the Ruby version number must be semantically viable and processable for a conversion.

**How to test the change?**

- CI should build the gem and produce a downloadable and installable artifact.
- Actual publishing via manual workflow is a bit trickier and can only be tested once the merge happens. Testing the push was done by temporarily circumventing the `if` condition, the result is visible [here for the package list](https://github.com/DataDog/dd-trace-rb/packages?ecosystem=rubygems) and [here for the specific package](https://github.com/DataDog/dd-trace-rb/pkgs/rubygems/ddtrace).


.github/workflows/build-gem.yml
```diff
@@ -0,0 +1,138 @@
+name: Build gem
```
ivoanjo, line 1: Minor: Perhaps 

```suggestion
name: Package experimental gem
```

to make it a bit more clear what we're doing here?

.github/workflows/build-gem.yml
```diff
@@ -0,0 +1,138 @@
...
+          # Obtain context information
+          git_ref='${{ github.ref }}'
+          git_branch="$(echo "${git_ref}" | sed -e 's#^refs/heads/##')"
+          git_sha='${{ github.sha }}'
+          gha_run_id='${{ github.run_id }}'
+
+          # Output info for CI debug
+          echo git_ref="${git_ref}"
+          echo git_branch="${git_branch}"
+          echo git_sha="${git_sha}"
```
ivoanjo, line 40: Minor: I noticed this results in really long versions (since the sha is quite big) -- gha4745388214.g2f501acf6b5a9d41bb800eb091474cf7961d6f62.use.github.packages.for.gems .

I would suggest shortening to the first [n = 12 chars](https://stackoverflow.com/questions/18134627/how-much-of-a-git-sha-is-generally-considered-necessary-to-uniquely-identify-a) :)

ivoanjo (commented):
There's a couple of CI steps breaking, you may want a look. Otherwise this seems reasonable, although it still needs the review from security that we discussed IRL.

Finally, just a question: what do you think is still missing? E.g. if you had infinite time on this, what would you change? (I'm curious if there's some later improvement we should keep on the radar)

.github/workflows/build-gem.yml
lloeki: An improvement I wanted to make was building both the alpha gem and the final gem.

This way when doing an actual release:

- we do not have to build the gem locally
- we are confident that the artifact is produced from a clean environment, and fully tested (more tests than `gem install` should probably be added)
- we can download it straight from the artifact, and eventually even have it added automatically (by that I mean something like a manual workflow instead of download+drag+drop, not necessarily _entirely_ automatically) to the GitHub release entry.

.github/workflows/build-gem.yml
lloeki: I thought about that as well. I will do it.

I was also thinking about shortening the branch name by having it hashed + taking a prefix as well. This way we could keep the last part for humans but still be able to put a stable branch-originating segment anywhere we'd want (I was thinking about per branch grouping without having that big dotted name in the middle)

.github/workflows/build-gem.yml
```diff
@@ -0,0 +1,160 @@
...
+  GEM_HOST: 'https://rubygems.pkg.github.com/DataDog'
+
+jobs:
+  build:
+    strategy:
+      fail-fast: false
+      matrix:
+        type:
+          - final
+          - alpha
```
lloeki, line 25: Regarding this `alpha` name:

I used to find it fitting semantically: we do `beta` as part of a manual release process, so these versions are more `alpha` and thus should sort before? But I'm not so convinced anymore that such ordering matters, or is even consistent with reality.

I also found it lacking, because `1.42.0.alpha.foobar` kind of makes one think that `1.42.0` is the next version when it is actually the last one released.

That's why I also liked `dev` (because they emanate from unmerged branches) or `ci` (because they are built only from CI).

Examples to compare and contrast (taking `1.10.1` instead of the current `1.11.0.beta1` which muddies the thing):

- `1.10.1.alpha.gha4745620001.gfe6c0807f40b.foo.bar.branch.name`
- `1.10.1.ci.gha4745620001.gfe6c0807f40b.foo.bar.branch.name`
- `1.10.1.dev.gha4745620001.gfe6c0807f40b.foo.bar.branch.name`

So I'm increasingly inclined to move from `alpha` to either `dev` or `ci`. WDYT?

GustavoCaso: @lloeki Thanks for the reasoning. 

I'm more inclined to `dev`. For me, it stipulates that this version is still under development and is not ready to be used in production and is just for development purposes

.github/workflows/build-gem.yml
```diff
@@ -0,0 +1,160 @@
...
+          PRE='${{ matrix.type }}'
+          BUILD="gha${gha_run_id}.g${git_sha}.${git_branch_sanitized}"
+
+          # Output info for CI debug
+          echo PRE="${PRE}"
+          echo BUILD="${BUILD}"
+
+          # Patch in components
+          sed lib/ddtrace/version.rb -i -e "s/^\([\t ]*PRE\) *= */\1 = \'${PRE}\' # /"
+          sed lib/ddtrace/version.rb -i -e "s/^\([\t ]*BUILD\) *= */\1 = \'${BUILD}\' # /"
```
GustavoCaso, line 78: What do you think about using ENV variables rather than manipulating files? 

```ruby
module DDTrace
  module VERSION
    MAJOR = 1
    MINOR = 11
    PATCH = 0
    PRE = ENV['PRE_VERSION'] || 'beta1'
    BUILD = ENV['BUILD_VERSION'] 

    STRING = [MAJOR, MINOR, PATCH, PRE, BUILD].compact.join('.')
end
```

The `Patch version` job could store those values, and on the `Build gem` job, we could use them. 

`PRE_VERSION=${}, BUILD_VERSION=${} bundle exec rake build`

This pattern could be used as well for `Patch gem host`

.github/workflows/build-gem.yml
lloeki: Agreed. Done.

.github/workflows/build-gem.yml
ivoanjo: This is interesting! For that we'd need to tweak the version numbering logic, but otherwise seems like this PR would leave us in a good shape :+1:

.github/workflows/build-gem.yml
ivoanjo: Classic meta-programing oopsie heh :)

.github/workflows/build-gem.yml
ivoanjo: I kinda like having the full branch name since it's more obvious a) it's not a regular release and b) what that release has in particular.

.github/workflows/build-gem.yml
```diff
@@ -0,0 +1,165 @@
...
+          # - BUILD has CI run id for traceability, prefixed by `gha`
+          #   for identification.
+          # - BUILD has commit next for traceability, prefixed git-describe
+          #   style by `g` for identification.
+          # - BUILD has branch name last since it has to be separated
+          #   by dots and thus has variable version segment size and
+          #   unpredictable ordering; it can thus be reliably extracted
+          #   and does not impair readability in lists
+          PRE='${{ matrix.type }}'
+          BUILD="gha${gha_run_id}.g${git_sha_short}.${git_branch_sanitized}"
```
ivoanjo, line 71: Minor, looking at your comment in another thread:

> - `1.10.1.alpha.gha4745620001.gfe6c0807f40b.foo.bar.branch.name`
> - `1.10.1.ci.gha4745620001.gfe6c0807f40b.foo.bar.branch.name`
> - `1.10.1.dev.gha4745620001.gfe6c0807f40b.foo.bar.branch.name`

Is it for me that the `g` in front of the sha hash a bit potentially confusing since it looks like part of the hash? (I know the hash is hex but lol it still looks like part of it visually, or if you double-click to select it)

So I'd suggest removing it maybe, but up to you :)

lloeki: As I mentioned, this is inspired by the format of `git describe`:

```
$ git describe --tags
v1.11.0.beta1-9-g2fcff4acc
```

The `g` prefix denotes that this is a git sha, so that it can readily be identified among other pieces of data that could look like a git sha (e.g I was suggesting hashing the branch name).

Note: I prefixed the run id with `gha` for similar reasons.

ivoanjo: It's not particularly obvious to me, but since we need to pick something, let's go with your approach ;)

lib/ddtrace/version.rb
```diff
@@ -6,8 +6,9 @@ module VERSION
     MINOR = 11
     PATCH = 0
     PRE = 'beta1'
+    BUILD = nil
```
marcotc, line 9: I recommend leaving a comment here saying that PRE and BUILD can be dynamically changed during our gem build GH Action.

.circleci/config.yml
```diff
@@ -176,21 +176,6 @@ step_run_all_tests: &step_run_all_tests
         " >> .rspec-local
 
         bundle exec rake ci
-step_release_docs: &step_release_docs
-  run:
-    name: Upload release docs
-    command: S3_DIR=trace bundle exec rake release:docs
```
marcotc, line 182: Can you also delete the respective rake task `release:docs`?

marcotc (approved):
This is awesome!
Just left a few nits to address.

.circleci/config.yml
lloeki: Done

lib/ddtrace/version.rb
lloeki: Done

PR #2801, opened 2023-04-19
Remote configuration handle Client SyncError

GustavoCaso:
**What does this PR do?**

Whenever a Client::SyncError is raised when calling `client.sync` the remote component rescues it, logs the sync error information and continues with the execution.

The remote configuration backend expects to maintain the state between requests even if there is a SyncError.

In the case of an unhandled error, we are still resting the client's state.

**Motivation**

**Additional Notes**

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/core/remote/component.rb
+++ b/lib/datadog/core/remote/component.rb
@@ -30,9 +30,14 @@ module Datadog
           @worker = Worker.new(interval: settings.remote.poll_interval_seconds) do
             begin
               @client.sync
+            rescue Client::SyncError => e
+              Datadog.logger.error do
+                "remote worker client sync error: #{e.message} location: #{Array(e.backtrace).first}. skipping sync"
+              end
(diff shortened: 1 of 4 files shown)


lib/datadog/core/remote/component.rb
```diff
@@ -30,9 +30,14 @@ def initialize(settings, agent_settings)
           @worker = Worker.new(interval: settings.remote.poll_interval_seconds) do
             begin
               @client.sync
+            rescue Client::SyncError => e
+              Datadog.logger.error do
+                "remote worker client sync error: #{e.message} location: #{Array(e.backtrace).first}. skipping sync"
```
ivoanjo, line 35: Minor: I suggest still including the error class, as in the future we may actually catch a subclass of `SyncError` and we'd be throwing away that information

lloeki (approved):
LGTM

PR #2803, opened 2023-04-20
Fix steep parallel crash

lloeki:
**What does this PR do?**

Fix steep parallel crash

**Motivation**

Parallell 1.23 seems to annoy steep:

```
cannot load such file -- parallel/processor_count (LoadError)
```

**Additional Notes**

Maybe an upgrade to steep could help.

**How to test the change?**

This should not crash:

```
bundle exec steep check
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -128,6 +128,10 @@ group :check do
   if RUBY_VERSION >= '2.7.0' && RUBY_PLATFORM != 'java' && !steep_ci_workaround
     gem 'rbs', '~> 2.8.1', require: false
     gem 'steep', '~> 1.3.1', require: false
+
+    # parallel 1.23 seems to annoy steep:
+    # cannot load such file -- parallel/processor_count (LoadError)
+    gem 'parallel', '< 1.23'
   end
 end
 


GustavoCaso (approved):

Gemfile
```diff
@@ -128,6 +128,10 @@ group :check do
   if RUBY_VERSION >= '2.7.0' && RUBY_PLATFORM != 'java' && !steep_ci_workaround
     gem 'rbs', '~> 2.8.1', require: false
     gem 'steep', '~> 1.3.1', require: false
+
+    # parallel 1.23 seems to annoy steep:
```
marcotc, line 132: Can you add the current steep version to the comment, so we can easily, in the future, judge if steep has changed enough versions for us to give it a try at removing this version lock?

marcotc (approved):

Gemfile
lloeki: Done

lloeki: Duh, I did not add the change to the index, so `--amend` did not pick it up:

```
diff --git a/Gemfile b/Gemfile
index 3c920f125..2bbd0059a 100644
--- a/Gemfile
+++ b/Gemfile
@@ -129,7 +129,7 @@ group :check do
     gem 'rbs', '~> 2.8.1', require: false
     gem 'steep', '~> 1.3.1', require: false
 
-    # parallel 1.23 seems to annoy steep:
+    # parallel 1.23 seems to annoy steep 1.3.2:
     # cannot load such file -- parallel/processor_count (LoadError)
     gem 'parallel', '< 1.23'
   end
```

I'll add it later.

PR #2805, opened 2023-04-21
[APPSEC-9177] Enable Remote configuration by default

GustavoCaso:
**What does this PR do?**

Even with remote configuration enabled by default,  on this version of remote configuration and appsec. 
If AppSec is disabled, the remote configuration worker thread won't start.

Since in AppSec, we do not support the feature of remote activation. There is no need to have a background thread opening HTTP connections to the agent to request remote configuration information.

This might change once we support ASM remote activation or other components, such as tracer, start using remote configuration since there would be remote configuration information to listen to. 

**Motivation**

**Additional Notes**

**How to test the change?**

CI


lloeki (approved):
LGTM

PR #2807, opened 2023-04-21
Negotiate /info later

lloeki:
**What does this PR do?**

- Extract negotiation out of the Remote component
- Reach to the `/info` endpoint later in the Remote component lifecycle

**Motivation**

- API endpoint negotiation is independent from Remote Configuration
- The check is very early and should be done much later in the lifecycle, notably:
  - when RC is live and healthy, be redone in case of sudden errors (e.g agent configuration changed)
  - when RC is started but failing to connect, retried regularly (e.g missing agent that starts later, agent has been upgraded or reconfigured to enable remote configuration)

**Additional Notes**

Nope

**How to test the change?**

Start and/or restart the agent at various points in the app lifecycle (before, after), with different configurations or versions of the agent.

### DIFF
--- a/spec/datadog/core/remote/component_spec.rb
+++ b/spec/datadog/core/remote/component_spec.rb
@@ -11,6 +11,10 @@ RSpec.describe Datadog::Core::Remote::Component do
   describe '.build' do
     subject(:component) { described_class.build(settings, agent_settings) }
 
+    after do
+      component.shutdown! if component
+    end
+
     context 'remote disabled' do
       let(:remote) do
         mock = double('remote')
@@ -130,6 +134,10 @@ RSpec.describe Datadog::Core::Remote::Component do
   describe '#initialize' do
     subject(:component) { described_class.new(settings, capabilities, agent_settings) }
 
+    after do
+      component.shutdown!
+    end
+
     context 'worker' do
       let(:worker) { component.instance_eval { @worker } }
       let(:client) { double }
@@ -143,10 +151,6 @@ RSpec.describe Datadog::Core::Remote::Component do
 
         expect(worker).to receive(:start).and_call_original
         expect(worker).to receive(:stop).and_call_original
-      end
-
-      after do
-        component.shutdown!
       end
 
       context 'when client sync succeeds' do


lib/datadog/core/remote/component.rb
```diff
@@ -20,23 +21,36 @@ def initialize(settings, agent_settings)
           transport_options = {}
           transport_options[:agent_settings] = agent_settings if agent_settings
 
+          negotiation = Negotiation.new(settings, agent_settings)
```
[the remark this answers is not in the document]
marcotc: I can see the tracing transport needing negotiation to find out what version of the flushing API the agent supports.

lib/datadog/core/remote/component.rb
```diff
@@ -20,23 +21,36 @@ def initialize(settings, agent_settings)
...
           @client = Client.new(transport_v7, capabilities)
+          healthy = false
+          Datadog.logger.debug { "new remote configuration client: #{@client.id}" }
+
           @worker = Worker.new(interval: settings.remote.poll_interval_seconds) do
+            unless healthy || negotiation.endpoint?('/v0.7/config')
+              @barrier.lift
+
+              next
+            end
```
GustavoCaso, line 40: It looks like we want to lift the barrier only when the client is unhealthy and the agent does not respond to the config endpoint: 

```ruby
irb(main):059:1* unless false || true
irb(main):060:1*   puts "hello"
irb(main):061:0> end
=> nil
irb(main):062:1* unless true || true
irb(main):063:1*   puts "hello"
irb(main):064:0> end
=> nil
irb(main):065:1* unless true || false
irb(main):066:1*   puts "hello"
irb(main):067:0> end
=> nil
irb(main):068:1* unless false || false
irb(main):069:1*   puts "hello"
irb(main):070:0> end
hello
```

Are we doing that to unblock callers of `Datadog::Core::Remote.active_remote.barrier(:once)` in case the agent does not support remote configuration?

lib/datadog/core/remote/negotiation.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require_relative '../transport/http'
+
+module Datadog
+  module Core
+    module Remote
+      class Negotiation
```
GustavoCaso, line 8: We should add unit tests for this class 😄

lib/datadog/core/remote/component.rb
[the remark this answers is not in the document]
lloeki: I kind of agree, although I'm not fond of actual singletons and globals. I was thinking of making it a component, like logging (If you squint you can kind of see it because of `new`'s signature ;)

> I can see the tracing transport needing negotiation to find out what version of the flushing API the agent supports.

Definitely, that is my thought. Actually that was my though even back when I worked on `core/transport` (where you can see I called the transport for `/info` `Negotiation`...)

It's a bigger change, that should be part of the general movement to refactor `ddtrace/transport` and `core/transport` so as a first step I put it isolated in `core/remote` because that was way out of scope for this PR.

lib/datadog/core/remote/component.rb
lloeki: > when the client is unhealthy and the agent does not respond to the config endpoint

Correct: `¬(A ∧ B) ⇔ ¬A ∨ ¬B`

lloeki: > Are we doing that to unblock callers of Datadog::Core::Remote.active_remote.barrier(:once) in case the agent does not support remote configuration?

Yep:

- It'll try to negotiate again and again until it's "healthy"
- a client is not considered "healthy" until the first sync happens
- If a client fails in such a way that we must recreate one, then it's back to the initial state and thus considered not healthy, and will try to negotiate again (because the failure may be that the agent was stopped or dropped RC)

lib/datadog/core/remote/negotiation.rb
lloeki: Sure. I was hoping to have these tested by the `Component` spec but it made more sense to stub the `endpoint?` over there so now `Negotiation` is not tested.

lloeki:
Rebased

marcotc (approved):

lib/datadog/core/remote/negotiation.rb
lloeki: I will address that in a PR tomorrow

lloeki: PR: https://github.com/DataDog/dd-trace-rb/pull/2818

PR #2808, opened 2023-04-24
Dalli quantize optimization

TonyCTHsu:
**What does this PR do?**

**Motivation**

**Additional Notes**

**How to test the change?**

### DIFF
--- a/lib/datadog/tracing/contrib/dalli/quantize.rb
+++ b/lib/datadog/tracing/contrib/dalli/quantize.rb
@@ -10,7 +10,21 @@ module Datadog
 
           def format_command(operation, args)
             placeholder = "#{operation} BLOB (OMITTED)"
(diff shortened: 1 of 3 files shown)


lib/datadog/tracing/contrib/dalli/quantize.rb
```diff
@@ -10,7 +10,21 @@ module Quantize
 
           def format_command(operation, args)
             placeholder = "#{operation} BLOB (OMITTED)"
-            command = [operation, *args].join(' ').strip
+            command = +operation.to_s
```
marcotc, line 13: [String#+@](https://ruby-doc.org/3.2.2/String.html#method-i-2B-40) is only supported in Ruby 2.3+.

ivoanjo:
This has been stale for quite a long time so I'm going to go ahead and close it.

PR #2811, opened 2023-04-25
Add remote configuration tags

lloeki:
**What does this PR do?**

Add remote configuration tags

**Motivation**

- version is mangled through semver, report the exact original gem version string
- ruby runtime information is gold to debug issues

**Additional Notes**

Ideally runtime information should go through telemetry but it is currently lacking in some departments.

**How to test the change?**

- Automatically: CI
- Manually: start a process with RC, go to the RC admin, and see the tags.

### DIFF
--- a/lib/datadog/core/remote/client.rb
+++ b/lib/datadog/core/remote/client.rb
@@ -118,12 +118,17 @@ module Datadog
             service: Datadog.configuration.service,
             env: Datadog.configuration.env,
             tags: [
+              "platform:#{native_platform}", # native platform
               "ruby.tracer.version:#{Core::Environment::Identity.tracer_version}",
               "ruby.runtime.platform:#{RUBY_PLATFORM}",
               "ruby.runtime.version:#{RUBY_VERSION}",
               "ruby.runtime.engine.name:#{RUBY_ENGINE}",
               "ruby.runtime.engine.version:#{defined?(RUBY_ENGINE_VERSION) ? RUBY_ENGINE_VERSION : RUBY_VERSION}",
               "ruby.rubygems.platform.local:#{Gem::Platform.local}",
+              "ruby.gem.libddwaf.version:#{Gem.loaded_specs['libddwaf'].version}",
+              "ruby.gem.libddwaf.platform:#{Gem.loaded_specs['libddwaf'].platform}",
+              "ruby.gem.libdatadog.version:#{Gem.loaded_specs['libdatadog'].version}",
+              "ruby.gem.libdatadog.platform:#{Gem.loaded_specs['libdatadog'].platform}",
             ],
           }
 
@@ -152,6 +157,45 @@ module Datadog
             cached_target_files: state.cached_target_files,
           }
         end
+
+        def native_platform
+          os = if RUBY_ENGINE == 'jruby'
+                 os_name = java.lang.System.get_property('os.name')
+
+                 case os_name
+                 when /linux/i then 'linux'
+                 when /mac/i   then 'darwin'
+                 else raise PlatformError, "unsupported JRuby os.name: #{os_name.inspect}"
+                 end
+               else
+                 Gem::Platform.local.os
+               end
+
+          version = if os != 'linux'
+                      nil
+                    elsif RUBY_PLATFORM =~ /linux-(.+)$/
+                      # Old rubygems don't handle non-gnu linux correctly
+                      Regexp.last_match(1)
+                    else
+                      'gnu'
+                    end
+
+          cpu = if RUBY_ENGINE == 'jruby'
+                  os_arch = java.lang.System.get_property('os.arch')
+
+                  case os_arch
+                  when 'amd64' then 'x86_64'
+                  when 'aarch64' then os == 'darwin' ? 'arm64' : 'aarch64'
+                  else raise PlatformError, "unsupported JRuby os.arch: #{os_arch.inspect}"
+                  end
+                else
+                  Gem::Platform.local.cpu
+                end
+
+          [cpu, os, version].compact.join('-')
+        end
+
+        class PlatformError < RuntimeError; end
       end
     end
   end
--- a/lib/datadog/core/remote/client.rb
+++ b/lib/datadog/core/remote/client.rb
@@ -108,7 +108,7 @@ module Datadog
 
         private
 
-        def payload
+        def payload # rubocop:disable Metrics/MethodLength
           state = repository.state
 
           client_tracer = {
@@ -117,7 +117,14 @@ module Datadog
             tracer_version: Core::Environment::Identity.tracer_version,
             service: Datadog.configuration.service,
             env: Datadog.configuration.env,
-            tags: [], # TODO: add nice tags!
+            tags: [
+              format('ruby.tracer.version:%s', Core::Environment::Identity.tracer_version),
+              format('ruby.runtime.platform:%s', RUBY_PLATFORM),
+              format('ruby.runtime.version:%s', RUBY_VERSION),
+              format('ruby.runtime.engine.name:%s', RUBY_ENGINE),
+              format('ruby.runtime.engine.version:%s', RUBY_ENGINE_VERSION),
+              format('ruby.rubygems.platform.local:%s', Gem::Platform.local.to_s),
+            ],
           }
 
           app_version = Datadog.configuration.version
--- a/lib/datadog/core/remote/client.rb
+++ b/lib/datadog/core/remote/client.rb
@@ -111,25 +111,29 @@ module Datadog
         def payload # rubocop:disable Metrics/MethodLength
           state = repository.state
 
+          client_tracer_tags = [
+            "platform:#{native_platform}", # native platform
+            # "asm.config.rules:#{}", # TODO: defined|undefined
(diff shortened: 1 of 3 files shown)


lloeki:
Duh, `RUBY_ENGINE_VERSION` does not exist on Ruby <= 2.2 🤦

lib/datadog/core/remote/client.rb
```diff
@@ -117,7 +117,17 @@ def payload
             tracer_version: Core::Environment::Identity.tracer_version,
             service: Datadog.configuration.service,
             env: Datadog.configuration.env,
-            tags: [], # TODO: add nice tags!
+            tags: [
+              format('ruby.tracer.version:%s', Core::Environment::Identity.tracer_version),
+              format('ruby.runtime.platform:%s', RUBY_PLATFORM),
```
marcotc, line 122: Sorry, but I had to suggest? Any reason to not use string interpolation?
```suggestion
              "ruby.runtime.platform:#{RUBY_PLATFORM}",
```

marcotc: Ok, I actually have a non-nit-picking point here:
```ruby
# frozen_string_literal: true
require 'benchmark/ips'

Benchmark.ips do |x|
  x.time = 5
  x.warmup = 0.5

  x.report('format') { format('ruby.runtime.platform:%s', RUBY_PLATFORM) }
  x.report('#{}') { "ruby.runtime.platform:#{RUBY_PLATFORM}" }

  x.compare!
end

# Ruby 3.2.1
# Comparison:
#                  #{}:  8359588.7 i/s
#               format:  3939488.9 i/s - 2.12x  slower
```

lib/datadog/core/remote/client.rb
```diff
@@ -117,7 +117,17 @@ def payload
...
+              format('ruby.runtime.version:%s', RUBY_VERSION),
+              format('ruby.runtime.engine.name:%s', RUBY_ENGINE),
+              format(
+                'ruby.runtime.engine.version:%s',
+                (defined?(RUBY_ENGINE_VERSION) ? RUBY_ENGINE_VERSION : RUBY_VERSION)
```
marcotc, line 127: Would it make sense to use:
```suggestion
                Datadog::Core::Environment::Ext::ENGINE_VERSION
```

lib/datadog/core/remote/client.rb
```diff
@@ -117,7 +117,17 @@ def payload
...
+              ),
+              format('ruby.rubygems.platform.local:%s', Gem::Platform.local.to_s),
```
marcotc, line 129: `ruby.rubygems.platform.local` or `ruby.gem.platform.local`? I'm not sure.

GustavoCaso: Technically the `Gem::Platform` class is part of the rubygems. But it is also true that `ruby.gem.platform.local` seems more comfortable to read. 

Naming is hard. If we use the tags to help filter on the dashboards, I would use `ruby.gem.platform.local` to simplify typing less characters.

Would it be possible to have a case where `RUBY_PLATFORM` and `Gem::Platform.local.to_s` differ? Asking, because if that is not the case, we might be able to remove one of those 😄

lloeki: Haha I had the same thoughts, but ultimately I was thinking of using the `ruby.gem.` prefix for _actual gems_, so the component after would be the gem name: `ruby.gem.<gem name>.<gem property>`. 

Examples:

- `ruby.gem.libddwaf.version`
- `ruby.gem.libddwaf.platform` # could be "from source" `ruby` gem platform or a binary gem platform like `arm64-darwin`
- `ruby.gem.libddwaf.loaded` # or something like that. Use case is: currently this should be `false` when above is `ruby` gem platform since we bail out and don't build from source, therefore even though the gem is there it's failing to provide any facility, so ASM is inoperative.
- `ruby.gem.libdatadog.version`
- `ruby.gem.libdatadog.platform`

So with this distinction you may understand why I put `ruby.rubygems.` there, `Gem::Platform` being actually part of `rubygems`.

lloeki: > Would it be possible to have a case where RUBY_PLATFORM and Gem::Platform.local.to_s

Yes, it is important to have both to be able to distinguish these cases (and many others):

- On old Ruby, `RUBY_PLATFORM == 'x86_64-linux'` even on `musl`
- On old rubygems, `RUBY_PLATFORM == 'x86_64-linux-musl'` becomes `Gem::Platform.local.to_s == 'x86_64-linux'`
- On darwin, `RUBY_PLATFORM == 'x86_64-darwin22'` becomes `Gem::Platform.local.to_s == 'x86_64-darwin-22'`
- On Alpine's apk ruby package, `RUBY_PLATFORM == 'x86_64-linux-musl'` but `Gem::Platform.local == 'ruby'`, forcing every gem to install from source

lib/datadog/core/remote/client.rb
lloeki: Very true!

lloeki: I'm going to archive that one to my [nascent "what's faster?" repo](https://github.com/lloeki/ruby-benchmarks) if you don't mind.

lloeki: Done

lib/datadog/core/remote/client.rb
lloeki: I'm not too sure, I thought about it but it feels very weird overall: these constants are kind of all over the place and there are helper methods in e.g `core/environment/identity`. It feels like this area is overdue for a refactoring.

I was hoping to punt and hope we'd drop 2.1 and 2.2 soon enough.

lloeki: Also I seem to recall these constants referencing constants referencing constants being returned by methods made some tests gnarly.

I understand the performance angle, but I was thinking maybe a solution would be to memoize behind methods in a dedicated instance. Sure a method call is a tiny bit slower than constant resolution, but then again we're calling `Core::Environment::Identity.lang*` (which I feel is weirdly designed and named, so, larger refactoring in perspective)

lib/datadog/core/remote/client.rb
```diff
@@ -108,16 +108,32 @@ def sync
...
+            "ruby.runtime.engine.name:#{RUBY_ENGINE}",
+            "ruby.runtime.engine.version:#{ruby_engine_version}",
+            "ruby.rubygems.platform.local:#{Gem::Platform.local}",
+            "ruby.gem.libddwaf.version:#{gem_spec('libddwaf').version}",
+            "ruby.gem.libddwaf.platform:#{gem_spec('libddwaf').platform}",
+            "ruby.gem.libdatadog.version:#{gem_spec('libdatadog').version}",
+            "ruby.gem.libdatadog.platform:#{gem_spec('libdatadog').platform}",
+          ]
+
           client_tracer = {
```
GustavoCaso, line 130: We could probably hoist `client_tracer` altogether since between requests the values would be the same. I noticed you already took care of hoisting  `ruby_engine_version`, `gem_spec`, and `native_platform`, but since the `payload` gets called for every request we might as well make sure to be as performant as possible 

The only values that might change are the ones in `state`

lib/datadog/core/remote/client.rb
```diff
@@ -145,6 +161,57 @@ def payload
...
+                      'gnu'
+                    end
+
+          cpu = if RUBY_ENGINE == 'jruby'
+                  os_arch = java.lang.System.get_property('os.arch')
+
+                  case os_arch
+                  when 'amd64' then 'x86_64'
+                  when 'aarch64' then os == 'darwin' ? 'arm64' : 'aarch64'
+                  else raise PlatformError, "unsupported JRuby os.arch: #{os_arch.inspect}"
```
GustavoCaso, line 203: We raise `PlatformError` and never rescue it. I'm worried this could lead to endless retries since the worker loop rescue `StandardError` and reset the client state. But we will end up with the same problem over and over.

Should we rescue it? 

Also, I see the test do not exercise the case where something fails, we should account for that as well in the tests

lib/datadog/core/remote/client.rb
GustavoCaso: This comment is not a blocker for this PR

lib/datadog/core/remote/client.rb
lloeki: Replaced by returning the original value.

lib/datadog/core/remote/client.rb
lloeki: > We could probably hoist client_tracer altogether since between requests the values would be the same.

Some keys of `client_tracer` may change if `Datadog.configure` is called multiple times and we decide to start the component at another time than the first request.

GustavoCaso: Not a blocker, but the only thing we are reading from the `configuration` is the service and the env. I'm not sure about those changes within a running app. I mean, technically they could, but I'm not they would.

GustavoCaso (approved):
LGTM

marcotc (approved):
We'll probably refactor much of this to reuse in Telemetry, because some of the fields are complex to calculate.

PR #2815, opened 2023-04-25
Telemetry:Disable for non-HTTP agent connection

marcotc:
This PR addresses the debug error log that happens when `ddtrace` is configured to communicate with the agent through UDS: `Unable to send telemetry event to agent: Failed to open TCP connection to :80 (Connection refused - connect(2) for nil port 80)`.

This PR does not enable the `Telemetry` client when UDS connection is used, instead disabling Telemetry.

### DIFF
--- a/lib/datadog/core/telemetry/http/transport.rb
+++ b/lib/datadog/core/telemetry/http/transport.rb
@@ -10,6 +10,7 @@ module Datadog
     module Telemetry
       module Http
         # Class to send telemetry data to Telemetry API
+        # Currently only supports the HTTP protocol.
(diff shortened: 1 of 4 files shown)


lib/datadog/core/telemetry/http/transport.rb
ivoanjo, line 1: Minor: Would be nice to also receive the `agent_settings` in initialize ;)

ivoanjo (approved):
:+1: LGTM, thanks for taking a look at this!

TonyCTHsu (approved):

PR #2817, opened 2023-04-26
make sure ENV variables are strings

GustavoCaso:
**What does this PR do?**

Fix the problem with our GitLab pipeline. It complains that the value definition must be a string or a hash

image

**Motivation**

**Additional Notes**

**How to test the change?**

### DIFF
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -52,33 +52,33 @@ only-profiling-benchmarks:
   extends: .benchmarks
   variables:
     DD_RELENV_CONFIGURATION: only-profiling
-    DD_PROFILING_ENABLED: true
+    DD_PROFILING_ENABLED: "true"
 
 profiling-and-tracing-benchmarks:
   extends: .benchmarks
   variables:
     DD_RELENV_CONFIGURATION: profiling-and-tracing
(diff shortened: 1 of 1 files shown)


.gitlab/benchmarks.yml
```diff
@@ -52,33 +52,33 @@ only-profiling-benchmarks:
...
-    DD_APPSEC_ENABLED: true
+    DD_APPSEC_ENABLED: "true"
 
 tracing-and-appsec-and-no-remote-configuration-benchmarks:
   extends: .benchmarks
   variables:
     DD_RELENV_CONFIGURATION: tracing-and-appsec-and-no-remote-configuration
-    DD_APPSEC_ENABLED: true
+    DD_APPSEC_ENABLED: "true"
     DD_REMOTE_CONFIGURATION_ENABLED: false
```
ivoanjo, line 74: I guess we should do the same for this one:

```suggestion
    DD_REMOTE_CONFIGURATION_ENABLED: "false"
```

ivoanjo (approved):
:+1: LGTM

.gitlab/benchmarks.yml
GustavoCaso: 🤦 You are right

PR #2819, opened 2023-04-27
Memoize tracer_version_semver2

lloeki:
**What does this PR do?**

Memoize tracer_version_semver2

**Motivation**

Followup to https://github.com/DataDog/dd-trace-rb/pull/2795#discussion_r1178058677

**Additional Notes**

Memoized in client instead of `Core::Environment::Identity` because the latter would make it global in the module singleton, and complicates testing. Also, none are memoized in there.

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/core/remote/client.rb
+++ b/lib/datadog/core/remote/client.rb
@@ -135,7 +135,7 @@ module Datadog
           client_tracer = {
             runtime_id: Core::Environment::Identity.id,
             language: Core::Environment::Identity.lang,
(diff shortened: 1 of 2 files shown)


GustavoCaso (approved):

ivoanjo:
Minor: Why not actually change `Datadog::Core::Environment::Identity.tracer_version_semver2` to a constant? It's not like the version itself changes during the lifetime of a Ruby process, so it feels a bit weird to cache it somewhere else instead of at the source.

PR #2820, opened 2023-04-27
fix flaky test by using expect API to make sure the original code doe…

GustavoCaso:
**What does this PR do?**

Fix flaky test 

The test break can be reproduced locally running `bundle exec appraisal ruby-3.2.0-core-old rake "spec:main[--seed 60499]"`

image

While working on the fix for the flaky test, I noticed we blindly call `finalize` on the old processor when calling `reconfigure`. There could be cases where the `old_processor` is `nil` . I added a test case and changed the code to avoid getting an unhandled exception in production.

**Motivation**

**Additional Notes**

**How to test the change?**

### DIFF
--- a/lib/datadog/appsec/component.rb
+++ b/lib/datadog/appsec/component.rb
@@ -53,7 +53,7 @@ module Datadog
           if new && new.ready?
             old = @processor
             @processor = new
-            old.finalize
+            old.finalize if old
           end
         end
       end
--- a/spec/datadog/appsec/remote_spec.rb
+++ b/spec/datadog/appsec/remote_spec.rb
@@ -164,12 +164,15 @@ RSpec.describe Datadog::AppSec::Remote do
 
         context 'when there is no ASM_DD information' do
           let(:transaction) { repository.transaction { |repository, transaction| } }
+
           it 'uses the rules from the appsec settings' do
+            ruleset = ''
             expect(Datadog::AppSec::Processor::RuleLoader).to receive(:load_rules).with(
               ruleset: Datadog.configuration.appsec.ruleset
-            ).at_least(:once).and_call_original
+            ).and_return(ruleset)
 
             changes = transaction
+            expect(Datadog::AppSec).to receive(:reconfigure).with(ruleset: ruleset)
             receiver.call(repository, changes)
           end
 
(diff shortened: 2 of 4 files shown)


spec/datadog/appsec/remote_spec.rb
```diff
@@ -164,12 +164,15 @@
...
+
           it 'uses the rules from the appsec settings' do
+            ruleset = ''
             expect(Datadog::AppSec::Processor::RuleLoader).to receive(:load_rules).with(
               ruleset: Datadog.configuration.appsec.ruleset
-            ).at_least(:once).and_call_original
+            ).and_return(ruleset)
 
             changes = transaction
+            expect(Datadog::AppSec).to receive(:reconfigure).with(ruleset: ruleset)
```
lloeki, line 175: `.and_call_original` or `.and_return(something)`?

spec/datadog/appsec/remote_spec.rb
```diff
@@ -164,12 +164,15 @@
 
         context 'when there is no ASM_DD information' do
           let(:transaction) { repository.transaction { |repository, transaction| } }
+
           it 'uses the rules from the appsec settings' do
+            ruleset = ''
```
lloeki, line 169: I'd use a bogus value like `'foo'` in there instead of an empty string.

lib/datadog/appsec/component.rb
```diff
@@ -53,7 +53,7 @@ def reconfigure(ruleset:)
           if new && new.ready?
             old = @processor
             @processor = new
-            old.finalize
+            old.finalize if old
```
lloeki, line 56: Oh, right, `create_processor` can return `nil` (and also theoretically if `shutdown!` is called multiple times)

lloeki: While checking whether the processor being nil is properly checked ([which it is](https://github.com/DataDog/dd-trace-rb/blob/3c84e41991b5b176b685e0047d14a802164830b7/lib/datadog/appsec/contrib/rack/request_middleware.rb#L40)), I noticed we [create an immediately thrown away processor](https://github.com/DataDog/dd-trace-rb/blob/3c84e41991b5b176b685e0047d14a802164830b7/lib/datadog/appsec/contrib/rack/request_middleware.rb#L31-L33)...

This might have come from an automatic merge resolution, and line 31 should be removed.

@GustavoCaso can you fix that in this PR?

lloeki (approved):
LGTM, but see the comments.

spec/datadog/appsec/remote_spec.rb
lloeki: NP, just wanted a clarification that it was not an oversight.

PR #2822, opened 2023-04-28
Properly check for remote configuration ASM content

GustavoCaso:
**What does this PR do?**

Could you check for remote configuration ASM content to decide whether the update includes rules override or exc

Added regression specs and improved the spec coverage for the rest of the AppSec remote products.

Today the remote configuration backend send ASM product updates that we can filter using the product path `config_id`. 

Example:
image

`policy-blocking-by-id-recommended` wasn't accounted for previously, so we failed to properly update the WAF with the information. Rather than checking the `config_id`, we are checking the content. If the content has the key `rules_override`, we treat it as an override, else we treat it as an exclusion. 

**Motivation**

**Additional Notes**

**How to test the change?**


lib/datadog/appsec/remote.rb
```diff
@@ -63,18 +65,20 @@ def receivers
...
-            asm_data_config_types = ['blocked_ips', 'blocked_users']
-            asm_overrides_config_types = ['blocking', 'disabled_rules']
-
             repository.contents.each do |content|
               case content.path.product
               when 'ASM_DD'
                 rules << parse_content(content)
               when 'ASM_DATA'
-                data << parse_content(content) if asm_data_config_types.include?(content.path.config_id)
+                data << parse_content(content) if ASM_DATA_CONFIG_TYPES.include?(content.path.config_id)
```
Julio-Guerra, line 73: ASM_DATA files need further merging, as explained [here](https://datadoghq.atlassian.net/wiki/spaces/APS/pages/2846623263/ASM+DATA+-+Blocked+IP+and+User+Data). Is it done somewhere else?

Julio-Guerra, line 73: Still filename-based instead of content-based with the field name `rules_data`

lib/datadog/appsec/remote.rb
```diff
@@ -63,18 +65,20 @@ def receivers
...
               when 'ASM'
-                overrides << parse_content(content) if asm_overrides_config_types.include?(content.path.config_id)
-                exclusions << parse_content(content) if content.path.config_id == 'exclusion_filters'
+                parsed_content = parse_content(content)
+
+                if parsed_content['rules_override']
+                  overrides << parsed_content
+                else
+                  exclusions << parsed_content
+                end
```
lloeki, line 81: I feel like something's amiss here, I'm expecting something more like:

```
               if parsed_content['rules_override']
                  overrides['rules_override'] << parsed_content['rules_override']
                end
                if parsed_content['exclusions']
                  exclusions['exclusions'] << parsed_content['exclusions']
                end
```

Otherwise overrides end up including exclusions instead of merging them in exclusions.

lib/datadog/appsec/remote.rb
lloeki: seems like we miss at least `custom_rules` on the ASM schema as well. The hotfix won't cut it: we need to audit this deeper.

lloeki (changes requested):

lloeki (approved):

PR #2824, opened 2023-04-28
Fix block response content negotiation

lloeki:
**What does this PR do?**

Reimplement response negotiation

This covers:
- unparseable Accept headers
- missing Accept headers
- proper ordering to handle priority
- defaulting to the first map element
- specs to avoid regressions

**Motivation**

A swapped argument prompted a deeper analysis of the response negotiation implementation, which uncovered many corner case issues.

**Additional Notes**

Specs were added to cover the most important main and corner cases 

**How to test the change?**

Perform a request that generates a blocking response, e.g (with remote configuration disabled) using:

```
    c.appsec.ip_denylist = [
      '127.0.0.1',
      '::1',
    ]
```

### DIFF
--- a/lib/datadog/appsec/assets/blocked.html
+++ b/lib/datadog/appsec/assets/blocked.html
@@ -1,4 +1,4 @@
-<!-- Sorry, you’ve been blocked -->
+<!-- Sorry, you've been blocked -->
 <!DOCTYPE html>
 <html lang="en">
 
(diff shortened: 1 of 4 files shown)


GustavoCaso (approved):
LGTM. Should we improve our integration tests from the contrib folder to exercise the different cases for the negotiation?

PR #2831, opened 2023-05-03
add APM Test Agent to CI for Integration Tests

wconti27:
**What does this PR do?**

This PR adds the APM Test Agent to run alongside integration/plugin tests within CI testing, handling all produced traces during a test run and asserting with any enabled checks. This change was made by modifying the FauxWriter Test Class to emit traces to the test agent.

 The APM Test Agent is being added to every DD APM Tracer with the primary purpose of providing a centralized location for writing tests which will subsequently be used for asserting on handled traces. This strategy will help bring consistency to Datadog APM Tagging/ integration support.  

**Motivation**
We're currently working on integrating the test agent into each tracer's test suites, and for the test agent to handle all traces produced by integrations during testing. Within the Test-Agent, we are then creating checks for Service Naming, Unified Naming Conventions tagging, etc in order to be a centralized location for asserting on integration traces. Allows us to leverage traces produced from integrations during testing.

**How to test the change?**
TO-DO: Verified manually within CircleCI, [such as here](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/10041/workflows/3309a3cf-2e3c-4bd0-b3a5-619022ac069d/jobs/372744/parallel-runs/5?filterBy=ALL&invite=true#step-10-1556), that integration traces are being emitted to and handled by the Test Agent.


spec/support/faux_writer.rb
```diff
@@ -20,19 +22,20 @@ def write(trace)
...
       super(trace) if @options[:call_original]
       @traces << trace
+      @options[:transport].send_traces(trace)
     end
   end
 
   def traces(action = :clear)
-    traces = @traces
-    @traces = [] if action == :clear
+    traces = @mutex.synchronize { @traces.dup }
```
marcotc, line 30: Why is this mutex necessary?

marcotc:
How does the "APM Test Agent" related to https://github.com/DataDog/system-tests? As far as I understand, system-tests are used to assert consistent behaviours across all tracers, and it does so by capturing data with a testing agent.

spec/support/faux_writer.rb
wconti27: accidental, removed this

wconti27:
> How does the "APM Test Agent" related to https://github.com/DataDog/system-tests? As far as I understand, system-tests are used to assert consistent behaviours across all tracers, and it does so by capturing data with a testing agent.

System Tests is different in the sense that its limited to only testing web frameworks and a select few integrations. What the Test Agent allows is for us to collect traces for all supported integrations across all tracers during testing (the Test Agent is being added at the moment to other tracers' integrations test suites as well), then assert on these traces with checks written centrally within the Test Agent. The Test Agent only has a few checks currently (mainly header checks like trace count and content length), but we'll be expanding it to have checks for Service Naming and Unified Naming for consistent tagging across tracers.

marcotc:
> The Test Agent only has a few checks currently (mainly header checks like trace count and content length), but we'll be expanding it to have checks for Service Naming and Unified Naming for consistent tagging across tracers.

What kind of signal do we get when one of these "checks" fail?

.circleci/config.yml
```diff
@@ -396,7 +436,7 @@ orbs:
         steps:
           - run:
               name: Wait for container on port <<parameters.port>>
```
GustavoCaso, line 438: We could `<<parameters.host>>` in the name as well

docker-compose.yml
```diff
@@ -186,6 +187,7 @@ services:
     command: /bin/bash
     depends_on:
       - ddagent
+      - testagent
```
GustavoCaso, line 190: Any reason why not using the test agent in all the tracer versions?

.circleci/config.yml
```diff
@@ -176,6 +190,21 @@ step_run_all_tests: &step_run_all_tests
...
+      RESPONSE_CODE=$(echo "$RESPONSE" | awk 'END {print $NF}')
+      if [[ $RESPONSE_CODE -eq 200 ]]; then
+        echo "All Check Traces returned successful! (HTTP 200)"
+      else
+        echo "Check Traces failed with response code: $RESPONSE_CODE"
+        echo "Failures:"
+        cat response.txt
+        exit 1
+      fi
+    when: always
```
GustavoCaso, line 207: What are the cases when the trace agent considers a trace to have failed? 

Should it make more sense to move this check into the spec suite in ruby?

wconti27: > > The Test Agent only has a few checks currently (mainly header checks like trace count and content length), but we'll be expanding it to have checks for Service Naming and Unified Naming for consistent tagging across tracers.
> 
> What kind of signal do we get when one of these "checks" fail?

Hi, please see [here](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/10224/workflows/60c4cfaa-c497-4ab5-afb8-81e82cebfdcc/jobs/379472#:~:text=Get%20APM%20Test-,Agent,-Trace%20Check%20Results) for the signal, where we query the Test-Agent to see if there were any failing traces submitted. Please keep in mind, there are so many failures since I increased the `X-Datadog-Trace-Count` header by 1 in order to purpos. efully fail the Test-Agent checks, so every received trace is failing.

wconti27:
> > The Test Agent only has a few checks currently (mainly header checks like trace count and content length), but we'll be expanding it to have checks for Service Naming and Unified Naming for consistent tagging across tracers.
> 
> What kind of signal do we get when one of these "checks" fail?

Hi, please see [here](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/10224/workflows/60c4cfaa-c497-4ab5-afb8-81e82cebfdcc/jobs/379472#:~:text=Get%20APM%20Test-,Agent,-Trace%20Check%20Results) for the signal, where we query the Test-Agent to see if there were any failing traces submitted. Please keep in mind, there are so many failures since I increased the `X-Datadog-Trace-Count` header by 1 in order to purpos. efully fail the Test-Agent checks, so every received trace is failing.

.circleci/config.yml
```diff
@@ -90,12 +92,24 @@ test_containers:
...
     image: datadog/agent
+    name: ddagent
     environment:
       - DD_APM_ENABLED=true
       - DD_BIND_HOST=0.0.0.0
       - DD_API_KEY=00000000000000000000000000000000
       - DD_HOSTNAME=dd-trace-rb-ci
   - &agent_port 8126
+  - &container_testagent
+    image: williamconti549/dd-apm-test-agent:pool-failures
```
GustavoCaso, line 103: Will the image be replaced with the official one once we are happy with the implementation? https://github.com/orgs/DataDog/packages?repo_name=dd-apm-test-agent

wconti27: Yes it will, just waiting on a PR to be approved in the Test-Agent repository (which this image was produced from). I will switch the images once that PR is merged and released in the Test-Agent

wconti27:
> @wconti27 if you rebase master on your branch, the system tests jobs from CI should be green.
> 
> A part from that do you need any help with this PR?

@GustavoCaso Thank you for the help so far! There is one more piece of functionality I need to add, which is to get the Datadog specific environment at the time of a trace's creation, then to include that environment in the trace headers sent to the Test-Agent. An example would be [this](https://github.com/DataDog/dd-trace-rb/blob/c38d420b0187da4962a5e70386724f936f32388a/spec/datadog/tracing/contrib/service_name_settings_examples.rb#L11) test, where the `DD_TRACE_SPAN_ATTRIBUTE_SCHEMA` env variable is set. Then, within the trace headers: `X-Datadog-Trace-Env-Variables: "DD_TRACE_SPAN_ATTRIBUTE_SCHEMA=v0;DD_OTHER_VAR=VALUE"`

.circleci/config.yml
wconti27: Yup, added this

spec/support/faux_writer.rb
```diff
@@ -20,6 +35,7 @@ def write(trace)
     @mutex.synchronize do
       super(trace) if @options[:call_original]
       @traces << trace
+      @options[:transport].send_traces(trace)
```
GustavoCaso, line 38: @wconti27 I wonder if we need this line here. 

The call to `super(trace`) will call the transport to send the traces. It depends on which transport we are using. Since it could be the case that `call_original` is set to true, there might be cases where we are sending the traces twice.

.circleci/config.yml
```diff
@@ -271,6 +299,7 @@ orbs:
           - *container_mongo
           - *container_memcached
           - *container_agent
```
GustavoCaso, line 301: Do we still need to use the `container_agent`?

spec/datadog/tracing/contrib/support/tracer_helpers.rb
```diff
@@ -92,6 +92,11 @@ def clear_traces!
       # with mock assertions.
       config.around do |example|
         example.run.tap do
+          traces = fetch_traces(tracer)
+          traces.each do |trace|
+            # write traces after the test to the agent in order to not mess up assertions
+            tracer.writer.write(trace)
+          end
```
GustavoCaso, line 99: I'm not sure we want to execute this step all the time, since there might be cases where the agent or the test agent is not running

GustavoCaso:
@wconti27 I don't particularly appreciate that we are hardcoding `'http://testagent:9126'` for `WebMock.

Is there a way to globally configure WebMock in the spec_helper?

GustavoCaso:
I will cc @marcotc on this PR so he can also have a look. Since he is more knowledgeable about the tracing code than me

spec/datadog/tracing/contrib/support/tracer_helpers.rb
wconti27: Added a check for this

wconti27:
> @wconti27 I don't particularly appreciate that we are hardcoding `'http://testagent:9126'` for `WebMock.
> 
> Is there a way to globally configure WebMock in the spec_helper?

I wasn't able to find a way to configure WebMock globally, but was able to remove all hardcoded references within each contrib test.

wconti27:
> I will cc @marcotc on this PR so he can also have a look. Since he is more knowledgeable about the tracing code than me

@marcotc cc Can you please review? Thanks!

docker-compose.yml
```diff
@@ -370,9 +385,22 @@ services:
       - "8126"
     ports:
       - "127.0.0.1:${DD_METRIC_AGENT_PORT}:8125/udp"
-      - "127.0.0.1:${DD_TRACE_AGENT_PORT}:8126"
+      - "127.0.0.1:8126:8126"
```
marcotc, line 388: Is this change needed? If so, why?

docker-compose.yml
```diff
@@ -370,9 +385,22 @@ services:
...
     volumes:
       - ddagent_var_run:/var/run/datadog
+
+  testagent:
+    image: ghcr.io/datadog/dd-apm-test-agent/ddapm-test-agent:latest
+    ports:
+        - "127.0.0.1:9126:9126"
```
marcotc, line 395: I suggest creating a new `DD_TEST_AGENT_PORT` variable to configure the host port.

marcotc:
```suggestion
        - "127.0.0.1:${DD_TEST_AGENT_PORT}:9126"
```

docker-compose.yml
```diff
@@ -370,9 +385,22 @@ services:
...
+    environment:
+        - LOG_LEVEL=DEBUG
+        - TRACE_LANGUAGE=ruby
+        - DD_TRACE_AGENT_URL=http://ddagent:8126
```
marcotc, line 399: Because we reference `ddagent` here, I believe we are missing a
```yaml
depends_on:
  - ddagent
```
for the `testagent` docker service.

spec/datadog/tracing/contrib/elasticsearch/transport_spec.rb
```diff
@@ -10,8 +10,8 @@
 
 RSpec.describe 'Elasticsearch::Transport::Client tracing' do
   before do
-    WebMock.enable!
-    WebMock.disable_net_connect!
+    call_web_mock_function_with_agent_host_exclusions { |options| WebMock.enable! options }
+    call_web_mock_function_with_agent_host_exclusions { |options| WebMock.disable_net_connect! options }
```
marcotc, line 14: Is `call_web_mock_function_with_agent_host_exclusions` here because one would run this test case without having the agent running locally? If so, that's not a concern as these tests are not meant to successfully run with an agent running.

spec/support/faux_writer.rb
```diff
@@ -1,11 +1,28 @@
...
 # FauxWriter is a dummy writer that buffers spans locally.
 class FauxWriter < Datadog::Tracing::Writer
+  include NetworkHelpers
+
   def initialize(options = {})
-    options[:transport] ||= FauxTransport.new
+    options[:transport] ||= if test_agent_running?
+                              Datadog::Transport::HTTP.default(
+                                headers: {
+                                  'X-Datadog-Trace-Env-Variables' => ENV.to_h.select { |key, _| key.start_with?('DD_') }
```
marcotc, line 16: Is `X-Datadog-Trace-Env-Variables` supposed to be an official tracer feature that `dd-trace-rb` does not support yet?

marcotc:
@wconti27, it looks like the `macOS` CI failures are related to the changes in this PR: https://github.com/DataDog/dd-trace-rb/actions/runs/5181100819/jobs/9336151879?pr=2831#step:5:4046, as the request in the stack trace came from `./spec/support/network_helpers.rb:41:in 'check_availability_by_http_request'`

spec/support/faux_writer.rb
wconti27: This is not an official tracer feature. The purpose of this is to header is to include the full tracer configuration, which is later used in the Test Agent for assertions (like service-naming by checking the Schema Version used).

wconti27: @marcotc Though on second thought, to achieve this, I believe that I need to include the tracer settings used as well since they can override / be used in addition to the configurable environment variables, correct?

spec/datadog/tracing/contrib/elasticsearch/transport_spec.rb
wconti27: This is used to allow http web traffic to the test agent, which would otherwise be disabled by webmock for these tests.

docker-compose.yml
wconti27: removed this change

docker-compose.yml
wconti27: added this

spec/support/faux_writer.rb
marcotc: Hmmm, do we know what tracer configuration we are looking for?

It is correct that `ENV` won't give you the tracer configuration.

I ask because we have a lot of tracer configuration options (I'd guess 100+). We probably don't need them all.

The effective configuration used by the tracer is in the `Datadog.configuration` object (it has a `.to_h` to give you all options, for example)

wconti27: Mainly the configuration that we would like to send over is basic datadog stuff like DD_SERVICE, the span attribute schema version being used (DD_TRACE_SPAN_ATTRIBUTE_SCHEMA), and a few more service naming specific variables.

marcotc: Ok, the best way here is to collect the require data individually from `Datadog.configure` (e.g. `Datadog.configure.service`) and send each one as a header.

wconti27: great, I made this change!

marcotc (approved):

marcotc (approved):

PR #2833, opened 2023-05-04
Measured tag for client/producer spans

zarirhamza:
Adds `_dd.measured` tag for client/producer spans as part of a greater intiative for improved future stats collection across all tracers.

For Ruby, this change should not impact anything in terms of customer visibility as most client/producer spans were top level spans already. However, this addition is necessary to explicitly define measured spans as it covers the changes from previous PRs that added the new `service_name` default values which removed the top_level status from these spans.

### DIFF
--- a/lib/datadog/tracing/contrib/active_record/events/sql.rb
+++ b/lib/datadog/tracing/contrib/active_record/events/sql.rb
@@ -58,6 +58,9 @@ module Datadog
                 Contrib::Analytics.set_sample_rate(span, configuration[:analytics_sample_rate])
               end
 
+              # Measure service stats
+              Contrib::Analytics.set_measured(span)
+
               # Find out if the SQL query has been cached in this request. This meta is really
               # helpful to users because some spans may have 0ns of duration because the query
               # is simply cached from memory, so the notification is fired with start == finish.
--- a/spec/datadog/tracing/contrib/sidekiq/client_tracer_spec.rb
+++ b/spec/datadog/tracing/contrib/sidekiq/client_tracer_spec.rb
@@ -27,7 +27,7 @@ RSpec.describe 'ClientTracerTest' do
     expect(span.get_tag('sidekiq.job.queue')).to eq('default')
     expect(span.status).to eq(0)
     expect(span).to be_root_span
-    expect(span.get_metric('_dd.measured')).to be_nil
+    expect(span.get_metric('_dd.measured')).to not_be nil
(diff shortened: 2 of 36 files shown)


spec/datadog/tracing/contrib/sidekiq/client_tracer_spec.rb
```diff
@@ -27,7 +27,7 @@
     expect(span.get_tag('sidekiq.job.queue')).to eq('default')
     expect(span.status).to eq(0)
     expect(span).to be_root_span
-    expect(span.get_metric('_dd.measured')).to be_nil
+    expect(span.get_metric('_dd.measured')).to not_be nil
```
TonyCTHsu, line 30: Prefer stronger assertion `.to eq(...)`, instead of `.to not_be nil`.

zarirhamza: Fixed

lib/datadog/tracing/contrib/active_record/events/sql.rb
```diff
@@ -58,6 +58,9 @@ def process(span, event, _id, payload)
                 Contrib::Analytics.set_sample_rate(span, configuration[:analytics_sample_rate])
               end
 
+              # Measure service stats
+              Contrib::Analytics.set_measured(span)
```
marcotc, line 62: @zarirhamza, setting `measured` for spans that can be service-entry spans (can have a different service name than their parent span) causes double counting of trace metrics. Because of this, we were instructed to not set `measured` on these kind of spans.
Do you know if this has changed? If not, this proposed change will cause double counting of trace metrics for the spans in this PR.

zarirhamza: With the service naming changes, in v1, these spans would no longer be service entry spans since all spans would default to a global value. I was under the impression that setting a measured tag on a service entry span would not result in any problem, because the agent should have logic against double calculation of metrics on the same span? Do you have any reference as to why this wouldn't be the case?

marcotc:
We are waiting for our internal team to discuss if the changes in PR affect only 0, v1, or both versions.

zarirhamza:
This feature is no longer necessary as the newest agent version computes necessary stats for the targeted spans. Because the feature which removes the top level status is opt-in, there is no need to provide the measured tag. Thus, closing this PR without any merge.

Issue #2722, opened 2023-03-27
AppSec throws Datadog::AppSec::Processor::AlreadyActiveContextError with nested apps


lloeki:
**Current behaviour**

From [this discussion](https://github.com/DataDog/dd-trace-rb/pull/2617/files#r1137390655)

```
   error.type => Datadog::AppSec::Processor::AlreadyActiveContextError,
   error.message => Datadog::AppSec::Processor::AlreadyActiveContextError,
   error.stack => /usr/local/bundle/gems/ddtrace-1.10.1/lib/datadog/appsec/processor.rb:90:in `activate_context': Datadog::AppSec::Processor::AlreadyActiveContextError (Datadog::AppSec::Processor::AlreadyActiveContextError)
```

**Expected behaviour**

No exception.

**Steps to reproduce**

- Use ddtrace 1.10.1
- Mount a Sinatra app within a Rails app.
- Activate Rails and Sinatra integrations
- Hit the Sinatra endpoint

**Workarounds**

Any one of these:

- Disable Sinatra integration
- Revert to 1.10.0

**Environment**

* **ddtrace version: 1.10.1**

PR #2835, opened 2023-05-08
[APPSEC-9343] [WIP] Initial suggestion for allowing appsec monitoring nested apps

GustavoCaso:
**What does this PR do?**

This is an initial suggestion for fixing https://github.com/DataDog/dd-trace-rb/issues/2722

I tested the code against a local Rails app that mounts a Sinatra App.

The configuration on the Rails app looks like this:

```ruby
Datadog.configure do |c|
  c.tracing.instrument :rails, service_name: 'gustavo-nested-apps'
  c.appsec.enabled = true
  c.appsec.instrument :rails
  c.appsec.instrument :sinatra
end
```

```ruby
# config/routes.rb

require_relative '../app/rack_apps/my_sinatra_app'

Rails.application.routes.draw do
  root "home#index"
  mount RackApps::MySinatraApp.new => '/api'
end
```

We do not get a `Datadog::AppSec::Processor::AlreadyActiveContextError` with the suggested changes.

Because:
- We have removed that particular error from the codebase 
- We changed the API of `Processor#activate_context`. The new API, rather than checking for an active context creates a new context every time we call the method, store it on the globally accessible `Thread.current[:datadog_current_waf_context]`, and make sure once the block has executed to finalize the context and reset ``Thread.current[:datadog_current_waf_context]`

This new API ensures that each request gets its new context and does not break the ability to access the active context globally outside of a rack context. Such as for `identity.set_user` appsec event within the [Monitor's watcher](https://github.com/DataDog/dd-trace-rb/blob/013813d0c5b9ceb5809beb3b9c4225be907b03f5/lib/datadog/appsec/monitor/gateway/watcher.rb#L25) 

**Motivation**

**Additional Notes**

I added a basic integration test to validate that things work as expected. If we like the approach I would improve the test coverage for nested apps.

**How to test the change?**

### DIFF
--- a/lib/datadog/appsec/contrib/rack/request_middleware.rb
+++ b/lib/datadog/appsec/contrib/rack/request_middleware.rb
@@ -31,62 +31,61 @@ module Datadog
 
             processor = nil
             ready = false
-            context = nil
 
             Datadog::AppSec.reconfigure_lock do
               processor = Datadog::AppSec.processor
 
-              if !processor.nil? && processor.ready?
-                context = processor.activate_context
-                env['datadog.waf.context'] = context
-                ready = true
-              end
+              ready = true if !processor.nil? && processor.ready?
             end
 
             # TODO: handle exceptions, except for @app.call
 
             return @app.call(env) unless ready
 
-            gateway_request = Gateway::Request.new(env)
+            context = nil
 
-            add_appsec_tags(processor, active_trace, active_span, env)
+            processor.activate_context do |active_context|
+              context = active_context
+              env['datadog.waf.context'] = context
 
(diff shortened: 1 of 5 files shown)


lib/datadog/appsec/contrib/rack/request_middleware.rb
```diff
@@ -31,62 +31,61 @@ def call(env)
...
             end
 
             # TODO: handle exceptions, except for @app.call
 
             return @app.call(env) unless ready
 
-            gateway_request = Gateway::Request.new(env)
+            context = nil
+
+            processor.activate_context do |active_context|
```
GustavoCaso, line 47: We make sure the `active_context` is valid for the lifecycle of the request

PR #2836, opened 2023-05-10
[APPSEC-9343] Appsec support nested apps

GustavoCaso:
**What does this PR do?**

Fix https://github.com/DataDog/dd-trace-rb/issues/2722

I tested the code against a local Rails app that mounts a Sinatra App.

The configuration on the Rails app looks like this:

```ruby
Datadog.configure do |c|
  c.tracing.instrument :rails, service_name: 'gustavo-nested-apps'
  c.appsec.enabled = true
  c.appsec.instrument :rails
  c.appsec.instrument :sinatra
end
```

```ruby
# config/routes.rb

require_relative '../app/rack_apps/my_sinatra_app'

Rails.application.routes.draw do
  root "home#index"
  mount RackApps::MySinatraApp.new => '/api'
end
```

To ensure we are not re-executing the request middleware code multiple times with the context of nested apps, we need to make sure we bail out at the proper time. 

That removes the 🐛 that some customers are experiencing, which raises `Datadog::AppSec::Processor::AlreadyActiveContextError`

Also, by making sure we continue the execution of the request without adding any extra calls to the WAF, we make sure to only calls the WAF the right amount of times 😄 

**Motivation**

**Additional Notes**

I added a basic integration test to validate that things work as expected.

**How to test the change?**

### DIFF
--- a/lib/datadog/appsec/contrib/rack/request_middleware.rb
+++ b/lib/datadog/appsec/contrib/rack/request_middleware.rb
@@ -33,19 +33,9 @@ module Datadog
             ready = false
             context = nil
 
-            # If a customer has mounted a nested Rack app within their main app
-            # and instrumented boths apps:
-            #
-            # Datadog.configure do |c|
-            #   c.appsec.enabled = true
-            #   c.appsec.instrument :rails
-            #   c.appsec.instrument :sinatra
-            # end
-            #
-            # We have to avoid calling processor.activate_context twice.
-            # To achive that we check the value of env['datadog.waf.context']
-            # If the waf context is part of the env, it means that we are on the nested app
-            # context. In that case we need process the request without any new appsec related code.
+            # For a given request, keep using the first Rack stack context for
+            # nested apps. Don't set `context` local variable so that on popping
+            # out of this nested stack we don't finalize the parent's context
             return @app.call(env) if env['datadog.waf.context']
 
             Datadog::AppSec.reconfigure_lock do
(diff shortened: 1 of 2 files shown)


lib/datadog/appsec/contrib/rack/request_middleware.rb
```diff
@@ -33,6 +33,21 @@ def call(env)
             ready = false
             context = nil
 
+            # If a customer has mounted a nested Rack app within their main app
+            # and instrumented boths apps:
```
lloeki, line 37: We generally don't refer to "customers" in code:

```suggestion
            # If a user has mounted a nested Rack app within their main app
            # and instrumented boths apps:
```

In fact we can remove the reference altogether with a passive tone:

```suggestion
            # If a nested Rack app is mounted within a main app
            # and both are instrumented:
```

lib/datadog/appsec/contrib/rack/request_middleware.rb
```diff
@@ -33,6 +33,21 @@ def call(env)
...
+            # Datadog.configure do |c|
+            #   c.appsec.enabled = true
+            #   c.appsec.instrument :rails
+            #   c.appsec.instrument :sinatra
+            # end
+            #
+            # We have to avoid calling processor.activate_context twice.
+            # To achive that we check the value of env['datadog.waf.context']
+            # If the waf context is part of the env, it means that we are on the nested app
+            # context. In that case we need process the request without any new appsec related code.
```
lloeki, line 48: I feel it can be distilled into:

```suggestion
            # For a given request, keep using the first Rack stack context for
            # nested apps. Don't set `context` local variable so that on popping
            # out of this nested stack we don't finalize the parent's context
```

lloeki (approved):
LGTM, just a few notes about the comment

PR #2848, opened 2023-05-15
Tracing Dynamic Configuration

marcotc:
Adds support for the `ddtrace` to receive `APM_LIBRARY` Dynamic Configuration, through the existing Remote Configuration channel, and applies these changes to the local application.

Dynamic Configuration has precedence over locally configured values.

Dynamic Configuration can be disabled at any point in time, restore the locally configured values.

**This PR passes our new system-tests for Dynamic Configuration: https://github.com/DataDog/system-tests/pull/1172**

# Notes

All components are safely reconfigured on-the-fly; `ddtrace` doesn't need to be reconfigured (`Datadog.configure{}`) for them to apply.

In fact, invoking `Datadog.configure{}` to apply Remote Configuration has several issues:
* Dynamic Configuration today does not utilize [The Update Framework](https://theupdateframework.github.io/specification/latest/)'s feature that allow us to know which values were changed, so we are forced to apply all values all the time, this is problematic because it would mean a `Datadog.configure{}` every 5 seconds. We could track the value changes ourselves, but now we are adding an unnecessary complexity layer on a system that has already solved it for us.
* The Remote Configuration configuration work is in itself a Component, thus it gets shut down on `Datadog.configure{}`, which causes the reconfiguration process itself to be aborted (the active thread is killed) and `ddtrace` ends up in an inconsistent state.
* Moving `Datadog.configure{}` to a separate execution context (e.g. another thread) opens room for race conditions, especially when the new Remote Configuration component initializes and tries to apply Remote Configuration values very quickly. I tried adding synchronization mechanisms, buy my tests were consistently deadlocking.

Moving Remote Configuration outside of Components is likely an option, but the complexity is quite large was we'd need a separate lifecycle management mechanism for it, and it depends on configuration values (e.g. `agent.host`, `remote.enabled`), which are naturally configured alongside the Component lifecycle management. I did not attempt this, as it seemed like its own project, but I think it's worth discussing.


spec/datadog/tracing/configuration/dynamic_spec.rb
```diff
@@ -0,0 +1,81 @@
...
+                   config_key: :log_injection,
+                   value: false
+end
+
+RSpec.describe Datadog::Tracing::Configuration::Dynamic::TracingHeaderTags do
+  include_examples 'tracing dynamic simple option',
+                   name: 'tracing_header_tags',
+                   env_var: 'DD_TRACE_HEADER_TAGS',
+                   config_key: :header_tags,
+                   value: RSpec::Matchers::BuiltIn::Match.new(->(header_tags) { header_tags.to_s == 'my-header:my-tag' })
```
marcotc, line 54: I want to refactor `option :header_tags` from https://github.com/DataDog/dd-trace-rb/pull/2935/files#diff-e39a5f803ff255c0c4f1872deaae5df5cd8fb110947b92ce918a6f03b8e96e35 back into a simple value.
The idea of having a never-nil object to always perform HTTP header tag was attractive for the Rack and Sinatra integrations, but it made it super hard to handle it in configuration testing.

I don't want to do it immediately because it will take a bit of time, and it is an internal-only refactor, with no visible side-effects.

marcotc:
I'm still battling a few flaky tests, which seem unrelated to this PR 🤞. And fighting appraisal to get one test gem (`rackup`) to fit amongst all other dependency versions, but overall, the PR will not materially change.

TonyCTHsu (approved):

PR #2850, opened 2023-05-15
Add telemetry heartbeat interval configuration

marcotc:
_This PR is part of project to enable Tracing remote configuration._

This PR exposes a configuration for the interval between Telemetry heartbeat requests.

This is used for system-tests currently.

I didn't not expose it in our public documentation because there's no valid use case today for clients manually configuring this setting: there's public documentation on how to completely disable telemetry, but the heartbeat reporting frequency should be managed by Datadog, and we should the reporting frequency is always optimal for both clients and Datadog.


lib/datadog/core/configuration/settings.rb
```diff
@@ -475,6 +475,15 @@ def initialize(*_)
             o.default { env_to_bool(Core::Telemetry::Ext::ENV_ENABLED, true) }
             o.lazy
           end
+
+          # The interval in seconds when telemetry must be sent.
+          #
+          # @default `DD_TELEMETRY_HEARTBEAT_INTERVAL` environment variable, otherwise `60`.
+          # @return [Float]
+          option :heartbeat_interval do |o|
```
GustavoCaso, line 483: This comment is not a blocker but more of an open question. 

Should we try to standardize the name of our configuration options when there are similar options for other components? 

For example the remote component has an interval as well, but the configuration is `poll-interval_seconds` https://github.com/DataDog/dd-trace-rb/blob/6c3b37b5195dcfd39c17a60125230d92a72b6223/lib/datadog/core/configuration/settings.rb#L496

I think would be good to have a more consistent configurations name.

cc @lloeki

marcotc: We've been trying to keep configuration methods close to environment variable names, which in this case is `DD_TELEMETRY_HEARTBEAT_INTERVAL`.

`DD_TELEMETRY_HEARTBEAT_INTERVAL` is implemented across multiple languages already, so we can't change the env var.

Also in this case, the heartbeat is not pulling, it's pushing data to the agent.

I'm not against the suggestion, and agree we should standardize these. We've been trying to standardize them so they look like the env vars so far.

ivoanjo: +1 on including units on things that don't have more specific types (e.g. seconds if it's just a "naked" integer). 

I agree with Marco that the public naming of this is harder to go back and fix, but I would suggest that elsewhere internally in our code we could use `heartbeat_interval_seconds`.

lib/datadog/core/configuration/settings.rb
```diff
@@ -475,6 +475,15 @@ def initialize(*_)
             o.default { env_to_bool(Core::Telemetry::Ext::ENV_ENABLED, true) }
             o.lazy
           end
+
+          # The interval in seconds when telemetry must be sent.
+          #
+          # @default `DD_TELEMETRY_HEARTBEAT_INTERVAL` environment variable, otherwise `60`.
+          # @return [Float]
```
ivoanjo, line 482: Would it make sense to document here that is "For internal use only." like we do for some of the other settings?

Additionally, one pattern I've been using for profiling is to shove weird settings into an "advanced" group to make it more clear that usually you shouldn't/don't need to mess with those.

marcotc:
I added `_seconds` to the internal option name.

lib/datadog/core/telemetry/client.rb
```diff
@@ -16,12 +16,13 @@ class Client
...
         # @param enabled [Boolean] Determines whether telemetry events should be sent to the API
-        def initialize(enabled: true)
+        # @param interval [Float] How frequently heartbeats will be reported, in seconds.
+        def initialize(heartbeat_interval:, enabled: true)
           @enabled = enabled
           @emitter = Emitter.new
           @stopped = false
           @unsupported = false
-          @worker = Telemetry::Heartbeat.new(enabled: @enabled) do
+          @worker = Telemetry::Heartbeat.new(enabled: @enabled, interval: heartbeat_interval) do
```
ivoanjo, line 25: Minor: I suggest propagating the `_seconds` prefix internally as well:

```suggestion
          @worker = Telemetry::Heartbeat.new(enabled: @enabled, heartbeat_interval_seconds: heartbeat_interval_seconds) do
```

lib/datadog/core/telemetry/heartbeat.rb
```diff
@@ -10,9 +10,7 @@ module Telemetry
       class Heartbeat < Core::Worker
         include Core::Workers::Polling
 
-        DEFAULT_INTERVAL_SECONDS = 60
-
-        def initialize(enabled: true, interval: DEFAULT_INTERVAL_SECONDS, &block)
+        def initialize(interval:, enabled: true, &block)
           # Workers::Polling settings
           self.enabled = enabled
           # Workers::IntervalLoop settings
```
ivoanjo, line 16: Minor: I suggest propagating the `_seconds` prefix internally as well:

(lol github doesn't allow me to suggest down to the next line ;_;)

```suggestion
        def initialize(heartbeat_interval_seconds:, enabled: true, &block)
          # Workers::Polling settings
          self.enabled = enabled
          # Workers::IntervalLoop settings
          interval = heartbeat_interval_seconds
```

spec/datadog/core/telemetry/client_spec.rb
```diff
@@ -3,8 +3,9 @@
 require 'datadog/core/telemetry/client'
 
 RSpec.describe Datadog::Core::Telemetry::Client do
-  subject(:client) { described_class.new(enabled: enabled) }
+  subject(:client) { described_class.new(enabled: enabled, heartbeat_interval: heartbeat_interval) }
```
ivoanjo, line 6: **Note: This test was actually breaking because the let below was called `heartbeat_interval_seconds` and not `heartbeat_interval`.**

Minor: I suggest propagating the `_seconds` prefix internally as well:

```suggestion
  subject(:client) { described_class.new(enabled: enabled, heartbeat_interval_seconds: heartbeat_interval_seconds) }
```

spec/datadog/core/telemetry/client_spec.rb
```diff
@@ -21,8 +22,8 @@
       client.worker.join
     end
 
-    context 'when no params provided' do
-      subject(:client) { described_class.new }
+    context 'with default parameters' do
+      subject(:client) { described_class.new(heartbeat_interval: heartbeat_interval) }
```
ivoanjo, line 26: **Note: This test was actually breaking because the let above was called `heartbeat_interval_seconds` and not `heartbeat_interval`.**

Minor: I suggest propagating the `_seconds` prefix internally as well:

```suggestion
      subject(:client) { described_class.new(heartbeat_interval_seconds: heartbeat_interval_seconds) }
```

spec/datadog/core/telemetry/client_spec.rb
```diff
@@ -164,7 +165,8 @@
     let(:worker) { instance_double(Datadog::Core::Telemetry::Heartbeat) }
 
     before do
-      allow(Datadog::Core::Telemetry::Heartbeat).to receive(:new).and_return(worker)
+      allow(Datadog::Core::Telemetry::Heartbeat).to receive(:new)
+        .with(enabled: enabled, interval: heartbeat_interval).and_return(worker)
```
ivoanjo, line 169: **Note: This test was actually breaking because the let above was called `heartbeat_interval_seconds` and not `heartbeat_interval`.**

Minor: I suggest propagating the `_seconds` prefix internally as well:

```suggestion
      allow(Datadog::Core::Telemetry::Heartbeat).to receive(:new)
        .with(enabled: enabled, heartbeat_interval_seconds: heartbeat_interval_seconds).and_return(worker)
```

spec/datadog/core/telemetry/heartbeat_spec.rb
```diff
@@ -6,7 +6,7 @@
   subject(:heartbeat) { described_class.new(enabled: enabled, interval: interval, &block) }
```
ivoanjo, line 6: Minor: I suggest propagating the `_seconds` prefix internally as well:

```suggestion
  subject(:heartbeat) { described_class.new(enabled: enabled, heartbeat_interval_seconds: heartbeat_interval_seconds, &block) }
```

spec/datadog/core/telemetry/heartbeat_spec.rb
```diff
@@ -6,7 +6,7 @@
...
 
   let(:enabled) { true }
-  let(:interval) { 60 }
+  let(:interval) { 1.2 }
```
ivoanjo, line 9: Minor: I suggest propagating the `_seconds` prefix internally as well:

```suggestion
  let(:heartbeat_interval_seconds) { 1.2 }
```

spec/datadog/core/telemetry/heartbeat_spec.rb
```diff
@@ -16,11 +16,11 @@
 
   describe '.new' do
     context 'when using default settings' do
-      subject(:heartbeat) { described_class.new(&block) }
+      subject(:heartbeat) { described_class.new(interval: interval, &block) }
```
ivoanjo, line 19: Minor: I suggest propagating the `_seconds` prefix internally as well:

```suggestion
      subject(:heartbeat) { described_class.new(heartbeat_interval_seconds: heartbeat_interval_seconds, &block) }
```

GustavoCaso (approved):
@marcotc LGTM. Thanks for adding `_seconds`. I agree with Ivo on propagating internally 😄

PR #2841, opened 2023-05-12
Fix slow CI and broken images 

GustavoCaso:
**What does this PR do?**

- Fix the docker ruby images for 2.3 and 2.4 for the integrations apps
- Fix AppSec slow tests https://github.com/DataDog/dd-trace-rb/pull/2843

**Motivation**

**Additional Notes**

**How to test the change?**

PR #2853, opened 2023-05-16
[APPSEC-9989] Allow setting remote configuration service name

lloeki:
**What does this PR do?**

Allow setting remote configuration service name

**Motivation**

Service name set by `DD_SERVICE` or `Datadog.configure { |c| c.service = ... }` may be absent or differ from integration service name configuration in advanced sub-service naming configuration cases.

**Additional Notes**

Having it default to Rack's service name could be a thing, but was found to be a bit unreliable in such advanced configurations. It needs a bit more testing so we'll land this as a stopgap.

This does not address the case of `request_queuing`.

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/core/configuration/settings.rb
+++ b/lib/datadog/core/configuration/settings.rb
@@ -549,6 +549,14 @@ module Datadog
             o.default { env_to_float(Core::Remote::Ext::ENV_POLL_INTERVAL_SECONDS, 5.0) }
             o.lazy
           end
+
+          # Declare service name to bind to remote configuration. Use when
+          # DD_SERVICE does not match the correct integration for which remote
+          # configuration applies.
+          #
+          # @default `nil`.
+          # @return [String,nil]
+          option :service
         end
 
         # TODO: Tracing should manage its own settings.
--- a/lib/datadog/core/remote/client.rb
+++ b/lib/datadog/core/remote/client.rb
@@ -136,7 +136,7 @@ module Datadog
             runtime_id: Core::Environment::Identity.id,
             language: Core::Environment::Identity.lang,
             tracer_version: tracer_version_semver2,
-            service: Datadog.configuration.service,
+            service: service_name,
             env: Datadog.configuration.env,
             tags: client_tracer_tags,
           }
@@ -165,6 +165,10 @@ module Datadog
             },
             cached_target_files: state.cached_target_files,
           }
+        end
+
+        def service_name
+          Datadog.configuration.remote.service || Datadog.configuration.service
         end
 
         def tracer_version_semver2
--- a/sig/datadog/core/configuration/settings.rbs
+++ b/sig/datadog/core/configuration/settings.rbs
@@ -12,6 +12,10 @@ module Datadog
           def poll_interval_seconds: () -> ::Float
 
           def poll_interval_seconds=: (::Float) -> void
+
+          def service: () -> ::String?
+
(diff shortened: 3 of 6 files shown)


GustavoCaso:
You should rebase master on your PR to fix some of the Circle CI jobs. 

Here are the PRs:
https://github.com/DataDog/dd-trace-rb/pull/2841
https://github.com/DataDog/dd-trace-rb/pull/2844

lib/datadog/core/configuration/settings.rb
```diff
@@ -497,6 +497,17 @@ def initialize(*_)
...
+          # Declare service name to bind to remote configuration. Use when
+          # DD_SERVICE does not match the correct integration for which remote
+          # configuration applies.
+          #
+          # @default `nil`.
+          # @return [String,nil]
+          option :service do |o|
+            o.default { nil }
+            o.lazy
+          end
```
GustavoCaso, line 510: Do we need to set `default` and `lazy`? It seems we can simply have `option :service`

ivoanjo:
Can you share a few more details about the use cases for this feature? It's kinda weird to have multiple service names in use for a single Ruby process.

lib/datadog/core/configuration/settings.rb
lloeki: Indeed, fixed

lloeki:
Rebased.

lloeki:
> Can you share a few more details about the use cases for this feature? It's kinda weird to have multiple service names in use for a single Ruby process.

- A lot of users enjoy slicing and dicing their MVC monolith with different service names per "layer", typically overriding names per integration
- Nested Rack apps are also a thing, where service names may differ
- There is the case of request queuing, which should be addressed by #2858

ivoanjo:
> > Can you share a few more details about the use cases for this feature? It's kinda weird to have multiple service names in use for a single Ruby process.
>
>- A lot of users enjoy slicing and dicing their MVC monolith with different service names per "layer", typically overriding names per integration
>- Nested Rack apps are also a thing, where service names may differ
>- There is the case of request queuing, which should be addressed by #2858 

Thanks for the extra context but... how does that apply here? E.g. I can see why customers would want to reconfigure different names for different subsets of their apps, but isn't the remote configuration service name just one for all of the app?

That's the part I don't quite understand -- in what situations would the top-level `Datadog.configure { |c| c.service = ... }` be different from the `c.remote.service`?

ivoanjo:
So in those examples, what should we use for `c.remote.service`? Would you see `foo_web` being the `c.remote.service` in examples 1 & 2? What about in example 3?

lloeki:
> Would you see `foo_web` being the `c.remote.service` in examples 1 & 2?

Yes

> What about in example 3?

`foo_rack`

This comes from the fact that backend-side service names are inferred from the service attribute on spans (service entry spans precisely) and `DD_SERVICE` is not there at all in those cases, so backend can't map to that, and it creates a discrepancy between RC services and APM/ASM services.

I tried looking at the tracing configuration to have a sane automated RC service name but it's tricky in many ways, could still fail for odd configurations, and assumes that RC maps only to tracing.

The "real" solution is far far away and involves transport, backend, library, and general concepts changes, so we're stuck with that.

GustavoCaso (approved):

ivoanjo:
(Btw me and Loic discussed my concerns directly over chat, we're aligned on the trade-off being done here)

Issue #2677, opened 2023-03-09
RSpec tracing setup broke on v1.10.0 - undefined method `appsec'


Drowze:
**Current behaviour**
On v1.10.0 I'm getting an error using the same setup to trace RSpec I've used on v1.9.0

**Expected behaviour**
Shouldn't get an error

**Steps to reproduce**
The following raises an error:
```ruby
require "datadog/ci"
Datadog.configure { |c| c.ci.enabled = true ; c.ci.instrument :rspec }
```
Error:
```
[...]

PR #2854, opened 2023-05-16
Check if `:appsec` setting is enabled before accessing it in remote component

GustavoCaso:
**What does this PR do?**

Fixes a regression previously fixed https://github.com/DataDog/dd-trace-rb/issues/2677.

Since the introduction of remote components, the CI configuration for some projects has been broken. 

The issue comes from the remote component checking `appsec` settings. In the context of CI, most users do not load appsec, which is causing issues. This PR fixes it 

**Motivation**

**Additional Notes**

**How to test the change?**

### DIFF
--- a/lib/datadog/core/remote/client/capabilities.rb
+++ b/lib/datadog/core/remote/client/capabilities.rb
@@ -23,7 +23,7 @@ module Datadog
           private
 
           def register(settings)
-            if settings.appsec.enabled
+            if settings.respond_to?(:appsec) && settings.appsec.enabled
               register_capabilities(Datadog::AppSec::Remote.capabilities)
               register_products(Datadog::AppSec::Remote.products)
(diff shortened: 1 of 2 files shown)


marcotc (approved):

lib/datadog/core/remote/client/capabilities.rb
```diff
@@ -23,7 +23,7 @@ def initialize(settings)
           private
 
           def register(settings)
-            if settings.appsec.enabled
+            if settings.respond_to?(:appsec) && settings.appsec.enabled
```
ivoanjo, line 26: To be honest, this is why I'm not a big fan of patching behavior into core classes, especially conditionally; it makes is a bit hard to reason about what's there or what's not and only really gives the illusion of things being separate.

GustavoCaso: Agree 👍

PR #2855, opened 2023-05-16
Telemetry:Do not send Dependency `hash` when `version` is present

marcotc:
`Datadog::Core::Telemetry::V1::Dependency` is currently report both gem `version` and gem `hash` values.

The spec states that `hash` should only be used when the value for `version` cannot be provided, which doesn't happen with Ruby Gems.

This PR removes `hash`, cleaning up our telemetry data and preventing issues during data processing.

### DIFF
--- a/lib/datadog/core/telemetry/v1/dependency.rb
+++ b/lib/datadog/core/telemetry/v1/dependency.rb
@@ -17,9 +17,10 @@ module Datadog
 
           # @param name [String] Module name
           # @param version [String] Version of resolved module
-          # @param hash [String] Dependency hash
+          # @param hash [String] Dependency hash, in case `version` is not available
           def initialize(name:, version: nil, hash: nil)
             raise ArgumentError, ERROR_NIL_NAME_MESSAGE if name.nil?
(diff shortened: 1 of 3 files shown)


lib/datadog/core/telemetry/v1/dependency.rb
```diff
@@ -17,9 +17,10 @@ class Dependency
 
           # @param name [String] Module name
           # @param version [String] Version of resolved module
-          # @param hash [String] Dependency hash
+          # @param hash [String] Dependency hash, in case `version` is not available
           def initialize(name:, version: nil, hash: nil)
             raise ArgumentError, ERROR_NIL_NAME_MESSAGE if name.nil?
+            raise ArgumentError, 'if both :version and :hash exist, use :version only' if version && hash
```
ivoanjo, line 23: I soft-wonder if we should just remove the support for the `hash` entirely, since it would never be applicable to Ruby, right?

marcotc: I'm not so sure it won't be.

But I think at the end of day, we'll probably move to `telemetry/v2` altogether before (which changes most models), before this could become an issue.

PR #2858, opened 2023-05-17
Tag AppSec on Rack top-level span

lloeki:
**What does this PR do?**
Tag AppSec on Rack top-level span

**Motivation**
AppSec discovers service name based on top-level span. Tagging at trace level results in tags on the trace root span, which is usually Rack's top-level span.

When request queuing is enabeld, though, a synthetic span is added in front, representing a service outside the application. This tag then gets mistakenly tagged with AppSec information, resulting in much confusion.

The situation gets even trickier with partial flushing, as partially flushed spans may receive a trace tag.

Ensuring we tag the proper span should solve both issues.

**Additional Notes**

Starting this PR as draft since I am still figuring out how to handle `Datadog::Kit` tagging in a backward compatible manner.

**How to test the change?**

Add `request_queuing: true` to Rack or Rails tracing integration, then:

```
curl -vvv -H 'User-Agent: Arachni/v2' -H "X-Request-Start: $(ruby -e 'puts "%d" % (Time.now.to_f * 1000)')" -H "X-Queue-Start: $(ruby -e 'puts "%d" % (Time.now.to_f * 1000)')" 'http://user:pass@127.0.0.1:9292/'
```

### DIFF
--- a/lib/datadog/appsec/contrib/rack/request_body_middleware.rb
+++ b/lib/datadog/appsec/contrib/rack/request_body_middleware.rb
@@ -17,7 +17,7 @@ module Datadog
           end
 
           def call(env)
-            context = env['datadog.waf.context']
+            context = env['datadog.appsec.scope']
 
             return @app.call(env) unless context
 
--- a/lib/datadog/appsec/contrib/rack/request_body_middleware.rb
+++ b/lib/datadog/appsec/contrib/rack/request_body_middleware.rb
@@ -17,7 +17,7 @@ module Datadog
           end
 
           def call(env)
-            context = env['datadog.appsec.scope']
+            context = env[Datadog::AppSec::Ext::SCOPE_KEY]
 
             return @app.call(env) unless context
 
--- a/lib/datadog/appsec/contrib/rack/request_middleware.rb
+++ b/lib/datadog/appsec/contrib/rack/request_middleware.rb
@@ -31,19 +31,19 @@ module Datadog
 
             processor = nil
             ready = false
-            context = nil
+            scope = nil
 
-            # For a given request, keep using the first Rack stack context for
+            # For a given request, keep using the first Rack stack scope for
             # nested apps. Don't set `context` local variable so that on popping
             # out of this nested stack we don't finalize the parent's context
-            return @app.call(env) if env['datadog.waf.context']
+            return @app.call(env) if active_scope(env)
 
             Datadog::AppSec.reconfigure_lock do
               processor = Datadog::AppSec.processor
 
               if !processor.nil? && processor.ready?
-                context = processor.activate_context
-                env['datadog.waf.context'] = context
+                scope = Datadog::AppSec::Scope.activate_scope(active_trace, active_span, processor)
+                env['datadog.appsec.scope'] = scope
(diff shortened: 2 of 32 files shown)


lib/datadog/kit/appsec/events.rb
```diff
@@ -14,56 +14,107 @@ module Events
...
+            trace = appsec_scope.trace
+            span = appsec_scope.span
+          end
+
+          trace ||= Datadog::Tracing.active_trace
+          span ||= trace.active_span || Datadog::Tracing.active_span
+
+          if trace.trace_id != span.trace_id
+            raise ArgumentError, "span #{span.span_id} does not belong to trace #{trace.trace_id}"
+          end
```
GustavoCaso, line 38: We should probably extra this to a separate helper method within appsec 😄

lib/datadog/kit/appsec/events.rb
```diff
@@ -14,56 +14,101 @@ module Events
...
+        def self.track_login_success(trace = nil, span = nil, user:, **others)
+          if (appsec_scope = Datadog::AppSec.active_scope)
+            trace = appsec_scope.trace
+            span = appsec_scope.service_entry_span
+          end
+
+          trace ||= Datadog::Tracing.active_trace
+          span ||= trace.active_span || Datadog::Tracing.active_span
+
+          raise ArgumentError, "span #{span.span_id} does not belong to trace #{trace.id}" if trace.id != span.trace_id
```
GustavoCaso, line 36: Should we raise or log an message letting the user that they have done something odd?

lib/datadog/appsec/contrib/rack/request_body_middleware.rb
```diff
@@ -17,7 +17,7 @@ def initialize(app, opt = {})
           end
 
           def call(env)
-            context = env['datadog.waf.context']
+            context = env['datadog.appsec.scope']
```
lloeki: Thanks @GustavoCaso !

lib/datadog/kit/appsec/events.rb
lloeki: Yeah, I went for the quick fix route here, as I find it awkward currently to hide that code behind the wrong abstraction.

lib/datadog/appsec/contrib/rack/request_middleware.rb
```diff
@@ -70,33 +70,37 @@ def call(env)
...
+              Datadog::AppSec::Scope.deactivate_scope
             end
           end
           # rubocop:enable Metrics/AbcSize,Metrics/PerceivedComplexity,Metrics/CyclomaticComplexity,Metrics/MethodLength
 
           private
 
+          def active_scope(env)
+            env['datadog.appsec.scope']
+          end
```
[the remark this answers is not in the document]
lloeki: Agreed. Let's address this later and get the release done.

lib/datadog/kit/appsec/events.rb
lloeki: As with the active scope (prev. active context) sanity checks, this should not happen, ever. If we swallow it we'll never be aware.

PR #2860, opened 2023-05-17
[PROF-7622] Add support for profiling Ruby 3.3.0-preview1

### LINKED ISSUE #2845: Profiling cannot be enabled on Ruby 3.3.0-preview1, complaining of missing JIT support
**Current behaviour**

When trying to enable profiling on Ruby 3.3.0-preview1, the following error is shown:

> WARN -- ddtrace: [ddtrace] Profiling was requested but is not supported, profiling disabled: Your ddtrace installation is missing support for the Continuous Profiler because your Ruby has been compiled without JIT support (--disable-jit-support). The profiling native extension requires a Ruby compiled with JIT support, even if the JIT is not in use by the application itself. For help solving this issue, please contact Datadog support at <https://docs.datadoghq.com/help/>. You can also check out the Continuous Profiler troubleshooting page at <https://dtdg.co/ruby-profiler-troubleshooting>.

**Expected behaviour**

Profiling should work on Ruby 3.3.0-preview1.

**Steps to reproduce**

1. Install dd-trace-rb
2. Try to enable profiling

**Environment**

* **ddtrace version:** master
* **Configuration block (`Datadog.configure ...`):** N/A
* **Ruby version:** 3.3.0-preview1
* **Operating system:** Ubuntu 22.04.2 LTS
* **Relevant library versions:** N/A

ivoanjo:
**What does this PR do?**:

Ruby 3.3.0-preview1 is the first release of the 3.3 branch. It includes a pretty big change: it drops MJIT (one of the two JIT implementations being shipped, the other being YJIT).

This affects the profiler because we were (ab)using an internal header that Ruby included only for supporting MJIT, and this header is no more.

The solution here is to use the same solution we were using for Ruby < 2.6. Quoting the `NativeExtensionDesign.md` file:

> Because these private header files are not included in regular Ruby installations, we have two different workarounds:
>
> 1. for Ruby versions >= 2.6 we make use use the Ruby private MJIT header
> 2. for Ruby versions < 2.6 (legacy Rubies) we make use of the `debase-ruby_core_source` gem

Thus, the strategy going forward will be to also use the `debase-ruby_core_source` gem to support Ruby 3.3+.

Furthermore, there were a few changes to the APIs exported by Ruby around ractors, that I also had to fix.

**Motivation**:

Make sure we keep up-to-date with latest Rubies, and that we don't block our, and our client's experiments with these versions.

**Additional Notes**:

Because we aren't yet running Ruby 3.3 on our CI, you'll have to take my word for it that this change is correct. (Or check for yourself locally).

Separately from this PR, I'll submit a PR to add the 3.3.0-preview1 headers to `debase-ruby_core_source`, as I've been testing using a local fork.

Furthermore, I plan to bootstrap Ruby 3.3 testing in our CI so that we can validate that support for 3.3 keeps working.

But I don't see any reason to delay this PR before those two others.

**How to test the change?**:

This change can only be tested locally at the moment: install Ruby 3.3.0-preview1 in your favorite way, and then point your `Gemfile` to use <https://github.com/DataDog/debase-ruby_core_source/tree/datadog/add-ruby-3.3.0-preview1-headers>, and you'll be able to run the profiling specs as usual.

Fixes #2845

### DIFF
--- a/ext/ddtrace_profiling_native_extension/extconf.rb
+++ b/ext/ddtrace_profiling_native_extension/extconf.rb
@@ -128,6 +128,12 @@ if RUBY_PLATFORM.include?('linux')
   $defs << '-DHAVE_PTHREAD_GETCPUCLOCKID'
 end
 
+# On older Rubies, we did not need to include the ractor header (this was built into the MJIT header)
+$defs << '-DNO_RACTOR_HEADER_INCLUDE' if RUBY_VERSION < '3.3'
+
+# On older Rubies, some of the Ractor internal APIs were directly accessible
+$defs << '-DUSE_RACTOR_INTERNAL_APIS_DIRECTLY' if RUBY_VERSION < '3.3'
+
 # On older Rubies, there was no struct rb_native_thread. See private_vm_api_acccess.c for details.
 $defs << '-DNO_RB_NATIVE_THREAD' if RUBY_VERSION < '3.2'
 
@@ -217,7 +223,8 @@ if Datadog::Profiling::NativeExtensionHelpers::CAN_USE_MJIT_HEADER
 
   create_makefile EXTENSION_NAME
 else
-  # On older Rubies, we use the debase-ruby_core_source gem to get access to private VM headers.
+  # The MJIT header was introduced on 2.6 and removed on 3.3; for other Rubies we rely on
+  # the debase-ruby_core_source gem to get access to private VM headers.
   # This gem ships source code copies of these VM headers for the different Ruby VM versions;
   # see https://github.com/ruby-debug/debase-ruby_core_source for details
 
@@ -228,7 +235,11 @@ else
 
   Debase::RubyCoreSource
     .create_makefile_with_core(
-      proc { have_header('vm_core.h') && have_header('iseq.h') },
+      proc do
+        have_header('vm_core.h') &&
+        have_header('iseq.h') &&
+        (RUBY_VERSION < '3.3' || have_header('ractor_core.h'))
+      end,
       EXTENSION_NAME,
     )
 end
--- a/ext/ddtrace_profiling_native_extension/native_extension_helpers.rb
+++ b/ext/ddtrace_profiling_native_extension/native_extension_helpers.rb
@@ -12,8 +12,8 @@ module Datadog
       # Can be set to force rubygems to fail gem installation when profiling extension could not be built
       ENV_FAIL_INSTALL_IF_MISSING_EXTENSION = 'DD_PROFILING_FAIL_INSTALL_IF_MISSING_EXTENSION'
 
-      # Older Rubies don't have the MJIT header, used by the JIT compiler, so we need to use a different approach
-      CAN_USE_MJIT_HEADER = RUBY_VERSION >= '2.6'
+      # The MJIT header was introduced on 2.6 and removed on 3.3; for other Rubies we rely on debase-ruby_core_source
+      CAN_USE_MJIT_HEADER = RUBY_VERSION.start_with?('2.6', '2.7', '3.0.', '3.1.', '3.2.')
 
       LIBDATADOG_VERSION = '~> 2.0.0.1.0'
 
--- a/ext/ddtrace_profiling_native_extension/private_vm_api_access.c
+++ b/ext/ddtrace_profiling_native_extension/private_vm_api_access.c
@@ -735,8 +735,7 @@ check_method_entry(VALUE obj, int can_be_svar)
     // to allow us to ensure that we're always operating on the main ractor (if Ruby has ractors)
     // Modifications:
     // * None
-    bool
-    rb_ractor_main_p_(void)
+    bool rb_ractor_main_p_(void)
     {
         VM_ASSERT(rb_multi_ractor_p());
         rb_execution_context_t *ec = GET_EC();
--- a/ext/ddtrace_profiling_native_extension/private_vm_api_access.c
+++ b/ext/ddtrace_profiling_native_extension/private_vm_api_access.c
@@ -12,7 +12,8 @@
   // Pick up internal structures from the private Ruby MJIT header file
   #include RUBY_MJIT_HEADER
 #else
-  // On older Rubies, use a copy of the VM internal headers shipped in the debase-ruby_core_source gem
+  // The MJIT header was introduced on 2.6 and removed on 3.3; for other Rubies we rely on
+  // the debase-ruby_core_source gem to get access to private VM headers.
 
   // We can't do anything about warnings in VM headers, so we just use this technique to suppress them.
   // See https://nelkinda.com/blog/suppress-warnings-in-gcc-and-clang/#d11e364 for details.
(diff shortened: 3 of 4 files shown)


ivoanjo:
Upstream PR for `debase-ruby_core_source` gem: https://github.com/ruby-debug/debase-ruby_core_source/pull/8

ext/ddtrace_profiling_native_extension/native_extension_helpers.rb
```diff
@@ -12,8 +12,8 @@ module NativeExtensionHelpers
       # Can be set to force rubygems to fail gem installation when profiling extension could not be built
       ENV_FAIL_INSTALL_IF_MISSING_EXTENSION = 'DD_PROFILING_FAIL_INSTALL_IF_MISSING_EXTENSION'
 
-      # Older Rubies don't have the MJIT header, used by the JIT compiler, so we need to use a different approach
-      CAN_USE_MJIT_HEADER = RUBY_VERSION >= '2.6'
+      # The MJIT header was introduced on 2.6 and removed on 3.3; for other Rubies we rely on debase-ruby_core_source
+      CAN_USE_MJIT_HEADER = RUBY_VERSION.start_with?('2.6', '2.7', '3.0.', '3.1.', '3.2.')
```
GustavoCaso, line 16: Why do you add a last `.` for version 3 and above? 

What do you think of `RUBY_VERSION > '2.6' && RUBY_VERSION <= '3.2'`?

ivoanjo: I'm adding a `.`because technically in around 7 years maybe Ruby `3.10` will come around and `3.10` would match `3.1` as well. But I'm not particularly consistent in doing this everywhere.

> What do you think of RUBY_VERSION > '2.6' && RUBY_VERSION <= '3.2'?

No strong opinions, if you think it's worth it, I can change it.

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
```diff
@@ -712,8 +726,29 @@ check_method_entry(VALUE obj, int can_be_svar)
...
   // This API and definition are exported as a public symbol by the VM BUT the function header is not defined in any public header, so we
   // repeat it here to be able to use in our code.
-  bool rb_ractor_main_p_(void);
-  extern struct rb_ractor_struct *ruby_single_main_ractor;
+  #ifndef USE_RACTOR_INTERNAL_APIS_DIRECTLY
+    // Disable fast path for detecting multiple Ractors. Unfortunately this symbol is no longer visible on modern Ruby
+    // versions, so we need to do a bit more work.
+    struct rb_ractor_struct *ruby_single_main_ractor = NULL;
+
+    // Taken from upstream ractor.c at commit a1b01e7701f9fc370f8dff777aad6d39a2c5a3e3 (May 2023, Ruby 3.3.0-preview1)
```
GustavoCaso, line 734: I noticed we only place the commit sha, would it be useful to have the full URL?

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
```diff
@@ -712,8 +726,29 @@ check_method_entry(VALUE obj, int can_be_svar)
...
+    // to allow us to ensure that we're always operating on the main ractor (if Ruby has ractors)
+    // Modifications:
+    // * None
+    bool
+    rb_ractor_main_p_(void)
```
GustavoCaso, line 739:
```suggestion
    bool rb_ractor_main_p_(void)
```

GustavoCaso, line 739:
```suggestion
    bool rb_ractor_main_p_(void)
```

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
ivoanjo: Hmm... I haven't considered, because usually I leave this to be able to have a reference when comparing to earlier/later changes. So usually I don't find a lot of value in looking at this commit, but looking at others that happened before or after it.

On the other hand, it wouldn't hurt, so I don't mind adding it if you find it relevant.

ext/ddtrace_profiling_native_extension/native_extension_helpers.rb
ivoanjo: Actually we do:

```[1] pry(main)> ruby_version = "3.10.0"
=> "3.10.0"
[2] pry(main)> ruby_version > '2.6' && ruby_version <= '3.2'
=> true
```

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
GustavoCaso: I am happy either way. I noticed because I have to go and look at the code for the ruby source and then search for the commit.

Of course, that was my flow. Others might have the code locally and simply checkout to that specific commit and check. 

So I leave it to you 😄

ext/ddtrace_profiling_native_extension/extconf.rb
```diff
@@ -128,6 +128,12 @@ def add_compiler_flag(flag)
   $defs << '-DHAVE_PTHREAD_GETCPUCLOCKID'
 end
 
+# On older Rubies, we did not need to include the ractor header (this was built into the MJIT header)
+$defs << '-DNO_RACTOR_HEADER_INCLUDE' if RUBY_VERSION < '3.3'
+
+# On older Rubies, some of the Ractor internal APIs were directly accessible
+$defs << '-DUSE_RACTOR_INTERNAL_APIS_DIRECTLY' if RUBY_VERSION < '3.3'
+
```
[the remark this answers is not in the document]
ivoanjo: To be honest, I still wonder a bit if this is the right approach. I've been trying to make the checks "semantic" e.g. have a define that says "this structure is missing" and then using that in the code

```c
#ifdef STRUCTURE_MISSING
... do stuff
```

rather than doing it implicitly using the Ruby version

```c
#ifdef RUBY_2_7+
... do stuff
```

But I'm still not completely convinced on the approach I'm using right now.

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
ivoanjo: Ok so for now I'll be lazy, but I promise to add it if we find a third person that looks at it and also finds it missing :)

ext/ddtrace_profiling_native_extension/native_extension_helpers.rb
GustavoCaso: As a last suggestion, which is not at all necessary, we could use `Gem::Version` instance.

```ruby
ruby_version = Gem::Version.new("3.10.0")
=> Gem::Version.new("3.10.0")
irb(main):007:0> ruby_version > '2.6' && ruby_version <= '3.2'
=> false
```

GustavoCaso (approved):
Great work 🎉

ext/ddtrace_profiling_native_extension/native_extension_helpers.rb
ivoanjo: We talked about this on slack, but let me copy this here for posteriority:

> Even though the simplified version has potential for bugs, my thinking is:
> 
> * We never got to `.10` in the past on a Ruby release, that I know of (at least minor version -- patch version yes, like 2.4.10)
> * We set the maximum version that can be used in the gem, so it's not like dd-trace-rb 1.11 is ever going to be used with Ruby 3.10
> * If ever Ruby 3.8/3.9 approaches, we can do a search in the entire codebase by `RUBY_VERSION` and fix this anyway, it's not like it's _that_ hard to find (and would be caught by tests in a lot of cases)

PR #2861, opened 2023-05-18
Fix system tests CI jobs

GustavoCaso:
**What does this PR do?**

Since the system-tests changed to run their test in the host, rather than on a Docker container, rather than on a Docker container.
https://github.com/DataDog/system-tests/pull/958

We can no longer build the runner as a Docker image and push it to the registry.

This PR removes the step that builds the runner image as a Docker image and ensures all Python dependencies are still before running the tests.

The biggest downside of not being able to have the runner image as a Docker image, is that for every test scenario we have, we need to download the python packages. Making the tests suite take longer than before 😢 

With the fix the total time for the jobs is 15min. Not bad, but I will like to try other improvements on subsequent PRs, for now I think is important to unblock CI for running system tests. 

**Motivation**

Make sure we are running system tests 😄 

**Additional Notes**

**How to test the change?**

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -17,7 +17,6 @@ jobs:
       fail-fast: false
       matrix:
         image:
-          - runner
           - agent
     runs-on: ubuntu-latest
     name: Build (${{ matrix.image }})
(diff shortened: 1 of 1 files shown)


.github/workflows/system-tests.yml
```diff
@@ -216,7 +211,7 @@ jobs:
         run: |
           docker image list
       - name: Run scenario
-        run: ./run.sh ${{ matrix.scenario }}
+        run: ./build.sh --images runner && ./run.sh ${{ matrix.scenario }}
```
ivoanjo, line 214: In a few other repos I see the build step being separate from the run step. Should we do that here? Does it enable some kind of caching?

ivoanjo (approved):
I don't know a lot about the system tests, but I agree that any progress is better than just having our CI being broken, so here's my :+1: and hopes that we can continue to iterate on this to have speedy results.

.github/workflows/system-tests.yml
GustavoCaso: We are currently using the build command to cache the `agent` and `weblog (and all the variations)` as part of the `build-harness` and `build-apps`

The `./build.sh --images runner` makes sure to install all the Python dependencies in the host machine to be able to execute `./run.sh ${{ matrix.scenario }}`

Here we would benefit from sharing the Python package dependencies between the different `./run.sh` jobs. Since each job step happens on a

lloeki: That's one of the main reasons of using Docker in CI instead of running on host.

The main reason to run on host was to facilitate development, at the expense of CI. The system tests repo's own CI uses a single job that does build once, then runs all scenarios _linearly_, missing opportunities for parallelization.

I'm trying to restore the availability of previous behaviour in https://github.com/DataDog/system-tests/pull/1109, which _also_ solves reaching out to a non-localhost Docker when running on host. See https://github.com/DataDog/system-tests/pull/1095#issuecomment-1514595189.

ivoanjo: That sounds way better! Our CI seems to be expanding and slowing down, so all work to make it faster is makes me very very happy :)

PR #2862, opened 2023-05-18
System tests jobs reduce total execution time

GustavoCaso:
**What does this PR do?**

**Motivation**

**Additional Notes**

**How to test the change?**


GustavoCaso:
Running it this way makes the whole system-test GH action take twice as much as it is today. Using the matrix to store the scenarios helps to parallelize the work, even if, on each job, we have to pull docker images and Python dependencies. 

Closing it

PR #2865, opened 2023-05-19
Add RBS checks

lloeki:
**What does this PR do?**

- Re-add CI checks for RBS files
- Add a few more, for reporting

**Motivation**

To ensure `sig` does not drift away from `lib` these checks are essential. They will also encourage us to do typing of new files.

**Additional Notes**

Follow-up to #2859 

**How to test the change?**

CI

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -10,4 +10,7 @@ jobs:
         with:
           ruby-version: '3.2'
           bundler-cache: true # runs 'bundle install' and caches installed gems automatically
-      - run: bundle exec rake typecheck
+      - name: Check for stale signature files
+        run: bundle exec rake rbs:stale
+      - name: Check types
+        run: bundle exec rake typecheck
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -15,4 +15,6 @@ jobs:
       - name: Check for missing signature files
         run: bundle exec rake rbs:missing
       - name: Check types
-        run: bundle exec rake typecheck
+        run: bundle exec rake steep:check
+      - name: Record stats
+        run: bundle exec rake steep:stats[md] >> $GITHUB_STEP_SUMMARY
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -12,5 +12,7 @@ jobs:
           bundler-cache: true # runs 'bundle install' and caches installed gems automatically
       - name: Check for stale signature files
         run: bundle exec rake rbs:stale
+      - name: Check for missing signature files
+        run: bundle exec rake rbs:missing
       - name: Check types
         run: bundle exec rake typecheck
--- a/tasks/rbs.rake
+++ b/tasks/rbs.rake
@@ -19,10 +19,10 @@ namespace :rbs do  # rubocop:disable Metrics/BlockLength
     if stale.any?
       warn <<-EOS
         +------------------------------------------------------------------------------+
-        |  **Hello there, fellow contributor who just triggered a signature error**    |
+        | **Hello there, fellow contributor who just triggered a type signature error**|
         |                                                                              |
-        | It looks like you removed a file from `lib` and this left a lingering        |
-        | signature file in `sig`. If this file is stale you can clean up              |
+        | It looks like you removed a file from `lib/` and this left a lingering       |
+        | signature file in `sig/`. If this file is stale you can clean up             |
         | automatically with:                                                          |
         |                                                                              |
         |   bundle exec rake rbs:clean                                                 |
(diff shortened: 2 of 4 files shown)


lloeki:
This job now outputs type check stats as well (for non-ignored files):

Screenshot 2023-05-19 at 17 16 28

https://github.com/DataDog/dd-trace-rb/actions/runs/5025668471/attempts/1#summary-13612863377

.github/workflows/check.yml
```diff
@@ -10,4 +10,11 @@ jobs:
         with:
           ruby-version: '3.2'
           bundler-cache: true # runs 'bundle install' and caches installed gems automatically
-      - run: bundle exec rake typecheck
+      - name: Check for stale signature files
+        run: bundle exec rake rbs:stale
+      - name: Check for missing signature files
+        run: bundle exec rake rbs:missing
```
lloeki, line 16: From: https://github.com/DataDog/dd-trace-rb/pull/2859#discussion_r1198833448

> We can tweak the check inside the GitHub action using ${{github.actor}} information. That way we do not bother the community with adding types. Not suggesting doing on this PR

> +1 if this can be gated only to the "usual suspects" I have no concerns at all with the approach and think it's a great idea.

I am concerned that this conditional skip would lead to an immediate failure on the merge commit of a contributor PR.

I don't think it's too much to ask for rake `rbs:stale` and `rake rbs:prototype` to be called (which is now suggested in the failure), and the user adding an `ignore` to `Steepfile` as suggested by `rake steep:check` if types are too gnarly to fix e.g prototyping fails to generate a passing skeleton.

Note that the latter case may happen already if someone changes an existing, previously type checked file and type checks suddenly don't pass.

GustavoCaso: I believe what @ivoanjo was concerned about was not with telling users to run those commands `clean`, and `prototype`  but more with the idea of forcing users, not maintainers (ourselves), to make sure to type (make sure no `untyped` is present in the files they have modified with the PR. 

The main idea comes for gradually adding types in the `sig` files rather than just having `untyped`

GustavoCaso (approved):
Amazing work @lloeki 🎉 Love the output table. Still I think having it only as part of the action, we are going to eventually forget about it. 

I'm still in favour of adding a separate task that only runs if we are the ones opening the PR to make sure we remove `untyped` methods. (Separate PR)

tasks/rbs.rake
```diff
@@ -16,7 +16,29 @@ namespace :rbs do  # rubocop:disable Metrics/BlockLength
...
     stale.each { |sig| puts sig }
 
-    exit 1 if stale.any?
+    if stale.any?
+      warn <<-EOS
+        +------------------------------------------------------------------------------+
+        |  **Hello there, fellow contributor who just triggered a signature error**    |
+        |                                                                              |
+        | It looks like you removed a file from `lib` and this left a lingering        |
+        | signature file in `sig`. If this file is stale you can clean up              |
```
ivoanjo, line 25: Minor:

```suggestion
        | It looks like you removed a file from `lib/` and this left a lingering       |
        | signature file in `sig/`. If this file is stale you can clean up             |
```

tasks/rbs.rake
```diff
@@ -34,7 +56,28 @@ namespace :rbs do  # rubocop:disable Metrics/BlockLength
...
     missing.each { |lib| puts lib }
 
-    exit 1 if missing.any?
+    if missing.any?
+      warn <<-EOS
+        +------------------------------------------------------------------------------+
+        |  **Hello there, fellow contributor who just triggered a signature error**    |
+        |                                                                              |
+        | It looks like you created a file in `lib` that has no matching signature     |
+        | file in `sig`. You can automatically generate missing signature files with:  |
```
ivoanjo, line 65: Minor:

```suggestion
        | It looks like you created a file in `lib/` that has no matching signature    |
        | file in `sig/`. You can automatically generate missing signature files with:  
```

tasks/rbs.rake
```diff
@@ -34,7 +56,28 @@ namespace :rbs do  # rubocop:disable Metrics/BlockLength
 
     missing.each { |lib| puts lib }
 
-    exit 1 if missing.any?
+    if missing.any?
+      warn <<-EOS
+        +------------------------------------------------------------------------------+
+        |  **Hello there, fellow contributor who just triggered a signature error**    |
```
ivoanjo, line 62: Minor:

```suggestion
        | **Hello there, fellow contributor who just triggered a type signature error**|
```

tasks/rbs.rake
```diff
@@ -16,7 +16,29 @@ namespace :rbs do  # rubocop:disable Metrics/BlockLength
 
     stale.each { |sig| puts sig }
 
-    exit 1 if stale.any?
+    if stale.any?
+      warn <<-EOS
+        +------------------------------------------------------------------------------+
+        |  **Hello there, fellow contributor who just triggered a signature error**    |
```
ivoanjo, line 22: Minor:

```suggestion
        | **Hello there, fellow contributor who just triggered a type signature error**|
```

ivoanjo (approved):
:+1: Thanks a lot, this is even better than what I had in mind :)

lloeki:
Thanks!

> Still I think having it only as part of the action, we are going to eventually forget about it.

Yeah, that's a first step. I have other ideas such as computing coverage % diff, a rake task that processes the stats CSV output and lists files in violation of specific criteria (e.g untyped > 0, coverage < 100%...), and have that posted as a comment on the discussion like Codecov does.

PR #2866, opened 2023-05-19
Fix sql comment propagation `full` mode when tracing is disabled

TonyCTHsu:
**What does this PR do?**

When sql comment propagation with `full` mode, but tracing is disabled, ends up with `TypeError` during build `traceparent`. 

 When tracing is disabled, there are not sufficient data to construct `traceparent`.


lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -28,8 +28,16 @@ def self.prepend_comment(sql, span_op, trace_op, mode)
             }
 
             if mode.full?
-              tags[Ext::KEY_TRACEPARENT] =
-                Tracing::Distributed::TraceContext.new(fetcher: nil).send(:build_traceparent, trace_op.to_digest)
+              # When tracing is disabled, trace_operation is a dummy object that does not contain data to build traceparent
+              if datadog_configuration.tracing.enabled
```
GustavoCaso, line 32: Propagation could happen even if tracing is disabled?

TonyCTHsu: Yes, when tracing is disabled. Our API returns dummy objects, see
https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/tracer.rb#L136
https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/tracer.rb#L518-L527

GustavoCaso: Interesting. 

But if tracing is disabled, do we call the tracing library for instrumenting MySQL and PG? 

It feels weird to see the places `prepend_comment` is called and have tracing disabled.
https://github.com/DataDog/dd-trace-rb/blob/99997b77c69d22787bdae96402a43f126bcca935/lib/datadog/tracing/contrib/pg/instrumentation.rb#L99

https://github.com/DataDog/dd-trace-rb/blob/99997b77c69d22787bdae96402a43f126bcca935/lib/datadog/tracing/contrib/mysql2/instrumentation.rb#L52

I would assume that if tracing is disabled, we wouldn't even instrument MySQL and PG. Is that not correct?

GustavoCaso: If we return a dummy trace and span operation when tracing is disabled, we should make them behave like a dummy object that responds to the same methods as a regular trace or span operation. Those methods must probably be no-op. Basically, create an object [duck type](https://chetan.medium.com/duck-typing-in-ruby-212a25b90cec) like a span or trace, rather than check if tracing is enabled through the codebase.

That change would be a more involved one, that should probably be tackled on a separate PR.

I will approve the PR, but I want to make sure that we try to tackle this in the future in a more sustainable way.

GustavoCaso (approved):

PR #2867, opened 2023-05-19
Remove dynamic input used in regular expression

marcotc:
This PR removes the usage of `base_url`, which is retrieve from `Rack::Request` object, to compose the dynamic regular expression `/^#{base_url}/`.

Dynamically created regular expression from untrusted sources are an attack vector.

This PR replaces this regular expression with a plain string search for the `base_url` prefix.


lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -304,7 +304,15 @@ def parse_url(env, original_env)
                        else
                          # normally REQUEST_URI starts at the path, but it
                          # might contain the full URL in some cases (e.g WEBrick)
-                         request_uri.sub(/^#{base_url}/, '')
+                         #
+                         # DEV: Use `request_uri.delete_prefix(base_url)`, supported in Ruby 2.5+
```
marcotc, line 308: I've received push back in the past with moving some refinements to a shared location.
A `String#delete_prefix` refinement already exists in `ddtrace`, but was kept close to its usage at the time: https://github.com/DataDog/dd-trace-rb/blob/53a90d91b2a531e282cb2be806c34c657d74d638/lib/datadog/tracing/distributed/trace_context.rb#L76-L87

This PR could be simplified to use the refinement, but that refinement declaration would have to be moved to a more shareable location than it is today.

This PR becomes
```diff
- request_uri.sub(/^#{base_url}/, '')
+ request_uri.delete_prefix(base_url)
```
with the refinement.

lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -304,7 +304,15 @@ def parse_url(env, original_env)
...
+                         # rubocop:disable Style/IfInsideElse
+                         if request_uri.rindex(base_url, 0)
```
ivoanjo, line 310: Minor: I'm curious, is `.rindex(base_url, 0)` an optimization vs `start_with?`?

ivoanjo (approved):
:+1: LGTM

TonyCTHsu (approved):

lib/datadog/tracing/contrib/rack/middlewares.rb
marcotc: I extracted it to its own method.

lib/datadog/tracing/contrib/rack/middlewares.rb
marcotc: I think I got this from an old Ruby pattern, but I just tested and `start_with?` is faster (and more legible), so I changed it.
Thank you!

lib/datadog/core/utils/backport.rb
```diff
@@ -0,0 +1,33 @@
...
+module Datadog
+  module Core
+    module Utils
+      # Methods from future versions of Ruby implemented in for older rubies.
+      #
+      # This helps keep the project using newer APIs for never rubies and
+      # facilitates cleaning up when support for an older versions of Ruby is removed.
+      module Backport
+        # `String` class backports.
+        module String
```
ivoanjo, line 12: Minor: As more of a "futurology" comment (e.g. thinking of how this may perhaps evolve in the future), maybe it would make sense to have a more top-level module to contain a bag of backports (e.g. `Datadog::Core::Backport`) or even per-minimum-Ruby-version-to-avoid-the-backport (`Datadog::Core::BackportFrom24`) versus having it scoped into `Utils` > `String` per-class.

ivoanjo (approved):
LGTM :+1:

lib/datadog/core/utils/backport.rb
marcotc: I'm not sure I understand the suggestion, could you expand the examples a bit more?

PR #2206, opened 2022-08-04
Fix CI with rubocop verison

TonyCTHsu:
**What does this PR do?**

Set Rubocop version `1.32.0` to fix CI

**Motivation**

Rubocop version `1.33.0` breaks CI at this moment

PR #2870, opened 2023-05-22
Update robocop version to work with ruby 3.2

GustavoCaso:
**What does this PR do?**

I Updated the rubocop version for the newer version of Ruby. 

I primarily work on the latest version of ruby, and my IDE was complaining the rubocop version installed did not recognise ruby `3.2`

I added a condition on our `Gemfile` to fix that. 

Here is the [rubocop file](https://github.com/rubocop/rubocop/blob/v1.33.0/lib/rubocop/target_ruby.rb#L7) that specifies which ruby version is supported. The update I did was a bit on the conservative side. There is a newer version of Rubocop [1.51.0](https://github.com/rubocop/rubocop/releases/tag/v1.51.0). 

If we feel courageous, we can try the latest version 😄 

**Motivation**

**Additional Notes**

**How to test the change?**

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -82,14 +82,9 @@ else
   gem 'yard', ['~> 0.9', '< 0.9.27'] # yard 0.9.27 starts pulling webrick as a gem dependency
 end
 
-if RUBY_VERSION >= '2.4.0'
-  if RUBY_VERSION >= '2.6.0'
-    gem 'rubocop', '~> 1.34.0', require: false
-    gem 'rubocop-packaging', '~> 0.5.2', require: false
-  else
-    gem 'rubocop', ['~> 1.10', '< 1.33.0'], require: false
-    gem 'rubocop-packaging', '~> 0.5', require: false
-  end
+if RUBY_VERSION >= '2.6.0'
+  gem 'rubocop', '~> 1.34.0', require: false
+  gem 'rubocop-packaging', '~> 0.5.2', require: false
   gem 'rubocop-performance', '~> 1.9', require: false
   gem 'rubocop-rspec', '~> 2.2', require: false
 end
--- a/Gemfile
+++ b/Gemfile
@@ -83,8 +83,13 @@ else
 end
 
 if RUBY_VERSION >= '2.4.0'
-  gem 'rubocop', ['~> 1.10', '< 1.33.0'], require: false
-  gem 'rubocop-packaging', '~> 0.5', require: false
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -83,8 +83,13 @@ else
 end
 
 if RUBY_VERSION >= '2.4.0'
-  gem 'rubocop', ['~> 1.10', '< 1.33.0'], require: false
-  gem 'rubocop-packaging', '~> 0.5', require: false
+  if RUBY_VERSION >= '2.6.0'
+    gem 'rubocop', '~> 1.34.0', require: false
+    gem 'rubocop-packaging', '~> 0.5.2', require: false
```
GustavoCaso, line 88: We need a higher version of `rubocop-packaing` because the library uses a deprecated method, so it is annoying to see those messages on your terminal. 

Looks like @marcotc fixed upstream https://github.com/utkarsh2102/rubocop-packaging/pull/43 😉

spec/datadog/ci/ext/environment_spec.rb
```diff
@@ -30,7 +30,7 @@
         context "for fixture #{File.basename(filename)}" do
           # Create a context for each example inside the JSON fixture file
           JSON.parse(f.read).each_with_index do |(env, tags), i|
-            context "##{i}" do
+            context i.to_s do
```
GustavoCaso, line 33: Rubocop was complaining of [RSpec/ContextMethod](https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/ContextMethod) cop.

Since `i` doe snot represent a method name, I decided to drop the `#` sign

Gemfile
ivoanjo: To be honest, I'm not sure it's worth keeping the alternative version for 2.4/2.5, given that anyway it'll probably not work as we may start referencing new cops and whatnot in the rubocop configuration.

Loic actually suggested something like this in #2206, which was the last time someone poked this.

Thus, consider making the `if RUBY_VERSION >= 2.6.0` the top-level check and drop the part for 2.4/2.5 :)

ivoanjo (approved):
:+1: LGTM

Gemfile
GustavoCaso: Changed https://github.com/DataDog/dd-trace-rb/pull/2870/commits/8ed2408c5d1ed864a4a270030d720167b4d9af5c

PR #2873, opened 2023-05-24
[PROF-7361] Implement "no signals" workaround for running CPU Profiling 2.0 without signals

ivoanjo:
**What does this PR do?**:

This PR adds a new mode to the profiler, which we're calling the "no signals" workaround.

It also makes the new CPU Profiling 2.0 the default for Ruby 2.3, 2.4 and 2.5, and automatically enables the "no signals" workaround on these Ruby versions.

**Motivation**:

This mode is a workaround for a few incompatibility issues with the new CPU Profiling 2.0 profiler. In #2702, when we made CPU Profiling 2.0 the default mode, we listed two cases where we were falling back to the legacy profiler:

* [Some gems use native libraries that don't handle the `EINTR` error code that can be caused by a signal handler interrupting a system call](https://docs.datadoghq.com/profiler/profiler_troubleshooting/ruby/#unexpected-run-time-failures-and-errors-from-ruby-gems-that-use-native-extensions-in-dd-trace-rb-1110)
* Ruby 2.5 or below are missing an API that allows the profiler to detect if the current thread is holding on to the Global VM Lock in the signal handler, which we suspect may lead to very rare crashes

The objective of the "no signals workaround" is to allow the new CPU Profiling 2.0 to be used by customers in the above situations, both of which are related to signal handling.

Thus, we're able to migrate every customer to the new codebase, allowing them to access the new features provided by the profiler, and allowing us to completely retire the legacy profiler.

(I'll still keep the legacy profiler alive for a couple more releases, but the plan is to delete it soon).

**Additional Notes**:

The "no signals workaround" mode works by making the CPU Profiling 2.0 profiler behave in a similar way to the legacy profiler when deciding to take samples.

That is, the CPU Profiling 2.0 profiler by default runs in a background thread that **does not hold the global VM lock** and thus is running in parallel with the Ruby application.

When it decides it wants to take a sample, it sends a unix SIGPROF signal to the currently-active Ruby thread, temporarily interrupting what that thread is doing to ask the VM to take a sample.

This allows us to get higher quality data, because it's the profiler that decides when it wants to take a sample and interrupts the code to do so.

The "no signals workaround" instead modifies the above to have the profiler **grab the global VM lock** when it wants to take a sample, instead of interrupting the currently-active Ruby thread.

This means that instead of sampling immediately, the profiler needs to wait for a Ruby VM scheduling switch point, and for the Ruby VM scheduler to actually schedule our thread to execute.

This means that the profiler may need to wait dozens or hundreds of milliseconds to sample. But... it works!

And the above is effectively how the legacy profiler worked -- since it was written in pure Ruby code, it always needed to wait for its turn to run.

Implementing the new behavior is not a lot of work because we already had the scaffolding to do it as part of being able to sample in periods of idleness (see #2468). So most of the work on this was validating and adding the needed configuration.

**How to test the change?**:

Validate that in the situations described above where the legacy profiler would be used no longer happen -- instead, the new profiler gets used, and the "no signals workaround" gets enabled.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -482,7 +482,17 @@ static void *run_sampling_trigger_loop(void *state_ptr) {
   while (atomic_load(&state->should_run)) {
     state->stats.trigger_sample_attempts++;
 
-    if (!state->no_signals_workaround_enabled) {
+    if (state->no_signals_workaround_enabled) {
+      // In the no_signals_workaround_enabled mode, the profiler never sends SIGPROF signals.
+      //
+      // This is a fallback for a few incompatibilities and limitations -- see the code that decides when to enable
+      // `no_signals_workaround_enabled` in `Profiling::Component` for details.
+      //
+      // Thus, we instead pretty please ask Ruby to let us run. This means profiling data can be biased by when the Ruby
+      // scheduler chooses to schedule us.
+      state->stats.trigger_simulated_signal_delivery_attempts++;
+      grab_gvl_and_sample(); // Note: Can raise exceptions
+    } else {
       current_gvl_owner owner = gvl_owner();
       if (owner.valid) {
         // Note that reading the GVL owner and sending them a signal is a race -- the Ruby VM keeps on executing while
@@ -500,16 +510,6 @@ static void *run_sampling_trigger_loop(void *state_ptr) {
         state->stats.trigger_simulated_signal_delivery_attempts++;
         idle_sampling_helper_request_action(state->idle_sampling_helper_instance, grab_gvl_and_sample);
       }
-    } else {
-      // In the no_signals_workaround_enabled mode, the profiler never sends SIGPROF signals.
-      //
-      // This is a fallback for a few incompatibilities and limitations -- see the code that decides when to enable
-      // `no_signals_workaround_enabled` in `Profiling::Component` for details.
-      //
-      // Thus, we instead pretty please ask Ruby to let us run. This means profiling data can be biased by when the Ruby
-      // scheduler chooses to schedule us.
-      state->stats.trigger_simulated_signal_delivery_attempts++;
-      grab_gvl_and_sample(); // Note: Can raise exceptions
     }
 
     sleep_for(minimum_time_between_signals);
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -482,22 +482,34 @@ static void *run_sampling_trigger_loop(void *state_ptr) {
   while (atomic_load(&state->should_run)) {
     state->stats.trigger_sample_attempts++;
 
-    current_gvl_owner owner = gvl_owner();
-    if (owner.valid) {
-      // Note that reading the GVL owner and sending them a signal is a race -- the Ruby VM keeps on executing while
-      // we're doing this, so we may still not signal the correct thread from time to time, but our signal handler
-      // includes a check to see if it got called in the right thread
-      pthread_kill(owner.owner, SIGPROF);
+    if (!state->no_signals_workaround_enabled) {
+      current_gvl_owner owner = gvl_owner();
+      if (owner.valid) {
+        // Note that reading the GVL owner and sending them a signal is a race -- the Ruby VM keeps on executing while
+        // we're doing this, so we may still not signal the correct thread from time to time, but our signal handler
+        // includes a check to see if it got called in the right thread
+        pthread_kill(owner.owner, SIGPROF);
+      } else {
+        // If no thread owns the Global VM Lock, the application is probably idle at the moment. We still want to sample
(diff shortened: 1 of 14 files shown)


ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -476,22 +482,34 @@ static void *run_sampling_trigger_loop(void *state_ptr) {
   while (atomic_load(&state->should_run)) {
     state->stats.trigger_sample_attempts++;
 
-    current_gvl_owner owner = gvl_owner();
-    if (owner.valid) {
-      // Note that reading the GVL owner and sending them a signal is a race -- the Ruby VM keeps on executing while
-      // we're doing this, so we may still not signal the correct thread from time to time, but our signal handler
-      // includes a check to see if it got called in the right thread
-      pthread_kill(owner.owner, SIGPROF);
+    if (!state->no_signals_workaround_enabled) {
```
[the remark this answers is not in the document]
ivoanjo: Since having that option disabled is the "uncommon path", I decided to keep the common path at the top, but no strong feelings either way :)

lib/datadog/profiling/component.rb
```diff
@@ -162,32 +148,70 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
...
+        true
+      end
+
+      private_class_method def self.no_signals_workaround_enabled?(settings) # rubocop:disable Metrics/MethodLength
+        setting_value = settings.profiling.advanced.no_signals_workaround_enabled
+        legacy_ruby_that_should_use_workaround = RUBY_VERSION.start_with?('2.3.', '2.4.', '2.5.')
+
+        unless [true, false, :auto].include?(setting_value)
+          Datadog.logger.error(
+            "Ignoring invalid value for profiling no_signals_workaround_enabled setting: #{setting_value.inspect}"
```
GustavoCaso, line 160: Should we tell which are the valid values?

lib/datadog/profiling/collectors/cpu_and_wall_time_worker.rb
```diff
@@ -37,12 +38,17 @@ def initialize(
             )
           end
 
+          if no_signals_workaround_enabled
+            Datadog.logger.debug('Profiling no signals workaround is in use. Profiling data will have lower precision.')
+          end
```
GustavoCaso, line 43: It looks like when initializing the component `lib/datadog/profiling/component.rb` we log enough information about what would happen if we use `no_signals_workaround`

Do we still need this log line?

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -476,22 +482,34 @@ static void *run_sampling_trigger_loop(void *state_ptr) {
...
-      // for an uncontrolled amount of time. (This can still happen to the IdleSamplingHelper, but the
-      // CpuAndWallTimeWorker will still be free to interrupt the Ruby VM and keep sampling for the entire blocking period).
+      // This is a fallback for a few incompatibilities and limitations -- see the code that decides when to enable
+      // `no_signals_workaround_enabled` in `Profiling::Component` for details.
+      //
+      // Thus, we instead pretty please ask Ruby to let us run. This means profiling data can be biased by when the Ruby
+      // scheduler chooses to schedule us.
       state->stats.trigger_simulated_signal_delivery_attempts++;
-      idle_sampling_helper_request_action(state->idle_sampling_helper_instance, grab_gvl_and_sample);
+      grab_gvl_and_sample(); // Note: Can raise exceptions
```
GustavoCaso, line 512: I'll set up another quick chat to go over some of the details of what this method does. I believe I more or less understand what it is doing, but I'm happy to go more into the details 😄

GustavoCaso (approved):
Minor cosmetic comments. The overall code LGTM 😄

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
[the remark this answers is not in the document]
GustavoCaso: I always find confusing the `!` part.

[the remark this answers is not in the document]
ivoanjo: Changed in 5e4b13195995f3079b0273c688482ff913bd841d

lib/datadog/profiling/component.rb
ivoanjo: Changed in 69fce3cd39493c30545f23698ca1e42f8193bb36

lib/datadog/profiling/collectors/cpu_and_wall_time_worker.rb
ivoanjo: Good catch. I initially thought this may be useful for debugging, but yeah thinking about it again let's just remove it 09fef75491022bcee0126b84297fe5253df949a0 .

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
ivoanjo: Yup, happy to do it!

PR #2874, opened 2023-05-24
[ci-visibility] Update gitlab's pipeline URL extraction

juan-fernandez:
**What does this PR do?**

Remove string replace we had for gitlab's pipelines URLs

**Motivation**
This was something we did to support legacy URLs, but they're now removed:  https://docs.gitlab.com/ee/update/deprecations.html#legacy-urls-replaced-or-removed

We'll keep backwards compatibility in the backend, but no sense to keep sending these tags now.

**Additional Notes**

**How to test the change?**


lib/datadog/ci/ext/environment.rb
```diff
@@ -211,11 +214,23 @@ def extract_buildkite(env)
...
             }.to_json
           }
+
+          extra_tags = env
+            .select { |key, value| key.start_with?('BUILDKITE_AGENT_META_DATA_') }
+            .map { |key, value| "#{key.to_s.sub('BUILDKITE_AGENT_META_DATA_', '').downcase}:#{value}"}
+            .sort_by { |key| key.length }
+
+          if extra_tags.length > 0
+            tags[TAG_NODE_LABELS] = JSON.generate(extra_tags)
```
GustavoCaso, line 230: `JSON.generate(extra_tags)` and `extra_tags.to_json` are analogous. Since it looks like we are using `to_json` in other parts of this file, should we try being consistent and use `to_json` here as well 😄 ?

lib/datadog/ci/ext/environment.rb
```diff
@@ -308,6 +325,11 @@ def extract_jenkins(env)
             name = name.gsub("/#{normalize_ref(branch)}", '') if branch
             name = name.split('/').reject { |v| v.nil? || v.include?('=') }.join('/')
           end
+
+          if !env['NODE_LABELS'].nil?
+            node_labels = JSON.generate(env['NODE_LABELS'].split(' '))
```
GustavoCaso, line 330: Sam as https://github.com/DataDog/dd-trace-rb/pull/2874/files#r1206521728

lib/datadog/ci/ext/environment.rb
juan-fernandez: I tried that first, but it seems `.to_json` does not work for lists. `.to_s` seems to work but it adds some extra spaces I don't want: maybe there's a workaround?

GustavoCaso: Interesting

I just did a simple example in `irb`

```
require 'json'
=> true
irb(main):002:0> test = ["mytag:my-value", "myothertag:my-other-value"]
=> ["mytag:my-value", "myothertag:my-other-value"]
irb(main):003:0> test.to_json
=> "[\"mytag:my-value\",\"myothertag:my-other-value\"]"
```

That is what you want, right?

juan-fernandez: ok I messed it up somehow 😆 . I'll change it, thanks

GustavoCaso (approved):
LGTM

PR #2875, opened 2023-05-24
Bump debase-ruby_core_source dependency to 3.2.1

ivoanjo:
**What does this PR do?**:

This PR bumps our dependency on `debase-ruby_core_source` to version 3.2.1. This version is needed to support profiling Ruby 3.3.0-preview1.

Previously, we were allowing a range of versions for this dependency (>= 0.10.16, <= 3.2.0). This made sense since any of those versions worked for us, since we were only interested in using this gem to support Ruby < 2.6.

But, we're relying on this gem to support Ruby 3.3+, so to avoid confusing situations where a customer could have an older version of the `debase-ruby_core_source` on their `Gemfile.lock` and thus break the profiler, we're now requiring the latest version of the gem.

**Motivation**:

Support profiling on Ruby 3.3.0-preview1.

**Additional Notes**:

I've updated the appraisal gemfiles as well.

**How to test the change?**:

Validate that CI is green.


GustavoCaso (approved):
LGTM

PR #2876, opened 2023-05-25
Remove unused AppSec instrumentation options

GustavoCaso:
**What does this PR do?**

Non of our instruments accept any options.

We are cargo culting a pattern from tracing
that is not used in AppSec.

If we ever need to provide instrumentation optinos for AppSec, we could implement it or bring this code back again.

Also, exposing ENV variables that are not used for anything is a bad pattern :(

**Motivation**

**Additional Notes**

**How to test the change?**


lib/datadog/appsec/extensions.rb
```diff
@@ -30,9 +30,9 @@ def initialize(settings)
 
         # Writer methods
 
-        def instrument(name, options = {})
+        def instrument(name, _options = {})
```
GustavoCaso, line 33: I left the `_options` because this is a public API. I do not want to break the contract 😄

lib/datadog/appsec/configuration.rb
```diff
@@ -66,9 +66,7 @@ def obfuscator_value_regex=(value)
         end
 
         def [](key)
-          found = @instruments.find { |e| e.name == key }
-
-          found.options if found
+          @instruments.find { |e| e.name == key }
         end
```
ivoanjo, line 70: Would it make sense to remove this method? It's not like anything was using it, as otherwise, it would've broken, and if we're changing its semantics, we might as well.

lib/datadog/appsec/extensions.rb
ivoanjo: Should we perhaps just go with `_unused`?

lib/datadog/appsec/contrib/configuration/settings.rb
```diff
@@ -1,20 +0,0 @@
...
-  module AppSec
-    module Contrib
-      module Configuration
-        # Common settings for all integrations
-        # TODO: move to datadog/appsec/configuration or remove?
-        class Settings
-          include Datadog::Core::Configuration::Base
-
-          option :enabled, default: true
-          option :service_name
```
ivoanjo, line 15: Can the appsec instrumentations not be disabled/use the service name?

ivoanjo (commented):
Left a few notes, but overall this looks reasonable!

I'm very +1 +1 +1 about removing things that we're not using. As you said, we can always add it later, git will remember it if we need to reference this, and I'm a big believer in YAGNI.

lib/datadog/appsec/contrib/configuration/settings.rb
GustavoCaso: Not at the moment https://docs.datadoghq.com/security/application_security/enabling/ruby/?tab=rails#get-started

lib/datadog/appsec/contrib/sinatra/patcher.rb
```diff
@@ -106,7 +107,7 @@ def process_route(*)
                 self.response = AppSec::Response.negotiate(env).to_sinatra_response
 
                 # interrupt request and return response to dispatch! for consistency
-                throw(Ext::ROUTE_INTERRUPT, response)
+                throw(Datadog::AppSec::Contrib::Sinatra::Ext::ROUTE_INTERRUPT, response)
```
GustavoCaso, line 110: I found out about this 🐛 while working on removing all the unused code 😄 

Thank god it hasn't reached production 😌

lib/datadog/appsec/contrib/configuration/settings.rb
lloeki: Since AppSec now uses Tracing spans as transport, AppSec has no way to actually define, use, or report service names, so no for the service name.

Disabling could theoretically be useful, but would largely be equivalent to commenting out, with the difference that it is evaluated so would be runtime-checked.

The only cases that would be remotely useful would be:

- forcefully disabling Rack when it is added automatically by Rails, but that's not supported anyway (AppSec would not work, it _needs_ the Rack integration), but you get the theory.
- disabling select instrumentations when auto-instrumenting is in use (e.g disabling Rails and/or Sinatra in a Rails app with nested Sinatra ones, leaving only Rack or something)

lib/datadog/appsec/contrib/sinatra/patcher.rb
lloeki: From a cursory look I don't see the bug:

Ruby would try to resolve `Ext::ROUTE_INTERRUPT` this way:

- `Ext::ROUTE_INTERRUPT` starts with `Ext`, so try to resolve that:
  - look for `Ext` in each component of `Datadog::AppSec::Contrib::Sinatra::Patcher`, popping each until `Ext` is found
    - `Ext` not in `Datadog::AppSec::Contrib::Sinatra::Patcher` => move on
    - `Ext` in `Datadog::AppSec::Contrib::Sinatra` => use that
- `ROUTE_INTERRUPT` in `Datadog::AppSec::Contrib::Sinatra::Ext`, so use that
- constant resolved!

```
module Foo
  module Bar
    module Baz
      def self.test
        X::Y
      end
    end

    module X
      Y = 'y'
    end
  end
end

p Foo::Bar::Baz.test # => "y"
```

Not that I don't condone the change, because the moment you introduce `Ext` at the `Patcher` level for some reason it would find that `Ext` instead and blow up:

```
module Foo
  module Bar
    module Baz
      def self.test
        X::Y
      end

      module X; end
    end

    module X
      Y = 'y'
    end
  end
end

p Foo::Bar::Baz.test # => NameError
```

lib/datadog/appsec/contrib/configuration/settings.rb
GustavoCaso: Thanks for the extra context @lloeki

lib/datadog/appsec/contrib/rails/ext.rb
```diff
@@ -5,7 +5,6 @@ module Rails
         # Rack integration constants
         module Ext
           APP = 'rails'.freeze
```
ivoanjo, line 7: I know this is a pattern we've used elsewhere in ddtrace but... it's a really high overhead creating a whole new module to contain `'rails'` (or `'rack'`).

Not for this PR of course, but some food for thought if we still think these add any value.

ivoanjo (commented):
Looks very reasonable! Left one final note

lib/datadog/appsec/extensions.rb
ivoanjo: Either is OK -- to be honest my comment was more towards not having the `= {}` but I realize now that wasn't particularly clear ;)

PR #2877, opened 2023-05-25
[APPSEC-9545] Devise integration and automatic user events 

GustavoCaso:
**What does this PR do?**

The PR adds a new AppSec integration `devise`. The integration is responsible for instrumenting User events: `login`, and `singup`. 

The PR also adds a couple of things:
- A new AppSec configuration `automated_track_user_events`
- Add a AppSec user event `track_signup`

**Motivation**

**Additional Notes**

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/appsec/configuration/settings.rb
+++ b/lib/datadog/appsec/configuration/settings.rb
@@ -202,6 +202,8 @@ module Datadog
           dsl.instruments.each do |instrument|
             # TODO: error handling
             registered_integration = Datadog::AppSec::Contrib::Integration.registry[instrument.name]
+            next unless registered_integration
+
             @integrations << Integration.new(registered_integration)
 
             # TODO: move to a separate apply step
--- a/lib/datadog/appsec/configuration/settings.rb
+++ b/lib/datadog/appsec/configuration/settings.rb
@@ -103,6 +103,7 @@ module Datadog
           trace_rate_limit: 100, # traces/s
           obfuscator_key_regex: DEFAULT_OBFUSCATOR_KEY_REGEX,
           obfuscator_value_regex: DEFAULT_OBFUSCATOR_VALUE_REGEX,
+          automated_track_user_events: :safe,
         }.freeze
 
         ENVS = {
@@ -113,6 +114,7 @@ module Datadog
           'DD_APPSEC_TRACE_RATE_LIMIT' => [:trace_rate_limit, Settings.integer],
           'DD_APPSEC_OBFUSCATION_PARAMETER_KEY_REGEXP' => [:obfuscator_key_regex, Settings.string],
           'DD_APPSEC_OBFUSCATION_PARAMETER_VALUE_REGEXP' => [:obfuscator_value_regex, Settings.string],
+          'DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING' => [:automated_track_user_events, Settings.string],
         }.freeze
 
         # Struct constant whisker cast for Steep
@@ -181,6 +183,11 @@ module Datadog
           _ = @options[:obfuscator_value_regex]
         end
 
+        def automated_track_user_events
+          # Cast for Steep
+          _ = @options[:automated_track_user_events]
+        end
+
         def merge(dsl)
           dsl.options.each do |k, v|
             unless v.nil?
@@ -199,10 +206,10 @@ module Datadog
 
             # TODO: move to a separate apply step
             klass = registered_integration.klass
-            if klass.loaded? && klass.compatible?
-              instance = klass.new
-              instance.patcher.patch
-            end
+            next unless klass.loaded? && klass.compatible?
+
+            instance = klass.new
+            instance.patcher.patch unless instance.patcher.patched?
           end
 
           self
--- a/lib/datadog/appsec/contrib/devise/ext.rb
+++ b/lib/datadog/appsec/contrib/devise/ext.rb
@@ -7,8 +7,6 @@ module Datadog
         # Devise integration constants
         module Ext
           APP = 'devise'
-          DISABLED_MODE = 'disabled'
-          EXTENDED_MODE = 'extended'
         end
       end
     end
--- a/lib/datadog/appsec/contrib/devise/ext.rb
+++ b/lib/datadog/appsec/contrib/devise/ext.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+module Datadog
+  module AppSec
+    module Contrib
+      module Devise
+        # Devise integration constants
+        module Ext
+          APP = 'devise'
+          DISABLED_MODE = 'disabled'
+          EXTENDED_MODE = 'extended'
+        end
+      end
+    end
+  end
+end
--- a/lib/datadog/appsec/contrib/devise/patcher.rb
+++ b/lib/datadog/appsec/contrib/devise/patcher.rb
@@ -3,18 +3,12 @@
 require_relative '../patcher'
 require_relative 'resource'
 require_relative 'ext'
-require_relative '../../../kit/appsec/events'
+require_relative 'tracking'
 
 module Datadog
   module AppSec
     module Contrib
       module Devise
-        AUTOMATED_USER_EVENT_TAGGING_BLOCK = proc do |trace, event|
-          trace.set_tag("_dd.appsec.events.#{event}.auto.mode", AppSec.settings.automated_track_user_events)
-        end
-
-        private_constant :AUTOMATED_USER_EVENT_TAGGING_BLOCK
-
         # Hook in devise validate method
         module AuthenticablePatch
           # rubocop:disable Metrics/MethodLength, Metrics/PerceivedComplexity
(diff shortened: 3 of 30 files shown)


lib/datadog/appsec/configuration/settings.rb
```diff
@@ -113,6 +114,7 @@ def duration(base = :ns, type = :integer)
           'DD_APPSEC_TRACE_RATE_LIMIT' => [:trace_rate_limit, Settings.integer],
           'DD_APPSEC_OBFUSCATION_PARAMETER_KEY_REGEXP' => [:obfuscator_key_regex, Settings.string],
           'DD_APPSEC_OBFUSCATION_PARAMETER_VALUE_REGEXP' => [:obfuscator_value_regex, Settings.string],
+          'DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING' => [:automated_track_user_events, Settings.string],
```
lloeki, line 117: Not too fond of that naming. Is `DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING` a specified env var?

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -199,10 +206,10 @@ def merge(dsl)
...
             # TODO: move to a separate apply step
             klass = registered_integration.klass
-            if klass.loaded? && klass.compatible?
-              instance = klass.new
-              instance.patcher.patch
-            end
+            next unless klass.loaded? && klass.compatible?
+
+            instance = klass.new
+            instance.patcher.patch unless instance.patcher.patched?
```
lloeki, line 212: Curious about `unless instance.patcher.patched?`

IIRC `instance.patcher.patch` should be idempotent

lib/datadog/appsec/contrib/devise/ext.rb
```diff
@@ -0,0 +1,16 @@
...
+
+module Datadog
+  module AppSec
+    module Contrib
+      module Devise
+        # Devise integration constants
+        module Ext
+          APP = 'devise'
+          DISABLED_MODE = 'disabled'
+          EXTENDED_MODE = 'extended'
```
lloeki, line 11: `safe` does not appear here, so this doesn't feel exhaustive. It feels like it should be an exhaustively listing enum.

lib/datadog/appsec/contrib/devise/patcher.rb
```diff
@@ -0,0 +1,160 @@
...
+            active_trace = defined?(Datadog::Tracing) && Datadog::Tracing.active_trace
+
+            return result unless active_trace
+
+            devise_resource = resource ? Resource.new(resource) : nil
+
+            event_information = {}
+            user_id = nil
+
+            if AppSec.settings.automated_track_user_events == Ext::EXTENDED_MODE && devise_resource
```
lloeki, line 35: Wondering if we should have predicate methods in there someday instead of comparing values, like `AppSec.settings.automated_track_user_events.extended?`

lloeki, line 35: Wondering if we should have predicate methods in there someday instead of comparing values, like `AppSec.settings.automated_track_user_events.extended?`

lib/datadog/appsec/contrib/devise/patcher.rb
```diff
@@ -0,0 +1,160 @@
...
+
+module Datadog
+  module AppSec
+    module Contrib
+      module Devise
+        AUTOMATED_USER_EVENT_TAGGING_BLOCK = proc do |trace, event|
+          trace.set_tag("_dd.appsec.events.#{event}.auto.mode", AppSec.settings.automated_track_user_events)
+        end
+
+        private_constant :AUTOMATED_USER_EVENT_TAGGING_BLOCK
```
lloeki, line 16: To me this looks like a horrible way to define a method 🙀

lib/datadog/appsec/contrib/devise/patcher.rb
```diff
@@ -0,0 +1,160 @@
...
+            user_id = devise_resource.id if devise_resource && devise_resource.id
+
+            if result
+              if user_id
+                Datadog::Kit::AppSec::Events.track_login_success(
+                  active_trace,
+                  user: { id: user_id.to_s },
+                  **event_information,
+                  &AUTOMATED_USER_EVENT_TAGGING_BLOCK
+                )
```
lloeki, line 52: Hmm, this makes `Datadog::AppSec` depend on `Datadog::Kit`, creating a circular dependency.

The main idea behind `Datadog::Kit` is to provide users with a glue API for users to peruse. The dependency order should be:

`Datadog::Core` < `Datadog::{AppSec,Tracing,...}` < `Datadog::Kit`

IOW, components are implemented in terms of `Core`, and `Kit` is implemented atop components.

I think this is why this block argument looks so unwieldy.

lib/datadog/appsec/contrib/devise/patcher.rb
```diff
@@ -0,0 +1,160 @@
...
+    module Contrib
+      module Devise
+        AUTOMATED_USER_EVENT_TAGGING_BLOCK = proc do |trace, event|
+          trace.set_tag("_dd.appsec.events.#{event}.auto.mode", AppSec.settings.automated_track_user_events)
+        end
+
+        private_constant :AUTOMATED_USER_EVENT_TAGGING_BLOCK
+
+        # Hook in devise validate method
+        module AuthenticablePatch
```
lloeki, line 19: I think there's a typo?

```suggestion
        module AuthenticatablePatch
```

lib/datadog/appsec/extensions.rb
```diff
@@ -24,6 +24,21 @@ def appsec
       # Merges {Datadog::AppSec::Configuration::Settings} and {Datadog::AppSec::Configuration::DSL}
       # into a single read/write object.
       class AppSecAdapter
+        VALID_AUTOMATED_TRACK_USER_EVENTS_VALUES = [
+          'safe',
+          'extended',
+          'disabled'
+        ].freeze
+
+        private_constant :VALID_AUTOMATED_TRACK_USER_EVENTS_VALUES
```
lloeki, line 33: There's the enum!

lib/datadog/kit/appsec/events.rb
```diff
@@ -57,9 +77,13 @@ def self.track_login_failure(trace, user_id:, user_exists:, **others)
...
         #   be :track.
-        def self.track(event, trace, **others)
-          trace.set_tag("appsec.events.#{event}.track", 'true')
+        def track(event, trace, **others, &custom_track_tags_block)
+          custom_track_tags_block ||= proc do |trace_op, event_name|
+            trace_op.set_tag("appsec.events.#{event_name}.sdk", 'true')
+          end
 
+          trace.set_tag("appsec.events.#{event}.track", 'true')
+          custom_track_tags_block.call(trace, event)
```
lloeki, line 86: It seems to me this API is now very abstraction-leaking. It looks like we're attempting to shapeshifting `track` into something it was not intended to do.

Were it internal I would not be too worried as we could refactor it at our own discretion, but this is veeeery public API so is bound to be used and abused.

lloeki (approved):
I'm OK to have this merged as is because it works and would need costly rebasing if it drags along, but several comments need addressing in short order, especially public API things.

Also watch out for conflicts with #2858

lib/datadog/appsec/configuration/settings.rb
GustavoCaso: Yes is part of the RFC

GustavoCaso: Yes is part of the RFC

lib/datadog/appsec/contrib/devise/patcher.rb
GustavoCaso: I thought about it but didn't want to add much logic to our custom configuration. If you feel strongly about it I can add it

lib/datadog/kit/appsec/events.rb
```diff
@@ -7,8 +7,11 @@ module Kit
     module AppSec
       # Tracking events
       module Events
+        module_function
```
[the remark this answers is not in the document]
GustavoCaso: Thanks for the feedback 😄

lib/datadog/appsec/contrib/devise/patcher.rb
lloeki: Clarification:

```
def tag_automated_user_event(trace, event)
  trace.set_tag("_dd.appsec.events.#{event}.auto.mode", AppSec.settings.automated_track_user_events)
end

...

method(:tag_automated_user_event) # => proc
```

lib/datadog/appsec/configuration/settings.rb
lloeki: Makes sense

lib/datadog/appsec/contrib/devise/ext.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Datadog
+  module AppSec
+    module Contrib
+      module Devise
+        # Devise integration constants
+        module Ext
+          APP = 'devise'
```
lloeki, line 9: Is this constant used anywhere?

GustavoCaso: None of the `Ext::APP` constants are used at all. We can decide to remove them all, but since is the standard I prefer to remove it on a separate PR

lloeki: Yeah we should remove them (and any other unused constant), they're confusing as hell.

marcotc:
General question: should `devise` be also a traced gem, from your experience working with it?

GustavoCaso:
@marcotc 

> General question: should devise be also a traced gem, from your experience working with it?

Yes. I want to add the integration myself, but for the sake of reducing the amount of code on this PR I wanted to do as a separate PR

PR #2879, opened 2023-05-30
Use first valid extracted style for distributed tracing

TonyCTHsu:
**What does this PR do?**

Currently, when extracting from multiple contexts, our propagation would validate against each other to ensure the trace id and span id are identical. This approach ensure that if user includes `Datadog` proprietary propagation, those proprietary tags would be propagated and the user will remain having full functionalities that Datadog offers.

If there is a mismatched being detected, since we are not able to identify the source of truth, it would leads to a broken distributed trace. The third scenario on the right illustrated a possible scenario with a mismatched detected given 128 bits trace ID upstream.

128-bit-trace-id-conflict

This PR changes the propagation behaviour by removing the validation phase for multiple extractions. This means the first successful extraction would always wins. W3C trace context propagation is fully compatible but user prioritizing `B3` extraction would lose Datadog's proprietary tags. 

System test for trace header precedence: https://github.com/DataDog/system-tests/pull/1454

### DIFF
--- a/lib/datadog/tracing/distributed/propagation.rb
+++ b/lib/datadog/tracing/distributed/propagation.rb
@@ -72,8 +72,10 @@ module Datadog
         #
         # @param data [Hash]
         def extract(data)
-          trace_digest = nil
-          dd_trace_digest = nil
+          return unless data
+          return if data.empty?
+
+          extracted_trace_digest = nil
 
           ::Datadog.configuration.tracing.distributed_tracing.propagation_extract_style.each do |style|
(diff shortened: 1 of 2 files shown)


marcotc:
Note here: check if there are any system-tests that cover this change (or if the Ruby tracer has been skipped in any relevant system-test because this change wasn't implemented before).

marcotc (approved):
With system-tests coverage, I'm confident this will bring us closer to how Datadog advertises propagation precedence: https://github.com/DataDog/system-tests/pull/1454

PR #2882, opened 2023-05-31
Change log injection hook to after_initialize

TonyCTHsu:
**What does this PR do?**

This PR provides the changes required to configure rails log injection in `after_initialize` hook, so that it would respect the configuration from `config/initializers/datadog.rb`


marcotc (approved):

januszm:
I see that this issue was closed but even today I see `dd` tags injected to logs in development environment when using semantic_logger

```
17:06:35 web.1  | 2025-12-30 17:06:35.855824 D [30436:puma srv tp 002] {dd: {env: "", service: "myapp-core", version: "", trace_id: "0", span_id: "0"}, ddsource: ruby} (1.129ms) ActiveRecord -- Project Load -- { ... }
```

and the only way to remove these dd tags is to add

```ruby
Datadog.configuration.tracing.log_injection = false
```

on top of the config/development.rb file.

```
datadog (2.23.0)
datadog-ruby_core_source (3.4.2)
```

PR #2888, opened 2023-06-01
Log pkg-config command when building profiling native extension

ivoanjo:
**What does this PR do?**:

This PR adds a new log message to the profiling native extension build log (`mkmf.log`) containing the pkg-config command detected by Ruby.

**Motivation**:

I'm currently debugging a customer issue where Ruby does not seem to be detecting the correct path for `pkg-config`.

I'm adding this extra logging so we can figure out where Ruby thinks `pkg-config` is.

**Additional Notes**:

N/A

**How to test the change?**:

Check that new log message shows up in `mkmf.log`.

Here's an example where Ruby doesn't find `pkg-config`:

```
[ddtrace] Ruby detected the pkg-config command is nil
```

...and here's one where it does:

```
[ddtrace] Ruby detected the pkg-config command is "pkg-config"
```


lloeki (approved):
Simple enough ^^

PR #2890, opened 2023-06-01
Use relative URI for server.request.uri.raw

lloeki:
**What does this PR do?**

Reduce the raw URI to a relative URI

**Motivation**

Rules for path passlisting match against `server.request.uri.raw` but do not take into account absolute URIs (that include scheme and authority), instead assuming path only (or rather, relative URI, since strangely they could technically match against the query string, which is not part of the URI path definition). Witness the `regex:` key of such a rule:

```
 "conditions": [
    {
      "operator": "match_regex",
      "parameters": {
        "inputs": [
          {
            "address": "server.request.uri.raw"
          }
        ],
        "options": {
          "case_sensitive": false
        },
        "regex": "^/error.*"
      }
    }
  ],
```

Which should have read something like:

```
"regex": "^([a-z]+://[^/]+)?/error.*"
```

As a workaround we partially parse the URI and strip the scheme and authority components.

**Additional Notes**

None

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/appsec/contrib/rack/reactive/request.rb
+++ b/lib/datadog/appsec/contrib/rack/reactive/request.rb
@@ -28,7 +28,7 @@ module Datadog
               end
             end
 
-            def self.subscribe(op, waf_context)
+            def self.subscribe(op, waf_context) # rubocop:disable Metrics/MethodLength
               op.subscribe(*ADDRESSES) do |*values|
                 Datadog.logger.debug { "reacted to #{ADDRESSES.inspect}: #{values.inspect}" }
                 headers = values[0]
@@ -38,10 +38,15 @@ module Datadog
                 cookies = values[3]
                 client_ip = values[4]
 
+                # rules such as coming from the passlist feature assume a relative URI
+                if uri_raw && (m = %r{^(?<scheme>[a-z]+)://(?<authority>[^/]+)?(?<relative_uri>/.*)}.match(uri_raw))
+                  relative_uri = m['relative_uri']
+                end
+
                 waf_args = {
                   'server.request.cookies' => cookies,
                   'server.request.query' => query,
-                  'server.request.uri.raw' => uri_raw,
+                  'server.request.uri.raw' => relative_uri,
                   'server.request.headers' => headers,
                   'server.request.headers.no_cookies' => headers_no_cookies,
                   'http.client_ip' => client_ip,
--- a/lib/datadog/appsec/contrib/rack/reactive/request.rb
+++ b/lib/datadog/appsec/contrib/rack/reactive/request.rb
@@ -20,7 +20,7 @@ module Datadog
               catch(:block) do
                 op.publish('request.query', gateway_request.query)
                 op.publish('request.headers', gateway_request.headers)
-                op.publish('request.uri.raw', gateway_request.relative_uri)
+                op.publish('request.uri.raw', gateway_request.fullpath)
                 op.publish('request.cookies', gateway_request.cookies)
                 op.publish('request.client_ip', gateway_request.client_ip)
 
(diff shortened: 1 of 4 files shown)


lib/datadog/appsec/contrib/rack/reactive/request.rb
```diff
@@ -38,10 +38,15 @@ def self.subscribe(op, waf_context)
...
+                # rules such as coming from the passlist feature assume a relative URI
+                if uri_raw && (m = %r{^(?<scheme>[a-z]+)://(?<authority>[^/]+)?(?<relative_uri>/.*)}.match(uri_raw))
+                  relative_uri = m['relative_uri']
+                end
+
                 waf_args = {
                   'server.request.cookies' => cookies,
                   'server.request.query' => query,
-                  'server.request.uri.raw' => uri_raw,
+                  'server.request.uri.raw' => relative_uri,
```
GustavoCaso, line 49: I suggest moving this logic to `Gateway::Request` object

Rack::Request have a method call `:fullpath` that actually returns what we want:
https://github.com/rack/rack/blob/main/lib/rack/request.rb#L592

So after adding the method `fullpath` to `Gateway::Request` object, we can use it in:

```ruby
def self.publish(op, gateway_request)
  catch(:block) do
    op.publish('request.query', gateway_request.query)
    op.publish('request.headers', gateway_request.headers)
    op.publish('request.uri.raw', gateway_request.fullpath)
    op.publish('request.cookies', gateway_request.cookies)
    op.publish('request.client_ip', gateway_request.client_ip)
    nil
  end
end
```

This has the added benefit that we do not use a custom regex in the critical path, and that we use the already provided `Rack::Request` methods

GustavoCaso (changes requested):
I believe using a regex is not the best path forward to obtaining the value we want. `Rack::Request#fullpath` is the method we should use here

lib/datadog/appsec/contrib/rack/reactive/request.rb
lloeki: > I suggest moving this logic to Gateway::Request object

Now that specs are passing, I'm attempting to do just that.

Sadly `fullpath` relies on parsing the URL and even _recomposes_ it from `path` + `query_string` parsed elements, and our internal address RFC mandates that the `server.request.uri` address MUST NOT be parsed.

```
      def fullpath
        query_string.empty? ? path : "#{path}?#{query_string}"
      end
```

GustavoCaso: Previously we were using `url` methods, which also recomposes from `base_url` and `fullpath`

```ruby
def url
  base_url + fullpath
end
```

So I don't see the difference or issue to just using `fullpath`

lloeki: > We were using already parsed values before

Ah, correct, we were using `request.url`, which is actually made from `fullpath`:

```
      def url
        base_url + fullpath
      end
```

lloeki: > We were using already parsed values before

Ah, correct, we were using [`request.url`](https://github.com/rack/rack/blob/54990d39446caf046b5cbf19fd030204f850b480/lib/rack/request.rb#L584-L586), which is actually made from `fullpath`:

```
      def url
        base_url + fullpath
      end
```

lloeki: A note that we have another parser here which attempts to use `Request-URI` instead: https://github.com/DataDog/dd-trace-rb/blob/9bebd8c84f4743eab535c2ce4f06289e8899812c/lib/datadog/tracing/contrib/rack/middlewares.rb#L271-L311

PR #2895, opened 2023-06-05
Add appsec.blocked tag to span

GustavoCaso:
**What does this PR do?**

Make sure to add the `appsec.blocked` at the span when the request has been blocked.

The main change is on the different watcher files:

```
if span
  span.set_tag('appsec.blocked', 'true') if result.actions.include?('block')
  span.set_tag('appsec.event', 'true')
end
```

The line `span.set_tag('appsec.blocked', 'true') if result.actions.include?('block')` is the one adding the tag.

The rest of the changes are improving the codebase:

- Added `Event.build_service_entry_tags` helper method to reduce the size of `Event.record_via_span`
- Extract the shared examples for our integration specs to a separate file so they can be reuse by all the rack integrations specs

**Motivation**

Make sure our code is up to date with RFCs and system tests

**Additional Notes**

**How to test the change?**

CI


lloeki (approved):
LGTM, save for a bit about naming that would bring confusion to future readers.

PR #2898, opened 2023-06-07
[APPSEC-9935] AppSec, make sure to use service entry span.

GustavoCaso:
**What does this PR do?**

Use the `scope.service_entry_span` when tagging a span with appsec related tags.

**Motivation**

Make sure AppSec features work as expected, and confront to the system tests specs 

**Additional Notes**

**How to test the change?**

CI


lloeki (approved):
LGTM code-wise.

That said, I did not do it on the original PR that tagged on service entry span because I thought it was specified that these tags should be on the closest span where the event happened, not the service entry span.

Therefore if it is so, and since this appears to be borne out of system tests failing, I'm questioning whether the system tests actually respect the RFC.

PR #2900, opened 2023-06-07
[APPSEC-9936] Update AppSec response content_type resolution

GustavoCaso:
**What does this PR do?**

Based on the [internal] specs. The default content type and the response of the blocked page should be `JSON`. 

I update the code that handles that, I also added a bunch os new specs. I fixed an issue in which spaces between the `,` for separating multiple HTTP accept values.

**Motivation**

**Additional Notes**

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/appsec/response.rb
+++ b/lib/datadog/appsec/response.rb
@@ -36,13 +36,13 @@ module Datadog
           Response.new(
             status: 403,
             headers: { 'Content-Type' => content_type },
-            body: [Datadog::AppSec::Assets.blocked(format: FORMAT_MAP[content_type])]
+            body: [Datadog::AppSec::Assets.blocked(format: CONTENT_TYPE_TO_FORMAT[content_type])]
           )
         end
 
         private
 
-        FORMAT_MAP = {
+        CONTENT_TYPE_TO_FORMAT = {
           'application/json' => :json,
           'text/html' => :html,
           'text/plain' => :text,
@@ -58,10 +58,12 @@ module Datadog
           accepted = accept_types.map { |m| Utils::HTTP::MediaRange.new(m) }.sort!.reverse!
 
           accepted.each do |range|
-            match = FORMAT_MAP.keys.find { |key| range === key }
+            type_match = CONTENT_TYPE_TO_FORMAT.keys.find { |type| range === type }
 
-            return match if match
+            return type_match if type_match
           end
+
+          DEFAULT_CONTENT_TYPE
         rescue Datadog::AppSec::Utils::HTTP::MediaRange::ParseError
           DEFAULT_CONTENT_TYPE
         end
--- a/lib/datadog/appsec/response.rb
+++ b/lib/datadog/appsec/response.rb
@@ -43,20 +43,22 @@ module Datadog
         private
 
         FORMAT_MAP = {
-          'text/plain' => :text,
-          'text/html' => :html,
           'application/json' => :json,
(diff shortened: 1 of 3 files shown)


lib/datadog/appsec/response.rb
```diff
@@ -43,20 +43,22 @@ def negotiate(env)
...
           return DEFAULT_CONTENT_TYPE unless env.key?('HTTP_ACCEPT')
 
-          accepted = env['HTTP_ACCEPT'].split(',').map { |m| Utils::HTTP::MediaRange.new(m) }.sort!.reverse!
+          accept_types = env['HTTP_ACCEPT'].split(',').map(&:strip)
+
+          accepted = accept_types.map { |m| Utils::HTTP::MediaRange.new(m) }.sort!.reverse!
 
-          accepted.each_with_object(DEFAULT_CONTENT_TYPE) do |range, _default|
-            match = FORMAT_MAP.keys.find { |type| range === type }
+          accepted.each do |range|
```
lloeki, line 60: `accepted.each_with_object(DEFAULT_CONTENT_TYPE)` was making it so it returns `DEFAULT_CONTENT_TYPE` when no match has occured.

`accepted.each` now instead returns `accepted`, which is an array of `MediaRange`, so:

- the return value of `content_type` is now inconsistent: when a match is found it returns a content type but when none is found it returns an array of `MediaRange`
- when there is no match it does not return the default value anymore, which will break `FORMAT_MAP[content_type]`

lib/datadog/appsec/response.rb
```diff
@@ -43,20 +43,22 @@ def negotiate(env)
...
+            match = FORMAT_MAP.keys.find { |key| range === key }
```
lloeki, line 61: `FORMAT_MAP` keys are content types, so I'm questioning the name change here.

I would even change it that way to make it clearer:

```
type_match = FORMAT_MAP.keys.find { |type| range === type }

return type_match if type_match
```

lloeki (changes requested):
Overall OK to change the default, but I believe there's a bug introduced here.

lib/datadog/appsec/response.rb
```diff
@@ -36,30 +36,34 @@ def negotiate(env)
...
-          accepted = env['HTTP_ACCEPT'].split(',').map { |m| Utils::HTTP::MediaRange.new(m) }.sort!.reverse!
+          accept_types = env['HTTP_ACCEPT'].split(',').map(&:strip)
 
-          accepted.each_with_object(DEFAULT_CONTENT_TYPE) do |range, _default|
-            match = FORMAT_MAP.keys.find { |type| range === type }
+          accepted = accept_types.map { |m| Utils::HTTP::MediaRange.new(m) }.sort!.reverse!
 
-            return match if match
+          accepted.each do |range|
+            type_match = CONTENT_TYPE_TO_FORMAT.keys.find { |type| range === type }
```
ivoanjo, line 61: I noticed `MediaRange#===` with a `String` as input will parse that input into a `MediaRange` before doing anything else. Aka, we keep parsing the items in `CONTENT_TYPE_TO_FORMAT` all the type on this check. Maybe something to improve?

ivoanjo (approved):
LGTM :+1:

lib/datadog/appsec/response.rb
GustavoCaso: Sounds good 😄

PR #2909, opened 2023-06-15
[APPSEC-10049] Appsec handle tracing disabled and appsec events

GustavoCaso:
**What does this PR do?**

With `1.11.x`, we introduce a new way to manage the AppSec scope: https://github.com/DataDog/dd-trace-rb/pull/2858, which fundamentally changed how we tag spans and traces. Before that change, we used to tag traces rather than spans. 

When tracing is disabled, fetching the `active_trace` works, but the `active_span` is `nil`. So to keep working as before, we need to ensure that appsec works as expected when tracing is disabled, making everything a no-op as contradicting as it sounds since AppSec needs tracing as the transport layer. So in those weird cases where someone disable tracing and not appsec, things should not break. 

**Motivation**

**Additional Notes**

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/appsec/contrib/rack/request_middleware.rb
+++ b/lib/datadog/appsec/contrib/rack/request_middleware.rb
@@ -53,7 +53,7 @@ module Datadog
 
             gateway_request = Gateway::Request.new(env)
 
-            add_appsec_tags(processor, scope.trace, scope.service_entry_span, env)
+            add_appsec_tags(processor, scope, env)
 
             request_return, request_response = catch(::Datadog::AppSec::Ext::INTERRUPT) do
               Instrumentation.gateway.push('rack.request', gateway_request) do
@@ -88,7 +88,7 @@ module Datadog
             request_return
           ensure
             if scope
-              add_waf_runtime_tags(scope.service_entry_span, scope.processor_context)
+              add_waf_runtime_tags(scope)
               Datadog::AppSec::Scope.deactivate_scope
             end
           end
@@ -116,8 +116,11 @@ module Datadog
             Datadog::Tracing.active_span
           end
 
-          def add_appsec_tags(processor, trace, span, env)
-            return unless trace
+          def add_appsec_tags(processor, scope, env)
+            span = scope.service_entry_span
+            trace = scope.trace
+
+            return unless trace && span
 
             span.set_tag('_dd.appsec.enabled', 1)
             span.set_tag('_dd.runtime_family', 'ruby')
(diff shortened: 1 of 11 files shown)


lib/datadog/kit/appsec/events.rb
```diff
@@ -10,108 +10,128 @@ module Events
...
-        #   used as defaults.
-        # @param user [Hash<Symbol, String>] User information to pass to
-        #   Datadog::Kit::Identity.set_user. Must contain at least :id as key.
-        # @param others [Hash<String || Symbol, String>] Additional free-form
-        #   event information to attach to the trace.
-        def self.track_login_success(trace = nil, span = nil, user:, **others)
-          if (appsec_scope = Datadog::AppSec.active_scope)
-            trace = appsec_scope.trace
-            span = appsec_scope.service_entry_span
+        class << self
```
lloeki, line 13: Can you split the 78ca839b502d4a467c97dbf95b58f0707773a07c in two?

1. move to `class << self`
2. do the intended change

It's kind of hard to read all mixed up, even when ignoring whitespace changes

lib/datadog/appsec/event.rb
```diff
@@ -48,6 +48,8 @@ def self.record(span, *events)
       end
 
       def self.record_via_span(span, *events)
+        return unless span
```
lloeki, line 51: I'd do that at or around the call site.

I mean `record_via_span` doesn't make sense to be called without a span, which interestingly pointed me to the more pressing issue: the rate limiter will still rate limit even though it doesn't record!

lloeki: I also think we should address that using the `contrib/*/integration_spec.rb` shared examples:

I prepped some `let` here:

https://github.com/DataDog/dd-trace-rb/blob/78ca839b502d4a467c97dbf95b58f0707773a07c/spec/datadog/appsec/contrib/rack/integration_test_spec.rb#L22

... and had a TODO here:

https://github.com/DataDog/dd-trace-rb/blob/78ca839b502d4a467c97dbf95b58f0707773a07c/spec/datadog/appsec/contrib/rack/integration_test_spec.rb#L148

... for it to be implemented as an additional nested context like that one, except it's not appsec disabled but tracing:

https://github.com/DataDog/dd-trace-rb/blob/78ca839b502d4a467c97dbf95b58f0707773a07c/spec/datadog/appsec/contrib/support/integration/shared_examples.rb#L26

lib/datadog/appsec/contrib/rack/request_middleware.rb
```diff
@@ -117,7 +117,7 @@ def active_span
           end
 
           def add_appsec_tags(processor, trace, span, env)
-            return unless trace
+            return unless trace && span
```
lloeki, line 120: wondering if we should pass the scope and check for these from that instead.

same for `add_waf_runtime_tags`

lloeki (commented):
Looks logically reasonable, but left some comments as it seems we need to be more consistent with such checks in some areas.

lloeki (approved):
LGTM! Thanks for the changes :)

PR #2910, opened 2023-06-16
fix possibel leaky spesc for appsec enabled value

GustavoCaso:
**What does this PR do?**

Trying to fix those annoying leaky specs for AppSec regarding if is runing or not 😡 

**Motivation**

**Additional Notes**

**How to test the change?**


spec/datadog/tracing/integration_spec.rb
```diff
@@ -909,6 +909,11 @@ def sample!(trace)
         end
       end
 
+      after do
+        Datadog.configuration.reset!
+        Datadog.configuration.appsec.send(:reset!)
```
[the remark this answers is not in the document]
GustavoCaso: This is one of the reasons I mentioned to you that I want to write an RFC for having a core configuration. AppSec has its own way of adding extensions. 

So Ideally yes, but at the moment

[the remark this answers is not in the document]
ivoanjo: Indeed appsec configuration is... in flow I guess. Can I convince you to add `Datadog.configuration.appsec.send(:reset!)` to `Datadog::Configuration#reset!`?

While doing that doesn't solve tech debt, it seems reasonable to do until we can do a better solution.

(But, on the other hand, if I can't convince you, and just to clarify, I don't think it's a blocking issue, happy to approve either way, if we really have to )

lib/datadog/core/configuration/base.rb
```diff
@@ -91,6 +91,7 @@ def dig(*options)
 
           def reset!
             reset_options!
+            Datadog::AppSec.settings.send(:reset!)
```
ivoanjo, line 94: Thanks for doing this change. Just in case, let me ask -- are we sure we can _always_ call this method on `Datadog::Appsec`?

E.g. I seem to remember a few cases where the constants were not loaded in some cases and we need to add a few `defined?(...)` checks.

GustavoCaso: The only case I can think of is when people follow the [CI documentation ](https://docs.datadoghq.com/continuous_integration/tests/ruby/?tab=azurepipelines#instrumenting-your-tests) since they only require `datadog/ci`. 

I'm not sure how many times people would call `reset!` on those cases, but I will add the `defined?` just in case

GustavoCaso: Done [599b534](https://github.com/DataDog/dd-trace-rb/pull/2910/commits/599b5342550a7293ad8f868a82190b3780b67af2)

PR #2912, opened 2023-06-19
Add frozen_string_literal comment to Ext module files

GustavoCaso:
**What does this PR do?**

There is an effort to make all strings frozen safely, but since we don't have full control of what the customers do with those strings, we can not safely make all files use the comment `frozen_string_literal: true`

The `Ext` modules have all the strings frozen, using `.freeze`. It is easier to add the comment and remove all the calls to `.freeze`

cc @marcotc 
**Motivation**

**Additional Notes**

**How to test the change?**


marcotc (approved):
:perfect:

PR #2913, opened 2023-06-19
[PROF-7786] Include _dd.profiling.enabled tag

ivoanjo:
**What does this PR do?**:

This PR implements the Datadog-internal
"RFC: Identifying whether a span’s host has profiling enabled" similar to what
[dd-trace-java](https://github.com/DataDog/dd-trace-java/pull/5395) does by including a `_dd.profiling.enabled` tag when profiling is enabled.

This is similar to the existing `_dd.appsec.enabled` tag.

**Motivation**:

This enables a better user experience in the "Code Hotspots" tab, e.g. for users that run profiling in a subset of their fleet.

This way we can immediately tell them "hey, there's no profiling data" rather than making them wait a few minutes and then telling them that we didn't find anything.

**Additional Notes**:

N/A

**How to test the change?**:

Validate that the `_dd.profiling.enabled` tag shows up in the UX.

### DIFF
--- a/lib/datadog/tracing/trace_operation.rb
+++ b/lib/datadog/tracing/trace_operation.rb
@@ -450,7 +450,9 @@ module Datadog
           service: service,
           tags: meta,
           metrics: metrics,
-          root_span_id: !partial ? root_span && root_span.id : nil
+          root_span_id: !partial ? root_span && root_span.id : nil,
+          profiling_enabled:
+            !!(defined?(Datadog::Profiling) && Datadog::Profiling.respond_to?(:enabled?) && Datadog::Profiling.enabled?), # rubocop:disable Style/DoubleNegation
         )
       end
 
--- a/lib/datadog/tracing/trace_operation.rb
+++ b/lib/datadog/tracing/trace_operation.rb
@@ -64,6 +64,7 @@ module Datadog
         sampled: nil,
         sampling_priority: nil,
         service: nil,
+        profiling_enabled: nil,
         tags: nil,
         metrics: nil
       )
@@ -84,6 +85,7 @@ module Datadog
         @sample_rate = sample_rate
         @sampling_priority = sampling_priority
         @service = service
+        @profiling_enabled = profiling_enabled
 
         # Generic tags
         set_tags(tags) if tags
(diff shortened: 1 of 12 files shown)


lib/datadog/tracing/trace_operation.rb
```diff
@@ -450,7 +450,9 @@ def build_trace(spans, partial = false)
           service: service,
           tags: meta,
           metrics: metrics,
-          root_span_id: !partial ? root_span && root_span.id : nil
+          root_span_id: !partial ? root_span && root_span.id : nil,
+          profiling_enabled:
+            !!(defined?(Datadog::Profiling) && Datadog::Profiling.respond_to?(:enabled?) && Datadog::Profiling.enabled?), # rubocop:disable Style/DoubleNegation
```
marcotc, line 455: I think this should be checked earlier and cached carefully if possible.

In today's world, I believe we can move this into the `Tracing::Tracer` object initialization process with no loss of correctness, then pass the value here.

ivoanjo: Good point. Fixed in cbe0317323587b27a542255211f4b5636a36ff56.

I had to do it lazily instead of during initialization to avoid any potential of loops (Profiling::Component taking an optional_tracer and doing something similar in the other direction), but otherwise it was a straightforward change.

marcotc (approved):
A bit of coupling, but our products are coupled a bit at the end of day.

ivoanjo:
Thanks Marco! I hope the trade-off for improving the user experience for the "Code Hotspots" when looking at at a trace is worth it. Especially since we're going to bring a lot of really cool features to the UX next -- stay tuned :)

GustavoCaso:
@ivoanjo @marcotc Maybe profiling could take a similar approach as ASM for adding tags in a way that is not as a couple.

Could the profiler code append the tag outside the code of tracing?

ivoanjo:
> Could the profiler code append the tag outside the code of tracing?

Not right now, because the profiler never gets called by the app, as it never instruments anything.

You'll never see a profiler method sandwiched in between customer code, like you do for tracing/asm; it only sits in the background as a periodic observer.

delner:
I think we need to revisit this. I don't think this is acceptable in its current state, but I want to review.

PR #2915, opened 2023-06-19
Add support for Option precedence

marcotc:
_This is part of https://github.com/DataDog/dd-trace-rb/pull/2848_

With Remote Configuration, it will be possible to override local options remotely.

As a side effect, a decision has also been made to do not allow local options to override remote configuration, if a remote configuration is applied.

Effectively, any specific configuration option that is configured by Remote Configuration can only be reconfigured by Remote Configuration. This behavior only applies to the specific option that was changed by Remote Configuration, not the whole configuration set as a whole.

The effective configuration precedence is now:
1. Remote Configuration.
2. Custom options inside a `Datadog.configure` block.
3. Environment variables.

https://github.com/DataDog/dd-trace-rb/pull/2920 has been merged into this branch. It uses precedence to check if the current option is using the default value or not.


lib/datadog/core/configuration/option.rb
```diff
@@ -9,17 +9,36 @@ class Option
...
         end
 
-        def set(value)
+        # Overrides the current value for this option if the `precedence` is equal or higher than
+        # the previously set value.
+        # The first call to `#set` will always store the value regardless of precedence.
+        #
+        # @param value [Object] the new value to be associated with this option
+        # @param precedence [Precedence] from what precedence order this new value comes from
+        def set(value, precedence: Precedence::PROGRAMMATIC)
```
ivoanjo, line 35: Since this API is not public (...I think...?) perhaps it would make sense to not have a default for `precedence:`? I think anyway you've covered all callers already :)

lib/datadog/core/configuration/option.rb
```diff
@@ -9,17 +9,36 @@ class Option
         attr_reader \
           :definition
 
+        # Option setting precedence. Higher number means higher precedence.
+        module Precedence
+          REMOTE_CONFIGURATION = 2
+          PROGRAMMATIC = 1
+          DEFAULT = 0
```
ivoanjo, line 16: Perhaps worth documenting that "default" here includes defaults in the code as well as environment variables (e.g. there's two levels that are somewhat collapsed into one, that we don't explicitly represent).

lib/datadog/core/configuration/option.rb
```diff
@@ -9,17 +9,36 @@ class Option
...
 
-        def set(value)
+        # Overrides the current value for this option if the `precedence` is equal or higher than
+        # the previously set value.
+        # The first call to `#set` will always store the value regardless of precedence.
+        #
+        # @param value [Object] the new value to be associated with this option
+        # @param precedence [Precedence] from what precedence order this new value comes from
+        def set(value, precedence: Precedence::PROGRAMMATIC)
+          return @value if precedence < @precedence_set # Cannot override higher precedence value
```
ivoanjo, line 36: On the topic of warnings when there are clashes between settings (comment above), I think we should at least print a warning when a setting gets ignored here.

E.g. it's very counter-intuitive that I would run a `Datadog.configure { }` block 10 minutes after I started my application and my setting would get outright ignored because in the meanwhile it got set via remote configuration.

lib/datadog/core/configuration/option.rb
```diff
@@ -9,17 +9,36 @@ class Option
         attr_reader \
           :definition
 
+        # Option setting precedence. Higher number means higher precedence.
+        module Precedence
+          REMOTE_CONFIGURATION = 2
+          PROGRAMMATIC = 1
```
ivoanjo, line 15: Minor: I wonder if there should exist an internal local override mode that is the highest priority (even if we don't document it)? Even if only for testing, it seems weird that the highest priority is remote configuration and that's the latest word on the subject.

docs/GettingStarted.md
```diff
@@ -2135,17 +2135,24 @@ LogJob.perform_async('login')
```
ivoanjo, line 2135: You mention in the PR description:

> This is part of https://github.com/DataDog/dd-trace-rb/pull/2848, and is being merged into a feature branch.

To be honest, this seems kinda harmless to merge directly to master to avoid having an even bigger feature branch that needs to be updated, rebased, etc.

Just a suggestion from my side, since this technique is what I almost always end up doing ;)

ivoanjo (commented):
Looks reasonable! I left a bunch of suggestions, but I don't think any of them are blocking -- let me know if you'd prefer to merge this as-is and I'll be happy to approve it :)

lib/datadog/core/configuration/option.rb
```diff
@@ -63,6 +84,10 @@ def context_exec(*args, &block)
         def context_eval(&block)
           @context.instance_eval(&block)
         end
+
+        # Used for testing
+        attr_reader :precedence_set
+        private :precedence_set
```
GustavoCaso, line 90: I believe this could be rewritten to: 

```ruby
private
attr_reader :precedence_set
```

I think is idiomatic ruby

docs/GettingStarted.md
marcotc: Into master we go!

docs/GettingStarted.md
```diff
@@ -2135,17 +2135,24 @@ LogJob.perform_async('login')
 
 ## Additional configuration
 
-To change the default behavior of Datadog tracing, you can set environment variables, or provide custom options inside a `Datadog.configure` block, e.g.:
+To change the default behavior of Datadog tracing, you can use, in order of precedence:
```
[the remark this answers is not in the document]
marcotc: 👍

docs/GettingStarted.md
```diff
@@ -2135,17 +2135,24 @@ LogJob.perform_async('login')
...
+    c.tracing.test_mode.enabled = (ENV['RACK_ENV'] == 'test')
+    end
+    ```
+3. Environment variables.
 
-  c.tracing.report_hostname = true
-  c.tracing.test_mode.enabled = (ENV['RACK_ENV'] == 'test')
-end
-```
+**If a higher precedence value is set for an option, setting that option with a lower precedence value will not change its effective value.**
```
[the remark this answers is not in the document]
marcotc: I implemented it as a debug log, 🙇.

lib/datadog/core/configuration/option.rb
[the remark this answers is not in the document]
marcotc: 🎶  Enum, num, num...

marcotc: Done!

lib/datadog/core/configuration/option.rb
marcotc: Done!

lib/datadog/core/configuration/option.rb
marcotc: That can make sense!
I don't think there's a use case for that yet, because we reset settings in between each test run.

But I'm not opposed to adding it if a use case arises.

lib/datadog/core/configuration/option.rb
marcotc: I tried, but Rubocop won't let me, not even with (`AllowModifiersOnSymbols: true`, which is the default) https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/AccessModifierDeclarations

lib/datadog/core/configuration/option.rb
marcotc: I think the default being "the user will override anything, except Remote Configuration" is the proper default here.

I don't like that this is a public API, but hiding it also prevents users writing extensions from being able to set Precedence::DEFAULT values, which is a valid use case.

I don't think this is the perfect API, but it currently models correctly the model we are going for for remote vs local settings.

lib/datadog/core/configuration/option.rb
ivoanjo: Yeah, good point, no use adding more complexity for a speculative use case.

spec/datadog/core/configuration/base_spec.rb
```diff
@@ -142,6 +142,81 @@
...
+              expect(configuration.fake_test.using_default?(:without_default)).to be(false)
+            end
+          end
+
+          context 'when not set but accessed' do
+            it 'returns true' do
+              configuration.fake_test.without_default
+              expect(configuration.fake_test.using_default?(:without_default)).to be(true)
+            end
+          end
```
ivoanjo, line 200: I believe this came from @GustavoCaso's PR, but... it's weird to me to adopt these semantics. E.g. when would it be useful to have "no default" change from false to true depending on it being accessed or not?

This seems like a potential foot-gun being armed and primed to go.

spec/datadog/core/configuration/options_spec.rb
```diff
@@ -114,6 +114,23 @@
...
+
+          context 'with precedence' do
+            subject(:set_option) { options_object.set_option(name, value, precedence: precedence) }
+            let(:precedence) { 123 }
+
+            it 'sets the option precedence' do
+              set_option
+              expect(options_object.options[name].send(:precedence_set)).to eq(precedence)
+            end
+          end
```
ivoanjo, line 133: I wonder if we should enforce that precedence is one of the defined ones? Just to make sure nobody thinks to provide their own custom level and wreck havoc.

ivoanjo (approved):
I've left a few more notes for your consideration, but I think it's in good enough shape that I'll press the magic button and let you make the decisions :+1:

spec/datadog/core/configuration/options_spec.rb
GustavoCaso: I agree with that. We should probably validate that the precedence is part of the ones defined by us 😄

spec/datadog/core/configuration/base_spec.rb
GustavoCaso: That is an excellent point @ivoanjo 

Maybe we should always return `false` when the option hasn't been accessed or configured once. That way we always rely on the precedence value for returning the result. 

```ruby
def using_default?(option)
  return options[option].default_precedence? if options[option]

  false
end
```

@ivoanjo @marcotc any thoughts?

ivoanjo: It's somewhat interesting to me that you propose that the default be `false`.

So, part of my confusion on what to suggest here is that up until now I've only seen the implementation of this API, and not really what we're going to do with it.

From some our chats, I thought the intention is for `using_default?` was to be used first as a way for remote configuration to decide "has the customer provided their own config directly or not?".

And in that situation, it kinda looks to me that we'd want `using_default?` to be `true` as a fallback when not accessed. That's becasuse, _in practice there's no default_ but the intention seems to be (?) that we're using this check as a proxy for _did the customer set something via code or not?_, and the answer to me is that they did not, and thus, it should count as `true` (= using default).

But again I may be misunderstanding all of the above :)

GustavoCaso: @ivoanjo, thanks for the feedback. 

You are right! We should default to `true` since the values haven't been yet configured or accessed, meaning the user or remote configuration hasn't been configured yet.

> I thought the intention is for using_default? was to be used first as a way for remote configuration to decide "has the customer provided their own config directly or not?".

That use case, while valid, is not why I introduced the code. In AppSec we need to check if the user has configured a certain setting. 

> And in that situation, it kinda looks to me that we'd want using_default? to be true as a fallback when not accessed.

Right!

Let me update the code to reflect the conversation above

GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/2915/commits/bb39eca0555d94c44a63dd9c0903b5293d746654

GustavoCaso:
@marcotc makes sure to pull before doing any more work. I added a new commit changing some behaviour from `using_default?`

lib/datadog/core/configuration/option.rb
```diff
@@ -6,20 +6,46 @@ module Configuration
...
       class Option
-        attr_reader \
-          :definition
+        attr_reader :definition
+
+        # Option setting precedence. Higher number means higher precedence.
+        module Precedence
+          REMOTE_CONFIGURATION = [2, :remote_configuration].freeze
+          PROGRAMMATIC = [1, :programmatic].freeze
+          DEFAULT = [0, :default].freeze
```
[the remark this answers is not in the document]
marcotc: Tempting, but it's ok with the two components because I use the human readable component in the debug log, and it looks nice and clean.

docs/GettingStarted.md
```diff
@@ -2844,6 +2851,14 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
 
 Without this flag, profiles for short-lived Resque jobs will not be available as Resque kills worker processes before they have a chance to submit this information.
 
+## Remote Configuration
+
+<!--
+TODO: Flush out consistent language regarding Remote Configuration documentation before
+releasing the Remote Configuration feature.
+This could possibly be a simple link to a central documentation page.
+-->
```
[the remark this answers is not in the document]
marcotc: Thank you, I read your previous comment but didn't actually implement the changes.
This section is now replaced by a link to https://docs.datadoghq.com/agent/remote_config.
🙇

lib/datadog/core/configuration/option.rb
```diff
@@ -6,20 +6,46 @@ module Configuration
...
+        # @param precedence [Precedence] from what precedence order this new value comes from
+        def set(value, precedence: Precedence::PROGRAMMATIC)
+          # Cannot override higher precedence value
+          if precedence[0] < @precedence_set[0]
+            Datadog.logger.debug do
+              "Option '#{definition.name}' not changed to '#{value}' (precedence: #{precedence[1]}) because the higher " \
+                "precedence value '#{@value}' (precedence: #{@precedence_set[1]}) was already set."
+            end
+
+            return @value
```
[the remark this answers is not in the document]
marcotc: I was concerned with our own internal usage of [`Datadog.configure {}`](https://github.com/search?q=repo%3ADataDog%2Fdd-trace-rb+%2FDatadog.configure%5Cb%2F+path%3A%2F%5Elib%5C%2Fdatadog%5C%2Ftracing%5C%2Fcontrib%5C%2F%2F&type=code), which the user has no control over. 
These, in my mind, could trigger this log line.

But inspecting them closer, none of the internal use cases change any of the options that Remote Configuration change today, and they don't seem likely to be supported through Remote Configuration in the future.

I've bumped it up to INFO: I'm not confident WARN is the best option because Remote Configuration overriding your local configuration leaves the application in a stable state, so the WARN might not actually be of concern. I don't want to have to introduce an introspection API for users to avoid triggering the WARN ("let me check if this option was set by remote configuration before setting it locally, otherwise ddtrace will WARN").

spec/datadog/core/configuration/options_spec.rb
marcotc: I changed the test to use a real precedence value, as to not lead people into thinking it can be anything.

lib/datadog/core/configuration/option.rb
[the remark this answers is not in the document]
ivoanjo: Seems reasonable!

PR #2918, opened 2023-06-20
`peer.service` extraction logic w/AWS

zarirhamza:
Adds `peer.service` extraction logic and implements it in the AWS integration

`peer.service` is a tag that needs to be set by extracting some values from certain "precursor" tags which vary based on span type and of which the complete list is already agreed upon internally which are checked for in this PR but may need to be fully implemented in the future for certain integrations. This PR implements the logic of checking and extracting these values within the `set_peer_service_from_source` method which checks span type and the existence of various "precursor" tags before setting `peer.service` and the source from which it was extracted.

Note however that `peer.service` is only extracted when not explicitly set by the user which is checked by the `should_set_peer_service` method which ensures that the existence of any `peer.service` tags implies that no further ones will be set. In addition, changing span.service ties directly with `peer.service` so we check if that value is unique to also imply `peer.service` is changed

Both the aforementioned methods are thoroughly tested as well.

The two methods are called in `set_peer_service` which in the future will be called in all target integrations. This PR has one complete working example with AWS as well as updated tests to cover all various AWS services.

NOTE: Please do not merge without @delner direct approval but feel free to review and ask questions for context to me.

### DIFF
--- a/lib/datadog/tracing/contrib/ext.rb
+++ b/lib/datadog/tracing/contrib/ext.rb
@@ -30,6 +30,24 @@ module Datadog
           TAG_RABBITMQ_ROUTING_KEY = 'messaging.rabbitmq.routing_key'
           TAG_KAFKA_BOOTSTRAP_SERVERS = 'messaging.kafka.bootstrap.servers'
         end
+
+        module SpanAttributeSchema
+          PEER_SERVICE_SOURCE_AWS = Array[Aws::Ext::TAG_QUEUE_NAME,
+            Aws::Ext::TAG_TOPIC_NAME,
+            Aws::Ext::TAG_STREAM_NAME,
+            Aws::Ext::TAG_TABLE_NAME,
+            Aws::Ext::TAG_BUCKET_NAME,
+            Aws::Ext::TAG_RULE_NAME,
+            Aws::Ext::TAG_STATE_MACHINE_NAME,]
+
+          # TODO: consolidate all db.name tags to 1 tag name and add to array here
+          PEER_SERVICE_SOURCE_DB = Array[Tracing::Contrib::Ext::DB::TAG_INSTANCE]
+
+          # TODO: add kafka bootstrap servers tag
+          PEER_SERVICE_SOURCE_MSG = Array[]
+
+          PEER_SERVICE_SOURCE_RPC = Array[Tracing::Contrib::Ext::RPC::TAG_SERVICE]
+        end
       end
     end
   end
--- a/lib/datadog/tracing/contrib/ext.rb
+++ b/lib/datadog/tracing/contrib/ext.rb
@@ -30,24 +30,6 @@ module Datadog
           TAG_RABBITMQ_ROUTING_KEY = 'messaging.rabbitmq.routing_key'
           TAG_KAFKA_BOOTSTRAP_SERVERS = 'messaging.kafka.bootstrap.servers'
         end
-
-        module SpanAttributeSchema
-          PEER_SERVICE_SOURCE_AWS = Array[Aws::Ext::TAG_QUEUE_NAME,
-            Aws::Ext::TAG_TOPIC_NAME,
-            Aws::Ext::TAG_STREAM_NAME,
-            Aws::Ext::TAG_TABLE_NAME,
-            Aws::Ext::TAG_BUCKET_NAME,
-            Aws::Ext::TAG_RULE_NAME,
-            Aws::Ext::TAG_STATE_MACHINE_NAME,]
-
-          # TODO: consolidate all db.name tags to 1 tag name and add to array here
-          PEER_SERVICE_SOURCE_DB = Array[Tracing::Contrib::Ext::DB::TAG_INSTANCE]
-
-          # TODO: add kafka bootstrap servers tag
-          PEER_SERVICE_SOURCE_MSG = Array[]
-
-          PEER_SERVICE_SOURCE_RPC = Array[Tracing::Contrib::Ext::RPC::TAG_SERVICE]
-        end
       end
     end
   end
--- a/lib/datadog/tracing/contrib/ext.rb
+++ b/lib/datadog/tracing/contrib/ext.rb
@@ -28,6 +28,7 @@ module Datadog
         module Messaging
           TAG_SYSTEM = 'messaging.system'
           TAG_RABBITMQ_ROUTING_KEY = 'messaging.rabbitmq.routing_key'
+          TAG_KAFKA_BOOTSTRAP_SERVERS = 'messaging.kafka.bootstrap.servers'
         end
       end
     end
--- a/lib/datadog/tracing/contrib/ext.rb
+++ b/lib/datadog/tracing/contrib/ext.rb
@@ -12,13 +12,20 @@ module Datadog
           TAG_SYSTEM = 'db.system'
           TAG_STATEMENT = 'db.statement'
           TAG_ROW_COUNT = 'db.row_count'
+          PEER_SERVICE_SOURCES = Array[TAG_INSTANCE,
+            Tracing::Metadata::Ext::NET::TAG_DESTINATION_NAME,
+            Tracing::Metadata::Ext::TAG_PEER_HOSTNAME,
+            Tracing::Metadata::Ext::NET::TAG_TARGET_HOST,].freeze
         end
 
         module RPC
           TAG_SYSTEM = 'rpc.system'
           TAG_SERVICE = 'rpc.service'
           TAG_METHOD = 'rpc.method'
-
+          PEER_SERVICE_SOURCES = Array[TAG_SERVICE,
+            Tracing::Metadata::Ext::NET::TAG_DESTINATION_NAME,
+            Tracing::Metadata::Ext::TAG_PEER_HOSTNAME,
+            Tracing::Metadata::Ext::NET::TAG_TARGET_HOST,].freeze
           module GRPC
             TAG_STATUS_CODE = 'rpc.grpc.status_code'
             TAG_FULL_METHOD = 'rpc.grpc.full_method'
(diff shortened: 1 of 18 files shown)


lib/datadog/tracing/contrib/span_attribute_schema.rb
```diff
@@ -22,6 +22,75 @@ def default_span_attribute_schema?
...
+        def set_peer_service(span)
+          should_set_peer_service(span) && set_peer_service_from_source(span)
+          # if above
+          # then remap + remapped from (SKIP)
+          # else
+          # debug that peer service could not be set
+          # end
+        end
+
+        def should_set_peer_service(span)
```
marcotc, line 36: In the spirit of Ruby patterns, a `?` is normally how we define boolean methods that answer a simple question:
```suggestion
        def set_peer_service?(span)
```

lib/datadog/tracing/contrib/span_attribute_schema.rb
```diff
@@ -22,6 +22,75 @@ def default_span_attribute_schema?
...
+              Aws::Ext::TAG_TOPIC_NAME,
+              Aws::Ext::TAG_STREAM_NAME,
+              Aws::Ext::TAG_TABLE_NAME,
+              Aws::Ext::TAG_BUCKET_NAME,
+              Aws::Ext::TAG_RULE_NAME,
+              Aws::Ext::TAG_STATE_MACHINE_NAME,]
+          when span.get_tag(Tracing::Contrib::Ext::DB::TAG_SYSTEM)
+            sources = Array[Tracing::Contrib::Ext::DB::TAG_INSTANCE] # DB_NAME tag?
+          when span.get_tag(Tracing::Contrib::Ext::Messaging::TAG_SYSTEM)
+            sources = Array[] # kafka bootstrap servers
```
marcotc, line 73: What does this comment mean?
Is it a call for future action? If so, what action?
Is this documenting that `kafka bootstrap servers` have empty sources? If so, we should write that in the comment.

lib/datadog/tracing/contrib/span_attribute_schema.rb
```diff
@@ -22,6 +22,75 @@ def default_span_attribute_schema?
...
+          when span.get_tag(Aws::Ext::TAG_AWS_SERVICE)
+            sources = Array[Aws::Ext::TAG_QUEUE_NAME,
+              Aws::Ext::TAG_TOPIC_NAME,
+              Aws::Ext::TAG_STREAM_NAME,
+              Aws::Ext::TAG_TABLE_NAME,
+              Aws::Ext::TAG_BUCKET_NAME,
+              Aws::Ext::TAG_RULE_NAME,
+              Aws::Ext::TAG_STATE_MACHINE_NAME,]
+          when span.get_tag(Tracing::Contrib::Ext::DB::TAG_SYSTEM)
+            sources = Array[Tracing::Contrib::Ext::DB::TAG_INSTANCE] # DB_NAME tag?
```
marcotc, line 71: Is this comment here a `TODO`? If so, let's make sure it says so.

lib/datadog/tracing/contrib/span_attribute_schema.rb
```diff
@@ -22,6 +22,75 @@ def default_span_attribute_schema?
...
+        def set_peer_service_from_source(span)
+          case
+          when span.get_tag(Aws::Ext::TAG_AWS_SERVICE)
+            sources = Array[Aws::Ext::TAG_QUEUE_NAME,
+              Aws::Ext::TAG_TOPIC_NAME,
+              Aws::Ext::TAG_STREAM_NAME,
+              Aws::Ext::TAG_TABLE_NAME,
+              Aws::Ext::TAG_BUCKET_NAME,
+              Aws::Ext::TAG_RULE_NAME,
+              Aws::Ext::TAG_STATE_MACHINE_NAME,]
```
marcotc, line 69: For all `source = ...` in this method, because the value set to `sources` is always the same, and it is a complex value (a new array is created on every invocation), we should store the value in a constant, so that we don't have to recompute it on every invocation of `set_peer_service_from_source`.

Example:
```suggestion
            sources = PEER_SERVICE_SOURCE_AWS
```

lib/datadog/tracing/contrib/span_attribute_schema.rb
```diff
@@ -22,6 +22,75 @@ def default_span_attribute_schema?
...
+            (span.get_tag(Tracing::Metadata::Ext::TAG_KIND) == Tracing::Metadata::Ext::SpanKind::TAG_PRODUCER)) &&
+              (Datadog.configuration.tracing.span_attribute_schema ==
+                  Tracing::Configuration::Ext::SpanAttributeSchema::VERSION_ONE) # OR if env var is set
+            return true
+          end
+
+          false
+        end
+
+        def set_peer_service_from_source(span)
```
marcotc, line 60: I think that this method does exactly is not super clear to me just by reading it, could you please add some method level documentation?

lib/datadog/tracing/contrib/span_attribute_schema.rb
```diff
@@ -22,6 +22,75 @@ def default_span_attribute_schema?
...
+          sources.each do |source|
+            source_val = span.get_tag(source)
+            next unless source_val && source_val != ''
+
+            span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE, source_val)
+            span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE_SOURCE, source)
+            break
+          end
+          true
+        end
```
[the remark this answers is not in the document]
marcotc:
```suggestion
          sources.each do |source|
            source_val = span.get_tag(source)
            next unless source_val && source_val != ''

            span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE, source_val)
            span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE_SOURCE, source)
            return true
          end
          false
        end
```

marcotc:
@zarirhamza, the description you wrote for the PR is very good!
To facilitate our understanding of the source code in the future, could you capture some of that description in the code itself (as comments, for example)?

For example, it would be great to have this explanation _"Note however that peer.service is only extracted when not explicitly set by the user which is checked by the should_set_peer_service method which ensures that the existence of any peer.service tags implies that no further ones will be set."_ paraphrased in the relevant method/line so we know why this check is there and what it's trying to check for.

lib/datadog/tracing/contrib/span_attribute_schema.rb
zarirhamza: These are all markers for tags that do not exist yet but should be added in future PRs as precursors. Will change them to TODO comments

lib/datadog/tracing/contrib/span_attribute_schema.rb
zarirhamza: I put the constant within the same file if that's ok. Otherwise let me know a more suitable location for it, thanks!

lib/datadog/tracing/contrib/span_attribute_schema.rb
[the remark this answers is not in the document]
zarirhamza: good catch!

lib/datadog/tracing/contrib/span_attribute_schema.rb
zarirhamza: Nvm moved it to the overall contrib constants file all good

zarirhamza:
> @zarirhamza, the description you wrote for the PR is very good! To facilitate our understanding of the source code in the future, could you capture some of that description in the code itself (as comments, for example)?
> 
> For example, it would be great to have this explanation _"Note however that peer.service is only extracted when not explicitly set by the user which is checked by the should_set_peer_service method which ensures that the existence of any peer.service tags implies that no further ones will be set."_ paraphrased in the relevant method/line so we know why this check is there and what it's trying to check for.

Took a look and added more information. Feel free to take another look and let me know if all concerns were addressed.

lib/datadog/tracing/contrib/span_attribute_schema.rb
```diff
@@ -22,6 +22,76 @@ def default_span_attribute_schema?
...
+          if ps && (ps != '')
+
+            # if peer.service is not equal to span.service we know it is edited
+            if ps != span.service
+              span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE_SOURCE, Tracing::Metadata::Ext::TAG_PEER_SERVICE)
+              return false
+            end
+
+            # if span.service is not the global service, we know it was changed to change the peer.service value
+            if (ps == span.service) && (span.service != Datadog.configuration.service)
```
marcotc, line 44: I believe `ps == span.service` is not needed here, given we return early in the `if` condition prior if `ps != span.service`:
```suggestion
            if span.service != Datadog.configuration.service
```

zarirhamza: That is correct, fixed!

delner (changes requested):
Moving in the right direction. Some small changes, and perhaps some slightly bigger ones worth looking at.

lib/datadog/tracing/contrib/ext.rb
```diff
@@ -28,6 +28,7 @@ module GRPC
         module Messaging
           TAG_SYSTEM = 'messaging.system'
           TAG_RABBITMQ_ROUTING_KEY = 'messaging.rabbitmq.routing_key'
+          TAG_KAFKA_BOOTSTRAP_SERVERS = 'messaging.kafka.bootstrap.servers'
```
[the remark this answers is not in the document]
zarirhamza: Agreed, we can move the Kafka constant to the respective integration and I don't think Ruby even has a rabbit integration so we can ignore that for now

lib/datadog/tracing/contrib/span_attribute_schema.rb
```diff
@@ -22,6 +36,84 @@ def default_span_attribute_schema?
           Datadog.configuration.tracing.span_attribute_schema ==
             Tracing::Configuration::Ext::SpanAttributeSchema::DEFAULT_VERSION
         end
+
+        # TODO: implement function in all integrations with spankind
+        def set_peer_service(span, sources)
+          set_peer_service?(span) && set_peer_service_from_source(span, sources)
```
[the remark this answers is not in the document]
zarirhamza: We actually need to keep the function `set_peer_service` for future manipulation (i.e remapping). However, we can roll the check into the `set_peer_service_from_source` function:

```ruby
def set_peer_service(span, sources)
  set_peer_service_from_source(span, sources)

  #add remapping logic here later
end

def set_peer_service_from_source(span, sources)
  return unless set_peer_service?(span)

  #Remaining logic of function continues here
end
```

delner (approved):
This is now looking much improved; concise, simple. It also looks a lot easier to reason about.

I think this is good to merge to your feature branch. When we have the rest of the complimentary changes on it (and we think it's ready to release), we'll merge the whole thing to master.

Thanks for all your hard work on this one @zarirhamza!

PR #2919, opened 2023-06-20
Add default_option method to Core::Configuration

GustavoCaso:
**What does this PR do?**

This PR adds  the `default_option?` to `Core::Configuration` 

**Motivation**

I will port AppSec (ASM) configuration to use the one in `Core::Configuration`. The one key feature that the ASM configuration has that the core configuration has is the ability to check if a value has been set or if it is using the default value.

Here is the code ASM:

https://github.com/DataDog/dd-trace-rb/blob/f7d1e5e42419e9adced5e05d82eda8bdd59d5b48/lib/datadog/appsec/configuration.rb#L88-L90

Once the code is merged, we can migrate the ASM configuration to use the core configuration 

**Additional Notes**

**How to test the change?**

### DIFF
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -6,8 +6,7 @@ module Datadog
       # Represents an instance of an integration configuration option
       # @public_api
       class Option
-        attr_reader \
-          :definition
+        attr_reader :definition, :is_set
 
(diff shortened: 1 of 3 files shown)


GustavoCaso:
Close in favour of https://github.com/DataDog/dd-trace-rb/pull/2920

PR #2920, opened 2023-06-20
Add `using_default?` method to Core::Configuration

GustavoCaso:
**What does this PR do?**

This PR adds  the `using_default?` to `Core::Configuration` 

**Motivation**

I will port AppSec (ASM) configuration to use the one in `Core::Configuration`. The one key feature that the ASM configuration has that the core configuration has is the ability to check if a value has been set or if it is using the default value.

Here is the code ASM:

https://github.com/DataDog/dd-trace-rb/blob/f7d1e5e42419e9adced5e05d82eda8bdd59d5b48/lib/datadog/appsec/configuration.rb#L88-L90

Once the code is merged, we can migrate the ASM configuration to use the core configuration 

**Additional Notes**

**How to test the change?**

### DIFF
--- a/lib/datadog/core/configuration/options.rb
+++ b/lib/datadog/core/configuration/options.rb
@@ -84,6 +84,12 @@ module Datadog
             self.class.options.key?(name)
           end
 
+          def using_default?(option)
+            return options[option].default_precedence? if options[option]
+
+            !self.class.options[option].default.nil?
+          end
+
           def options_hash
             self.class.options.merge(options).each_with_object({}) do |(key, _), hash|
               hash[key] = get_option(key)
--- a/spec/datadog/core/configuration/base_spec.rb
+++ b/spec/datadog/core/configuration/base_spec.rb
@@ -142,6 +142,81 @@ RSpec.describe Datadog::Core::Configuration::Base do
         end
       end
 
+      describe '#using_default?' do
+        let(:configuration) do
+          Class.new do
+            include Datadog::Core::Configuration::Base
+
+            settings :fake_test do
+              option :enabled, default: true
+              option :without_default
+            end
+          end.new
+        end
+
+        context 'with default option' do
+          context 'when not set' do
+            it 'returns true' do
+              expect(configuration.fake_test.using_default?(:enabled)).to be(true)
+            end
+          end
+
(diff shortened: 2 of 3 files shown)


spec/datadog/core/configuration/base_spec.rb
```diff
@@ -142,6 +142,81 @@
...
+              configuration.fake_test.reset!
+              expect(configuration.fake_test.default_option?(:enabled)).to be(true)
+            end
+          end
+        end
+
+        context 'without default option' do
+          context 'when not set' do
+            it 'returns true' do
+              expect(configuration.fake_test.default_option?(:without_default)).to be(false)
```
ivoanjo, line 191: The test description doesn't match the assertion :)

lib/datadog/core/configuration/options.rb
```diff
@@ -84,6 +84,12 @@ def option_defined?(name)
             self.class.options.key?(name)
           end
 
+          def default_option?(option)
```
ivoanjo, line 87: Minor: Should this be `using_default?` or `default_value?`? `default_option?` sounds a bit weird to me because it seems to be asserting on the option, rather than what it is set to.

ivoanjo (commented):
Left a few notes!

lib/datadog/core/configuration/options.rb
```diff
@@ -84,6 +84,12 @@ def option_defined?(name)
...
+            return options[option].default_precedence? if options[option]
+
+            !self.class.options[option].default.nil?
```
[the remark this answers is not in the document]
GustavoCaso: I thought about it more and leaving it as it made sense. 

In the case of an option with a default value that has yet to be accessed. It makes sense to return `true` since the next time the option is accessed, the returned value is the default value.  If we have set a value by any change, the result from calling `default_option?` would be false.

In the case of an option with a default value that has yet to be accessed. It makes sense to return `false` since there is no default set.

spec/datadog/core/configuration/base_spec.rb
GustavoCaso: Ahh the wonder of copy pasta 🤣

marcotc (approved):

PR #2922, opened 2023-06-21
[APPSEC-3634] Replace AppSec custom configuration with Core::Configuration

GustavoCaso:
**What does this PR do?**

Replace the custom code from ASM mode to define settings.

**Motivation**

To have a consistent way to define settings across components. That way any addition to the core configuration is propagated to all components.

**Additional Notes**

**How to test the change?**

CI and System test should be 🟢


lloeki (approved):
Looks fine to me!

PR #2927, opened 2023-06-22
[PROF-7645] Include state of "no signals" workaround when reporting profiles

ivoanjo:
**What does this PR do?**:

In #2873 we introduced an alternative mode to the profiler, called the "no signals" workaround.

This PR makes use of the new libdatadog internal metadata reporting support to send the state of this setting along with reported profiles.

**Motivation**:

We'll use this to track how many customers are using the "no signals" workaround, as well as when debugging issues.

**Additional Notes**:

This PR will break in CI since it depends on the libdatadog support for the internal metadata reporting, which has not been released yet as of this writing (https://github.com/datadog/libdatadog/pull/181).

My plan is to open a separate PR to update the libdatadog version in use by ddtrace, and then rebase this PR on top of it.

This PR has been tested locally, and it's working and ready for review.

Once it's approved, I'll make sure it's merged only after the new libdatadog version is available and CI is green.

**How to test the change?**:

This change includes test coverage. Testing it locally requires building libdatadog locally as well -- feel free to ping me and I can guide you through it.


marcotc (approved):
Looks good!

I do notice that this is quite a bit of work to send an already existing value ("no signal") to the profiler payload.

ivoanjo:
> I do notice that this is quite a bit of work to send an already existing value ("no signal") to the profiler payload.

Yeap, that's a good point. We did consider sending it as tag in profiles (some `_dd.profiling...` thing), but this new mechanism allows us to more easily omit the information from the profiling UX (because we don't think it's relevant to expose to customers at the moment).

Half of the work in this PR is propagating the `internal_metadata_json`, which in the future could be reused for other such flags.

Getting the configuration to the exporter also needed some work -- I've been wondering if it makes sense to introduce a tiny object to represent the resolved profiler settings (e.g. different from configuration because some settings may be the result of dynamic probing/decisions) that could then be passed to the multiple objects that need it. That would make it easier to "pass along" more settings by modifying only the settings object and whatever needs to consume it, and not have to change all of the plumbing in between. I haven't decided to jump on this one yet but... it's on the back of my mind.

This all to say: yes "+110 −25" to send a fancy tag is a bit of work indeed -- I hope it's not representative of future additions!

PR #2928, opened 2023-06-23
Make sure the GRPC client interceptor actually returns the response

KJTsanaktsidis:
**What does this PR do?**
Datadog::Tracing::Contrib::GRPC::DatadogInterceptor::Client#trace is currently causing all GRPC client operations to return 0.0 instead of whatever they were supposed to return. This is a bit of a Ruby gotcha.

This method returns :three

```ruby
def broken_method
    begin
        :one
    rescue => e
        :two
    else
        :three
    end
end
```

So when `#trace` is putting `yield` in a `begin` block with an `else`, it's not actually returning the result of yield; it's returning the result of the else block. I guess this was broken in https://github.com/DataDog/dd-trace-rb/pull/2620

Fix it by explicitly returning the result of yield.

**Motivation**
All my GRPC client methods started returning `0.0` after upgrading ddtrace 😱 

**Additional Notes**

**How to test the change?**
Now my GRPC client methods don't return 0.0! I included a test in this patch.

### DIFF
--- a/lib/datadog/tracing/contrib/grpc/datadog_interceptor/client.rb
+++ b/lib/datadog/tracing/contrib/grpc/datadog_interceptor/client.rb
@@ -29,15 +29,15 @@ module Datadog
                 annotate!(trace, span, keywords, formatter)
 
                 begin
-                  yield
+                  result = yield
                 rescue StandardError => e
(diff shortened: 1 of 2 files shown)


marcotc (approved):
Thank you, @KJTsanaktsidis!

PR #2929, opened 2023-06-23
Actions:Allow type check to run at PR level

marcotc:
All external PRs are stuck because "Check types" does run for external contributors, not even after manual approval (see https://github.com/DataDog/dd-trace-rb/pull/2928, for example):
Screenshot 2023-06-23 at 2 09 41 PM

Interestingly CodeQL does work correctly, running on approval:
Screenshot 2023-06-23 at 2 10 37 PM

My research was not 100% conclusive, but I believe that the `on` trigger is the reason.

This PR change the `on` trigger to look more similar to CodeQL's, specifically regarding having an explicit `pull_request` trigger.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,5 +1,10 @@
 name: Check
-on: [push]
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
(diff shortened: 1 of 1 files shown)


ivoanjo (approved):
Seems reasonable! Thanks for improving the experience for contributors :)

LGTM :+1:

PR #2931, opened 2023-06-26
Remove `lazy` option from the configuration option

GustavoCaso:
**What does this PR do?**

This PR removes the `lazy` option in favour of checking the content of `default.` 

These changes are part of an internal discussion of the ruby team 😄 

**Motivation**

The `lazy` option is superfluous, the check to either evaluate `default` as a block or a static value could be achieved by checking the type of value `default` is. `default.is_a?(Proc)`

There is an edge case: 
The default value is a block.

When we provide a block to `default` and do not specify it as `lazy`, we store the block as the default value. I added a new option, `experimental_default_proc`, to circumvent that edge case. Pushing us to be more explicit about those edge cases makes the code more readable and easier to grok.

Since this new option is experimental and we might not need it in the near future, I gave it a name that it clear to the reader that they should not rely on it and that is subject to change.

I added a deprecation warning to `lazy,` so if anyone is using the core code to build their settings, they can migrate to the new system. 

**Additional Notes**

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/core/configuration/option_definition.rb
+++ b/lib/datadog/core/configuration/option_definition.rb
@@ -13,7 +13,6 @@ module Datadog
           :default,
           :delegate_to,
           :depends_on,
-          :lazy,
           :name,
           :on_set,
           :resetter,
@@ -24,7 +23,6 @@ module Datadog
           @default = meta[:default]
           @delegate_to = meta[:delegate_to]
           @depends_on = meta[:depends_on] || []
-          @lazy = meta[:lazy] || false
           @name = name.to_sym
           @on_set = meta[:on_set]
           @resetter = meta[:resetter]
@@ -48,7 +46,6 @@ module Datadog
             @delegate_to = nil
             @depends_on = []
             @helpers = {}
-            @lazy = false
             @name = name.to_sym
             @on_set = nil
             @resetter = nil
@@ -78,10 +75,6 @@ module Datadog
             @helpers[name] = block
           end
 
-          def lazy(value = true)
-            @lazy = value
-          end
-
           def on_set(&block)
             @on_set = block
           end
@@ -105,7 +98,6 @@ module Datadog
             default(options[:default]) if options.key?(:default)
             delegate_to(&options[:delegate_to]) if options.key?(:delegate_to)
             depends_on(*options[:depends_on]) if options.key?(:depends_on)
-            lazy(options[:lazy]) if options.key?(:lazy)
             on_set(&options[:on_set]) if options.key?(:on_set)
             resetter(&options[:resetter]) if options.key?(:resetter)
             setter(&options[:setter]) if options.key?(:setter)
@@ -121,7 +113,6 @@ module Datadog
               default: @default,
               delegate_to: @delegate_to,
               depends_on: @depends_on,
-              lazy: @lazy,
               on_set: @on_set,
               resetter: @resetter,
               setter: @setter,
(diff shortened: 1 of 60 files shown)


lib/datadog/core/configuration/settings.rb
```diff
@@ -490,7 +472,7 @@ def initialize(*_)
         # @default `->{ Time.now }`
         # @return [Proc<Time>]
         option :time_now_provider do |o|
-          o.default { ::Time.now }
+          o.default { -> { ::Time.now } }
```
GustavoCaso, line 475: This is a weird use case. We want the default value to be a block, so we need to have a block that returns a block. 

Before we did not mark the settings as `lazy`, so rather than execute the block, we store the block as the default value. To achieve the same thing, we have to wrap the block into a block.

marcotc:
:wave: @GustavoCaso, why is lazy being removed?

GustavoCaso:
@marcotc I am preparing an RFC to reduce some of the complexity of our core configuration code and options. By any means, this is something we have to do, I instead want the guild to agree, but for the cases of the `lazy` option, I wanted to have an example PR for everyone to have a look at.

lloeki:
In my mind referred to when a proc (or callable, really) is evaluated:

- default may be a callable
- some settings may allow callables to be set as values (IIRC there's transport)

`lazy` decides whether these are evaluated upon the setting being set or upon the setting being called once.

So:

- I don't really see a solid argument for a callable being evaluated upon being set
- if a non-callable is set then by definition the value is evaluated right away

If we hold the above as true then it looks like `lazy` is indeed superfluous.

lib/datadog/core/configuration/option_definition.rb
```diff
@@ -11,9 +11,9 @@ class OptionDefinition
 
         attr_reader \
           :default,
+          :experimental_default_proc,
```
ivoanjo, line 14: Here or somewhere on this file, I think it would be useful to document (or repeat from the PR description) the reason why we're calling this experimental, and what are our thoughts concerning the next steps (e.g. either removing it, or renaming it back to non-experimental).

lib/datadog/core/configuration/option_definition.rb
```diff
@@ -78,8 +86,13 @@ def helper(name, *_args, &block)
...
 
-          def lazy(value = true)
-            @lazy = value
+          def lazy(_value = true)
+            Datadog::Core.log_deprecation do
+              'The `lazy` option is deprecated. You no longer need to specify an option to be lazy '\
+              "when the default value is a block.\n"\
+              'In the case edge case that you need the default value to be a block, please use the the expriemntal option:'\
+              'experimental_default_proc'
+            end
```
ivoanjo, line 95: Two notes on this:

* "You no longer need to" should perhaps be "you no longer can" -- this option is gone, it's not that you wouldn't ever need it
* Is it worth telling people to use the experimental option? Should we instead tell them that we don't support it all until we figure out what we should do?

ivoanjo (commented):
Overall I really like this diff! 

I think the missing bit is being a bit more clear on our advice to customers (do we want them to use `experimental_default_proc`? And are they aware it may disappear soon?).

This is one situation where the warnings, as well as the PR description should be considered with advanced customers in mind as the readers, not only our team.

ivoanjo (approved):
:+1: LGTM. It may be worth to triple-check that everyone is happy with this internally, but I see no reason not to merge this in :)

PR #2935, opened 2023-06-29
Add:HTTP header tagging with DD_TRACE_HEADER_TAGS for servers

marcotc:
**What does this PR do?**

This PR adds support for saving the value of HTTP headers as span tags by configuring the environment variable `DD_TRACE_HEADER_TAGS`.

This PR covers HTTP servers: Rack and Sinatra in our case.

**_A follow up PR will cover HTTP clients._**

**Motivation**

Saving HTTP headers as span tags has been supported for Rack and Sinatra, but there was no standard configuration at application-level.

At Datadog, we have the `DD_TRACE_HEADER_TAGS` variable defined across all tracers as the recommended way to configure HTTP header capture.

This PR brings the Ruby tracer on par with other languages that already support this configuration.

**Also, this is one of the features supported by Remote Configuration.**

**How to test the change?**

`rake spec:main` covers the utility methods.
Rack and Sinatra test suites cover the integration testing for this feature.


lib/datadog/tracing/contrib/rack/header_tagging.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
```
marcotc, line 1: This code was moved here from `lib/datadog/tracing/contrib/sinatra/env.rb` and `lib/datadog/tracing/contrib/sinatra/headers.rb`.
Sinatra already had helpers to perform header matching for Rack-style headers.
Rack was also performing the matching, but in a less organized fashion.

Because the `DD_TRACE_HEADER_TAGS` option has to be added to both Rack and Sinatra, this PR merges both header matching implementations into this file.

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -165,6 +166,23 @@ def self.extended(base)
...
+              # * A key value pair, "header name:tag name", which is mapped to the span tag `tag name`.
+              #
+              # You can mix the two types of header declaration in the same list.
+              # Tag names will be normalized based on the [Datadog tag normalization rules](https://docs.datadoghq.com/getting_started/tagging/#defining-tags).
+              #
+              # @default `DD_TRACE_HEADER_TAGS` environment variable, otherwise an empty set of tags
+              # @return [Array<String>]
+              option :header_tags do |o|
+                o.default { env_to_list(Configuration::Ext::ENV_HEADER_TAGS, nil, comma_separated_only: true) }
+                o.lazy
```
GustavoCaso, line 182: No need to use the `lazy` option anymore 😄

lib/datadog/tracing/configuration/http.rb
```diff
@@ -0,0 +1,84 @@
...
+
+              [span_tag, header_value] if header_value
+            end.compact
+          end
+
+          # Returns false if this class was explicitly configured
+          # or left without configuration.
+          def configured?
+            !@header_tags.equal?(EMPTY)
+          end
```
GustavoCaso, line 69: We would be able to remove this logic once we have https://github.com/DataDog/dd-trace-rb/pull/2970 in master

lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -138,8 +139,6 @@ def call(env)
           # rubocop:disable Metrics/MethodLength
           def set_request_tags!(trace, request_span, env, status, headers, response, original_env)
             request_header_collection = Header::RequestHeaderCollection.new(env)
```
GustavoCaso, line 141: It looks like we could pass `request_header_collection`  to `HeaderTagging.tag_request_headers(request_span, env, configuration)`

`HeaderTagging.tag_request_headers(request_span, request_header_collection, configuration)`. Internally in `tag_request_headers` we use the `env` to construct a `Header::RequestHeaderCollection`. We can avoid the duplicate `Header::RequestHeaderCollection`

lib/datadog/tracing/contrib/sinatra/env.rb
```diff
@@ -1,6 +1,7 @@
 require 'time'
 
 require_relative '../../metadata/ext'
+require_relative '../rack/header_collection'
```
GustavoCaso, line 4: It looks like we are not using `header_collection` on this file 😄

lib/datadog/tracing/configuration/http.rb
```diff
@@ -0,0 +1,84 @@
...
+
+          # Receives a case insensitive hash with the request headers and returns
+          # a list of tag names and values that can be set in a span.
+          def request_tags(headers)
+            @request_headers.map do |header_name, span_tag|
+              # Case-insensitive search. {RequestHeaderCollection} already ensures case-insensitiveness.
+              header_value = headers[header_name]
+
+              [span_tag, header_value] if header_value
+            end.compact
```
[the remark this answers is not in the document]
marcotc: I'm not sure, do you think the reduce easier to read?
```ruby
@request_headers.reduce([]) do |array, (header_name, span_tag)|
  # Case-insensitive search
  header_value = headers[header_name]

  array << [span_tag, header_value] if header_value
  array
end
```

lib/datadog/tracing/configuration/http.rb
marcotc: Oh yeah, I think so! 👍 
I'll work on this when both are merged!

lib/datadog/tracing/contrib/sinatra/env.rb
marcotc: Nice catch, this is an artifact from early development. 👍

lib/datadog/tracing/contrib/rack/middlewares.rb
marcotc: I gave it a shot!
I had to make changes inside `HeaderTagging#tag_request_headers` to prevent it from double wrapping the argument, but I think it's a good change.

lib/datadog/tracing/contrib/rack/header_tagging.rb
```diff
@@ -0,0 +1,58 @@
...
+                     whitelist.each_with_object({}) do |header, result|
+                       if headers.key?(header)
+                         result[Tracing::Metadata::Ext::HTTP::ResponseHeaders.to_tag(header)] = headers[header]
+                       else
+                         # Try a case-insensitive lookup
+                         uppercased_header = header.to_s.upcase
+                         matching_header = headers.keys.find { |h| h.upcase == uppercased_header }
+                         if matching_header
+                           result[Tracing::Metadata::Ext::HTTP::ResponseHeaders.to_tag(header)] = headers[matching_header]
+                         end
```
GustavoCaso, line 47: Could we use the `Core::Utils::Hash::CaseInsensitiveWrapper` here? I mostly as, because we add a dedicated module for case insensitive look up but we are doing a hand roll one in the else branch.

spec/datadog/tracing/configuration/http_spec.rb
```diff
@@ -0,0 +1,122 @@
+require 'spec_helper'
+
+require 'datadog/tracing/configuration/http'
+require 'datadog/core/utils/hash'
+
+RSpec.describe Datadog::Tracing::Configuration::HTTP::HeaderTags do
+  subject(:http_header_tags) { described_class.new(header_tags) }
```
GustavoCaso, line 7: I find this `subject` a little confusing. It looks like the real subjects are the ones defined for the shared examples.

Suggestion, should we use a let instead here?

marcotc: Moved to `let`! 👍

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -165,6 +166,22 @@ def self.extended(base)
...
+              #
+              # You can mix the two types of header declaration in the same list.
+              # Tag names will be normalized based on the [Datadog tag normalization rules](https://docs.datadoghq.com/getting_started/tagging/#defining-tags).
+              #
+              # @default `DD_TRACE_HEADER_TAGS` environment variable, otherwise an empty set of tags
+              # @return [Array<String>]
+              option :header_tags do |o|
+                o.default { env_to_list(Configuration::Ext::ENV_HEADER_TAGS, nil, comma_separated_only: true) }
+                o.setter { |header_tags, _| Configuration::HTTP::HeaderTags.new(header_tags) }
+              end
```
GustavoCaso, line 183: If you rebase master on it. We can use the latest configuration options
```suggestion
              option :header_tags do |o|
                o.env Configuration::Ext::ENV_HEADER_TAGS
                o.type :array
                o.default []
                o.setter { |header_tags, _| Configuration::HTTP::HeaderTags.new(header_tags) }
              end
```

If you rebase or merge master, you can use the latest configuration options.

lib/datadog/tracing/contrib/rack/header_tagging.rb
```diff
@@ -0,0 +1,53 @@
...
+    module Contrib
+      module Rack
+        # Matches Rack-style headers with a matcher and sets matching headers into a span.
+        module HeaderTagging
+          def self.tag_request_headers(span, env, configuration)
+            # Wrap env in a case-insensitive Rack-style accessor.
+            headers = env.is_a?(Header::RequestHeaderCollection) ? env : Header::RequestHeaderCollection.new(env)
+
+            # Use global DD_TRACE_HEADER_TAGS if integration-level configuration is not provided
+            tags = if configuration.using_default?(:headers) && Datadog.configuration.tracing.header_tags.configured?
```
GustavoCaso, line 14: If you apply this comment https://github.com/DataDog/dd-trace-rb/pull/2935/files#r1270699898

You could change the code to: 
```suggestion
            tags = if configuration.using_default?(:headers) && !Datadog.configuration.tracing.using_default(:header_tags)
```

And we can get rid of the `configured?` method all together

marcotc:
Opened a system-tests PR for this issue^: https://github.com/DataDog/system-tests/pull/1430

lib/datadog/tracing/contrib/rack/header_tagging.rb
marcotc: Let's try it!

lib/datadog/tracing/configuration/http.rb
```diff
@@ -0,0 +1,78 @@
...
+        # allow for ease of utilization by each HTTP integration.
+        #
+        # This class process configuration, stores the result, and provides methods to
+        # utilize this configuration.
+        class HeaderTags
+          # @param header_tags [Array<String>] The list of strings from DD_TRACE_HEADER_TAGS.
+          def initialize(header_tags)
+            @request_headers = {}
+            @response_headers = {}
+            @header_tags = header_tags || EMPTY
```
GustavoCaso, line 19: After https://github.com/DataDog/dd-trace-rb/pull/2935/commits/a01cde6209eeeac9cc592284c879a1785b872708 we can get rid of `EMPTY` now 😄

marcotc: This suggestion kind of worked, but I need to fix a small thing in `Options` for it to work: https://github.com/DataDog/dd-trace-rb/pull/2994

marcotc: I rebased this PR on that fix, but everything here remains unchanged.

TonyCTHsu (approved):

PR #2936, opened 2023-06-29
[EXPERIMENTAL] Add `env_var` and `deprecated_env_var` configuration options

GustavoCaso:
**What does this PR do?**

I'm exploring trying to try solve some problems from a private discussion about how to track the configuration option using the ENV variable or the fallback.

The PR explores adding `env_var` and `deprecatated_env_var` as configuration options. Those options would be used to fetch the value from ENV. If the values are present, we treat them as being set programmatically by the customer. That way, we can later use the `using_default?`  method to know if a configuration is using the default value or the one set by the customer. 

I had to refactor the functions `env_to_*` into `val_to_*`, since they no longer deal with ENV variables. The methods still live under `Core::Environment::VariableHelpers`; that namespace is wrong. 

The intent of the PR is for people to see the approach and comment on anything they would change or miss. 

If people like this addition, I will address all the missing and failing specs and any feedback from comments. Finish adding `o.default` to some settings

**Motivation**

**Additional Notes**

**How to test the change?**

### DIFF
--- a/lib/datadog/core/configuration/settings.rb
+++ b/lib/datadog/core/configuration/settings.rb
@@ -79,7 +79,8 @@ module Datadog
         # @default `DD_API_KEY` environment variable, otherwise `nil`
         # @return [String,nil]
         option :api_key do |o|
-          o.default { ENV.fetch(Core::Environment::Ext::ENV_API_KEY, nil) }
+          o.type :string, nil: true
+          o.env_var Core::Environment::Ext::ENV_API_KEY
         end
 
         # Datadog diagnostic settings.
@@ -98,7 +99,9 @@ module Datadog
           # @default `DD_TRACE_DEBUG` environment variable, otherwise `false`
           # @return [Boolean]
           option :debug do |o|
-            o.default { env_to_bool(Datadog::Core::Configuration::Ext::Diagnostics::ENV_DEBUG_ENABLED, false) }
+            o.env_var Datadog::Core::Configuration::Ext::Diagnostics::ENV_DEBUG_ENABLED
+            o.default false
+            o.type :bool
             o.on_set do |enabled|
               # Enable rich debug print statements.
               # We do not need to unnecessarily load 'pp' unless in debugging mode.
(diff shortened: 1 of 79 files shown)


delner:
`env_var` is great. Might I suggest adding type conversion as well? e.g.

```ruby
env_var 'DD_TRACE_ENABLED', type: :bool
env_var 'DD_TRACE_HEADER_STYLES', type: :array

# Alternatively...

env_var 'DD_TRACE_HEADER_STYLES', type: Array
```

lib/datadog/core/environment/variable_helpers.rb
```diff
@@ -7,54 +7,36 @@ module Environment
       # Defines helper methods for environment
       # @public_api
       module VariableHelpers
```
ivoanjo, line 9: Note that APIs on this file are marked as "@public_api" which means that customers are allowed to use them, and we're not supposed to break them between minor releases.

This doesn't mean you can't apply your changes. One possibility is to make sure that the old versions of the methods are at least still available (and probably marked as deprecated).

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -51,14 +53,13 @@ def self.extended(base)
                 #   otherwise `['Datadog','b3multi','b3']`.
                 # @return [Array<String>]
                 option :propagation_extract_style do |o|
-                  o.default do
+                  o.deprecated_env_var Tracing::Configuration::Ext::Distributed::ENV_PROPAGATION_STYLE_EXTRACT_OLD
```
ivoanjo, line 56: I may have counted them wrong, but it seems to me we're adding `deprecated_env_var` to support only 2 configuration settings -- this one and `ENV_PROPAGATION_STYLE_INJECT_OLD`.

It seems a lot of work for only these two use-cases... Could we perhaps use custom logic on the `setter` or somewhere else to avoid having the configuration system need to support this?

ivoanjo (commented):
Left a few notes :)

Overall I don't have very strong thoughts in terms of this option vs what we had before (e.g. I'm not convinced this one is _better_, but also not convinced it's _worse_, just different).

lib/datadog/core/configuration/settings.rb
```diff
@@ -406,7 +444,8 @@ def initialize(*_)
           o.setter { |v| v.to_s if v }
 
           # NOTE: service also gets set as a side effect of tags. See the WORKAROUND note in #initialize for details.
-          o.default { ENV.fetch(Core::Environment::Ext::ENV_SERVICE, Core::Environment::Ext::FALLBACK_SERVICE_NAME) }
+          o.env_var Core::Environment::Ext::ENV_SERVICE
+          o.default { Core::Environment::Ext::FALLBACK_SERVICE_NAME }
```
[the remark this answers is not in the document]
GustavoCaso: The intent of this PR is not to modify existing default blocks. Once we merge the PR that removes `lazy` as a option, we can have a PR that goes over the defaults values that should not be a block.

GustavoCaso:
Close in favour of https://github.com/DataDog/dd-trace-rb/pull/2970

PR #2937, opened 2023-06-29
[PROF-7440] Bootstrap support for timeline feature in profiler, off by default

ivoanjo:
**What does this PR do?**:

This PR bootstraps support for collecting data to power the upcoming timeline feature. Specifically it:

* Introduces a new `DD_PROFILING_EXPERIMENTAL_TIMELINE_ENABLED` environment variable, as well as a `c.profiling.advanced.experimental_timeline_enabled` setting to turn on collection of this data.
* Changes the `ThreadContext` collector to include timestamps of when cpu/wall-time events happen as a label in profiling samples

**The new changes are off by default**, and are not expected to have any impact on customer applications (hence this is safe to merge into master and release).

**Motivation**:

The key part of the "timeline" feature is to start assigning timestamps to when things happen, rather than a profile being an aggregation of 60 seconds of events, where you can't really tell what came before and what came after.

**Additional Notes**:

At this moment, it's not possible for customers to see this data in the profiling UX. But with the correct set of backend branches and feature flags I was able to see what looks like correct timeline data from this change. The plan is to obviously iterate a lot from here on, to get to the point where we'll be enabling the feature by default.

**How to test the change?**:

The changes to add the new parameter and the timestamps include test coverage. The overall feature needs a few internal things to see working, ping me if you want to try it out.

### DIFF
--- a/benchmarks/profiler_sample_loop_v2.rb
+++ b/benchmarks/profiler_sample_loop_v2.rb
@@ -17,8 +17,9 @@ class ProfilerSampleLoopBenchmark
 
   def create_profiler
     @recorder = Datadog::Profiling::StackRecorder.new(cpu_time_enabled: true, alloc_samples_enabled: true)
-    @collector =
-      Datadog::Profiling::Collectors::ThreadContext.new(recorder: @recorder, max_frames: 400, tracer: nil, endpoint_collection_enabled: false)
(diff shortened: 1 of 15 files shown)


GustavoCaso (approved):
LGTM!. The code is very easy to follow and understand. Thank you for the wonderful comments

PR #2939, opened 2023-06-29
Add fallback name for main thread in profiling data

ivoanjo:
**What does this PR do?**:

This PR adds a fallback name of "main" to the main Ruby thread.

The main Ruby thread is first thread created in the VM, and the one that is returned by `Thread.main` (when in the main Ractor).

**Motivation**:

As more profiling features start to show the name in the UX, it's annoying to often have the main thread not have a name.

**Additional Notes**:

This PR sits atop #2937 to avoid any merge conflicts (since both touch labels and the `ThreadContext` collector state), but is otherwise completely independent of that one.

**How to test the change?**:

You can see it in action by profiling a Ruby app that never names its main thread, and then confirming that profiling data for that thread is shown under the name `main` in the Datadog UX.


GustavoCaso (approved):
🚢

PR #2940, opened 2023-06-29
OpenSearch Integration

### LINKED ISSUE #2628: Support for opensearch gem
**Is your feature request related to a problem? Please describe.**
No
**Describe the goal of the feature**
DataDog can trace the `opensearch` gem out of the box

**Describe alternatives you've considered**
None

**Additional context**

https://github.com/opensearch-project/opensearch-ruby is a fork of elasticsearch-ruby. ddtrace has a tracer built in for elasticsearch, but it is hardcoded to the elasticsearch constants, and therefore does not support the opensearch gem.

sarahchen6:
**What does this PR do?**
This PR implements the OpenSearch integration with dd-trace-rb code by creating and populating two directories: `/lib/datadog/tracing/contrib/opensearch/` and `/spec/datadog/tracing/contrib/opensearch/`.

**Motivation**
OpenSearch has had over 100 million downloads since its fork from Elasticsearch in 2021. It is a widely used project with roughly 500 contributors. Thus, providing OpenSearch integration would benefit a substantial number of users and expand Datadog’s tracing client to offer more valuable trace requests and development visibility.

fixes #2628 

**Additional Notes**
Under Github's checks, there are currently still failures with the Library Injection code.

**How to test the change?**
This code is covered by unit tests located in the `/spec/datadog/tracing/contrib/opensearch/` directory. To test the change on your machine, run `docker compose run --rm tracer-3.1 /bin/bash` in your terminal to enter the docker environment. Once inside, run `bundle exec appraisal ruby-3.1.2-contrib rspec spec/datadog/tracing/contrib/opensearch/patcher_spec.rb` to test the patcher code. Run `bundle exec appraisal ruby-3.1.2-contrib rspec spec/datadog/tracing/contrib/opensearch/quantize_spec.rb` to test the quantization code.

This PR has also been tested by CircleCI.


lib/datadog/tracing/contrib/opensearch/patcher.rb
```diff
@@ -0,0 +1,120 @@
...
+              response = nil
+              Tracing.trace('opensearch.query', service: service) do |span|
+                begin
+                  # set generic tags
+                  span.set_tag(Tracing::Metadata::Ext::TAG_COMPONENT, Ext::TAG_COMPONENT)
+                  span.set_tag(Tracing::Metadata::Ext::TAG_KIND, Tracing::Metadata::Ext::SpanKind::TAG_CLIENT)
+                  span.set_tag(Contrib::Ext::DB::TAG_SYSTEM, Ext::TAG_SYSTEM)
+
+                  # set argument tags
+                  span.set_tag(Datadog::Tracing::Contrib::OpenSearch::Ext::TAG_METHOD, method)
```
marcotc, line 46: minor comment:
```suggestion
                  span.set_tag(OpenSearch::Ext::TAG_METHOD, method)
```

The way Ruby searches for constants (constants are anything starting with an uppercase letter) is by starting from the let-most part (in this case `Datadog`, for `Datadog::Tracing::Contrib::OpenSearch::Ext::TAG_METHOD`), and starts walking up the module/class declaration where this is declared.

In this case, it's:
```
module Datadog
  module Tracing
    module Contrib
      module OpenSearch
        module Patcher
          module Client
```

Ruby will look for `Datadog` (from `Datadog::Tracing::Contrib::OpenSearch::Ext::TAG_METHOD`), in `Client::Datadog` but not find it, then `Patcher::Datadog` but not find it, then `OpenSearch::Datadog` but not find it and so on. It will eventually find `Datadog` as a root module.
When `Datadog` is found, it will then navigate from left to right finding each nested constant from `Datadog::Tracing::Contrib::OpenSearch::Ext::TAG_METHOD`, until it reaches `TAG_METHOD`.

At the end of day, what we really want here is `TAG_METHOD`, we don't care about all the rest of the hierarchy.
Because we only need `TAG_METHOD` here, we can find it easier in the class hierarchy by shortening our constant lookup to: `OpenSearch::Ext::TAG_METHOD`.

This means that Ruby will look for `OpenSearch` up the current module/class declaration, and it will find it here:
```
module Datadog
  module Tracing
    module Contrib
      module OpenSearch # <--
        module Patcher
          module Client
```

Which will translate to `Datadog::Tracing::Contrib::OpenSearch`
Then Ruby will try to follow the constant lookup from left to right, looking up `Ext` inside `Datadog::Tracing::Contrib::OpenSearch`, which is correct, then finding the correct `TAG_METHOD`.

The rule here is to always declare a common ancestor, in this case `OpenSearch` is the common ancestor.

As a side note, one might be tempted to simply declare `TAG_METHOD`, but the issue is that `TAG_METHOD` is not directly nested to any of the scoped modules:
```
module Datadog
  module Tracing
    module Contrib
      module OpenSearch
        module Patcher
          module Client
```
Thus lookup will fail.

`Ext::TAG_METHOD` would actually work here, but leaves it ambiguous where in the hierarchy chain `Ext` was found.

```
module Datadog # was `Ext` found here?
  module Tracing # or here?
    module Contrib # maybe here?
      module OpenSearch # or here?
        module Patcher # here maybe?
          module Client # probably not here?
```

Adding the extra `OpenSearch` to it allows us to easily find `OpenSearch` in the hierarchy and know that's our anchor point.

As a second side note, if we need to refer to a global module, but we have a module with the name name in our current hierarchy, we prefix it with `::`. In our case, simply writing `OpenSearch` will retrieve us `Datadog::Tracing::Contrib::OpenSearch`, not the `OpenSearch` client from `opensearch-ruby`. To get that client, we need to do `::OpenSearch`. This is only the case because are inside `Datadog::Tracing::Contrib::OpenSearch` in this file.

lib/datadog/tracing/contrib/opensearch/patcher.rb
```diff
@@ -0,0 +1,120 @@
...
+                  tag_params(params, span)
+                  tag_body(body, span)
+
+                  # parse url, and remove username/password
+                  original_url = transport.get_connection.full_url(path, {})
+                  url = URI.parse(original_url)
+                  host = url.host
+                  port = url.port
+                  scheme = url.scheme
+                  url.user = nil
```
marcotc, line 58: Document why we are doing `url.user = nil` here.

lib/datadog/tracing/contrib/opensearch/patcher.rb
```diff
@@ -0,0 +1,120 @@
...
+          # Patches OpenSearch::Transport::Client module
+          module Client
+            # rubocop:disable Metrics/MethodLength
+            def perform_request(*args)
+              service = datadog_configuration[:service_name]
+              method = args[0]
+              path = args[1]
+              params = args[2]
+              body = args[3]
+              response = nil
```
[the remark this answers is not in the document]
sarahchen6: Line 73 -- `response = super` fails if `response` is not declared earlier

lib/datadog/tracing/contrib/opensearch/patcher.rb
```diff
@@ -0,0 +1,120 @@
...
+
+          # Patches OpenSearch::Transport::Client module
+          module Client
+            # rubocop:disable Metrics/MethodLength
+            def perform_request(*args)
+              service = datadog_configuration[:service_name]
+              method = args[0]
+              path = args[1]
+              params = args[2]
+              body = args[3]
```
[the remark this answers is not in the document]
sarahchen6: `body` and `params` get overridden in the `tag_params` and `tag_body` methods, so it didn't work to remove all of the parameters back into the `def perform_request` line. I could move them all back to the line and then reset `body` and `params` separately?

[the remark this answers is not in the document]
delner: Reading the code in `tag_params`:

```
params = JSON.generate(params) unless params.is_a?(String)
```

This shouldn't mutate the original params. The arg `params` in this function is a local variable, that points to the params passed in (what we don't want to alter.) However, `JSON.generate` returns a new object, and doing `params=` changes the variable to point at that new object; it does not modify the old `params` object it used to point at. It should be okay.

It would be concerning if we did something like `params.reject!` where a `!` signifies a modification to the object being passed in. There are other functions that can also mutate an object (even without a `!`). It's good that you're being considerate that this is not happening here.

spec/datadog/tracing/contrib/opensearch/patcher_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+
+    # Create a client index before every test case
+    unless client.indices.exists?(index: index_name)
+      client.indices.create(
+        index: index_name,
+        body: index_body
+      )
+    end
+
+    clear_traces!
```
delner, line 21: Interesting, not sure I've seen the need to do this before. Why do we need this?

spec/datadog/tracing/contrib/opensearch/quantize_spec.rb
```diff
@@ -0,0 +1,112 @@
+require 'datadog/tracing/contrib/support/spec_helper'
+require 'opensearch'
+
+require 'ddtrace'
+require 'datadog/tracing/contrib/opensearch/quantize'
+
+RSpec.describe Datadog::Tracing::Contrib::OpenSearch::Quantize do
```
delner, line 7: If we copy over `quantize` code from ElasticSearch, then it makes sense to keep this spec to check its behavior. Otherwise, if we promote this to `Contrib`, this spec should live there instead.

lib/datadog/tracing/contrib/opensearch/ext.rb
```diff
@@ -0,0 +1,32 @@
...
+          TAG_SYSTEM = 'opensearch'
+          TAG_METHOD = 'http.method'
+          TAG_PATH = 'http.url_details.path'
+          TAG_PARAMS = 'opensearch.params'
+          TAG_BODY = 'opensearch.body'
+          TAG_URL = 'http.url'
+          TAG_HOST = 'http.url_details.host'
+          TAG_PORT = 'http.url_details.port'
+          TAG_SCHEME = 'http.url_details.scheme'
+          TAG_RESPONSELENGTH = 'http.response.content_length'
```
delner, line 27: Minor: I think this should be `TAG_RESPONSE_CONTENT_LENGTH` for consistency.

delner (changes requested):
Overall this is shaping up very nicely!

The only significant change I see so far that's an issue is the dependence upon Elasticsearch. We should be able to instrument OpenSearch without importing the instrumentation for Elasticsearch. See my comments about our options.

lib/datadog/tracing/contrib/opensearch/patcher.rb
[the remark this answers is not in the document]
sarahchen6: Ahh I see, thanks!

spec/datadog/tracing/contrib/opensearch/quantize_spec.rb
sarahchen6: Code was copied over from `quantize` in Elasticsearch

lib/datadog/tracing/contrib/opensearch/quantize.rb
```diff
@@ -0,0 +1,31 @@
...
+  module Tracing
+    module Contrib
+      module OpenSearch
+        # Removes sensitive data from OpenSearch strings (i.e. url and body).
+        module Quantize
+          PLACEHOLDER = '/?'
+
+          module_function
+
+          include Contrib::Elasticsearch::Quantize
```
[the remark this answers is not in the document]
sarahchen6: Code has been duplicated in OpenSearch

lib/datadog/tracing/contrib/opensearch/patcher.rb
```diff
@@ -0,0 +1,116 @@
...
+                  tag_params(params, span)
+                  tag_body(body, span)
+
+                  # Parse url
+                  original_url = transport.get_connection.full_url(path, {})
+                  url = URI.parse(original_url)
+                  host = url.host
+                  port = url.port
+                  scheme = url.scheme
+                  # Set url.user to nil to remove sensitive information
```
marcotc, line 53: Can you specify what information is being removed here?

sarahchen6: Changed to "# Set url.user to nil to remove sensitive information (i.e. user's username and password)"!

marcotc (approved):
Awesome work, @sarahchen6! 🙇

docs/GettingStarted.md
```diff
@@ -506,7 +507,8 @@ For a list of available integrations, and their configuration options, please re
 | MongoDB                    | `mongo`                    | `>= 2.1`                 | `>= 2.1`                  | *[Link](#mongodb)*                  | *[Link](https://github.com/mongodb/mongo-ruby-driver)*                         |
 | MySQL2                     | `mysql2`                   | `>= 0.3.21`              | *gem not available*       | *[Link](#mysql2)*                   | *[Link](https://github.com/brianmario/mysql2)*                                 |
 | Net/HTTP                   | `http`                     | *(Any supported Ruby)*   | *(Any supported Ruby)*    | *[Link](#nethttp)*                  | *[Link](https://ruby-doc.org/stdlib-2.4.0/libdoc/net/http/rdoc/Net/HTTP.html)* |
-| Postgres                     | `pg`                   | `>= 0.18.4`              | *gem not available*       | *[Link](#postgres)*                   | *[Link](https://github.com/ged/ruby-pg)*                  |
+| OpenSearch                 | `opensearch-ruby`          | `>= 1.0.0`               | `>= 1.0.0`                | *[Link](#opensearch)*               | *[Link](https://github.com/opensearch-project/opensearch-ruby)*                        |
```
delner, line 510: Very minor: align the `|` with the rest of the table?

docs/GettingStarted.md
```diff
@@ -1433,6 +1435,35 @@ If you wish to configure each connection object individually, you may use the `D
 client = Net::HTTP.new(host, port)
 Datadog.configure_onto(client, **options)
 ```
+
+### OpenSearch
```
delner, line 1439: Docs look good! :+1:

lib/datadog/tracing/contrib/opensearch/configuration/settings.rb
```diff
@@ -0,0 +1,45 @@
...
+            option :analytics_sample_rate do |o|
+              o.default { env_to_float(Ext::ENV_ANALYTICS_SAMPLE_RATE, 1.0) }
+              o.lazy
+            end
+
+            option :quantize, default: {}
+
+            option :service_name do |o|
+              o.default do
+                Contrib::SpanAttributeSchema.fetch_service_name(
```
delner, line 33: Curious to see this in here. My understanding is this (`SpanAttributeSchema`) was still on a feature branch? Or is this really in `master`? cc @zarirhamza

lib/datadog/tracing/contrib/opensearch/patcher.rb
```diff
@@ -0,0 +1,116 @@
...
+
+            def tag_params(params, span)
+              if params
+                params = JSON.generate(params) unless params.is_a?(String)
+                span.set_tag(OpenSearch::Ext::TAG_PARAMS, params)
+              end
+            end
+
+            def tag_body(body, span)
+              if body
```
delner, line 96: Minor style: You could put `if body` as `return unless body` and remove the indentation here. Same applies to `tag_params`.

delner (commented):
Some minor suggestions, but otherwise looks great. Only thing I need to double check is the use of `SpanAttributeSchema`.... although this should happen soon, I didn't think this was yet merged to `master`. Want to check with @zarirhamza on this.

docs/GettingStarted.md
sarahchen6: fixed all the `|`s to align

lib/datadog/tracing/contrib/opensearch/configuration/settings.rb
delner: Answering my question, this looks like it was a prototype of sorts merged months ago.

delner (approved):
:tada:

PR #2941, opened 2023-06-29
Add gitlab pipeline for host injection

TonyCTHsu:
**What does this PR do?**

Add gitlab pipeline for host injection


.gitlab/build-deb-rpm.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+if [ -n "$CI_COMMIT_TAG" ] && [ -z "$RUBY_PACKAGE_VERSION" ]; then
+  RUBY_PACKAGE_VERSION=${CI_COMMIT_TAG##v}
+fi
+
+source common_build_functions.sh
+
+# Not sure how it is called
```
marcotc, line 9: I don't think I understand what this comment means here.

.gitlab/build-deb-rpm.sh
```diff
@@ -0,0 +1,17 @@
...
+fpm_wrapper "datadog-apm-library-ruby" "$RUBY_PACKAGE_VERSION" \
```
marcotc, line 10: Where does the `fpm_wrapper` command come from?
I ask because googling this gave me like no results: https://www.google.com/search?q=%22fpm_wrapper%22

.gitlab/install_ddtrace_deps.rb
```diff
@@ -0,0 +1,78 @@
+# require "fileutils"
+require "open3"
+require "bundler"
+
+lock_file_path = "./Gemfile.lock"
+install_dir = "./vendor"
+
+# lock_file_path = "/opt/datadog/apm/library/ruby/Gemfile.lock"
+# install_dir = "./vendor"
```
marcotc, line 9: Is this commented out code helpful?
If so, we should document when to uncomment it.

.gitlab/install_ddtrace_deps.rb
```diff
@@ -0,0 +1,78 @@
...
+
+lock_file_parser = Bundler::LockfileParser.new(Bundler.read_file(lock_file_path))
+
+gem_version_mapping = lock_file_parser.specs.each_with_object({}) do |spec, hash|
+  hash[spec.name] = spec.version.to_s
+  hash
+end
+
+def install_ddtrace_deps_without_native_extensions(gem_version_mapping: ,location: )
```
marcotc, line 18: I think this method name is misleading, `install_ddtrace_deps_without_native_extensions`, because `libdatadog` and `libddwaf` **do have** native extensions, but they are already precompiled for the user:
Screenshot 2023-07-04 at 10 33 04 AM.

I think a better name would be something like: `install_ddtrace_deps_without_local_compilation`.

Same for `install_ddtrace_deps_with_native_extensions`: `install_ddtrace_deps_with_local_compilation`.

marcotc: Same kind of language should be applied for `INSTALL_DDTRACE_NON_NATIVE_DEPS` as well.

lib-injection/host_inject.rb
```diff
@@ -0,0 +1,97 @@
+return if ENV["DD_TRACE_SKIP_LIB_INJECTION"] == "true"
+
+Gem.paths = {
+  "GEM_PATH" => "/opt/datadog/apm/library/ruby/#{RUBY_VERSION.split(".")[0..1].join(".")}:/opt/datadog/apm/library/ruby:#{ENV["GEM_PATH"]}"
```
marcotc, line 4: What is this parting doing `RUBY_VERSION.split(".")[0..1].join(".")`?

Could you document it or extract it to a variable to make it easier to read?

lib-injection/host_inject.rb
```diff
@@ -0,0 +1,97 @@
...
+}
+
+ENV["GEM_PATH"] = Gem.path.join(":")
```
marcotc, line 7: This is a hard change to the user's environment variable, is this intended like this?
If so, can we document what exactly we want to accomplish in this line? I ask because overriding the GEM_PATH will have serious side effects, so other might question this same line in the future.

lib-injection/host_inject.rb
```diff
@@ -0,0 +1,97 @@
...
+  require "open3"
+  require "bundler"
+  require "shellwords"
+  require "fileutils"
+
+  failure_prefix = "Datadog lib injection failed:"
+  support_message = "For help solving this issue, please contact Datadog support at https://docs.datadoghq.com/help/."
+
+  unless Bundler::SharedHelpers.in_bundle?
+    STDOUT.puts "[ddtrace] Not in bundle... skipping host injection" if ENV["DD_TRACE_DEBUG"] == "true"
```
marcotc, line 19:
```suggestion
    debug_log "[ddtrace] Not in bundle... skipping host injection"
```

I recommend extracting debug logging to a method, as it is used multiple times in this file.

lib-injection/host_inject.rb
```diff
@@ -0,0 +1,97 @@
...
+    return
+  end
+
+  if Bundler.frozen_bundle?
+    warn "[ddtrace] #{failure_prefix} Cannot inject with frozen Gemfile, run `bundle config unset deployment` to allow lib injection. To learn more about bundler deployment, check https://bundler.io/guides/deploying.html#deploying-your-application. #{support_message}"
```
marcotc, line 24: Should we also suggest that he user goes through our regular installation process, in case they want to keep `deployment` mode intact?

Something like: "run `bundle config unset deployment` or add `gem 'ddtrace', require: 'ddtrace/auto_instrument'` to your `Gemfile`"

lib-injection/host_inject.rb
```diff
@@ -0,0 +1,97 @@
+return if ENV["DD_TRACE_SKIP_LIB_INJECTION"] == "true"
```
ivoanjo, line 1: This file seems like it copies a lot over from `auto_inject.rb`. Which in this case... may be easier than shipping multiple files.

Nevertheless, I suggest trying to refactor the common bits so that we could keep them in-sync by a simple copy/paste, as right now they're in a similar-but-not-the-same situation, which is a slightly dangerous middle ground to be in, because we can easily introduce bugs when trying to keep them in-sync, or forget to update one of them.

.gitlab/build-deb-rpm.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
```
ivoanjo, line 1: Is the intention here to build a `rpm` (or `deb`) with the gems already installed and the dependencies pre-compiled?

Because that usually means that we need to have variants for every Ruby supported, it's usually problematic to pick a native extension that was compiled for some Ruby version and then use it a different Ruby version.

.gitlab-ci.yml
```diff
@@ -1,22 +1,74 @@
 stages:
```
ivoanjo, line 1: I wish we had a bit more documentation about how this change works and how it's getting used.

As someone working in the codebase, I have no idea why we need `host_inject.rb` and why it's different from `auto_inject.rb`, how and when we're releasing things as rpm/deb, and how to support customers than run into trouble with these.

marcotc: A `lib-injection/README.md` could help here^.

.gitlab-ci.yml
```diff
@@ -1,22 +1,119 @@
...
+      - vendor
+
+# Hack, remove later
+.package:
+  before_script:
+    - mkdir -p packaging
+    - cd packaging
+    - if [ $CI_PROJECT_NAME != "auto_inject" ]; then cp /packaging/* .; fi
+    - if [ -z "$CI_COMMIT_TAG" ]; then export CI_VERSION_SUFFIX="~pipeline.${CI_PIPELINE_ID}.beta.${CI_COMMIT_SHORT_SHA}"; fi
+
```
randomanderson, line 101:
```suggestion
```

This can be removed. Package signing should work

.gitlab/build-deb-rpm.sh
randomanderson: It comes from: https://github.com/DataDog/auto_inject/blob/9e0dc0f689b5acb49a1ffee53081abe994754421/packaging/common_build_functions.sh#L7

lib-injection/README.md
```diff
@@ -0,0 +1,30 @@
+# Host injection
+
+Our gitlab build pipeline ships a pre-installed tracer deb/rpm package.
```
marcotc, line 3:
```suggestion
Our gitlab build pipeline ships a pre-installed `ddtrace` deb/rpm package.
```

Is it only tracer, or all of `ddtrace` that is shipped?

marcotc, line 3:
```suggestion
The gitlab build pipeline ships pre-installed tracer deb and rpm packages.
```

marcotc: Do we ship rpm too, or only deb?

lib-injection/README.md
```diff
@@ -0,0 +1,30 @@
...
+| OS    | Debian 10+ |
+| Arch  | `amd64` |
+
+
+### Packaging
+
+In order to ship `ddtrace` and its dependencies as a pre-install package, we need a few tweaks in our build pipeline.
+
+* Use multiple custom built Ruby images to build native extensions. Those images are based on Debian `buster` to support older distribution and Ruby is compiled as a static library with `--disbale-shared` option which disables the creation of shared libraries (also known as dynamic libraries or DLLs).
+* Install `ffi` gem with its built-in `libffi` instead of system `libffi`.
```
marcotc, line 19:
```suggestion
* Install `ffi` gem with its built-in `libffi` native extension, instead of using system's `libffi`.
```

lib-injection/README.md
```diff
@@ -0,0 +1,30 @@
...
+* After installation, rename directories to remove the `-static` suffix so user’s ruby can detect those  `.so` files and make sure files have read permission.
```
marcotc, line 20: Who creates `-static` directories?
Is it a specific part of `ddtrace`?
If so, can we add it here?

lib-injection/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+### Injection
+
+The host inject script would add `ddtrace` to your Ruby on Rails application's `Gemfile` to [instrument your application](https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/ruby/#rails-or-hanami-applications)).
+
+* Bundler version must be >= 2.3
+* Does not support vendoring gem (Bundler's deployment mode or `BUNDLE_PATH`)
```
marcotc, line 27:
```suggestion
* Does not support vendoring gem ([Bundler's Deployment Mode](https://www.bundler.cn/man/bundle-install.1.html#DEPLOYMENT-MODE) or `BUNDLE_PATH`)
```

lib-injection/host_inject.rb
```diff
@@ -0,0 +1,113 @@
...
+require "rubygems"
+
+# This would return MAJOR.MINOR without PATCH version, for example 3.1, 3.2
+ruby_version = RUBY_VERSION.split(".")[0..1].join(".")
```
marcotc, line 5:
```suggestion
ruby_version = RbConfig::CONFIG["ruby_version"]
```

I recommend using the same logic that bundler uses to resolve this partial ruby version: https://github.com/rubygems/rubygems/blob/be63f08d78b0fd9cf85cd8aa68aab9b850177069/lib/rubygems/defaults.rb#L38

lib-injection/host_inject.rb
```diff
@@ -0,0 +1,113 @@
...
+
+# Look for pre-installed tracers
+Gem.paths = {
+  "GEM_PATH" => "/opt/datadog/apm/library/ruby/#{ruby_version}:/opt/datadog/apm/library/ruby:#{ENV["GEM_PATH"]}"
+}
+
+ENV["GEM_PATH"] = Gem.path.join(":")
```
marcotc, line 12:
```suggestion
# Also apply to the environment variable, to guarantee any spawned processes will respected the modified `GEM_PATH`.
ENV["GEM_PATH"] = Gem.path.join(":")
```

randomanderson (approved):
As far as `.gitlab-ci.yml` and the bash scripts are concerned, LGTM. I can't comment on the ruby code

lib-injection/README.md
TonyCTHsu: When `gem install msgpack` with Ruby compiled with `--disabled-shared`, the native extension would be stored at `extensions/x86_64-linux/3.2.0-static/` instead of `extensions/x86_64-linux/3.2.0/`. 

This would be an issue for other Ruby installation to lookup for native extension. Eventually, it would raise error about could not find native extension but gem has been installed.

.gitlab/build-deb-rpm.sh
```diff
@@ -0,0 +1,37 @@
...
+  exit 1
+fi
+
+# Loop through the directories
+for dir in "${directories[@]}"; do
+  # Remove the "./" prefix and the "-static" suffix from the directory name
+  new_dir="${dir#./}"
+  new_dir="${new_dir%-static}"
+  # Rename the directory
+  mv "$dir" "$new_dir"
```
ivoanjo, line 24: Minor: I wonder if instead of moving, we could symlink them instead. That seems like it may make the gems work with both `-static` and non-static Rubies...?

lib-injection/README.md
```diff
@@ -0,0 +1,30 @@
+# Host injection
+
+Our gitlab build pipeline ships a pre-installed tracer deb/rpm package.
+
+Currently, we support
+
+| Environment| version |
+|---|---|
+| Ruby  | `2.7`, `3.0`, `3.1`, `3.2`|
+| OS    | Debian 10+ |
```
ivoanjo, line 10: Are we still building `.rpm`? Because Debian doesn't use `rpm`, so it looks like this list is out-of-sync with what we're actually generating?

.gitlab-ci.yml
```diff
@@ -1,22 +1,110 @@
...
+    - if: $CI_PIPELINE_SOURCE == "merge_request_event" || $CI_PIPELINE_SOURCE == "push"
+      changes:
+        - .gitlab/Dockerfile-*
+      when: manual
+      allow_failure: true
+  tags: [ "runner:docker" ]
+  image: $DOCKER_REGISTRY/docker:20.10.13
+  parallel:
+    matrix:
+      - RUBY_VERSION: [ '3.2.2', '3.1.4', '3.0.6', '2.7.8' ]
```
ivoanjo, line 39: I suggest adding `# ADD NEW RUBIES HERE` to places in this file which we'll need to update when newer Rubies come out. We already have this comment in other places in the codebase and it helps us not forget to add new Rubies when there's a new release out :)

marcotc (approved):

.gitlab-ci.yml
```diff
@@ -1,22 +1,110 @@
...
+    - .gitlab/install-gem-to-folder.sh
+  artifacts:
+    paths:
+      - vendor
+
+install-version-dependant-gems:
+  image: $RUBY_CUSTOM_IMAGE_BASE/$RUBY_VERSION:current
+  parallel:
+    matrix:
+      - RUBY_VERSION: ["2.7.8", "3.0.6", "3.1.4"]
```
ivoanjo, line 79: :thinking: Are we missing 3.2.2 here?

.gitlab/Dockerfile-2.7.8
```diff
@@ -0,0 +1,95 @@
+#
+# NOTE: THIS DOCKERFILE IS GENERATED VIA "apply-templates.sh"
```
ivoanjo, line 2: Hey, I don't see this `apply-templates.sh` anywhere -- mind adding a note/link to where it can be found?

.gitlab/install_ddtrace_deps.rb
```diff
@@ -0,0 +1,60 @@
...
+  if status.success?
+    next
+  else
+    exit 1
+  end
+end
+
+FileUtils.cd(versioned_path.join("extensions/#{Gem::Platform.local.to_s}"), verbose: true) do
+  FileUtils.ln_sf Gem.extension_api_version, ruby_api_version
+end
```
ivoanjo, line 60: Maybe leave here a comment explaining why this is needed? It's very subtle :)

lib-injection/README.md
```diff
@@ -0,0 +1,30 @@
...
+| Arch  | `amd64` |
+
+
+### Packaging
+
+In order to ship `ddtrace` and its dependencies as a pre-install package, we need a few tweaks in our build pipeline.
+
+* Use multiple custom built Ruby images to build native extensions. Those images are based on Debian `buster` to support older distribution and Ruby is compiled as a static library with `--disbale-shared` option which disables the creation of shared libraries (also known as dynamic libraries or DLLs).
```
ivoanjo, line 18: Minor:

```suggestion
* Use multiple custom built Ruby images to build native extensions. Those images are based on Debian `buster` to support older distribution and Ruby is compiled as a static library with `--disable-shared` option which disables the creation of shared libraries (also known as dynamic libraries or DLLs).
```

lib-injection/host_inject.rb
```diff
@@ -0,0 +1,116 @@
...
+
+  unless Bundler::SharedHelpers.in_bundle?
+    debug_log "[ddtrace] Not in bundle... skipping host injection"
+    return
+  end
+
+  unless Bundler::CLI.commands["add"] && Bundler::CLI.commands["add"].options.key?("require")
+    debug_log "[ddtrace] You are currently using Bundler version #{Bundler::VERSION} which is not supported by host injection. Please upgrade >= 2.3, check https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md#enhancements-31"
+    return
+  end
```
ivoanjo, line 37: I'm not sure the link to the changelog adds a lot -- e.g. what do we want customers to learn from that link?

ivoanjo (commented):
Left a few notes!

.gitlab-ci.yml
ivoanjo: (If not, may be worth leaving a comment explaining why?)

.gitlab/Dockerfile-2.7.8
```diff
@@ -0,0 +1,91 @@
+# This is copied from official Docker image, but compile Ruby with `--disable-shared` instead.
```
ivoanjo, line 1: May be worth leaving a link here to the upstream file, so we can pick up changes in the future?

ivoanjo (commented):
I gave a final after-merging pass, and added a couple of comments :)

PR #2942, opened 2023-06-30
Fix environment variable mock in test

TonyCTHsu:
**What does this PR do?**

**Motivation**

**Additional Notes**

**How to test the change?**

### DIFF
--- a/spec/datadog/tracing/contrib/rails/support/base.rb
+++ b/spec/datadog/tracing/contrib/rails/support/base.rb
@@ -6,9 +6,6 @@ if ENV['USE_SIDEKIQ']
   require 'datadog/tracing/contrib/sidekiq/server_tracer'
 end
 
-require 'lograge' if ENV['USE_LOGRAGE'] == true
-require 'rails_semantic_logger' if ENV['USE_SEMANTIC_LOGGER'] == true
-
 RSpec.shared_context 'Rails base application' do
   if Rails.version >= '6.0'
     require 'datadog/tracing/contrib/rails/support/rails6'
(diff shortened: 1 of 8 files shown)


spec/datadog/tracing/contrib/rails/support/base.rb
```diff
@@ -6,9 +6,6 @@
   require 'datadog/tracing/contrib/sidekiq/server_tracer'
 end
 
-require 'lograge' if ENV['USE_LOGRAGE'] == true
-require 'rails_semantic_logger' if ENV['USE_SEMANTIC_LOGGER'] == true
```
ivoanjo, line 10: I'm curious -- why are these requires no longer needed?

I'm asking because I remember that in the past a few things "auto-probed" what was enabled, and thus it looked like the tests were actually testing what we wanted but since something wasn't loaded it wasn't actually being tested.

spec/datadog/tracing/contrib/rails/support/rails3.rb
```diff
@@ -1,7 +1,5 @@
 require 'rails/all'
 
-require 'ddtrace' if ENV['TEST_AUTO_INSTRUMENT'] == true
-
 if ENV['USE_SIDEKIQ']
```
ivoanjo, line 3: Should `USE_SIDEKIQ` be cleaned up as well? Right now it's kinda weird that it accepts anything, e.c. `USE_SIDEKIQ=false` will enabled it.

marcotc: That seems like a reasonable ask.

PR #2943, opened 2023-07-03
Implement `peer.service`

zarirhamza:
Implements the `set_peer_service!` function in all target integrations. These are integrations that produce spans that have `span.kind` values of either `client` or `producer`. The checklist to show the progress for integrations to be implemented are:

- [x] AWS
- [x] Dalli
- [x] ~~Delayed_Job~~ (Needs precursor tags to be implemented first so skipped for now, to be in separate PR)
- [x] Elasticsearch
- [x] Ethon
- [x] Excon
- [x] Faraday
- [x] ~~GRPC~~ (Needs precursor tags to be implemented first so skipped for now, to be in separate PR)
- [x] HTTP
- [x] Httpclient
- [x] Httprb
- [x] ~~Kafka~~ (Needs precursor tags to be implemented first so skipped for now, to be in separate PR)
- [x] MongoDB
- [x] Mysql2
- [x] Opensearch
- [x] PG
- [x] Presto
- [x] Redis
- [x] Rest_client
- [x] ~~Sidekiq~~ (Needs precursor tags to be implemented first so skipped for now, to be in separate PR)

Please note that few libraries will be implemented in the future upon implementation of their own library specific precursor tags

We also remove the old integrations that had `peer.service` and `peer.hostname` but no longer fall underneath the categorization of a `peer.service` producing integration by not being an integration that produces spans with `span.kind` of value `client` or `producer`

- [x] Active_record (hostname values are duplicated in other tags, so simple removal of `peer.*` tags)
- [x] Active_support (only `peer.service` was set with no testing, so removed)
- [x] Sequel (replaced `peer.hostname` with proper `network.destination.host` tag to keep information available in span)
- [x] Racecar (only `peer.service` was set with no testing, so removed)

Please do not merge without direct approval from @delner, but feel free for anyone to take a look at it


delner (approved):
@zarirhamza and I review this together. Changes overall make sense, and I think it's acceptable to release as it is right now. However, it comes with some caveats which we need to address.

- We removed `peer.service` from certain ORMs, with the understanding that they are not "leaf" or "client" spans (last span before an external service.) I believe this is correct in the long term (ORMs are internal), but its worth calling out that, *as long as they have the potential to be a "leaf" span, we must provide default behavior that renames their `service` to something else*. (Which is currently the case today.)
   - This is because if these ORMs reflect `service` to match `DD_SERVICE` by default (as they probably should), then the "Time Spent %" graphs will incorrectly attribute DB time to the app.
- `Contrib::SpanAttributeSchema.set_peer_service!` should probably not pass a `span` as an argument, because it obfuscates its dependency on the state of the `span`. Specifically, if the span did not set certain tags before calling `set_peer_service!`, then it will not behave correctly. It's not unreasonable to anticipate this function call might be refactored, and moved some place (before other tags) where it would malfunction. This is a form of coupling.
   - This is not a blocker to merge, but we should refactor this ASAP to proactively address the issue. @zarirhamza has agreed we will undertake this after we release this feature.

delner:
@zarirhamza I suggest rebasing this on the latest master before merging.

PR #2946, opened 2023-07-04
Add:HTTP header tagging with DD_TRACE_HEADER_TAGS for clients

marcotc:
_I recommend reviewing #2935 before this one, as most of the context is in there._

**What does this PR do?**

As a follow up to #2935, this PR adds the ability for clients to save the value of HTTP clients in their application's HTTP clients as span tags by configuring the `DD_TRACE_HEADER_TAGS` environment variable.

**Motivation**

This PR brings HTTP clients on par with HTTP servers, which are configurable through `DD_TRACE_HEADER_TAGS` since #2935.

Also, as soon as Remote Configure is available for Ruby Tracing, this configuration will be configurable remotely.

**Additional Notes**

**How to test the change?**

The test suite of each affected HTTP client has been modified to cover this new use case.
Running their test suites cover all changes in this PR.


lib/datadog/tracing/contrib/ethon/easy_patch.rb
```diff
@@ -168,6 +183,26 @@ def analytics_enabled?
...
               datadog_configuration[:analytics_sample_rate]
             end
+
+            # `#response_headers` returns a "\n" concatenated String containing:
+            # * The HTTP Status-Line.
+            # * The response headers.
+            # * A trailing "\n".
+            #
+            # This method extracts only the headers from it.
+            def parse_response_headers
```
TonyCTHsu, line 193: Is this method tested somewhere? 

Our ethon integration test seems to provide response_headers as a hash, which probably is not covering the implementation of this method.

marcotc: This method is tested in `spec/datadog/tracing/contrib/ethon/integration_test_spec.rb`.

There was actually no test for `response_headers` itself, this is the first time we are testing real header usage for Ethon.

TonyCTHsu (approved):

lib/datadog/tracing/contrib/ethon/easy_patch.rb
```diff
@@ -58,6 +59,12 @@ def complete
                     set_span_error_message("Request has failed with HTTP error: #{response_code}")
                   end
                 end
+
+                @datadog_span.set_tags(
+                  Datadog.configuration.tracing.header_tags.response_tags(
+                    Core::Utils::Hash::CaseInsensitiveWrapper.new(parse_response_headers)
+                  )
+                )
```
GustavoCaso, line 67: If `header_tags` is not configured, this would be a no-op, right?

GustavoCaso: Same for the `request_tags`

marcotc: That is correct, both methods return an empty list, and `#set_tags` naturally does nothing with the empty list.

GustavoCaso (approved):
code LGTM. We need to make sure to fix system test failure on https://github.com/DataDog/dd-trace-rb/pull/2935 before merging both in master

marcotc:
@GustavoCaso @lloeki, regrading the system-test failure in this PR, the issue is that we perform HTTP header tagging for `sinatra.request` spans, but that is not a service entry span in the Weblog test variant. Service entry spans get tagged by AppSec with extra HTTP headers, https://github.com/DataDog/dd-trace-rb/blob/c0fc32eb440b6bd1342b72d689207d657b3a2272/lib/datadog/appsec/event.rb#L64, but `sinatra.request` does not receive these tags.

And because Weblog variants have the environment variable `DD_TRACE_HEADER_TAGS=user-agent`, we are tagging   `http.request.headers.user-agent => Arachni/v1 rid/OAOALNJYOEAONLTFJKDKPHVCPORINRTNRCOO` in the `sinatra.request`, as well as in the top-level `rack.request` span.

The issue happens because [`get_rid_from_span`](https://github.com/DataDog/system-tests/blob/072b94247b4179d589a7bddf058b25087c526c59/utils/interfaces/_core.py#L182) successfully returns the rid (request id) for this test for the `sinatra.request`. These means that `sinatra.request` is now selected (alongside the top-level `rack.request` span) as relevant for test assertion.

This causes the test assertion to fail because the remaining expected headers (`"content-type", "content-length", "content-language"`), at https://github.com/DataDog/system-tests/blob/072b94247b4179d589a7bddf058b25087c526c59/tests/appsec/test_traces.py#L354, are not present in `sinatra.request` because AppSec does not tag non-service entry spans.

I'm not too sure how this was working before, but I don't see where we could be setting all the 3 tags in all spans generated by [this Weblog test run](https://github.com/DataDog/system-tests/blob/072b94247b4179d589a7bddf058b25087c526c59/tests/appsec/test_traces.py#L354).

For reference, this is the stdout with the spans that are failing for this test (test name `Test_CollectRespondHeaders`):
```[
 Name: sinatra.route
Span ID: 3547715443187073528
Parent ID: 991106531156891986
Trace ID: 556950475597441551
Type: web
Service: weblog
Resource: GET /headers
Error: 0
Start: 1689891643793740032
End: 1689891643793774080
Duration: 3.370000001723383e-05
Tags: [
   key1 => val1,
   key2  =>  val2,
   env => system-tests,
   version => 1.0.0,
   sinatra.app.name => Sinatra::Application,
   sinatra.route.path => /headers,
   component => sinatra,
   operation => route]
Metrics: [ _dd.measured => 1.0],
 
 Name: sinatra.request
Span ID: 991106531156891986
Parent ID: 2903024419441692001
Trace ID: 556950475597441551
Type: web
Service: weblog
Resource: GET /headers
Error: 0
Start: 1689891643792504832
End: 1689891643793895936
Duration: 0.0013905150000255162
Tags: [
   key1 => val1,
   key2  =>  val2,
   env => system-tests,
   version => 1.0.0,
   http.request.headers.user-agent => Arachni/v1 rid/OAOALNJYOEAONLTFJKDKPHVCPORINRTNRCOO,
   component => sinatra,
   operation => request,
   http.url => /headers,
   http.method => GET,
   sinatra.route.path => /headers,
   http.status_code => 200]
Metrics: [ _dd.measured => 1.0],
 
 Name: rack.request
Span ID: 2903024419441692001
Parent ID: 0
Trace ID: 556950475597441551
Type: web
Service: weblog
Resource: GET /headers
Error: 0
Start: 1689891643791098624
End: 1689891643794927616
Duration: 0.003829038999981549
Tags: [
   key1 => val1,
   key2  =>  val2,
   env => system-tests,
   version => 1.0.0,
   _dd.runtime_family => ruby,
   _dd.appsec.waf.version => 1.9.0,
   http.client_ip => 172.18.0.1,
   _dd.appsec.event_rules.version => 1.7.0,
   appsec.event => true,
   _dd.appsec.json => {"triggers":[{"rule":{"id":"ua0-600-12x","name":"Arachni","tags":{"type":"security_scanner","category":"attack_attempt"}},"rule_matches":[{"operator":"match_regex","operator_value":"^Arachni\\/v","parameters":[{"address":"server.request.headers.no_cookies","key_path":["user-agent"],"value":"Arachni/v1 rid/OAOALNJYOEAONLTFJKDKPHVCPORINRTNRCOO","highlight":["Arachni/v"]}]}]}]},
   http.request.headers.host => localhost:7777,
   http.request.headers.accept-encoding => identity,
   http.request.headers.user-agent => Arachni/v1 rid/OAOALNJYOEAONLTFJKDKPHVCPORINRTNRCOO,
   http.host => localhost,
   http.useragent => Arachni/v1 rid/OAOALNJYOEAONLTFJKDKPHVCPORINRTNRCOO,
   network.client.ip => 172.18.0.1,
   http.response.headers.content-type => text/plain,
   http.response.headers.content-language => en-US,
   http.response.headers.content-length => 15,
   _dd.origin => appsec,
   component => rack,
   operation => request,
   span.kind => server,
   http.method => GET,
   http.url => /headers,
   http.base_url => http://localhost:7777,
   http.status_code => 200]
Metrics: [
   _dd.appsec.enabled => 1.0,
   _dd.appsec.waf.timeouts => 0.0,
   _dd.appsec.waf.duration => 250.304,
   _dd.appsec.waf.duration_ext => 520.405,
   _dd.measured => 1.0,
   _dd.top_level => 1.0]]
```

# A few solutions

1. Remove HTTP header tagging from `sinatra.request` spans, but Sinatra currently supports `:headers` configuration, and `DD_TRACE_HEADER_TAGS` is meant to control the `:headers` field (when an integration-specific `:headers` value is not provided).
2. Tweak the system test to skip `sinatra.request` checking.
3. Tweak weblog to make `sinatra.request` a service-entry span by changing the sinatra integration `service_name`.

None seem too clean, to be honest.

GustavoCaso:
@marcotc Thanks so much for the detailed information on the current issue with the system test

I have a few questions: 

> Tweak weblog to make sinatra.request a service-entry span by changing the sinatra integration service_name.

How would that solve the issue? What is the underlying thing that happens underneath would make the test pass? 

> Tweak the system test to skip sinatra.request checking.

It might be risky, since this is affecting all languages, but it might be worth a try. Since we want the span with `_dd.top_level` and the right `user-agent` tag. What do you think? 

Alternatively, if we see ourselves blocked by this, we could skip the test until we have a better solution. 
`@bug("sinatra" in context.weblog_variant, reason="wrong span selected for assertions")`

marcotc:
Opened a system-tests PR for this issue^: https://github.com/DataDog/system-tests/pull/1430

PR #2949, opened 2023-07-05
fix flaking elasticsearch contrib test flakiness

wconti27:
fix test flakiness by reverting to not sending traces to apm-test-agent


marcotc (approved):
I'm happy to unblock us, given this issue is affecting every single CI run.

PR #2950, opened 2023-07-05
[PROF-7440] Use invoke location as a fallback for nameless threads in the profiler

ivoanjo:
**What does this PR do?**:

Naming Ruby threads is optional, and a lot of Ruby code still creates threads and doesn't name them.

As a workaround for Ruby threads that have no name, this PR makes the profiler use the "thread invoke location", which is what Ruby shows in the `to_s`/`inspect` of a thread, for instance:

```
[1] pry(main)> Thread.list.last.name
=> nil
[2] pry(main)> Thread.list.last.to_s
=> "#<Thread:0x00005934cc0143e8 thread-name.rb:3 sleep>"
```

The "invoke location" is the file and line where the block that starts the thread is defined (`thread-name.rb:3` for the example above).

**Motivation**:

Having nameless threads is annoying because we show this information in several places of our UX, and with the timeline feature this will be even more visible.

**Additional Notes**:

Since Ruby doesn't have an API to get this information (other than calling `to_s` and then parsing its output), and because we want to avoid allocating any Ruby objects while the profiler is taking a sample, I chose to add to `private_vm_api_access.c` an API to get this information that is basically a heavily simplified version of `Thread#to_s`.

**How to test the change?**:

This change includes code coverage. Also, profile any app that starts and doesn't name threads, and you'll start seeing threads named "some_file.rb:some_line" in the profiler.

### DIFF
--- a/ext/ddtrace_profiling_native_extension/private_vm_api_access.c
+++ b/ext/ddtrace_profiling_native_extension/private_vm_api_access.c
@@ -798,14 +798,11 @@ static const rb_iseq_t *maybe_thread_invoke_proc_iseq(VALUE thread_value) {
   return iseq;
 }
 
-VALUE invoke_file_location_for(VALUE thread) {
+VALUE invoke_location_for(VALUE thread, int *line_location) {
   const rb_iseq_t *iseq = maybe_thread_invoke_proc_iseq(thread);
 
-  return iseq != NULL ? rb_iseq_path(iseq) : Qnil;
-}
+  if (iseq == NULL) return Qnil;
 
-VALUE invoke_line_location_for(VALUE thread) {
-  const rb_iseq_t *iseq = maybe_thread_invoke_proc_iseq(thread);
-
-  return iseq != NULL ? rb_iseq_first_lineno(iseq) : Qnil;
+  *line_location = NUM2INT(rb_iseq_first_lineno(iseq));
+  return rb_iseq_path(iseq);
 }
--- a/ext/ddtrace_profiling_native_extension/private_vm_api_access.c
+++ b/ext/ddtrace_profiling_native_extension/private_vm_api_access.c
@@ -776,9 +776,15 @@ check_method_entry(VALUE obj, int can_be_svar)
 static const rb_iseq_t *maybe_thread_invoke_proc_iseq(VALUE thread_value) {
   rb_thread_t *thread = thread_struct_from_object(thread_value);
 
-  if (thread->invoke_type != thread_invoke_type_proc) return NULL;
+  #ifndef NO_THREAD_INVOKE_ARG // Ruby 2.6+
+    if (thread->invoke_type != thread_invoke_type_proc) return NULL;
 
-  VALUE proc = thread->invoke_arg.proc.proc;
+    VALUE proc = thread->invoke_arg.proc.proc;
+  #else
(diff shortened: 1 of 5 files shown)


ext/ddtrace_profiling_native_extension/private_vm_api_access.c
```diff
@@ -767,3 +767,40 @@ check_method_entry(VALUE obj, int can_be_svar)
...
+//
+// It's useful to have here because not all of the methods above are accessible to extensions + to avoid the
+// array allocation that iseq_location did to contain its return value.
+static const rb_iseq_t *maybe_thread_invoke_proc_iseq(VALUE thread_value) {
+  rb_thread_t *thread = thread_struct_from_object(thread_value);
+
+  #ifndef NO_THREAD_INVOKE_ARG // Ruby 2.6+
+    if (thread->invoke_type != thread_invoke_type_proc) return NULL;
+
+    VALUE proc = thread->invoke_arg.proc.proc;
```
marcotc, line 782: proc proc proc

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
```diff
@@ -767,3 +767,40 @@ check_method_entry(VALUE obj, int can_be_svar)
...
+  const rb_iseq_t *iseq = maybe_thread_invoke_proc_iseq(thread);
+
+  return iseq != NULL ? rb_iseq_path(iseq) : Qnil;
+}
+
+VALUE invoke_line_location_for(VALUE thread) {
+  const rb_iseq_t *iseq = maybe_thread_invoke_proc_iseq(thread);
+
+  return iseq != NULL ? rb_iseq_first_lineno(iseq) : Qnil;
+}
```
marcotc, line 806:
```suggestion
VALUE invoke_file_location_for(const rb_iseq_t *iseq) {
  return iseq != NULL ? rb_iseq_path(iseq) : Qnil;
}

VALUE invoke_line_location_for(const rb_iseq_t *iseq) {
  return iseq != NULL ? rb_iseq_first_lineno(iseq) : Qnil;
}
```

Do you think it makes sense to only invoke `maybe_thread_invoke_proc_iseq` once and reuse the value when calling `invoke_file_location_for` and `invoke_line_location_for`?

I understand it exposes `maybe_thread_invoke_proc_iseq` outside of this file, but all the code in question here is internal, and `maybe_thread_invoke_proc_iseq` performs quite a few operations that are being done twice today with no added value.

In either case, the overhead is probably small, but I wanted to chime in because all work here is internal anyway.

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
ivoanjo: proc proc? prooooc!

ext/ddtrace_profiling_native_extension/private_vm_api_access.c
ivoanjo: Agreed, on top of doing the lookup twice, having the two methods made it kinda awkward to the caller as well. I've merged the two functions into one in 87bd64e1f0eea09538ddaf73028bc6104ec73fee, let me know what you think.

marcotc (approved):

PR #2952, opened 2023-07-06
[APPSEC-10361] Appsec automated user events check for UUID in safe mode

GustavoCaso:
**What does this PR do?**

For `safe` automated user tracking events, we want to reduce the possibility of reporting PII information as much as possible. The RFC states that we could report any integer that we know 100% has been autogenerated, we could make the assumption that every ID or UUID coming from Rails is autogenerated, but since we can not be 100% sure, in `safe` mode we only report IDs if they are UUID.

This PR extracts the logic of gathering event information into a new module, `Devise::EventInformation`.

**Motivation**

**Additional Notes**

I realised the devise specs were not running in CI 🤦 

The second commit enables the devise integration specs in CI https://github.com/DataDog/dd-trace-rb/pull/2952/commits/0b9c43cdc31cad5a9446885896dff21be97b4e84. 

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/appsec/contrib/devise/event_information.rb
+++ b/lib/datadog/appsec/contrib/devise/event_information.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+module Datadog
+  module AppSec
+    module Contrib
+      module Devise
+        # Module to extract event information from the resource
+        module EventInformation
+          UUID_REGEX = /^\h{8}-\h{4}-\h{4}-\h{4}-\h{12}$/.freeze
+
+          SAFE_MODE = 'safe'
+          EXTENDED_MODE = 'extended'
+
+          def self.extract(resource, mode)
+            event = {}
+
+            return event unless resource
+
+            resource_id = resource.id
+
+            case mode
+            when EXTENDED_MODE
+              resource_email = resource.email
+              resource_username = resource.username
+
+              event[:id] = resource_id if resource_id
+              event[:email] = resource_email if resource_email
+              event[:username] = resource_username if resource_username
+            when SAFE_MODE
+              event[:id] = resource_id if resource_id && resource_id.to_s =~ UUID_REGEX
+            else
+              Datadog.logger.warn(
+                "Invalid automated user evenst mode: `#{mode}`. "\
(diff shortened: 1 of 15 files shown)


lib/datadog/appsec/contrib/devise/event_information.rb
```diff
@@ -0,0 +1,44 @@
...
+            case mode
+            when EXTENDED_MODE
+              resource_email = resource.email
+              resource_username = resource.username
+
+              event[:id] = resource_id if resource_id
+              event[:email] = resource_email if resource_email
+              event[:username] = resource_username if resource_username
+            when SAFE_MODE
+              event[:id] = resource_id if resource_id && resource_id.to_s =~ UUID_REGEX
```
lloeki, line 30: Certainly not for this PR but there is an opportunity here to define matchers (e.g as procs) and iterate over them, in order to easily extend the list of known safe bits of info.

lloeki: Also, for my understanding, wasn't there a thing specified where known unsafe bits of info would get hashed to pseudonymous information?

lib/datadog/appsec/contrib/devise/event_information.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+module Datadog
+  module AppSec
+    module Contrib
+      module Devise
+        # Module to extract event information from the resource
+        module EventInformation
```
lloeki, line 8: WDYT of the class name simply be `Devise::Event`?

lloeki (approved):
Overall LGTM, left a few notes.

lib/datadog/appsec/contrib/devise/event_information.rb
GustavoCaso: Sounds good

lib/datadog/appsec/contrib/devise/event_information.rb
GustavoCaso: > Also, for my understanding, wasn't there a thing specified where known unsafe bits of info would get hashed to pseudonymous information?

In the end for simplify we decided not to do it on the first iteration

PR #2953, opened 2023-07-06
Change `peer.service` logic

zarirhamza:
Changes existing and future `peer.service` logic to follow the new v1 version instead.

Removes all usage of old `peer.service` values

Updates all tests accordingly


delner (approved):
My understanding is that this PR will change today's `peer.service` tag default behavior (what we refer to as "v0") to behave like "v1" (which will feature a variety of different values based on the kind of span produced.)

This should be OK, as the `peer.service` tag has not been meaningfully employed anywhere in the Datadog UX prior to this point (Ruby was the only one to implement before.)

PR #2956, opened 2023-07-10
Fix profiler libmysqlclient version detection with mysql2-aurora gem

ivoanjo:
**What does this PR do?**:

This PR tweaks the libmysqlclient version detection code added in #2770 to also work when the `mysql2-aurora` gem is in use.

**Motivation**:

The way that the mysql2-aurora gem installs itself leads to the "no signals" workaround (#2873) being incorrectly enabled for customers that do have a modern version of libmysqlclient.

**Additional Notes**:

The mysql2-aurora gem likes to monkey patch itself in replacement of `Mysql2::Client`, and uses `method_missing` to delegate to the original BUT unfortunately does not implement `respond_to_missing?` and thus one of our checks (`respond_to?(:info)`) was incorrectly failing.

**How to test the change?**:

This change includes code coverage. This can also be reproduced easily by adding the `mysql2-aurora` gem to the `Gemfile` and then running a trivial Ruby app:

```
$ DD_PROFILING_ENABLED=true DD_TRACE_DEBUG=true bundle exec ruby -e "require 'mysql2/aurora';
require 'datadog/profiling/preload'"

 # Before

DEBUG -- ddtrace: [ddtrace] Requiring `mysql2` to check if the
`libmysqlclient` version it uses is compatible with profiling
 WARN -- ddtrace: [ddtrace] Enabling the profiling "no signals"
workaround because an incompatible version of the mysql2 gem is
installed. Profiling data will have lower quality. To fix this,
upgrade the libmysqlclient in your OS image to version 8.0.0 or above.

 # After

DEBUG -- ddtrace: [ddtrace] Requiring `mysql2` to check if the
`libmysqlclient` version it uses is compatible with profiling
DEBUG -- ddtrace: [ddtrace] The `mysql2` gem is using a compatible
version of the `libmysqlclient` library (8.0.33)
```

### DIFF
--- a/Steepfile
+++ b/Steepfile
@@ -659,6 +659,7 @@ target :ddtrace do
   library 'google-protobuf'
   library 'protobuf-cucumber'
   library 'mysql2'
+  library 'mysql2-aurora'
   library 'opentracing'
   library 'concurrent-ruby'
   library 'faraday'
--- a/lib/datadog/profiling/component.rb
+++ b/lib/datadog/profiling/component.rb
@@ -207,7 +207,7 @@ module Datadog
         if Gem.loaded_specs['mysql2'] && incompatible_libmysqlclient_version?(settings)
           Datadog.logger.warn(
             'Enabling the profiling "no signals" workaround because an incompatible version of the mysql2 gem is ' \
-            'installed. Profiling data will have lower quality.' \
+            'installed. Profiling data will have lower quality. ' \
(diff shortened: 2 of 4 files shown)


ivoanjo:
Also reported issue upstream: https://github.com/alfa-jpn/mysql2-aurora/issues/13

marcotc (approved):

PR #2960, opened 2023-07-12
Improve our SafeDup module

GustavoCaso:
**What does this PR do?**

When working on https://github.com/DataDog/dd-trace-rb/pull/2936 I noticed that calling `SafeDup.frozen_or_dup({a: :b})` on ruby 3.2 would break; because it tries to call `+(-){a: :b}` which doesn't work. 

I had to ensure the `SafeDup` module worked for all objects. 

I decided to extract that work into a separate PR to ease the work of reviewing. 

The method `frozen_or_dup` and `frozen_dup` for ruby 2.3 and forward only expect to receive String values.  This could lead to future errors. 
By checking if the value is a String, we can warranty the same behaviour for all kinds of objects

I extracted the refinements into the `backport` file under the `BackportFrom24` namespace 

**Motivation**

**Additional Notes**

Interesting read on the whole `dup` and `clone` debacle for ruby versions https://blog.arkency.com/2017/03/prototypes-in-ruby/

**How to test the change?**


lib/datadog/core/utils/safe_dup.rb
```diff
@@ -1,40 +1,39 @@
...
-
         # String#+@ was introduced in Ruby 2.3
         if String.method_defined?(:+@) && String.method_defined?(:-@)
           def self.frozen_or_dup(v)
-            # If the string is not frozen, the +(-v) will:
-            # - first create a frozen deduplicated copy with -v
-            # - then it will dup it more efficiently with +v
-            v.frozen? ? v : +(-v)
+            case v
+            when String
```
marcotc, line 12: Can you capture here why `String` is handled separately from other types?

lib/datadog/core/utils/safe_dup.rb
```diff
@@ -1,40 +1,39 @@
...
+              v.frozen? ? v : +(-v)
+            else
+              v.frozen? ? v : Core::BackportFrom24.dup(v)
+            end
           end
 
           def self.frozen_dup(v)
-            -v if v
+            case v
+            when String
```
marcotc, line 24: Same here.

lib/datadog/core/utils/safe_dup.rb
GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/2960/commits/f4429a6d9501cc9e20985d92cdd71a0b30a40fa9

lib/datadog/core/utils/safe_dup.rb
GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/2960/commits/f4429a6d9501cc9e20985d92cdd71a0b30a40fa9

lib/datadog/core/backport.rb
```diff
@@ -21,5 +21,25 @@ def string_delete_prefix(string, prefix)
         end
       end
     end
+
+    # This module is used to provide features from Ruby 2.4+ to older Rubies
+    module BackportFrom24
+      module_function
+
+      if RUBY_VERSION < '2.4'
+        def dup(value)
```
ivoanjo, line 30: So about using `module_function` and `dup` here... 

Using `module_function` means that we can use the helpers from this module directly (e.g. `BackportFrom24.dup`) or by including it in a class. But with the current naming, you'd be overwriting `Kernel#dup`.

Aka either:
* Name this `kernel_dup(value)` (or `object_dup`?) so that including it as a class from other module doesn't have disastrous effects
* Don't make this a `module_function`, just a good ole `self.dup`

Now since I've been annoying you in this PR from the beginning in this PR about explicitness, you'll probably not be very surprised if I tell you I'm not a fan of module function and that I prefer that in some code we do 
`Core::BackportFrom24.dup(...)` rather than some mysterious `kernel_dup`, but I'll leave the decision to you :)

(I'll add that yeah the other `BackportFrom25` also uses `module_function`, but it never actually gets included anywhere, so it's a tiny diff to also convert it to sue `self` :P)

lib/datadog/core/backport.rb
```diff
@@ -21,5 +21,25 @@ def string_delete_prefix(string, prefix)
...
+          case value
+          when NilClass, TrueClass, FalseClass, Integer, Float
```
ivoanjo, line 32: I'm not sure if it's worth caring, but out of pure curiosity I went and checked and `Bignum` (e.g. `2**300.dup`) also breaks on 2.3 and works on modern Rubies.

lib/datadog/core/utils/safe_dup.rb
ivoanjo: Minor: Isn't `case v .... when String ... else ...` the same as `if String === v ... else`? (Same applies below as well)

lib/datadog/tracing/span_operation.rb
```diff
@@ -457,8 +445,8 @@ def build_span
           duration: duration,
           end_time: @end_time,
           id: @id,
-          meta: meta.dup,
-          metrics: metrics.dup,
+          meta: Core::Utils::SafeDup.frozen_or_dup(meta),
+          metrics: Core::Utils::SafeDup.frozen_or_dup(metrics),
```
ivoanjo, line 449: Wait... This change isn't equivalent -- the previous version only `dup`'d stuff, whereas now we're also freezing it. I think freezing seems like the right way to go, but worth triple-checking this is the case (which maybe you have, but I wanted to call it out!)

sig/datadog/core/backport.rbs
```diff
@@ -3,5 +3,9 @@ module Datadog
     module BackportFrom25
       def self?.string_delete_prefix: (::String string, ::String prefix) -> ::String
     end
+
+    module BackportFrom24
+      def self?.dup: (untyped value) -> untyped
```
ivoanjo, line 8: Minor: Should this be `(Object value) -> Object`?

spec/datadog/core/utils/safe_dup_spec.rb
```diff
@@ -1,62 +1,260 @@
...
+        it 'returns the original input' do
+          input = { a: :b }.freeze
+
+          result = described_class.frozen_or_dup(input)
+
+          expect(input).to be_frozen
+
+          expect(result).to eq(input)
+          expect(result).to be(input)
+          expect(result).to be_frozen
```
ivoanjo, line 78: Minor: The other checks are redundant -- if result and input are the same object (`be` check), then they will be `eq`, and we don't need to check that both are frozen.

```suggestion
          expect(result).to be(input)
          expect(result).to be_frozen
```

(Same note applies to a few other cases below)

spec/datadog/core/utils/safe_dup_spec.rb
```diff
@@ -1,62 +1,260 @@
...
+        end
+      end
+
+      context 'when given a string' do
+        it 'returns a non frozen array' do
```
ivoanjo, line 183:
```suggestion
      context 'when given a non-frozen array' do
        it 'returns a non-frozen copy of that array' do
```

spec/datadog/core/utils/safe_dup_spec.rb
```diff
@@ -1,62 +1,260 @@
...
+
+          expect(input).to be_frozen
+
+          expect(result).to eq(input)
+          expect(result).to be(input)
+          expect(result).to be_frozen
+        end
+      end
+
+      context 'when given a array' do
```
ivoanjo, line 212: Minor:
```suggestion
      context 'when given an array' do
```

ivoanjo (approved):
Looks great to me! I think the main two things to look at are the `module_function` thing (rename or remove) and making sure the change in span operation is correct, otherwise the rest is all smaller suggestions and nitpicks :)

lib/datadog/core/backport.rb
GustavoCaso: Yeah, I thought about adding it as well. But I wasn't sure if we use `Bignum` on our codebase. I'm guessing not, but adding it just in case makes sense to avoid future 💥

lib/datadog/tracing/span_operation.rb
GustavoCaso: @ivoanjo we do not return a frozen object here. `frozen_or_dup` only dups the object if is not frozen already. `frozen_dup` would be the one that would return a frozen duplicate if is the object is not already frozen

lib/datadog/core/backport.rb
ivoanjo: Yeap I vaguely remembered that. Yay for modern rubies!

lib/datadog/tracing/span_operation.rb
ivoanjo: Urghhh you're right, I read it as being `frozen_dup`. Please ignore my moment of insanity.

lib/datadog/core/utils/safe_dup.rb
```diff
@@ -1,40 +1,47 @@
...
-            # If the string is not frozen, the +(-v) will:
-            # - first create a frozen deduplicated copy with -v
-            # - then it will dup it more efficiently with +v
-            v.frozen? ? v : +(-v)
+            # For the case of a String we use the methods +@ and -@.
+            # Those methods are only for String objects
+            # they are faster and chepaer on the memory side.
+            # Check the benchmark on
+            # https://github.com/DataDog/dd-trace-rb/pull/2704
+            if v === String
```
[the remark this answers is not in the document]
ivoanjo: Interestingly enough, since this code path is a fast path, the tests don't catch this error ;)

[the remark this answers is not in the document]
GustavoCaso: I will use `v.is_a?(String)` is quite readable

marcotc (approved):

PR #2961, opened 2023-07-13
Prevent telemetry requests from being traced

marcotc:
Telemetry has its own custom transport, and it was missing a flag that lets the `Net::HTTP` tracing integration know to skip instrumenting the telemetry request.

This PR adds that flag, as an HTTP header, similar to how other `ddtrace` components operate.

There is also a bit of refactoring around this flag, to easy clarity during future use and simplify the circuit breaker implementation.

### DIFF
--- a/spec/datadog/core/telemetry/http/transport_spec.rb
+++ b/spec/datadog/core/telemetry/http/transport_spec.rb
@@ -27,9 +27,10 @@ RSpec.describe Datadog::Core::Telemetry::Http::Transport do
     let(:env) { instance_double(Datadog::Core::Telemetry::Http::Env, body: payload, path: path) }
     let(:headers) do
       {
-        Datadog::Core::Telemetry::Http::Ext::HEADER_CONTENT_TYPE => 'application/json',
-        Datadog::Core::Telemetry::Http::Ext::HEADER_DD_TELEMETRY_API_VERSION => 'v1',
-        Datadog::Core::Telemetry::Http::Ext::HEADER_DD_TELEMETRY_REQUEST_TYPE => request_type,
+        'Content-Type' => 'application/json',
+        'DD-Telemetry-API-Version' => 'v1',
+        'DD-Telemetry-Request-Type' => 'app-started',
+        'DD-Internal-Untraced-Request' => '1',
       }
     end
     let(:hostname) { 'foo' }
--- a/spec/datadog/tracing/contrib/http/circuit_breaker_spec.rb
+++ b/spec/datadog/tracing/contrib/http/circuit_breaker_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe Datadog::Tracing::Contrib::HTTP::CircuitBreaker do
 
     context 'given a normal request' do
       before do
-        allow(circuit_breaker).to receive(:datadog_http_request?)
+        allow(circuit_breaker).to receive(:internal_request?)
           .with(request)
           .and_return(false)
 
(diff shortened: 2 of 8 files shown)


spec/datadog/tracing/contrib/http/circuit_breaker_spec.rb
```diff
@@ -74,45 +74,19 @@
           it { is_expected.to be false }
         end
       end
-    end
-
-    context 'integration' do
```
marcotc, line 79: This integration test was moved to `spec/datadog/tracing/contrib/http/request_spec.rb` and refactored.

spec/datadog/tracing/contrib/http/circuit_breaker_spec.rb
```diff
@@ -74,45 +74,19 @@
...
-          let(:transport) do
-            Datadog::Transport::HTTP.default do |t|
-              t.adapter :unix, '/tmp/ddagent/trace.sock'
-            end
-          end
+        context 'is missing' do
+          let(:request) { ::Net::HTTP::Post.new('/some/path') }
 
-          it { is_expected.to be true }
+          it { is_expected.to be false }
```
marcotc, line 89: This diff is confusing because of the test removed: this PR did not change this assertion from `true` to `false`. See side-by-side for more clarity.

lloeki (approved):
LGTM!

spec/datadog/core/telemetry/http/transport_spec.rb
```diff
@@ -27,9 +27,10 @@
     let(:env) { instance_double(Datadog::Core::Telemetry::Http::Env, body: payload, path: path) }
     let(:headers) do
       {
-        Datadog::Core::Telemetry::Http::Ext::HEADER_CONTENT_TYPE => 'application/json',
-        Datadog::Core::Telemetry::Http::Ext::HEADER_DD_TELEMETRY_API_VERSION => 'v1',
-        Datadog::Core::Telemetry::Http::Ext::HEADER_DD_TELEMETRY_REQUEST_TYPE => request_type,
+        'Content-Type' => 'application/json',
+        'DD-Telemetry-API-Version' => 'v1',
+        'DD-Telemetry-Request-Type' => 'app-started',
+        'DD-Internal-Untraced-Request' => '1',
```
GustavoCaso, line 33: Curious why you decided to remove the constants. Not opposed to the change, I just want to know your reasoning 😄

GustavoCaso (approved):

spec/support/network_helpers.rb
```diff
@@ -43,7 +43,7 @@ def call_web_mock_function_with_agent_host_exclusions
   def check_availability_by_http_request(host, port)
     uri = URI("http://#{host}:#{port}/info")
     request = Net::HTTP::Get.new(uri)
-    request['X-Datadog-Untraced-Request'] = true
+    request[Datadog::Transport::Ext::HTTP::HEADER_DD_INTERNAL_UNTRACED_REQUEST] = true
```
TonyCTHsu, line 46: Should the value be `true` or string `'1'` ?

marcotc: String, thank you!

spec/datadog/core/telemetry/http/transport_spec.rb
marcotc: When you use to the System Under Test (SUT) as test assertions, you can end up with test assertions that are couple to your SUT.

This means you could have wrong assertions, because they are simply reflecting something incorrect in your SUT, and regurgitating it back to you in a test case.

This doesn't mean we should replace every reference from the SUT from our test assertions, but when feasible, it's good to avoid it.

In a few cases, it can create unmaintainable tests, and it might be fine to use the SUT, but I tend to avoid it if possible.

PR #2966, opened 2023-07-13
Fixes `web_service_name` default for rack integration

zarirhamza:
The introduction of the `SpanAttributeSchema` module, all service names in integrations default to global `$DD_SERVICE` when enabled. 

This PR adds that functionality to the `web_service_name` option as well for rack such that intermediate spans can also be flattened to `$DD_SERVICE` when enabled


zarirhamza:
Closing due to recognizing this is intended behavior

PR #2967, opened 2023-07-13
Internal:Fix environment variable for remote configuration polling interval

marcotc:
`DD_REMOTE_CONFIGURATION_POLL_INTERVAL_SECONDS` was changed to `DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS`.

Some new system-tests fail when enabled for Ruby because `DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS` is not processed by `ddtrace`, making the test timeout.

Also, this PR marks this setting as internal, because we don't have a valid use case for tweaking remote configuration pooling rate outside of testing purposes.


lib/datadog/core/configuration/settings.rb
```diff
@@ -541,8 +541,9 @@ def initialize(*_)
 
           # Tune remote configuration polling interval.
           #
-          # @default `DD_REMOTE_CONFIGURATION_POLL_INTERVAL_SECONDS` environment variable, otherwise `5.0` seconds.
+          # @default `DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS` environment variable, otherwise `5.0` seconds.
           # @return [Float]
+          # @!visibility private
```
ivoanjo, line 546: Should we document here a stronger "don't tune this, this setting is only used for internal validation and changing for other use-cases it is unsupported" or something like that?

lib/datadog/core/remote/ext.rb
```diff
@@ -5,7 +5,7 @@ module Core
     module Remote
       module Ext
         ENV_ENABLED = 'DD_REMOTE_CONFIGURATION_ENABLED'
-        ENV_POLL_INTERVAL_SECONDS = 'DD_REMOTE_CONFIGURATION_POLL_INTERVAL_SECONDS'
+        ENV_POLL_INTERVAL_SECONDS = 'DD_REMOTE_CONFIG_POLL_INTERVAL_SECONDS'
```
ivoanjo, line 8: This makes this setting somewhat inconsistent with `DD_REMOTE_CONFIGURATION_ENABLED`. I've searched on github and it seems internally we have both `DD_REMOTE_CONFIGURATION_ENABLED` and `DD_REMOTE_CONFIG_ENABLED`...

Maybe worth raising with the remote config folks what they want to do about this?

ivoanjo (approved):
:+1: LGTM

lib/datadog/core/configuration/settings.rb
GustavoCaso: Since this is only used from system tests and system tests only use ENV variables, we might get away by not exposing the setting at all. 

We could read the ENV variable. here: https://github.com/DataDog/dd-trace-rb/blob/86f9593268706f2b1a288a6ebf9e6dad29ff2e0d/lib/datadog/core/remote/component.rb#L33 and if it not present default to `5.0`

That way, there is no need to document that is private we just make it private 😎

lib/datadog/core/remote/ext.rb
GustavoCaso: I just asked on their Slack channel

GustavoCaso (approved):
LGTM. Left a comment on how to better hide this setting, but if we decide not to follow that path I'm happy with how it is right now 😄

lib/datadog/core/configuration/settings.rb
ivoanjo: That's a good point, although you lose the consistency of always going through the settings. I guess we could at least move such settings to an `internal` or even `debug` group, if we wanted to still keep them.

marcotc: I live having it go through `settings` for consistency. It will get better with configuration validation 😉.

I've added stronger wording for it being private.

PR #2970, opened 2023-07-14
Core configuration add env var and casting functionality

GustavoCaso:
**What does this PR do?**

The PR adds `env_var` and `deprecated_env_var` options to our internal configuration DSL, using the `type`  attribute to coerce ENV variables and validate the value set is validity.

There are a lot of files changes so I split the PR into smaller, more digestible commits:
- Add the logic for `env_var`, `deprecated_env_var` and type coercion https://github.com/DataDog/dd-trace-rb/pull/2970/commits/6e2f67941cb3fcb36cc1e5ff380dd4d7bd73e16e
- Update SpanAttributeSchema to not look for ENV variables https://github.com/DataDog/dd-trace-rb/pull/2970/commits/e25ab8294fb80065bf3e9e9151f5b55e7d86c8bf
- Update configuration to use the new options https://github.com/DataDog/dd-trace-rb/pull/2970/commits/424431d30d19579c65309808e7c725e3083b43bc
- Update specs https://github.com/DataDog/dd-trace-rb/pull/2970/commits/57b0aa7bbbb46079e91fe124c73cbf680660b36e

### Supported types 

For validation

```
:int
:float
:array
:string
:block
:hash
:bool
:symbol
```

For coercion. I used the same coercion types as we currently supported within `Core::Environment::VariableHelpers`

```
:int
:float
:array
:bool
```

Examples of configuration possibilities.

#### Example 1

```ruby
option :enabled do |o|
  o.type :bool
  o.env_var 'DD_APPSEC_ENABLED'
  o.default DEFAULT_APPSEC_ENABLED
end
```

We would try to extract the value from the `DD_APPSEC_ENABLED` ENV var, if the env var is present we would coerce it to a boolean value. In the case, the customer wants to configure using the configure block `c.appsec.enabled = true`. We would validate that the value provided is a boolean value. If the value doesn't match a boolean value, we would raise an exception with a useful message. 

#### Example 2

```ruby
option :enabled do |o|
  o.env_var 'DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING'
  o.default DEFAULT_APPSEC_AUTOMATED_TRACK_USER_EVENTS_ENABLED
  o.setter do |value|
    if value.is_a?(String)
      value != 'disabled'
    else
      !!value # rubocop:disable Style/DoubleNegation
    end
  end
end
```

This example does not use the `type` attribute, because the expected values are complex and could be multiple types. In this case, rather than pushing the complexity into the internal of the DSL, we do not set `type`, which would not try coercing ENV var or validate the provided value. We push the complexity into the `setter` block in this case. That way, those settings options deserve special attention; they could do so using the `setter` block.

#### Example 3

```ruby
option :default_rate do |o|
    o.type :float, nil: true
    o.env_var Tracing::Configuration::Ext::Sampling::ENV_SAMPLE_RATE
  end
```

This example signal uses the type option `nil: true` that signals to the option code that the value could be a Float or nil.

#### Example 4

```ruby
 option :tags do |o|
  o.type :hash, nilable: true
  o.env Core::Environment::Ext::ENV_TAGS
  o.env_parser do |env_value|
    values = if env_value.include?(',')
               env_value.split(',')
             else
               env_value.split(' ') # rubocop:disable Style/RedundantArgument
             end

    values.map! do |v|

      v.empty? ? nil : v
    end

    values.compact!
    values.each_with_object({}) do |tag, tags|
      key, value = tag.split(':', 2)
      tags[key] = value if value && !value.empty?
    end
  end
  o.setter do |new_value, old_value|
    raw_tags = new_value || {}

    env_value = env
    version_value = version
    service_name = service_without_fallback

    # Override tags if defined
    raw_tags[Core::Environment::Ext::TAG_ENV] = env_value unless env_value.nil?
    raw_tags[Core::Environment::Ext::TAG_VERSION] = version_value unless version_value.nil?

    # Coerce keys to strings
    string_tags = raw_tags.collect { |k, v| [k.to_s, v] }.to_h

    # Cross-populate tag values with other settings
    if env.nil? && string_tags.key?(Core::Environment::Ext::TAG_ENV)
      self.env = string_tags[Core::Environment::Ext::TAG_ENV]
    end

    if version_value.nil? && string_tags.key?(Core::Environment::Ext::TAG_VERSION)
      self.version = string_tags[Core::Environment::Ext::TAG_VERSION]
    end

    if service_name.nil? && string_tags.key?(Core::Environment::Ext::TAG_SERVICE)
      self.service = string_tags[Core::Environment::Ext::TAG_SERVICE]
    end

    # Merge with previous tags
    (old_value || {}).merge(string_tags)
  end
end
```

This example uses `env_parser` block. In cases where the parsing of env variables is more complex and can not be achieved using the default types, we can push the complexity to the `env_parser` block. 

In this example the tags ENV variable stores the information on a string with the format `tag1:value1,tag2:value2`, which gets converted into a hash `{"tag1" => "value1", "tag2" => "value2"}`

#### Breaking changes 
Also, I want to emphasize that **all these changes do not introduce any breaking change** 🎉 , since we haven't altered any existing code previously used for configuration.  

**Motivation**

 
**Additional Notes**

The logic for coercion inside of `option.rb` is very similar that the one of `Core::Environment::VariableHelpers`. Why did I not use that one? 

As of today, the code from `Core::Environment::VariableHelpers` is public, so any alteration to that code would have been considered a breaking change. Keeping internal of our configuration DSL private would allow us to iterate faster without worrying about breaking changes. 

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/appsec/configuration/settings.rb
+++ b/lib/datadog/appsec/configuration/settings.rb
@@ -34,7 +34,7 @@ module Datadog
             settings :appsec do
               option :enabled do |o|
                 o.type :bool
-                o.env_var 'DD_APPSEC_ENABLED'
+                o.env 'DD_APPSEC_ENABLED'
                 o.default DEFAULT_APPSEC_ENABLED
               end
 
@@ -52,7 +52,7 @@ module Datadog
               end
 
               option :ruleset do |o|
-                o.env_var 'DD_APPSEC_RULES'
+                o.env 'DD_APPSEC_RULES'
                 o.default DEFAULT_APPSEC_RULESET
               end
 
@@ -67,7 +67,7 @@ module Datadog
               end
 
               option :waf_timeout do |o|
-                o.env_var 'DD_APPSEC_WAF_TIMEOUT' # us
+                o.env 'DD_APPSEC_WAF_TIMEOUT' # us
                 o.default DEFAULT_APPSEC_WAF_TIMEOUT
                 o.setter do |v|
                   Datadog::Core::Utils::Duration.call(v.to_s, base: :us)
@@ -75,32 +75,32 @@ module Datadog
               end
 
               option :waf_debug do |o|
-                o.env_var 'DD_APPSEC_WAF_DEBUG'
+                o.env 'DD_APPSEC_WAF_DEBUG'
                 o.default DEFAULT_APPSEC_WAF_DEBUG
                 o.type :bool
               end
 
               option :trace_rate_limit do |o|
                 o.type :int
-                o.env_var 'DD_APPSEC_TRACE_RATE_LIMIT' # trace/s
+                o.env 'DD_APPSEC_TRACE_RATE_LIMIT' # trace/s
                 o.default DEFAULT_APPSEC_TRACE_RATE_LIMIT
               end
 
               option :obfuscator_key_regex do |o|
                 o.type :string
-                o.env_var 'DD_APPSEC_OBFUSCATION_PARAMETER_KEY_REGEXP'
+                o.env 'DD_APPSEC_OBFUSCATION_PARAMETER_KEY_REGEXP'
                 o.default DEFAULT_OBFUSCATOR_KEY_REGEX
               end
 
               option :obfuscator_value_regex do |o|
                 o.type :string
-                o.env_var 'DD_APPSEC_OBFUSCATION_PARAMETER_VALUE_REGEXP'
+                o.env 'DD_APPSEC_OBFUSCATION_PARAMETER_VALUE_REGEXP'
                 o.default DEFAULT_OBFUSCATOR_VALUE_REGEX
               end
 
               settings :track_user_events do
                 option :enabled do |o|
-                  o.env_var 'DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING'
+                  o.env 'DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING'
                   o.default DEFAULT_APPSEC_AUTOMATED_TRACK_USER_EVENTS_ENABLED
                   o.setter do |value|
                     if value.is_a?(String)
@@ -112,7 +112,7 @@ module Datadog
                 end
 
                 option :mode do |o|
-                  o.env_var 'DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING'
+                  o.env 'DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING'
                   o.default DEFAULT_APPSEC_AUTOMATED_TRACK_USER_EVENTS_MODE
                   o.setter do |v|
                     if APPSEC_VALID_TRACK_USER_EVENTS_MODE.include?(v.to_s)
--- a/lib/datadog/appsec/configuration/settings.rb
+++ b/lib/datadog/appsec/configuration/settings.rb
@@ -67,7 +67,7 @@ module Datadog
               end
 
               option :waf_timeout do |o|
-                o.env_var 'DD_APPSEC_WAF_TIMEOUT'
+                o.env_var 'DD_APPSEC_WAF_TIMEOUT' # us
                 o.default DEFAULT_APPSEC_WAF_TIMEOUT
                 o.setter do |v|
                   Datadog::Core::Utils::Duration.call(v.to_s, base: :us)
@@ -82,7 +82,7 @@ module Datadog
 
               option :trace_rate_limit do |o|
                 o.type :int
-                o.env_var 'DD_APPSEC_TRACE_RATE_LIMIT'
+                o.env_var 'DD_APPSEC_TRACE_RATE_LIMIT' # trace/s
                 o.default DEFAULT_APPSEC_TRACE_RATE_LIMIT
               end
 
--- a/lib/datadog/appsec/configuration/settings.rb
+++ b/lib/datadog/appsec/configuration/settings.rb
@@ -33,10 +33,9 @@ module Datadog
           base.class_eval do
             settings :appsec do
               option :enabled do |o|
-                o.default { env_to_bool('DD_APPSEC_ENABLED', DEFAULT_APPSEC_ENABLED) }
-                o.setter do |v|
-                  v ? true : false
-                end
+                o.type :bool
+                o.env_var 'DD_APPSEC_ENABLED'
+                o.default DEFAULT_APPSEC_ENABLED
               end
 
               define_method(:instrument) do |integration_name|
@@ -53,73 +52,71 @@ module Datadog
               end
 
               option :ruleset do |o|
-                o.default { ENV.fetch('DD_APPSEC_RULES', DEFAULT_APPSEC_RULESET) }
+                o.env_var 'DD_APPSEC_RULES'
+                o.default DEFAULT_APPSEC_RULESET
               end
 
               option :ip_denylist do |o|
-                o.default { [] }
+                o.type :array
+                o.default []
               end
 
               option :user_id_denylist do |o|
-                o.default { [] }
+                o.type :array
+                o.default []
               end
 
               option :waf_timeout do |o|
-                o.default { ENV.fetch('DD_APPSEC_WAF_TIMEOUT', DEFAULT_APPSEC_WAF_TIMEOUT) } # us
+                o.env_var 'DD_APPSEC_WAF_TIMEOUT'
+                o.default DEFAULT_APPSEC_WAF_TIMEOUT
                 o.setter do |v|
                   Datadog::Core::Utils::Duration.call(v.to_s, base: :us)
                 end
               end
 
               option :waf_debug do |o|
-                o.default { env_to_bool('DD_APPSEC_WAF_DEBUG', DEFAULT_APPSEC_WAF_DEBUG) }
-                o.setter do |v|
-                  v ? true : false
-                end
+                o.env_var 'DD_APPSEC_WAF_DEBUG'
+                o.default DEFAULT_APPSEC_WAF_DEBUG
+                o.type :bool
               end
 
               option :trace_rate_limit do |o|
-                o.default { env_to_int('DD_APPSEC_TRACE_RATE_LIMIT', DEFAULT_APPSEC_TRACE_RATE_LIMIT) } # trace/s
+                o.type :int
+                o.env_var 'DD_APPSEC_TRACE_RATE_LIMIT'
+                o.default DEFAULT_APPSEC_TRACE_RATE_LIMIT
               end
 
               option :obfuscator_key_regex do |o|
-                o.default { ENV.fetch('DD_APPSEC_OBFUSCATION_PARAMETER_KEY_REGEXP', DEFAULT_OBFUSCATOR_KEY_REGEX) }
+                o.type :string
+                o.env_var 'DD_APPSEC_OBFUSCATION_PARAMETER_KEY_REGEXP'
+                o.default DEFAULT_OBFUSCATOR_KEY_REGEX
               end
 
               option :obfuscator_value_regex do |o|
-                o.default do
-                  ENV.fetch(
-                    'DD_APPSEC_OBFUSCATION_PARAMETER_VALUE_REGEXP',
-                    DEFAULT_OBFUSCATOR_VALUE_REGEX
-                  )
-                end
+                o.type :string
+                o.env_var 'DD_APPSEC_OBFUSCATION_PARAMETER_VALUE_REGEXP'
+                o.default DEFAULT_OBFUSCATOR_VALUE_REGEX
               end
 
               settings :track_user_events do
                 option :enabled do |o|
-                  o.default do
-                    ENV.fetch(
-                      'DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING',
-                      DEFAULT_APPSEC_AUTOMATED_TRACK_USER_EVENTS_ENABLED
-                    )
-                  end
-                  o.setter do |v|
-                    if v
-                      v.to_s != 'disabled'
+                  o.env_var 'DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING'
+                  o.default DEFAULT_APPSEC_AUTOMATED_TRACK_USER_EVENTS_ENABLED
+                  o.setter do |value|
+                    if value.is_a?(String)
+                      value != 'disabled'
                     else
-                      false
+                      !!value # rubocop:disable Style/DoubleNegation
                     end
                   end
                 end
 
                 option :mode do |o|
-                  o.default do
-                    ENV.fetch('DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING', DEFAULT_APPSEC_AUTOMATED_TRACK_USER_EVENTS_MODE)
-                  end
+                  o.env_var 'DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING'
+                  o.default DEFAULT_APPSEC_AUTOMATED_TRACK_USER_EVENTS_MODE
                   o.setter do |v|
-                    string_value = v.to_s
-                    if APPSEC_VALID_TRACK_USER_EVENTS_MODE.include?(string_value)
-                      string_value
+                    if APPSEC_VALID_TRACK_USER_EVENTS_MODE.include?(v.to_s)
+                      v.to_s
                     else
                       Datadog.logger.warn(
                         'The appsec.track_user_events.mode value provided is not supported.' \
--- a/lib/datadog/ci/configuration/settings.rb
+++ b/lib/datadog/ci/configuration/settings.rb
@@ -17,7 +17,7 @@ module Datadog
             settings :ci do
               option :enabled do |o|
                 o.type :bool
-                o.env_var CI::Ext::Settings::ENV_MODE_ENABLED
+                o.env CI::Ext::Settings::ENV_MODE_ENABLED
                 o.default false
               end
 
--- a/lib/datadog/ci/configuration/settings.rb
+++ b/lib/datadog/ci/configuration/settings.rb
@@ -16,7 +16,9 @@ module Datadog
           base.class_eval do
             settings :ci do
               option :enabled do |o|
-                o.default { env_to_bool(CI::Ext::Settings::ENV_MODE_ENABLED, false) }
+                o.type :bool
+                o.env_var CI::Ext::Settings::ENV_MODE_ENABLED
+                o.default false
               end
 
               # DEV: Alias to Datadog::Tracing::Contrib::Extensions::Configuration::Settings#instrument.
@@ -34,12 +36,11 @@ module Datadog
               # TODO: Deprecate in the next major version, as `instrument` better describes this method's purpose
               alias_method :use, :instrument
 
-              option :trace_flush do |o|
-                o.default { nil }
-              end
+              option :trace_flush
 
               option :writer_options do |o|
-                o.default { {} }
+                o.type :hash
+                o.default({})
               end
             end
           end
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -138,7 +138,7 @@ module Datadog
           valid_type = validate(@definition.type, value)
 
           unless valid_type
-            raise_error = if @definition.type_options[:nil]
+            raise_error = if @definition.type_options[:nilable]
                             !value.is_a?(NilClass)
                           else
                             true
@@ -146,7 +146,7 @@ module Datadog
           end
 
           if raise_error
-            error_msg = if @definition.type_options[:nil]
+            error_msg = if @definition.type_options[:nilable]
                           "The option #{@definition.name} support this type `#{@definition.type}` "\
                                       "and `nil` but the value provided is #{value.class}"
                         else
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative '../utils/safe_dup'
+
 module Datadog
   module Core
     module Configuration
@@ -49,7 +51,7 @@ module Datadog
           end
 
           old_value = @value
-          (@value = context_exec(value, old_value, &definition.setter)).tap do |v|
+          (@value = context_exec(validate_type(value), old_value, &definition.setter)).tap do |v|
             @is_set = true
             @precedence_set = precedence
             context_exec(v, old_value, &definition.on_set) if definition.on_set
@@ -62,7 +64,7 @@ module Datadog
           elsif definition.delegate_to
             context_eval(&definition.delegate_to)
           else
-            set(default_value, precedence: Precedence::DEFAULT)
+            set_value_from_env_or_default
           end
         end
 
@@ -84,7 +86,7 @@ module Datadog
           if definition.default.instance_of?(Proc)
             context_eval(&definition.default)
           else
-            definition.experimental_default_proc || definition.default
+            definition.experimental_default_proc || Core::Utils::SafeDup.frozen_or_dup(definition.default)
           end
         end
 
@@ -94,12 +96,108 @@ module Datadog
 
         private
 
+        def coerce_env_variable(value)
+          case @definition.type
+          when :int
+            value.to_i
+          when :float
+            value.to_f
+          when :array
+            values = if value.include?(',')
+                       value.split(',')
+                     else
+                       value.split(' ') # rubocop:disable Style/RedundantArgument
+                     end
+
+            values.map! do |v|
+              v.gsub!(/\A[\s,]*|[\s,]*\Z/, '')
+
+              v.empty? ? nil : v
+            end
+
+            values.compact!
+            values
+          when :bool
+            string_value = value
+            string_value = string_value.downcase
+            string_value == 'true' || string_value == '1' # rubocop:disable Style/MultipleComparison
+          else
+            value
+          end
+        end
+
+        def validate_type(value)
+          raise_error = false
+
+          valid_type = validate(@definition.type, value)
+
+          unless valid_type
+            raise_error = if @definition.type_options[:nil]
+                            !value.is_a?(NilClass)
+                          else
+                            true
+                          end
+          end
+
+          if raise_error
+            error_msg = if @definition.type_options[:nil]
+                          "The option #{@definition.name} support this type `#{@definition.type}` "\
+                                      "and `nil` but the value provided is #{value.class}"
+                        else
+                          "The option #{@definition.name} support this type `#{@definition.type}` "\
+                          "but the value provided is #{value.class}"
+                        end
+
+            raise ArgumentError, error_msg
+          end
+
+          value
+        end
+
+        def validate(type, value)
+          case type
+          when :string
+            value.is_a?(String)
+          when :int
+            value.is_a?(Integer)
+          when :float
+            value.is_a?(Float)
+          when :array
+            value.is_a?(Array)
+          when :hash
+            value.is_a?(Hash)
+          when :bool
+            value.is_a?(TrueClass) || value.is_a?(FalseClass)
+          when :block
+            value.is_a?(Proc)
+          when :nil
+            value.is_a?(NilClass)
+          when :symbol
+            value.is_a?(Symbol)
+          else
+            true
+          end
+        end
+
         def context_exec(*args, &block)
           @context.instance_exec(*args, &block)
         end
 
         def context_eval(&block)
           @context.instance_eval(&block)
+        end
+
+        def set_value_from_env_or_default
+          if definition.env_var && ENV[definition.env_var]
+            set(coerce_env_variable(ENV[definition.env_var]), precedence: Precedence::PROGRAMMATIC)
+          elsif definition.deprecated_env_var && ENV[definition.deprecated_env_var]
+            Datadog::Core.log_deprecation do
+              "#{definition.deprecated_env_var} environment variable is deprecated, use #{definition.env_var} instead."
+            end
+            set(coerce_env_variable(ENV[definition.deprecated_env_var]), precedence: Precedence::PROGRAMMATIC)
+          else
+            set(default_value, precedence: Precedence::DEFAULT)
+          end
         end
 
         # Used for testing
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -195,17 +195,17 @@ module Datadog
           value = nil
           precedence = nil
 
-          if definition.env_var && ENV[definition.env_var]
-            value = coerce_env_variable(ENV[definition.env_var])
+          if definition.env && ENV[definition.env]
+            value = coerce_env_variable(ENV[definition.env])
             precedence = Precedence::PROGRAMMATIC
           end
 
-          if value.nil? && definition.deprecated_env_var && ENV[definition.deprecated_env_var]
-            value = coerce_env_variable(ENV[definition.deprecated_env_var])
+          if value.nil? && definition.deprecated_env && ENV[definition.deprecated_env]
+            value = coerce_env_variable(ENV[definition.deprecated_env])
             precedence = Precedence::PROGRAMMATIC
 
             Datadog::Core.log_deprecation do
-              "#{definition.deprecated_env_var} environment variable is deprecated, use #{definition.env_var} instead."
+              "#{definition.deprecated_env} environment variable is deprecated, use #{definition.env} instead."
             end
           end
 
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -121,8 +121,10 @@ module Datadog
             string_value = value
             string_value = string_value.downcase
             string_value == 'true' || string_value == '1' # rubocop:disable Style/MultipleComparison
-          else
+          when :string, NilClass
             value
+          else
+            raise ArgumentError, "The option #{@definition.name} is using an unknown type option `#{@definition.type}`"
           end
         end
 
@@ -168,12 +170,14 @@ module Datadog
             value.is_a?(Hash)
           when :bool
             value.is_a?(TrueClass) || value.is_a?(FalseClass)
-          when :block
+          when :proc
             value.is_a?(Proc)
           when :symbol
             value.is_a?(Symbol)
-          else
+          when NilClass
             true
+          else
+            raise ArgumentError, "The option #{@definition.name} is using an unknown type option `#{@definition.type}`"
           end
         end
 
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -190,16 +190,26 @@ module Datadog
         end
 
         def set_value_from_env_or_default
+          value = nil
+          precedence = nil
+
           if definition.env_var && ENV[definition.env_var]
-            set(coerce_env_variable(ENV[definition.env_var]), precedence: Precedence::PROGRAMMATIC)
-          elsif definition.deprecated_env_var && ENV[definition.deprecated_env_var]
+            value = coerce_env_variable(ENV[definition.env_var])
+            precedence = Precedence::PROGRAMMATIC
+          end
+
+          if value.nil? && definition.deprecated_env_var && ENV[definition.deprecated_env_var]
+            value = coerce_env_variable(ENV[definition.deprecated_env_var])
+            precedence = Precedence::PROGRAMMATIC
+
             Datadog::Core.log_deprecation do
               "#{definition.deprecated_env_var} environment variable is deprecated, use #{definition.env_var} instead."
             end
-            set(coerce_env_variable(ENV[definition.deprecated_env_var]), precedence: Precedence::PROGRAMMATIC)
-          else
-            set(default_value, precedence: Precedence::DEFAULT)
           end
+
+          option_value = value.nil? ? default_value : value
+
+          set(option_value, precedence: precedence || Precedence::DEFAULT)
         end
 
         # Used for testing
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -162,10 +162,8 @@ module Datadog
           case type
           when :string
             value.is_a?(String)
-          when :int
-            value.is_a?(Integer)
-          when :float
-            value.is_a?(Float)
+          when :int, :float
+            value.is_a?(Numeric)
           when :array
             value.is_a?(Array)
           when :hash
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -99,8 +99,10 @@ module Datadog
         def coerce_env_variable(value)
           case @definition.type
           when :int
+            # DEV-2.0: Change to a more strict coercion method. Integer(value).
             value.to_i
           when :float
+            # DEV-2.0: Change to a more strict coercion method. Float(value).
             value.to_f
           when :array
             values = if value.include?(',')
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -170,8 +170,6 @@ module Datadog
             value.is_a?(TrueClass) || value.is_a?(FalseClass)
           when :block
             value.is_a?(Proc)
-          when :nil
-            value.is_a?(NilClass)
           when :symbol
             value.is_a?(Symbol)
           else
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -97,6 +97,8 @@ module Datadog
         private
 
         def coerce_env_variable(value)
+          return context_exec(value, &@definition.env_parser) if @definition.env_parser
+
           case @definition.type
           when :int
             # DEV-2.0: Change to a more strict coercion method. Integer(value).
--- a/lib/datadog/core/configuration/option_definition.rb
+++ b/lib/datadog/core/configuration/option_definition.rb
@@ -94,7 +94,7 @@ module Datadog
             @default = block || value
           end
 
-          def experimental_default_proc(&block)
+          def experimental_default_proc(block)
             @experimental_default_proc = block
           end
 
--- a/lib/datadog/core/configuration/option_definition.rb
+++ b/lib/datadog/core/configuration/option_definition.rb
@@ -18,6 +18,7 @@ module Datadog
           :experimental_default_proc,
           :env_var,
           :deprecated_env_var,
+          :env_parser,
           :delegate_to,
           :depends_on,
           :name,
@@ -32,6 +33,7 @@ module Datadog
           @experimental_default_proc = meta[:experimental_default_proc]
           @env_var = meta[:env_var]
           @deprecated_env_var = meta[:deprecated_env_var]
+          @env_parser = meta[:env_parser]
           @delegate_to = meta[:delegate_to]
           @depends_on = meta[:depends_on] || []
           @name = name.to_sym
@@ -58,6 +60,7 @@ module Datadog
           def initialize(name, options = {})
             @env_var = nil
             @deprecated_env_var = nil
+            @env_parser = nil
             @default = nil
             @experimental_default_proc = nil
             @delegate_to = nil
@@ -88,6 +91,10 @@ module Datadog
 
           def deprecated_env_var(value)
             @deprecated_env_var = value
+          end
+
+          def env_parser(&block)
+            @env_parser = block
           end
 
           def default(value = nil, &block)
@@ -142,6 +149,7 @@ module Datadog
             default(options[:default]) if options.key?(:default)
             env_var(options[:env_var]) if options.key?(:env_var)
             deprecated_env_var(options[:deprecated_env_var]) if options.key?(:deprecated_env_var)
+            env_parser(&options[:env_parser]) if options.key?(:env_parser)
             delegate_to(&options[:delegate_to]) if options.key?(:delegate_to)
             depends_on(*options[:depends_on]) if options.key?(:depends_on)
             lazy(options[:lazy]) if options.key?(:lazy)
@@ -161,6 +169,7 @@ module Datadog
               experimental_default_proc: @experimental_default_proc,
               env_var: @env_var,
               deprecated_env_var: @deprecated_env_var,
+              env_parser: @env_parser,
               delegate_to: @delegate_to,
               depends_on: @depends_on,
               on_set: @on_set,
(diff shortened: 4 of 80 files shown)


lib/datadog/core/configuration/option.rb
```diff
@@ -94,6 +96,87 @@ def default_precedence?
 
         private
 
+        def coerce_env_variable(value)
+          case @definition.type
+          when :int
+            value.to_i
```
marcotc, line 102: Will this mask empty strings as `0`?
```ruby
''.to_i
# => 0
```

I think this is `to_i` here because `env_to_int`'s current implementation is `to_i`. This is fair, as we don't want to break compatibility, despite this implementation not being strictly correct.

If we need to keep `to_i` here to ensure backwards compatibility, can I ask you to document it here (and also around `#to_f` below) that we want to make this more strict in ddtrace 2.0? All invalid values becoming zero is definitely not what we want here.

lib/datadog/core/configuration/settings.rb
```diff
@@ -431,36 +468,54 @@ def initialize(*_)
...
-          o.default do
-            tags = {}
-
-            # Parse tags from environment
-            env_to_list(Core::Environment::Ext::ENV_TAGS, comma_separated_only: false).each do |tag|
-              key, value = tag.split(':', 2)
-              tags[key] = value if value && !value.empty?
-            end
+          o.env_var Core::Environment::Ext::ENV_TAGS
+          o.setter do |new_value, old_value|
```
marcotc, line 472: Why does `:tags` need its own custom setter now? Before it was happily using `env_to_list`..

lib/datadog/tracing/contrib/span_attribute_schema.rb
```diff
@@ -7,16 +7,18 @@ module Contrib
...
-        def fetch_service_name(env, default)
-          ENV.fetch(env) do
-            if Datadog.configuration.tracing.span_attribute_schema ==
-                Tracing::Configuration::Ext::SpanAttributeSchema::VERSION_ONE
-              Datadog.configuration.service
-            else
-              default
-            end
+        # rubocop:disable Style/OptionalArguments
+        def fetch_service_name(value = nil, default)
```
marcotc, line 11: Can you have a leading positional argument with a default value?

I'm confused on what `SpanAttributeSchema.fetch_service_name('my-value')` will do.

I guess Rubocop didn't like it either. Maybe it's a non-intuitive pattern.

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -53,73 +52,71 @@ def self.add_settings!(base)
...
               end
 
               option :user_id_denylist do |o|
-                o.default { [] }
+                o.type :array
+                o.default []
               end
 
               option :waf_timeout do |o|
-                o.default { ENV.fetch('DD_APPSEC_WAF_TIMEOUT', DEFAULT_APPSEC_WAF_TIMEOUT) } # us
```
marcotc, line 68: We probably want to capture the units (`us`), even if just as a comment like before. Same for the `trace_rate_limit` option below.

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -53,73 +52,71 @@ def self.add_settings!(base)
...
-                    if v
-                      v.to_s != 'disabled'
+                  o.env_var 'DD_APPSEC_AUTOMATED_USER_EVENTS_TRACKING'
+                  o.default DEFAULT_APPSEC_AUTOMATED_TRACK_USER_EVENTS_ENABLED
+                  o.setter do |value|
+                    if value.is_a?(String)
+                      value != 'disabled'
                     else
-                      false
+                      !!value # rubocop:disable Style/DoubleNegation
```
marcotc, line 109: This cop is not great, we might as well disable it (doesn't have to be done in this PR).

lib/datadog/core/configuration/option.rb
```diff
@@ -94,6 +96,87 @@ def default_precedence?
...
+          when :hash
+            value.is_a?(Hash)
+          when :bool
+            value.is_a?(TrueClass) || value.is_a?(FalseClass)
+          when :block
+            value.is_a?(Proc)
+          when :symbol
+            value.is_a?(Symbol)
+          else
+            true
```
marcotc, line 176: I think we should emit a message through `Datadog.logger` here to let the user know that their type wasn't recognized.

marcotc (commented):
Validation is always something we lacked, and could bite `ddtrace` users easily.

Thank you so much for this work, @GustavoCaso!

I'd like to hear others opinions on this one, and I think @delner is probably a "mandatory" reviewer here too :)

lib/datadog/core/configuration/settings.rb
GustavoCaso: The `tags` is a complex setting. It accepts an ENV var that gets parsed as a list, and for customers configuring it directly though `Datadog.configured` block, it accepts a hash with the different tags. 

To keep the DSL simple and avoid extra complexity to support all cases. The settings that accept multiple value types do not have to define any `type` and push all the logic into the `setter` block. 

I'd like to explain this setting change in detail since it is the most complex one from the coed base.

Old code:
```ruby
option :tags do |o|
  o.default do
    tags = {}

    # Parse tags from environment
    env_to_list(Core::Environment::Ext::ENV_TAGS, comma_separated_only: false).each do |tag|
      key, value = tag.split(':', 2)
      tags[key] = value if value && !value.empty?
    end

    # Override tags if defined
    tags[Core::Environment::Ext::TAG_ENV] = env unless env.nil?
    tags[Core::Environment::Ext::TAG_VERSION] = version unless version.nil?

    tags
  end

  o.setter do |new_value, old_value|
    # Coerce keys to strings
    string_tags = new_value.collect { |k, v| [k.to_s, v] }.to_h

    # Cross-populate tag values with other settings
    if env.nil? && string_tags.key?(Core::Environment::Ext::TAG_ENV)
      self.env = string_tags[Core::Environment::Ext::TAG_ENV]
    end

    if version.nil? && string_tags.key?(Core::Environment::Ext::TAG_VERSION)
      self.version = string_tags[Core::Environment::Ext::TAG_VERSION]
    end

    if service_without_fallback.nil? && string_tags.key?(Core::Environment::Ext::TAG_SERVICE)
      self.service = string_tags[Core::Environment::Ext::TAG_SERVICE]
    end

    # Merge with previous tags
    (old_value || {}).merge(string_tags)
  end
end
```

We can see the `default`  block, parse the ENV to a list and transform its values into a hash, plus some overrides. This block is only executed if the value is not provided by `Datadog.configure` block.
The `setter`, in this case, always assumes that it receives a `hash` either because the ENV has been parsed in the `default` block or the customer has provided over `Datadog.configure` block.


New code:
```ruby
option :tags do |o|
  o.env_var Core::Environment::Ext::ENV_TAGS
  o.setter do |new_value, old_value|
    tag_list = case new_value
               when String
                 values = if new_value.include?(',')
                            new_value.split(',')
                          else
                            new_value.split(' ') # rubocop:disable Style/RedundantArgument
                          end

                 values.map! do |v|

                   v.empty? ? nil : v
                 end

                 values.compact!
                 values.each_with_object({}) do |tag, tags|
                   key, value = tag.split(':', 2)
                   tags[key] = value if value && !value.empty?
                 end
               when Hash
                 new_value
               else
                 {}
               end

    env_value = env
    version_value = version
    service_name = service_without_fallback

    # Override tags if defined
    tag_list[Core::Environment::Ext::TAG_ENV] = env_value unless env_value.nil?
    tag_list[Core::Environment::Ext::TAG_VERSION] = version_value unless version_value.nil?

    # Coerce keys to strings
    string_tags = tag_list.collect { |k, v| [k.to_s, v] }.to_h

    # Cross-populate tag values with other settings
    if env.nil? && string_tags.key?(Core::Environment::Ext::TAG_ENV)
      self.env = string_tags[Core::Environment::Ext::TAG_ENV]
    end

    if version_value.nil? && string_tags.key?(Core::Environment::Ext::TAG_VERSION)
      self.version = string_tags[Core::Environment::Ext::TAG_VERSION]
    end

    if service_name.nil? && string_tags.key?(Core::Environment::Ext::TAG_SERVICE)
      self.service = string_tags[Core::Environment::Ext::TAG_SERVICE]
    end

    # Merge with previous tags
    (old_value || {}).merge(string_tags)
  end
end
```

We do not specify any `type` on the setting. The `setter` block is responsible for accounting for the case the value is a string coming from reading the ENV variable, and a `Hash` from `Datadog.configure` block. In the case of any of those, we default to an empty hash. 
Everything below that checking of type of value we are dealing with is identical as the old code 😄

lib/datadog/appsec/configuration/settings.rb
GustavoCaso: Agree with that. I will take off that on a separate PR

lib/datadog/tracing/contrib/span_attribute_schema.rb
GustavoCaso: Great point 😄 

Done https://github.com/DataDog/dd-trace-rb/pull/2970/commits/c27577e592813e14faec446da2f391820d205c73

lib/datadog/core/configuration/option_definition.rb
```diff
@@ -74,6 +82,14 @@ def depends_on(*values)
             @depends_on = values.flatten
           end
 
+          def env_var(value)
```
lloeki, line 85: `<nitpick level=420% bikeshed=yes>`

I feel like `environment` (mimicking Python's `os.environment`) or `env` (mimicking Ruby's `ENV`) look cleaner in the DSL?

Examples:

```
option :foo_bar do |o|
  o.type :string
  o.environment 'DD_FOO_BAR'
end
```

```
option :foo_bar do |o|
  o.type :string
  o.env 'DD_FOO_BAR'
end
```

`</nitpick>`

lib/datadog/core/configuration/option.rb
```diff
@@ -94,6 +96,87 @@ def default_precedence?
...
+          end
+        end
+
+        def validate_type(value)
+          raise_error = false
+
+          valid_type = validate(@definition.type, value)
+
+          unless valid_type
+            raise_error = if @definition.type_options[:nil]
```
lloeki, line 135: I might prefer `:nilable` (or `:allow_nil`) here.

Example:

```
o.type :float, nilable: true
```

lib/datadog/core/configuration/option.rb
lloeki: Definitely worth adding a `# TODO:` comment explaining that.

lib/datadog/core/configuration/option.rb
lloeki: I see two cases here:

- an option type is not enforced, `type` is presumably `nil` and `validate` should return `true` because all value types are accepted:

  ```
  when nil
    true
  ```

  Is that something we want, or should we disallow unenforced types altogether?
  
- an option definition receives an unknown type, e.g:
  
  ```
  option :foo do |o|
    o.type :frobz
  end
  ```
  
  Then the type cannot be validated, but that error is _on us_, because the option definition is incorrect, so to me `validate` should blow up:
  
  ```
  else
    raise ConfigurationSettingError, "unknown type: #{type}"
  ```

lloeki (approved):
Overall it looks good, noted some possible improvements

lib/datadog/core/configuration/option.rb
GustavoCaso: Right now, when the `type` is `nil` it is a way to signal that we do not want any validation or coercion.

> Is that something we want, or should we disallow unenforced types altogether?

That would be the goal, unfortunately, the current configuration options include some complex settings that coercion and validation differ. An example is the `tags` coerce the environment variable into an array and the supported value for setting the option via the configure block is a hash. 

A possible solution would be to add some functionality to the coercion to be able to get the string value to a hash. 

As for invalid type coercion of validation I like the idea to raise an exception ex: `ConfigurationSettingError `

lib/datadog/core/configuration/option.rb
GustavoCaso: I'm ok with whatever the team decides. 
`nilable: true` sounds good

lib/datadog/core/configuration/option.rb
GustavoCaso: Added the code to raise an exception if the type value is invalid here: https://github.com/DataDog/dd-trace-rb/pull/2970/commits/7a476a00f0bf7c728f80c34660507e61c8891d97

lloeki: > when the `type` is `nil` it is a way to signal that we do not want any validation or coercion.

I realised that when seeing the other discussion about delegating complex type checking to the setter, but thanks for the clarification.

lib/datadog/core/configuration/option.rb
```diff
@@ -94,6 +96,91 @@ def default_precedence?
...
+          when :bool
+            value.is_a?(TrueClass) || value.is_a?(FalseClass)
+          when :proc
+            value.is_a?(Proc)
+          when :symbol
+            value.is_a?(Symbol)
+          when NilClass
+            true
+          else
+            raise ArgumentError, "The option #{@definition.name} is using an unknown type option `#{@definition.type}`"
```
lloeki, line 180: Should we have a specific exception (or exceptions) for such configuration validation errors?

I'm thinking of at least a few:

- `Configuration::Error`: generic error
- `Configuration::TypeError`: when type does not match
- `Configuration::ValueError`: when value is of the correct type but fails to pass a value check, esp. reusable for those `setter`-time checks
- `Configuration::ConsistencyError`: when e.g two settings are mutually exclusive but both are set in inconsistent ways
- etc...

Those would be automatic for simple types, but be especially reusable for those `setter`-time checks that can be much more rich.

lib/datadog/core/configuration/settings.rb
marcotc: Makes sense!
I suggest leaving a comment saying we should deprecate allowing this to be with anything that does not reflect the environment variable String format (e.g. `c.tags = Hash`), as this will complicate long-term maintenance of this setting quite a lot.

lib/datadog/core/configuration/option_definition.rb
marcotc: One more vote for `env`.

lib/datadog/core/configuration/option.rb
marcotc: 👍 for `nilable`

lib/datadog/core/configuration/option_definition.rb
```diff
@@ -16,24 +16,30 @@ class OptionDefinition
...
           :setter,
-          :type
+          :type,
+          :type_options
 
         def initialize(name, meta = {}, &block)
           @default = meta[:default]
           @experimental_default_proc = meta[:experimental_default_proc]
+          @env_var = meta[:env_var]
+          @deprecated_env_var = meta[:deprecated_env_var]
```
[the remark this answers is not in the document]
marcotc: Both being lists is a great suggestion, actually. I think we should do it.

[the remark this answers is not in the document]
marcotc: Deprecation should emit a warning and use the value when needed (only after all `@env_var`s have been checked first).

lib/datadog/core/configuration/option.rb
marcotc: If there's enough repetition, the error-specific types seems fine.

If they are not very reusable today, then I think it's fine to leave them as they are today.

spec/datadog/core/telemetry/collector_spec.rb
```diff
@@ -10,6 +10,7 @@
 require 'datadog/core/telemetry/v1/integration'
 require 'datadog/core/telemetry/v1/product'
 require 'ddtrace/transport/ext'
+require 'datadog/profiling/profiler'
```
ivoanjo, line 13: Minor: Is this really needed? I don't see any references to profiling in the collector, other than to its settings which are not under `datadog/profiling/profiler`.

lib/datadog/core/configuration/option.rb
```diff
@@ -94,6 +96,91 @@ def default_precedence?
...
+        end
+
+        def validate(type, value)
+          case type
+          when :string
+            value.is_a?(String)
+          when :int
+            value.is_a?(Integer)
+          when :float
+            value.is_a?(Float)
```
ivoanjo, line 166: So I think strict `:float` validation is a bit problematic -- in a lot of cases we're using floats to allow for higher definition of values, but an integer is as valid as a float in those situations when you don't need the decimal part.

A good example is the number of defaults and specs that needed to change to pass this validation -- I think we'll be causing undue annoyance to customers here.

My suggestion is -- either accept either an int or a float when a `:float` is specified OR rename `:float` to `:numeric` and only check for `is_a?(Numeric)`.

lib/datadog/core/configuration/option.rb
```diff
@@ -94,6 +96,91 @@ def default_precedence?
 
         private
 
+        def coerce_env_variable(value)
```
ivoanjo, line 99: I believe the intention here is to replace `Datadog::Core::Environment::VariableHelpers`. Should we go ahead and mark all those methods as deprecated for removal?

lib/datadog/core/configuration/option.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
```
ivoanjo, line 1: Meta note on this PR: This is the kind of PR that has the potential to have undetected conflicts with other ongoing PRs -- e.g. any PRs that touch settings that branched out before this PR have the potential to cause issues.

Here's my suggestion -- before merging this PR to master:

1. Diff master and the base of the `core-configuration-add-env-var-and-casting-functionality` branch, and see if there's any changes that haven't been picked up
2. (if needed) Merge master into this branch, include any updates
3. Do a quick round of open PRs and see if any touch the settings; If they do, ask them to hold off until this gets merged, and after that, ask them to rebase on top of master.

lib/datadog/core/configuration/option.rb
ivoanjo: To be honest, what's the value of the more specific types? My thinking is -- if you messed up with configuration, you need to fix it.

We don't expect customers to be rescuing any of the more specific types, so if you're not going to be rescuing them, is there a lot of value in having different types?

ivoanjo (commented):
I like the idea, although I wish the PR diff was waaaay smaller lol. 

There's a handful of files where I ran out of time and didn't go as deep as I wanted today, but I'm posting my current notes as will do a second pass on those files tomorrow.

lib/datadog/ci/configuration/settings.rb
```diff
@@ -34,12 +36,11 @@ def self.add_settings!(base)
...
 
-              option :trace_flush do |o|
-                o.default { nil }
-              end
+              option :trace_flush
 
               option :writer_options do |o|
-                o.default { {} }
+                o.type :hash
+                o.default({})
```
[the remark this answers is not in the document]
GustavoCaso: I tried that but rubocop suggest to use `{}` 😢

lib/datadog/tracing/contrib/aws/configuration/settings.rb
```diff
@@ -13,21 +13,29 @@ module Configuration
...
-              o.default do
+              o.type :string, nil: true
+              o.env_var Ext::ENV_SERVICE_NAME
+              o.setter do |value|
                 Contrib::SpanAttributeSchema.fetch_service_name(
-                  Ext::ENV_SERVICE_NAME,
+                  value,
                   Ext::DEFAULT_PEER_SERVICE_NAME
                 )
               end
```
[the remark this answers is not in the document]
GustavoCaso: > Also: why use o.setter and instead of o.default?

We use setter rather than default because:
- We changed the `Contrib::SpanAttributeSchema.fetch_service_name` to not check ENV directly, but just values provided to it. 
- The `default` block is not executed with the customer setting value. 

So here is the setter `value` could be the ENV variable `Ext::ENV_SERVICE_NAME` or what ever the customer configured via `Datadog.configure` block

spec/datadog/core/telemetry/collector_spec.rb
GustavoCaso: When running `bundle exec rake spec:main` was complaining that the constant was missing.

Let me double check

lib/datadog/core/configuration/option.rb
GustavoCaso: I agree with that. There is still one setting that still uses `Datadog::Core::Environment::VariableHelpers`

```ruby
settings :client_ip do
  # Whether client IP collection is enabled. When enabled client IPs from HTTP requests will
  #   be reported in traces.
  #
  # Usage of the DD_TRACE_CLIENT_IP_HEADER_DISABLED environment variable is deprecated.
  #
  # @see https://docs.datadoghq.com/tracing/configure_data_security#configuring-a-client-ip-header
  #
  # @default `DD_TRACE_CLIENT_IP_ENABLED` environment variable, otherwise `false`.
  # @return [Boolean]
  option :enabled do |o|
    o.type :bool
    o.default do
      disabled = env_to_bool(Tracing::Configuration::Ext::ClientIp::ENV_DISABLED)

      enabled = if disabled.nil?
                  false
                else
                  Datadog::Core.log_deprecation do
                    "#{Tracing::Configuration::Ext::ClientIp::ENV_DISABLED} environment variable is deprecated, use #{Tracing::Configuration::Ext::ClientIp::ENV_ENABLED} instead."
                  end

                  !disabled
                end

      # ENABLED env var takes precedence over deprecated DISABLED
      env_to_bool(Tracing::Configuration::Ext::ClientIp::ENV_ENABLED, enabled)
    end
  end
end
```

I guess it is ok to mark the methods from `Datadog::Core::Environment::VariableHelpers` as deprecated. 

I would like to that on separate PR to avoid touching more parts of the codebase on this single PR

lib/datadog/core/configuration/option.rb
GustavoCaso: Sounds good. Thanks for the suggestion

lib/datadog/core/configuration/option.rb
```diff
@@ -94,6 +96,91 @@ def default_precedence?
...
+                     end
+
+            values.map! do |v|
+              v.gsub!(/\A[\s,]*|[\s,]*\Z/, '')
+
+              v.empty? ? nil : v
+            end
+
+            values.compact!
+            values
```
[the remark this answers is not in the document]
GustavoCaso: Very good point @ivoanjo 

Will see what I can do

lib/datadog/appsec/configuration/settings.rb
marcotc: https://github.com/DataDog/dd-trace-rb/pull/2979

lib/datadog/ci/configuration/settings.rb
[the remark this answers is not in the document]
marcotc: Both look fine to me. 🤷

lib/datadog/core/configuration/option.rb
marcotc: I like :numeric in this case, it captures that both integers and floats are accepted in one type.

GustavoCaso: I'm going to leave the distinction of `:int` and `:float` as for ENV variable parsing and coercion, we still have the distinction between int and floats, but I will change the validation portion to: 

```ruby
when :int, :float
  value.is_a?(Numeric)
end
```

lib/datadog/ci/configuration/settings.rb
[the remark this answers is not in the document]
ivoanjo: Ahh Rubocop... I so so dislike half of its default settings...

spec/datadog/core/telemetry/collector_spec.rb
ivoanjo: Aaahh you're right. I had only looked at the `environment_logger.rb` and tried to run the spec and did not see the same result you did.

I missed that:
1) The `environment_logger_spec.rb` directly mentions the profiler
2) The `Datadog::Profiling::Profiler` class is only loaded [if profiling is supported](https://github.com/DataDog/dd-trace-rb/blob/7632a804d7a31ea706ac3b42fa46a449b67b8962/lib/datadog/profiling.rb#L191). This is why the spec worked for me but failed for you -- on my machine the profiler is set up correctly so it's supported, and on yours it hasn't been compiled :)

lib/datadog/core/configuration/option.rb
GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/2970/commits/15246fc9b9ecd88deeeff161e9fd3e8bf2b7cc5e

lib/datadog/core/configuration/settings.rb
GustavoCaso: I decided to add `env_parser` block to account for complex env parsing logic. 

https://github.com/DataDog/dd-trace-rb/pull/2970/commits/d6ab986ab909ee6fabbcba4cb64d19107225bc65

lib/datadog/core/configuration/option.rb
GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/2970/commits/345454af4bfc3e57b86157aa6d89f07892d24488

lib/datadog/core/configuration/option_definition.rb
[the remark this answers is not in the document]
GustavoCaso: @marcotc @lloeki I support this. 


To keep the PR smaller (if that even possible 🤣 ), I will add that extra functionality on a separate PR or once we find the use case for it

lib/datadog/core/configuration/option.rb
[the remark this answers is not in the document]
GustavoCaso: This use case where tackled on https://github.com/DataDog/dd-trace-rb/pull/2970/commits/d6ab986ab909ee6fabbcba4cb64d19107225bc65 Thanks to the use of `env_parser` block

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -50,23 +51,16 @@ def self.extended(base)
                 #   otherwise `['Datadog','b3multi','b3']`.
                 # @return [Array<String>]
                 option :propagation_extract_style do |o|
-                  o.default do
-                    # DEV-2.0: Change default value to `tracecontext, Datadog`.
-                    # Look for all headers by default
```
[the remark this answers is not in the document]
GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/2970/commits/15246fc9b9ecd88deeeff161e9fd3e8bf2b7cc5e

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -91,18 +85,10 @@ def self.extended(base)
                 # @default `DD_TRACE_PROPAGATION_STYLE_INJECT` environment variable (comma-separated list), otherwise `['Datadog']`.
                 # @return [Array<String>]
                 option :propagation_inject_style do |o|
-                  o.default do
-                    # DEV-2.0: Change default value to `tracecontext, Datadog`.
```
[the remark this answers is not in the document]
GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/2970/commits/15246fc9b9ecd88deeeff161e9fd3e8bf2b7cc5e

lib/datadog/core/configuration/option.rb
GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/2970/commits/765a2870cf8063db8121e42083ede9c7243e2427

lib/datadog/core/configuration/option.rb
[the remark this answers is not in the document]
lloeki: `comma_separated_only`, `array_can_have_spaces` could become `sep: ','` or something (I'm thinking something like [bash `IFS`](https://tldp.org/LDP/abs/html/internalvariables.html#IFSREF) here)

lib/datadog/core/configuration/option.rb
lloeki: > I'm going to leave the distinction of :int and :float as for ENV variable parsing and coercion, we still have the distinction between int and floats, but I will change the validation portion to:

I think it is a good and simple example highlighting how coercion is a thing specific to `ENV` (which can contain only strings which contain some form of serialized values) whereas Ruby type checking is a thing specific to the setting itself. Coercion of `ENV` resulting in a type accepted by the setting is only incidental.

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -53,73 +52,71 @@ def self.add_settings!(base)
...
+                o.type :array
+                o.default []
               end
 
               option :waf_timeout do |o|
-                o.default { ENV.fetch('DD_APPSEC_WAF_TIMEOUT', DEFAULT_APPSEC_WAF_TIMEOUT) } # us
+                o.env_var 'DD_APPSEC_WAF_TIMEOUT' # us
+                o.default DEFAULT_APPSEC_WAF_TIMEOUT
                 o.setter do |v|
                   Datadog::Core::Utils::Duration.call(v.to_s, base: :us)
```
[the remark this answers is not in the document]
lloeki: > in our other libraries

Except in Go, coz it's [parsed as a `time.Duration`](https://github.com/DataDog/dd-trace-go/blob/be1878f26eff1be2fde73bc1819d2e6db792f4f9/internal/appsec/config.go#L115)) using [`time.ParseDuration`](https://pkg.go.dev/time#ParseDuration), which [doesn't accept unitless strings](https://play.golang.com/p/CJfhzjMBmOy), and the env var got used with a suffix in system tests coz Go was the first one there, and then it got in the wild, so I had to implement this parser otherwise both system tests and Go customers having already set env vars in some global way would crash.

But they [fixed since](https://github.com/DataDog/dd-trace-go/blob/be1878f26eff1be2fde73bc1819d2e6db792f4f9/internal/appsec/config.go#L111-L113) by appending `us` when it's missing and IIRC system tests have been adjusted to not use the suffix.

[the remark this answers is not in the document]
lloeki: > Consider maybe going back to o.type :int in the future?

I'm wondering if an alternative would be to have a `o.type :duration` shorthand and apply that wherever we have durations.

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -53,73 +52,71 @@ def self.add_settings!(base)
...
               option :ruleset do |o|
-                o.default { ENV.fetch('DD_APPSEC_RULES', DEFAULT_APPSEC_RULESET) }
+                o.env 'DD_APPSEC_RULES'
+                o.default DEFAULT_APPSEC_RULESET
               end
 
               option :ip_denylist do |o|
-                o.default { [] }
+                o.type :array
+                o.default []
```
lloeki, line 61: There's a difference between block and value here:
1. `[]` would be `eval`'d when that line is evaluated, so ever produces a single array
2.`{ [] }` would be `eval`'d when the block is evaluated, so may or may not produce a single array

This led me to wonder if we should `freeze` that default in the 1. case to ensure it does not get mutated.

This then led me to wonder if we should in fact (transparently) `freeze` all configuration setting values, because once they are set they should not be mutated.

WDYT?

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -33,10 +33,9 @@ def self.add_settings!(base)
           base.class_eval do
             settings :appsec do
               option :enabled do |o|
-                o.default { env_to_bool('DD_APPSEC_ENABLED', DEFAULT_APPSEC_ENABLED) }
-                o.setter do |v|
-                  v ? true : false
-                end
+                o.type :bool
+                o.env 'DD_APPSEC_ENABLED'
+                o.default DEFAULT_APPSEC_ENABLED
```
lloeki, line 38: This feels weird, seems like it should read `o.default true`.

Indeed whatever else depends on `DEFAULT_APPSEC_ENABLED` should instead read the configuration setting.

lib/datadog/core/configuration/option.rb
lloeki: > Meta note on this PR

Meta-question: this has nothing to do with the `# frozen_string_literal: true` comment, and only uses that first line as anchor for a comment, correct?

Meta-suggestion: maybe it should go as a review comment in "Finish your review"? but then it's not a "threaded" discussion I guess (subsequent discussion would happen in the global PR discussion tab).

lib/datadog/core/configuration/settings.rb
```diff
@@ -399,14 +434,15 @@ def initialize(*_)
...
+          o.env Core::Environment::Ext::ENV_SERVICE
+          o.default Core::Environment::Ext::FALLBACK_SERVICE_NAME
 
           # There's a few cases where we don't want to use the fallback service name, so this helper allows us to get a
           # nil instead so that one can do
           # nice_service_name = Datadog.configuration.service_without_fallback || nice_service_name_default
           o.helper(:service_without_fallback) do
             service_name = service
-            service_name unless service_name.equal?(Core::Environment::Ext::FALLBACK_SERVICE_NAME)
+            service_name unless service_name === Core::Environment::Ext::FALLBACK_SERVICE_NAME
```
lloeki, line 445: Is that `===` (a.k.a match operator) intended?

lib/datadog/core/configuration/option.rb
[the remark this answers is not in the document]
ivoanjo: I don't think we should make the separator more configurable -- this way we push everyone to use commas, and they need to go out of their way if they really need to support others (but can, if needed).

Thus, the complexity is moved from the core (supporting arbitrary separators) to the setting (if you wanna do weird stuff, you pay the cost of doing weird stuff, locally).

lib/datadog/core/configuration/option.rb
ivoanjo: > Meta-suggestion: maybe it should go as a review comment in "Finish your review"? but then it's not a "threaded" discussion I guess (subsequent discussion would happen in the global PR discussion tab).

Yes, it's for the whole review, and I only picked elsewhere because I find people sometimes miss the comment posted with the review itself, and I thought this comment was important enough not to be missed.

Plus, as you pointed out, it's really annoying to not have threading for those comments.

So I'm just working around GitHub UX limitations >_>

lib/datadog/appsec/configuration/settings.rb
GustavoCaso: This is legacy from when the appsec setting was not using the core configuration code. 

We can remove all those constants on a separate PR 😄

lib/datadog/core/configuration/settings.rb
GustavoCaso: I believe it was an issue with some tests. I had to change it to make them pass. 

I reverted to using `equal?` https://github.com/DataDog/dd-trace-rb/pull/2970/commits/87d739d1526c9ef3c143946a641f9e909eeedbb7

lib/datadog/appsec/configuration/settings.rb
GustavoCaso: I had to wrap it on `Core::Utils::SafeDup.frozen_or_dup(...)` because a test validates that when modifying the default value doesn't modify it by reference. 

```ruby
describe '#writer_options' do
  subject(:writer_options) { settings.tracing.test_mode.writer_options }

  it { is_expected.to eq({}) }

  context 'when modified' do
    it 'does not modify the default by reference' do
      settings.tracing.test_mode.writer_options[:foo] = :bar
      expect(settings.tracing.test_mode.writer_options).to_not be_empty
      expect(settings.tracing.test_mode.options[:writer_options].default_value).to be_empty
    end
  end
end
```

The options were like this before our changes:
```ruby
option :writer_options do |o|
  o.default { {} }
end
```

This is how they are after our changes:
```ruby
option :writer_options do |o|
  o.type :hash
  o.default({})
end
```

So I believe this was an oversight on my part in which using a default value as a block also meant that calling `default_value` would return a hash every time. To fix that particular spec, I had to use `Core::Utils::SafeDup.frozen_or_dup`

@lloeki suggested that we should be freezing the configuration value it gets set, so any modification to the configuration value should have to use `dup` first

lib/datadog/core/configuration/option.rb
```diff
@@ -102,6 +191,29 @@ def context_eval(&block)
...
+          precedence = nil
+
+          if definition.env && ENV[definition.env]
+            value = coerce_env_variable(ENV[definition.env])
+            precedence = Precedence::PROGRAMMATIC
+          end
+
+          if value.nil? && definition.deprecated_env && ENV[definition.deprecated_env]
+            value = coerce_env_variable(ENV[definition.deprecated_env])
+            precedence = Precedence::PROGRAMMATIC
```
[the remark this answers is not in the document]
GustavoCaso: That comment should be changed. 

If a customer has modified the configuration value via ENV variable, it has been configured programmatically since the customer is not using the default value. 

For example: 

```ruby
option :enabled do |o|
  o.type :bool
  o.env 'DD_APPSEC_ENABLED'
  o.default false
end
```

If the customer changes the value using `Datadog.configuration.appsec.enabled = true` or via `DD_APPSEC_ENABLED`, we would want the helper function `using_default?` to return false `Datadog.configuration.appsec.using_default?(:enabled) #=> false`, since the intent of `DEFAULT = [0, :default].freeze` is to signal the customer hasn't set the value in anyway

lib/datadog/core/configuration/option.rb
```diff
@@ -94,6 +96,93 @@ def default_precedence?
...
+            value.to_i
+          when :float
+            # DEV-2.0: Change to a more strict coercion method. Float(value).
+            value.to_f
+          when :array
+            values = if value.include?(',')
+                       value.split(',')
+                     else
+                       value.split(' ') # rubocop:disable Style/RedundantArgument
+                     end
```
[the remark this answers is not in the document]
GustavoCaso: > Since the default is not to support the spaces-variant?

Where is that defined? 

The code for `env_to_list` did not define any default value for `def env_to_list(var, default = [], comma_separated_only:, deprecation_warning: true)`

Doing a search on [`comma_separated_only`](https://github.com/search?q=repo%3ADataDog%2Fdd-trace-rb%20comma_separated_only&type=code) shows three instances of `true`, and four instances of `false`.

I'm happy to remove the if condition, I just to make sure to understand if we have to

lib/datadog/core/configuration/option_definition.rb
```diff
@@ -74,11 +85,23 @@ def depends_on(*values)
...
+          def env_parser(&block)
+            @env_parser = block
+          end
+
           def default(value = nil, &block)
             @default = block || value
           end
 
-          def experimental_default_proc(&block)
+          def experimental_default_proc(block)
```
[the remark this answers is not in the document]
GustavoCaso: Good question 🤦

lib/datadog/core/configuration/option_definition.rb
```diff
@@ -112,23 +135,28 @@ def setter(&block)
             @setter = block
           end
 
-          def type(value = nil)
+          def type(value, type_options = {})
```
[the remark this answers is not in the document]
GustavoCaso: Great suggestion 😄

lib/datadog/core/configuration/settings.rb
GustavoCaso: > As for having the parser as a block for env, no strong feelings

I think the decision of having the parser as an optional block argument for `env` is an interesting option, since it allows for better organisation, but could lead to repetition if we have a `env` and `deprecated_env`  option that has a custom parsing logic. 

```ruby
option :test do |o|
  o.env HELLO_WORLD do |value|
    custom logic 
  end
  o.deprecated_env HELLO_WORLD_OLD do |value|
    custom logic 
  end
end
```

Another point of view, is that it allow for individual parsing logic f the deprecated env var has a different format than new one for example. 

Another thing we can do (@ivoanjo might love it 😉) is remove `deprecated_env` and add support for defining multiple env + return the `env_var` to the `env` block. That would allow to move the complexity into the env block including the deprecation warning

```ruby
option :foo do |o|
  o.env ['NEW_ENV_VAR', 'OLD_ENV_VAR'] do |env_var, value|
    if env_var == 'OLD_ENV_VAR'
      # log deprecation
      # custom parsing logic if needed 
    else 
      # parse new env var
    end
  end
end
``` 

What do you guys think?

lib/datadog/core/configuration/option.rb
```diff
@@ -94,6 +96,93 @@ def default_precedence?
...
+          end
+
+          if raise_error
+            error_msg = if @definition.type_options[:nilable]
+                          "The option #{@definition.name} support this type `#{@definition.type}` "\
+                                      "and `nil` but the value provided is #{value.class}"
+                        else
+                          "The option #{@definition.name} support this type `#{@definition.type}` "\
+                          "but the value provided is #{value.class}"
+                        end
```
ivoanjo, line 155: Right now this is not particularly user-friendly:

```
$ cat example.rb 
require 'ddtrace'

Datadog.configure { |c| c.profiling.enabled = 'potato' }
$ bundle exec ruby example.rb 
Traceback (most recent call last):
  7: from example.rb:3:in `<main>'
  6: from lib/datadog/tracing/contrib/extensions.rb:64:in `configure'
  5: from lib/datadog/core/configuration.rb:82:in `configure'
  4: from example.rb:3:in `block in <main>'
  3: from lib/datadog/core/configuration/options.rb:47:in `block in default_helpers'
  2: from lib/datadog/core/configuration/options.rb:70:in `set_option'
  1: from lib/datadog/core/configuration/option.rb:54:in `set'
lib/datadog/core/configuration/option.rb:157:in `validate_type': The option enabled support this type `bool` but the value provided is String (ArgumentError)
```

I suggest approaching this from the POV of a user getting this error and what they'd need to fix it:

```
The setting `.profiling.enabled` inside the `Datadog.configure` block expects a bool, but a `String` was provided ('potato'). Please update your `configure` block. Alternatively, you can disable this validation using `...env var here...`. For help, see <link to section of doc where we can tell users to get in touch with support>.
```

I did a quick experiment and I could inject the name of the containing group (although not the full path) by doing

```diff
           # e.g. `settings :foo { option :bar }` --> `config.foo.bar`
           # @param [Symbol] name option name. Methods will be created based on this name.
           def settings(name, &block)
-            settings_class = new_settings_class(&block)
+            settings_class = new_settings_class(name, &block)
 
             option(name) do |o|
               o.default { settings_class.new }
@@ -39,8 +39,9 @@ module Datadog
 
           private
 
-          def new_settings_class(&block)
+          def new_settings_class(name = nil, &block)
             Class.new { include Configuration::Base }.tap do |klass|
+              klass.instance_variable_set(:@name, name)
               klass.instance_eval(&block) if block
```

which is why I was able to prototype `.profiling.enabled`.

ivoanjo (commented):
More comments...

lib/datadog/appsec/configuration/settings.rb
ivoanjo: I think freezing instead of dup'ing seems reasonable as well, although I wonder if somewhere we may be doing something weird that doesn't like it (but perhaps in that situation it'd be worth fixing that code instead of not freezing)

lib/datadog/core/configuration/option.rb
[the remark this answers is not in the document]
ivoanjo: I see. Seems reasonable -- either updating the comment or perhaps introducing a new environment variable level seems OK.

lib/datadog/core/configuration/option.rb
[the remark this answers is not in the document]
ivoanjo: I mean the default for the DSL should be to not support it, because in my view the "spaces separator" is the weird use-case, so it should require extra work on the part of the settings that want it.

(As I mention below, accepting spaces on top of commas has the potential of creating more inconsistency across datadog libraries)

lib/datadog/core/configuration/option_definition.rb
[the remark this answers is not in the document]
ivoanjo: Should there be some test coverage to make sure we don't have a regression on this?

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -91,18 +99,22 @@ def self.extended(base)
...
+                    values = value.split(',')
 
+                    values.map! do |v|
+                      v.gsub!(/\A[\s,]*|[\s,]*\Z/, '')
+
+                      v.empty? ? nil : v
+                    end
+                    values.compact!
+                    values
+                  end
```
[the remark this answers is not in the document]
GustavoCaso: Great point @ivoanjo will change the code to reflect that

lib/datadog/tracing/contrib/faraday/middleware.rb
```diff
@@ -64,7 +64,9 @@ def annotate!(span, env, options)
           end
 
           def handle_response(span, env, options)
-            span.set_error(["Error #{env[:status]}", env[:body]]) if options.fetch(:error_handler).call(env)
+            if options[:error_handler] && options[:error_handler].call(env)
+              span.set_error(["Error #{env[:status]}", env[:body]])
+            end
```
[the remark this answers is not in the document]
GustavoCaso: I found the problem. I accidentally removed the way to define `experimental_default_proc` inline. 

https://github.com/DataDog/dd-trace-rb/pull/2970/files/87d739d1526c9ef3c143946a641f9e909eeedbb7#diff-7e0fbcca4a616a5c1dae88b8c33fd4f534021639f9370b0c9496281ec25c9ae1L124

 On the PR that extracted just the logic out https://github.com/DataDog/dd-trace-rb/pull/2983, we no loger have that issue 😄

lib/datadog/core/configuration/option_definition.rb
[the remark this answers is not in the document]
GustavoCaso: I found the problem. I accidentally removed the way to define `experimental_default_proc` inline. 

https://github.com/DataDog/dd-trace-rb/pull/2970/files/87d739d1526c9ef3c143946a641f9e909eeedbb7#diff-7e0fbcca4a616a5c1dae88b8c33fd4f534021639f9370b0c9496281ec25c9ae1L124

 On the PR that extracted just the logic out https://github.com/DataDog/dd-trace-rb/pull/2983, we no longer have that issue 😄

lib/datadog/core/configuration/option.rb
GustavoCaso: Done on https://github.com/DataDog/dd-trace-rb/pull/2983

lib/datadog/core/configuration/option_definition.rb
delner: I like `env_var` but I will defer to the team. It's not something I feel particularly strongly about.

GustavoCaso:
This PR has been split into smaller ones to facilitate the review process.

https://github.com/DataDog/dd-trace-rb/pull/2983
https://github.com/DataDog/dd-trace-rb/pull/2987
https://github.com/DataDog/dd-trace-rb/pull/2988

marcotc:
@GustavoCaso, because this PR was split, can this specific one be closed then?

PR #2972, opened 2023-07-14
Add support for unsetting options

marcotc:
For Remote Configuration, support for Option precedence was introduce in #2915.

In addition to supporting some options having higher precedence than others, it's also necessary to allow for options to be removed, based on precedence.
The tricky part is that the previous value has to be restored.

The main use-case is for when Remote Configuration is enabled in the UI, but then disabled while the application is running. In this case, it's desirable to return the application to the same state it would have if Remote Configuration was never enabled.

It is unfortunately, not enough for Remote Configuration itself to store the previous value right before it sets a new one, given the `Datadog.configuration` object can be modified by the user at any point of the application lifecycle.

This PR keeps track of all values set, one per precedence, and allows us to unset a specific precedence, restoring to a previous value if applicable.


spec/datadog/core/configuration/option_spec.rb
```diff
@@ -221,6 +223,98 @@
...
+                  expect(get).to eq(@highest_value)
+                  expect(option.send(:precedence_set)).to eq(state[0])
+                end
+              end
+            end
+          end
+        end
+      end
+    end
+  end
```
GustavoCaso, line 316:
```suggestion
{
  empty_state: [],
  remote_configuration: [Datadog::Core::Configuration::Option::Precedence::REMOTE_CONFIGURATION],
  programatic: [Datadog::Core::Configuration::Option::Precedence::PROGRAMMATIC],
  default: [Datadog::Core::Configuration::Option::Precedence::DEFAULT],
  remote_and_programattic: [
    Datadog::Core::Configuration::Option::Precedence::REMOTE_CONFIGURATION,
    Datadog::Core::Configuration::Option::Precedence::PROGRAMMATIC
  ],
  remote_and_default: [
    Datadog::Core::Configuration::Option::Precedence::REMOTE_CONFIGURATION,
    Datadog::Core::Configuration::Option::Precedence::DEFAULT
  ],
  programatice_and_default: [
    Datadog::Core::Configuration::Option::Precedence::PROGRAMMATIC,
    Datadog::Core::Configuration::Option::Precedence::DEFAULT
  ],
  all: [
    Datadog::Core::Configuration::Option::Precedence::REMOTE_CONFIGURATION,
    Datadog::Core::Configuration::Option::Precedence::PROGRAMMATIC,
    Datadog::Core::Configuration::Option::Precedence::DEFAULT
  ]
}.each do |name, precedences|
    context "for #{name} set" do
      before do
        # See this Option with many values set a different precedences.
        precedences.each do |precedence|
          # For convenience, the option value is set to the same object as the precedence.
          value = precedence

          @highest_value ||= value
          option.set(value, precedence: precedence)
        end
      end

      # Far all scenarios, try to remove each precedence and assert the correct behavior.
      Datadog::Core::Configuration::Option::Precedence::LIST.each do |precedence|
        context "unsetting '#{precedence[1]}'" do
          subject!(:unset) { option.unset(precedence) }
          let(:precedence) { precedence }
          let(:get) { option.get }

          if precedences.empty?
            context 'when no value is set' do
              it 'resets the option' do
                expect(get).to eq(default)
                expect(option.send(:precedence_set)).to eq(Datadog::Core::Configuration::Option::Precedence::DEFAULT)
              end
            end
          elsif precedence < precedences[0]
            context 'when a value with lower precedence is unset' do
              it 'does not modify the option value' do
                expect(get).to eq(@highest_value)
                expect(option.send(:precedence_set)).to eq(precedences[0])
              end
            end
          elsif precedence == precedences[0]
            context 'the highest precedence value is unset' do
              if precedences.size == 1
                context 'removing the only value set' do
                  it 'resets the option' do
                    expect(get).to eq(default)
                    expect(option.send(:precedence_set)).to eq(Datadog::Core::Configuration::Option::Precedence::DEFAULT)
                  end
                end
              else
                it 'falls back to lower precedence value' do
                  expect(get).to eq(precedences[1])
                  expect(option.send(:precedence_set)).to eq(precedences[1])
                end
              end
            end
          elsif precedence > precedences[0]
            context 'when a nonexistent value with higher precedence is unset' do
              it 'does not modify the option value' do
                expect(get).to eq(@highest_value)
                expect(option.send(:precedence_set)).to eq(precedences[0])
              end
            end
          end
        end
      end
    end
  end
end
```

Cosmetic change comment 😄 

I find it more readable to see the list of combination that we are running on the tests upfront than for it to be dynamically generated. 

If we add a new precedence value, I know it has the issue that we manually have to come to add the combination. Not sure how often we do that, but for starters, I would go for readability than automation 😄 

As being a cosmetic suggestion is not a blocker for me

GustavoCaso:
Should calling `reset` on the option also reset the `@value_per_precedence` state?

spec/datadog/core/configuration/option_spec.rb
marcotc: I tried this at first, but thought it would look silly if we added one more precedence (say ENV_VAR), because it grows non-linearly.

But you know what, we don't need to test every combination, as a lot of them are redundant.
For 3 possible values, the redundancy is not too bad, but for 4 or more it there's a lot of waste.

Let me take in these changes.

marcotc:
> Should calling `reset` on the option also reset the `@value_per_precedence` state?

Yes, otherwise a `option.reset; option.set(new_value, precedence: DEFAULT); assert option.get == new_value` can fail precedence set before `reset` was higher than `DEFAULT`. (there's a test that fails if I remove the `#value_per_precedence` reset, for example)

lib/datadog/core/configuration/option.rb
```diff
@@ -38,22 +58,55 @@ def initialize(definition, context)
...
+              "Option '#{definition.name}' not changed to '#{value}' (precedence: #{precedence.name}) because the higher " \
+                "precedence value '#{@value}' (precedence: #{@precedence_set.name}) was already set."
             end
 
+            # But if it happens, we have to store the lower precedence value `value`
+            # because it's possible to revert to it by `#unset`ting
+            # the existing, higher-precedence value.
+            # Effectively, we always store one value pre precedence.
+            old_value = @value_per_precedence[precedence]
+            @value_per_precedence[precedence] = context_exec(value, old_value, &definition.setter)
```
GustavoCaso, line 75: We execute the setter block twice when setting lower precedence, once here and the other on line 98, when we unset and find the closer precedence value store and restore the option to the previous value. 

This could become an issue if the setter blocks logic runs expensive code or modify the internal state.

marcotc: Well pointed out!
Actually this would be a correctness issue, given `setter` would be run twice for the same value, probably yield unexpected results.

marcotc: Fixed it, thank you again!

GustavoCaso (approved):
Great work @marcotc

lib/datadog/core/configuration/option.rb
```diff
@@ -8,17 +8,33 @@ module Configuration
...
-          PROGRAMMATIC = [1, :programmatic].freeze
+          PROGRAMMATIC = Value.new(1, :programmatic).freeze
 
           # Configuration that comes either from environment variables,
           # or fallback values.
-          DEFAULT = [0, :default].freeze
+          DEFAULT = Value.new(0, :default).freeze
+
+          # All precedences, sorted from highest to lowest
+          LIST = [REMOTE_CONFIGURATION, PROGRAMMATIC, DEFAULT].freeze
```
ivoanjo, line 37: Minor: Maybe call `sort` here to enforce that the list follows the defined order? 

```suggestion
          LIST = [REMOTE_CONFIGURATION, PROGRAMMATIC, DEFAULT].sort.reverse.freeze
```

lib/datadog/core/configuration/option.rb
```diff
@@ -38,22 +58,54 @@ def initialize(definition, context)
...
-              "Option '#{definition.name}' not changed to '#{value}' (precedence: #{precedence[1]}) because the higher " \
-                "precedence value '#{@value}' (precedence: #{@precedence_set[1]}) was already set."
+              "Option '#{definition.name}' not changed to '#{value}' (precedence: #{precedence.name}) because the higher " \
+                "precedence value '#{@value}' (precedence: #{@precedence_set.name}) was already set."
             end
 
+            # But if it happens, we have to store the lower precedence value `value`
+            # because it's possible to revert to it by `#unset`ting
+            # the existing, higher-precedence value.
+            # Effectively, we always store one value pre precedence.
```
ivoanjo, line 73: This is indeed a weird situation, but I think it makes the info message be a bit... misleading. Because it looks like the setting was ignored, but it's possible that it still may be acted on in the future.

So I would maybe suggest updating the message to mention this? Something along the lines of "Warning: Not changing option XXX to value YYY because it's being controlled via remote configuration. YYY will be used if remote configuration is disabled".

This makes the warning message less general than it currently is, but right now there shouldn't be any other way for this to happen, right? You can't really set a "default" after-the-fact, so the only option here seems to be programmatic being set after remote configuration. (And we can always add a spec that breaks if this assumption fails, so we don't forget to change something, if e.g. we add more precedence levels)

lib/datadog/core/configuration/option.rb
```diff
@@ -38,22 +58,54 @@ def initialize(definition, context)
...
+            end
+
+            # If no value is left to fall back on, reset this option
+            reset
           end
+
+          # ... otherwise, we are either unsetting a higher precedence value that is not
+          # yet set, thus there's nothing to do; or we are unsetting a lower precedence
+          # value, which also does not change the current value.
         end
```
ivoanjo, line 109: This seems to be a bit more generic than it needs to be. Since the use case of unsetting is disabling remote configuration, have you considered only allowing those values to be unset?

ivoanjo (commented):
:cold_sweat: :cold_sweat: :cold_sweat: Apologies for coming late into a one-week-old PR and then adding a bunch of comments.

I still wanted to leave them; obviously they don't block anything (it's already merged even!) but food for thought, hopefully :P

PR #2973, opened 2023-07-14
RC:Add apply state and apply error

marcotc:
This PR adds the `apply_state` and `apply_error` fields to the configuration content payload we sent back to the agent.

This allows downstream systems to know if the configuration was correctly applied, and if not, why.

This is used to display to users that Remote Configuration was correctly applied, or if it was not, what was the reason.

### DIFF
--- a/lib/datadog/core/remote/configuration/content.rb
+++ b/lib/datadog/core/remote/configuration/content.rb
@@ -41,6 +41,7 @@ module Datadog
           # Sets this configuration as successfully applied.
           def applied
             @apply_state = ApplyState::ACKNOWLEDGED
+            @apply_error = nil
           end
 
           # Sets this configuration as not successfully applied, with
--- a/lib/datadog/core/remote/configuration/content.rb
+++ b/lib/datadog/core/remote/configuration/content.rb
@@ -18,12 +18,14 @@ module Datadog
             end
           end
 
-          attr_reader :path, :data, :hashes
+          attr_reader :path, :data, :hashes, :apply_state, :apply_error
           attr_accessor :version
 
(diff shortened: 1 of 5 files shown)


lib/datadog/core/remote/configuration/content.rb
```diff
@@ -36,6 +38,30 @@ def length
...
+          module ApplyState
+            # Default state of configurations.
+            # Set until the component consuming the configuration has acknowledged it was applied.
+            UNACKNOWLEDGED = 1
+
+            # Set when the configuration has been successfully applied.
+            ACKNOWLEDGED = 2
+
+            # Set when the configuration has been unsuccessfully applied.
+            ERROR = 3
```
ivoanjo, line 62: Minor: I'm curious, do we need these integers in the protocol? Could they be symbols (:error, :unacknowledged, ...) instead?

lib/datadog/core/remote/configuration/content.rb
```diff
@@ -36,6 +38,30 @@ def length
             @length ||= @data.size
           end
 
+          # Sets this configuration as successfully applied.
+          def applied
+            @apply_state = ApplyState::ACKNOWLEDGED
```
TonyCTHsu, line 43: Should we also clear the `@apply_error` ?

TonyCTHsu (approved):

lib/datadog/core/remote/configuration/content.rb
marcotc: They are numbers, as part of the protocol. Backend expects these specific values.

PR #2974, opened 2023-07-17
Fix manual log injection for 128 bit trace_id

TonyCTHsu:
**What does this PR do?**

https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/ruby/#for-logging-in-ruby-applications

We exposed `Datadog::Tracing.correlation#trace_id` as public API, which user ends up with different injection result from manual log injection(returning the entire 128 bit trace ID).

This PR enforce the coercion of lower 64 bits trace ID on this API to ensure the correctness of manual log injection.

### DIFF
--- a/lib/datadog/tracing/correlation.rb
+++ b/lib/datadog/tracing/correlation.rb
@@ -23,7 +23,6 @@ module Datadog
           :span_resource,
           :span_service,
           :span_type,
-          :trace_id,
           :trace_name,
           :trace_resource,
           :trace_service,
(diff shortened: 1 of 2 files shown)


lib/datadog/tracing/correlation.rb
```diff
@@ -23,7 +23,6 @@ class Identifier
           :span_resource,
           :span_service,
           :span_type,
-          :trace_id,
```
marcotc, line 26: I think removing `trace_id` from the public api of this class is problematic.

I think a better solution is to allow `trace_id` to continue to be called, but return a string when `Datadog.configuration.tracing.trace_id_128_bit_logging_enabled` is true. This means users that opt into this new feature get different behavior, but only after enabling this option setting.

lib/datadog/tracing/correlation.rb
```diff
@@ -65,22 +64,20 @@ def to_log_format
...
-            end
+        # DEV-2.0: This public method was returning an Integer, but with 128 bit trace id it would return a String.
+        def trace_id
+          if Datadog.configuration.tracing.trace_id_128_bit_logging_enabled &&
+              !Tracing::Utils::TraceId.to_high_order(@trace_id).zero?
+            Kernel.format('%032x', @trace_id)
+          else
+            Tracing::Utils::TraceId.to_low_order(@trace_id)
+          end
         end
```
marcotc, line 81: I think this calculation should be repeated for multiple calls of `Identifier#trace_id`.
I suggest either moving this to `#initialize` and saving the result to `trace_id`, or caching it here `@trace_id ||= if Datadog.configuration.tracing.trace_id_128_bit_logging_enabled...`.

marcotc (approved):

PR #2975, opened 2023-07-17
Allow blocking response configuration

lloeki:
**What does this PR do?**

Allow blocking response configuration:
- change status code
- change response body
- specify redirect target (location header)

**Motivation**

Enable users to customise attack response.

**Additional Notes**

- Initial implementation operates via static configuration.
- Then remote configuration support can be added by altering configuration fetching in `AppSec::Response`

**How to test the change?**

There will be specs

### DIFF
--- a/lib/datadog/appsec/configuration/settings.rb
+++ b/lib/datadog/appsec/configuration/settings.rb
@@ -95,6 +95,62 @@ module Datadog
                 end
               end
 
+              settings :block do
+                # HTTP status code to block with
+                option :status do |o|
+                  o.default { 403 }
+                end
+
+                # only applies to redirect status codes
+                option :location do |o|
+                  o.setter { |v| URI(v) unless v.nil? }
+                end
+
+                # only applies to non-redirect status codes with bodies
+                option :templates do |o|
+                  o.default do
+                    json = ENV.fetch(
+                      'DD_APPSEC_HTTP_BLOCKED_TEMPLATE_JSON',
+                      :json
+                    )
+
+                    html = ENV.fetch(
+                      'DD_APPSEC_HTTP_BLOCKED_TEMPLATE_HTML',
+                      :html
+                    )
+
+                    text = ENV.fetch(
+                      'DD_APPSEC_HTTP_BLOCKED_TEMPLATE_TEXT',
+                      :text
+                    )
+
+                    {
+                      'application/json' => json,
+                      'text/html' => html,
+                      'text/plain' => text,
+                    }
+                  end
+                  o.setter do |v|
+                    next if v.nil?
+
+                    # TODO: should merge with o.default to allow overriding only one mime type
+
+                    v.each do |k, w|
+                      case w
+                      when :json, :html, :text
+                        next
+                      when String, Pathname
+                        next if File.exist?(w.to_s)
+
+                        raise(ArgumentError, "appsec.templates.#{k}: file not found: #{w}")
+                      else
+                        raise ArgumentError, "appsec.templates.#{k}: unexpected value: #{w.inspect}"
+                      end
+                    end
+                  end
+                end
+              end
+
               settings :track_user_events do
                 option :enabled do |o|
                   o.default do
--- a/lib/datadog/appsec/response.rb
+++ b/lib/datadog/appsec/response.rb
@@ -33,10 +33,16 @@ module Datadog
 
           Datadog.logger.debug { "negotiated response content type: #{content_type}" }
 
+          headers = { 'Content-Type' => content_type }
+          headers['Location'] = location.to_s if redirect?
+
+          body = []
+          body << content(content_type) unless redirect?
+
           Response.new(
-            status: 403,
-            headers: { 'Content-Type' => content_type },
-            body: [Datadog::AppSec::Assets.blocked(format: CONTENT_TYPE_TO_FORMAT[content_type])]
+            status: status,
+            headers: headers,
+            body: body,
           )
         end
 
@@ -49,6 +55,7 @@ module Datadog
         }.freeze
 
         DEFAULT_CONTENT_TYPE = 'application/json'
+        REDIRECT_STATUS = [301, 302, 303, 307, 308].freeze
 
         def content_type(env)
           return DEFAULT_CONTENT_TYPE unless env.key?('HTTP_ACCEPT')
@@ -66,6 +73,37 @@ module Datadog
           DEFAULT_CONTENT_TYPE
         rescue Datadog::AppSec::Utils::HTTP::MediaRange::ParseError
           DEFAULT_CONTENT_TYPE
+        end
+
+        def status
+          Datadog.configuration.appsec.block.status
+        end
+
+        def redirect?
+          REDIRECT_STATUS.include?(status)
+        end
+
+        def location
+          Datadog.configuration.appsec.block.location
+        end
+
+        def content(content_type)
+          setting = Datadog.configuration.appsec.block.templates[content_type]
+
+          case setting
+          when :html, :json, :text
+            Datadog::AppSec::Assets.blocked(format: setting)
+          when String, Pathname
+            path = setting.to_s
+
+            cache[path] ||= (File.open(path, 'rb', &:read) || '')
+          else
+            raise ArgumentError, "unexpected type: #{content_type.inspect}"
+          end
+        end
+
+        def cache
+          @cache ||= {}
         end
       end
     end
--- a/sig/datadog/appsec/response.rbs
+++ b/sig/datadog/appsec/response.rbs
@@ -3,9 +3,10 @@ module Datadog
     class Response
       attr_reader status: ::Integer
       attr_reader headers: ::Hash[::String, ::String]
-      attr_reader body: ::Array[untyped]
+      attr_reader body: ::Array[::String]
 
-      def initialize: (status: ::Integer, ?headers: ::Hash[::String, ::String], ?body: ::Array[untyped]) -> void
+
+      def initialize: (status: ::Integer, ?headers: ::Hash[::String, ::String], ?body: ::Array[::String]) -> void
       def to_rack: () -> ::Array[untyped]
       def to_sinatra_response: () -> ::Sinatra::Response
       def to_action_dispatch_response: () -> ::ActionDispatch::Response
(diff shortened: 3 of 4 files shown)


lib/datadog/appsec/configuration/settings.rb
```diff
@@ -95,6 +95,62 @@ def self.add_settings!(base)
...
 
+              settings :block do
+                # HTTP status code to block with
+                option :status do |o|
+                  o.default { 403 }
+                end
+
+                # only applies to redirect status codes
+                option :location do |o|
+                  o.setter { |v| URI(v) unless v.nil? }
```
lloeki, line 106: Steep insists that the `v` in `URI(v)` can be `nil` here

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -95,6 +95,62 @@ def self.add_settings!(base)
...
+                      'text/html' => html,
+                      'text/plain' => text,
+                    }
+                  end
+                  o.setter do |v|
+                    next if v.nil?
+
+                    # TODO: should merge with o.default to allow overriding only one mime type
+
+                    v.each do |k, w|
```
lloeki, line 138: Steep insists that the `v` in `v.each` can be `nil` here

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -95,6 +95,62 @@ def self.add_settings!(base)
...
+                    {
+                      'application/json' => json,
+                      'text/html' => html,
+                      'text/plain' => text,
+                    }
+                  end
+                  o.setter do |v|
+                    next if v.nil?
+
+                    # TODO: should merge with o.default to allow overriding only one mime type
```
lloeki, line 136: Without this, if `templates` is set it overrides the whole content, e.g the following removes `text/plain` and `application/json`

```
    c.appsec.block.templates = {
      'text/html' => 'block.html',
    }
```

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -95,6 +95,62 @@ def self.add_settings!(base)
...
+
+                    v.each do |k, w|
+                      case w
+                      when :json, :html, :text
+                        next
+                      when String, Pathname
+                        next if File.exist?(w.to_s)
+
+                        raise(ArgumentError, "appsec.templates.#{k}: file not found: #{w}")
```
lloeki, line 145: Checking for file existence early, so that it doesn't burn into flames at runtime.

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -95,6 +95,62 @@ def self.add_settings!(base)
...
+                      else
+                        raise ArgumentError, "appsec.templates.#{k}: unexpected value: #{w.inspect}"
```
lloeki, line 147: Checking for consistency early, so that it doesn't burn into flames at runtime.

lib/datadog/appsec/response.rb
```diff
@@ -67,6 +74,37 @@ def content_type(env)
         rescue Datadog::AppSec::Utils::HTTP::MediaRange::ParseError
           DEFAULT_CONTENT_TYPE
         end
+
+        def status
+          Datadog.configuration.appsec.block.status
```
lloeki, line 79: We can priorise the RC configuration value here

lib/datadog/appsec/response.rb
```diff
@@ -67,6 +74,37 @@ def content_type(env)
...
+        end
+
+        def redirect?
+          REDIRECT_STATUS.include?(status)
+        end
+
+        def location
+          Datadog.configuration.appsec.block.location
```
lloeki, line 87: We can priorise the RC configuration value here

lib/datadog/appsec/response.rb
```diff
@@ -67,6 +74,37 @@ def content_type(env)
...
+        end
+
+        def content(content_type)
+          setting = Datadog.configuration.appsec.block.templates[content_type]
```
lloeki, line 91: We can priorise the RC configuration value here

lib/datadog/appsec/response.rb
```diff
@@ -67,6 +74,37 @@ def content_type(env)
...
+
+          case setting
+          when :html, :json, :text
+            Datadog::AppSec::Assets.blocked(format: setting)
+          when String, Pathname
+            path = setting.to_s
+
+            cache[path] ||= (File.open(path, 'rb', &:read) || '')
+          else
+            raise ArgumentError, "unexpected type: #{content_type.inspect}"
```
lloeki, line 101: Ideally that should not happen given the early checks done on configuration.

lib/datadog/appsec/response.rb
```diff
@@ -67,6 +74,37 @@ def content_type(env)
...
+          end
+        end
+
+        def cache
+          @cache ||= {}
```
lloeki, line 106: That cache is in the singleton, but I don't think it's too bad in this case, even when testing.

lloeki (commented):
Left some comments for reviewers to peruse

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -95,6 +95,62 @@ def self.add_settings!(base)
                 end
               end
 
+              settings :block do
+                # HTTP status code to block with
+                option :status do |o|
+                  o.default { 403 }
```
GustavoCaso, line 101: No need for this to be a block 😄

lib/datadog/appsec/configuration/settings.rb
lloeki: I mean, `URI(v) unless v.nil?` should make it so that `v` in `URI(v)` is not, but Steep still believes it can.

lib/datadog/appsec/configuration/settings.rb
lloeki: Ah that was aimed at a few lines below. Similarly as above `next if v.nil?` ensures that afterwards `v` is not `nil` but Steep still complains.

lib/datadog/appsec/configuration/settings.rb
lloeki: Ah, true.

lib/datadog/appsec/response.rb
```diff
@@ -67,6 +74,37 @@ def content_type(env)
...
+          when String, Pathname
+            path = setting.to_s
+
+            cache[path] ||= (File.open(path, 'rb', &:read) || '')
+          else
+            raise ArgumentError, "unexpected type: #{content_type.inspect}"
+          end
+        end
+
+        def cache
```
marcotc, line 105: Can you document in this method what characteristics this cache has?
I see it's in a singleton, so `@cache` will only ever grow, but it looks like it's only based on configuration changes, which require direct user intervention.

I'm suggesting that we write something in code so that we don't have to dig out these answers in the future.

GustavoCaso:
@lloeki here is an alternative suggestion for configuration of this feature.

```ruby
settings :block do
  # HTTP status code to block with
  option :status do |o|
    o.type :int
    o.default 403
  end

  # only applies to redirect status codes
  option :location do |o|
    o.type :string
    o.setter { |v| URI(v) unless v.nil? }
  end

  # only applies to non-redirect status codes with bodies
  settings :templates do
    option :html do |o|
      o.env 'DD_APPSEC_HTTP_BLOCKED_TEMPLATE_HTML'
      o.type :string, nilable: true
      o.setter do |value|
        unless File.exist?(value)
          raise(ArgumentError, "appsec.templates.html: file not found: #{value}")
        end
      end
    end

    option :json do |o|
      o.env 'DD_APPSEC_HTTP_BLOCKED_TEMPLATE_JSON'
      o.type :string, nilable: true
      o.setter do |value|
        unless File.exist?(value)
          raise(ArgumentError, "appsec.templates.json: file not found: #{value}")
        end
      end
    end

    option :html do |o|
      o.env 'DD_APPSEC_HTTP_BLOCKED_TEMPLATE_TEXT'
      o.type :string, nilable: true
      o.setter do |value|
        unless File.exist?(value)
          raise(ArgumentError, "appsec.templates.text: file not found: #{value}")
        end
      end
    end
  end
end
```

It separates each type into its own `option`. It allows us to control the setting with a single ENV variable. Also, it removes the need to have to maintain a hash for all possible templates.

The code for the moment in which we need to retrieve either the default template or the configured one the code could look like this:

```ruby
def content(content_type)
  content_format = content_type.split('/')[1]).to_sym

  raise ArgumentError, "unexpected type: #{content_type.inspect}" unless valid_format?(content_format)

  using_default = Datadog.configuration.appsec.block.templates.using_default?(content_format)

  if using_default
    Datadog::AppSec::Assets.blocked(format: content_format)
  else
    path = Datadog.configuration.appsec.block.templates.send(content_format).to_s

    cache[path] ||= (File.open(path, 'rb', &:read) || '')
  end
end
```

This new code uses `using_default?` to check if the template for that content type has been configured. We can do this because we have split each template on its separate configuration `option`.

What do you think 😄 ?

GustavoCaso:
Actually, @lloeki looking at the documentation on the DD site https://docs.datadoghq.com/security/application_security/threats/library_configuration/#configure-a-custom-blocking-page-or-payload

The documented way to configure the blocking response is `appsec.http.blocked.template.html`, and `appsec.http.blocked.template.json`. So it looks like we might have to provide those entry points for consistency

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -95,6 +95,46 @@ def self.add_settings!(base)
                 o.default DEFAULT_OBFUSCATOR_VALUE_REGEX
               end
 
+              settings :block do
+                settings :templates do
+                  option :html do |o|
+                    o.env 'DD_APPSEC_HTTP_BLOCKED_TEMPLATE_HTML'
+                    o.type :string, nilable: true
```
GustavoCaso, line 102: We need to provide `nilable` because we use  `using_default?` when checking whether we need to use the default value. 

https://github.com/DataDog/dd-trace-rb/pull/2975/files#diff-83b38af05def4a7362ff8abda84f27d84964c9fce7781eb7c4a27ecbfc235c66R77

Since `using_default?` [computes the option](https://github.com/DataDog/dd-trace-rb/blob/8cf3b72d9f5079076eaa67533e62a6906bdd62e6/lib/datadog/core/configuration/options.rb#L93), we needed to provide something that would make it work. Another solution would be to give an empty string as default.

marcotc (approved):

PR #2977, opened 2023-07-18
Telemetry:add app-client-configuration-change event

marcotc:
This PR adds the telemetry event `app-client-configuration-change`, which reports that configuration at the Client Library (`ddtrace`) has changed.
This event captures the changes themselves and a change source (called `origin` in the event).

The first use case is reporting changes from Dynamic Configuration. These have the origin `remote_config`.
The changes are a list of 2-tuples (`List<String, Integer | Boolean | String>`) representing the configuration that was just set.

One thing to noticed is that `app-client-configuration-change` is a Telemetry V2 event, but `ddtrace` currently implements Telemetry V1.
To aid with the migration to V2, the Datadog backend supports a few V2 events through the V1 pipeline. This is one of them.

This PR does not actually report the new event, only creates the API to trigger this event.

This event is crucial to Dynamic Configuration because it reports back to Datadog that the changes have been applied successfully to the current Ruby process.

### DIFF
--- a/lib/datadog/core/telemetry/client.rb
+++ b/lib/datadog/core/telemetry/client.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require_relative 'emitter'
 require_relative 'heartbeat'
 require_relative '../utils/forking'
@@ -63,6 +65,13 @@ module Datadog
           return if !@enabled || forked?
 
           @emitter.request(:'app-integrations-change')
+        end
+
+        # Report configuration changes caused by Dynamic Configuration.
+        def dynamic_configuration_change!(changes)
+          return if !@enabled || forked?
+
+          @emitter.request('app-client-configuration-change', data: { changes: changes, origin: 'remote_config' })
         end
 
         private
--- a/lib/datadog/core/telemetry/client.rb
+++ b/lib/datadog/core/telemetry/client.rb
@@ -67,8 +67,8 @@ module Datadog
           @emitter.request(:'app-integrations-change')
         end
 
-        # Report configuration changes caused by Dynamic Configuration.
-        def dynamic_configuration_change!(changes)
+        # Report configuration changes caused by Remote Configuration.
+        def client_configuration_change!(changes)
           return if !@enabled || forked?
 
           @emitter.request('app-client-configuration-change', data: { changes: changes, origin: 'remote_config' })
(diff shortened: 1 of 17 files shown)


lib/datadog/core/telemetry/client.rb
```diff
@@ -65,6 +67,13 @@ def integrations_change!
           @emitter.request(:'app-integrations-change')
         end
 
+        # Report configuration changes caused by Dynamic Configuration.
+        def dynamic_configuration_change!(changes)
```
GustavoCaso, line 71: Looks like there is a mismatch in the naming. `dynamic_configuration` and `remote_config`

spec/datadog/core/telemetry/emitter_spec.rb
```diff
@@ -74,6 +74,17 @@
...
+
+    context 'with data' do
+      subject(:request) { emitter.request(request_type, data: data) }
+      let(:data) { { changes: ['test-data'] } }
+      let(:request_type) { 'app-client-configuration-change' }
+
+      it 'sends data to the emitter' do
+        request
+        expect(http_transport).to have_received(:request).with(payload: include('test-data'), request_type: anything)
+      end
```
GustavoCaso, line 86: Should this test focus on testing that `data` gets propagated to `Datadog::Core::Telemetry::Event.new.telemetry_request`? If feels that checking against `http_transport` is skipping the real change on `Emitter`

spec/datadog/core/telemetry/emitter_spec.rb
```diff
@@ -74,6 +74,17 @@
         end
       end
     end
+
+    context 'with data' do
+      subject(:request) { emitter.request(request_type, data: data) }
+      let(:data) { { changes: ['test-data'] } }
+      let(:request_type) { 'app-client-configuration-change' }
+
+      it 'sends data to the emitter' do
```
GustavoCaso, line 83: The test description mentions emitter, but the asserting against `http_transport`.
Is that correct?

GustavoCaso (approved):
Left a few comments but overall LGTM

lib/datadog/core/telemetry/client.rb
marcotc: Renamed it to match the telemetry event.

spec/datadog/core/telemetry/emitter_spec.rb
marcotc: changed it.

spec/datadog/core/telemetry/emitter_spec.rb
marcotc: Renamed to match!

marcotc:
Thank you, @GustavoCaso!

PR #2979, opened 2023-07-18
Lint:Disable double negation cop

marcotc:
`!!` is the de facto way in Ruby to cast an Object to boolean.

We should allow this pattern in the repository.


GustavoCaso (approved):

GustavoCaso:
Thanks @marcotc

ivoanjo (approved):
+1 +1 +1 thank you! :)

PR #2907, opened 2023-06-14
Adds `DD_TRACE_REMOVE_CLIENT_SERVICE_NAMES_ENABLED`

zarirhamza:
Adds `DD_TRACE_REMOVE_CLIENT_SERVICE_NAMES_ENABLED` environment variable which defaults to `FALSE`.

When set to `TRUE`, it enables the service name changes without being forced to upgrade span attribute schema version t which may come with unwanted additional feature changes.

Adds test appropriately to test various scenarios when enabling/disabling new environment variable

Please do not merge without direct approval from @delner, but feel free for anyone to take a look at it

PR #2982, opened 2023-07-19
Implement `peer.service` and associated features

zarirhamza:
`peer.service` is now a tag that extracts its value from a subset of source tags that are unique per integration.

- Implement `peer.service` extraction logic in SpanAttributeSchema` #2953

- Replace `peer.service` tag in target integrations that create spans with `span.kind:client` or `span.kind:producer `over the old `peer.service` tag that simply used to reflect `span.service` #2943

- Remove all usages of `peer.service` from integrations that do not meet the above definition #2943

- Implement remapper for `peer.service` that allows users to define key:value pairs which remap the value of `peer.service` to a new remapped value #2958

- Add `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` so that users can enable global service name defaults to `DD_SERVICE`  #2907

Needs explicit @delner approval


lib/datadog/core/environment/variable_helpers.rb
```diff
@@ -98,6 +98,53 @@ def env_to_list(var, default = [], comma_separated_only:, deprecation_warning: t
...
+        #
+        #
+        # @param [String] var environment variable
+        # @param [Array<String>] var list of environment variables
+        # @param [Array<Object>] default the default value if the keys in `var` are not present in the environment
+        # @param [Boolean] deprecation_warning when `var` is a list, record a deprecation log when
+        #   the first key in `var` is not used.
+        # @return [Hash<String>] if the environment value is a valid list of key:value entries
+        # @return [default] if the environment value is not found
+        def env_to_hash(var, default = [], comma_separated_only:, deprecation_warning: true)
```
GustavoCaso, line 121: Thanks, @ivoanjo, for the call. 

Yes, @zarirhamza, I think having a way to convert ENV variable values into a hash is a great idea. Still , since we want to deprecate this file, let's wait for this to get merged https://github.com/DataDog/dd-trace-rb/pull/2983, and add that functionality there

lib/datadog/core/environment/variable_helpers.rb
```diff
@@ -98,6 +98,53 @@ def env_to_list(var, default = [], comma_separated_only:, deprecation_warning: t
...
+            value = ENV[var]
+
+            values = if value.include?(',') || comma_separated_only
+                       value.split(',')
+                     else
+                       value.split(' ') # rubocop:disable Style/RedundantArgument
+                     end
+
+            values_hash = Hash.new(0)
+            values.map! do |v|
```
GustavoCaso, line 133: We do not need a `map!` That modifies the existing array.

We can get away with: 
```ruby
values.each.with_object({}) do |v, hash|

  unless v.empty?
    pair = v.split(':', 2)
    hash[pair[0].to_sym] = pair[1]
  end
end
```

lib/datadog/tracing/contrib/grpc/formatting.rb
```diff
@@ -114,7 +114,7 @@ def format_resource_name(grpc_full_method)
             def extract_grpc_service(grpc_full_method)
               parts = grpc_full_method.split('/')
               if parts.length < 3
-                VALUE_UNKNOWN
+                ''
```
GustavoCaso, line 117: What is the implication of this change?

lib/datadog/core/environment/variable_helpers.rb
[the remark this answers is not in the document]
zarirhamza: I spoke to @GustavoCaso and this PR needs to be out by next week urgently. If #2970 is merged before that I will gladly rebase on top of that and make the according changes, but if that doesn't happen this PR will take priority and merge

lib/datadog/tracing/contrib/grpc/formatting.rb
zarirhamza: This is because we don't want to set the standard of setting values for tags that should be empty. Right now, we use the idea that if a tag is empty then the value doesn't exist, but if we set the 'unknown' value then, we have 2 conflicts:

- The user could for some odd reason set the actual value of the tag to 'unknown' which can create a collision
- The existing logic assumes that any tag with a value is a valid defined tag, so we have propagations of the value 'unknown' to other places where in reality, we would want to skip the value.

lib/datadog/core/environment/variable_helpers.rb
zarirhamza: Addressed above

lib/datadog/core/environment/variable_helpers.rb
```diff
@@ -98,6 +98,53 @@ def env_to_list(var, default = [], comma_separated_only:, deprecation_warning: t
...
+              v.gsub!(/\A[\s,]*|[\s,]*\Z/, '')
+
+              unless v.empty?
+                pair = v.split(':', 2)
+                values_hash[pair[0].to_sym] = pair[1]
```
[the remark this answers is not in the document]
zarirhamza: I just thought that was the standard, but yes we can remove this

lib/datadog/tracing/contrib/span_attribute_schema.rb
```diff
@@ -1,26 +1,132 @@
...
+        private_class_method :active_version
+
+        # Contains interface of methods to be implemented
+        module Base
+          REFLEXIVE_SOURCES = [Tracing::Metadata::Ext::TAG_PEER_SERVICE].freeze
+          NO_SOURCE = [].freeze
+          IMPLEMENT_ERROR = 'SpanAttributeSchema Version must implement fetch_service_name'
+
+          def self.extended(base)
+            base.private_class_method :not_empty_tag?, :set_peer_service_from_source, :filter_peer_service_sources
```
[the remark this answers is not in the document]
zarirhamza: I have to push back on these changes because we are already aware of changes coming down the pipeline which would require separate usage of `v0` and `v1`. Removing the concept entirely here would require a refactor in the very near future.

Originally, we did have differences between `set_peer_service!` but we condensed it into one base class. While it seems excessive at the moment, I think it sets up for the future very nicely and future modules (`v2` onwards) so we should keep it.

`set_peer_service!` does change the internal `span` state too.

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -445,6 +445,34 @@ def self.extended(base)
...
+              # @return [Hash]
+              option :peer_service_mapping do |o|
+                o.default do
+                  env_to_hash(
+                    Tracing::Configuration::Ext::SpanAttributeSchema::ENV_PEER_SERVICE_MAPPING,
+                    {},
+                    comma_separated_only: true
+                  )
+                end
+              end
```
GustavoCaso, line 461:
```suggestion
              option :peer_service_mapping do |o|
                o.env Tracing::Configuration::Ext::SpanAttributeSchema::ENV_PEER_SERVICE_MAPPING
                o.type :hash
                o.default({})
              end
```

After rebasing `master`, we can use the latest configuration options

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -445,6 +445,34 @@ def self.extended(base)
...
+                # Overrides default service name to global service name
+                #
+                # Allows for usage of v1 service name changes without
+                # being forced to update schema versions
+                #
+                # @default `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` environment variable, otherwise `false`
+                # @return [Boolean]
+                option :enabled do |o|
+                  o.default { env_to_bool(Tracing::Configuration::Ext::SpanAttributeSchema::ENV_GLOBAL_DEFAULT_SERVICE_NAME_ENABLED, false) }
+                end
```
GustavoCaso, line 474:
```suggestion
                option :enabled do |o|
                  o.env Tracing::Configuration::Ext::SpanAttributeSchema::ENV_GLOBAL_DEFAULT_SERVICE_NAME_ENABLED
                  o.default false
                end
```

After rebasing master, we can use the latest configuration options

lib/datadog/core/environment/variable_helpers.rb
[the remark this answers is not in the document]
zarirhamza: The above is giving me  ['key:value'] but I need {'key' => 'value'}. How about:

```Ruby
when :hash
  values = value.split(',') # By default we only want to support comma separated strings
  
  values.map! do |v|
  
    v.empty? ? nil : v
  end
  
  values.compact!
  values_hash = Hash.new(0)
  values.each do |v|
    pair = v.split(':', 2)
    values_hash[pair[0]] = pair[1]
  end
values_hash
```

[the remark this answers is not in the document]
zarirhamza: Nevermind, unrelated error. Will use original suggestion

lib/datadog/tracing/contrib/span_attribute_schema.rb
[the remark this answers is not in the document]
zarirhamza: After talking it over with @delner, we think its best to keep the module structure rather than using classes. We do not want to expose the `Base`. Using the class structure, we increase the surface area of the public API, allowing for things like `Base.set_peer_service!`. We want to be cautious against that so we decided to move forward with our module structure

lib/datadog/core/configuration/option.rb
```diff
@@ -98,6 +98,20 @@ def coerce_env_variable(value)
...
+              v.gsub!(/\A[\s,]*|[\s,]*\Z/, '')
+
+              v.empty? ? nil : v
+            end
+
+            values.compact!
+            values.each.with_object({}) do |v, hash|
+              pair = v.split(':', 2)
+              hash[pair[0]] = pair[1]
+            end
```
GustavoCaso, line 114: This new code is not tested. We need to add test cases in `spec/datadog/core/configuration/option_spec.rb`

lib/datadog/core/environment/variable_helpers.rb
```diff
@@ -98,6 +98,53 @@ def env_to_list(var, default = [], comma_separated_only:, deprecation_warning: t
...
+                pair = v.split(':', 2)
+                values_hash[pair[0]] = pair[1]
+              end
+            end
+
+            values_hash
+          else
+            default
+          end
+        end
```
GustavoCaso, line 146: We should remove this code. We are planning to deprecate this module soon, so adding new code is not the desired outcome

GustavoCaso (changes requested):
The code LGTM 😄 
But we need to make sure everything new that we add has test coverage and we remove the unused function from `lib/datadog/core/environment/variable_helpers.rb`

lib/datadog/tracing/contrib/span_attribute_schema.rb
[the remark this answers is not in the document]
GustavoCaso: I'm not the biggest fan of this approach, so here's me taking a stab at explaining why. (Note I don't consider this blocking -- e.g. feel free not to use any of this)

1. Private VS Public

> We do not want to expose the Base. Using the class structure, we increase the surface area of the public API, allowing for things like Base.set_peer_service!

Note that the current approach still exposes the Base. For instance, it exposes `Base.set_peer_service!` for everyone to call. If we truly want to make a class private, we need to use `private_constant`

```ruby
module A
  B = 1
  private_constant :B
end
=> A
irb(main):005:0> A::B
(irb):5:in `<main>': private constant A::B referenced (NameError)
```


2. Use of `extended`

```ruby
def self.extended(base)
  base.private_class_method :not_empty_tag?, :set_peer_service_from_source, :filter_peer_service_sources
end
```

The use of `extended` to make method private, seems like a smell, since that functionality could easily be achive by the use of `private` keyworkd. 
Also, it adds inconsistency because hides the method only when the `Base` module is included.

```ruby
module A
  module_function

  def self.extended(base)
    base.private_class_method :hello
  end

  def hello
    puts "hello"
  end
end

module B
  extend A
end
```

So `A.hello` would remain accessible while `B.hello` would be private

```
A.hello
hello
=> nil
irb(main):022:0> B.hello
(irb):22:in `<main>': private method `hello' called for B:Module (NoMethodError)
```

3. The use of `module_function` 

I'm not sure what the advantage of `module_function`` is in this case, as we only want to expose the methods at the singleton level `A.hello`` and the `V0`, `V1``, and `SpanAttributeSchema`` are not meant to be included in other classes or modules. The use of module_function basically converts a method defined for the instance available to the class, which I'm not sure is what we want in this case.

Here is an example of what `module_function` does


```ruby
module A
  module_function

  def hello
    puts "hello"
  end
end
```

Is the same as 

```ruby
module A
  def self.hello
    puts "hello"
  end

  def hello
    puts "hello"
  end
end
```

Putting all of my suggestions together, with the goal of hiding the internal stuff, I suggest keeping the code using modules, but removing the use of `extended`` and `module_function` + hide `Base` using `private_constant`.

```ruby
# frozen_string_literal: true

require_relative 'ext'

module Datadog
  module Tracing
    module Contrib
      # Contains methods for fetching values according to span attributes schema
      module SpanAttributeSchema
        # Contains interface of methods to be implemented
        module Base
          REFLEXIVE_SOURCES = [Tracing::Metadata::Ext::TAG_PEER_SERVICE].freeze
          NO_SOURCE = [].freeze
          IMPLEMENT_ERROR = 'SpanAttributeSchema Version must implement fetch_service_name'

          class << self
            def fetch_service_name(_env, _default)
              raise NotImplementedError, IMPLEMENT_ERROR
            end

            def set_peer_service!(span, sources)
              # Acquire all peer.service values as well as any potential remapping
              peer_service_val, peer_service_source = set_peer_service_from_source(span, sources)
              remap_val = Datadog.configuration.tracing.peer_service_mapping[peer_service_val]

              # Only continue to setting peer.service if actual source is found
              return false unless peer_service_source

              span.set_tag(Tracing::Contrib::Ext::Metadata::TAG_PEER_SERVICE_SOURCE, peer_service_source)

              # Set peer.service to remapped value if found otherwise normally set peer.service
              if remap_val
                span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE, remap_val)
                span.set_tag(Tracing::Contrib::Ext::Metadata::TAG_PEER_SERVICE_REMAP, peer_service_val)
              else
                span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE, peer_service_val)
              end
              true
            end

            private

            # set_peer_service_from_source: Implements the extraction logic to determine the peer.service value
            # based on the list of source tags passed as a parameter.
            #
            # If no values are found, it checks the default list for all spans before returning false for no result
            # Sets the source of where the information for peer.service was extracted from
            # Returns a peer.service value if successfully set or not
            def set_peer_service_from_source(span, sources = [])
              # Filter out sources based on existence of peer.service tag
              sources = filter_peer_service_sources(span, sources)

              # Find a possible peer.service source from the list of source tags passed in
              sources.each do |source|
                next unless not_empty_tag?(source_val)

                return source_val, source
              end
              false
            end

            # filter_peer_service_sources: returns filtered sources based on existence of peer.service tag
            # If peer.service exists, we do not read from any other source rather use peer.service as source
            # This is to prevent overwriting of pre-existing peer.service tags
            def filter_peer_service_sources(span, sources)
              ps = span.get_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE)
              # Do not override existing peer.service tag if it exists based on schema version
              return REFLEXIVE_SOURCES if not_empty_tag?(ps)

              # Check that peer.service is only set on spankind client/producer spans
              if (span.get_tag(Tracing::Metadata::Ext::TAG_KIND) == Tracing::Metadata::Ext::SpanKind::TAG_CLIENT) ||
                  (span.get_tag(Tracing::Metadata::Ext::TAG_KIND) == Tracing::Metadata::Ext::SpanKind::TAG_PRODUCER)
                return sources
              end

              NO_SOURCE
            end

            def not_empty_tag?(tag)
              tag && (tag != '')
            end
          end
        end

        private_constant :Base

        # Contains implementation of methods specific to v0
        module V0
          extend Base

          class << self
              ENV.fetch(env) do
                return Datadog.configuration.service if Datadog.configuration.tracing.global_default_service_name.enabled

                default
              end
            end
          end
        end

        # Contains implementation of methods specific to v1
        module V1
          extend Base

          class << self
            def fetch_service_name(env, _)
              ENV.fetch(env) { Datadog.configuration.service }
            end
          end
        end

        class << self
          def default_span_attribute_schema?
          end

            active_version.fetch_service_name(env, default)
          end

          def set_peer_service!(span, sources)
            active_version.set_peer_service!(span, sources)
          end

          private

          def active_version
            case Datadog.configuration.tracing.span_attribute_schema
            when 'v1'
              V1
            else
              V0 # Default Version
            end
          end
        end
      end
    end
  end
end
```

What do you think @zarirhamza @delner ?

lib/datadog/core/environment/variable_helpers.rb
zarirhamza: Done

lib/datadog/core/configuration/option.rb
zarirhamza: Done

delner (changes requested):
@zarirhamza and I talked about our plans for this. We identified there's a little more work to do to make `peer.service` configurable. This should be in-flight; once its merged in to this branch, I think this will be ready.

GustavoCaso:
As far as the code, it is LGTM. 

@zarirhamza, thanks for adding `type :hash` support to the core configuration options. 

As for the `peer.service` logic, I can not review if it is complete or not 😄  since I don't know how it should behave. I will defer that to @delner and @zarirhamza

lib/datadog/tracing/contrib/span_attribute_schema.rb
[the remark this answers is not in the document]
delner: A couple of points of clarification:

- > "For instance, it exposes `Base.set_peer_service!` for everyone to call."
   
   I don't believe this is strictly true. `set_peer_service!` is not a `module_function`, it is an instance method of the module. Thus you cannot call this statically. You would have to compose it into a module to access it; I'm not concerned with someone extending it, that's advanced usage.
- > So A.hello would remain accessible while B.hello would be private

    This code is not equivalent: the methods in `Base` are not `module_function` (like in the example), they are instance methods.

Otherwise, your code suggestion (at the bottom of your comment) looks reasonable. The only two things I'd call out as undesirable are (1) its unnecessary for `Base` to expose static-accessible methods, and (2) the use of `class << self` in `V0` and `V1`, which I think is a code smell; particularly it's a `Module` and not a `Class`, and it's... just ugly haha. Functionally I don't have an issue with it.

delner (approved):
Looks good. Thank you for all your hard work @zarirhamza!

PR #2983, opened 2023-07-20
Core configuration add env and type backend logic

GustavoCaso:
**What does this PR do?**

This PR extracts the logic for adding support for `env`, `deprecated_env`, `env_parser`, and `type` validation from https://github.com/DataDog/dd-trace-rb/pull/2970

For a more detail discussion on these changes please refer to the PR linked above 

**Motivation**

**Additional Notes**

**How to test the change?**

### DIFF
--- a/lib/datadog/core/configuration/base.rb
+++ b/lib/datadog/core/configuration/base.rb
@@ -34,8 +34,6 @@ module Datadog
                 value.reset! if value.respond_to?(:reset!)
                 value
               end
-
-              o.type settings_class
             end
           end
 
--- a/lib/datadog/core/configuration/base.rb
+++ b/lib/datadog/core/configuration/base.rb
@@ -25,7 +25,7 @@ module Datadog
           # e.g. `settings :foo { option :bar }` --> `config.foo.bar`
           # @param [Symbol] name option name. Methods will be created based on this name.
           def settings(name, &block)
-            settings_class = new_settings_class(&block)
+            settings_class = new_settings_class(name, &block)
 
             option(name) do |o|
               o.default { settings_class.new }
@@ -35,12 +35,15 @@ module Datadog
                 value
               end
             end
+
+            settings_class
           end
 
           private
 
-          def new_settings_class(&block)
+          def new_settings_class(name, &block)
             Class.new { include Configuration::Base }.tap do |klass|
+              klass.instance_variable_set(:@settings_name, name)
               klass.instance_eval(&block) if block
             end
           end
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -13,11 +13,11 @@ module Datadog
           # Remote configuration provided through the Datadog app.
           REMOTE_CONFIGURATION = [2, :remote_configuration].freeze
 
-          # Configuration provided in Ruby code, in this same process.
+          # Configuration provided in Ruby code, in this same process
+          # or via Environment variable
           PROGRAMMATIC = [1, :programmatic].freeze
 
-          # Configuration that comes either from environment variables,
-          # or fallback values.
+          # Configuration that comes from default values
           DEFAULT = [0, :default].freeze
         end
 
@@ -49,7 +49,7 @@ module Datadog
           end
 
           old_value = @value
-          (@value = context_exec(value, old_value, &definition.setter)).tap do |v|
+          (@value = context_exec(validate_type(value), old_value, &definition.setter)).tap do |v|
             @is_set = true
             @precedence_set = precedence
             context_exec(v, old_value, &definition.on_set) if definition.on_set
@@ -62,7 +62,7 @@ module Datadog
           elsif definition.delegate_to
             context_eval(&definition.delegate_to)
           else
-            set(default_value, precedence: Precedence::DEFAULT)
+            set_value_from_env_or_default
           end
         end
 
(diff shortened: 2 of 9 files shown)


lib/datadog/core/configuration/base.rb
```diff
@@ -34,15 +34,16 @@ def settings(name, &block)
                 value.reset! if value.respond_to?(:reset!)
                 value
               end
-
-              o.type settings_class
             end
+
+            settings_class
```
GustavoCaso, line 39: I return the `settings_class` to be able to access the `spec/datadog/core/configuration/base_spec.rb`. Previously the returned value was an `OptionDefintion`, which does not make much sense.

lib/datadog/core/configuration/option.rb
```diff
@@ -102,6 +191,33 @@ def context_eval(&block)
...
+            end
+          end
+
+          option_value = value.nil? ? default_value : value
+
+          set(option_value, precedence: precedence || Precedence::DEFAULT)
+        end
+
+        def skip_validation?
+          !!ENV['DD_EXPERIMENTAL_SKIP_CONFIGURATION_VALIDATION']
```
ivoanjo, line 218: To make sure that we don't trigger unless it's being set to the correct thing:

```suggestion
          ['true', '1'].include?(ENV.fetch('DD_EXPERIMENTAL_SKIP_CONFIGURATION_VALIDATION', '').strip)
```

lib/datadog/core/configuration/option.rb
```diff
@@ -13,11 +13,11 @@ module Precedence
           # Remote configuration provided through the Datadog app.
           REMOTE_CONFIGURATION = [2, :remote_configuration].freeze
 
-          # Configuration provided in Ruby code, in this same process.
+          # Configuration provided in Ruby code, in this same process
+          # or via Environment variable
           PROGRAMMATIC = [1, :programmatic].freeze
```
lloeki, line 18: `PROGRAMMATIC` feels very weird. A couple suggestions:

- it's configuration as code
  
  ```
  CODE = :code
  ```

- configuration is written by using a DSL:
  
  ```
  DSL = :dsl
  ```

- but the above are quite "technical" so to speak, so I quite like this one:
  
  ```
  LOCAL = :local
  ```

Why do I like it?:

- `local` is squarely set within the app
- `remote` is telecontrolled from some far away place
- `env` is from outside the app (literally the definition of "environment"), and possibly globally set at the infrastructure level
- if there is someday such a thing as configuring from, say, a YAML file, `file` is from outside the app as well

So overall I feel they are all quite short and clear when contextualised.

lloeki (approved):
Technically it looks good. I would like a few adjustments WRT some concepts and naming though.

lib/datadog/core/configuration/option.rb
```diff
@@ -102,6 +191,33 @@ def context_eval(&block)
...
+              "#{definition.deprecated_env} environment variable is deprecated, use #{definition.env} instead."
+            end
+          end
+
+          option_value = value.nil? ? default_value : value
+
+          set(option_value, precedence: precedence || Precedence::DEFAULT)
+        end
+
+        def skip_validation?
```
ivoanjo, line 217: Maybe not for this PR, but it may be worth using an `OnlyOnce` to drop a warning to customers when we detect this is being used. This way we push people to stop using it, as well as are able to see it if someone reports an issue and shares a log.

lib/datadog/core/configuration/option.rb
```diff
@@ -94,6 +94,95 @@ def default_precedence?
...
+          when :hash
+            value.is_a?(Hash)
+          when :bool
+            value.is_a?(TrueClass) || value.is_a?(FalseClass)
+          when :proc
+            value.is_a?(Proc)
+          when :symbol
+            value.is_a?(Symbol)
+          when NilClass
+            true
```
ivoanjo, line 180: Minor:

```suggestion
          when NilClass
            true # No validation is performed when option is typeless
```

lib/datadog/core/configuration/option.rb
```diff
@@ -102,6 +191,33 @@ def context_eval(&block)
           @context.instance_eval(&block)
         end
 
+        def set_value_from_env_or_default
+          value = nil
+          precedence = nil
+
+          if definition.env && ENV[definition.env]
+            value = coerce_env_variable(ENV[definition.env])
+            precedence = Precedence::PROGRAMMATIC
```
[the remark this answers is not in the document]
GustavoCaso: I don't dislike the idea. 

For now, I would not do much about it because there is another PR by @marcotc https://github.com/DataDog/dd-trace-rb/pull/2972 that uses the precedence to unset configuration values.

I'm not sure what weight `Precedence::ENVIRONMENT` should have. Same as `Precedence::PROGRAMMATIC`? 

What would happen when someone configures something via ENV var and later changed it using the `configure` block? 

Since those are question that we might not have the answer now I think is best to wait until making that change

lib/datadog/core/configuration/option_definition.rb
```diff
@@ -122,13 +148,16 @@ def apply_options!(options = {})
...
+            deprecated_env(options[:deprecated_env]) if options.key?(:deprecated_env)
+            env_parser(&options[:env_parser]) if options.key?(:env_parser)
             delegate_to(&options[:delegate_to]) if options.key?(:delegate_to)
             depends_on(*options[:depends_on]) if options.key?(:depends_on)
             lazy(options[:lazy]) if options.key?(:lazy)
             on_set(&options[:on_set]) if options.key?(:on_set)
             resetter(&options[:resetter]) if options.key?(:resetter)
             setter(&options[:setter]) if options.key?(:setter)
-            type(&options[:type]) if options.key?(:type)
+            type(options[:type]) if options.key?(:type)
```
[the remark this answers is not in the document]
GustavoCaso: I believe this extra complexity comes from allowing to define options either using a block or inline style:

```ruby
option :enabled, default: true, env: 'DD_ENABLED'
```

```ruby
option :enabled do |o|
  o.default true
  o. env 'DD_ENABLED'
end
```

We might as well choose one form and remove the extra complexity 😄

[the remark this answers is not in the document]
ivoanjo: Good point. I don't entirely dislike the one-line variant, I suspect we may be able to implement it without as much boilerplate, but we don't have to find out for now :P

lib/datadog/core/configuration/option.rb
```diff
@@ -143,16 +144,17 @@ def validate_type(value)
...
-                          "#{@definition.type} , but a `#{value.class}` was provided (#{value.inspect})."\
+                          "The setting `#{@definition.name}` inside your app's `Datadog.configure` block expects a "\
+                          "#{@definition.type}, but a `#{value.class}` was provided (#{value.inspect})."\
                         end
 
             error_msg = "#{error_msg} Please update your `configure` block. "\
-            'Alternatively, you can disable this validation using `DD_EXPERIMENTAL_SKIP_CONFIGURATION_VALIDATION=1`. '\
-            'For help, please open an issue on https://github.com/DataDog/dd-trace-rb/issues/new/choose.'
+            'Alternatively, you can disable this validation using the '\
+            '`DD_EXPERIMENTAL_SKIP_CONFIGURATION_VALIDATION=true`environment variable. '\
```
ivoanjo, line 156: Minor: Missed a space between ```...=true`environment```

GustavoCaso: Will fix on the following PR 😄

PR #2985, opened 2023-07-20
Refactor EnvironmentLogger V1

sarahchen6:
**What does this PR do?**
This PR moves the EnvironmentLogger module from the Core namespace to the Tracing namespace to remove product dependencies. It also updates its specs and any references to the module.

**Motivation**
The overall goal of this PR is to refactor the Core namespace so that it does not depend upon product implementation. This PR specifically addresses the EnvironmentLogger module.

**Additional Notes**
Currently the code still couples Tracing and Profiling; however, we are hoping to remove or separate the Profiling dependency soon.

**How to test the change?**
This change can be tested by running unit tests for the specific file `bundle exec rspec spec/datadog/tracing/diagnostics/environment_logger_spec.rb` and by running tests for the entire suite `bundle exec rake spec:main.`


lib/datadog/tracing/diagnostics/environment_logger.rb
```diff
@@ -74,14 +74,14 @@ def version
...
-          Core::Environment::Ext::LANG
+          Diagnostics::Ext::LANG
         end
 
         # Supported Ruby language version.
         # Will be distinct from VM version for non-MRI environments.
         # @return [String]
         def lang_version
-          Core::Environment::Ext::LANG_VERSION
+          Diagnostics::Ext::LANG_VERSION
```
marcotc, line 84: I recommend using `Datadog::Core::Environment::Ext::LANG_VERSION` instead of creating yet another `LANG_VERSION` just for this use.

Different ruby runtimes (cruby, jruby, truffleruby) has slightly different semantic for these fields, so it's good to hide all the complexity in one place and let all of Ruby Datadog products get consistent versions values.

I suggest the same for `def lang` above.

sarahchen6: Good point. Just adjusted both constants, and updated the environment/ext.rb file as well.

sarahchen6:
See #3020

PR #2986, opened 2023-07-20
Implement `peer_service` option

zarirhamza:
Allows users to set `peer.service` manually via the `peer_service` option in all target integrations where `peer.service` is correctly implemented.

Please do not merge without @delner approval but feel free to review


spec/datadog/tracing/contrib/dalli/instrumentation_spec.rb
```diff
@@ -78,13 +78,7 @@
   end
 
   describe 'when multiplexed configuration is provided' do
-    let(:service_name) { 'multiplex-service' }
-
-    before do
```
delner, line 83: Is removing this block correct? My understanding is it wouldn't be testing multiplexing then. The test becomes a false negative.

lib/datadog/tracing/contrib/aws/configuration/settings.rb
```diff
@@ -32,6 +32,8 @@ class Settings < Contrib::Configuration::Settings
                 )
               end
             end
+
+            option :peer_service
```
delner, line 36: I think we should add an ENV var to back this value, to make it configurable without code. e.g. `DD_TRACE_AWS_PEER_SERVICE`. A code comment (explaining usage, effect, purpose) and addition to the AWS options table in `GettingStarted.md` would also be in order.

Same goes for the rest.

lib/datadog/tracing/contrib/aws/instrumentation.rb
```diff
@@ -40,6 +40,13 @@ def annotate!(span, context)
               handler.add_tags(span, params)
             end
 
+            if configuration[:peer_service]
```
delner, line 43: 👍 . You could probably collapse this to a line if you wanted to; doesn't matter much though.

delner (changes requested):
My comments are probably too early, but hopefully they're helpful.

lib/datadog/tracing/contrib/aws/instrumentation.rb
zarirhamza: I think for now we can keep it the way it is and mass refactor if we need to, but I also agree it's fine the way it is now

lib/datadog/tracing/contrib/aws/configuration/settings.rb
zarirhamza: Hmm ok if that's confirmed what we want to do, we would need to utilize the new options which is possible but just a thing to note. Then we would probably create a separate test that tests the env var. Will do though, saw this comment after I pushed my other changes

spec/datadog/tracing/contrib/dalli/instrumentation_spec.rb
zarirhamza: Yes you're right. These tests will fail, but if we include the new env vars with new tests then it should work out.

delner: Not sure I understand how you mean by work out. The test isn't testing multiplexing if it doesn't apply multiplexing?

lib/datadog/tracing/contrib/aws/configuration/settings.rb
delner: Is there a reason why would wouldn't want to do this?

zarirhamza: Nope, sounds like a good idea to me. I implemented it thoroughly, but we need to discuss our approach on how to document this properly in terms of comments and documentation

spec/datadog/tracing/contrib/dalli/instrumentation_spec.rb
zarirhamza: Nevermind, reverted back to the way it was before so no worries

lib/datadog/tracing/contrib/elasticsearch/patcher.rb
```diff
@@ -77,6 +77,13 @@ def perform_request(*args)
                     host = connection.host[:host] if connection
                     port = connection.host[:port] if connection
 
+                    if datadog_configuration[:peer_service]
+                      span.set_tag(
+                        Tracing::Metadata::Ext::TAG_PEER_SERVICE,
+                        datadog_configuration[:peer_service]
+                      )
+                    end
```
GustavoCaso, line 85: Is this something that could be encapsulated inside `SpanAttributeSchema`? 

Also, not necessary for this PR but I believe the name `SpanAttributeSchema` no longer reflect the core responsibility of that module. Maybe `PeerServiceTagger`?

zarirhamza: After discussing, we decided not to move forward with these suggestions. We want to avoid sending the span and configuration block to `SpanAttributeSchema` more than necessary, and regardless of the encapsulation, it still needs to be called every integration so there seems to be no need to change it.

While it may not reflect changes at the moment, there will be many changes in the future, fully showing the reason for calling it `SpanAttributeSchema`

delner (approved):
Looks good overall. Thanks! :+1:

PR #2987, opened 2023-07-21
Update AppSec Settings to use the `env`, and `type` options

GustavoCaso:
**What does this PR do?**

Build on top of https://github.com/DataDog/dd-trace-rb/pull/2983

Update the AppSec settings to use `env`, and `type` options to improve appsec settings integrity

**Motivation**

**Additional Notes**

**How to test the change?**

### DIFF
--- a/lib/datadog/appsec/configuration/settings.rb
+++ b/lib/datadog/appsec/configuration/settings.rb
@@ -11,15 +11,8 @@ module Datadog
         DEFAULT_OBFUSCATOR_KEY_REGEX = '(?i)(?:p(?:ass)?w(?:or)?d|pass(?:_?phrase)?|secret|(?:api_?|private_?|public_?)key)|token|consumer_?(?:id|key|secret)|sign(?:ed|ature)|bearer|authorization'
         DEFAULT_OBFUSCATOR_VALUE_REGEX = '(?i)(?:p(?:ass)?w(?:or)?d|pass(?:_?phrase)?|secret|(?:api_?|private_?|public_?|access_?|secret_?)key(?:_?id)?|token|consumer_?(?:id|key|secret)|sign(?:ed|ature)?|auth(?:entication|orization)?)(?:\s*=[^;]|"\s*:\s*"[^"]+")|bearer\s+[a-z0-9\._\-]+|token:[a-z0-9]{13}|gh[opsu]_[0-9a-zA-Z]{36}|ey[I-L][\w=-]+\.ey[I-L][\w=-]+(?:\.[\w.+\/=-]+)?|[\-]{5}BEGIN[a-z\s]+PRIVATE\sKEY[\-]{5}[^\-]+[\-]{5}END[a-z\s]+PRIVATE\sKEY|ssh-rsa\s*[a-z0-9\/\.+]{100,}'
         # rubocop:enable Layout/LineLength
-        DEFAULT_APPSEC_ENABLED = false
-        DEFAULT_APPSEC_RULESET = :recommended
-        DEFAULT_APPSEC_WAF_TIMEOUT = 5_000 # us
-        DEFAULT_APPSEC_WAF_DEBUG = false
-        DEFAULT_APPSEC_TRACE_RATE_LIMIT = 100 # traces/s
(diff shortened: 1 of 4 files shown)


spec/datadog/appsec/configuration/settings_spec.rb
```diff
@@ -551,7 +533,7 @@ def patcher
         end
 
         context 'when given a supported value' do
-          let(:track_user_events_mode) { :extended }
+          let(:track_user_events_mode) { 'extended' }
```
marcotc, line 536: There's a small change, where somewhere we are converting a `:extended` to a string, but now we just pass the string.
Is this material to the application running appsec, or simply a testing artifact?

marcotc (approved):

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -53,80 +45,80 @@ def self.add_settings!(base)
...
               option :obfuscator_value_regex do |o|
-                o.default do
-                  ENV.fetch(
-                    'DD_APPSEC_OBFUSCATION_PARAMETER_VALUE_REGEXP',
-                    DEFAULT_OBFUSCATOR_VALUE_REGEX
-                  )
-                end
+                o.type :string
+                o.env 'DD_APPSEC_OBFUSCATION_PARAMETER_VALUE_REGEXP'
+                o.default DEFAULT_OBFUSCATOR_VALUE_REGEX
```
ivoanjo, line 91: Not for this PR, but it kills me that we use `REGEXP` and `REGEX` lol

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -53,80 +45,80 @@ def self.add_settings!(base)
...
-                    string_value = v.to_s
-                    if APPSEC_VALID_TRACK_USER_EVENTS_MODE.include?(string_value)
-                      string_value
+                    if APPSEC_VALID_TRACK_USER_EVENTS_MODE.include?(v)
+                      v
                     else
                       Datadog.logger.warn(
                         'The appsec.track_user_events.mode value provided is not supported.' \
                         "Supported values are: safe | extended.\n" \
                         'Using default value safe'
```
ivoanjo, line 119: Minor: This was already like this, but since we're here, I suggest removing the newline from the log message, as a lot of logging tools can get tripped with newlines (e.g. show them out of order and whatnot), and thus having it all in a single line avoids any possible confusion.

```suggestion
                        'The appsec.track_user_events.mode value provided is not supported.' \
                        "Supported values are: safe | extended. " \
                        'Using default value safe'
```

spec/datadog/appsec/configuration/settings_spec.rb
```diff
@@ -52,7 +52,7 @@ def patcher
         context 'is not defined' do
           let(:appsec_enabled) { nil }
 
-          it { is_expected.to be described_class::DEFAULT_APPSEC_ENABLED }
+          it { is_expected.to eq false }
         end
```
ivoanjo, line 56: :clap: I really love having the defaults inlined in the tests. That's because it protects us from changing some default accidentally, or at least without considering its impact on the tests -- if the behavior of the implementation (default) changes, the tests should have to change too!

ivoanjo (approved):
:+1: LGTM

spec/datadog/appsec/configuration/settings_spec.rb
GustavoCaso: On the first iteration, we accepted either string or symbol as a possible value for `appsec.track_user_events.mode`. Internally we stored it as a string, and what ever check we did against those values expected a string.

https://github.com/DataDog/dd-trace-rb/blob/7a8771098170280b81c5e4acb9b6cab2393fbadd/lib/datadog/appsec/contrib/devise/event.rb#L41-L44

With this change, we signal that we only support string values. 

As a note, this feature hasn't ben release yet, so there are no users using it, so there is no breaking change or anything that could potentially affect users of appsec

lib/datadog/appsec/configuration/settings.rb
lloeki: I definitely relate to that feeling!

lloeki (approved):
LGTM!

PR #2988, opened 2023-07-21
Update core configuration to use env and type options

GustavoCaso:
**What does this PR do?**

This code has been extracted from https://github.com/DataDog/dd-trace-rb/pull/2970 to facilitate the review process. 

This PR focus on updating our current settings options to use `env` and `type` new functionality https://github.com/DataDog/dd-trace-rb/pull/2983

The only addition outside of updating the settings is making sure the default value gets duplicated before setting it https://github.com/DataDog/dd-trace-rb/pull/2988/commits/a118dbd768bf15a63dba06742fed5ece426fd9b7

There has been a discussion rather than duplicating. We should make sure that it is frozen to avoid unwanted changes to the value. https://github.com/DataDog/dd-trace-rb/pull/2970#discussion_r1267962923
That change would be done on a separate PR 

**Motivation**

**Additional Notes**

**How to test the change?**

### DIFF
--- a/lib/datadog/ci/configuration/settings.rb
+++ b/lib/datadog/ci/configuration/settings.rb
@@ -16,7 +16,9 @@ module Datadog
           base.class_eval do
             settings :ci do
               option :enabled do |o|
-                o.default { env_to_bool(CI::Ext::Settings::ENV_MODE_ENABLED, false) }
+                o.type :bool
+                o.env CI::Ext::Settings::ENV_MODE_ENABLED
+                o.default false
               end
 
               # DEV: Alias to Datadog::Tracing::Contrib::Extensions::Configuration::Settings#instrument.
@@ -34,12 +36,11 @@ module Datadog
               # TODO: Deprecate in the next major version, as `instrument` better describes this method's purpose
               alias_method :use, :instrument
 
-              option :trace_flush do |o|
-                o.default { nil }
-              end
+              option :trace_flush
 
               option :writer_options do |o|
-                o.default { {} }
+                o.type :hash
+                o.default({})
               end
             end
           end
--- a/lib/datadog/core/configuration/settings.rb
+++ b/lib/datadog/core/configuration/settings.rb
@@ -79,7 +79,8 @@ module Datadog
         # @default `DD_API_KEY` environment variable, otherwise `nil`
         # @return [String,nil]
         option :api_key do |o|
-          o.default { ENV.fetch(Core::Environment::Ext::ENV_API_KEY, nil) }
+          o.type :string, nilable: true
+          o.env Core::Environment::Ext::ENV_API_KEY
         end
 
         # Datadog diagnostic settings.
@@ -98,7 +99,9 @@ module Datadog
           # @default `DD_TRACE_DEBUG` environment variable, otherwise `false`
           # @return [Boolean]
           option :debug do |o|
-            o.default { env_to_bool(Datadog::Core::Configuration::Ext::Diagnostics::ENV_DEBUG_ENABLED, false) }
+            o.env Datadog::Core::Configuration::Ext::Diagnostics::ENV_DEBUG_ENABLED
+            o.default false
+            o.type :bool
             o.on_set do |enabled|
               # Enable rich debug print statements.
               # We do not need to unnecessarily load 'pp' unless in debugging mode.
@@ -115,7 +118,9 @@ module Datadog
             # @default `DD_HEALTH_METRICS_ENABLED` environment variable, otherwise `false`
             # @return [Boolean]
             option :enabled do |o|
-              o.default { env_to_bool(Datadog::Core::Configuration::Ext::Diagnostics::ENV_HEALTH_METRICS_ENABLED, false) }
+              o.env Datadog::Core::Configuration::Ext::Diagnostics::ENV_HEALTH_METRICS_ENABLED
+              o.default false
+              o.type :bool
             end
 
             # {Datadog::Statsd} instance to collect health metrics.
@@ -136,11 +141,12 @@ module Datadog
             # If `nil`, defaults to logging startup logs when `ddtrace` detects that the application
             # is *not* running in a development environment.
             #
-            # @default `DD_TRACE_STARTUP_LOGS` environment variable, otherwise `nil`
-            # @return [Boolean,nil]
+            # @default `DD_TRACE_STARTUP_LOGS` environment variable, otherwise `false`
+            # @return [Boolean]
             option :enabled do |o|
-              # Defaults to nil as we want to know when the default value is being used
-              o.default { env_to_bool(Datadog::Core::Configuration::Ext::Diagnostics::ENV_STARTUP_LOGS_ENABLED, nil) }
+              o.env Datadog::Core::Configuration::Ext::Diagnostics::ENV_STARTUP_LOGS_ENABLED
+              o.default false
+              o.type :bool
             end
           end
         end
@@ -154,7 +160,7 @@ module Datadog
           o.setter { |v| v.to_s if v }
 
           # NOTE: env also gets set as a side effect of tags. See the WORKAROUND note in #initialize for details.
-          o.default { ENV.fetch(Core::Environment::Ext::ENV_ENVIRONMENT, nil) }
+          o.env Core::Environment::Ext::ENV_ENVIRONMENT
         end
 
         # Internal `Datadog.logger` configuration.
@@ -189,7 +195,9 @@ module Datadog
           # @default `DD_PROFILING_ENABLED` environment variable, otherwise `false`
           # @return [Boolean]
           option :enabled do |o|
-            o.default { env_to_bool(Profiling::Ext::ENV_ENABLED, false) }
+            o.env Profiling::Ext::ENV_ENABLED
+            o.default false
+            o.type :bool
           end
 
           # @public_api
@@ -222,7 +230,9 @@ module Datadog
             #
             # @default `DD_PROFILING_MAX_FRAMES` environment variable, otherwise 400
             option :max_frames do |o|
-              o.default { env_to_int(Profiling::Ext::ENV_MAX_FRAMES, 400) }
+              o.type :int
+              o.env Profiling::Ext::ENV_MAX_FRAMES
+              o.default 400
             end
 
             # @public_api
@@ -234,7 +244,9 @@ module Datadog
                 # @default `DD_PROFILING_ENDPOINT_COLLECTION_ENABLED` environment variable, otherwise `true`
                 # @return [Boolean]
                 option :enabled do |o|
-                  o.default { env_to_bool(Profiling::Ext::ENV_ENDPOINT_COLLECTION_ENABLED, true) }
+                  o.env Profiling::Ext::ENV_ENDPOINT_COLLECTION_ENABLED
+                  o.default true
+                  o.type :bool
                 end
               end
             end
@@ -276,7 +288,9 @@ module Datadog
             #
             # @default `DD_PROFILING_FORCE_ENABLE_LEGACY` environment variable, otherwise `false`
             option :force_enable_legacy_profiler do |o|
-              o.default { env_to_bool('DD_PROFILING_FORCE_ENABLE_LEGACY', false) }
+              o.env 'DD_PROFILING_FORCE_ENABLE_LEGACY'
+              o.default false
+              o.type :bool
               o.on_set do |value|
                 if value
                   Datadog.logger.warn(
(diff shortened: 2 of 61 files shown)


lib/datadog/tracing/contrib/http/configuration/settings.rb
```diff
@@ -35,10 +41,26 @@ class Settings < Contrib::Configuration::Settings
             end
 
             option :error_status_codes do |o|
```
GustavoCaso, line 43: I haven't add the `type` to the `error_status_code` option, because in the tests the values used are Array and Range. I didn't wanted to change the backend logic for `:array` on this PR. So I will leave the decision to adding the `type` to the original author of this integration

lib/datadog/tracing/contrib/httpclient/configuration/settings.rb
```diff
@@ -35,10 +41,26 @@ class Settings < Contrib::Configuration::Settings
             end
 
             option :error_status_codes do |o|
```
GustavoCaso, line 43: Same as https://github.com/DataDog/dd-trace-rb/pull/2988/files#r1272116069

lib/datadog/tracing/contrib/delayed_job/configuration/settings.rb
```diff
@@ -13,20 +13,29 @@ module Configuration
...
+              o.env Ext::ENV_ANALYTICS_SAMPLE_RATE
+              o.default 1.0
             end
 
             option :service_name
             option :client_service_name
-            option :error_handler, experimental_default_proc: Tracing::SpanOperation::Events::DEFAULT_ON_ERROR
+            option :error_handler do |o|
+              o.type :proc
+              o.experimental_default_proc(&Tracing::SpanOperation::Events::DEFAULT_ON_ERROR)
```
ivoanjo, line 37: Not for this PR -- but are we're ready to make a decision on the `experimental_default_proc`? 

E.g. do we want to keep it as-is for now (probably rename it), or do you still have plans to change/remove it?

lib/datadog/tracing/contrib/opensearch/configuration/settings.rb
```diff
@@ -12,18 +12,21 @@ module Configuration
...
 
             option :analytics_sample_rate do |o|
-              o.default { env_to_float(Ext::ENV_ANALYTICS_SAMPLE_RATE, 1.0) }
-              o.lazy
+              o.type :float
+              o.env Ext::ENV_ANALYTICS_SAMPLE_RATE
+              o.default 1.0
             end
 
             option :quantize, default: {}
```
ivoanjo, line 32: @sarahchen6 Maybe worth adding a type here as a follow-up PR to Gustavo's?

spec/datadog/ci/configuration/settings_spec.rb
```diff
@@ -18,6 +18,10 @@
     end
 
     describe '#ci' do
+      after do
+        settings.reset!
+      end
+
```
ivoanjo, line 24: I don't think this is needed? `subject(:settings)` above seems to be defined as a new object that gets recreated for every testcase, so there shouldn't be any leftover state between tests that needs resetting.

Also, the tests still pass for me if I remove this :)

spec/datadog/core/configuration/option_spec.rb
```diff
@@ -34,6 +34,12 @@
   let(:setter_value) { double('setter_value') }
   let(:context) { double('configuration object') }
 
+  before do
+    allow(Datadog::Core::Utils::SafeDup).to receive(:frozen_or_dup) do |args, _block|
+      args
+    end
+  end
```
ivoanjo, line 41: Minor: May be worth leaving here a comment (as well as on `options_spec`) to explain why this is needed.

spec/datadog/core/configuration/settings_spec.rb
```diff
@@ -12,6 +12,10 @@
 RSpec.describe Datadog::Core::Configuration::Settings do
   subject(:settings) { described_class.new(options) }
 
+  after do
+    settings.reset!
+  end
```
ivoanjo, line 17: Similar to my comment above on ci/configuration/settings_spec.rb, I don't think this is needed? And the specs still pass for me if I remove this.

spec/datadog/tracing/configuration/settings_spec.rb
```diff
@@ -18,6 +18,10 @@
   let(:options) { {} }
 
   describe '#tracing' do
+    after do
+      settings.reset!
+    end
```
ivoanjo, line 23: Similar comment to others above :)

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -261,9 +259,10 @@ def self.extended(base)
                 # for resources not seen recently).
                 #
                 # @default `DD_TRACE_SAMPLE_RATE` environment variable, otherwise `nil`.
-                # @return [Float,nil]
+                # @return [Float]
```
ivoanjo, line 262: This can still be `nil` (so this change is incorrect...?)

lib/datadog/tracing/contrib/rails/configuration/settings.rb
```diff
@@ -24,30 +24,35 @@ def initialize(options = {})
...
             option :analytics_enabled do |o|
-              o.default { env_to_bool(Ext::ENV_ANALYTICS_ENABLED, nil) }
-
+              o.type :bool, nilable: true
+              o.env Ext::ENV_ANALYTICS_ENABLED
               o.on_set do |value|
                 # Update ActionPack analytics too
-                Datadog.configuration.tracing[:action_pack][:analytics_enabled] = value
+                Datadog.configuration.tracing[:action_pack][:analytics_enabled] = value unless value.nil?
               end
```
ivoanjo, line 38: It's subtle, but since both this setting, as well as the action_pack setting allow `nil`, I think perhaps that `unless value.nil?` is perhaps not the best idea (because in a way, `nil` previously meant "disable everything" and now only means "disable only rails".

In practice in most cases it'll be equivalent, but my thinking is, why change it, if we can keep the exact same previous behavior?

lib/datadog/core/configuration/settings.rb
```diff
@@ -136,11 +141,12 @@ def initialize(*_)
             # If `nil`, defaults to logging startup logs when `ddtrace` detects that the application
             # is *not* running in a development environment.
             #
-            # @default `DD_TRACE_STARTUP_LOGS` environment variable, otherwise `nil`
-            # @return [Boolean,nil]
+            # @default `DD_TRACE_STARTUP_LOGS` environment variable, otherwise `false`
+            # @return [Boolean]
```
ivoanjo, line 145: See my comment on environment_logger; if you end up agreeing, this change to remove the `nil` probably needs reverting.

ivoanjo (commented):
It's in mega good shape! I left a few small notes, and I think after looking into them/discussing them, this is good to go.

One additional note: when running the specs manually I'm seeing a bunch of

`/lib/datadog/core/configuration/options.rb:27: warning: instance variable @settings_name not initialized`

I think this happens for top-level groups (e.g. `c.agent`) that don't get to have a custom class. I think the fix is as simple as `settings_name = defined?(@settings_name) && @settings_name`, as we do in a few other places in the codebase.

lib/datadog/core/configuration/option.rb
```diff
@@ -137,7 +139,7 @@ def default_value
           if definition.default.instance_of?(Proc)
             context_eval(&definition.default)
           else
-            definition.experimental_default_proc || definition.default
+            definition.experimental_default_proc || Core::Utils::SafeDup.frozen_or_dup(definition.default)
```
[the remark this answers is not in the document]
GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/2999

lib/datadog/tracing/contrib/delayed_job/configuration/settings.rb
GustavoCaso: For now, no plans to remove it. We should rename it to `default_proc` 😄

lib/datadog/core/diagnostics/environment_logger.rb
```diff
@@ -38,12 +38,11 @@ def log_error!(type, error)
...
-            if startup_logs_enabled.nil?
-              !repl? # Suppress logs if we running in a REPL
-            else
-              startup_logs_enabled
+            if Datadog.configuration.diagnostics.startup_logs.using_default?(:enabled)
+              return !repl? # Suppress logs if we running in a REPL
             end
+
+            Datadog.configuration.diagnostics.startup_logs.enabled
           end
```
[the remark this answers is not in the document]
GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/2988/commits/165138d90e99ad6cadc95de464bac50216a4bb69

lib/datadog/core/configuration/settings.rb
GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/2988/commits/165138d90e99ad6cadc95de464bac50216a4bb69

lib/datadog/tracing/contrib/rails/configuration/settings.rb
GustavoCaso: Done [00b1168](https://github.com/DataDog/dd-trace-rb/pull/2988/commits/00b1168187a12a8bc75e8a65a7063f3845e6c3b1)

On a private Zoom call, we also discussed that `rails` and `action_pack` supporting `nil` for `analytics_enabled` is weird and that we should use a boolean value instead.

That change should be done on a separate PR

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -396,13 +404,16 @@ def self.extended(base)
...
                   o.default do
                     disabled = env_to_bool(Tracing::Configuration::Ext::ClientIp::ENV_DISABLED)
 
                     enabled = if disabled.nil?
                                 false
                               else
-                                Datadog.logger.warn { "#{Tracing::Configuration::Ext::ClientIp::ENV_DISABLED} environment variable is deprecated, found set to #{disabled}, use #{Tracing::Configuration::Ext::ClientIp::ENV_ENABLED}=#{!disabled}" }
+                                Datadog::Core.log_deprecation do
+                                  "#{Tracing::Configuration::Ext::ClientIp::ENV_DISABLED} environment variable is deprecated, use #{Tracing::Configuration::Ext::ClientIp::ENV_ENABLED} instead."
+                                end
```
marcotc, line 416: 👍

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -380,8 +385,11 @@ def self.extended(base)
               # This option is recommended for internal use only.
               #
               # @default `{}`
-              # @return [Hash,nil]
-              option :writer_options, default: ->(_i) { {} }
+              # @return [Hash]
+              option :writer_options do |o|
+                o.type :hash
```
marcotc, line 390: This was changed to not nillable, but after inspecting our codebase, this change has no impact as the default `{}` is accepted everywhere `nil` would.

lib/datadog/ci/configuration/settings.rb
```diff
@@ -34,12 +36,11 @@ def self.add_settings!(base)
               # TODO: Deprecate in the next major version, as `instrument` better describes this method's purpose
               alias_method :use, :instrument
 
-              option :trace_flush do |o|
-                o.default { nil }
-              end
+              option :trace_flush
```
marcotc, line 39: When nothing is configured like this, is the type nillable?

marcotc (approved):
Nothing really blocking this, awesome job!

lib/datadog/tracing/contrib/delayed_job/configuration/settings.rb
ivoanjo: Ack! Created https://github.com/DataDog/ruby-guild/issues/29 to record it (Datadog-only link)

lib/datadog/ci/configuration/settings.rb
ivoanjo: When the there's no type, the validation is disabled completely (https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/core/configuration/option.rb#L237) so it's a bit different from nillable (because nillable is a modifier after you select a specific type)

lib/datadog/core/configuration/settings.rb
```diff
@@ -137,10 +142,10 @@ def initialize(*_)
             # is *not* running in a development environment.
             #
             # @default `DD_TRACE_STARTUP_LOGS` environment variable, otherwise `nil`
-            # @return [Boolean,nil]
+            # @return [Boolean, nil]
             option :enabled do |o|
-              # Defaults to nil as we want to know when the default value is being used
```
ivoanjo, line 142: Minor: May be worth keeping this comment?

ivoanjo (approved):
:+1: LGTM!

For future PRs, I'd suggest not force-pushing after the first review is posted, as it makes it hard to see what has changed between revisions.

I'm the "trust, but verify" kinda person so I always like to give a look to the overall diff to try to spot new accidental issues, and this makes it hard to know if the older commits have changed :)

PR #2989, opened 2023-07-21
Fix:Elasticsearch Webmock testing

marcotc:
Elasticsearch tests started failing due to Webmock assertions recently.

This is caused by Elasticsearch using Webmock, but other processes like Telemetry still being executed in the background. This PR allows agent calls to go through, while still respecting Elasticsearch Webmock stubs.

### DIFF
--- a/spec/datadog/tracing/contrib/elasticsearch/transport_spec.rb
+++ b/spec/datadog/tracing/contrib/elasticsearch/transport_spec.rb
@@ -14,7 +14,7 @@ RSpec.describe 'Elasticsearch::Transport::Client tracing' do
     # TODO: Find why using testagent changes the elasticsearch test service name despite no differences between other tests
     Datadog.configuration.agent.host = 'not-testagent'
     WebMock.enable!
-    WebMock.disable_net_connect!
+    WebMock.allow_net_connect!
   end
 
   after do


spec/datadog/tracing/contrib/elasticsearch/transport_spec.rb
```diff
@@ -14,7 +14,7 @@
     # TODO: Find why using testagent changes the elasticsearch test service name despite no differences between other tests
     Datadog.configuration.agent.host = 'not-testagent'
     WebMock.enable!
-    WebMock.disable_net_connect!
+    WebMock.allow_net_connect!
```
ivoanjo, line 17: This seems reasonable but... is it only weird for me that we start telemetry during random specs?

Should we have a more generic "don't run telemetry during testing"?

(Also -- does this change impact the TODOs on line 13 and 14?)

ivoanjo (approved):
I kinda suspect we need a more generic fix for telemetry, but no reason to delay this fix until that one happens :)

spec/datadog/tracing/contrib/elasticsearch/transport_spec.rb
GustavoCaso: Should we remove the `WebMock.allow_net_connect!` from the `after` block?

marcotc: > Should we remove the `WebMock.allow_net_connect!` from the `after` block?

Not today.
This combination of webmock settings is how our spec_helper sets up all tests:
https://github.com/DataDog/dd-trace-rb/blob/d6fcf015a2eeaa7f9a853378d70bc757fc8a8f71/spec/spec_helper.rb#L60-L61

marcotc: > This seems reasonable but... is it only weird for me that we start telemetry during random specs?
> 
> Should we have a more generic "don't run telemetry during testing"?
> 
> (Also -- does this change impact the TODOs on line 13 and 14?)

Agreed!

PR #2992, opened 2023-07-22
Allow for dynamic log injection configuration

### LINKED ISSUE #2542: unable to disable log_injection
**Current behaviour**
logs include dd_trace details, like
```
[dd.service=finance_it dd.trace_id=470572845010387719 dd.span_id=2086160090212041436]   Rendered layouts/_some_page.html.haml (333.6ms)
```
even though `Datadog.configuration.tracing.log_injection` is set to false.

**Expected behaviour**
`Datadog.configuration.tracing.log_injection` should be respected

**Steps to reproduce**
set `Datadog.configuration.tracing.log_injection` to true, observe logs still contain the trace details.

**Environment**

* **ddtrace version: 1.8.0**
* **Ruby version: 2.5.7**
* **Operating system: OSX Ventura, 13.1**

** further **

This seems to be due to this code, https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/rails/patcher.rb#L45 , which runs before any of the configuration values can be set.

All I'm trying to do here is keep the logs less cluttered for development. Is there a recommended way to disable datadog for development? Should I just do it at the gemfile level?

marcotc:
Fixes #2542

As part of Dynamic Configuration, we want to allow users to enable or disable log injection at any time.

The main limitation we have in Ruby Tracing is that we need to patch an integration to enable log injection.

This PR enables all integrations that perform log correlation to always inject the correlation patch methods, but allow it to be safely disabled when `Datadog.configuration.tracing.log_injection` is set to `false` at runtime.

Rails, for example, was making decision at the application startup that couldn't be changed later. This PR addresses that.

We still depend on users having instrumented the integration itself for correlation to work.

Our `SemanticLogger` and `Lograge` integrations already support dynamic configuration of log injection; only `Rails` core logger and `ActiveJob` logging needed changes.

### DIFF
--- a/lib/datadog/tracing/contrib/rails/patcher.rb
+++ b/lib/datadog/tracing/contrib/rails/patcher.rb
@@ -41,12 +41,7 @@ module Datadog
               # Sometimes we don't want to activate middleware e.g. OpenTracing, etc.
               add_middleware(app) if Datadog.configuration.tracing[:rails][:middleware]
 
-              # Initialize Rails::Rack::Logger with a mutable taggers
-              # that can be modified by after_initialize hook
-              #
-              # https://github.com/rails/rails/blob/e88857bbb9d4e1dd64555c34541301870de4a45b/railties/lib/rails/rack/logger.rb#L16-L19
-              #
-              Contrib::Rails::LogInjection.set_mutatable_default(app)
+              Rails::LogInjection.configure_log_tags(app)
             end
           end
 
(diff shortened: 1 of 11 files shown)


lib/datadog/tracing/contrib/rails/patcher.rb
```diff
@@ -90,10 +73,6 @@ def after_intialize(app)
               # Finish configuring the tracer after the application is initialized.
               # We need to wait for some things, like application name, middleware stack, etc.
               setup_tracer
-
-              # `after_initialize` will respect the configuration from `config/initializers/datadog.rb`
-              # and add tags to `::ActiveSupport::TaggedLogging`
-              add_tags_to_logger(app) if Datadog.configuration.tracing.log_injection
```
marcotc, line 96: We don't have to try to inject tag logs at the `after_initialize` stage anymore because we always inject a safe injection proc early, at `before_initialize`.
This simplifies our logic quite a bit.

TonyCTHsu (approved):

PR #2993, opened 2023-07-24
[PROF-7440] Add fallback name/invoke location for unnamed threads started in native code

ivoanjo:
**What does this PR do?**:

This PR adds a fallback invoke location (which gets used as a name) for threads that a) Are unnamed, and b) Were started from native code.

The added invoke location is the fixed string ~`(Unnamed thread from native gem)`~ `(Unnamed thread from native code)`.

**Motivation**:

We previously (in #2950) started using invoke location for threads that do not have a name BUT when a thread is started from native code it would have an empty name.

Having an empty name is not particularly helpful, so I decided to instead have a simple placeholder that at least provides an hint on what the thread may be.

**Additional Notes**:

We don't yet do any resolving of symbols from native code, but in the future I guess we may try to see if the native code has a name at the C level, and use that as a fallback, but that's much more than a one-liner change :)

**How to test the change?**:

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
@@ -783,7 +783,7 @@ static void initialize_context(VALUE thread, struct per_thread_context *thread_c
     // If the first function of a thread is native code, there won't be an invoke location, so we use this fallback.
     // NOTE: In the future, I wonder if we could take the pointer to the native function, and try to see if there's a native
     // symbol attached to it.
-    snprintf(thread_context->thread_invoke_location, THREAD_INVOKE_LOCATION_LIMIT_CHARS, "%s", "(Unnamed thread from native gem)");
+    snprintf(thread_context->thread_invoke_location, THREAD_INVOKE_LOCATION_LIMIT_CHARS, "%s", "(Unnamed thread from native code)");
   }
 
   thread_context->thread_invoke_location_char_slice = (ddog_CharSlice) {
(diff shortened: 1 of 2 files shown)


ext/ddtrace_profiling_native_extension/collectors_thread_context.c
```diff
@@ -777,8 +779,11 @@ static void initialize_context(VALUE thread, struct per_thread_context *thread_c
       StringValueCStr(invoke_file_location),
       invoke_line_location
     );
-  } else {
-    snprintf(thread_context->thread_invoke_location, THREAD_INVOKE_LOCATION_LIMIT_CHARS, "%s", "");
+  } else if (thread != state->main_thread) {
+    // If the first function of a thread is native code, there won't be an invoke location, so we use this fallback.
+    // NOTE: In the future, I wonder if we could take the pointer to the native function, and try to see if there's a native
+    // symbol attached to it.
+    snprintf(thread_context->thread_invoke_location, THREAD_INVOKE_LOCATION_LIMIT_CHARS, "%s", "(Unnamed thread from native gem)");
```
marcotc, line 786: Is it always a thread from a "gem"? Could this thread have originated from the current application's native code?

marcotc (approved):

ext/ddtrace_profiling_native_extension/collectors_thread_context.c
ivoanjo: You know what, your question gave me quite a bit of pause.

This being Ruby, the answer is of course "everything is possible", but then I started thinking it would be kinda awkward it would be, since you wouldn't get the automatic compilation behavior on install like you get for gems. On the other hand, I can imagine if you need to interact with some weird hardware (like for a point of sale), you may end up doing that anyway.

In the end, as we already had another placeholder "In native code", I changed this one to be `(Unnamed thread from native code)` to match it in  c52c7329079e801dffac1256dd887ad283610ccd .

PR #2994, opened 2023-07-24
Fix:Options#using_default? on env var

marcotc:
Today, `Options#using_default?` always returns `false` if the first interaction with that option is a call to `Options#using_default?`.

This is an issue because the underlying `Option` did not have the chance to try to resolve its environment variable. If it did, it might actually not be using the `default:` value, but the `env:` value instead.

This PR ensures that `Options#using_default?` always returns a reliable result under every condition.


lib/datadog/core/configuration/options.rb
```diff
@@ -68,18 +68,15 @@ def options
           end
 
           def set_option(name, value, precedence: Configuration::Option::Precedence::PROGRAMMATIC)
-            add_option(name) unless options.key?(name)
-            options[name].set(value, precedence: precedence)
+            resolve_option(name).set(value, precedence: precedence)
```
marcotc, line 71: This is a small refactor to this file, because `add_option(name) unless options.key?(name)` was too common and not trivial to understand.

spec/datadog/core/configuration/options_spec.rb
```diff
@@ -255,6 +255,52 @@
...
+
+              it { is_expected.to be(false) }
+            end
+          end
+        end
+
+        context 'when the option is not defined' do
+          it { expect { using_default? }.to raise_error(described_class::InvalidOptionError) }
+        end
+      end
```
GustavoCaso, line 302:
```suggestion
describe '#using_default?' do
        subject(:using_default?) { options_object.using_default?(name) }

        let(:name) { :foo }

        context 'when the option is defined' do
          before { options_class.send(:option, name, meta) }

          let(:meta) { {} }

          context 'and a value is set' do
            before { options_object.set_option(name, 'something') }

            it { is_expected.to be(false) }
          end

          context 'and a value is not set' do
            context 'and no default value is configured' do
              it { is_expected.to be(true) }
            end

            context 'and a default value is configured' do
              let(:meta) { { default: 'anything' } }

              it { is_expected.to be(true) }

              context 'and an environment variable is configured' do
                let(:meta) { { default: 'anything', env: 'TEST_ENV_VAR' } }

                context 'and an environmet variable is set' do
                  around do |example|
                    ClimateControl.modify('TEST_ENV_VAR' => 'anything') { example.run }
                  end

                  it { is_expected.to be(false) }
                end

                context 'and an environment variable is not set' do
                  it { is_expected.to be(true) }
                end
              end
            end

            context 'an environment variable is configured' do
              let(:meta) { { env: 'TEST_ENV_VAR' } }

              context 'and an environmet variable is set' do
                around do |example|
                  ClimateControl.modify('TEST_ENV_VAR' => 'anything') { example.run }
                end

                it { is_expected.to be(false) }
              end

              context 'and an environment variable is not set' do
                it { is_expected.to be(true) }
              end
            end
          end
        end
```

The suggestion includes:
- Remove calls to `super` inside `let`
- Changed some test description to be more specific 
- Add a case for when the `env` option is set but the ENV variable is not configured 
- Add a context for `env` is defined without `default` value

GustavoCaso: Also, I wrote some specs on https://github.com/DataDog/dd-trace-rb/blob/37d556d1245e189fa840799f9995384fdb4aaf4a/spec/datadog/core/configuration/base_spec.rb#L145-L218

What do you think about removing those? I feel they test the same thing (without the `env` part) I think your specs are better suited to test the feature fully.

GustavoCaso (approved):
Thank you so much for fixing this @marcotc 🎉 

I left a few comments regarding the spec 😄

spec/datadog/core/configuration/options_spec.rb
marcotc: Thank you for the updates, applied!

> What do you think about removing those?

I think it's fine to remove those, as they seem a bit redundant like you said. I went ahead and did that.

PR #2996, opened 2023-07-25
Fix a format of CHANGELOG

y-yagi:
**What does this PR do?**

Fix the format of CHANGELOG of v1.12.1.

**before**

before

**after**

after

**Motivation**

To make changes easier to read.

**Additional Notes**

None

**How to test the change?**

Doc change only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,18 +8,18 @@
 ## [1.12.1] - 2023-06-14
 
 ### Added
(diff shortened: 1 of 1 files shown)


ivoanjo (approved):
Huh... seems like there was a small copy-paste accident there!

Thanks for noticing and sending a fix our way @y-yagi! I'll wait for CI to finish executing and then merge this :)

PR #3000, opened 2023-07-25
Disable `v1` schema version

zarirhamza:
After discussing with @delner, we determined to not release `v1` schema verison since it is not fully implemented yet. 

Nonetheless, all current features of v1 can be achieved by enabling the `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` environment variable.

Thus, we removed all references to v1 within `SpanAttributeSchema` and updated relevant tests to  not refer to `v1` but rather to use the `DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED` environment variable instead

With this, we can finally implement proper documentation as well.

Please do not merge without @delner approval, but feel free to review


GustavoCaso:
@delner @zarirhamza I would highly recommend removing the code for `V1` it is not fully implemented. 

The fact that it is not referenced by the code when calling `SpanAttributeSchema` doesn't mean it is not out there for library users to reference it on their code. 

Also, I'm a huge proponent of not adding code that would be later used, since there might be the case that it won't, and we might need to remember to clean up. If it happens that we end up needing it, we can always come back to GH and copy the code from the commits history. 

Having said this, I think we could have the `SpanAttributeSchema` module be something like: 

```ruby
# frozen_string_literal: true

require_relative 'ext'

module Datadog
  module Tracing
    module Contrib
      # Contains methods for fetching values according to span attributes schema
      module SpanAttributeSchema
        REFLEXIVE_SOURCES = [Tracing::Metadata::Ext::TAG_PEER_SERVICE].freeze
        NO_SOURCE = [].freeze

        def self.default_span_attribute_schema?
        end

          ENV.fetch(env) do
            return Datadog.configuration.service if Datadog.configuration.tracing.global_default_service_name.enabled

            default
          end
        end

        def self.set_peer_service(span, sources)
          # Check if peer.service value is actually set prior to checking for remapping
          if (peer_service_val = set_peer_service_from_source(span, sources))
            if (remap_val = Datadog.configuration.tracing.peer_service_mapping[peer_service_val.to_sym])
              span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE, remap_val)
              span.set_tag(Tracing::Contrib::Ext::Metadata::TAG_PEER_SERVICE_REMAP, peer_service_val)
            end
            return true
          end
          false
        end

        # set_peer_service_from_source: Implements the extraction logic to determine the peer.service value
        # based on the list of source tags passed as a parameter.
        #
        # If no values are found, it checks the default list for all spans before returning false for no result
        # Sets the source of where the information for peer.service was extracted from
        # Returns a peer.service value if successfully set or not
        def self.set_peer_service_from_source(span, sources = [])
          # Filter out sources based on existence of peer.service tag
          sources = filter_peer_service_sources(span, sources)

          # Find a possible peer.service source from the list of source tags passed in
          sources.each do |source|
            next unless not_empty_tag?(source_val)

            span.set_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE, source_val)
            span.set_tag(Tracing::Contrib::Ext::Metadata::TAG_PEER_SERVICE_SOURCE, source)
            return source_val
          end
          false
        end

        private_class_method :set_peer_service_from_source

        # filter_peer_service_sources: returns filtered sources based on existence of peer.service tag
        # If peer.service exists, we do not read from any other source rather use peer.service as source
        # This is to prevent overwriting of pre-existing peer.service tags
        def self.filter_peer_service_sources(span, sources)
          ps = span.get_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE)
          # Do not override existing peer.service tag if it exists based on schema version
          return REFLEXIVE_SOURCES if not_empty_tag?(ps)

          # Check that peer.service is only set on spankind client/producer spans
          if (span.get_tag(Tracing::Metadata::Ext::TAG_KIND) == Tracing::Metadata::Ext::SpanKind::TAG_CLIENT) ||
              (span.get_tag(Tracing::Metadata::Ext::TAG_KIND) == Tracing::Metadata::Ext::SpanKind::TAG_PRODUCER)
            return sources
          end

          NO_SOURCE
        end

        private_class_method :filter_peer_service_sources

        def self.not_empty_tag?(tag)
          tag && (tag != '')
        end

        private_class_method :not_empty_tag? 
      end
    end
  end
end
```

Also, I suggest removing all the skip tests as well.

zarirhamza:
I agree, the more I think about it, the idea of just leaving unused code in the repo seems unsettling to me. Functionality wise there are no changes, so I think its fine if we implement @GustavoCaso's changes and remove all v1 related testing as well. There should be no reference to the constant `v1` in the repo.

lib/datadog/tracing/contrib/extensions.rb
```diff
@@ -108,6 +108,46 @@ def configure(&block)
...
+            def self.included(base)
+              base.class_eval do
+                settings :contrib do
+                  # Schema version for span attributes that enables various features. Only returns 'v0' for now
+                  #
+                  # @default `v0`
+                  # @return [String]
+                  option :span_attribute_schema do |o|
+                    o.default 'v0'
+                  end
```
GustavoCaso, line 120: We only use this value on `spec/support/network_helpers.rb`. Should we remove the possibility of someone setting a value different `v0` altogether and leave the `v0` value hardcoded  on `spec/support/network_helpers.rb`? Mainly to avoid users using a setting that has no functionality yet. And avoid the possibility to have issues once we actually introduce different version 😄

lib/datadog/tracing/contrib/extensions.rb
```diff
@@ -108,6 +108,46 @@ def configure(&block)
           module Settings
             InvalidIntegrationError = Class.new(StandardError)
 
+            def self.included(base)
+              base.class_eval do
+                settings :contrib do
```
GustavoCaso, line 113: Mostly curious about the `contrib` namespace. Is the `contrib` namespace part of the specs? Also, could we add the settings directly on `lib/datadog/tracing/configuration/settings.rb` ?

GustavoCaso:
@zarirhamza Thank you so much for taking into consideration my suggestion 😄 

The code LGTM, but as I do not have a full understanding of the feature in place, I would delegate to others for the ✅

lib/datadog/tracing/contrib/extensions.rb
zarirhamza: `tracing/contrib` and `tracing/configuration` are actually two separate things. We want to avoid coupling since `tracing/configuration` relates to configuration details for ALL of tracing whereas `tracing/contrib` affects only contrib related libraries. 

`peer.service` and `service_name` are related solely to integrations under the `contrib` namespace, so I think it's best if we leave those setting here. Some customers desire a core tracing package (with only tracing capabilities without contrib integrations) that doesn't include any unnecessary code. 

The code is being tested as well underneath `extensions_spec.rb` corresponding to the file

lib/datadog/tracing/contrib/extensions.rb
zarirhamza: Implemented!

lib/datadog/tracing/contrib/extensions.rb
GustavoCaso: That makes sense. 

Thanks for the explanation 😄

delner (approved):

PR #3002, opened 2023-07-26
Remove unused Barrier#wait_next functionality

ivoanjo:
**What does this PR do?**:

This PR removes the `Barrier#wait_next` functionality which we ended up not making use of.

**Motivation**:

In #2990 we ran into issues with its current specs, and since we're not using this functionality, it seems like removing it for now seems to be easier than investing more time to fix flakiness in the specs.

**Additional Notes**:

N/A

**How to test the change?**:

Existing tests already cover the `Barrier` APIs.

### DIFF
--- a/lib/datadog/core/remote/component.rb
+++ b/lib/datadog/core/remote/component.rb
@@ -62,13 +62,9 @@ module Datadog
           end
         end
 
-        def barrier(kind)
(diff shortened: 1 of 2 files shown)


lloeki (approved):
LGTM

PR #3003, opened 2023-07-26
Continue with remote configuration when agent is not reachable

lloeki:
**What does this PR do?**

Continue with remote configuration when agent is not reachable or has no RC enabled 

**Motivation**

If the agent is not available on startup, RC will bail out and never be able to recover even if the agent appears or gets fixed down the road, thus requiring an app restart.

**Additional Notes**

If the agent is not available or RC is not enabled in the agent this will continuously produce error logs. This is expected as there is misconfiguration that the user must address (either by disabling RC or by fixing the agent)

**How to test the change?**

- start app with rc enabled and without agent (or with agent lacking rc)
- a log message should appear that RC is not as expected
- start rc thread by hitting the app
- error messages should appear continously
- start agent with RC enabled
- RC should recover

### DIFF
--- a/lib/datadog/core/remote/component.rb
+++ b/lib/datadog/core/remote/component.rb
@@ -123,15 +123,11 @@ module Datadog
 
             negotiation = Negotiation.new(settings, agent_settings)
 
-            unless negotiation.endpoint?('/v0.7/config')
-              Datadog.logger.error do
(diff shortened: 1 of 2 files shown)


lloeki:
8f596b56f4f46f51a2ebea88d430f5ee0e6dcaf2 has been merged with a similar effect. Closing.

PR #3004, opened 2023-07-26
Refactor EnvironmentLogger V2

sarahchen6:
**What does this PR do?**
Unfinished version.. not passing any tests, and tests are uncompleted! 

General idea is to move data extraction components to respective products (i.e. collect tracing data under Tracing namespace and collect profiling data under Profiling namespace). Everything is still logged under Core.

**Motivation**

**Additional Notes**

**How to test the change?**


sarahchen6:
See #3020

PR #3008, opened 2023-07-28
Start RC on demand

lloeki:
**What does this PR do?**

Start RC on demand.

**Motivation**

System tests fail on Sinatra because of corrupted RC state + unfortunate timing when a forking puma is involved. Starting on demand ensures that it happens in the process that needs RC + that no resource is wasted when RC is not needed.

**Additional Notes**

The above can be reevaluated.

**How to test the change?**

System tests should pass

### DIFF
--- a/spec/datadog/core/configuration/components_spec.rb
+++ b/spec/datadog/core/configuration/components_spec.rb
@@ -1098,15 +1098,6 @@ RSpec.describe Datadog::Core::Configuration::Components do
         let(:info_response) { { endpoints: ['/v0.7/config'] }.to_json }
       end
 
-      context 'enabled' do
-        before { allow(settings.remote).to receive(:enabled).and_return(true) }
-
-        it 'starts the remote manager' do
-          expect(components.remote).to receive(:start)
-          startup!
-        end
-      end
-
       context 'disabled' do
         before { allow(settings.remote).to receive(:enabled).and_return(false) }
 
--- a/lib/datadog/core/configuration/components.rb
+++ b/lib/datadog/core/configuration/components.rb
@@ -110,9 +110,6 @@ module Datadog
           else
             @logger.debug('Profiling is disabled')
           end
-
-          # Start remote configuration worker
-          @remote.start if @remote
         end
 
         # Shuts down all the components in use.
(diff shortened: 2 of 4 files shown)


GustavoCaso (approved):
LGTM. Great find

spec/datadog/core/configuration/components_spec.rb
```diff
@@ -1098,15 +1098,6 @@
...
       end
 
-      context 'enabled' do
-        before { allow(settings.remote).to receive(:enabled).and_return(true) }
-
-        it 'starts the remote manager' do
-          expect(components.remote).to receive(:start)
-          startup!
-        end
-      end
```
ivoanjo, line 1108: I wonder if we should have a negative test as well. I don't often advocate for those, but since this seems an important detail, it seems relevant to make sure a test fails if someone ever remembers to re-add something like the line that was removed?

lloeki: that would make sense

lloeki:
Relevant tests now pass, previous behaviour is restored (lazy-start RC worker when AppSec middleware receives a request).

This behaviour is not correct for tracing, which as expected now shows in the new RC integration spec failures; attempting to address that by lazy-starting in the same way as AppSec for Tracing, which will cater for the http request case.

This will not cover non-web thing like Sidekiq & al, but the goal right now is to restore CI sanity to work from solid ground. If we can get to green right now we can merge and address behaviour in later PRs.

lloeki:
Remaining failures regarding `rid` should be addressed by https://github.com/DataDog/dd-trace-rb/pull/3005 which is co-dependent.

spec/datadog/core/configuration/components_spec.rb
lloeki: Trying to reach green ASAP RN so I won't do that in this PR.

GustavoCaso:
Reminder that we might/should merge to fix the header issues with sinatra apps in system tests https://github.com/DataDog/system-tests/pull/1430

lloeki:
Ah, I might have been confused and that might be the one actually, thanks.

lloeki:
Merging, so that at least things are correct for web things now.

spec/datadog/core/configuration/components_spec.rb
ivoanjo: Good call, not a blocker at all!

PR #3012, opened 2023-07-28
Improve integration testing for dynamic configuration

marcotc:
This PR makes the integration tests for dynamic configuration a bit less noisy and a bit less flaky.

The tests only fail for JRuby 9.2, but the noisy threads is for all versions of Ruby.

### DIFF
--- a/spec/datadog/tracing/contrib/suite/integration_spec.rb
+++ b/spec/datadog/tracing/contrib/suite/integration_spec.rb
@@ -235,9 +235,8 @@ RSpec.describe 'contrib integration testing' do
       context 'for log_injection_enabled' do
         let(:tracing_sampling_rate) { 0.0 }
         let(:io) { StringIO.new }
-        let(:appender) { SemanticLogger.add_appender(io: io) }
         let(:logger) do
-          appender
+          SemanticLogger.add_appender(io: io)
           SemanticLogger['TestClass']
         end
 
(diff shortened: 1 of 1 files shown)


spec/datadog/tracing/contrib/suite/integration_spec.rb
```diff
@@ -39,7 +39,12 @@
       Datadog.configure { |c| c.remote.poll_interval_seconds = 0.001 }
     end
 
-    after { WebMock.disable! }
+    after do
+      # Ensure RC background worker is stopped before we disable webmock
+      # to avoid failed HTTP requests, trying to make a real remote call.
+      Datadog.shutdown!
```
marcotc, line 45: This currently generates an error in a background thread, after the test assertions finish, because the RC worker thread tries to make an unmocked RC call after we call `WebMock.disable!`. This PR ensures the RC worker is closed before `WebMock.disable!` is called.

spec/datadog/tracing/contrib/suite/integration_spec.rb
```diff
@@ -242,7 +247,10 @@ def build(*dynamic_configurations)
           end
         end
 
-        after { SemanticLogger.remove_appender(appender) }
+        after do
+          SemanticLogger.remove_appender(appender)
+          SemanticLogger::Logger.processor.thread.kill
```
marcotc, line 252: This is a leaky thread, emitting warnings on test run.

ekump (approved):

TonyCTHsu (approved):

PR #3013, opened 2023-07-31
Bump stuff in nix shell

lloeki:
**What does this PR do?**

- Bump nix channel
- Bump ruby to 3.2
- Add simple test in CI to ensure no breakage happens
- Introduce Nix derivation caching via cachix (this adds a secret to push to cachix)

**Motivation**

Keeping things up to date

**Additional Notes**

Derivations are cached publicly, therefore care must be taken not to store secrets as files, but that's already the current practice knowing that:

- it's a public repo with public CI
- we generate docker images that could capture the filesystem state and are published
- all secrets should be CI env vars

**How to test the change?**

- Manually

```
nix-shell
bundle install
# do stuff
```

- CI

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -18,6 +18,10 @@ jobs:
           - unstable
     name: Test Nix (${{ matrix.os }}, ${{ matrix.nix }})
     runs-on: ${{ matrix.os }}
+    env:
+      SKIP_SIMPLECOV: 1
+      DD_INSTRUMENTATION_TELEMETRY_ENABLED: false
+      DD_REMOTE_CONFIGURATION_ENABLED: false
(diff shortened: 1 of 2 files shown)


ivoanjo (approved):
LGTM :+1: 

Since I'm the only one on the team on ubuntu, I would suggest maybe removing the ubuntu testing -- it's one less CI configuration to take time on our PRs.

(That also works around the issue with the current profiler compilation incompatibilities with nixos. At least on my local setting there seems to be several incompatibilities with the weird way that nix builds its ruby image, and with no reported customers currently on nix I'm not sure it's worth the investment to fix.)

ivoanjo:
@lloeki is this still relevant?

PR #3017, opened 2023-08-01
ddtrace gem 0.x does not support for Ruby 3.2+

tnir:
**What does this PR do?**

Changes in #2971 are incorrect. This indicates Ruby 3.2 is now supported only with ddtrace gem 1.0.0 or higher.

**Motivation**

Accuracy is required.

**Additional Notes**

https://rubygems.org/gems/ddtrace/versions/0.54.2

**How to test the change?**

n/a

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -122,7 +122,7 @@ To contribute, check out the [contribution guidelines][contribution docs] and [d
 
 | Type  | Documentation              | Version | Support type                         | Gem version support |
 | ----- | -------------------------- | -----   | ------------------------------------ | ------------------- |
-| MRI   | https://www.ruby-lang.org/ | 3.2     | Full                                 | Latest              |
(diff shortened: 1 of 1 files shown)


ivoanjo:
Hey @tnir :wave: 

The intention of that table is to show, for old Rubies, which versions of dd-trace-rb you should run, and what level of support you should expect.

Our intention is that "Latest" means the latest release (1.13.0 at time of writing), which is the version we recommend our customers are always on (unless they have a legacy Ruby). We never suggest staying on any older version, as once a new version comes out, we don't fix or patch older versions.

Thus, we prefer to keep this as "Latest" instead of indicating `> 1.0.0` because none of the other versions are recommended.

Hopefully that clarifies why we don't want to change the current doc?

estherk15 (commented):
No docs review needed.

ivoanjo:
Let's close this for now, I think the above discussion on why we recommend the latest and never any specific old version (other than for deprecated Rubies) I believe clarifies our position enough.

tnir:
Okay, I see.

PR #3018, opened 2023-08-01
Implements `_dd.base_service` tag

zarirhamza:
Implements `_dd.base_service` tag which captures the global service name IFF the span.service is not equivalent to the global service name, implying it was overridden and a fake service was implemented.


zarirhamza:
System tests were changed a bit during development so after consulting @wconti27, I applied the fix for that in this PR as well. No impact should be made to functionality

delner (approved):
@zarirhamza and I reviewed this together. May need a refactor later to reduce repetition, create consistency, but approach seems "OK" for now.

PR #3019, opened 2023-08-01
Register the OpenSearch contrib integration

segiddins:
Otherwise, attempting to use the integration gives the following error:

```
/Users/segiddins/.gem/ruby/3.2.2/gems/ddtrace-1.13.0/lib/datadog/tracing/contrib/extensions.rb:210:in `fetch_integration': 'opensearch' is not a valid integration. (Datadog::Tracing::Contrib::Extensions::Configuration::Settings::InvalidIntegrationError)
```

**What does this PR do?**

**Motivation**
Failures on https://github.com/rubygems/rubygems.org/pull/3976

**Additional Notes**

**How to test the change?**


ivoanjo:
(The CI issue is fixed by #3021 -- there was a new release of the rake-compiler gem which we don't pin since it's only a development dependency, and broke older Rubies)

TonyCTHsu (approved):

PR #3022, opened 2023-08-02
Fix hash log_tags warning

TonyCTHsu:
**What does this PR do?**

Skip warning for log injection configuration when Semantic Logger's named tags override Rails built-in config.log_tags with a hash value


marcotc (approved):

lib/datadog/tracing/contrib/rails/log_injection.rb
```diff
@@ -8,9 +8,12 @@ module LogInjection
           module_function
 
           # Use `app.config.log_tags` to inject propagation tags into the default Rails logger.
-          def configure_log_tags(app)
-            app.config.log_tags ||= [] # Can be nil, we initialized it if so
-            app.config.log_tags << proc { Tracing.log_correlation if Datadog.configuration.tracing.log_injection }
+          def configure_log_tags(app_cofig)
```
marcotc, line 11:
```suggestion
          def configure_log_tags(app_config)
```

marcotc (approved):

PR #3024, opened 2023-08-02
Run minitest integration test on CI

TonyCTHsu:
**What does this PR do?**

Add minitest integration test in our test matrix, which will be picked up on CI.


marcotc (approved):
Did you mean to have the target git branch as `1.13.1` instead?

PR #3026, opened 2023-08-04
Improve OpenTelemetry integration compatiblity

marcotc:
**What does this PR do?**

This PR adds changes that ensure the OpenTelemetry integration passes all relevant system-tests (https://github.com/DataDog/system-tests/pull/1456).

This improves conformity for OpenTelemetry -> Datadog tracer translation.

**Motivation**

OpenTelelemtry system-tests ensure that all requires features for GA are implemented by all traces.
Before this PR, the Ruby trace did not conform to all required OpenTelemetry -> Datadog API translations.

**Additional Notes**

**How to test the change?**

The best way is to run https://github.com/DataDog/system-tests/pull/1456 pointing to this branch.

The integration test in this repo act more like unit tests, given the complex cases are captured in system-tests.


lib/datadog/opentelemetry/trace.rb
```diff
@@ -0,0 +1,60 @@
...
+          digest = if parent_span
+                     digest_with_parent_span(trace, parent_span)
+                   else
+                     trace.to_digest
+                   end
+
+          # Create a new TraceOperation, attached to the current Datadog Tracer.
+          # This is important because trace flushing is performed by the active
+          # tracer instance.
+          Tracing.send(:tracer).send(:start_trace, continue_from: digest)
```
TonyCTHsu, line 26: Could you elaborate the difference between this approach and leveraging our public API, such as 
`Datadog::Tracing.continue_trace!(digest)`?

lib/datadog/opentelemetry/sdk/trace/span.rb
```diff
@@ -0,0 +1,67 @@
...
+
+          # Status code can only change into an error state.
+          # Other change operations should be ignored.
+          span.set_error(status.description) if status && status.code == ::OpenTelemetry::Trace::Status::ERROR
+        end
+
+        private
+
+        def datadog_set_attribute(key)
+          return unless @attributes # Return if attributes are currently disabled by OpenTelemetry.
```
TonyCTHsu, line 42: Does it make also check if `@attributes` could be empty?

lib/datadog/opentelemetry/trace.rb
marcotc: It's not needed actually, I think it was required during development, but I changed it back to using ` Datadog::Tracing.continue_trace!(digest)` and it passes tests.

lib/datadog/opentelemetry/sdk/trace/span.rb
marcotc: I made so that it checks for the existence of the variable.

marcotc:
@TonyCTHsu can you take another look here?

TonyCTHsu (approved):

PR #3032, opened 2023-08-08
Add extra service configuration for RC

lloeki:
**What does this PR do?**

Add extra service configuration for RC

**Motivation**

Service names can be mixed and matched in many ways, sometimes in complex ways that cannot be automatically inferred. For this information to be properly relayed in the most complex cases manual configuration is necessary.

**Additional Notes**

It may be possible to automate population of this field in the future (depends on configuration system evolution).

**How to test the change?**

Here's a typical configuration example, where:

- main service name is different from sub-services provided by integrations
- remote service name is made to match main service name
- rack and rails service names are different, and subject to remote configuration for this process as well
- request queuing is handled, creates a synthetic span which belongs to another service in another process, and so is not provided as extra service for RC.

```
  Datadog.configure do |c|
    c.service = 'web'

    c.tracing.enabled = true
    c.tracing.instrument :rack, service_name: 'web-rack', request_queuing: true, web_service_name: 'nginx'
    c.tracing.instrument :rails, service_name: 'web-rails'

    c.remote.enabled = true
    c.remote.service = 'web'
    c.remote.extra_services = ['web-rack', 'web-rails']
  end
```

### DIFF
--- a/lib/datadog/core/configuration/settings.rb
+++ b/lib/datadog/core/configuration/settings.rb
@@ -646,23 +646,13 @@ module Datadog
           # @default `nil`.
           # @return [Array<String>,nil]
           option :extra_services do |o|
+            o.type :array
+            o.default []
             o.setter do |v|
-              if v.nil?
-                []
-              elsif !v.is_a?(Array)
-                Datadog.logger.warn(
-                  "Unexpected remote.extra_services type #{v.class}, expected Array. " \
-                  'Using default value `[]`'
-                )
+              if v.any? { |e| !e.is_a?(String) }
+                t = v.find { |e| !e.is_a?(String) }
 
-                []
-              elsif (t = v.find { |e| !e.is_a?(String) })
-                Datadog.logger.warn(
-                  "Unexpected remote.extra_services element type #{t.class}, expected String. " \
-                  'Using default value `[]`'
-                )
-
-                []
+                raise ArgumentError, "Unexpected remote.extra_services element type #{t.class}, expected String."
               else
                 v
               end
--- a/lib/datadog/core/configuration/settings.rb
+++ b/lib/datadog/core/configuration/settings.rb
@@ -635,6 +635,13 @@ module Datadog
           # @default `nil`.
           # @return [String,nil]
           option :service
+
+          # Declare additional service names to bind to remote configuration. Use when
+          # integrations are configured to use names that differ from DD_SERVICE.
+          #
+          # @default `nil`.
+          # @return [Array<String>,nil]
+          option :extra_services
         end
 
         # TODO: Tracing should manage its own settings.
--- a/lib/datadog/core/configuration/settings.rb
+++ b/lib/datadog/core/configuration/settings.rb
@@ -634,14 +634,36 @@ module Datadog
           #
           # @default `nil`.
           # @return [String,nil]
-          option :service
+          option :service do |o|
+            o.type :string, nilable: true
+          end
 
           # Declare additional service names to bind to remote configuration. Use when
           # integrations are configured to use names that differ from DD_SERVICE.
           #
           # @default `nil`.
           # @return [Array<String>,nil]
-          option :extra_services
+          option :extra_services do |o|
+            o.setter do |v|
+              if !v.is_a?(Array)
+                Datadog.logger.warn(
+                  "Unexpected remote.extra_services type #{v.class}, expected Array. " \
+                  'Using default value `[]`'
+                )
+
+                []
+              elsif (t = v.find { |e| !e.is_a?(String) })
+                Datadog.logger.warn(
+                  "Unexpected remote.extra_services element type #{t.class}, expected String. " \
+                  'Using default value `[]`'
+                )
+
+                []
+              else
+                v
+              end
+            end
+          end
         end
 
         # TODO: Tracing should manage its own settings.
--- a/lib/datadog/core/configuration/settings.rb
+++ b/lib/datadog/core/configuration/settings.rb
@@ -645,7 +645,9 @@ module Datadog
           # @return [Array<String>,nil]
           option :extra_services do |o|
             o.setter do |v|
-              if !v.is_a?(Array)
+              if v.nil?
+                []
+              elsif !v.is_a?(Array)
                 Datadog.logger.warn(
                   "Unexpected remote.extra_services type #{v.class}, expected Array. " \
                   'Using default value `[]`'
--- a/lib/datadog/core/configuration/settings.rb
+++ b/lib/datadog/core/configuration/settings.rb
@@ -638,8 +638,10 @@ module Datadog
             o.type :string, nilable: true
           end
 
-          # Declare additional service names to bind to remote configuration. Use when
-          # integrations are configured to use names that differ from DD_SERVICE.
+          # Declare additional service names to fold under the main service
+          # name used by remote configuration. Use when integrations are
+          # configured to use names that differ from DD_SERVICE and should be
+          # considered as a single remote configuration controllable service.
           #
           # @default `nil`.
           # @return [Array<String>,nil]
--- a/spec/datadog/core/remote/client_spec.rb
+++ b/spec/datadog/core/remote/client_spec.rb
@@ -588,6 +588,24 @@ RSpec.describe Datadog::Core::Remote::Client do
               end
             end
 
+            context 'with remote extra_services setting' do
+              it 'returns client_tracer' do
+                expect(Datadog.configuration.remote).to receive(:service).and_return('foo').at_least(:once)
+                expect(Datadog.configuration.remote).to receive(:extra_services).and_return(['foo']).at_least(:once)
+
+                expected_client_tracer = {
+                  :runtime_id => Datadog::Core::Environment::Identity.id,
+                  :language => Datadog::Core::Environment::Identity.lang,
+                  :tracer_version => Datadog::Core::Environment::Identity.tracer_version_semver2,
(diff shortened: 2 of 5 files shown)


delner:
What is `extra_services`? It wasn't clear from my immediate read. I'm also curious how this lines up against some of our other service name changes.

lib/datadog/core/configuration/settings.rb
```diff
@@ -611,14 +611,36 @@ def initialize(*_)
...
+            o.type :string, nilable: true
+          end
 
           # Declare additional service names to bind to remote configuration. Use when
           # integrations are configured to use names that differ from DD_SERVICE.
           #
           # @default `nil`.
           # @return [Array<String>,nil]
-          option :extra_services
+          option :extra_services do |o|
```
ivoanjo, line 623: Since `extra_services` also effectively includes `service`, should we deprecate `service` and ask people to use `extra_services` instead?

spec/datadog/core/remote/client_spec.rb
```diff
@@ -588,6 +588,24 @@
               end
             end
 
+            context 'with remote extra_services setting' do
+              it 'returns client_tracer' do
+                expect(Datadog.configuration.remote).to receive(:service).and_return('foo').at_least(:once)
```
lloeki, line 593: I don't quite get why this is necessary.

Without it it fails with:

```
  1) Datadog::Core::Remote::Client#sync #payload no sync errors client client_tracer with remote extra_services setting returns client_tracer
     Failure/Error: expect(client_payload[:client_tracer].tap { |h| h.delete(:tags) }).to eq(expected_client_tracer)
     
       expected: {:env=>nil, :extra_services=>["foo"], :language=>"ruby", :runtime_id=>"803326d7-3898-48eb-8e73-c4b2854fa7df", :service=>nil, :tracer_version=>"1.14.0"}
            got: {:env=>nil, :extra_services=>["foo"], :language=>"ruby", :runtime_id=>"803326d7-3898-48eb-8e73-c4b2854fa7df", :service=>"rspec", :tracer_version=>"1.14.0"}
     
       (compared using ==)
     
       Diff:
       @@ -2,6 +2,6 @@
        :extra_services => ["foo"],
        :language => "ruby",
        :runtime_id => "803326d7-3898-48eb-8e73-c4b2854fa7df",
       -:service => nil,
       +:service => "rspec",
        :tracer_version => "1.14.0",
```

lib/datadog/core/configuration/settings.rb
lloeki: The duplication is sorted out backend-side.

Hopefully `extra_services` gets to disappear if we manage to figure a way to automate it, while `service` could still be needed for a while. I would not want the user configuration to break then.

spec/datadog/core/remote/client_spec.rb
marcotc: `rspec` is the fallback service name: https://github.com/DataDog/dd-trace-rb/blob/ff32e137b07d261d10179ea22176a875654521a7/lib/datadog/core/environment/ext.rb#L26

and it might be coming from here: https://github.com/DataDog/dd-trace-rb/blob/ff32e137b07d261d10179ea22176a875654521a7/lib/datadog/tracing/tracer.rb#L51

lloeki: What I don't get is why it consistently works for the other spec examples but consistently not for this one 🤔

lloeki: What I don't get is why it consistently works for the other spec examples but consistently not for this one 🤔

lib/datadog/core/configuration/settings.rb
lloeki: To clarify: since this can be tricky it would be in scope for a second PR following this one.

lib/datadog/core/configuration/settings.rb
```diff
@@ -634,7 +634,38 @@ def initialize(*_)
...
-          option :service
+          option :service do |o|
+            o.type :string, nilable: true
+          end
+
+          # Declare additional service names to bind to remote configuration. Use when
+          # integrations are configured to use names that differ from DD_SERVICE.
+          #
+          # @default `nil`.
+          # @return [Array<String>,nil]
```
marcotc, line 645: If we would like to make this option not public, let's add the `# @!visibility private` tag to it.

lloeki: Good point, although I'm not sure making it non-public is a good thing either, so I tried to describe it in a more user friendly way.

ivoanjo: What does this mean for an option like this one to be "non-public"?

lib/datadog/core/configuration/settings.rb
```diff
@@ -634,7 +634,40 @@ def initialize(*_)
...
+
+          # Declare additional service names to fold under the main service
+          # name used by remote configuration. Use when integrations are
+          # configured to use names that differ from DD_SERVICE and should be
+          # considered as a single remote configuration controllable service.
+          #
+          # @default `nil`.
+          # @return [Array<String>,nil]
+          option :extra_services do |o|
+            o.setter do |v|
```
GustavoCaso, line 649: You could use `o.type :array, nilable: true` that way you can remove the `elsif !v.is_a?(Array)` branch from the `setter` block

lib/datadog/core/configuration/settings.rb
GustavoCaso: The description is clear enough of what the code would do.

GustavoCaso: > What does this mean for an option like this one to be "non-public"?

By making it not public, we can change the name/implementation without worrying about breaking the customer's application.

We could name it something like `experimental_extra_services` to let customers know that this option is somewhat not fully clear. Still we will need help to rename it easily in the future. 

What about adding as part of the comment that the use of this option is only meant for edge cases in which RC is not working due to lack of understanding between the tracer and the backend, and that we might remove it in the future once is no longer necessary?

ivoanjo: > What about adding as part of the comment that the use of this option is only meant for edge cases in which RC is not working due to lack of understanding between the tracer and the backend, and that we might remove it in the future once is no longer necessary?

This looks a bit weird -- seems like a "public-ish/private-ish"? My question was exactly that it's weird to add a version and for it to not be public.

We may still recommend strongly that people don't use it (via docs, logs, etc), but if it's in the settings, it seems to me that it would automatically be public?

marcotc (approved):
I'm ok with it, as long as it is either hidden or public and well documented (as situational, advanced option).

lib/datadog/core/configuration/settings.rb
lloeki: I agree with Ivo, e.g having support direct users to use a non-public API sounds odd; I can see users feeling less confident with that than if it is public.

lib/datadog/core/configuration/settings.rb
lloeki: I was under the impression that the `setter` would either override `o.type` behaviour or be called after.

Also somehow I kind of preferred having the logic fully laid out in a single place. With `o.type :array, nilable: true` it becomes a bit unclear to me how corner cases would operate.

That makes me think, maybe it should not even be `nilable` as well as `o.default` to `[]`, the idea being to guarantee that when calling the getter you get an `Array`, even if empty.

lib/datadog/core/configuration/settings.rb
```diff
@@ -646,23 +646,13 @@ def initialize(*_)
...
-                )
-
-                []
-              elsif (t = v.find { |e| !e.is_a?(String) })
-                Datadog.logger.warn(
-                  "Unexpected remote.extra_services element type #{t.class}, expected String. " \
-                  'Using default value `[]`'
-                )
+              if v.any? { |e| !e.is_a?(String) }
+                t = v.find { |e| !e.is_a?(String) }
```
lloeki, line 653: The previous code had an interesting corner case: with `[nil]` it would return `nil`, which would be falsey, thus pass the condition and not produce the expected error!

The double iteration is fine since it's on the unhappy path.

lib/datadog/core/configuration/settings.rb
```diff
@@ -646,23 +646,13 @@ def initialize(*_)
...
 
-                []
+                raise ArgumentError, "Unexpected remote.extra_services element type #{t.class}, expected String."
```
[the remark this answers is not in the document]
GustavoCaso: > we should support at least simple element runtime typing, e.g using o.type 'Array[String]' RBS syntax

That could be an excellent north start for our DSL, but I would only place a little effort now. Having a complex type-checking system is a huge task. For now, the idea behind `o.type :array` without the types inside the array forces us to move the logic to the setter block. Once we see enough use cases for a more sophisticated type check system we could add it. 

> we should attempt to unify to raising instead of logging when items fail the runtime typecheck

The internal type check system already raises https://github.com/DataDog/dd-trace-rb/blob/40778dbaa753510eb37c3322101b0eab49d89965/lib/datadog/core/configuration/option.rb#L214

As to when custom setter blocks fails, I think is up to the implementer to decided

lib/datadog/core/configuration/settings.rb
lloeki: Attempted here: https://github.com/DataDog/dd-trace-rb/pull/3168

ivoanjo:
I've discussed this with Loic, and since we ended up not using it, I'm going to go ahead and close this PR for now :)

PR #3035, opened 2023-08-08
Use Fiber instead of Thread for IntervalLoop spec

delner:
**What does this PR do?**

Changes the `IntervalLoop` spec to use a `Fiber` instead of a `Thread`.

**Motivation**

This spec is considered flaky; it appears to hang often in CI, causing intermittent failures. Opting for a fiber should hopefully allow for a faster, more consistent test.

**Additional Notes**

It should not affect any production code; the use of `Fiber` is only in the test suite, not in the main library.

**How to test the change?**

`bundle exec rake spec:main`

### DIFF
--- a/spec/datadog/core/workers/interval_loop_spec.rb
+++ b/spec/datadog/core/workers/interval_loop_spec.rb
@@ -35,16 +35,26 @@ RSpec.describe Datadog::Core::Workers::IntervalLoop do
       end
     end
 
-    shared_context 'perform loop in thread' do
-      before do
-        # Start the loop in a thread, give it time to warm up.
-        @thread = Thread.new { worker.perform }
-        sleep(0.1)
+    shared_context 'perform loop in fiber' do
+      let(:fiber) do
+        Fiber.new do
+          worker.perform
+        end
       end
 
-      after do
-        @thread.kill
-        @thread.join
+      let(:task) do
+        proc do |*args|
+          worker_spy.perform(*args)
+          Fiber.yield
+        end
+      end
+
+      before do
+        # Start the loop in a fiber.
+        fiber.resume
+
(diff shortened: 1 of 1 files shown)


marcotc (approved):

ivoanjo:
I'm not sure about this change... The whole point of `IntervalLoop` is to be used in a background thread, and so I worry that by changing this to a `Fiber` we're not testing the class in the same way it's supposed to actually be used in production.

Even if we're reasonably happy with what it does now, what if someone next week changes the class, and our tests don't catch it because the change works on a `Fiber` but not on a `Thread`?

ivoanjo:
Fair enough... Although I'm still worried about promoting a style of testing threaded code that's very different from the actual way we use it.

> With (1), I have no guarantee, as I'm at the mercy of the scheduler. I'm just hoping it ran through the loop once (after giving it a headstart) before asserting.

One alternative I would suggest is keeping the `Thread`, but actually coordinating with it, e.g. something along the lines of:

```ruby
      before do
        perform_called = Queue.new
        allow(worker_spy).to receive(:perform) { perform_called << true }
        @thread = Thread.new { worker.perform }
        perform_called.pop
      end
```

This should make the `before` not actually finish until we the thread executed as expected.

My suggestion is not blocking btw :)

delner:
I appreciate the suggestion. All things considered, for now, I think I want to stick with the `Fiber` because it matches the cooperative need well, and should be more reliable than a thread.

I scanned the codebase for other uses of `Thread` like this, and didn't see any other great candidates offhand. I think this is a bit exceptional. We could consider this an experiment to see how well `Fiber` works, and we have the option of using your `Thread` suggestion if it doesn't. I'm optimistic that it should work fine.

PR #3036, opened 2023-08-08
Add development environment detection

marcotc:
This PR adds a utility to detect if `ddtrace` is running in a development environment (an environment that is unlikely to need to send traces, profiles, perform appsec checking).

This will enable us to selectively disable heavy features (background workers, network requests) performed by `ddtrace` when the environment does not need it. This is especially helpful when running tests with `ddtrace` installed.

### DIFF
--- a/spec/datadog/core/environment/execution_spec.rb
+++ b/spec/datadog/core/environment/execution_spec.rb
@@ -47,7 +47,9 @@ RSpec.describe Datadog::Core::Environment::Execution do
 
       context 'when in an IRB session' do
         it 'returns true' do
-          _, err, = Open3.capture3('irb', '--noprompt', '--noverbose', stdin_data: repl_script)
+          _, err = Bundler.with_clean_env do # Ruby 2.6 does not have irb by default in a bundle, but has it outside of it.
+            Open3.capture3('irb', '--noprompt', '--noverbose', stdin_data: repl_script)
+          end
           expect(err).to end_with('true')
         end
       end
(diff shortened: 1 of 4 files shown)


spec/datadog/core/environment/execution_spec.rb
```diff
@@ -0,0 +1,83 @@
...
+      # otherwise we'll have no real test for non-RSpec cases.
+      around do |example|
+        begin
+          original = $PROGRAM_NAME
+          $PROGRAM_NAME = 'not-rspec'
+          example.run
+        ensure
+          $PROGRAM_NAME = original
+        end
+      end
```
ivoanjo, line 30: This made me think that one possible alternative implementation strategy is to maybe have a proc-per-thing we want to detect, e.g.

```ruby
RSPEC_PROBE = proc { $PROGRAM_NAME.end_with?(RSPEC_PROGRAM_NAME) }
MINITEST_PROBE = proc { ... }

def test?(test_probes: [RSPEC_PROBE, MINITEST_PROBE])
  test_probes.any?(&:call)
end
```

then we could override individual probes (or even dependency inject them).

This to say -- it came to mind; I think the current version is great as well so as usual with my random suggestions, feel free to ignore :)

marcotc:
The `EnvironmentLogger` is undergoing a big refactoring by @sarahchen6 as we speak, I didn't want to refactor it as to avoid conflicts, but I agree that they will have to converge in the future.

spec/datadog/core/environment/execution_spec.rb
marcotc: I agree that this suggested approach looks good!

I think the file is small now to keep it as it, but it's definitely a great alternative.

ivoanjo (approved):
:+1: LGTM

TonyCTHsu (approved):

Issue #2823, opened 2023-04-28
New telemetry and remote config features breaking webmock-enabled tests


HoneyryderChuck:
**Current behaviour**

[httpx tests datadog and webmock integrations in the same scope](https://gitlab.com/os85/httpx/-/tree/master/integration_tests). It was never a problem until yesterday, when [the CI started complaining due to unrelated missing stubs for net-http](https://gitlab.com/os85/httpx/-/tree/master/integration_tests). 

**Expected behaviour**

The tests shouldn't have had problems, like they never had so far.

I suspect it didn't happen because:

* webmock tests are wrapped with "enable!/disable!"; so it shouldn't have impacted datadog tests, as wemock is supposed to be disabled. [I opened a ticket here](https://github.com/bblimke/webmock/issues/1025) for that.
* The telemetry feature being enabled by default means that the request is performed every time by default, even during tests, where I think it shouldn't. This, coupled with the  webmock bug above, exposed both situations.

I've since set `DD_INSTRUMENTATION_TELEMETRY_ENABLED=false`.

**Steps to reproduce**

```
# within the httpx repo
> bundle install
> bundle exec rake integration_tests
```

**Environment**

* **ddtrace version:**: 1.11.0
* **Configuration block (`Datadog.configure { |c| c.tracing.instrument(:httpx, options, &blk) }`):**
* **Ruby version:**: 3.2.1
* **Operating system:** Mac latest

delner:
@HoneyryderChuck Thanks for the report here.

Is tracing enabled in CI for the purposes of just testing the `httpx` instrumentation? Or are you using the CI tracing feature here?

If for the prior, you may want to use a "test friendly" configuration instead. The goal would be to pipe traces and spans into a buffer that can be inspected/reset by the test, as a a replacement of the background worker that writes traces to HTTP.

We have a "test mode" feature: see https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#additional-configuration for a description. Though it may not fully satisfy what you're looking for here... (we built it mostly for the CI visibility case mentioned above.)

I think this could be a good feature request: "add a test suite friendly mechanism to pipe spans to buffer instead of HTTP." If that sounds like that fits the bill, I'd be happy to consider its implementation.

HoneyryderChuck:
> If for the prior, you may want to use a "test friendly" configuration instead.

[...]

PR #3039, opened 2023-08-09
Disable Telemetry and Remote Configuration in development environments

marcotc:
**What does this PR do?**

By default, it disables the Telemetry and Remote Configuration background workers, which can send HTTP requests, in any environment that is detected as a development environment. See https://github.com/DataDog/dd-trace-rb/pull/3036 for the information about the detection logic.

**Motivation**

Fixes https://github.com/DataDog/dd-trace-rb/issues/2823

`ddtrace` can cause the test suite for the application to fail when `ddtrace` sends unexpected HTTP requests in the background. These requests are rarely desired and only cause noise add overhead.

**Additional Notes**

It's still possible to enable these features in a development environment by using their respective environment variables.

**How to test the change?**

Without the agent running, run any project's test suite with `ddtrace` enabled and noticed that there's no warning about Telemetry or Remote Configuration not reaching the agent (because their background workers are not running at all).

The same is true for an `irb` or `pry` environment.

### DIFF
--- a/lib/datadog/core/configuration/settings.rb
+++ b/lib/datadog/core/configuration/settings.rb
@@ -2,6 +2,7 @@ require 'logger'
 
 require_relative 'base'
 require_relative 'ext'
+require_relative '../environment/execution'
 require_relative '../environment/ext'
 require_relative '../runtime/ext'
 require_relative '../telemetry/ext'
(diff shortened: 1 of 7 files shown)


marcotc:
@delner, do you think we should also set `DD_TRACE_ENABLED` to `false` by default in development environments?

ivoanjo (commented):
A few more specs need to be adjusted, but I really like the approach!

TonyCTHsu (approved):

ivoanjo (approved):
:+1: LGTM

PR #3041, opened 2023-08-10
Improve quantization

lloeki:
**What does this PR do?**

Small bits of improvement to quantization and obfuscation.

**Motivation**

Stumbled upon a few areas that were lacking.

**Additional Notes**

Regex has been checked for linearity with the new Ruby 3.3 `Regexp.linear_time?`.

**How to test the change?**

```
env SKIP_SIMPLECOV=1 bundle exec rspec spec/datadog/tracing/contrib/utils/quantization/http_spec.rb
```

### DIFF
--- a/lib/datadog/tracing/contrib/utils/quantization/http.rb
+++ b/lib/datadog/tracing/contrib/utils/quantization/http.rb
@@ -134,18 +134,18 @@ module Datadog
                  (?:"|%22)?
               )
               (?: # common keys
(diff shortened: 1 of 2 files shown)


ivoanjo:
> Regex has been checked for linearity with the new Ruby 3.3 Regexp.linear_time?.

Maybe consider adding a test for this? We already have 3.3 in CI, and we could skip the test on all other Rubies -- we'd still benefit from no accidental changes :)

TonyCTHsu (approved):

PR #3044, opened 2023-08-10
enable APM Integrations Service Naming and Peer Service Tests

wconti27:
This PR enables APM Service Naming and Peer Service Tests from APM Test Agent.


spec/datadog/tracing/contrib/span_attribute_schema_examples.rb
```diff
@@ -13,13 +13,17 @@
     context 'test the default' do
       it do
         expect(span.service).to eq('rspec')
+        span.set_tag('_expected_service_name', 'rspec')
+        span.set_tag('_remove_integration_service_names_enabled', 'true')
```
marcotc, line 17: Could you explain what's happening here?

spec/support/network_helpers.rb
```diff
@@ -52,20 +52,22 @@ def check_availability_by_http_request(host, port)
...
-      trace_variables = "DD_SERVICE=#{dd_service},DD_SPAN_ATTRIBUTE_SCHEMA=#{dd_span_attribute_schema}"
-    else
-      trace_variables += ",DD_SERVICE=#{dd_service},DD_SPAN_ATTRIBUTE_SCHEMA=#{dd_span_attribute_schema}"
+  # @return [String] Key/Value pairs representing relevant Tracer Configuration
+  def parse_tracer_config(trace)
+    dd_env_variables = ENV.to_h.select { |key, _| key.start_with?('DD_') }
+    trace.spans.each do |s|
+      if s && s.meta && s.meta['_expected_service_name'] && s.meta['_remove_integration_service_names_enabled'] == 'true'
+        dd_env_variables['DD_SERVICE'] = s.meta['_expected_service_name']
+        dd_env_variables['DD_TRACE_SPAN_ATTRIBUTE_SCHEMA'] = 'v1'
```
marcotc, line 63: Because this is running in a `trace.spans` loop, `dd_env_variables['DD_SERVICE']` will be overwritten multiple times, with the last span winning. Is that expected?

spec/datadog/tracing/contrib/span_attribute_schema_examples.rb
wconti27: I updated this file to include a comment. Here we are grabbing the expected service name and putting it in the span meta in order to later send it to the test agent for service name testing to ensure consistency for service name when the `_remove_integration_service_names_enabled ` is set to true.

spec/support/network_helpers.rb
wconti27: Updated this to not overwrite service name if a valid service name has already been found.

spec/datadog/tracing/contrib/span_attribute_schema_examples.rb
marcotc: Ok, I understand so far.

When should I use these two tags myself?

marcotc (approved):
Thank you for refactoring and simplifying the PR!

PR #3045, opened 2023-08-10
Tracing: Exclude query string for `http.url` tag from `net/http` instrumentation 

TonyCTHsu:
**What does this PR do?**

Exclude query string for `http.url` tag from `net/http` instrumentation

### DIFF
--- a/lib/datadog/tracing/contrib/http/instrumentation.rb
+++ b/lib/datadog/tracing/contrib/http/instrumentation.rb
@@ -3,6 +3,7 @@ require 'uri'
 require_relative '../../metadata/ext'
 require_relative '../analytics'
 require_relative '../http_annotation_helper'
+require_relative '../utils/quantization/http'
 
 module Datadog
   module Tracing
@@ -79,8 +80,10 @@ module Datadog
 
               span.set_tag(Tracing::Metadata::Ext::TAG_COMPONENT, Ext::TAG_COMPONENT)
               span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_REQUEST)
-
-              span.set_tag(Tracing::Metadata::Ext::HTTP::TAG_URL, request.path)
+              span.set_tag(
+                Tracing::Metadata::Ext::HTTP::TAG_URL,
+                Contrib::Utils::Quantization::HTTP.url(request.path, { query: { exclude: :all } })
+              )
               span.set_tag(Tracing::Metadata::Ext::HTTP::TAG_METHOD, request.method)
 
               host, port = host_and_port(request)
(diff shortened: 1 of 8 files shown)


spec/datadog/tracing/contrib/ethon/easy_patch_spec.rb
```diff
@@ -248,4 +248,15 @@
       expect(span.get_tag('out.host')).to eq('example.com')
     end
   end
+
+  context 'when query string in url' do
+    it 'does not collect string' do
```
marcotc, line 253:
```suggestion
    it 'does not collect the query string' do
```

spec/datadog/tracing/contrib/ethon/easy_patch_spec.rb
```diff
@@ -248,4 +248,15 @@
...
+      easy = EthonSupport.ethon_easy_new(url: 'http://example.com/sample/path?foo=bar')
+
+      easy.perform
+
+      expect(span.get_tag('http.url')).to eq('/sample/path')
+      expect(span.get_tag('out.host')).to eq('example.com')
```
marcotc, line 259: I don't think we need to test for hostname here, given this test is asserting on query strings.

marcotc (approved):
Great work, @TonyCTHsu! 🙇

lib/datadog/tracing/contrib/http/instrumentation.rb
```diff
@@ -79,8 +80,10 @@ def annotate_span_with_request!(span, request, request_options)
 
               span.set_tag(Tracing::Metadata::Ext::TAG_COMPONENT, Ext::TAG_COMPONENT)
               span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_REQUEST)
-
-              span.set_tag(Tracing::Metadata::Ext::HTTP::TAG_URL, request.path)
+              span.set_tag(
+                Tracing::Metadata::Ext::HTTP::TAG_URL,
+                Contrib::Utils::Quantization::HTTP.url(request.path, { query: { exclude: :all } })
```
delner, line 85: This seems fine as a default, and you've added tests to make sure there aren't any query strings in other integrations.

Is it worth putting this same quantization line in those integrations as well? For consistency? Or is there a drawback?

delner (approved):
Perhaps you should address @marcotc 's comment first, and I have one question here, but otherwise this seems acceptable.

lib/datadog/tracing/contrib/http/instrumentation.rb
TonyCTHsu: Currently, other http client integration do not offer settings for quantization, each of them leverage their public interface to provide data. 

I believe we should refactor and centraliz the qunatization mechanism when we offer the feature.

PR #3049, opened 2023-08-14
Ekump/move agent settings resolver to components

ekump:
**What does this PR do?**
This PR removes the dependency of the tracing component in the Core `AgentSettingsResolver` and introduces component specific resolvers for Tracing and Profiling. The Core AgentSettingsResolver no longer takes into account the tracing transport proc. 

The `AgentSettingsResolvers` in `Tracing` and `Profiling` are identical and the introduction of tech debt is intentional. Over time, components should be refactored to not be dependent on settings from other components. Also, it is a goal to remove the transport options proc in a future major release. If profiling and/or tracing reach a point where they no longer require special agent configuration, they can revert to using the core `AgentSettingsResolver`.

Methods that are common to all resolvers have been moved to a mixin called `AgentSettingsResolverCommon` to reduce code duplication. 

**Motivation**
This PR is part of a larger effort to remove component dependencies from the `Core` namespace. There currently exists a circular dependency in the library where `Core` depends on components, and components depend on `Core`.

**Additional Notes**
There should be no functional change from this PR. 

The lines changed stats are a bit misleading as the majority of the code is copied from the pre-existing AgentSettingsResolver to the component specific ones, along with their respective tests.

**How to test the change?**
No functionality should have changed, so existing tests should continue to pass.


lib/datadog/core/configuration/agent_settings_resolver.rb
```diff
@@ -276,21 +171,9 @@ def warn_if_configuration_mismatch(detected_configurations_in_priority_order)
...
-        def log_warning(message)
-          logger.warn(message) if logger
-        end
-
-        def http_scheme?(uri)
-          ['http', 'https'].include?(uri.scheme)
-        end
-
         # Expected to return nil (not false!) when it's not http
         def parsed_http_url
```
ekump, line 175: TODO: remove method

lib/datadog/core/transport/http/builder.rb
```diff
@@ -70,7 +70,7 @@ def initialize
 
           def adapter(config, *args, **kwargs)
             @default_adapter = case config
-                               when Core::Configuration::AgentSettingsResolver::AgentSettings
+                               when Core::Configuration::AgentSettingsResolverCommon::BaseAgentSettings
```
ekump, line 73: I don't love that we directly reference `BaseAgentSettings` outside of the resolvers, but since we don't know which resolver generated AgentSettings here, I'm not sure of a better way to maintain existing functionality.

lib/datadog/profiling/agent_settings_resolver.rb
```diff
@@ -0,0 +1,335 @@
...
+require 'uri'
+
+require_relative '../core/configuration/settings'
+require_relative '../core/configuration/agent_settings_resolver_common'
+require_relative '../tracing/configuration/ext'
+require_relative '../../ddtrace/transport/ext'
+
+module Datadog
+  module Profiling
+    # This class unifies all the different ways that users can configure how we talk to the agent.
```
ekump, line 12: TODO: Add comment explaining why this is here and the same as tracing's resolver.

lib/datadog/tracing/sync_writer.rb
```diff
@@ -23,7 +23,7 @@ class SyncWriter
       # @param [Datadog::Transport::Traces::Transport] transport a custom transport instance.
       #   If provided, overrides `transport_options` and `agent_settings`.
       # @param [Hash<Symbol,Object>] transport_options options for the default transport instance.
-      # @param [Datadog::Core::Configuration::AgentSettingsResolver::AgentSettings] agent_settings agent options for
+      # @param [Datadog::Core::Configuration::AgentSettingsResolverCommon::AgentSettings] agent_settings agent options for
```
ekump, line 26: TODO: Fix this comment. Should reference `Tracing::Configuration::AgentSettingsResolver::AgentSettings`

lib/datadog/profiling/agent_settings_resolver.rb
```diff
@@ -0,0 +1,340 @@
...
+    # about it and pick a value based on the following priority: code > environment variable > defaults.
+    #
+    # TODO: Decouple Tracing. Profiling relies on the deprecated_for_removal_transport_configuration_proc, which in turn
+    # relies on code and constants in the Tracing namespace. This class is an exact copy of
+    # Datadog::Tracing::Configuration::AgentSettingsResolver. When the proc is removed, this resolver may be replaced
+    # with the implementation in Core.
+    class AgentSettingsResolver
+      include Core::Configuration::AgentSettingsResolverCommon
+
+      AgentSettings = Class.new(BaseAgentSettings) do
```
ekump, line 29: Tracing and Profiling override the BaseAgentSettings to include the `deprecated_for_removal_transport_configuration_proc`. This seemed like the least bad way to extend AgentSettings without forcing the version in Core to also have an attribute for the proc.

lib/datadog/core/configuration/ext.rb
```diff
@@ -19,6 +19,11 @@ module Metrics
 
         module Transport
           ENV_DEFAULT_HOST = 'DD_AGENT_HOST'
+          # DD_TRACE_AGENT_PORT is the only env var available for setting the agent port.
+          # It is the same as Datadog::Tracing::Configuration::Ext::Transport::ENV_DEFAULT_PORT
+          # but is also here as we do not want Core to have a dependency on Tracing.
+          ENV_DEFAULT_PORT = 'DD_TRACE_AGENT_PORT'
```
marcotc, line 25: I think the `_TRACE_` in `DD_TRACE_AGENT_PORT` is a (legacy?) misnomers, given this port is the communication port for all of APM products with the agent. It's documented as `APM Receiver`: https://github.com/DataDog/documentation/blob/318dffe3a8c86aa81fc3b6619c9f19ab4e48af41/content/en/agent/guide/network.md?plain=1#L354-L355

I'd read this environment variable as `DD_APM_AGENT_PORT`.

I don't want to throw a wrench at this specific change, just want to bring up the fact that this this port is not tracing-specific.

ekump: So this change is just copying the consts for the existing env vars into the core namespace so we can reference it in core without including any tracing code and maintain backwards compatibility. I agree the name is probably not accurate since the port isn't tracing specific anymore. From conversations i've had, changing the actual env var to remove the `_tracing_` part would require effort beyond the scope of this project.

ivoanjo:
I'm still a bit unsure about the path we're following in this PR.

The `AgentSettingsResolver` is already an extremely complex beast ([where it's easy to miss things](https://github.com/DataDog/dd-trace-rb/pull/3037) and there's [still missing behavior](https://github.com/DataDog/dd-trace-rb/pull/3037#issuecomment-1672283380)). (And I'm partially at fault about that, having built it)

But at least it's a self-contained beast -- all of the weird configuration semantics we've concocted are in a single place, which reads a few things and outputs a single decision that every component can use (an instance of `AgentSettingsResolver::AgentSettings`).

This PR breaks the resolver into at least 4 files (without counting specs), with a lot of copy-pasted stuff. It seems to me that will make it a lot harder to maintain.

This PR I suspect also breaks UDS for remote configuration (and maybe telemetry), and any other products that want to make use of it. I'm really concerned we're degrading the user experience here -- right now, regardless of how customers configure it, if they have working traces, or profiles, or ..., they have working everything (or at least they should be able to, if we set up things right).

With this change, different components of dd-trace-rb will behave differently, which seems to me a bit weird.

One thing I didn't quite get is, is the added complexity and refactor to break the dependency to the `c.tracing.transport_options` from the `AgentSettingsResolver` in the core? Or is there any other things we're trying to separate?

Could we perhaps deprecate or change a bit the settings we offer to customers, in a way that would allow us to keep a single `AgentSettingsResolver` in the core for every component, but still offer additional flexibility for tracing?

delner:
> This PR I suspect also breaks UDS for remote configuration (and maybe telemetry), and any other products that want to make use of it.

I think this is a fair concern; what can we do to test this is not the case, @ivoanjo? I'd like an actionable way to gain confidence and answer definitively that it is not breaking either of these things.

delner:
> The AgentSettingsResolver is already an extremely complex beast (https://github.com/DataDog/dd-trace-rb/pull/3037 and there's https://github.com/DataDog/dd-trace-rb/pull/3037#issuecomment-1672283380). (And I'm partially at fault about that, having built it)
>
> But at least it's a self-contained beast -- all of the weird configuration semantics we've concocted are in a single place, which reads a few things and outputs a single decision that every component can use (an instance of AgentSettingsResolver::AgentSettings).

I think it's fair to say compartmentalizing risk is a good thing. And I think this is a good strategy to contain a problem before it becomes more complicated. This was the right move at the time.

Still, the purpose of this PR is to now take it a step further, and "tame" the beast that is the `AgentSettingsResolver`, which is virtually unmaintainable right now. Containment is not a viable strategy to reduce the complexity. The general approach here (being break it into smaller, simpler parts) seems like a sensible way to make it more manageable/simpler, and complements the goal of reducing `Core` knowledge of products. If there's a better strategy we can employ that accomplishes the same goals to the same or greater affect, I'm open to considering that.

ivoanjo:
>> This PR I suspect also breaks UDS for remote configuration (and maybe telemetry), and any other products that want to make use of it.
>
> I think this is a fair concern; what can we do to test this is not the case, @ivoanjo? I'd like an actionable way to gain confidence and answer definitively that it is not breaking either of these things.

Customers setting up UDS via code are following https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#using-the-unix-domain-socket-uds-adapter .

(We should also support customers that configure their http settings via [a proc as well](https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#using-the-nethttp-adapter)). 

One possible way of validating this is having integration tests that check that every product (aka every thing that uses `AgentSettingResolver`) is able to correctly communicate with the Datadog agent in both of those cases.

> Still, the purpose of this PR is to now take it a step further, and "tame" the beast that is the AgentSettingsResolver, which is virtually unmaintainable right now.

To be clear, regardless of my concerns on the maintainability of the code itself (With a few tweaks I can live with that), my biggest concern is that we mustn't regress on having every component of dd-trace-rb be able to communicate with the agent, regardless of which of the many, many, many ways customers chose to provide configuration.

delner:
> Customers setting up UDS via code are following https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#using-the-unix-domain-socket-uds-adapter .

> (We should also support customers that configure their http settings via [a proc as well](https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#using-the-nethttp-adapter)).

> One possible way of validating this is having integration tests that check that every product (aka every thing that uses AgentSettingResolver) is able to correctly communicate with the Datadog agent in both of those cases.

Okay, let's spec out what it takes to test these things, see if @ekump's changes cause any kind of regression.

ivoanjo:
I was thinking about this PR earlier today, and I think I can add a bit more context here, and a suggestion or two.

I think part of the mess here is that the `c.tracing.transport_options`, for historical reasons, started its life inside tracing (as tracing was at some point the only thing this gem did), but [is documented and used by the whole gem](https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/ruby/#configuring-the-transport-layer).

I realized that when working on the `AgentSettingsResolver` and that's why it was stored as `deprecated_for_removal_transport_configuration_proc` -- as a reminder that there's definitly something wrong that we should move away from.

Perhaps this is the time finally to get rid of it? Here's a suggestion I was thinking about. (For reference, I refer to the "c.tracing.transport_options" we use right now as "old transport_options".)

1. Create a replacement for the old transport_options inside core that can be used by all products. It doesn't have to look exactly like `transport_options`, but should probably allow for the same range of configurability for uds/https.

2. Create a replacement for the old transport_options only for tracing, that allows the extra features that we want to provide only for that product.

3. Update [documentation](https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/ruby/#configuring-the-transport-layer) to point customers on when to using the options 1 and 2. Remove all references to the old transport options from that section.

4. Deprecate the old transport options for removal in 2.x. Furthermore, when users use it, add a nice message telling them this is going to be removed and perhaps even showing how they need to update their configuration to adopt the new, recommended ways. We can also add a reference to using this to telemetry, so we could have metrics around usage of the deprecated setting.

Doing the above would allow us to move customers to the new solution, while making sure existing configurations would work for the remainder or the 1.x series. Customers following our instructions now would smoothly go over to 2.x; customers ignoring it would need to change their config when they upgrade.

GustavoCaso:
I really like @ivoanjo suggestion for a more gradual migration, especially considering the 2.x release and being able to deprecate old code. 

Having such a migration plan would allow us to achieve what we want in the future, which to have a generic transport options and custom options for products that need it.

ekump:
Replaced by #3067

PR #3055, opened 2023-08-17
Report Linux native thread ids in profiles for Ruby < 3.1

ivoanjo:
**What does this PR do?**:

This PR adds support for reading the actual native thread id (tid) for Ruby threads in Ruby < 3.1.

Starting in Ruby 3.1, Ruby stores the tid for every thread it creates, and even exposes a `Thread#native_thread_id` with this information.

The tid is relevant because it's what shows up in OS-level tools, such as task managers, or even the Datadog native profiler, thus enabling the correlation of what a thread was doing from both inside and outside the VM.

As mentioned above, up until now, the profiler was only able to report this tid for Ruby 3.1+. For older Rubies, we used a fallback id, which was still as unique as the tid, but was not useful when correlating what a thread was doing with other tools.

This profiler introduces a new `LinuxTidFallback` class, which, together with some dark magic implemented in `linux_tid_from_pthread.c` (see comments on that file for details), is able to retrieve the tid for threads, even though the Linux libc developers really don't want us to do that.

**Motivation**:

Having a tid, as I mentioned above, allows us to correlate what happens inside the Ruby VM with what external tools can observe as well.

The tid is especially visible in the upcoming profiler timeline feature, which is why I decided to finally take a stab at improving this as a R&D week project.

**Additional Notes**:

This new feature is on by default, but can be disabled via configuration.

My intention with making it configurable is as a "just-in-case" lever for support; I do not recommend disabling it.

Although the profiler only officially supports Linux, we like to keep it running on macOS and others for ease of development and experimentation. This feature appropriately degrades on those OSs and disables itself.

There's a few situations, Linux-permissions-wise, where this approach may not work. In those cases (running in CircleCI is one of them), we also degrade gracefully back to the previous behavior.

**How to test the change?**:

This change includes test coverage. You can also see it running by running the profiler on Linux + Ruby < 3.1 and observing that the tids reported are actual linux tids that are seen in any task manager tool such as `htop` (do remember to turn on the feature to see individual threads, since most tools don't show them by default).

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
@@ -8,6 +8,7 @@
 #include "private_vm_api_access.h"
 #include "stack_recorder.h"
 #include "time_helpers.h"
+#include "linux_tid_fallback.h"
 
 // Used to trigger sampling of threads, based on external "events", such as:
 // * periodic timer for cpu-time and wall-time
@@ -105,6 +106,8 @@ struct thread_context_collector_state {
   monotonic_to_system_epoch_state time_converter_state;
   // Used to identify the main thread, to give it a fallback name
   VALUE main_thread;
+  // Used to obtain the thread id on Linux for legacy Rubies
+  VALUE linux_tid_fallback;
 
   struct stats {
     // Track how many garbage collection samples we've taken.
@@ -157,7 +160,8 @@ static VALUE _native_initialize(
   VALUE max_frames,
   VALUE tracer_context_key,
   VALUE endpoint_collection_enabled,
-  VALUE timeline_enabled
+  VALUE timeline_enabled,
+  VALUE linux_tid_fallback
 );
 static VALUE _native_sample(VALUE self, VALUE collector_instance, VALUE profiler_overhead_stack_thread);
 static VALUE _native_on_gc_start(VALUE self, VALUE collector_instance);
@@ -218,7 +222,7 @@ void collectors_thread_context_init(VALUE profiling_module) {
   // https://bugs.ruby-lang.org/issues/18007 for a discussion around this.
   rb_define_alloc_func(collectors_thread_context_class, _native_new);
 
-  rb_define_singleton_method(collectors_thread_context_class, "_native_initialize", _native_initialize, 6);
+  rb_define_singleton_method(collectors_thread_context_class, "_native_initialize", _native_initialize, 7);
   rb_define_singleton_method(collectors_thread_context_class, "_native_inspect", _native_inspect, 1);
   rb_define_singleton_method(collectors_thread_context_class, "_native_reset_after_fork", _native_reset_after_fork, 1);
   rb_define_singleton_method(testing_module, "_native_sample", _native_sample, 2);
@@ -262,6 +266,7 @@ static void thread_context_collector_typed_data_mark(void *state_ptr) {
   st_foreach(state->hash_map_per_thread_context, hash_map_per_thread_context_mark, 0 /* unused */);
   rb_gc_mark(state->thread_list_buffer);
   rb_gc_mark(state->main_thread);
+  rb_gc_mark(state->linux_tid_fallback);
 }
 
 static void thread_context_collector_typed_data_free(void *state_ptr) {
@@ -310,6 +315,7 @@ static VALUE _native_new(VALUE klass) {
   state->timeline_enabled = true;
   state->time_converter_state = (monotonic_to_system_epoch_state) MONOTONIC_TO_SYSTEM_EPOCH_INITIALIZER;
   state->main_thread = rb_thread_main();
+  state->linux_tid_fallback = Qnil;
 
   return TypedData_Wrap_Struct(klass, &thread_context_collector_typed_data, state);
 }
@@ -321,7 +327,8 @@ static VALUE _native_initialize(
   VALUE max_frames,
   VALUE tracer_context_key,
   VALUE endpoint_collection_enabled,
-  VALUE timeline_enabled
+  VALUE timeline_enabled,
+  VALUE linux_tid_fallback
 ) {
   ENFORCE_BOOLEAN(endpoint_collection_enabled);
   ENFORCE_BOOLEAN(timeline_enabled);
@@ -338,6 +345,7 @@ static VALUE _native_initialize(
   state->recorder_instance = enforce_recorder_instance(recorder_instance);
   state->endpoint_collection_enabled = (endpoint_collection_enabled == Qtrue);
   state->timeline_enabled = (timeline_enabled == Qtrue);
+  state->linux_tid_fallback = linux_tid_fallback;
 
   if (RTEST(tracer_context_key)) {
     ENFORCE_TYPE(tracer_context_key, T_SYMBOL);
@@ -766,9 +774,16 @@ static struct per_thread_context *get_context_for(VALUE thread, struct thread_co
 }
 
 static void initialize_context(VALUE thread, struct per_thread_context *thread_context, struct thread_context_collector_state *state) {
+  uint64_t native_thread_id = native_thread_id_for(thread);
+
+  if (state->linux_tid_fallback != Qnil) { // Only available when needed
+    pid_t maybe_tid = linux_tid_fallback_for(state->linux_tid_fallback, pthread_id_for(thread));
+    if (maybe_tid > 0) native_thread_id = maybe_tid;
+  }
+
   // Why does the thread id actually contain two ids?
   //
-  // The `native_thread_id_for` is the OS-level thread id, and can be used to match up a thread reported by the profiler
+  // The `native_thread_id` is the OS-level thread id, and can be used to match up a thread reported by the profiler
   // to system-level tools, such as the task manager or even the Datadog native profiler.
   // (It's usually the same as `Thread#native_thread_id` on modern Ruby versions.)
   //
@@ -777,7 +792,7 @@ static void initialize_context(VALUE thread, struct per_thread_context *thread_c
   // `Thread#object_id` for the thread object to provide disambiguation. The `object_id` can also come in handy when
   // matching up profiler results to the Ruby threads in our own tests, since not all Ruby versions we test on
   // expose `Thread#native_thread_id`.
-  snprintf(thread_context->thread_id, THREAD_ID_LIMIT_CHARS, "%"PRIu64" (%lu)", native_thread_id_for(thread), (unsigned long) thread_id_for(thread));
+  snprintf(thread_context->thread_id, THREAD_ID_LIMIT_CHARS, "%"PRIu64" (%lu)", native_thread_id, (unsigned long) thread_id_for(thread));
   thread_context->thread_id_char_slice = (ddog_CharSlice) {.ptr = thread_context->thread_id, .len = strlen(thread_context->thread_id)};
 
   int invoke_line_location;
@@ -836,6 +851,7 @@ static VALUE _native_inspect(DDTRACE_UNUSED VALUE _self, VALUE collector_instanc
     state->time_converter_state.delta_to_epoch_ns
   ));
   rb_str_concat(result, rb_sprintf(" main_thread=%"PRIsVALUE, state->main_thread));
+  rb_str_concat(result, rb_sprintf(" linux_tid_fallback=%"PRIsVALUE, state->linux_tid_fallback));
 
   return result;
 }
--- a/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
@@ -766,6 +766,17 @@ static struct per_thread_context *get_context_for(VALUE thread, struct thread_co
 }
 
 static void initialize_context(VALUE thread, struct per_thread_context *thread_context, struct thread_context_collector_state *state) {
+  // Why does the thread id actually contain two ids?
+  //
+  // The `native_thread_id_for` is the OS-level thread id, and can be used to match up a thread reported by the profiler
+  // to system-level tools, such as the task manager or even the Datadog native profiler.
+  // (It's usually the same as `Thread#native_thread_id` on modern Ruby versions.)
+  //
+  // BUT! Ruby actually reuses native threads (https://ivoanjo.me/blog/2022/11/26/ruby-reuses-native-threads/) and thus
+  // the native thread id is not enough (and could cause confusion). Thus, we also include the result of
+  // `Thread#object_id` for the thread object to provide disambiguation. The `object_id` can also come in handy when
+  // matching up profiler results to the Ruby threads in our own tests, since not all Ruby versions we test on
+  // expose `Thread#native_thread_id`.
   snprintf(thread_context->thread_id, THREAD_ID_LIMIT_CHARS, "%"PRIu64" (%lu)", native_thread_id_for(thread), (unsigned long) thread_id_for(thread));
   thread_context->thread_id_char_slice = (ddog_CharSlice) {.ptr = thread_context->thread_id, .len = strlen(thread_context->thread_id)};
 
--- a/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
@@ -8,7 +8,7 @@
 #include "private_vm_api_access.h"
 #include "stack_recorder.h"
 #include "time_helpers.h"
-#include "linux_tid_fallback.h"
+#include "linux_tid_override.h"
 
 // Used to trigger sampling of threads, based on external "events", such as:
 // * periodic timer for cpu-time and wall-time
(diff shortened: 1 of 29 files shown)


spec/datadog/profiling/linux_tid_fallback_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+  # It also breaks the specs with a clear error message when the API is not available; as otherwise the failures
+  # would be a bit cryptic.
+  #
+  # Finally, it adds a DD_PROFILING_SKIP_LINUX_TID_FALLBACK_TESTING that folks can use for local testing, in case
+  # their setup doesn't provide this API either.
+  def check_for_process_vm_readv
+    return if Datadog::Profiling::LinuxTidFallback::Testing._native_can_use_process_vm_readv?
+
+    if ENV['CIRCLECI'] == 'true'
+      skip "Skipping LinuxTidFallback specs because process_vm_readv doesn't work on CircleCI"
```
AlexJF, line 94: Q: How hard would it be to support mocking `process_vm_readv`? You already seem to have native things dedicated to `::Testing`. If you accessed `process_vm_ready` through a function pointer, it sounds like you could have a `::Testing::mockProcessVMReadv` method that could replace that pointer during a CIRCLECI rspec run?  Might be better than skipping altogether?

ext/ddtrace_profiling_native_extension/collectors_thread_context.c
```diff
@@ -766,7 +774,25 @@ static struct per_thread_context *get_context_for(VALUE thread, struct thread_co
...
+  // The `native_thread_id` is the OS-level thread id, and can be used to match up a thread reported by the profiler
+  // to system-level tools, such as the task manager or even the Datadog native profiler.
+  // (It's usually the same as `Thread#native_thread_id` on modern Ruby versions.)
+  //
+  // BUT! Ruby actually reuses native threads (https://ivoanjo.me/blog/2022/11/26/ruby-reuses-native-threads/) and thus
+  // the native thread id is not enough (and could cause confusion). Thus, we also include the result of
+  // `Thread#object_id` for the thread object to provide disambiguation. The `object_id` can also come in handy when
+  // matching up profiler results to the Ruby threads in our own tests, since not all Ruby versions we test on
+  // expose `Thread#native_thread_id`.
+  snprintf(thread_context->thread_id, THREAD_ID_LIMIT_CHARS, "%"PRIu64" (%lu)", native_thread_id, (unsigned long) thread_id_for(thread));
```
marcotc, line 795: Is this string format (`native_thread_id (thread.object_id)`) free-form? I ask because I assume there is thread correlation happening the Datadog UI down the pipeline.

ext/ddtrace_profiling_native_extension/linux_tid_from_pthread.c
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  static short find_tid_offset_in_buffer(struct tid_probe_info *tid_probe_info, short start_position) {
+    if (!tid_probe_info->buffer_read_success) return -1;
+
+    short buffer_size = STRUCT_PTHREAD_READ_SIZE;
+    pid_t *struct_pthread_base = (pid_t *) tid_probe_info->buffer;
+    pid_t reference_tid = tid_probe_info->tid;
+
+    for (pid_t *tid_ptr = struct_pthread_base + start_position; tid_ptr < struct_pthread_base + (buffer_size / sizeof(pid_t)); tid_ptr++) {
```
marcotc, line 178: Is it safe to assume that the tread_id we are looking for is of size `sizeof(pid_t)`?

ext/ddtrace_profiling_native_extension/linux_tid_from_pthread.c
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  static bool read_struct_pthread(pthread_t thread, void *struct_pthread_buffer) {
+    return read_safely((void *) thread, struct_pthread_buffer, STRUCT_PTHREAD_READ_SIZE);
+  }
+
+  static short find_tid_offset_in_buffer(struct tid_probe_info *tid_probe_info, short start_position) {
+    if (!tid_probe_info->buffer_read_success) return -1;
+
+    short buffer_size = STRUCT_PTHREAD_READ_SIZE;
```
marcotc, line 174: Given we are reading 2048 bytes, is it possible that we read from a region of memory that belongs to a different data structure (beyond `struct pthread`)?

ext/ddtrace_profiling_native_extension/linux_tid_from_pthread.c
```diff
@@ -0,0 +1,209 @@
...
+    int unused_flags = 0;
+    int number_of_iovecs = 1;
+
+    short num_read = process_vm_readv(ddtrace_gettid(), &read_into, number_of_iovecs, &read_from, number_of_iovecs, unused_flags);
+
+    return num_read == buffer_size;
+  }
+
+  static bool read_struct_pthread(pthread_t thread, void *struct_pthread_buffer) {
+    return read_safely((void *) thread, struct_pthread_buffer, STRUCT_PTHREAD_READ_SIZE);
```
marcotc, line 168: Given we are asking to read 2048 bytes, is it possible that we are asking `process_vm_readv` to read the desired `struct pthread` plus memory that could possibly resided in a protected page?
My concern is: are we reading too much, and possibly triggering page errors because of that?

ext/ddtrace_profiling_native_extension/linux_tid_from_pthread.c
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  bool read_safely(void *read_from_ptr, void *read_into_buffer, short buffer_size) {
+    struct iovec read_into = {.iov_base = read_into_buffer, .iov_len = buffer_size};
+    struct iovec read_from = {.iov_base = read_from_ptr, .iov_len = buffer_size};
+
+    int unused_flags = 0;
+    int number_of_iovecs = 1;
+
+    short num_read = process_vm_readv(ddtrace_gettid(), &read_into, number_of_iovecs, &read_from, number_of_iovecs, unused_flags);
```
marcotc, line 162: Ok so hear me out: what if we replaced all reads with `process_vm_readv`, so that segfaults can never exist ever again?!

ext/ddtrace_profiling_native_extension/linux_tid_from_pthread.c
```diff
@@ -0,0 +1,209 @@
...
+//   (or -1 if not found/available)
+//
+// The intuition for figuring out the offset inside in `setup_linux_tid_from_pthread_offset` is that:
+//
+// 1. We can take a thread for which we DO know the tid, like the caller thread. Let's call it a reference thread.
+// 2. We cast the reference thread's `pthread_t` to a pointer.
+// 3. We follow that pointer and all goes well, we expect that at some offset from the memory `pthread_t` was pointing
+//    at, we'll find the reference thread's tid.
+// 4. Since the offset for every thread is the same (the layout of `struct pthread` doesn't change), we can now use this
+//    offset to get the tid any other thread we want, given its `pthread_t`.
```
marcotc, line 35: image

spec/datadog/profiling/linux_tid_fallback_spec.rb
ivoanjo: It would definitely be doable -- we already have the `struct linux_tid_fallback_state` that is stored inside instances of the `LinuxTidFallback` object, so we could store the right function pointer to use there. We could then introduce an initialize method for that class that had a flag to allow mocking, or something like that.

What makes me hesitate a bit is that's adding more complexity to the "production" code path, just for testing, whereas right now the extra stuff for testing is mostly off to the side.

Also, mocking would only catch regressions with not calling `LinuxTidFallback`, which... yeah I'm not sure how often it would happen.

So I'm a bit undecided on this one; do you think it's worth the effort for the extra bit of coverage?

ext/ddtrace_profiling_native_extension/collectors_thread_context.c
ivoanjo: It is free-form. You're right: this format will make it harder to correlate ids with other tools.

The current trade-off is that we don't need to special-case handling of ids for Ruby -- e.g. we could submit two ids as two different labels, but then the backend would need special code for Ruby to distinguish/show/render them.

My thinking is: once we have a better idea of what we would want to with ids, we'd migrate Ruby over (one thing we also don't signal to the backend is when the tid is a real one vs when it's a fallback one that the backend won't be able to correlate). This would mean the feature wouldn't work/work as well for older versions but... that's already the case today, since we've been using this format for a few releases.

ext/ddtrace_profiling_native_extension/linux_tid_from_pthread.c
ivoanjo: I believe so, for two kinds of reasons:

1. This comes from the linux kernel:

* It is a `pid_t` on both glibc ([link](https://github.com/bminor/glibc/blob/4290aed05135ae4c0272006442d147f2155e70d7/nptl/descr.h#L174))
* And an `int` on musl ([link](https://github.com/bminor/musl/blob/718f363bc2067b6487900eddc9180c84e7739f80/src/internal/pthread_impl.h#L35)) ([And then musl typedefs `int` as `pid_t`](https://github.com/bminor/musl/blob/718f363bc2067b6487900eddc9180c84e7739f80/include/alltypes.h.in#L46)
* The Linux kernel also uses `int` [as a typedef for it](https://github.com/torvalds/linux/blob/89bf6209cad66214d3774dac86b6bbf2aec6a30d/include/uapi/asm-generic/posix_types.h#L28).

2. We're limited by our read of exactly `STRUCT_PTHREAD_READ_SIZE` bytes. If somehow the `pid_t` info is not there, or the size/alignment changes so we don't find it, the code correctly handles this by reporting `-1` and thus we disable this approach.

ext/ddtrace_profiling_native_extension/linux_tid_from_pthread.c
ivoanjo: It is! This is the cool thing about the `process_vm_readv` approach -- it really is reading how many bytes we ask, regardless of what's there, and returning an error if it couldn't.

So yeah, we may not read the whole structure, or we may grab extra stuff that isn't relevant. Worst case we don't find what we need (-1) or in very rare situations we may pick the wrong offset, but that seems quite rare seeing as we would need all 3 reference threads to agree on that wrong offset, and even then since this is only used for display the consequence is just wrong data in the UX.

ext/ddtrace_profiling_native_extension/linux_tid_from_pthread.c
ivoanjo: Yes -- it's possible. But the error is limited to a return code of `process_vm_readv`, so worst case, the workaround gets disabled, and the previous fallback behavior is used.

I claim this is ok for such a best-effort feature -- the idea is that we can improve the previous behavior for a subset of customers (hopefully lots?), but if we can't for some reason (API not available, weird memory that we can't read), we're not worse than we were.

ext/ddtrace_profiling_native_extension/linux_tid_from_pthread.c
ivoanjo: This is actually not that wild of an idea. I can't remember any reported segfaults from the profiler sampling code, but definitely using `process_vm_readv` would increase safety when reading VM data structures (in fact, we experimented with `process_vm_readv` for implementing at least one other profiler, but ended up taking a different approach).

I'd say the main disadvantage is that `process_vm_readv` can require a few more security permissions (see: it's disabled on CircleCI) thus it would complicate/restrict onboarding, which is a shame.

ext/ddtrace_profiling_native_extension/linux_tid_from_pthread.c
ivoanjo: Haha I had fun during R&D week!

ext/ddtrace_profiling_native_extension/linux_tid_from_pthread.c
```diff
@@ -0,0 +1,221 @@
...
+  #include <stdio.h>
+  #include <string.h>
+  #include <sys/syscall.h>
+  #include <sys/uio.h>
+  #include <unistd.h>
+  #include "linux_tid_from_pthread.h"
+
+  // This value seems enough for both glibc and musl (and it's safe to read with `process_vm_readv`)
+  // Must be a multiple of sizeof(pid_t)
+  #define STRUCT_PTHREAD_READ_SIZE 2048
```
marcotc, line 87: Can we document why `2048`?
From the current comment, I'm not sure how large glibc's and musl's versions are at the time of this writing (for an example architecture).
Also, links to glibc and musl data structures would be helpful here.

My main thing here is: I'm not sure what 2048 is supposed to be. Is it close to glibc's struct size? is it double, just to be safe? Is it like 10x more, to be even more safe? is it half the size, to avoid reading garbage?

marcotc (approved):

ext/ddtrace_profiling_native_extension/linux_tid_from_pthread.c
ivoanjo: Extremely fair! In hindsight I agree I was handwaving too much here. I've added this as a comment in b31620734c13bbdf59f7b9710f9b23537bd0c2f8 :

````
In particular, I've picked 2048 after examining the layouts of the structures using gdb. This value is a bit more
than double the needed for glibc, and is way higher than needed for alpine, so we have some leeway for the field
moving around in different libc versions and architectures.

* On Ubuntu 22.04's x86-64 glibc 2.35:
 ```
 gdb /lib/x86_64-linux-gnu/libc.so.6
 (gdb) print sizeof(struct pthread)
 $1 = 2496
 (gdb) ptype /o struct pthread
 /* offset      |    size */  type = struct pthread {
 /*    720      |       4 */    pid_t tid;
 ```
* On alpine 3.18.3's x86-64 musl 1.2.4:
  ```
 gdb /lib/libc.musl-x86_64.so.1
 (gdb) print sizeof(struct __pthread)
 $1 = 200
 (gdb) ptype /o struct __pthread
 /* offset      |    size */  type = struct __pthread {
 /*     48      |       4 */    int tid;
 ```

In the future, we may want to tune this down if we see that process_vm_readv fails because it can't read the whole
2048 bytes (e.g. on musl?), but it's a trade-off between overfitting and overreading :)
````

It's cool that gdb is able to print the structure and its layout :)

spec/datadog/profiling/linux_tid_fallback_spec.rb
AlexJF: No strong opinion but I think mocking would also get you some CI coverage of the main behaviour on `:thread_id/uses the real native thread id` from `thread_contex_spec.rb` (unless that one works in CI and I'm just dumb)?

Without the mock, it seems like CI would be blind to deliberate or accidental changes to all of this functionality? Ofc the best approach would be to ensure some part of CI could actually meet the conditions to exercise this and get a full functional test  😅 

> that's adding more complexity to the "production" code path, just for testing, whereas right now the extra stuff for testing is mostly off to the side.

I understand the worry but OTOH, it sounds like the extra complexity would just be an extra pointer whose override would just be done under a testing run (similar to how you'd test a critical system using dependency injection), there'd be no extra testing-only loops or branches in production hotpath.

ext/ddtrace_profiling_native_extension/linux_tid_from_pthread.c
nsavoire: When calling process_vm_readv, you could split your single remote iovec into multiple iovecs that don't cross page boundaries. That way even if the second remote iovec fails because of a page error, you would still get data from the first remote iovec (and then you are guaranteed that pthread_t object is fully present in that first iovec).

ivoanjo: Interesting... Is there an easy way to determine where in a the page my pointer is?

Or would you suggest just breaking on 4k boundaries since that's the minimum for most architectures?

nsavoire: You can call `sysconf(_SC_PAGESIZE)` to get the page size and then do some pointer arithmetic to figure out the boundary (since pages are aligned on page size).
To get the next page boundary from `thread` you could do:
`(((uintptr_t)thread) & ~(page_size - 1)) + page_size`

ivoanjo: Nice! Let me take a stab at doing that

spec/datadog/profiling/linux_tid_fallback_spec.rb
ivoanjo: I'm convinced, let me take a stab at it!

ivoanjo:
It's been almost a year since I've opened this. This was an R&D week thing, and I ended up not getting back to it, and then kinda fell in the background.

In the meanwhile, two things happened:
* Ruby added M:N threads (https://bugs.ruby-lang.org/issues/19842) which will make native thread ids be way less relevant going forward (at least if it becomes adopted by the community -- it's unclear when that'll happen)
* I've learned from @sanchda there's actually alternative ways of doing what `process_vm_readv` does, that we could use as a fallback when `process_vm_readv` is not available/disabled.

So, rather than leave this PR in eternal zombie maybe-I'll-get-back-to-it-maybe-I-won't, I'll close it for now. It'll stay behind as documenting the idea, and in the future we may consider resurrecting in some form or other.

Thanks everyone for the amazing feedback, btw! I learned a lot going through the discussion above.

PR #3056, opened 2023-08-17
include peer service configurations in telemetry payload

wconti27:
Adds the new peer service configurations to telemetry data


ivoanjo (commented):
Hey @wconti27, mind adding tests to `collector_spec.rb` as well? It's rather trivial, but I think it helps make sure we don't accidentally break stuff during refactoring.

spec/datadog/core/telemetry/collector_spec.rb
```diff
@@ -176,6 +176,48 @@
...
+    context 'DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED' do
+      around do |example|
+        ClimateControl.modify(
+          DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED: dd_trace_remove_integration_service_names_enabled
+        ) do
+          example.run
+        end
+      end
+
+      let(:dd_trace_remove_integration_service_names_enabled) { nil }
```
ivoanjo, line 189: Minor: I don't think this is needed? (Maybe a left-over from a previous test?) (Same applies below)

spec/datadog/core/telemetry/collector_spec.rb
```diff
@@ -176,6 +176,48 @@
...
+      let(:dd_trace_peer_service_mapping) { nil }
+      context 'when set' do
+        let(:dd_trace_peer_service_mapping) { 'key:value' }
+        it { is_expected.to include(:DD_TRACE_PEER_SERVICE_MAPPING => 'key:value') }
+      end
+
+      context 'when nil is not included' do
+        let(:dd_trace_peer_service_mapping) { nil }
+        it { is_expected.to_not include(:DD_TRACE_PEER_SERVICE_MAPPING) }
+      end
```
ivoanjo, line 219: Minor: Should we send an empty value when not present? Omitting the key can be confusing sometimes, as you don't know if it means that the data is not being collected or it is being collected, just not being used.

ivoanjo (approved):
:+1: LGTM

spec/datadog/core/telemetry/collector_spec.rb
wconti27: yeah that makes sense

lib/datadog/core/telemetry/collector.rb
```diff
@@ -43,11 +43,20 @@ def application
...
         def configurations
+          peer_service_enabled = Datadog.configuration.tracing.contrib.global_default_service_name.enabled
+
           hash = {
             DD_AGENT_HOST: Datadog.configuration.agent.host,
             DD_AGENT_TRANSPORT: agent_transport,
             DD_TRACE_SAMPLE_RATE: format_configuration_value(Datadog.configuration.tracing.sampling.default_rate),
+            DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED: peer_service_enabled
           }
+          peer_service_mapping_str = ''
```
marcotc, line 54: Do you need this value to be an empty string, instead of `nil`?

I ask because `nil` will remove it from the Hash, reducing the payload size needed to be flushed.

lib/datadog/core/telemetry/collector.rb
```diff
@@ -43,11 +43,20 @@ def application
...
+          unless Datadog.configuration.tracing.contrib.peer_service_mapping.empty?
+            peer_service_mapping = Datadog.configuration.tracing.contrib.peer_service_mapping
```
marcotc, line 56: Can you extract a local variable for `Datadog.configuration`, so this method becomes more succinct?
```suggestion
          config = Datadog.configuration

          hash = {
            DD_AGENT_HOST: Datadog.configuration.agent.host,
            DD_AGENT_TRANSPORT: agent_transport,
            DD_TRACE_SAMPLE_RATE: format_configuration_value(Datadog.configuration.tracing.sampling.default_rate),
            DD_TRACE_REMOVE_INTEGRATION_SERVICE_NAMES_ENABLED: config.tracing.contrib.global_default_service_name.enabled
          }
          peer_service_mapping_str = ''
          unless config.tracing.contrib.peer_service_mapping.empty?
            peer_service_mapping = config.tracing.contrib.peer_service_mapping
```

lib/datadog/core/telemetry/collector.rb
marcotc: Also, it would make it slightly simpler to write this, given we don't have to initialize `peer_service_mapping_str` anymore if we are ok with `nil`.

wconti27: Other tracers include the empty string since it makes it easier to recognize if the configuration is actually supported

spec/datadog/core/telemetry/collector_spec.rb
wconti27: @marcotc See here

marcotc (approved):

PR #3060, opened 2023-08-17
Correctly set `rails.cache.backend` span tag for multiple stores

marcotc:
**What does this PR do?**

This PR fixes the span tag `rails.cache.backend` in `rails.cache` spans to correctly represent the cache store object being instrumented.

**Motivation**

Today, we always set `rails.cache.backend` to `Rails.configuration.cache_store`, which is a global application value.
This works correctly for applications with a single cache store, but incorrectly reports this tag when multiple stores are used.

**Additional Notes**

**How to test the change?**

All changes are covered by testing.


Gemfile
```diff
@@ -1,122 +1,52 @@
...
-  if RUBY_VERSION >= '2.7.0' # Bundler 1.x fails to find that versions >= 3.8.0 are not compatible because of binary gems
-    gem 'google-protobuf', ['~> 3.0', '!= 3.7.0', '!= 3.7.1']
-  elsif RUBY_VERSION >= '2.3.0'
-    gem 'google-protobuf', ['~> 3.0', '!= 3.7.0', '!= 3.7.1', '< 3.19.2']
-  else
-    gem 'google-protobuf', ['~> 3.0', '!= 3.7.0', '!= 3.7.1', '< 3.8.0']
-  end
-end
+# This file was generated by Appraisal
+
```
GustavoCaso, line 2: Could you elaborate on why the need for this change? Should we extract this change into a separate PR?

lib/datadog/tracing/contrib/active_support/cache/instrumentation.rb
```diff
@@ -112,7 +108,26 @@ def read(*args, &block)
...
+                  # As `self` is the store object, we can reverse engineer
+                  # the original symbol by converting the class name to snake case:
+                  # e.g. ActiveSupport::Cache::RedisStore -> active_support/cache/redis_store
+                  # In this case `redis_store` is the store name.
+                  #
+                  # Because there's no API retrieve only the class name
+                  # (only `RedisStore`, and not `ActiveSupport::Cache::RedisStore`)
+                  # the easiest way to retrieve the store symbol is to convert the fully qualified
+                  # name using Rails-provided methods, then extracting the last part.
+                  store: self.class.name.underscore.match(/active_support\/cache\/(.*)/)[1]
```
GustavoCaso, line 130: Would using regex for each command executed on the cache cause performance issues for the customer's application? 

We could memoize the result on an instance variable 

```ruby
def store_name
  return @store_name if defined?(@store_name)
  @store_name = self.class.name.underscore.match(/active_support\/cache\/(.*)/)[1]
  @store_name
end
```

Also, seems we always are getting the last portion of the match. Could we use `split(''/'').last` rather than a regex? 

```ruby
```ruby
def store_name
  return @store_name if defined?(@store_name)
  @store_name = self.class.name.underscore.split('/').last
  @store_name
end
```
```

GustavoCaso: Should we add a comment to explain that `underscore` is from `Rails::ActiveSupport` and that this codes works because is executed within the context of Rails

GustavoCaso:
I noticed that on the `lib/datadog/tracing/contrib/active_support/cache/patcher.rb` file, we have an outdated method
`cache_store_class` which takes `symbol` but no longer do anything with it. 

https://github.com/DataDog/dd-trace-rb/blob/e3d2c37fb5996053b28ea90d12a8ef488561005b/lib/datadog/tracing/contrib/active_support/cache/patcher.rb#L31-L45

Since we are changing related files, should we update it? 

Since that change is not related to the main purpose of the PR, fell free to ignore it

Gemfile
marcotc: Sorry, @GustavoCaso, because it's a draft PR, I make a bunch of mess before getting it ready.
Mucking about with the Gemfile makes my testing easier locally.

marcotc: I also push my changes at the end of day in case my laptop explodes overnight.

lib/datadog/tracing/contrib/active_support/cache/instrumentation.rb
marcotc: Caching is great, I made those changes.

marcotc: Regarding "split+last" vs regex, I'm keeping the regex because it's the version that more closely mirrors ActiveSupport's logic (`require "active_support/cache/#{store}"`).

And because we are only doing it once, as we are now caching, the performance impact will be almost nil.

marcotc:
> I noticed that on the `lib/datadog/tracing/contrib/active_support/cache/patcher.rb` file, we have an outdated method `cache_store_class` which takes `symbol` but no longer do anything with it.
> 
> https://github.com/DataDog/dd-trace-rb/blob/e3d2c37fb5996053b28ea90d12a8ef488561005b/lib/datadog/tracing/contrib/active_support/cache/patcher.rb#L31-L45
> 
> Since we are changing related files, should we update it?
> 
> Since that change is not related to the main purpose of the PR, fell free to ignore it

This is because there's a conditional override for that method, depending if the gem https://github.com/redis-store/redis-activesupport is installed or not. It's still needed for such compatibility:
https://github.com/DataDog/dd-trace-rb/blob/94158dbfa631cba05bae43f5bdea10e7d99a2ec8/lib/datadog/tracing/contrib/active_support/cache/redis.rb#L31-L37

GustavoCaso:
> This is because there's a conditional override for that method, depending if the gem https://github.com/redis-store/redis-activesupport is installed or not. It's still needed for such compatibility:
[dd-trace-rb/lib/datadog/tracing/contrib/active_support/cache/redis.rb](https://github.com/DataDog/dd-trace-rb/blob/94158dbfa631cba05bae43f5bdea10e7d99a2ec8/lib/datadog/tracing/contrib/active_support/cache/redis.rb#L31-L37)

Thanks for the context. It would be nice to note that rails 5.2.0+ has redis cache support. Maybe adding a comment that once we do not support old Rails version we can remove such code 😄

GustavoCaso (approved):
LGTM

marcotc:
> > This is because there's a conditional override for that method, depending if the gem [redis-store/redis-activesupport](https://github.com/redis-store/redis-activesupport) is installed or not. It's still needed for such compatibility:
> > [dd-trace-rb/lib/datadog/tracing/contrib/active_support/cache/redis.rb](https://github.com/DataDog/dd-trace-rb/blob/94158dbfa631cba05bae43f5bdea10e7d99a2ec8/lib/datadog/tracing/contrib/active_support/cache/redis.rb#L31-L37)
> 
> Thanks for the context. It would be nice to note that rails 5.2.0+ has redis cache support. Maybe adding a comment that once we do not support old Rails version we can remove such code 😄

@GustavoCaso I documented this in code, good call!

pbstriker38:
This fails for custom stores that are not in `active_support/cache/`

PR #3062, opened 2023-08-18
Ensure Rails testing is considered a development environment

marcotc:
**What does this PR do?**

This PR detects if a Rails application is running in `development` or `test` mode, and uses this information to inform `ddtrace` internally in order improve logging and background worker management.

**Motivation**

Currently, we consider local development Rails console as a production environment, which is incorrect.

**Additional Notes**

This checking ensures `staging`, which is a common non-stardard Rails environment is considered "not development" by `ddtrace`.

**How to test the change?**

There are unit test covering this change.


sig/datadog/core/environment/execution.rbs
```diff
@@ -15,6 +15,10 @@ module Datadog
         def self.minitest?: () -> bool
 
         def self.rails_development?: -> bool
+
+        RAILS_ENV_DEVELOPMENT: Set[String]
+
+        def self.rails_env_development?: -> untyped
```
anmarchenko, line 21:
```suggestion
        def self.rails_env_development?: -> bool
```

anmarchenko (approved):
I think we need to go forward with that because there are still issues with telemetry and RC being reported.

Do you think we would need to add a check for Cucumber in `test?` method?

sig/datadog/core/environment/execution.rbs
marcotc: Added, thank you!

marcotc:
> Do you think we would need to add a check for Cucumber in `test?` method?

That's a good idea, I'll open a separate PR.

PR #3065, opened 2023-08-22
Detect WebMock to disable telemetry and remote configuration 

TonyCTHsu:
**What does this PR do?**

It has been reported by the community that `ddtrace` is breaking test suite with `WebMock`, since some components are sending http request in the background with `net/http` .

Hence this PR make an effort the detect `WebMock` to provide a better default value for `ddtrace` telemetry and remote configuration. 

`WebMock` does not offer an public API such as `WebMock.enabled?` to check the status. When `WebMock.enable!`, it iterates through http adapters to enable them. The net/http  adapter implements enablement by swapping the reference for `Net::HTTP`

### DIFF
--- a/lib/datadog/core/environment/execution.rb
+++ b/lib/datadog/core/environment/execution.rb
@@ -23,7 +23,7 @@ module Datadog
           #   2. Checking if `Net::HTTP` is referring to the original one
           #   => ::Net::HTTP.equal?(::WebMock::HttpLibAdapters::NetHttpAdapter::OriginalNetHTTP)
           def webmock_enabled?
-            defined?(::WebMock) &&
+            defined?(::WebMock::HttpLibAdapters::NetHttpAdapter) &&
               defined?(::Net::HTTP) &&
               ::Net::HTTP.equal?(::WebMock::HttpLibAdapters::NetHttpAdapter.instance_variable_get(:@webMockNetHTTP))
           end
--- a/lib/datadog/core/environment/execution.rb
+++ b/lib/datadog/core/environment/execution.rb
@@ -10,7 +10,7 @@ module Datadog
           # This can be used to make decisions about when to enable
           # background systems like worker threads or telemetry.
           def development?
-            !!(repl? || test? || rails_development? || webmock_enabled?)
+            !!(webmock_enabled? || repl? || test? || rails_development?)
           end
 
           # WebMock stores the reference to `Net::HTTP` with constant `OriginalNetHTTP`, and when WebMock enables,
--- a/lib/datadog/core/environment/execution.rb
+++ b/lib/datadog/core/environment/execution.rb
@@ -10,7 +10,22 @@ module Datadog
           # This can be used to make decisions about when to enable
           # background systems like worker threads or telemetry.
           def development?
-            !!(repl? || test? || rails_development?)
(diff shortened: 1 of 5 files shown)


spec/datadog/core/environment/execution_spec.rb
```diff
@@ -114,5 +118,94 @@
...
+    context 'when missing constant `Net::HTTP`' do
+      it do
+        hide_const('::Net::HTTP')
+        expect(described_class).not_to be_webmock_enabled
+      end
+    end
+
+    context 'when `WebMock` and `Net::HTTP` constants both exist' do
+      before do
+        WebMock.enable!
```
anmarchenko, line 152: could it make sense to disable WebMock in `after` block? wouldn't it spill in the global scope for another tests?

lib/datadog/core/environment/execution.rb
```diff
@@ -10,7 +10,22 @@ class << self
...
+          # the adapter swaps `Net::HTTP` reference to its mock object, @webMockNetHTTP.
+          #
+          # Hence, we can detect by
+          #   1. Checking if `Net::HTTP` is referring to mock object
+          #   => ::Net::HTTP.equal?(::WebMock::HttpLibAdapters::NetHttpAdapter.instance_variable_get(:@webMockNetHTTP))
+          #
+          #   2. Checking if `Net::HTTP` is referring to the original one
+          #   => ::Net::HTTP.equal?(::WebMock::HttpLibAdapters::NetHttpAdapter::OriginalNetHTTP)
+          def webmock_enabled?
+            defined?(::WebMock) &&
```
marcotc, line 26: We check for `defined?(::WebMock)`, but then access `::WebMock::HttpLibAdapters::NetHttpAdapter`, which I think can be dangerous. I suggest:

```suggestion
            defined?(::WebMock::HttpLibAdapters::NetHttpAdapter) &&
```

spec/datadog/core/environment/execution_spec.rb
```diff
@@ -114,5 +118,94 @@
...
+              require 'webmock'
+              WebMock.enable!
+
+              lib = File.expand_path('lib', __dir__)
+              $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+              require 'datadog/core/environment/execution'
+
+              STDOUT.print "WebMock::VERSION:"
+              STDOUT.print WebMock::VERSION
+              STDOUT.print "=>"
```
marcotc, line 199: Less code to maintain 🤷 
```suggestion
              STDOUT.print "WebMock::VERSION:#{WebMock::VERSION}"
```

spec/datadog/core/environment/execution_spec.rb
```diff
@@ -114,5 +118,94 @@
...
+      end
+
+      it do
+        WebMock.disable!(except: [:net_http])
+
+        expect(described_class).to be_webmock_enabled
+      end
+    end
+
+    [1, 2, 3].each do |version|
```
marcotc, line 178: Are there significant changes from webmock version 1, 2, and 3?

I ask because if their APIs are all the same, we can just test with 3, since 1 and 2 will never receive new updates in the future.

marcotc (approved):

PR #3067, opened 2023-08-22
ekump/remove hard dep on tracing in agent resolver

ekump:
**What does this PR do?**
This PR has no functional changes visible to customers. 

- It moves certain transport related constants to the core module from `tracing` and `ddtrace` modules. The existing constants in those modules are still present, but now reference core. The AgentSettingsResolver in core is updated to reference the core constants, thus reversing dependencies. 

- It creates product specific `AgentSettingsResolver` classes that inherit from the existing `AgentSettingsResolver` in core. HOWEVER, this PR does not introduce any new product specific resolver functionality. It is expected that any product specific agent settings will live in these child classes. In the future, if we come to the conclusion that we have no need for product-specific resolvers, we can remove them and revert back to a single core `AgentSettingsResolver`. 

- The reference to `settings.tracing`is made a soft dependency. If/when tracing is no longer part of the core settings, the resolver should continue to work without error. It just won't be able to reference `transport_options`.

**Motivation**
This is part of a larger effort to remove product dependencies from core. 

**How to test the change?**
No functionality should have changed. Existing tests have been updated to reference the correct resolvers and should continue to pass. An additional test was added to the core resolver unit tests to ensure the resolver continues to work without `settings.tracing` present.


lib/datadog/core/configuration/components.rb
```diff
@@ -81,13 +82,17 @@ def build_telemetry(settings, agent_settings, logger)
         def initialize(settings)
           @logger = self.class.build_logger(settings)
 
+          # This agent_settings is intended for use within Core. If you require
+          # agent_settings within a product outside of core you should extend
+          # the Core resolver from within your product/component's namespace.
           agent_settings = AgentSettingsResolver.call(settings, logger: @logger)
```
ekump, line 88: I ultimately decided to not create component-specific resolvers for telemetry and remote. It felt odd since they already live in core and it's not obvious to me they'd ever have specific agent settings to resolve beyond what is provided in core. Happy to add them if people disagree.

lib/datadog/core/configuration/components.rb
```diff
@@ -81,13 +82,17 @@ def build_telemetry(settings, agent_settings, logger)
         def initialize(settings)
           @logger = self.class.build_logger(settings)
 
+          # This agent_settings is intended for use within Core. If you require
+          # agent_settings within a product outside of core you should extend
+          # the Core resolver from within your product/component's namespace.
```
ekump, line 87: I ultimately decided to not create component-specific resolvers for telemetry and remote. It felt odd since they already live in core and it's not obvious to me they'd ever have specific agent settings to resolve beyond what is provided in core. Happy to add them if people disagree.

spec/datadog/profiling/agent_settings_resolver_spec.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'datadog/profiling/agent_settings_resolver'
+require 'datadog/core/configuration/settings'
```
ekump, line 5: On the one hand, this file is pointless. On the other hand, it feels odd to add a class without a unit test file for it. Happy to kill it.

lib/datadog/core/configuration/agent_settings_resolver.rb
```diff
@@ -201,11 +200,15 @@ def timeout_seconds
...
         def deprecated_for_removal_transport_configuration_proc
-          if settings.tracing.transport_options.is_a?(Proc) && transport_options.adapter.nil?
-            settings.tracing.transport_options
+          if transport_options_settings.is_a?(Proc) && transport_options.adapter.nil?
+            transport_options_settings
           end
         end
 
+        def transport_options_settings
+          @transport_options_settings ||= settings.tracing&.transport_options
```
delner: Is safe navigation supported in all our Rubies? I think we still support 2.2 which doesn't have this.

lib/datadog/core/configuration/components.rb
delner: I think that's reasonable. :+1: We can always add them later if we really need them.

lib/datadog/core/configuration/components.rb
```diff
@@ -81,13 +82,17 @@ def build_telemetry(settings, agent_settings, logger)
...
           agent_settings = AgentSettingsResolver.call(settings, logger: @logger)
 
           @remote = Remote::Component.build(settings, agent_settings)
-          @tracer = self.class.build_tracer(settings, agent_settings)
+          @tracer = self.class.build_tracer(settings, logger: @logger)
```
delner, line 91: Why do we pass `@logger`? Is this because it might need to log during initialization?

lib/datadog/core/configuration/ext.rb
```diff
@@ -19,6 +19,22 @@ module Metrics
 
         module Transport
```
delner, line 20: I would consider changing this name from `Transport`: this name narrows its purpose unnecessarily to just use with `Transport`, and what its describing doesn't really pertain to `Transport`. I might suggest something like `Agent`/`TraceAgent`/`APMAgent`.

delner (commented):
Great start! A few comments so far.

lib/datadog/core/configuration/agent_settings_resolver.rb
ekump: You're correct. CI caught it and I updated to not use safe navigation.

lib/datadog/core/configuration/components.rb
ekump: Correct. It gets used by the AgentSettingsResolver initializer, which now gets done in the specific component initializers, rather than here.

lib/datadog/core/configuration/ext.rb
ekump: I like the idea of using `Agent` as a module name here. But, this file on [master](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/core/configuration/ext.rb) already has `ENV_DEFAULT_HOST` defined under `Datadog::Core::Configuration::Ext::Transport` and it is referenced in the upgrade guide [here](https://github.com/DataDog/dd-trace-rb/blob/master/docs/UpgradeGuide.md?plain=1#L661). Is there any harm in changing this, and updating the doc? 

Alternatively, we could leave `ENV_DEFAULT_HOST` under `Transport` and move the items relevant to my PR into `Agent`, but that does feel a little bit odd.

lib/datadog/core/configuration/agent_settings_resolver.rb
```diff
@@ -201,11 +200,15 @@ def timeout_seconds
...
         # doesn't work, we include the proc directly in the agent settings result.
         def deprecated_for_removal_transport_configuration_proc
-          if settings.tracing.transport_options.is_a?(Proc) && transport_options.adapter.nil?
-            settings.tracing.transport_options
+          if transport_options_settings.is_a?(Proc) && transport_options.adapter.nil?
+            transport_options_settings
           end
         end
 
+        def transport_options_settings
```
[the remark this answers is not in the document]
marcotc: Seem like a fine name to me, given what it does. :)

lib/datadog/core/configuration/ext.rb
marcotc: > it is referenced in the upgrade guide [here](https://github.com/DataDog/dd-trace-rb/blob/master/docs/UpgradeGuide.md?plain=1&rgh-link-date=2023-08-23T14%3A31%3A09Z#L661). Is there any harm in changing this, and updating the doc?

Since we directly instructed clients to use `Datadog::Core::Configuration::Ext::Transport::ENV_DEFAULT_HOST`, we shouldn't break this contract in `ddtrace` 1.x.

> Alternatively, we could leave ENV_DEFAULT_HOST under Transport and move the items relevant to my PR into Agent, but that does feel a little bit odd.

I suggest doing this, despite this solution being a compromise.
I had to do something similar for distributed trace headers: https://github.com/DataDog/dd-trace-rb/blob/f3c7f9672ff4a9d6b5fbc6a2de3a276d294a061e/lib/datadog/tracing/distributed/headers/ext.rb#L7
These constants were moved to their respective implementation classes (e.g. https://github.com/DataDog/dd-trace-rb/blob/f3c7f9672ff4a9d6b5fbc6a2de3a276d294a061e/lib/datadog/tracing/distributed/headers/ext.rb#L20 was moved to https://github.com/DataDog/dd-trace-rb/blob/f3c7f9672ff4a9d6b5fbc6a2de3a276d294a061e/lib/datadog/tracing/distributed/b3_multi.rb#L13), but I'm not able to completely remove them from their original location because they are very much public.

As long as we have a comment documentation that we should remove these constants on the next major release, it's ok to have them duplicated for the time being.

lib/datadog/profiling/agent_settings_resolver.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require_relative '../core/configuration/agent_settings_resolver'
+
+module Datadog
+  module Profiling
+    # Any profiling specific agent settings should be added here rather than in the core resolver
```
marcotc, line 7: I think this description can be more "plain", simply explaining what this class does, and not necessarily referencing its history, something like "this class encapsulates how the profiler communicates with the Datadog agent."

lib/datadog/core/configuration/ext.rb
ekump: @marcotc That's a good point. I don't think I fully understood that users are directly references the constants like that. I'll move `ENV_DEFAULT_HOST` back under transport.

lib/datadog/profiling/agent_settings_resolver.rb
ekump: That's a fair point. I'll update it.

spec/datadog/profiling/agent_settings_resolver_spec.rb
marcotc: 🪓 it! Nothing to test, nothing to see here.

spec/datadog/tracing/configuration/agent_settings_resolver_spec.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+require 'datadog/tracing/configuration/agent_settings_resolver'
+require 'datadog/core/configuration/settings'
```
marcotc, line 5: 🪓

marcotc (approved):
Great work, @ekump! 🙇

lib/datadog/core/configuration/agent_settings_resolver.rb
```diff
@@ -201,8 +201,12 @@ def timeout_seconds
...
         def deprecated_for_removal_transport_configuration_proc
-          if settings.tracing.transport_options.is_a?(Proc) && transport_options.adapter.nil?
-            settings.tracing.transport_options
+          transport_options_settings if transport_options_settings.is_a?(Proc) && transport_options.adapter.nil?
+        end
+
+        def transport_options_settings
+          @transport_options_settings ||= begin
+            settings.tracing.transport_options if settings.respond_to?(:tracing) && settings.tracing
           end
```
ivoanjo, line 210: This is a bit of a regression in the making... Right now, we always load `tracing/configuration/settings` from `core/configuration/settings` so there is no way for `settings.respond_to?(:tracing)` to ever be `false`.

So currently, this code is equivalent to what was before.

BUT, I suggest being careful with next steps. [I had suggested in a previous comment](https://github.com/DataDog/dd-trace-rb/pull/3049#issuecomment-1683553855), we're still  [documenting the `transport_options`](https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/ruby/#configuring-the-transport-layer) as a valid way of configuring dd-trace-rb -- that includes all products, so we need to be careful with the way we plan to make `settings.respond_to?(:tracing)` start returning `false` to not impact existing customers.

(This assuming we plan to touch that before 2.0; if we move things around for 2.0 that seems fine)

lib/datadog/core/configuration/components.rb
```diff
@@ -81,13 +82,17 @@ def build_telemetry(settings, agent_settings, logger)
...
+
           @profiler = Datadog::Profiling::Component.build_profiler_component(
             settings: settings,
-            agent_settings: agent_settings,
+            logger: @logger,
             optional_tracer: @tracer,
           )
```
ivoanjo, line 97: I'd like to request that the profiler still receives the `agent_settings` from the core, as I don't plan on maintaining any profiling-specific changes.

(My whole intention with the creation of the `AgentSettingsResolver` was exactly to have one solution that fits most/all products, rather than needing to keep a separate resolver)

lib/datadog/profiling/agent_settings_resolver.rb
```diff
@@ -0,0 +1,11 @@
...
+
+require_relative '../core/configuration/agent_settings_resolver'
+
+module Datadog
+  module Profiling
+    # This class encapsulates any profiling specific agent settings
+    class AgentSettingsResolver < Datadog::Core::Configuration::AgentSettingsResolver
+    end
+  end
+end
```
ivoanjo, line 11: As I mentioned above, I'd like to request that this gets deleted -- I really don't want to maintain my own copy of the `AgentSettingsResolver`.

sig/datadog/profiling/component.rbs
```diff
@@ -1,42 +1,23 @@
...
-
-      def self.incompatible_libmysqlclient_version?: (untyped settings) -> bool
-
-      def self.load_pprof_support: () -> void
+      def self.no_signals_workaround_enabled?: (untyped settings) -> (false | true)
+      def self.incompatible_libmysqlclient_version?: (untyped settings) -> (true | untyped)
+      def self.load_pprof_support: () -> untyped
     end
   end
 end
```
ivoanjo, line 23: Hey, what happened here? A lot of stuff already had specific types and got replaced with `untyped` O_o?

ivoanjo (commented):
Left a few notes.

sig/datadog/profiling/component.rbs
ivoanjo: Yes -- a lot of the typing on Profiling I've done manually.

ekump: Ah, ok. Sorry. I'll fix this.

lib/datadog/core/configuration/agent_settings_resolver.rb
ekump: We don't plan on removing `transport_options` prior to 2.0. But we do plan on refactoring the Settings class and removing hard dependencies on tracing (I haven't fully researched the idea yet, it's on the list). This should have no effect on users prior to 2.0. 

I don't think this is a dangerous change given then fact that our tests for the `transport_options` proc still expect `settings.tracing` to exist and the rest of resolver code treats `transport_options` as optional.

lib/datadog/core/configuration/components.rb
ekump: What do you think about profiling using the core resolver, but having the profiler component instantiating it, rather than having it passed in? 

`build_profiler_component` would change from

`agent_settings = Datadog::Profiling::AgentSettingsResolver.call(settings, logger: logger)`
to
`agent_settings = Datadog::Core::Configuration::AgentSettingsResolver.call(settings, logger: logger)`.

I believe we are going to look at ways to decouple [Core::Configuration::Components](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/core/configuration/components.rb) prior to 2.0. I think part of that will entail finding a way for components to better encapsulate their own initialization.

lib/datadog/profiling/agent_settings_resolver.rb
ekump: If profiling has no intention of implementing any profiling-specific logic for the agent then I think that's reasonable. I can remove in a follow-up PR.

lib/datadog/core/configuration/agent_settings_resolver.rb
ivoanjo: Yeah, agreed!

lib/datadog/core/configuration/components.rb
ekump: Passing the existing `AgentSettings` to profiling for now doesn't seem problematic, especially since profiling isn't seeking any custom behavior.  I can create a follow-up PR that will revert `build_profiler_component` to its previous state. 

I think we're going to have to rethink how we instantiate components and how they get settings sooner, rather than later. But, we can cross that bridge when we get there.

PR #3071, opened 2023-08-23
Refactor ActiveSupport Cache instrumentation

marcotc:
While trying to make improvements to the ActiveSupport::Cache instrumentation, I noticed that there was a lot of repeated code, and that it used a custom message passing pattern that is not strictly necessary.

This PR refactors this integration by moving all the duplicate behavior into encapsulated methods and removing a bunch of lines in the process :)

### DIFF
--- a/lib/datadog/tracing/contrib/active_support/cache/instrumentation.rb
+++ b/lib/datadog/tracing/contrib/active_support/cache/instrumentation.rb
@@ -10,96 +10,63 @@ module Datadog
       module ActiveSupport
         module Cache
           # Defines instrumentation for ActiveSupport caching
-          # rubocop:disable Lint/RescueException
           module Instrumentation
             module_function
 
-            def start_trace_cache(payload)
-              return unless enabled?
-
-              # In most of the cases Rails ``fetch()`` and ``read()`` calls are nested.
-              # This check ensures that two reads are not nested since they don't provide
-              # interesting details.
-              # NOTE: the ``finish_trace_cache()`` is fired but it already has a safe-guard
-              # to avoid any kind of issue.
-              current_span = Tracing.active_span
-              return if current_span.try(:name) == Ext::SPAN_CACHE &&
-                (
-                  payload[:action] == Ext::RESOURCE_CACHE_GET &&
-                  current_span.try(:resource) == Ext::RESOURCE_CACHE_GET ||
-                  payload[:action] == Ext::RESOURCE_CACHE_MGET &&
-                  current_span.try(:resource) == Ext::RESOURCE_CACHE_MGET
-                )
-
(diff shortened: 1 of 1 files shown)


lib/datadog/tracing/contrib/active_support/cache/instrumentation.rb
```diff
@@ -10,96 +10,63 @@ module Contrib
       module ActiveSupport
         module Cache
           # Defines instrumentation for ActiveSupport caching
-          # rubocop:disable Lint/RescueException
           module Instrumentation
             module_function
 
-            def start_trace_cache(payload)
```
marcotc, line 17: Look at the new code separately from old code is probably the easiest way to review this file.

lib/datadog/tracing/contrib/active_support/cache/instrumentation.rb
```diff
@@ -10,96 +10,63 @@ module Contrib
...
-                if defined?(::Rails)
-                  store, = *Array.wrap(::Rails.configuration.cache_store).flatten
-                  span.set_tag(Ext::TAG_CACHE_BACKEND, store)
-                end
+            # In most of the cases, `#fetch()` and `#read()` calls are nested.
+            # Instrument both does not add any value.
+            # This method checks if these two operations are nested.
+            def nested_read?
+              current_span = Tracing.active_span
+              current_span && current_span.name == Ext::SPAN_CACHE && current_span.resource == Ext::RESOURCE_CACHE_GET
```
anmarchenko, line 44: I think this condition changed a bit from what it was before refactoring:

```ruby
    return if current_span.try(:name) == Ext::SPAN_CACHE &&
    (
        payload[:action] == Ext::RESOURCE_CACHE_GET &&
        current_span.try(:resource) == Ext::RESOURCE_CACHE_GET ||
        payload[:action] == Ext::RESOURCE_CACHE_MGET &&
        current_span.try(:resource) == Ext::RESOURCE_CACHE_MGET
    )
```

It used `current_span.try(:resource)` - I am not very familiar with codebase so I have no idea whether `current_span` really could have no `resource` method

anmarchenko (approved):
It feels my heart with joy to see cleaner code! Thank you.

lib/datadog/tracing/contrib/active_support/cache/instrumentation.rb
marcotc: It cannot: `current_span` can be `nil`, but not have missing methods. It's our very own [SpanOperation](https://github.com/DataDog/dd-trace-rb/blob/afc4a025c7c8717050901f1419b7b21856aa9a2a/lib/datadog/tracing/span_operation.rb).

PR #3074, opened 2023-08-24
[PROF-7307] Include 'ruby vm type' in profiler allocation samples

ivoanjo:
**What does this PR do?**:

This PR builds extends the existing support for the `ThreadContext` collector to sample allocations to also record the "ruby vm type" of sampled objects.

This "ruby vm type" is what Ruby internally represents in its `enum ruby_value_type`, see
https://github.com/ruby/ruby/blob/84a12d657848dfb54f8cc556d344f017a793e95a/include/ruby/internal/value_type.h#L112

**Motivation**:

I'm working on tagging stacks of sampled objects with both the object class, as well as the ruby vm type.

I'm not yet sure how we may expose the ruby vm type in the UX, but having it is useful for development and debugging, and I can imagine it being useful in advanced use cases (e.g. a customer downloading a pprof to look at this data).

As libdatadog interns labels (https://github.com/DataDog/libdatadog/pull/205), the cost of tracking this data is quite low.

**Additional Notes**:

Allocation profiling is still not something that can be enabled; e.g. right now we can only exercise it directly by calling the `ThreadContext` collector.

**How to test the change?**:

### DIFF
--- a/ext/ddtrace_profiling_native_extension/libdatadog_helpers.c
+++ b/ext/ddtrace_profiling_native_extension/libdatadog_helpers.c
@@ -3,38 +3,7 @@
 #include <ruby.h>
 
 const char *ruby_value_type_to_string(enum ruby_value_type type) {
-  switch (type) {
-    case(RUBY_T_NONE    ): return "T_NONE";
-    case(RUBY_T_OBJECT  ): return "T_OBJECT";
-    case(RUBY_T_CLASS   ): return "T_CLASS";
-    case(RUBY_T_MODULE  ): return "T_MODULE";
-    case(RUBY_T_FLOAT   ): return "T_FLOAT";
-    case(RUBY_T_STRING  ): return "T_STRING";
-    case(RUBY_T_REGEXP  ): return "T_REGEXP";
-    case(RUBY_T_ARRAY   ): return "T_ARRAY";
-    case(RUBY_T_HASH    ): return "T_HASH";
-    case(RUBY_T_STRUCT  ): return "T_STRUCT";
-    case(RUBY_T_BIGNUM  ): return "T_BIGNUM";
-    case(RUBY_T_FILE    ): return "T_FILE";
-    case(RUBY_T_DATA    ): return "T_DATA";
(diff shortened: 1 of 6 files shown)


ext/ddtrace_profiling_native_extension/libdatadog_helpers.c
```diff
@@ -0,0 +1,73 @@
...
+    case(RUBY_T_FIXNUM  ): return "T_FIXNUM";
+    case(RUBY_T_UNDEF   ): return "T_UNDEF";
+    case(RUBY_T_IMEMO   ): return "T_IMEMO";
+    case(RUBY_T_NODE    ): return "T_NODE";
+    case(RUBY_T_ICLASS  ): return "T_ICLASS";
+    case(RUBY_T_ZOMBIE  ): return "T_ZOMBIE";
+    #ifdef RUBY_T_MOVED // Introduced in Ruby 2.7
+    case(RUBY_T_MOVED   ): return "T_MOVED";
+    #endif
+                  default: return "T_UNKNOWN_OR_MISSING_RUBY_VALUE_TYPE_ENTRY";
```
marcotc, line 36: Is this a default created for this PR? I couldn't find it the ruby's source code.

If so, I'd say keep it different from the others, and I find it quite hard to parse such a long screaming case string.

I suggest something like `"Unknown Ruby Value"`, or something that will unmistakably read as "Ha, this is not a valid Ruby VM type".

marcotc (approved):

ext/ddtrace_profiling_native_extension/libdatadog_helpers.c
ivoanjo: Changed in d73a72f9527b9c5751f98b06a76e45189c080aed to `BUG: Unknown value for ruby_value_type`. As you suspected, the only objective here was to have an obvious wrong value that we could grep in the codebase and tie back to this function, so we can make it look even more as an obvious wrong value.

ivoanjo:
I was playing with the merge queue feature, please ignore! :)

PR #3076, opened 2023-08-24
Fix bug where Tracing::AgentSettingsResolver wasn't being passed param logger

ekump:
**What does this PR do?**
Fixes a bug introduced in 94158dbfa631cba05bae43f5bdea10e7d99a2ec8 where the tracing component was passing the AgentSettingsResolver `@logger` which isn't initialized at that point. It should be passing AgentSettingsResolver the `logger` passed to the method instead.

Also, re-add types that were accidentally removed in 94158dbfa631cba05bae43f5bdea10e7d99a2ec8 for profiling/component.rbs.

**Motivation**
CI is emitting warnings about `@logger` not being initialized.

**Additional Notes**

**How to test the change?**
CI should no longer emit warnings. Spec tests added to make sure correct logger is passed.

### DIFF
--- a/sig/datadog/profiling/component.rbs
+++ b/sig/datadog/profiling/component.rbs
@@ -28,8 +28,6 @@ module Datadog
 
       def self.enable_gc_profiling?: (untyped settings) -> bool
 
-      def self.print_new_profiler_warnings: () -> void
-
       def self.enable_new_profiler?: (untyped settings) -> bool
 
       def self.no_signals_workaround_enabled?: (untyped settings) -> bool
--- a/sig/datadog/profiling/component.rbs
+++ b/sig/datadog/profiling/component.rbs
@@ -1,23 +1,42 @@
 module Datadog
   module Profiling
     module Component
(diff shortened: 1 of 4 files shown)


marcotc (approved):

sig/datadog/profiling/component.rbs
```diff
@@ -1,23 +1,42 @@
...
+      def self.build_profiler_transport: (
+        untyped settings,
+        Datadog::Profiling::AgentSettingsResolver::AgentSettings agent_settings
+      ) -> untyped
 
-      def self.enable_gc_profiling?: (untyped settings) -> (true | false)
+      def self.enable_gc_profiling?: (untyped settings) -> bool
 
-      def self.enable_new_profiler?: (untyped settings) -> (false | true)
+      def self.print_new_profiler_warnings: () -> void
```
GustavoCaso, line 31: I think that method no longer is present on the Profiler component if that is the case we can get rid of its definition here 😄

ekump: Good catch, removed!

ivoanjo (approved):
:+1: LGTM

PR #3085, opened 2023-08-28
Remove `depends_on` from Core Configuration option

GustavoCaso:
**What does this PR do?**

Based on internal conversations, we agreed that `depends_on` is a feature from the past and that is no longer used. It has been replaced with the different components. Because of that we agreed that it was good to remove it 🔥 

**Motivation:**

We are cleaning up the codebase from options we no longer use. 

I was able to remove the following: 
- The `depends_on` option from the configuration DSL
- `Configuration::DependencyResolver`. 
- `Configuration::OptionDefinitionSet`. It was inherited from Hash, and the only functionality added was to organize the options in order using the `depends_on` attribute. I replaced it with a normal Hash.
- `Configuration::OptionSet`. It was inherited from Hash, but we were not extending the class with any particular functionality. I replaced it with a normal Hash.

**Additional Notes:**

**How to test the change?**

CI must be 🟢 

**For Datadog employees:**


marcotc (approved):
Let's make sure to include in this in our changelog.

Because this feature hasn't been functional to its advertised intent for quite some time, I'm 👍 to remove it in its current form.

ivoanjo:
Nice, hooray for less complexity! :)

PR #3092, opened 2023-08-30
Split appraisal groups for each ruby version

TonyCTHsu:
**Motivation:**

From internal discussion, we want to horizontally scale our appraisal groups, instead of vertically. 

This means 

1. Large appraisal groups would be break into smaller ones
2. Stop sharing group definition between ruby versions (CRuby and JRuby).

**What does this PR do?**

This PR remove the if/else statement from Appraisal file to check each ruby version and read from its own appraisal definition from `appraisal/*.rb` 

I will create more PR to 

1. Remove unnecessary gems from CRuby and JRuby definition.
2. Break `contrib` groups into smaller groups. 

PS: 

* There should not be changes for lockfiles. However, those are changes from bump libddwaf version https://github.com/DataDog/dd-trace-rb/pull/3087


gemfiles/jruby_9.2.8.0_opentelemetry.gemfile.lock
```diff
@@ -1,181 +0,0 @@
-GIT
```
TonyCTHsu, line 1: Deleting this file since it is obsolete

anmarchenko (approved):
great first step, thank you!

Appraisals
```diff
@@ -65,1595 +58,14 @@ def self.gem_cucumber(version)
...
-  # ADD NEW RUBIES HERE
-end
-
 ruby_runtime = if defined?(RUBY_ENGINE_VERSION)
                  "#{RUBY_ENGINE}-#{RUBY_ENGINE_VERSION}"
                else
                  "#{RUBY_ENGINE}-#{RUBY_VERSION}" # For Ruby < 2.3
                end
 
+instance_eval IO.read("appraisal/#{ruby_runtime}.rb")
```
ivoanjo, line 67: Have you considered not including the Ruby minor version in the file names? E.g. Ruby 3.2 instead of 3.2.0?

That would allow us to easily bump Ruby versions we use for CI testing without having to do big mass renames (I know appraisals already does this crime, but ahaha since we're adding this new one, we can get it right from the start ;)

appraisal/jruby-9.4.0.0.rb
```diff
@@ -0,0 +1,143 @@
+appraise 'rails61-mysql2' do
```
ivoanjo, line 1: Minor: Since we're not actually testing with JRuby 9.4, I would suggest just keeping a `# todo: Actually test JRuby 9.4` in this file for now, as it otherwise seems the stuff here is actually supposed to do something

Appraisals
TonyCTHsu: That seems reasonable. I think I will take one small step at a time.

appraisal/jruby-9.4.0.0.rb
ivoanjo: Ah, I meant actually replacing the whole file with a comment; Since we're not using it, this file right now is just a lie :P

PR #3093, opened 2023-08-30
Extract aws appraisal group

TonyCTHsu:
**Motivation**

From internal discussion, we want to horizontally scale our appraisal groups, instead of vertically. This means we are breaking large appraisal groups into smaller ones.

**What does this PR do?**

Since `contrib` is the largest group to slim down and `aws-sdk` includes a lot of AWS dependencies. This PR extracts `aws-sdk` from `contrib` group to an `aws` group. 

I grouped `shoryuken` into `aws` group as well, since its depends on `aws-sdk`.


tasks/appraisal.rake
```diff
@@ -82,6 +82,23 @@ namespace :appraisal do # rubocop:disable Metrics/BlockLength
     end
   end
 
+  desc 'Generate Appraisal gemfile.lock. Takes a version list as argument, defaults to all'
```
TonyCTHsu, line 85: I added this rake task to help me with the migration.

GustavoCaso:
> I grouped shoryuken into aws group as well, since its depends on aws-sdk.

I understand why you might group them, as it would reduce the time it takes to update those dependencies. 
But if we want to split appraisal groups by integrations, it is better to stick with the plan rather than focusing on which gem depends on what since Apprisal takes care of handling the dependencies via bundler.

Also, the task `bundle exec appraisal aws rake spec:shoryuken` seems weird I would expect `bundle exec appraisal shoryuken rake spec:shoryuken`

TonyCTHsu:
For removing `shoryuken` from `contrib` group: Since `shoryuken` depends on `aws-sdk`, removing only `aws-sdk` does not contribute much in term of reducing the size of `contrib`. 

For adding `shoryuken` to `aws` group: I do not want to add a separate group at this moment.

In the short term, my goal right now is to reduce the size of `contrib`. I want to quickly reach to a stage that smaller groups gives us reasonable time spent on resolving dependencies. 

In the long term, I don't have a problem with having individual group for each integration, because it bring us fine granularity we currently do not have.

marcotc (approved):

PR #1583, opened 2021-07-09
Add error_handler option to GRPC tracer configuration

fteem:
## Description

Exposes an `error_handler` option to the GRPC tracer configuration. It uses similar approach to the `error_handler`s for Sidekiq and Faraday.

The motivation behind this is that currently `GRPC::NotFound` errors are being reported as server-side errors, even though (in the semantics of our application) this error is acceptable. Having these errors appear in APM just adds noise and diminishes any other _real_ errors that might appear. 

Here's how that ☝️  looks like in Datadog's UI:

Screen Shot 2021-07-09 at 11 10 45

By having the `error_handler` exposed as an option, we can set it to:

```ruby
c.use :grpc, service_name: 'grpc-service', error_handler: proc { |span, error|
  span.set_error(error) unless span.nil? || error.is_a?(GRPC::NotFound)
}
```

This ☝️ will allow us to ignore the `GPRC::NotFound` error from appearing in APM, while correctly marking any other errors.

N.B. I tried to follow the [contributing guidelines](https://github.com/DataDog/dd-trace-rb/blob/edb8b89debe0e3635d8c6c0e81182efc54413ba2/CONTRIBUTING.md#have-a-patch) as close as possible, but please feel free to let me know if any other changes are needed. 🙏  Thank you in advance!

PR #3095, opened 2023-08-31
fix(grpc): allow custom error_handler for client interceptor

rfwroo:
**What does this PR do?**

In https://github.com/DataDog/dd-trace-rb/pull/1583, the option of passing a custom error_handler to gRPC interceptors was added. This is useful for e.g. selectively marking certain gRPC status codes as an APM error.

However, the change was implemented only for gRPC **server** interceptors. This patch:

* updates `DatadogInterceptor::Client` to pass the `error_handler` into the `Tracing.trace` call
* updates `DatadogInterceptor::Base` to read the error_handler from the interceptor itself (i.e. what's passed [here](https://github.com/DataDog/dd-trace-rb/blob/22d20fbd94a490d6a260b9f92c824b5cfd7122d4/lib/datadog/tracing/contrib/grpc/datadog_interceptor.rb#L20)) instead of the default/global configuration. It's not clear why this wasn't needed for the Server interceptor.

**Motivation:**

To allow `grpc.client` spans to be selectively marked as errors based on gRPC status codes.

**Additional Notes:**

N/A

**How to test the change?**

Example:

```ruby
interceptor = Datadog::Tracing::Contrib::GRPC::DatadogInterceptor::Client.new do |c|
  c.error_handler = proc { |span, error| span.set_error(error) unless error.is_a?(GRPC::NotFound) }
end

# then pass interceptor into gRPC stub
```

**For Datadog employees:**

### DIFF
--- a/spec/datadog/tracing/contrib/grpc/datadog_interceptor/server_spec.rb
+++ b/spec/datadog/tracing/contrib/grpc/datadog_interceptor/server_spec.rb
@@ -4,6 +4,8 @@ require 'datadog/tracing/contrib/analytics_examples'
 require 'datadog/tracing/contrib/environment_service_name_examples'
 require 'datadog/tracing/contrib/span_attribute_schema_examples'
 
+require_relative 'shared_examples'
+
 require 'grpc'
 require 'ddtrace'
 
@@ -70,11 +72,12 @@ RSpec.describe 'tracing on the server connection' do
     end
 
     context 'with an error' do
-      subject(:request_response) do
+      let(:request_response) do
         server.request_response(**keywords) { raise error_class, 'test error' }
       end
 
       let(:error_class) { stub_const('TestError', Class.new(StandardError)) }
+      let(:span_kind) { 'server' }
 
       context 'without an error handler' do
         it do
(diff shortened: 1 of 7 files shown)


TonyCTHsu:
👋 @rfwroo , this looks promising. Let me take some time to play with and give feedback about this 😄

marcotc (approved):
Thank you so much, @rfwroo! 🙇

TonyCTHsu (approved):

TonyCTHsu:
👋 @rfwroo , do you think it make to add a new configuration settings for the client's error handler instead of sharing the same error handler?

rfwroo:
Hey @TonyCTHsu,

> do you think it make sense to add a new configuration settings for the client's error handler instead of sharing the same error handler?

Assuming that you mean the global error handler (i.e. `datadog_configuration[:error_handler]` from [here](https://github.com/DataDog/dd-trace-rb/pull/3095/files#diff-9ca56b81aff94d0152e581e1cca304ed0bf294545c6d076e22fd38bb4803bbfaR62))?

Allowing a global error handler for each span kind sounds neat 👍 

I could follow up with a separate PR if you're happy to merge this as-is?

TonyCTHsu:
👋 @rfwroo, Sorry for my late reply.

Currently, there is only [one `error_handler` configuration](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/grpc/configuration/settings.rb#L49-L52) for the grpc integration. The `error_handler` method, is [defined in the parent object](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/grpc/datadog_interceptor.rb#L61-L63) `DatadogInterceptor::Base` and is [only used by  server interceptor
](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/grpc/datadog_interceptor/server.rb#L27) 
Applying global error handler to client interceptor is **NOT** actually adding a new function, it changes the existing behaviour for users already have error handler defined, which could be an undesired side-effect.

In order to have a smoother upgrade to other users, could you implement a new configuration option `client_error_handler` instead of sharing the same error handler? We could make it explicit that the `error_handler` would be only applying to server from the document.

rfwroo:
@TonyCTHsu that sounds good. Would it also make sense to add a new `server_error_handler` configuration option, and deprecate the existing `error_handler` option?

TonyCTHsu:
> Would it make sense to also add a new `server_error_handler` configuration option, and deprecate the existing `error_handler` option instead?

We want to avoid changes for the existing user, so deprecating without removing it makes sense to me. 😄

spec/datadog/tracing/contrib/grpc/datadog_interceptor/server_spec.rb
```diff
@@ -70,11 +72,12 @@
     end
 
     context 'with an error' do
-      subject(:request_response) do
+      let(:request_response) do
```
TonyCTHsu, line 75: Is there any specific reason to change `subject` to `let` ?

TonyCTHsu (approved):

spec/datadog/tracing/contrib/grpc/datadog_interceptor/server_spec.rb
rfwroo: It's actually more consistent now with the other example blocks in these files, but mostly because the original subject had to be redefined on line 96.

Happy to swap this back if you prefer though.

TonyCTHsu: 👋 @rfwroo , Thanks for the clarification. I think its fine and not a blocker at all. I have triggered our workflow. I believe it is ready to be merged after all the checks are successful. 

Really appreciate the great work! 👍

PR #3099, opened 2023-08-31
[APPSEC-9341] Enable configuring blocking response via Remote Configuration 

GustavoCaso:
**What does this PR do?**

Update ASM to be able to configure the blocking response via Remote Configuration.

**Motivation:**

Allow customers to change the blocking response via the DD UI

The PR includes multiple changes:
- Update the ASM remote capabilities 
- Store in memory the actions configured via the DD UI. `AppSec::Processor::Actions` apart from storing in memory, we need to merge existing actions with the ones coming from the Remote Configuration and merging.  
The logic for merging replaces existing actions with new ones coming from remote configuration and keeps the ones not updated. 
- Modify the `AppSec::Response.negotiate` method to account for the WAF result actions and the configured information

**Additional Notes:**

I've removed unused returned values from the methods `Reactive::<Class>.subscribe`, and `Reactive::<Class>.publish`

The returned values from the reactive engine are plain arrays ex: `[:block, event]` or `[:monitor, event]` those could be refactored into their own classes `AppSec::Processor::Actions::Block` and `AppSec::Processor::Actions::Monitor`. I left that exercise to a future PR. 

**How to test the change?**

**For Datadog employees:**

### DIFF
--- a/lib/datadog/appsec/monitor/gateway/watcher.rb
+++ b/lib/datadog/appsec/monitor/gateway/watcher.rb
@@ -24,7 +24,7 @@ module Datadog
                 scope = Datadog::AppSec.active_scope
 
                 AppSec::Reactive::Operation.new('identity.set_user') do |op|
-                  Monitor::Reactive::SetUser.subscribe(op, scope.processor_context) do |result, _block|
+                  Monitor::Reactive::SetUser.subscribe(op, scope.processor_context) do |result|
                     if result.status == :match
                       # TODO: should this hash be an Event instance instead?
                       event = {
@@ -44,10 +44,10 @@ module Datadog
                     end
                   end
 
-                  _result, block = Monitor::Reactive::SetUser.publish(op, user)
+                  block = Monitor::Reactive::SetUser.publish(op, user)
                 end
 
-                throw(Datadog::AppSec::Ext::INTERRUPT, [nil, [:block, event]]) if block
+                throw(Datadog::AppSec::Ext::INTERRUPT, [nil, [[:block, event]]]) if block
 
                 ret, res = stack.call(user)
 
--- a/lib/datadog/appsec.rb
+++ b/lib/datadog/appsec.rb
@@ -23,12 +23,12 @@ module Datadog
         appsec_component.processor if appsec_component
       end
 
-      def reconfigure(ruleset:)
+      def reconfigure(ruleset:, actions:)
         appsec_component = components.appsec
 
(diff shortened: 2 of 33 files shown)


lib/datadog/appsec/monitor/gateway/watcher.rb
```diff
@@ -44,10 +44,10 @@ def watch_user_id(gateway = Instrumentation.gateway)
                     end
                   end
 
-                  _result, block = Monitor::Reactive::SetUser.publish(op, user)
+                  block = Monitor::Reactive::SetUser.publish(op, user)
                 end
 
-                throw(Datadog::AppSec::Ext::INTERRUPT, [nil, [:block, event]]) if block
+                throw(Datadog::AppSec::Ext::INTERRUPT, [nil, [[:block, event]]]) if block
```
lloeki, line 50: Curious about the double array here: was it missing before or is that an intended API change?

lloeki (approved):
LGTM, just a Q.

I think for those catch/throw protocols we should (but not in this PR) move to instances instead of hashes and arrays, it should make things easier to follow (and properly type-checked)

lib/datadog/appsec/monitor/gateway/watcher.rb
GustavoCaso: In was missing 😨

GustavoCaso: While missing the functionality still worked because we did not use the values 

```ruby
[:block, {}].any? {|action, _event| action == :block }
# => true
```

We now care about the returned value with the code changes on the request middleware. We use `find` rather than `any?`

```ruby
blocked_event = request_response.find { |action, _options| action == :block }
```


Old code
```ruby
blocked_event = [:block, {}].find {|action, _event| action == :block }
#=> :block
```

New Code
```ruby
blocked_event = [[:block, {}]].find {|action, _event| action == :block }
# => [:block, {}]
```

GustavoCaso:
> I think for those catch/throw protocols we should (but not in this PR) move to instances instead of hashes and arrays, it should make things easier to follow (and properly type-checked)

Absolutely 😄 

I mentioned in the PR description 

> The returned values from the reactive engine are plain arrays ex: [:block, event] or [:monitor, event] those could be refactored into their own classes AppSec::Processor::Actions::Block and AppSec::Processor::Actions::Monitor. I left that exercise to a future PR.

PR #3100, opened 2023-09-01
[CIAPP-5371] CI visibility: validate git tags

anmarchenko:
**What does this PR do?**

In cases when it is not possible to automatically extract git environment information for CI visibility we allow users to supply this info via environment variables, see docs [here](https://docs.datadoghq.com/continuous_integration/troubleshooting/#data-appears-in-test-runs-but-not-tests). 

DD_GIT_REPOSITORY_URL and DD_GIT_COMMIT_SHA are required but currently not validated. When these values are not provided or SHA is not valid CI-APP backend silently drops these traces. End users get no feedback in these cases.

We add the following validations to these tags:
- DD_GIT_REPOSITORY_URL must be present
- DD_GIT_COMMIT_SHA must be present, must be 40 characters long must be a valid hex number

In case these values are incorrect we continue to send traces but we log an error message explaining the problem.

**Motivation:**

Improve end user experience and prevent confusion when test spans are not shown in UI without any reason.

**How to test the change?**
Covered by unit tests.

**For Datadog employees:**

### DIFF
--- a/spec/datadog/ci/ext/environment_spec.rb
+++ b/spec/datadog/ci/ext/environment_spec.rb
@@ -6,6 +6,13 @@ require 'datadog/ci/ext/environment'
 RSpec.describe Datadog::CI::Ext::Environment do
   FIXTURE_DIR = "#{File.dirname(__FILE__)}/fixtures/" # rubocop:disable all
 
+  let(:logger) { instance_double(Datadog::Core::Logger) }
+  before do
+    allow(Datadog).to receive(:logger).and_return(logger)
+    allow(logger).to receive(:debug)
+    allow(logger).to receive(:error)
+  end
+
   describe '.tags' do
     subject(:tags) do
       ClimateControl.modify(environment_variables) { described_class.tags(env) }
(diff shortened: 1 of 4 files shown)


spec/datadog/ci/ext/environment_spec.rb
```diff
@@ -110,49 +117,139 @@
...
+              'DD_GIT_COMMIT_COMMITTER_DATE' => '2021-06-19T18:35:10+00:00',
             }
-          )
+          end
+
+          it 'returns user provided metadata' do
+            is_expected.to eq(
+              {
+                'ci.workspace_path' => "#{Dir.pwd}/spec/datadog/ci/ext/fixtures/git",
+                'git.branch' => env['DD_GIT_BRANCH'],
```
TonyCTHsu, line 150: Can we change the assertion with the actual value instead of an indirection through environment variable?

TonyCTHsu (approved):

spec/datadog/ci/ext/environment_spec.rb
anmarchenko: sorry, I missed your comment :( 

I think in this case though it would make it only more magical to put an actual value here as these env vars are defined in the same context

PR #3103, opened 2023-09-01
[PROF-7307] Wire up allocation sampling into `CpuAndWallTimeWorker`

ivoanjo:
**What does this PR do?**

In #2657 we added support for the `Collectors::ThreadContext` to sample a given object allocation.

But until now, triggering these kinds of allocations was only done in our test suite.

This PR adds the missing part: actually sampling objects as Ruby allocates them.

In the `Collectors::CpuAndWallTimeWorker`, we already had a `TracePoint` that is called on every new object allocation (`RUBY_INTERNAL_EVENT_NEWOBJ`), but we were only using it to count allocations.

This PR changes the `TracePoint` code to actually call into the `Collectors::ThreadContext` and thus allow allocation sampling to happen.

(Also included is a crash fix + another tiny tweak. I suggest reviewing commit-by-commit as it makes more sense!)

**Motivation:**

This PR is another step towards the allocation profiling feature. 

**Additional Notes:**

It's not yet possible to turn on allocation profiling via configuration.

I am reserving that change for later. But it's easy to do: change the `profiling/component.rb` to use `alloc_samples_enabled: true` and `allocation_sample_every: 1`.

This will result in a big overhead impact to the application. I'll be working on this in later PRs.

In particular, the sampling approach we're using (`allocation_sample_every`), is only a placeholder; it will get replaced before we get to beta. This is why I don't want to add it as a real user-available setting -- because it's temporary.

**How to test the change?**

This change includes test coverage. You can also manually check it out, by following the instructions in the "Additional Notes".

(There's a feature flag needed for this data to show up in the Datadog UX).

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
@@ -206,6 +206,7 @@ static VALUE _native_reset_after_fork(DDTRACE_UNUSED VALUE self, VALUE collector
 static VALUE thread_list(struct thread_context_collector_state *state);
 static VALUE _native_sample_allocation(DDTRACE_UNUSED VALUE self, VALUE collector_instance, VALUE sample_weight, VALUE new_object);
 static VALUE _native_new_empty_thread(VALUE self);
+ddog_CharSlice ruby_value_type_to_class_name(enum ruby_value_type type);
 
 void collectors_thread_context_init(VALUE profiling_module) {
   VALUE collectors_module = rb_define_module_under(profiling_module, "Collectors");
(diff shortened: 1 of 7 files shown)


ext/ddtrace_profiling_native_extension/collectors_thread_context.c
```diff
@@ -1122,17 +1123,30 @@ void thread_context_collector_sample_allocation(VALUE self_instance, unsigned in
...
+      // Thus, we need to make sure there's actually a class before getting its name.
+
+      if (klass != 0) {
+        const char *name = rb_obj_classname(new_object);
+        size_t name_length = name != NULL ? strlen(name) : 0;
+
+        if (name_length > 0) {
+          class_name = (ddog_CharSlice) {.ptr = name, .len = name_length};
+        } else {
+          // @ivoanjo: I'm not sure this can ever happen, but just-in-case
```
ekump, line 1139: Is it possible for this to happen with anonymous classes?

ekump (approved):

ext/ddtrace_profiling_native_extension/collectors_thread_context.c
ivoanjo: Good question! It shouldn't happen -- we get the Ruby autogenerated name for them:

```ruby
[1] pry(main)> Class.new.to_s
=> "#<Class:0x00005b721ef2a558>"
```

Here's the output of an actual profile with such a class:

```
allocation class:[#<Class:0x00005dd2c7da98a8>] ruby vm type:[T_OBJECT] thread id:[123612321531712 (2560)] thread name:[main]
```

PR #3104, opened 2023-09-04
[PROF-8139] Upgrade to libdatadog 4

ivoanjo:
**What does this PR do?**

This PR upgrades dd-trace-rb to use libdatadog 4. This release includes a number of memory footprint improvements, but there were also a few breaking API changes, which needed adjustments in our code.

**Motivation:**

Use the latest libdatadog release.

**Additional Notes:**

✅ Done ~~**I'm opening this PR as a draft until the libdatadog 4 release is up on rubygems.org, but I've already tested it locally.**~~

✅ Done ~~**After libdatadog 4 gets released on rubygems.org, we should update the appraisals and then mark the PR as non-draft**.~~

The backwards incompatible changes were mainly these 3:
* Instead of having a `ddog_prof_Location` be able to include multiple `ddog_prof_Line`s, now a location can only include a single line.

  In our code we already always used 1-line-per-location only, so we only had to adjust to the `ddog_pprof_Line` getting removed.

* The `ddog_prof_Profile_new` now returns a `ddog_prof_Profile`, instead of a pointer; conversely, a bunch of other APIs now want a pointer to a `ddog_prof_Profile` instead of the `ddog_prof_Profile` directly.

* A bunch of APIs started returning `ddog_prof_Profile_Result`.

**How to test the change?**

Our existing test coverage already exercises all of these code paths.

**For Datadog employees:**

### DIFF
--- a/ext/ddtrace_profiling_native_extension/stack_recorder.c
+++ b/ext/ddtrace_profiling_native_extension/stack_recorder.c
@@ -165,10 +165,10 @@ static const uint8_t all_value_types_positions[] = {CPU_TIME_VALUE_ID, CPU_SAMPL
 // Contains native state for each instance
 struct stack_recorder_state {
   pthread_mutex_t slot_one_mutex;
-  ddog_prof_Profile *slot_one_profile;
+  ddog_prof_Profile slot_one_profile;
 
   pthread_mutex_t slot_two_mutex;
-  ddog_prof_Profile *slot_two_profile;
+  ddog_prof_Profile slot_two_profile;
 
   short active_slot; // MUST NEVER BE ACCESSED FROM record_sample; this is NOT for the sampler thread to use.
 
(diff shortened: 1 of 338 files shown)


ekump (approved):
My C is a bit rusty, but LGTM.

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -386,8 +387,9 @@ static VALUE _native_serialize(DDTRACE_UNUSED VALUE _self, VALUE recorder_instan
   VALUE start = ruby_time_from(ddprof_start);
   VALUE finish = ruby_time_from(ddprof_finish);
 
-  if (!ddog_prof_Profile_reset(args.profile, NULL /* start_time is optional */ )) {
-    return rb_ary_new_from_args(2, error_symbol, rb_str_new_cstr("Failed to reset profile"));
+  ddog_prof_Profile_Result reset_result = ddog_prof_Profile_reset(args.profile, NULL /* start_time is optional */ );
+  if (reset_result.tag == DDOG_PROF_PROFILE_RESULT_ERR) {
+    return rb_ary_new_from_args(2, error_symbol, rb_sprintf("Failed to reset profile: %"PRIsVALUE, get_error_details_and_drop(&reset_result.err)));
   }
```
GustavoCaso, line 393: Could we use `reset_profile` here? 

Maybe the function `reset_profile` could return if the profiler successfully rest and if not the caller can choose what to do

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -428,7 +430,7 @@ void record_sample(VALUE recorder_instance, ddog_prof_Slice_Location locations,
 
   sampler_unlock_active_profile(active_slot);
 
-  if (result.tag == DDOG_PROF_PROFILE_ADD_RESULT_ERR) {
+  if (result.tag == DDOG_PROF_PROFILE_RESULT_ERR) {
```
GustavoCaso, line 433: This was not mentioned in the breaking changes on the PR description. Is this expected?

ivoanjo: Yes! This was slightly folded under

> A bunch of APIs started returning ddog_prof_Profile_Result.

The API changed to no longer return an `AddResult` and use a regular `Result` (diff of `profiling.h` header):

```diff
-struct ddog_prof_Profile_AddResult ddog_prof_Profile_add(struct ddog_prof_Profile *profile,
+struct ddog_prof_Profile_Result ddog_prof_Profile_add(struct ddog_prof_Profile *profile,
                                                          struct ddog_prof_Sample sample);
```

Error handling on result structure is similar, but the constant to indicate the error is different.

ext/ddtrace_profiling_native_extension/stack_recorder.c
ivoanjo: Simplified as 80d8cecf27988952a8cc180655fd17788629a64c

GustavoCaso (approved):

PR #3108, opened 2023-09-05
Add description for test task

TonyCTHsu:
**What does this PR do?**

When `bundle exec rake -T test`, the information is missing because the rake task does not have a description. So I added a description with example to help people understand how to invoke the task. 

```
dd-trace-rb tonycthsu/add-test-task-description % bundle exec rake -T test
rake spec:minitest              # Run RSpec code examples
rake test[rake_task,task_args]  # Run test with its coresponding dependencies, example: `bundle exec rake test[spec:redis]`
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -270,6 +270,7 @@ TEST_METADATA = {
 }.freeze
 # rubocop:enable Layout/HashAlignment
 
+desc 'Run test with its coresponding dependencies, example: `bundle exec rake test[spec:redis]`'
 task :test, [:rake_task, :task_args] do |_, args|
   spec_task = args.rake_task
   spec_arguments = args.task_args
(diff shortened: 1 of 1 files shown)


anmarchenko (approved):

Rakefile
```diff
@@ -270,6 +270,7 @@ TEST_METADATA = {
 }.freeze
 # rubocop:enable Layout/HashAlignment
 
+desc 'Run test with its coresponding dependencies, example: `bundle exec rake test[spec:redis]`'
```
GustavoCaso, line 273: Depending on the shell call `test[spec:redis]` might break.

In Zsh:

```
bundle exec rake test[spec:redis]
zsh: no matches found: test[spec:redis]
```

To fix that we can wrap the brackets with singles quotes 

`bundle exec rake test'[spec:redis]'` I thin is good to update the example to avoid that error from happening

PR #3109, opened 2023-09-05
Fix missing endpoint profiling when request_queuing is enabled in rack instrumentation

ivoanjo:
**What does this PR do?**

This PR tweaks the logic that decides if the profiler should collect the resource from an active trace to also consider the span type `proxy`, which is used when the `request_queuing` feature is in use in the rack instrumentation.

Without this fix, the endpoint profiling feature (e.g. allowing customers to drill into their profiles by endpoint) does not work in combination with `request_queuing`, and the feature is hidden in the profiler UX.

**Motivation:**

Make sure that customers that enable `request_queuing` are still able to look at their profiles per-endpoint.

**Additional Notes:**

N/A

**How to test the change?**

The change includes test coverage. If you're interested in testing it out with a real http app, you can do so by enabling the `request_queuing` feature:

```ruby
Datadog.configure do |c|
  c.tracing.instrument :rack, request_queuing: true
end
```

and then perform a request to the app including the `X-Request-Start` header:

```
$ curl --header "X-Request-Start: t=1693901946000" http://localhost:8080/foo
```

This will generate a `queue` span, and with this change, the endpoint will show up in the right sidebar when looking at profiles for this application. Without this change, it would not show up at all for requests with the `queue` span.

**For Datadog employees:**

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
@@ -201,7 +201,7 @@ static long cpu_time_now_ns(struct per_thread_context *thread_context);
 static long thread_id_for(VALUE thread);
 static VALUE _native_stats(VALUE self, VALUE collector_instance);
 static void trace_identifiers_for(struct thread_context_collector_state *state, VALUE thread, struct trace_identifiers *trace_identifiers_result);
-static bool is_type_web(VALUE root_span_type);
+static bool should_collect_resource(VALUE root_span_type);
 static VALUE _native_reset_after_fork(DDTRACE_UNUSED VALUE self, VALUE collector_instance);
 static VALUE thread_list(struct thread_context_collector_state *state);
 static VALUE _native_sample_allocation(DDTRACE_UNUSED VALUE self, VALUE collector_instance, VALUE sample_weight, VALUE new_object);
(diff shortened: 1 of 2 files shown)


ext/ddtrace_profiling_native_extension/collectors_thread_context.c
```diff
@@ -1051,11 +1051,21 @@ static void trace_identifiers_for(struct thread_context_collector_state *state,
...
+// ddtrace integrations, but worth mentioning just in case :)
+static bool should_collect_resource(VALUE root_span_type) {
   ENFORCE_TYPE(root_span_type, T_STRING);
 
-  return RSTRING_LEN(root_span_type) == strlen("web") &&
-    (memcmp("web", StringValuePtr(root_span_type), strlen("web")) == 0);
+  int root_span_type_length = RSTRING_LEN(root_span_type);
+  const char *root_span_type_value = StringValuePtr(root_span_type);
+
+  return (root_span_type_length == strlen("web") && (memcmp("web", root_span_type_value, strlen("web")) == 0)) ||
```
ekump, line 1067: Since you are potentially doing 2 `strlen()` calls per string per call does it make sense to make these constants so we don't evaluate them every function call?

GustavoCaso: Fair point

ivoanjo: Since this is `strlen` on a constant string, the compiler will optimize it away, even at `-O0`, see: https://godbolt.org/z/Wb351Monv :)

ivoanjo:
I've rebased this on top of current master without any other changes so as to incorporate the new CI job which is a requirement for merging.

PR #3110, opened 2023-09-05
rename system test action to a single word

GustavoCaso:
**What does this PR do?**

Rename the system test GH action to a single word to test if we can use this action in our branch protection. Currently, we can select individual jobs, but not the entire action.

image

**Motivation:**

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**


ivoanjo (approved):
:+1: LGTM

PR #3113, opened 2023-09-05
Instrument gems that are required after `Datadog.configure`

marcotc:
**What does this PR do?**

When `Datadog.configure` tries to instrument a gem but that gem is not loaded it currently gives up.

With this PR, it now registers a hook on the `Kernel.require` method that will instrument the respective gem when that gem is loaded. 

**Motivation:**

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**

**To do**
- [ ] Test this with all combinations of integrations we support (e.g. load `activesupport` before `rails`. Load `rails` before `activesupport`. Same for `typhoeus` and `ethon`.)


lib/datadog/tracing/contrib/kernel.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        # Helper class to register callbacks for `Kernel#require`.
+        # Having a separate class allows for easier testing.
+        class Instance
+          def initialize
+            @on_require = {}
+          end
+
+          # Executes the callback for when a gem is loaded.
+          def require(name)
```
p-datadog, line 65: How does this handle nested requires (application -> gem -> individual files within the gem)?

ivoanjo:

ivoanjo:
This has been around for a long time, so I'm going to go ahead and close it -- please do reopen if this is still useful!

PR #3114, opened 2023-09-06
Refactor: Simplify `Profiler` instance

ivoanjo:
**What does this PR do?**

This PR simplifies the `Profiler` class to receive only a single `worker`, instead of a `collectors` array.

**Motivation:**

The `collectors` array was added a long time ago when we expected that we may add more `collectors` in the future.

We never ended up making use of this extra flexibility, so I've decided to simplify the code to remove this feature.

**Additional Notes:**

N/A

**How to test the change?**

This change includes code coverage. Since this is the core of starting the profiler, it's pretty visible when it doesn't work ;)

**For Datadog employees:**

### DIFF
--- a/benchmarks/profiler_sample_loop.rb
+++ b/benchmarks/profiler_sample_loop.rb
@@ -28,7 +28,8 @@ class ProfilerSampleLoopBenchmark
     Datadog.shutdown!
 
     # Call collection directly
-    @stack_collector = Datadog.send(:components).profiler.collectors.first
+    # Note -- this is going to be deprecated/removed soon
+    @stack_collector = Datadog.send(:components).profiler.send(:worker)
     @recorder = @stack_collector.recorder
   end
 
--- a/lib/datadog/profiling.rb
+++ b/lib/datadog/profiling.rb
@@ -68,7 +68,8 @@ module Datadog
 
     def self.enabled?
       profiler = Datadog.send(:components).profiler
-      !!(profiler.scheduler.running? if profiler)
+      # Use .send(...) to avoid exposing the attr_reader as an API to the outside
+      !!(profiler.send(:scheduler).running? if profiler)
     end
 
     private_class_method def self.replace_noop_allocation_count
--- a/lib/datadog/profiling/component.rb
+++ b/lib/datadog/profiling/component.rb
@@ -70,6 +70,7 @@ module Datadog
 
         no_signals_workaround_enabled = false
         timeline_enabled = false
+        worker = nil
 
         if enable_new_profiler?(settings)
           no_signals_workaround_enabled = no_signals_workaround_enabled?(settings)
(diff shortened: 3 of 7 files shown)


lib/datadog/profiling/component.rb
```diff
@@ -97,7 +98,7 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
           load_pprof_support
 
           recorder = build_profiler_old_recorder(settings)
-          collector = build_profiler_oldstack_collector(settings, recorder, optional_tracer)
+          worker = build_profiler_oldstack_collector(settings, recorder, optional_tracer)
```
GustavoCaso, line 101: Should we rename `build_profiler_oldstack_collector` to `build_profiler_oldstack_worker`?

benchmarks/profiler_sample_loop.rb
```diff
@@ -28,7 +28,8 @@ def create_profiler
     Datadog.shutdown!
 
     # Call collection directly
-    @stack_collector = Datadog.send(:components).profiler.collectors.first
+    # Note -- this is going to be deprecated/removed soon
+    @stack_collector = Datadog.send(:components).profiler.send(:worker)
     @recorder = @stack_collector.recorder
```
GustavoCaso, line 33:
```suggestion
    @stack_worker = Datadog.send(:components).profiler.send(:worker)
    @recorder = @stack_worker.recorder
```

lib/datadog/profiling.rb
```diff
@@ -68,7 +68,8 @@ def self.allocation_count
 
     def self.enabled?
       profiler = Datadog.send(:components).profiler
-      !!(profiler.scheduler.running? if profiler)
+      # Use .send(...) to avoid exposing the attr_reader as an API to the outside
+      !!(profiler.send(:scheduler).running? if profiler)
```
GustavoCaso, line 72: Should we add an instance method to the profiler that returns if it is running? 

`profiler.running?`

The method internally can call the private scheduler attribute and any other ways for it to check if is running.

That way we can get away from using `send`

lib/datadog/profiling/profiler.rb
```diff
@@ -1,35 +1,35 @@
...
       include Datadog::Core::Utils::Forking
 
-      attr_reader \
-        :collectors,
-        :scheduler
+      private
 
-      def initialize(collectors, scheduler)
-        @collectors = collectors
+      attr_reader :worker, :scheduler
```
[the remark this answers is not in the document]
ivoanjo: I actually do like having the `attr_reader`s, when they exist, at the top as a form of documentation: "Here's the things we're storing in this class".

lib/datadog/profiling/component.rb
ivoanjo: I don't think it's worth it -- the new profiler has been released, and on by default for almost two quarters.

I was keeping the old code around as a safety policy against "unknown-unknowns", e.g. a customer running into an issue where the new profiler really couldn't work for them, but after almost two quarters, this did not happen and the only reason this code hasn't yet been deleted is time ;) 

So TL;DR: I don't think it's worth spending any more time on code that's about to be deleted.

benchmarks/profiler_sample_loop.rb
ivoanjo: For the same reason as https://github.com/DataDog/dd-trace-rb/pull/3114/files#r1318276131, I don't think it's worth spending time on code that's about to be deleted ;)

lib/datadog/profiling.rb
GustavoCaso: Sounds good 😄 

We can always revisit this if we ever need from another component

GustavoCaso (approved):

PR #3115, opened 2023-09-07
Improve injection logging message and doc

TonyCTHsu:
**Motivation:**
After discussion with PM and the support team, improve the logging message for clarification.

**What does this PR do?**

For the known constraints. 

1. Log message to standard output. 
2. Describe the constraint 

For unknown error: 

1. Given suggestion to contact support team.


lib-injection/auto_inject.rb
```diff
@@ -2,22 +2,29 @@
...
     return
   end
 
   require 'bundler'
+  require "bundler/cli"
   require 'shellwords'
 
   if Bundler.frozen_bundle?
-    STDERR.puts "[ddtrace] #{failure_prefix} Cannot inject with frozen Gemfile, run `bundle config unset deployment` to allow lib injection. To learn more about bundler deployment, check https://bundler.io/guides/deploying.html#deploying-your-application. #{support_message}"
+    STDERR.puts "[ddtrace] Injection failed: Frozen `Gemfile` because Bundler is configured with `deployment`."
```
wantsui, line 22: What about the message to be clear that it's unable to inject?
[ddtrace] Injection failed: Unable to inject into a Frozen `Gemfile` (Bundler is configured with `deployment`) 

?

wantsui: Also will there be a guide explaining what to do for frozen Gemfiles?

wantsui (commented):
I left some small comments for now.

lib-injection/README.md
```diff
@@ -1,4 +1,12 @@
-# Host injection
+### Injection
+
+The inject script would add `ddtrace` to your application's `Gemfile` to [instrument your application](https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/ruby/#rails-or-hanami-applications)).
+
+* Supports `Ruby on Rails` / `Hanami` application
+* Supports `Bundler` version must be >= 2.3
+* Does not support vendoring gem ([Bundler's Deployment Mode](https://www.bundler.cn/man/bundle-install.1.html#DEPLOYMENT-MODE) or setting `BUNDLE_PATH`)
```
wantsui, line 7: Is there an example of what vendoring gem means? (Ie: It's not clear to me that this would affect the frozen gemfile scenario).

lib-injection/auto_inject.rb
```diff
@@ -2,22 +2,29 @@
...
+    return
+  end
+
+  unless Bundler::CLI.commands["add"] && Bundler::CLI.commands["add"].options.key?("require")
+    STDERR.puts "[ddtrace] Injection failed: Bundler version #{Bundler::VERSION} is not supported. Please upgrade >= 2.3."
```
marcotc, line 27:
```suggestion
    STDERR.puts "[ddtrace] Injection failed: Bundler version #{Bundler::VERSION} is not supported. Upgrade to Bundler >= 2.3 to enable injection."
```

lib-injection/README.md
marcotc: Aren't these two different things?
* Vendoring with BUNDLE_PATH.
* Deployment mode.

If they are different, we should make it clear that these are not both the same kind of "vendoring gem".

lib-injection/host_inject.rb
```diff
@@ -23,30 +23,26 @@ def debug_log(msg)
...
   require "fileutils"
 
-  failure_prefix = "Datadog lib injection failed:"
   support_message = "For help solving this issue, please contact Datadog support at https://docs.datadoghq.com/help/."
 
   unless Bundler::SharedHelpers.in_bundle?
-    debug_log "[ddtrace] Not in bundle... skipping host injection"
+    debug_log "[ddtrace] Not in bundle... skipping injection"
     return
   end
```
ivoanjo, line 31: Since a lot of stuff between both files is copy/pasted, consider adding a note atop both files mentioning that and reminding people to keep them in sync.

ivoanjo (commented):
Left a few notes!

lib-injection/auto_inject.rb
```diff
@@ -60,18 +67,18 @@
...
+      STDERR.puts "[ddtrace] Injection failed: Unable to add ddtrace. Error output:\n#{output.split("\n").map {|l| "[ddtrace] #{l}"}.join("\n")}\n#{support_message}"
     end
   ensure
     # Remove the copies
     FileUtils.rm datadog_gemfile
     FileUtils.rm datadog_lockfile
   end
 rescue Exception => e
-  STDERR.puts "[ddtrace] #{failure_prefix} #{e.class.name} #{e.message}\nBacktrace: #{e.backtrace.join("\n")}\n#{support_message}"
+  STDERR.puts "[ddtrace] Injection failed: #{e.class.name} #{e.message}\nBacktrace: #{e.backtrace.join("\n")}\n#{support_message}"
```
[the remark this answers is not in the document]
TonyCTHsu: I think having both `ddtrace` and `Datadog` is redundant. I keep `ddtrace` because it is more specific.

lib-injection/auto_inject.rb
```diff
@@ -60,18 +67,18 @@
     )
 
     if status.success?
-      STDOUT.puts '[ddtrace] Datadog lib injection successfully added ddtrace to the application.'
+      STDOUT.puts '[ddtrace] Injection adds ddtrace to the application successfully.'
```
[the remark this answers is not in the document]
TonyCTHsu: I can approach our doc writers(maybe for the next iteration), can you live with this for now? Or you believe it is a blocker?

lib-injection/auto_inject.rb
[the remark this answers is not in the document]
ivoanjo: My thinking is, the person using lib injection or seeing this log message may not even know anything about directly:

1. They may be someone that is trying to enable lib injection for everyone, and perhaps don't know the details (e.g. that a "ddtrace" component is added)
2. They may be someone that works on a Ruby app and does not know what lib injection is, or that other team is adding it to their app. So again they may not know what "ddtrace" is.

So this is why I think it's nice to mention Datadog as well. But... not blocking, I'll leave it up to you!

lib-injection/auto_inject.rb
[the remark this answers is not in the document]
ivoanjo: I don't think it's blocking, but I would suggest dropping them a link to this text after merging it anyway -- and if they have any feedback, we can incorporate it :)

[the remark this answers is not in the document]
brett0000FF: @TonyCTHsu pinged me to take a look at this log message. Here are two options:

`[ddtrace] Injection succeeded: ddtrace has been added to the application.`

or

`[ddtrace] Successfully injected ddtrace into the application.`

I think either of these are clear and informative. The first option aligns better with the structure of our failure log messages, but I don't think we can go wrong with either. 👍

PR #3120, opened 2023-09-08
Conti/configure test agent locally

wconti27:
**What does this PR do?**
As discussed in the meeting, we want to configure the local testing environment to be the same as remote CI. To make these changes, tests are now pointing towards the APM Test Agent. Additionally, we are utilizing proxying capabilities present within the APM Test Agent to proxy all received requests to a real Datadog Agent. Responses from the real Datadog Agent are returned by the APM Test Agent to the Ruby test suite.

**Motivation:**
Converge remote and local testing environments to be the same

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**


wconti27:
@TonyCTHsu @marcotc Should we also be proxying requests to the Datadog Agent from the APM Test Agent during remote CI testing?

ivoanjo:
Hey @wconti27, I've been a bit aside from this work, and I'm curious, what's the advantage/difference of using the test agent for these kinds of tests?

TonyCTHsu:
I was able to ran the test successfully on my local development environment, but I need a bit more time to check the code if things could be even simpler

ivoanjo:
> Hey @wconti27, I've been a bit aside from this work, and I'm curious, what's the advantage/difference of using the test agent for these kinds of tests?

Update: We discussed this over chat, thanks for the context!

TonyCTHsu:
👋 @wconti27 , our code base should not contain logic about whether it is a real datadog agent or test agent.

Changes for `FauxWriter` at `spec/support/faux_writer.rb` 

* Refactor the `initialize` method to remove referencing `disable_test_agent`. Consider injecting the transport from caller, instead of checking the keys

---

❌  This is implicit 
```ruby
call_web_mock_function_with_agent_host_exclusions { |options| WebMock.enable! options }
```

✅ This is expressive
```ruby
WebMock.enable!(allow: agent_url)
```

Changes for `NetworkHelpers` need to made: 

* Replace method `call_web_mock_function_with_agent_host_exclusions` with method `agent_url` so the caller could be more explicit about the operation with  
* Change constant declaration `TEST_AGENT_HOST` and `TEST_AGENT_PORT` to methods
* Find a better home for method `parse_tracer_config`, since it does not make sense to be within the `NetworkHelpers` module

wconti27:
@TonyCTHsu Ready for another review, all your comments have been addressed

spec/support/faux_writer.rb
```diff
@@ -3,22 +3,11 @@
...
 require 'support/faux_transport'
-require 'support/network_helpers'
 
 # FauxWriter is a dummy writer that buffers spans locally.
 class FauxWriter < Datadog::Tracing::Writer
-  include NetworkHelpers
-
   def initialize(options = {})
-    if ENV['DD_AGENT_HOST'] == 'testagent' && !options[:disable_test_agent] && test_agent_running?
-      options[:transport] ||= Datadog::Transport::HTTP.default do |t|
```
TonyCTHsu, line 14: Do we ever need this piece of functionality?

spec/support/network_helpers.rb
```diff
@@ -21,44 +18,24 @@ def available_port
...
-    response.is_a?(Net::HTTPSuccess)
-  rescue SocketError
-    false
+  # @return [Integer] agent port
+  def agent_port
+    ENV['DD_TRACE_AGENT_PORT']
   end
 
-  # Gets the Datadog Trace Configuration and returns a comma separated string of key/value pairs.
+  # Yields the agent url to use for testing
```
TonyCTHsu, line 35:
```suggestion
  # Returns the agent url to use for testing
```

TonyCTHsu (commented):
I think we could remove any reference to `disable_test_agent` in our codebase.

spec/support/faux_writer.rb
wconti27: No

marcotc (approved):
Thank you, @wconti27!

PR #3127, opened 2023-09-12
peer.service adjustment for sql propagation with DBM

zarirhamza:
This PR addresses the issue of DBM relying on `service_name` for their own service names prior to `peer.service`. Now since all services are flattened, the DBM map is inaccurate and a better tag to create a more accurate map is to use `peer.service`

### DIFF
--- a/lib/datadog/tracing/contrib/mysql2/instrumentation.rb
+++ b/lib/datadog/tracing/contrib/mysql2/instrumentation.rb
@@ -53,12 +53,18 @@ module Datadog
                 span.set_tag(Tracing::Metadata::Ext::NET::TAG_TARGET_HOST, query_options[:host])
                 span.set_tag(Tracing::Metadata::Ext::NET::TAG_TARGET_PORT, query_options[:port])
 
+                Contrib::SpanAttributeSchema.set_peer_service!(span, Ext::PEER_SERVICE_SOURCES)
+
                 propagation_mode = Contrib::Propagation::SqlComment::Mode.new(comment_propagation)
 
                 Contrib::Propagation::SqlComment.annotate!(span, propagation_mode)
-                sql = Contrib::Propagation::SqlComment.prepend_comment(sql, span, trace_op, propagation_mode)
-
-                Contrib::SpanAttributeSchema.set_peer_service!(span, Ext::PEER_SERVICE_SOURCES)
+                sql = Contrib::Propagation::SqlComment.prepend_comment(
+                  sql,
+                  span,
+                  trace_op,
+                  propagation_mode,
+                  span.get_tag('peer.service')
+                )
 
                 super(sql, options)
               end
--- a/lib/datadog/tracing/contrib/mysql2/instrumentation.rb
+++ b/lib/datadog/tracing/contrib/mysql2/instrumentation.rb
@@ -62,8 +62,7 @@ module Datadog
                   sql,
                   span,
                   trace_op,
-                  propagation_mode,
-                  span.get_tag('peer.service')
+                  propagation_mode
                 )
 
                 super(sql, options)
--- a/lib/datadog/tracing/contrib/propagation/sql_comment.rb
+++ b/lib/datadog/tracing/contrib/propagation/sql_comment.rb
@@ -21,7 +21,7 @@ module Datadog
             return sql unless mode.enabled?
 
             tags = {
-              Ext::KEY_DATABASE_SERVICE => span_op.get_tag('peer.service') || span_op.service,
+              Ext::KEY_DATABASE_SERVICE => span_op.get_tag(Tracing::Metadata::Ext::TAG_PEER_SERVICE) || span_op.service,
               Ext::KEY_ENVIRONMENT => datadog_configuration.env,
               Ext::KEY_PARENT_SERVICE => datadog_configuration.service,
               Ext::KEY_VERSION => datadog_configuration.version
--- a/lib/datadog/tracing/contrib/propagation/sql_comment.rb
+++ b/lib/datadog/tracing/contrib/propagation/sql_comment.rb
@@ -17,11 +17,11 @@ module Datadog
 
           # Inject span_op and trace_op instead of TraceDigest to improve memory usage
           # for `disabled` and `service` mode
-          def self.prepend_comment(sql, span_op, trace_op, mode)
+          def self.prepend_comment(sql, span_op, trace_op, mode, peer_service)
             return sql unless mode.enabled?
 
             tags = {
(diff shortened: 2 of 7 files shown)


lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -17,11 +17,11 @@ def self.annotate!(span_op, mode)
 
           # Inject span_op and trace_op instead of TraceDigest to improve memory usage
           # for `disabled` and `service` mode
-          def self.prepend_comment(sql, span_op, trace_op, mode)
+          def self.prepend_comment(sql, span_op, trace_op, mode, peer_service)
```
TonyCTHsu, line 20: The span has already be injected into the method. No need to  add another parameter

TonyCTHsu (changes requested):

lib/datadog/tracing/contrib/propagation/sql_comment.rb
zarirhamza: Good catch, done

spec/datadog/tracing/contrib/sql_comment_propagation_examples.rb
```diff
@@ -1,8 +1,12 @@
-RSpec.shared_examples_for 'with sql comment propagation' do |span_op_name:, error: nil|
+RSpec.shared_examples_for 'with sql comment propagation' do |span_op_name:, peer_service:, error: nil|
```
TonyCTHsu, line 1: What is the reason for adding `peer_service` ?

lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -21,7 +21,7 @@ def self.prepend_comment(sql, span_op, trace_op, mode)
             return sql unless mode.enabled?
 
             tags = {
-              Ext::KEY_DATABASE_SERVICE => span_op.service,
+              Ext::KEY_DATABASE_SERVICE => span_op.get_tag('peer.service') || span_op.service,
```
marcotc, line 24: Is `'peer.service'` already a constant somewhere, if so we we should reference that constant instead of using an inline string.

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
```diff
@@ -79,6 +78,7 @@
         let(:mode) { 'service' }
 
         it do
+          allow(span_op).to receive(:get_tag).with('peer.service').and_return(nil)
```
TonyCTHsu, line 81: The test was simple enough so we could use a mock object. Since it is becoming more complex, replace `span_op` from mock object with a real object like [this](https://github.com/DataDog/dd-trace-rb/pull/3127/files#diff-8ac548c14d9fa2993b3cdc2768cf4ff4dc4be995602c7daf448b03dee7477d76R8)

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
```diff
@@ -102,6 +103,63 @@
       end
     end
 
+    context 'when tracing is enabled with peer.service' do
```
TonyCTHsu, line 106: The description does not match the scenario, it only need a span_op  with a specific tag

lib/datadog/tracing/contrib/mysql2/instrumentation.rb
```diff
@@ -53,12 +53,17 @@ def query(sql, options = {})
                 span.set_tag(Tracing::Metadata::Ext::NET::TAG_TARGET_HOST, query_options[:host])
                 span.set_tag(Tracing::Metadata::Ext::NET::TAG_TARGET_PORT, query_options[:port])
 
+                Contrib::SpanAttributeSchema.set_peer_service!(span, Ext::PEER_SERVICE_SOURCES)
```
TonyCTHsu, line 56: Do we have test for this?

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
zarirhamza: I disagree, the main thing here is to mock a situation when tracing is enabled with `peer.service`, ensuring that the `peer.service` value is getting set as the `dddbs` tag.

> it only need a span_op with a specific tag

That's what the peer.service feature does. It adds a tag

lib/datadog/tracing/contrib/mysql2/instrumentation.rb
zarirhamza: Several, peer.service is a completely independent feature already thoroughly fleshed out. 

We test the function [here](https://github.com/DataDog/dd-trace-rb/blob/master/spec/datadog/tracing/contrib/span_attribute_schema_spec.rb)

We test specifically `mysql2` [here](https://github.com/DataDog/dd-trace-rb/blob/115d4a8c30492e6c9e67ddd2b0620e7cc8747cf3/spec/datadog/tracing/contrib/mysql2/patcher_spec.rb#L102) and we also test more [here](https://github.com/DataDog/dd-trace-rb/blob/115d4a8c30492e6c9e67ddd2b0620e7cc8747cf3/spec/datadog/tracing/contrib/mysql2/patcher_spec.rb#L76) and several places so no worries about that

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
TonyCTHsu: Sorry, I meant that the test could be simplified, for example, 

We don't need 
```
        Datadog.configure do |c|
          c.env = 'production'
          c.service = "Traders' Joe"
          c.version = '1.0.0'
        end
```

lib/datadog/tracing/contrib/mysql2/instrumentation.rb
TonyCTHsu: If we already have the `peer.service` tag, why do we need to add this line? Isn't it already in place?

zarirhamza: It's not being added, it's just being moved to calculate `peer.service` before we do propagation so a value exists within the span to check. Previously we made the last step of every span to calculate `peer.service` but in this case, sql propagation requires `peer.service` to be calculated first and not vice versa

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
TonyCTHsu: The description should be 
```
context 'when given a span operation tagged with`peer.service`' do
```

TonyCTHsu (approved):

PR #3130, opened 2023-09-14
Extract db appraisal group

TonyCTHsu:
**Motivation:**

From internal discussion, we want to horizontally scale our appraisal groups, instead of vertically. This means we are breaking large appraisal groups into smaller ones.

**What does this PR do?**

This PR extracts `relational_db` group from `contrib` group. This group includes a range of gems that are depending on a relational database. 

**Additional Notes**

I started off with extracting `activesupport` group, but found there are strong coupling in our test with our db adatpers(`mysql2`, `pg` and `sqlite`) while `activerecord` does not declare dependencies for them. It would be strange to include those adapters into the `activesupport` group. Hence, I pivot to extract those relational database related gems. I hesitate to include other db related gems(`mongodb`, `presto`/`trino`, `dalli`) at this moment. 

The original `bundle exec rake test:autoinstrument` is failing because the test cases were setup with a Sinatra application making db request but now the appraisal group `contrib` no longer contains those db gems. I consider making db request to be excessive for test case to verify the auto instrument behaviour, so I fix the task by removing the those  db calls and change the dependent group to `sinatra` and increase the ruby version support to latest version.


GustavoCaso:
> I hesitate to include other db related gems(mongodb, presto/trino, dalli) at this moment.

Can you explain why you hesitated?

TonyCTHsu:
> > I hesitate to include other db related gems(mongodb, presto/trino, dalli) at this moment.
> 
> Can you explain why you hesitated?

The reason I group those gems because they require database adapter (`mysql2`, `pg` and `sqlite`) to work properly. With such group, I can test the instrumentation for high level ORM (`active_record`, `sequel` and `delayed_job` ) and instrumentation for low level drivers  (`mysql2`, `pg` and `sqlite`). 

The ones excluded (`mongodb`, `presto/trino`, `dalli`) do not required those drivers or they contain their own drivers.

GustavoCaso (approved):

PR #3131, opened 2023-09-14
[APPSEC-10967] ASM API security. Schema extraction

GustavoCaso:
**What does this PR do?**

This PR adds the necessary bits to enable API Security's new feature. Schema Extraction 🎉 

To enable that, we need the latest version of libddwaf 1.14.0. At the end request lifecycle, we would call the WAF to extract the schema of the different WAF addresses. Calling schema extraction is gated via specific configuration: `DD_EXPERIMENTAL_API_SECURITY_ENABLED` and `DD_API_SECURITY_REQUEST_SAMPLE_RATE`

The schema extraction information is appended to the top-level span attributes so the backend can later process it.

At the moment, the processor configuration is not part of the static WAF rules, so we need to make sure we always include it as part of the Handle rules. We store the processor configuration at the same pace as the WAF rule. 

To better review the changes, I split the work into separate commits. I recommend reading individual commits to make it easier to understand the changes. Another suggestion is to filter out all the `.lock` file changes. Here is a handy [link](https://github.com/DataDog/dd-trace-rb/pull/3131/files?file-filters%5B%5D=.gemspec&file-filters%5B%5D=.json&file-filters%5B%5D=.rb&file-filters%5B%5D=.rbs&show-viewed-files=true) to filter out those files 

**Motivation:**

**Additional Notes:**

There are a few missing features of the API Secuirty RFC that would be implemented on the following PRs
- Compress schema information
- Add support for the new WAF address `server.response.body`

**How to test the change?**

**For Datadog employees:**

### DIFF
--- a/ddtrace.gemspec
+++ b/ddtrace.gemspec
@@ -62,7 +62,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'debase-ruby_core_source', '= 3.2.1'
 
   # Used by appsec
-  spec.add_dependency 'libddwaf', '~> 1.11.0.0.0'
+  spec.add_dependency 'libddwaf', '~> 1.14.0.0.0'
 
   # Used by profiling (and possibly others in the future)
   # When updating the version here, please also update the version in `native_extension_helpers.rb` (and yes we have a test for it)
--- a/lib/datadog/appsec/assets.rb
+++ b/lib/datadog/appsec/assets.rb
@@ -10,6 +10,10 @@ module Datadog
         read("waf_rules/#{kind}.json")
       end
 
+      def waf_processors
+        read('waf_rules/processors.json')
+      end
+
       def blocked(format: :html)
         (@blocked ||= {})[format] ||= read("blocked.#{format}")
       end
--- a/lib/datadog/appsec/contrib/rack/request_middleware.rb
+++ b/lib/datadog/appsec/contrib/rack/request_middleware.rb
@@ -75,6 +75,16 @@ module Datadog
 
             _response_return, response_response = Instrumentation.gateway.push('rack.response', gateway_response)
 
+            result = scope.processor_context.extract_schema
+
+            if result
+              scope.processor_context.events << {
+                trace: scope.trace,
+                span: scope.service_entry_span,
+                waf_result: result,
+              }
+            end
+
             scope.processor_context.events.each do |e|
               e[:response] ||= gateway_response
               e[:request]  ||= gateway_request
--- a/lib/datadog/appsec/event.rb
+++ b/lib/datadog/appsec/event.rb
@@ -62,6 +62,7 @@ module Datadog
 
           # prepare and gather tags to apply
           service_entry_tags = build_service_entry_tags(event_group)
+
           # complex types are unsupported, we need to serialize to a string
           triggers = service_entry_tags.delete('_dd.appsec.triggers')
           span.set_tag('_dd.appsec.json', JSON.dump({ triggers: triggers }))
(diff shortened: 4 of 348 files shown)


lib/datadog/appsec/processor/rule_merger.rb
```diff
@@ -16,8 +18,13 @@ def initialize(version1, version2)
           end
         end
 
+        DEFAULT_PROCESSORS = JSON.parse(Datadog::AppSec::Assets.waf_processors)
```
lloeki, line 21: I'd call that `DEFAULT_WAF_PROCESSORS` to materialize the distinction with `AppSec::Processor`

lib/datadog/appsec/assets.rb
```diff
@@ -10,6 +10,10 @@ def waf_rules(kind = :recommended)
         read("waf_rules/#{kind}.json")
       end
 
+      def waf_processors
+        read('waf_rules/processors.json')
```
lloeki, line 14: Ha our `Appsec::Processor` naming is a bit unfortunate. Let's make sure there's no confusion in our code.

lib/datadog/appsec/processor/rule_merger.rb
lloeki: I would also refrain from reading+parsing that file + setting that constant at file load time. I'd rather have that be a memoized method.

ddtrace.gemspec
```diff
@@ -62,7 +62,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'debase-ruby_core_source', '= 3.2.1'
 
   # Used by appsec
-  spec.add_dependency 'libddwaf', '~> 1.11.0.0.0'
+  spec.add_dependency 'libddwaf', '~> 1.14.0.0.0'
```
lloeki, line 65: Make sure to update `libddwaf` signatures in `vendor/rbs`

lib/datadog/appsec/contrib/rack/request_middleware.rb
```diff
@@ -75,6 +75,16 @@ def call(env)
...
 
+            result = scope.processor_context.extract_schema
+
+            if result
+              scope.processor_context.events << {
+                trace: scope.trace,
+                span: scope.service_entry_span,
+                waf_result: result,
+              }
+            end
```
lloeki, line 86: I'm confused by this, the extracted schema is a waf result, and stored in the events list?

lib/datadog/appsec/event.rb
```diff
@@ -104,8 +105,15 @@ def self.build_service_entry_tags(event_group)
...
 
           # accumulate triggers
+          waf_result = event[:waf_result]
           tags['_dd.appsec.triggers'] ||= []
-          tags['_dd.appsec.triggers'] += event[:waf_result].events
+          tags['_dd.appsec.triggers'] += waf_result.events
+
+          waf_result.derivatives.each do |key, value|
+            tags[key] = JSON.dump(value)
+          end
```
lloeki, line 114: Hmm I think I get it now, but coming back to this after a long time I feel some confusion.

lib/datadog/appsec/processor.rb
```diff
@@ -38,9 +36,30 @@ def run(input, timeout = WAF::LibDDWAF::DDWAF_RUN_TIMEOUT)
...
+        def extract_schema
+          return unless extract_schema?
+
+          input = {
+            'waf.context.processor' => {
+              'extract-schema' => true
+            }
+          }
+
+          _code, res = @context.run(input, WAF::LibDDWAF::DDWAF_RUN_TIMEOUT)
```
lloeki, line 48: This `libddwaf` API looks _very_ strange, which might be the root cause of my confusion.

lloeki (approved):
LGTM, a few notes but nothing consequential or worth a block.

PR #3132, opened 2023-09-14
Support Rack 3 by adding to test matrix

TonyCTHsu:
**Motivation:**

Support Rack 3 

**What does this PR do?**

1. Rack is tested against version 1, 2, 3
2. Skip Rack 1.x with Ruby 3.3, because non-deprecated form of Regexp.new does not backport to Rack 1.x. 
https://github.com/rack/rack/pull/1998

**Additional notes**

Appsec test requires `gem 'rack-contrib'` in the group.


lib/datadog/tracing/contrib/rack/header_tagging.rb
```diff
@@ -37,10 +37,16 @@ def self.tag_response_headers(span, headers, configuration)
...
-                       unless header_value.nil?
-                         header_tag = Tracing::Metadata::Ext::HTTP::ResponseHeaders.to_tag(header)
-                         result[header_tag] = header_value
-                       end
+
+                       next if header_value.nil?
+
+                       header_tag = Tracing::Metadata::Ext::HTTP::ResponseHeaders.to_tag(header)
+
+                       result[header_tag] = if header_value.is_a? Array
```
marcotc, line 45: We should document that this is done to maintain the value format between Rack 2 and 3.

marcotc (approved):

PR #3143, opened 2023-09-19
Update CODEOWNERS file

GustavoCaso:
**What does this PR do?**

Update the CODEOWNER files to assign teams better when specific files are modified. 

We still have a catch-all `* @DataDog/ruby-guild`, meaning every team would get a ping for files not specified on the other rules.

Of course we can tweak and improve those rules if we want 

**Motivation:**

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**


.github/CODEOWNERS
```diff
@@ -1,5 +1,30 @@
...
+lib/datadog/tracing/* @DataDog/apm-ruby
+lib/datadog/tracing.rb @DataDog/apm-ruby
+lib/datadog/opentelemetry/* @DataDog/apm-ruby
+lib/datadog/opentelemetry.rb @DataDog/apm-ruby
+lib/datadog/opentracer/* @DataDog/apm-ruby
+lib/datadog/opentracer.rb @DataDog/apm-ruby
+lib-injection/* @DataDog/apm-ruby
+lib/datadog/profiling/* @DataDog/profiling-rb
+lib/datadog/profiling.rb @DataDog/profiling-rb
+ext/* @DataDog/profiling-rb
```
ekump, line 19: This question is probably outside the scope of this specific PR, but should we rename `ext/` since it appears to be a profiling only folder?

ekump (approved):

.github/CODEOWNERS
marcotc: We can't as `ext/` is a standard for Ruby gems with native extensions: https://guides.rubygems.org/gems-with-extensions/

.github/CODEOWNERS
```diff
@@ -1,5 +1,30 @@
...
 /docs/Compatibility.md @DataDog/documentation
-/docs/GettingStarted.md @DataDog/documentation
\ No newline at end of file
+/docs/GettingStarted.md @DataDog/documentation
+
+# Library
+lib/datadog/appsec/* @DataDog/asm-ruby
+lib/datadog/appsec.rb @DataDog/asm-ruby
+lib/datadog/tracing/* @DataDog/apm-ruby
+lib/datadog/tracing.rb @DataDog/apm-ruby
```
marcotc, line 11: I just created the https://github.com/orgs/DataDog/teams/tracing-ruby team, we can update to use this new team in this file.

lloeki: Updated.

.github/CODEOWNERS
lloeki: My rationale here is:

- `ext` being a generic folder for extensions, it should be owned by the guild.
- `ext/ddtrace_profiling_loader` and `ext/ddtrace_profiling_native_extension` should be owned by `profiling-rb`

lloeki: That said, currently profiling is the sole user of `libdatadog`, but that may/should change, therefore I would like to encourage ownership to be split:

- extension code pertaining to profiling should live in its own profiling-owned subdirectories
- extension code pertaining to libdatadog in general should live in guild-owned subdirectories

.github/CODEOWNERS
```diff
@@ -1,5 +1,30 @@
...
+# Library
+lib/datadog/appsec/* @DataDog/asm-ruby
+lib/datadog/appsec.rb @DataDog/asm-ruby
+lib/datadog/tracing/* @DataDog/tracing-ruby
+lib/datadog/tracing.rb @DataDog/tracing-ruby
+lib/datadog/opentelemetry/* @DataDog/tracing-ruby
+lib/datadog/opentelemetry.rb @DataDog/tracing-ruby
+lib/datadog/opentracer/* @DataDog/tracing-ruby
+lib/datadog/opentracer.rb @DataDog/tracing-ruby
+lib-injection/* @DataDog/tracing-ruby
```
lloeki, line 16: I am questioning this a bit: while initially developed by tracing automatic injection sounds like a feature that applies to everyone, not just tracing.

.github/CODEOWNERS
```diff
@@ -1,5 +1,30 @@
-* @DataDog/apm-ruby
+* @DataDog/ruby-guild
 
 # Public Documentation
 /docs/Compatibility.md @DataDog/documentation
-/docs/GettingStarted.md @DataDog/documentation
\ No newline at end of file
+/docs/GettingStarted.md @DataDog/documentation
+
+# Library
```
lloeki, line 7: Looks like this file is missing `sig` ownership

lloeki: Added!

marcotc (approved):

PR #3147, opened 2023-09-20
ekump/upgrade to rubocop 1.50.0

ekump:
**What does this PR do?**
Upgrades Rubocop from 1.34.0 to 1.50.0

**Motivation:**
While working on another branch, I encountered a bug where attempting to disable `Lint/MissingSuper` led to a `Lint/RedundantCopDisableDirective` error that could not be suppressed. 

**Additional Notes:**
Uncovered a test in `profiling/integration_spec.rb` that may have been running as expected.

**How to test the change?**
`bundle exec rake rubocop`

**For Datadog employees:**


lib/datadog/tracing/contrib/utils/quantization/http.rb
```diff
@@ -160,11 +160,11 @@ def obfuscate_query(query, options = {})
                  (?:"|%22)                                     # closing '"' at end of value
               )
              |(?: # other common secret values
-                 bearer(?:\s|%20)+[a-z0-9._\-]+
+                 bearer(?:\s|%20)+[a-z0-9._-]+
```
ekump, line 163: The regex changes are just removing redundant escape characters

spec/datadog/profiling/integration_spec.rb
```diff
@@ -255,9 +255,8 @@ def stack_frame_to_function_id(backtrace_location)
           is_expected.to have(4).items
 
           # All but last are unique
-          (0..-2).each do |i|
+          3.times do |i|
```
ekump, line 258: @ivoanjo wanted to bring this to your attention. `(0..-2).each` never executes.

spec/support/platform_helpers.rb
```diff
@@ -6,15 +6,15 @@ module PlatformHelpers
   # Ruby runtime engines
 
   def mri?
-    RUBY_ENGINE == 'ruby'.freeze
+    'ruby'.freeze == RUBY_ENGINE
```
marcotc, line 9: I think this is a bug of the https://www.rubydoc.info/github/bbatsov/RuboCop/RuboCop/Cop/Style/YodaCondition cop.
And I believe it happens because it doesn't know that `#freeze` is not really a complex method call, and `'ruby'.freeze` is still an immediate value.

marcotc: For this file, can I suggest adding `# frozen_string_literal: true` to the top, and reverting the YodaCondition changes?

marcotc (approved):
Minor nit.

spec/datadog/profiling/integration_spec.rb
ivoanjo: Oh wow, thanks for spotting it! The good news is -- this test is only for the old profiler codepath, that is off by default and no customers should be using anymore.

The new profiler has similar specs, but in different files, and doesn't have this bug as far as I can spot it.

(The old profiler, along with this file are going to be deleted in Q4; I've mostly been finishing up a few other features before I come and remove all of this.)

spec/support/platform_helpers.rb
ekump: sure!

PR #3149, opened 2023-09-20
Action:Update gemfiles when modified

marcotc:
*This PR is ready for review, but I'd like to merge it only after 1.15.0, in case it somehow gets in the way of the release process (all the branching and tagging that happens during a release).*

**What does this PR do?**

Ensures that `appraisal/*` and `gemfiles/*` are always up to date with changes to `Gemfile`, `Appraisals`, and `ddtrace.gemspec`.

This means it's no longer possible to forget to generate gemfiles when modifying dependencies.

Also, the rather long process of updating gemfiles (> 20 minutes) is now done in a GitHub Action instead of in the developer's computer.

**Motivation:**

Updating gemfiles when modifying dependencies is a very slow operation. So slow that we sometimes, for the sake of time, do not update them at all in our PRs.

Slow but completely mechanical operations are better done by machines. This operation is one of such cases.

This also opens the door for automating `Gemfile.lock` bumps (`bundle update`), to automatically test against the latest versions of integrated gems: it should be possible to compose `Gemfile.lock` bumps and PR creation.

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**


docker-compose.yml
```diff
@@ -33,7 +33,7 @@ services:
       TEST_REDIS_HOST: redis
       TEST_REDIS_OLD_HOST: redis_old
       TEST_REDIS_OLD_PORT: 6379
-      DDTRACE_CI: true
+      DDTRACE_CI: 'true'
```
marcotc, line 36: GitHub Actions's Docker runner correctly complained that environment variables should not be native boolean values. For some reason, the local docker runners for MacOS does not complain about this.

.github/workflows/update-gemfiles.yml
```diff
@@ -0,0 +1,63 @@
...
+        name: Update gemfiles/*.gemfile
+        run: bundle exec rake appraisal:generate
+      - if: steps.pr.outputs.pr_found == 'true'
+        name: Update gemfiles/*.gemfile.lock
+        run: bundle exec rake appraisal:lock
+      - if: steps.pr.outputs.pr_found == 'true'
+        name: Commit gemfiles changes, if any, back to the branch
+        uses: stefanzweifel/git-auto-commit-action@3ea6ae190baf489ba007f7c92608f33ce20ef04a # v4.16.0
+        with:
+          commit_message: Update gemfiles/*
```
ekump, line 63: I'm not super familiar with git workflows, so maybe I'm just missing it, but is this workflow required to complete before being able to merge to master? I think that's reasonably important, so we don't wind up commiting gemfile changes to a branch that's already been merged, and thus it never makes it to master.

ekump (approved):

.github/workflows/update-gemfiles.yml
marcotc: That's a good point. To enable a mandatory GitHub check the job should be able to always succeed.
The way it is currently implemented, it won't run on PRs that don't change gem files, thus the check won't pass.

I'll change it so that it runs (and passes) if no changes to gems have taken place.

marcotc: Fixed it, thank you!

marcotc:
Reminder: Add ` Update Gemfiles` as a required status check after merging this PR.

TonyCTHsu:
👋 @marcotc Thanks for the great effort. 

I was trying out this with another branch. https://github.com/DataDog/dd-trace-rb/pull/3200. I found something worth mentioning. 

1. When the commit contains changes in `appraisal/*.rb` for each ruby version  => the GHA was not triggered, I have to manually make additional changes (unnecessary changes to `Gemfile`) to trigger it. 

2. We can squash these two into a single step (we don't need to run `bundle exec rake appraisal:generate` ), since the lock task already contains the gemfile generation. 

```
       - if: steps.pr.outputs.pr_found == 'true'
        name: Update gemfiles/*.gemfile
        run: bundle exec rake appraisal:generate
      - if: steps.pr.outputs.pr_found == 'true'
        name: Update gemfiles/*.gemfile.lock
        run: bundle exec rake appraisal:lock
```

3. Optional, currently the autogenerated file would only contains x86 platform because we are using  x86 machines to run the task. We have a task to add platform `bundle exec rake appraisal:platform`, however, the task has less flexibility. We can add the platform part later

ivoanjo:
With 1.15.0 out, is it time to merge this one?

marcotc:
@TonyCTHsu I addressed all 3 issues, thank you so much for testing this!

PR #3150, opened 2023-09-21
ekump/decouple core transport

ekump:
**What does this PR do?**
Properly namespace the various transport modules to aid in decoupling Tracing from Core. **There should be no functional changes.**

`Datadog::Transport` (ddtrace/transport) was primarily used by Tracing so it has been moved to `Datadog::Tracing::Transport`. The exception to this is the ext file which is a public API. We should discuss moving this in 2.0 (should be relatively easy). 

`Datadog::Core::Transport` was only used by `Datadog::Core::Remote`. It has been moved to `Datadog::Core::Remote::Transport`.

Files common to both Remote and Tracing were moved to `Datadog::Core::Transport`.

**Motivation:**
Decouple Tracing from Core

**Additional Notes:**
The scope of this PR is primarily decoupling. There is probably a fair amount of work that can be done to simplify the transport code in general, but that can be done separately. 

**How to test the change?**

**For Datadog employees:**


.rubocop_todo.yml
```diff
@@ -32,12 +32,6 @@ Lint/MissingSuper:
     - 'lib/datadog/profiling/pprof/converter.rb'
```
marcotc, line 32: Nice cleanup!

marcotc (approved):
Awesome work, @ekump!

anmarchenko (approved):
nice, I will migrate to use Net::HTTP adapter from Core for agentless mode as soon as it gets merged

lib/datadog/core/remote/transport/http.rb
```diff
@@ -0,0 +1,179 @@
...
+require_relative '../../transport/http/adapters/test'
+
+# TODO: Improve negotiation to allow per endpoint selection
+#
+# Since endpoint negotiation happens at the `API::Spec` level there can not be
+# a mix of endpoints at various versions or versionless without describing all
+# the possible combinations as specs. See http/api.
+#
+# Below should be:
+# require_relative '../../transport/http/api'
```
GustavoCaso, line 19: Is this comment relevant?

lib/datadog/core/remote/transport/http.rb
```diff
@@ -0,0 +1,179 @@
...
+require_relative 'http/api'
+
+# TODO: Decouple transport/http/builder
+#
+# See http/builder
+#
+# Below should be:
+# require_relative '../../transport/http/builder'
```
GustavoCaso, line 27: Is this comment relevant?

lib/datadog/core/remote/transport/http/api.rb
```diff
@@ -0,0 +1,58 @@
...
+#
+# Separate classes are needed because transport/http/traces includes
+# Trace::API::Instance which closes over and uses a single spec, which is
+# negotiated as either /v3 or /v4 for the whole API at the spec level, but we
+# need an independent toplevel path at the endpoint level.
+#
+# Separate classes are needed because of `include Trace::API::Instance`.
+#
+# Below should be:
+# require_relative '../../../../ddtrace/transport/http/api/spec'
```
GustavoCaso, line 16: Is this comment relevant?  And any other comment that state what the actual require should look like 😄

GustavoCaso:
Amazing work @ekump 🎉 

I wondered if you explored avoiding the duplicated portions of `core/transport` and `tracing/transport` and trying to work with what we currently have to not have the code duplicated? Or if we plan on working on subsequent PR to have a more flexible core transport that remote, tracing, and CI can use?

lib/datadog/core/remote/transport/http/api.rb
ekump: These comments were [already in place](https://github.com/DataDog/dd-trace-rb/blob/ba9fc95f49ec48fd28ca69ecec1a16a0024d3252/lib/datadog/core/transport/http/api.rb#L17). I just updated them to reflect this PR's changes. If we think they're superfluous, I'm happy to remove them.

ekump:
> Amazing work @ekump 🎉
> 
> I wondered if you explored avoiding the duplicated portions of `core/transport` and `tracing/transport` and trying to work with what we currently have to not have the code duplicated? Or if we plan on working on subsequent PR to have a more flexible core transport that remote, tracing, and CI can use?

@GustavoCaso - I think there is a fair amount of interest from everyone to a) simplify transport and b) do what we can to reduce the duplicate code between remote and tracing. But, the scope of this PR was just decoupling. I'm hoping we'll be able to revisit the transport code sooner, rather than later to give it more attention.

PR #3153, opened 2023-09-25
[APPSEC-10967] ASM parse response body 

GustavoCaso:
**What does this PR do?**

Parsed the response body and passed it to the waf as `server.response.body`.

We only parse bodies that are either an `Array` or a `Rack::BodyProxy`. Since we might not be the last middleware in the customer application, we can not consume the response body directly by calling `each`. To circumvent that, we make a copy of the body. 

Parsing the response body could lead to performance implications for our customers. Since we have yet to learn how this would impact our customers, I added a configuration entry for them to skip the response body parsing altogether. 

This documentation would remain undocumented, and we would only mention it to customers if they experience any performance degradation.

**Motivation:**

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**

### DIFF
--- a/lib/datadog/appsec/contrib/rack/gateway/response.rb
+++ b/lib/datadog/appsec/contrib/rack/gateway/response.rb
@@ -20,7 +20,7 @@ module Datadog
             end
 
             def parsed_body
-              return unless body.instance_of?(Array)
+              return unless body.instance_of?(Array) || body.instance_of?(::Rack::BodyProxy)
               return unless supported_response_type
 
               body_dup = body.dup # avoid interating over the body. This is just in case code.
--- a/lib/datadog/appsec/contrib/rack/gateway/response.rb
+++ b/lib/datadog/appsec/contrib/rack/gateway/response.rb
@@ -20,7 +20,16 @@ module Datadog
             end
 
             def parsed_body
-              return unless body.instance_of?(Array) || body.instance_of?(::Rack::BodyProxy)
+              return unless Datadog.configuration.appsec.parse_response_body
+
+              unless body.instance_of?(Array) || body.instance_of?(::Rack::BodyProxy)
+                if result.timeout
+                  Datadog.logger.debug do
+                    "Unable to parse response body because of unsupported body type: #{body.class}"
+                  end
+                end
+                return
+              end
               return unless supported_response_type
 
               body_dup = body.dup # avoid interating over the body. This is just in case code.
--- a/lib/datadog/appsec/contrib/rack/gateway/response.rb
+++ b/lib/datadog/appsec/contrib/rack/gateway/response.rb
@@ -22,19 +22,19 @@ module Datadog
             def parsed_body
               return unless Datadog.configuration.appsec.parse_response_body
 
-              unless body.instance_of?(Array) || body.instance_of?(::Rack::BodyProxy)
+              unless body.instance_of?(Array)
                 Datadog.logger.debug do
                   "Response body type unsupported: #{body.class}"
                 end
                 return
               end
+
               return unless json_content_type?
 
-              body_dup = body.dup # avoid interating over the body. This is just in case code.
               result = ''.dup
               all_body_parts_are_string = true
 
-              body_dup.each do |body_part|
+              body.each do |body_part|
                 if body_part.is_a?(String)
                   result.concat(body_part)
                 else
@@ -65,7 +65,8 @@ module Datadog
             ].freeze
 
             def json_content_type?
-              VALID_JSON_TYPES.include?(headers['content-type'])
+              content_type = headers['content-type']
+              VALID_JSON_TYPES.any? { |valid_type| content_type.include?(valid_type) }
             end
           end
         end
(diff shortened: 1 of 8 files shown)


lib/datadog/appsec/contrib/rack/gateway/response.rb
```diff
@@ -19,9 +19,58 @@ def initialize(body, status, headers, scope:)
...
 
+            def parsed_body
+              return unless Datadog.configuration.appsec.parse_response_body
+
+              unless body.instance_of?(Array) || body.instance_of?(::Rack::BodyProxy)
+                if result.timeout
+                  Datadog.logger.debug do
+                    "Unable to parse response body because of unsupported body type: #{body.class}"
+                  end
+                end
```
lloeki, line 30: Hmm, that doesn't seem like the intended log message there.

lib/datadog/appsec/contrib/rack/gateway/response.rb
```diff
@@ -19,9 +19,58 @@ def initialize(body, status, headers, scope:)
...
+                return
+              end
+              return unless supported_response_type
+
+              body_dup = body.dup # avoid interating over the body. This is just in case code.
```
lloeki, line 35: For `Array` this is not needed, so is that for `Rack::BodyProxy`?

lib/datadog/appsec/contrib/rack/gateway/response.rb
```diff
@@ -19,9 +19,58 @@ def initialize(body, status, headers, scope:)
...
             end
+
+            private
+
+            def supported_response_type
+              json? || text?
+            end
+
+            def json?
+              headers['content-type'].include?('json')
```
lloeki, line 68: I'd rather be strict and explicitly list supported content types. I think those would be:

- `application/json` (the official one [registered at IANA](https://www.iana.org/assignments/media-types/media-types.xhtml))
- `text/json` because someone from the team added it? I don't think I've seen it in the (Ruby) wild, ever.

lib/datadog/appsec/contrib/rack/gateway/response.rb
```diff
@@ -19,9 +19,58 @@ def initialize(body, status, headers, scope:)
...
+            end
+
+            def text?
+              headers['content-type'].include?('text')
```
lloeki, line 72: I'd rather be strict and explicitly list supported content types.

Indeed as is this would match at least `text/plain` and `text/html` but is there any point of passing those raw to libddwaf in the non-raw address that aims to carry parsed, structured data?

Is it for `text/xml`, which according to [RFC 3023](https://www.rfc-editor.org/rfc/rfc3023) says:

> If an XML document -- that is, the unprocessed, source XML document
   -- is readable by casual users, text/xml is preferable to
   application/xml.  MIME user agents (and web user agents) that do not
   have explicit support for text/xml will treat it as text/plain, for
   example, by displaying the XML MIME entity as plain text.
   Application/xml is preferable when the XML MIME entity is unreadable
   by casual users.

lib/datadog/appsec/contrib/rack/gateway/response.rb
```diff
@@ -19,9 +19,58 @@ def initialize(body, status, headers, scope:)
...
+                else
+                  all_body_parts_are_string = false
+                  break
+                end
+              end
+
+              return unless all_body_parts_are_string
+
+              if json?
+                JSON.parse(result)
```
lloeki, line 51: That `json?` tests that folks have been advertising the body as being JSON, but if content is broken JSON it'd blow up with a `JSON::ParserError`.

We should guard against that and `return`.

lib/datadog/appsec/contrib/rack/gateway/response.rb
```diff
@@ -19,9 +19,58 @@ def initialize(body, status, headers, scope:)
...
+              else
+                result
```
lloeki, line 53: Then it's not really parsed, is it? Either it's raw (`server.response.body.raw`) or it's parsed (`server.response.body`).

Since the goal is to parse it - notably to perform schema extraction - it seems to me there's not much point doing that.

lloeki (approved):
Overall ok, but I'm concerned about the fuzziness around content type conditions.

lib/datadog/appsec/contrib/rack/gateway/response.rb
GustavoCaso: Yes. 

Since `Rack::BodyProxy` acts as a proxy. If we call `each` directly, we might iterate over the body and consume it. 

https://github.com/rack/rack/blob/main/lib/rack/body_proxy.rb#L45-L58

If we call to `to_ary` we also close the body, and we do not want that, since we might not be the last middleware. That is why I thought using `dup` would save us here

lib/datadog/appsec/contrib/rack/gateway/response.rb
GustavoCaso: You are right, since we do not parse, there is no schema information to extract from it 

I will remove the entire support for `text/*` content-type

lib/datadog/appsec/contrib/rack/gateway/response.rb
lloeki: But then you're duping the `Rack::BodyProxy` instance, which would still point to the same underlying wrapped object, which we don't know the type of, and may not be traversable twice.

PR #3154, opened 2023-09-25
Make sure that file name and defined class on the file matches.

GustavoCaso:
By following the Ruby convention that most Ruby developers are used to makes it easy to find the files and classes on our codebase.

The Ruby Convention states:

`hello.rb` -> class or module  `Hello`
`helloworld.rb` -> class or module  `Helloworld`
`hello_world.rb` -> class or module  `HelloWorld`
`hello/world.rb` -> class or module  `Hello::World`

I excluded specific files to avoid having to do a big rename. For example, to have a class name be OpenTelemtry, we would need the filename or directory to be open_telemetry.
Those changes make looking at the git history hard and might not provide much value.

Also, specific names in the industry will not match the ruby convention, such as MongoDB.

I added all the reasonable configurations to this cop and added the list of files that need to change in `.rubocop_todo.yml`. 

That way, this PR only makes sure that no new files that not follow the ruby convention can be added to the repository, as well as having a list of files to slowly change 😄 

**What does this PR do?**

**Motivation:**

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**


ivoanjo (approved):
:+1: LGTM

PR #3155, opened 2023-09-25
Update rubocop for appraisals

TonyCTHsu:
**What does this PR do?**

This PR propagates previous changes for rubocop to appraisal dependencies.

https://github.com/DataDog/dd-trace-rb/pull/3147/files#diff-d09ea66f8227784ff4393d88a19836f321c915ae10031d16c93d67e6283ab55f


ekump (approved):
Sorry about this! I had no idea this needed to be done when updating gems.

ivoanjo:
> Sorry about this! I had no idea this needed to be done when updating gems.

Technically, it shouldn't! It doesn't make a lot of sense, actually -- none of the tests we run depends on rubocop, it's just an artifact of how it's setup currently. (Which would be cool to fix, thus reducing a bunch of noise on the appraisals :D )

PR #3161, opened 2023-09-27
Polling the presence of gem to trigger gitlab job

TonyCTHsu:
**Motivation:**

Building lib injection package depends on the gem ddtrace being published at rubygems.org.

During ddtrace release, the process is triggered by a tag push event from the repository but failed since the gem has not yet been published. It needs to be manually retriggered after the gem is published to rubygems.org successfully.

**What does this PR do?**

Create a job that makes polling request to rubygems.org and blocks to pipeline until the the gem has been published.


randomanderson (approved):
👍

marcotc (approved):
I think, ideally, this step should be triggered after rubygems.org receives the new gem version.
rubygems.org publishing is not automatic today, so it's not possible to trigger this PR's job after the rubygems.org. But when we automate rubygems.org publishing, we should trigger this PR's job automatically, no through polling.

PR #3164, opened 2023-09-27
Refactor rails test

TonyCTHsu:
**Motivation:**

Preparing for Rails 7 test. 

The current setup for Rails in test is very confusing. There are nested conditionals in order to share code between versions. It is not only hard to debug, but also hard to write code that works for all versions. 

The pain could be mitigated by identifying the abstraction and split each version into its own file.  

Sharing configuration code between different Rails versions is problematic. Since different versions get access to different configuration interfaces, objects and utility functions. For instance: `::ActiveSupport::Subscriber.detach_from` is available from 6+ and `ActiveSupport::Logger::SimpleFormatter` is available after 4+. 

The goal would be making the code easy for developer to setup a new version(Rails 7) by adding `rails7.rb` and remove an old version (Rails 3) by deleting `rails3.rb`

**What does this PR do?**

1. Creates a single entry point for setting up a Rails app in test and fanout to each version to streamline the logic flow.
2. Stop sharing global configuration by pushing logic into individual Rails version and remove version checks
3. Only extract shared utility functions
4. Update each version configuration implementation with advance objects.
5. After the changes, obsolete code can be identified easily and removed.

**Additional Notes:**

This PR does not address some existing issue including: 

1. Loading `ddtrace` in auto instrumentation scenario
2. Configuring Rails with ENV by mocking ENV with non string values


marcotc (approved):
Cleaner and smaller, awesome work!

PR #3168, opened 2023-09-28
Automatically populate extra services

lloeki:
**What does this PR do?**

Attempt to automatically populate service names from the process instrumented service names.

**Motivation:**

Follow-up to #3032 which is very manual and obscure to customers.

**Additional Notes:**

This may catch much more than is needed.

**How to test the change?**

- CI
- play with `c.remote.extra_services`

**For Datadog employees:**

### DIFF
--- a/lib/datadog/core/configuration/settings.rb
+++ b/lib/datadog/core/configuration/settings.rb
@@ -643,13 +643,15 @@ module Datadog
           # configured to use names that differ from DD_SERVICE and should be
           # considered as a single remote configuration controllable service.
           #
-          # @default `nil`.
+          # @default `nil`. Automatically gathers service names from tracing configuration; use `[]` to disable.
           # @return [Array<String>,nil]
           option :extra_services do |o|
-            o.type :array
-            o.default []
+            o.type :array, nilable: true
+            o.default nil
             o.setter do |v|
-              if v.any? { |e| !e.is_a?(String) }
+              if v.nil?
+                v
+              elsif v.any? { |e| !e.is_a?(String) }
                 t = v.find { |e| !e.is_a?(String) }
 
                 raise ArgumentError, "Unexpected remote.extra_services element type #{t.class}, expected String."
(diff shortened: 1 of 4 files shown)


lloeki:
As it stands I found the following corner case:

```
  c.service = 'puma'
  c.tracing.instrument :rack, service_name: 'rack', request_queuing: true, web_service_name: 'nginx'
  c.remote.service = 'rack'
```

RC sees `rack` but not `puma` which is only normal since `puma` is not advertised via an integration.

Arguably, `nginx` should not appear (and indeed does not) as it's an ingress synthetic span.

lloeki:
Found another case that we can't easily automatically; the public tracing API allows:

```
  Datadog::Tracing.trace 'my-span-name', service: 'my-service-name' do
    # things being traced
  end
```

The service name is then only obtainable by gathering a list of service names via either:

- examining the trace itself once it's complete
- have `Datadog::Tracing.trace` (or any one of its sub-calls, down to the one that sets the service name metadata) and gather the argument value

And subsequently use that list to further populate `extra_services`.

marcotc:
> Found another case that we can't easily automatically; the public tracing API allows:
> 
> ```
>   Datadog::Tracing.trace 'my-span-name', service: 'my-service-name' do
>     # things being traced
>   end
> ```
> 
> The service name is then only obtainable by gathering a list of service names via either:
> 
>     * examining the trace itself once it's complete
> 
>     * have `Datadog::Tracing.trace` (or any one of its sub-calls, down to the one that sets the service name metadata) and gather the argument value
> 
> 
> And subsequently use that list to further populate `extra_services`.

We have code like this that collects service names from flushed traces: https://github.com/DataDog/dd-trace-rb/blob/e1a35966e453abb36df69153c4e2a34e16e767c3/lib/datadog/tracing/writer.rb#L141

Do you think it makes sense to add these as well to `extra_services`?

ivoanjo:
Do we want to make an explicit go/no-go decision on this one?

ivoanjo:
It's been more than a year, and I've not heard further discussion on this idea of "extra services" for a long while, so I'm going to assume we decided against this idea.

Going to go ahead and close this PR :)

PR #3172, opened 2023-09-29
[PROF-8299] Remove legacy profiler codepath

ivoanjo:
**What does this PR do?**

For a while, dd-trace-rb has actually been shipping two profilers.

Internally, we've been calling them "Ruby CPU Profiling 2.0" and "legacy profiler".

This PR removes the "legacy profiler", leaving the "Ruby CPU Profiling 2.0" profiler as the only option avaialble.

**Motivation:**

Back in April, the "Ruby CPU Profiling 2.0" went GA and thus became the default profiler that gets used.

At the time, I decided to keep the "legacy profiler" around just in case of unknown-unknowns: in case there was a big issue with "CPU Profiling 2.0" profiler that would make it not usable to some of our customers.

That was back in April, and here's us at the end of September and we did not see any issues, and adoption of newer versions of the gem (and thus the "CPU Profiling 2.0" profiler) is quite high.

So I'm confident we can remove the "legacy profiler", which is quite a nice cleanup to the codebase.

**Additional Notes:**

* The removed files in `lib/datadog/core` were actually not used by anything else in ddtrace. At some point in the past we thought we may reuse them, but we clearly didn't, so I think it's time to remove them. (These classes were not part of our public API, so they can be removed before 2.0).

* The protobuf/pprof decoding code was moved to `spec/` since the profiling specs still make use of it. The encoding part was deleted.

* There were a few tiny loose ends that were waiting for the "legacy profiler" to be deleted to be fixed/cleaned up. I've gone ahead and included those as well.

* The profiler no longer needs the `google-protobuf` gem (but we used it for the specs). Once this cleanup gets released, I'll do a PR to update https://docs.datadoghq.com/profiler/enabling/ruby/ to reflect this.

**How to test the change?**

Since the "legacy profiler" was off by default, the only thing to be tested is that the profiler is not affected by these changes.

**For Datadog employees:**


GustavoCaso (approved):
Amazing cleanup 🔥

PR #3174, opened 2023-09-29
Address graphql test failure in 3.3 preview

TonyCTHsu:
**What does this PR do?**

Add `racc` for graphql test with Ruby 3.3 preview 2

https://github.com/rmosolgo/graphql-ruby/issues/4650


ivoanjo (approved):
Thanks for this!

I think another alternative could be to add `racc` to the appraisal group (since that fixes the issue as I showed in the ticket), but I think this one is fine as well to "stop the bleeding" and bring our CI back to green.

PR #3175, opened 2023-09-29
Shorten appraisal filenames with only major and minor version of Ruby

TonyCTHsu:
**What does this PR do?**

Shorten the appraisal filenames with only major and minor version of Ruby, so that when we upgrade the patch version, no changes need to be made.


ivoanjo (approved):
:+1: Thanks for the nice improvement, LGTM!

ivoanjo:
@TonyCTHsu asked me to merge this in, so here it goes! :)

PR #3183, opened 2023-10-04
[PROF-7440] Add workaround for incorrect invoke location when logging gem is in use

ivoanjo:
**What does this PR do?**

In #2950, we added a fallback for nameless threads: we use their "invoke location", as shown by Ruby in the default `Thread#to_s`. The invoke location is the file and line of the first method of the thread.

This fallback has an issue: when thread creation is monkey patched. One common source of this is the `logging` gem. When the `logging` gem monkey patches thread creation, every thread will have the same invoke location, which will point to the `logging` gem.

This made the fallback invoke location worse than not having anything.

To work around this, this PR changes the profiler so that when the invoke location belongs to the `logging` gem, it's not used, and a simpler `(Unnamed thread)` placeholder is used instead.

**Motivation:**

This issue came up when testing the timeline feature with some of the internal Ruby apps.

**Additional Notes:**

In the future we could probably explore a more generic fix (e.g. using `Thread.method(:new).source_location` or something like that to detect redefinition) but doing that from the profiler native code is a bit more work so I decided to go with a simpler approach.

**How to test the change?**

Change includes test coverage. You can also see the thread name difference on an app by creating a simple thread before/after loading the `logging` gem and see the fallback name for the thread.

**For Datadog employees:**

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
@@ -815,6 +815,20 @@ static struct per_thread_context *get_context_for(VALUE thread, struct thread_co
   return thread_context;
 }
 
+#define LOGGING_GEM_PATH "/lib/logging/diagnostic_context.rb"
+
+// The `logging` gem monkey patches thread creation, which makes the `invoke_location_for` useless, since every thread
+// will point to the `logging` gem. When that happens, we avoid using the invoke location.
+static bool is_logging_gem_monkey_patch(VALUE invoke_file_location) {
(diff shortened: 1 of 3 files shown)


ekump (approved):
A hacky solution, but LGTM to unblock the 1.15 release. This is better than no fix at all. Might be worth adding a comment to the code indicating that it's likely a temporary solution?

ivoanjo:
> A hacky solution, but LGTM to unblock the 1.15 release. This is better than no fix at all. Might be worth adding a comment to the code indicating that it's likely a temporary solution?

Thanks for the review + agreed. Documented it in 4ff3cf630556ba369d32d66515398441848afdda.

PR #3184, opened 2023-10-05
Respect service_name in active_record.instantiation events

ozydingo:
**What does this PR do?**

Use the configured value of `service_name` for `active_record.instantiation` instrumentation

**Motivation:**

In our application, we configure the service name. This is being respected by all operations under rails and active_record, including `mysql.query` (active_record), but also `rack.request`, `active_job.perform`, `rails.render_template`, and so on, but not `active_record.instantiation`, which ends up in a separate, unique service. This is confusing and makes it harder to reason about our APM page.

**Additional Notes:**

We are configuring our application similarly to

```rb
  Datadog.configure do |c|
    # ...
    c.tracing.instrument :rails, service_name: "foo"
    c.tracing.instrument :active_record, service_name: "foo"  
    # ...
  end
```

And we are currently working around this limitation with a monkey patch to `Datadog::Tracing::Contrib::ActiveRecord::Events`:

```rb
        def process(span, event, _id, payload)
          span.service = "foo"
          super
        end
```

**How to test the change?**

There wasn't a spec for `active_record.instantiation` that I could see, so I added one. (Still trying to work out some local docker issues, so also trying to use this PR for CI)

**For Datadog employees:**

### DIFF
--- a/lib/datadog/tracing/contrib/active_record/events/instantiation.rb
+++ b/lib/datadog/tracing/contrib/active_record/events/instantiation.rb
@@ -31,6 +31,7 @@ module Datadog
 
             def process(span, event, _id, payload)
               span.resource = payload.fetch(:class_name)
+              span.service = configuration[:service_name] if configuration[:service_name]
               span.span_type = Ext::SPAN_TYPE_INSTANTIATION
               span.set_tag(Tracing::Metadata::Ext::TAG_COMPONENT, Ext::TAG_COMPONENT)
               span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_INSTANTIATION)
--- a/spec/datadog/tracing/contrib/active_record/instantiation_spec.rb
+++ b/spec/datadog/tracing/contrib/active_record/instantiation_spec.rb
@@ -1,0 +1,71 @@
+require 'datadog/tracing/contrib/support/spec_helper'
+require 'datadog/tracing/contrib/analytics_examples'
+require 'datadog/tracing/contrib/integration_examples'
+require 'datadog/tracing/contrib/span_attribute_schema_examples'
+require 'ddtrace'
+
+require 'spec/datadog/tracing/contrib/rails/support/deprecation'
+
+require_relative 'app'
+
+RSpec.describe 'ActiveRecord instantiation instrumentation' do
+  let(:configuration_options) { {} }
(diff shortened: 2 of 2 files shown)


ozydingo:
I'm having a lot of trouble getting these tests to work as I want them. I don't understand the abstractions that govern some of these tests; any help would be appreciated (also happy to roll back the addition of the active_record.instantiation spec since it wasn't even there before!)

p-datadog:
Hi @ozydingo , thank you very much for this PR.

I have extracted the tests in #3397, the main change I had to make is to retrieve a record from the database since the instantiation event fires when a model instance is populated with data loaded from the database, not when a new (empty) model instance is created.

For the code change, the service name of the instantiation span should match the service name of the application. This is more thoroughly explained in #3359 and #3397 adds a note to documentation to this effect.

I am not entirely sure from your description whether you are trying to hardcode the service name to the application service name or the SQL query service name (which is presumably overridden); if the explanation in #3359 is unclear can you please add the configuration of dd-trace-rb here and clarify which exact names are causing confusion?

ozydingo:
Thanks @p-datadog ! To clarify here, we have our `active_record` instrumentation with a custom service name

```
  Datadog.configure do |c|
    ...
    c.tracing.instrument :active_record, service_name: "foo"
```

Upon doing this, we noticed the `active_record.instantiation` spans were still showing up under the application name (`"threeplaymedia_app3"`) For various reasons, we don't want to use that service name. And while all other operations started showing up under the `foo` service, the operation `active_record.instantiation` was still showing up under the default service name (`threeplaymedia_app3`), and it was the _only_ remaining operation doing so. This makes this service a phantom in our datadog instance, and created a confusing an unnecessary service division between our customized service name and the unconfiguratble `active_record.instantiation` operation.

We mokney-patched the datadog lib to force the `active_record` instrumentation event to be the same "foo"; doing this successfully aligned `active_record.instantiaion` spans to the same service name as the other spans.

```
  module MonkeyPatches
    module Datadog
      module ActiveRecordInstatiation
        def process(span, event, _id, payload)
          span.service = "foo"
          super
        end
      end
    end
  end
```

I think the comments in #3359 adequately cover this, but I'll add my own color in case it helps build the case.

p-datadog:
I will close this based on the majority of the diff having been merged in #3397 , let's continue the discussion in #3359 regarding how the service names should be configured for instantiation.

PR #3188, opened 2023-10-05
skip passing waf addresses when the value is empty

GustavoCaso:
**What does this PR do?**

We remove empty and nil WAF Addresses before sending it to the WAF to be processed.

**Motivation:**

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**

### DIFF
--- a/lib/datadog/appsec/processor.rb
+++ b/lib/datadog/appsec/processor.rb
@@ -22,9 +22,16 @@ module Datadog
 
           start_ns = Core::Utils::Time.get_time(:nanosecond)
 
-          cleaned_input = remove_empty_entries(input)
+          input.reject! do |_, v|
+            case v
+            when TrueClass, FalseClass
+              false
+            else
+              v.nil? ? true : v.empty?
+            end
+          end
 
-          _code, res = @context.run(cleaned_input, timeout)
+          _code, res = @context.run(input, timeout)
 
           stop_ns = Core::Utils::Time.get_time(:nanosecond)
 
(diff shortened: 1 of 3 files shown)


lib/datadog/appsec/processor.rb
```diff
@@ -60,6 +62,18 @@ def extract_schema?
...
         end
+
+        def remove_empty_entries(entries)
+          entries.each_with_object({}) do |(k, v), acc|
+            acc[k] = v unless empty_value?(v)
+          end
+        end
+
+        def empty_value?(v)
+          return true unless v
```
lloeki, line 73: That would return `true` for `v == false`

lib/datadog/appsec/processor.rb
```diff
@@ -60,6 +62,18 @@ def extract_schema?
           Datadog.configuration.appsec.api_security.enabled &&
             Datadog.configuration.appsec.api_security.sample_rate.sample?
         end
+
+        def remove_empty_entries(entries)
+          entries.each_with_object({}) do |(k, v), acc|
+            acc[k] = v unless empty_value?(v)
```
lloeki, line 68: Why not just `unless v.nil? || v.empty?`

lib/datadog/appsec/processor.rb
```diff
@@ -22,7 +22,9 @@ def run(input, timeout = WAF::LibDDWAF::DDWAF_RUN_TIMEOUT)
 
           start_ns = Core::Utils::Time.get_time(:nanosecond)
 
-          _code, res = @context.run(input, timeout)
+          cleaned_input = remove_empty_entries(input)
```
lloeki, line 25: Looks like `remove_empty_entries` is merely `.reject { |k, v| v.nil? || v.empty? }` or did I miss something?

Also since `input` is thrown away in that case I would suggest mutating in place:

```suggestion
          input.reject! { |_, v| v.nil? || v.empty? }
```

lloeki (approved):
LGTM as a solution, although as is the code seems to have some holes.

lib/datadog/appsec/processor.rb
GustavoCaso: Absolutely true 😄

PR #3189, opened 2023-10-05
Support Opensearch 3

TonyCTHsu:
**Motivation:**

Support `opensearch-ruby` 3.x. 

**What does this PR do?**

`opensearch-ruby` 3.x merge the transport gem into it main gem. The target of instrumentation remains the same.

1. Add `opensearch-ruby` 3.x to our test matrix
2. Implement to get test green


lib/datadog/tracing/contrib/opensearch/integration.rb
```diff
@@ -18,8 +18,8 @@ class Integration
           register_as :opensearch, auto_patch: true
 
           def self.version
-            Gem.loaded_specs['opensearch-transport'] \
-              && Gem.loaded_specs['opensearch-transport'].version
+            Gem.loaded_specs['opensearch-ruby'] \
```
TonyCTHsu, line 21: Since the transport gem merges into the main gem, this is the changes required to have 3.x instrumented.

ekump (approved):

PR #3190, opened 2023-10-06
[PROF-8328] Import java-profiler PID controller and port it to C

ivoanjo:
**What does this PR do?**

This PR:

* Imports the `pidController.h` and `pidController.cpp` from the [java-profiler](https://github.com/DataDog/java-profiler/) library
* Ports the `PidController` class to C, to match the rest of the Ruby profiler

I've recorded the porting steps in multiple commits, and as much as possible tried to keep the conventions and structure of the original code.

In practice, this does not yet change anything in the Ruby profiler, since these files are not yet used.

**Motivation:**

I plan to use the PID controller to control how often we take samples in the allocation profiler, like the Java profiler does.

**Additional Notes:**

I plan to ask the folks from the Java profiler to review this one :)

**How to test the change?**

This code is not yet used; it'll get tested when it gets integrated into the allocation profiler.

**For Datadog employees:**


jbachorik (approved):
Looking good!

PR #3194, opened 2023-10-06
Reduce permissions to Action for creating next milestone

marcotc:
Similar to #3193, this PR reduces the permission to the `create-next-milestone` GitHub Action.

This action does not carry the same concern as #3193 because it does not execute based on external pull requests, but this is a step towards changing our default GitHub Actions token to read only.

### DIFF
--- a/.github/workflows/create-next-milestone.yml
+++ b/.github/workflows/create-next-milestone.yml
@@ -5,6 +5,9 @@ on:
 
 jobs:
   create_next_milestone:
+    permissions:
+      issues: write
+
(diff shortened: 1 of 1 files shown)


ivoanjo (approved):
:+1: LGTM, checks out, [milestone creation is listed as needing this permission](https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-issues)

GustavoCaso (approved):

PR #3197, opened 2023-10-09
Shut down profiler if any components failed

ivoanjo:
**What does this PR do?**

This PR changes the profiler `Collectors::CpuAndWallTimeWorker` and `Scheduler` components to signal the top `Profiler` instance when they failed, so the `Profiler` can stop the other component as well.

**Motivation:**

While never reported by a customer, it ocurred to me that right now if one of the component fails, the other will keep working in a weird state.

Of particular concern, if the scheduler failed, the `StackRecorder` would trigger a memory leak: it would start accumulating samples and never flush them.

This PR changes the profiler so that if one component stops with a failure, the other will do so as well.

**Additional Notes:**

I'm not especially happy with how this change turned out.

1. The error handling in the scheduler somewhat duplicates similar logic in the `Core::Workers::Async` helper, which it uses.

   I wanted to avoid changing the shared helper just for this use-case, which I managed, but at the cost of a bit of duplication (including a duplicated log message, sigh).

2. I don't like the `on_failure_proc: nil`, but adding the argument would mean changing A LOT of tests, which is doable but has its own ugliness as well.

   I've considered passing the `on_failure_proc` via the constructor, but it's not entirely straightforward: the `Profiler` instance is created after the two others, so we'd need a way to break the loop.

In the end, I decided to move forward with something, as I think it's important to plug this gap.

**How to test the change?**

**For Datadog employees:**

### DIFF
--- a/lib/datadog/profiling/scheduler.rb
+++ b/lib/datadog/profiling/scheduler.rb
@@ -45,22 +45,31 @@ module Datadog
         self.enabled = enabled
       end
 
-      def start
-        perform
+      def start(on_failure_proc: nil)
+        perform(on_failure_proc)
       end
 
-      def perform
-        # A profiling flush may be called while the VM is shutting down, to report the last profile. When we do so,
-        # we impose a strict timeout. This means this last profile may or may not be sent, depending on if the flush can
-        # successfully finish in the strict timeout.
-        # This can be somewhat confusing (why did it not get reported?), so let's at least log what happened.
-        interrupted = true
-
+      def perform(on_failure_proc)
         begin
-          flush_and_wait
-          interrupted = false
-        ensure
-          Datadog.logger.debug('#flush was interrupted or failed before it could complete') if interrupted
+          # A profiling flush may be called while the VM is shutting down, to report the last profile. When we do so,
+          # we impose a strict timeout. This means this last profile may or may not be sent, depending on if the flush can
+          # successfully finish in the strict timeout.
+          # This can be somewhat confusing (why did it not get reported?), so let's at least log what happened.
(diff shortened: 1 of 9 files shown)


lib/datadog/profiling/scheduler.rb
```diff
@@ -45,22 +45,31 @@ def initialize(
...
+          ensure
+            Datadog.logger.debug('#flush was interrupted or failed before it could complete') if interrupted
+          end
+        rescue Exception => e # rubocop:disable Lint/RescueException
+          Datadog.logger.warn(
+            'Profiling::Scheduler thread error. ' \
+            "Cause: #{e.class.name} #{e.message} Location: #{Array(e.backtrace).first}"
+          )
+          on_failure_proc&.call
+          raise
```
AlexJF, line 72: Why the extra begin, couldn't we have added the `rescue` to the other block? The only functional change I can see would be in the order of the debug vs warn lines but that doesn't seem that important?

AlexJF (approved):

lib/datadog/profiling/scheduler.rb
ivoanjo: You're right -- cleaned it up + added a spec for it in 844cf50d66cd4c1fbf46e7fac984c9da2ab3695c .

marcotc (approved):

ivoanjo:
Update: I had tried to merge master into this branch but that triggered a bunch of gemfile updates that were missing.

Instead, I pulled the gemfile updates into a separate PR (#3213) and rebased this branch on top of master once that other PR got merged.

PR #3204, opened 2023-10-17
ASM make sure to append content type and length information

GustavoCaso:
**What does this PR do?**

On ASM, append HTTP request information as part of the top-level span. 

The code that collects the request header assumes we are interested in all header requests starting with `HTTP_`.  Those headers are set by the client following the [RFC3875 section 4.1.18](https://tools.ietf.org/html/rfc3875#section-4.1.18) 

Unfortunately, that is not the case for specific headers that are specified by the server. For example, `Content-Type` and `Content-Length` are not prefixed with `HTTP_`  

The variables are called CGI server variables. [Here is the list](https://www6.uniovi.es/~antonio/ncsa_httpd/cgi/env.html)

This PR appends the content type and content length information and improves our specs.

**Motivation:**

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**

### DIFF
--- a/lib/datadog/appsec/contrib/rack/gateway/request.rb
+++ b/lib/datadog/appsec/contrib/rack/gateway/request.rb
@@ -40,9 +40,13 @@ module Datadog
             end
 
             def headers
-              request.env.each_with_object({}) do |(k, v), h|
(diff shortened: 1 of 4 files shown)


lloeki (approved):
LGTM

PR #3207, opened 2023-10-18
Name `Datadog::Core::Remote::Worker` thread

ivoanjo:
**What does this PR do?**

This PR sets the name of the thread started by the `Datadog::Core::Remote::Worker` class (to be the name of the class).

This follows the same pattern we use for other dd-trace-rb internal threads.

**Motivation:**

The profiler shows thread names in multiple places, and nameless threads make it harder to see the data in several cases, so as much as possible we want to avoid them.

One particular case is that for the timeline feature, we collapse all dd-trace-rb threads by default, but that collapsing feature relies on thread names, and so the remote worker thread was not being properly collapsed (which is how I spotted it).

**Additional Notes:**

I've chosen to still include the Ruby 2.2/2.1 support because I plan to backport this to the 1.x branch.

**How to test the change?**

This change includes test coverage. You can also enable the profiler and check that when looking at thread names (right sidebar in single and aggregated profiles; left side for timeline) you will see the thread for this class properly named.

**For Datadog employees:**

### DIFF
--- a/lib/datadog/core/remote/worker.rb
+++ b/lib/datadog/core/remote/worker.rb
@@ -29,6 +29,7 @@ module Datadog
           @starting = true
 
           @thr = Thread.new { poll(@interval) }
+          @thr.name = self.class.name unless Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.3')
 
           @started = true
           @starting = false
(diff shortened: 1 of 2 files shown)


ekump (approved):

ivoanjo:
Update: I had tried to merge master into this branch but that triggered a bunch of gemfile updates that were missing.

Instead, I pulled the gemfile updates into a separate PR (#3213) and rebased this branch on top of master once that other PR got merged.

PR #3211, opened 2023-10-23
Disable Ruby 3.3 integration due to protobuf compilation error

marcotc:
Disable integration test for Ruby 3.3. that cannot install its gemfiles until https://github.com/protocolbuffers/protobuf/issues/14509 is addressed.


anmarchenko (approved):

ivoanjo (approved):
:+1: We should probably keep an eye out for the fix, timeline-wise.

E.g. if by the beginning of December this isn't fixed, we I think we should consider other solutions, so that we have test coverage by the time Ruby 3.3 stable gets released.

PR #3218, opened 2023-10-24
Action:Fix PR milestone action

marcotc:
This PR fixes the permission for the `.github/workflows/add-milestone-to-pull-requests.yml` workflow, which attaches milestones to PRs.

I think there's something weird going on here because the octokit calls from the workflow returned 200, but did not modify the milestone AND the listed permission required for updating a pull request do not seem to coincide with reality: I think it's a GH bug.

### Other comments 

It also upgrades the GitHub actions used in the workflow as it was using a very very old version, which used node12 and GitHub has been complaining that node12 is deprecated:
```
The following actions uses node12 which is deprecated and will be forced to run on node16: actions/github-script@0.9.0. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/
```

### DIFF
--- a/.github/workflows/add-milestone-to-pull-requests.yml
+++ b/.github/workflows/add-milestone-to-pull-requests.yml
@@ -8,6 +8,8 @@ on:
 jobs:
   add_milestone_to_merged:
     permissions:
+      # These two permissions are needed to modify milestone, even though only one should be enough
+      issues: write
       pull-requests: write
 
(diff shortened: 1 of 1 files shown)


ivoanjo (approved):
:+1: LGTM Thanks for looking into this!

GustavoCaso (approved):

PR #3226, opened 2023-10-25
Hide public documentation for App Analytics

marcotc:
[App Analytics](https://docs.datadoghq.com/tracing/legacy_app_analytics/) has been deprecated since ddtrace 1.0.

We've removed all explicit mention of it in our public documentation in ddtrace but missed a few lines here and there.

This PR removes or hides App Analytics information as much as possible, without breaking existing users.
It's important to keep support for App Analytics tagging still because the migration to other ingestion mechanism is not automatic, and require user configuration in the Datadog App.

There's no change in behaviour with this PR.


ekump (approved):

buraizu (approved):

GustavoCaso (approved):

delner:
If this is marked for 2.0, why deprecate and not remove?

PR #3212, opened 2023-10-24
RFC: Remove tracer threads integration test

ivoanjo:
**What does this PR do?**

As discussed during last week's meeting, I'm opening this PR to discuss the removal of this long-standing flaky test.

See
<https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/12281/workflows/f4763f8c-d024-405b-bc00-cae3c5cecc0c/jobs/462118> for a recent example of it failing.

**Motivation:**

Avoid CI flakiness by either fixing or disabling/removing flaky tests.

**Additional Notes:**

N/A

**How to test the change?**

Observe that CI is green :)

**For Datadog employees:**
- [ ] If this PR touches code that signs or publishes builds or packages, or handles
credentials of any kind, I've requested a review from `@DataDog/security-design-and-guidance`.
- [x] This PR doesn't touch any of that.

PR #3228, opened 2023-10-25
Attempt to remove flaky test for checking that tracer threads are correctly shutdown

ekump:
**What does this PR do?**
~~Instead of checking number of threads before and after tracer is started and shutdown, we grab the object IDs for the threads the tracer creates and check that they are no longer present after the tracer shuts down. This test has been flaky for some time.~~

Add back `Timeout.ensure_timeout_thread_created if Timeout.respond_to?(:ensure_timeout_thread_created)` to spec_helper.rb that was removed in https://github.com/DataDog/dd-trace-rb/pull/2991. The Timeout thread can lead to flaky tests where we are checking what threads are present before and after actions as it could be created during a test. This ensures the thread is always there before a test starts for consistent observation.

Also, improve the existing test for tracing thread shutdown by checking the thread objects themselves and not just the size of `Threads.list` and make the error output clearer. 

**Motivation:**
Flaky tests make people sad. This PR addresses #3212 

**Additional Notes:**

**How to test the change?**
I ran the rspec test several times and didn't observe any false failures. The previous test would fail at least 1 in every 10 times locally. 

**For Datadog employees:**

### DIFF
--- a/spec/datadog/integration_spec.rb
+++ b/spec/datadog/integration_spec.rb
@@ -22,20 +22,13 @@ RSpec.describe 'Datadog integration' do
     end
 
     context 'for threads' do
-      let!(:original_threads) { Thread.list }
-      let(:start_tracer) do
-        Datadog::Tracing.trace('test.op') {}
-
-        new_threads = Thread.list - original_threads
-        tracer_threads.concat(new_threads)
-      end
-      let(:tracer_threads) { [] }
+      let(:original_threads) { Thread.list }
+      let!(:original_threads_inspect) { inspect_threads(original_threads) } # Store result as stack trace will change
+      let(:threads) { Thread.list }
 
       def inspect_threads(threads)
-        threads.map.with_index { |t, idx| "#{idx}=#{t.object_id}:#{t.backtrace}" }.join(';')
+        threads.map.with_index { |t, idx| "#{idx}=#{t.backtrace}" }.join(';')
       end
-
-      subject(:shutdown) { Datadog.shutdown! }
 
       it 'closes tracer threads' do
         start_tracer
@@ -43,10 +36,11 @@ RSpec.describe 'Datadog integration' do
 
         shutdown
 
-        post_shutdown_threads = Thread.list
+        post_shutdown_new_threads = Thread.list - original_threads
 
-        expect(post_shutdown_threads & tracer_threads).to be_empty,
-          "Tracer threads not terminated: #{inspect_threads(post_shutdown_threads)}"
+        expect(post_shutdown_new_threads).to be_empty,
+          "Expected #{original_threads.size} threads but got #{threads.size} threads. "\
+            "extra_threads: (#{inspect_threads(post_shutdown_new_threads)})"
       end
     end
 
--- a/spec/datadog/integration_spec.rb
+++ b/spec/datadog/integration_spec.rb
@@ -23,12 +23,19 @@ RSpec.describe 'Datadog integration' do
 
     context 'for threads' do
       let(:original_threads) { Thread.list }
-      let!(:original_threads_inspect) { inspect_threads(original_threads) } # Store result as stack trace will change
-      let(:threads) { Thread.list }
+      let(:start_tracer) do
+        Datadog::Tracing.trace('test.op') {}
+
+        new_threads = Thread.list - original_threads
+        tracer_threads.concat(new_threads.map(&:object_id))
+      end
+      let(:tracer_threads) { [] }
 
       def inspect_threads(threads)
-        threads.map.with_index { |t, idx| "#{idx}=#{t.backtrace}" }.join(';')
+        threads.map.with_index { |t, idx| "#{idx}=#{t.object_id}:#{t.backtrace}" }.join(';')
       end
+
+      subject(:shutdown) { Datadog.shutdown! }
 
       it 'closes tracer threads' do
         start_tracer
@@ -36,9 +43,10 @@ RSpec.describe 'Datadog integration' do
 
         shutdown
 
-        expect(threads.size).to eq(original_threads.size),
-          "Expected #{original_threads.size} threads (#{original_threads_inspect}), "\
-                  "got #{threads.size} threads (#{inspect_threads(threads)})"
+        post_shutdown_threads = Thread.list.map(&:object_id)
+
+        expect(post_shutdown_threads & tracer_threads).to be_empty,
+          "Tracer threads not terminated: #{inspect_threads(Thread.list)}"
       end
     end
 
--- a/spec/datadog/integration_spec.rb
+++ b/spec/datadog/integration_spec.rb
@@ -22,12 +22,12 @@ RSpec.describe 'Datadog integration' do
     end
 
     context 'for threads' do
-      let(:original_threads) { Thread.list }
+      let!(:original_threads) { Thread.list }
       let(:start_tracer) do
         Datadog::Tracing.trace('test.op') {}
 
         new_threads = Thread.list - original_threads
-        tracer_threads.concat(new_threads.map(&:object_id))
+        tracer_threads.concat(new_threads)
       end
       let(:tracer_threads) { [] }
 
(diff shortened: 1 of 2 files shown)


spec/datadog/integration_spec.rb
```diff
@@ -23,22 +23,30 @@ def wait_for_tracer_sent
 
     context 'for threads' do
       let(:original_threads) { Thread.list }
-      let!(:original_threads_inspect) { inspect_threads(original_threads) } # Store result as stack trace will change
-      let(:threads) { Thread.list }
+      let(:start_tracer) do
+        Datadog::Tracing.trace('test.op') {}
+
+        new_threads = Thread.list - original_threads
```
ivoanjo, line 29: Be careful with the lazy initialization of `let`s. Since `original_threads` is only referenced after starting the tracer, this line is the equivalent of `new_threads = Thread.list - Thread.list`.

spec/datadog/integration_spec.rb
```diff
@@ -23,22 +23,30 @@ def wait_for_tracer_sent
...
+        tracer_threads.concat(new_threads.map(&:object_id))
```
ivoanjo, line 30: I'm not quite sure I follow the need to use `object_id`s instead of the thread objects. What's the intention?

If you're worried that keeping the actual thread objects referenced is a problem, it's actually not: `Thread.list` only returns threads that are alive, even though more `Thread` objects may still exist:

```ruby
$ pry
[1] pry(main)> Thread.list
=> [#<Thread:0x0000786c3b447c60 run>]
[2] pry(main)> thread = Thread.new { sleep 2 }
=> #<Thread:0x0000786c3b195b20 (pry):2 run>
[3] pry(main)> Thread.list
=> [#<Thread:0x0000786c3b447c60 run>, #<Thread:0x0000786c3b195b20 (pry):2 sleep>]
[4] pry(main)> sleep 2
=> 2
[5] pry(main)> Thread.list
=> [#<Thread:0x0000786c3b447c60 run>]
[6] pry(main)> thread
=> #<Thread:0x0000786c3b195b20 (pry):2 dead>
```

spec/datadog/integration_spec.rb
```diff
@@ -23,22 +23,30 @@ def wait_for_tracer_sent
...
 
         shutdown
 
-        expect(threads.size).to eq(original_threads.size),
-          "Expected #{original_threads.size} threads (#{original_threads_inspect}), "\
-                  "got #{threads.size} threads (#{inspect_threads(threads)})"
+        post_shutdown_threads = Thread.list.map(&:object_id)
+
+        expect(post_shutdown_threads & tracer_threads).to be_empty,
+          "Tracer threads not terminated: #{inspect_threads(Thread.list)}"
```
ivoanjo, line 49: I would suggest calling to `Thread.list` only once, because otherwise if a thread finishes between the call on line 46 and the call on line 49, we'll get a really confusing error message.

ivoanjo (commented):
Thanks for looking into this, left a few notes!

spec/datadog/integration_spec.rb
ekump: By default, I prefer to compare identifiers instead of objects, but you're correct that it's unnecessary here.

spec/datadog/integration_spec.rb
```diff
@@ -22,23 +22,31 @@ def wait_for_tracer_sent
...
 
         shutdown
 
-        expect(threads.size).to eq(original_threads.size),
-          "Expected #{original_threads.size} threads (#{original_threads_inspect}), "\
-                  "got #{threads.size} threads (#{inspect_threads(threads)})"
+        post_shutdown_threads = Thread.list
+
+        expect(post_shutdown_threads & tracer_threads).to be_empty,
+          "Tracer threads not terminated: #{inspect_threads(post_shutdown_threads)}"
```
ivoanjo, line 49: So just to make sure I'm following the change here: The previous flaky test was checking that between the start of the test and after calling `shutdown`, there were no threads remaining.

The new version is instead checking that between the start of the test, and after calling `Datadog::Tracing.trace` no new threads remain. Thus, the `wait_for_tracer_sent` and `shutdown` parts aren't covered anymore. And presumably the flakiness was coming from there and since we're not counting them, the test will pass?

spec/datadog/integration_spec.rb
ivoanjo: Tip: You don't need to manually compare `object_id`, RSpec has helpers for that: https://dev.to/andy4thehuynh/the-difference-between-rspec-eq-and-be-3jep :)

spec/datadog/integration_spec.rb
ekump: As I understand it, the previous iteration of this test was simply checking the number of threads present in `Threads.list` before and after the tracer started and shutdown. Also, as you pointed out in my refactor, the old test was lazily evaluating the instantiation of `original_threads`.

I refactored the test to check the specific threads created when the tracer starts are no longer present after shutdown. My thesis is that unrelated threads may have been in `Threads.list` at the time of observation when the test executes, causing the flakiness. The `wait_for_tracer_sent` and `shutdown` are still executed and necessary for the test to pass in the refactored version.

ivoanjo: > I refactored the test to check the specific threads created when the tracer starts are no longer present after shutdown.

Indeed! What I'm pointing out is -- I'm not sure _when_ the extra threads that caused the test to be flaky were created.

So if those threads happened in the `wait_for_tracer_sent` or `shutdown` phases, we now ignore them, which is presumably why the flakiness went away -- right?

> as you pointed out in my refactor, the old test was lazily evaluating the instantiation of original_threads.

Nitpick xD but this was actually not the case! The old test had a `let!(:original_threads_inspect)` which called the `original_threads`, making it so that indirectly, `original_threads` was a `let!`. The issue I pointed out actually only happened after the inspect was removed.

ekump: I dug into what the extra threads were and it turns out it's Ruby's [Timeout](https://ruby-doc.org/stdlib-3.1.0/libdoc/timeout/rdoc/Timeout.html) module. 

Adding `Timeout.ensure_timeout_thread_created if Timeout.respond_to?(:ensure_timeout_thread_created)` to our spec_helper ensures the thread is there before the test is run so it is included in `original_threads` and therefore should stop the flakiness.

I mostly reverted my test changes as @marcotc pointed out it was testing only the threads created at startup are no longer present. But it's possible threads created after startup could still exist and my test would incorrectly pass.

spec/spec_helper.rb
```diff
@@ -280,3 +280,11 @@ def initialize(*args, &block)
...
 # Helper matchers
 RSpec::Matchers.define_negated_matcher :not_be, :be
+
+# The Ruby Timeout class uses a long-lived class-level thread that is never terminated.
+# Creating it early here ensures tests that tests that check for leaking threads are not
+# triggered by the creation of this thread.
+#
+# This has to be one once for the lifetime of this process, and was introduced in Ruby 3.1.
+# Before 3.1, a thread was created and destroyed on every Timeout#timeout call.
+Timeout.ensure_timeout_thread_created if Timeout.respond_to?(:ensure_timeout_thread_created)
```
ivoanjo, line 290: Since you mentioned this was Marco's suggestion, I wonder what changed from https://github.com/DataDog/dd-trace-rb/pull/2991 ?

I remember we had this in the past and removed it again :eyes:

spec/datadog/integration_spec.rb
ivoanjo: Nice!

spec/spec_helper.rb
ekump: I don't believe the call to `ensure_timeout_thread_created` was removed because it was causing any problems. #2991 was attempting to solve the timeout thread being a false positive in the leaky thread report. In addition to the call to `ensure_timeout_thread_created` the old code was also [assigning a thread group to the timeout thread](https://github.com/DataDog/dd-trace-rb/pull/2991/files#diff-89eebfcbc0f14b6d989517837ca1e94fce4e2ce9a03233641cd936f2b8d2ed94L290-L298). This thread group was used to filter it out of the leaky thread report. However, the timeout code itself [overwrites the thread group with its own](https://github.com/ruby/timeout/blob/54835eaf4036c9f10503541d8c2bcf0b7bcdd6ab/lib/timeout.rb#L120). #2991 added [this check for the thread group](https://github.com/DataDog/dd-trace-rb/pull/2991/files#diff-89eebfcbc0f14b6d989517837ca1e94fce4e2ce9a03233641cd936f2b8d2ed94R146-R147) set in the timeout code, so it doesn't get reported as leaky. 

I believe the call to `ensure_timeout_thread_created` was removed as it was seen as unnecessary and the assertion of threads being present before/after certain conditions in tests wasn't taken into consideration at the time. @marcotc feel free to correct me if I mischaracterized your thinking here.

ivoanjo (approved):
:+1: LGTM

marcotc (approved):

PR #3230, opened 2023-10-26
[APPSEC-11887] Add Telemetry metric support

GustavoCaso:
**What does this PR do?**

Adds support for telemetry metrics. 

Telemetrics metrics is enabled by default and can be configured using `c.telemetry.metrics = <bool>` or using `DD_TELEMETRY_METRICS_ENABLED`

Each product will be able to send metrics using a straightforward API.

```ruby
telemetry_client.add_<metric_type>_metric(namespace ["tracers", "general", "telemetry", "iast", "appsec"], metric_key String, value Numeric, tags Hash)
```

The metrics will be flushed at the interval configured for telemetry. I decided to reuse the same interval as telemetry to avoid having too many configurations for telemetry, also the RFC did not specify anything about this particular option.

**Motivation:**

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**


lib/datadog/core/telemetry/metric_queue.rb
```diff
@@ -0,0 +1,55 @@
...
+            existing_metric.update_value(value)
+            @metrics[metric_klass.request_type][namespace][name] = existing_metric
+          else
+            new_metric = metric_klass.new(name, tags)
+            new_metric.update_value(value)
+            @metrics[metric_klass.request_type][namespace][name] = new_metric
+          end
+        end
+
+        def build_metrics_payload
```
marcotc, line 29: Do we have to reset the metric counters when we flush?

GustavoCaso: Yes. 

We do in the client right after sending the metrics. 
https://github.com/DataDog/dd-trace-rb/blob/27142bdee303ad61fe395fe43c3f22624b3b5b6c/lib/datadog/core/telemetry/client.rb#L120-L124

lib/datadog/core/telemetry/client.rb
```diff
@@ -74,13 +94,48 @@ def client_configuration_change!(changes)
...
 
           @emitter.request(:'app-heartbeat')
         end
+
+        def flush_metrics!
+          return if !@metrics_enabled || forked?
+
+          # Send metrics
+          @metric_queue.build_metrics_payload do |metric_type, payload|
+            @emitter.request(metric_type, payload: payload)
```
anmarchenko, line 134: sorry if I missed something obvious, but what if another metric arrives when this request is running and GIL is released? would this metric be lost when we call `@metric_queue = MetricQueue.new` afterwards?

anmarchenko (commented):
thank you for the great work! 

I have only one question about sync, probably something I missed

marcotc (approved):

lib/datadog/core/telemetry/emitter.rb
```diff
@@ -22,12 +22,18 @@ def initialize(http_transport: Datadog::Core::Telemetry::Http::Transport.new)
         # Retrieves and emits a TelemetryRequest object based on the request type specified
         # @param request_type [String] the type of telemetry request to collect data for
         # @param data [Object] arbitrary object to be passed to the respective `request_type` handler
-        def request(request_type, data: nil)
+        def request(request_type, data: nil, payload: nil)
+          if payload && data
+            Datadog.logger.debug('Can not provide data and payload when generating temetry request')
```
ivoanjo, line 27: There's a typo in the message:

```suggestion
            Datadog.logger.debug('Cannot provide data and payload when generating telemetry request')
```

...but more than that, providing `data` and `payload` looks like a bug. So if it's a bug, have you considered raising an `ArgumentError` to make the error visible? That would make it more obvious when we're doing the wrong thing, whereas I can imagine the debug log here being lost under the radar.

lib/datadog/core/telemetry/client.rb
```diff
@@ -12,26 +15,42 @@ class Client
...
           @unsupported = false
           @worker = Telemetry::Heartbeat.new(enabled: @enabled, heartbeat_interval_seconds: heartbeat_interval_seconds) do
             heartbeat!
           end
+
+          Metric::Rate.interval = heartbeat_interval_seconds
+
+          @metric_queue = MetricQueue.new
+
+          @metrics_worker = Telemetry::MetricWorker.new(
```
ivoanjo, line 41: Minor: The inconsistency between `metric` and `metrics` is a bit weird in these names

lib/datadog/core/telemetry/metric_worker.rb
```diff
@@ -0,0 +1,35 @@
...
+
+require_relative '../worker'
+require_relative '../workers/polling'
+
+module Datadog
+  module Core
+    module Telemetry
+      # Periodically (every DEFAULT_INTERVAL_SECONDS) sends metrics to the telemetry API.
+      class MetricWorker < Core::Worker
+        include Core::Workers::Polling
```
ivoanjo, line 11: So I'm not the one to talk a lot, the profiler already uses 3 threads. But... do we really really want to add another background thread to dd-trace-rb?

If we enable every product, we may be getting close to 9 or 10 now which seems to add up to quite a bit...

lib/datadog/core/telemetry/metric_queue.rb
```diff
@@ -0,0 +1,56 @@
...
+      # Stores all the metrics information by request_type during a time interval
+      class MetricQueue
+        attr_reader :metrics
+
+        def initialize
+          @metrics = {
+            'generate-metrics' => {},
+            'distributions' => {},
+          }
+        end
```
ivoanjo, line 15: Is this data structure a bit of a data race?

* Can multiple threads be writing to it through `@metrics` at the same time?
* Can the reader thread calling `build_metrics_payload` run at the same time as a metric is being written?

lib/datadog/core/telemetry/metric.rb
```diff
@@ -0,0 +1,126 @@
...
+        # GenerateMetricType
+        class GenerateMetricType < Base
+          class << self
+            def request_type
+              'generate-metrics'
+            end
+          end
+        end
+
+        # DistributionsMetricType
```
ivoanjo, line 54: Minor: It seems weird to add these comments just to "shut up" rubocop -- we're actually making our code worse to make rubocop "happy".

ivoanjo (commented):
I did a quick pass and left a few notes!

lib/datadog/core/telemetry/emitter.rb
GustavoCaso: I thought about raising an exception. But I went against it because using an exception to control the flow felt weird. 

We already handle the exception when executing the HTTP operation, and if things fail, we log and return a value that, by the look of it, we never check, so I thought just adding some logging would be enough.

If you feel like we should through an exception I'm happy to change the code 😄

lib/datadog/core/telemetry/metric_worker.rb
GustavoCaso: That is an excellent point.

And since this iteration, we use the same interval time as the heartbeat, we could use the same thread to send the heartbeat and the metrics. 

I will change the code to reflect that. I might have to rename `Heartbeat` to something like `TelemetryWorker` 😄

lib/datadog/core/telemetry/metric_queue.rb
GustavoCaso: Indeed, it is not thread-safe 😄 

Let me change it a bit to make it thread-safe 🔒

lib/datadog/core/telemetry/emitter.rb
ivoanjo: I'm always a bit wary of "swallowing" errors -- in some cases we need to do it, but I'm not entirely convinced this is one of them. Do you have any concerns here with using an exception?

lib/datadog/core/telemetry/metric_worker.rb
GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/3230/commits/4826434cdde86dad5e31d3ea1eaeee822e5e3bf5

lib/datadog/core/telemetry/metric_queue.rb
GustavoCaso: Done https://github.com/DataDog/dd-trace-rb/pull/3230/commits/1574d153ef9ab931c46b1886eac54d3ef557a19e

anmarchenko (approved):

lib/datadog/core/telemetry/emitter.rb
ivoanjo: > If we accidentally misuse the API, we should not break the customers' application.

Well... I'm not sure if I agree that this is the best way of doing that :D

Although, I'll clarify that I don't think changing this is a blocker for the rest of the PR -- to me it's more of a "in general I think we should avoid swallowing errors". :thinking: 

My reasoning is -- if we want to make absolutely sure we don't break customer's apps, I think the way to go isto have a begin/rescue that encompasses the whole API call, starting from the customer code. Otherwise we may always have some accidental exception that bubbles back up. So it's a bit weird to me to say "here we won't raise an exception because" instead of "here we always rescue because".

spec/datadog/core/telemetry/event_spec.rb
```diff
@@ -62,6 +65,26 @@
...
 
+      ['generate-metrics', 'distributions'].each do |request_type|
+        context request_type do
+          let(:request_type) { request_type }
+
+          context 'when payload is nil' do
+            let(:payload) { nil }
+
+            it 'raise ArgumentError' do
+              expect { telemetry_request }.to raise_error(ArgumentError)
```
ivoanjo, line 76: Minor: When reusing system exceptions, it's not a bad idea to try to match a bit on the message, so that we don't accidentally make the test pass by triggering another `ArgumentError` in the test code itself (e.g. if we pass the wrong number of arguments to the test code, for instance)

```suggestion
            it 'raises an ArgumentError' do
              expect { telemetry_request }.to raise_error(ArgumentError, /needs to provide the payload/)
```

spec/datadog/core/telemetry/metric_queue_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+          1,
+          { foo: :bar },
+          metric_klass
+        )
+
+        expect(metric_queue.send(:metrics)[metric_klass.request_type]['test_namespace']).to_not be_nil
+        expect(metric_queue.send(:metrics)[metric_klass.request_type]['test_namespace']['test_metric_name']).to_not be_nil
+
+        metric_instace = metric_queue.send(:metrics)[metric_klass.request_type]['test_namespace']['test_metric_name']
+        expect(metric_instace).to be_a(metric_klass)
```
ivoanjo, line 34: Minor: Since the last check won't pass unless the previous ones do, I suggest keeping only the last check. Especially since in this case we're asserting on private state of the class, it's IMHO a good idea to minimize the number of assertions, so it's easier to fix if we need to tweak the representation.

spec/datadog/core/telemetry/metric_queue_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+        metric_queue.add_metric(
+          'test_namespace',
+          'test_metric_name',
+          2,
+          { foo: :bar },
+          metric_klass
+        )
+
+        metric_instace = metric_queue.send(:metrics)[metric_klass.request_type]['test_namespace']['test_metric_name']
+        expect(metric_instace).to be_a(metric_klass)
```
ivoanjo, line 60: Minor:

```suggestion
        metric_instance = metric_queue.send(:metrics)[metric_klass.request_type]['test_namespace']['test_metric_name']
        expect(metric_instance).to be_a(metric_klass)
```

spec/datadog/core/telemetry/metric_queue_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+      end
+    end
+  end
+
+  describe '#build_metrics_payload' do
+    it 'yields metric_type and assiciated payload' do
```
ivoanjo, line 66: Minor:

```suggestion
    it 'yields metric_type and associated payload' do
```

spec/datadog/core/telemetry/metric_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+require 'datadog/core/telemetry/metric'
+
+RSpec.describe Datadog::Core::Telemetry::Metric::Count do
+  subject(:metric) { described_class.new('tests', { foo: :bar }) }
+  let(:generate_metric_type) { 'generate-metrics' }
+
+  describe '.request_type' do
+    subject(:request_type) { metric.class.request_type }
+
+    it { is_expected.to eq(generate_metric_type) }
```
ivoanjo, line 12: Minor: Since this is only used once, consider inlining it? (Same applies to other examples in this file)

```suggestion
  describe '.request_type' do
    subject(:request_type) { metric.class.request_type }

    it { is_expected.to eq('generate-metrics') }
```

spec/datadog/core/telemetry/metric_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+  end
+
+  describe '#update_value' do
+    it 'increment values' do
+      expect(metric.values).to be_nil
+      metric.update_value(1)
+      expect(metric.values).to_not be_nil
+      expect(metric.values[0][1]).to eq 1
```
ivoanjo, line 20: Minor: This assertion is kinda not needed, since the next one can only pass if this one passes (same note for others in the file)

```suggestion
      expect(metric.values[0][1]).to eq 1
```

spec/datadog/core/telemetry/metric_spec.rb
```diff
@@ -0,0 +1,184 @@
...
+      expect(metric.values[0][1]).to eq 0.1
+      metric.update_value(1)
+      expect(metric.values[0][1]).to eq 0.2
+    end
+
+    it 'stores the timestamp as part of the values array' do
+      expect(metric).to receive(:timestamp).and_return(1234)
+      metric.update_value(1)
+      expect(metric.values[0][0]).to eq 1234
+    end
```
ivoanjo, line 82: Should we test that the timestamp gets correctly updated a second time when new values get added? (Same for others in this file)

ivoanjo (commented):
Re-did a full review! It's a lot of comments -- I had not done a full review on my previous pass, had only checked the thready bits :sweat_smile:

lib/datadog/core/telemetry/client.rb
```diff
@@ -12,25 +14,37 @@ class Client
...
-        def initialize(heartbeat_interval_seconds:, enabled: true)
+        def initialize(heartbeat_interval_seconds:, metrics_enabled:, enabled: true)
           @enabled = enabled
+          @metrics_enabled = metrics_enabled
           @emitter = Emitter.new
           @stopped = false
           @unsupported = false
-          @worker = Telemetry::Heartbeat.new(enabled: @enabled, heartbeat_interval_seconds: heartbeat_interval_seconds) do
+          @worker = Telemetry::Worker.new(
+            enabled: @enabled || @metrics_enabled,
```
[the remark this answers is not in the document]
GustavoCaso: The fact that the telemetry is split into two separate features means that the worker could be enabled if one or the other is enabled.

Then, each of the methods that the worker calls after each interval checks the correct instance variable to make sure we do not report things that we shouldn't

https://github.com/DataDog/dd-trace-rb/blob/f4ddb4b66c86d9388e150a0ae47820d7a2c84b08/lib/datadog/core/telemetry/client.rb#L123-L124

https://github.com/DataDog/dd-trace-rb/blob/f4ddb4b66c86d9388e150a0ae47820d7a2c84b08/lib/datadog/core/telemetry/client.rb#L117-L118

This change was done after the thread about using a single background thread to handle all background tasks for telemetry. https://github.com/DataDog/dd-trace-rb/pull/3230#discussion_r1374860235

sig/datadog/core/telemetry/client.rbs
```diff
@@ -1,21 +1,27 @@
 module Datadog
   module Core
     module Telemetry
```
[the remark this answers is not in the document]
GustavoCaso: I tried that, but it created a cascade of failures from Steep, that are not related with the changes on this PR, so I rather fix the typing aspect of telemetry on a separate PR 😄

spec/datadog/core/telemetry/metric_queue_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+
+RSpec.describe Datadog::Core::Telemetry::MetricQueue do
+  let(:metric_queue) { described_class.new }
+  let(:metric_klass) { Datadog::Core::Telemetry::Metric::Count }
+  subject(:empty_metrics_queue) do
+    {
+      'generate-metrics' => {},
+      'distributions' => {},
+    }
+  end
```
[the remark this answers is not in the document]
GustavoCaso: It should not be the subject 🤦 I will change into a let

lib/datadog/core/telemetry/metric.rb
```diff
@@ -0,0 +1,126 @@
...
+            'gauge'
+          end
+        end
+
+        # Distribution metric are a metric type that aggregate values during the interval.
+        class Distribution < DistributionsMetricType
+          def update_value(value)
+            @values ||= []
+            @values << value
+          end
```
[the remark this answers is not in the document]
GustavoCaso: That is a good point. When reading the RFC I did not see any limit requirements, but I can ask for more clarification

lib/datadog/core/telemetry/metric.rb
```diff
@@ -0,0 +1,126 @@
...
+
+        # Count metric sup all the submitted values in a time interval
+        class Count < GenerateMetricType
+          def update_value(value)
+            if @values
+              @values[0][1] += value
+            else
+              @values = [[timestamp, value]]
+            end
+          end
```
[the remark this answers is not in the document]
GustavoCaso: From talking with the instrumentation people:
> The data you're sending is:
- The total calls in the interval period (10s)
- The time at which the interval period ends (the timestamp)
So updating the timestamp when you increment the count would give some strange results

lib/datadog/core/telemetry/client.rb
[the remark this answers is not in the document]
ivoanjo: Yes! To clarify, I got the feeling that this was a deliberate change (it has specs validating it even -- good job there also!).

My question is more, let's turn this around and think about the customer POV. Our default is to enable telemetry, so, that I can think of, there's only one situation where telemetry is disabled: the customer has made a deliberate decision to disable it.

This decision may have been because [it interfered with some of their tests](https://github.com/DataDog/dd-trace-rb/issues/2823) or they are not interested at all in Datadog collecting this data, and we document that in our `GettingStarted.md`:

> Allows you to enable sending telemetry data to Datadog. Can be disabled, as documented [here](https://docs.datadoghq.com/tracing/configure_data_security/#telemetry-collection)

So consider that a customer has expressed their intention: I do not want telemetry being sent. It's weird that as we add some extra data for telemetry, we decide that they said no only to a specific subset of telemetry, and not to all of it, and we turn some part of telemetry on again.

It's a bit like those websites where you unsubscribe but they come up with a new mailing list and add you there -- because you only unsubscribed to the old one, the new one is different!

Happy to discuss this over a zoom call, if it helps (I may be missing some background on these decisions).

lib/datadog/core/telemetry/client.rb
```diff
@@ -74,13 +88,52 @@ def client_configuration_change!(changes)
...
+          @metric_queue.add_metric(namespace, name, value, tags, Metric::Rate)
+        end
+
+        def add_gauge_metric(namespace, name, value, tags)
+          return if !@metrics_enabled || forked?
+
+          @metric_queue.add_metric(namespace, name, value, tags, Metric::Gauge)
+        end
+
+        def add_distribution_metric(namespace, name, value, tags)
```
marcotc, line 109: I think we can remove the `add_` prefix here. It sounds more "ruby" that way.
What do you think?

anmarchenko:
I will close this PR and take over this feature as part of SDTEST-107 epic "Ruby telemetry". I will open PRs implementing metrics support one by one in smaller pieces.

PR #3242, opened 2023-11-07
Fix concurrent ruby future propagation without `active_trace`

TonyCTHsu:
**What does this PR do?**

Concurrent ruby propagation breaks with `NoMethodError` when there is no `active_trace` to propagates. 

Fix with conditionals.

### DIFF
--- a/spec/datadog/tracing/contrib/concurrent_ruby/integration_test_spec.rb
+++ b/spec/datadog/tracing/contrib/concurrent_ruby/integration_test_spec.rb
@@ -135,6 +135,18 @@ RSpec.describe 'ConcurrentRuby integration tests' do
           end
         end
       end
+
+      context 'when propagates without an active trace' do
+        it 'creates a root span' do
+          future = Concurrent::Promises.future do
+            tracer.trace('inner_span') {}
+          end
+
+          future.wait
+
+          expect(inner_span).to be_root_span
+        end
+      end
     end
   end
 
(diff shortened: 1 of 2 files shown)


spec/datadog/tracing/contrib/concurrent_ruby/integration_test_spec.rb
```diff
@@ -135,6 +135,18 @@
           end
         end
       end
+
+      context 'when propagates without an active trace' do
+        it 'creates a root span' do
```
lloeki, line 140: Fix looks good but that sounds weird. I would expect:

- 'does nothing'
- 'does not raise'
- 'calls executor'

spec/datadog/tracing/contrib/concurrent_ruby/integration_test_spec.rb
```diff
@@ -135,6 +135,18 @@
...
+          future = Concurrent::Promises.future do
+            tracer.trace('inner_span') {}
+          end
+
+          future.wait
+
+          expect(inner_span).to be_root_span
```
lloeki, line 147: I understand that it indirectly test that it did not crash (because if it crashed the test would blow up), but does that actually test that it didn't append a span to the trace?

marcotc (approved):
We can address @lloeki concerns, but the fix itself looks good.

TonyCTHsu:
I am merging this to dish out a release as soon as possible.

PR #3245, opened 2023-11-07
Setup clangd-based native dev env via rake native_dev:setup

AlexJF:
**What does this PR do?**
Add some helpers to facilitate development on the native extensions of `dd-trace-rb` using the [clangd language server](https://clangd.llvm.org/installation).

1. `rake native_dev:setup` task to generate the required `compile_commands.json` files through [bear](https://github.com/rizsotto/Bear) (following the [clangd getting started guide](https://clangd.llvm.org/installation#project-setup)).
2. `rake native_dev:clean` task to cleanup the files generated by the previous task (and integration into global `rake clean`).
3. Disable automatic clang formatting on the `ext/*` directories since the current source code format does not align with clang and it'll try to reformat whole files on every save.
4. Add a `.gitignore` specific to native extensions to ignore the `compilation-flags.json` files created from this setup and keep them out of source control.

**Motivation:**
Auto complete, navigation and diagnostic support during native extension development:

image

image

image

**Additional Notes:**

**How to test the change?**
1. Install [bear](https://github.com/rizsotto/Bear).
1. Run `rake native_dev:setup`.
2. Setup clangd LSP support in your IDE: [vscode](https://github.com/clangd/vscode-clangd) | [neovim](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#clangd) | [sublime](https://github.com/sublimelsp/LSP-clangd).
3. Check that completion and diagnostics work while editing files in `ext/`.

I tested it with `neovim` on my local setup.

**For Datadog employees:**


ivoanjo (commented):
Left a few notes! I'm curious about the choice to develop in `ext/` instead of in `tmp/` :thinking:

Rakefile
```diff
@@ -602,14 +602,62 @@ namespace :changelog do
...
+native_exts = []
+
+native_exts << Rake::ExtensionTask.new("ddtrace_profiling_native_extension.#{RUBY_VERSION}_#{RUBY_PLATFORM}") do |ext|
   ext.ext_dir = 'ext/ddtrace_profiling_native_extension'
 end
 
-Rake::ExtensionTask.new("ddtrace_profiling_loader.#{RUBY_VERSION}_#{RUBY_PLATFORM}") do |ext|
+native_exts << Rake::ExtensionTask.new("ddtrace_profiling_loader.#{RUBY_VERSION}_#{RUBY_PLATFORM}") do |ext|
   ext.ext_dir = 'ext/ddtrace_profiling_loader'
 end
```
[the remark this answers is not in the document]
AlexJF: I'd personally prefer to keep it. It looks less complex with the latest version and makes for full coverage and easy change if any other native extension appears in the future. But not a strong opinion.

Rakefile
```diff
@@ -602,14 +602,62 @@ namespace :changelog do
...
+    NATIVE_CLEAN << "#{ext.ext_dir}/extconf.h"
+    NATIVE_CLEAN << "#{ext.ext_dir}/mkmf.log"
+
+    compile_commands_task = file "#{ext.ext_dir}/compile_commands.json" => makefile_task do
+      generate_ext_compile_commands(ext)
+    end
+
+    NATIVE_CLEAN << compile_commands_task.name
+    compile_commands_task
+  end
```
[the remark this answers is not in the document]
AlexJF: We talked a bit off-hand and agreed that trying to keep all the intermediate steps to `tmp` and then just move the desired `compilation-flags.json` back to src and keeping those git-ignored seemed like a cleaner approach. Lastest commit implements this.

Rakefile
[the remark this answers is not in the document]
ivoanjo: Not a strong opinion either, so let's keep it!

Rakefile
```diff
@@ -602,14 +602,51 @@ namespace :changelog do
...
+
+  desc 'Setup dev environment for native extensions.'
+  task setup: compile_commands_tasks
+
+  desc 'Clean dev environment for native extensions.'
+  task :clean do
+    Rake::Cleaner.cleanup_files(NATIVE_CLEAN)
+  end
```
ivoanjo, line 645: Minor: Maybe this could be removed? E.g. do you think we'll often `native_dev:clean` instead of just `clean`?

Rakefile
```diff
@@ -602,14 +602,51 @@ namespace :changelog do
   end
 end
 
-Rake::ExtensionTask.new("ddtrace_profiling_native_extension.#{RUBY_VERSION}_#{RUBY_PLATFORM}") do |ext|
+native_exts = []
```
ivoanjo, line 605: Minor: I would suggest making this into a constant, to make it clear that it's in the global scope.

(Note that the array itself pointed to by the constant is still mutable, so everything else would work fine as-is)

Rakefile
```diff
@@ -602,14 +602,51 @@ namespace :changelog do
...
 end
 
+NATIVE_CLEAN = ::Rake::FileList[]
+namespace :native_dev do
+  compile_commands_tasks = native_exts.map do |ext|
+    tmp_dir_dd_native_dev = "#{ext.tmp_dir}/dd_native_dev"
+    directory tmp_dir_dd_native_dev
+    NATIVE_CLEAN << tmp_dir_dd_native_dev
+
+    compile_commands_task = file "#{ext.ext_dir}/compile_commands.json" => [tmp_dir_dd_native_dev] do |t|
```
ivoanjo, line 622: I guess one downside of this approach is that it works only for a specific Ruby version, so if we need to switch around, we'll need to re-run the rake task every time (and remember to do so).

On the other hand, I think that's fine ATM -- I'm guessing it may be non-trivial to make this per-ruby-version.

ivoanjo (approved):
Thanks for the great contribution, LGTM :+1:

Rakefile
AlexJF: Yeah I think doing some automation here may require either encoding the switch operation as a rake task or formally aligning on an official way to handle different Ruby versions (e.g. asdf) so we could use some kind of hook system.

Both options sound way too complicated for the benefit atm 😅

Rakefile
AlexJF: Needed to change things a little bit because rubocop doesn't really like unfrozen constants but the end result is what you suggested 👍

PR #3246, opened 2023-11-07
[PROF-8543] Ruby heap profiling PoC

AlexJF:
**What does this PR do?**
A PoC for having live heap object counting in the Ruby profiler.

**Motivation:**
We want to add heap profiling capabilities to the Ruby profiler.

**Additional Notes:**
* Lacks tests.
* Written by someone that hasn't touched C in a long time.
* Collected heap profiles are not aligned or synced with any GC data so they contain a lot of "noise" from objects that will go away in the next GC. This is something we plan to improve upon.

**How to test the change?**

1. Grab/write any Ruby app
2. Configure it to use dd-trace-rb from this PR
    1. Checkout the branch locally
    2. Add `gem "ddtrace", :path => "<path to checkedout dd-trace-rb>` to the Gemfile of the chosen app 
4. Run it with ddtracerb and the right envs like:
    ```
    export DD_PROFILING_ENABLED="true"
    export DD_PROFILING_EXPERIMENTAL_ALLOCATION_ENABLED="true"
    export DD_PROFILING_EXPERIMENTAL_HEAP_ENABLED="true"
    bundle exec ddtracerb exec ruby <path to main.rb>
    ```

Tested this locally along with backend and frontend PRs to support this end-to-end:
image

**For Datadog employees:**


ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+      .skip = false,
+  });
+}
+
+void start_heap_allocation_recording(heap_recorder* heap_recorder, VALUE new_obj, unsigned int weight, ddog_CharSlice *class_name) {
+  fprintf(stderr, "Started recording allocation of %lu with weight %u\n", new_obj, weight);
+  partial_heap_recording *active_recording = &heap_recorder->active_recording;
+  active_recording->obj = new_obj;
+  active_recording->weight = weight;
+  active_recording->class_name = class_name;
```
ivoanjo, line 280: Minor: I usually favor this style to fill up structures, since it has slightly less typing and has the advantage that anything you don't set gets set to zero, rather than being accidental garbage (insert snarky remark about being surprised in C about getting sane behavior) 

```suggestion
  heap_recorder->active_recording = (partial_heap_recording) {
    .obj = new_obj,
    .weight = weight,
    .class_name = class_name,
  }
```

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+    // Recording ended without having been started?
+    rb_raise(rb_eRuntimeError, "Ended a heap recording that was not started");
+    return;
+  }
+
+  int weight = active_recording->weight;
+
+  // From now on, mark active recording as invalid so we can short-circuit at any point and
+  // not end up with a still active recording. new_obj still holds the object for this recording
+  active_recording->obj = 0;
```
ivoanjo, line 298: Really minor: When dealing with `VALUE`s, I usually prefer to set them to `Qnil` explicitly, rather than using 0. More of a "I find it easier to think with the Ruby concept" kinda thing -- just sharing, up to you if you prefer zero ;)

Note that 0 is usually what `Qfalse` gets typedef'd to (see https://github.com/ruby/ruby/blob/master/include/ruby/internal/special_consts.h#L83 ).

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+
+  heap_stack *heap_stack = heap_stack_init(locations);
```
ivoanjo, line 300: I guess not for V1, but I'm thinking we can probably dedupe the locations at this step, even without acquiring the lock?

That is, if the lock in practice is protecting the reader (which is flushing the profile) from the writer (which is taking a sample), then it's ok if the writer accesses the hashtables, as long as it doesn't write to it.

(Of course we'd probably need to document that very clearly -- that on the flush path we would not be allowed to mutate the hashtables)

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+  int error = pthread_mutex_trylock(&heap_recorder->records_mutex);
+  if (!error) {
+    // We were able to get a lock to heap_records so lets flush any pending samples
+    // that might have been queued previously before adding this new one.
+    flush_queue(heap_recorder);
```
ivoanjo, line 305: Minor: More of a style thing, but the current structure reads a bit weird to me

```
error = ...
if (!error) {
  happy path steps
} else {
  error fallback
  return
}

more happy path steps
```

I'd suggest doing `if (error) { .... return; }` and keeping the happy path all together.

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+    } else {
+      ENFORCE_SUCCESS_GVL(error)
+    }
+    return;
+  }
+
+  // If we got this far, we got a write lock so we can commit the record
+  commit_allocation(heap_recorder, heap_stack, new_obj, weight);
+
+  pthread_mutex_unlock(&heap_recorder->records_mutex);
```
ivoanjo, line 320: Ahhh C :rofl:

```suggestion
  ENFORCE_SUCCESS_GVL(pthread_mutex_unlock(&heap_recorder->records_mutex));
```

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+// ==============
+heap_stack* heap_stack_init(ddog_prof_Slice_Location locations) {
+  heap_stack *stack = ruby_xcalloc(1, sizeof(heap_stack));
+  stack->frames = ruby_xcalloc(locations.len, sizeof(heap_frame));
+  stack->frames_len = locations.len;
+  for (uint64_t i = 0; i < locations.len; i++) {
+    const ddog_prof_Location *location = &locations.ptr[i];
+    heap_frame *frame = &stack->frames[i];
+    frame->name = ruby_strdup(location->function.name.ptr);
+    frame->filename = ruby_strdup(location->function.filename.ptr);
```
ivoanjo, line 433: There's a couple of assumptions here that I'm not 100% sure about:
* That `ptr` is always a null-terminated string: I think so... but... technically... we're getting this from a Ruby string, and I think it's possible for Ruby strings to not be null-terminated? Or at least I'm not entirely confident.
* That there's always a value in the `ptr`

When we feed the locations to libdatadog, we give it a pointer and a length (which I'm not sure is always non-zero) so neither of those problems would occur, but here since we're just using the pointers without the length I'm a bit more unsure.

(Note that it doesn't mean we need to change stuff here -- maybe we just test these things when building the locations so we don't have to care here)

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+  unsigned int weight;
+  heap_record *heap_record;
+} object_record;
+static object_record* object_record_init(VALUE, unsigned int, heap_record*);
+static void object_record_free(object_record*);
+
+typedef struct {
+  VALUE obj;
+  unsigned int weight;
+  ddog_CharSlice *class_name;
```
ivoanjo, line 60: We're not using the class name yet, but do note this pointer is pointing at something that's stack-allocated, so it's only valid while taking the sample (aka it'll need to be copied if we want to keep it for later).

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+  pthread_mutex_unlock(&heap_recorder->records_mutex);
+}
+
+void commit_allocation(heap_recorder *heap_recorder, heap_stack *heap_stack, VALUE obj, unsigned int weight) {
+  heap_record *heap_record = NULL;
+  if (!st_lookup(heap_recorder->heap_records, (st_data_t) heap_stack, (st_data_t*) &heap_record)) {
+    heap_record = heap_record_init(heap_stack);
+    if (st_insert(heap_recorder->heap_records, (st_data_t) heap_stack, (st_data_t) heap_record)) {
+      rb_raise(rb_eRuntimeError, "Duplicate heap stack tracking: %p", heap_stack);
+      return;
```
ivoanjo, line 185: I'm curious, when would this happen?

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+  st_index_t hash = seed;
+  for (uint64_t i = 0; i < stack->frames_len; i++) {
+    hash = heap_frame_hash(&stack->frames[i], hash);
+  }
+  return hash;
+}
+
+st_index_t heap_stack_hash_st(st_data_t key) {
+  return heap_stack_hash((heap_stack *) key, FNV1_32A_INIT);
+}
```
ivoanjo, line 489: Minor: I wonder if we should keep the hash somewhere, to avoid recomputing over and over (as it may get a bit expensive)

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+    heap_stack_free(heap_stack);
+  }
+
+  object_record *object_record = object_record_init(obj, weight, heap_record);
+  if (st_insert(heap_recorder->object_records, (st_data_t) obj, (st_data_t) object_record) != 0) {
+    // Object already tracked?
+    object_record_free(object_record);
+    rb_raise(rb_eRuntimeError, "Duplicate heap object tracking: %lu", obj);
+    return;
+  }
```
ivoanjo, line 200: If my suspicion that the free tracepoint may be skipped sometimes, we may see this one happening. :thinking:

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+}
+
+void commit_free(heap_recorder *heap_recorder, VALUE obj) {
+  st_data_t key = (st_data_t) obj;
+  object_record *object_record = NULL;
+  if (!st_delete(heap_recorder->object_records, (st_data_t*) &key, (st_data_t*) &object_record)) {
+    // Object not tracked?
+    rb_raise(rb_eRuntimeError, "Committing free of untracked object");
+    return;
+  }
```
ivoanjo, line 215: This is even weirder, since it gets called from `record_heap_free` and that one already looks up in the map and checks that there's something there, so it would be really weird to see it fail here.

Note that if it turns out to be the case that the `RUBY_INTERNAL_EVENT_FREEOBJ` tracepoint gets called during GC, this one will make the VM explode since we're not allowed to allocate during GC (and we're allocating an exception).

P.s.: No need for `return` after raise, since it'll longjmp out :)

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+  heap_recorder->queued_samples_len = 0;
+}
+
+static void enqueue_sample(heap_recorder *heap_recorder, sample new_sample) {
+  fprintf(stderr, "Enqueuing sample for %lu (weight=%u free=%i)\n", new_sample.obj, new_sample.weight, new_sample.free);
+  if (heap_recorder->queued_samples_len >= MAX_QUEUE_LIMIT) {
+    fprintf(stderr, "Dropping sample on the floor.\n");
+    return;
+  }
```
ivoanjo, line 249: This may be problematic if we're dropping a free. We could prioritize frees here (e.g. clobber allocations in the buffer), but we'd always hit a limit still.

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+      sample *queued_sample = &heap_recorder->queued_samples[i];
+      if (queued_sample->obj == obj) {
+        queued_sample->skip = true;
+        break;
+      }
+    }
+
+    // free of an untracked object, return early
+    return;
+  }
```
ivoanjo, line 340: I guess rare situation, but if for some reason the background thread really gets behind, it seems possible that object at location A is allocated, then freed, then location A is reused while the queue remains unflushed.

In such a case, this check be incorrect, since we'd skip the first allocation, but not the second one.

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,489 @@
...
+  int cmp = strcmp(f1->name, f2->name);
+  if (cmp != 0) {
+    return cmp;
+  }
+  cmp = strcmp(f1->filename, f2->filename);
+  if (cmp != 0) {
+    return cmp;
+  }
+  return (int) (f1->line - f2->line);
+}
```
ivoanjo, line 404: Minor: May be worth starting with the line first, since it's the cheapest check.

ivoanjo (commented):
Left a few more notes! :)

ext/ddtrace_profiling_native_extension/stack_recorder.h
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <datadog/profiling.h>
+#include <ruby.h>
```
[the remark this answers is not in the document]
AlexJF: Any LSP will complain about `VALUE` being undefined otherwise. It indeed isn't needed (at least at the moment) during compilation because all codepaths that include `stack_recorder.h` include `ruby.h` first. That being said, because of the `ifndef/define`, this extra include should have no impact on the compiled result.

ext/ddtrace_profiling_native_extension/heap_recorder.c
AlexJF: 👍 

Adding a `ruby_strndup` helper to deal with this for now.

ext/ddtrace_profiling_native_extension/heap_recorder.c
AlexJF: The ruby hash table implementation already keeps this around to avoid recomputations on existing items: https://github.com/ruby/ruby/blob/903b0931a116691386e71fa30fb1698bbd785853/st.c#L133-L137

So this is only called on an external key (which is used on external to st.c lookup/insertion/update/deletion) and these keys would almost always be new objects and thus not have hashes precomputed anyway so I think this is fine?

lib/datadog/core/configuration/settings.rb
```diff
@@ -334,6 +334,8 @@ def initialize(*_)
             #
             # @default `true` on Ruby 2.x and 3.1.4+, 3.2.3+ and 3.3.0+; `false` for Ruby 3.0 and unpatched Rubies.
             option :allocation_counting_enabled do |o|
+              o.type :bool
+              o.env 'DD_PROFILING_EXPERIMENTAL_ALLOCATION_ENABLED'
               o.default do
                 RUBY_VERSION.start_with?('2.') ||
                   (RUBY_VERSION.start_with?('3.1.') && RUBY_VERSION >= '3.1.4') ||
```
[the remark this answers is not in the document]
AlexJF: Done in latest commit :+1:

ext/ddtrace_profiling_native_extension/heap_recorder.c
ivoanjo: Ah, this is one of those `throw "Should never happen"` that in a few months actually shows up in prod :rofl:

ext/ddtrace_profiling_native_extension/heap_recorder.c
ivoanjo: Seems reasonable (and neat that this is done for us!)

AlexJF:
Replaced with

PR #3255, opened 2023-11-10
Propagation:Ensure W3C `tracestate` is always propagated

marcotc:
**What does this PR do?**

This PR ensures that the `tracestate` header received, even if not part of the first valid propagation style, is still propagated downstream. This respects [the spec for `tracestate`](https://www.w3.org/TR/2021/REC-trace-context-1-20211123/#mutating-the-tracestate-field), which states:
```
Vendors receiving a tracestate request header MUST send it to outgoing requests.
```

This is only an issue when W3C extraction style is proceeded by another style. And this is currently the default (`Datadog,tracecontext`).
Now with this change, when Datadog headers match, we will now continue searching for `tracecontext` headers to enrich the downstream propagation context.

**Additional Notes:**

This PR also introduces the `DD_TRACE_PROPAGATION_EXTRACT_FIRST` configuration, to allow users to restore the prior behavior of not inspecting any headers after the first one was extracted.

**How to test the change?**

All changes are covered by unit tests and system-tests (https://github.com/DataDog/system-tests/pull/1795).

**For Datadog employees:**


aliciascott (approved):
Looks good

lib/datadog/tracing/trace_digest.rb
```diff
@@ -141,6 +141,37 @@ def initialize(
 
         freeze
       end
+
+      # Creates a copy of this object, modifying the provided fields.
+      # @param field_value_pairs [Hash<String>] the fields to be overwritten
+      # @return [TraceDigest] returns a copy of this object with the `field_value_pairs` modified
+      def merge(field_value_pairs)
+        # DEV: Because we want to sometimes freeze the values provided to `TraceDigest`, it's best
```
ekump, line 149: non-blocking: Is there any reason to be concerned about performance here?

ekump (approved):

lib/datadog/tracing/trace_digest.rb
marcotc: It's not really performance, but more about maintainability.
This code here would have to be duplicated here: https://github.com/DataDog/dd-trace-rb/blob/8b6432a4cd51b9e59f6fef596d0ba3d0a44197c8/lib/datadog/tracing/trace_digest.rb#L122-L140
But it has to decide which fields can be used as is, or which ones need to be `dup.freeze`.

This is just to make sure we don't duplicate the logic and forget to copy some values but not others.

PR #3266, opened 2023-11-20
set 128-bit trace_id to true by default

ekump:
**What does this PR do?**
Changes the default configuration for 128-bit trace_id generation to `true`. 

**Motivation:**
Part of Datadog's migration to W3C/128-bit tracers for OTEL compatibility. 

**Additional Notes:**
When using 128-bit IDs internally, we propagate the 64-bit version in `x-datadog-tags`, which we do not do for 64-bit IDs. Some unit tests had to be updated to reflect this. 

**How to test the change?**
Since this changes default behavior you just have to use the tracer with default settings to see the change. 

**For Datadog employees:**


spec/datadog/tracing/distributed/datadog_spec.rb
```diff
@@ -103,6 +109,8 @@
         end
 
         context 'nil' do
+          include_context 'with 128-bit trace_id generation disabled'
```
TonyCTHsu, line 112: I would prefer making examples more explicit by repeating 
```
before do
  Datadog.configure { |c| c.tracing.trace_id_128_bit_generation_enabled = false }
end
```
Instead of adding an indirection
```
include_context 'with 128-bit trace_id generation disabled'
```

lib/datadog/tracing/distributed/datadog.rb
```diff
@@ -46,11 +46,18 @@ def inject!(digest, data)
           data[@sampling_priority_key] = digest.trace_sampling_priority.to_s if digest.trace_sampling_priority
           data[@origin_key] = digest.trace_origin.to_s if digest.trace_origin
 
-          build_tags(digest).tap do |tags|
-            inject_tags!(tags, data) unless tags.empty?
+          begin
```
TonyCTHsu, line 49: Is there any reason for this change?

TonyCTHsu (commented):
Generally fine!

spec/datadog/opentelemetry_spec.rb
```diff
@@ -339,18 +339,21 @@
...
         subject(:inject) { ::OpenTelemetry.propagation.inject(carrier) }
         let(:carrier) { {} }
-
+        let(:headers) do
+          {
+            'x-datadog-parent-id' => Datadog::Tracing.active_span.id.to_s,
+            'x-datadog-sampling-priority' => '1',
+            'x-datadog-tags' => '_dd.p.dm=-0,_dd.p.tid=' +
+              Datadog::Tracing::Utils::TraceId.to_high_order(Datadog::Tracing.active_trace.id).to_s(16),
+            'x-datadog-trace-id' => Datadog::Tracing::Utils::TraceId.to_low_order(Datadog::Tracing.active_trace.id).to_s,
```
ekump, line 348: @marcotc I added a [helper for this](https://github.com/DataDog/dd-trace-rb/pull/3266/files#diff-e7ca68de5426a8ab8980827a9671f1b74ffe6c5b6d3a91eaf0201be601e407abR141) but it's scoped to `Tracing::Contrib` so I can't / shouldn't use it outside of that.

spec/datadog/opentelemetry_spec.rb
```diff
@@ -339,18 +339,21 @@
       describe '#inject' do
         subject(:inject) { ::OpenTelemetry.propagation.inject(carrier) }
         let(:carrier) { {} }
-
+        let(:headers) do
```
TonyCTHsu, line 342: This is fine, but as a reminder. `let` would cache the result when first invoked, this could potentially cause stale values within a test life cycle if not paying attention, especially the data depends on a variable `Datadog::Tracing.active_trace` that changes with context.

I recommend 

```
def headers
  {
    ...
  }
end
```

drichards-87 (approved):

lib/datadog/tracing/distributed/datadog.rb
ekump: Closing the loop on this: Spoke with Tony and we decided to revert this and update the test instead. The likelihood of tags not being a hash is extremely unlikely, so we shouldn't be generating any exceptions in `build_tags`.

spec/datadog/opentelemetry_spec.rb
ekump: Good callout, updated the test.

spec/datadog/tracing/distributed/datadog_spec.rb
```diff
@@ -185,7 +195,7 @@
           end
 
           context 'with invalid tags' do
-            let(:tags) { 'not_a_tag_hash' }
+            let(:tags) { { 'key with=spaces' => 'value' } }
```
TonyCTHsu, line 198: 👍

TonyCTHsu (approved):

spec/datadog/tracing/contrib/support/tracer_helpers.rb
```diff
@@ -136,5 +136,16 @@ def use_real_tracer!
...
+    # Wraps call to Tracing::Utils::TraceId.to_low_order for better test readability
+    def low_order_trace_id(trace_id)
+      Datadog::Tracing::Utils::TraceId.to_low_order(trace_id)
+    end
+
+    # Wraps call to Tracing::Utils::TraceId.to_high_order and converts to hex
+    # for better test readability
+    def high_order_hex_trace_id(trace_id)
+      Datadog::Tracing::Utils::TraceId.to_high_order(trace_id).to_s(16)
+    end
```
marcotc, line 149: I think these helpers are related to tracing, and not specifically contrib, as the 128-bit trace id generation is a part of core tracing.

I suggest moving these helpers to https://github.com/DataDog/dd-trace-rb/blob/master/spec/support/tracer_helpers.rb instead.

spec/datadog/tracing/distributed/datadog_spec.rb
marcotc: I agree here, given `c.tracing.trace_id_128_bit_generation_enabled = false` reads pretty much the same as `with 128-bit trace_id generation disabled`.

marcotc (approved):

spec/datadog/tracing/distributed/datadog_spec.rb
ekump: I disagree. I think the shared_context is a bit clearer in communicating intentions with the tests as well as removes some clutter. But since the actual context is so small I don't think it's the end of the world to revert this.

PR #3267, opened 2023-11-20
Tracing: Add b3 for default extraction

TonyCTHsu:
**What does this PR do?**

Add back b3 extraction which was removed from https://github.com/DataDog/dd-trace-rb/pull/3248

**Motivation:**

This would prevent breaking changes for user application while upgrading.

**For Datadog employees:**


marcotc (approved):

marcotc:
I think we forgot to update the documentation: https://github.com/DataDog/dd-trace-rb/blob/749020950eb5b806f9b37e6af47a6312b3c4cee1/docs/GettingStarted.md?plain=1#L2068

PR #3273, opened 2023-11-22
Convert OTel semantic convention to Datadog's

marcotc:
**What does this PR do?**

This PR parses the relevant [OpenTelemetry semantic conventions for tracing](https://opentelemetry.io/docs/specs/semconv/general/trace/) and converts those values into appropriate Datadog APM values.

Most changes come from parsing OTel Span attributes to derive a Datadog Span operation name.
For example, from [Semantic Conventions for Database Client Calls](https://opentelemetry.io/docs/specs/semconv/database/database-spans/#connection-level-attributes): 
If an OTel span has the attribute `db.system` and the Span kind `:client`,
then the respective Datadog Span will have the operation name `"#{otel_span.attributes['db.system']}.query"`.

This PR also introduces a few reserved attributes that can be set on an OTel Span to modify a Datadog Span field, for example `resource.name` can be set to modify the Datadog Span's resource name.

There are also a few fixes to attribute parsing, for cross-language consistency.

**Motivation:**

OpenTelemetry consistent, both with the OpenTelemetry semantic conventions and with other Datadog APM languages.

**How to test the change?**

There are unit test and system-tests for this change: https://github.com/DataDog/system-tests/pull/1456

**For Datadog employees:**


lib/datadog/opentelemetry/sdk/trace/span.rb
```diff
@@ -36,6 +36,78 @@ def status=(s)
...
+          else
+            [[key, value]]
+          end
+        end
+
+        # Create a meaningful Datadog operation name from the OpenTelemetry
+        # semantic convention for span kind and span attributes.
+        # @see https://opentelemetry.io/docs/specs/semconv/general/trace/
+
+        # rubocop:disable Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
```
TonyCTHsu, line 59: Perhaps we could revisit this later to break this method into smaller pieces.

lib/datadog/opentelemetry/sdk/span_processor.rb
```diff
@@ -90,19 +92,35 @@ def start_datadog_span(span)
...
 
-          name = attributes.key?('operation.name') ? attributes['operation.name'] : span.name
-          kwargs[:resource] = attributes['resource.name'] if attributes.key?('resource.name')
+          kwargs[:resource] = attributes.key?('resource.name') ? attributes['resource.name'] : span.name
           kwargs[:service] = attributes['service.name'] if attributes.key?('service.name')
           kwargs[:type] = attributes['span.type'] if attributes.key?('span.type')
 
+          attributes.reject! { |key, _| OpenTelemetry::Trace::Span::DATADOG_SPAN_ATTRIBUTE_OVERRIDES.include?(key) }
+
+          # DEV: There's no `flat_map!`; we have to split it into two operations
```
TonyCTHsu, line 116: For later: Would it be possible to complete this with a single operation with `reduce` or `each_with_object`?

TonyCTHsu (approved):

lib/datadog/opentelemetry/sdk/span_processor.rb
marcotc: This is an interesting line because I built it to support arbitrarily nested arrays, but that's actually not required for OTel, so I might simplify it later. But I wasn't able to in time for this PR.

PR #3281, opened 2023-11-24
[PROF-8667] Heap Profiling - Part 1 - Setup

AlexJF:
**What does this PR do?**

This PR paves the way for the introduction of heap profiling functionality by:
* Introduction of a setting for controlling heap profiling (`DD_PROFILING_EXPERIMENTAL_HEAP_ENABLED`, false by default).
* Refactoring of settings related to allocation profiling (`DD_PROFILING_EXPERIMENTAL_ALLOCATION_ENABLED` and `DD_PROFILING_EXPERIMENTAL_ALLOCATION_SAMPLE_RATE`) and improving the warnings on broken rubies.
  * As a result of this refactoring, allocation counting is now tied with allocation profiling and can no longer be enabled stand-alone.
* Introduction of a heap recorder component (with noop implementation for now) in the native profiling extension and plugging it in on top of the existing allocation profiling functionality.
* Interaction with the heap recorder component to collect new `heap-live-samples` data at profile serialization time.
* The necessary tests to have coverage for this functionality (some marked as pending until the proper implementation is added)

Future PRs will gradually build on the heap recorder implementation to actually enable heap data collection.

**Motivation:**
Start adding heap profiling capabilities in easy-to-review chunks.

**Additional Notes:**

**How to test the change?**
The only visible result of this PR is the inclusion of `heap-live-samples` values in the resulting Ruby profiles when an app is executed with:

```
```

**For Datadog employees:**


lib/datadog/profiling/component.rb
```diff
@@ -110,6 +116,52 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
...
+            'experimental, not recommended, and will increase overhead!'
+          )
+        end
+
+        if allocation_profiling_enabled && !Ext.allocation_sampling_supported?
+          Datadog.logger.warn(
+            "Current Ruby version (#{RUBY_VERSION}) does not officially support allocation profiling but it was " \
+            'requested. There may be unexpected problems during execution.'
+          )
+        end
```
ivoanjo, line 144: I'm thinking: the list of issues with the tracepoint basically could be grouped into

a) It breaks if you're using Ractors
b) It breaks if you're on Ruby 3.2.0 to 3.2.2

I know this is not what we do today, but what do you think of maybe:

1. Never allowing the feature to be enabled on Ruby 3.2.0 to 3.2.2, even if enabled. "Sorry, not supported, please use 3.1.x/3.2.3 or 3.3.0+"? (Caveat: 3.2.3 is actually not out, but the backport is in the branch, so we'd be excluding the latest stable Ruby for now but 3.3 will be out in December so maybe that's ok-ish for the state of this feature?)
2. Then we're free to be more exact in this error message for a) and say that due to VM bugs, if you're using this feature together with Ractors, expect issues, including crashes.

lib/datadog/profiling/component.rb
```diff
@@ -110,6 +116,52 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
         end
       end
 
+      private_class_method def self.get_allocation_sample_every(settings)
+        allocation_sample_rate = settings.profiling.advanced.experimental_allocation_sample_rate
+
+        if allocation_sample_rate <= 0
+          raise("Allocation sample rate must be a positive integer. Was #{allocation_sample_rate}")
+        end
```
ivoanjo, line 124: Minor: Here and below, consider raising an `ArgumentError`, as it's a bit more idiomatic?

ext/ddtrace_profiling_native_extension/collectors_thread_context.c
```diff
@@ -1203,6 +1203,8 @@ void thread_context_collector_sample_allocation(VALUE self_instance, unsigned in
     }
   }
 
+  track_obj_allocation(state->recorder_instance, new_object, sample_weight);
```
ivoanjo, line 1206: Minor: More of a naming thing, but `track_obj_allocation` may be a bit confusing during `sample_allocation`. Maybe `track_object`? `heap_recorder_track`? Something like that?

ivoanjo (commented):
I think this is looking great! Did a first pass but ran out of time, so here's the comments I got so far :)

ivoanjo (commented):
Here's another full detailed pass! I'm still missing a pass at the specs, will do so asap :)

spec/datadog/core/configuration/settings_spec.rb
```diff
@@ -497,39 +497,108 @@
         end
       end
 
-      describe '#allocation_counting_enabled' do
-        subject(:allocation_counting_enabled) { settings.profiling.advanced.allocation_counting_enabled }
```
ivoanjo, line 501: I think it may worth leaving a quick test in place, similar to what we have above:

```ruby
      describe '#allocation_counting_enabled=' do
        it 'logs a warning informing customers this no longer does anything' do
          expect(Datadog.logger).to receive(:warn).with(/no longer does anything/)
          settings.profiling.advanced.allocation_counting_enabled = true
        end
      end
```

In the spirit of, in the rare case that a customer is using this + we accidentally delete it, we'd still have a test that would prevent us from doing a backwards-incompatible change. 

(Of course for 2.0 we'll just rip-off the bandaid)

spec/datadog/core/configuration/settings_spec.rb
```diff
@@ -497,39 +497,108 @@
...
+      end
+
+      describe '#experimental_allocation_sample_rate=' do
+        it 'updates the #experimental_allocation_sample_rate setting' do
+          expect { settings.profiling.advanced.experimental_allocation_sample_rate = 100 }
+            .to change { settings.profiling.advanced.experimental_allocation_sample_rate }
+            .from(50)
+            .to(100)
         end
       end
```
ivoanjo, line 604: Not for this PR for sure! :arrow_up: These tests are boiler plate-y, but also not that hard to maintain. I've thought about maybe doing a bit of fancy code to reuse (e.g. test all boolean options trivially, but never went for it ;)

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
```diff
@@ -56,13 +60,16 @@
...
 
   describe '#start' do
+    let(:expected_worker_initialization_error) { nil }
+
     subject(:start) do
       cpu_and_wall_time_worker.start
       wait_until_running
     end
 
+    @skip_cleanup = false
```
ivoanjo, line 70: I'm not sure this is in the right scope, actually, since it's outside everything in the dsl. Might not be doing anything?

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
```diff
@@ -497,6 +501,62 @@
...
+        test_num_allocated_object.times { CpuAndWallTimeWorkerSpec::TestStruct.new }
+        allocation_line = __LINE__ - 1
+
+        cpu_and_wall_time_worker.stop
+
+        total_heap_sampled_for_test_struct = samples_for_thread(samples_from_pprof(recorder.serialize!), Thread.current)
+          .select { |s| s.locations.first.lineno == allocation_line && s.locations.first.path == __FILE__ }
+          .sum { |s| s.values[:'heap-live-samples'] }
+
+        expect(total_heap_sampled_for_test_struct).to eq test_num_allocated_object
```
ivoanjo, line 531: Minor: I'm curious, is the sum needed? I would expect the heap samples to aggregate to the same one sample with the total of all of them, since they have the same labels + stack. (like the allocation profiling samples do)

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
```diff
@@ -747,61 +807,83 @@
...
-        # the current thread were not affected by this change
+          before_allocations = described_class._native_allocation_count
+          100.times(&new_object)
+          after_allocations = described_class._native_allocation_count
 
-        expect(after_t1 - before_t1).to be 100
-        expect(after_allocations - before_allocations).to be < 10
+          expect(before_allocations).to be nil
+          expect(after_allocations).to be nil
+        end
```
ivoanjo, line 886: It does make sense to add a test for when this is disabled. But I don't quite get why we need to allocate anything at all as `_native_allocation_count` in this situation should return `nil` regardless of 0 or many objects being allocated. :eyes:

spec/datadog/profiling/spec_helper.rb
```diff
@@ -72,8 +72,14 @@ def samples_for_thread(samples, thread)
     samples.select { |sample| object_id_from(sample.labels.fetch(:'thread id')) == thread.object_id }
   end
 
-  def build_stack_recorder
-    Datadog::Profiling::StackRecorder.new(cpu_time_enabled: true, alloc_samples_enabled: true)
+  # We disable heap_sample collection by default in tests since it requires some extra mocking/
+  # setup for it to properly work.
+  def build_stack_recorder(heap_samples_enabled = false)
```
ivoanjo, line 77: Minor: It's nice to use keyword arguments for methods with boolean to make them self-documenting (e.g. `build_stack_recorder(heap_samples_enabled: false)` vs `build_stack_recorder(false)`).

Of course maybe your setup shows the arg names anyway, ehehe I should get it working as well (and it would be amazing if github did more of that......).

```suggestion
  def build_stack_recorder(heap_samples_enabled: false)
```

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -302,6 +336,74 @@ def sample_types_from(decoded_profile)
...
+    describe 'heap samples' do
+      let(:metric_values1) { { 'cpu-time' => 101, 'cpu-samples' => 1, 'wall-time' => 789, 'alloc-samples' => 10 } }
+      let(:metric_values2) { { 'cpu-time' => 102, 'cpu-samples' => 2, 'wall-time' => 790, 'alloc-samples' => 11 } }
+      let(:labels) { { 'label_a' => 'value_a', 'label_b' => 'value_b', 'state' => 'unknown' }.to_a }
+      let(:sample_rate) { 50 }
+
+      let(:a_string) { 'a beautiful string' }
+      let(:another_string) { 'a fearsome string' }
+      let(:an_array) { [1..10] }
+      let(:another_array) { [-10..-1] }
```
ivoanjo, line 348: Btw I'm not sure if you really want this or `(1..10).to_a`. `1..10` is a literal for a range object so these arrays only have a single object:

```ruby
[5] pry(main)> [1..10].size
=> 1
[6] pry(main)> [1..10][0].class
=> Range
```

spec/datadog/profiling/component_spec.rb
```diff
@@ -119,34 +120,146 @@
...
+        describe '#allocation_counting_enabled=' do
+          it 'updates the #allocation_counting_enabled setting' do
+            settings.profiling.advanced.allocation_counting_enabled = true
+
+            expect { settings.profiling.advanced.allocation_counting_enabled = false }
+              .to change { settings.profiling.advanced.allocation_counting_enabled }
+              .from(true)
+              .to(false)
+          end
+        end
```
ivoanjo, line 261: Wait, this doesn't belong here? :eyes: :rofl:

ivoanjo (commented):
Did a pass on the specs!

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -474,10 +551,55 @@ void record_endpoint(VALUE recorder_instance, uint64_t local_root_span_id, ddog_
...
+    grab_gvl_and_raise_ddogerr_and_drop("recording heap sample", &result.err);
+  }
+}
+
+static void build_heap_profile_without_gvl(struct stack_recorder_state *state, ddog_prof_Profile *profile) {
+  heap_recorder_iteration_context iteration_context = {
+    .state = state,
+    .profile = profile
+  };
+  heap_recorder_for_each_live_object(state->heap_recorder, add_heap_sample_to_active_profile_without_gvl, (void*) &iteration_context, false);
```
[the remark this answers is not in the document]
AlexJF: Yeah this was more of a way to flag that possibility rather than a necessary parameter (i.e. instead of naming it `for_each_live_object_no_gvl`, that parameter showed that was a possibility).

ext/ddtrace_profiling_native_extension/libdatadog_helpers.h
```diff
@@ -24,6 +24,12 @@ inline static VALUE get_error_details_and_drop(ddog_Error *error) {
   return result;
 }
 
+// Utility function to be able to easily raise a ddog_Error outside the GVL.
+// The message will look like 'Libstreaming error while <while_context>: <error.msg>'
```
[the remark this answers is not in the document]
AlexJF: Aahah 🤦 too much backend dev where that is the only lib* we deal with.

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,91 @@
...
+  partial_heap_recording active_recording;
+};
+
+// ==========================
+// Heap Recorder External API
+//
+// WARN: Except for ::heap_recorder_for_each_live_object, we always assume interaction with these APIs
+// happens under the GVL.
+//
+// ==========================
```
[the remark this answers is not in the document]
AlexJF: We do!

ext/ddtrace_profiling_native_extension/stack_recorder.c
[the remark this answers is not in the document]
ivoanjo: Aahh, I guess it works, I'd say maybe drop a comment on the method explaining that?

ext/ddtrace_profiling_native_extension/heap_recorder.c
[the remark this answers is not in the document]
ivoanjo: :see_no_evil:

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
AlexJF: Indeed! I wrote the more complicated stack_recorder version first and didn't realize the simpler context here.

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
AlexJF: Yeah this may be me being overly nitpicky about not relying on implementation details during tests.

The tests above validate that:
* `when CpuAndWallTimeWorker has not been started it is expected to be nil`
* `when CPUAndWallTimeWorker has been started and allocation profiling is enabled it returns the number of allocations between two calls of the mehod`

Treating the cpu and wall time worker as a black-box I can derive 2 logical possibilities from those tests:
* `nil` only returned when not started.
* `nil` always returned when allocation profiling disabled.

A test of the form: `when CPUAndWallTimeWorker has been started and allocation profiling is disabled it still returns nil after a bunch of allocations` completely disambiguates and covers the behaviour.

Doing just `when CPUAndWallTimeWorker has been started and allocation profiling is disabled it still returns nil` kinda leaves you wondering: ok but what happens after the first allocations are tracked? I feel like you'd need a separate `when CPUAndWallTimeWorker has been started and allocation profiling is enabled, it is set to 0 before we see allocations` or something.

spec/datadog/profiling/stack_recorder_spec.rb
AlexJF: Indeed! If it weren't for your eagle eyes, I would have only noticed it later when I added the size tests and maybe waste an hour trying to understand what was happening 😂

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
ivoanjo: I see. To be fair, it wouldn't be unreasonable to have a test that says "when enabled it returns >= 0" ;P

---

That said, if the intention is to test "even after allocations, it'll still be nil", it seems very reasonable. But then.... why check the before allocations as well? Does this test need to be this thorough?

P.s.: I just spotted that the test isn't quite correct for another reason: we're not starting cpu and wall time worker >_> (oops)

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -599,6 +721,9 @@ static VALUE _native_reset_after_fork(DDTRACE_UNUSED VALUE self, VALUE recorder_
   reset_profile(&state->slot_one_profile, /* start_time: */ NULL);
   reset_profile(&state->slot_two_profile, /* start_time: */ NULL);
 
+  // After a fork, all tracked live objects should still be alive in the child process so we purposefully DON'T
+  // reset the heap recorder. The live heap it is tracking is still alive after all.
```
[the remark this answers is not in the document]
AlexJF: I'll redo a bunch of things in this Part 1 to accommodate for the no-free-tracepoint approach. I'll also add a `heap_recorder_after_fork` function where we can centralize our decisions and thought process even if it ends up being a no-op in the end.

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -616,6 +741,17 @@ static VALUE _native_record_endpoint(DDTRACE_UNUSED VALUE _self, VALUE recorder_
   return Qtrue;
 }
 
+static VALUE _native_track_obj_allocation(DDTRACE_UNUSED VALUE _self, VALUE recorder_instance, VALUE new_obj, VALUE weight) {
+  ENFORCE_TYPE(weight, T_FIXNUM);
+  track_object(recorder_instance, new_obj, NUM2UINT(weight));
+  return Qtrue;
+}
```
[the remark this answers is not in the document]
AlexJF: Since this is supposed to only be using for testing, I thought the flexibility of being able to pass arbitrary values could be interesting to check some of the safety features/excludes we could add in the future when we start asking for the heap size of the object?

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
AlexJF: Ahah true that, also just spotted another weirdness, hierarchy-related. Test ends up being:

"_native_allocation_count when allocation profiling is enabled when CpuAndWallTimeWorker has been started when allocation profiling is disabled still returns nil after a bunch of allocations"

🙈 

I'll rework this.

ext/ddtrace_profiling_native_extension/stack_recorder.c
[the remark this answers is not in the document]
ivoanjo: Ah, yeah, that sounds good -- I was thinking "we may not support these things so perhaps it's better to not allow them here", but if the approach is to "let's actually make sure to have this tested" it's even better :)

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
ivoanjo: Yeah it's the usual problem of testing that things don't happen -- there may be multiple reasons for them not to happen, and sometimes it's not happening for the wrong reason ;)

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -474,10 +551,55 @@ void record_endpoint(VALUE recorder_instance, uint64_t local_root_span_id, ddog_
...
+      .locations = iteration_data.locations,
+      .values = (ddog_Slice_I64) {.ptr = metric_values, .len = context->state->enabled_values_count},
+      .labels = {0},
+    },
+    0
+  );
+
+  if (result.tag == DDOG_PROF_PROFILE_RESULT_ERR) {
+    grab_gvl_and_raise_ddogerr_and_drop("recording heap sample", &result.err);
+  }
```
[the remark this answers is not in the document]
AlexJF: Redid this keeping the out-of-GVL iteration and only raising outside of the records_mutex!

ivoanjo (approved):
I've left a last round of comments. Most of them are minor, and the others are more for later than for this PR.

I think this is in good shape to go in. Here sir, have my :+1:

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -0,0 +1,113 @@
...
+    heap_recorder *heap_recorder,
+    bool (*for_each_callback)(heap_recorder_iteration_data stack_data, void *extra_arg),
+    void *for_each_callback_extra_arg,
+    bool with_gvl) {
+  if (heap_recorder == NULL) {
+    return;
+  }
+
+  // TODO: Implement
+}
```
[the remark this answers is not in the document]
AlexJF: Yeah I was thinking in similar lines once I realized we'd need to flush/update first, while having the GVL to properly calculate sizes and then export that updated data outside the GVL to libdatadog.

Didn't want to commit to it here but indeed think it's worth implementing on the next PR 👍

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -302,6 +336,88 @@ def sample_types_from(decoded_profile)
...
+          sum_alloc_samples = 0
+
+          non_heap_samples.each do |s|
+            sum_cpu_time += s.values[:'cpu-time']
+            sum_alloc_samples += s.values[:'alloc-samples']
+          end
+
+          expect(sum_cpu_time).to be > 0
+          expect(sum_alloc_samples).to eq(@num_allocations * sample_rate)
+        end
```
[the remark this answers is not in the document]
AlexJF: I tend to agree with the intention in general. But in this case, because of the extra work needed to "mock" allocation tracking in these tests, this section is the only one where we actually have `heap_samples_enabled`.

So all the other tests in this suite that validate non-heap profiling do not cover `heap_samples_enabled=true` case which is also the only case where we do "out-of-band" sample writing that could, in theory, overwrite/mess with something that happens in the normal recording. The intention of this test is to serve as a canary for such a problem.

An alternative might be adding a positive test case with `heap_samples_enabled=true` to the existing test context?

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -472,10 +526,73 @@ void record_endpoint(VALUE recorder_instance, uint64_t local_root_span_id, ddog_
...
 
+#define MAX_LEN_HEAP_ITERATION_ERROR_MSG 256
+
+// Heap recorder iteration context allows us access to stack recorder state and profile being serialized
+// during iteration of heap recorder live objects.
+typedef struct heap_recorder_iteration_context {
+  struct stack_recorder_state *state;
+  ddog_prof_Profile *profile;
+  bool error;
+  char errorMsg[MAX_LEN_HEAP_ITERATION_ERROR_MSG];
```
ivoanjo, line 537: Minor: Suggest using snake case ;)

```suggestion
  char error_msg[MAX_LEN_HEAP_ITERATION_ERROR_MSG];
```

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -197,11 +203,11 @@ static VALUE _native_new(VALUE klass);
 static void initialize_slot_concurrency_control(struct stack_recorder_state *state);
 static void initialize_profiles(struct stack_recorder_state *state, ddog_prof_Slice_ValueType sample_types);
 static void stack_recorder_typed_data_free(void *data);
-static VALUE _native_initialize(DDTRACE_UNUSED VALUE _self, VALUE recorder_instance, VALUE cpu_time_enabled, VALUE alloc_samples_enabled);
+static VALUE _native_initialize(DDTRACE_UNUSED VALUE _self, VALUE recorder_instance, VALUE cpu_time_enabled, VALUE alloc_samples_enabled, VALUE heap_samples_enabled);
 static VALUE _native_serialize(VALUE self, VALUE recorder_instance);
 static VALUE ruby_time_from(ddog_Timespec ddprof_time);
 static void *call_serialize_without_gvl(void *call_args);
-static struct active_slot_pair sampler_lock_active_profile();
+static struct active_slot_pair sampler_lock_active_profile(struct stack_recorder_state *state);
```
ivoanjo, line 210: I just realized this was wrong before, and am slightly horrified the compiler didn't tell me about it.

ivoanjo (approved):
:+1: Did a final tiny pass, I think this is ready!

ext/ddtrace_profiling_native_extension/stack_recorder.c
AlexJF: Yeah the compiler sure is picky about what it decides to complain about 😂

ivoanjo:
YAY!

PR #3287, opened 2023-11-28
[PROF-8667] Heap Profiling - Part 2 - Heap Recorder

AlexJF:
**What does this PR do?**

This PR follows https://github.com/DataDog/dd-trace-rb/pull/3281 by actually implementing the heap recorder native module to support tracking the number (not size yet) of live heap objects.

**How does it work?**

The heap recorder essentially plugs into the allocation sampling mechanism and will record the object ids (via `Object::object_id`) and allocation stacktraces of all objects whose allocation we sampled.

At profile serialization time, we flush/update the heap recorder, at which point we make use of [`ObjectSpace::_id2ref`](https://github.com/ruby/ruby/blob/57748ef2a26cbdfe075347c6f7064eb419b4949a/gc.c#L4683) to try to obtain an object reference from our stored ids. This will fail only in these cases:
1. The id isn't valid (shouldn't happen since we record the id directly by calling `Object::object_id`.
2. The object the id refers to is no longer alive -> This leads us to clean up the associated recorded live object record and acts as a more robust alternative to listening to free tracepoints (which are easy to miss).
3. The VM is operating in multi-ractor mode and the object we sampled is not shareable between ractors -> We consciously choose not to support ractors initially for this feature so we'll also handle these as free.

For all ids for which we were still able to retrieve a valid object reference, we'll iterate through them and generate corresponding libdatadog samples with `heap-live-samples` values (properly scaled according to allocation sampling rate).

**NOTE**: This implementation relies on the new object id mechanism introduced in Ruby >= 2.7 which guarantees uniqueness (i.e. never re-used) and stickiness across eventual memory compaction. Prior to Ruby 2.7 object ids were related to memory locations and it'd thus be very easy to "miss" frees if the memory location got re-used for a different object. If that happened, we'd keep on reporting the original object as alive when it really wasn't.

**Motivation:**
Actually implement basic heap profiling (count only, no size yet).

**Additional Notes:**

**How to test the change?**
Non-zero `heap-live-samples` values should now be included in the resulting Ruby profiles when an app is executed with:

```
```

**For Datadog employees:**

### DIFF
--- a/ext/ddtrace_profiling_native_extension/heap_recorder.c
+++ b/ext/ddtrace_profiling_native_extension/heap_recorder.c
@@ -1,23 +1,129 @@
 #include "heap_recorder.h"
 #include <pthread.h>
 #include "ruby/st.h"
-#include "ruby/util.h"
 #include "ruby_helpers.h"
 #include <errno.h>
+
+#define MAX_FRAMES_LIMIT 10000
+
+// A compact representation of a stacktrace frame for a heap allocation.
+typedef struct {
+  char *name;
+  char *filename;
+  int64_t line;
+} heap_frame;
+static st_index_t heap_frame_hash(heap_frame*, st_index_t seed);
+
+// A compact representation of a stacktrace for a heap allocation.
+//
+// We could use a ddog_prof_Slice_Location instead but it has a lot of
+// unused fields. Because we have to keep these stacks around for at
+// least the lifetime of the objects allocated therein, we would be
+// incurring a non-negligible memory overhead for little purpose.
+typedef struct {
+  uint64_t frames_len;
+  heap_frame frames[];
+} heap_stack;
+static heap_stack* heap_stack_new(ddog_prof_Slice_Location);
+static void heap_stack_free(heap_stack*);
+static st_index_t heap_stack_hash(heap_stack*, st_index_t);
+
+enum heap_record_key_type {
+  HEAP_STACK,
+  LOCATION_SLICE
+};
+// This struct allows us to use two different types of stacks when
+// interacting with a heap_record hash.
+//
+// The idea is that we'll always want to use heap_stack-keys when
+// adding new entries to the hash since that's the compact stack
+// representation we rely on internally.
+//
+// However, when querying for an existing heap record, we'd save a
+// lot of allocations if we could query with the
+// ddog_prof_Slice_Location we receive in our external API.
+//
+// To allow this interchange, we need a union and need to ensure
+// that whatever shape of the union, the heap_record_key_cmp_st
+// and heap_record_hash_st functions return the same results for
+// equivalent stacktraces.
+typedef struct {
+  enum heap_record_key_type type;
+  union {
+    // key never owns this if set
+    heap_stack *heap_stack;
+    // key never owns this if set
+    ddog_prof_Slice_Location *location_slice;
+  };
+} heap_record_key;
+static heap_record_key* heap_record_key_new(heap_stack*);
+static void heap_record_key_free(heap_record_key*);
+static int heap_record_key_cmp_st(st_data_t, st_data_t);
+static st_index_t heap_record_key_hash_st(st_data_t);
+static const struct st_hash_type st_hash_type_heap_record_key = {
+    heap_record_key_cmp_st,
+    heap_record_key_hash_st,
+};
+
+// Need to implement these functions to support the location-slice based keys
+static st_index_t ddog_location_hash(ddog_prof_Location, st_index_t seed);
+static st_index_t ddog_location_slice_hash(ddog_prof_Slice_Location, st_index_t seed);
+
+// A heap record is used for deduping heap allocation stacktraces across multiple
+// objects sharing the same allocation location.
+typedef struct {
+  // How many objects are currently tracked by the heap recorder for this heap record.
+  uint64_t num_tracked_objects;
+  // stack is owned by the associated record and gets cleaned up alongside it
+  heap_stack *stack;
+} heap_record;
+static heap_record* heap_record_new(heap_stack*);
+static void heap_record_free(heap_record*);
+
+// An object record is used for storing data about currently tracked live objects
+typedef struct {
+  long obj_id;
+  heap_record *heap_record;
+  live_object_data object_data;
+} object_record;
+static object_record* object_record_new(long, heap_record*, live_object_data);
+static void object_record_free(object_record*);
 
 // Allows storing data passed to ::start_heap_allocation_recording to make it accessible to
 // ::end_heap_allocation_recording.
 //
-// obj != Qnil flags this struct as holding a valid partial heap recording.
+// obj_id != 0 flags this struct as holding a valid partial heap recording.
 typedef struct {
-  VALUE obj;
+  long obj_id;
   live_object_data object_data;
 } partial_heap_recording;
 
 struct heap_recorder {
+  // Map[key: heap_record_key*, record: heap_record*]
+  // NOTE: We always use heap_record_key.type == HEAP_STACK for storage but support lookups
+  // via heap_record_key.type == LOCATION_SLICE to allow for allocation-free fast-paths.
+  st_table *heap_records;
+
+  // Map[obj_id: long, record: object_record*]
+  st_table *object_records;
+
+  // Lock protecting writes to above record tables
+  pthread_mutex_t records_mutex;
+
   // Data for a heap recording that was started but not yet ended
   partial_heap_recording active_recording;
+
+  // Reusable location array, implementing a flyweight pattern for things like iteration.
+  ddog_prof_Location *reusable_locations;
 };
+static heap_record* get_or_create_heap_record(heap_recorder*, ddog_prof_Slice_Location);
+static void cleanup_heap_record_if_unused(heap_recorder*, heap_record*);
+static int st_heap_record_entry_free(st_data_t, st_data_t, st_data_t);
+static int st_object_record_entry_free(st_data_t, st_data_t, st_data_t);
+static int st_object_record_entry_free_if_invalid(st_data_t, st_data_t, st_data_t);
+static int st_object_records_iterate(st_data_t, st_data_t, st_data_t);
+static int update_object_record_entry(st_data_t*, st_data_t*, st_data_t, int);
+static void commit_allocation(heap_recorder*, heap_record*, long, live_object_data);
 
 // ==========================
 // Heap Recorder External API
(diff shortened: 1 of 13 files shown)


ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -491,6 +459,10 @@ void object_record_free(object_record *record) {
   ruby_xfree(record);
 }
 
+void object_record_mark(object_record *record) {
+  rb_gc_mark(record->obj_id);
+}
```
ivoanjo, line 464: I... suspect this may not be needed in practice. Looking at the implementation of `rb_obj_id`/`cached_object_id` the id is a regular Ruby integer that gets incremented by some number every time a new id is needed. This means this integer starts life as an immediate (e.g. not something allocated on the heap), and only if it gets too big to fit in an immediate, will it get promoted to a heap-allocated `Bignum`.

Thus, I'm thinking one potential optimization might be to check that the object id we get is indeed an immediate, and in that case we don't need to care about marking.

If it ever gets to bignum, we could either start doing the marking in such a situation (so we could have a flag that's "seen a bignum" to control the marking), or we could just not support that for now and stop heap profiling if we ever get to that -- I'm not entirely sure the exact number, but let's assume it's somewhere around 2^60 -- that's a lot of objects allocated and a very long time for an app to live.

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -332,41 +352,25 @@ typedef struct {
...
-    //
-    // 3. An unknown thing happened.
-    //
-    // Except for #3 where we're kinda screwed anyway, dropping the existing object record is
-    // acceptable behaviour and is equivalent to treating this allocation as a combined
-    // free+allocation.
-    VALUE obj = (VALUE) (*key);
-    object_record *existing_record = (object_record*) (*value);
-    commit_free(update_data->heap_recorder, obj, existing_record);
+    rb_raise(rb_eRuntimeError, "Object ids are supposed to be unique. We got 2 allocation recordings with the same id");
```
ivoanjo, line 355: I've mentioned it on slack, but this will break for Ruby <= 2.6. I think for now if needed we can disable heap profiling for such Rubies -- I think it makes sense to have a good implementation for modern Rubies, and we can later do a pass on older ones if there's enough interest. (E.g. rather than trying to do a one size fits all)

ext/ddtrace_profiling_native_extension/ruby_helpers.c
```diff
@@ -118,3 +118,39 @@ char* ruby_strndup(const char *str, size_t size) {
 
   return dup;
 }
+
+VALUE _id2ref(VALUE obj_id) {
+  // Call ::ObjectSpace._id2ref natively. It will raise if the id is no longer valid
+  VALUE module_object_space = rb_define_module("ObjectSpace");
```
ivoanjo, line 124: Minor: If we do end up calling `_id2ref` through `rb_funcall`, we can cache this somewhere

ivoanjo (commented):
Left a few notes, but this is a pretty interesting alternative!

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -29,43 +135,80 @@ struct heap_recorder {
...
   }
 
+  VALUE ruby_obj_id = rb_obj_id(new_obj);
+  if (!FIXNUM_P(ruby_obj_id)) {
+    rb_raise(rb_eRuntimeError, "Detected a bignum object id. These are not supported by heap profiling.");
+  }
+
   heap_recorder->active_recording = (partial_heap_recording) {
-    .obj = new_obj,
+    .obj_id = FIX2LONG(ruby_obj_id),
```
ivoanjo, line 211: Really minor: Since we now that `ruby_obj_id` is a fixnum and thus an immediate, we could use it directly, without converting it to a long (which is just a shift anyway). That would allow us to still check for `Qnil` and whatnot in a few places.

ivoanjo (commented):
Here's a pass on everything but the `heap_recorder.c` -- I was halfway there still. Will come back tomorrow to finish it :)

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -110,8 +280,426 @@ void heap_recorder_for_each_live_object(
...
+st_index_t char_slice_hash(ddog_CharSlice char_slice, st_index_t seed) {
+  return st_hash(char_slice.ptr, char_slice.len, seed);
+}
+
+// WARN: Must be kept in-sync with ::heap_frame_hash
+st_index_t ddog_location_hash(ddog_prof_Location location, st_index_t seed) {
+  st_index_t hash = char_slice_hash(location.function.name, seed);
+  hash = char_slice_hash(location.function.filename, hash);
+  hash = st_hash(&location.line, sizeof(location.line), hash);
+  return hash;
```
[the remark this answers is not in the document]
AlexJF: Good questions/notes!

> Use `rb_hash_start(0)`

Yeah reading on murmur nomenclature I don't see a particular reason why one seed would be better than others outside of zero-padding collisions which we don't really care about here given the keys are not user-controllable. So maybe I'll simplify to not have that weird define around. Also note that `rb_hash_start` is not used in any of the `st` hashing at all (and anyway is currently implemented as an identity function).

> `rb_hash_uint` for the line

Yeah good idea.

> They use `rb_hash_end`

`st_hash` does that internally (by calling murmur_finish). Chaining `st_hashes` does lead to potentially too much finishing in that case so there may be some superfluous prime multiplication occuring there.

TL;DR: I'm not worried about correctness/collisions here but I'll add a TODO note to re-evaluate/look for performance opportunities here once we have a benchmarking suite for heap recording stuff.

ext/ddtrace_profiling_native_extension/heap_recorder.c
AlexJF: Ah right. It does bother me a bit from a "hygiene" perspective to store `VALUE` stuff around without GC marking them. It also makes it easy to forget/typo and start passing the actual object ref into those APIs. 

I personally don't think the ability to check for Qnil rather than 0 is a big enough pro to warrant it?

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -1,23 +1,129 @@
...
 #include "ruby_helpers.h"
 #include <errno.h>
 
+#define MAX_FRAMES_LIMIT 10000
+
+// A compact representation of a stacktrace frame for a heap allocation.
+typedef struct {
+  char *name;
+  char *filename;
+  int64_t line;
```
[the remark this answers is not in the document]
AlexJF: I had initially kept the typing from the ddog_prof_Location struct. 

I think what you say makes sense but going with a `int32_t` since even the ruby native methods we're using seem to operate on signed `ints` (despite seeming to use 0 as invalid/unknown markers). I don't think having 2M vs 4M lines of code in a Ruby source file makes enough of a difference anyway?

ext/ddtrace_profiling_native_extension/ruby_helpers.c
```diff
@@ -108,3 +108,61 @@ void raise_syserr(
...
+VALUE _id2ref(VALUE obj_id) {
+  if (module_object_space == Qnil) {
+    module_object_space = rb_define_module("ObjectSpace");
+  }
+
+  // Call ::ObjectSpace._id2ref natively. It will raise if the id is no longer valid
+  return rb_funcall(module_object_space, rb_intern("_id2ref"), 1, obj_id);
+}
+
+VALUE _id2ref_failure(DDTRACE_UNUSED VALUE _unused1, DDTRACE_UNUSED VALUE _unused2) {
```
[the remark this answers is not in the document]
ivoanjo: I think so -- I don't know what happened there :rofl:

ext/ddtrace_profiling_native_extension/heap_recorder.c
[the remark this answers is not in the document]
ivoanjo: Sounds good.

> TL;DR: I'm not worried about correctness/collisions here but I'll add a TODO note to re-evaluate/look for performance opportunities here once we have a benchmarking suite for heap recording stuff.

Yeah, I think it looks fine so I'm not sure it's worth spending a lot of time on it, but... the fact that getting it wrong is rather invisible is annoying so yeah.

ext/ddtrace_profiling_native_extension/heap_recorder.c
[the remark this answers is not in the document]
ivoanjo: Yeah, that seems fine!

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -110,8 +283,469 @@ void heap_recorder_for_each_live_object(
...
+// WARN: Must be kept in-sync with ::heap_frame_hash
+st_index_t ddog_location_hash(ddog_prof_Location location, st_index_t seed) {
+  st_index_t hash = char_slice_hash(location.function.name, seed);
+  hash = char_slice_hash(location.function.filename, hash);
+  // Convert ddog_prof line type to the same type we use for our heap_frames to
+  // ensure we have compatible hashes
+  int32_t line_as_int32 = (int32_t) location.line;
+  hash = st_hash(&line_as_int32, sizeof(line_as_int32), hash);
+  return hash;
+}
```
ivoanjo, line 568: Minor: Since each of these methods need to be kept in sync with the other, maybe worth moving the methods that need to be kept in sync together -- e.g. both string hashes together, then both frame/location hashes.

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -110,8 +283,469 @@ void heap_recorder_for_each_live_object(
...
+// Heap Stack API
+// ==============
+heap_stack* heap_stack_new(ddog_prof_Slice_Location locations) {
+  uint16_t frames_len = locations.len;
+  if (frames_len > MAX_FRAMES_LIMIT) {
+    // Ensure we never attempt to store a heap stack that goes over the MAX_FRAMES_LIMIT
+    // This should not be happening anyway since MAX_FRAMES_LIMIT should be shared with
+    // the stacktrace construction mechanism
+    frames_len = MAX_FRAMES_LIMIT;
+  }
```
ivoanjo, line 581: Minor: We can probably raise if this happens (we allocate stacks with the gvl acquired anyway?)

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -110,8 +283,469 @@ void heap_recorder_for_each_live_object(
...
+    return (int) (st1->frames_len - st2->frames_len);
+  }
+  for (uint64_t i = 0; i < st1->frames_len; i++) {
+    int cmp = heap_frame_cmp(&st1->frames[i], &st2->frames[i]);
+    if (cmp != 0) {
+      return cmp;
+    }
+  }
+  return 0;
+}
```
ivoanjo, line 617: Hey, is it me or is this method not used in the current version at all? (Maybe it's needed for a follow-up PR? Or maybe we can delete it?)

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -110,8 +283,469 @@ void heap_recorder_for_each_live_object(
...
+heap_record_key* heap_record_key_new(heap_stack *stack) {
+  heap_record_key *key = ruby_xmalloc(sizeof(heap_record_key));
+  key->type = HEAP_STACK;
+  key->heap_stack = stack;
+  return key;
+}
+
+void heap_record_key_free(heap_record_key *key) {
+  ruby_xfree(key);
+}
```
ivoanjo, line 649: Minor: It looks to me that maybe we could have the `heap_record_key` own the `stack` as well?

E.g. whenever a `heap_record` key gets malloc'd, we know it's because there's a new stack being tracked, and they will also get freed together, so we could make freeing the stack become the responsibility of the key and simplify this a bit I guess.

ivoanjo (approved):
Took me a while, but this looks great! Left a final set of comments/suggestions, but here sir, take my :+1:

ext/ddtrace_profiling_native_extension/heap_recorder.c
AlexJF: There is a kind of 1:1 mapping between a heap-allocated heap_record_key and a heap_record so whether `stack` is owned by key or record seems a bit equivalent from a "created together, freed together".

IMO, the more flexible nature of `heap_record_key` with its potential non-heap-allocated stacks or even not having a `heap_stack` at all, makes it slightly more confusing to give ownership to. Thus why I preferred to keep ownership to the record and not the key.

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -739,6 +741,36 @@ static VALUE _native_track_object(DDTRACE_UNUSED VALUE _self, VALUE recorder_ins
...
+          .name = char_slice_from_ruby_string(name),
+          .filename = char_slice_from_ruby_string(filename),
+        }
+    };
+  }
+  ddog_prof_Slice_Location ddog_locations = {
+    .len = locations_len,
+    .ptr = locations_arr,
+  };
+  heap_recorder_testonly_assert_hash_matches(ddog_locations);
```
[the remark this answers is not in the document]
AlexJF: Good catch! We are running in C99 so I think we can rely on variable length arrays here and keep everything sane?

ext/ddtrace_profiling_native_extension/heap_recorder.c
AlexJF: Right! Since we may be comparing any combination of stacks vs locations I added a new more flexible cmp function which made this one unused! 🙇

ext/ddtrace_profiling_native_extension/stack_recorder.c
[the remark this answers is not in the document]
ivoanjo: Yeah, I think that's reasonable.

To be honest in the past I've experimented with variable length arrays in ddtrace and did not fully grasp what happens on stack overflows. With what I know today, maybe what I saw was an interaction with Ruby's detection of stack overflows...? Need to retest again.

This to say, I'm not 100% confident of using them for the actual production code yet, but for the testing code I don't think it matters at all.

PR #3306, opened 2023-12-02
Permit Faraday on_error to be passed into Tracing.trace

alexevanczuk:
**What does this PR do?**
This PR allows clients to pass in an `on_error` to the `faraday` integration to change the behavior when `Faraday` raises an error due to an unavailable server.

**Motivation:**
Currently, the `faraday` integration permits an `error_handler` which allows someone to change the behavior when their application hits a server and gives them a 4xx or 5xx back.

However, say you run this in ruby:
```ruby
require 'faraday'
Faraday.new(url: 'http://localhost:1234').get('/')
```

Assuming you have nothing running on that port, you'll get this exception:
```
/Users/alexevanczuk/.rbenv/versions/3.2.2/lib/ruby/3.2.0/net/http.rb:1271:in `initialize': Failed to open TCP connection to localhost:1234 (Connection refused - connect(2) for "localhost" port 1234) (Faraday::ConnectionFailed)
```

The user may still want to override error behavior in these cases. In our case, our application will do one of the other:
1. It will `rescue Faraday::ConnectionFailed`, causing this error to not propagate through the application – yet Faraday still has an error span that we want to ignore.
2. It will not rescue this error, in which case the error propagates through and the rails middleware will attach an error to the span. However, Faraday also sets an error – so we end up with a duplicate error where ideally we just have one.

**Additional Notes:**
Let me know if you have any other ideas on the API. I tried to make it clear how this is different from `error_handler` and I didn't want to break any public API for folks, so I thought this made the most sense as a new setting.

**How to test the change?**
Automated tests

**For Datadog employees:**


alexevanczuk:
@urseberry Could you look at this one too? Thank you so much!

docs/GettingStarted.md
```diff
@@ -838,6 +838,7 @@ connection.get('/foo')
...
 | `split_by_domain`     |                                 | Uses the request domain as the service name when set to `true`.                                                                                                                             | `false`   |
 | `error_handler`       |                                 | A `Proc` that accepts a `response` parameter. If it evaluates to a *truthy* value, the trace span is marked as an error. By default only sets 5XX responses as errors.                      | `nil`     |
+| `on_error`      | `Proc`                  | While `error_handler` is a way to control how errors are set based on the response to a request, `on_error` is applied to other types of `Faraday` errors. For example, if the request fails because the server is not available, `on_error` will e called instead of `error_handler`. This configuration overrides error handling behavior when a span raises an error. Provided `span` and `error` as arguments. Sets error on the span by default. | `proc { |span, error| span.set_error(error) unless span.nil? }` |
```
alai97, line 841:
```suggestion
| `on_error`      | `Proc`                  | While `error_handler` is a way to control how errors are set based on the response to a request, `on_error` is applied to other types of `Faraday` errors. For example, if the request fails because the server is not available, `on_error` will be called instead of `error_handler`. This configuration overrides error handling behavior when a span raises an error. Provided `span` and `error` as arguments. Sets error on the span by default. | `proc { |span, error| span.set_error(error) unless span.nil? }` |
```

alai97 (approved):
Looks good for docs, one suggestion. Thanks for opening this PR, @alexevanczuk!

alexevanczuk:
Merged your suggestion @alai97 , let me know if ya need anything else from me!

lib/datadog/tracing/contrib/faraday/middleware.rb
```diff
@@ -24,7 +24,9 @@ def call(env)
             # Do this once to reduce expensive regex calls.
             request_options = build_request_options!(env)
 
-            Tracing.trace(Ext::SPAN_REQUEST) do |span, trace|
+            on_error = datadog_configuration[:error_handler]
```
marcotc, line 27: Did you mean to use 
```suggestion
            on_error = datadog_configuration[:on_error]
```
here instead?

alexevanczuk: Bumping this @marcotc !

marcotc: > Reason being:
> 
>     * `error_handler` is a faraday configuration option that has a different interface:
> 
> 
> > A Proc that accepts a response parameter. If it evaluates to a truthy value, the trace span is marked as an error. By default only sets 5XX responses as errors.
> > [From docs](https://docs.datadoghq.com/tracing/trace_collection/dd_libraries/ruby/#faraday)
> 
>     * `on_error` is a new configuration API for the faraday integration. It's interface is a proc that accepts the span and the error, as required by `Tracing.trace`.
> 
> 
> These two error handlers catch different categories of errors.

This sounds correct to me, which comes back to my original suggestion:
`on_error = datadog_configuration[:error_handler]` is "A `Proc` that accepts a `response` parameter. If it evaluates to a *truthy* value, the trace span is marked as an error. By default only sets 5XX responses as errors." but we are passing it to `Tracing.trace(Ext::SPAN_REQUEST, on_error: on_error)`, which expects a proc that receives a span and an error object.
This means that the configured `datadog_configuration[:error_handler]` proc, which likely looks like `->(response){...}` will actually receive `proc.call(span,error)` and will most likely fail.

alexevanczuk: Oh I see... that's my bad 🤦🏼

alexevanczuk: Not sure why my test passed though 🤔

alexevanczuk: @marcotc Could you help me with something?

- I want to add a test that ensures when there is a no-op `on_error` and we get a `Faraday::ConnectionFailed` but an `error_handler` that always returns `true`, that no error is registered (since this is an error with the `faraday` client rather than with the endpoint being called itself
- However, I'm having trouble getting my tests to run as expected
- I add a `raise` in `call` of `lib/datadog/tracing/contrib/faraday/middleware.rb` and run `bundle exec rake spec:all` and the raise is never hit – I'm not sure why the test is not running
- `bundle exec rspec path/to/file_spec.rb`, the canonical way to run rspec tests, does not work in this repo
- There appears to be no way to run a specific test, based on the `Rakefile`

Can you let me know if I'm missing something? Are these tests not actually running? Is there no way to run individual tests within this repo? Would be great to update the contributor docs to make this a bit clearer since the test configuration is non-standard (i.e. cannot run `bundle exec rspec` successfully).

Thanks for your help!

alexevanczuk: @marcotc Wanted to bump this!

alexevanczuk: @marcotc bump!

TonyCTHsu:
👋 @alexevanczuk 

I have merged a PR to standardize for `on_error` settings. https://github.com/DataDog/dd-trace-rb/pull/3341. 

This would bring a consistent experience for configuring instrumentation for our 2.0 release. 

What do you think?

alexevanczuk:
Interesting @TonyCTHsu , so it sounds like we are combining on error and error handler?? And since on error is a superset of error handler, on error becomes the API to implement the behavior previously implemented with error handler? 

Let me know if that's the right interpretation

It looks like this would probably work for us since we only need the more generic error handler to stop any errors from these integrations since they propagate to the general datadog trace anyways.

TonyCTHsu:
👋 @alexevanczuk , Thanks for getting back. 

>so it sounds like we are combining on error and error handler??

Yes, and no. Currently, we have a mix of usage for `on_error` and `error_handler`, that causes confusion and frustration. 

The changes from #3341 basically standardize the usage by 

1. Eliminating `error_handler` option
2. Expose `on_error` option which align with our public API, that **will be invoke when an exception is raised.** 
3. For http clients (faraday), it provides an `error_status_codes` option to allow customization with http status from headers. (I believe this is the most frequent usage for the current `error_handler`https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/faraday/middleware.rb#L79 that yields `Faraday::Env` to be evaluated  )

alexevanczuk:
I think that makes sense @TonyCTHsu .

Any thoughts on when 2.0 will be released? If not anytime soon, wondering if you're still okay with this merging? Would be great to remove this noise from our error signal in the meantime.

TonyCTHsu:
👋 @alexevanczuk , 

I have backported the changes for `faraday` from `2.0` branch and I am going to make it part of `1.20.0` release scheduled next Monday.

alexevanczuk:
Amazing, thanks @TonyCTHsu . In that case, it sounds like we're good to close this branch because it's incorporated in the release next Monday.

Thanks for doing this, I look forward to incorporating it!

TonyCTHsu:
👋 @alexevanczuk , I actually modify this branch with the backport changes....

TonyCTHsu:
👋 @alexevanczuk , `1.20.0` has been released! Give it a try 👍

PR #3310, opened 2023-12-07
Allow the dynamic sampling rate overhead target to be set

KJTsanaktsidis:
**What does this PR do?**
This PR adds an option ~`dynamic_sampling_rate_overhead_target_percentage`~ `overhead_target_percentage` which allows the dynamic sampling rate in the profiler to be controlled. 

If this is used to reduce the sampling rate, the flushing interval is increased to compensate, so that the size of the flushed pprof files is roughly the same.

**Motivation:**
The profiler has a sophisticated dynamic sampling rate mechanism to tune the sample collection rate in order to target a desired level of wall-clock overhead time. Unfortunately, the default value of 2% is too much for our application; we find a noticeable and unfortunately not acceptable impact on our application's p95/p99 response times.

**Additional Notes:**

**How to test the change?**

(Edit by @ivoanjo): Setting `DD_PROFILING_UPLOAD_PERIOD` to > 60 will emit less profiles, and setting `DD_PROFILING_OVERHEAD_TARGET_PERCENTAGE` will to something different will make the profiler sample less/more than usual.

**For Datadog employees:**


ivoanjo:
Nice! I've sent a few notes privately, but I think this is in great shape :D

AlexJF (approved):
LGTM. Was a bit worried by the

> If this is used to reduce the sampling rate, the flushing interval is increased to compensate, so that the size of the flushed pprof files is roughly the same.

Part of the PR description (worth removing given it's no longer true I think!) but I see this approach was changed after (what I assume to be) the initial interactions with @ivoanjo. 

It's not a big problem to change the timeouts but it makes it extremely hard at the moment to have good metric plots if you don't have a fixed interval that you can rollup over (you'd potentially get a lot of see-saw action in any graphs coming from profile metrics).

ivoanjo:
> It's not a big problem to change the timeouts but it makes it extremely hard at the moment to have good metric plots if you don't have a fixed interval that you can rollup over (you'd potentially get a lot of see-saw action in any graphs coming from profile metrics).

Good point -- I guess this is another good reason for why this setting is not quite recommended unless you have a very specific use-case.

ivoanjo:
Merging away! :tada:

ekump:
@KJTsanaktsidis This has been included in our latest release. Thank you for your contribution!

PR #3313, opened 2023-12-08
[NO-TICKET] Redesign GC profiling to add timeline support and reduce overhead

ivoanjo:
**What does this PR do?**

This PR redesigns the garbage collection profiling feature to add support for timeline and reduce its overhead.

Here's how it looks in timeline mode:

image

Here's how it looks in flamegraph mode:

image

A TL;DR description of how it works now is below:

1. Wiring it up

  * Ruby exposes native-level tracepoint events for when it starts/stops doing GC (`RUBY_INTERNAL_EVENT_GC_ENTER`/`RUBY_INTERNAL_EVENT_GC_EXIT`).
  * Inside the `CpuAndWallTimeWorker` we create a tracepoint that listens for these events and calls the `on_gc_start` and `on_gc_finish` events on the `ThreadContext` collector.
  * The `RUBY_INTERNAL_EVENT_GC_ENTER`/`RUBY_INTERNAL_EVENT_GC_EXIT` tracepoint events are emitted while Ruby is doing GC, which means we can't allocate anything. The code we use is thus built to not allocate anything.
  * We can't do all the work we want in `on_gc_finish` because we can't allocate anything, so that function returns a flag that tells the `CpuAndWallTimeWorker` to call `sample_after_gc` as soon as the GC finishes.

2. Tracking

  * When a thread starts doing GC (`on_gc_start`), we record its current cpu-time and wall-time.
  * When a thread stops doing GC (`on_gc_finish`), we record how much cpu-time and wall-time elapsed.
  * When `sample_after_gc` gets called, we turn the data recorded above + some metadata from Ruby into a sample event to represent GC as a fake Garbage Collection thread.

3. Overhead

  * To avoid impacting applications a lot, the Ruby GC works incrementally. That is, rather than doing a full GC (marking, sweeping) in a single step, Ruby breaks this into smaller steps.
  * We get told about every smaller step: we get a pair of `on_gc_start`/`on_gc_finish` calls for each step.
  * In the past, we always called `sample_after_gc` after every `on_gc_finish` to record the GC. This was a big source of overhead: in a sample application that allocates as much as possible in a loop I was able to observe more than 50k calls to `on_gc_start`/`on_gc_finish` per second, which obviously would represent a lot of overhead if we create one GC sample for each of these.
  * To avoid this overhead, we coalesce/accumulate the cpu/wall-time from multiple calls to `on_gc_start`/`on_gc_finish` and only create a sample to represent this data from time to time (default is every 10ms or every full GC cycle, whatever happens first).

Compared to the old implementation, here's what changed:

* Garbage collection events get attributed to a placeholder "Garbage Collection" thread
* We no longer record "Garbage Collection" samples on the thread that triggered GC. (And thus no longer include the stack of that thread when recording GC information). This enabled a nice cleanup of the code in the stack collector
* GCs across several threads can be included in the same sample
* Timeline support is included
* Extra metadata about the GC can be seen in the timeline
* Overhead is lowered, even during heavy periods of GC action

**Motivation:**

I worked on this as part of the Profiler R&D week. I've been wanting to go back and fix the problems with the gc profiling implementation for a long time :)

**Additional Notes:**

I'm opening this PR on top of @AlexJF's https://github.com/DataDog/dd-trace-rb/pull/3281 because otherwise both were going to conflict, given the changes to the stack recorder. The two PRs are otherwise completely independent.

I'll wait until https://github.com/DataDog/dd-trace-rb/pull/3281 is merged to master to merge this PR as well.

**How to test the change?**

Test coverage is included.

During development, this tiny benchmark was useful:

```ruby
require 'benchmark/ips'
require 'ddtrace'

puts RUBY_DESCRIPTION

Datadog.configure do |c|
  c.profiling.enabled = true
  c.profiling.advanced.allocation_counting_enabled = false
  c.telemetry.enabled = false
end

class Potato
  def initialize(val)
    @potato = val
  end
end

Benchmark.ips do |x|
  x.config(time: 10, warmup: 2)

  x.report("DD_PROFILING_FORCE_ENABLE_GC=#{ENV['DD_PROFILING_FORCE_ENABLE_GC']} Potato.new(10)", "Potato.new(10)")

  x.compare!
end
```

**For Datadog employees:**

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
@@ -555,27 +555,29 @@ bool thread_context_collector_on_gc_finish(VALUE self_instance) {
   // how often this happens -- see on_gc_start.
   if (thread_context == NULL) return false;
 
-  if (thread_context->gc_tracking.cpu_time_at_start_ns == INVALID_TIME &&
-    thread_context->gc_tracking.wall_time_at_start_ns == INVALID_TIME) {
+  long cpu_time_at_start_ns = thread_context->gc_tracking.cpu_time_at_start_ns;
+  long wall_time_at_start_ns = thread_context->gc_tracking.wall_time_at_start_ns;
+
+  if (cpu_time_at_start_ns == INVALID_TIME && wall_time_at_start_ns == INVALID_TIME) {
     // If this happened, it means that on_gc_start was either never called for the thread OR it was called but no thread
     // context existed at the time. The former can be the result of a bug, but since we can't distinguish them, we just
     // do nothing.
     return false;
   }
 
-  // Here we record the wall-time second and in on_gc_start we record it first to try to avoid having wall-time be slightly < cpu-time
-  long cpu_time_at_finish_ns = cpu_time_now_ns(thread_context);
-  long wall_time_at_finish_ns = monotonic_wall_time_now_ns(DO_NOT_RAISE_ON_FAILURE);
-
-  long gc_cpu_time_elapsed_ns = cpu_time_at_finish_ns - thread_context->gc_tracking.cpu_time_at_start_ns;
-  long gc_wall_time_elapsed_ns = wall_time_at_finish_ns - thread_context->gc_tracking.wall_time_at_start_ns;
-
   // Mark thread as no longer in GC
   thread_context->gc_tracking.cpu_time_at_start_ns = INVALID_TIME;
   thread_context->gc_tracking.wall_time_at_start_ns = INVALID_TIME;
 
+  // Here we record the wall-time second and in on_gc_start we record it first to try to avoid having wall-time be slightly < cpu-time
+  long cpu_time_at_finish_ns = cpu_time_now_ns(thread_context);
+  long wall_time_at_finish_ns = monotonic_wall_time_now_ns(DO_NOT_RAISE_ON_FAILURE);
+
   // If our end timestamp is not OK, we bail out
   if (wall_time_at_finish_ns == 0) return false;
+
+  long gc_cpu_time_elapsed_ns = cpu_time_at_finish_ns - cpu_time_at_start_ns;
+  long gc_wall_time_elapsed_ns = wall_time_at_finish_ns - wall_time_at_start_ns;
 
   // Wall-time can go backwards if the system clock gets changed (and we observed spurious jumps back on macOS as well)
   // so let's ensure we don't get negative values for time deltas.
--- a/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
@@ -190,7 +190,6 @@ static void trigger_sample_for_thread(
   VALUE stack_from_thread,
   struct per_thread_context *thread_context,
   sample_values values,
-  sample_type type,
   long current_monotonic_wall_time_ns,
   ddog_CharSlice *ruby_vm_type,
   ddog_CharSlice *class_name
@@ -496,7 +495,6 @@ void update_metrics_and_sample(
     stack_from_thread,
     thread_context,
     (sample_values) {.cpu_time_ns = cpu_time_elapsed_ns, .cpu_or_wall_samples = 1, .wall_time_ns = wall_time_elapsed_ns},
-    SAMPLE_REGULAR,
     current_monotonic_wall_time_ns,
     NULL,
     NULL
@@ -667,7 +665,6 @@ static void trigger_sample_for_thread(
   VALUE stack_from_thread, // This can be different when attributing profiler overhead using a different stack
   struct per_thread_context *thread_context,
   sample_values values,
-  sample_type type,
   long current_monotonic_wall_time_ns,
   // These two labels are only used for allocation profiling; @ivoanjo: may want to refactor this at some point?
   ddog_CharSlice *ruby_vm_type,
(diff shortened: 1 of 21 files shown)


ext/ddtrace_profiling_native_extension/collectors_thread_context.c
```diff
@@ -509,67 +528,90 @@ void thread_context_collector_on_gc_start(VALUE self_instance) {
...
+  // Here we record the wall-time second and in on_gc_start we record it first to try to avoid having wall-time be slightly < cpu-time
+  long cpu_time_at_finish_ns = cpu_time_now_ns(thread_context);
+  long wall_time_at_finish_ns = monotonic_wall_time_now_ns(DO_NOT_RAISE_ON_FAILURE);
+
+  long gc_cpu_time_elapsed_ns = cpu_time_at_finish_ns - thread_context->gc_tracking.cpu_time_at_start_ns;
+  long gc_wall_time_elapsed_ns = wall_time_at_finish_ns - thread_context->gc_tracking.wall_time_at_start_ns;
+
+  // Mark thread as no longer in GC
+  thread_context->gc_tracking.cpu_time_at_start_ns = INVALID_TIME;
+  thread_context->gc_tracking.wall_time_at_start_ns = INVALID_TIME;
```
AlexJF, line 575: Minor/styling nit: We could keep variables closer to their uses by moving these lines above the previous?

1. Mark thread as no longer in GC
2. CPU and wall time at finish
3. Return false if end timestamp not OK
4. Elapsed calculation.

ext/ddtrace_profiling_native_extension/collectors_thread_context.c
```diff
@@ -578,70 +620,41 @@ VALUE thread_context_collector_sample_after_gc(VALUE self_instance) {
...
+    state->sampling_buffer,
+    state->recorder_instance,
+    (sample_values) {
+      .cpu_time_ns = state->gc_tracking.accumulated_cpu_time_ns,
+      .cpu_or_wall_samples = 1,
+      .wall_time_ns = state->gc_tracking.accumulated_wall_time_ns,
+      .timeline_wall_time_ns = state->gc_tracking.accumulated_wall_time_ns,
+    },
+    (sample_labels) {.labels = slice_labels, .state_label = NULL, .end_timestamp_ns = end_timestamp_ns},
+    (ddog_prof_Function) {.name = DDOG_CHARSLICE_C(""), .filename = gc_placeholder}
```
AlexJF, line 651: Minor: We have this in at least 3 places and it's encoding a behaviour logic that "synthetic frames have an empty name and non-empty filename". Probably worth extracting to a libdatadog helper at some point?

AlexJF (approved):

ext/ddtrace_profiling_native_extension/collectors_thread_context.c
```diff
@@ -578,70 +620,41 @@ VALUE thread_context_collector_sample_after_gc(VALUE self_instance) {
...
-  if (sampled_any_thread) state->stats.gc_samples++;
+  ddog_CharSlice gc_placeholder = DDOG_CHARSLICE_C("Garbage Collection");
+  record_placeholder_stack(
+    state->sampling_buffer,
+    state->recorder_instance,
+    (sample_values) {
+      .cpu_time_ns = state->gc_tracking.accumulated_cpu_time_ns,
+      .cpu_or_wall_samples = 1,
+      .wall_time_ns = state->gc_tracking.accumulated_wall_time_ns,
+      .timeline_wall_time_ns = state->gc_tracking.accumulated_wall_time_ns,
```
AlexJF, line 648: I don't get the need for both? Is this a backend limitation? Why does it work fine for non-GC frames?

ext/ddtrace_profiling_native_extension/collectors_thread_context.c
ivoanjo: Fixed in 70479a5bfcbba7ecb53dccd197182589357ef759

ext/ddtrace_profiling_native_extension/collectors_thread_context.c
ivoanjo: This is a great question. I've pushed a note in 391db5278f61e1f25d7099e823e111ef96a1e9cc to address this:

```
      // This event gets both a regular cpu/wall-time duration, as a normal cpu/wall-time sample would, as well as a
      // timeline duration.
      // This is done to enable two use-cases:
      // * regular cpu/wall-time makes this event show up as a regular stack in the flamegraph
      // * the timeline duration is used when the event shows up in the timeline
```

In practice, we could change the backend to use get the cpu/wall-time timeline duration and use it for the timeline event, but doing it this way keeps us consistent with what other runtimes do for timeline (e.g. dotnet), and thus makes it easy to share code across them.

ext/ddtrace_profiling_native_extension/collectors_thread_context.c
ivoanjo: Fixed in 807a5194b44eaf0b525b7a3cd3b553d8d29baa23 by passing in only a single string to `record_placeholder_stack`. We may need to change it at some point, but for now this simplifies the code a bit ;)

marcotc (approved):
Awesome work, @ivoanjo!

ivoanjo:
Update: This PR was originally opened on top of #3281 but there were a number of changes to that branch, so I chose to rebase on top of master and force-push.

The rebase was clean other than a manual fix in `profiling/component.rb` + a new spec from master that needed updating ( bfdd1b73e95214e3a670eda0d444f8a5ac1d17cc ). No other changes were made.

ivoanjo:
Thanks y'all for the help and reviews. Merging away!

PR #3320, opened 2023-12-12
[PROF-8667] Split profiling tests into ractor and non-ractor suites.

AlexJF:
**What does this PR do?**

* Separate the profiling test suite from main to allow for profiling-specific customizations.
* Tag all tests that use `Ractor.new` with `ractors => true` tag.
* Introduce two spec tasks for profiling:
  * `profiling:main` - Run all tests EXCLUDING `ractors` tagged tests.
  * `profiling:ractors` - Run all `ractors` tagged tests.

**Motivation:**

As soon as you execute `Ractor.new`, the VM [transitions to a multi-ractor state](https://github.com/ruby/ruby/blob/b9f2c67106ac8277b0e83a5701a43d016d829131/ractor.c#L1877-L1892). This transition cannot be reversed and will change the behaviour of the VM. For example, [`ObjectSpace::_id2ref` will start to raise errors on unshareable objects](https://github.com/ruby/ruby/blob/b9f2c67106ac8277b0e83a5701a43d016d829131/gc.c#L4746C1-L4747). 

This essentially introduces an unavoidable cross-test side-effect that may make tests flaky depending on execution order. By containing tests that create actors to their own rspec run, we scope this side-effect to those tests that would always trigger it anyway.

**Additional Notes:**

**How to test the change?**

* `bundle exec rake spec:profiling`
* `bundle exec rake spec:profiling:main`
* `bundle exec rake spec:profiling:ractors`
* CI: 
  * `spec:profiling:main`: https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/12856/workflows/933470cd-2e63-4891-babb-9aca834e0660/jobs/483058/parallel-runs/4?filterBy=ALL
    image
  * `spec:profiling:ractors`: https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/12856/workflows/933470cd-2e63-4891-babb-9aca834e0660/jobs/483058/parallel-runs/5?filterBy=ALL
    image

**For Datadog employees:**

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -17,6 +17,12 @@ TEST_METADATA = {
   'appsec:main' => {
     '' => '✅ 2.1 / ✅ 2.2 / ✅ 2.3 / ✅ 2.4 / ✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ jruby'
   },
+  'profiling:main' => {
+    '' => '✅ 2.1 / ✅ 2.2 / ✅ 2.3 / ✅ 2.4 / ✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ jruby'
+  },
+  'profiling:ractors' => {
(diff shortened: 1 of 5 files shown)


ivoanjo:
Looks reasonable! The CI issue I believe it's because we're still trying to run the profiling stuff with `spec:main` but it's not getting compiled.

Perhaps we shouldn't run in `spec:main`? It's a bit of a misnomer to be honest (it's kinda "core + tracing + ...") but I don't think it would be a very hard edge, since anyone changing something that impacts profiling would still see it in CI, even if they ran `spec:main` and it said everything was OK when it wasn't.

ivoanjo (approved):
:+1: LGTM

PR #3322, opened 2023-12-12
Return string for `trace_id` and `span_id` from `Correlation::Identifier`

TonyCTHsu:
**2.0 Upgrade Guide notes**

When manually correlating logs with `Datadog::Tracing.correlation.trace_id` and `Datadog::Tracing.correlation.span_id`, make sure to check if your usage still compatible after returning string instead of integer.

🚨 Breaking changes: Other fields(except `service`, `env`, `version`, `trace_id` and `span_id`) removed. 

**What does this PR do?**

Return string for `trace_id` and `span_id` from `Correlation::Identifier`

Remove other fields that are not consider public API.

**For Datadog employees:**


lib/datadog/tracing/correlation.rb
```diff
@@ -48,7 +48,7 @@ def initialize(
           # Dup and freeze strings so they aren't modified by reference.
           @env = Core::Utils::SafeDup.frozen_dup(env || Datadog.configuration.env)
           @service = Core::Utils::SafeDup.frozen_dup(service || Datadog.configuration.service)
-          @span_id = span_id || 0
+          @span_id = (span_id || 0).to_s
```
marcotc, line 51: @TonyCTHsu I don't remember: why are we converting the numbers to strings?

ekump: @marcotc's question suggests that we should include that in the release notes somewhere. 

Apparently, Ruby integers can natively handle 128-bit numbers, so I agree. It's not clear why we need to convert them to strings.

TonyCTHsu: @marcotc @ekump The value returned by those methods are meant to be injected into log entries. 

Hence, returning string instead of integer provides finer control for implementing different encoding.

ekump: >The value returned by those methods are meant to be injected into log entries

Is that the only supported use for the return value? This is part of the public API, so should we be considering more generic usage? 

Generally speaking, I don't think public API methods should be too opinionated on what is done with their return values (within reason) and we should be returning values as the types that best reflect what they are. This contributes to an API being predictable and intuitive for an engineer to use. If I know `trace_ids` are 128-bit integers, and I see a method called `trace_id` I would assume/expect it to return an integer. 

It should be the responsibility of the calling method to convert it to a string and/or encode it a particular way unless we have a good reason to do otherwise.

marcotc: @ekump we have the method `#to_log_format` which is a helper method you use when you want to inject the correlation into your logs safely, that method is opinionated.

marcotc: Here an interesting bit, OTel uses 128-bit trace ids as well, but it represents them as [16-char strings](https://github.com/open-telemetry/opentelemetry-ruby/blob/ff9a7d3af31ad83ed6e07445a9e1609bec53d779/api/lib/opentelemetry/trace.rb#L24-L32). OTel always serializes the trace id as base 16.

lib/datadog/tracing/correlation.rb
```diff
@@ -48,7 +48,7 @@ def initialize(
           # Dup and freeze strings so they aren't modified by reference.
```
marcotc, line 48: Let's document that "this class is for usage with log correlation. To continue from a trace, users should use TraceDigest instead."

marcotc (approved):

ekump (approved):

lib/datadog/tracing/correlation.rb
```diff
@@ -21,42 +23,21 @@ class Identifier
           :env,
           :service,
           :span_id,
-          :span_name,
```
kloktech-mark, line 24: @TonyCTHsu  Hi there, may I ask the reason these fields are removed?  Reason I asked was because we use them in our semantic logger formatter to enrich the log so that we can view the logs by resource or service.

Is it possible to add them back or is there away for correlation to retrieve trace details so we can continue to enrich our log?

p-datadog: @kloktech-mark I think you will get a better answer to your question if you create a new issue (https://github.com/DataDog/dd-trace-rb/issues/new/choose).

PR #3328, opened 2023-12-13
[PROF-8667] Heap Profiling - Part 3 - Snapshot system

AlexJF:
**What does this PR do?**

This PR follows https://github.com/DataDog/dd-trace-rb/pull/3287 and reduces the conflict of serialization and sampling which was forcing us to potentially drop some allocation samples on the floor. 

**How does it work?**

Initially the assumption was that sampling and serialization operated on the same structures and thus needed to be protected by a mutex. Due to us never wanting to delay sampling code, we had noted the need for a queueing system to not be forced to drop data on the floor. This was the only realistic implementation back when we were relying on free tracepoints which could invalidate object_records at any point in time.

@ivoanjo noted that:
* With the current implementation where staleness checking does not rely on free tracepoints but is done synchronously and under GVL before iteration.
* `st_copy` operations seem to be extremely efficient, requiring at most 2 mallocs and 2 memcpy.
* We can guarantee a non-concurrent and strict ordering of `heap_recorder_prepare_iteration`, `heap_recorder_for_each_live_object` and `heap_recorder_finish_iteration` (since they all run from the same serialization thread and we have only one such thread).

Given the above, we can trivially do a copy/snapshot of `object_records` during `heap_recorder_prepare_iteration` and have the actual iteration operate on that snapshot without ever conflicting with any concurrent sampling that might happen in-between (`object_records_snapshot` is read-only after construction, mutations will only be recorded in `object_records` during any sampling that happens concurrently with serialization).

**Motivation:**
Never wanting to block the allocation tracepoint and not wanting to drop relevant samples on the floor.

**Additional Notes:**

**How to test the change?**
It's quite hard to manually or deterministically trigger the conditions whereupon serialization could conflict with sampling and which justified this entire endeavour. Hopefully the fact that all existing specs plus the new "long serialization simulation" specs can provide enough assurance?

**For Datadog employees:**

### DIFF
--- a/ext/ddtrace_profiling_native_extension/heap_recorder.c
+++ b/ext/ddtrace_profiling_native_extension/heap_recorder.c
@@ -1,23 +1,129 @@
 #include "heap_recorder.h"
 #include <pthread.h>
 #include "ruby/st.h"
-#include "ruby/util.h"
 #include "ruby_helpers.h"
 #include <errno.h>
+
+#define MAX_FRAMES_LIMIT 10000
+
+// A compact representation of a stacktrace frame for a heap allocation.
+typedef struct {
+  char *name;
+  char *filename;
+  int64_t line;
+} heap_frame;
+static st_index_t heap_frame_hash(heap_frame*, st_index_t seed);
+
+// A compact representation of a stacktrace for a heap allocation.
+//
+// We could use a ddog_prof_Slice_Location instead but it has a lot of
+// unused fields. Because we have to keep these stacks around for at
+// least the lifetime of the objects allocated therein, we would be
+// incurring a non-negligible memory overhead for little purpose.
+typedef struct {
+  uint64_t frames_len;
+  heap_frame frames[];
+} heap_stack;
+static heap_stack* heap_stack_new(ddog_prof_Slice_Location);
+static void heap_stack_free(heap_stack*);
+static st_index_t heap_stack_hash(heap_stack*, st_index_t);
+
+enum heap_record_key_type {
+  HEAP_STACK,
+  LOCATION_SLICE
+};
+// This struct allows us to use two different types of stacks when
+// interacting with a heap_record hash.
+//
+// The idea is that we'll always want to use heap_stack-keys when
+// adding new entries to the hash since that's the compact stack
+// representation we rely on internally.
+//
+// However, when querying for an existing heap record, we'd save a
+// lot of allocations if we could query with the
+// ddog_prof_Slice_Location we receive in our external API.
+//
+// To allow this interchange, we need a union and need to ensure
+// that whatever shape of the union, the heap_record_key_cmp_st
+// and heap_record_hash_st functions return the same results for
+// equivalent stacktraces.
+typedef struct {
+  enum heap_record_key_type type;
+  union {
+    // key never owns this if set
+    heap_stack *heap_stack;
+    // key never owns this if set
+    ddog_prof_Slice_Location *location_slice;
+  };
+} heap_record_key;
+static heap_record_key* heap_record_key_new(heap_stack*);
+static void heap_record_key_free(heap_record_key*);
+static int heap_record_key_cmp_st(st_data_t, st_data_t);
+static st_index_t heap_record_key_hash_st(st_data_t);
+static const struct st_hash_type st_hash_type_heap_record_key = {
+    heap_record_key_cmp_st,
+    heap_record_key_hash_st,
+};
+
+// Need to implement these functions to support the location-slice based keys
+static st_index_t ddog_location_hash(ddog_prof_Location, st_index_t seed);
+static st_index_t ddog_location_slice_hash(ddog_prof_Slice_Location, st_index_t seed);
+
+// A heap record is used for deduping heap allocation stacktraces across multiple
+// objects sharing the same allocation location.
+typedef struct {
+  // How many objects are currently tracked by the heap recorder for this heap record.
+  uint64_t num_tracked_objects;
+  // stack is owned by the associated record and gets cleaned up alongside it
+  heap_stack *stack;
+} heap_record;
+static heap_record* heap_record_new(heap_stack*);
+static void heap_record_free(heap_record*);
+
+// An object record is used for storing data about currently tracked live objects
+typedef struct {
+  long obj_id;
+  heap_record *heap_record;
+  live_object_data object_data;
+} object_record;
+static object_record* object_record_new(long, heap_record*, live_object_data);
+static void object_record_free(object_record*);
 
 // Allows storing data passed to ::start_heap_allocation_recording to make it accessible to
 // ::end_heap_allocation_recording.
 //
-// obj != Qnil flags this struct as holding a valid partial heap recording.
+// obj_id != 0 flags this struct as holding a valid partial heap recording.
 typedef struct {
-  VALUE obj;
+  long obj_id;
   live_object_data object_data;
 } partial_heap_recording;
 
 struct heap_recorder {
+  // Map[key: heap_record_key*, record: heap_record*]
+  // NOTE: We always use heap_record_key.type == HEAP_STACK for storage but support lookups
+  // via heap_record_key.type == LOCATION_SLICE to allow for allocation-free fast-paths.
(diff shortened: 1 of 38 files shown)


ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -101,6 +103,13 @@ typedef struct {
...
+typedef struct {
+  // Has ownership of this, needs to clean-it-up if not transferred.
+  heap_record *heap_record;
+  long obj_id;
+  live_object_data object_data;
+} uncommitted_sample;
+
 struct heap_recorder {
   // Map[key: heap_record_key*, record: heap_record*]
   // NOTE: We always use heap_record_key.type == HEAP_STACK for storage but support lookups
```
ivoanjo, line 115: Was already like this before, but while going through this PR, I realized that actually the comment on `records_mutex` is innacurate -- it actually is only really projecting the `object_records`, since for the `heap_records` we rely on the GVL instead right now.

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -101,6 +103,13 @@ typedef struct {
   live_object_data object_data;
 } partial_heap_recording;
 
+typedef struct {
+  // Has ownership of this, needs to clean-it-up if not transferred.
+  heap_record *heap_record;
+  long obj_id;
+  live_object_data object_data;
+} uncommitted_sample;
```
ivoanjo, line 111: To be honest, I've been starting at this PR for quite a while and thinking... is there a way we can avoid the additional complexity of the queue?

I'm starting to think that maybe there's an option. Looking at the implementation of `rb_st_copy` (which calls `st_replace`) and it seems like it would be blazing fast -- it has 2 malloc and 2 memcopys and it's done.

What if, instead of having to deal with the lock, we created a copy of the `object_records` at the end of `heap_recorder_flush` and then used that copy when calling `heap_recorder_for_each_live_object`?

This is all a bit pie in the sky, but the tradeoff of one copy of the `object_records` per flush would be no locking needed + no queueing needed.

That said, I may be missing something so I'll give a pass on this PR still and leave the final decision to you.

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -347,20 +347,23 @@ def sample_types_from(decoded_profile)
...
+
       before do
         allocations = [a_string, an_array, 'a fearsome string', [-10..-1], a_hash, { 'z' => -1, 'y' => '-2', 'x' => false }]
         @num_allocations = 0
         allocations.each_with_index do |obj, i|
-          # Heap sampling currently requires this 2-step process to first pass data about the allocated object...
-          described_class::Testing._native_track_object(stack_recorder, obj, sample_rate)
-          # ...and then pass data about the allocation stacktrace (with 2 distinct stacktraces)
+          # Sample allocations with 2 distinct stacktraces
+          # (the 2nd parameter is there only to avoid a rubocop issue with identical branches)
```
ivoanjo, line 362: Minor: It's actually possible to shut rubocop up with an inline comment, e.g.

```ruby
        def self.pkg_config_missing?(command: $PKGCONFIG) # rubocop:disable Style/GlobalVars
```

... I guess it would be a bit cleaner here to do that than passing dummy args in?

ivoanjo (approved):
Here sir, take my :+1: . I've left a few notes on how I don't quite think this is the way to go, but leave the decision in your hands :)

ext/ddtrace_profiling_native_extension/heap_recorder.c
AlexJF: Hmm yeah you may be onto something here. I had assumed the map copy would be slower but it seems to be quite fast indeed. I'll follow your suggestion and maybe use this as a benchmarking case to move our pie closer to the ground 😂

ext/ddtrace_profiling_native_extension/heap_recorder.c
AlexJF: Nice flag, fixed it upstream.

ext/ddtrace_profiling_native_extension/heap_recorder.c
AlexJF: Implemented!

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -605,13 +617,15 @@ static void build_heap_profile_without_gvl(struct stack_recorder_state *state, d
...
-    // We wait until we're out of the iteration to grab the gvl and raise. This is important because during
-    // iteration we first grab the records_mutex and raising requires grabbing the GVL. When sampling, we are
-    // in the opposite situation: we have the GVL and may need to grab the records_mutex for mutation. This
-    // different ordering can lead to deadlocks. By delaying the raise here until after we no longer hold
-    // records_mutex, we prevent this different-lock-acquisition-order situation.
+  bool iterated = heap_recorder_for_each_live_object(state->heap_recorder, add_heap_sample_to_active_profile_without_gvl, (void*) &iteration_context);
+  // We wait until we're out of the iteration to grab the gvl and raise. This is important because during
+  // iteration we may potentially acquire locks in the heap recorder and we could reach a deadlock if the
+  // same locks are acquired by the heap recorder while holding the gvl (since we'd be operating on the
+  // same locks but acquiring them in different order).
```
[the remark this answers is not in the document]
AlexJF: It is but I still think it's healthy to keep this structure to try and reduce the dependency on internal details of heap recorder and thus futureproof this against  any potential changes on that side. 

I added the "potentially" keyword with that intention 😅

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -353,27 +353,47 @@ def sample_types_from(decoded_profile)
...
+      end
+
+      after do |example|
+        # This is here to facilitate troubleshooting when this test fails. Otherwise
+        # it's very hard to understand what may be happening.
+        if example.exception
+          puts('Heap recorder debugging info:')
+          puts(described_class::Testing._native_debug_heap_recorder(stack_recorder))
+        end
       end
```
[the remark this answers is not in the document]
AlexJF: This troubleshooting would be helpful for the great majority of the tests in this context which is why I added it via an after. Strong feelings in keeping it this way vs adding extra parameters to different expectations in individual tests?

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -187,21 +197,14 @@ void heap_recorder_after_fork(heap_recorder *heap_recorder) {
...
-  // as unlocking mutexes a thread doesn't own is undefined behaviour. What we can do is
-  // create a new lock and start using it from now on-forward. This is fine because at this
-  // point in the fork-handling logic, all tracepoints are disabled and no-one should be
-  // iterating on the recorder state so there are no writers/readers that may race with
-  // this reinitialization.
-  heap_recorder->records_mutex = (pthread_mutex_t) PTHREAD_MUTEX_INITIALIZER;
+  // outside of the GVL, any state specific to that interaction may be incosistent after fork
+  // (e.g. an acquired lock for thread safety). Iteration operates on object_records_snapshot
+  // though and that one will be updated on next heap_recorder_prepare_iteration so there's
+  // nothing for us to do here.
```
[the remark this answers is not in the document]
AlexJF: Yes good catch!

spec/datadog/profiling/stack_recorder_spec.rb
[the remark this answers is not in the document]
ivoanjo: No strong feelings at all in leaving as-is! Just sharing that tidbit, as it may come in handy :)

ext/ddtrace_profiling_native_extension/heap_recorder.c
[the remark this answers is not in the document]
AlexJF: Done in [`4b3d636` (#3329)](https://github.com/DataDog/dd-trace-rb/pull/3329/commits/4b3d636fb5eeb4b3e63a5937a2d5c01d956311df)

PR #3329, opened 2023-12-13
[PROF-8667] Heap Profiling - Part 4 - Labels

AlexJF:
**What does this PR do?**

This PR follows https://github.com/DataDog/dd-trace-rb/pull/3328 by adding `allocation class` and `gc gen age` labels to the heap samples included in the Ruby profiles.

**How does it work?**

For `allocation class`, we simply hold on to the resolved class that was already being computed during the normal allocation sampling path.

For `gc gen age`, we record the GC Generation/Epoch (i.e. `GC.count`) when we saw the allocation happening. Then, when writing the heap samples in a new profile, we calculate the difference between the allocation generation and the current generation and that gives us the age of the recorded object in terms of "how many GCs have run since we saw this object being allocated".

**Motivation:**
* Understand the type of objects that are being kept alive in our heap.
* Understand the age dynamics of the objects being kept alive in heap (are they all "new" and likely

**Additional Notes:**

**How to test the change?**
Heap samples included in profiles when an app is executed with:

```
```

should now include the above labels. This can be trivially tested by downloading these profiles and looking at them with some pprof parsing tool like https://github.com/felixge/pprofutils

**For Datadog employees:**

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_thread_context.c
@@ -1244,7 +1244,7 @@ void thread_context_collector_sample_allocation(VALUE self_instance, unsigned in
     }
   }
 
-  track_object(state->recorder_instance, new_object, sample_weight);
+  track_object(state->recorder_instance, new_object, sample_weight, optional_class_name);
 
   trigger_sample_for_thread(
     state,
--- a/ext/ddtrace_profiling_native_extension/heap_recorder.c
+++ b/ext/ddtrace_profiling_native_extension/heap_recorder.c
@@ -200,8 +200,11 @@ void heap_recorder_after_fork(heap_recorder *heap_recorder) {
   // will be the thread holding on to the GVL. Since we support iteration on the heap recorder
   // outside of the GVL, any state specific to that interaction may be incosistent after fork
   // (e.g. an acquired lock for thread safety). Iteration operates on object_records_snapshot
-  // though and that one will be updated on next heap_recorder_prepare_iteration so there's
-  // nothing for us to do here.
+  // though and that one will be updated on next heap_recorder_prepare_iteration so we really
+  // only need to finish any iteration that might have been left unfinished.
+  if (heap_recorder->object_records_snapshot != NULL) {
+    heap_recorder_finish_iteration(heap_recorder);
+  }
 }
 
 void start_heap_allocation_recording(heap_recorder *heap_recorder, VALUE new_obj, unsigned int weight, ddog_CharSlice *alloc_class) {
--- a/ext/ddtrace_profiling_native_extension/heap_recorder.c
+++ b/ext/ddtrace_profiling_native_extension/heap_recorder.c
@@ -4,6 +4,7 @@
 #include "ruby_helpers.h"
 #include <errno.h>
 #include "collectors_stack.h"
+#include "libdatadog_helpers.h"
 
 // A compact representation of a stacktrace frame for a heap allocation.
 typedef struct {
@@ -92,15 +93,6 @@ typedef struct {
 static object_record* object_record_new(long, heap_record*, live_object_data);
 static void object_record_free(object_record*);
 
-// Allows storing data passed to ::start_heap_allocation_recording to make it accessible to
-// ::end_heap_allocation_recording.
-//
-// obj_id != 0 flags this struct as holding a valid partial heap recording.
-typedef struct {
-  long obj_id;
-  live_object_data object_data;
-} partial_heap_recording;
-
 struct heap_recorder {
   // Map[key: heap_record_key*, record: heap_record*]
   // NOTE: We always use heap_record_key.type == HEAP_STACK for storage but support lookups
@@ -125,7 +117,7 @@ struct heap_recorder {
   st_table *object_records_snapshot;
 
   // Data for a heap recording that was started but not yet ended
-  partial_heap_recording active_recording;
+  object_record *partial_object_record;
 
   // Reusable location array, implementing a flyweight pattern for things like iteration.
   ddog_prof_Location *reusable_locations;
@@ -138,7 +130,7 @@ static int st_object_record_entry_free_if_invalid(st_data_t, st_data_t, st_data_
 static int st_object_records_iterate(st_data_t, st_data_t, st_data_t);
 static int st_object_records_debug(st_data_t key, st_data_t value, st_data_t extra);
 static int update_object_record_entry(st_data_t*, st_data_t*, st_data_t, int);
-static void commit_allocation(heap_recorder*, heap_record*, long, live_object_data);
+static void commit_allocation(heap_recorder*, heap_record*, object_record*);
 
 // ==========================
 // Heap Recorder External API
@@ -156,11 +148,7 @@ heap_recorder* heap_recorder_new(void) {
   recorder->object_records = st_init_numtable();
   recorder->object_records_snapshot = NULL;
   recorder->reusable_locations = ruby_xcalloc(MAX_FRAMES_LIMIT, sizeof(ddog_prof_Location));
-  recorder->active_recording = (partial_heap_recording) {
-    .obj_id = 0, // 0 is actually the obj_id of false, but we'll never track that one in heap so we use
-                 // it as invalid/unset value.
-    .object_data = {0},
-  };
+  recorder->partial_object_record = NULL;
 
   return recorder;
 }
@@ -176,11 +164,20 @@ void heap_recorder_free(heap_recorder *heap_recorder) {
     heap_recorder_finish_iteration(heap_recorder);
   }
 
+  // Clean-up all object records
   st_foreach(heap_recorder->object_records, st_object_record_entry_free, 0);
   st_free_table(heap_recorder->object_records);
 
+  // Clean-up all heap records (this includes those only referred to by queued_samples)
   st_foreach(heap_recorder->heap_records, st_heap_record_entry_free, 0);
   st_free_table(heap_recorder->heap_records);
+
+  if (heap_recorder->partial_object_record != NULL) {
+    // If there's a partial object record, clean it up as well
+    object_record_free(heap_recorder->partial_object_record);
+  }
+
+  ruby_xfree(heap_recorder->reusable_locations);
 
   ruby_xfree(heap_recorder);
 }
@@ -207,7 +204,7 @@ void heap_recorder_after_fork(heap_recorder *heap_recorder) {
   // nothing for us to do here.
 }
 
-void start_heap_allocation_recording(heap_recorder *heap_recorder, VALUE new_obj, unsigned int weight) {
+void start_heap_allocation_recording(heap_recorder *heap_recorder, VALUE new_obj, unsigned int weight, ddog_CharSlice *alloc_class) {
   if (heap_recorder == NULL) {
     return;
   }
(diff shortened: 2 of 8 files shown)


ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -141,9 +145,11 @@ static int st_object_record_entry_free(st_data_t, st_data_t, st_data_t);
 static int st_object_record_entry_free_if_invalid(st_data_t, st_data_t, st_data_t);
 static int st_object_records_iterate(st_data_t, st_data_t, st_data_t);
 static int update_object_record_entry(st_data_t*, st_data_t*, st_data_t, int);
-static void commit_allocation(heap_recorder*, heap_record*, long, live_object_data);
+static void commit_allocation(heap_recorder*, heap_record*, long, live_object_data*);
 static void flush_queue(heap_recorder*);
-static bool enqueue_allocation(heap_recorder*, heap_record*, long, live_object_data);
+static bool enqueue_allocation(heap_recorder*, heap_record*, long, live_object_data*);
```
ivoanjo, line 150: Minor: I wonder if `commit_allocation` and similar methods should receive an `object_record*` directly?

That way we would only need to pass 1 argument around, instead of having to keep "slicing" the object record's fields to pass them around until we actually create the object record :)

(I'm reminded of the horse scene from "the cell" which always made me :cold_sweat: )

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -89,9 +90,10 @@ static void heap_record_free(heap_record*);
 typedef struct {
   long obj_id;
   heap_record *heap_record;
-  live_object_data object_data;
+  // Has ownership of this, needs to clean-it-up if not transferred.
+  live_object_data *object_data;
 } object_record;
```
ivoanjo, line 95: Hmm, I wonder if it's time to revisit the need of `live_object_data` to exist stand-alone.

Prior to this change, it was just something we could pass a reference to, so you'd get only that part, and we didn't need to expose the whole `object_record`.

Which... is fine. But now we're doing two allocations and have to track the lifetime of two things -- is it perhaps time to inline the fields from `live_object_data` into `object_record` directly?

(And yeah, we also reference `live_object_data` from a few other structures, which makes me think we could either inline the fields or replace those structures with `object_record` as well)

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -177,12 +183,22 @@ void heap_recorder_free(heap_recorder *heap_recorder) {
...
   st_foreach(heap_recorder->heap_records, st_heap_record_entry_free, 0);
   st_free_table(heap_recorder->heap_records);
 
+  // Clean-up dangling object datas that are not part of object records yet
+  if (heap_recorder->active_recording.obj_id != 0) {
+    live_object_data_free(heap_recorder->active_recording.object_data);
+  }
+  for (size_t i = 0; i < heap_recorder->queued_samples_len; i++) {
+    live_object_data_free(heap_recorder->queued_samples[i].object_data);
+  }
```
ivoanjo, line 200: If we do keep the queue (as per discussion in earlier PRs), I wonder if we should consider having the `queued_samples` not be an array of structures, but instead an array of `object_record *`. We're going to do those allocations anyway, so we may as well do them as part of enqueuing, rather than waiting until later maybe?

ext/ddtrace_profiling_native_extension/libdatadog_helpers.h
```diff
@@ -34,3 +34,7 @@ size_t read_ddogerr_string_and_drop(ddog_Error *error, char *string, size_t capa
 // ruby_value_type that Ruby uses so that we can also use this for debugging.
 const char *ruby_value_type_to_string(enum ruby_value_type type);
 ddog_CharSlice ruby_value_type_to_char_slice(enum ruby_value_type type);
+
+// Returns a dynamically allocated string from the provided char slice.
+// WARN: The returned string must be explicitly freed with ruby_xfree.
+char* string_from_char_slice(ddog_CharSlice slice);
```
ivoanjo, line 40: Minor: Since the implementation is so trivial, should we just move it here and make it `inline static`?

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -551,14 +554,34 @@ static bool add_heap_sample_to_active_profile_without_gvl(heap_recorder_iteratio
   int64_t metric_values[ALL_VALUE_TYPES_COUNT] = {0};
   uint8_t *position_for = context->state->position_for;
 
-  metric_values[position_for[HEAP_SAMPLES_VALUE_ID]] = iteration_data.object_data.weight;
+  metric_values[position_for[HEAP_SAMPLES_VALUE_ID]] = iteration_data.object_data->weight;
+
+  const live_object_data *object_data = iteration_data.object_data;
+  ddog_prof_Label labels[2];
+  size_t num_labels = 2;
```
ivoanjo, line 561: Minor:

```suggestion
  size_t num_labels = 2;
  ddog_prof_Label labels[num_labels];
```

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -393,13 +397,67 @@ def sample_allocation(obj, _unused = 0)
...
+          expect(array_sample.labels).to match(
+            {
+              :'allocation class' => 'Array',
+              :'gc gen age' => approximate_age_based_on_allocation_order(1)
+            }
+          )
+
+          hash_sample = heap_samples.find { |s| s.labels[:'allocation class'] == 'Hash' }
+          expect(hash_sample).not_to be nil
+          expect(array_sample.values).to match(
```
ivoanjo, line 450: Copy-pasta ;)

```suggestion
          expect(hash_sample.values).to match(
```

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -393,13 +397,67 @@ def sample_allocation(obj, _unused = 0)
...
+            "be between #{expected_age} and #{expected_age + flaky_margin}"
+          end
+        end
+
+        it 'include the stack, sample counts and labels for the objects still left alive' do
+          # There should be 3 different allocation class labels so we expect 3 different heap samples
+          expect(heap_samples.size).to eq(3)
+
+          string_sample = heap_samples.find { |s| s.labels[:'allocation class'] == 'String' }
+          expect(string_sample).not_to be nil
```
ivoanjo, line 421: Minor: Since this testcase is already quite complex, I'd suggest removing this + the other matching assertions.

Anyway, if this happens there's other things that would fail, so at least we reduce a bit the complexity :)

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -393,13 +397,67 @@ def sample_allocation(obj, _unused = 0)
...
+
+          match do |gc_gen_age|
+            expected_age = num_allocations - allocation_order
+            gc_gen_age.between?(expected_age, expected_age + flaky_margin)
+          end
+          description do
+            expected_age = num_allocations - allocation_order
+            "be between #{expected_age} and #{expected_age + flaky_margin}"
+          end
+        end
```
ivoanjo, line 414: So... I'm a bit divided about this. It does work, and it's quite fancy, but I wonder if we could test the same property without needing this kind of machinery.

For instance, if we're testing the age of objects gets calculated correctly, perhaps we could assert in a separate testcase that the ages of the 3 relevant samples are in order of age (e.g. `ages.size == ages.size.uniq && ages.sort == ages`) and that the last age is within the expected value -- something like that?

ivoanjo (approved):
Left some notes, but this looks sane! :+1:

ext/ddtrace_profiling_native_extension/heap_recorder.c
AlexJF: Yeah this makes sense and allows me to remove yet another struct!

ext/ddtrace_profiling_native_extension/heap_recorder.c
AlexJF: Right. I went with an intermediate approach where live_object_data still exists but as an inner data struct of `object_record`  and without a separate lifecycle. Thus I can also re-use it as part of the public API around iteration (where users should not have access to the underlying heap_record and such).

I think this is cleaner as it allows us to keep documentation centralized and easily extend it with more data later without needing to inflate arguments of `object_record_new`.

ext/ddtrace_profiling_native_extension/heap_recorder.c
AlexJF: Killed the queue in part 3!

ext/ddtrace_profiling_native_extension/collectors_thread_context.c
```diff
@@ -1210,7 +1210,7 @@ void thread_context_collector_sample_allocation(VALUE self_instance, unsigned in
     }
   }
 
-  track_object(state->recorder_instance, new_object, sample_weight);
+  track_object(state->recorder_instance, new_object, sample_weight, class_name);
```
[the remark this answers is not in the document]
AlexJF: Went with optional class names for now.

ext/ddtrace_profiling_native_extension/stack_recorder.c
AlexJF: Needed a slight re-work to deal with optional class names.

spec/datadog/profiling/stack_recorder_spec.rb
AlexJF: Yeah this is a good suggestion 👍 Done and moved the age accuracy checks into their own example as I felt the existing one was becoming quite bloated.

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -418,15 +426,72 @@ def sample_allocation(obj)
...
+              :"heap-live-samples" => sample_rate,
+            )
+          )
+          expect(hash_sample.labels).to match(
+            {
+              :'allocation class' => 'Hash',
+              :'gc gen age' => be_a(Integer).and(be >= 0),
+            }
+          )
+        end
```
ivoanjo, line 470: Minor: Since the checks are pretty much the same, they can be simplified a bit:

```ruby
          expect(heap_samples.map { |it| it.labels[:'allocation class'] }).to include('String', 'Array', 'Hash')
          expect(heap_samples.map(&:labels)).to all(match(:'gc gen age' => be_a(Integer).and(be >= 0)))
```

ivoanjo (approved):
LGTM :+1: Left a few final comments/suggestions

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -372,18 +372,26 @@ def sample_allocation(obj)
...
         # reason but it might be that some machine context/register ends up still pointing to
         # that last entry and thus manages to get it marked in the first GC.
+        # To reduce the likelihood of this happening we'll:
+        # * Allocate some more stuff and clear again
+        # * Sleep a bit
+        # * Do another GC
+        allocations = ["another fearsome interpolated string: #{sample_rate}", (-20..-10).to_a,
+                       { 'a' => 1, 'b' => '2', 'c' => true }, Object.new]
+        allocations.clear
+        sleep(0.5)
```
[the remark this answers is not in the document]
AlexJF: Yeah that is fair!

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -616,6 +642,7 @@ static void build_heap_profile_without_gvl(struct stack_recorder_state *state, d
     .profile = profile,
     .error = false,
     .error_msg = {0},
+    .profile_gen = rb_gc_count(),
```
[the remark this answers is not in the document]
AlexJF: Yep makes a lot of sense!

spec/datadog/profiling/stack_recorder_spec.rb
AlexJF: Ah yes failed to spot this after splitting the tests 😅

PR #3332, opened 2023-12-14
Stricter validation for integer and float configuration

marcotc:
**2.0 Upgrade Guide notes**
Configuration options of types ':int' and ':float' now have stricter type validation:
* Fractional numbers are no longer considered valid for ':int' options.
* Empty values (nil or empty string) are no longer considered valid for ':int' or ':float' options.

This applies to both environment variable and programmatic configuration.
Here's the list of all affected options:
[Insert all :int and :float options at release time!!!]

The environment variable `DD_EXPERIMENTAL_SKIP_CONFIGURATION_VALIDATION` has been removed.

**Motivation:**
It was possible to configure options with empty values (e.g. `DD_TRACE_SAMPLE_RATE=`) and end up with a sampling rate of `0.0`, because of the logic `value.to_f`. The same applies to integer options (`value.to_i`).
Non-numeric values would also be converted to zero.
This means users do not know what their effective configuration will be, and are not given feedback on how to address configuration issues.

**How to test the change?**
There are unit tests for all the changes.

**For Datadog employees:**

### DIFF
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -196,7 +196,7 @@ module Datadog
 
         def coerce_env_var_int(value)
           # Strict integer parsing
-          Integer(value)
+          Integer(value, 10)
         rescue ArgumentError => e
           # Try also parsing as a whole floating point numbers
           begin
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -168,11 +168,9 @@ module Datadog
               hash[pair[0]] = pair[1]
             end
           when :int
-            # DEV-2.0: Change to a more strict coercion method. Integer(value).
-            value.to_i
+            coerce_env_var_int(value)
           when :float
-            # DEV-2.0: Change to a more strict coercion method. Float(value).
-            value.to_f
+            Float(value)
           when :array
             values = value.split(',')
 
@@ -191,9 +189,31 @@ module Datadog
           when :string, NilClass
             value
           else
-            raise ArgumentError,
+            raise InvalidDefinitionError,
               "The option #{@definition.name} is using an unsupported type option for env coercion `#{@definition.type}`"
           end
+        end
+
+        def coerce_env_var_int(value)
+          # Strict integer parsing
+          Integer(value)
+        rescue ArgumentError => e
+          # Try also parsing as a whole floating point numbers
+          begin
+            f = Float(value)
+
+            # Check for whole numbers
+            raise ArgumentError, "#{value} is not a whole number" unless f.truncate == f
+
+            f
+          rescue ArgumentError
+            # It's neither an integer nor a whole float
+            raise e
+          end
+        end
+
+        # For errors caused by illegal option declarations
+        class InvalidDefinitionError < StandardError
         end
 
         def validate_type(value)
@@ -235,8 +255,10 @@ module Datadog
           case type
           when :string
             value.is_a?(String)
-          when :int, :float
-            value.is_a?(Numeric)
+          when :int
+            value.is_a?(Integer) || (value.is_a?(Float) && (value.truncate == value))
+          when :float
+            value.is_a?(Float) || value.is_a?(Integer) || value.is_a?(Rational)
           when :array
             value.is_a?(Array)
           when :hash
@@ -279,13 +301,16 @@ module Datadog
         def set_value_from_env_or_default
           value = nil
           precedence = nil
+          effective_env = nil
 
           if definition.env && ENV[definition.env]
+            effective_env = definition.env
             value = coerce_env_variable(ENV[definition.env])
             precedence = Precedence::PROGRAMMATIC
           end
 
           if value.nil? && definition.deprecated_env && ENV[definition.deprecated_env]
+            effective_env = definition.deprecated_env
             value = coerce_env_variable(ENV[definition.deprecated_env])
             precedence = Precedence::PROGRAMMATIC
 
@@ -297,6 +322,10 @@ module Datadog
           option_value = value.nil? ? default_value : value
 
           set(option_value, precedence: precedence || Precedence::DEFAULT)
+        rescue ArgumentError
+          raise ArgumentError,
+            "Expected environment variable #{effective_env} to be a #{@definition.type}, " \
+                              "but '#{ENV[effective_env]}' was provided"
         end
 
         def skip_validation?
--- a/lib/datadog/core/configuration/option.rb
+++ b/lib/datadog/core/configuration/option.rb
@@ -168,7 +168,7 @@ module Datadog
               hash[pair[0]] = pair[1]
             end
           when :int
-            coerce_env_var_int(value)
+            Integer(value, 10)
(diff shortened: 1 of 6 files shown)


lib/datadog/core/configuration/option.rb
```diff
@@ -191,11 +189,33 @@ def coerce_env_variable(value)
...
+        def coerce_env_var_int(value)
+          # Strict integer parsing
+          Integer(value)
+        rescue ArgumentError => e
+          # Try also parsing as a whole floating point numbers
+          begin
+            f = Float(value)
+
+            # Check for whole numbers
+            raise ArgumentError, "#{value} is not a whole number" unless f.truncate == f
```
ivoanjo, line 206: To be honest... This seems like a bit of a surprising semantics. If I'm in a rush and see something on my config that I want to update:

```
DD_PROFILING_HOW_MUCH_IT_ROCKS=100.0
```

and then I go, oh, I want to make it rock a bit less

```
DD_PROFILING_HOW_MUCH_IT_ROCKS=95.5
```

and it blows up -- this seems weird. That .0 that someone on my team left misled me into thinking this value was supposed to be a float, and then it turns out it isn't and it doesn't work.

lib/datadog/core/configuration/option.rb
```diff
@@ -191,11 +189,33 @@ def coerce_env_variable(value)
...
           else
-            raise ArgumentError,
+            raise InvalidDefinitionError,
               "The option #{@definition.name} is using an unsupported type option for env coercion `#{@definition.type}`"
           end
         end
 
+        def coerce_env_var_int(value)
+          # Strict integer parsing
+          Integer(value)
```
ivoanjo, line 199: I learned (depressingly) the other day that `Integer(value)` actually supports hex and octal literals => e.g. `Integer("010") => 8` and `Integer("0x10") => 16`.

TL;DR the correct form is `Integer(value, 10)` to specifically ask it to parse only as base 10. Now I'll go back to wondering why this isn't the default >_>

ivoanjo (commented):
I'm a bit short on time today so I didn't look at this very thoroughly but here's a quick pass :)

lib/datadog/core/configuration/option.rb
marcotc: What do you think about allowing whole floats as integer @TonyCTHsu @ekump? (e.g. `100.0` is allowed for `:int` configuration options). Or should be strict and say that this is not allowed (e.g. `100.0` would be rejected, and the user should replace it with an explicit `100`)

lib/datadog/core/configuration/option.rb
marcotc: I think if only `0x10` (and base 10) were allowed, we wouldn't bat an eye, but octals ruin it for everyone.

marcotc: Fixed it!

lib/datadog/core/configuration/option.rb
marcotc: @ivoanjo, how about a warning on whole floats? I think that might be a good compromise.

ivoanjo: > how about a warning on whole floats?

I would still kinda prefer not to accept 'em, but I can live with that as a compromise ;)

lib/datadog/core/configuration/option.rb
ivoanjo: Yeah, it's hard to accidentally have hex, but so easy to accidentally have octal >_>

lib/datadog/core/configuration/option.rb
ivoanjo: We had discussed moving this to a warning, which I think is better than the current raise. If we do keep the raise, I would suggest making it a bit clear what needs to be fixed in the current error -- e.g. "Setting expects only integer settings, you've provided #{value}"?

lib/datadog/core/configuration/option.rb
```diff
@@ -191,11 +189,33 @@ def coerce_env_variable(value)
...
+            f = Float(value)
+
+            # Check for whole numbers
+            raise ArgumentError, "#{value} is not a whole number" unless f.truncate == f
+
+            f
+          rescue ArgumentError
+            # It's neither an integer nor a whole float
+            raise e
+          end
```
ivoanjo, line 212: Is it just me, or would this `rescue` swallow our `raise ArgumentError, "#{value} is not a whole number" unless f.truncate == f` above and replace it with the original `Integer` parsing error?

(I guess it may be worth adding a spec to validate that the right error/message gets presented?)

lib/datadog/core/configuration/option.rb
```diff
@@ -191,11 +189,33 @@ def coerce_env_variable(value)
...
+          Integer(value, 10)
+        rescue ArgumentError => e
+          # Try also parsing as a whole floating point numbers
+          begin
+            f = Float(value)
+
+            # Check for whole numbers
+            raise ArgumentError, "#{value} is not a whole number" unless f.truncate == f
+
+            f
```
ivoanjo, line 208: Minor: Output an `Integer` in the end?

```suggestion
            f.to_i
```

lib/datadog/core/configuration/option.rb
```diff
@@ -235,8 +255,10 @@ def validate(type, value)
           case type
           when :string
             value.is_a?(String)
-          when :int, :float
-            value.is_a?(Numeric)
+          when :int
+            value.is_a?(Integer) || (value.is_a?(Float) && (value.truncate == value))
+          when :float
+            value.is_a?(Float) || value.is_a?(Integer) || value.is_a?(Rational)
```
ivoanjo, line 261: This makes me think we should change `validate_type` above to coerce int and float to the correct types.

It seems like a bit of a potential sharp edge that someone can set a `Rational` on a `:float` setting, or a `Float` on `:int` but our code may not deal very well with it.

On the other hand, it's nice to allow these; hence why I think here would be a nice place to coerce after validation?

ivoanjo (commented):
Left a few final notes :)

marcotc:
@ivoanjo I talked to @TonyCTHsu and I was convinced to have stricter Integer validation, which aligns with your original suggestion.

lib/datadog/core/configuration/option.rb
```diff
@@ -235,8 +232,10 @@ def validate(type, value)
           case type
           when :string
             value.is_a?(String)
-          when :int, :float
-            value.is_a?(Numeric)
+          when :int
+            value.is_a?(Integer)
+          when :float
+            value.is_a?(Float) || value.is_a?(Rational)
```
ivoanjo, line 238: I think allowing users to provide integer values for floats is a nice UX touch -- e.g. if you're providing a percentage, often you may want to say 50 and not 50.0 or something like that.

So I don't think we should lose that.

ivoanjo (approved):
I think we should still allow users to provide `Integer` to float settings when doing it via the configure block, but otherwise this looks good to me :+1:

ivoanjo (approved):
:+1: LGTM

PR #3333, opened 2023-12-15
[PROF-8667] Heap Profiling - Part 5 - Size

AlexJF:
**What does this PR do?**

This PR follows https://github.com/DataDog/dd-trace-rb/pull/3329 by adding `live-heap-size` values to heap samples. These values give an idea of how much estimated heap memory the weighted samples tracked by `live-heap-samples` are using.

**How does it work?**

On each `heap_recorder_flush`, for all object records associated with objects that were deemed to be still alive, we call `rb_obj_memsize_of` to get an updated value for that object's estimated size in memory.

When writing a heap sample to a profile on serialization, we use the weighted object size (i.e. multiplied by the sampling weight of that object) as the value for the `heap-live-size` profile type for that sample.

**Motivation:**
Go one step further from understanding the approximate number of objects alive in the heap, to understanding the approximate space those objects are using in heap.

**Additional Notes:**

**How to test the change?**
Heap samples included in profiles when an app is executed with:

```
```

should now include values for `heap-live-size` profile type. This can be trivially tested by downloading these profiles and looking at them with some pprof parsing tool like https://github.com/felixge/pprofutils

**For Datadog employees:**

### DIFF
--- a/ext/ddtrace_profiling_native_extension/heap_recorder.c
+++ b/ext/ddtrace_profiling_native_extension/heap_recorder.c
@@ -94,6 +94,10 @@ static object_record* object_record_new(long, heap_record*, live_object_data);
 static void object_record_free(object_record*);
 
 struct heap_recorder {
+  // Config
+  // Whether the recorder should try to determine approximate sizes for tracked objects.
+  bool size_enabled;
+
   // Map[key: heap_record_key*, record: heap_record*]
   // NOTE: We always use heap_record_key.type == HEAP_STACK for storage but support lookups
   // via heap_record_key.type == LOCATION_SLICE to allow for allocation-free fast-paths.
@@ -149,6 +153,7 @@ heap_recorder* heap_recorder_new(void) {
   recorder->object_records_snapshot = NULL;
   recorder->reusable_locations = ruby_xcalloc(MAX_FRAMES_LIMIT, sizeof(ddog_prof_Location));
   recorder->partial_object_record = NULL;
+  recorder->size_enabled = true;
 
   return recorder;
 }
@@ -180,6 +185,10 @@ void heap_recorder_free(heap_recorder *heap_recorder) {
   ruby_xfree(heap_recorder->reusable_locations);
 
   ruby_xfree(heap_recorder);
+}
+
+void heap_recorder_set_size_enabled(heap_recorder *heap_recorder, bool size_enabled) {
+  heap_recorder->size_enabled = size_enabled;
 }
 
 // WARN: Assumes this gets called before profiler is reinitialized on the fork
@@ -250,15 +259,7 @@ void end_heap_allocation_recording(struct heap_recorder *heap_recorder, ddog_pro
   commit_allocation(heap_recorder, heap_record, partial_object_record);
 }
 
-typedef struct {
-  // A reference to the heap recorder so we can access extra stuff to cleanup unused records.
-  heap_recorder *heap_recorder;
-
-  // Whether we should update object sizes as part of the update iteration or not.
-  bool update_sizes;
-} prepare_iteration_context;
-
-void heap_recorder_prepare_iteration(heap_recorder *heap_recorder, bool update_sizes) {
+void heap_recorder_prepare_iteration(heap_recorder *heap_recorder) {
   if (heap_recorder == NULL) {
     return;
   }
@@ -268,11 +269,7 @@ void heap_recorder_prepare_iteration(heap_recorder *heap_recorder, bool update_s
     rb_raise(rb_eRuntimeError, "New heap recorder iteration prepared without the previous one having been finished.");
   }
 
-  prepare_iteration_context context = (prepare_iteration_context) {
-    .heap_recorder = heap_recorder,
-    .update_sizes = update_sizes,
-  };
-  st_foreach(heap_recorder->object_records, st_object_record_update, (st_data_t) &context);
+  st_foreach(heap_recorder->object_records, st_object_record_update, (st_data_t) heap_recorder);
 
   heap_recorder->object_records_snapshot = st_copy(heap_recorder->object_records);
   if (heap_recorder->object_records_snapshot == NULL) {
@@ -376,7 +373,7 @@ static int st_object_record_entry_free(DDTRACE_UNUSED st_data_t key, st_data_t v
 static int st_object_record_update(st_data_t key, st_data_t value, st_data_t extra_arg) {
   long obj_id = (long) key;
   object_record *record = (object_record*) value;
-  prepare_iteration_context *context = (prepare_iteration_context*) extra_arg;
+  heap_recorder *recorder = (heap_recorder*) extra_arg;
 
   VALUE ref;
 
@@ -388,7 +385,7 @@ static int st_object_record_update(st_data_t key, st_data_t value, st_data_t ext
     heap_record->num_tracked_objects--;
 
     // One less object using this heap record, it may have become unused...
-    cleanup_heap_record_if_unused(context->heap_recorder, heap_record);
+    cleanup_heap_record_if_unused(recorder, heap_record);
 
     object_record_free(record);
     return ST_DELETE;
@@ -396,9 +393,12 @@ static int st_object_record_update(st_data_t key, st_data_t value, st_data_t ext
 
   // If we got this far, entry is still valid
 
-  if (context->update_sizes) {
-    // if we were asked to update sizes, do so...
+  if (recorder->size_enabled && !record->object_data.is_frozen) {
+    // if we were asked to update sizes and this object was not already seen as being frozen,
+    // update size again.
     record->object_data.size = ruby_obj_memsize_of(ref);
+    // Check if it's now frozen so we skip a size update next time
+    record->object_data.is_frozen = RB_OBJ_FROZEN(ref);
   }
 
   return ST_CONTINUE;
--- a/ext/ddtrace_profiling_native_extension/heap_recorder.c
+++ b/ext/ddtrace_profiling_native_extension/heap_recorder.c
@@ -126,7 +126,7 @@ static heap_record* get_or_create_heap_record(heap_recorder*, ddog_prof_Slice_Lo
 static void cleanup_heap_record_if_unused(heap_recorder*, heap_record*);
 static int st_heap_record_entry_free(st_data_t, st_data_t, st_data_t);
 static int st_object_record_entry_free(st_data_t, st_data_t, st_data_t);
-static int st_object_record_entry_free_if_invalid(st_data_t, st_data_t, st_data_t);
+static int st_object_record_update(st_data_t, st_data_t, st_data_t);
 static int st_object_records_iterate(st_data_t, st_data_t, st_data_t);
 static int st_object_records_debug(st_data_t key, st_data_t value, st_data_t extra);
 static int update_object_record_entry(st_data_t*, st_data_t*, st_data_t, int);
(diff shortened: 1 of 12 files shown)


ext/ddtrace_profiling_native_extension/ruby_helpers.c
```diff
@@ -167,3 +167,39 @@ bool ruby_ref_from_id(VALUE obj_id, VALUE *value) {
...
+  switch (BUILTIN_TYPE(obj)) {
+    case T_OBJECT:
+    case T_MODULE:
+    case T_CLASS:
+    case T_ICLASS:
+    case T_STRING:
+    case T_ARRAY:
+    case T_HASH:
+    case T_REGEXP:
+    case T_DATA:
```
ivoanjo, line 185: I soft-wonder if we'll run into trouble with `T_DATA` (or data/typeddata) objects, since  the size for these is up to native extension authors. Since these methods don't get used very often, they may be buggy or inefficient. I guess... let's maybe keep an eye out on these ones?

ext/ddtrace_profiling_native_extension/ruby_helpers.c
```diff
@@ -167,3 +167,39 @@ bool ruby_ref_from_id(VALUE obj_id, VALUE *value) {
 
   return true;
 }
+
+// Not part of public headers but is externed from Ruby
+size_t rb_obj_memsize_of(VALUE obj);
+
+// Wrapper around rb_obj_memsize_of to avoid hitting crashing paths.
+size_t ruby_obj_memsize_of(VALUE obj) {
```
ivoanjo, line 175: To be fair, the crashing path is only when an object is invalid -- all of the Ruby types are accounted for in `rb_obj_memsize_of(...)` so I'm not sure it's worth having this wrapper around it -- `rb_bug(...)` is a pretty big thing that the VM only uses when it truly believes there's something wrong at the VM/native level.

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -153,13 +153,15 @@ static VALUE error_symbol = Qnil; // :error in Ruby
...
+#define HEAP_SIZE_VALUE_ID 5
 
-static const ddog_prof_ValueType all_value_types[] = {CPU_TIME_VALUE, CPU_SAMPLES_VALUE, WALL_TIME_VALUE, ALLOC_SAMPLES_VALUE, HEAP_SAMPLES_VALUE};
+static const ddog_prof_ValueType all_value_types[] = {CPU_TIME_VALUE, CPU_SAMPLES_VALUE, WALL_TIME_VALUE, ALLOC_SAMPLES_VALUE, HEAP_SAMPLES_VALUE, HEAP_SIZE_VALUE};
 
 // This array MUST be kept in sync with all_value_types above and is intended to act as a "hashmap" between VALUE_ID and the position it
 // occupies on the all_value_types array.
 // E.g. all_value_types_positions[CPU_TIME_VALUE_ID] => 0, means that CPU_TIME_VALUE was declared at position 0 of all_value_types.
-static const uint8_t all_value_types_positions[] = {CPU_TIME_VALUE_ID, CPU_SAMPLES_VALUE_ID, WALL_TIME_VALUE_ID, ALLOC_SAMPLES_VALUE_ID, HEAP_SAMPLES_VALUE_ID};
+static const uint8_t all_value_types_positions[] = {CPU_TIME_VALUE_ID, CPU_SAMPLES_VALUE_ID, WALL_TIME_VALUE_ID, ALLOC_SAMPLES_VALUE_ID, HEAP_SAMPLES_VALUE_ID, HEAP_SIZE_VALUE_ID};
```
ivoanjo, line 164: Heads-up -- this will conflict with my GC timeline branch, consider merging master into this PR / rebasing on top of master ;)

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -344,8 +347,8 @@ def sample_types_from(decoded_profile)
       let(:labels) { { 'label_a' => 'value_a', 'label_b' => 'value_b', 'state' => 'unknown' }.to_a }
 
       let(:a_string) { 'a beautiful string' }
-      let(:an_array) { [1..10] }
-      let(:a_hash) { { 'a' => 1, 'b' => '2', 'c' => true } }
+      let(:an_array) { (1..100).to_a }
```
ivoanjo, line 350: Minor: I believe doing `(1..100).to_a.compact` will guarantee that the underlying array has a capacity of exactly 100.

ivoanjo (approved):
:+1: LGTM.

I'm curious about the performance impact of this one.

ext/ddtrace_profiling_native_extension/ruby_helpers.c
AlexJF: There's an extra "hidden" `rb_bug` call for the `T_NODE` case which was the one that worried me a bit more. The fact that they use `rb_bug` in that place may be a hint that it should be impossible for us to accidentally track one. But I'm also not 100% sure if that's the case or if they just assumed whoever called `obj_memsize_of` would be doing that check first.

ivoanjo: You're right -- I had missed that one. I don't know enough about RNode; intuitively if they're calling `rb_bug` it would be weird if we ever got handed one of those objects but I definitely see your concern.

I guess it may worth leaving a bit more context as a comment on why we're doing this, but I'm convinced :)

ext/ddtrace_profiling_native_extension/ruby_helpers.c
```diff
@@ -166,3 +166,39 @@ bool ruby_ref_from_id(VALUE obj_id, VALUE *value) {
...
+    case T_DATA:
+    case T_MATCH:
+    case T_FILE:
+    case T_RATIONAL:
+    case T_COMPLEX:
+    case T_IMEMO:
+    case T_FLOAT:
+    case T_SYMBOL:
+    case T_BIGNUM:
+    // case T_NODE: -> Throws exception in rb_obj_memsize_of
```
ivoanjo, line 193: Minor: It's not really an exception, it actually fully crashes the VM ;)

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -214,6 +217,7 @@ static VALUE _native_initialize(
   VALUE cpu_time_enabled,
   VALUE alloc_samples_enabled,
   VALUE heap_samples_enabled,
+  VALUE heap_sizes_enabled,
```
ivoanjo, line 220: Minor: Just to keep it in-sync with what's below...

```suggestion
  VALUE heap_size_enabled,
```

...I really dislike that C forces us to do this (or otherwise order our code in really awkward ways :(((

lib/datadog/profiling/component.rb
```diff
@@ -85,6 +85,7 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
           cpu_time_enabled: RUBY_PLATFORM.include?('linux'), # Only supported on Linux currently
           alloc_samples_enabled: allocation_profiling_enabled,
           heap_samples_enabled: heap_profiling_enabled,
+          heap_size_enabled: heap_profiling_enabled,
```
ivoanjo, line 88: Note to our future selves, a separate config for this gets introduced in https://github.com/DataDog/dd-trace-rb/pull/3360 .

ext/ddtrace_profiling_native_extension/stack_recorder.c
```diff
@@ -422,7 +430,18 @@ static VALUE _native_initialize(
     state->position_for[HEAP_SAMPLES_VALUE_ID] = next_enabled_pos++;
   } else {
     state->position_for[HEAP_SAMPLES_VALUE_ID] = next_disabled_pos++;
+  }
+
+  if (heap_size_enabled == Qtrue) {
+    enabled_value_types[next_enabled_pos] = (ddog_prof_ValueType) HEAP_SIZE_VALUE;
+    state->position_for[HEAP_SIZE_VALUE_ID] = next_enabled_pos++;
+    state->heap_size_enabled = true;
```
ivoanjo, line 438: Really minor: I guess we could also store this in the heap recorder itself?

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -361,25 +373,34 @@ static int st_object_record_entry_free(DDTRACE_UNUSED st_data_t key, st_data_t v
...
     object_record_free(record);
     return ST_DELETE;
   }
 
+  // If we got this far, entry is still valid
+
+  if (context->update_sizes) {
+    // if we were asked to update sizes, do so...
+    record->object_data.size = ruby_obj_memsize_of(ref);
+  }
```
ivoanjo, line 402: Minor: One optimization that came to mind would be to not recompute the size of frozen objects (given that, by definition, they become immutable).

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -340,16 +353,16 @@ def sample_types_from(decoded_profile)
...
       let(:sample_rate) { 50 }
       let(:metric_values) do
         { 'cpu-time' => 101, 'cpu-samples' => 1, 'wall-time' => 789, 'alloc-samples' => sample_rate, 'timeline' => 42 }
       end
       let(:labels) { { 'label_a' => 'value_a', 'label_b' => 'value_b', 'state' => 'unknown' }.to_a }
 
       let(:a_string) { 'a beautiful string' }
-      let(:an_array) { (1..10).to_a }
-      let(:a_hash) { { 'a' => 1, 'b' => '2', 'c' => true } }
+      let(:an_array) { (1..100).to_a.compact }
```
ivoanjo, line 364: Minor: Huh, interesting, I would've expected Ruby to right-size the underlying array (e.g. not need the `compact`) when doing `#to_a` on a range.

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -431,6 +447,32 @@ def sample_allocation(obj)
...
         end
 
+        it 'include accurate object sizes' do
+          string_sample = heap_samples.find { |s| s.labels[:'allocation class'] == 'String' }
+          expect(string_sample.values[:'heap-live-size']).to be_between(
+            # 18 UTF-8 characters at minimum.
+            (18 * 2) * sample_rate,
+            # Add some extra padding (per char and object-wide) for extra data.
+            (18 * 4 + 40) * sample_rate
+          )
```
ivoanjo, line 457: So at first pass I was a bit confused here with the math -- why 18*2 (if utf-8 represents ascii chars with 1 byte), etc. I realized actually the size of `a_string` is 40, because it fits within a single slot.

A quick experiment with `ObjectSpace.memsize_of` says you can have 23 ascii chars in there, which is a bit like an array -- 24 bytes is the same 3 words that an array can fit.

All that to say: maybe instead of asserting directly on values, we could rely on Ruby's value to be accurate and directly use it in the equation?

```suggestion
          expect(string_sample.values[:'heap-live-size'])
            .to eq(ObjectSpace.memsize_of(a_string) * sample_rate) 
```

(Same for the others?)

ivoanjo (approved):
Gave it another pass! LGTM :+1:

spec/datadog/profiling/stack_recorder_spec.rb
AlexJF: It might not be needed but you had suggested calling it would enforce it in a previous comment so I added it for good measure

spec/datadog/profiling/stack_recorder_spec.rb
AlexJF: Hmm yeah I was reluctant to use `memsize_of` directly here because that gives us no notion of whether the values returned by `memsize_of` are "reasonable" (or if they'd all be fixed to, for example, 40 due to some issue/bug).

I do agree that the current range-checking has quite arbitrary (and potentially flaky) bounds though so maybe better to do what you suggest and leave correctness checks for some other more "aggregated"/E2E test.

spec/datadog/profiling/stack_recorder_spec.rb
ivoanjo: It's indeed harmless. Usually an array gets created it gets right-sized, and only later appends trigger the growth behavior (although I haven't looked at Range#to_a so I'm assuming it calls the right APIs to do that ;) )

PR #3335, opened 2023-12-15
Remove deprecated HTTP Transport methods

marcotc:
**2.0 Upgrade Guide notes**
Removed unused, non-public methods:
* `Datadog::Tracing::Transport::HTTP#default_hostname`
* `Datadog::Tracing::Transport::HTTP#default_port`
* `Datadog::Tracing::Transport::HTTP#default_url`
* `Datadog::Core::Remote::Transport::HTTP#default_hostname`
* `Datadog::Core::Remote::Transport::HTTP#default_port`
* `Datadog::Core::Remote::Transport::HTTP#default_url`

**Motivation:**
These methods are unused and sit very oddly in the transport object themselves.

**Additional Notes:**
These methods were never public.

**How to test the change?**
All changes are covered by tests (meaning no change in behavior, since these methods are unused).

**For Datadog employees:**


ekump (approved):

lib/datadog/core/remote/transport/http.rb
```diff
@@ -135,33 +135,6 @@ def default_adapter
             Datadog::Core::Transport::Ext::HTTP::ADAPTER
           end
 
-          def default_hostname(logger: Datadog.logger)
-            logger.warn(
-              'Deprecated for removal: Using #default_hostname for configuration is deprecated and will ' \
-                'be removed on a future ddtrace release.'
-            )
-
-            DO_NOT_USE_ENVIRONMENT_AGENT_SETTINGS.hostname
```
ivoanjo, line 144: Minor: I wonder if `DO_NOT_USE_ENVIRONMENT_AGENT_SETTINGS` still gets used in practice, or if it is a no-op default that always gets replaced by whatever we pass in.

marcotc: When I played around with dependency injection, I did find I believe 1 or 2 places that really still use it, and all the other references were non-functional (value is passed, but never used).

PR #3344, opened 2023-12-20
Fix  `error_status_codes` options

TonyCTHsu:
**2.0 Upgrade Guide notes**

**What does this PR do?**

Fix and standardize the way to configure an array of ranges for `error_status_codes` from environment variables.

### DIFF
--- a/lib/datadog/tracing/contrib/excon/configuration/settings.rb
+++ b/lib/datadog/tracing/contrib/excon/configuration/settings.rb
@@ -2,6 +2,8 @@
 
 require_relative '../../../span_operation'
 require_relative '../../configuration/settings'
+require_relative '../../status_range_matcher'
+require_relative '../../status_range_env_parser'
 require_relative '../ext'
 
 module Datadog
@@ -39,20 +41,11 @@ module Datadog
             option :error_status_codes do |o|
               o.env Ext::ENV_ERROR_STATUS_CODES
               o.default 400...600
-              o.env_parser do |value|
-                values = if value.include?(',')
-                           value.split(',')
-                         else
-                           value.split
-                         end
-                values.map! do |v|
-                  v.gsub!(/\A[\s,]*|[\s,]*\Z/, '')
-
-                  v.empty? ? nil : v
-                end
-
-                values.compact!
-                values
+              o.setter do |v|
+                Tracing::Contrib::StatusRangeMatcher.new(v) if v
+              end
+              o.env_parser do |v|
+                Tracing::Contrib::StatusRangeEnvParser.call(v) if v
               end
             end
 
--- a/lib/datadog/tracing/contrib/faraday/configuration/settings.rb
+++ b/lib/datadog/tracing/contrib/faraday/configuration/settings.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 require_relative '../../configuration/settings'
+require_relative '../../status_range_matcher'
+require_relative '../../status_range_env_parser'
 require_relative '../ext'
 
 module Datadog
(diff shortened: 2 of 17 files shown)


lib/datadog/tracing/contrib/status_range_matcher.rb
```diff
@@ -0,0 +1,25 @@
...
+      class StatusRangeMatcher
+        def initialize(ranges)
+          @ranges = Array(ranges)
+        end
+
+        def include?(status)
+          @ranges.any? do |e|
+            case e
+            when Range
+              e.include? status
```
marcotc, line 16: How about using === for both integer and range? I think it works for both.

marcotc (approved):

lib/datadog/tracing/contrib/status_range_matcher.rb
lloeki: Compared to current code I feel like it kind of obscures the reading though: personally I prefer reading the current version which is immediately obvious, whereas I had to check that  `2 === 2` and `(2..4) === 2` both worked.

lib/datadog/tracing/contrib/faraday/configuration/settings.rb
```diff
@@ -42,20 +44,11 @@ class Settings < Contrib::Configuration::Settings
             option :error_status_codes do |o|
               o.env Ext::ENV_ERROR_STATUS_CODES
               o.default 400...600
```
lloeki, line 46: `400...600` includes `600`, is that intended?

lib/datadog/tracing/contrib/excon/configuration/settings.rb
```diff
@@ -39,20 +41,11 @@ class Settings < Contrib::Configuration::Settings
             option :error_status_codes do |o|
               o.env Ext::ENV_ERROR_STATUS_CODES
               o.default 400...600
```
lloeki, line 43: `400...600` includes `600`, is that intended?

lib/datadog/tracing/contrib/excon/configuration/settings.rb
```diff
@@ -39,20 +41,11 @@ class Settings < Contrib::Configuration::Settings
...
-                values.map! do |v|
-                  v.gsub!(/\A[\s,]*|[\s,]*\Z/, '')
-
-                  v.empty? ? nil : v
-                end
-
-                values.compact!
-                values
+              o.setter do |v|
+                Tracing::Contrib::StatusRangeMatcher.new(v) if v
```
lloeki, line 45: I would add a runtime type check here, to enforce that either one of these is true:

- v is an `Integer`
- v is a `Range` with `.first` (and last, but that's internally checked by `Range`) being `Integer`

lib/datadog/tracing/contrib/status_range_env_parser.rb
```diff
@@ -0,0 +1,33 @@
...
+                v = e.split('-')
+
+                case v.length
+                when 0 then next
+                when 1 then array << Integer(v.first)
+                when 2 then array << (Integer(v.first)..Integer(v.last))
+                else
+                  Datadog.logger.debug(
+                    "Invalid error_status_codes configuration: Unable to parse #{value}, containing #{v}."
+                  )
```
lloeki, line 23: Somehow I feel like this logging should happen in `o.env_parser`.

Also, I think we might be inconsistent and in some places env parse fails / conf type checks raise and in others they log but IIRC we mostly raise? I might be wrong on that one but I'd rather be consistent.

If we chose to always log, I feel like we should raise `EnvValueError` / `ParseError` / `TypeError` / whatever exceptions in such pieces of code, and then the caller (the config system) of `o.env_parser` / `o.setter` / ... should handle that kind of exception uniformly and log.

TonyCTHsu:
Going to refactor later

PR #3345, opened 2023-12-21
Add Rails routing instrumentation

lloeki:
**2.0 Upgrade Guide notes**

**What does this PR do?**

- Add a Rails routing span, inspired by Sinatra
- Fetch the resolved route from Journey and tag it to the routing span

**Motivation:**

API Catalog + API Security need it.

**Additional Notes:**

First commit by myself is crude, but implements the core of the solution, this should get the ball rolling for teh APM team to take over.

**How to test the change?**

To be implemented by APM team.

**For Datadog employees:**


lloeki:
Things that should be made working (if they don't work already):

- `get '/foo/:id' => 'mycontroller#myaction'` (`rails.route` should == `/whatevs/:id/foo`)
- `resource 'whatevs' do get 'foo' => 'mycontroller#myaction' end` and stuff like that (`rails.route` should == `/whatevs/:id/foo`)
- `mount '/foo' => Rack app || Sinatra app || Rails engine (`rails.route` should == `/foo`)
- `root 'mycontroller#myaction'` (`rails.route` should == `/`)
- whatever happens when no route matches
- this is for Rails 7, I think it works the same for some earlier major versions but will probably fail on Rails 3 and 4 and maybe fail on Rails 5 (depending on when they started to use Journey / if Journey changed, which I can't recall)

lib/datadog/tracing/contrib/grape/endpoint.rb
```diff
@@ -101,6 +101,9 @@ def endpoint_run(name, start, finish, id, payload)
 
                 span.set_tag(Tracing::Metadata::Ext::HTTP::TAG_METHOD, request_method)
                 span.set_tag(Tracing::Metadata::Ext::HTTP::TAG_URL, path)
+
+                # TEMP REMOVE ONCE SENT TO RACK // ENSURE APPLICATION ROOT IS PREPENDED IN RACK VIA URL#
```
marcotc, line 105: To accomplish this, I suggest adding the Grape route to the active Rack `env`.

This way the Rack integration can read that `env` value and set the tag appropriately.

marcotc:
I read the requirements and I think we should have the tag `http.route` only on the Rack span, for Rack-based web frameworks. The `http.route` is always scoped by `service`, so routes-by-service should be canonical with the top-most instrumented web span (Rack in this case).

This way there's no ambiguity to which route is effective for a `service`.

When Rack is nested with another Rack stack, the last one (top-most Rack span) will win and count towards the Rack route for the service. We'd still set the route tag on all Rack spans that we create, but the API Catalog would only read the top-level one. This is for a trace with a `service` and multiple Rack stacks.
If there are multiple `services`, then each top-most Rack span would be read by the API Catalog naturally, so things would still work as expected.

lib/datadog/tracing/contrib/rails/patcher.rb
```diff
@@ -5,11 +5,59 @@
...
+
+        # Patcher to trace rails routing done by JourneyRouter
+        module JourneyRouterPatch
+          def find_routes(*args, **kwargs)
+            result = super
+
+            if Datadog::Tracing.enabled? && (span = Datadog::Tracing.active_span)
+              if result.any?
+                datadog_route = result.first[2].path.spec.to_s
+              end
```
lloeki, line 43: Otherwise `datadog_route` is `nil`, but we still record it as the router failing to match.

lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -93,13 +93,38 @@ def call(env)
...
+              # note above
+              if last_route
+                # TODO: concatenate better? (according to spec I think it's fine /-wise)
+                composite_route = last_script_name + last_route
+                composite_path = last_script_name + last_route
+
+                request_span.set_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE, composite_route)
+
+                # TODO: should it be composite_route?
+                request_span.resource = env['REQUEST_METHOD'] + ' ' + composite_path
```
lloeki, line 124: As is it doesn't work because of the various cases of resource setting in `ensure` below, but I believe this is what it should be on the Rack span, which handles HTTP.

I'm not sure about resource naming, it's all very confusing, especially when it ends up being `Controller#action` which to me makes no sense at all however I try to slice it.

lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -93,13 +93,38 @@ def call(env)
...
             # Copy the original env, before the rest of the stack executes.
             # Values may change; we want values before that happens.
             original_env = env.dup
 
             # call the rest of the stack
             status, headers, response = @app.call(env)
+
+            # TODO: might be wrong if say rails routed to sinatra but sinatra
+            #       fails to find a route, so I dumbly checked for 404?
+            if status != 404 && (routed = Thread.current[:datadog_http_routing].last)
```
lloeki, line 107: We should probably log `Thread.current[:datadog_http_routing]` because from what I've seen it can be very surprising at times and without this full thing nor the intermediate routing spans and tags it's absolutely impossible to debug anything.

zarirhamza:
Most TODOs are addressed but things that are left:

- Need to confirm that `Thread.current` is the best way to proceed with propagating values
- Determine if it's even feasible to change the resource name for the spans without breaking dashboards
- TESTING

marcotc:
The current error in CI happens because `ActionDispatch::Journey` is not available in Rails 3: https://github.com/rails/rails/tree/v3.2.22.5/actionpack/lib/action_dispatch

marcotc:
System tests for Rails 4 are failing with this error:
```
D, [2024-01-13T01:11:48.291837 #19] DEBUG -- ddtrace: [ddtrace] (/binaries/dd-trace-rb/lib/datadog/tracing/transport/traces.rb:105:in `encode_trace') Flushing trace: [{"error":1,"meta":{"key1":"val1","key2 ":" val2","env":"system-tests","version":"1.0.0","_dd.runtime_family":"ruby","_dd.appsec.waf.version":"1.14.0","http.client_ip":"127.0.0.1","_dd.appsec.event_rules.version":"1.8.0","error.type":"NoMethodError","error.message":"undefined method `env' for #<Hash:0x00007f6681922e28>","error.stack":"/binaries/dd-trace-rb/lib/datadog/tracing/contrib/rails/patcher.rb:19:in `find_routes': undefined method `env' for #<Hash:0x00007f6681922e28> (NoMethodError)\n\tfrom /usr/local/bundle/gems/actionpack-4.1.16/lib/action_dispatch/journey/router.rb:59:in `call'\n\tfrom /usr/local/bundle/gems/actionpack-4.1.16/lib/action_dispatch/routing/route_set.rb:690:in `call'\n\tfrom /usr/local/bundle/gems/warden-1.2.7/lib/warden/manager.rb:36:in `block in call'\n\tfrom /usr/local/bundle/gems/warden-1.2.7/lib/warden/manager.rb:35:in `catch'\n\tfrom /usr/local/bundle/gems/warden-1.2.7/lib/warden/manager.rb:35:in `call'\n\tfrom /usr/local/bundle/gems/rack-1.5.5/lib/rack/etag.rb:23:in `call'\n\tfrom /usr/local/bundle/gems/rack-1.5.5/lib/rack/conditionalget.rb:25:in `call'\n\tfrom /usr/local/bundle/gems/rack-1.5.5/lib/rack/head.rb:11:in `call'\n\tfrom /usr/local/bundle/gems/actionpack-4.1.16/lib/action_dispatch/middleware/params_parser.rb:27:in `call'\n\tfrom /usr/local/bundle/gems/actionpack-4.1.16/lib/action_dispatch/middleware/flash.rb:254:in `call'\n\tfrom /usr/local/bundle/gems/rack-1.5.5/lib/rack/session/abstract/id.rb:225:in `context'\n\tfrom /usr/local/bundle/gems/rack-1.5.5/lib/rack/session/abstract/id.rb:220:in `call'\n\tfrom /usr/local/bundle/gems/actionpack-4.1.16/lib/action_dispatch/middleware/cookies.rb:562:in `call'\n\tfrom /usr/local/bundle/gems/activerecord-4.1.16/lib/active_record/query_cache.rb:36:in `call'\n\tfrom /usr/local/bundle/gems/activerecord-4.1.16/lib/active_record/connection_adapters/abstract/connection_pool.rb:621:in `call'\n\tfrom /usr/local/bundle/gems/actionpack-4.1.16/lib/action_dispatch/middleware/callbacks.rb:29:in `block in call'\n\tfrom /usr/local/bundle/gems/activesupport-4.1.16/lib/active_support/callbacks.rb:82:in `run_callbacks'\n\tfrom /usr/local/bundle/gems/actionpack-4.1.16/lib/action_dispatch/middleware/callbacks.rb:27:in `call'\n\tfrom /usr/local/bundle/gems/actionpack-4.1.16/lib/action_dispatch/middleware/remote_ip.rb:76:in `call'\n\tfrom /binaries/dd-trace-rb/lib/datadog/tracing/contrib/rails/middlewares.rb:19:in `call'\n\tfrom /usr/local/bundle/gems/actionpack-4.1.16/lib/action_dispatch/middleware/debug_exceptions.rb:17:in `call'\n\tfrom /usr/local/bundle/gems/actionpack-4.1.16/lib/action_dispatch/middleware/show_exceptions.rb:30:in `call'\n\tfrom /usr/local/bundle/gems/railties-4.1.16/lib/rails/rack/logger.rb:38:in `call_app'\n\tfrom /usr/local/bundle/gems/railties-4.1.16/lib/rails/rack/logger.rb:20:in `block in call'\n\tfrom /usr/local/bundle/gems/activesupport-4.1.16/lib/active_support/tagged_logging.rb:68:in `block in tagged'\n\tfrom /usr/local/bundle/gems/activesupport-4.1.16/lib/active_support/tagged_logging.rb:26:in `tagged'\n\tfrom /usr/local/bundle/gems/activesupport-4.1.16/lib/active_support/tagged_logging.rb:68:in `tagged'\n\tfrom /usr/local/bundle/gems/railties-4.1.16/lib/rails/rack/logger.rb:20:in `call'\n\tfrom /usr/local/bundle/gems/actionpack-4.1.16/lib/action_dispatch/middleware/request_id.rb:21:in `call'\n\tfrom /usr/local/bundle/gems/rack-1.5.5/lib/rack/methodoverride.rb:21:in `call'\n\tfrom /usr/local/bundle/gems/rack-1.5.5/lib/rack/runtime.rb:17:in `call'\n\tfrom /usr/local/bundle/gems/activesupport-4.1.16/lib/active_support/cache/strategy/local_cache_middleware.rb:26:in `call'\n\tfrom /usr/local/bundle/gems/rack-1.5.5/lib/rack/sendfile.rb:112:in `call'\n\tfrom /binaries/dd-trace-rb/lib/datadog/appsec/contrib/rack/request_middleware.rb:60:in `block (2 levels) in call'\n\tfrom /binaries/dd-trace-rb/lib/datadog/appsec/instrumentation/gateway.rb:37:in `block in push'\n\tfrom /binaries/dd-trace-rb/lib/datadog/appsec/contrib/rack/gateway/watcher.rb:56:in `block in watch_request'\n\tfrom /binaries/dd-trace-rb/lib/datadog/appsec/instrumentation/gateway.rb:19:in `call'\n\tfrom /binaries/dd-trace-rb/lib/datadog/appsec/instrumentation/gateway.rb:43:in `block (2 levels) in push'\n\tfrom /binaries/dd-trace-rb/lib/datadog/appsec/instrumentation/gateway.rb:47:in `push'\n\tfrom /binaries/dd-trace-rb/lib/datadog/appsec/contrib/rack/request_middleware.rb:59:in `block in call'\n\tfrom /binaries/dd-trace-rb/lib/datadog/appsec/contrib/rack/request_middleware.rb:58:in `catch'\n\tfrom /binaries/dd-trace-rb/lib/datadog/appsec/contrib/rack/request_middleware.rb:58:in `call'\n\tfrom /binaries/dd-trace-rb/lib/datadog/tracing/contrib/rack/middlewares.rb:106:in `call'\n\tfrom /usr/local/bundle/gems/railties-4.1.16/lib/rails/engine.rb:514:in `call'\n\tfrom /usr/local/bundle/gems/railties-4.1.16/lib/rails/application.rb:144:in `call'\n\tfrom /usr/local/bundle/gems/puma-5.6.1/lib/puma/configuration.rb:252:in `call'\n\tfrom /usr/local/bundle/gems/puma-5.6.1/lib/puma/request.rb:77:in `block in handle_request'\n\tfrom /usr/local/bundle/gems/puma-5.6.1/lib/puma/thread_pool.rb:340:in `with_force_shutdown'\n\tfrom /usr/local/bundle/gems/puma-5.6.1/lib/puma/request.rb:76:in `handle_request'\n\tfrom /usr/local/bundle/gems/puma-5.6.1/lib/puma/server.rb:441:in `process_client'\n\tfrom /usr/local/bundle/gems/puma-5.6.1/lib/puma/thread_pool.rb:147:in `block in spawn_thread'\n","component":"rack","operation":"request","span.kind":"server","http.method":"GET","http.url":"/","http.base_url":"http://localhost:7777","http.status_code":"500","http.useragent":"curl/7.52.1","http.request.headers.user-agent":"curl/7.52.1","_dd.p.dm":"-1","language":"ruby","runtime-id":"76a3d8d8-5749-4c60-add4-5e5d39574210","_dd.p.tid":"65a1e35400000000"},"metrics":{"_dd.appsec.enabled":1.0,"_dd.appsec.waf.timeouts":0.0,"_dd.appsec.waf.duration":330.015,"_dd.appsec.waf.duration_ext":568.77,"_dd.measured":1.0,"_dd.top_level":1.0,"_dd.agent_psr":1.0,"process_id":19.0,"_sampling_priority_v1":1.0,"_dd.profiling.enabled":0.0},"name":"rack.request","parent_id":0,"resource":"GET 500","service":"weblog","span_id":2355416730313891479,"trace_id":4463725121677521616,"type":"web","start":1705108308270437492,"duration":8158101}]
```

marcotc:
I think I fixed system tests, let's see what they say.

lib/datadog/tracing/contrib/rails/patcher.rb
```diff
@@ -5,11 +5,22 @@
...
 
 module Datadog
   module Tracing
     module Contrib
       module Rails
+        # Patcher to trace rails routing done by JourneyRouter
+        module JourneyRouterPatch
+          def find_routes(*args)
+            result = super
+            integration_route = result.first[2].path.spec.to_s.gsub(/\(\.:format\)/, '') if result.any?
```
TonyCTHsu, line 18: If `result.any?` returns `false`, is it still required to push into `Thread.current[:datadog_http_routing]`?

lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -93,13 +96,27 @@ def call(env)
...
             original_env = env.dup
 
             # call the rest of the stack
             status, headers, response = @app.call(env)
+
+            if status != 404 && (routed = Thread.current[:datadog_http_routing].last)
+              last_script_name = routed[1]
+              last_route = routed[2]
+
+              if last_script_name == '' && env['SCRIPT_NAME'] != ''
```
TonyCTHsu, line 112: Can we add a comment explaining about such exceptional case?

lib/datadog/tracing/contrib/rails/ext.rb
```diff
@@ -12,6 +12,10 @@ module Ext
           ENV_ANALYTICS_ENABLED = 'DD_TRACE_RAILS_ANALYTICS_ENABLED'
           ENV_ANALYTICS_SAMPLE_RATE = 'DD_TRACE_RAILS_ANALYTICS_SAMPLE_RATE'
           ENV_DISABLE = 'DISABLE_DATADOG_RAILS'
+          SPAN_ROUTE = 'rails.route'
```
TonyCTHsu, line 15: How are those constants useful?

TonyCTHsu (dismissed):
I believe there are quite some tests missing. 

1. Add tests for the changes on `grape`, `sinatra`, `rack` 
2. Add test case for path containing path parameter
3. Add test case for integrated behaviour, for example: mounting Rails engine or Sinatra app on Rails app 

Consider those scenarios written in test suite would help me understand the outcome.

lib/datadog/tracing/contrib/rails/patcher.rb
```diff
@@ -41,6 +52,9 @@ def before_initialize(app)
               # Sometimes we don't want to activate middleware e.g. OpenTracing, etc.
               add_middleware(app) if Datadog.configuration.tracing[:rails][:middleware]
 
+              # ActionDispatch::Journey is not available or incompatible in Rails < 4.2.
+              ActionDispatch::Journey::Router.prepend(JourneyRouterPatch) if Integration.version >= Gem::Version.new('4.2')
```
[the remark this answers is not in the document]
lloeki: Sounds sensible. This is the kind of decision I cannot take, not being a stakeholder in `Tracing`.

lib/datadog/tracing/contrib/rails/patcher.rb
lloeki: Depends on whether you want to record the trip to the router even though it failed to match, which can come in really useful when it comes to debugging.

lib/datadog/tracing/contrib/rails/ext.rb
lloeki: Maybe they're not anymore. I do believe that having all intermediate routing spans is super useful.

Ideally:
- These should probably tag `SCRIPT_NAME` and `PATHINFO` as well.
- I would also add a Rack `UrlMap` instrumentation that creates a `rack.map` span with a `rack.map.path` (not `route` as it's not a route, but that's debatable) for having full trace clarity on routing behaviour observability.

lloeki:
> I believe there are quite some tests missing.

Certainly. Reminder that my involvement here is limited to "make it work", as I had a clear path towards an implementation based on prior experience doing the exact same thing.

Overall, steps are:

- Make it work (on me, done)
- Make it right (on Zarir, with APM Tracing help)
- Make it fast (if needed)

> Add test case for integrated behaviour, for example: mounting Rails engine or Sinatra app on Rails app

Using https://github.com/lloeki/minimal-rack/tree/datadog as a base, such tests should cover:

```
curl -vvv 127.0.0.1:3000/hello/world                                   # rails route (get) => rails controller
curl -vvv 127.0.0.1:3000/rack/hello/world                              # rails route (mount) => rack app
curl -vvv 127.0.0.1:3000/sinatra/hello/world                           # rails route (mount) => sinatra app
curl -vvv 127.0.0.1:3000/grape/hello/world                             # rails route (mount) => grape app
curl -vvv 127.0.0.1:3000/rack/sinatra/hello/world                      # rails route (mount) => rack app (map) => sinatra app
curl -vvv 127.0.0.1:3000/rack/grape/hello/world                        # rails route (mount) => rack app (map) => grape app
curl -vvv 127.0.0.1:3000/engine/simple/hello/world                     # rails route (mount) => rails engine (route) => rails engine controller
curl -vvv 127.0.0.1:3000/engine/endpoint/hello/world                   # rails route (mount) => rails engine (endpoint) => rack app
curl -vvv 127.0.0.1:3000/engine/endpoint/grape/hello/world             # rails route (mount) => rails engine (endpoint) => rack app (map) => grape app
curl -vvv 127.0.0.1:3000/engine/endpoint/sinatra/hello/world           # rails route (mount) => rails engine (endpoint) => rack app (map) => sinatra app
curl -vvv 127.0.0.1:3000/rack/rack/rack/rack/rack/hello/world.         # deep rack calls
curl -vvv 127.0.0.1:3000/rack/engine/endpoint/sinatra/hello/world.     # rails mounts rack app at `/rack`, rack app maps `engine/endpoint` to rails engine (which is just a rack app itself) that wraps rack app, which maps `/sinatra` to sinatra app

```

lib/datadog/tracing/contrib/rack/middlewares.rb
lloeki: I can't seem to wrap my head around it either.

In the conditioned code:

- `last_route` would always be `''`
- `last_script_name = env['PATH_INFO']` means there's presumably interesting info in `PATH_INFO` but what of the `env['SCRIPT_NAME']` value, which is `!= ''`?

I would also be wary of what the `env['PATH_INFO']` and `env['SCRIPT_NAME']` values would be assumed to be in case of nesting: if they've been mutated at each stage of nesting they may not be the original values at the beginning of the request. This may or may not be what's intended to be relied upon here, but that's very unclear.

Also I can't recall what happens when e.g NGiNX is in front and proxying under a path, e.g NGiNX `example.com/foo/(.*)` is proxied to `app:3000/$1` (`SCRIPT_NAME` may be already set to `/foo` CGI-style - that's the part I can't recall - even at the very top of the Rack stack)

lloeki: > if they've been mutated at each stage of nesting they may not be the original values

At least Rack `map` ensures they're being restored to the previous value. Dunno about other cases.

https://github.com/rack/rack/blob/3897649e8740e560a5fa142f972121a119b26b5c/lib/rack/urlmap.rb#L81-L84

hestonhoffman (approved):
No review necessary from the docs team

spec/datadog/core/environment/execution_spec.rb
```diff
@@ -107,6 +107,7 @@
             gemfile(true) do
               source 'https://rubygems.org'
               gem 'spring', '>= 2.0.2'
+              gem 'concurrent-ruby', '#{Gem.loaded_specs['concurrent-ruby'].version}'
```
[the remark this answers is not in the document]
zarirhamza: Apparently this fixes a system test failure according to @marcotc

lib/datadog/tracing/contrib/rack/middlewares.rb
zarirhamza: > last_route would always be ''

I changed the struct of the item in `Thread.current` so that is in the form {integration, full_path, integration_route} where full_path -> env['SCRIPT_INFO'] and integration_route -> the end route value according to that integration. So last_route will not be '' ever except for the exception case detailed below

zarirhamza: Update with comment but quick summary is that for request of nested rack requests ONLY, we essentially get just the root of the first rack request and all other nested rack requests are ignored. Leading to things like the route of `/rack/rack/hello/world` to be `/rack` which is incorrect. This fixes it by ensuring that there exists a route and appending it only when certain conditions are met in a nested rack request ONLY

lib/datadog/tracing/contrib/grape/endpoint.rb
```diff
@@ -94,20 +95,27 @@ def endpoint_run(name, start, finish, id, payload)
...
                 span.set_tag(Ext::TAG_ROUTE_ENDPOINT, api_view) unless api_view.nil?
                 span.set_tag(Ext::TAG_ROUTE_PATH, path)
                 span.set_tag(Ext::TAG_ROUTE_METHOD, request_method)
 
                 span.set_tag(Tracing::Metadata::Ext::HTTP::TAG_METHOD, request_method)
                 span.set_tag(Tracing::Metadata::Ext::HTTP::TAG_URL, path)
+
+                if Thread.current.key?(:datadog_http_routing) && Thread.current[:datadog_http_routing].is_a?(Array)
+                  Thread.current[:datadog_http_routing] << [:grape, endpoint.env['SCRIPT_NAME'], integration_route]
+                end
```
marcotc, line 110:
```suggestion
                trace.set_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE, endpoint.env['SCRIPT_NAME'] + integration_route)
```

Would this work instead?

marcotc (approved):

lib/datadog/tracing/contrib/sinatra/tracer.rb
```diff
@@ -73,6 +73,10 @@ def route_eval
 
                 Contrib::Analytics.set_measured(span)
 
+                _, path = env['sinatra.route'].split(' ', 2)
+                trace.set_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE, path)
+                trace.set_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE_PATH, env['SCRIPT_NAME'])
+                # binding.pry
```
ivoanjo, line 79: Leftover pry call >_> :wink:

ivoanjo (commented):
Should we update this PR title and description? Looking at the contents, it seems we ended up changing a lot more than just the Rails routing (e.g. this change is relevant for customers not using Rails as well!)

lib/datadog/tracing/contrib/sinatra/tracer.rb
zarirhamza: Fixed in #3456

PR #3346, opened 2023-12-22
[PROF-8669] Benchmarks for core heap profiling phases

AlexJF:
**What does this PR do?**
Add a couple of benchmarks focusing on the core phases of heap profiling:
* What's the impact on the sampling side.
* What's the impact on serialization side (both within GVL - preparation - and outside the GVL - full serialization).

The benchmarks are currently running through the normal flow (i.e. involving stack recorder and such) which means they also capture some extra work unrelated to heap profiling like stack walking, generation of non-heap samples or ddprof serialization. I think this is valuable to get a feeling of overall overhead in normal operation and not focus too much on local optimizations that may have a very small impact in total runtime.

**Motivation:**
* Get a feeling for heap profiling overheads.
* Do some extra stress testing of the heap recorder mechanisms.

**Additional Notes:**

**How to test the change?**

#### Sampling

```
❯ bundle exec ruby benchmarks/profiler_heap_sampling.rb     
DogStatsD reporting ❌ disabled
no heap - single stack
                        37.125k i/100ms
 heap - single stack    27.479k i/100ms
no heap - single stack
                        361.009k (± 6.2%) i/s -      3.601M in  10.014774s
 heap - single stack    280.275k (± 3.3%) i/s -      2.803M in  10.013193s

Comparison:
no heap - single stack:   361008.9 i/s
 heap - single stack:   280274.7 i/s - 1.29x  slower

DogStatsD reporting ❌ disabled
no heap - many stacks
                         4.243k i/100ms
  heap - many stacks     3.048k i/100ms
no heap - many stacks
                         43.056k (± 0.8%) i/s -    432.786k in  10.052483s
  heap - many stacks     29.916k (± 3.6%) i/s -    301.752k in  10.099722s

Comparison:
no heap - many stacks:    43055.6 i/s
  heap - many stacks:    29916.3 i/s - 1.44x  slower

DogStatsD reporting ❌ disabled
heap profiling - sample_rate=1
                         2.834k i/100ms
heap profiling - sample_rate=10
                         4.034k i/100ms
heap profiling - sample_rate=100
                         3.906k i/100ms
heap profiling - sample_rate=1
                         29.322k (± 3.8%) i/s -    294.736k in  10.066480s
heap profiling - sample_rate=10
                         39.827k (± 3.4%) i/s -    399.366k in  10.039527s
heap profiling - sample_rate=100
                         41.736k (± 2.6%) i/s -    417.942k in  10.020877s

Comparison:
heap profiling - sample_rate=100:    41736.3 i/s
heap profiling - sample_rate=10:    39827.3 i/s - same-ish: difference falls within error
heap profiling - sample_rate=1:    29321.7 i/s - 1.42x  slower
```

#### Serialization

```
❯ bundle exec ruby benchmarks/profiler_heap_serialization.rb
DogStatsD reporting ❌ disabled
full serialization without heap profiling
                        30.506k i/100ms
heap sampling serialization preparation - samples
                       214.000  i/100ms
heap sampling serialization preparation - samples+size
                       223.000  i/100ms
full serialization with heap profiling - samples
                        11.000  i/100ms
full serialization with heap profiling - samples+size
                        11.000  i/100ms
full serialization without heap profiling
                        299.932k (± 2.6%) i/s -      3.020M in  10.076043s
heap sampling serialization preparation - samples
                          2.162k (± 4.7%) i/s -     21.614k in  10.023727s
heap sampling serialization preparation - samples+size
                          2.163k (± 3.6%) i/s -     21.631k in  10.015366s
full serialization with heap profiling - samples
                        109.297  (± 1.8%) i/s -      1.100k in  10.067532s
full serialization with heap profiling - samples+size
                        107.649  (± 0.9%) i/s -      1.078k in  10.015311s

Comparison:
full serialization without heap profiling:   299932.4 i/s
heap sampling serialization preparation - samples+size:     2162.9 i/s - 138.67x  slower
heap sampling serialization preparation - samples:     2161.7 i/s - 138.75x  slower
full serialization with heap profiling - samples:      109.3 i/s - 2744.21x  slower
full serialization with heap profiling - samples+size:      107.6 i/s - 2786.20x  slower

DogStatsD reporting ❌ disabled
heap profiling serialization - single stack
                        10.000  i/100ms
heap profiling serialization - many stacks
                         1.000  i/100ms
heap profiling serialization - single stack
                        105.570  (± 1.9%) i/s -      1.060k in  10.044343s
heap profiling serialization - many stacks
                          6.222  (± 0.0%) i/s -     63.000  in  10.126989s

Comparison:
heap profiling serialization - single stack:      105.6 i/s
heap profiling serialization - many stacks:        6.2 i/s - 16.97x  slower

DogStatsD reporting ❌ disabled
heap profiling serialization with sampling rate of 1
                         1.000  i/100ms
heap profiling serialization with sampling rate of 10
                         5.000  i/100ms
heap profiling serialization with sampling rate of 100
                        50.000  i/100ms
heap profiling serialization with sampling rate of 1
                          6.173  (± 0.0%) i/s -     62.000  in  10.044963s
heap profiling serialization with sampling rate of 10
                         60.810  (± 1.6%) i/s -    610.000  in  10.034120s
heap profiling serialization with sampling rate of 100
                        513.250  (± 1.8%) i/s -      5.150k in  10.037699s

Comparison:
heap profiling serialization with sampling rate of 100:      513.3 i/s
heap profiling serialization with sampling rate of 10:       60.8 i/s - 8.44x  slower
heap profiling serialization with sampling rate of 1:        6.2 i/s - 83.14x  slower
```

**For Datadog employees:**
- [x] If this PR touches code that signs or publishes builds or packages, or handles

### DIFF
--- a/benchmarks/profiler_heap_sampling.rb
+++ b/benchmarks/profiler_heap_sampling.rb
@@ -1,0 +1,183 @@
+# Used to quickly run benchmark under RSpec as part of the usual test suite, to validate it didn't bitrot
+VALIDATE_BENCHMARK_MODE = ENV['VALIDATE_BENCHMARK'] == 'true'
+
+return unless __FILE__ == $PROGRAM_NAME || VALIDATE_BENCHMARK_MODE
+
+require 'benchmark/ips'
+require 'ddtrace'
+require 'pry'
+require_relative 'dogstatsd_reporter'
+
+METRIC_VALUES = { 'cpu-time' => 0, 'cpu-samples' => 0, 'wall-time' => 0, 'alloc-samples' => 1, 'timeline' => 0 }.freeze
+OBJECT_CLASS = 'object'.freeze
+
+def sample_object(recorder, depth = 0)
+  if depth <= 0
+    Datadog::Profiling::StackRecorder::Testing._native_track_object(
+      recorder,
+      Object.new,
+      1,
+      OBJECT_CLASS,
+    )
+    Datadog::Profiling::Collectors::Stack::Testing._native_sample(
+      Thread.current,
+      recorder,
+      METRIC_VALUES,
+      [],
+      [],
+      400,
+      false
+    )
+  else
+    sample_object(recorder, depth - 1)
+  end
+end
+
+# This benchmark measures the performance of the sampling operations with different flavours of heap profiling
+# with single stacks
+Benchmark.ips do |x|
+  benchmark_time = VALIDATE_BENCHMARK_MODE ? { time: 0.01, warmup: 0 } : { time: 10, warmup: 2 }
+  x.config(
+    **benchmark_time,
+    suite: report_to_dogstatsd_if_enabled_via_environment_variable(benchmark_name: 'profiler_heap_sampling')
+  )
+
+  recorder_without_heap = Datadog::Profiling::StackRecorder.new(
+    cpu_time_enabled: false,
+    alloc_samples_enabled: false,
+    heap_samples_enabled: false,
+    heap_size_enabled: false,
+    heap_sample_every: 1,
+    timeline_enabled: false,
+  )
+
+  recorder_with_heap = Datadog::Profiling::StackRecorder.new(
+    cpu_time_enabled: false,
+    alloc_samples_enabled: true,
+    heap_samples_enabled: true,
+    heap_size_enabled: true,
+    heap_sample_every: 1,
+    timeline_enabled: false,
+  )
+
+  x.report('no heap - single stack') do
+    sample_object(recorder_without_heap)
+  end
+
+  x.report('heap - single stack') do
+    sample_object(recorder_with_heap)
+  end
+
+  x.save! 'profiler-heap-sampling-results-single-stack.json' unless VALIDATE_BENCHMARK_MODE
+  x.compare!
+end
+
+# This benchmark measures the performance of the sampling operations with different flavours of heap profiling
+# with many stacks
+Benchmark.ips do |x|
+  benchmark_time = VALIDATE_BENCHMARK_MODE ? { time: 0.01, warmup: 0 } : { time: 10, warmup: 2 }
+  x.config(
+    **benchmark_time,
+    suite: report_to_dogstatsd_if_enabled_via_environment_variable(benchmark_name: 'profiler_heap_sampling_many_stacks')
+  )
+
+  recorder_without_heap = Datadog::Profiling::StackRecorder.new(
+    cpu_time_enabled: false,
+    alloc_samples_enabled: false,
+    heap_samples_enabled: false,
+    heap_size_enabled: false,
+    heap_sample_every: 1,
+    timeline_enabled: false,
+  )
+
+  recorder_with_heap = Datadog::Profiling::StackRecorder.new(
+    cpu_time_enabled: false,
+    alloc_samples_enabled: true,
+    heap_samples_enabled: true,
+    heap_size_enabled: true,
+    heap_sample_every: 1,
+    timeline_enabled: false,
+  )
+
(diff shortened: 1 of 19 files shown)


benchmarks/profiler_heap_sampling.rb
```diff
@@ -0,0 +1,183 @@
...
+      [],
+      [],
+      400,
+      false
+    )
+  else
+    sample_object(recorder, depth - 1)
+  end
+end
+
```
ivoanjo, line 35: I'm thinking this benchmark may be a bit... inaccurate in its current form. That's because "no heap" is basically sampling the same thing over and over again, which due to deduping means it's basically bumping a counter somewhere inside libdatadog and thus every iteration is truly the same.

BUT the "heap" variant is actually tracking more and more and more objects. Which means that every iteration is probably going to be slightly slower than the last one, and thus it's kinda hard to make an apples-to-apples comparison. In fact, making the code go faster may make it go slower, because it will allocate even more and thus have more data to wrangle with -- something like that?

I took a peek at the benchmark/ips docs to see if there was some way to fix the number of times the benchmark runs, instead of the time. Because if we said "go for X iterations", at least we could have some comparison of how much time X samples take and it would go lower with each optimization. I didn't find an easy way of doing it.

Hmmm... Thoughts?

(Edit: I guess we could add a `_native_reset_heap_recorder` or something like that that did the simplest thing it could to "untrack" the object and the stack?)

benchmarks/profiler_heap_serialization.rb
```diff
@@ -0,0 +1,249 @@
...
+  when :object
+    BenchmarkObject.new(depth)
+  end
+end
+
+def create_benchmark_live_objects(num_objects: 1_000_000)
+  live_objects = []
+
+  num_objects.times { live_objects << random_object }
+end
```
ivoanjo, line 55: I think this ended up not being used?

benchmarks/profiler_heap_serialization.rb
```diff
@@ -0,0 +1,249 @@
...
+    timeline_enabled: false,
+  )
+
+  live_objects = Array.new(VALIDATE_BENCHMARK_MODE ? 100 : 10_000) { BenchmarkObject.new(0) }
+
+  live_objects.each_with_index do |obj, _i|
+    sample_object(recorder_without_heap, obj)
+    sample_object(recorder_with_heap_samples_only, obj)
+    sample_object(recorder_with_full_heap, obj)
+  end
```
ivoanjo, line 117: Wait, I'm somewhat confused by what's going on here. Each `BenchmarkObject` holds inside it references to a bunch of objects that are the result of calling `random_value`.

But... since we only really sample the `BenchmarkObject` instance itself... how does all that fancyness differ from sampling `Array.new(...) { Object.new }`? As far as the profiler goes, it never sees anything other that instances of `BenchmarkObject`, so all of the work above to randomly do stuff seems mostly a no-op? Am I missing something O_o...?

ivoanjo (commented):
Left a few notes! I'm curious about a few things in the current benchmarks.

(Btw this probably needs a rebase!)

benchmarks/profiler_heap_sampling.rb
AlexJF: Hmm yeah great point :/

We also face a similar issue with the serialization side of things where subsequent iterations are naturally less heavy because the previous one cleared a series of objects.

May need to do some of this stuff outside ips as it feels like we'd want to either support fixing iterations to a set number of enforce the execution of a `before_iteration` that doesn't actively count towards timings.

AlexJF:
Will get back to this later with a different PR

PR #3350, opened 2023-12-27
ekump/add config gaps from transport proc

ekump:
**2.0 Upgrade Guide notes**
The `c.tracing.transport_options` option has been removed. The library can no longer be configured with a custom transport adapter. The default, or test adapter must be used. All agent transport configuration options can be  set via the `Datadog.configure` block, or environment variables. 

The following options have been added to replace configuration options previously only available via transport_options:
- `c.agent.timeout_seconds` or `DD_TRACE_AGENT_TIMEOUT_SECONDS`
- `c.agent.uds_path` or ~`DD_TRACE_AGENT_UDS_PATH`~
- `c.agent.use_ssl` or ~`DD_AGENT_USE_SSL`~

Setting uds path or the use of SSL via DD_TRACE_AGENT_URL continues to work as it did in 1.x.

To use the test adapter set `c.agent.tracing.test_mode.enabled = true`  

**What does this PR do?**
This PR removes the `transport_proc`. Users can no longer define custom transporters. All configuration must be done via either env vars, or programmatic configuration. The following options can now be set without `transport_proc`.
- `uds_path`
- `use_ssl` (HTTP only)
- `timeout_seconds`

Additionally, product-specific `AgentSettingsResolvers` have been removed as there is no added functionality, and the concept of non-core resolvers doesn't really make sense at this time. 

**Motivation:**
Simplification of agent communication configuration. No reasonable justification presented to support custom transport procs. 

**Additional Notes:**

**How to test the change?**
Configure agent communication via newly provided options. Library should function as normal.

**For Datadog employees:**


docs/GettingStarted.md
```diff
@@ -2492,41 +2491,20 @@ To use, first configure your trace Agent to listen by Unix socket, then configur
...
-    t.adapter :test, []
-  }
-end
-```
-
-#### Using a custom transport adapter
-
-Custom adapters can be configured with:
+The `Test` adapter is a no-op transport that can optionally buffer requests. For use in test suites or other
+non-production environments. It is configured by setting c.tracing.test_mode.enabled to true.
```
alai97, line 2503:
```suggestion
The `Test` adapter is a no-op transport that can optionally buffer requests in test suites or other
non-production environments. It is configured by setting `c.tracing.test_mode.enabled` to true.
```

docs/GettingStarted.md
```diff
@@ -2473,14 +2473,13 @@ See [Additional Configuration](#additional-configuration) for more details.
 
 #### Using the Net::HTTP adapter
 
-The `Net` adapter submits traces using `Net::HTTP` over TCP. It is the default transport adapter.
+The `Net` adapter submits traces using `Net::HTTP` over TCP. It is the default transport adapter. Use of `Net::HTTP` is
+implied by setting hostname and port.
```
alai97, line 2477:
```suggestion
implied by setting the host name and port.
```

alai97 (approved):
Looks good, two copy suggestions!

ivoanjo:
Hey! I'm a bit behind on a few other PRs and tasks, but would like to give a careful pass on this one, since I've written most of the `AgentSettingsResolver`.

Do you mind holding off on merging before I can give a review?

ekump:
> Hey! I'm a bit behind on a few other PRs and tasks, but would like to give a careful pass on this one, since I've written most of the `AgentSettingsResolver`.
> 
> Do you mind holding off on merging before I can give a review?

@ivoanjo No problem, this doesn't need to be merged ASAP.

lib/datadog/core/configuration/agent_settings_resolver.rb
```diff
@@ -107,17 +84,13 @@ def configured_hostname
...
-              value: transport_options.hostname,
-            ),
             DetectedConfiguration.new(
               friendly_name: "'c.agent.host'",
               value: settings.agent.host
             ),
             DetectedConfiguration.new(
               friendly_name: "#{Datadog::Core::Configuration::Ext::Agent::ENV_DEFAULT_URL} environment variable",
-              value: parsed_http_url && parsed_http_url.hostname
+              value: parsed_http_url&.hostname
```
marcotc, line 93: IS THIS MODERN RUBY???

lib/datadog/core/configuration/ext.rb
```diff
@@ -27,11 +27,16 @@ module Agent
           # env var has "trace" in it, but it really applies to all products
           ENV_DEFAULT_PORT = 'DD_TRACE_AGENT_PORT'
           ENV_DEFAULT_URL = 'DD_TRACE_AGENT_URL'
+          ENV_DEFAULT_USE_SSL = 'DD_TRACE_AGENT_USE_SSL'
+          ENV_DEFAULT_TIMEOUT_SECONDS = 'DD_TRACE_AGENT_TIMEOUT_SECONDS'
+          ENV_DEFAULT_UDS_PATH = 'DD_TRACE_AGENT_UDS_PATH'
```
marcotc, line 32: Out of curiosity, as these cross-language, or created by us?

marcotc (approved):
Very nice clean up!

lib/datadog/core/configuration/ext.rb
ivoanjo: I was going to comment on something similar -- if these are cross language, no concerns at all.

But if they're new and specific to Ruby, I would perhaps hesitate to add them as they're subsets of `DD_TRACE_AGENT_URL`, which is already supported by all libraries.

In my view, part of the problem with the transport proc started was it wasn't possible to set these values via code, exactly because it wasn't possible to set `DD_TRACE_AGENT_URL` via code. If you wanted to, you could already set them via env vars.

Thus, I think the new options via code look reasonable, but it's weird to add the new env vars because:

a) It creates inconsistency between client libraries. These env vars look very generic, so I can imagine a customer copy-pasting their existing Ruby config to another service, and then get confused when it doesn't work. (This has happened in the past with other env vars -- customers open an issue reporting "hey I use DD_FOO with python/java etc but it doesn't work for ruby -- help?)

b) It creates more duplication of settings -- having `DD_TRACE_AGENT_URL` and `DD_TRACE_AGENT_UDS_PATH` means introducing precedence, etc. And as you can kinda tell from the `AgentSettingsResolver`, there's already quite a lot of complexity in all this business.

Note that all of this comment is about the env vars -- I think it's fine to have the new settings for code (e.g. I think it's ok to not have a 1:1 correspondence between every code setting and env var; it's definitely not new to our settings)

ekump: @marcotc These are not cross-language, I made them up. Searching through github it looks like I should probably replace `DD_TRACE_AGENT_UDS_PATH` with `DD_TRACE_AGENT_UNIX_DOMAIN_SOCKET`. For timeout, Python uses `DD_TRACE_AGENT_TIMEOUT_SECONDS` but PHP uses `DD_TRACE_AGENT_TIMEOUT_VAL`. 

@ivoanjo I think your points are valid and I'm in favor of simplifying the resolver as much as possible. I also agree that having multiple ways to set options via env var is a bit weird. But I think we may need the SSL and timeout env vars.
- It is possible to set your host via the cross-language vars `DD_TRACE_AGENT_HOST` and `DD_TRACE_AGENT_PORT` for an http connection. If a user does this, shouldn't they also be able to set the use of SSL via env var? 
- You can't set timeout via `DD_TRACE_AGENT_URL`. I don't think we want to go down the path of query strings?
- The only justification for `DD_TRACE_AGENT_UDS_PATH` is consistency with the http functionality. I don't feel strongly, but if we do keep the timeout and ssl env vars, I don't see the incremental added complexity of a UDS path    var as a big deal. 

If we want to adopt a policy of "If you want to configure via ENV vars and you want to either use SSL for http connections or you want to use UDS then you must use `DD_TRACE_AGENT_URL`" I don't object. I would just want to run it by TEE first to make sure we aren't missing an edge case.

ivoanjo: > It is possible to set your host via the cross-language vars DD_TRACE_AGENT_HOST and DD_TRACE_AGENT_PORT for an http connection. If a user does this, shouldn't they also be able to set the use of SSL via env var?

It's a very good point! I think this is all the result of very organic evolution. Having said that, using ssl is a quite advanced configuration: Afaik the datadog agent doesn't have built-in support for serving ssl. So if you're using ssl, it already means you have some proxying http server in front the agent, which probably means you know what you're doing.

For that reason, I would kinda lean towards letting `DD_TRACE_AGENT_URL` take care of ssl (and uds).

---

As for timeout: one one side, we never actually got asked to support it via env, which probably suggests it's not often used. So I would kinda wait for someone to ask.

But, I think it's reasonable if you want to preeempt that and go ahead. In that case, I'd probably suggest following the python convention, just because the PHP one is awful (cmon -- every setting is a value! it's like suffixing things with `_ENV_VAR` lol).

ekump: @ivoanjo I think that's reasonable. I removed `DD_TRACE_AGENT_UDS_PATH` and `DD_TRACE_AGENT_USE_SSL` but left `DD_TRACE_AGENT_TIMEOUT_SECONDS` as is.

marcotc: Thank you so much for the improvements, @ekump!

marcotc (approved):

lib/datadog/core/transport/ext.rb
```diff
@@ -12,6 +12,7 @@ module HTTP
           ADAPTER = Datadog::Core::Configuration::Ext::Agent::HTTP::ADAPTER
           DEFAULT_HOST = Datadog::Core::Configuration::Ext::Agent::HTTP::DEFAULT_HOST
           DEFAULT_PORT = Datadog::Core::Configuration::Ext::Agent::HTTP::DEFAULT_PORT
+          DEFAULT_TIMEOUT_SECONDS = Datadog::Core::Configuration::Ext::Agent::HTTP::DEFAULT_TIMEOUT_SECONDS
```
ivoanjo, line 15: It's weird that we're duplicating these constants here. Maybe consider removing them? After all, the key responsibility of `AgentSettingsResolver` is to provide the correct information in the `AgentSettingsResolver::AgentSettings` object, and no other parts of the code should care about defaults or whatnot.

lib/datadog/core/transport/http/adapters/net.rb
```diff
@@ -16,14 +16,11 @@ class Net
...
-            DEFAULT_TIMEOUT = 30
-
             # @deprecated Positional parameters are deprecated. Use named parameters instead.
             def initialize(hostname = nil, port = nil, **options)
               @hostname = hostname || options.fetch(:hostname)
               @port = port || options.fetch(:port)
-              @timeout = options[:timeout] || DEFAULT_TIMEOUT
+              @timeout = options[:timeout] || Datadog::Core::Transport::Ext::HTTP::DEFAULT_TIMEOUT_SECONDS
               @ssl = options.key?(:ssl) ? options[:ssl] == true : false
             end
```
ivoanjo, line 26: Looking at the code, I don't think anyone is calling this `initialize` anymore -- it always gets called by `build`. Being that the case, my suggestion would be to actually simplify initialize to directly receive the `agent_settings`.

It's a bit less code + no need to access the timeout constant, it always gets read from the agent settings resolver.

lib/datadog/core/configuration/agent_settings_resolver.rb
```diff
@@ -149,6 +118,73 @@ def configured_port
...
+              value = false
+            else
+              log_warning("Invalid value for #{friendly_name} (#{value.inspect}). Ignoring this configuration.")
+              value = nil
+            end
+          end
+
+          DetectedConfiguration.new(friendly_name: friendly_name, value: value)
+        end
+
```
ivoanjo, line 187: Minor: This method does not seem to be in use anymore ;)
```suggestion
```

lib/datadog/core/configuration/agent_settings_resolver.rb
```diff
@@ -175,40 +214,35 @@ def port
...
           end
         end
 
-        # Defaults to +nil+, letting the adapter choose what default
-        # works best in their case.
-        def timeout_seconds
-          transport_options.timeout_seconds
-        end
+        def parsed_url_uds_path
+          # return nil unless !mixed_http_and_uds? && parsed_url && unix_scheme?(parsed_url)
```
ivoanjo, line 228: Minor: Leftover code?
```suggestion
```

spec/datadog/core/configuration/agent_settings_resolver_spec.rb
```diff
@@ -22,15 +23,15 @@
...
   end
 
   let(:adapter) { :net_http }
   let(:hostname) { '127.0.0.1' }
   let(:port) { 8126 }
   let(:uds_path) { nil }
+  let(:timeout_seconds) { 30 }
 
   before do
     # Environment does not have existing unix socket for the base testing case
```
ivoanjo, line 37: :arrow_down: the " 'when settings.tracing is not present'" block below can now be removed as it was protecting against the tracing section of the settings not being present, e.g. this bit that got removed:

image

ivoanjo (commented):
Thanks for this amazing cleanup, and for the patience for waiting for my review!

One thing I noticed is that there's some telemetry bits that were copy-pasted that completely ignore the settings we have -- e.g. the timeout and ssl.

On a quick check the changes to plumb it in don't seem that big, but technically that mess already existed, so up to you if you feel like fixing it (extra brownie points in my book! xD) or leaving it for another day ;)

lib/datadog/core/configuration/agent_settings_resolver.rb
```diff
@@ -26,25 +25,22 @@ class AgentSettingsResolver
...
               super(
                 adapter,
                 ssl,
                 hostname,
                 port,
                 uds_path,
-                timeout_seconds,
-                deprecated_for_removal_transport_configuration_proc
+                timeout_seconds
               )
```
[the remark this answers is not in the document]
ekump: I've updated this and it simplifies the code a bit, but I believe we want the struct instances to be frozen? So the need for the initializer remains.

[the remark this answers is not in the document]
ivoanjo: One way to remove the initializer would be to move the `.freeze` call to just after we call `AgentSettings.new(...)` in `AgentSettingsResolver#call`.

But I think we can also leave it as-is, only 5 more rubies to drop until we can use [`Data.define`](https://www.shakacode.com/blog/ruby-3-2-adds-a-new-data-class/) :rofl:

ivoanjo (approved):
It may be worth giving a pass at the docs + at the telemetry code (as per my suggestions above) but I don't see anything blocking so here sir, thanks and LGTM :+1:

docs/GettingStarted.md
```diff
@@ -2473,14 +2473,13 @@ See [Additional Configuration](#additional-configuration) for more details.
...
 ```ruby
 Datadog.configure do |c|
-  c.tracing.transport_options = proc { |t|
-    # Hostname, port, and additional options. :timeout is in seconds.
-    t.adapter :net_http, '127.0.0.1', 8126, timeout: 30
-  }
+  c.agent.host = '127.0.0.1'
+  c.agent.port = 8126
 end
 ```
```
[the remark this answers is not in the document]
ekump: @ivoanjo Updated the docs to not mention the adapters directly.

@alai97 Would you mind giving the changes to `GettingsStarted.md` another look?

PR #3357, opened 2024-01-02
[PROF-8037] Profiling system info support

AlexJF:
**2.0 Upgrade Guide notes**

**What does this PR do?**

**Motivation:**

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**


lib/datadog/core/configuration/settings.rb
```diff
@@ -212,11 +212,13 @@ def initialize(*_)
             # This was used prior to the GA of the new CPU Profiling 2.0 profiler. The CPU Profiling 2.0 profiler does not
             # use or need this setting and thus it doesn't do anything.
             option :max_events do |o|
-              o.after_set do
-                Datadog.logger.warn(
-                  'The profiling.advanced.max_events setting has been deprecated for removal and no ' \
-                  'longer does anything. Please remove it from your Datadog.configure block.'
-                )
+              o.after_set do |_, _, precedence|
+                unless precedence == Datadog::Core::Configuration::Option::Precedence::DEFAULT
```
AlexJF, line 216: Without this, a simple `get` on this option (which we do because of the profiler settings dump) will trigger this codepath: 

https://github.com/DataDog/dd-trace-rb/blob/28dc18783c6dd6e6740c180e71a190346e75d9c0/lib/datadog/core/configuration/option.rb#L117

Which will eventually trigger `after_set` even for unset/default values:

https://github.com/DataDog/dd-trace-rb/blob/28dc18783c6dd6e6740c180e71a190346e75d9c0/lib/datadog/core/configuration/option.rb#L299

An alternative could be to just check for a nil value as well. But given that we'll be introducing more and more remote config support which means yet another type of precedence, having that info available in the `after_set` block sounds interesting.

ivoanjo: Yeah... I'm not particularly excited about adding more features to the options system, but I do agree that this seems like it would be relevant for remote configuration and the only alternative I could think of:

```ruby
# doing something like this in the collector itself
(v.class.options.keys - [:force_enable_new_profiler, :legacy_transport_enabled, :force_enable_legacy_profiler, :allocation_counting_enabled]).each { |key| v.get_option(key) }
```

...is super ugly as well.

My suggestion is to keep the approach you have now, but please don't forget to add test coverage for the new feature in `option_spec` ;)

lib/datadog/profiling/collectors/system_info.rb
```diff
@@ -0,0 +1,101 @@
...
+          @application_info ||= {
+            start_time: START_TIME.iso8601,
+            env: @settings.env,
+            service: @settings.service,
+            version: @settings.version,
+          }
+        end
+
+        def collect_profiler_info
+          unless @profiler_info
```
ivoanjo, line 64: As this class reuses the info hashes it gives out on `system_info`, it may be worth sprinkling `.freeze` on all of them, just to avoid accidental mutations somewhere ;)

lib/datadog/profiling/collectors/system_info.rb
```diff
@@ -0,0 +1,101 @@
...
+              collected_value = collect_settings_recursively(value)
+              hash[key] = collected_value unless collected_value.nil?
+            end
+            collected_hash.empty? ? nil : collected_hash
+          elsif v.is_a?(Enumerable)
+            collected_list = v
+              .map { |value| collect_settings_recursively(value) }
+              .reject(&:is_nil?)
+            collected_list.empty? ? nil : collected_list
+          end
```
ivoanjo, line 96: I'm curious: why omit the `nil`s?

ivoanjo (commented):
Left a few notes! I'm excited to finally have this for Ruby! :heart:

lib/datadog/profiling/flush.rb
```diff
@@ -34,6 +36,7 @@ def initialize(
         @code_provenance_data = code_provenance_data
         @tags_as_array = tags_as_array
         @internal_metadata_json = JSON.fast_generate(internal_metadata.map { |k, v| [k, v.to_s] }.to_h)
+        @system_info_json = JSON.fast_generate(system_info)
```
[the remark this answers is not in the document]
AlexJF: Yeah backend is built for full flexibility here (so long as it's valid JSON). I'll remove the enforcement from internal_metadata as well while I'm at it.

lib/datadog/profiling/collectors/system_info.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        def refresh
+          @system_info = {
+            platform: collect_platform_info,
+            runtime: collect_runtime_info,
+            application: collect_application_info,
+            profiler: collect_profiler_info,
+          }
+          self
+        end
```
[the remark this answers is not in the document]
AlexJF: That was the initial idea but yeah I'll simplify for this initial iteration and freeze stuff at initialization time. Later we can add refreshments 😅

lib/datadog/profiling/collectors/system_info.rb
AlexJF: I'm not sure myself. I think it was to try and keep the info as compact as possible but I'll recheck without the nil omitting.

AlexJF: Nothing blew up so removed nil checks!

AlexJF:
image

Profile sent with latest version of this PR

lib/datadog/profiling/exporter.rb
```diff
@@ -65,6 +69,7 @@ def flush
           code_provenance_data: uncompressed_code_provenance,
           tags_as_array: Datadog::Profiling::TagBuilder.call(settings: Datadog.configuration).to_a,
           internal_metadata: internal_metadata,
+          info: info,
```
ivoanjo, line 72: I wonder if we should cache the `internal_metadata` and the `info` as JSON in the constructor, since right now we repeat the work of encoding it every time.

spec/datadog/profiling/exporter_spec.rb
```diff
@@ -24,6 +25,8 @@
     collector
   end
   let(:internal_metadata) { { no_signals_workaround_enabled: no_signals_workaround_enabled } }
+  let(:info) { { profiler: { running_under_test: true } } }
+  let(:info_collector) { instance_double(Datadog::Profiling::Collectors::Info, info: info) }
   let(:no_signals_workaround_enabled) { false }
   let(:logger) { Datadog.logger }
   let(:options) { {} }
```
ivoanjo, line 32: Minor: May be worth changing the `'returns a flush containing the data from the recorders'` to also validate it includes the info (and I guess the internal_metadata too, which I realized I missed :P)

ivoanjo (approved):
:+1: LGTM

lib/datadog/profiling/exporter.rb
```diff
@@ -43,6 +45,9 @@ def initialize(
         @last_flush_finish_at = nil
         @created_at = time_provider.now.utc
         @internal_metadata = internal_metadata
+        # NOTE: At the time of this comment collected info does not change over time so we'll hardcode
+        #       it on startup to prevent serializing the same info on every flush.
+        @info_json = JSON.fast_generate(info_collector.info)
```
ivoanjo, line 50: Minor: Not that we plan to mutate it, but since it lives for a while it may be worth as a "just-in-case"

```suggestion
        @info_json = JSON.fast_generate(info_collector.info).freeze
```

(Or alternatively, maybe the `Flush` should instead do it to every one of its arguments?)

ivoanjo (approved):
:+1: LGTM

PR #3362, opened 2024-01-03
[NO-TICKET] Fix "no signals" workaround detection when mariadb is in use

### LINKED ISSUE #3334: "no signals" workaround used with MariaDB-devel-11.2.2-1.el9
**Current behaviour**
Similar to #2954, the "no signals" workaround is being enabled when MariaDB-devel (latest) is in use.
`
WARN -- ddtrace: [ddtrace] Enabling the profiling "no signals" workaround because an incompatible version of the mysql2 gem is installed. Profiling data will have lower quality. To fix this, upgrade the libmysqlclient in your OS image to version 8.0.0 or above.
`
**Expected behaviour**
The "no signals" workaround should not be enabled.

**Steps to reproduce**
```
irb(main):001> Mysql2::VERSION
=> "0.5.5"
irb(main):002>
irb(main):003> Mysql2::Client.respond_to?(:info)
=> true
irb(main):004>
irb(main):005> Mysql2::Client.info[:version]
=> "3.3.8"
irb(main):006>
irb(main):007> Mysql2::Client.respond_to?(:info) # I'm curious to see if this changes after info has been called the first time
=> true
irb(main):008>
irb(main):009> (Mysql2::Client.methods - Object.methods).sort
=> [:default_query_options, :escape, :info]
irb(main):010>
irb(main):011> (Mysql2::Client.instance_methods - Object.methods).sort
=>
[:abandon_results!,
 :affected_rows,
 :async_result,
 :automatic_close=,
 :automatic_close?,
 :close,
 :closed?,
 :encoding,
 :escape,
 :find_default_ca_path,
 :info,
 :last_id,
 :more_results?,
 :next_result,
 :parse_connect_attrs,
 :parse_flags_array,
 :parse_ssl_mode,
 :ping,
 :prepare,
 :query,
 :query_info,
 :query_info_string,
 :query_options,
 :read_timeout,
 :reconnect=,
 :select_db,
 :server_info,
 :session_track,
 :set_server_option,
 :socket,
 :ssl_cipher,
 :store_result,
 :thread_id,
 :warning_count]
irb(main):012>
```

Note: client.info has a header_version field that contains the MariaDB version string.
`Mysql2::Client.info
=> {:id=>30308, :version=>"3.3.8", :header_version=>"11.2.2"}
`

**Environment**

* **ddtrace version:**  ddtrace-1.17.0
* **Ruby version:** 3..1.2-p20
* **Operating system:** RHEL 9.3 UBI

```
Name        : MariaDB-devel
Version     : 11.2.2
Release     : 1.el9
Architecture: x86_64
Install Date: Sun Dec 10 08:02:44 2023
Group       : Applications/Databases
Size        : 56828812
License     : GPLv2
Signature   : RSA/SHA512, Mon Nov 20 22:10:12 2023, Key ID f1656f24c74cd1d8
Source RPM  : MariaDB-devel-11.2.2-1.el9.src.rpm
Build Date  : Sun Nov 19 09:13:20 2023
Build Host  : rhel9-amd64
Vendor      : MariaDB Foundation
URL         : http://mariadb.org
Summary     : MariaDB database development files`
```

ivoanjo:
**What does this PR do?**

This PR fixes #3334 by tweaking the libmysqlclient detection code to not trigger when the mariadb version of libmysqlversion is in use.

The detection was triggering because the versioning schema used by mariadb is completely different from the libmysqlclient (e.g. latest releases report version 3.x, whereas mysql is 5.x or above).

**Motivation:**

For context, how we got here, is that legacy versions of the libmysqlclient library require the profiler "no signals" workaround to work without issues. To provide a seamless experience, we autodetect such versions and automatically apply the "no signals" workaround.

Unfortunately, this detection code incorrectly flagged the libmysqlclient version provided by mariadb as being incompatible.

In fact, that library has diverged a lot from the mysql version and we don't have any reports of issues that require the "no signals" workaround.

**Additional Notes:**

It's kinda fiddly to detect the mariadb version of libmysqlclient, see the added comments for more details.

**How to test the change?**

The change includes test coverage. To test it locally, I've additionally used the `dokken/centos-stream-9:latest` docker image, and then installed ruby + ruby-devel + gcc + make + MariaDB-devel from <https://mariadb.org/download/?t=repo-config&d=CentOS+Stream&v=11.2>.

**For Datadog employees:**

Fixes #3334

### DIFF
--- a/lib/datadog/profiling/component.rb
+++ b/lib/datadog/profiling/component.rb
@@ -331,9 +331,12 @@ module Datadog
 
           return true unless mysql2_client_class && mysql2_client_class.respond_to?(:info)
 
-          libmysqlclient_version = Gem::Version.new(mysql2_client_class.info[:version])
+          info = mysql2_client_class.info
+          libmysqlclient_version = Gem::Version.new(info[:version])
 
-          compatible = libmysqlclient_version >= Gem::Version.new('8.0.0')
+          compatible =
+            libmysqlclient_version >= Gem::Version.new('8.0.0') ||
+            looks_like_mariadb?(info, libmysqlclient_version)
 
           Datadog.logger.debug(
             "The `mysql2` gem is using #{compatible ? 'a compatible' : 'an incompatible'} version of " \
@@ -371,6 +374,35 @@ module Datadog
 
           2.0
         end
+      end
+
+      # To add just a bit more complexity to our detection code, in https://github.com/DataDog/dd-trace-rb/issues/3334
+      # a user reported that our code was incorrectly flagging the mariadb variant of libmysqlclient as being
+      # incompatible. In fact we have no reports of the mariadb variant needing the "no signals" workaround,
+      # so we flag it as compatible when it's in use.
+      #
+      # A problem is that there doesn't seem to be an obvious way to query the mysql2 gem on which kind of
+      # libmysqlclient it's using, so we detect it by looking at the version.
+      #
+      # The info method for mysql2 with mariadb looks something like this:
+      # `{:id=>30308, :version=>"3.3.8", :header_version=>"11.2.2"}`
+      #
+      # * The version seems to come from https://github.com/mariadb-corporation/mariadb-connector-c and the latest
+      # one is 3.x.
+      # * The header_version is what people usually see as the "mariadb version"
+      #
+      # As a comparison, for libmysql the info looks like:
+      # * `{:id=>80035, :version=>"8.0.35", :header_version=>"8.0.35"}`
+      #
+      # Thus our detection is version 4 or older, because libmysqlclient 4 is almost 20 years old so it's most probably
+      # not that one + header_version being 10 or newer, since according to https://endoflife.date/mariadb that's a
+      # sane range for modern mariadb releases.
+      private_class_method def self.looks_like_mariadb?(info, libmysqlclient_version)
+        header_version = Gem::Version.new(info[:header_version]) if info[:header_version]
+
+        !!(header_version &&
+          libmysqlclient_version < Gem::Version.new('5.0.0') &&
+          header_version >= Gem::Version.new('10.0.0'))
       end
     end
   end
--- a/sig/datadog/profiling/component.rbs
+++ b/sig/datadog/profiling/component.rbs
@@ -42,6 +42,7 @@ module Datadog
       def self.incompatible_passenger_version?: () -> bool
       def self.flush_interval: (untyped settings) -> ::Numeric
       def self.valid_overhead_target: (::Float overhead_target_percentage) -> ::Float
+      def self.looks_like_mariadb?: ({ header_version: ::String? }, ::Gem::Version) -> bool
     end
   end
 end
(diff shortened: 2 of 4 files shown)


lib/datadog/profiling/component.rb
```diff
@@ -372,6 +375,35 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
...
+      # * `{:id=>80035, :version=>"8.0.35", :header_version=>"8.0.35"}`
+      #
+      # Thus our detection is version 4 or older, because libmysqlclient 4 is almost 20 years old so it's most probably
+      # not that one + header_version being 10 or newer, since according to https://endoflife.date/mariadb that's a
+      # sane range for modern mariadb releases.
+      private_class_method def self.looks_like_mariadb?(info, libmysqlclient_version)
+        header_version = Gem::Version.new(info[:header_version]) if info[:header_version]
+
+        !!(header_version &&
+          libmysqlclient_version < Gem::Version.new('5.0.0') &&
```
AlexJF, line 404: Based on https://github.com/brianmario/mysql2/blob/79f78f940685396f1b5f30ec502544bb7e3ba9cf/ext/mysql2/client.c#L148 it seems like one may be able to use `mysql2` without linking/using the connector-c library in which case the reported version may be `>= 10.0.0` even for MariaDB.

Doesn't seem like they have any better way to distinguish between mariadb and mysql themselves in the gem based on the version gymnastics being done around that line.

So looks like eventually any logic we add here will get out of date. 

How crazy would it be to add a settings knob to complement this auto-detection and enabling users to force-set one flavour or the other?

ivoanjo: Ah, good call out!

Although looking at the code:

```c
  if (version >= 100000                         // MariaDB (all versions numbered 10.x)
    || (version >= 30000 && version < 40000)    // MariaDB Connector/C (all versions numbered 3.x)
```

...I think we'd be safe, although somewhat by accident.

The problem in #3334 is that the version check is testing for libmysqlclient older than 8.x and using that to choose to enable the "no signals" workaround; because mariadb connector/c reports 3.x, the logic would trigger.

But if for mariadb without connector/c the version reported is 10.x, we're fine there too, because it's >= 8.x.

> How crazy would it be to add a settings knob to complement this auto-detection and enabling users to force-set one flavour or the other?

There's already a [`settings.advanced.no_signals_workaround_enabled`](https://github.com/DataDog/dd-trace-rb/blob/7088ace6a50857b7e0931fa5eb67587ad0cae3d3/lib/datadog/core/configuration/settings.rb#L410) that can be set to `true` or `false` to override the autodetection.

Were you thinking of something like that or did you mean having something more specific to only override the mysql2 detection?

AlexJF: Ah right. The workaround applies globally so no point making anything specific to mysql right? Just need to be aware of it when dealing with eventual support cases 👍

AlexJF (approved):

lib/datadog/profiling/component.rb
ivoanjo: Yeah the global one has worked so far. It's also documented in https://docs.datadoghq.com/profiler/profiler_troubleshooting/ruby/#unexpected-failures-or-errors-from-ruby-gems-that-use-native-extensions-in-dd-trace-rb-1110 .

PR #3360, opened 2024-01-03
[prof-heap] Misc improvements/fixes

AlexJF:
**What does this PR do?**
* When an object id conflict is not allowed, raise with a more informative error message.
* Allow configuration of heap size collection (`DD_PROFILING_EXPERIMENTAL_HEAP_SIZE_ENABLED`) and heap sample rate (`DD_PROFILING_EXPERIMENTAL_HEAP_SAMPLE_RATE`).

**Motivation:**
* Debugging profiler shutdown on Ruby 2.7 workloads due to obj id conflicts.
* More knobs to enable different benchmarking flavours.

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**
- [ ] If this PR touches code that signs or publishes builds or packages, or handles
credentials of any kind, I've requested a review from `@DataDog/security-design-and-guidance`.
- [ ] This PR doesn't touch any of that.

Unsure? Have a question? Request a review!

PR #3366, opened 2024-01-05
[prof-heap] Fix issues stemming from rb_gc_force_recycle

AlexJF:
**2.0 Upgrade Guide notes**

**What does this PR do?**

**Motivation:**

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**


ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -493,10 +546,20 @@ static int update_object_record_entry(DDTRACE_UNUSED st_data_t *key, st_data_t *
...
-    VALUE new_inspect = object_record_inspect(update_data->new_object_record);
-    rb_raise(rb_eRuntimeError, "Object ids are supposed to be unique. We got 2 allocation recordings with "
-        "the same id.\nprevious=%"PRIsVALUE"\nnew=%"PRIsVALUE"\n", existing_inspect, new_inspect);
+    #ifdef HAVE_WORKING_RB_GC_FORCE_RECYCLE
+      // In this case, it's possible for an object id to be re-used. This is not great since we only
+      // look at a small amount of allocation samples which means we may have live object records
+      // being attributed to the wrong stack/labels.
+      // In this particular case, though, we saw the re-used allocation so prefer the new object_record
+      // and cleanup the old one.
+      on_committed_object_record_cleanup(update_data->heap_recorder, existing_record);
```
ivoanjo, line 555: I left a comment about this #3360; I'm somewhat repeating it here in case you missed it -- sorry if you actually did see it but explicitly decided not to do something about it ;)

I somewhat worry that this can hide bugs, e.g. this call to `on_committed_object_record_cleanup` assumes that object id reuse happened because of recycling without actually testing for it.

It's a bit awkward (needs a few changes), but one way of propagating this info would be to introduce a specific struct for `partial_object_record` and then record in `start_heap_allocation_recording` when the bit was set. That way we could get here and do `if (bit was set) { clean record } else { /* fallthrough to raise, it's a bug */ }`.

lib/datadog/profiling/component.rb
```diff
@@ -213,6 +213,17 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
...
+        if RUBY_VERSION < '3.1'
+          Datadog.logger.warn(
+            "Current Ruby version (#{RUBY_VERSION}) supports forced object recycling which has a bug that the " \
+            'heap profiler is forced to work around to remain accurate. This workaround should be safe and ' \
+            'should even partially fix a runtime memory leak around object id mapping. Details about the workaround ' \
+            "can be found in https://github.com/DataDog/dd-trace-rb/pull/3360. If you're not comfortable with this, " \
+            'disable heap profiler or upgrade to Ruby >= 3.1 where forced object recycling was completely' \
+            'removed (https://bugs.ruby-lang.org/issues/18290).'
+          )
+        end
```
ivoanjo, line 225: I... don't quite like this warning as-is :thinking: 

Yes, there are unknown-unknowns in what we're doing, but I think anyone reading this warning would need to dive really, really, really deep to make an informed decision about this.

Furthermore, those customers that do decide "I am comfortable with this", will still see the giant scary warning every time, which is a bit meh.

On the other hand, I do see the relevance in documenting what we're doing in a user-visible way.

I'm thinking we could move documenting this to [the public docs](https://docs.datadoghq.com/profiler/profiler_troubleshooting/ruby/). Alternatively, we could change this to a debug, not warn-level log -- this would not show up by default, but if folks are investigating issues in ddtrace, they would get the message when they enable debug logging.

(I also have a few suggestions for the text itself, but I think the discussion on where we place it and when we show it may end up making some of the suggestions outdated so I'll hold those :wink: )

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -545,6 +545,42 @@ def sample_allocation(obj)
...
+          # discard those tracked heap samples, ending with no matching recycled samples at all.
+          heap_samples_in_test_matcher = lambda { |sample|
+            (sample.values[:'heap-live-samples'] || 0) > 0 && sample.locations.any? do |location|
+              location.lineno == allocation_line && location.path == __FILE__
+            end
+          }
+
+          relevant_sample = heap_samples.find(&heap_samples_in_test_matcher)
+          expect(relevant_sample).to be nil
+        end
```
ivoanjo, line 583: I really like that we're adding coverage for this, even if it's awkward.

Having said that, I somewhat worry that in its current form this test is somewhat brittle, since it relies on allocating a lot and then hoping at least some of those 60k objects hit the issue.

I have some suggestions on how to change this, but I realize it may be a bit of work so... yeah, I guess up to you for how deep you want to go on this rabbit hole lol.

It seems to me that there's a few different cases that may be interesting to test:

1. That if we try to sample a recycled object, we will set its RUBY_FL_SEEN_OBJ_ID bit
2. That we'll correctly handle sampling a recycled object that was born on the spot of a tracked object (== that has the same id)
3. That we'll correctly clean up object tracking during serialization if we spot that a recycled object

For 1, one way that occurs to me of creating a recycled object is to allocate objects in a loop, get their object_id, and testing their `RUBY_FL_SEEN_OBJ_ID`. Eventually that should give us a recycled object.

For 2. and 3. I think we can modify the above to also track every object that gets created. Once we spot a recycled object, we could then either try to sample it to test 2. or try to flush the profile to test 3.

One trick for testing some of these weirder cases that worked well in the `CpuAndWallTimeWorker` and `ThreadContext` collectors was to introduce counters ("stats") to be able to observe/assert that we took the expected codepath for some weird situation.

ivoanjo (commented):
Really happy to see this working! Left a few notes :)

ext/ddtrace_profiling_native_extension/heap_recorder.c
```diff
@@ -415,19 +448,39 @@ static int st_object_record_update(st_data_t key, st_data_t value, st_data_t ext
...
+    // safely say re-use happened and thus treat it as an implicit free of the object
+    // we were tracking (the original one which got recycled).
+    if (RB_FL_ABLE(ref) && !RB_FL_TEST(ref, RUBY_FL_SEEN_OBJ_ID)) {
+      on_committed_object_record_cleanup(recorder, record);
+      return ST_DELETE;
+    }
+
+    // NOTE: The bug mentioned above may also lead to a slow leaking of records in the
+    // objspace obj_to_id and id_to_obj tables. We could try also force-setting the seen
+    // obj id bit here to reduce the impact of this leak.
```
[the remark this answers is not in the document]
AlexJF: I left this note here because there are 2 places where we could set the bit:
* When we sample a new allocation on a recycled slot. Here we have to set a bit otherwise our tracking breaks.
* When we detect that an object we were tracking got recycled into a new object for which we didn't get an allocation. This is the situation where this note is. We don't really need to set the bit here for our heap profiler to work. We could set it to further improve leak impact but it could be argued it would be an extra unnecessary meddling with runtime internals.

Since we're already setting in that other place, I'd vote to also set it in this codepath but wanted to run that by you first and see if you shared this opinion 😄

spec/datadog/profiling/stack_recorder_spec.rb
AlexJF: Went with a slightly different approach but I got the intended level of granularity and seems to work locally

ext/ddtrace_profiling_native_extension/heap_recorder.c
AlexJF: Did it by adding an explicit struct for a `recording`. I think in the end it makes some other intents clearer so I think it's a win overall.

ext/ddtrace_profiling_native_extension/heap_recorder.c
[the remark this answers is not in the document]
AlexJF: Did so and added a test for it.

ext/ddtrace_profiling_native_extension/extconf.rb
```diff
@@ -192,6 +192,14 @@ def add_compiler_flag(flag)
   $defs << '-DUSE_LEGACY_RB_VM_FRAME_METHOD_ENTRY'
 end
 
+# On older Rubies, rb_gc_force_recycle allowed to free objects in a way that
+# would be invisible to free tracepoints, finalizers and without cleaning
+# obj_to_id_tbl mappings.
+$defs << '-DHAVE_WORKING_RB_GC_FORCE_RECYCLE' if RUBY_VERSION < '3.1'
+
+# On older Rubies, there was no RUBY_SEEN_OBJ_ID flag
+$defs << '-DNO_SEEN_OBJ_ID_FLAG' if RUBY_VERSION < '2.7'
```
ivoanjo, line 201: Minor: It's not a pattern we've used before on this file, but since these two defines will usually be used together

```c
#if defined(HAVE_WORKING_RB_GC_FORCE_RECYCLE) && ! defined(NO_SEEN_OBJ_ID_FLAG)
```

...I wonder if we should go for a `'-DNEED_GC_FORCE_RECYCLE_BUG_WORKAROUND' if RUBY_VERSION < '3.1' && RUBY_VERSION > '2.6'` or something like that?

lib/datadog/profiling/component.rb
ivoanjo: Ack, sounds like a plan :+1:

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -545,6 +549,108 @@ def sample_allocation(obj)
...
+            sample_line = __LINE__ - 1
+
+            recycled_heap_sample_matcher = lambda { |sample|
+              (sample.values[:'heap-live-samples'] || 0) > 0 && sample.locations.any? do |location|
+                location.lineno == sample_line && location.path == __FILE__
+              end
+            }
+
+            recycled_sample = heap_samples.find(&recycled_heap_sample_matcher)
+            expect(recycled_sample).not_to be nil
```
ivoanjo, line 620: Really minor: The objects we're using here come from `profiling/spec_helper.rb` (from `samples_from_pprof`). Since these kinds of checks are coming up often, it may be useful to add them has helpers for those classes, so we can simplify the assertions:

```ruby
recycled_sample = heap_samples.find { |s| s.values?(:heap-live-samples) && s.location?(line: sample_line, path: __FILE__) }
```

ivoanjo (approved):
:+1: Really happy that this seems to have gone from a really annoying issue to a small roadbump :)

spec/datadog/profiling/stack_recorder_spec.rb
AlexJF: Added this!

ext/ddtrace_profiling_native_extension/extconf.rb
AlexJF: I feel this makes things more awkward in multiple senses:
* Feel like we reduce the maintenance burden of `extconf.rb` by trying to keep things more directly linked to runtime properties rather than our own "concepts" which people would need to dig deeper to determine conditions for (e.g. it's "straightforward" to determine whether a ruby version has a working `rb_gc_force_recycle` or has a `FL_SEEN_OBJ_ID` constant but less so to know whether it needs a `gc_force_recycle_bug_workaround` or what exactly that is).
* Some of the tests for this workaround rely on native functions registered in the `StackRecoder::Testing` module to proxy to runtime functionality (`rb_gc_force_recycle` and checking of `FL_SEEN_OBJ_ID`). By keeping the separate macros we can keep these functions registered with all Ruby versions and ensuring their results are consistent with Rubies that may be missing that functionality (by either doing no-ops or returning false in the case of flag checks). With a single workaround macro it feels like we'd need to either conditionally register these macros only in those cases where we know we'll be using them (which IMO adds a synchronization burden between a spec and its native code) or we'd have to tie the fallback functionality of those functions with less clear conditionals (e.g. why should `has_seen_id_flag` return false if `!NEED_GC_FORCE_RECYCLE_BUG_WORKAROUND`?).

So I'd vote for keeping things separate 😅 That being said, I did add a `CAN_APPLY_GC_FORCE_RECYCLE_BUG_WORKAROUND` macro to encode the conditions instead of having them in multiple places.

ivoanjo: +1 That's very reasonable :)

ivoanjo (approved):

PR #3381, opened 2024-01-11
Move propagation public API to contrib

marcotc:
`Datadog::Tracing::Propagation::HTTP` has moved to `Datadog::Tracing::Contrib::HTTP`.

**2.0 Upgrade Guide notes**

In 1.x:

```ruby
Datadog::Tracing::Propagation::HTTP.inject!
Datadog::Tracing::Propagation::HTTP.extract
```

In 2.0:
```ruby
Datadog::Tracing::Contrib::HTTP.inject
Datadog::Tracing::Contrib::HTTP.extract
```

**Motivation:**

The original intention was simply to remove the file `lib/datadog/tracing/propagation/http.rb`: 
https://github.com/DataDog/dd-trace-rb/blob/916472f9db3d73fa92ad068df7495f25c7a45f86/lib/datadog/tracing/propagation/http.rb#L5-L14
Which has been moved closer to its HTTP integration inside `tracing/contrib`.

It turns out there more issues with the 1.x approach than simply moving a constant:
1. The object `HTTP = Tracing::Contrib::HTTP::Distributed::Propagation.new` is a constant singleton, but it actually depends on [4 dynamic settings](https://github.com/DataDog/dd-trace-rb/blob/916472f9db3d73fa92ad068df7495f25c7a45f86/docs/GettingStarted.md?plain=1#L2068-L2071) and has to update its behavior whenever `Datadog.configure` is called.
    1. This was accomplished in 1.x by [re-evalating the global configuration](https://github.com/DataDog/dd-trace-rb/blob/916472f9db3d73fa92ad068df7495f25c7a45f86/lib/datadog/tracing/distributed/propagation.rb#L86-L95) on every call, which is slow.
1. Reconfiguring anything inside contrib was not supported in 1.x. Only top-level components have a life-cycle hook for reconfiguration, and they all live inside [a monolithic `component.rb` file](https://github.com/DataDog/dd-trace-rb/blob/916472f9db3d73fa92ad068df7495f25c7a45f86/lib/datadog/tracing/component.rb).
    1. Adding integration-specific hooks in `lib/datadog/tracing/component.rb` is incorrect today, as integration are like "plugins", independent from the core tracer.
1. Integrations did not officially defined a public API.
    1. This was simply done by creating a public method anywhere in the integration's modules and documenting it, but there was no official Public API documentation of these integration-specific APIs.

**What does this PR do?**

This PR creates a component life-cycle hook for contrib integrations:
```ruby
Contrib::Component.register('my-integration') do |config|
  # config.tracing.any_setting_here
end
```

This hook can be added to any integration and will be called when components are reconfigured.

This PR also publishes two official integration public APIs `Datadog::Tracing::Contrib::HTTP` and `Datadog::Tracing::Contrib::GRPC`. Each with only their respective `inject` and `extract` methods today.

**How to test the change?**
All changes are tested, including integration apps (which indeed caught a meaningful bug!).

**For Datadog employees:**


lib/datadog/tracing/contrib/http.rb
```diff
@@ -0,0 +1,44 @@
...
+module Datadog
+  module Tracing
+    module Contrib
+      # `Net/HTTP` integration public API
+      module HTTP
+        # Inject distributed headers into the given request
+        # @param digest [Datadog::Tracing::TraceDigest] the trace to inject
+        # @param data [Hash] the request to inject
+        def self.inject(digest, data)
+          raise 'Please invoke Datadog.configure at least once before calling this method' unless @propagation
```
marcotc, line 16: I'm experimenting with raising an actionable error, instead of possibly lazy loading all of `ddtrace` inside this method.

lib/datadog/tracing/contrib/sinatra/tracer_middleware.rb
```diff
@@ -18,15 +18,9 @@ def initialize(app, opt = {})
...
           end
 
-          # rubocop:disable Metrics/AbcSize
           # rubocop:disable Metrics/MethodLength
           def call(env)
-            # Set the trace context (e.g. distributed tracing)
-            if configuration[:distributed_tracing] && Tracing.active_trace.nil?
-              original_trace = Tracing::Propagation::HTTP.extract(env)
-              Tracing.continue_trace!(original_trace)
-            end
```
marcotc, line 28: I got these [`build_and_test_integration` errors](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/13253/workflows/8a67b3dd-a5be-44b6-a9e9-d2d758d58e35/jobs/496520) and was puzzled for a bit.

The error was simple, that I forgot to update the line `original_trace = Tracing::Propagation::HTTP.extract(env)` with the move of `HTTP`, but I examined this error closer and more importantly why none of our tests exercised this code path, except for `build_and_test_integration`.

I believe this block, which runs when Sinatra is the top-level span, is not needed anymore, given we've introduced a `sinatra/framework.rb` file that always instruments Rack when Sinatra integration is activated: https://github.com/DataDog/dd-trace-rb/blob/916472f9db3d73fa92ad068df7495f25c7a45f86/lib/datadog/tracing/contrib/sinatra/framework.rb#L14-L28

jhgilbert (approved):

lib/datadog/tracing/distributed/propagation.rb
```diff
@@ -83,40 +82,35 @@ def extract(data)
...
-              end
-            rescue => e
-              ::Datadog.logger.error(
-                "Error extracting distributed trace data. Cause: #{e} Location: #{Array(e.backtrace).first}"
+          @propagation_extract_style.each do |propagator|
+            # First extraction?
+            unless extracted_trace_digest
+              extracted_trace_digest = propagator.extract(data)
+              next
+            end
```
ekump, line 90: Could you do
```ruby
extracted_trace_digest ||= propagator.extract(data)
```
here? It is a bit terser and clearer to me.

ekump (approved):
couple of non-blocking suggestions, but LGTM

lib/datadog/tracing/distributed/propagation.rb
marcotc: I tried to use a suggestion, but I don't understand how to use the `||=` operator here. The early `next` is still conditional on the original value of `extracted_trace_digest`.

lib/datadog/core/configuration/components.rb
```diff
@@ -99,6 +99,8 @@ def initialize(settings)
           @health_metrics = self.class.build_health_metrics(settings)
           @telemetry = self.class.build_telemetry(settings, agent_settings, logger)
           @appsec = Datadog::AppSec::Component.build_appsec_component(settings)
+
+          self.class.configure_tracing(settings)
```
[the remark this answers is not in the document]
marcotc: I left this separate because `tracing` is not `tracer`.
The `build_tracer` method explicitly does the work to return a tracer instance.

The new method `configure_tracing` does work to configure any generic needs for the tracing product.
Ideally, the `tracing` product can configure itself without explicitly having to return a tracer instance, that is then managed by the core. This is a larger task that I'm not sure how to do today.

I could move this logic to build tracer, but that is somewhat incorrect today. The way core components work today, they manage the life cycle initialization of everything. And because `configure_tracing` is something that needs to be configured, but does not fall into the category of any other existing thing, it is being managed by the core components.

At the end of the day, I did not add this to `build_tracer` because it would be misleading to have it there. While we want with components is that they can be enabled in independently from the core tracing, adding it directly to the tracer building interface would not be correct.

Again, I'm following today's standard, and how the code is structured. With a different compartmentalization of products, this would likely live somewhere else.

Issue #3359, opened 2024-01-02
ActiveRecord SQL spans don't have the same service name as Instantiation spans


dannyrife:
**Current behaviour**
ActiveRecord SQL calls reflect the configured service name, but instantiations do not

The difference is clear comparing the contrib code: sql gets the settings, instantiation does not:
https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/active_record/events/sql.rb#L34
vs
https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/active_record/events/instantiation.rb#L32

**Expected behaviour**
Since this is part of the active record instrumentations, the service name should match the other instrumentation for active record spans.

**Steps to reproduce**
Instrument active record and create an instantiation something like below should make something that creates instantiations
```
class ApplicationRecord < ActiveRecord::Base
  self.abstract_class = true
end
```

**Environment**

* **ddtrace version: ddtrace (~> 1.14.0)**
* **Configuration block (`Datadog.configure ...`):**
* **Ruby version: 3.2.2 and 3.1.3**
* **Operating system: Alpine and Rocky Linux**
* **Relevant library versions:  ddtrace (~> 1.14.0) **

benoittgt:
Hello @dannyrife 

I took a look at your issue but there is two comments in specs that question myself. I am wondering if it was intended. 

https://github.com/DataDog/dd-trace-rb/blob/a6f2af005202ba806e52448c86bf2e28c37f04e9/spec/datadog/tracing/contrib/rails/database_spec.rb#L85

and

https://github.com/DataDog/dd-trace-rb/blob/a6f2af005202ba806e52448c86bf2e28c37f04e9/spec/datadog/tracing/contrib/rails/database_spec.rb#L113

But the comment could also be outdated when I see the `span.service` value at the time.

https://github.com/DataDog/dd-trace-rb/commit/f64a90c88369dad6bcf6109524f52a10361033fc#diff-480dabeac4cb83aeebfd219880fbbaafff3c0b3a83cef7b2fa0e3819e3b516a5R60-R61

Here is draft diff. I can open the pull request if it's the correct way of doing this. https://github.com/DataDog/dd-trace-rb/compare/master...benoittgt:dd-trace-rb:service-name-on-instantiation?expand=1

TonyCTHsu:
[...]

PR #3383, opened 2024-01-15
Pass the service name to instantiation spans

benoittgt:
**What does this PR do?**

This PR fix https://github.com/DataDog/dd-trace-rb/issues/3359 

This is a draft because I think we need to have a discussion first on the issue.

**Motivation:**

Having the same service name for "instantiation spans" and "sql spans".

**Additional Notes:**

I don't know how this change can impact end user. 

**How to test the change?**

It has been tested in spec suite but should be tested E2E. 😄

**For Datadog employees:**

### DIFF
--- a/spec/datadog/tracing/contrib/rails/database_spec.rb
+++ b/spec/datadog/tracing/contrib/rails/database_spec.rb
@@ -82,8 +82,7 @@ RSpec.describe 'Rails database' do
         span, = spans
         expect(span.name).to eq('active_record.instantiation')
         expect(span.span_type).to eq('custom')
-        # Because no parent, and doesn't belong to database service
-        expect(span.service).to eq(tracer.default_service)
+        expect(span.service).to eq(database_service)
         expect(span.resource).to eq('Article')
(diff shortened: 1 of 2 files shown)


spec/datadog/tracing/contrib/rails/database_spec.rb
```diff
@@ -110,7 +109,7 @@
 
           expect(instantiation_span.name).to eq('active_record.instantiation')
           expect(instantiation_span.span_type).to eq('custom')
-          expect(instantiation_span.service).to eq(tracer.default_service) # Because within parent
```
benoittgt, line 113: I think we could test also the fallback when 
```
service_name: database_service
```
is not defined.

TonyCTHsu:
I left a comment on the original issue. 
https://github.com/DataDog/dd-trace-rb/issues/3359#issuecomment-1898561195

benoittgt:
Closing because of https://github.com/DataDog/dd-trace-rb/issues/3359#issuecomment-1898561195

PR #3384, opened 2024-01-16
[PROF-8917] Add support for the libdatadog crash tracker

ivoanjo:
**What does this PR do?**

This PR adds support for the libdatadog crash tracker feature (off by default).

The crash tracker works by detecting when the Ruby VM reaches a segmentation fault, reporting the crash information as a last profile before the VM dies.

All of the interesting work is in <https://github.com/DataDog/libdatadog/pull/282>, this PR basically just wires things up.

**Motivation:**

This will be a useful tool when debugging VM crashes.

**Additional Notes:**

I'm opening this PR as a draft as the libdatadog support has not yet landed/been released. Also, there's a few open questions on:

* fork handling
* when to shut down

**How to test the change?**

(You'll need to build <<https://github.com/DataDog/libdatadog/pull/282> until the crash tracker gets included in a libdatadog release)

To test the crash tracker with an actual crash, try running the following on Ruby 2.6:

```bash
$ DD_PROFILING_ENABLED=true DD_PROFILING_EXPERIMENTAL_CRASH_TRACKING_ENABLED=true DD_TRACE_DEBUG=true bundle exec ddtracerb exec ruby -e 'Process.detach(fork { exit! }).instance_variable_get(:@foo)'
```

This should also work in the future but right now it doesn't work correctly; still looking into why:

```bash
$ DD_PROFILING_ENABLED=true DD_PROFILING_EXPERIMENTAL_CRASH_TRACKING_ENABLED=true DD_TRACE_DEBUG=true bundle exec ddtracerb exec ruby -e 'Process.kill("SEGV", Process.pid)'
```

**For Datadog employees:**


ivoanjo:
Update: I've updated the PR description to match the current state of the code.

I think we're in good shape -- the test suite is green when running with libdatadog master. We can merge as soon as a new libdatadog release is out. Marking PR ready for review.

lib/datadog/profiling/crashtracker.rb
```diff
@@ -0,0 +1,69 @@
...
+          return
+        end
+
+        begin
+          self.class._native_start_or_update_on_fork(
+            action: action,
+            exporter_configuration: exporter_configuration,
+            path_to_crashtracking_receiver_binary: path_to_crashtracking_receiver_binary,
+            tags_as_array: tags_as_array,
+          )
```
AlexJF, line 61: Wouldn't it be easier to have `_native_start_or_update_on_fork` be an actual method taking only a single parameter and using the private attr_readers for the rest?

Related minor: There's also currently an implicit mismatch between the native implementation which is essentially forced to work as a singleton given the crashtracker APIs and the ruby one which will happily support multiple instances if so desired but with repeated starts leading to either crashtracker update errors or silent overrides (haven't tested which happens). I think this may now become the first instance of something that would prevent us from having (if we so desired) multiple running profiler instances simultaneously? This may in practice not be that important, especially because crash tracking failures are currently just logging and not actually propagating exceptions but I feel there's a mismatch between lifecycles here.

Maybe worth "singletonning" the profiler to ensure you can never have 2 running instances of it and give a more understandable error rather than a "failed to register crashtracker with different settings" one that makes the underlying issue slightly more obscure?

lib/datadog/profiling/profiler.rb
```diff
@@ -49,6 +53,9 @@ def component_failed(failed_component)
           'See previous log messages for details.'
         )
 
+        # We explicitly not stop the crash tracker in this situation, under the assumption that, if a component failed,
+        # we're operating in a degraded state and crash tracking may still be helpful.
```
AlexJF, line 57: Related to previous minor comment about conflicting models/lifecycles between profiler and crash tracker.

After this, in theory, there could be a reconfiguration of `Datadog` through `Datadog.configure` and thus a restarting of profiler right? But if this were to happen, there's now a dangling crash tracking start that was never stopped and thus all future reconfigurations will fail.

In practice this does sound like enough of a edge case for us to be able to just go with unsupported usage so don't think it's a blocker in any way.

spec/datadog/profiling/crashtracker_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+        end
+      end
+    end
+
+    include_context 'HTTP server'
+
+    let(:request) { messages.first }
+
+    let(:hostname) { '127.0.0.1' }
+    let(:port) { '6006' }
```
AlexJF, line 154: Love this integration test but I've been hit by enough flakes due to "port in use" to really believe you can get away with this fixed port indefinitely. Should be easy enough to let WEBrick::HTTPServer pick one and update the crashtracker configuration with it before starting the test?

spec/datadog/profiling/profiler_spec.rb
```diff
@@ -51,6 +87,20 @@
...
+        expect(scheduler).to receive(:stop).ordered
+
+        expect(optional_crashtracker).to receive(:stop).ordered
+
+        shutdown!
+      end
+    end
   end
 
   describe 'Component failure handling' do
```
AlexJF, line 106: If we want to keep the crashtracker alive in these situations (which I think makes sense), may want to add a test for this case to make sure we don't inadvertently remove it later on.

AlexJF (approved):

lib/datadog/profiling/component.rb
```diff
@@ -110,6 +112,22 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
           )
       end
 
+      private_class_method def self.build_crashtracker(settings, transport)
+        return unless settings.profiling.advanced.experimental_crash_tracking_enabled
+
+        unless transport.respond_to?(:exporter_configuration)
```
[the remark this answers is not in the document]
ivoanjo: Actually, I'm not quite sure:

> * A refactoring of the default transport could lead to a situation where a warning starts being emitted about usage of custom transport when that is not the case.

Yes, but our tests would catch that (I actually went in and just tried it). Specifically

> 1) Datadog::Profiling::Component.build_profiler_component with :enabled true when crash tracking is enabled initializes the crash tracker
> 2) Datadog::Profiling::Component.build_profiler_component with :enabled true when crash tracking is enabled initializes the profiler instance with the crash tracker
> 3) Datadog::Profiling::HttpTransport#exporter_configuration returns the current exporter configuration

both fail because on top of the warning, the crash tracker doesn't get enabled. So I think this covers us in terms of "oops we changed the API and forgot about this detail".

> * A custom transport could be transparently used with no warnings even though it may not actually be fully compatible with crash tracking.

Note that this check happens only when the crash tracker is enabled, so if you have a custom transport that doesn't expose `exporter_configuration` you'll not get a warning by default.

And if you do see the warning, you can either expose an `exporter_configuration` on the custom transport OR disable the crash tracker to make the issue go away.

ext/datadog_profiling_native_extension/crashtracker.c
```diff
@@ -0,0 +1,80 @@
...
+  ENFORCE_TYPE(tags_as_array, T_ARRAY);
+  ENFORCE_TYPE(path_to_crashtracking_receiver_binary, T_STRING);
+  ENFORCE_TYPE(action, T_SYMBOL);
+
+  if (action != start_action && action != update_on_fork_action) rb_raise(rb_eArgError, "Unexpected action: %+"PRIsVALUE, action);
+
+  VALUE version = ddtrace_version();
+  ddog_prof_Endpoint endpoint = endpoint_from(exporter_configuration);
+
+  // This needs to come last, after all things that can raise exceptions, as otherwise it can leak
```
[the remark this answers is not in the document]
ivoanjo: Sounds good, fixed in 1b501cce2ae8d7fa144561ab4dfbfcd1ea1de48b

ext/datadog_profiling_native_extension/crashtracker.c
```diff
@@ -0,0 +1,80 @@
...
+    .tags = &tags,
+  };
+
+  ddog_prof_CrashtrackerResult result =
+    action == start_action ?
+      ddog_prof_Crashtracker_init(config, metadata) :
+      ddog_prof_Crashtracker_update_on_fork(config, metadata);
+
+  // Clean up before potentially raising any exceptions
+  ddog_Vec_Tag_drop(tags);
```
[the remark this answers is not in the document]
ivoanjo: Sounds good, fixed in 1b501cce2ae8d7fa144561ab4dfbfcd1ea1de48b

ext/datadog_profiling_native_extension/crashtracker.c
```diff
@@ -0,0 +1,80 @@
...
+  if (action != start_action && action != update_on_fork_action) rb_raise(rb_eArgError, "Unexpected action: %+"PRIsVALUE, action);
+
+  VALUE version = ddtrace_version();
+  ddog_prof_Endpoint endpoint = endpoint_from(exporter_configuration);
+
+  // This needs to come last, after all things that can raise exceptions, as otherwise it can leak
+  ddog_Vec_Tag tags = convert_tags(tags_as_array);
+
+  ddog_prof_CrashtrackerConfiguration config = {
+    .create_alt_stack = false, // This breaks the Ruby VM's stack overflow detection
```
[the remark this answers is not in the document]
ivoanjo: Added more context in ce15ade38d4b3bedd4868c41abecfca0d0bde02c

lib/datadog/profiling/crashtracker.rb
ivoanjo: > Wouldn't it be easier to have _native_start_or_update_on_fork be an actual method taking only a single parameter and using the private attr_readers for the rest? Or does this have some unexpected consequences when dealing with native extensions?

It's definitely _possible_. I'm somewhat on the wall on it being easier. Specifically, I think it's a bit harder to debug if we're going across from Ruby to native and then back.

So I've been kinda trying to keep this pattern of, when we need to go into native, the Ruby code gathers all things that we need, and hands them over as a bunch of simple types (strings, arrays, hashes) to the native function.

But in practice in many cases (and I think this one as well), the difference isn't thaaat huge so while I slightly favor this approach, I think it'd be fine if we changed to your suggested approach (which would be a lot more object-oriented).

> Related minor: There's also currently an implicit mismatch between the native implementation which is essentially forced to work as a singleton given the crashtracker APIs and the ruby one which will happily support multiple instances if so desired but with repeated starts leading to either crashtracker update errors or silent overrides (haven't tested which happens). I think this may now become the first instance of something that would prevent us from having (if we so desired) multiple running profiler instances simultaneously? This may in practice not be that important, especially because crash tracking failures are currently just logging and not actually propagating exceptions but I feel there's a mismatch between lifecycles here.
>
> Maybe worth "singletonning" the profiler to ensure you can never have 2 running instances of it and give a more understandable error rather than a "failed to register crashtracker with different settings" one that makes the underlying issue slightly more obscure?

So this is an interesting point.

We actually already have a check that makes the profiler a singleton. Specifically the `CpuAndWallTimeWorker` has some global state in native code (`active_sampler_instance` / `active_sampler_instance_state`) and if you do try to start more than one you'll get

> Could not start CpuAndWallTimeWorker: There's already another instance of CpuAndWallTimeWorker active in a different thread

In practice, you'd only hit this if you were directly invoking `Datadog::Profiling::Component.build_profiler_component` or even creating any of the other classes by themselves. None of this is supposed to be how the profiler gets used in-production: the profiler is supposed to be created via the usual ddtrace components, and those only keep one active instance of the profiler around at any one time.

So while yes, there's ways in which multiple `Crashtracker` instances could be active, in practice this should never happen in user applications, unless the profiler is being used piecemeal, which is not really supported.

The crashtracker itself doesn't completely mind `start` being called multiple times, it's a "last writer wins" for the configuration. And stop only works once, since it stops the only one active:

```
$ bundle exec pry
[1] pry(main)> require 'ddtrace'
[3] pry(main)> ct1 = Datadog::Profiling::Crashtracker.new(exporter_configuration: [:agent, 'http://localhost:6006'], tags: {})
=> #<Datadog::Profiling::Crashtracker:0x0000633690cc9be0
 @exporter_configuration=[:agent, "http://localhost:6006"],
 @path_to_crashtracking_receiver_binary="libdatadog-crashtracking-receiver",
 @tags_as_array=[]>
[4] pry(main)> ct2 = Datadog::Profiling::Crashtracker.new(exporter_configuration: [:agent, 'http://localhost:6006'], tags: {})
=> #<Datadog::Profiling::Crashtracker:0x0000633690d94890
 @exporter_configuration=[:agent, "http://localhost:6006"],
 @path_to_crashtracking_receiver_binary="libdatadog-crashtracking-receiver",
 @tags_as_array=[]>
[5] pry(main)> ct1.start
=> true
[6] pry(main)> ct1.start
=> true
[7] pry(main)> ct2.start
=> true
[8] pry(main)> ct2.start
=> true
[9] pry(main)> ct1.stop
=> true
[10] pry(main)> ct1.stop
E, [2024-04-04T09:46:42.795319 #116895] ERROR -- ddtrace: [ddtrace] (crashtracker.rb:41:in `rescue in stop') Failed to stop crash tracking: Failed to stop the crash tracker: ddog_prof_Crashtracker_shutdown failed: No crashtracking previous signal handlers
=> true
[12] pry(main)> ct2.stop
E, [2024-04-04T09:46:48.329567 #116895] ERROR -- ddtrace: [ddtrace] (crashtracker.rb:41:in `rescue in stop') Failed to stop crash tracking: Failed to stop the crash tracker: ddog_prof_Crashtracker_shutdown failed: No crashtracking previous signal handlers
=> true
```

I did, on an earlier version of this PR, use a singleton, but found that it was a bit more awkward to enable and test. I've pushed bd0537e7c0d96caa3cbac2d0137c375e7c9c039c to more clearly document what's up here; if you're still not convinced, let me know, and I can take a stab at turning it again into a singleton and see how that looks.

lib/datadog/profiling/profiler.rb
ivoanjo: > After this, in theory, there could be a reconfiguration of Datadog through Datadog.configure and thus a restarting of profiler right? But if this were to happen, there's now a dangling crash tracking start that was never stopped and thus all future reconfigurations of crash tracker may fail?

Actually, not! This comment is about stopping the profiler via `component_failed`, whereas the components stop the profiler via `shutdown!`, which has been modified to stop the crashtracker.

Thus, the semantics are: if the profiler stops itself because it ran into an issue, the crash tracker states. But once ddtrace decides to clean up the profiler and/or restart it, the crash tracker is correctly updated.

spec/datadog/profiling/crashtracker_spec.rb
ivoanjo: Changed in 0b40fbfdfb32e38d05fb8f8eced9ca2bbd720e0b

AlexJF (approved):
LGTM and thanks for all the clarifications!

lib/datadog/profiling/crashtracker.rb
```diff
@@ -0,0 +1,69 @@
...
+          Datadog.logger.debug('Crash tracking stopped successfully')
+        rescue => e
+          Datadog.logger.error("Failed to stop crash tracking: #{e.message}")
+        end
+      end
+
+      private
+
+      def start_or_update_on_fork(action:)
+        unless path_to_crashtracking_receiver_binary
```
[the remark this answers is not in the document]
ivoanjo: Hmmm maybe? I wasn't very deliberate about this at the time, but coming back to it with a fresh head, the thing is -- there's a bunch of things that can go wrong, e.g. maybe the path exists as a string, but it's not correct; or when we try to start the other process, it doesn't; etc.

All of those other errors, we can only see when we call `_native_start_or_update_on_fork`.

So any validation we do in `#initialize`, will always be kinda incomplete. Being that the case, why go to the trouble of requiring callers to wrap the call to new with a `begin/rescue` OR us providing a `.build` that does all that, just to do half of the work? ;)

ivoanjo:
Update: I've changed this PR to be on top of https://github.com/DataDog/dd-trace-rb/pull/3627 ( libdatadog v9 upgrade) + updated it for libdatadog v9 compatibility.

I think once libdatadog 9 support lands in master, this is finally 100% good to go :)

spec/datadog/core/configuration/settings_spec.rb
```diff
@@ -765,6 +765,41 @@
...
+          end
+        end
+      end
+
+      describe '#experimental_crash_tracking_enabled=' do
+        it 'updates the #experimental_crash_tracking_enabled setting' do
+          expect { settings.profiling.advanced.experimental_crash_tracking_enabled = true }
+            .to change { settings.profiling.advanced.experimental_crash_tracking_enabled }
+            .from(false)
+            .to(true)
```
danielsn, line 800: Ruby is kinda a fun language

spec/datadog/profiling/crashtracker_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+
+    context 'when called in a fork' do
+      before { crashtracker.start }
+      after { crashtracker.stop }
+
+      it 'starts a second crash tracker for the fork' do
+        expect_in_fork do
+          crashtracker.reset_after_fork
+
+          expect(`pgrep -f libdatadog-crashtracking-receiver`.lines.size).to be 2
```
danielsn, line 94: Is it possible to have other processes running that could be captured here?

spec/datadog/profiling/crashtracker_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+    it 'reports crashes via http' do
+      fork_expectations = proc do |status:, stdout:, stderr:|
+        expect(Signal.signame(status.termsig)).to eq('SEGV').or eq('ABRT')
+        expect(stderr).to include('[BUG] Segmentation fault')
+      end
+
+      expect_in_fork(fork_expectations: fork_expectations) do
+        crashtracker.start
+
+        Process.kill('SEGV', Process.pid)
```
danielsn, line 182: Would it also make sense to test an actual crash (e.g. null ptr deref)

spec/datadog/profiling/crashtracker_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+      end
+
+      crash_report = JSON.parse(request.body, symbolize_names: true)[:payload].first
+
+      expect(crash_report[:stack_trace]).to_not be_empty
```
danielsn, line 187: Could also check that the correct signal type was reported

spec/datadog/profiling/profiler_spec.rb
```diff
@@ -21,6 +24,19 @@
       start
     end
 
+    context 'when a crash tracker instance is provided' do
+      let(:optional_crashtracker) { instance_double(Datadog::Profiling::Crashtracker) }
+
+      it 'signals the crash tracker to start before other components' do
```
danielsn, line 30: nice

danielsn (commented):
LGTM. A few comments/questions

ext/datadog_profiling_native_extension/crashtracker.c
```diff
@@ -0,0 +1,108 @@
...
+  };
+
+  ddog_prof_EnvVar ld_library_path_env = {
+    .key = DDOG_CHARSLICE_C("LD_LIBRARY_PATH"),
+    .val = char_slice_from_ruby_string(ld_library_path),
+  };
+
+  ddog_prof_CrashtrackerReceiverConfig receiver_config = {
+    .args = {},
+    .env = {.ptr = &ld_library_path_env, .len = 1},
```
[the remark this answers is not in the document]
ivoanjo: Good question -- I don't particularly have a preference.

On one side, inheriting the full env would be useful if the crash tracker wanted to log some extra info from there. On the other hand, we can always add anything else we want extra later.

Dealer's choice? Do you think it'd be useful for me to pass along the env that Ruby has + with the addition of the ld_library_path?

ext/datadog_profiling_native_extension/libdatadog_helpers.c
```diff
@@ -60,3 +62,87 @@ size_t read_ddogerr_string_and_drop(ddog_Error *error, char *string, size_t capa
   ddog_Error_drop(error);
   return error_msg_size;
 }
+
+__attribute__((warn_unused_result))
+ddog_prof_Endpoint endpoint_from(VALUE exporter_configuration) {
```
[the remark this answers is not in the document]
ivoanjo: I can imagine some uses for it in some weird debugging cases... :thinking:

From my side, I'm fine with not having that ability, so I don't particularly think it's worth the effort.

lib/datadog/core/configuration/settings.rb
```diff
@@ -415,6 +415,16 @@ def initialize(*_)
...
               o.default 60
             end
+
+            # Enables reporting of information when the Ruby VM crashes.
+            #
+            # @default `DD_PROFILING_EXPERIMENTAL_CRASH_TRACKING_ENABLED` environment variable as a boolean,
+            # otherwise `false`
+            option :experimental_crash_tracking_enabled do |o|
+              o.type :bool
+              o.env 'DD_PROFILING_EXPERIMENTAL_CRASH_TRACKING_ENABLED'
```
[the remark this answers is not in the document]
ivoanjo: Yeah, that's a good idea. Did you have something in mind? I'm thinking I could open a PR to add to https://docs.datadoghq.com/profiler/profiler_troubleshooting/ruby/ .

lib/datadog/profiling/profiler.rb
ivoanjo: I think the current behavior from the native side is quite reasonable/useful: last-writer-wins-singleton means that even if there's a weird bug on our Ruby code the crash tracker can always be started and stopped by a new instance.

spec/datadog/profiling/crashtracker_spec.rb
ivoanjo: That I know of, there's no Ruby-level way to cause a segfault. I even thought to check the built-in ffi fiddle gem, which actually is neat enough to handle its own potential segfaults cleanly and turn them into Ruby exceptions:

> spec/datadog/profiling/crashtracker_spec.rb:188:in `[]': NULL pointer dereference (Fiddle::DLError)

So to actually do this, I would need to add a `cause_crash` method on the native side of the Profiler, which.... seems like a sharp edge that is not great to leave lying around >_> ?

spec/datadog/profiling/crashtracker_spec.rb
ivoanjo: Added in a2ae7305e2cadbb30f80bca5d8be673f6dda88c2

spec/datadog/profiling/crashtracker_spec.rb
ivoanjo: It is, although we never run our tests in parallel inside the same container in CI, so only if you're running the test suite locally would this potentially happen.

I've added an extra check at the beginning of these tests to check that no existing crash tracker was already running: 5431d2b07ce1b60039a4bf8680bd7e56e225656e .

lib/datadog/profiling/component.rb
```diff
@@ -110,6 +112,23 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
...
+          Datadog.logger.warn(
+            'Cannot enable profiling crash tracking as a custom settings.profiling.exporter.transport is configured'
```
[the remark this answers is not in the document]
ivoanjo: Yes, really good question -- I've added a big comment to the code to explain this in 09976ccf474643f30a9ad1ec42d3e7d319f0fe4a :)

ivoanjo:
CI is now green on this! Took a while but we made it :D

danielsn (approved):

spec/datadog/profiling/crashtracker_spec.rb
lloeki: You didn't try hard enough? :D

```
ruby -r fiddle -e 'Fiddle.free(42)'
-e:1: [BUG] Segmentation fault at 0x0000000000000022
```

ivoanjo: Ah, interesting, I didn't think of that :D . Thanks @lloeki for the suggestion, I'll make a note to shoot a small PR with your suggestion :)

PR #3386, opened 2024-01-17
Improve performance of gathering the ClassCount metric

ivoanjo:
**What does this PR do?**

This PR tweaks the `ClassCount.available?` method to cache its result.

The result is expected to be always the same for a given Ruby VM, so we don't need to keep checking again and again.

**Motivation:**

While looking at our GitLab test deployments, I noticed that on this big app, calling `ObjectSpace.count_objects` does have some cost (because this method goes through the whole Ruby heap).

I then noticed we _always_ call it twice in a row -- we call `available?` and then call `value`:

image

It's trivial to avoid this, so I decided to open this tiny PR.

**Additional Notes:**

N/A

**How to test the change?**

Existing test coverage is enough :)

**For Datadog employees:**

### DIFF
--- a/gemfiles/jruby_9.2_activesupport.gemfile.lock
+++ b/gemfiles/jruby_9.2_activesupport.gemfile.lock
@@ -11,7 +11,7 @@ GIT
 PATH
   remote: ..
   specs:
-    ddtrace (1.18.0)
+    ddtrace (1.19.0)
       datadog-ci (~> 0.6.0)
       debase-ruby_core_source (= 3.3.1)
       libdatadog (~> 5.0.0.1.0)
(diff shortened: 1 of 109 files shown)


AlexJF (approved):

ekump (approved):

ivoanjo:
Update: Rebased on top of current master to pull in #3387 and get a green CI again.

ivoanjo:
CI is broken... Unrelated to this PR... Again... Fix is in #3390, I'll rebase once it's in master...

ivoanjo:
So it appears that we were missing a few gemfile updates, that our github action pushed into this PR. I've looked into them and they look good. Going ahead and merging this PR.

PR #3392, opened 2024-01-22
Frozen string literals 

TonyCTHsu:
**2.0 Upgrade Guide notes**

🚨 In 2.0, all the strings are frozen. Make sure your code does not mutate them.


lib/datadog/tracing/contrib/qless/integration.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
```
ekump, line 2: Is this file and the qless_job.rb file below supposed to be added back as part of this PR?

marcotc (approved):

PR #3403, opened 2024-01-25
Remove `use` alias for `instrument`

TonyCTHsu:
**2.0 Upgrade Guide notes**

🚨 Breaking change: Remove `use` method for configuration, replace with `instrument`. 

For example

```ruby
Datadog.configure do |c|
  c.tracing.use :mysql2
end
```
to
```ruby
Datadog.configure do |c|
  c.tracing.instrument :mysql2
end
```

### DIFF
--- a/gemfiles/ruby_2.5_graphql_1.12.gemfile.lock
+++ b/gemfiles/ruby_2.5_graphql_1.12.gemfile.lock
@@ -11,7 +11,7 @@ GIT
 PATH
   remote: ..
   specs:
-    ddtrace (1.19.0)
+    ddtrace (1.20.0)
       base64
       debase-ruby_core_source (= 3.3.1)
       libdatadog (~> 5.0.0.1.0)
--- a/gemfiles/ruby_2.6_graphql_1.12.gemfile.lock
+++ b/gemfiles/ruby_2.6_graphql_1.12.gemfile.lock
@@ -11,7 +11,7 @@ GIT
 PATH
   remote: ..
   specs:
-    ddtrace (1.19.0)
+    ddtrace (1.20.0)
       base64
       debase-ruby_core_source (= 3.3.1)
       libdatadog (~> 5.0.0.1.0)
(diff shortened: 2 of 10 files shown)


marcotc (approved):

ivoanjo:
Can we consider leaving this one in? The maintenance cost seems to be very low -- it's one line, and I'm wondering if having this one line is the difference between a lot of customers smooth no-changes-needed upgrade from 1.x to 2.x?

So it seems like a nice win, even if it's slightly "ugly" to keep both.

TonyCTHsu:
> Can we consider leaving this one in?

@ivoanjo , I believe the alias was introduced for backward compatibility at 1.x for 0.x. and  I think it is fair to consider keeping this considered the maintenance cost is low. However, I prefer to remove it because I believe it is also fairly easy to change them in user's configuration.

ivoanjo:
I'll say that in my opinion no changes > easy changes ;)

p-datadog:
If people actually use `use`, it seems that it's removed with no deprecation period leading to errors when these users upgrade 1.0 -> 2.0. Perhaps adding a deprecation warning to master branch advising users to replace `use` with `instrument` would be good?

ekump:
I'm in favor of moving forward with this change. It may seem insignificant, but it does make our public interface and docs less confusing. 

> If people actually use use, it seems that it's removed with no deprecation period leading to errors when these users upgrade 1.0 -> 2.0

Users have to opt-in to a major release and will need to follow the upgrade guide. This change, along with all the other breaking changes shouldn't really surprise anyone. That being said, we should reasonably do what we can to ease the transition.

> Perhaps adding a deprecation warning to master branch advising users to replace use with instrument would be good?

If it is relatively easy to tell if a customer is using the alias, I think this is a good idea.

TonyCTHsu:
PR for the logging deprecation warning: https://github.com/DataDog/dd-trace-rb/pull/3438

PR #3405, opened 2024-01-25
Consolidate Datadog::Core::Configuration::Ext::Transport into Datadog::Core::Configuration::Agent

marcotc:
**2.0 Upgrade Guide notes**

1.x:
```ruby
Datadog::Core::Configuration::Ext::Transport::ENV_DEFAULT_HOST
```

2.0:
```ruby
Datadog::Core::Configuration::Agent::ENV_DEFAULT_HOST
```

**Motivation:**

The constant `ENV_DEFAULT_HOST` was the only constant that resided in a different namespace.

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**


TonyCTHsu (approved):

lib/datadog/core/configuration/ext.rb
```diff
@@ -17,17 +17,12 @@ module Metrics
           ENV_DEFAULT_PORT = 'DD_METRIC_AGENT_PORT'
         end
 
-        # DEV-2.0: This module only exists for backwards compatibility with the public API.
-        # It should be consolidated into the Agent module below.
-        module Transport
-          ENV_DEFAULT_HOST = 'DD_AGENT_HOST'
-        end
-
         module Agent
```
ivoanjo, line 20: I wonder if we should move most of these to the `AgentSettingsResolver`.

My thinking is all, no other component is supposed to know about them -- they're all supposed to use whatever they get told by the resolver to use. :thinking:

The statsd client is a bit of a sore spot as per the [discussion here](https://github.com/DataDog/ruby-guild/issues/96#:~:text=everything%20in%20ddtrace.-,Gaps%20we%20have%20today%3A,-a.%20Some%20settings) -- it was never updated to use the AgentSettingsResolver (which I suspect it should by default? -- at least for the hostname, as the port is different).

ivoanjo (approved):
LGTM :+1:

PR #3406, opened 2024-01-25
HTTP adapter renamed to :http

marcotc:
**2.0 Upgrade Guide notes**

In 1.x:
```ruby
c.tracing.transport_options = proc do |t|
  t.adapter :net_http
end
```

In 2.0:
```ruby
c.tracing.transport_options = proc do |t|
  t.adapter :http
end
```

**Motivation:**

The fact that the HTTP transport currently uses `Net::HTTP` is an implementation detail.

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**

### DIFF
--- a/lib/datadog/core/configuration/ext.rb
+++ b/lib/datadog/core/configuration/ext.rb
@@ -30,7 +30,7 @@ module Datadog
           ENV_DEFAULT_TIMEOUT_SECONDS = 'DD_TRACE_AGENT_TIMEOUT_SECONDS'
 
           module HTTP
-            ADAPTER = :net_http # DEV: Rename to simply `:http`, as Net::HTTP is an implementation detail.
+            ADAPTER = :http
             DEFAULT_HOST = '127.0.0.1'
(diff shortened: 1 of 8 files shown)


ivoanjo:
Hey, I think PR's upgrade notes description is off, since `c.tracing.transport_options` has been removed in #3350 :thinking:

ekump:
>Hey, I think PR's upgrade notes description is off, since c.tracing.transport_options has been removed in https://github.com/DataDog/dd-trace-rb/pull/3350 🤔

@ivoanjo is correct. Users can no longer explicitly set an adapter anymore.

ekump (approved):
Not to be pedantic, but I wonder if `http` is accurate enough since UDS also technically uses http? Perhaps sticking with http is necessary for consistency with other tracers and other datadog docs, so not blocking an LGTM from me.

marcotc:
This change is not really important, so I'll just remove the possibly inaccurate comment instead https://github.com/DataDog/dd-trace-rb/pull/3411

PR #3410, opened 2024-01-26
Enforce using bundler 2.3.26 for ruby 3.3 container

TonyCTHsu:
**What does this PR do?**

Currently, our appraisal lock files are bundled with `2.3.27` in Ruby 3.3 containers, however, our dev environment is using `2.3.26`. 

This change will make our automated Github action to use Bundler `2.3.26`


p-datadog:
This looks good to me.

marcotc (approved):

PR #3422, opened 2024-01-30
Put all docker-compose services on an internal network.

p-datadog:
**What does this PR do?**

Instead of exposing ports to the host via `port:`, expose them all only to the internal network and arrange for all services to be on this network.

**Motivation:**

This permits running docker-compose when the host machine also is running datadog-agent, mysql server, etc. The docker-compose services should now be completely isolated from the host services.

**Additional Notes:**

This PR should not change behavior for any operations performed inside docker. However, if any users run the docker-compose command and then access the component services from the host machine for some reason, this will no longer work.

Also, there may be cached docker state that could cause docker-compose to fail to start the environment if you already started it previously to applying this PR. I fixed this by either renaming the directory with the checked out dd-trace-rb tree (which creates a new namespace for the containers and also networks) or by `rm -rf`ing the entire `/var/lib/docker/*`.

**How to test the change?**

1. Install datadog agent or mysql server on the host machine and start them.
2. Run `docker-compose run --rm tracer-3.0 /bin/bash` 

**For Datadog employees:**

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,8 @@
 version: '3.2'
+
+networks:
+  dd-trace-rb:
+
 services:
   tracer-2.1:
     image: ghcr.io/datadog/dd-trace-rb/ruby:2.1.10-dd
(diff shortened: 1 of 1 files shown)


marcotc:
> This PR should not change behavior for any operations performed inside docker. However, if any users run the docker-compose command and then access the component services from the host machine for some reason, this will no longer work.

I depend on this for my workflow, I'll give this PR a try and see if I can still access docker-compose services (like elastic-search) from the host somehow.

ivoanjo:
Should we make a decision on this PR? E.g. either explicitly merge it or close it?

p-datadog:
I will close this for now since I don't actively need the proposed functionality at the moment.

PR #3424, opened 2024-01-31
[CIVIS-8604] APM source code integration

anmarchenko:
**2.0 Upgrade Guide notes**
In order to use the new source code integration feature set the following environment variables:
- DD_GIT_REPOSITORY_URL
- DD_GIT_COMMIT_SHA 

**What does this PR do?**
Adds source code integration tags to APM traces according to the following requirements:
- if DD_GIT_REPOSITORY_URL env var is defined, its value should be sanitized by removing possible URL credentials (user/password information if it can be successfully parsed into an URL) and set in the _dd.git.repository_url of the first span of every trace chunk
- if DD_GIT_COMMIT_SHA env var is defined, its value should be set in _dd.git.commit.sha of the first span of every trace chunk

**Additional Notes:**
Why are we not using trace_operation.set_tags?

Trace-level tags are being set in the **root span** of the trace, not the **first span** as defined by the requirements. Moreover, if flush is partial then these tags are not set at all:

```ruby
    def set_trace_tags!
      # If the root span wasn't specified, don't set this. We don't want to
      # misset or overwrite the tags of a span that is in the middle of the
      # trace.
      return if partial?

      root_span.set_tags(trace.send(:meta))
      root_span.set_tags(trace.send(:metrics))
    end
```

In order to provide 100% compatibility with Datadog APM backend without edge cases I decided to implement additional logic in `Datadog::Tracing::Transport::TraceFormatter` to set git tags for the first span of trace that is being serialized.

**How to test the change?**
Tested using system-tests locally
From junit report:

```xml
<testcase classname="tests.parametric.test_tracer.Test_TracerSCITagging" name="tests.parametric.test_tracer.Test_TracerSCITagging.test_tracer_repository_url_environment_variable[library_env0]" time="0.048">
<properties>
<property name="dd_tags[systest.case.outcome]" value="passed" />
<property name="dd_tags[systest.case.skip_reason]" value="" />
</properties>
</testcase>
```

**For Datadog employees:**

### DIFF
--- a/lib/datadog/core/environment/git.rb
+++ b/lib/datadog/core/environment/git.rb
@@ -8,15 +8,13 @@ module Datadog
     module Environment
       # Retrieves git repository information from environment variables
       module Git
-        module_function
-
-        def git_repository_url
+        def self.git_repository_url
           return @git_repository_url if defined?(@git_repository_url)
 
           @git_repository_url = Utils::Url.filter_basic_auth(ENV[Datadog::Core::Git::Ext::ENV_REPOSITORY_URL])
         end
 
-        def git_commit_sha
+        def self.git_commit_sha
           return @git_commit_sha if defined?(@git_commit_sha)
 
           @git_commit_sha = ENV[Datadog::Core::Git::Ext::ENV_COMMIT_SHA]
--- a/lib/datadog/core/environment/git.rb
+++ b/lib/datadog/core/environment/git.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require_relative '../git/ext'
+require_relative '../utils/url'
+
+module Datadog
+  module Core
+    module Environment
+      # Retrieves garbage collection statistics
+      module Git
+        module_function
+
+        def git_repository_url
+          return @git_repository_url if defined?(@git_repository_url)
+
+          @git_repository_url = Utils::Url.filter_sensitive_info(ENV[Datadog::Core::Git::Ext::ENV_REPOSITORY_URL])
+        end
+
+        def git_commit_sha
+          return @git_commit_sha if defined?(@git_commit_sha)
+
+          @git_commit_sha = ENV[Datadog::Core::Git::Ext::ENV_COMMIT_SHA]
+        end
+      end
+    end
+  end
+end
--- a/lib/datadog/core/environment/git.rb
+++ b/lib/datadog/core/environment/git.rb
@@ -6,14 +6,14 @@ require_relative '../utils/url'
 module Datadog
   module Core
     module Environment
-      # Retrieves garbage collection statistics
+      # Retrieves git repository information from environment variables
       module Git
         module_function
 
         def git_repository_url
           return @git_repository_url if defined?(@git_repository_url)
 
-          @git_repository_url = Utils::Url.filter_sensitive_info(ENV[Datadog::Core::Git::Ext::ENV_REPOSITORY_URL])
+          @git_repository_url = Utils::Url.filter_basic_auth(ENV[Datadog::Core::Git::Ext::ENV_REPOSITORY_URL])
         end
 
         def git_commit_sha
--- a/lib/datadog/core/utils/url.rb
+++ b/lib/datadog/core/utils/url.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Core
+    module Utils
+      # Helpers class that provides methods to proces URLs
+      # such as filtering sensitive information.
+      module Url
+        def self.filter_sensitive_info(url)
+          return nil if url.nil?
+
+          url.gsub(%r{((https?|ssh)://)[^/]*@}, '\1')
+        end
+      end
+    end
+  end
+end
--- a/lib/datadog/core/utils/url.rb
+++ b/lib/datadog/core/utils/url.rb
@@ -1,15 +1,23 @@
 # frozen_string_literal: true
 
+require 'uri'
+
 module Datadog
   module Core
     module Utils
-      # Helpers class that provides methods to proces URLs
+      # Helpers class that provides methods to process URLs
       # such as filtering sensitive information.
       module Url
-        def self.filter_sensitive_info(url)
+        def self.filter_basic_auth(url)
           return nil if url.nil?
 
-          url.gsub(%r{((https?|ssh)://)[^/]*@}, '\1')
+          URI(url).tap do |u|
+            u.user = nil
+            u.password = nil
+          end.to_s
+        # Git scheme: git@github.com:DataDog/dd-trace-rb.git
+        rescue URI::InvalidURIError
+          url
         end
       end
     end
--- a/lib/datadog/profiling/ext.rb
+++ b/lib/datadog/profiling/ext.rb
@@ -23,6 +23,8 @@ module Datadog
           FORM_FIELD_TAG_RUNTIME_VERSION = 'runtime_version'
           FORM_FIELD_TAG_SERVICE = 'service'
           FORM_FIELD_TAG_VERSION = 'version'
+          FORM_FIELD_TAG_GIT_REPOSITORY_URL = 'git.repository_url'
+          FORM_FIELD_TAG_GIT_COMMIT_SHA = 'git.commit.sha'
 
           PPROF_DEFAULT_FILENAME = 'rubyprofile.pprof'
           CODE_PROVENANCE_FILENAME = 'code-provenance.json'
--- a/lib/datadog/profiling/ext.rb
+++ b/lib/datadog/profiling/ext.rb
@@ -23,8 +23,8 @@ module Datadog
           FORM_FIELD_TAG_RUNTIME_VERSION = 'runtime_version'
           FORM_FIELD_TAG_SERVICE = 'service'
           FORM_FIELD_TAG_VERSION = 'version'
-          FORM_FIELD_TAG_GIT_REPOSITORY_URL = 'git.repository_url'
-          FORM_FIELD_TAG_GIT_COMMIT_SHA = 'git.commit.sha'
+          TAG_GIT_REPOSITORY_URL = 'git.repository_url'
+          TAG_GIT_COMMIT_SHA = 'git.commit.sha'
 
           PPROF_DEFAULT_FILENAME = 'rubyprofile.pprof'
           CODE_PROVENANCE_FILENAME = 'code-provenance.json'
--- a/lib/datadog/profiling/tag_builder.rb
+++ b/lib/datadog/profiling/tag_builder.rb
@@ -2,14 +2,12 @@
 
 require_relative '../core/utils'
 require_relative '../core/environment/git'
-require_relative '../core/git/ext'
 
 module Datadog
   module Profiling
     # Builds a hash of default plus user tags to be included in a profile
     module TagBuilder
       include Datadog::Profiling::Ext::Transport::HTTP # Tag name constants
-      include Datadog::Core::Git::Ext # Git tags constants
 
       def self.call(
         settings:,
@@ -47,8 +45,8 @@ module Datadog
         tags[FORM_FIELD_TAG_ENV] = env if env
         tags[FORM_FIELD_TAG_SERVICE] = service if service
         tags[FORM_FIELD_TAG_VERSION] = version if version
-        tags[TAG_REPOSITORY_URL] = git_repository_url if git_repository_url
-        tags[TAG_COMMIT_SHA] = git_commit_sha if git_commit_sha
+        tags[FORM_FIELD_TAG_GIT_REPOSITORY_URL] = git_repository_url if git_repository_url
+        tags[FORM_FIELD_TAG_GIT_COMMIT_SHA] = git_commit_sha if git_commit_sha
 
         # Make sure everything is an utf-8 string, to avoid encoding issues in native code/libddprof/further downstream
         user_tags.merge(tags).map do |key, value|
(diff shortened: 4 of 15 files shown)


lib/datadog/core/utils/url.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Core
+    module Utils
+      # Helpers class that provides methods to process URLs
+      # such as filtering sensitive information.
+      module Url
+        def self.filter_sensitive_info(url)
```
ekump, line 9: @marcotc or @TonyCTHsu do we not already have something to handle scrubbing sensitive info from urls?

TonyCTHsu: If we are only concerned with basic auth, I would prefer leveraging `URI` object instead regex.

```ruby
          URI.parse(url).tap do |u|
            u.user = nil
            u.password = nil
          end.to_s
```

This implement is only failing the test example `without protocol and without credentials`, but I think raising `URI::InvalidURIError` is correct since it is indeed not a URI.

TonyCTHsu: BTW, it looks like `Datadog::Tracing::Contrib::Utils::Quantization` does not contain functionality to filter out basic auth info.

anmarchenko: yeah, we can use URI approach here, as long as I make the specification tests pass

TonyCTHsu (approved):

lib/datadog/core/environment/git.rb
```diff
@@ -0,0 +1,27 @@
...
+  module Core
+    module Environment
+      # Retrieves git repository information from environment variables
+      module Git
+        module_function
+
+        def git_repository_url
+          return @git_repository_url if defined?(@git_repository_url)
+
+          @git_repository_url = Utils::Url.filter_basic_auth(ENV[Datadog::Core::Git::Ext::ENV_REPOSITORY_URL])
```
ivoanjo, line 16: I suspect this can be a bit expensive...

As we're exposing `Git.git_repository_url` as a stateless helper function, this means that once per call, we're going to be reparsing/filtering the same repository url that never changes. Furthermore, when it's in git scheme (as per `url.rb`), we'll be throwing and handling an exception.

Ideally we would do this validation once (perhaps as part of the settings system?) and then reuse the known-to-be-valid value after that.

anmarchenko: Calling this parsing logic every time git_repository_url is called wasn't my intention at all. From my tests I see that caching in the module variable works and this code is called once per process, so it returns cached `@git_repository_url` every time. You can see here I have to remove these cached variables with some magic in tests in order to test different scenarios: https://github.com/DataDog/dd-trace-rb/pull/3424/files#diff-b489b0fca6f1f662bc61fa750ef1f1fbbe60e8fa7894e99b13606d1d05140ba0R8

So my intention is to write code that will be executed **at most once** per process. I chose for this the same approach that you implemented for ClassCount:

```ruby
module Datadog
  module Core
    module Environment
      # Retrieves number of classes from runtime
      module ClassCount
        def self.value
          ::ObjectSpace.count_objects[:T_CLASS]
        end

        def self.available?
          return @class_count_available if defined?(@class_count_available)

          @class_count_available =
            ::ObjectSpace.respond_to?(:count_objects) && ::ObjectSpace.count_objects.key?(:T_CLASS)
        end
      end
    end
  end
end
```

Please tell me if this does not work as intended and the ClassCount optimization was done with a different goal in mind.

anmarchenko: @ivoanjo please tell me if this approach would also work for profiling to reuse this info and if not, I am happy to make changes so that it works for every product

TonyCTHsu: > Git.git_repository_url as a stateless helper function 

I think this is not correct. Although naming with `util` suggest it is a stateless helper function, but the value is cached by its implementation. 

I think the performance impact should not be blocking this implementation.

ivoanjo: Ah, derp, you're completely right. It does work as you state :)

The one case it wouldn't ~exactly work like that is module inclusion. I've been trying to avoid the use of `module_function`, since I don't think it makes any sense at all that you can `include Core::Environment::Git` anywhere to directly access `@git_repository_url`. And that one would mean the cache would belong to every instance of whatever class included that module.

I actually got rid of `module_function` in the file you shared exactly for that reason.

---

(More as a generic comment, I think needing to `remove_instance_variable` in the tests is a code smell and in general I still kinda favour having state in objects, not class/modules BUT it's not like this pattern isn't in use in a lot of places in ddtrace, so I'm ok with saying "let's fix it another day).

ivoanjo: > @ivoanjo please tell me if this approach would also work for profiling to reuse this info and if not, I am happy to make changes so that it works for every product

Yeah, it works. If you want to take a stab at the profiling part (and I'd thank you), the right place to modify is: 
https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/profiling/tag_builder.rb

Let me know / happy to help if you run into any troubles with it.

anmarchenko: Yeah, I will be happy to do it in TagBuilder too and I will also remove `module_function` too.

I hate using `remove_instance_variable` too, but I think here we have a tradeoff between needs to keep global state somewhere/test the code/make it performant.

lib/datadog/profiling/tag_builder.rb
```diff
@@ -1,12 +1,15 @@
 # frozen_string_literal: true
 
 require_relative '../core/utils'
+require_relative '../core/environment/git'
+require_relative '../core/git/ext'
 
 module Datadog
   module Profiling
     # Builds a hash of default plus user tags to be included in a profile
     module TagBuilder
```
anmarchenko, line 10: @ivoanjo could you review these changes? I tried to follow the conventions I saw in this file but I am not sure if I'm doing it correctly

spec/datadog/profiling/tag_builder_spec.rb
```diff
@@ -78,5 +78,33 @@
...
+
+      context 'when git environment is not available' do
+        before do
+          allow(Datadog::Core::Environment::Git).to receive(:git_repository_url).and_return(nil)
+          allow(Datadog::Core::Environment::Git).to receive(:git_commit_sha).and_return(nil)
+        end
+
+        it 'includes the git repository URL and commit SHA' do
+          expect(call).to_not include('_dd.git.repository_url', '_dd.git.commit.sha')
+        end
```
ivoanjo, line 106: I suspect this is incorrect...

The `_dd.` prefix I think is only for tracing? Looking at the [source code integration docs for Ruby](https://docs.datadoghq.com/integrations/guide/source-code-integration/?tab=ruby) the correct names should be `git.repository_url` and `git.commit.sha`.

I just tried it locally, by quickly doing this

```diff
-        tags[TAG_REPOSITORY_URL] = git_repository_url if git_repository_url
-        tags[TAG_COMMIT_SHA] = git_commit_sha if git_commit_sha
+        tags['git.repository_url'] = git_repository_url if git_repository_url
+        tags['git.commit.sha'] = git_commit_sha if git_commit_sha
```

and it worked for me.

Here's a [quick link to the profiles I got](https://ddstaging.datadoghq.com/profiling/search?query=service%3Aivoanjo-testing%2A%20&env=staging&my_code=disabled&service=ivoanjo-testing-new&version=2&viz=stream&start=1708005474337&end=1708009074337&paused=false):

* `ivoanjo-testing-old` works, sets the `DD_TAGS` manually
* `ivoanjo-testing-new` is this PR, doesn't work
* `ivoanjo-testing-new-no-prefix` is this PR + the change above

image

lib/datadog/profiling/tag_builder.rb
ivoanjo: See https://github.com/DataDog/dd-trace-rb/pull/3424/#discussion_r1491157422 :)

spec/datadog/profiling/tag_builder_spec.rb
ivoanjo: I... think the `_dd.` prefix for tracing is just a "flag" to tell the backend this is internal stuff to now show by default in the UX, but I'm not 100% sure ;)

anmarchenko: @ivoanjo fixed

lib/datadog/profiling/ext.rb
```diff
@@ -23,6 +23,8 @@ module HTTP
           FORM_FIELD_TAG_RUNTIME_VERSION = 'runtime_version'
           FORM_FIELD_TAG_SERVICE = 'service'
           FORM_FIELD_TAG_VERSION = 'version'
+          FORM_FIELD_TAG_GIT_REPOSITORY_URL = 'git.repository_url'
```
anmarchenko, line 26: @ivoanjo I am not sure if these constants should be called `FORM_FIELD_*` but others in these file are

anmarchenko: on the other hand I start to think that constants in Core must not have the `_dd` prefix if it is only for tracing

so in this case I would use Core::Git::Ext::GIT_REPOSITORY_URL for profiling and then append '_dd.' prefix when formatting traces in tracing component

wdyt?

spec/datadog/profiling/tag_builder_spec.rb
ivoanjo: Thanks, [can confirm it works](https://ddstaging.datadoghq.com/profiling/search?query=service%3Aivoanjo-testing%2A%20&env=staging&event=AgAAAY2xV_V1B-Lw9QAAAAAAAAAYAAAAAEFZMnhWX2FwQUFEVjJvRWVVN3hYblFBQQAAACQAAAAAMDE4ZGIxNTgtMDQyNy00NDdhLWE1Y2EtYmE5OTYwMmYwZDg3&my_code=disabled&p_tl_tb=1708077344531000064&p_tl_tf_f_1=3866099968&p_tl_tf_us_0=0&p_tl_tf_us_1=3866099968&profileId=AY2xV_apAADV2oEeU7xXnQAA&profileViz=flame_graph&service=ivoanjo-testing-new&version=2&viz=stream&start=1708073757483&end=1708077357483&paused=false)!

lib/datadog/profiling/ext.rb
anmarchenko: @ivoanjo I renamed the constants - now ready for a next pass :)

lib/datadog/tracing/transport/trace_formatter.rb
```diff
@@ -203,6 +226,14 @@ def find_root_span(trace)
...
           root_span || trace.spans.last
         end
+
+        def git_repository_url
+          Core::Environment::Git.git_repository_url
+        end
+
+        def git_commit_sha
+          Core::Environment::Git.git_commit_sha
+        end
```
ivoanjo, line 236: Minor: Maybe it's just me, but it seems weird to add ~8 lines of code to be able to write `git_repository_url/git_commit_sha` instead of directly calling `Core::Environment::Git.git_repository_url/git_commit_sha` above? :eyes:

lib/datadog/tracing/transport/trace_formatter.rb
```diff
@@ -54,6 +60,11 @@ def format!
           tag_sampling_priority!
           tag_profiling_enabled!
 
+          return trace unless first_span
+
+          tag_git_repository_url!
+          tag_git_commit_sha!
+
           trace
```
ivoanjo, line 68: Minor: It's very much stylistic thing, but at least for me this seems to read a bit better?

```suggestion
          if first_span
            tag_git_repository_url!
            tag_git_commit_sha!
          end

          trace
```

Also -- I'm not quite sure I understand, when would a trace not have any spans? (e.g. `trace.spans.first == nil`)

spec/datadog/tracing/transport/trace_formatter_spec.rb
```diff
@@ -96,7 +120,8 @@
         it do
           is_expected.to have_attributes(
             trace: trace,
-            root_span: root_span
+            root_span: root_span,
+            first_span: first_span
```
ivoanjo, line 124: Minor: This... kinda seems like coupling on an internal detail, e.g. if we ever decide to access the `first_span` directly instead of keeping it as an instance variable, it's kinda weird to have to change the tests?

ivoanjo (approved):
:+1: LGTM

Two notes:
* Since this touches a few things on the tracing side, probably worth getting a :+1: from one of the tracing folks
* Would it be worth to backport this to master? E.g. otherwise announcing this to customers/updating our docs will be delayed until whenever 2.0 is available >_>

lib/datadog/tracing/transport/trace_formatter.rb
anmarchenko: yes, it is not worth it to add these methods

About the points above: we reviewed this with @TonyCTHsu and he ack'd this feature from Tracing side
I will backport it to master as soon as I merge it to 2.0 later today

anmarchenko: no, actually I found the reason I did
this looks awful:

```ruby
        def tag_git_repository_url!
          return if Core::Environment::Git.git_repository_url.nil?

          first_span.set_tag(Core::Git::Ext::TAG_REPOSITORY_URL, Core::Environment::Git.git_repository_url)
        end
```

ivoanjo: Still looks better than having two extra methods IMHO but I'll leave the final decision up to you since it's a very small detail ;)

PR #3428, opened 2024-02-01
[PROF-8942] Enable Ruby timeline by default

ivoanjo:
**What does this PR do?**

This PR enables the timeline feature for the Continuous Profiler by default.

It introduces a new setting, `timeline_enabled`, which can also be controlled by setting the `DD_PROFILING_TIMELINE_ENABLED` environment variable.

It also deprecates the old `experimental_timeline_enabled` + it's environment variable counterpart.

**Motivation:**

With the latest round of improvements to libdatadog, we're confident about enabling this feature by default.

**Additional Notes:**

~~I'm opening this PR as a draft as it still depends on https://github.com/DataDog/libdatadog/pull/293 being released + a few more rounds of internal validation.~~ ✅ Fixed

~~This PR also will need to be updated with the changes from https://github.com/DataDog/dd-trace-rb/pull/3357 OR the reverse will need to happen -- whichever one is merged first.~~ ✅ Fixed

**How to test the change?**

Try to profile any application, validate that you get timeline data without further configuration.

**For Datadog employees:**

### DIFF
--- a/lib/datadog/core/configuration/settings.rb
+++ b/lib/datadog/core/configuration/settings.rb
@@ -424,9 +424,28 @@ module Datadog
             #
             # @default `DD_PROFILING_EXPERIMENTAL_TIMELINE_ENABLED` environment variable as a boolean, otherwise `false`
             option :experimental_timeline_enabled do |o|
+              o.after_set do
+                # FIXME: TODO: This needs to be updated to match up with https://github.com/DataDog/dd-trace-rb/pull/3357
+                Datadog.logger.warn(
+                  'The profiling.advanced.experimental_timeline_enabled setting has been deprecated for removal ' \
+                  'and no longer does anything. Please remove it from your Datadog.configure block. ' \
+                  'The timeline feature counting is now controlled by the `timeline_enabled` setting instead.'
+                )
+              end
+            end
+
+            # Controls data collection for the timeline feature.
+            #
+            # If you needed to disable this, please tell us why on <https://github.com/DataDog/dd-trace-rb/issues/new>,
+            # so we can fix it!
+            #
+            # TODO: In the future this setting may be deprecated and the timeline may no longer be optional.
+            #
+            # @default `DD_PROFILING_TIMELINE_ENABLED` environment variable as a boolean, otherwise `true`
+            option :timeline_enabled do |o|
               o.type :bool
-              o.env 'DD_PROFILING_EXPERIMENTAL_TIMELINE_ENABLED'
-              o.default false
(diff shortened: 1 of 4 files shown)


ivoanjo:
Marking this as good to go. For Datadog folks, search google drive for "Review of Ruby profiler timeline feature before going GA (2024-02)" for a review of benchmarks with the timeline feature.

Note to self: after this gets merged, we need to do a pass on our benchmarking configurations, due to the new default/new environment variable.

lib/datadog/core/configuration/settings.rb
```diff
@@ -424,9 +424,29 @@ def initialize(*_)
             #
             # @default `DD_PROFILING_EXPERIMENTAL_TIMELINE_ENABLED` environment variable as a boolean, otherwise `false`
             option :experimental_timeline_enabled do |o|
+              o.after_set do |_, _, precedence|
```
AlexJF, line 427: Slightly off-topic: I've noticed we don't do this in the other deprecated settings but wonder if we should:

Keep the `o.env` setting so that we warn if deprecated env variables are left lying around?

lib/datadog/core/configuration/settings.rb
```diff
@@ -424,9 +424,29 @@ def initialize(*_)
...
+                end
+              end
+            end
+
+            # Controls data collection for the timeline feature.
+            #
+            # If you needed to disable this, please tell us why on <https://github.com/DataDog/dd-trace-rb/issues/new>,
+            # so we can fix it!
+            #
+            # TODO: In the future this setting may be deprecated and the timeline may no longer be optional.
```
AlexJF, line 443: Minor: Is this TODO useful? It's a given that in the future anything may happen 😂

AlexJF (approved):

lib/datadog/core/configuration/settings.rb
ivoanjo: It's an interesting idea...

Although at this point, I was thinking that for 2.0 we'd remove all these settings :thinking:. Do you think we should keep them around, potentially with the `o.env` as well?

lib/datadog/core/configuration/settings.rb
ivoanjo: You're right, it's a bit too much futurology. Removed in c862874ec8b0d94241bb2d801a191d339e755fe4 .

lib/datadog/core/configuration/settings.rb
AlexJF: Nah good time to clean up house but I think it'd make sense to take this into account during the lifetime of 2.x

PR #3417, opened 2024-01-29
GraphQL: Instrument with `GraphQL::Tracing::DataDogTrace`

TonyCTHsu:
**2.0 Upgrade Guide notes**

GraphQL instrumentation replace the deprecated `GraphQL::Tracing::DataDogTracing`  with  `GraphQL::Tracing::DataDogTrace`. An option `with_deprecated_tracer` can be configured if user want to continue using `GraphQL::Tracing::DataDogTracing`.  Since `GraphQL::Tracing::DataDogTrace` is only available after `graphql-ruby(>= 2.0.19)`, user prior to this version will fallback to the deprecated `GraphQL::Tracing::DataDogTracing`

PR #3432, opened 2024-02-02
Use GraphQL::Tracing::DataDogTrace when it's available

rmosolgo:
**What does this PR do?**

Uses the new GraphQL-Ruby `trace_with` API _directly_, instead of via `use ...`., to avoid a deprecation warning. 

**Motivation:**

Someone reported the deprecation warning to me 

**Additional Notes:**

Although this changes how the module is applied, it doesn't actually change behavior on recent GraphQL-Ruby versions, because GraphQL-Ruby was applying this new module even when the old one was attached: https://github.com/rmosolgo/graphql-ruby/blob/1a0b0dea2f9134c63f20228001c785f9cbaea3d9/lib/graphql/tracing/platform_tracing.rb#L82-L85

So, I think users have already been using the new code. 

**How to test the change?**

Unit tests 

**For Datadog employees:**

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -902,18 +902,8 @@ If you prefer to individually configure the tracer settings for a schema (e.g. y
 ```ruby
 # Class-based schema
 class YourSchema < GraphQL::Schema
-  use(
-    GraphQL::Tracing::DataDogTracing,
-    service: 'graphql'
-  )
-end
-```
-
-```ruby
-# .define-style schema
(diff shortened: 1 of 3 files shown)


TonyCTHsu:
👋 @rmosolgo , I already have this in our 2.0 release. https://github.com/DataDog/dd-trace-rb/pull/3417

aliciascott (approved):
looks good for docs!

rmosolgo:
That looks great, @TonyCTHsu, thanks!

PR #3434, opened 2024-02-05
[PROF-9045] Better handle spikes within dynamic sampling adjustment windows

AlexJF:
**What does this PR do?**
Makes dynamic sampling adjustments in our dynamic discrete sampler more flexible by allowing these to be preempted based on number of samples, not only based on time.

**Motivation:**
Before this change preemption would occur every second. This means that very quick spikes/changes in allocation patterns that mostly fit within an adjustment window (i.e. take less than a second) would be sampled purely according to the previous understanding of the sampling load.

At the extreme, imagine a situation where the baseline is virtually zero which means we can sample events with probability 1 and be sure we are below the overhead. Then a big event spike occurs within an window. Because we'd be operating under the assumption that we can sample everything that comes, we'd end up sampling the entire spike, overloading our system until the next readjustment tries to compensate for it. We suspect this behaviour is the one responsible for the more noisy latencies around >p99 seen in https://github.com/DataDog/dd-trace-rb/pull/3395#issuecomment-1918877615

**Additional Notes:**
This PR tells a story:
1. We add [a test](https://github.com/DataDog/dd-trace-rb/pull/3434/commits/59b5fade8f53a240e2f70f47004efbf75c2f1fc5) simulating the scenario described above in motivation. Low baseline followed by <1s spike. This test [fails](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/13517/workflows/59ed058b-5f79-4ff5-9316-43fffb8fc782/jobs/505215/parallel-runs/4/steps/4-114?invite=true#step-114-73415_132) with current code as we'd be sampling the entire spike:
    ```
     1) Datadog::Profiling::Collectors::DiscreteDynamicSampler when under a variable load with a big spike that fits within an adjustment window will readjust preemptively with smaller windows to prevent sampling overload
       Failure/Error: expect(stats[:num_samples]).to be < 1000
     
         expected: < 1000
              got:   2500
    ```

2. We then introduce [adjustment preemption based on number of samples](https://github.com/DataDog/dd-trace-rb/pull/3434/commits/ed53cd1479d3d0e4ef81c8144213ca71c42d860a), [fixing](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/13520/workflows/6bdf2eeb-eb65-4636-9f5c-de6b593692ef/jobs/505322?invite=true#step-114-66634_82) the above test.
3. We then add [an extra expectation](https://github.com/DataDog/dd-trace-rb/blob/7e759e77d9259121993cebbbe1d7e116bf1b51ba/spec/datadog/profiling/collectors/discrete_dynamic_sampler_spec.rb#L184-L187) on our new test to verify the extra readjustments don't make us forget the past too quickly. This test [fails](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/13521/workflows/9d5ead5f-99a6-4724-b1ce-b08eda1d529e/jobs/505359/parallel-runs/4/steps/4-114?invite=true#step-114-73377_132) with our naive preemption: each readjustment carries the same "memory importance" so when we do them more frequently, our view into the past decreases.
    ```
      1) Datadog::Profiling::Collectors::DiscreteDynamicSampler when under a variable load with a big spike that fits within an adjustment window will readjust preemptively with smaller windows to prevent sampling overload
     Failure/Error: expect(sampler_current_events_per_sec).to be < 1250
     
       expected: < 1250
            got:   3730.1696912114626
    ```
4. We change the exponential moving average code to [take into account the width of a readjustment window](https://github.com/DataDog/dd-trace-rb/pull/3434/commits/dd820cfe27109779521388bf6490f83e8fb23cd7). This means we can now readjust more frequently but that we should still keep the same "length" of past in our memory, even if we now do 100 adjustments rather than the previous 10. This makes the previous test [pass](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/13522/workflows/cda410f0-6501-4d68-a5f4-89e3b20e16ae/jobs/505408?invite=true#step-114-43112_82).

Shipping the final PR to our reliability environment, we can definitely observe a significant impact in controlling the >p99 latencies and bring them in-line with those of non-allocation-sampling flavours (right-side is this PR, left is current master):

ddstaging datadoghq com_notebook_7340586_dynamic-sampling-allocation_range=14460000 start=1707129105661 live=true

**How to test the change?**

**For Datadog employees:**


ivoanjo (approved):
:+1: LGTM

This was one of those PRs where I had to stare at it for quite a while for it to make sense, but I think I stared enough at it and it does make sense :rofl:

ext/ddtrace_profiling_native_extension/collectors_discrete_dynamic_sampler.c
```diff
@@ -190,9 +196,15 @@ static void maybe_readjust(discrete_dynamic_sampler *sampler, long now) {
   //   then probability will be > 1 (but we should clamp to 1 since probabilities higher than 1 don't make sense).
   // * If app is eventing a lot or our sampling overhead is big, then as time_to_sample_all_events_ns grows, sampling_probability will
   //   tend to 0.
+
+  // Use our stored data to extrapolate to a full adjustment window view
+  long window_time_ns = ADJUSTMENT_WINDOW_NS;
+  long sampling_window_time_ns = window_time_ns * sampler->events_per_ns * sampler->sampling_probability * sampler->sampling_time_ns;
```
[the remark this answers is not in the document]
AlexJF: Nice spot. Went back to the formula and fully simplified it to point out it's actually not dependent on window_time_ns at all 😂

ext/ddtrace_profiling_native_extension/collectors_discrete_dynamic_sampler.c
```diff
@@ -190,16 +214,22 @@ static void maybe_readjust(discrete_dynamic_sampler *sampler, long now) {
...
+  if (max_sampling_overhead == 0) {
+    // if we aren't allowed any sampling overhead at all, probability has to be 0
     sampler->sampling_probability = 0;
   } else {
     // otherwise apply the formula described above (protecting against div by 0)
-    sampler->sampling_probability = time_to_sample_all_events_ns == 0 ? 1. :
-      double_min_of(1., max_allowed_time_for_sampling_ns / time_to_sample_all_events_ns);
+    sampler->sampling_probability = avg_sampling_overhead == 0 ? 1. :
+      double_min_of(1., max_sampling_overhead / avg_sampling_overhead);
   }
```
ivoanjo, line 233: Are we missing the "protecting against div by 0" bit here? I'll admit it's a pretty situation to have `events_per_ns` or `sampling_time_ns` or `(1 - max_sampling_overhead)` be zero, but I can't convince myself exactly that they can't ever go there...?

ivoanjo (approved):
Ok so I spent yet more time staring hard into this and it seems reasonable still :) :+1: LGTM

ext/ddtrace_profiling_native_extension/collectors_discrete_dynamic_sampler.c
ivoanjo: Ergh, on a closer look, it's there, I don't know how I missed it, please ignore temporary insanity.

PR #3439, opened 2024-02-06
[PROF-9035] Stop worker on clock failure

AlexJF:
**What does this PR do?**
When detecting an interaction that involves a clock time of 0 (returned on error), put the worker into a failure state and ultimately lead to it stopping.

**Motivation:**
We can't trust things if we have non working clocks and better to shut things down than to continue consuming resources.

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**


ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -1053,3 +1063,20 @@ static VALUE rescued_sample_allocation(VALUE tracepoint_data) {
...
 }
+
+static void stop_if_allocation_sampling_failure(struct cpu_and_wall_time_worker_state *state) {
+  const char* error = discrete_dynamic_sampler_error(&state->allocation_sampler);
+  if (error != NULL) {
+    VALUE error_rbstr = rb_str_new_cstr("allocation sampler - ");
+    rb_str_cat_cstr(error_rbstr, discrete_dynamic_sampler_error(&state->allocation_sampler));
+    stop_state(state, rb_exc_new_str(rb_eRuntimeError, error_rbstr));
+  }
+}
```
ivoanjo, line 1074: Having to remember to check the error is a bit awkward. One technique we use for libdatadog, and for instance in https://github.com/DataDog/dd-trace-rb/blob/master/ext/ddtrace_profiling_native_extension/clock_id.h is to return a struct from the APIs that immediately flags if there was an error or not.

Wouldn't say it's a blocker for this PR, but may be cool to experiment with maybe?

ivoanjo (approved):
:+1: LGTM

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -1061,3 +1092,29 @@ static VALUE rescued_sample_allocation(VALUE tracepoint_data) {
...
+  // discrete sampler in the future.
+  VALUE error_rbstr = rb_str_new_cstr("allocation sampler - ");
+  rb_str_cat_cstr(error_rbstr, error);
+
+  if (can_raise) {
+    rb_raise(rb_eRuntimeError, "%"PRIsVALUE, error_rbstr);
+  } else {
+    // If we can't raise an immediate exception at the calling site, use the asynchronous flow through the main worker loop.
+    stop_state(state, rb_exc_new_str(rb_eRuntimeError, error_rbstr));
+  }
```
ivoanjo, line 1111: Minor: Very tiny cleanup ;)

```suggestion
  VALUE exception = rb_exc_new_str(rb_eRuntimeError, rb_sprintf("allocation sampler - %s", error));

  if (can_raise) {
    rb_exc_raise(exception);
  } else {
    // If we can't raise an immediate exception at the calling site, use the asynchronous flow through the main worker loop.
    stop_state(state, exception);
  }
```

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -1061,3 +1092,29 @@ static VALUE rescued_sample_allocation(VALUE tracepoint_data) {
...
+}
+
+static VALUE _native_force_allocation_sampler_failure(DDTRACE_UNUSED VALUE self) {
+  struct cpu_and_wall_time_worker_state *state = active_sampler_instance_state; // Read from global variable, see "sampler global state safety" note above
+
+  discrete_dynamic_sampler_testing_force_fail(&state->allocation_sampler);
+
+  return Qnil;
+}
```
ivoanjo, line 1120: This assumes that `active_sampler_instance_state` is not `NULL` which... may not necessarily be the case. E.g. it's probably safer to receive the worker instance as an argument.

ext/ddtrace_profiling_native_extension/collectors_discrete_dynamic_sampler.c
```diff
@@ -13,12 +13,27 @@
...
+  long now = monotonic_wall_time_now_ns(DO_NOT_RAISE_ON_FAILURE);
+  return _discrete_dynamic_sampler_init(sampler, debug_name, now);
 }
 
-static void _discrete_dynamic_sampler_reset(discrete_dynamic_sampler *sampler, long now_ns) {
+static const char* _discrete_dynamic_sampler_reset(discrete_dynamic_sampler *sampler, long now_ns) {
+  if (now_ns <= 0) {
+    sampler->error = ERR_CLOCK_FAIL;
+    return sampler->error;
+  }
```
ivoanjo, line 35:
```suggestion
  if (now_ns <= 0) return sampler->error = ERR_CLOCK_FAIL;
```

(I'll show myself out :rofl:)

ext/ddtrace_profiling_native_extension/collectors_discrete_dynamic_sampler.c
```diff
@@ -28,6 +43,7 @@ static void _discrete_dynamic_sampler_reset(discrete_dynamic_sampler *sampler, l
     // to compute stats. Otherwise, we'd readjust on the next event that comes and thus be operating
     // with very incomplete information
     .last_readjust_time_ns = now_ns,
+    .error = now_ns > 0 ? NULL : ERR_CLOCK_FAIL,
```
ivoanjo, line 46: Wait... why do we need to check for this again, if we've already checked at the beginning of the function?

ext/ddtrace_profiling_native_extension/collectors_discrete_dynamic_sampler.c
```diff
@@ -36,29 +52,41 @@ static void _discrete_dynamic_sampler_reset(discrete_dynamic_sampler *sampler, l
...
 
-static void _discrete_dynamic_sampler_set_overhead_target_percentage(discrete_dynamic_sampler *sampler, double target_overhead, long now_ns) {
+static const char* _discrete_dynamic_sampler_set_overhead_target_percentage(discrete_dynamic_sampler *sampler, double target_overhead, long now_ns) {
   if (target_overhead <= 0 || target_overhead > 100) {
     rb_raise(rb_eArgError, "Target overhead must be a double between ]0,100] was %f", target_overhead);
   }
   sampler->target_overhead = target_overhead;
-  _discrete_dynamic_sampler_reset(sampler, now_ns);
+  return _discrete_dynamic_sampler_reset(sampler, now_ns);
 }
```
ivoanjo, line 70: Hmmm, this seems a bit inconsistent -- we're reporting the `now_ns` via the new mechanism of returning a string, but if the `target_overhead` doesn't make sense, we raise directly :eyes:

ext/ddtrace_profiling_native_extension/collectors_discrete_dynamic_sampler.c
```diff
@@ -36,29 +52,41 @@ static void _discrete_dynamic_sampler_reset(discrete_dynamic_sampler *sampler, l
...
+static discrete_dynamic_sampler_should_sample_result _discrete_dynamic_sampler_should_sample(discrete_dynamic_sampler *sampler, long now_ns) {
+  if (now_ns <= 0) {
+    sampler->error = ERR_CLOCK_FAIL;
+  }
+
+  if (sampler->error) {
+    return (discrete_dynamic_sampler_should_sample_result) {
+      .error = sampler->error,
+    };
+  }
```
ivoanjo, line 88: Minor: Me doing code golfing on a Friday afternoon :rofl: 

```suggestion
  if (now_ns <= 0) {
    return (discrete_dynamic_sampler_should_sample_result) {
      .error = (sampler->error = ERR_CLOCK_FAIL),
    };
  }
```

ivoanjo (commented):
Left a few more notes.

I must say, after going through this PR it's definitely... a very big change to enable the discrete sampler to say "hey, the clock is broken".

I kinda wonder if instead we should move the responsibility of checking the clock back to the `CpuAndWallTimeWorker`. It already needs to check for errors when calling the discrete sampler so it's not like we're "hiding" a lot of complexity away from it.

On the other hand I kinda realize my suggestion may be ~slightly cruel after you spent all this time on this PR (and I did a bit more on top as I was reviewing it -- the thought only kinda occurred to me at the end).

Also, the current mechanism is definitely more flexible in terms of allowing many more error states to be introduced, whereas my suggestion only really solves the issue of the clock error.

I leave the decision up to you ;)

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -318,6 +321,11 @@ static VALUE _native_new(VALUE klass) {
 
   reset_stats_not_thread_safe(state);
 
+  const char *error = discrete_dynamic_sampler_init(&state->allocation_sampler, "allocation");
+  if (error) {
+    handle_allocation_sampler_error(state, error, false);
+  }
```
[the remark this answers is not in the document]
AlexJF: Rewrote the core logic according to main comment but still felt like I needed to address this given that:
* We intend on `.new` on these native objects to be fully functional even if not initialized.
* Proper creation of the discrete sampler requires initialization.

I addressed the NPE in `disable_tracepoints` and added an explicit check on startup for any delayed errors that may already be there. I think that would cover it?

ext/ddtrace_profiling_native_extension/collectors_discrete_dynamic_sampler.c
```diff
@@ -310,6 +290,7 @@ VALUE discrete_dynamic_sampler_state_snapshot(discrete_dynamic_sampler *sampler)
 // Below here is boilerplate to expose the above code to Ruby so that we can test it with RSpec as usual.
 
 static VALUE _native_new(VALUE klass);
+static VALUE _native_initialize(VALUE klass, VALUE now);
```
ivoanjo, line 293: Minor: The argument here is actually the instance ;)

```suggestion
static VALUE _native_initialize(VALUE self, VALUE now);
```

ext/ddtrace_profiling_native_extension/collectors_discrete_dynamic_sampler.c
```diff
@@ -347,29 +331,39 @@ static const rb_data_type_t sampler_typed_data = {
 static VALUE _native_new(VALUE klass) {
   sampler_state *state = ruby_xcalloc(sizeof(sampler_state), 1);
 
-  discrete_dynamic_sampler_init(&state->sampler, "test sampler");
-
   return TypedData_Wrap_Struct(klass, &sampler_typed_data, state);
 }
```
ivoanjo, line 335: Minor: Should we still perhaps call `_init` here (possibly with a dummy time), just to avoid exposing the non-initialized sampler to Ruby code?

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -325,6 +330,14 @@ static VALUE _native_new(VALUE klass) {
 
   reset_stats_not_thread_safe(state);
 
+  long now = monotonic_wall_time_now_ns(DO_NOT_RAISE_ON_FAILURE);
+  if (now == 0) {
+    // NOTE: Throwing an exception on new means leaking memory as noted above so
+    //       lets let things carry on but store a delayed error.
+    delayed_error(state, ERR_CLOCK_FAIL);
```
ivoanjo, line 337: :thinking: Since in this case, the reason to not raise exceptions is to avoid leaking the `state`, we could `free(state)` and then go ahead and raise the exception.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -408,13 +422,20 @@ static VALUE _native_sampling_loop(DDTRACE_UNUSED VALUE _self, VALUE instance) {
     }
   }
 
+  // If we already got a delayed exception registered even before starting, raise before starting
+  if (state->failure_exception != Qnil) {
+    disable_tracepoints(state);
+    rb_exc_raise(state->failure_exception);
+  }
```
ivoanjo, line 429: Minor: I would maybe check for this even before checking the `old_state` -- the thinking being, if the current instance is in error already, it's not worth even complaining if there's other instances.

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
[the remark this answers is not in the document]
ivoanjo: Sounds good!

ivoanjo (approved):
:+1: LGTM

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
AlexJF: Ah yeah wasn't sure if it was just that malloc or there was some other internal thing that would cause a leak around the runtime 😅

PR #3440, opened 2024-02-06
[PROF-9044] Handle timing hiccups

AlexJF:
**What does this PR do?**
Introduce a mechanism so that point-in-time hiccups in timing information cannot effectively lead to disabling of the sampler for whole minutes.

**Motivation:**
VM suspension, system overload and other noise may occasionally lead to timing hiccups where time where the application was suspended is misattributed to time spent sampling (e.g. VM suspension occurs in between calls to `should_sample` and `after_sample`). 

This (hopefully) rare occurrence can skew our sampling statistics so much that the sampler would be effectively disabled for whole minutes. We want to limit this impact and try to have the sampler ignore sampling times that are deemed too big (our reference target is that, if forced to sample, we want to sample at least 1 event per second).

**Additional Notes:**
This PR tells a story:

1. We start by adding a test describing a scenario where due to some weird circumstance (e.g. suspension), the sampling time for a single event is reported as being 1000 seconds. With existing code the test fails with:
    ```
    1) Datadog::Profiling::Collectors::DiscreteDynamicSampler a weird hiccup shouldn't be able to disable sampling for whole minutes
     Failure/Error: expect(stats[:num_samples]).to be >= 60
     
       expected: >= 60
            got:    0
     # ./spec/datadog/profiling/collectors/discrete_dynamic_sampler_spec.rb:324:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:230:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:115:in `block (2 levels) in <top (required)>'
    ```
2. We then introduce a mechanism that limits the maximum amount of sampling time we're willing to consider. This maximum amount is chosen so as to represent a practical minimum sampling rate of 1 sample per adjustment window (ergo 1 sample per second). With this change the test passes:
    ```
      a weird hiccup shouldn't be able to disable sampling for whole minutes

     Finished in 0.08911 seconds (files took 1.03 seconds to load)
     1 example, 0 failures
    ```

**How to test the change?**

**For Datadog employees:**


spec/datadog/profiling/collectors/discrete_dynamic_sampler_spec.rb
```diff
@@ -15,10 +15,22 @@
...
-    start_ns = (@now * 1e9).to_i
-    end_ns = start_ns + (sampling_seconds * 1e9).to_i
-    sampler._native_after_sample(end_ns) / 1e9 if sampler._native_should_sample(start_ns)
+    sampled = false
+    if sampler._native_should_sample(to_ns(@now))
+      sampler._native_after_sample(to_ns(@now + sampling_seconds))
+      @now += sampling_seconds
+      sampled = true
+    end
+    sampled
```
ivoanjo, line 33: Minor: Tiny simplification ;)

```suggestion
    return false unless sampler._native_should_sample(to_ns(@now))
    
    sampler._native_after_sample(to_ns(@now + sampling_seconds))
    @now += sampling_seconds
    true
```

spec/datadog/profiling/collectors/discrete_dynamic_sampler_spec.rb
```diff
@@ -6,7 +6,7 @@
 
   before do
     skip_if_profiling_not_supported(self)
-    @now = Time.now.to_f
+    @now = Time.now.to_r
```
ivoanjo, line 9: I'm not sure this change is needed -- in fact, this spec works as well with `to_i`, and looking at the code it kinda makes sense since `@now` is used to calculate time differences so any point in time is expected to yield the same results.

spec/datadog/profiling/collectors/discrete_dynamic_sampler_spec.rb
```diff
@@ -245,19 +256,22 @@ def sampler_current_events_per_sec
...
     # time for us to be able to take that probing hit assuming things remain the same. Each adjustment window
-    # with no sampling activity earns us 0.02 seconds of budget. Therefore we need 4 of these to go by before
-    # we see the next probing sample.
-    stats = simulate_load(duration_seconds: 3, events_per_second: 4, sampling_seconds: 2)
-    expect(stats[:num_samples]).to eq(0)
-    stats = simulate_load(duration_seconds: 1, events_per_second: 4, sampling_seconds: 2)
+    # with no sampling activity earns us 0.02 seconds of budget. Therefore in theory we'd need 4 of these windows
+    # to go by before we see the next probing sample. However, with our new minimum sampling target, we are actually
+    # clamping the 0.08 sampling_seconds for an individual event to just 0.02 so we expect to see a sample in the next
+    # window already
```
ivoanjo, line 273: Minor: The reference to "minimum sampling target" makes sense in the context of this PR, but may be a bit confusing when reading later so I've tweaked the text a bit ;)

```suggestion
    # to go by before we see the next probing sample. However, an additional factor comes in -- the minimum sampling
    # target -- and we are actually clamping the 0.08 sampling_seconds for an individual event to just 0.02 so we expect
    # to see a sample in the next window already
```

ext/ddtrace_profiling_native_extension/collectors_discrete_dynamic_sampler.c
```diff
@@ -327,6 +347,8 @@ VALUE discrete_dynamic_sampler_state_snapshot(discrete_dynamic_sampler *sampler)
     ID2SYM(rb_intern("sampling_probability")),            /* => */ DBL2NUM(sampler->sampling_probability * 100),
     ID2SYM(rb_intern("events_since_last_readjustment")),  /* => */ ULONG2NUM(sampler->events_since_last_readjustment),
     ID2SYM(rb_intern("samples_since_last_readjustment")), /* => */ ULONG2NUM(sampler->samples_since_last_readjustment),
+    ID2SYM(rb_intern("max_sampling_time_ns")),            /* => */ LONG2NUM(sampler->max_sampling_time_ns),
```
ivoanjo, line 350: Minor: Since we expose this, may be worth adding a spec that validates its value?

ivoanjo (approved):
:+1: LGTM

The macOS CI failures are unrelated -- https://github.com/DataDog/dd-trace-rb/pull/3444 .

spec/datadog/profiling/collectors/discrete_dynamic_sampler_spec.rb
AlexJF: Uh yeah remnant from some troubleshooting I was doing on an earlier flaky version of a test (floats will approximate the timings from `Time.now` so I tried to keep to rationals to ensure no approximation was done. but then I simply rewrote the test to not care about it 😂 )

PR #3445, opened 2024-02-07
Fix minitest detection spec

TonyCTHsu:
**What does this PR do?**

Fix failing spec on MacOS and CircleCI, where `steep` is bundled. 

`steep` depends on  `activesupport`, which depends on `minitest`. So the test is activated on CI. 

With no version constraint on the version, Minitest release 5.22.1 on Feb 6
https://github.com/minitest/minitest/blob/master/History.rdoc#label-5.22.1+-2F+2024-02-06

Our test suite failed with such change. 

To fix it, add a dummy test case so the process status would be successful.

**Note**: This is something we should be working on continuing improving our environment.


ivoanjo (approved):
:+1: Thanks for debugging this!

ekump (approved):

marcotc (approved):

PR #3449, opened 2024-02-08
Fix flakiness of gc_force_recycle tests

AlexJF:
**What does this PR do?**
Increases the probability to re-use a recycled object slot.

**Motivation:**
Reduce flakiness seen in gc_force_recycle tests.

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**

### DIFF
--- a/spec/datadog/profiling/stack_recorder_spec.rb
+++ b/spec/datadog/profiling/stack_recorder_spec.rb
@@ -613,6 +613,7 @@ RSpec.describe Datadog::Profiling::StackRecorder do
                 end
               end
             end
+            raise 'could not allocate an object in a recycled slot'
           end
 
           it 'enforces seen id flag on objects on recycled slots that get sampled' do
--- a/spec/datadog/profiling/stack_recorder_spec.rb
+++ b/spec/datadog/profiling/stack_recorder_spec.rb
@@ -554,6 +554,26 @@ RSpec.describe Datadog::Profiling::StackRecorder do
             described_class::Testing._native_has_seen_id_flag(obj)
           end
 
+          # This method attempts to allocate an object on a recycled heap slot.
+          #
+          # Heap slot recycling was a troublesome feature that has been removed from Rubies >= 3.1
+          # in which an object could be freed through a fast-path that bypassed a lot of runtime
+          # machinery such as finalizers or object id tracking and thus introduced a fair amount
+          # of buggy behaviour. Some of this buggy behaviour manifests when a recycled slot gets
+          # re-used by a new live object: the new live object id will be the same as the id of
+          # the object that was recycled, violating a core constraint of Ruby objects: object ids
+          # are unique and non-repeatable.
+          #
+          # Recycling an object slot is easy (accomplished by a rb_gc_force_recycle native method call).
+          # More difficult is allocating an object on a recycled slot. Ruby gives us no control on
+          # where to allocate an object so we have to play a probability game. This method attempts to
+          # maximize our chances of quickly getting an object in a recycled slot by:
+          # 1. Force recycling 1000 objects.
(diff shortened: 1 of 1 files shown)


spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -555,25 +555,39 @@ def has_seen_id_flag(obj)
...
+              # Force recycle the given object
+              described_class::Testing._native_gc_force_recycle(obj)
+            end
 
-            # Repeatedly allocate objects until we find one that resolves to the same id
+            # Repeatedly allocate objects until we find one that resolves to the id of one of
+            # the force recycled objects
+            objs = []
             loop do
               # Instead of doing this one at a time which would be slow given id2ref will
```
ivoanjo, line 578: Should we perhaps move this inside the loop? I'm thinking in cases where we get "unlucky", this may cause memory to spike a lot (and may even lead to an OOM kill).

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -555,25 +555,39 @@ def has_seen_id_flag(obj)
           end
 
           def create_obj_in_recycled_slot(should_sample_original: false)
```
ivoanjo, line 557: Minor: In retrospect, I think it would be nice to have here a comment explaining what this bizarre method is trying to do, since it's really weird.

ivoanjo (approved):
:+1: LGTM

spec/datadog/profiling/stack_recorder_spec.rb
AlexJF: Will add one!

spec/datadog/profiling/stack_recorder_spec.rb
AlexJF: It being outside the loop was intentional: I don't want GC to be able to reclaim slots from a previous iteration otherwise we risk entering an infinite loop (e.g. a GC runs between each iteration and during an iteration all 1000 objects get allocated on slots cleaned up by the previous GC).

Given this method forcibly creates 1000 recycled slots and on each iteration we will create 1000 new objects, I'd hope that before deciding to expand heap pages and potentially triggering an OOM it'd decide to simply re-use the free slots it already has.

ivoanjo: That's fair. What do you think of perhaps limiting the number of iterations? E.g. going from something like

```ruby
loop do
  ...
end
```

to

```ruby
100.times do 
  ...
end
raise 'Could not cause object to be recycled!'
```

I think if by the time we get to some large number of objects, if we haven't managed to trigger the issue, it seems cleaner to fail the test, rather than letting the process balloon and die.

(Only a suggestion, happy to keep the `loop` if you're not convinced!)

ivoanjo: Really minor: I think it may be worth blowing up explicitly, as right now if it the looping fails, it'll return `nil`, which will still make the specs fail but in a rather non-obvious way.

```
  1) Datadog::Profiling::StackRecorder#serialize heap samples and sizes when enabled on Rubies supporting rb_gc_force_recycle enforces seen id flag on untracked objects that replace tracked recycled objects
     Failure/Error: expect(has_seen_id_flag(recycled_obj)).to be true
     
       expected true
            got false
```

AlexJF: Makes sense! Added an explicit failure:

```
  5) Datadog::Profiling::StackRecorder#serialize heap samples and sizes when enabled on Rubies supporting rb_gc_force_recycle correctly handles lifecycle of objects on recycled slots that get sampled
     Failure/Error: raise 'could not allocate an object in a recycled slot'
     
     RuntimeError:
       could not allocate an object in a recycled slot
```

PR #3452, opened 2024-02-08
Build lib injection artifact on demand

TonyCTHsu:
**What does this PR do?**

Change the Gitlab pipeline to build artifacts on demand.

1. Lib injection artifacts does not depends on downloading gem from rubygems.org
2. Patch gem version before building gem for preleases
3. Prepare artifacts by installing from local gem
4. For supported ruby versions, resolve and bundle dependencies under its own namespace


.gitlab/install_ddtrace_deps.rb
```diff
@@ -1,53 +1,79 @@
...
+FileUtils.mkdir_p(versioned_path, verbose: true)
+
+gemfile_file_path = versioned_path.join('Gemfile')
+
+File.open(gemfile_file_path, 'w') do |file|
+  file.write("source \"https://rubygems.org\"\n")
+  file.write("gem 'ddtrace', '#{ENV.fetch('RUBY_PACKAGE_VERSION')}', path: '#{current_path}'")
+end
+
+$stdout.puts '=== Reading Gemfile ==='
```
marcotc, line 24:
```suggestion
puts '=== Reading Gemfile ==='
```
Do we need `$stdout` declared explicitly for a particular reason?

marcotc (approved):

PR #3466, opened 2024-02-20
[PROF-9163] Fix missing profiling code hotspots when using ddtrace+otel

ivoanjo:
**What does this PR do?**

This PR fixes the profiler not sending code hotspots data when ddtrace is being used with OpenTelemetry (e.g. [when following these instructions](https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/otel_instrumentation/ruby/)).

Note that in particular this only works in apps that use

```ruby
require 'datadog/opentelemetry'
```

**Motivation:**

Since in this mode, otel data goes through ddtrace, code hotspots should be correctly supported. The only reason this didn't work is that otel spans were handled slightly differently, and the profiler was not correctly accounting for this.

**Additional Notes:**

N/A

**How to test the change?**

I've updated the `Rakefile` to run the profiler test suite with and without opentelemetry, and added test coverage for this.

To test it manually, you can add `gem 'opentelemetry-sdk'` to your Gemfile, and the specs will run.

Here's also a trivial app you can use for testing:

```ruby
require 'opentelemetry/sdk'
require 'datadog/opentelemetry'

OpenTelemetry::SDK.configure
MyAppTracer = OpenTelemetry.tracer_provider.tracer('otel-testing')
MyAppTracer.in_span("do_work") do |span|
  sleep 2
end
```


AlexJF (approved):
LGTM and nice test coverage!

PR #3421, opened 2024-01-30
2.0

marcotc:
**What does this PR do?**

This PR contains all the changes for 2.0.0

PR #3472, opened 2024-02-21
Add tracing microbenchmarks

marcotc:
**What does this PR do?**
Adds a tracing microbenchmark that measures the total time the tracing takes to create a custom trace, until it reaches the writer. This does not measure the writer cost (it's stubbed out).

**Motivation:**

Microbenchmarks allows us to quickly compare the performance between two commits, making it feasible to run on every PR change. It also allows for quick iteration when performing performance improvements.

**How to test the change?**

You can see the [PR comment below](https://github.com/DataDog/dd-trace-rb/pull/3472#issuecomment-1958201602) with the results for this PR, which are uninteresting because it doesn't make Ruby prod changes.

When merged, the [2.0 PR](https://github.com/DataDog/dd-trace-rb/pull/3421) will continuously display the performance changes compared to master.

**For Datadog employees:**

### DIFF
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -169,6 +169,7 @@ benchmarks:
     - ./steps/run-benchmarks.sh
     - ./steps/analyze-results.sh
     - "./steps/upload-results-to-s3.sh || :"
+    - "./steps/upload-results-to-benchmarking-api.sh || :"
     - "./steps/post-pr-comment.sh || :"
   artifacts:
     name: "artifacts"
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -5,6 +5,7 @@
 variables:
   GITLAB_BENCHMARKS_CI_IMAGE: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/benchmarking-platform:ruby-gitlab
   GITLAB_DDPROF_BENCHMARK_CI_IMAGE: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/benchmarking-platform:ruby-ddprof-benchmark
+  BASE_CI_IMAGE: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/benchmarking-platform:dd-trace-rb
 
 .benchmarks:
   stage: macrobenchmarks
@@ -153,3 +154,32 @@ ddprof-benchmark:
     FF_USE_LEGACY_KUBERNETES_EXECUTION_STRATEGY: "true" # Important tweak for stability of benchmarks
     LATEST_COMMIT_ID: $CI_COMMIT_SHA
     KUBERNETES_SERVICE_ACCOUNT_OVERWRITE: dd-trace-rb
+benchmarks:
+  stage: benchmarks
+  when: on_success
+  tags: ["runner:apm-k8s-tweaked-metal"]
+  image: $BASE_CI_IMAGE
+  interruptible: true
+  timeout: 1h
+  script:
+    - export ARTIFACTS_DIR="$(pwd)/artifacts" && (mkdir "${ARTIFACTS_DIR}" || :)
+    - git config --global url."https://gitlab-ci-token:${CI_JOB_TOKEN}@gitlab.ddbuild.io/DataDog/".insteadOf "https://github.com/DataDog/"
+    - git clone --branch dd-trace-rb https://github.com/DataDog/benchmarking-platform platform && cd platform
+    - ./steps/capture-hardware-software-info.sh
+    - ./steps/run-benchmarks.sh
+    - ./steps/analyze-results.sh
+    - "./steps/upload-results-to-s3.sh || :"
+    - "./steps/post-pr-comment.sh || :"
+  artifacts:
+    name: "artifacts"
+    when: always
+    paths:
+      - artifacts/
+    expire_in: 3 months
+  variables:
+    KUBERNETES_SERVICE_ACCOUNT_OVERWRITE: dd-trace-rb
+    FF_USE_LEGACY_KUBERNETES_EXECUTION_STRATEGY: "true"
+    UPSTREAM_BRANCH: $CI_COMMIT_REF_NAME
+    UPSTREAM_PROJECT_ID: $CI_PROJECT_ID
+    UPSTREAM_PROJECT_NAME: $CI_PROJECT_NAME
+    UPSTREAM_COMMIT_SHA: $CI_COMMIT_SHA
--- a/benchmarks/Dockerfile
+++ b/benchmarks/Dockerfile
@@ -7,15 +7,15 @@ RUN echo 'APT::Get::Assume-Yes "true";' > /etc/apt/apt.conf.d/90circleci \
 ENV DEBIAN_FRONTEND=noninteractive
 
 # Install required packages
-RUN set -ex; \
-        apt-get update; \
-        mkdir -p /usr/share/man/man1; \
+RUN set -ex && \
+        apt-get update && \
+        mkdir -p /usr/share/man/man1 && \
         apt-get install -y --no-install-recommends \
             git mercurial xvfb \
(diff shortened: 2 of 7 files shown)


spec/datadog/profiling/validate_benchmarks_spec.rb
```diff
@@ -20,4 +20,8 @@
   describe 'profiler_sample_serialize' do
     it('runs without raising errors') { expect_in_fork { load './benchmarks/profiler_sample_serialize.rb' } }
   end
+
+  describe 'tracing_trace' do
+    it('runs without raising errors') { expect_in_fork { load './benchmarks/tracing_trace.rb' } }
+  end
```
ivoanjo, line 26: Minor: Placing this test here means it's also covered by the `skip_if_profiling_not_supported` e.g. it probably will be skipped on most folks' machines because you need to manually build libdatadog for profiler to run on macOS.

I would suggest either introducing a `context 'profiling benchmarks'` to contain the profiling stuff, and thus allowing this benchmark to be outside of it (I guess in the process also moving this file to `spec/datadog/`?), or alternatively introduce a separate `tracing/validate_benchmarks_spec.rb` that does not need the profiling checks at all.

P.s.: I guess we can remove the `, if: (RUBY_VERSION >= '2.4.0')` on line 3? :)

benchmarks/tracing_trace.rb
```diff
@@ -0,0 +1,48 @@
+# Used to quickly run benchmark under RSpec as part of the usual test suite, to validate it didn't bitrot
+VALIDATE_BENCHMARK_MODE = ENV['VALIDATE_BENCHMARK'] == 'true'
+
+return unless __FILE__ == $PROGRAM_NAME || VALIDATE_BENCHMARK_MODE
+
+require 'benchmark/ips'
+
+# Load the local version of the gem
+lib = File.expand_path('lib', __dir__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
```
ivoanjo, line 10: Minor: I'm not sure this is needed if you run with `bundle exec`. And running without `bundle exec` is not a great since it may pull random versions of stuff, so I'm not sure it's worth supporting (?)

benchmarks/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US:en
+
+# Install RubyGems
+RUN mkdir -p "$GEM_HOME" && chmod -R 777 "$GEM_HOME"
+
+# Upgrade RubyGems and Bundler
+RUN gem update --system 3.3.26
+RUN gem install bundler -v '~> 2.3.26'
+ENV BUNDLE_SILENCE_ROOT_WARNING 1
```
ivoanjo, line 34: Minor: I actually suspect that for modern docker Ruby images this is no longer needed, but I guess we may keep it around just for consistency with other docker images?

benchmarks/tracing_trace.rb
```diff
@@ -0,0 +1,48 @@
...
+    def write(trace)
+      # no-op
+    end
+
+    ::Datadog::Tracing::Writer.prepend(self)
+  end
+
+  def run_benchmark
+    Benchmark.ips do |x|
+      benchmark_time = VALIDATE_BENCHMARK_MODE ? { time: 0.001, warmup: 0 } : { time: 10.5, warmup: 2 }
```
ivoanjo, line 24: Minor: 10.5 seconds for the benchmark seems like an oddly-specific number? :rofl:

benchmarks/tracing_trace.rb
```diff
@@ -0,0 +1,48 @@
...
+      x.config(**benchmark_time)
+
+      def trace(x, depth)
+        x.report(
+          "#{depth} span trace",
+          (depth.times.map { "Datadog::Tracing.trace('op.name') {" } + depth.times.map { "}" }).join
+        )
```
ivoanjo, line 31: This is a really clever and clean way of doing this! :D

.gitlab/benchmarks.yml
ivoanjo, line 1: These changes are a bit harder to review! Maybe @ddyurchenko can give a quick pass?

ivoanjo (approved):
:+1: LGTM

.gitlab/benchmarks.yml
ddyurchenko: :+1: Taking a look at it.

ddyurchenko: @ivoanjo Looks good! :tada: 

I added [a script for uploading results to BP UI](https://github.com/DataDog/benchmarking-platform/commit/a3e517511fba59d4959875b850a5c95ac2bd1a6a).

Please change

```
    - ./steps/analyze-results.sh
    - "./steps/upload-results-to-s3.sh || :"
    - "./steps/post-pr-comment.sh || :"
```

to

```
    - ./steps/analyze-results.sh
    - "./steps/upload-results-to-s3.sh || :"
    - "./steps/upload-results-to-benchmarking-api.sh || :"
    - "./steps/post-pr-comment.sh || :"
```

ddyurchenko (commented):
Looks great, I suggest also adding uploading to BP UI (see [my comment above](https://github.com/DataDog/dd-trace-rb/pull/3472/files#r1499287343)), so we have long term trends tracking.

benchmarks/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+# Install required packages
+RUN set -ex; \
+        apt-get update; \
+        mkdir -p /usr/share/man/man1; \
+        apt-get install -y --no-install-recommends \
+            git mercurial xvfb \
+            locales sudo openssh-client ca-certificates tar gzip parallel \
+            net-tools netcat unzip zip bzip2 gnupg curl wget \
+            tzdata rsync vim less jq; \
+        rm -rf /var/lib/apt/lists/*;
```
p-datadog, line 18: Out of curiosity, why does this use `;` and not `&&` between commands?

marcotc: No reason, just copy/paste from existing solution.
I changed it to `&&` as suggested.

.gitlab/benchmarks.yml
marcotc: Done, thank you so much!

benchmarks/Dockerfile
marcotc: BUNDLE_SILENCE_ROOT_WARNING is still relevant, but you reminded me to upgrade the versions of RubyGems and Bundler.

benchmarks/tracing_trace.rb
marcotc: Good suggestion, done!

spec/datadog/profiling/validate_benchmarks_spec.rb
marcotc: I didn't notice the file location, sorry about that!
I created `tracing/validate_benchmarks_spec.rb` as suggested.

benchmarks/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+RUN mkdir -p "$GEM_HOME" && chmod -R 777 "$GEM_HOME"
+
+# Upgrade RubyGems and Bundler
+RUN gem update --system 3.5.6
+RUN gem install bundler -v '~> 2.5.6'
+ENV BUNDLE_SILENCE_ROOT_WARNING 1
+
+# Reinstall a recent version of the trace to help Docker cache dependencies.
+# Bump this version periodically.
+RUN gem install ddtrace -v 1.20.0
```
ekump, line 38: non-blocking: Do you think this is worth automating at some point? Or even updating manually as part of our release process? I can see us easily forgetting about this.

ekump (approved):

benchmarks/Dockerfile
marcotc: It's not really important to update this line, it's only used to install the dependencies of the gem in a best-effort fashion, so that `RUN bundle install` a few lines down ideally performs less work. It doesn't affect correctness.

Issue #3430, opened 2024-02-01
Segmentation faults in Sidekiq even when no jobs running


mnovelo:
**Current behaviour**

We have a ddtrace in our Sidekiq services (deployed using AWS Fargate). We've noticed that there have been some segfaults at seemingly random times. What most caught my interest was a seg fault when the Sidekiq service was idle.

**Expected behaviour**

No seg faults

**Steps to reproduce**

No idea. This Sidekiq service that had seg fault was sitting idle for 6 hours when it suddenly had a seg fault

**How does `ddtrace` help you?**

We like ddtrace and Datadog for APM because we get insights into our running services that we wouldn't be as easily able to through just CloudWatch logs

**Environment**

* **ddtrace version:** 1.19.0
* **Configuration block (`Datadog.configure ...`):** Provided below
* **Ruby version:** 3.2.3
* **Operating system:** ruby:3.2.3-slim Docker image running Fargate Linux ARM64 instances
* **Relevant library versions:** Gemfile.lock provided below

```ruby
# frozen_string_literal: true

if ENV['DD_AGENT_HOST'].present? && !ENV.fetch('PLATFORM_ENV', 'dev').eql?('dev')
  require 'ddtrace/auto_instrument'
  require 'datadog/statsd'
  require 'datadog/appsec'
  require 'ddtrace'

  Datadog.configure do |c|
    c.version = ENV.fetch('BUILD_VERSION', nil) if ENV['BUILD_VERSION'].present?
    c.env = ENV.fetch('PLATFORM_ENV', nil)
    c.service = if (sidekiq_role = ENV.fetch('SIDEKIQ_ROLE', nil)).present?
                  "sidekiq-#{sidekiq_role.downcase}"
                else
                  'portal'
                end

    c.tracing.sampling.default_rate = case ENV.fetch('PLATFORM_ENV', nil)
                                      when 'prod'
                                        0.5
                                      when 'test', 'labs'
                                        0.25
                                      else
                                        0
                                      end
    c.profiling.enabled = true
    c.runtime_metrics.enabled = true
    c.appsec.enabled = true

    c.appsec.instrument(:rails)

    c.tracing.instrument(:rack, {
                           distributed_tracing: true,
                           request_queuing: true,
                           headers: {
                             response: %w[Content-Type X-Request-Id X-Amzn-Trace-Id],
                           },
                         })
    c.tracing.instrument(:rails, {
                           distributed_tracing: true,
                           request_queuing: true,
[...]

PR #3473, opened 2024-02-22
[PROF-9147] Add workaround for Ruby VM bug causing crash in gc_finalize_deferred

ivoanjo:
**What does this PR do?**

This PR adds a workaround for a VM bug (https://bugs.ruby-lang.org/issues/19991) that has affected at least two customers (the other being https://github.com/DataDog/dd-trace-rb/issues/3430).

The workaround is... not very pretty... but I believe will get the job done, for the reasons explained in the big comment I added.

**Motivation:**

The good thing about this crash is that it seems extremely rare.

The bad thing is that without this workaround, only moving to Ruby 3.3+ is the fix, and we still have a lot of customers on older Ruby versions.

We've documented in https://docs.datadoghq.com/profiler/profiler_troubleshooting/ruby/#segmentation-faults-in-gc_finalize_deferred-in-ruby-versions-26-to-32 that the no signals workaround can be used to work around the issue but:

a. Enabling it for everyone on < 3.3 is really crappy, as it yields biased samples and the crash only happens in the presence of specific types of Ruby objects (e.g. objects with finalizers) that not all apps will run into.

b. For folks that are affected by this issue, it may take quite a while for them to connect the bug to the profiler, since their crash log will point at `gc_finalize_deferred` and won't mention datadog at all.

**Additional Notes:**

N/A

**How to test the change?**

This is the million dollar question. I am unable to reproduce the issue locally just by running the profiler.

I was able to simulate it by changing Ruby's `vm_trace.c` as follows:

```diff
@@ -1748,6 +1748,8 @@ rb_workqueue_register(unsigned flags, rb_postponed_job_func_t func, void *data)
     return TRUE;
 }

+static void dummy_method(void *_unused) { }
+
 void
 rb_postponed_job_flush(rb_vm_t *vm)
 {
@@ -1775,7 +1777,11 @@ rb_postponed_job_flush(rb_vm_t *vm)
             while ((index = vm->postponed_job_index) > 0) {
                 if (ATOMIC_CAS(vm->postponed_job_index, index, index-1) == index) {
                     rb_postponed_job_t *pjob = &vm->postponed_job_buffer[index-1];
-                    (*pjob->func)(pjob->data);
+                    rb_postponed_job_func_t the_func = (*pjob->func);
+                    if (the_func != dummy_method) rb_postponed_job_register_one(0, dummy_method, NULL);
+                    //if (the_func != dummy_method) rb_postponed_job_register_one(0, dummy_method, GET_VM()->objspace);
+                    (the_func)(pjob->data);
                 }
```

The idea being that it's equivalent to the profiler SIGPROF interruption landing right between reading `pjob->func` and `pjob->data`.

I then ran the following Ruby snippet (which triggers `gc_finalize_deferred`):

```ruby
class UsesFinalizer
  DUMMY_FINALIZER = proc { }
  def initialize = ObjectSpace.define_finalizer(self, DUMMY_FINALIZER)
end
UsesFinalizer.new while true
```

...and this gets me a crash that matches the ones linked in
* https://bugs.ruby-lang.org/issues/19991
* https://github.com/DataDog/dd-trace-rb/issues/3430

If instead of passing `NULL`, we uncommend the second option that passes in `GET_VM()->objspace`, the crash goes away.

Thus I believe that my fix works and is equivalent to what the experiment above does.

I plan to reach out to both customers that saw this issue and see if they'd be up for testing the fix.

**For Datadog employees:**

### DIFF
--- a/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -225,6 +225,11 @@ static VALUE rescued_sample_allocation(VALUE tracepoint_data);
 static VALUE active_sampler_instance = Qnil;
 static struct cpu_and_wall_time_worker_state *active_sampler_instance_state = NULL;
 
+// See handle_sampling_signal for details on what this does
+#ifdef NO_POSTPONED_TRIGGER
+  static void *gc_finalize_deferred_workaround;
+#endif
+
 // Used to implement CpuAndWallTimeWorker._native_allocation_count . To be able to use cheap thread-local variables
 // (here with `__thread`, see https://gcc.gnu.org/onlinedocs/gcc/Thread-Local.html), this needs to be global.
 //
@@ -243,6 +248,8 @@ void collectors_cpu_and_wall_time_worker_init(VALUE profiling_module) {
     if (sample_from_postponed_job_handle == POSTPONED_JOB_HANDLE_INVALID || after_gc_from_postponed_job_handle == POSTPONED_JOB_HANDLE_INVALID) {
       rb_raise(rb_eRuntimeError, "Failed to register profiler postponed jobs (got POSTPONED_JOB_HANDLE_INVALID)");
     }
+  #else
+    gc_finalize_deferred_workaround = objspace_ptr_for_gc_finalize_deferred_workaround();
   #endif
 
   VALUE collectors_module = rb_define_module_under(profiling_module, "Collectors");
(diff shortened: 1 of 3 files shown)


AlexJF (approved):
Smart!

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -517,7 +524,32 @@ static void handle_sampling_signal(DDTRACE_UNUSED int _signal, DDTRACE_UNUSED si
...
+    // `rb_postponed_job_register_one(gc_finalize_deferred, objspace)`.
+    // Clobbering this call means that `gc_finalize_deferred` would get called with `NULL`, causing a segmentation fault.
+    //
+    // Note that this is quite rare: our signal needs to land at exactly the point where the VM has read the function
+    // to execute, but has yet to read the arguments. @ivoanjo: I could only reproduce it by manually changing the VM
+    // code to simulate this happening.
+    //
+    // Thus, our workaround is simple: we pass in objspace as our argument, just in case the clobbering happens.
+    // In the happy path, we never use this argument so it makes no difference. In the buggy path, we avoid crashing the VM.
+    int result = rb_postponed_job_register(0, sample_from_postponed_job, gc_finalize_deferred_workaround /* instead of NULL */);
```
p-datadog, line 552: This took me some time to understand: the `gc_finalize_deferred_workaround` argument will be used in the call to the *other* function that is in the postponed queue, and the only function (in the Ruby core itself) that has a meaningful argument takes ObjectSpace.

So, if there is some other library/extension which also uses postponed jobs, it could still crash - and could possibly be given wrong argument (i.e. ObjectSpace instead of whatever the correct argument was stored earlier).

p-datadog (approved):

ext/ddtrace_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
ivoanjo: > So, if there is some other library/extension which also uses postponed jobs, it could still crash - and could possibly be given wrong argument (i.e. ObjectSpace instead of whatever the correct argument was stored earlier).

Yes! Although as per the discussion in https://github.com/ruby/ruby/pull/8949#issuecomment-1821441370 it seems this is quite rare, so if we ever run into one of those (which I really hope we don't) the [no signals workaround](https://docs.datadoghq.com/profiler/profiler_troubleshooting/ruby/#segmentation-faults-in-gc_finalize_deferred-in-ruby-versions-26-to-32) will still work to avoid crashes.

ivoanjo:
Thanks everyone for reviewing! I've additionally gotten good feedback in https://github.com/DataDog/dd-trace-rb/issues/3430#issuecomment-1969325331 that this fix works in practice, so I'm going to go ahead and merge it in.

PR #3479, opened 2024-02-23
update compatibility doc with support policy

ekump:
**2.0 Upgrade Guide notes**
We have updated our [compatibility page](https://docs.datadoghq.com/tracing/trace_collection/compatibility/ruby/) to include our library [support policy](https://docs.datadoghq.com/tracing/trace_collection/compatibility/ruby/#support-policy). 

**What does this PR do?**
Adds a support policy to our [compatibility document](https://docs.datadoghq.com/tracing/trace_collection/compatibility/ruby) and make minor tweaks to compatibility content to conform to the accepted support policy RFC (a non-public doc I can share via private means) to avoid duplicating content.

**Motivation:**
Better communicate Ruby language version and architecture support for ddtrace.

**Additional Notes:**
The support policy does not explain how long previous library major versions will be supported. That conversation is ongoing and will be addressed in a separate PR.

**How to test the change?**

**For Datadog employees:**

### DIFF
--- a/docs/Compatibility.md
+++ b/docs/Compatibility.md
@@ -4,156 +4,236 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
 
 ### Supported Ruby interpreters
 
-| Type  | Documentation              | Version | Support type                         | Gem version support |
-| ----- | -------------------------- | -----   | ------------------------------------ | ------------------- |
-| MRI   | https://www.ruby-lang.org/ | 3.3     | Full                                 | Latest              |
-|       |                            | 3.2     | Full                                 | Latest              |
-|       |                            | 3.1     | Full                                 | Latest              |
-|       |                            | 3.0     | Full                                 | Latest              |
-|       |                            | 2.7     | Full                                 | Latest              |
-|       |                            | 2.6     | Full                                 | Latest              |
-|       |                            | 2.5     | Full                                 | Latest              |
-|       |                            | 2.4     | EOL                                  | < 2.0.0             | [comment]: <> (# DEV-2.0: Update with release date)
-|       |                            | 2.3     | EOL                                  | < 2.0.0             |
-|       |                            | 2.2     | EOL                                  | < 2.0.0             |
-|       |                            | 2.1     | EOL                                  | < 2.0.0             |
-|       |                            | 2.0     | EOL since June 7th, 2021             | < 0.50.0            |
-|       |                            | 1.9.3   | EOL since August 6th, 2020           | < 0.27.0            |
-|       |                            | 1.9.1   | EOL since August 6th, 2020           | < 0.27.0            |
-| JRuby | https://www.jruby.org      | 9.3     | Full                                 | Latest              |
-|       |                            | 9.2     | Full                                 | Latest              |
+| Type  | Documentation              | Version | Support type        | Gem version support |
+|-------|----------------------------|---------|---------------------|---------------------|
+| MRI   | https://www.ruby-lang.org/ | 3.3     | [GA](#support-ga)   | Latest              |
+|       |                            | 3.2     | [GA](#support-ga)   | Latest              |
+|       |                            | 3.1     | [GA](#support-ga)   | Latest              |
+|       |                            | 3.0     | [GA](#support-ga)   | Latest              |
+|       |                            | 2.7     | [GA](#support-ga)   | Latest              |
+|       |                            | 2.6     | [GA](#support-ga)   | Latest              |
+|       |                            | 2.5     | [GA](#support-ga)   | Latest              |
+|       |                            | 2.4     | [EOL](#support-eol) | < 2.0.0             | [comment]: <> (# DEV-2.0: Update with release date)
+|       |                            | 2.3     | [EOL](#support-eol) | < 2.0.0             |
+|       |                            | 2.2     | [EOL](#support-eol) | < 2.0.0             |
+|       |                            | 2.1     | [EOL](#support-eol) | < 2.0.0             |
+|       |                            | 2.0     | [EOL](#support-eol) | < 0.50.0            |
+|       |                            | 1.9.3   | [EOL](#support-eol) | < 0.27.0            |
+|       |                            | 1.9.1   | [EOL](#support-eol) | < 0.27.0            |
+| JRuby | https://www.jruby.org      | 9.3     | [GA](#support-ga)   | Latest              |
+|       |                            | 9.2     | [GA](#support-ga)   | Latest              |
 
 ### Supported web servers
 
-| Type      | Documentation                     | Version      | Support type |
-| --------- | --------------------------------- | ------------ | ------------ |
-| Puma      | http://puma.io/                   | 2.16+ / 3.6+ | Full         |
-| Unicorn   | https://bogomips.org/unicorn/     | 4.8+ / 5.1+  | Full         |
-| Passenger | https://www.phusionpassenger.com/ | 5.0+         | Full         |
+| Type      | Documentation                     | Version      | Support type      |
+|-----------|-----------------------------------|--------------|-------------------|
+| Puma      | http://puma.io/                   | 2.16+ / 3.6+ | [GA](#support-ga) |
+| Unicorn   | https://bogomips.org/unicorn/     | 4.8+ / 5.1+  | [GA](#support-ga) |
+| Passenger | https://www.phusionpassenger.com/ | 5.0+         | [GA](#support-ga) |
 
 ### Supported tracing frameworks
 
-| Type        | Documentation                                   | Version               | Gem version support |
-| ----------- | ----------------------------------------------- | --------------------- | ------------------- |
-| OpenTracing | https://github.com/opentracing/opentracing-ruby | 0.4.1+                | < 2.0.0            |
+| Type        | Documentation                                   | Version | Support type        | Gem version support |
+|-------------|-------------------------------------------------|---------|---------------------|---------------------|
+| OpenTracing | https://github.com/opentracing/opentracing-ruby | 0.4.1+  | [EOL](#support-eol) | < 2.0.0             |
 
-*Full* support indicates all tracer features are available.
+### Supported operating systems
 
-*Deprecated* indicates support will transition to *Maintenance* in a future release.
+| OS           | Support type                        | Package version |
+|--------------|-------------------------------------|-----------------|
+| Linux x86-64 | [GA](#support-ga)                   | Latest          |
+| Linux ARM64  | [GA](#support-ga)                   | Latest          |
+| MacOS        | Dev environments only               | Latest          |
+| MS Windows   | [Unsupported](#support-unsupported) | Latest          |
 
-*Maintenance* indicates only critical bugfixes are backported until EOL.
+### Supported Datadog agent versions
 
-*EOL* indicates support is no longer provided.
-
-### Apple macOS support
-
-Use of `ddtrace` on macOS is supported for development, but not for production deployments.
-
-### Microsoft Windows support
-
-Using `ddtrace` on Microsoft Windows is currently unsupported. We'll still accept community contributions and issues,
-but will consider them as having low priority.
+| Datadog Agent version                                                    | Support type      | Package version |
+|--------------------------------------------------------------------------|-------------------|-----------------|
+| [7.x](https://docs.datadoghq.com/agent/basic_agent_usage/?tab=agentv6v7) | [GA](#support-ga) | Latest          |
+| [6.x](https://docs.datadoghq.com/agent/basic_agent_usage/?tab=agentv6v7) | [GA](#support-ga) | Latest          |
+| [5.x](https://docs.datadoghq.com/agent/basic_agent_usage/?tab=agentv5)   | [GA](#support-ga) | Latest          |
 
 ## Integrations
 
 For a list of available integrations, and their configuration options, refer to the following:
 
-| Name                       | Key                        | Versions Supported: MRI  | Versions Supported: JRuby | How to configure                    | Gem source                                                                     |
-| -------------------------- | -------------------------- | ------------------------ | --------------------------| ----------------------------------- | ------------------------------------------------------------------------------ |
-| Action Cable               | `action_cable`             | `>= 5.0`                 | `>= 5.0`                  | [Link][2]                           | [Link](https://github.com/rails/rails/tree/master/actioncable)               |
-| Action Mailer              | `action_mailer`            | `>= 5.0`                 | `>= 5.0`                  | [Link][3]                           | [Link](https://github.com/rails/rails/tree/master/actionmailer)              |
-| Action Pack                | `action_pack`              | `>= 4.0`                 | `>= 4.0`                  | [Link][4]                           | [Link](https://github.com/rails/rails/tree/master/actionpack)                |
-| Action View                | `action_view`              | `>= 4.0`                 | `>= 4.0`                  | [Link][5]                           | [Link](https://github.com/rails/rails/tree/master/actionview)                |
-| Active Job                 | `active_job`               | `>= 4.2`                 | `>= 4.2`                  | [Link][6]                           | [Link](https://github.com/rails/rails/tree/master/activejob)                 |
-| Active Model Serializers   | `active_model_serializers` | `>= 0.9`                 | `>= 0.9`                  | [Link][7]                           | [Link](https://github.com/rails-api/active_model_serializers)                |
-| Active Record              | `active_record`            | `>= 4.0`                 | `>= 4.0`                  | [Link][8]                           | [Link](https://github.com/rails/rails/tree/master/activerecord)              |
-| Active Support             | `active_support`           | `>= 4.0`                 | `>= 4.0`                  | [Link][9]                           | [Link](https://github.com/rails/rails/tree/master/activesupport)             |
-| AWS                        | `aws`                      | `>= 2.0`                 | `>= 2.0`                  | [Link][10]                          | [Link](https://github.com/aws/aws-sdk-ruby)                                  |
-| Concurrent Ruby            | `concurrent_ruby`          | `>= 0.9`                 | `>= 0.9`                  | [Link][11]                          | [Link](https://github.com/ruby-concurrency/concurrent-ruby)                  |
-| Dalli                      | `dalli`                    | `>= 2.0`                 | `>= 2.0`                  | [Link][12]                          | [Link](https://github.com/petergoldstein/dalli)                              |
-| DelayedJob                 | `delayed_job`              | `>= 4.1`                 | `>= 4.1`                  | [Link][13]                          | [Link](https://github.com/collectiveidea/delayed_job)                        |
-| Elasticsearch              | `elasticsearch`            | `>= 1.0`                 | `>= 1.0`                  | [Link][14]                          | [Link](https://github.com/elastic/elasticsearch-ruby)                        |
-| Ethon                      | `ethon`                    | `>= 0.11`                | `>= 0.11`                 | [Link][15]                          | [Link](https://github.com/typhoeus/ethon)                                    |
-| Excon                      | `excon`                    | `>= 0.50`                | `>= 0.50`                 | [Link][16]                          | [Link](https://github.com/excon/excon)                                       |
-| Faraday                    | `faraday`                  | `>= 0.14`                | `>= 0.14`                 | [Link][17]                          | [Link](https://github.com/lostisland/faraday)                                |
-| Grape                      | `grape`                    | `>= 1.0`                 | `>= 1.0`                  | [Link][18]                          | [Link](https://github.com/ruby-grape/grape)                                  |
-| GraphQL                    | `graphql`                  | `>= 2.2.6`, `2.1.11+`,`2.0.28+`, `1.13.21+`| `>= 2.2.6`, `2.1.11+`,`2.0.28+`, `1.13.21+`| [Link][19]                          | [Link](https://github.com/rmosolgo/graphql-ruby)                             |
-| gRPC                       | `grpc`                     | `>= 1.7`                 | *gem not available*       | [Link][20]                          | [Link](https://github.com/grpc/grpc/tree/master/src/rubyc)                   |
-| hanami                     | `hanami`                   | `>= 1`, `< 2`            | `>= 1`, `< 2`             | [Link][21]                          | [Link](https://github.com/hanami/hanami)                                     |
-| http.rb                    | `httprb`                   | `>= 2.0`                 | `>= 2.0`                  | [Link][22]                          | [Link](https://github.com/httprb/http)                                       |
-| httpclient                 | `httpclient`               | `>= 2.2`                 | `>= 2.2`                  | [Link][23]                          | [Link](https://github.com/nahi/httpclient)                                   |
-| httpx                      | `httpx`                    | `>= 0.11`                | `>= 0.11`                 | [Link][24]                          | [Link](https://gitlab.com/honeyryderchuck/httpx)                             |
-| Kafka                      | `ruby-kafka`               | `>= 0.7.10`              | `>= 0.7.10`               | [Link][25]                          | [Link](https://github.com/zendesk/ruby-kafka)                                |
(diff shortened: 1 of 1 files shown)


docs/Compatibility.md
```diff
@@ -4,156 +4,236 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
...
-| Sinatra                    | `sinatra`                  | `>= 1.4`                 | `>= 1.4`                  | [Link][45]                          | [Link](https://github.com/sinatra/sinatra)                                   |
-| Sneakers                   | `sneakers`                 | `>= 2.12.0`              | `>= 2.12.0`               | [Link][46]                          | [Link](https://github.com/jondot/sneakers)                                   |
-| Stripe                     | `stripe`                   | `>= 5.15.0`              | `>= 5.15.0`               | [Link][47]                          | [Link](https://github.com/stripe/stripe-ruby)                                |
-| Sucker Punch               | `sucker_punch`             | `>= 2.0`                 | `>= 2.0`                  | [Link][48]                          | [Link](https://github.com/brandonhilkert/sucker_punch)                       |
-
+| Name                       | Key                        | Versions Supported: MRI                     | Versions Supported: JRuby                   | How to configure | Gem source                                                                   |
```
ekump, line 61: no content changes to this table, just cleaning up formatting.

hestonhoffman (approved):
LGTM

docs/Compatibility.md
```diff
@@ -4,156 +4,247 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
...
+
+Datadog APM for Ruby is built upon dependencies defined in specific versions of the host operating system, Ruby
+runtime, certain Ruby libraries, and the Datadog Agent/API. When these versions are no longer supported by their
+maintainers, Datadog APM for Ruby limits its support for these as well.
+
+#### Levels of Support
+
+| **Level**                                              | **Support provided**                                                                                                                                                         |
+|--------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| <span id="support-unsupported">Unsupported</span>      | No implementation. [Contact our customer support team for special requests.](https://www.datadoghq.com/support/)                                                             |
```
brett0000FF, line 215: Related to the previous comment could you please use reference style links instead of in-line links? For example, this would be `[Contact our customer support team for special requests.][49]` with a reference like `[49]: https://www.datadoghq.com/support/`.

This way it's easier to maintain the page when we need to update links. 👍

brett0000FF (approved):
Thanks! Approved but I left two minor comments about the links:

- Moving reference links to the end of the page.
- Changing in-line links to reference style links.

marcotc (approved):

docs/Compatibility.md
```diff
@@ -4,156 +4,247 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
...
+| <span id="support-beta">Beta</span>                    | Initial implementation. May not yet contain all features. Support for new features, bug & security fixes provided on a best-effort basis.                                    |
+| <span id="support-ga">General Availability (GA)</span> | Full implementation of all features. Full support for new features, bug & security fixes.                                                                                    |
```
lloeki, line 217: To me "GA" sounds like:
- a point in time
- it's feature oriented, not version oriented

e.g in an announcement: "we announce general availability of feature foo" == "feature foo is now generally available".

Inspired by Node, I would like to suggest calling this "current".

docs/Compatibility.md
```diff
@@ -4,156 +4,247 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
...
+#### Library version support
+
+Datadog APM for Ruby will provide GA support for the latest major version and maintenance support for the previous major
+version of the library.
+
+| Gem Version | Support type                        |
+|-------------|-------------------------------------|
+| 2.x         | [GA](#support-ga)                   |
+| 1.x         | [Maintenance](#support-maintenance) |
+| 0.x         | [EOL](#support-eol)                 |
```
lloeki, line 244: For clarity, I feel there's a need to distinguish between "GA support" in the 2.x line. As is, it sounds like a hypothetical 2.0, 2.1, 2.2 would receive the exact same level of support as a hypothetical 2.3, i.e any bug fix that is landed on a 2.3.y are going to be all backported to 2.0.y, 2.1.y, 2.2.y.

docs/Compatibility.md
```diff
@@ -4,156 +4,247 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
...
+| <span id="support-unsupported">Unsupported</span>      | No implementation. [Contact our customer support team for special requests.](https://www.datadoghq.com/support/)                                                             |
+| <span id="support-beta">Beta</span>                    | Initial implementation. May not yet contain all features. Support for new features, bug & security fixes provided on a best-effort basis.                                    |
+| <span id="support-ga">General Availability (GA)</span> | Full implementation of all features. Full support for new features, bug & security fixes.                                                                                    |
+| <span id="support-maintenance">Maintenance</span>      | Full implementation of existing features. Does not receive new features. Support for critical bug & security fixes only.                                                     |
+| <span id="support-legacy">Legacy</span>                | Legacy implementation. May have limited function, but no maintenance provided. [Contact our customer support team for special requests.](https://www.datadoghq.com/support/) |
```
lloeki, line 219: Seems like `Legacy` != `Maintenance`, but I feel a bit confused as to how `Legacy` is distinct from `EOL`. It sounds like a LTS-upon-request-otherwise-EOL?

docs/Compatibility.md
```diff
@@ -4,156 +4,247 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
...
+runtime, certain Ruby libraries, and the Datadog Agent/API. When these versions are no longer supported by their
+maintainers, Datadog APM for Ruby limits its support for these as well.
+
+#### Levels of Support
+
+| **Level**                                              | **Support provided**                                                                                                                                                         |
+|--------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| <span id="support-unsupported">Unsupported</span>      | No implementation. [Contact our customer support team for special requests.](https://www.datadoghq.com/support/)                                                             |
+| <span id="support-beta">Beta</span>                    | Initial implementation. May not yet contain all features. Support for new features, bug & security fixes provided on a best-effort basis.                                    |
```
lloeki, line 216: > Initial implementation. May not yet contain all features.

Sounds odd to me, as in it looks like an alpha in disguise.

A definition I would have:

- `alpha`: "Checkpoint"/"early testing" phase. Partial featureset *regarding a target* (e.g for vX.Y we target having feature foo bar baz but only foo+bar are implemented), or feature is partially implemented (known bug, known missing corner case/subfeature, known missing support for an otherwise supported runtime or framework)
- `beta`: "Ironing out" phase. Complete featureset *regarding a target* (e.g for vX.Y we have feature foo bar baz), feature is fully implemented (all corner cases addressed, all runtimes/frameworks supported) but they are not battle-tested (there may be bugs, some might even be known and spelled out in release notes, but will be fixed later)
- `rc`:  Qualification phase. Complete featureset regarding a target, implementation has no known bugs, final testing phase to attempt to surface unknown issues. If none found, promoted "as-is" to GA.

docs/Compatibility.md
lloeki, line 215: > Unsupported: No implementation

I dont' quite grok this.

In fact throughout the table I feel like making reference to no/initial/full implementation/features very much muddies the point.

To me, both of the following feel much more synthetic and clearer at the same time:

https://guides.rubyonrails.org/maintenance_policy.html
https://nodejs.org/en/about/previous-releases

docs/Compatibility.md
ekump: That's fair. What do you think is the best way to communicate that this only applies to the latest minor versions? I could change the column header `Gem Version` to `Gem Major Version` and drop the `.x`?

We could also add a sentence in the text above the table along the lines of `Respective levels of support for major versions are provided via the latest minor versions. Datadog will generally not backport changes to previous minor versions via patch releases. For example if 2.3 is the latest version of 2.x, changes will not be backported to 2.2, 2.1, etcetera. `

docs/Compatibility.md
ekump: Now that you mention it, I agree. GA typically indicates a point in time. @delner any objection here?

docs/Compatibility.md
marcotc: In the context of Datadog APM, `beta` means a fully featured feature, but missing enough real-world data before labelling it as full release.

docs/Compatibility.md
```diff
@@ -4,156 +4,259 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
...
+|       |                            | 2.7       | [GA](#support-ga)   | Latest              |
+|       |                            | 2.6       | [GA](#support-ga)   | Latest              |
+|       |                            | 2.5       | [GA](#support-ga)   | Latest              |
+|       |                            | 2.4       | [EOL](#support-eol) | < 2.0.0             |
+|       |                            | 2.3       | [EOL](#support-eol) | < 2.0.0             |
+|       |                            | 2.2       | [EOL](#support-eol) | < 2.0.0             |
+|       |                            | 2.1       | [EOL](#support-eol) | < 2.0.0             |
+|       |                            | 2.0       | [EOL](#support-eol) | < 0.50.0            |
+|       |                            | 1.9.3     | [EOL](#support-eol) | < 0.27.0            |
+|       |                            | 1.9.1     | [EOL](#support-eol) | < 0.27.0            |
```
ivoanjo, line 22: Minor: Let's merge these together, it doesn't seem very relevant to have them separate anymore?

```suggestion
|       |                            | 1.9       | [EOL](#support-eol) | < 0.27.0            |```

File: docs/Compatibility.md [hunk 8]
```diff
@@ -4,156 +4,259 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
...
 
-*Full* support indicates all tracer features are available.
+### Supported operating systems
 
-*Deprecated* indicates support will transition to *Maintenance* in a future release.
+| OS           | Support type                        | Package version |
+|--------------|-------------------------------------|-----------------|
+| Linux x86-64 | [GA](#support-ga)                   | Latest          |
+| Linux ARM64  | [GA](#support-ga)                   | Latest          |
+| MacOS        | Dev environments only               | Latest          |
```
ivoanjo, line 46: Minor: It's actually `macOS` ;)

```suggestion
| macOS        | Dev environments only               | Latest          |
```

docs/Compatibility.md
```diff
@@ -4,156 +4,259 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
...
+| OpenTracing | https://github.com/opentracing/opentracing-ruby | 0.4.1+  | [EOL](#support-eol) | < 2.0.0             |
 
-*Full* support indicates all tracer features are available.
+### Supported operating systems
 
-*Deprecated* indicates support will transition to *Maintenance* in a future release.
+| OS           | Support type                        | Package version |
+|--------------|-------------------------------------|-----------------|
+| Linux x86-64 | [GA](#support-ga)                   | Latest          |
+| Linux ARM64  | [GA](#support-ga)                   | Latest          |
```
ivoanjo, line 45: Should we add some statement regarding Linux on other architectures? (Dev environments only? Talk to us?)

docs/Compatibility.md
```diff
@@ -4,156 +4,259 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
...
-| Sneakers                   | `sneakers`                 | `>= 2.12.0`              | `>= 2.12.0`               | [Link][46]                          | [Link](https://github.com/jondot/sneakers)                                   |
-| Stripe                     | `stripe`                   | `>= 5.15.0`              | `>= 5.15.0`               | [Link][47]                          | [Link](https://github.com/stripe/stripe-ruby)                                |
-| Sucker Punch               | `sucker_punch`             | `>= 2.0`                 | `>= 2.0`                  | [Link][48]                          | [Link](https://github.com/brandonhilkert/sucker_punch)                       |
+- **Major version updates** (e.g. `1.0.0` to `2.0.0`) may change support for any runtime
+  from [Beta](#support-beta)/[GA](#support-ga)
+  to [Maintenance](#support-maintenance)/[Legacy](#support-legacy)/[EOL](#support-eol).
+- **Minor version updates** (e.g. `1.0.0` to `1.1.0`) may change support for any runtime
+  from [Beta](#support-beta)/[GA](#support-ga) to [Maintenance](#support-maintenance).
```
ivoanjo, line 138: I still think we're backing ourselves into a corner here. As a new Ruby comes out each year, this means that either we may need to carry a given Ruby version around for longer than we wanted to (sound familiar?) or we'll be forced to bump the major version sooner/when otherwise we wouldn't really need to.

Imagine in the extreme that we do no more features for ddtrace ever, only maintenance fixes. It's weird to think that a customer that onboards today on Ruby 3.3 will need to go through several "major" releases where actually nothing changes, because we just did a minor fix here or there, but we did them only because we wanted to drop support for older Rubies.

Furthermore, I don't really think this is a violation of semver, because support for this is built-into rubygems! Rubygems will not install a ddtrace version that doesn't work for a user's Ruby version, so if we drop support on a minor version (as many upstream libraries do), users would never accidentally pick up a version that wouldn't work for them.

Can we discuss this a bit before moving forward with this part?

docs/Compatibility.md
```diff
@@ -4,156 +4,259 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
...
+#### Library version support
+
+Datadog APM for Ruby will provide GA support for the latest major version and maintenance support for the previous major
+version of the library.
+
+| Gem Version | Support type                        |
+|-------------|-------------------------------------|
+| 2.x         | [GA](#support-ga)                   |
+| 1.x         | [Maintenance](#support-maintenance) |
+| 0.x         | [EOL](#support-eol)                 |
```
ivoanjo, line 150: Is 1.x going to be maintained for the whole of the 2.x series? That seems like a big commitment to make at this point?

ivoanjo (commented):
Left a few notes!

docs/Compatibility.md
ekump: I don't think we should mark any architectures as dev-only if we don't have some sort of CI in place for it. I could add a sentence below the table that says:

Need linux support for a CPU architecture not listed? [Contact our customer support team for special requests.](https://www.datadoghq.com/support)

ivoanjo: Seems reasonable!

docs/Compatibility.md
ekump: Yes, that was my (and most of the guild's ) understanding. Maintenance implies critical bug and security fixes. I don't expect that to be overly-burdensome. If you strongly disagree we should discuss with the guild.

ivoanjo: I just spotted the thread where this was discussed, and I thought the scope was larger, so I agree with the current plan :+1:

docs/Compatibility.md
ivoanjo: We discussed this privately via slack, so I'm happy with the current text :+1:

docs/Compatibility.md
```diff
@@ -4,156 +4,258 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
...
+|       |                            | 3.0       | [GA](#support-ga)   | Latest              |
+|       |                            | 2.7       | [GA](#support-ga)   | Latest              |
+|       |                            | 2.6       | [GA](#support-ga)   | Latest              |
+|       |                            | 2.5       | [GA](#support-ga)   | Latest              |
+|       |                            | 2.4       | [EOL](#support-eol) | < 2.0.0             |
+|       |                            | 2.3       | [EOL](#support-eol) | < 2.0.0             |
+|       |                            | 2.2       | [EOL](#support-eol) | < 2.0.0             |
+|       |                            | 2.1       | [EOL](#support-eol) | < 2.0.0             |
+|       |                            | 2.0       | [EOL](#support-eol) | < 0.50.0            |
+|       |                            | 1.9       | [EOL](#support-eol) | < 0.27.0            |```
```
ivoanjo, line 21: Stray ```` ``` ````

```suggestion
|       |                            | 1.9       | [EOL](#support-eol) | < 0.27.0            |
```

ivoanjo (approved):
:+1: LGTM

docs/Compatibility.md
delner: No strong opinion on my end. I will defer. `Current` is fine.

docs/Compatibility.md
delner: I like the explicit nature of the expanded table, showing each version. You could collapse these into ranges where the info is repetitive e.g. `2.1-2.4`. But that's a minor issue.

docs/Compatibility.md
delner: `Legacy` is unimplemented; it was a postulated level of support. If not applicable, we might just want to delete it under the principle of `less is more`.

ekump: @delner I think that makes sense. We could always re-introduce the tier if we ever wind up using it.

docs/Compatibility.md
delner: @lloeki I agree with your descriptions in the engineering sense of how these terms are actually used. I would suggest that this policy does not describe versions (as we'd attach to our releases) but a general purpose label used for general messaging, particularly to those less initiated in the world of developer terminology.

TL;DR different target audience.

docs/Compatibility.md
delner: @ekump your interpretation is correction. This was meant to address things that might be asked, but is not yet (or might never be) supported.

I'm fine with the name `Unimplemented` instead, if that seems clearer.

delner (approved):
Some minor suggestions/concurrence with some other suggestions, but the bulk of this looks good. My only other suggestion would be to strike `APM` from where we see `Datadog APM for Ruby`, given this library contains much more than just APM.

docs/Compatibility.md
ekump: @delner As I understand it, this table is supposed to convey the level of support available in major versions. I agree with @lloeki that it's unclear in its current form that normally, that support is provided via new minor releases. Would listing every single version of the library help convey this information better?

docs/Compatibility.md
ekump: I have no objection to either `Current` or `Latest`. I lean slightly towards `Latest` if if provides better consistency. @lloeki you get the deciding vote.

lloeki: Either one is fine to me. `Latest` sounds just as good, let's go for that.

docs/Compatibility.md
lloeki: The annoying part with explicitly listing versions including minor is that we'd have to update the doc on every release, which would be annoying.

> `Gem Major Version`

That would have been my suggestion, but it also makes it less explicit that only the latest major is supported. 

Counterpoint though: that's what [node does](https://nodejs.org/en/about/previous-releases) and it seems fine.

I think a visual representation [such as this one](https://raw.githubusercontent.com/nodejs/Release/main/schedule.svg?sanitize=true) would reaaaally drive the point home.

docs/Compatibility.md
lloeki: Agreed, less source of confusion.

docs/Compatibility.md
lloeki: I don't know what the status for profiling/libdatadog is, but for libddwaf it goes like this:

a) We have binaries for supported platofrms
b) `libddwaf` is able to be compiled on basically any platform, the `ruby` platform gem can† do that.

If customers figure out a way to have it build (which is essentially having a C++ compiler + Cmake) then they can have ASM work on ppc64le, RISC-V, FreeBSD, Solaris, what have you.

† It's not done yet but I'm planning to do it to support the people that use `force_ruby_platform` on supported platforms. The above will come as a consequence of that.

ivoanjo: > I don't know what the status for profiling/libdatadog is, but for libddwaf it goes like this:

The current status is we're shipping libdatadog binaries for [x86-64, arm64] x [linux glibc, linux musl], so that's what profiler supports out of the box.

(It's undocumented, but if you can manually compile libdatadog to more platforms, and profiler does usually work on them as well).

---

Having said that, @lloeki I think @ekump has a reasonable point about the difference between "we have binaries for obscure platforms" and "we're actively testing and supporting them". If we're not testing on Solaris, and would struggle if someone came to us with a bug that only triggered on Solaris, should we state we *support* Solaris?

I think it's OK to have a "best-effort, come talk to us if this really important to you" tier, but I think it makes sense to be clear in this table about how much effort we would spend on issues for these platforms.

docs/Compatibility.md
lloeki: I understand (and glad we agree on the technical terminology).

That said, I feel like it can create confusion. For the target audience I would fold this under the umbrella `prerelease` term.

docs/Compatibility.md
lloeki: > If we're not testing on Solaris, and would struggle if someone came to us with a bug that only triggered on Solaris, should we state we support Solaris?

I agree, hence why I said these fall under "unsupported".

The conundrum I have is that once I add support for the `ruby` platform to properly support `{x86_64,aarch64}-linux-{gnu,musl}` the *other* platforms are going to start working but we don't want ambiguity as to whether they're supported through the fact that the `ruby` platform is.

> I think it's OK to have a "best-effort, come talk to us if this really important to you" tier, but I think it makes sense to be clear in this table about how much effort we would spend on issues for these platforms.

I like the "best effort" angle, especially if it encourages external contributions. And I agree, this is all about clarity in communicating the effort we are able to expend (both in terms of human resources e.g dev/debug time and technical resources e.g CI).

docs/Compatibility.md
```diff
@@ -4,156 +4,260 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
...
 
-*Full* support indicates all tracer features are available.
+### Supported operating systems
 
-*Deprecated* indicates support will transition to *Maintenance* in a future release.
+| OS           | Support type                        | Package version |
+|--------------|-------------------------------------|-----------------|
+| Linux x86-64 | [GA](#support-ga)                   | Latest          |
+| Linux ARM64  | [GA](#support-ga)                   | Latest          |
+| macOS        | Dev environments only               | Latest          |
```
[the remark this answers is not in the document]
lloeki: > adding one column listing the list of platforms for each OS

Maybe this can be two columns: cpu arch and libc

Which makes me think the macOS "dev env only" in "support type" is a bit weird, maybe "environment" should be its own column so that this can read "GA".

Also "support type" and "package version" feel redundant, it seems like one implies the other (and conversely).

```
| OS      | CPU             | libc      | environment | Support type                        |
|---------|-----------------|-----------|-------------|-------------------------------------|
| Linux   | x86_64, aarch64 | gnu, musl | prod, dev   | [GA](#support-ga)                   |
| macOS   | x86_64, arm64   | darwin    | dev         | [GA](#support-ga)                   |
```

Thoughts?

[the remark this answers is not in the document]
ekump: > Both x86_64-darwin and arm64-darwin are supported.

The current 1.0 compatibility doc states 

```Use of ddtrace on macOS is supported for development, but not for production deployments.```

I just moved that statement to fit a tabular format. It is my understanding that while ddtrace may run, we don't officially support macOS in production. 

If that is not the case, I can update the doc. @delner is there more to the story about why we don't support macOS in production?

docs/Compatibility.md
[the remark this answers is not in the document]
ekump: I saw somewhere that `ARM64` was acceptable, but now I can't find it...so maybe it was a fever dream. Happy to change it to the official name.

docs/Compatibility.md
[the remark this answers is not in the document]
ekump: I'll create a ticket so we can track enhancing this table in the near future.

docs/Compatibility.md
ekump: updated so there is both `unsupported` and `unimplemented`.

PR #3486, opened 2024-02-26
Improve log correlation performance

TonyCTHsu:
Inspired by https://github.com/DataDog/dd-trace-rb/pull/3447

**What does this PR do?**

This is a performance improvement for correlation object. 

1. Creating correlation object from trace operation (skipping TraceDigest)
2. Improve `version`, `env` and `service` without duplicate in memory

### DIFF
--- a/lib/datadog/tracing/correlation.rb
+++ b/lib/datadog/tracing/correlation.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require_relative 'utils'
-require_relative 'metadata/ext'
 require_relative '../core/logging/ext'
 
 module Datadog
@@ -36,11 +35,11 @@ module Datadog
           version: nil
         )
           # Dup and freeze strings so they aren't modified by reference.
-          @env = Core::Utils::SafeDup.frozen_dup(env || Datadog.configuration.env)
-          @service = Core::Utils::SafeDup.frozen_dup(service || Datadog.configuration.service)
+          @env = env || Datadog.configuration.env
+          @service = service || Datadog.configuration.service
           @span_id = (span_id || 0).to_s
           @trace_id = trace_id || 0
-          @version = Core::Utils::SafeDup.frozen_dup(version || Datadog.configuration.version)
+          @version = version || Datadog.configuration.version
         end
 
         def to_h
--- a/gemfiles/ruby_2.5_rails4_mysql2.gemfile.lock
+++ b/gemfiles/ruby_2.5_rails4_mysql2.gemfile.lock
@@ -69,7 +69,7 @@ PATH
   specs:
     datadog (2.0.0.beta1)
       debase-ruby_core_source (= 3.3.1)
-      libdatadog (~> 6.0.0.2.0)
+      libdatadog (~> 7.0.0.1.0)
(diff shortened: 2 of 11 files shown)


delner:
Just trying to make sure I have a complete understanding of the change. I think the broad strokes are `TraceOperation --> TraceDigest --> Correlation::Identifier`.

 - `TraceDigest` is a full-copy of the the trace state basically, whose purpose is propagation or some other kind of snapshotting of the trace.
 - `Correlation::Identifier` is a much smaller struct, whose purpose is to be printed into logs, or consumed by other processes (aka profiling or ASM)

The new `to_correlation` function is basically what `identifier_from_digest` was minus the need for a `TraceDigest`. That makes sense, in that it skips a lot of the allocation involved in building an immutable `TraceDigest` object. I'm on board with this.

Still, some questions in my mind were:

1. **Do we lose data that would've been in the `TraceDigest` by skipping it?**
    - *Answer: I don't think so.* It looks like there's no place in code that has ever populated `service`, `env`, `version`, or `source` on a `Correlation::Identifier` explicitly. One could argue `identifier_from_digest` (from which this was copied) is coded incorrectly, and that it should've populated those values from the digest, not from `Datadog.configuration`. I think this is a bug, and probably should be corrected. But you could this is severable from the new `to_correlation`. But maybe `to_correlation` should be populating those fields from the `active_span` or the `TraceOperation` instead.
2. **Is building a `Correlation::Identifier` directly in the `Tracer` a breach of encapsulation? Should it have to go through a function analogous to `Correlation.identifier_from_digest`?**
    - *Answer: I don't think so.* `Identifier` is a simple struct with some helpers attached to it. I don't think it needs to be hidden behind a helper function.

My own read on the history of `Correlation` seems to suggest it started off as a really simplistic struct attached to `Context` way back in `0.19`, but has snowballed a bit picking up additional fields out of convenience. Then we (by that I mean I) naively decided to rebase it upon `TraceDigest` because it seemed like a nice, logical, clean thing to do at the time.

Bottom line: I think the change is fine. It's certainly nice to see a speed up in throughput like that. Consider addressing the issue outlined in (2) though.

lib/datadog/tracing/correlation.rb
```diff
@@ -21,26 +21,24 @@ class Identifier
...
-          env: nil,
-          service: nil,
-          span_id: nil,
-          trace_id: nil,
-          version: nil
-        )
-          # Dup and freeze strings so they aren't modified by reference.
-          @env = Core::Utils::SafeDup.frozen_dup(env || Datadog.configuration.env)
-          @service = Core::Utils::SafeDup.frozen_dup(service || Datadog.configuration.service)
+        def initialize(trace_id: nil, span_id: nil)
```
marcotc, line 27: How about keeping the constructor configurable, but not dupe configuration values to increase performance:
```ruby
@env = env ? Core::Utils::SafeDup.frozen_dup(env) : configuration.env
```

TonyCTHsu: I removed `Core::Utils::SafeDup.frozen_dup` completely

marcotc (approved):

PR #3488, opened 2024-02-26
w3c: Adds p datadog member to tracestate

mabdinur:
W3C Phase 2 support

The following operations will allow us to reconstruct traces that contain non datadog spans and use w3c tracecontext headers:
- Injects last seen datadog span id in tracestate headers
- Extracts last seen datadog span id and uses this value to set the `_dd.parent_id` distributed tracing tag

Tests: https://github.com/DataDog/system-tests/pull/2181

**2.0 Upgrade Guide notes**

**What does this PR do?**

**Motivation:**

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**


lib/datadog/tracing/distributed/trace_context.rb
```diff
@@ -282,7 +288,8 @@ def parse_sampled_flag(trace_flags)
           trace_flags & TRACE_FLAGS_SAMPLED
         end
 
-        # @return [Array<String,Integer,String,Hash>] returns 4 values: tracestate, sampling_priority, origin, tags.
+        # @return [Array<String,Integer,String,String,Hash>] returns 4 values:
```
marcotc, line 291:
```suggestion
        # @return [Array<String,Integer,String,String,Hash>] returns 5 values:
```
:)

lib/datadog/tracing/distributed/trace_context.rb
```diff
@@ -42,7 +42,9 @@ def extract(data)
 
           return unless trace_id # Could not parse traceparent
 
-          tracestate, sampling_priority, origin, tags, unknown_fields = extract_tracestate(fetcher[@tracestate_key])
+          tracestate, sampling_priority, origin, dd_parent_id, tags, unknown_fields = extract_tracestate(
```
marcotc, line 45:
```suggestion
          tracestate, sampling_priority, origin, parent_id, tags, unknown_fields = extract_tracestate(
```
I think it's OK to shorten it, without the `dd_` prefix, because everything in this file tends to be `dd` anyway.

marcotc:
It looks like the span_id value in the traceparent: https://github.com/DataDog/dd-trace-rb/blob/c9530f68267982d1c2918c714e10e6c03c8047c1/lib/datadog/tracing/distributed/trace_context.rb#L108
will always be equal to the value of the new tag: https://github.com/DataDog/dd-trace-rb/blob/c9530f68267982d1c2918c714e10e6c03c8047c1/lib/datadog/tracing/distributed/trace_context.rb#L150

Doesn't this mean the new tagged introduced in this PR is actually redundant, given the same `span_id` is already captured in the main traceparent?

In other words, when can the traceparent span_id and the new tag span_id diverge?

marcotc:
I think I found my answer.

It looks like when both values diverge, when extracting remote distributed headers, we should use the value of the tag instead of the value from the traceparent:
https://github.com/DataDog/system-tests/blob/f2c546afe09dce0ad7319c0b6108908c378638da/tests/parametric/test_headers_tracecontext.py#L780-L781

We should add a unit test to cover this specific behaviour in `spec/datadog/tracing/distributed/trace_context_spec.rb`.

lib/datadog/tracing/distributed/trace_context.rb
mabdinur: 😅

mabdinur:
> I think I found my answer.
> 
> It looks like when both values diverge, when extracting remote distributed headers, we should use the value of the tag instead of the value from the traceparent: https://github.com/DataDog/system-tests/blob/f2c546afe09dce0ad7319c0b6108908c378638da/tests/parametric/test_headers_tracecontext.py#L780-L781
> 
> We should add a unit test to cover this specific behaviour in `spec/datadog/tracing/distributed/trace_context_spec.rb`.

Ideally both values should never diverge. When we inject distributed tracing headers, traceparent and `x-datadog` headers SHOULD always be consistent (if they are not then this is a bug we need to fix). 

We are adding this new dd member to detect the following scenario: 
 - Host A: Injects tracecontext headers via ddtrace -> Host B: Injects/Extracts tracecontext headers via another vendor (ex: Dynatrace) -> Host C: Extracts tracecontext headers via  ddtrace

In the scenario above Datadog Trace Intake services receive spans from Host A and Host C (host B uses Dynatrace). Although Host A and Host C generate spans in the same trace the relationship between both trace chunks are lost. By updating the tracestate to also include the span_id then we can ensure third party vendors do not "break" trace propagation for Datadog spans. The span_id in the tracestate is extracted as a distributed tracing tag and sent to the backend/agent. This will allow Datadog UI to better visualize the trace.

This big trade off here is that we are adding 20 new characters to every tracestate header.

marcotc:
@mabdinur makes sense, thank you!

I think my comment is still correct and should be implemented:
> It looks like when both values diverge, when extracting remote distributed headers, we should use the value of the tag instead of the value from the traceparent: [DataDog/system-tests@f2c546a/tests/parametric/test_headers_tracecontext.py#L780-L781](https://github.com/DataDog/system-tests/blob/f2c546afe09dce0ad7319c0b6108908c378638da/tests/parametric/test_headers_tracecontext.py#L780-L781)

lib/datadog/tracing/distributed/trace_context.rb
```diff
@@ -54,6 +56,9 @@ def extract(data)
             end
           end
 
+          tags ||= {}
+          tags[Tracing::Metadata::Ext::Distributed::TAG_DD_PARENT_ID] = tp_parent_id || '0000000000000000'
+
```
bouwkast, line 61: This may be me not understanding Ruby code very well or the overall flow here, but won't this always be creating a `_dd.parent_id` tag and adding it the span below (`TraceDigest.new`)?

I can't really follow if this is won't do that if the `datadog` headers are present - I'd imagine that it is correct behavior as that is what the system tests assert, I just can't tell where that is here.

bouwkast (commented):
This looks good to me, I just have a question on when exactly this would be adding the `_dd.parent_id` tag, I couldn't tell for sure if it was always being added even if `datadog` headers would have been used.

lib/datadog/tracing/distributed/trace_context.rb
mabdinur: I believe this tag will be set if the `tracontext` headers are being and used in the local context. It does not take `datadog` headers in to account. Should we save this work for phase 3?

lib/datadog/tracing/distributed/trace_context.rb
```diff
@@ -142,6 +147,7 @@ def build_trace_flags(digest)
         # @see https://www.w3.org/TR/trace-context/#tracestate-header
         def build_tracestate(digest)
           tracestate = String.new('dd=')
+          tracestate << "p:#{format('%016x', digest.span_id)};"
```
mabdinur, line 150:
```suggestion
          unless digest.is_remote
            tracestate << "p:#{format('%016x', digest.span_id)};"
          else if digest.trace_distributed_tags.key?(Tracing::Metadata::Ext::Distributed::TAG_DD_PARENT_ID)
            tracestate << "p:#{digest.trace_distributed_tags[Tracing::Metadata::Ext::Distributed::TAG_DD_PARENT_ID]};"
          end
```

lib/datadog/tracing/distributed/trace_context.rb
```diff
@@ -142,6 +147,11 @@ def build_trace_flags(digest)
         # @see https://www.w3.org/TR/trace-context/#tracestate-header
         def build_tracestate(digest)
           tracestate = String.new('dd=')
+          unless digest.is_remote
```
marcotc, line 150: Following up from this PR https://github.com/DataDog/dd-trace-rb/pull/3516, it looks like `digest.is_remote` will almost always be `false`; it will only be `true` if we are injecting a recently extracted `TraceDigest`. All other `TraceDigest`s will have `digest.is_remote == false`, because they were created internally.

mabdinur: Yup that's the desired behavior. We expect the first case to be hit the vast majority of the time:
- If a TraceDigest is created from an internal span then TraceDigest.span_id is the last datadog span id in the trace
- If a TraceDigest represents a span from a remote host then the TraceDigest.span_id may not be the last datadog span id in the trace. In this scenario we must check the incoming for the last datadog span id (`p:.......` field).
- If the digest represents a remote span and the last datadog parent id is not propagated in the incoming tracestate then do nothing. We can not set the last datadog span id, this information has been lost.

mabdinur: I just added tests for this logic

marcotc (approved):

PR #3490, opened 2024-02-26
Rename library to datadog

delner:
Rename the library from `ddtrace` to `datadog` to reflect the inclusive nature of many Datadog tools (tracing, profiling, ASM, etc).

**2.0 Upgrade Guide notes**

Users should use `datadog` instead of `ddtrace` in their requires, where applicable e.g.:

 - `gem 'ddtrace'` --> `gem 'datadog'`
 - `require 'ddtrace'` --> `require 'datadog'`
 - `DDTrace::VERSION` --> `Datadog::VERSION`

It otherwise preserves all the same functionality & default behavior present in `ddtrace`.

**What does this PR do?**

A 1-1 name change of the `ddtrace` gem to `datadog`, updating relevant file names, constants, comments, and other references to the Ruby gem.

**Motivation:**

`ddtrace` as a name has not been reflective of the many products we include within the package, and the name was in conflict with the semantic conventions of Ruby (where gem name & file paths should mirror constants.)

**Additional Notes:**

TBD

**How to test the change?**

 TBD


delner:
**Notes about development**

There are quite a few changes here, and I think its worth segmenting them into what's a minimum, and what's additional.

**Minimum**

- [X] Change gem from `ddtrace` to `datadog`
- [X] Migrate remaining `ddtrace`.rb` and `ddtrace/` files to `datadog` files, update `require`s.
- [X] Update references to `DDTrace` to `Datadog`
- [X] Update documentation to reflect new behavior.
- [X] Update builds, tools, tests, and other automated processes.
- [X] Update code comments (for less confusion)

**Possible changes**

- [X] Update the profiling extension artifacts to reflect `datadog_profiling_native_extension` instead of `ddtrace_profiling_native_extension`.
- [X] Update the `ddtracerb` bin to `ddprofrb`

(I've applied the above two changes in their own commit, just to trial them. We can back these out if undesirable.)

- [ ] Update the profiling C code to reflect `datadog` instead of `ddtrace`

**Open questions**

1. Should we rename `Datadog::Core::Environment::Ext::TRACER_VERSION` (which was an alias to `DDTrace::VERSION::STRING`) to `GEM_VERSION` instead? Seems more accurate, inclusive of what the constant represents.
2. What should the behavior of `datadog/auto_instrument` be? Currently activates profiling and tracing automatically. Should this file be the "auto activate whatever is in the lib" trigger? (a.k.a more inclusive of just tracing/profiling?)
3. Should we rename the `ddtracerb` bin? This file's sole purpose is to codelessly activate profiling, when used to wrap a Ruby process from shell. Should to be renamed to `ddprofrb` to reflect its actual behavior?
4. Should we rename the Faraday middleware named `ddtrace` to `datadog`? This middleware is trace specific. If renamed to `datadog`, and CI or ASM also want to add their own middleware, its name is ambiguous.
5. There are a bunch of comments that were updated in transport code that suggest changing `require_relative` paths to `ddtrace`: paths which no longer exist. Not sure how to update these appropriately. (cc @ekump )
6. Should profiling artifacts/constants be updated to reflect `datadog` instead of `ddtrace`? There are a lot of direct references to the package name in the C code. I think this is a straight find-replace operation. (cc @AlexJF @ivoanjo )

ekump (commented):
Looks fine at first glance for non-profiling changes.

integration/apps/rack/Gemfile
```diff
@@ -23,10 +23,10 @@ gem 'resque'
...
 
 gem 'dogstatsd-ruby'
-# Choose correct specs for 'ddtrace' demo environment
-gem 'ddtrace', *Datadog::DemoEnv.gem_spec('ddtrace')
+# Choose correct specs for 'datadog' demo environment
+gem 'datadog', *Datadog::DemoEnv.gem_spec('datadog')
 
-# Needed for ddtrace profiling
+# Needed for datadog profiling
 google_protobuf_versions = [
```
ivoanjo, line 30: Minor: Actually no longer needed ;)

integration/apps/rspec/Dockerfile-ci
```diff
@@ -6,6 +6,6 @@ FROM ${BASE_IMAGE}
 COPY . /vendor/dd-trace-rb
 
 # Install dependencies
-# Setup specific version of ddtrace, if specified.
-ENV DD_DEMO_ENV_GEM_LOCAL_DDTRACE /vendor/dd-trace-rb
+# Setup specific version of datadog, if specified.
+ENV DD_DEMO_ENV_GEM_LOCAL_DATADOG /vendor/dd-trace-rb
```
ivoanjo, line 10: I guess these `dd-trace-rb` paths should be updated too

lib/datadog/version.rb
```diff
@@ -1,13 +1,12 @@
...
 
-module DDTrace
+module Datadog
   module VERSION
     MAJOR = 1
     MINOR = 20
     PATCH = 0
     PRE = nil
     BUILD = nil
-    # PRE and BUILD above are modified for dev gems during gem build GHA workflow
```
ivoanjo, line 10: Minor: I think this comment is helpful to keep around, as otherwise it doesn't look like `PRE` or `BUILD` are very useful.

ivoanjo (commented):
:+1: This seems reasonable to me!

My two main concerns are:

a. Smooth `ddprofrb` migration path
b. Changes to profiling stuff makes it hard to move stuff between master and 2.0 branch

And actually I think we can solve both a. and b. by moving most of the profiling changes to the master branch (e.g. as a separate PR).

Starting with a., we could introduce a `bin/ddprofrb`, side-by-side with the old `bin/ddtracerb`. The old `ddtracerb` would print a deprecation warning when getting used, and so customers on the latest 1.x would be able to move to the new name even before moving to 2.0, and would not be "surprised" when it's gone in 2.0. Then in the 2.0 branch we can remove it.

In the case of b., almost all of the changes to profiling (name of the extensions, `build_ddtrace_profiling_native_extension` script, etc) are internal things that are not part of any public API, so by making those changes in the master branch we'd simplify back/forward porting AND also make this PR a lot simpler as well.

Thoughts?

integration/apps/rack/Gemfile
delner: As in delete these arrays? Can you make a code suggestion? I can apply it once I see the pattern of it.

delner:
@ivoanjo I'm onboard with your plan. I think we were hoping to do another 1.x release to add deprecation warnings anyways, so introducing the new `ddprofrb` there would fit well. Then we can open a separate PR that removes `ddtracerb` on the 2.0 branch.

If the other names are internal, that's fine. Just wanted to make sure any artifacts/APIs produced by these profiling files wouldn't be considered breaking if their names/constants changed. When would you like to change these names?

integration/apps/rack/Gemfile
ivoanjo: We can clean it all, the dependency is no longer needed:

```diff
-# Needed for ddtrace profiling
-google_protobuf_versions = [
-  '~> 3.0',
-  '!= 3.7.0.rc.2',
-  '!= 3.7.0.rc.3',
-  '!= 3.7.0',
-  '!= 3.7.1',
-  '!= 3.8.0.rc.1',
-  '!= 3.24.0.rc.1',
-]
-if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.4')
-  gem 'google-protobuf', *google_protobuf_versions
-else
-  # Bundler resolves incorrect version (too new, incompatible with Ruby <= 2.3)
-  gem 'google-protobuf', *google_protobuf_versions, '< 3.19.2'
-end
-
```

integration/apps/rspec/Dockerfile-ci
delner: Technically these reflect the repo name; we can change these now (say to `datadog-rb`) if we want, but it would be out of step with the current repo name. I recommend holding off and doing repo renaming changes separately.

delner:
Rebased this branch on the latest 2.0 and added one more name change of `TRACER_VERSION` to `GEM_DATADOG_VERSION`. All changes should be complete, just need a passing build now.

Handing off to @TonyCTHsu and @marcotc for any relevant system tests, benchmarks, etc...

urseberry (approved):
Documentation updates look good.

marcotc (approved):

PR #3494, opened 2024-02-28
Fix unstable benchmarks and comparison with 1.x

marcotc:
**What does this PR do?**

This PR documents why the benchmark that includes the writer yields unstable results.

This PR also fixes test comparisons with `master` branch running `1.x`. This is only needed because the benchmarks are run against both `1.X` and `2.X`. If benchmarks already existed `1.x` this wouldn't be an issue, but because they were recently introduced, they do not exist in master branch. We can revert this fix when `2.x` is merged to `master` but we need minor changes to ensure immediately compatibility.

**Motivation:**

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**

### DIFF
--- a/benchmarks/tracing_trace.rb
+++ b/benchmarks/tracing_trace.rb
@@ -23,12 +23,15 @@ class TracingTraceBenchmark
     end
   end
 
+  def benchmark_time(time: 10.5, warmup: 2)
+    VALIDATE_BENCHMARK_MODE ? { time: 0.001, warmup: 0 } : { time: time, warmup: warmup }
+  end
+
   def benchmark_no_writer
     ::Datadog::Tracing::Writer.prepend(NoopWriter)
 
     Benchmark.ips do |x|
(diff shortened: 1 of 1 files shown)


benchmarks/tracing_trace.rb
```diff
@@ -23,15 +23,19 @@ def open
     end
   end
 
-  def benchmark_time(time: 10.5, warmup: 2)
+  # @param [Integer] time in seconds. The default is 12 seconds because having over 105 samples allows the
+  #   benchmarking platform to calculate helpful aggregate stats. Because benchmark-ips tries to run one iteration
+  #   pre 100 ms, this means we'll have around 120 samples (give or take a small margin of error).
```
ekump, line 28: should that be `per 100ms`?

ekump (approved):

benchmarks/tracing_trace.rb
marcotc: Yes! Thank you!

PR #3497, opened 2024-02-29
Revert "peer.service adjustment for sql propagation with DBM"

ynnadkrap:
Reverts DataDog/dd-trace-rb#3127

`peer.service` is not a meaningful attribute for DBM as the data model is evolving to rely on `peer.db.name` and `peer.hostname` instead. The original PR is causing issues that we can avoid by reverting back to the old behavior.


ivoanjo:
Update: There's still some ongoing discussion on this change, so tagging with do-not-merge.

PR #3500, opened 2024-03-01
Remove SpanOperation.new(child_of)

marcotc:
**2.0 Upgrade Guide notes**

Removed argument `child_of` from `SpanOperation.new`.

**What does this PR do?**

This PR removes the unused argument `child_of` from `SpanOperation.new`.

**Motivation:**

This argument is only used for testing.

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**

### DIFF
--- a/lib/datadog/tracing/span_operation.rb
+++ b/lib/datadog/tracing/span_operation.rb
@@ -41,7 +41,6 @@ module Datadog
 
       def initialize(
         name,
-        child_of: nil,
         events: nil,
         on_error: nil,
         parent_id: 0,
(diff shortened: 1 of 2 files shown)


lib/datadog/tracing/span_operation.rb
```diff
@@ -83,12 +82,6 @@ def initialize(
         # Set tags if provided.
         set_tags(tags) if tags
 
-        # Only set parent if explicitly provided.
-        # We don't want it to override context-derived
-        # IDs if it's a distributed trace w/o a parent span.
-        parent = child_of
-        self.parent = parent if parent
```
marcotc, line 90: I tried to completely remove `#parent` as well, but it's still used for `service_entry` calculation.

ekump (approved):

delner:
We talked about this a little more, and I think I may have misunderstood things. There's still a `parent_id`; `child_of` was a parallel argument that was more cumbersome to use. I'm fine with this as is, and worst case, we can reintroduce if needed.

PR #3502, opened 2024-03-01
Rename Profiling files to reflect "datadog" instead of "ddtrace"

delner:
**What does this PR do?**

Changes the names/filepaths of the profiling code to reflect "datadog" instead of "ddtrace". Should not functionally change anything, or contain any breaking changes.

**Motivation:**

Per @ivoanjo 's suggestion that we extract the profiling rename from a larger rename PR. This change seems isolated and internal, which should make it a good candidate for its own PR.

**Additional Notes:**

This is an early step towards a larger rename to `datadog` in 2.0. There may be more changes to the C code (constant names) or other references to `ddtrace`. Those changes will happen in another PR.


ivoanjo (approved):
:+1: LGTM

PR #3504, opened 2024-03-04
Remove private `build_span_events`

TonyCTHsu:
**What does this PR do?**

Streamline the method calling by removing private `build_span_events`

**Motivation:**

This private method `build_span_events` is only invoked without any arguments

### DIFF
--- a/lib/datadog/tracing/tracer.rb
+++ b/lib/datadog/tracing/tracer.rb
@@ -376,23 +376,36 @@ module Datadog
       )
         trace = _trace || start_trace(continue_from: continue_from)
 
-        options = {
-          events: SpanOperation::Events.new,
-          on_error: on_error,
-          resource: resource,
-          service: service,
-          tags: resolve_tags(tags),
-          type: type,
-        }
+        events = SpanOperation::Events.new
 
         if block
           # Ignore start time if a block has been given
-          trace.measure(name, **options, &block)
+          trace.measure(
+            name,
+            events: events,
+            on_error: on_error,
+            resource: resource,
+            service: service,
+            tags: resolve_tags(tags),
+            type: type,
+            &block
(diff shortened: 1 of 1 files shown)


delner (commented):
What's the reason for this?

I can see `build_span_events` as redundant, but this does more than that. It eliminates the use of `resolve_tags` and options for `**options` over named arguments which are known to have high performance cost.

TonyCTHsu:
1. `resolve_tags` is not eliminated, https://github.com/DataDog/dd-trace-rb/pull/3504/files#diff-fb264450c82590f2bf44f09afc63b6a01bd80e4916333f7c19441240694412a1R384

2. I can remove the double splat usage for `option`.

lib/datadog/tracing/tracer.rb
```diff
@@ -376,48 +376,27 @@ def start_span(
...
-            start_time: start_time,
-            tags: resolve_tags(tags),
-            type: type
-          )
-
-          span.start(start_time)
-          span
+          trace.build_span(name, start_time: start_time, **options).tap do |span|
+            span.start(start_time)
+          end
```
marcotc, line 395: Using `**kwargs` it's slower than passing all the arguments.

Also, using `tap` is slower than calling the method directly, then returning the original value.

marcotc: .I recommend passing the values by name, similar to how they were before, and not using `tap`.

marcotc: Don't take my word for it, our microbenchmarks should tell us the real numbers: `benchmarks/tracing_trace.rb`

marcotc (approved):
No externally facing changes. 👍

delner (approved):

PR #3510, opened 2024-03-06
[PROF-9263] Add experimental support for profiling code hotspots when used with opentelemetry ruby gem

ivoanjo:
**What does this PR do?**

This PR adds experimental support for getting profiling code hotspots data (including endpoint profiling) when profiling processes being traced using the opentelemetry ruby gem **_directly_**.

Note that this differs from the recommended way of using opentelemetry with the ddtrace library, which is to follow the instructions from https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/otel_instrumentation/ruby/ .

The key difference is -- this PR makes code hotspots work even for setups that opt to not use `require 'datadog/opentelemetry'` (which is the recommended and easier way).

The approach taken here is similar to #2342 and #3466: we peek inside the implementation of the opentelemetry gem to extract the information we need (namely the span id, local root span id, trace type, and trace endpoint). This approach is potentially brittle, which is why the code is written very defensively, with the aim of never breaking the application (or profiling) if something is off -- it just won't collect code hotspots.

**Motivation:**

We have a customer interested in running this setup, so hopefully they'll be able to test this PR and validate if it works for them.

Furthermore, I'm hoping to see if the opentelemetry Ruby folks would be open to tweaking their APIs to be more friendlier to tools such as the profiler, but for now I opted for getting our hands dirt.

**Additional Notes:**

I'm opening this PR as draft until we can get feedback from the customer and see if this works for them.

**How to test the change?**

On top of the added test coverage, I was able to see code hotspots working for the following sinatra example app:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'rackup'
  gem 'dogstatsd-ruby'
  gem 'datadog', git: 'https://github.com/datadog/dd-trace-rb', branch: 'ivoanjo/prof-9263-otlp-ruby-code-hotspots'
  gem 'sinatra'
  gem 'opentelemetry-api'
  gem 'opentelemetry-sdk'
  gem 'opentelemetry-instrumentation-sinatra'
  gem 'opentelemetry-exporter-otlp'
  gem 'pry'
end

require 'sinatra/base'
require 'opentelemetry/sdk'
require 'pry'

Datadog.configure do |c|
  c.service = 'ivoanjo-testing-opentelemetry-test'
  c.profiling.enabled = true
end

# Configure OpenTelemetry
OpenTelemetry::SDK.configure do |c|
  c.service_name = 'ivoanjo-testing-opentelemetry-test'
  c.use 'OpenTelemetry::Instrumentation::Sinatra'
end

class MyApp < Sinatra::Base
  get '/' do
    OpenTelemetry::Trace.current_span.add_attributes({'runtime-id' => Datadog::Core::Environment::Identity.id})
    sleep 1
    'Hello, OpenTelemetry!'
  end
end

MyApp.run!
```

After doing a few requests, here's how this looks:

> image

---

> image


ext/datadog_profiling_native_extension/collectors_thread_context.c
```diff
@@ -734,6 +765,11 @@ static void trigger_sample_for_thread(
   struct trace_identifiers trace_identifiers_result = {.valid = false, .trace_endpoint = Qnil};
   trace_identifiers_for(state, thread, &trace_identifiers_result);
 
+  if (!trace_identifiers_result.valid) {
```
AlexJF, line 768: Worth/possible doing a bit of extra work at the start to arrive at a sticky decision here and short-circuit constantly failing `trace_identifiers_for` with all its rb_var_get if ddtrace is not used for tracing at all? 

Or thinking is that we want to support situations where there's a mix of ddtrace and pure-ot traces or the ability to change between one and the other dynamically (e.g. via a FF)?

AlexJF (commented):
LGTM! And so safe you should be designing baby furniture 😄

ext/datadog_profiling_native_extension/collectors_thread_context.c
ivoanjo: I think your suggestion makes sense.

My intent here in checking both is that the profiler may start quite early in the app lifecycle, so we may not know which one is going to be used yet.

> Or thinking is that we want to support situations where there's a mix of ddtrace and pure-ot traces and/or the ability to change between one and the other dynamically (e.g. via a feature flag)?

I'm not sure mixing is even possible at this point, since the ddtrace otel support monkey patches itself pretty deep into opentelemetry (which is why I needed to contort a bit to be able to test both).

For that reason, and after our last discussion, I think it makes sense to stop checking opentelemetry once we see data coming from ddtrace traces.

The reverse is harder to figure out, actually. It would be weird, but not impossible, for an app that started with opentelemetry to then switch over to ddtrace.

---

TL;DR: I'll wait for feedback from our customer on how this is working before acting on this comment, just in case we end up going in a completely different direction BUT I'll definitely come back to it before marking the PR as non-draft.

ivoanjo:
We're waiting for a bit more feedback on if this is the right approach before going forward. If there's not a lot of movement in a few months we can close this PR.

ivoanjo:
Update: I've rebased this PR on top of the v2.3.0 stable release to help testing.

ivoanjo:
This PR is going to be superceded by https://github.com/DataDog/dd-trace-rb/pull/3984 . I'm going to leave it open for a while longer to give customers testing this feature time to migrate to a release containing the newer version of this PR.

ivoanjo:
Erghh our automation was a bit too automated -- I've restored the branch, and removed this from the 2.4.0 milestone (since it was https://github.com/DataDog/dd-trace-rb/pull/3984 that got merged).

PR #3515, opened 2024-03-08
Patch Sinatra only once

lloeki:
**What does this PR do?**

Sinatra needs to be patched once only.

**Motivation:**

Sinatra sets things up at first-request time, which may cause concurrent requests to call `configure` and break RC (thread leak).

**Additional Notes:**

The integration [hooks](https://github.com/DataDog/dd-trace-rb/blob/9e4a267ce0d93c3bd4603b02b3e39d40ae7c937b/lib/datadog/tracing/contrib/sinatra/patcher.rb#L68-L70) onto [`Sinatra::Base#setup_middleware`](https://github.com/DataDog/dd-trace-rb/blob/9e4a267ce0d93c3bd4603b02b3e39d40ae7c937b/lib/datadog/tracing/contrib/sinatra/patcher.rb#L26-L29), which then (much later, as `Sinatra::Base#setup_middleware` appears to be invoked at first request) calls our [`Contrib::Sinatra::Framework.setup`](https://github.com/DataDog/dd-trace-rb/blob/9e4a267ce0d93c3bd4603b02b3e39d40ae7c937b/lib/datadog/tracing/contrib/sinatra/framework.rb#L11-L16), which calls `Datadog.configure` and [`activate_rack!`](https://github.com/DataDog/dd-trace-rb/blob/9e4a267ce0d93c3bd4603b02b3e39d40ae7c937b/lib/datadog/tracing/contrib/sinatra/framework.rb#L23-L29) which invokes `config.tracing.instrument :rack`, which ultimately inserts the Rack middleware.

The trouble appears to be that this `Datadog.configure` call is not concurrency-protected enough, which leads to RC workers being started then mid-way the RC component being rebuilt with inconsistent access through `active_remote`.

**How to test the change?**

- CI
- test in a Sinatra app

**For Datadog employees:**

### DIFF
--- a/lib/datadog/tracing/contrib/configurable.rb
+++ b/lib/datadog/tracing/contrib/configurable.rb
@@ -41,7 +41,7 @@ module Datadog
 
           # Create or update configuration associated with `matcher` with
           # the provided `options` and `&block`.
-          def configure(matcher = :default, options = {}, &block)
+          def configure(matcher = :default, options = {}, tracing = nil, &block)
             config = if matcher == :default
                        default_configuration
                      else
@@ -51,6 +51,12 @@ module Datadog
 
             # Apply the settings
             config.configure(options, &block)
+
+            # Configure dependents
+            if tracing && respond_to?(:subconfigure)
+              subconfigure(tracing, config)
+            end
+
             config
           end
 
--- a/lib/datadog/tracing/contrib/extensions.rb
+++ b/lib/datadog/tracing/contrib/extensions.rb
@@ -63,6 +63,10 @@ module Datadog
             # Reconfigure core settings
             super(&block)
 
+            # Configuration should be considered RO past this point:
+            # - Patcher's `patch` must not modify configuration.
+            # - use Integration's `subconfigure` instead
+
             # Activate integrations
             configuration = self.configuration.tracing
 
@@ -158,7 +162,7 @@ module Datadog
               unless integration.nil? || !integration.default_configuration.enabled
                 configuration_name = options[:describes] || :default
                 filtered_options = options.reject { |k, _v| k == :describes }
-                integration.configure(configuration_name, filtered_options, &block)
+                integration.configure(configuration_name, filtered_options, self, &block)
                 instrumented_integrations[integration_name] = integration
 
                 # Add to activation list
--- a/lib/datadog/tracing/contrib/sinatra/integration.rb
+++ b/lib/datadog/tracing/contrib/sinatra/integration.rb
@@ -31,6 +31,14 @@ module Datadog
             Configuration::Settings.new
           end
 
+          def subconfigure(tracing, sinatra)
+            tracing.instrument(
+              :rack,
+              service_name: sinatra.service_name,
+              distributed_tracing: sinatra.distributed_tracing,
+            )
+          end
+
           def patcher
             Patcher
           end
--- a/lib/datadog/tracing/contrib/sinatra/patcher.rb
+++ b/lib/datadog/tracing/contrib/sinatra/patcher.rb
@@ -9,19 +9,6 @@ module Datadog
   module Tracing
     module Contrib
       module Sinatra
-        # Set tracer configuration at a late enough time
-        module TracerSetupPatch
-          ONLY_ONCE_PER_APP = Hash.new { |h, key| h[key] = Core::Utils::OnlyOnce.new }
-
-          def setup_middleware(*args, &block)
-            super.tap do
-              ONLY_ONCE_PER_APP[self].run do
-                Contrib::Sinatra::Framework.setup
-              end
-            end
-          end
-        end
-
         # Hook into builder before the middleware list gets frozen
         module DefaultMiddlewarePatch
           ONLY_ONCE_PER_APP = Hash.new { |h, key| h[key] = Core::Utils::OnlyOnce.new }
@@ -42,8 +29,6 @@ module Datadog
         module Patcher
           include Contrib::Patcher
 
-          PATCH_ONLY_ONCE = Core::Utils::OnlyOnce.new
-
           module_function
 
           def target_version
@@ -51,13 +36,10 @@ module Datadog
           end
 
           def patch
-            PATCH_ONLY_ONCE.run do
-              require_relative 'tracer'
-              register_tracer
+            require_relative 'tracer'
+            register_tracer
 
-              patch_default_middlewares
-              setup_tracer
-            end
+            patch_default_middlewares
           end
 
           def register_tracer
--- a/lib/datadog/tracing/contrib/sinatra/framework.rb
+++ b/lib/datadog/tracing/contrib/sinatra/framework.rb
@@ -7,27 +7,6 @@ module Datadog
         # - handle configuration entries which are specific to Datadog tracing
         # - instrument parts of the framework when needed
         module Framework
-          # Configure Rack from Sinatra, but only if Rack has not been configured manually beforehand
-          def self.setup
-            Datadog.configure do |datadog_config|
-              sinatra_config = config_with_defaults(datadog_config)
-              activate_rack!(datadog_config, sinatra_config)
-            end
-          end
-
-          def self.config_with_defaults(datadog_config)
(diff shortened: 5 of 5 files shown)


lloeki:
There are two things happening here that should be split:

- setting the configuration settings for Sinatra+Rack:
  - saying that this integration is enabled
  - setting any passed options, or setting to default values
- inserting our middlewares into the Rack stack, which can only be done when there is a Rack stack (which Sinatra appears to do lazily, hence the hook on `setup_middlewares` and the cause of this happening when requests start to be received)

That would go a long way towards this class of issue disappearing.

lib/datadog/tracing/contrib/sinatra/patcher.rb
```diff
@@ -42,18 +42,22 @@ def setup_middleware(*args, &block)
...
+            PATCH_ONLY_ONCE.run do
+              require_relative 'tracer'
+              register_tracer
 
-            patch_default_middlewares
-            setup_tracer
+              patch_default_middlewares
+              setup_tracer
+            end
           end
```
ivoanjo, line 61: Hmm... I actually think this is a bit of a no-op, compared to what we're doing now?

E.g. since we're still monkey patching the sinatra classes with the `TracerSetupPatch`, we'll still be calling `Datadog.configure` multiple times for each sinatra app.

I don't quite see how this moves the needle from where we are right now :thinking:

lib/datadog/tracing/contrib/extensions.rb
```diff
@@ -63,6 +63,10 @@ def configure(&block)
             # Reconfigure core settings
             super(&block)
 
+            # Configuration should be considered RO past this point:
+            # - Patcher's `patch` must not modify configuration.
+            # - use Integration's `subconfigure` instead
+
```
lloeki, line 69: I was wondering about calling `subconfigure` here instead:

- gives access to full config, not just tracing's, which would allow cross-product integration subconfiguration (e.g appsec could enable tracing)
- config is final state after whole `configure`'s `&block` is executed vs at the time the `instrument :sinatra` line is

lib/datadog/tracing/contrib/sinatra/integration.rb
```diff
@@ -31,6 +31,14 @@ def new_configuration
             Configuration::Settings.new
           end
 
+          def subconfigure(tracing, sinatra)
```
lloeki, line 34: Lousy name, `configure` is already used.

I considered merely `def configure` and `super`ing but that requires giving arguments here that are sorta internal WRT this `Integration` DSL-ish descriptive class. So I preferred not leaking internal details.

lib/datadog/tracing/contrib/sinatra/integration.rb
```diff
@@ -31,6 +31,14 @@ def new_configuration
...
+            tracing.instrument(
+              :rack,
+              service_name: sinatra.service_name,
+              distributed_tracing: sinatra.distributed_tracing,
+            )
```
lloeki, line 39: Most probably:
- other integrations are affected by this: Rails, Grape, etc...
- a similar pattern should be used by e.g `AppSec` integrations.

lib/datadog/tracing/contrib/sinatra/patcher.rb
```diff
@@ -53,7 +40,6 @@ def patch
             register_tracer
 
             patch_default_middlewares
-            setup_tracer
           end
 
           def register_tracer
```
ivoanjo, line 45: :arrow_down: Don't forget to remove `setup_tracer`, below

lib/datadog/tracing/contrib/sinatra/integration.rb
ivoanjo: Maybe `instrument_dependencies` or `instrument_components`? It looks like in most cases this is what we end up doing here.

lib/datadog/tracing/contrib/sinatra/integration.rb
```diff
@@ -31,6 +31,14 @@ def new_configuration
             Configuration::Settings.new
           end
 
+          def subconfigure(tracing, sinatra)
+            tracing.instrument(
+              :rack,
+              service_name: sinatra.service_name,
```
ivoanjo, line 37: I wonder if we should deprecate setting the rack `service_name` from the sinatra service name for 2.0?

E.g. instead suggest customers configure a top-level service name instead, or if they really want to, they can configure a service name in rack directly.

lib/datadog/tracing/contrib/configurable.rb
```diff
@@ -41,7 +41,7 @@ def configurations
 
           # Create or update configuration associated with `matcher` with
           # the provided `options` and `&block`.
-          def configure(matcher = :default, options = {}, &block)
+          def configure(matcher = :default, options = {}, tracing = nil, &block)
```
ivoanjo, line 44: Should this be `tracing_configuration` or `tracing_instrumentation` instead maybe?

ivoanjo (commented):
This seems quite reasonable!

I'll admit that I'm less confident about some of the specific details of the changes to how we configure integrations.

Having said that, I think it's a good insight that we have these ad-hoc hooks for "oh I'm setting up sinatra, and now need to setup rack; or I'm setting up rails and now need to setup active_...whatever" and making them part of the API avoids integrations needing to solve the problem on their own.

Thoughts on next steps? Perhaps pull in the tracing folks to make a decision on moving forward (or not) with this change?

lib/datadog/tracing/contrib/sinatra/integration.rb
lloeki: It's not so much about passing a manually configured service name as much as when that `sinatra.service_name` is defaulted and inferred e.g from the app class or something.

Same for Rails where it's the application name as a default.

Tangentially there may be a conundrum to resolve here in such default cases: ordering may matter when there's both Sinatra and Rails in the same app, as both would set their service name and one would win. But that's a problem for another day.

lib/datadog/tracing/contrib/sinatra/integration.rb
lloeki: Hmm.

Indeed in this case we do `c.instrument`, but that may not always be the case, it could also do `c.some_setting = :foo` in which case `instrument_*` is misleading.

Technically this is really a callback within the `Datadog.configure` block, so that we don't nest configure anymore. In that case I believe the name should include `configure` in some way for clarity. Since for `Integration`s there's a `def configure` that is called before, so maybe `def after_configure`? That would give clarity that i.e that it happens after `Integration#configure` has been called.

Functionally this is about setting up dependencies, so it may go beyond just `c.something = :foo` or `c.instrument`. So `def dependencies` or `def configure_dependencies` might cut it.

e.g I'll need to apply similar changes to `AppSec` so it makes me think that I'm yet unsure how that would unfold, therefore I'm trying to keep an open-ended approach where integrations could have dependencies across products. e.g `AppSec` or `CI` may decide that they'd want to enable tracing and/or enable specific tracing integrations when tracing is enabled to ensure Things Just Work.

lib/datadog/tracing/contrib/sinatra/patcher.rb
lloeki: Good catch, removed.

lib/datadog/tracing/contrib/configurable.rb
lloeki: Yeah it's a bit ambiguous. Since this is a `Configuration` object and the other name here is `config` I went with `tracing_config`.

lib/datadog/tracing/contrib/sinatra/integration.rb
ivoanjo: Good points. I like the `after_configure`, seems like it's a reasonable description for what's happening.

ivoanjo:
Hey @lloeki any thoughts on getting this in?

ivoanjo:
This has been stale for a long time, let's close it -- we can always revive if we decide to revive this effort. I do think this is very valuable -- the way we rebuild the whole library for an integration change is a leftover of simpler times that does not fit with the current scope of the library.

PR #3519, opened 2024-03-12
[PROF-9136] Fix rare remote configuration worker thread leak

ivoanjo:
**What does this PR do?**

This PR fixes the remote configuration worker thread leak that is discussed in more detail in the internal
"dd-trace-rb Remote::Worker thread leak investigation" doc.

In a nutshell, there's a potential race between threads that call `Datadog::Core::Remote::Tie.boot` and threads that call `Datadog.configure` that can, in really unlucky situations, mean that the `Remote::Worker thread gets leaked during library reconfiguration.

This rare issue is O(number of times Datadog.configure gets called) while requests are being processed in a web-app.

I was only able to reproduce it with the weird reproducer below, and by adding `sleep` to some parts of our code, so while we have observed it in production, we expect it to be rare.

Being more specific, this issue is related to a race between the remote configuration thread being shut down, and other parts of the code base trying to restart it.

In practice, the remote configuration would previously always get started by calls to `Datadog::Core::Remote.active_remote.barrier(:once)`. During dd-trace-rb reconfiguration, it’s possible for the remote configuration thread to be shut down, and before the new configuration kicks in, for any other thread to call
`Datadog::Core::Remote.active_remote.barrier(:once)`, restarting the remote configuration thread that belonged to the old configuration.

This PR fixes this issue introducing a `@stop_requested` variable that flags a "terminal state" for the remote configuration worker.

Thus, the default behavior of `Remote::Worker.start` is changed from always starting the thread to not starting the thread if the same worker was stopped before.

**Motivation:**

Fix issue.

**Additional Notes:**

We are also working to remove cases where `Datadog.configure` gets called during request handling, e.g. #3515 .

**How to test the change?**

I've added unit test coverage for the worker not restarting itself again.

To validate that this fixes the worker thread leak, the following reproducer can be used:

```diff
diff --git a/lib/datadog/core/configuration.rb b/lib/datadog/core/configuration.rb
index 7a20444784..fe0ca8ba0f 100644
--- a/lib/datadog/core/configuration.rb
+++ b/lib/datadog/core/configuration.rb
@@ -254,6 +254,12 @@ module Datadog
         components = Components.new(settings)

         old.shutdown!(components)
+        puts "After shutdown, sleeping!"
+
+        sleep 1
+
+        puts "Woke up after shutdown!"
+
         components.startup!(settings)
         components
       end
diff --git a/remote-worker-thread-leak.rb b/remote-worker-thread-leak.rb
new file mode 100644
index 0000000000..3552c654b5
--- /dev/null
+++ b/remote-worker-thread-leak.rb
@@ -0,0 +1,22 @@
+require 'ddtrace'
+
+configure_thread = Thread.new do
+  15.times {
+    Datadog.configure { Thread.pass }
+    Thread.pass
+  }
+end
+
+trigger_rc_thread = Thread.new do
+  loop {
+    sleep 0.5
+    Datadog::Core::Remote.active_remote.barrier(:once)
+    Thread.pass
+  }
+end
+
+configure_thread.join
+trigger_rc_thread.kill
+trigger_rc_thread.join
+
+puts Thread.list
```

Before the fix, the list of threads printed would show a lot of leaked remote worker configuration threads. With the fix, the threads no longer leak, and if you enable debug logging you will see that the restart is being prevented and the debug log I added will show up

```
D, [2024-03-12T16:27:02.951950 #742185] DEBUG -- ddtrace: [ddtrace]
(dd-trace-rb/lib/datadog/core/remote/worker.rb:28:in `start')
remote worker: refusing to restart after previous stop
```

This debug message is harmless and shows the fix is working fine.

**For Datadog employees:**

### DIFF
--- a/lib/datadog/core/remote/worker.rb
+++ b/lib/datadog/core/remote/worker.rb
@@ -11,7 +11,7 @@ module Datadog
 
           @starting = false
           @started = false
-          @stop_requested = false
+          @stopped = false
 
           @interval = interval
           raise ArgumentError, 'can not initialize a worker without a block' unless block
@@ -24,7 +24,7 @@ module Datadog
 
           acquire_lock
 
-          if @stop_requested
+          if @stopped
             Datadog.logger.debug('remote worker: refusing to restart after previous stop')
             return
           end
@@ -60,7 +60,7 @@ module Datadog
 
           @started = false
           @thr = nil
-          @stop_requested = true
+          @stopped = true
 
           Datadog.logger.debug { 'remote worker stopped' }
         ensure
--- a/lib/datadog/core/remote/worker.rb
+++ b/lib/datadog/core/remote/worker.rb
@@ -19,12 +19,12 @@ module Datadog
           @block = block
         end
 
-        def start(allow_restart: false)
+        def start
           Datadog.logger.debug { 'remote worker starting' }
 
           acquire_lock
 
-          if @stop_requested && !allow_restart
+          if @stop_requested
             Datadog.logger.debug('remote worker: refusing to restart after previous stop')
             return
           end
--- a/lib/datadog/core/remote/worker.rb
+++ b/lib/datadog/core/remote/worker.rb
@@ -10,8 +10,8 @@ module Datadog
           @thr = nil
 
           @starting = false
-          @stopping = false
           @started = false
+          @stop_requested = false
 
           @interval = interval
           raise ArgumentError, 'can not initialize a worker without a block' unless block
@@ -19,10 +19,15 @@ module Datadog
           @block = block
         end
 
-        def start
+        def start(allow_restart: false)
           Datadog.logger.debug { 'remote worker starting' }
 
           acquire_lock
+
+          if @stop_requested && !allow_restart
+            Datadog.logger.debug('remote worker: refusing to restart after previous stop')
+            return
+          end
 
           return if @starting || @started
 
@@ -46,8 +51,6 @@ module Datadog
 
           acquire_lock
 
-          @stopping = true
-
           thread = @thr
 
           if thread
@@ -56,8 +59,8 @@ module Datadog
           end
 
           @started = false
-          @stopping = false
           @thr = nil
+          @stop_requested = true
 
           Datadog.logger.debug { 'remote worker stopped' }
         ensure
--- a/sig/datadog/core/remote/worker.rbs
+++ b/sig/datadog/core/remote/worker.rbs
@@ -12,7 +12,7 @@ module Datadog
 
         def initialize: (interval: ::Float) { () -> void } -> void
 
-        def start: () -> void
+        def start: (?allow_restart: bool) -> void
 
         def stop: () -> void
 
--- a/sig/datadog/core/remote/worker.rbs
+++ b/sig/datadog/core/remote/worker.rbs
@@ -12,7 +12,7 @@ module Datadog
 
         def initialize: (interval: ::Float) { () -> void } -> void
 
-        def start: (?allow_restart: bool) -> void
+        def start: () -> void
 
         def stop: () -> void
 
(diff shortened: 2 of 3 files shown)


AlexJF (approved):

lib/datadog/core/remote/worker.rb
```diff
@@ -10,20 +10,25 @@ def initialize(interval:, &block)
           @thr = nil
 
           @starting = false
-          @stopping = false
           @started = false
+          @stop_requested = false
```
lloeki, line 14: Not fond of what the naming implies here ^^', I feel it introduces incidental complexity.

Here's my train of thought.

This sequence diagram shows how I view the state machine†:

```
{ ¬started, ¬starting, ¬stopping }     # 1.
           |
           |
        #start
           |
           V
{ ¬started, starting, ¬stopping }      # 2.
           |
           |
        success
           |
           V
{ started, ¬starting, ¬stopping }      # 3.
           |
           |
         #stop
           |
           V
{ started, ¬starting, stopping }       # 4.
           |
           |
         success
           |
           V
{ ¬started, ¬starting, ¬stopping }     # == 1.
```

In that context, `stop_requested` captures a transition (IOW `#stop` has been called) instead of representing a state. In effect this leads to a representational issue: `@stop_requested` will be forever `true` even when a restartable worker has been restarted and no `#stop` has been issued to that second `#start` state.

AIUI what `stop_requested` intends to introduce is:

- a distinction between the first and last state of the diagram, turning the last one into a 5. then making 5. equivalent to 1. by a conditional on `@allow_restart`
- a distinction between 3. and 4. which became identical due to the removal of `@stopping`.

As a consequence I feel introducing a `stopped` state would better represent all possible states:

```
{ ¬started, ¬starting, ¬stopping, ¬stopped }     # 1.
           |
           |
        #start
           |
           V
{ ¬started, starting, ¬stopping, ¬stopped }      # 2.
           |
           |
        success
           |
           V
{ started, ¬starting, ¬stopping, ¬stopped }      # 3.
           |
           |
         #stop
           |
           V
{ started, ¬starting, stopping, ¬stopped }       # 4.
           |
           |
         success
           |
           V
{ ¬started, ¬starting, ¬stopping, stopped }      # 5.
           |
           |
        #start
           |
           V
{ ¬started, starting, ¬stopping, stopped }       # 6.
           |
           |
        success
           |
           V
{ started, ¬starting, ¬stopping, ¬stopped }      # == 3.
```

It then follows that:

- `@allow_restart` is only effective when `stopped` (and not effective when `¬stopped`)
- the current _implementation_ of `stop_requested` is equivalent to `stopping OR stopped`
  - ... except when the worker is `started`; in that cas I feel a stop count (or maybe better, a start count) could be useful
- the ambiguity about `stop_requested` on a restartable worker is resolved and fully captured by `stopping`

I also feel it makes it easier to implement helper methods from these fundamental state primitives (e.g synthesising `#stop_requested` if needed)

† `success` is actually a diamond with an `error` branch, elided so as not to overload the sequence diagram, and also because it happens via exceptional code flow.

lloeki (commented):
Thanks for tackling that one!

I seem to remember we talked about protecting concurrent access with a lock inside `Datadog::Some::Thing.active_remote`. @ivoanjo do you feel this PR is instead enough to solve the issue?

lib/datadog/core/remote/worker.rb
```diff
@@ -10,20 +10,25 @@ def initialize(interval:, &block)
...
 
           @interval = interval
           raise ArgumentError, 'can not initialize a worker without a block' unless block
 
           @block = block
         end
 
-        def start
+        def start(allow_restart: false)
```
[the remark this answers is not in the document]
ivoanjo: Removed in 760e084bdf5267ec962f66da7a62815330a4d70c

lib/datadog/core/remote/worker.rb
ivoanjo: Changed `@stop_requested` to `@stopped` in 22f33301baac1da4fc555e22a12cef66c85260fc

lloeki (approved):

PR #3520, opened 2024-03-13
[NO-TICKET] Replace `profiling.advanced.experimental_allocation_enabled` with `profiling.allocation_enabled` and remove experimental warning

ivoanjo:
**What does this PR do?**

In the next dd-trace-rb release, allocation profiling is no longer experimental. To reflect this, this PR includes two changes:

* It replaces the `profiling.advanced.experimental_allocation_enabled` with the new `profiling.allocation_enabled` option
* It turns the previous warning when turning on allocation profiling into a debug log message

**Motivation:**

Update code to match non-experimental status of feature.

**Additional Notes:**

I've decided to move the new setting outside the `advanced` configuration group, as enabling allocation profiling is not considered an "advanced" option.

To allow for smooth upgrades, I've left `profiling.advanced.experimental_allocation_enabled` in place with a deprecation warning, but unfortunately I was not able to find a nice way to automatically set the new `profiling.allocation_enabled` when the old one was set.

This means that users that had previously-enabled allocation profiling via code will need to update their configuration after upgrading, otherwise they won't get allocation profiling.

On the bright side, I was still able to support the old environment variable, so any customers that enabled the feature previously using that path will still get allocation profiling, plus a deprecation warning suggesting moving to the new environment variable.

**How to test the change?**

**For Datadog employees:**

### DIFF
--- a/lib/datadog/core/configuration/settings.rb
+++ b/lib/datadog/core/configuration/settings.rb
@@ -205,6 +205,22 @@ module Datadog
             option :transport
           end
 
+          # Can be used to enable/disable collection of allocation profiles.
+          #
+          # This feature is disabled by default
+          #
+          # @warn Due to bugs in Ruby we only recommend enabling this feature in
+          #       Ruby versions 2.x, 3.1.4+, 3.2.3+ and 3.3.0+
+          #       (more details in {Datadog::Profiling::Component.enable_allocation_profiling?})
+          #
+          # @default `DD_PROFILING_ALLOCATION_ENABLED` environment variable as a boolean, otherwise `false`
+          option :allocation_enabled do |o|
+            o.type :bool
+            o.deprecated_env 'DD_PROFILING_EXPERIMENTAL_ALLOCATION_ENABLED' # TODO: Remove this for dd-trace-rb 2.0
+            o.env 'DD_PROFILING_ALLOCATION_ENABLED'
+            o.default false
+          end
+
           # @public_api
           settings :advanced do
             # @deprecated No longer does anything, and will be removed on dd-trace-rb 2.0.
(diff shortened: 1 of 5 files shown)


lib/datadog/core/configuration/settings.rb
```diff
@@ -322,32 +338,32 @@ def initialize(*_)
...
               o.type :bool
-              o.env 'DD_PROFILING_EXPERIMENTAL_ALLOCATION_ENABLED'
               o.default false
+              o.after_set do |_, _, precedence|
+                unless precedence == Datadog::Core::Configuration::Option::Precedence::DEFAULT
+                  Datadog.logger.warn(
+                    'The profiling.advanced.experimental_allocation_enabled setting has been deprecated for removal and ' \
+                    'no longer does anything. Please remove it from your Datadog.configure block. ' \
+                    'Allocation profiling is now controlled by the profiling.allocation_enabled setting instead.'
+                  )
```
AlexJF, line 364: Related to your PR description, is the problem that `set_option(:allocation_enabled, value)` would act on the `advanced` settings, not on the parent `profiling` settings?

I'm still not fully acquainted with the option code but it could be interesting if you could do `allocation_enabled_option = option :allocation_enabled` and then call `allocation_enabled_option.set(value)` from here. Although yeah if not heavily restricted I can see how this can create a mess of setting value discoverability over time 😂

AlexJF (approved):

lib/datadog/core/configuration/settings.rb
ivoanjo: > Related to your PR description, is the problem that set_option(:allocation_enabled, value) would act on the advanced settings, not on the parent profiling settings?

Yes, exactly -- unless I saw it completely wrong, the option groups objects form a tree and are linked only to their children.

> I'm still not fully acquainted with the option code but it could be interesting if you could do allocation_enabled_option = option :allocation_enabled and then call allocation_enabled_option.set(value) from here. Although yeah if not heavily restricted I can see how this can create a mess of setting value discoverability over time 😂

Hmm... I'll admit I didn't try something like this. Let me give a final try to see if I can make it work.

ivoanjo: Ok so after thinking about it some more, here's how far I got:

1. `allocation_enabled_option = option :allocation_enabled` was a good try, but doesn't quite work because we get an `OptionDefinition` instance. The actual `Option` instance that keeps the data is created separately (in `resolve_option`) and so we don't actually have the object we need at this point.

2. It's possible to add a link to the parent group when defining the options -- [here](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/core/configuration/base.rb#L41) but I'm somewhat reluctant to add a feature to the options system (which is already quite complex) just for this

3. One thing that occurred to me would be inverting the order of the options. That is, `allocation_enabled`, since it's higher in the tree, is able to access the `experimental_allocation_enabled`, so one option would be for `allocation_enabled` to write its result into `experimental_allocation_enabled` and then have the code read the result from `experimental_allocation_enabled` until we deprecated it. That... would somewhat work, although it would interfere with the deprecation warning messages (e.g. not sure if we could still have them).

Overall... Given the current number of users we have using the experimental feature and that we're going to remove the option soon (e.g. for 2.0), I'm not sure it's worth trying really hard to support reading from the old config. (But let me know if you disagree!)

And overall our settings system is too damn complex >_>

ivoanjo: Given you approved, I'm going to go ahead and merge this, but happy to keep iterating on it as a separate PR if you're not convinced by my attempts above ;)

PR #3527, opened 2024-03-15
Fix lib injection artifact file permission

TonyCTHsu:
**What does this PR do?**

The change moved the packaging into /tmp the file permissions there [default to being wide open](https://unix.stackexchange.com/questions/71622/what-are-correct-permissions-for-tmp-i-unintentionally-set-it-all-public-recu). This PR changes the permission to `drwxr-xr-x`

### DIFF
--- a/.gitlab/build-deb-rpm.sh
+++ b/.gitlab/build-deb-rpm.sh
@@ -5,7 +5,7 @@ echo -n "$RUBY_PACKAGE_VERSION" > auto_inject-ruby.version
 
 source common_build_functions.sh
 
-chmod a+r -R ../tmp/*
+chmod -R a+r,go-w ../tmp/*
 
 fpm_wrapper "datadog-apm-library-ruby" "$RUBY_PACKAGE_VERSION" \
  --input-type dir \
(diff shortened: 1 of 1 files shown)


.gitlab/build-deb-rpm.sh
```diff
@@ -5,7 +5,7 @@ echo -n "$RUBY_PACKAGE_VERSION" > auto_inject-ruby.version
 
 source common_build_functions.sh
 
-chmod a+r -R ../tmp/*
+chmod -R a+r,go-w ../tmp/*
```
marcotc, line 8: Can you add a comment here explaining what the permission change is accomplishing, so we can reference it in the future?

ivoanjo: Is `../tmp` here the same as the top-level `/tmp` in the root of the filesystem?

If so, should we maybe use a subdirectory of `tmp`? E.g. so that we don't accidentally package other things that may have left files in `/tmp`?

TonyCTHsu: > Is ../tmp here the same as the top-level /tmp in the root of the filesystem?

No, this is under the `dd-trace-rb`

Issue #3529, opened 2024-03-15
dd-trace-rb 1.21.0 LoadError failures when running under a different user


AlexJF:
**Current behaviour**
* When the ddtrace gem is installed under a user (say `userA`).
* And the ruby app runs under a different user (say `userB`).
* Then `LoadError` errors are raised either during `ddtracerb`/`ddprofrb` execution or during requirement of a file included in the gem (such as `require 'ddtrace/auto_instrument'`).

**Expected behaviour**
* ddtrace gem files should be available by default independently of the user that installed them, unless user-specified permissions dictate otherwise.

**Steps to reproduce**
* Dockerfile
```
FROM ruby:3.3
ARG ddtrace_version

WORKDIR /app

RUN echo "source 'https://rubygems.org'\n\ngem 'ddtrace', '$ddtrace_version'" >> gems.rb

RUN bundle install

CMD bundle exec ddtracerb -h
```

* Try with 1.20.0 and a made-up user/group id (different from the root user who ran the bundle install in the Dockerfile), you get  the expected output of:

```
❯ docker run -u 502:20 -it $(docker build -q . --build-arg ddtrace_version=1.20.0)
Command '' is not valid for ddtrace.

  Usage: ddtracerb [command] [arguments]
    exec [command]: Executes command with tracing & profiling preloaded.
    help:           Prints this help message.
```

* Try with 1.21.0 with the same made-up user/group id:

```
❯ docker run -u 502:20 -it $(docker build -q . --build-arg ddtrace_version=1.21.0)
bundler: failed to load command: ddtracerb (/usr/local/bundle/bin/ddtracerb)
/usr/local/bundle/bin/ddtracerb:25:in `load': cannot load such file -- /usr/local/bundle/gems/ddtrace-1.21.0/bin/ddtracerb (LoadError)
        from /usr/local/bundle/bin/ddtracerb:25:in `<top (required)>'
        from /usr/local/lib/ruby/3.3.0/bundler/cli/exec.rb:58:in `load'
        from /usr/local/lib/ruby/3.3.0/bundler/cli/exec.rb:58:in `kernel_load'
        from /usr/local/lib/ruby/3.3.0/bundler/cli/exec.rb:23:in `run'
        from /usr/local/lib/ruby/3.3.0/bundler/cli.rb:451:in `exec'
        from /usr/local/lib/ruby/3.3.0/bundler/vendor/thor/lib/thor/command.rb:28:in `run'
[...]

PR #3531, opened 2024-03-18
[NO-TICKET] Add validation for gem file permissions

ivoanjo:
**What does this PR do?**

This PR adds validation so we catch the issue from #3529 before the incorrect packages are published on rubygems.org.

Specifically, it checks that every file in packaged `.gem` file has the expected permissions.

**Motivation:**

Avoid #3529 happening again.

**Additional Notes:**

N/A

**How to test the change?**

You can run `bundle exec rake build` to trigger this validation. Try setting incorrect permissions on one of the files and running, and you should see the validation failing.

**For Datadog employees:**

### DIFF
--- a/spec/ddtrace/gem_packaging_spec.rb
+++ b/spec/ddtrace/gem_packaging_spec.rb
@@ -1,0 +1,30 @@
+require 'rubygems'
+require 'rubygems/package'
+require 'rubygems/package/tar_reader'
+
+RSpec.describe 'gem release process (after packaging)' do
+  # TODO: This will need to be updated for the 2.0 branch
+  let(:gem_name) { 'ddtrace' }
+  let(:gem_version) { DDTrace::VERSION::STRING }
+  let(:packaged_gem_file) { "pkg/#{gem_name}-#{gem_version}.gem" }
+  let(:executable_permissions) { ['bin/ddprofrb', 'bin/ddtracerb'] }
+
+  it 'sets the right permissions on the gem files' do
+    Gem::Package::TarReader.new(File.open(packaged_gem_file)) do |tar|
+      data = tar.find { |entry| entry.header.name == 'data.tar.gz' }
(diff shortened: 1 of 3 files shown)


spec/ddtrace/gem_packaging_spec.rb
```diff
@@ -0,0 +1,30 @@
+require 'rubygems'
+require 'rubygems/package'
+require 'rubygems/package/tar_reader'
+
+RSpec.describe 'gem release process (after packaging)' do
+  # TODO: This will need to be updated for the 2.0 branch
```
AlexJF, line 6: Given the release task copies all gems under `/pkg` maybe we could run this check against all `*.gem` files in pkg atm? May be slightly annoying in that it may match on things not produced by the build but those things would be invalid gems and should be deleted sooner rather than later anyway?

AlexJF (approved):

spec/ddtrace/gem_packaging_spec.rb
ivoanjo: Good point. I still kept a check that the file we expect is there (just in case we run the spec out-of-order or something like that), but I've expanded it to check every file: https://github.com/DataDog/dd-trace-rb/pull/3531/commits/b059258dbc5652ae5a359c2a47bcc92c87c9b2a7

p-datadog (approved):

spec/ddtrace/gem_packaging_spec.rb
AlexJF: Yeah keeping the file check is smart indeed!

AlexJF (approved):

marcotc (approved):

ivoanjo:
I'm going to go ahead and merge this, since the failing CI jobs are unrelated and we're looking into them separately :)

PR #3534, opened 2024-03-18
[ci] Fix 2.1, 2.2 and 2.3 Dockerfiles

AlexJF:
**What does this PR do?**
Make Dockerfiles for 2.1, 2.2 and 2.3 work with debian bookworm (2.2 with bullseye due to https://github.com/ruby/setup-ruby/issues/496) rather than stretch, whose repos have been archived.

@ivoanjo pointed out we could have just rerouted the repos to debian archive but this way we stop relying on such old setups which makes things like docker setup and any eventual changes much easier since we're less likely to run into more unsupported/broken things.

**Motivation:**
Make github CI jobs work for 2.1, 2.2 and 2.3 again.

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**
- [ ] If this PR touches code that signs or publishes builds or packages, or handles
credentials of any kind, I've requested a review from `@DataDog/security-design-and-guidance`.
- [ ] This PR doesn't touch any of that.

Unsure? Have a question? Request a review!

PR #3541, opened 2024-03-20
[NO-TICKET] Fix Ruby 2.3 system tests by using Debian 11 bullseye as a base

ivoanjo:
**What does this PR do?**

This PR changes our Ruby 2.3 docker image to use Debian 11 "bullseye" as a base instead of Debian 12 "buster".

In https://github.com/DataDog/dd-trace-rb/pull/3534 we updated this image but this broke system-tests
(https://github.com/DataDog/dd-trace-rb/actions/runs/8346827939/job/22845592152) because they use an old nokogiri version that fails to build on modern systems (https://github.com/sparklemotion/nokogiri/issues/2105):

```
$ gem install nokogiri -v '= 1.10.10'

xml_document.c:495:14: error: conflicting types for 'canonicalize'; have 'VALUE(int,  VALUE *, VALUE)' {aka 'long unsigned int(int,  long unsigned int *, long unsigned int)'}
  495 | static VALUE canonicalize(int argc, VALUE* argv, VALUE self)
      |              ^~~~~~~~~~~~
In file included from /usr/include/features.h:489,
                 from /usr/include/x86_64-linux-gnu/bits/libc-header-start.h:33,
                 from /usr/include/stdlib.h:26,
                 from ./nokogiri.h:4,
                 from ./xml_document.h:4,
                 from xml_document.c:1:
/usr/include/x86_64-linux-gnu/bits/mathcalls.h:370:1: note: previous declaration of 'canonicalize' with type 'int(double *, const double *)'
  370 | __MATHDECL_1 (int, canonicalize,, (_Mdouble_ *__cx, const _Mdouble_ *__x));
      | ^~~~~~~~~~~~
make: *** [Makefile:240: xml_document.o] Error 1

make failed, exit code 2

Gem files will remain installed in /usr/local/bundle/gems/nokogiri-1.10.10 for inspection.
Results logged to /usr/local/bundle/extensions/x86_64-linux/2.3.0/nokogiri-1.10.10/gem_make.out
```

Rather than fighting nokogiri or the system-tests app, let's instead gently step back from Debian 12 to 11, which should still buy us plenty of time until we deprecate Ruby 2.3.

**Motivation:**

Fix CI/system-tests for Ruby 2.3.

**Additional Notes:**

N/A

**How to test the change?**

I've locally built the image with

```bash
$ docker build . -f Dockerfile-2.3.8 -t some-testing-name
```

and confirmed that nokogiri successfully installs.

**For Datadog employees:**

### DIFF
--- a/.circleci/images/primary/Dockerfile-2.3.8
+++ b/.circleci/images/primary/Dockerfile-2.3.8
@@ -1,7 +1,12 @@
 # Note: See the "Publishing updates to images" note in ./README.md for how to publish new builds of this container image
 
-# taken from https://hub.docker.com/layers/library/ruby/2.3.8-jessie/images/sha256-cff37ff7e88348a8dd630f9b2d1c679fe9e7b597cfb51a30c42bf2c2afb2eca2?context=explore
-FROM buildpack-deps:bookworm AS ruby-2.3.8-bookworm
+# Based on https://hub.docker.com/layers/library/ruby/2.3.8-jessie/images/sha256-cff37ff7e88348a8dd630f9b2d1c679fe9e7b597cfb51a30c42bf2c2afb2eca2?context=explore
+#
+# Important note: In https://github.com/DataDog/dd-trace-rb/pull/3534 we actually tried to use debian 12 (bookworm) here
+# but ran into issues because one of the test apps we use in system-tests uses a version of nokogiri that fails to
+# build on that debian version. If you change the base image, make sure that `gem install nokogiri -v '= 1.10.10'`
+# successfully installs.
+FROM buildpack-deps:bullseye AS ruby-2.3.8-bullseye
 
 # skip installing gem documentation
 RUN mkdir -p /usr/local/etc \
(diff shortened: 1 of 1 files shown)


AlexJF (approved):

lloeki (approved):

ivoanjo:
Notes on this PR:

1. System-tests will still fail since it won't pick up the new image until we publish it (https://github.com/DataDog/dd-trace-rb/actions/runs/8356497866/job/22873701994?pr=3541)

2. CircleCI tests for 2.1/2.2/2.3 are failing with many different reasons (missing ffi, missing database driver). I believe the fix here is to bust the CircleCI caches -- they are probably caching gems from a different debian base image and they need to be rebuilt. We need to bust the cache on a later PR. I say a later PR because this is a) unrelated to this change in particular; b) we're going to need to bust the cache again after publishing this new image.

The key bit to look out is for https://github.com/DataDog/dd-trace-rb/actions/runs/8356497863/job/22873697379?pr=3541 -- the new image build do succeed.

PR #3543, opened 2024-03-20
Add Rails 7.1 to the system test matrix

lloeki:
**2.0 Upgrade Guide notes**

**What does this PR do?**

Add Rails 7.1

**Motivation:**

missing coverage

**Additional Notes:**

- followup to https://github.com/DataDog/system-tests/pull/2242
- more to come with subsequent https://github.com/DataDog/system-tests/pull/2246

**How to test the change?**

CI should pass

**For Datadog employees:**

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -27,8 +27,9 @@ jobs:
           repository: 'DataDog/system-tests'
       - name: Pull released image
         run: |
-          docker pull ghcr.io/datadog/dd-trace-rb/system-tests/${{ matrix.image }}:latest
-          docker tag ghcr.io/datadog/dd-trace-rb/system-tests/${{ matrix.image }}:latest system_tests/${{ matrix.image }}:latest
+          if docker pull ghcr.io/datadog/dd-trace-rb/system-tests/${{ matrix.image }}:latest; then
+            docker tag ghcr.io/datadog/dd-trace-rb/system-tests/${{ matrix.image }}:latest system_tests/${{ matrix.image }}:latest
+          fi
       - name: Build image
         run: ./build.sh --images ${{ matrix.image }}
       - name: List images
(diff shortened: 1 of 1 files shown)


.github/workflows/system-tests.yml
```diff
@@ -27,8 +27,9 @@ jobs:
           repository: 'DataDog/system-tests'
       - name: Pull released image
         run: |
-          docker pull ghcr.io/datadog/dd-trace-rb/system-tests/${{ matrix.image }}:latest
-          docker tag ghcr.io/datadog/dd-trace-rb/system-tests/${{ matrix.image }}:latest system_tests/${{ matrix.image }}:latest
+          if docker pull ghcr.io/datadog/dd-trace-rb/system-tests/${{ matrix.image }}:latest; then
+            docker tag ghcr.io/datadog/dd-trace-rb/system-tests/${{ matrix.image }}:latest system_tests/${{ matrix.image }}:latest
+          fi
```
ivoanjo, line 32: I saw your description in the commit message

> This step is there to have a previous image as cache for subsequent
build steps. It fails if the image has not been pushed, which only
happens from master.

but I'm still not sure exactly why this is needed :thinking: . Why exactly do we need to tag the image here, and then again in later steps?

lloeki: That's because system tests have an "interesting" idea of image naming, internally encoded in `build.sh` and whatever various python bits in `utils`  `run.sh` ends up invoking:

```
# pulls the last published image, it may fail if there's none
docker pull ghcr.io/datadog/dd-trace-rb/system-tests/${{ matrix.image }}:latest

# if above pulled, tag the image as system tests expects it to be named internally
docker tag ghcr.io/datadog/dd-trace-rb/system-tests/${{ matrix.image }}:latest system_tests/${{ matrix.image }}:latest
```

This allows layers to be reused... in some limited cases because docker has complexities with caching: when not using BuildKit that would work right away, but when using BuildKit (which system tests now do) [docker caching backend shenanigans](https://docs.docker.com/build/cache/backends/) come into play.

ivoanjo: Thanks for the context! One thing that occurred to me from our discussion on slack is that it may be worth leaving here a comment clarifying that this step in effect is optional and if it fails it's OK -- it just takes longer because it doesn't use the cached images.

ivoanjo (approved):
:+1: LGTM

.github/workflows/system-tests.yml
lloeki: I'm planning some changes which should make it go away!

PR #3546, opened 2024-03-20
feat: Add span links to the Datadog API

mabdinur:
**What does this PR do?**

Introduces an API for generating and serializing span links using the Datadog API. Support for OTEL span links will be introduced in a future PR.

Inspiration: https://github.com/open-telemetry/opentelemetry-ruby/blob/b1d6d050c613ad81f51dbe6db498815deeb110d4/api/lib/opentelemetry/trace/link.rb#L13

**Motivation:**

Span Links provides a new mechanism to correlate spans. 

**Additional Notes:**

- This change requires  `datadog-agent v7.41`. datadog-agent<7.41 does not support deserializing span links. 
- This is not a breaking change to the v0.4 trace API. Older versions of the agent will ignore the span_link field in trace payloads.

**How to test the change?**

**For Datadog employees:**


spec/datadog/tracing/span_link_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+        it { is_expected.to eq(trace_id: 12, span_id: 34, flags: 0) }
+      end
+
+      context 'when trace_id >= 2^64' do
+        let(:trace_id) { 2**64 + 12 }
+        it { is_expected.to eq(trace_id: 12, trace_id_high: 1, span_id: 34, flags: 0) }
+      end
+    end
+
+    context 'required fields are set to nil' do
```
mabdinur, line 86:
```suggestion
    context 'when required fields are not set' do
```

lib/datadog/tracing/span.rb
```diff
@@ -136,7 +140,8 @@ def to_hash
           service: @service,
           span_id: @id,
           trace_id: @trace_id,
-          type: @type
+          type: @type,
+          span_links: @links.map!(&:to_hash)
```
marcotc, line 144: This is modifying the original `@links` list of objects, which you probably don't want. Use `map` instead.

lib/datadog/tracing/span_link.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      def initialize(
+        attributes: nil,
+        digest: nil
+      )
+        @span_id = digest&.span_id
+        @trace_id = digest&.trace_id
+        @trace_flags = digest&.trace_flags
+        @trace_state = digest&.trace_state && digest&.trace_state.dup
+        @dropped_attributes = 0
```
marcotc, line 46: I see that this attribute is not exposed as a public accessor. I think that if we are capturing it and populating it and returning it when serializing this object, we should have a properly documented public access accessor.

lib/datadog/tracing/transport/serializable_trace.rb
```diff
@@ -93,6 +93,8 @@ def to_msgpack(packer = nil)
           packer.write(span.meta)
           packer.write('metrics')
           packer.write(span.metrics)
+          packer.write('span_links')
+          packer.write(span.links.map!(&:to_hash))
```
marcotc, line 97: Same comment here about `map!`

marcotc (approved):

PR #3550, opened 2024-03-21
Temp disable publishing lib injection artifacts

TonyCTHsu:
**What does this PR do?**

On the verge of releasing `2.0.0.beta1`, disable publishing lib injection artifacts since they are not safe to be publicly available.

### DIFF
--- a/.github/workflows/release-lib-injection.yml
+++ b/.github/workflows/release-lib-injection.yml
@@ -1,33 +1,32 @@
+# name: "Release Library Injection"
+# on:
+#   push:
+#     tags:
+#       - 'v*.*.*'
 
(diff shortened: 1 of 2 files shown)


marcotc (approved):

.github/workflows/release-lib-injection.yml
```diff
@@ -1,33 +1,32 @@
+# name: "Release Library Injection"
```
marcotc, line 1: I suggest leaving a comment in code to remind us to re-enable this, and when we should re-enable it

TonyCTHsu: I forgot to push

PR #3551, opened 2024-03-21
Upgrade Telemetry to V2

marcotc:
This PR upgrades `ddtrace` to send Telemetry in the V2 schema version, instead of V1.

Also, this PR cleans up a lot of the existing telemetry code, simplifying its implementation and increasing type checking.

Done

- [x] System tests updated and passing: https://github.com/DataDog/system-tests/pull/2266
- [x] Update unit tests.

**Additional Notes:**

**How to test the change?**

**For Datadog employees:**


lib/datadog/core/telemetry/event.rb
```diff
@@ -1,81 +1,272 @@
...
+            tracing.sampling.default_rate
+            tracing.sampling.rate_limit
+          ].freeze
+
+          def configuration
+            config = Datadog.configuration
+
+            list = [
+              conf_value('DD_AGENT_HOST', config.agent.host),
+              conf_value('DD_AGENT_TRANSPORT', agent_transport(config)),
```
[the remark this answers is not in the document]
marcotc: > adding a config option now to configure a different transport

Sounds reasonable.
I don't think I can fit it into this PR, but it's a good requirement.

lib/datadog/core/configuration.rb
```diff
@@ -81,18 +81,22 @@ def configure
...
             else
               components = build_components(configuration)
-              components.telemetry.started!
+              start_telemetry = true
               components
             end
           )
         end
 
+        components.telemetry.started! if start_telemetry
```
marcotc, line 98: Telemetry doesn't have to start inside the global. In fact, while we're factoring it, I triggered a deadlock a few times while trying to read global information for telemetry purposes.

lib/datadog/core/configuration/option.rb
```diff
@@ -303,10 +303,6 @@ def skip_validation?
           ['true', '1'].include?(ENV.fetch('DD_EXPERIMENTAL_SKIP_CONFIGURATION_VALIDATION', '').strip)
         end
 
-        # Used for testing
```
marcotc, line 306: The time has come for someone besides tests to use the data!

integration/apps/rails-five/app/controllers/health_controller.rb
```diff
@@ -16,7 +16,6 @@ def detailed_check
       profiler_threads: Thread.list.map(&:name).select { |it| it && it.include?('Profiling') },
       telemetry_enabled: Datadog.configuration.telemetry.enabled,
       telemetry_client_enabled: Datadog.send(:components).telemetry.enabled,
-      telemetry_worker_enabled: Datadog.send(:components).telemetry.worker.enabled?
```
marcotc, line 19: The internal worker instance is now private, even it's an implementation concern.

lib/datadog/core/configuration/option.rb
```diff
@@ -8,7 +8,7 @@ module Configuration
       # Represents an instance of an integration configuration option
       # @public_api
       class Option
-        attr_reader :definition
+        attr_reader :definition, :precedence_set
```
delner, line 11: What is `precedence_set`?

lib/datadog/core/telemetry/client.rb
```diff
@@ -37,21 +42,24 @@ def disable!
         def started!
           return if !@enabled || forked?
 
-          res = @emitter.request(:'app-started')
+          res = @emitter.request(Event::AppStarted.new)
```
delner, line 45: I like the use of a class. Nit, but does it make sense to initialize an object?

lib/datadog/core/telemetry/event.rb
```diff
@@ -1,81 +1,277 @@
...
-require_relative 'v2/app_client_configuration_change'
-
 module Datadog
   module Core
     module Telemetry
-      # Class defining methods to construct a Telemetry event
       class Event
-        include Telemetry::Collector
+        # Base class for all Telemetry V2 events.
+        class Base
```
delner, line 8: Should this be a class? Or could it be a module? (Given it doesn't introduce state.)

lib/datadog/core/telemetry/event.rb
```diff
@@ -1,81 +1,277 @@
...
+          end
+        end
+
+        # Telemetry class for the 'app-started' event
+        class AppStarted < Base
+          def type
+            'app-started'
+          end
 
-        API_VERSION = 'v1'
```
delner, line 15: Is there any versioning of the API within our code?

lib/datadog/core/telemetry/http/transport.rb
```diff
@@ -39,10 +39,15 @@ def request(request_type:, payload:)
...
-              Datadog::Core::Transport::Ext::HTTP::HEADER_DD_INTERNAL_UNTRACED_REQUEST => '1',
-              Http::Ext::HEADER_CONTENT_TYPE => Http::Ext::CONTENT_TYPE_APPLICATION_JSON,
-              Http::Ext::HEADER_DD_TELEMETRY_API_VERSION => api_version,
-              Http::Ext::HEADER_DD_TELEMETRY_REQUEST_TYPE => request_type,
+              Core::Transport::Ext::HTTP::HEADER_DD_INTERNAL_UNTRACED_REQUEST => '1',
+              Ext::HEADER_CONTENT_TYPE => Http::Ext::CONTENT_TYPE_APPLICATION_JSON,
+              Ext::HEADER_DD_TELEMETRY_API_VERSION => api_version,
+              Ext::HEADER_DD_TELEMETRY_REQUEST_TYPE => request_type,
+              Ext::HEADER_CLIENT_LIBRARY_LANGUAGE => Core::Environment::Ext::LANG,
+              Ext::HEADER_CLIENT_LIBRARY_VERSION => DDTrace::VERSION::STRING,
```
delner, line 47: Just FYI, when ported to 2.x, this will need to be updated.

delner (commented):
Added some questions.

lib/datadog/core/telemetry/event.rb
marcotc: Yes, it's here: https://github.com/DataDog/dd-trace-rb/blob/fb2ca4dc49d813f0a61e27534d7d72dfd76e5088/lib/datadog/core/telemetry/http/ext.rb#L18

lib/datadog/core/telemetry/client.rb
marcotc: This pattern of initializing an object makes sense for the telemetry events like `AppClientConfigurationChange`: https://github.com/DataDog/dd-trace-rb/blob/fb2ca4dc49d813f0a61e27534d7d72dfd76e5088/lib/datadog/core/telemetry/client.rb#L82

`AppClientConfigurationChange` needs information to be provided at two different times: first he needs to know what has changed in the source of that change (`changes, 'remote_config'`), then later he needs to receive the auto-incrementing `seq_id`: https://github.com/DataDog/dd-trace-rb/blob/fb2ca4dc49d813f0a61e27534d7d72dfd76e5088/lib/datadog/core/telemetry/request.rb#L20

Because these two sets of information exist in different context, I need a way to start a first set (`changes, 'remote_config'`) and then merge it with the second set (`seq_id`).
So this case it makes sense to initialize the class, then call amended accepting a parameter.

Most events are simple, like you mention `AppStarted`, and don't actually need two sets of data to build the event, but in order to keep the API consistent, I had to add this flexibility to all events.

Let me know if this makes sense or if you have further suggestions.

lib/datadog/core/telemetry/event.rb
marcotc: I think the hierarchy here is that events are true children of a base class: an event is a subtype of a `Base` event.

It can technically work as a module, but I think that it would create the misleading impression that events just so happen to respect the `Base` interface, when in fact all events are a subtype of a `Base` event.

lib/datadog/core/configuration/option.rb
marcotc: This is the effective precedence of the value this option is currently set to.
For example, if this option was never set, the precedence will be "default".
If the same option is then set by programmatic configuration (`Datadog.configure{}`), it will change to "programmatic".
Then, again, if it is over, written by remote configuration, it will change to "remote configuration".

Now, the interesting part is that same option is then again remote configuration by programmatic configuration, it will not change value, given remote configuration has a higher precedence.

Let me know if this makes sense. Also, I expanded the in-line documentation of this class regarding precedence.

On another topic, it probably makes sense to simply rename me to `precedence`, what do you think? I would probably do that in a separate PR as, it will touch many files and related to this PR, but it's a very simple rename to do as a follow up.

marcotc:
Taking a look at the system test failures right now.

marcotc:
System tests should now pass.

marcotc:
The last failing system test (`test_app_dependencies_loaded_not_sent`) will be fixed when https://github.com/DataDog/system-tests/pull/2275 is merged.

lib/datadog/core/configuration/option.rb
delner: Gotcha. Was just curious. No strong opinions at this point, so if you're confident on this, I'm interested to see how it works.

lib/datadog/core/telemetry/client.rb
delner: I see: sometimes it has state attached. I suppose most of these events don't fire often, but if they ever fired as often as traces (crazy possibly, I know), then I was imagining this would apply more pressure to memory.

Still I do like the `event` as a parameter from a flexibility point of view.

Feedback is not blocking.

lib/datadog/core/telemetry/event.rb
delner: I see, this got me thinking about Sandi Metz's take on inheritance/composition. Some notes surrounding her discourse on the subject that I could find:

> *Deciding Between Inheritance and Composition*

> Inheritance gives you message delegation for free at the cost of maintaining a class hierarchy. Composition allows objects to have structural independence at the cost of explicit message delegation.

> > Composition contains far few built-in dependencies than inheritance; it is very often the best choice.

> With inheritance, a correctly modeled hierarchy will give you the benefit of propogating changes in the base class to all subclasses. This can also be a disadvantage when the hierarchy is modeled incorrectly, as a dramatic change to the base class due to a change in requirements will break sub-classes. Inheritance = built-in dependencies.

> > Enormous, broad-reaching changes of behavior can be achieved with very small changes in code. This is true, for better or for worse, whether you come to regret it or not.

> > Avoid writing frameworks that require users of your code to subclass your objects in order to gain your behavior. Their application's objects may already be arranged in a hierarchy; inheriting from your framework may not be possible.

On that very last point: is there a case where we might want to have independent trees of event behavior? I don't think immediately, no, but that would be a strong case for composition over inheritance. I suppose to me, it would be more a factor of keeping our options open in accordance to the principle of least harm: that inheritance would be less restraining.

Another way I think about the reason to do one over the other is state (e.g. instance variables).

 - Does the base class define or maintain state? (Internal or otherwise?)
 - Does it need some initialization (constructor or otherwise)?

If yes to either, then probably should be inheritance, because modules are not great at maintaining state. Otherwise consider using a module.

Some food for thought. Ultimately, we can always refactor and this doesn't constitute a breaking change, so no blocker on my side if we can always change later without much pain.

marcotc: > Deciding Between Inheritance and Composition

It seems like the composition that she is discussing is about having a delegator object. And I'm not sure that that applies here because the need is for a consistent interface. There's almost no behaviour being delegated to the base class, is for subclasses to respect interface.

Both classes and modules would fall under inheritance, based on her definition above I believe.

marcotc: Also, this whole thing is internal and constraint inside a single file, so factoring is trivial, and will likely not even affect unit tests.

delner (approved):

PR #3558, opened 2024-03-27
[PROF-9370] Enable Garbage Collection profiling by default

ivoanjo:
**What does this PR do?**

This PR turns the Garbage Collection profiling feature on by default.

Specifically it:

1. Introduces a new `c.profiling.advanced.gc_enabled` setting, which defaults to `true` (can also be toggled using the `DD_PROFILING_GC_ENABLED` env var)

2. Deprecates the previous `c.profiling.advanced.force_enable_gc_profiling` setting. This setting only served to turn the feature on, not turn it off, so it no longer makes sense to exist.

3. Does not turn on GC profiling on Rubies affected by https://bugs.ruby-lang.org/issues/18464 . Specifically, GC profiling can now only be turned on in Ruby versions 2.x, 3.1.4+, 3.2.3+ and 3.3+ . This avoids impacting known-buggy Rubies.

**Motivation:**

We've done through validation that this feature is working nicely and with low overhead, and it works great with the timeline view.

**Additional Notes:**

Regarding the impact of Ruby issue 18464, I explicitly chose to no longer allow GC profiling to be enabled on affected Rubies.

I did this so as to simplify the implementation of the settings.

Because we wanted to enable this feature by default, having a way to still enable it on these Rubies would mean either having another option (e.g. keep `force_enable_gc_profiling` around?), somehow introducing a tri-state value for the setting (like we do for the `no_signals_workaround_enabled`), or fiddling with precedences to distinguish the default value from a user-set one.

In the end, I decided to simplify and not do any of the above. We can still change our minds later and introduce it based on user feedback and needs, but unless a user really needs to be on 3.0.x for which there is no bug fix release, we really really really really instead recomend moving to the latest patch release of the 3.1.x or 3.2.x series.

**How to test the change?**

The change includes test coverage. You can also profile a simple script and observe that GC profiling is included by default, without any extra option having to be set.

**For Datadog employees:**

### DIFF
--- a/lib/datadog/profiling/component.rb
+++ b/lib/datadog/profiling/component.rb
@@ -111,19 +111,30 @@ module Datadog
       end
 
       private_class_method def self.enable_gc_profiling?(settings)
-        # See comments on the setting definition for more context on why it exists.
-        if settings.profiling.advanced.force_enable_gc_profiling
-          if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3')
-            Datadog.logger.debug(
-              'Profiling time/resources spent in Garbage Collection force enabled. Do not use Ractors in combination ' \
-              'with this option as profiles will be incomplete.'
-            )
-          end
+        return false unless settings.profiling.advanced.gc_enabled
 
-          true
-        else
-          false
+        # SEVERE - Only with Ractors
+        # On Ruby versions 3.0 (all), 3.1.0 to 3.1.3, and 3.2.0 to 3.2.2 gc profiling can trigger a VM bug
+        # that causes a segmentation fault during garbage collection of Ractors
+        # (https://bugs.ruby-lang.org/issues/18464). We don't allow enabling gc profiling on such Rubies.
+        # This bug is fixed on Ruby versions 3.1.4, 3.2.3 and 3.3.0.
+        if RUBY_VERSION.start_with?('3.0.') ||
+            (RUBY_VERSION.start_with?('3.1.') && RUBY_VERSION < '3.1.4') ||
+            (RUBY_VERSION.start_with?('3.2.') && RUBY_VERSION < '3.2.3')
+          Datadog.logger.warn(
+            "Current Ruby version (#{RUBY_VERSION}) has a VM bug where enabling GC profiling would cause "\
+            'crashes (https://bugs.ruby-lang.org/issues/18464). GC profiling has been disabled.'
+          )
+          return false
+        elsif RUBY_VERSION.start_with?('3.')
+          Datadog.logger.debug(
+            'In all known versions of Ruby 3.x, using Ractors may result in GC profiling unexpectedly ' \
+            'stopping (https://bugs.ruby-lang.org/issues/19112). Note that this stop has no impact in your ' \
+            'application stability or performance. This does not happen if Ractors are not used.'
+          )
         end
+
+        true
       end
 
       private_class_method def self.get_heap_sample_every(settings)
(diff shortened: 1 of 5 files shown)


lib/datadog/profiling/component.rb
```diff
@@ -111,19 +111,30 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
...
       private_class_method def self.enable_gc_profiling?(settings)
-        # See comments on the setting definition for more context on why it exists.
-        if settings.profiling.advanced.force_enable_gc_profiling
-          if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3')
-            Datadog.logger.debug(
-              'Profiling time/resources spent in Garbage Collection force enabled. Do not use Ractors in combination ' \
-              'with this option as profiles will be incomplete.'
-            )
-          end
+        return false unless settings.profiling.advanced.gc_enabled
```
AlexJF, line 114: For allocation we moved it out of advanced. Should we apply the same reasoning to gc?

lib/datadog/profiling/component.rb
```diff
@@ -111,19 +111,30 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
...
+        # (https://bugs.ruby-lang.org/issues/18464). We don't allow enabling gc profiling on such Rubies.
+        # This bug is fixed on Ruby versions 3.1.4, 3.2.3 and 3.3.0.
+        if RUBY_VERSION.start_with?('3.0.') ||
+            (RUBY_VERSION.start_with?('3.1.') && RUBY_VERSION < '3.1.4') ||
+            (RUBY_VERSION.start_with?('3.2.') && RUBY_VERSION < '3.2.3')
+          Datadog.logger.warn(
+            "Current Ruby version (#{RUBY_VERSION}) has a VM bug where enabling GC profiling would cause "\
+            'crashes (https://bugs.ruby-lang.org/issues/18464). GC profiling has been disabled.'
+          )
+          return false
```
AlexJF, line 128: Isn't this too aggressive since it only manifests when using Ractors (which bring other problems in those versions as well?).

Allocation suffers from the same issue and does not enforce false for this bug. We should either enforce there as well or we could have a separate flag governing disablement of both features related to ractor bugs?

AlexJF (approved):

lib/datadog/profiling/component.rb
ivoanjo: My thinking behind putting that one out of advanced and leaving this one in advanced, is that I'm not sure exactly when (if?) we're going to enable allocation profiling by default.

Thus, it's a setting that I expect many users will want to toggle until we change that.

---

Conversely, for GC profiling, given we are _enabling_ it by default I don't think anyone will need or should to touch this.

I'd say its existence is mostly a support mechanism, where if we find any issue that we suspect may be related to the feature, we have the means to disable it.

(This is why I left `timeline_enabled` under advanced as well).

LMK your thoughts, if you still think this distinction is too subtle, I don't particularly mind moving it.

lib/datadog/profiling/component.rb
ivoanjo: I'll concede it's a bit aggressive, and I shared my reasoning in the [Additional notes](https://github.com/DataDog/dd-trace-rb/pull/3558#:~:text=the%20timeline%20view.-,Additional%20Notes%3A,-Regarding%20the%20impact) for the commit.

TBH I think this feature is less important than allocation/heap, and so while we could put them on the same bucket, I... somewhat think the others deserve the extra attention, while this one doesn't quite?

AlexJF: Fair fair 😄 And as you mention, we can revisit this at any point in the future anyway.

lib/datadog/profiling/component.rb
AlexJF: Gotcha. I think it makes sense and just wanted to understand the pattern you were going with here so I could follow it in the future :)

ivoanjo:
Reminder to self: Remove separate configurations we have for GC profiling testing once this gets merged in

ivoanjo:
Merging this in! Thanks for the feedback (in this PR and in the review doc!)

PR #3565, opened 2024-04-02
Use strict base64 encoding (no newline)

lloeki:
**2.0 Upgrade Guide notes**

**What does this PR do?**

Use strict base64 encoding. 

**Motivation:**

MIME-style newlines in the middle of base64 makes for some unhappy results.

**Additional Notes:**

Reported case was with compressed data on appsec derivative results (e.g schema), introduced in https://github.com/DataDog/dd-trace-rb/commit/8c52213e047cde29a773714fc304135635d8dedd

This also applies to a few areas like RC capabilities, although the contents was smaller thus not affected (a trailing newline was handled with a `chomp`.

**How to test the change?**

CI covers it (added a non-regression test)

**For Datadog employees:**

### DIFF
--- a/lib/datadog/appsec/event.rb
+++ b/lib/datadog/appsec/event.rb
@@ -138,7 +138,7 @@ module Datadog
         private
 
         def compressed_and_base64_encoded(value)
-          Base64.encode64(gzip(value))
+          Base64.strict_encode64(gzip(value))
         rescue TypeError => e
           Datadog.logger.debug do
             "Failed to compress and encode value when populating AppSec::Event. Error: #{e.message}"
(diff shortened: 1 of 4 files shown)


ivoanjo (approved):
:+1: LGTM

I wonder if we should introduce a helper method for this -- not to avoid repeating code (it's one line), but because we wouldn't have to keep "remembering" to use the correct variant of base64 (although we'd need to remember to use the helper ;D )

lloeki:
> we wouldn't have to keep "remembering" to use the correct variant of base64

The correct version depends on what you're doing. It just turns out that the changes here all use the non-MIME one but it's really a case by case.

> (although we'd need to remember to use the helper ;D )

So true. It's also less explicit, hiding the fact that should be in plain sight and in context.

PR #3573, opened 2024-04-04
[PROF-9469] Skip heap samples with age 0

AlexJF:
**What does this PR do?**

Skip updating and serialization of heap samples for objects that were allocated in the current GC epoch/generation. The fact that these objects are still alive is usually just a happy accident from the fact that GC hasn't run yet and thus just add noise when you want to understand long-term heap usage and, more specifically, memory leaks.

In fact, allocation profiles alone can give you a rather accurate description of what the population of age == 0 should look like at any-one time. If no GCs ever ran, allocation profiles would be === to heap profiles as all objects would be technically alive.

**Motivation:**
* Objects with age == 0 tend to be mostly noise when investigating heap usage.
* Unless GCs occur at intervals <1 minute or you have big memory leaks, objects with age == 0 will naturally be much more common than objects with age > 0 and so end up being major contributors to sampling and serialization overhead. Skipping them can result in significant savings.

**Additional Notes:**

Some micro-benchmark results:

```
Comparison:
sample+serialize min_age_1 retain_every=10 heap_samples=true heap_size=true heap_sample_every=10 skip_end_gc=true:        5.7 i/s
sample+serialize min_age_0 retain_every=10 heap_samples=true heap_size=true heap_sample_every=10 skip_end_gc=true:        5.5 i/s - 1.02x  slower
sample+serialize min_age_1 retain_every=10 heap_samples=true heap_size=true heap_sample_every=1 skip_end_gc=true:        4.3 i/s - 1.33x  slower
sample+serialize min_age_0 retain_every=10 heap_samples=true heap_size=true heap_sample_every=1 skip_end_gc=true:        3.1 i/s - 1.85x  slower
```

**How to test the change?**

### DIFF
--- a/ext/datadog_profiling_native_extension/heap_recorder.c
+++ b/ext/datadog_profiling_native_extension/heap_recorder.c
@@ -10,6 +10,18 @@
   #define CAN_APPLY_GC_FORCE_RECYCLE_BUG_WORKAROUND
 #endif
 
+// Minimum age (in GC generations) of heap objects we want to include in heap
+// recorder iterations. Object with age 0 represent objects that have yet to undergo
+// a GC and, thus, may just be noise/trash at instant of iteration and are usually not
+// relevant for heap profiles as the great majority should be trivially reclaimed
+// during the next GC.
+#define ITERATION_MIN_AGE 1
+
+// Check to see if an object should not be included in a heap recorder iteration.
+// This centralizes the checking logic to ensure it's equally applied between
+// preparation and iteration codepaths.
+#define NOT_INCLUDED_IN_ITERATION(obj_record) obj_record->object_data.gen_age < ITERATION_MIN_AGE
+
 // A compact representation of a stacktrace frame for a heap allocation.
 typedef struct {
   char *name;
@@ -142,8 +154,6 @@ struct heap_recorder {
   // This enables us to calculate the age of iterated objects in the above snapshot by
   // comparing it against an object's alloc_gen.
   size_t iteration_gen;
-  // The minimum required age for an object to be included in the current iteration.
-  size_t iteration_min_gen_age;
 
   // Data for a heap recording that was started but not yet ended
   recording active_recording;
@@ -355,15 +365,12 @@ void end_heap_allocation_recording(struct heap_recorder *heap_recorder, ddog_pro
   commit_recording(heap_recorder, heap_record, active_recording);
 }
 
-size_t heap_recorder_prepare_iteration(heap_recorder *heap_recorder, size_t min_age) {
-  size_t current_gc_gen = rb_gc_count();
-
+void heap_recorder_prepare_iteration(heap_recorder *heap_recorder) {
   if (heap_recorder == NULL) {
-    return current_gc_gen;
+    return;
   }
 
-  heap_recorder->iteration_gen = current_gc_gen;
-  heap_recorder->iteration_min_gen_age = min_age;
+  heap_recorder->iteration_gen = rb_gc_count();
 
   if (heap_recorder->object_records_snapshot != NULL) {
     // we could trivially handle this but we raise to highlight and catch unexpected usages.
@@ -376,8 +383,6 @@ size_t heap_recorder_prepare_iteration(heap_recorder *heap_recorder, size_t min_
   if (heap_recorder->object_records_snapshot == NULL) {
     rb_raise(rb_eRuntimeError, "Failed to create heap snapshot.");
   }
-
-  return heap_recorder->iteration_gen;
 }
 
 void heap_recorder_finish_iteration(heap_recorder *heap_recorder) {
@@ -480,6 +485,20 @@ static int st_object_record_update(st_data_t key, st_data_t value, st_data_t ext
 
   VALUE ref;
 
+  size_t iteration_gen = recorder->iteration_gen;
+  size_t alloc_gen = record->object_data.alloc_gen;
+  // Guard against potential overflows given unsigned types here.
+  size_t age = alloc_gen < iteration_gen ? iteration_gen - alloc_gen : 0;
+  record->object_data.gen_age = age;
+
+  if (NOT_INCLUDED_IN_ITERATION(record)) {
+    // If an object won't be included in the current iteration, there's
+    // no point checking for liveness or updating its size, so exit early.
+    // NOTE: This means that there should be an equivalent check during actual
+    //       iteration otherwise we'd iterate/expose stale object data.
+    return ST_CONTINUE;
+  }
+
   if (!ruby_ref_from_id(LONG2NUM(obj_id), &ref)) {
     // Id no longer associated with a valid ref. Need to delete this object record!
     on_committed_object_record_cleanup(recorder, record);
@@ -540,12 +559,11 @@ static int st_object_records_iterate(DDTRACE_UNUSED st_data_t key, st_data_t val
   iteration_context *context = (iteration_context*) extra;
 
   const heap_recorder *recorder = context->heap_recorder;
-  size_t iteration_gen = recorder->iteration_gen;
-  size_t min_gen_age = recorder->iteration_min_gen_age;
-  size_t obj_gen_age = iteration_gen - record->object_data.alloc_gen;
 
-  if (obj_gen_age < min_gen_age) {
-    // Skip objects that don't meet minimum age configured during iteration preparation.
+  if (NOT_INCLUDED_IN_ITERATION(record)) {
+    // Skip objects that should not be included in iteration
+    // NOTE: This matches the short-circuiting condition in st_object_record_update
+    //       and prevents iteration over stale objects.
     return ST_CONTINUE;
   }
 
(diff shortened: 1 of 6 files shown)


ext/datadog_profiling_native_extension/stack_recorder.c
```diff
@@ -481,15 +480,17 @@ static VALUE _native_serialize(DDTRACE_UNUSED VALUE _self, VALUE recorder_instan
   serializer_set_start_timestamp_for_next_profile(state, finish_timestamp);
 
   // Prepare the iteration on heap recorder we'll be doing outside the GVL. The preparation needs to
-  // happen while holding on to the GVL.
-  heap_recorder_prepare_iteration(state->heap_recorder);
+  // happen while holding on to the GVL. Do not include in the iteration objects with age == 0. These
+  // objects that have yet to undergo a GC may just be noise/trash at this point and are usually not
+  // relevant for heap profiles as the great majority should be trivially reclaimed during the next GC.
+  size_t min_age = 1;
+  heap_recorder_prepare_iteration(state->heap_recorder, min_age);
```
ivoanjo, line 487: Minor: Since this constant ends up being repeated in `heap_recorder_prepare_iteration` below, may be worth adding a `#define`?

ext/datadog_profiling_native_extension/heap_recorder.h
```diff
@@ -113,8 +116,16 @@ void end_heap_allocation_recording(heap_recorder *heap_recorder, ddog_prof_Slice
...
+//   Minimum age (in GC epochs/generations) of objects we want to iterate over. Any objects whose
+//   age is below this value will not be seen by the next heap_recorder_for_each_live_object
+//   calls and thus, for efficiency, we'll also skip their updating (size and liveness checks).
+// @return gc_count() at which the iteration started being prepared. Note that preparation is
+//   not atomic and thus a GC could have executed in the middle so our notion of age can be off
+//   by a bit (hopefully at most just 1 generation/epoch).
+//
 // WARN: This must be called strictly before iteration. Failing to do so will result in exceptions.
-void heap_recorder_prepare_iteration(heap_recorder *heap_recorder);
+size_t heap_recorder_prepare_iteration(heap_recorder *heap_recorder, size_t min_gc_age);
```
ivoanjo, line 128: Minor: Since the return value doesn't end up getting used anywhere, maybe worth simplifying?

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -364,6 +376,8 @@ void heap_recorder_prepare_iteration(heap_recorder *heap_recorder) {
   if (heap_recorder->object_records_snapshot == NULL) {
     rb_raise(rb_eRuntimeError, "Failed to create heap snapshot.");
   }
+
+  return heap_recorder->iteration_gen;
```
ivoanjo, line 380: Minor: Given `iteration_gen` doesn't really change during iteration, may be easier to use `current_gc_gen` here (as otherwise the code slightly seems to suggest that they are different?)

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -137,6 +137,13 @@ struct heap_recorder {
   // mutation of the data so iteration can occur without acquiring a lock.
   // NOTE: Contrary to object_records, this table has no ownership of its data.
   st_table *object_records_snapshot;
+  // The GC gen/epoch/count in which we prepared the current iteration.
+  //
+  // This enables us to calculate the age of iterated objects in the above snapshot by
+  // comparing it against an object's alloc_gen.
+  size_t iteration_gen;
+  // The minimum required age for an object to be included in the current iteration.
+  size_t iteration_min_gen_age;
```
ivoanjo, line 146: Minor: It's not a lot of extra code to support this "configurable min age" capability, so I can live with it, but... It does kinda occur to me that this PR would be even smaller if we hardcoded the "1" constant inside the heap recorder instead of making it configurable and hardcoding it in the stack recorder :wink: 

E.g. I wonder if we're ever going to make use of this capability to set `min_age` to something else than 1 :thinking:

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -348,11 +355,16 @@ void end_heap_allocation_recording(struct heap_recorder *heap_recorder, ddog_pro
...
-    return;
+    return current_gc_gen;
   }
 
+  heap_recorder->iteration_gen = current_gc_gen;
+  heap_recorder->iteration_min_gen_age = min_age;
+
   if (heap_recorder->object_records_snapshot != NULL) {
     // we could trivially handle this but we raise to highlight and catch unexpected usages.
     rb_raise(rb_eRuntimeError, "New heap recorder iteration prepared without the previous one having been finished.");
```
ivoanjo, line 370: :arrow_down: Actually, if we're not going to include objects with age == 0, I think it would be an overall win to additionally tweak `st_object_record_update`, below, to also be aware of this.

E.g. why bother going `ruby_ref_from_id` + `ruby_obj_memsize_of` if we're not going to use the data anyway?

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -525,8 +539,17 @@ static int st_object_records_iterate(DDTRACE_UNUSED st_data_t key, st_data_t val
   const heap_stack *stack = record->heap_record->stack;
   iteration_context *context = (iteration_context*) extra;
 
-  ddog_prof_Location *locations = context->heap_recorder->reusable_locations;
+  const heap_recorder *recorder = context->heap_recorder;
+  size_t iteration_gen = recorder->iteration_gen;
+  size_t min_gen_age = recorder->iteration_min_gen_age;
+  size_t obj_gen_age = iteration_gen - record->object_data.alloc_gen;
+
+  if (obj_gen_age < min_gen_age) {
```
ivoanjo, line 547: So I'm not sure if this could happen _right now_ but it does make me slightly fearful to see this subtraction

```
  size_t obj_gen_age = iteration_gen - record->object_data.alloc_gen;
```

given that `size_t` is supposed to be unsigned.

That is -- what if we get really unlucky and an extra object gets allocated and sampled after we set `iteration_gen`: that would mean an overflow in `obj_gen_age`. I guess that would mean the object would be included, which is not that bad, but... it seems like a nice footgun lol.

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -510,6 +510,29 @@ def sample_allocation(obj)
...
+            end
+
+            sample_line = __LINE__ - 3
+
+            relevant_sample = heap_samples.find { |s| s.has_location?(path: __FILE__, line: sample_line) }
+            expect(relevant_sample).to be nil
+          ensure
+            GC.enable
+          end
+        end
```
ivoanjo, line 534: Minor: I wonder if it would be worth sampling an extra object + doing a GC before turning off GC, so that there's both a positive (expect to include that string) as well as a negative expectation (expect not to include the others), so that we don't accidentally make this test pass by e.g. setting `heap_samples` to an empty array.

ivoanjo (commented):
Looks good! I left my usual smattering of minor suggestions, one bigger suggestion and one maybe-a-bug note ;)

ext/datadog_profiling_native_extension/heap_recorder.h
AlexJF: Yeah I was using it until I realized we were ending up calculating age in 2 different places and could just include it as part of iteration info 👍

ext/datadog_profiling_native_extension/heap_recorder.c
AlexJF: I was wondering this myself. On the one hand, having the parameter makes the behaviour explicit and gives you an escape hatch to access the full data if needed (e.g. custom debugging/dumping logic). On the other hand, this behavior will be assumed in the UI through our descriptions and what not so any changes may lead to inconsistencies and we wouldn't want to really incentivize this.

I'll hardcode this inside heap for now and we could refer to the intermediate commit with the parameter in the future if we see a need in the future.

ext/datadog_profiling_native_extension/heap_recorder.c
AlexJF: 🤦 that was precisely my intention and why I marked PROF-9471 as a duplicate of PROF-9469. But I added the age check to the wrong place 😂

ext/datadog_profiling_native_extension/heap_recorder.c
AlexJF: Hmm. The preparation happens under the GVL and other than the object record snapshot it's not explicitly allocating anything. But yes you're right, even just the new st_table will likely trigger an allocation sample and would be sufficient to cause the situation you describe. Will adjust 👍

ivoanjo (approved):
:+1: Looks great!

I'm slightly curious if the change moved the micro-benchmark results needle, but feel free to ignore my curiosity haha

ivoanjo:
Ah! Somewhat sad, I was hoping this tweak would buy is a bit more headroom. I guess it might still on real-world usecases, but would've been cool to see in the microbenchmark as well :)

PR #3574, opened 2024-04-05
[NO-TICKET] Improve testing coverage for profiler sampling via signals

ivoanjo:
**What does this PR do?**

This PR improves the test coverage we have around taking samples in the profiler using signals.

While prototyping something yesterday I was surprised by the discovery that if I commented out triggering profiler samples via a signal:

```c
pthread_kill(owner.owner, SIGPROF);
```

that actually only one test of the test suite failed! (Yay, at least one!)

This is because our tests were *waiting* on a loop, and because most of the time they were waiting, sampling was happening via the idle sampling code path, not via the signal sending code path.

But this is a dangerous state to be in -- we don't want accidental changes to such core logic to not be caught clearly in tests.

Thus, this PR tightens out testing around sampling, to make sure we check for the existence of samples that were taken via the thread interruption code paths.

**Motivation:**

Improve test coverage for the profiler.

**Additional Notes:**

There's quite a few whitespace-only changes in this PR/commit, so I recommend looking at the diff while hiding whitespace.

**How to test the change?**

Comment out `pthread_kill`, and observe how tests now fail by having no samples.

### DIFF
--- a/ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -96,6 +96,7 @@ struct cpu_and_wall_time_worker_state {
   bool no_signals_workaround_enabled;
   bool dynamic_sampling_rate_enabled;
   bool allocation_profiling_enabled;
+  bool skip_idle_samples_for_testing;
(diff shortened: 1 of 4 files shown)


AlexJF (approved):
Ah nice catch!

PR #3583, opened 2024-04-11
[PROF-9472] Add recorder_stats and profile_stats to internal metadata

AlexJF:
**What does this PR do?**

Make use of the internal metadata mechanism to ship some statistics about the profile itself and the stack recorder. 

We first introduced this back in https://github.com/DataDog/dd-trace-rb/pull/3429 but that was scoped to worker stats which capture sampling-related statistics. This PR focuses on the serialization side of things + heap recorder if enabled.

**Motivation:**

Easier understanding of some of the internal ops that are otherwise hard/impossible to gather from the final pprof.

**Additional Notes:**

**How to test the change?**
Look at the `event.json` produced by any app being profiled with this PR and notice the extra fields in `internal`:

image

### DIFF
--- a/ext/datadog_profiling_native_extension/heap_recorder.c
+++ b/ext/datadog_profiling_native_extension/heap_recorder.c
@@ -159,12 +159,12 @@ struct heap_recorder {
   // Sampling state
   uint num_recordings_skipped;
 
-  struct stats {
-    size_t last_update_objects_alive;
-    size_t last_update_objects_dead;
-    size_t last_update_objects_skipped;
-    size_t last_update_objects_frozen;
-  } stats;
+  struct stats_last_update {
+    size_t objects_alive;
+    size_t objects_dead;
+    size_t objects_skipped;
+    size_t objects_frozen;
+  } stats_last_update;
 };
 static heap_record* get_or_create_heap_record(heap_recorder*, ddog_prof_Slice_Location);
 static void cleanup_heap_record_if_unused(heap_recorder*, heap_record*);
@@ -379,10 +379,8 @@ void heap_recorder_prepare_iteration(heap_recorder *heap_recorder) {
     rb_raise(rb_eRuntimeError, "New heap recorder iteration prepared without the previous one having been finished.");
   }
 
-  heap_recorder->stats.last_update_objects_alive = 0;
-  heap_recorder->stats.last_update_objects_dead = 0;
-  heap_recorder->stats.last_update_objects_skipped = 0;
-  heap_recorder->stats.last_update_objects_frozen = 0;
+  // Reset last update stats, we'll be building them from scratch during the st_foreach call below
+  heap_recorder->stats_last_update = (struct stats_last_update) {};
 
   st_foreach(heap_recorder->object_records, st_object_record_update, (st_data_t) heap_recorder);
 
@@ -445,10 +443,10 @@ VALUE heap_recorder_state_snapshot(heap_recorder *heap_recorder) {
     ID2SYM(rb_intern("num_heap_records")),   /* => */ LONG2NUM(heap_recorder->heap_records->num_entries),
 
     // Stats as of last update
-    ID2SYM(rb_intern("last_update_objects_alive")), /* => */ LONG2NUM(heap_recorder->stats.last_update_objects_alive),
-    ID2SYM(rb_intern("last_update_objects_dead")), /* => */ LONG2NUM(heap_recorder->stats.last_update_objects_dead),
-    ID2SYM(rb_intern("last_update_objects_skipped")), /* => */ LONG2NUM(heap_recorder->stats.last_update_objects_skipped),
-    ID2SYM(rb_intern("last_update_objects_frozen")), /* => */ LONG2NUM(heap_recorder->stats.last_update_objects_frozen),
+    ID2SYM(rb_intern("last_update_objects_alive")), /* => */ LONG2NUM(heap_recorder->stats_last_update.objects_alive),
+    ID2SYM(rb_intern("last_update_objects_dead")), /* => */ LONG2NUM(heap_recorder->stats_last_update.objects_dead),
+    ID2SYM(rb_intern("last_update_objects_skipped")), /* => */ LONG2NUM(heap_recorder->stats_last_update.objects_skipped),
+    ID2SYM(rb_intern("last_update_objects_frozen")), /* => */ LONG2NUM(heap_recorder->stats_last_update.objects_frozen),
   };
   VALUE hash = rb_hash_new();
   for (long unsigned int i = 0; i < VALUE_COUNT(arguments); i += 2) rb_hash_aset(hash, arguments[i], arguments[i+1]);
@@ -525,14 +523,14 @@ static int st_object_record_update(st_data_t key, st_data_t value, st_data_t ext
     // no point checking for liveness or updating its size, so exit early.
     // NOTE: This means that there should be an equivalent check during actual
     //       iteration otherwise we'd iterate/expose stale object data.
-    recorder->stats.last_update_objects_skipped++;
+    recorder->stats_last_update.objects_skipped++;
     return ST_CONTINUE;
   }
 
   if (!ruby_ref_from_id(LONG2NUM(obj_id), &ref)) {
     // Id no longer associated with a valid ref. Need to delete this object record!
     on_committed_object_record_cleanup(recorder, record);
-    recorder->stats.last_update_objects_dead++;
+    recorder->stats_last_update.objects_dead++;
     return ST_DELETE;
   }
 
@@ -580,9 +578,9 @@ static int st_object_record_update(st_data_t key, st_data_t value, st_data_t ext
     record->object_data.is_frozen = RB_OBJ_FROZEN(ref);
   }
 
-  recorder->stats.last_update_objects_alive++;
+  recorder->stats_last_update.objects_alive++;
   if (record->object_data.is_frozen) {
-    recorder->stats.last_update_objects_frozen++;
+    recorder->stats_last_update.objects_frozen++;
   }
 
   return ST_CONTINUE;
--- a/ext/datadog_profiling_native_extension/heap_recorder.c
+++ b/ext/datadog_profiling_native_extension/heap_recorder.c
@@ -565,6 +565,7 @@ static int st_object_record_update(st_data_t key, st_data_t value, st_data_t ext
       RB_FL_SET(ref, RUBY_FL_SEEN_OBJ_ID);
 
       on_committed_object_record_cleanup(recorder, record);
+      recorder->stats_last_update.objects_dead++;
       return ST_DELETE;
     }
 
--- a/ext/datadog_profiling_native_extension/heap_recorder.c
+++ b/ext/datadog_profiling_native_extension/heap_recorder.c
@@ -158,6 +158,13 @@ struct heap_recorder {
 
   // Sampling state
   uint num_recordings_skipped;
+
+  struct stats {
+    size_t last_update_objects_alive;
+    size_t last_update_objects_dead;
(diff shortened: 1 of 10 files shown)


lib/datadog/profiling/exporter.rb
```diff
@@ -52,7 +52,12 @@ def initialize(
...
-        start, finish, compressed_pprof = pprof_recorder.serialize
+        serialization_result = pprof_recorder.serialize
+        if serialization_result.nil?
+          Datadog.logger.debug('Skipped exporting profiling events due to failed serialization')
+          return
+        end
+        start, finish, compressed_pprof, profile_stats = serialization_result
         @last_flush_finish_at = finish
 
         return if compressed_pprof.nil? # We don't want to report empty profiles
```
ivoanjo, line 63: I think these two checks are redundant?

lib/datadog/profiling/exporter.rb
```diff
@@ -52,7 +52,12 @@ def initialize(
 
       def flush
         worker_stats = @worker.stats_and_reset_not_thread_safe
-        start, finish, compressed_pprof = pprof_recorder.serialize
+        serialization_result = pprof_recorder.serialize
+        if serialization_result.nil?
+          Datadog.logger.debug('Skipped exporting profiling events due to failed serialization')
```
ivoanjo, line 57: Minor: Since this is debugging, I guess it's fine, but it does somewhat duplicate the debug message we already print in `StackRecorder#serialize`.

lib/datadog/profiling/exporter.rb
```diff
@@ -75,6 +80,8 @@ def flush
           internal_metadata: internal_metadata.merge(
             {
               worker_stats: worker_stats,
+              profile_stats: profile_stats,
+              recorder_stats: pprof_recorder.stats,
```
ivoanjo, line 84: Minor: I wonder if we should just return the recorder stats from the `serialize` as well?

sig/datadog/profiling/stack_recorder.rbs
```diff
@@ -22,9 +22,16 @@ module Datadog
...
+        ::Time, # Start
+        ::Time, # Finish
+        String, # Serialized PProf
+        ::Hash[::Symbol, untyped], # Hash
+      ]
 
-      def self._native_serialize: (Datadog::Profiling::StackRecorder recorder_instance) -> [:ok | :error, untyped]
+      def serialize: () -> (serialized_profile_data | nil)
+
+      def self._native_serialize: (Datadog::Profiling::StackRecorder recorder_instance) -> ([:ok, serialized_profile_data] | [:error, ::String])
```
ivoanjo, line 34: Minor: This is not quite correct, the signature of `_native_serialize` is only (Time, Time, String)`, e.g. it does not include the stats hash.

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -827,7 +875,7 @@ def create_obj_in_recycled_slot(should_sample_original: false)
           ._native_sample(Thread.current, stack_recorder, metric_values, labels, numeric_labels, 400, false)
 
         # Sanity check: validate that data is there, to avoid the test passing because of other issues
-        sanity_check_samples = samples_from_pprof(stack_recorder.serialize.last)
+        sanity_check_samples = samples_from_pprof(stack_recorder.serialize[-2])
         expect(sanity_check_samples.size).to be 1
```
ivoanjo, line 879: Minor: Lol this is a really roundabout way of writing `.serialize[3]` ;)

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -852,6 +900,119 @@ def create_obj_in_recycled_slot(should_sample_original: false)
...
+        stats.values_at(
+          :serialization_time_ns_min,
+          :serialization_time_ns_max,
+          :serialization_time_ns_avg,
+          :serialization_time_ns_total
+        )
+
+      expect(serialization_time_min).to be <= serialization_time_avg
+      expect(serialization_time_avg).to be <= serialization_time_max
+      expect(serialization_time_total).to eq serialization_time_avg * num_serializations
```
ivoanjo, line 937: I wonder if this may cause some flakiness due to double rounding, e.g. it may be worth adding some `be_within(...)` here maybe?

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -852,6 +900,119 @@ def create_obj_in_recycled_slot(should_sample_original: false)
...
+        Datadog::Profiling::Collectors::Stack::Testing
+          ._native_sample(Thread.current, stack_recorder, { 'alloc-samples' => 1 }, [], [], 400, false)
+      end
+
+      it 'includes heap recorder snapshot' do
+        live_objects = []
+        live_heap_samples = 6
+        live_heap_samples.times do |i|
+          obj = "nice living string #{i}"
+          obj.freeze if i.odd? # Freeze every other string
```
ivoanjo, line 969: Using strings may not be a great idea here as people may start sprinkling `#frozen_string_literal` on files and break this. Perhaps a `Object.new` instead?

ivoanjo (approved):
Left my usual bunch of notes, but I think overall it LGTM :)

ext/datadog_profiling_native_extension/stack_recorder.c
```diff
@@ -570,6 +614,8 @@ void record_sample(VALUE recorder_instance, ddog_prof_Slice_Location locations,
     labels.end_timestamp_ns
   );
 
+  active_slot.stats->recorded_samples++;
```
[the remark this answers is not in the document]
AlexJF: I thought about this. I think at this point and paired with the rest of the internal info we have enough info to breakdown and thus can skip the extra complexity of tracking multiple types of samples:
* Heap samples written: `recorder_stats.heap_recorder_snapshot.last_update_objects_alive`.
* Allocation samples written: `worker_stats.allocation_sampled` 
* CPU/Wall samples: `profile_stats.recorded_samples - (last_update_objects_alive + allocation_sampled)`
And we can revisit in the future?

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -550,6 +578,11 @@ static int st_object_record_update(st_data_t key, st_data_t value, st_data_t ext
     record->object_data.is_frozen = RB_OBJ_FROZEN(ref);
```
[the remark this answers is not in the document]
AlexJF: Eagle-eyed Ivo strikes again! 👍

ext/datadog_profiling_native_extension/stack_recorder.c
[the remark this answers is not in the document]
ivoanjo: I guess if we have the others is this one useful at all? :package: 

I'll leave the decision to you ;)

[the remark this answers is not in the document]
AlexJF: IMO it is since it captures total libdatadog Adds for this profile and without it you can't really arrive at the `CPU/Wall samples` unless we tracked it in a separate metric.

ext/datadog_profiling_native_extension/stack_recorder.c
```diff
@@ -199,7 +221,9 @@ struct call_serialize_without_gvl_arguments {
 
   // Set by callee
   ddog_prof_Profile *profile;
+  stats_slot *stats;
```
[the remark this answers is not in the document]
AlexJF: Yep, slight refactor incoming which includes this.

ext/datadog_profiling_native_extension/stack_recorder.c
```diff
@@ -508,12 +543,20 @@ static VALUE _native_serialize(DDTRACE_UNUSED VALUE _self, VALUE recorder_instan
   // Cleanup after heap recorder iteration. This needs to happen while holding on to the GVL.
   heap_recorder_finish_iteration(state->heap_recorder);
 
+  long serialization_time_ns = monotonic_wall_time_now_ns(DO_NOT_RAISE_ON_FAILURE) - serialization_start_time_ns;
```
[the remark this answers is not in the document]
AlexJF: I don't know, I feel a negative number would be more valuable here as it does not attempt to mask a problem? In some of the tests dealing with 0 samples I've actually seen a few 0 nanosecond times as there's not a lot to do and we'd lose the capability to distinguish between them and timing errors?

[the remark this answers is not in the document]
AlexJF: Ah but if this happens we should not include it in lifetime stats which I think is what you were aiming at. I think that makes sense 👍

lib/datadog/profiling/exporter.rb
AlexJF: Yeah they would only be non-redundant if `compressed_pprof`  could be nil with everything else not nil but it comes from https://github.com/DataDog/dd-trace-rb/blob/0221fbebc9849f7a2b30984b1670eea1eeaf8ad6/ext/datadog_profiling_native_extension/stack_recorder.c?plain=1#L560 which is guaranteed not to be `Qnil`

lib/datadog/profiling/exporter.rb
AlexJF: Feels a bit icky from an API perspective since they are kinda separate from the profile itself?

sig/datadog/profiling/stack_recorder.rbs
AlexJF: But it does? https://github.com/DataDog/dd-trace-rb/blob/0221fbebc9849f7a2b30984b1670eea1eeaf8ad6/ext/datadog_profiling_native_extension/stack_recorder.c?plain=1#L571

spec/datadog/profiling/stack_recorder_spec.rb
AlexJF: Good call!

ext/datadog_profiling_native_extension/stack_recorder.c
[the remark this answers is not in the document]
ivoanjo: Ah. We do actually have this data I think; it's the sum of `CPU_SAMPLES_VALUE` in a profile (`cpu_or_wall_samples` counter).

But to be honest that gets reported in the pprof and we never did anything with it, so I'm not overly attached to it either. (And with timeline it's less useful, since you can recover this information by counting anyway).

ext/datadog_profiling_native_extension/stack_recorder.c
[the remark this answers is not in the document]
ivoanjo: Yup :+1:

sig/datadog/profiling/stack_recorder.rbs
ivoanjo: Aaah you're completely right, I missed it

ext/datadog_profiling_native_extension/ruby_helpers.h
```diff
@@ -82,6 +82,9 @@ NORETURN(
 #define ENFORCE_SUCCESS_HELPER(expression, have_gvl) \
   { int result_syserr_errno = expression; if (RB_UNLIKELY(result_syserr_errno)) raise_syserr(result_syserr_errno, have_gvl, ADD_QUOTES(expression), __FILE__, __LINE__, __func__); }
 
+#define RUBY_NUM_OR_NIL(val, condition, conv) ((val condition) ? conv(val) : Qnil)
+#define RUBY_AVG_OR_NIL(total, count) ((count == 0) ? Qnil : DBL2NUM(((double) total) / count))
```
ivoanjo, line 86: This is really cool, I like it!

ivoanjo (approved):
:+1: LGTM

This looks great! Thanks for the patience with my slowness + the second pass with a bunch of nice small tweaks.

PR #3587, opened 2024-04-12
Add sampling by trace tags

marcotc:
**What does this PR do?**
This PR expands the sampling rule engine to allow for matching on traces based on their tags and metrics.
In `ddtrace`, both tags (string value) and metrics (float value) are called `tags`, but the matcher has a few implementation details that makes it work with both.

**How to test the change?**

There are unit tests in this PR, as well as system tests: https://github.com/DataDog/system-tests/blob/373c58b012da02794385510cb083001f5a0b9013/tests/parametric/test_trace_sampling.py#L302


TonyCTHsu (approved):
LGTM

estherk15 (approved):

PR #3596, opened 2024-04-17
fix(span_links): ensures trace_flags is consistent with the sampling priority

mabdinur:
**What does this PR do?**

Currently the `SpanLink.trace_flags`  is set using `TraceDigest.trace_flags`. The `trace_flag` field in a trace_digest store sampling decisions propagated by tracecontext distributed headers. Therefore for non-remote spans and spans generated via datadog/b3 distributed tracing headers will have a `trace_flag` of `nil` even when a sampling decision exists. 

`SpanLink.trace_flag` must have one of the following values:
  - `nil` If there is NO sampling decision on the linked span
  -  `1` If the linked span is sampled
  - `0` If the linked span is NOT sampled (span will be dropped by the agent)

Since trace_flag is ONLY a function of sampling priority, we can ignore `TraceDigest.trace_flags` and always use sampling_priority to compute the valid when a span link is generated.

### DIFF
--- a/lib/datadog/tracing/span_link.rb
+++ b/lib/datadog/tracing/span_link.rb
@@ -46,7 +46,13 @@ module Datadog
       )
         @span_id = digest&.span_id
         @trace_id = digest&.trace_id
-        @trace_flags = digest&.trace_flags
+        @trace_flags = if digest.nil? || digest.trace_sampling_priority.nil?
+                         nil
+                       elsif digest.trace_sampling_priority > 0
+                         1
+                       else
+                         0
+                       end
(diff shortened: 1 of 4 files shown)


TonyCTHsu (approved):
The code is generally good. 

Question: When would `digest` be `nil`? I thinking about using null object pattern to avoid those `nil` check

mabdinur:
> The code is generally good.
> 
> Question: When would `digest` be `nil`? I thinking about using null object pattern to avoid those `nil` check

Digest shouldn't be nil. I'll make this change

PR #3597, opened 2024-04-17
[NO-TICKET] Remove deprecated profiler settings

ivoanjo:
**What does this PR do?**

This profiler removes a number of profiler settings that have been deprecated for a while.

**Motivation:**

On the 1.x series, we were keeping these settings around, even though they didn't do anything; for the 2.0 release, we're taking the opportunity to delete them, as, again, they don't do anything.

**Additional Notes:**

I've included the needed changes to the upgrade guide as well.

**How to test the change?**

These settings were already not in use, so the only change is really to delete any references to them.

### DIFF
--- a/docs/UpgradeGuide2.md
+++ b/docs/UpgradeGuide2.md
@@ -92,7 +92,7 @@ Datadog.configure do |c|
 end
 ```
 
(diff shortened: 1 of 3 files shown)


marcotc (approved):
🧹

PR #3598, opened 2024-04-17
Add dynamically configurable sampling rules

marcotc:
**What does this PR do?**

`tracing_sampling_rules` is now provided in the dynamic remote configuration payload, and configures the local trace sampling rules.

Also, each sampling rule has a new field, called `provenance`, which allows us to track through the decision-maker tag (`_dd.p.dm`) where is sampling decision came from. This even applies for locally configured rules, which get the default value of `local` provenance.

**How to test the change?**
There are units tests for all changes, plus system tests for integration testing: https://github.com/DataDog/system-tests/pull/2361


TonyCTHsu (approved):
Since I have already compile the changelog for 2.0 in 2.0.0.beta1. Could you add this entry under the [Unreleased] section at Changelog.md.

marcotc:
@TonyCTHsu would you like me to change https://github.com/DataDog/dd-trace-rb/blob/master/CHANGELOG.md directly in this PR?

PR #3602, opened 2024-04-19
Remove obsolete matrix

TonyCTHsu:
**What does this PR do?**

Update appraisal groups definition and tasks to remove obsolete ones.


ivoanjo (approved):
:+1: LGTM, thanks for the nice clean-up

marcotc (approved):
Thank you so much!
It's very hard to inspect if the changes to the Rakefile did not exclude any tests that it we not mean to, but the coverage report tells us that nothing was missed!

PR #3610, opened 2024-04-23
[PROF-9650] Enable endpoint profiling for Sidekiq and similar background job processors

ivoanjo:
**What does this PR do?**

We have an opt-in list for which kinds of traces we collect the resource from; this PR extends that list with the `worker` type used by background consumers such as Sidekiq.

image

**Motivation:**

Enable profiling for Sidekiq and similar background job processors.

**Additional Notes:**

We also collect a few Sidekiq internal things (such as `sidekiq.job_fetch`, `sidekiq.scheduled_poller_wait`, `sidekiq.scheduled_push`, `sidekiq.stop` and `sidekiq.heartbeat`).

We discussed if it would make sense to hide them or not, but for now we decided to still show them.

**How to test the change?**

This change includes test coverage; I also booted up sidekiq and tested manually using the following script:

```ruby
 # Running redis:
 # * `docker-compose up redis`
 # Running server:
 # * `DD_SERVICE=sidekiq-testing DD_ENV=staging DD_PROFILING_ENABLED=true bundle exec sidekiq -r ./sidekiq.rb`
 # Running client:
 # * `DD_TRACE_ENABLED=false DD_TRACING_ENABLED=false be ddprofrb exec irb -r ./sidekiq.rb`
 # * `ThisIsASidekiqJob.perform_async("hello", 10)`

require "sidekiq"
require 'datadog'

Datadog.configure { |c| c.tracing.instrument :sidekiq }

class ThisIsASidekiqJob
  include Sidekiq::Job

  def perform(some_arg = "potato", how_long = 1)
    puts "Start of job!"
    sleep how_long
    puts "End of job!"
  end
end
```

### DIFF
--- a/ext/datadog_profiling_native_extension/collectors_thread_context.c
+++ b/ext/datadog_profiling_native_extension/collectors_thread_context.c
@@ -217,7 +217,7 @@ static long thread_id_for(VALUE thread);
 static VALUE _native_stats(VALUE self, VALUE collector_instance);
 static VALUE _native_gc_tracking(VALUE self, VALUE collector_instance);
 static void trace_identifiers_for(struct thread_context_collector_state *state, VALUE thread, struct trace_identifiers *trace_identifiers_result);
-static bool should_collect_resource(VALUE root_span_type);
+static bool should_collect_resource(VALUE root_span);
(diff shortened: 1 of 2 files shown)


AlexJF (approved):

zachmccormick:
Looking forward to using this :)

PR #3614, opened 2024-04-30
Fix macro benchmark 

TonyCTHsu:
**What does this PR do?**

1. Fix marco benchmarking apps by injecting gemfiles
2. Reduce the amount of variants
3. Fix environment variables

### DIFF
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -30,6 +30,11 @@ variables:
     # Benchmark's env variables. Modify to tweak benchmark parameters.
     DD_TRACE_DEBUG: "false"
     DD_RUNTIME_METRICS_ENABLED: "true"
+    # Gitlab makes use of the rugged gem, which triggers the automatic no signals workaround use, see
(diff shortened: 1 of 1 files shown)


ddyurchenko (approved):
Looks good :+1: , need to change to `ruby/gitlab` before merge though. I would also ask @ivoanjo about changes to profiler job definitions, so there is no accidental breakage.

.gitlab/benchmarks.yml
```diff
@@ -65,92 +73,50 @@ only-tracing:
   extends: .benchmarks
   variables:
     DD_BENCHMARKS_CONFIGURATION: only-tracing
+    DD_TRACE_ENABLED: "true"
 
 only-profiling:
   extends: .benchmarks
   variables:
     DD_BENCHMARKS_CONFIGURATION: only-profiling
     DD_PROFILING_ENABLED: "true"
```
ddyurchenko, line 82: Originally it was

```
echo -e "\ngem 'dogstatsd-ruby'\ngem 'ddtrace', $TARGET_VERSION\n" >> /opt/gitlab/embedded/service/gitlab-rails/Gemfile
```

without auto_instrument.

TonyCTHsu: Profiling changes were discussed in slack thread.

PR #3616, opened 2024-04-30
Add sampling rule glob pattern support

marcotc:
**What does this PR do?**
This PR adds support for matching sample rules not only by simple strings, but also buy a simple matching pattern. The matching is also case insensitive now.

This brings there will be tracer to parody with other tracing languages.

The following special characters are supported:
 - `?`: matches any single character
 - `*`: matches any substring

**Additional Notes:**
Support for an arbitrary matcher your object has been removed, given the sampling APIs are now private and does not need such flexibility.

The glob pattern is the same for trace sampling and span sampling, so the code was refractored to share this complex matcher.

**How to test the change?**
They are unit tests and system tests for this change: https://github.com/DataDog/system-tests/pull/2415


lib/datadog/tracing/sampling/ext.rb
```diff
@@ -50,9 +50,9 @@ module Decision
           # Single Span Sampled.
           SPAN_SAMPLING_RATE = '-8'
           # Dynamically configured rule, explicitly created by the user.
-          REMOTE_USER_RULE = '-10'
+          REMOTE_USER_RULE = '-11'
           # Dynamically configured  rule, automatically generated by Datadog.
-          REMOTE_DYNAMIC_RULE = '-11'
+          REMOTE_DYNAMIC_RULE = '-12'
```
marcotc, line 55: The RFC was recently changed, and these numbers have been updated.

lib/datadog/tracing/sampling/matcher.rb
```diff
@@ -59,27 +105,10 @@ def tags_match?(trace)
...
 
-            matcher === tag
+            matcher.match?(tag)
           end
         end
       end
-
-      # A {Datadog::Tracing::Sampling::Matcher} that allows for arbitrary trace matching
-      # based on the return value of a provided block.
-      class ProcMatcher < Matcher
```
marcotc, line 69: Arbitrary object matching is not required anymore, that this helper is no longer needed.

TonyCTHsu (approved):

PR #3628, opened 2024-05-06
[PROF-9476] Managed string storage PoC

AlexJF:
**What does this PR do?**
Make use of new string storage APIs introduced in https://github.com/DataDog/libdatadog/pull/414 to prevent having to keep a lot of strings being duplicated and managed by heap recorder.

This is currently a PoC and not production ready.

**Motivation:**

A bottleneck analysis using Datadog's native profiler has shown that stacktrace management in heap recorder is one of the biggest bottlenecks with heap profiling:

Picture 1
Picture 2

```
Stacktrace storage in heap recorder (copying, hashing, cleanup)
~4% of entire CPU profile
~30% of the overhead during sampling
~16.3% of the overhead during flushing
```

This work is similar to that being done already in libdatadog and thus is work we can potentially remove entirely from the equation if libdatadog exposed APIs that allowed stacktrace storage to survive across profiles.

This first implementation accomplishes this solely on the strings table but could potentially be expanded to the entire stacktrace. Nevertheless, it already shows some great promise by allowing sampleRate=1 with negligible overhead compared to the currently merged implementation with sampleRate=10 in our test app:

image

image

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/ext/datadog_profiling_native_extension/heap_recorder.c
+++ b/ext/datadog_profiling_native_extension/heap_recorder.c
@@ -19,8 +19,8 @@
 
 // A compact representation of a stacktrace frame for a heap allocation.
 typedef struct {
-  char *name;
-  char *filename;
+  u_int32_t name;
+  u_int32_t filename;
   int32_t line;
 } heap_frame;
 static st_index_t heap_frame_hash(heap_frame*, st_index_t seed);
@@ -35,54 +35,20 @@ typedef struct {
   uint16_t frames_len;
   heap_frame frames[];
 } heap_stack;
-static heap_stack* heap_stack_new(ddog_prof_Slice_Location);
-static void heap_stack_free(heap_stack*);
+static heap_stack* heap_stack_new(heap_recorder*, ddog_prof_Slice_Location);
+static void heap_stack_free(heap_recorder*, heap_stack*);
 static st_index_t heap_stack_hash(heap_stack*, st_index_t);
 
 #if MAX_FRAMES_LIMIT > UINT16_MAX
   #error Frames len type not compatible with MAX_FRAMES_LIMIT
 #endif
 
-enum heap_record_key_type {
-  HEAP_STACK,
-  LOCATION_SLICE
+static int heap_stack_cmp_st(st_data_t, st_data_t);
+static st_index_t heap_stack_hash_st(st_data_t);
+static const struct st_hash_type st_hash_type_heap_stack = {
+    heap_stack_cmp_st,
+    heap_stack_hash_st,
 };
-// This struct allows us to use two different types of stacks when
-// interacting with a heap_record hash.
-//
-// The idea is that we'll always want to use heap_stack-keys when
-// adding new entries to the hash since that's the compact stack
-// representation we rely on internally.
-//
-// However, when querying for an existing heap record, we'd save a
-// lot of allocations if we could query with the
-// ddog_prof_Slice_Location we receive in our external API.
-//
-// To allow this interchange, we need a union and need to ensure
-// that whatever shape of the union, the heap_record_key_cmp_st
-// and heap_record_hash_st functions return the same results for
-// equivalent stacktraces.
-typedef struct {
-  enum heap_record_key_type type;
-  union {
-    // key never owns this if set
-    heap_stack *heap_stack;
-    // key never owns this if set
-    ddog_prof_Slice_Location *location_slice;
-  };
-} heap_record_key;
-static heap_record_key* heap_record_key_new(heap_stack*);
-static void heap_record_key_free(heap_record_key*);
-static int heap_record_key_cmp_st(st_data_t, st_data_t);
-static st_index_t heap_record_key_hash_st(st_data_t);
-static const struct st_hash_type st_hash_type_heap_record_key = {
-    heap_record_key_cmp_st,
-    heap_record_key_hash_st,
-};
-
-// Need to implement these functions to support the location-slice based keys
-static st_index_t ddog_location_hash(ddog_prof_Location, st_index_t seed);
-static st_index_t ddog_location_slice_hash(ddog_prof_Slice_Location, st_index_t seed);
 
 // A heap record is used for deduping heap allocation stacktraces across multiple
 // objects sharing the same allocation location.
@@ -93,7 +59,7 @@ typedef struct {
   heap_stack *stack;
 } heap_record;
 static heap_record* heap_record_new(heap_stack*);
-static void heap_record_free(heap_record*);
+static void heap_record_free(heap_recorder*, heap_record*);
 
 // An object record is used for storing data about currently tracked live objects
 typedef struct {
@@ -102,8 +68,8 @@ typedef struct {
   live_object_data object_data;
 } object_record;
 static object_record* object_record_new(long, heap_record*, live_object_data);
-static void object_record_free(object_record*);
-static VALUE object_record_inspect(object_record*);
+static void object_record_free(heap_recorder*, object_record*);
+static VALUE object_record_inspect(heap_recorder*, object_record*);
 static object_record SKIPPED_RECORD = {0};
 
 // A wrapper around an object record that is in the process of being recorded and was not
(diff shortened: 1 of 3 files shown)


ext/datadog_profiling_native_extension/heap_recorder.h
```diff
@@ -59,7 +59,7 @@ typedef struct {
 } heap_recorder_iteration_data;
 
 // Initialize a new heap recorder.
-heap_recorder* heap_recorder_new(void);
+heap_recorder* heap_recorder_new(const ddog_prof_ManagedStringStorage *string_storage);
```
ivoanjo, line 62: Minor: Doesn't `const ddog_prof_ManagedStringStorage *` mean the data being pointed to is immutable? :thinking:

Because I'm pretty sure we're mutating it via the libdatadog helpers

ext/datadog_profiling_native_extension/stack_recorder.c
```diff
@@ -193,6 +193,10 @@ struct stack_recorder_state {
   pthread_mutex_t mutex_slot_two;
   profile_slot profile_slot_two;
 
+  ddog_prof_ManagedStringStorage string_storage;
+  uint32_t label_key_allocation_class;
+  uint32_t label_key_gc_gen_age;
```
ivoanjo, line 198: Minor: It occurs to me that it may be nice for libdatadog to specify a typedef (of a one-item-struct) to represent this, so we get a better type than "a naked integer".

ext/datadog_profiling_native_extension/stack_recorder.c
```diff
@@ -1057,3 +1040,17 @@ static VALUE build_profile_stats(profile_slot *slot, long serialization_time_ns,
...
   return stats_as_hash;
 }
+
+static u_int32_t intern_or_raise(struct stack_recorder_state *state, const char *str) {
+  if (str == NULL) {
+    return 0;
+  }
+  ddog_CharSlice char_slice;
+  char_slice.len = strlen(str);
+  char_slice.ptr = str;
```
ivoanjo, line 1050: Minor: It may be worth introducing a macro version of this to allow the compiler to hardcode the strlen at compile-time for static strings, like `DDOG_CHARSLICE_C` does.

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -635,8 +598,10 @@ static int st_object_records_iterate(DDTRACE_UNUSED st_data_t key, st_data_t val
     locations[i] = (ddog_prof_Location) {
       .mapping = {.filename = DDOG_CHARSLICE_C(""), .build_id = DDOG_CHARSLICE_C("")},
       .function = {
-        .name = {.ptr = frame->name, .len = strlen(frame->name)},
-        .filename = {.ptr = frame->filename, .len = strlen(frame->filename)},
+        .name = DDOG_CHARSLICE_C(""),
+        .name_id = frame->name,
+        .filename = DDOG_CHARSLICE_C(""),
+        .filename_id = frame->filename,
```
ivoanjo, line 604: I guess we'll see how this evolves on the libdatadog but I still think it's really weird/bizarre/meh that libdatadog doesn't allow us to just use NULL for strings in situations where they can be empty, as this looks a lot more awkward because of it.

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -864,53 +818,24 @@ int heap_frame_cmp(heap_frame *f1, heap_frame *f2) {
...
-  return st_hash(char_slice.ptr, char_slice.len, seed);
-}
-
-// WARN: Must be kept in-sync with ::ddog_location_hash
 st_index_t heap_frame_hash(heap_frame *frame, st_index_t seed) {
-  st_index_t hash = string_hash(frame->name, seed);
-  hash = string_hash(frame->filename, hash);
+  st_index_t hash = st_hash(&frame->name, sizeof(frame->name), seed);
+  hash = st_hash(&frame->filename, sizeof(frame->filename), hash);
   hash = st_hash(&frame->line, sizeof(frame->line), hash);
```
ivoanjo, line 831: Minor: I guess we could probably hash the whole frame object in one now?

E.g. `st_hash(frame, sizeof(frame), ...)`?

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -923,8 +848,8 @@ heap_stack* heap_stack_new(ddog_prof_Slice_Location locations) {
   for (uint16_t i = 0; i < stack->frames_len; i++) {
     const ddog_prof_Location *location = &locations.ptr[i];
     stack->frames[i] = (heap_frame) {
-      .name = string_from_char_slice(location->function.name),
-      .filename = string_from_char_slice(location->function.filename),
+      .name = intern_or_raise(recorder, &location->function.name),
+      .filename = intern_or_raise(recorder, &location->function.filename),
```
ivoanjo, line 852: I guess for the PoC (or even for v1?) this is enough, although conceptually this could be done way earlier, e.g. in the stack recorder, since by definition we know all of these strings must have already been interned, because they've been sampled by the allocation profiler.

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -933,16 +858,15 @@ heap_stack* heap_stack_new(ddog_prof_Slice_Location locations) {
...
-void heap_stack_free(heap_stack *stack) {
-  for (uint64_t i = 0; i < stack->frames_len; i++) {
-    heap_frame *frame = &stack->frames[i];
-    ruby_xfree(frame->name);
-    ruby_xfree(frame->filename);
+void heap_stack_free(heap_recorder *recorder, heap_stack *stack) {
+  for (u_int16_t i = 0; i < stack->frames_len; i++) {
+    unintern_or_raise(recorder, stack->frames[i].filename);
+    unintern_or_raise(recorder, stack->frames[i].name);
   }
```
ivoanjo, line 865: Minor: More of a naming thing, but I had to check exactly what `unintern` does on the libdatadog side, since I was thinking "wait, does this mean we're deleting the string right now?"

Maybe release or something like that may be a bit more explicit? (Perhaps even decrement_usage?)

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -951,112 +875,30 @@ st_index_t heap_stack_hash(heap_stack *stack, st_index_t seed) {
...
+      return ((int) frame1->filename) - ((int) frame2->filename);
     }
-    int filename_cmp_result = strncmp(filename1, filename2, filename1_len);
-    if (filename_cmp_result != 0) {
-      return filename_cmp_result;
+
+    if (frame1->line != frame2->line) {
+      return ((int) frame1->line) - ((int)frame2->line);
     }
   }
```
ivoanjo, line 903: As an optimization here, now that this is all just a bunch of numbers, maybe we can just `memcmp` the two stacks? (I think this would be a stable comparison similar to the  current implementation?)

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -933,16 +858,15 @@ heap_stack* heap_stack_new(ddog_prof_Slice_Location locations) {
...
+    unintern_or_raise(recorder, stack->frames[i].name);
   }
+
   ruby_xfree(stack);
 }
 
-// WARN: Must be kept in-sync with ::ddog_location_slice_hash
 st_index_t heap_stack_hash(heap_stack *stack, st_index_t seed) {
   st_index_t hash = seed;
   for (uint64_t i = 0; i < stack->frames_len; i++) {
```
ivoanjo, line 872: Now that this is all a bunch of numbers, I wonder if we could hash it all in one go (rather than going frame by frame).

ivoanjo (commented):
Left a bunch of notes/ideas :)

ext/datadog_profiling_native_extension/heap_recorder.c
ivoanjo: Ah no, I just realized that the managed string storage is different from the regular profile string storage so no, actually, these strings have not been interned already (but they could be...)

ext/datadog_profiling_native_extension/heap_recorder.c
AlexJF: Yes this along with the cmp version are great points!

ivoanjo:
Closing in favor of https://github.com/DataDog/dd-trace-rb/pull/4331 (finally! haha!)

PR #3629, opened 2024-05-06
Add apm-idm-ruby as CODEOWNERS to contrib

marcotc:
**Motivation:**
The IDM team would like to have a clean notification inbox when it comes to code reviews in `dd-trace-rb`. This is being done across all languages.

**How to test the change?**


ivoanjo (commented):
So the internet is a bit confusing about this, but I suspect you may need a trailing `/` on those paths. The docs aren't clear on what exactly happens when you don't, and codeowners does sometimes have sharp edges where it doesn't actually match the whole subtree.

Issue #3581, opened 2024-04-10
ActiveRecord instrumentation is causing config resolve on every query


ngan:
**Current behaviour**
I was playing around with [`ActiveRecord::DatabaseConfigurations.register_db_config_handler`](https://edgeapi.rubyonrails.org/classes/ActiveRecord/DatabaseConfigurations.html#method-c-register_db_config_handler) and noticed an odd behavior in console. Every time I make a query to the database (eg. `User.first`), I see the proc that I've define get run. This means that every query means that the database configs is resolved and a brand new (redundant) `ActiveRecord::DatabaseConfigurations::HashConfig` object is built. This seems very inefficient.

`ddtrace` is resolving the connection for every `sql.active_record` event:
https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/active_record/configuration/resolver.rb#L122

**Expected behaviour**
I would expect that the configuration should have already been resolved by Rails and ddtrace would just get the information given that it gets the `connection` as part of the payload: https://guides.rubyonrails.org/active_support_instrumentation.html#sql-active-record

Or ddtrace should at least memoize the config. There shouldn't need to be a new to resolve every time and therefore generate a new configuration object ever time. If someone uses `register_db_config_handler` to do some more involved, this could have drastic performance costs.

**Steps to reproduce**

In a plain Rails app with a least 1 model, do `bundle open activerecord` then add a `puts` or whatever here:
https://github.com/rails/rails/blob/main/activerecord/lib/active_record/database_configurations.rb#L69

Then boot up console and perform a query through the model, eg `User.first`. You'll see it enter that code path everytime. In a plain Rails app, without `ddtrace` installed, this isn't the case.

* **ddtrace version:** 1.21.1
* **Configuration block (`Datadog.configure ...`):**
* **Ruby version:** 3.3.0
* **Operating system:** Linux

ngan:
Here's a reproduction script. You'll need to create a mysql database called `test`.

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  # If you want to test against edge Rails replace the previous line with this:
  # gem "rails", github: "rails/rails", branch: "main"

  gem "mysql2"

  gem "ddtrace"
end

require "ddtrace"
require "active_record"
require "active_record/database_configurations"
require "minitest/autorun"
require "logger"

Datadog.configure do |config|
[...]

PR #3630, opened 2024-05-08
Cache ActiveRecord configuration resolver

marcotc:
**What does this PR do?**
This PR adds caching to the ActiveRecord configuration resolver, resulting in only the first request with a specific ActiveRecord configuration performing the Rails resolution lookup.

This virtually eliminates the for ActiveRecord configuration resolution. This will improve the performance on every ActiveRecord span.

I can provide better performance numbers when this change has had a chance to run in our internal testing environment.

**Motivation:**
#3581

**How to test the change?**
All changes are covered by testing.

### DIFF
--- a/lib/datadog/tracing/contrib/configuration/resolver.rb
+++ b/lib/datadog/tracing/contrib/configuration/resolver.rb
@@ -80,16 +80,16 @@ module Datadog
           end
         end
 
-        # The {CachedResolver} is a mixin that provides caching functionality to the {Resolver} class.
+        # The {CachingResolver} is a mixin that provides caching functionality to the {Resolver} class.
         # This is useful when {Resolver#resolve} values that are expensive to compute.
         # This is a size-limited, FIFO cache.
         #
         # @example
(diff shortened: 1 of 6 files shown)


lib/datadog/tracing/contrib/configuration/resolver.rb
```diff
@@ -79,6 +79,49 @@ def parse_matcher(matcher)
...
+        # The {CachedResolver} is a mixin that provides caching functionality to the {Resolver} class.
+        # This is useful when {Resolver#resolve} values that are expensive to compute.
+        # This is a size-limited, FIFO cache.
+        #
+        # @example
+        #   class MyResolver < Datadog::Tracing::Contrib::Configuration::Resolver
+        #     prepend Datadog::Tracing::Contrib::Configuration::CachedResolver
+        #     # ...
+        #   end
+        module CachedResolver
```
p-datadog, line 92: I would call this `CachingResolver` because the resolver is what is doing the caching, not the thing that is being cached.

p-datadog (approved):

ngan:
🙏 thank you @marcotc

PR #3632, opened 2024-05-08
Ensure micro benchmarks always run

marcotc:
**What does this PR do?**
Micro benchmarks haven't been running recently because their GitLab job is set to `when: on_success`, which depends on every proceeding GitLab job to succeeded.

This is problematic when there are optional GitLab jobs that can fail.


ivoanjo (approved):
:+1: Seems reasonable

PR #3636, opened 2024-05-10
DEBUG-2334 Change Dispatcher constructor to directly accept receivers

p-datadog:
**What does this PR do?**

There is only one call site of Dispatcher and the caller immediately copies the receivers into the newly created dispatcher.

Instead pass the dispatchers directly into the constructor to make it clear that dispatchers are part of initialization.

**Motivation:**

- More clearly convey what is needed to initialize dispatcher
- Encapsulate data better (no more mutating receivers)

**How to test the change?**
Unit tests

### DIFF
--- a/lib/datadog/core/remote/dispatcher.rb
+++ b/lib/datadog/core/remote/dispatcher.rb
@@ -7,8 +7,8 @@ module Datadog
       class Dispatcher
         attr_reader :receivers
 
-        def initialize
-          @receivers = []
+        def initialize(receivers)
+          @receivers = receivers
         end
 
         def dispatch(changes, repository)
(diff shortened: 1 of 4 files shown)


lib/datadog/core/remote/dispatcher.rb
```diff
@@ -7,8 +7,8 @@ module Remote
       class Dispatcher
         attr_reader :receivers
 
-        def initialize
-          @receivers = []
+        def initialize(receivers)
+          @receivers = receivers
```
lloeki, line 11: The main difference here is that before the receiver list is guaranteed to be owned by the `Dispatcher` instance, whereas it is not anymore afterwards.

The only place it is used though is upon `Client#initialize` so the change is fair.

lloeki (approved):

PR #3638, opened 2024-05-10
Fix bundle add mechanism 

TonyCTHsu:
**What does this PR do?**

* Fix adding gem mechanism in host injection by invoking with `Bundler::CLI::Add`. 
* This reduce the overhead that previously need to resolved multiple times for each dependency.

### DIFF
--- a/lib-injection/host_inject.rb
+++ b/lib-injection/host_inject.rb
@@ -73,4 +73,5 @@ begin
   ).run
 rescue Exception => e
   warn "[datadog] Injection failed: #{e.class.name} #{e.message}\nBacktrace: #{e.backtrace.join("\n")}\n#{support_message}"
+  ENV['DD_TRACE_SKIP_LIB_INJECTION'] = 'true'
 end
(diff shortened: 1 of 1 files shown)


lib-injection/host_inject.rb
```diff
@@ -59,60 +60,17 @@ def debug_log(msg)
...
-          end.join("\n")}\n#{support_message}"
-        end
-      ensure
-        # Remove the copies
-        ::FileUtils.rm datadog_gemfile
-        ::FileUtils.rm datadog_lockfile
-      end
-    end
-  end
+  Bundler::CLI::Add.new(
```
delner, line 63: What does this do? Is it really equivalent to all that code above?

TonyCTHsu: This is what `bundle add datadog --version "2.0.0" --skip-install  --strict` does. 

It improves the previous implementation by 

1. Resolving once
2. Resolving within the process

marcotc (approved):
I've discussed with @TonyCTHsu and we get an equivalent outcome with the new implementation.

PR #3639, opened 2024-05-10
Add hanami app to integration CI configurations

p-datadog:
l

**What does this PR do?**

Adds hanami to CI configurations, as requested in https://github.com/DataDog/dd-trace-rb/pull/2546.

Several things needed to be fixed for the tests to pass:

1. https://github.com/DataDog/dd-trace-rb/pull/2546
2. There was an empty `hanami.rb` which looks to have been intended to mock Hanami framework (?). The app requires the actual hanami framework to function and this empty file was loaded instead of hanami itself. I removed the file.
3. Hanami integration app was missing unicorn configuration, which the test suite was expecting. I copied it from sinatra.
4. The controller in the test app was written to fail 50% of the time, presumably to provide an environment to test failure. I commented out the failure generation so that CI would pass every time. 

I also made the image build script work when ruby version isn't given and to provide a help message so that someone new to this area of the code could actually get it running locally.

**Motivation:**

Review of https://github.com/DataDog/dd-trace-rb/pull/2546.

**Test plan**
The tests are running on Ruby 2.6 and 2.7 only. Hanami 1 (1.3.5, more exactly) that is currently being used in these tests is not compatible with Ruby 3.


TonyCTHsu (approved):

TonyCTHsu (approved):

p-datadog:
I ran all of the tests in https://github.com/DataDog/dd-trace-rb/pull/3690.

PR #3640, opened 2024-05-10
DEBUG-2334 Add Debugger component

p-datadog:
This adds a component for Dynamic Instrumentation / Live Debugger using AppSec component as the model.

Included is a 'debugger' setting to turn the component on, and handling of DD_DYNAMIC_INSTRUMENTATION_ENABLED environment variable which is equivalent to the setting.

The component is not used by anything at this point and is/will remain off by default.

**2.0 Upgrade Guide notes**

**How to test the change?**
Unit tests only at this time.


marcotc (approved):
Great, clean bootstrapping!

PR #3643, opened 2024-05-14
Implement `latest_major` and `lastest_minor` tag for docker init container

TonyCTHsu:
**What**

Currently, images for docker init container are tagged with the semantic version (ie: `v2.1.1`) and `latest`. 

This PR would add 2 more tags

1. latest major, ie: `v2`
2. latest minor, ie: `v2.1`

### DIFF
--- a/.gitlab/check_gem_presence.rb
+++ b/.gitlab/check_gem_presence.rb
@@ -1,0 +1,37 @@
+#!/usr/bin/env ruby
+
+require 'bundler/inline'
+
+gemfile { gem 'gems', source: 'https://rubygems.org' }
+
+require 'rubygems'
+require 'gems'
+
+version = ARGV[0].chomp
+version = version.delete_prefix('v') if version.start_with?('v')
+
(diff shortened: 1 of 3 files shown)


anmarchenko (approved):

.gitlab/check_gem_presence.rb
```diff
@@ -0,0 +1,37 @@
...
+require 'gems'
+
+version = ARGV[0].chomp
+version = version.delete_prefix('v') if version.start_with?('v')
+
+candidate = Gem::Version.new(version)
+
+retry_count = 0
+max_retries = 60
+interval = 60
```
marcotc, line 17: Do we need wait a whole minute before retrying?
I ask in case `.gitlab/check_gem_presence.rb` is ever something a human is actively waiting on. If so, I suggest reducing the interval to something like 10 seconds (and increasing `max_retries` to match).

TonyCTHsu: We can change this frequency in the future.

PR #3646, opened 2024-05-14
Check require encapsulation of components 

p-datadog:
**What does this PR do?**
This PR adds tests that requiring individual components is successful. It also adds some requires that were missing in `appsec` to make the tests pass.

**Motivation:**
Troubleshooting configuration for debugging which I copied from appsec, appsec itself does not load cleanly.

**Additional Notes:**

**How to test the change?**

Added a unit test and a GH configuration to run it (since it needs a clean environment).

### DIFF
--- a/.github/workflows/require-encapsulation.yml
+++ b/.github/workflows/require-encapsulation.yml
@@ -1,0 +1,9 @@
+name: Check require encapsulation
+on: [push]
+jobs:
+  check-require-encapsulation:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - run: rm -f .rspec
+      - run: bundle exec rspec spec/require_encapsulation_check.rb
--- a/.github/workflows/require-encapsulation.yml
+++ b/.github/workflows/require-encapsulation.yml
@@ -1,5 +1,13 @@
 name: Check require encapsulation
-on: [push]
+
+on:
+  push:
+    branches:
+      - "**"
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+
 jobs:
   check-require-encapsulation:
     runs-on: ubuntu-latest
--- a/.github/workflows/require-encapsulation.yml
+++ b/.github/workflows/require-encapsulation.yml
@@ -13,5 +13,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.2'
+          bundler-cache: true # runs 'bundle install' and caches installed gems automatically
       - run: rm -f .rspec
       - run: bundle exec rspec spec/require_encapsulation_check.rb
--- a/lib/datadog/appsec/contrib/sinatra/integration.rb
+++ b/lib/datadog/appsec/contrib/sinatra/integration.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_relative '../../../tracing/contrib'
 require_relative '../integration'
 
 require_relative 'patcher'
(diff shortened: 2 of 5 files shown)


.github/workflows/require-encapsulation.yml
```diff
@@ -0,0 +1,21 @@
...
+  check-require-encapsulation:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: '3.2'
+          bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+      - run: rm -f .rspec
+      - run: bundle exec rspec spec/require_encapsulation_check.rb
```
lloeki, line 21: What's the rationale for a separate job?

If the file is named correctly it can be part of the usual `rake spec` tasks and tested across the matrix.

spec/require_encapsulation_check.rb
```diff
@@ -0,0 +1,31 @@
+require 'English'
+REQUIRES = %w[
+  datadog/appsec
+  datadog/core
+  datadog/kit
+  datadog/profiling
+  datadog/tracing
+].freeze
```
lloeki, line 8: This constant will be leaking globally, as is it should probably go into the `RSpec.describe` block.

Could even be a plain local variable, or even just `%w[ ... ].each do |req|`.

spec/require_encapsulation_check.rb
```diff
@@ -0,0 +1,31 @@
...
+
+RSpec.describe 'require encapsulation' do
```
lloeki, line 10: In a way I feel this description isn't quite correct, which led me to think that this spec isn't quite how we expect it to be.

Notably it puts some cross-component knowledge into that spec file. I think it might be better if each "require "works is isolated in each separate component:

- assert that `require datadog/core` works => in `spec/datadog/core_spec.rb`
- assert that `require datadog/appsec` works => in `spec/datadog/appsec_spec.rb`
- ... and so on

lib/datadog/appsec/contrib/sinatra/integration.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_relative '../../../tracing/contrib'
```
lloeki, line 3: I'm not sure about this `require` here: this integration file doesn't itself seem to need it.

lib/datadog/appsec/contrib/sinatra/patcher.rb
```diff
@@ -11,6 +11,7 @@
 require_relative 'gateway/route_params'
 require_relative 'gateway/request'
 require_relative '../../../tracing/contrib/sinatra/framework'
+require_relative '../../../tracing/contrib'
```
lloeki, line 14: That's a good catch!

But `AppSec` should be able to load without `Tracing` being present, so I think instead there's a bit of conditioning around the `Tracing` constants to be made below.

spec/require_encapsulation_check.rb
```diff
@@ -0,0 +1,31 @@
...
+    # Permit Datadog::VERSION to be defined but no other constants.
+    # See the note in gemspec about requiring 'datadog/version'.
+    expect(defined?(Datadog)).to eq 'constant'
+    expect(Datadog.constants).to eq([:VERSION])
+  end
+
+  REQUIRES.each do |req|
+    context req do
+      it 'loads' do
+        pid = fork do
```
lloeki, line 21: `fork` isn't available on Windows (which we don't support.. yet. I think it's a worthy goal) but more importantly it's not available on JRuby, for which we want coverage as well.

spec/require_encapsulation_check.rb
```diff
@@ -0,0 +1,31 @@
+require 'English'
```
lloeki, line 1:
```
require 'English'
```

That's in order to have `$CHILD_STATUS` instead of `$?`?`

lloeki (changes requested):
Good catches! But I think it needs some changes.

p-datadog:
New PR: https://github.com/DataDog/dd-trace-rb/pull/3738

PR #3647, opened 2024-05-14
Add more 2.x deprecation warnings

delner:
**What does this PR do?**

Adds some more deprecation warnings for changes outlined in [the upgrade guide](https://github.com/DataDog/dd-trace-rb/blob/master/docs/UpgradeGuide2.md). Includes:

- `Use of non-strings for 'env' configuration is deprecated. Use a string instead.`
- `Use of non-strings for 'service' configuration is deprecated. Use a string instead.`

**Motivation:**

To help users more easily through the 2.x upgrade process. These warnings should make it easier for users to identify breaking changes on existing 1.x deployments, and update their code in a 1.x/2.x compatible way, before undertaking a 2.x upgrade.

**Additional Notes:**

There are many changes in the upgrade guide. Some may be worth warning about, others may be more trouble than they are worth. I'm mostly focusing on actionable.

We also should be careful with how verbose some of these warnings could be, especially when they are in the instrumentation path (and could be triggered many times.) Perhaps we should add some more explicit "print once" behavior?


delner:
I'd like to rebase this on the changes implemented in #3675, so that these deprecation warnings can utilize the limit feature.

p-datadog (approved):
This LGTM

delner:
I've rebased the changes on top of the "log limit" functionality backported via #3780. This should prevent verbosity on the configuration path.

We could go further with changes here, but I think I want to stop here for now; I need to move onto other changes, and want to get this merged before it gets either too large or stale.

Should be ready for review.

ivoanjo:
Hey @delner is this still relevant?

p-datadog:
I will mark this PR as draft to clean up the list of PRs needing reviews since CI is red and I am not familiar with the 1.x branch to look into those failures at the moment.

ivoanjo:
I don't think we're gonna pick up this PR at this point. So I'll go ahead and close it -- cc @delner please do reopen if you think it's still worth taking a stab at it.

PR #3648, opened 2024-05-14
DEBUG-2334 rename "debugger" to "debugging" 

p-datadog:
DEBUG-2334 rename "debugger" to "debugging" because "debugger" is used by local debuggers
    
   E.g. byebug defines a method called "debugger" in all scopes which conflicts
   with the live debugging implementation.


marcotc (approved):

p-datadog:
This PR no longer applies cleanly due to the first PR being reverted out of master. Also due to the revert, I think it does not make sense to have a PR with a rename - the new naming could be added as a single change.

PR #3656, opened 2024-05-21
Update `rexml` 

TonyCTHsu:
**What does this PR do?**

Update `rexml` due to [CVE-2024-35176](https://www.ruby-lang.org/en/news/2024/05/16/dos-rexml-cve-2024-35176/)


p-datadog:
Behavior of StringScanner#scan changed for ruby 2.7 to accept a string. The following is accepted by 2.7 but not 2.6:

```
irb(main):003:0> require'strscan'
=> true
irb(main):004:0> StringScanner.new('x').scan 'y'
=> nil
```

This usage is employed by rexml (latest).

jruby 9.2 & 9.3 follow ruby 2.6 behavior and reject it. jruby 9.4 (master) no longer implements strscan internally as 9.2 & 9.3 did, instead relying on the gem with the same name (perhaps extracted from MRI).

Unfortunately installing `strscan` on jruby 9.2 doesn't seem to fix the issue - it looks like jruby is prioritizing its bundled implementation of `strscan` over the installed gem.

anmarchenko (approved):

p-datadog (approved):

PR #3657, opened 2024-05-21
configurations: support for opentelemetry environment variables

mabdinur:
**What does this PR do?**

This PR adds maps the following open telemetry environment variables to an existing Datadog configuration::
>         OTEL_SERVICE_NAME -> DD_SERVICE
>         OTEL_LOG_LEVEL -> DD_TRACE_DEBUG
>         OTEL_PROPAGATORS -> DD_TRACE_PROPAGATION_STYLE
>         OTEL_TRACES_SAMPLER -> DD_TRACE_SAMPLE_RATE
>         OTEL_TRACES_EXPORTER -> DD_TRACE_ENABLED
>         OTEL_RESOURCE_ATTRIBUTES -> DD_TAGS

**Notable deviations from the RFC**
- There is no need to remap OTEL_LOGS_EXPORTER and OTEL_METRICS_EXPORTER. The ruby tracer is fully compaitble with the otel metrics and logs apis.

- OTEL_SDK_DISABLED does not have a corresponding datadog configuration and will not be remapped.

-  The RFC states that OTEL_RESOURCE_ATTRIBUTES should be limited to 10 values (for cardinality reasons). This PR does not bound the number of resources converted to ddtags. This feature will be added in a follow up.

Note - If both OpenTelemetry and Datadog environment variables are set, the Datadog configurations will take precedence.

**Motivation:**

Improve otel compapatiblity.

**Additional Notes:**

**How to test the change?**


lib/datadog/core/configuration/option.rb
```diff
@@ -290,6 +292,10 @@ def set_value_from_env_or_default
             effective_env = definition.env
             value = coerce_env_variable(ENV[definition.env])
             precedence = Precedence::PROGRAMMATIC
+          elsif definition.backup_env && ENV[definition.backup_env]
+            effective_env = definition.backup_env
+            value = coerce_env_variable(ENV[definition.backup_env])
+            precedence = Precedence::PROGRAMMATIC
           end
```
marcotc, line 299: Do you think it makes sense to support a list for the value of `definition.env`, instead of creating `backup_env`?
This way we won't be limited to two environment variables, and it reduces the scope of our configuration API by not adding `backup_env`.

```suggestion
          if (definition_env = definition.env)
            Array(definition_env).each do |env|
              next unless (env_val = ENV[env])
              effective_env = env
              value = coerce_env_variable(env_val)
              precedence = Precedence::PROGRAMMATIC
              break
            end
          end
```

lib/datadog/core/configuration/settings.rb
```diff
@@ -456,6 +456,7 @@ def initialize(*_)
 
           # NOTE: service also gets set as a side effect of tags. See the WORKAROUND note in #initialize for details.
           o.env Core::Environment::Ext::ENV_SERVICE
+          o.backup_env Core::Environment::Ext::ENV_OTEL_SERVICE
```
marcotc, line 459: If my previous suggestion is applied, this is what this line would look like:
```suggestion
          o.env [Core::Environment::Ext::ENV_SERVICE, Core::Environment::Ext::ENV_OTEL_SERVICE]
```

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -90,13 +90,25 @@ def self.extended(base)
...
                 o.default []
                 o.after_set do |styles|
                   next if styles.empty?
 
                   # Make values case-insensitive
                   styles.map!(&:downcase)
 
+                  if effective_env(:propagation_style) == Configuration::Ext::Distributed::ENV_OTEL_PROPAGATION_STYLE
+                    styles.collect! do |s|
+                      s = 'b3' if s == 'b3single'
```
marcotc, line 103: `b3single` is not a valid APM Ruby nor OpenTelemetry propagator. Can't we just let it fall through as an invalid case, just like any order unsupported string value? I don't see it being different from `banana`.

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -90,13 +90,25 @@ def self.extended(base)
...
+                      if Configuration::Ext::Distributed::PROPAGATION_STYLE_SUPPORTED.include?(s)
+                        s
+                      else
+                        Datadog.logger.warn("Unsupported propagation style #{s} in #{Configuration::Ext::Distributed::ENV_OTEL_PROPAGATION_STYLE}")
+                        nil
+                      end
+                    end.compact!
+                  end
```
marcotc, line 111: How about a simpler version:
```suggestion
                  styles.select! do |s|
                    if Configuration::Ext::Distributed::PROPAGATION_STYLE_SUPPORTED.include?(s)
                      true
                    else
                      Datadog.logger.warn("Unsupported propagation style #{s}")
                      false
                    end
                  end
```

lib/datadog/tracing/configuration/settings.rb
mabdinur: Ahhh good point. b3single/b3 single header is a Datadog naming convention and it is deprecated (in some languages). It seems like some ddtrace tracers (ex: java) map propagation_style=b3 to b3 multi headers and propagation_style=b3single  to b3 headers. This was the source of confusion.

lib/datadog/core/configuration/option.rb
```diff
@@ -14,7 +14,7 @@ class Option
         # @!attribute [r] precedence_set
         #   When this option was last set, what was the value precedence used?
         #   @return [Precedence::Value]
-        attr_reader :definition, :precedence_set
+        attr_reader :definition, :precedence_set, :effective_env
```
delner, line 17: Nit, non-blocking: not sure the name is intuitive? `effective` can be ambiguous as to whether its present or past tense. I think I'd like to see something that demonstrates the "ex-post-facto" nature of this value. Top of mind suggestion: `resolved_env`? `env_used`?

lib/datadog/core/configuration/settings.rb
```diff
@@ -122,9 +122,16 @@ def initialize(*_)
           # @default `DD_TRACE_DEBUG` environment variable, otherwise `false`
           # @return [Boolean]
           option :debug do |o|
-            o.env Datadog::Core::Configuration::Ext::Diagnostics::ENV_DEBUG_ENABLED
+            o.env [Datadog::Core::Configuration::Ext::Diagnostics::ENV_DEBUG_ENABLED,
```
delner, line 125: I like this array syntax!

lib/datadog/core/configuration/settings.rb
```diff
@@ -438,7 +445,7 @@ def initialize(*_)
           # @default `DD_RUNTIME_METRICS_ENABLED` environment variable, otherwise `false`
           # @return [Boolean]
           option :enabled do |o|
-            o.env Core::Runtime::Ext::Metrics::ENV_ENABLED
+            o.env [Core::Runtime::Ext::Metrics::ENV_ENABLED]
```
delner, line 448: Nit non-blocking: are `[]` necessary here?

delner (commented):
I like the spirit of this! Only a question about name and a nit, but nothing major or blocking.

I think Ruby Platform folks should review this, and take ownership of the `Core::Configuration` implementation from here on out.

Suggestion: separate out the changes in `Core::Configuration` that relate to adding env precedence and `effective_env` support to its own PR? Then rebase this PR on that one? Makes it easier to split the ownership, limit blast radius, and do reverts if necessary.

lib/datadog/core/configuration/settings.rb
mabdinur: I was experimenting with the Array syntax and forgot to remove it from this line 😓

lib/datadog/core/configuration/settings.rb
```diff
@@ -122,9 +122,16 @@ def initialize(*_)
...
           option :debug do |o|
-            o.env Datadog::Core::Configuration::Ext::Diagnostics::ENV_DEBUG_ENABLED
+            o.env [Datadog::Core::Configuration::Ext::Diagnostics::ENV_DEBUG_ENABLED,
+                   Datadog::Core::Configuration::Ext::Diagnostics::ENV_OTEL_LOG_LEVEL]
             o.default false
             o.type :bool
+            o.env_parser do |value|
+              if value
+                value = value.strip.downcase
+                ['true', '1', 'debug'].include?(value)
```
marcotc, line 132: Can you document which of these values are for otel? For example, in this case, I think `debug` is for otel.

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -121,9 +131,20 @@ def self.extended(base)
...
-                o.env Tracing::Configuration::Ext::ENV_ENABLED
+                o.env [Tracing::Configuration::Ext::ENV_ENABLED, Tracing::Configuration::Ext::ENV_OTEL_TRACES_EXPORTER]
                 o.default true
                 o.type :bool
+                o.env_parser do |value|
+                  value = value&.downcase
+                  if ['none', 'false', '0'].include?(value)
+                    false
+                  elsif ['true', '1'].include?(value)
+                    true
```
marcotc, line 142: Can you document which of these values are for otel?

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -245,7 +266,27 @@ def self.extended(base)
...
+                      value = "parentbased_#{value}"
+                    end
+                    case value
+                    when 'parentbased_always_on'
+                      1.0
+                    when 'parentbased_always_off'
+                      0.0
+                    when 'parentbased_traceidratio'
+                      ENV.fetch(Tracing::Configuration::Ext::Sampling::OTEL_TRACES_SAMPLER_ARG, 1.0).to_f
+                    else
```
marcotc, line 286: Can you document which of these values are for otel? For example, in this case, I think `debug` is for otel.

lib/datadog/core/configuration/settings.rb
mabdinur, line 132:
```suggestion
                # debug can be extracted from ENV_OTEL_LOG_LEVEL
                # true and 1 are extracted from ENV_DEBUG_ENABLED
                ['true', '1', 'debug'].include?(value)
```

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -121,9 +131,20 @@ def self.extended(base)
               # @default `DD_TRACE_ENABLED` environment variable, otherwise `true`
               # @return [Boolean]
               option :enabled do |o|
-                o.env Tracing::Configuration::Ext::ENV_ENABLED
+                o.env [Tracing::Configuration::Ext::ENV_ENABLED, Tracing::Configuration::Ext::ENV_OTEL_TRACES_EXPORTER]
                 o.default true
                 o.type :bool
+                o.env_parser do |value|
+                  value = value&.downcase
+                  if ['none', 'false', '0'].include?(value)
```
mabdinur, line 139:
```suggestion
                  # none can be extracted from OTEL_TRACES_EXPORTER
                  # true, false, 1 and 0 can be extracted from ENV_DEBUG_ENABLED
                  # All other values will ignored and the default value will be used.
                  if ['none', 'false', '0'].include?(value)
```

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -245,7 +266,27 @@ def self.extended(base)
                 # @return [Float, nil]
                 option :default_rate do |o|
                   o.type :float, nilable: true
-                  o.env Tracing::Configuration::Ext::Sampling::ENV_SAMPLE_RATE
+                  o.env [Tracing::Configuration::Ext::Sampling::ENV_SAMPLE_RATE, Tracing::Configuration::Ext::Sampling::ENV_OTEL_TRACES_SAMPLER]
+                  o.env_parser do |value|
+                    # Parse the value as a float
+                    next if value.nil?
+
+                    value = value&.downcase
```
mabdinur, line 274:
```suggestion
                    value = value&.downcase
                    # OTEL_TRACES_SAMPLER uses strings to set the sample rate (ex: always_on) this value is then mapped to a float.
                    # DD_TRACE_SAMPLE_RATE sets the sample rate to float.
```

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -245,7 +266,27 @@ def self.extended(base)
...
+                    if ['always_on', 'always_off', 'traceidratio'].include?(value)
+                      Datadog.logger.warn("The value '#{value}' is not yet supported. 'parentbased_#{value}' will be used instead.")
+                      value = "parentbased_#{value}"
+                    end
+                    case value
```
mabdinur, line 279:
```suggestion
                    # OTEL_TRACES_SAMPLER can be set to always_on, always_off, traceidratio, and/or parentbased value. 
                    # These values are mapped to a sample rate.
                    # DD_TRACE_SAMPLE_RATE sets the sample rate to float.
                    case value
```

lib/datadog/core/configuration/settings.rb
```diff
@@ -517,7 +523,22 @@ def initialize(*_)
             values.compact!
             values.each_with_object({}) do |tag, tags|
               key, value = tag.split(':', 2)
-              tags[key] = value if value && !value.empty?
+              if value.nil?
+                # tags following the Opentelemetry spec are in the format key=value
```
mabdinur, line 527:
```suggestion
                # support tags/attributes delimited by the OpenTelemetry separator (`=`)
```

marcotc (approved):

PR #3672, opened 2024-05-29
Add new GraphQL tracer complying to span attributes specification

vpellan:
**What does this PR do?**

This PR adds a new GraphQL tracer that follows Datadog's span attributes specification. This can be activated with the option "with_unified_tracer: true". This new tracer also create a span for every field instead of only the non-scalar ones. 

**Motivation:**

Following the span attributes spec is required to use API Catalog

**Additional Notes:**

- Although this now follows the span attributes spec, it seems that the http layer does not entirely (http.route for example). Meaning that we will not be able to use it in API Catalog yet
- The new tracer is named "unified_trace" as it follows the **unified** naming convention but feel free to suggest a better name (datadog_trace is already used upstream on the graphql-ruby gem, and while we could use it too, it's better to avoid confusion)


p-datadog (commented):
I read this PR and it looks really good, discussed it with @vpellan and I had two suggestions:

1. https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md appears to be the source for https://docs.datadoghq.com/tracing/trace_collection/automatic_instrumentation/dd_libraries/ruby/, add the new `with_unified_tracer` option into the options table there
2. For the graphql methods overridden by the tracer, add catch-all arguments `*args, **kwargs` to permit graphql library to add additional arguments to the methods which currently (with the PR as it is right now) would not be possible because this PR effectively freezes method signatures.

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -0,0 +1,170 @@
+# frozen_string_literal: true
+
+require 'graphql/tracing'
+
+module Datadog
+  module Tracing
+    module Contrib
+      module GraphQL
+        # These methods will be called by the GraphQL runtime to trace the execution of queries
```
marcotc, line 9: Can you add a short blurb here about what exactly the `UnifiedTrace` represents, specially, if compared with the default `TracePatcher`?

I want to make sure we have enough context in the future to know whether we should switch to use this module by default, in the next major version. (actually, if it is desirable to switch to the `UnifiedTrace` by default, let's write that down in the comments here. We used the keyword `# DEV-3.0:` when looking into possible changes for the next major version upgrade, so we can do the same here, if it applies)

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -0,0 +1,170 @@
...
+            end
+          end
+
+          def execute_query(*args, query:, **kwargs)
+            trace(proc { super }, 'execute', query.selected_operation_name, query: query) do |span|
+              span.set_tag('graphql.source', query.query_string)
+              span.set_tag('graphql.operation.type', query.selected_operation.operation_type)
+              span.set_tag('graphql.operation.name', query.selected_operation_name) if query.selected_operation_name
+              query.provided_variables.each do |key, value|
+                span.set_tag("graphql.variables.#{key}", value)
```
marcotc, line 59: Do we have a concern with unbound carnality here? In case `query.provided_variables` is a very large list?

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -0,0 +1,170 @@
...
+            resource = if query
+                         query.selected_operation_name || fallback_transaction_name(query.context)
+                       else
+                         multiplex_resource(multiplex)
+                       end
+            trace(proc { super }, 'execute_lazy', resource, query: query, multiplex: multiplex)
+          end
+
+          def execute_field_span(callable, span_key, **kwargs)
+            platform_key = @platform_key_cache[UnifiedTrace].platform_field_key_cache[kwargs[:field]]
```
marcotc, line 74: `@platform_key_cache` doesn't seem to be initialized anywhere. Could you clarify where it comes from? (and probably document it in code, unless it's super trivial and I just personally missed it)

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -0,0 +1,170 @@
...
+          end
+
+          include ::GraphQL::Tracing::PlatformTrace
+
+          # Implement this method in a subclass to apply custom tags to datadog spans
+          # @param key [String] The event being traced
+          # @param data [Hash] The runtime data for this event (@see GraphQL::Tracing for keys for each event)
+          # @param span [Datadog::Tracing::SpanOperation] The datadog span for this event
+          # def prepare_span(key, data, span)
+          # end
```
marcotc, line 129: Is the `prepare_span` method commented out for a particular reason? Or did we just forget to uncomment it?

lib/datadog/tracing/contrib/graphql/unified_trace.rb
vpellan: There is two possibilities : First option, we use provided_variables, which takes every variable even the ones that are not used in the query. This could indeed create concern for unbound cardinality. Second option: We use variables.storage. This is a map of the variables that will be used in the query, thus reducing concerns about unbound cardinality, but with less information sent in the trace.

lib/datadog/tracing/contrib/graphql/unified_trace.rb
vpellan: It is initialized upstream, in `::GraphQL::Tracing::PlatformTrace`. I will add a comment to clarify it.

lib/datadog/tracing/contrib/graphql/unified_trace.rb
vpellan: The reason is to show how you can add custom tags to your spans. We can delete these comments and add it to the documentation instead.

p-datadog (commented):
Approving because the points I brought up are now addressed, but I find Marco's questions and suggestions valuable as well.

p-datadog (approved):

lib/datadog/tracing/contrib/graphql/unified_trace.rb
vpellan: Fixed !

marcotc (approved):
Awesome work, @vpellan!

PR #3675, opened 2024-05-30
Add limits to #log_deprecation

delner:
**What does this PR do?**

This adds a limiting feature to the `log_deprecation` function. By default, there is no limit. However, if you provide a `key:`, then it will prevent any other message logged with the same key thereafter.

**Motivation:**

Some deprecated behavior may be on a frequently used path that could produce a lot of noise. Instead of logging every time, or never logging at all, make it possible to warn once.

**Additional Notes:**

I extracted the existing `log_deprecation` function into its own module (`Core::Deprecations`) and recomposed it back into the original `Core` module. Now that the deprecation behavior was becoming more sophisticated, I didn't want the `core.rb` file to scope creep; thus I isolated the behavior to its own file and module.

**How to test the change?**

Run the test suite.

### DIFF
--- a/lib/datadog/core.rb
+++ b/lib/datadog/core.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_relative 'core/deprecations'
 require_relative 'core/extensions'
 
 # We must load core extensions to make certain global APIs
@@ -9,25 +10,7 @@ module Datadog
   # products. It is a dependency of each product. Contrast with Datadog::Kit
   # for higher-level features.
   module Core
-    class << self
-      # Records the occurrence of a deprecated operation in this library.
-      #
-      # Currently, these operations are logged to `Datadog.logger` at `warn` level.
-      #
-      # `disallowed_next_major` adds a message informing that the deprecated operation
-      # won't be allowed in the next major release.
-      #
-      # @yieldreturn [String] a String with the lazily evaluated deprecation message.
-      # @param [Boolean] disallowed_next_major whether this deprecation will be enforced in the next major release.
-      def log_deprecation(disallowed_next_major: true)
-        Datadog.logger.warn do
-          message = yield
-          message += ' This will be enforced in the next major release.' if disallowed_next_major
-          message
-        end
-        nil
-      end
-    end
+    extend Core::Deprecations
   end
 
   extend Core::Extensions
--- a/lib/datadog/core/deprecations.rb
+++ b/lib/datadog/core/deprecations.rb
@@ -1,0 +1,58 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Core
+    # Contains behavior for handling deprecated functions in the codebase.
+    module Deprecations
+      # Records the occurrence of a deprecated operation in this library.
+      #
+      # Currently, these operations are logged to `Datadog.logger` at `warn` level.
+      #
+      # `disallowed_next_major` adds a message informing that the deprecated operation
+      # won't be allowed in the next major release.
+      #
+      # @yieldreturn [String] a String with the lazily evaluated deprecation message.
+      # @param [Boolean] disallowed_next_major whether this deprecation will be enforced in the next major release.
+      # @param [Object] key A unique key for the deprecation. Only the first message with the same key will be logged.
+      def log_deprecation(disallowed_next_major: true, key: nil)
+        return unless log_deprecation?(key)
+
+        Datadog.logger.warn do
+          message = yield
+          message += ' This will be enforced in the next major release.' if disallowed_next_major
+          message
+        end
+
+        # Track the deprecation being logged.
+        deprecation_logged!(key)
+
+        nil
+      end
+
+      private
+
+      # Determines whether a deprecation message should be logged.
+      #
+      # Internal use only.
+      def log_deprecation?(key)
+        return true if key.nil?
+
+        # Only allow a deprecation to be logged once.
+        !logged_deprecations.key?(key)
+      end
+
+      def deprecation_logged!(key)
+        return if key.nil?
+
+        logged_deprecations[key] += 1
+      end
+
+      # Tracks what deprecation warnings have already been logged
+      #
+      # Internal use only.
+      def logged_deprecations
+        @logged_deprecations ||= Hash.new(0)
+      end
+    end
+  end
+end
--- a/sig/datadog/core/deprecations.rbs
+++ b/sig/datadog/core/deprecations.rbs
@@ -1,0 +1,22 @@
+module Datadog
+  module Core
+    module Deprecations
+      interface _Hashing
+        def hash: () -> ::Integer
+        def eql?: (untyped) -> bool
+        def nil?: () -> bool
+      end
+
+      type key = _Hashing
+
+      @logged_deprecations: ::Hash[key, ::Integer]
+      def log_deprecation: (?disallowed_next_major: bool, ?key: key?) { () -> String } -> void
+
+      private
+      def log_deprecation?: (key key) -> bool
+
+      def deprecation_logged!: (key key) -> void
+      def logged_deprecations: () -> ::Hash[key, ::Integer]
+    end
+  end
+end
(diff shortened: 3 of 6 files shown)


marcotc (approved):

sig/datadog/core/deprecations.rbs
```diff
@@ -0,0 +1,14 @@
+module Datadog
```
delner, line 1: Does this RBS file need adjustment? Not sure what to do with this, honestly.

sig/datadog/core/deprecations.rbs
```diff
@@ -0,0 +1,14 @@
...
+  module Core
+    module Deprecations
+      @_logged_deprecations: untyped
```
lloeki, line 4: - From `Hash.new(0)` the expected hash values are `Integer`.
- From `_logged_deprecations[key] += 1` the hash values are indeed `Integer` and the keys are `key`.

To avoid repetition, let's call the `key` type "`key`" by defining a type alias.

> Note: RBS type aliases are all lowercase, distinguishing them from Ruby types which are constants and thus start with an uppercase letter.

> Note: The apparent name clash is resolved at `.rbs` parse time, as there is no actual parsing ambiguity.

Also, from our discussion, the expected behaviour of `key` is to behave correctly as a hash key, which means it must to `#hash` and `#eql?`. Let's define an interface `_Hashing` for that.

> Note: the apparent name clash is resolved at `.rbs` parse time, as there is no actual parsing ambiguity.

> Note: `_Hashing` comes from [the RBS doc](https://github.com/ruby/rbs/blob/e5cc58296fb3080a52a1bdae1a47bc12dd41a0cf/docs/syntax.md#interface-declaration). 

This gives us:

```suggestion
      interface _Hashing
        def hash: () -> ::Integer
        def eql?: (untyped) -> bool
      end

      type key = _Hashing

      @_logged_deprecations: ::Hash[key, ::Integer]
```

lloeki (changes requested):

TonyCTHsu (commented):
I would like to see some changes to the method naming from this module. Currently, it is confusing because everything named with combination of `log` and `deprecation` and a bunch of `!`, `?` `_`.  Those knowledge should already been capture under the namespace of `Deprecation`.

lib/datadog/core/deprecations.rb
```diff
@@ -0,0 +1,58 @@
...
+      # Records the occurrence of a deprecated operation in this library.
+      #
+      # Currently, these operations are logged to `Datadog.logger` at `warn` level.
+      #
+      # `disallowed_next_major` adds a message informing that the deprecated operation
+      # won't be allowed in the next major release.
+      #
+      # @yieldreturn [String] a String with the lazily evaluated deprecation message.
+      # @param [Boolean] disallowed_next_major whether this deprecation will be enforced in the next major release.
+      # @param [Object] key A unique key for the deprecation. Only the first message with the same key will be logged.
```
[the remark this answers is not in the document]
delner: `[Object]` is effectively untyped. You can give it pretty much anything. It's really meant to take [nil, Symbol, String, Object (that responds to hash)].

delner:
@TonyCTHsu 

 - `everything named with combination of log and deprecation`: this is because its a composable module: it could conflict with other method names in the class its composed into.
 - `a bunch of !, ? _`: I agree that I can strike the underscore in the one name. But the `?` and `!` are Ruby convention for those methods.

delner:
Thanks @lloeki for the detailed feedback on the RBS! I will try to apply that, and submit again for your consideration.

lib/datadog/core/deprecations.rb
```diff
@@ -0,0 +1,58 @@
...
+      def deprecation_logged!(key)
+        return if key.nil?
+
+        _logged_deprecations[key] += 1
+      end
+
+      # Tracks what deprecation warnings have already been logged
+      #
+      # Internal use only.
+      def _logged_deprecations
```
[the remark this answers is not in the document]
lloeki: > For... private?

I kind of got the idea, but that's a bit awkward and unconventional in Ruby land and really doesn't prevent anything.

That's a very [Python](https://docs.python.org/3/tutorial/classes.html#tut-private) and [C/C++](https://stackoverflow.com/a/228797) thing :) where a leading underscore is sort of equivalent to Ruby's `private`.

Tangentially, [Go's way of doing it](https://go.dev/doc/effective_go#names) is that the visibility of a name outside a package is determined by whether its first character is upper case - in which case it is public - along with an [`internal` package name](https://tip.golang.org/doc/go1.4#internalpackages).

p-datadog (approved):
If I am understanding this PR correctly, each class/module that extends `Deprecations` will get its own tracker of what was already warned about. I think this would be useful to include in the documentation for `Deprecations`.

lib/datadog/core/deprecations.rb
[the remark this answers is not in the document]
delner: Yes, I'm aware. Unfortunately Ruby occasionally leaks `private` things when there's composition in play with `Module`s, which is why I was trying to flag this as private by some other means. But in this case, I think I was misremembering this as a gap with how `protected` works.

In any case, the underscore doesn't serve the purpose I need, hence I'll remove it.

sig/datadog/core/deprecations.rbs
delner: So IIUC, we don't do `@logged_deprecations: ::Hash[::Object, ::Integer]` because `Object` may not respond to the functions that make it "hashable", therefore, we define an interface to which objects used as Hash keys must conform to?

delner:
@p-datadog Technically true, but its intention is to be composed into the `Core` module itself, so I don't really expect anyone to extend it.

delner:
@lloeki the guide and explanation about how to properly RBS type this was great. I definitely learned some things about RBS. Thank you! I think this is ready for another look.

sig/datadog/core/deprecations.rbs
lloeki: Correct. In addition:

- a `BasicObject` child instance may actually be hashable
- an `Object` child instance may have their `hash` and `eql?` methods removed
- it gets precise and explicit as to what is actually required, both in naming (`Object` is unclear as to what's the requirement is) and in which methods (through the definition of the interface)

lib/datadog/core/deprecations.rb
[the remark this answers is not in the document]
lloeki: > `[Object]` is effectively untyped

`BasicObject` would like to have a word ;)

lib/datadog/core.rb
```diff
@@ -9,25 +10,7 @@ module Datadog
...
-      def log_deprecation(disallowed_next_major: true)
-        Datadog.logger.warn do
-          message = yield
-          message += ' This will be enforced in the next major release.' if disallowed_next_major
-          message
-        end
-        nil
-      end
-    end
+    extend Core::Deprecations
```
lloeki, line 13: Missed that one: probably this needs an update to `sig/datadog/core.rbs` with:

```
  extend Core::Deprecations
```

lloeki (approved):

sig/datadog/core/deprecations.rbs
```diff
@@ -0,0 +1,22 @@
+module Datadog
+  module Core
+    module Deprecations
```
lloeki, line 3: >> each class/module that extends Deprecations will get its own tracker of what was already warned about. I think this would be useful to include in the documentation for Deprecations.

> I don't really expect anyone to extend it.

It might be possible to enforce this via a [self type constraint](https://github.com/ruby/rbs/blob/master/docs/syntax.md#module-declaration):

> Module declaration takes optional *self type* parameter, which defines a constraint about a class when the module is mixed.

Something like (not quite sure):

```suggestion
    module Deprecations : Core
```

delner: Adding this with the change above causes:

```
sig/datadog/core.rbs:5:2: [error] Module self type constraint in type `::Datadog` doesn't satisfy: `singleton(::Datadog) <: ::Datadog::Core`
│ Diagnostic ID: RBS::ModuleSelfTypeError
│
└   extend Core::Deprecations
    ~~~~~~~~~~~~~~~~~~~~~~~~~
```

PR #3676, opened 2024-06-03
Injection guard rails

TonyCTHsu:
**What does this PR do?**

This PR adds some guard rails for host injection

1. Checking  Ruby runtime
2. Checking Ruby platform
3. Send telemetry with the provided executable from `ENV['DD_TELEMETRY_FORWARDER_PATH']`

In additionally, 

1. Fix false positive result while checking whether the dependency exists in the bundle
2. Write to another Gemfile instead of the existing Gemfile and swap the bundled Gemfile in order to make lib injection transactional to prevent corrupted file.

### DIFF
--- a/lib-injection/host_inject.rb
+++ b/lib-injection/host_inject.rb
@@ -39,7 +39,7 @@ begin
         tracer_version: tracer_version,
         pid: pid
       },
-      points: Array(events)
+      points: events
     }.to_json
 
     output, = Open3.capture2e(ENV.fetch('DD_TELEMETRY_FORWARDER_PATH', 'cat'), stdin_data: payload)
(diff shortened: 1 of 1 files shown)


lib-injection/host_inject.rb
```diff
@@ -1,48 +1,113 @@
...
+  end
+
+  def dd_skip_injection!
+    ENV['DD_TRACE_SKIP_LIB_INJECTION'] = 'true'
+  end
+
+  def dd_send_telemetry(events)
+    pid = Process.respond_to?(:pid) ? Process.pid : 0 # Not available on all platforms
+
+    tracer_version = if File.exist?('/opt/datadog/apm/library/ruby/version.txt')
```
marcotc, line 27: This `version.txt` is not familiar to me, where does it come from?

marcotc (approved):
Looks very good! Just one comment

lib-injection/host_inject.rb
TonyCTHsu: During the process for building the gem, we wrote it.
https://github.com/DataDog/dd-trace-rb/blob/master/.gitlab-ci.yml#L94

PR #3679, opened 2024-06-04
chore: add a mechanism to set multiple envs per configuration

mabdinur:
**What does this PR do?**

- Updates  OptionDefintion.env to accept a list of environment variables. With this change dd-trace-rb contributors can supply multiple envars for each configuration (in order of precedence).
- Introduces `resolved_env` instance method for the Options class. This will allow developers to detect which environment variable were used to set a configuration.

**Motivation:**

**Additional Notes:**

**How to test the change?**


lib/datadog/core/configuration/option.rb
```diff
@@ -284,16 +286,21 @@ def context_eval(&block)
...
-          if definition.env && ENV[definition.env]
-            effective_env = definition.env
-            value = coerce_env_variable(ENV[definition.env])
-            precedence = Precedence::PROGRAMMATIC
+          if definition.env
+            Array(definition.env).each do |env|
+              next unless (env_val = ENV[env])
+
+              resolved_env = env
+              value = coerce_env_variable(env_val)
```
mabdinur, line 296:
```suggestion
              next if ENV[env].nil?
              resolved_env = env
              value = coerce_env_variable(ENV[env])
```

marcotc (approved):

PR #3685, opened 2024-06-05
Add option to inherit service name from the parent span

marcotc:
**What does this PR do?**

This PR is the first step towards aligning the Ruby tracer with all other APM languages in inheriting the parent span's service name by default.

Making this change is a breaking change, so this is introduced as an option which default to "disabled".

This option should be made "enabled" by default in the next major version.

**How to test the change?**

Changing the setting `c.tracing.inherit_parent_service` to `true` allows you to see the new behaviour, with the span's service inheriting the value from the active parent span.


lib/datadog/tracing/tracer.rb
```diff
@@ -341,12 +347,13 @@ def bind_trace_events!(trace_op)
         events = trace_op.send(:events)
 
         events.span_before_start.subscribe do |event_span_op, event_trace_op|
-          event_trace_op.service ||= @default_service
```
marcotc, line 344: A small improvement here: we used to always set `@default_service` as the fallback service for a the newly created `TraceOperation`. By explicity setting the value of `event_trace_op.service` here, the `TraceOperation` can no longer use the root span service name as its default value, which is the intended behavior: https://github.com/DataDog/dd-trace-rb/blob/b3c5a3d57b7441da2d7c7ceb3e24acb4c5846bf5/lib/datadog/tracing/trace_operation.rb#L171-L173

This PR no longer explicitly sets the service field in the active `TraceOperation`, letting `event_trace_op.service` handle the fallback normally, as it was originally intended.

lib/datadog/tracing/tracer.rb
```diff
@@ -341,12 +347,13 @@ def bind_trace_events!(trace_op)
...
-          event_span_op.service ||= @default_service
           sample_trace(event_trace_op) if event_span_op && event_span_op.parent_id == 0
         end
 
         events.span_finished.subscribe do |event_span, event_trace_op|
+          # Fallback in case the service was never set
+          event_span.service ||= @default_service
```
marcotc, line 355: I moved the fallback `event_span.service ||= @default_service` from the span creation event to span completion evnt, because semantically a `nil` span service name has the meaning of "use whatever the default behaviour is for span service naming". But before, we were not allowing this to happen because we are setting the service name explicitly too early, in the `span_before_start` event.

cswatt (approved):
approved for docs

spec/datadog/tracing/tracer_spec.rb
```diff
@@ -259,29 +273,55 @@
...
+        subject!(:nested_spans) do
           tracer.trace('parent', service: 'service-parent') do
             tracer.trace('child1') { |s| s.set_tag('tag', 'tag_1') }
             tracer.trace('child2', service: 'service-child2') { |s| s.set_tag('tag', 'tag_2') }
           end
+        end
 
-          expect(spans).to have(3).items
+        context 'when inherit_parent_service is false' do
+          it 'propagates parent span and uses default service name' do
```
TonyCTHsu, line 284: Be explicit with conditions and useful when flipping the defaults

```
let(:tracer_options) { super().merge(inherit_parent_service: false) }
```

spec/datadog/tracing/tracer_spec.rb
```diff
@@ -259,29 +273,55 @@
       end
 
       context 'when nesting spans' do
-        it 'propagates parent span and uses default service name' do
+        subject!(:nested_spans) do
```
TonyCTHsu, line 276: Can we remove `!` here?

This brings confusion to me with the ordering of execution, such as "Would this be executed before the `let` in the nested context"?

docs/GettingStarted.md
```diff
@@ -1935,6 +1935,7 @@ For example, if `tracing.sampling.default_rate` is configured by [Remote Configu
...
 | `tracing.header_tags`                                  | `DD_TRACE_HEADER_TAGS`                                  | `Array`                               | `nil`                        | Record HTTP headers as span tags. See [Applying header tags to root spans][header tags] for more information.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
+| `tracing.inherit_parent_service`                       |                                                         | `Bool`                                | `false`                      | Inherit the parent span's service name when creating a new child span. |
```
TonyCTHsu, line 1938: Should we be more explicit about this option is not ready for production yet? 

And it might leads to breaking changes for user existing graph with various instrumentations

lib/datadog/tracing/span_operation.rb
```diff
@@ -439,6 +443,11 @@ def message
       # mutation by reference; when this span is returned,
       # we don't want this SpanOperation to modify it further.
       def build_span
+        # Use parent service if no service is set and @inherit_parent_service is set
```
TonyCTHsu, line 446: I am generally supportive for moving toward this direction. What I found a larger problem is the fact that `service` is a mutable field. 

Things get tricky when parent span changes its service after child spans are finished. For instance:

```ruby

Datadog::Tracing.trace('parent', service: 'service-parent') do |span_op, _|
  Datadog::Tracing.trace('child1') {}
  span_op.service = "BalaBoom!" # Change the service name after the child span is finished
end
```

I was not sure what the behaviour is like for instrumentation depending on `ActiveSupport::Notifications`, which I remember the service field is set in the after callback. 

Any thoughts about restricting the mutability of the service field?

delner: I agree with Tony; mutability is a problem. Generally speaking, I think moving towards an immutable world simplifies things quite a bit, the only question is how.

delner: Secondary question: how does this feature fit in with the "one service name" change we originally planned for 2.0? If we end up introducing that, is this obsolete?

lib/datadog/tracing/span_operation.rb
```diff
@@ -439,6 +443,11 @@ def message
...
+        service = self.service || (@parent&.service if @inherit_parent_service)
+
+        service_entry = @parent.nil? || (service && @parent.service != service)
```
delner, line 449: What is `@parent`? It concerns me that `SpanOperation` has a reference to `parent`, because of the deep coupling/nesting... more references means less garbage collection means more memory bloat. It also makes it harder to move data structures like this off the critical path to a background process, if there was an optimization opportunity to do so.

I know my original intention was to eliminate `@parent` altogether. But I can't remember if we couldn't and if I just had to minimize its usage. I'm not hot on the idea of expanding its usage...

p-datadog (approved):
The `lib` part of the diff LGTM

ivoanjo:

ivoanjo:
This has been around for a long time, so I'm going to go ahead and close it -- please do reopen if this is still useful!

PR #3693, opened 2024-06-07
Add Ruby 3.3 to matrix target in test-yjit

MITSUBOSHI:
Ruby 3.3 has already been released last year and has mayn improvement changes on yjit.
Thus, I added Ruby 3.3 to matrix test target in test-yjit.

https://github.com/ruby/ruby/blob/029d92b8988d26955d0622f0cbb8ef3213200749/doc/NEWS/NEWS-3.3.0.md?plain=1#L401


ivoanjo (approved):
Great idea, thanks for the PR :+1:

PR #3694, opened 2024-06-07
Move away from Debian Buster

lloeki:
**2.0 Upgrade Guide notes**

**What does this PR do?**

Move away from Debian Buster

**Motivation:**

- Debian Buster goes EOL on 2024-06-30.
- Ubuntu Jammy goes EOL in April 2025.
- Bullseye goes EOL in June 2026
- Bookrworm goes EOL in June 2028

This buys us a bit of time.

**Additional Notes:**

- Using the Docker Hub images, 2.5 is stuck on Buster. We may have to build our own Ruby.
- This should be back ported to `1-x-stable`, where 2.1-2.5 may need to receive the same treatment as above.

**How to test the change?**

Images should build and run.


TonyCTHsu (approved):
Generally agreed. Perhaps we need to change the file names as well.

p-datadog:
I think the file names should be changed so that they are correct (either by adjusting the patch version or removing it altogether), otherwise LGTM.

marcotc (approved):
I agree with @p-datadog,
> file names should be changed so that they are correct (either by adjusting the patch version or removing it altogether)

lloeki:
You're absolutely correct, will do.

PR #3695, opened 2024-06-10
[PROF-9821] Fix incorrect code provenance due to broken JSON monkey patch

ivoanjo:
**What does this PR do?**

This PR tweaks the
`Datadog::Profiling::Collectors::CodeProvenance::Library` class so that if it is encoded instance-filed-by-instance-field it results in a correct code provenance JSON file.

Specifically, the `path` argument is now stored as a `paths` array, to match what we expect in the JSON file.

**Motivation:**

We've discovered that in a certain combination of the oj and activesupport libraries, when oj is used to replace the standard library JSON gem, it incorrectly encodes our `Library` instances instance-field-by-instance-field instead of calling `#to_json`.

This resulted in incorrect code provenance files, which would be rejected by the backend.

Since the change is trivial (`path` => `paths` array), I've opted to change the shape of `Library` so that it still encodes correctly in the presence of a buggy JSON encoder monkey patch.

**Additional Notes:**

N/A

**How to test the change?**

I've added coverage to this issue. I did it in a bit of a roundabout way (using YAML), see details for why, but I claim it's a reasonable proxy for any encoder that encodes field-by-field, including oj.

### DIFF
--- a/spec/datadog/profiling/collectors/code_provenance_spec.rb
+++ b/spec/datadog/profiling/collectors/code_provenance_spec.rb
@@ -1,5 +1,6 @@
 require 'datadog/profiling/collectors/code_provenance'
 require 'json-schema'
+require 'yaml'
 
 RSpec.describe Datadog::Profiling::Collectors::CodeProvenance do
   subject(:code_provenance) { described_class.new }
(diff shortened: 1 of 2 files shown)


spec/datadog/profiling/collectors/code_provenance_spec.rb
```diff
@@ -187,5 +188,76 @@
...
+
+      it 'is still able to correctly encode a library instance' do
+        instance = library_class_without_to_json.new(
+          name: 'datadog',
+          kind: 'library',
+          version: '1.2.3',
+          path: '/example/path/to/datadog/gem',
+        )
+
+        serialized_without_to_json = YAML.dump(instance)
```
p-datadog, line 250: Is the reasoning here that YAML serialization is performed field by field, same as some JSON libraries would serialize, or that libraries like `oj` actually use YAML serialization logic? I am guessing it's the former and if so, a comment to this effect in the code would be helpful.

p-datadog (approved):

p-datadog (approved):

spec/datadog/profiling/collectors/code_provenance_spec.rb
ivoanjo: Yes! I've added a comment above the spec stating this:

```
    # So instead in this test we use YAML as an example of an encoder that doesn't use #to_json, and does it
    # field-by-field. Thus if the Library class doesn't match exactly what we want in the output, this test will fail.
```

PR #3699, opened 2024-06-10
Add debug info for log device

TonyCTHsu:
**What does this PR do?**

Adding debug info for `Logger::LogDevice` 

**Motivation:**

We have seen this flaky specs in CircleCI, especially with JRuby.

```
log writing failed. closed stream
```

The error cannot be reproduce locally but it seems to be pretty generic when trying to write to a closed stream. 

```ruby
logger = Logger.new('t.log')
logger.close       # => nil
logger.info('foo') # Prints "log writing failed. closed stream"
```

Source: https://github.com/ruby/ruby/blob/master/lib/logger/log_device.rb#L48


.circleci/config.yml
```diff
@@ -40,6 +40,7 @@ test_containers:
     DD_AGENT_HOST: testagent
     DD_TRACE_AGENT_PORT: 9126
     DATADOG_GEM_CI: true
+    RUBYOPT: -r/app/.circleci/patch_log_device
```
ivoanjo, line 43: I would suggest applying this in `spec_helper.rb`, as we have a few other similar tweaks for the test suite there; applying it via RUBYOPT may collide with other things we may be testing in the test suite. (Because it affects every ruby process that gets run, including e.g. bundler and whatnot)

.circleci/patch_log_device.rb
```diff
@@ -0,0 +1,16 @@
...
+      def write(message)
+        $stdout.puts("[Writing] ProcessID: #{Process.pid}, ThreadID: #{Thread.current.object_id}, Message: #{message}")
+        super(message)
+      end
+
+      def close
+        $stdout.puts("[Closing] ProcessID: #{Process.pid}, ThreadID: #{Thread.current.object_id}")
+        super
+      end
+    end
```
ivoanjo, line 14: I would suggest using the same `warn(...)` mechanism that the logger does, because we're not actually sure that `$stdout` is open at this time -- it may have been closed.

Also, I'd suggest printing every thread (`Thread.list`) and their backtrace, so you'll know what may be staying over and perhaps making the app not quit.

Also, instrumenting every `write` and `close` may get a bit slow/chatty? I'd suggest instrumenting only when there's an exception, e.g. when we know the issue has been triggered.

spec/datadog/core/environment/execution_spec.rb
```diff
@@ -72,7 +72,7 @@
             f.close
 
             out, = Open3.capture2e('pry', '-f', '--noprompt', f.path)
-            expect(out).to eq('ACTUAL:true')
+            expect(out).to be_end_with('ACTUAL:true')
           end
```
ivoanjo, line 76: Is this change needed because of the RUBYOPT change? :thinking:

ivoanjo (commented):
Thanks for picking this up! Left a few notes :)

ivoanjo:
Also, you may want to check that the monkey patch works by e.g. creating a logger on a new stream (such as `StringIO`) and then closing it, and then checking we get all the relevant info :)

.circleci/patch_log_device.rb
p-datadog: `warn` should also output to standard error rather than standard output which I think would be more correct?

lloeki: - `$stdout` or `STDOUT` may also be taken over (say by some mock or stub). `File.open(/dev/stdout)`? (or err)?
- either or both of `$stdout`/`err` may or may not be closed. Might be worth querying and logging their state via `$stdout.closed?` as well?
- what about logging to a _file_?

> Also, instrumenting every write and close may get a bit slow/chatty?

Let's get some (any) output first and see if we're having some patterns emerge.

> instrumenting only when there's an exception, e.g. when we know the issue has been triggered.

We dont' know what's causing it, or do we? Seems speculative to tie log output to a hypothetical.

PR #3701, opened 2024-06-11
Include `source_code_uri` in gemspec, version `changelog_uri`

Earlopain:
See #3684

If you compare https://rubygems.org/gems/datadog and https://rubygems.org/gems/ddtrace you will notice that `datadog` is missing the GitHub star count, and the "Source Code" link. Right now you can only go to source through the "Changelog" link. This should fix that, though the GitHub badge may depend on `spec.homepage`, not entirely sure. Either way, a bit better than now.

### DIFF
--- a/datadog.gemspec
+++ b/datadog.gemspec
@@ -29,7 +29,8 @@ Gem::Specification.new do |spec|
 
   if spec.respond_to?(:metadata)
     spec.metadata['allowed_push_host'] = 'https://rubygems.org'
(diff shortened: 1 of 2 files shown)


ivoanjo (approved):
Nice catch, thank you! :+1:

TonyCTHsu (approved):

ivoanjo:
Also pulled in #3697 to fix CI test failure :)

PR #3709, opened 2024-06-12
fix: ensure the correct ids are serialized and propagated

mabdinur:
**What does this PR do?**

- Ensures span_ids and trace_ids are encoded/decoded (integer <-> bytes) in a manner that is [consistent with the OpenTelemetry API](https://github.com/open-telemetry/opentelemetry-ruby/blob/5b2cc8ffe7614c27fd69f0b40d34368f0c06ff53/api/lib/opentelemetry/trace/span_context.rb#L41-L50)
  - Ensures 128bit trace_ids are not truncated to 64bits in the OpenTelemetry propagator
  - Ensures OpenTelemetry spans generate a corresponding Datadog span with the same Span ID and TraceID. This ensures OtelSpan.context returns the expected value. Regression test: https://github.com/DataDog/system-tests/pull/2548

**Motivation:**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/lib/datadog/opentelemetry/sdk/propagator.rb
+++ b/lib/datadog/opentelemetry/sdk/propagator.rb
@@ -41,9 +41,9 @@ module Datadog
           digest = @datadog_propagator.extract(carrier)
           return context unless digest
 
-          # Converts the {Numeric} Datadog id object to OpenTelemetry's byte array format.
-          trace_id = [format('%032x', digest.trace_id)].pack('H32')
-          span_id = [format('%016x', digest.span_id)].pack('H16')
+          # Converts the {Numeric} Datadog id object to OpenTelemetry's byte array format.  
+          trace_id = [digest.trace_id >> 64, digest.trace_id & 0xFFFFFFFFFFFFFFFF].pack('Q>Q>') # 128-bit unsigned, big-endian integer  
+          span_id = [digest.span_id].pack('Q>') # 64-bit unsigned, big-endian integer 
 
           if digest.trace_state || digest.trace_flags
             trace_flags = ::OpenTelemetry::Trace::TraceFlags.from_byte(digest.trace_flags)
--- a/lib/datadog/opentelemetry/sdk/propagator.rb
+++ b/lib/datadog/opentelemetry/sdk/propagator.rb
@@ -41,9 +41,11 @@ module Datadog
           digest = @datadog_propagator.extract(carrier)
           return context unless digest
 
-          # Converts the {Numeric} Datadog id object to OpenTelemetry's byte array format.  
-          trace_id = [digest.trace_id >> 64, digest.trace_id & 0xFFFFFFFFFFFFFFFF].pack('Q>Q>') # 128-bit unsigned, big-endian integer  
-          span_id = [digest.span_id].pack('Q>') # 64-bit unsigned, big-endian integer 
+          # Converts the {Numeric} Datadog id object to OpenTelemetry's byte array format.
+          # 128-bit unsigned, big-endian integer
+          trace_id = [digest.trace_id >> 64, digest.trace_id & 0xFFFFFFFFFFFFFFFF].pack('Q>Q>')
+          # 64-bit unsigned, big-endian integer
+          span_id = [digest.span_id].pack('Q>')
 
           if digest.trace_state || digest.trace_flags
             trace_flags = ::OpenTelemetry::Trace::TraceFlags.from_byte(digest.trace_flags)
--- a/lib/datadog/opentelemetry/sdk/propagator.rb
+++ b/lib/datadog/opentelemetry/sdk/propagator.rb
@@ -41,8 +41,9 @@ module Datadog
           digest = @datadog_propagator.extract(carrier)
           return context unless digest
 
-          trace_id = to_otel_id(digest.trace_id)
-          span_id = to_otel_id(digest.span_id)
+          # Converts the {Numeric} Datadog id object to OpenTelemetry's byte array format.
+          trace_id = [format('%032x', digest.trace_id)].pack('H32')
+          span_id = [format('%016x', digest.span_id)].pack('H16')
 
           if digest.trace_state || digest.trace_flags
             trace_flags = ::OpenTelemetry::Trace::TraceFlags.from_byte(digest.trace_flags)
@@ -77,14 +78,6 @@ module Datadog
         # DEV: Doesn't seem like it's used in production Otel code paths.
         def fields
           []
-        end
-
-        private
-
-        # Converts the {Numeric} Datadog id object to OpenTelemetry's byte array format.
-        # This method currently converts an unsigned 64-bit Integer to a binary String.
-        def to_otel_id(dd_id)
-          Array(dd_id).pack('Q')
         end
       end
     end
--- a/lib/datadog/opentelemetry/sdk/span_processor.rb
+++ b/lib/datadog/opentelemetry/sdk/span_processor.rb
@@ -85,11 +85,6 @@ module Datadog
           name, kwargs = span_arguments(span, attributes)
 
           datadog_span = Tracing.trace(name, **kwargs)
-          # The Datadog span must have the same ID as the OpenTelemetry span
-          # DEV: We need to set the span ID after the span is created
-          datadog_span.id = span.context.hex_span_id.to_i(16)
-          datadog_span.trace_id = span.context.hex_trace_id.to_i(16)
-
           datadog_span.set_error([nil, span.status.description]) unless span.status.ok?
           datadog_span.set_tags(span.attributes)
 
@@ -146,6 +141,9 @@ module Datadog
           attributes.flatten!(1)
 
           kwargs[:tags] = attributes.to_h
+
+          # DEV: The datadog span must have the same ID as the OpenTelemetry span
+          kwargs[:id] = span.context.hex_span_id.to_i(16)
 
           [name, kwargs]
         end
--- a/lib/datadog/opentelemetry/sdk/span_processor.rb
+++ b/lib/datadog/opentelemetry/sdk/span_processor.rb
@@ -70,7 +70,8 @@ module Datadog
           if parent_context.trace
             Tracing.send(:tracer).send(:call_context).activate!(parent_context.ensure_trace)
           else
-            Tracing.continue_trace!(nil)
+            otel_trace_id = span.context.hex_trace_id.to_i(16)
+            Tracing.continue_trace!(Datadog::Tracing::TraceDigest.new(trace_id: otel_trace_id, span_remote: false))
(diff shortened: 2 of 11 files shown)


spec/datadog/profiling/collectors/thread_context_spec.rb
```diff
@@ -641,13 +641,13 @@ def self.otel_sdk_available?
...
-                      @t1_local_root_span_id = Datadog::Tracing.correlation.span_id
+                    otel_tracer.in_span('profiler.test') do |root|
+                      @t1_local_root_span_id = root.context.hex_span_id.to_i(16)
                       otel_tracer.in_span('profiler.test.nested.1') do
                         Datadog::Tracing.trace('profiler.test.nested.2') do
                           otel_tracer.in_span('profiler.test.nested.3') do
-                            Datadog::Tracing.trace('profiler.test.nested.4') do
-                              @t1_span_id = Datadog::Tracing.correlation.span_id
+                            Datadog::Tracing.trace('profiler.test.nested.4') do |leaf|
+                              @t1_span_id = leaf.context.hex_span_id.to_i(16)
```
mabdinur, line 650:
```suggestion
                            Datadog::Tracing.trace('profiler.test.nested.4') do
                              @t1_span_id = Datadog::Tracing.correlation.span_id
```

lib/datadog/opentelemetry/sdk/propagator.rb
```diff
@@ -41,8 +41,9 @@ def extract(
           digest = @datadog_propagator.extract(carrier)
           return context unless digest
 
-          trace_id = to_otel_id(digest.trace_id)
-          span_id = to_otel_id(digest.span_id)
+          # Converts the {Numeric} Datadog id object to OpenTelemetry's byte array format.
+          trace_id = [format('%032x', digest.trace_id)].pack('H32')
+          span_id = [format('%016x', digest.trace_id)].pack('H16')
```
mabdinur, line 46:
```suggestion
          span_id = [format('%016x', digest.span_id)].pack('H16')
```

lib/datadog/opentelemetry/sdk/span_processor.rb
```diff
@@ -85,6 +85,10 @@ def start_datadog_span(span)
           name, kwargs = span_arguments(span, attributes)
 
           datadog_span = Tracing.trace(name, **kwargs)
+          # The Datadog span must have the same ID as the OpenTelemetry span
+          # DEV: We need to set the span ID after the span is created
+          datadog_span.id = span.context.hex_span_id.to_i(16)
+          datadog_span.trace_id = span.context.hex_trace_id.to_i(16)
```
mabdinur, line 91: Fix #1: Ensure otel and datadog spans have the same trace and span id. This will ensure `otel_span.context` returns valid ids

mabdinur: Note - This approach is very naive. Not sure if it introduces race conditions or breaks span activation.

lib/datadog/tracing/span_operation.rb
```diff
@@ -421,6 +413,12 @@ def message
         end
       end
 
+      # Provides a mechanism to override the automatically generated span ID
+      # @param id [Integer] the span ID
+
+      # Provides a mechanism to override the automatically generated trace ID
+      # @param trace_id [Integer] the trace ID
+
```
mabdinur, line 421:
```suggestion
```

lib/datadog/opentelemetry/sdk/propagator.rb
```diff
@@ -41,8 +41,9 @@ def extract(
           digest = @datadog_propagator.extract(carrier)
           return context unless digest
 
-          trace_id = to_otel_id(digest.trace_id)
-          span_id = to_otel_id(digest.span_id)
+          # Converts the {Numeric} Datadog id object to OpenTelemetry's byte array format.
+          trace_id = [format('%032x', digest.trace_id)].pack('H32')
+          span_id = [format('%016x', digest.span_id)].pack('H16')
```
mabdinur, line 46: Fix 2: Ensure otel spans are encoded in a manner that is consistent with the otel sdk: https://github.com/open-telemetry/opentelemetry-ruby/blob/5b2cc8ffe7614c27fd69f0b40d34368f0c06ff53/api/lib/opentelemetry/trace/span_context.rb#L42

marcotc, line 46: The main difference between the Datadog Ruby trace and the OpenTelemetry tracer is that ours store ID numbers as `Numeric`, while OpenTelemetry stores it as a byte array. We both convert IDs to Hex for serialization, the we don't use the same native representation internally.
With this in mind, we can't directly borrow from OpenTelemetry's implementation here, as we are not working with byte arrays, but with `Numeric`.

```suggestion
          # Converts the {Numeric} Datadog id object to OpenTelemetry's byte array format.
          trace_id = [digest.trace_id >> 64, digest.trace_id & 0xFFFFFFFFFFFFFFFF].pack('Q>Q>') # 128-bit unsigned, big-endian integer
          span_id = [digest.span_id].pack('Q>') # 64-bit unsigned, big-endian integer
```

lib/datadog/opentelemetry/sdk/span_processor.rb
marcotc: Setting the `trace_id` is what
```ruby
          if parent_context.trace
            Tracing.send(:tracer).send(:call_context).activate!(parent_context.ensure_trace)
          else
            Tracing.continue_trace!(nil)
          end
```
does above. Any changes to the `trace_id` have to be done on the active trace.

For `span_id`, there's no API today, so I suggest adding a `_span_id` argument to `trace`, so that we can set it.

lib/datadog/opentelemetry/sdk/propagator.rb
mabdinur: I think my suggestion does the same thing but in a more inefficient way. It converts a numeric to a zero padded hexidecimal and then encodes the hex into the binary format. This was my workaround for supporting large unsigned integers of unkown length. I like your approach a lot better. I'll merge this suggestion.

lib/datadog/opentelemetry/sdk/span_processor.rb
mabdinur: I'll pass in the trace_id of the otel span to Tracing.continue_trace!(nil) via trace_digest

mabdinur:
[Profiling tests](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/15051/workflows/38cb8d43-e9e9-45db-bcfc-05fbe8a0723e/jobs/553568 ) are failing due to 64bit unsigned span_ids being converted to signed integers.

@marcotc Is this a bug in the profiler?

lib/datadog/tracing/span_operation.rb
```diff
@@ -25,16 +25,8 @@ class SpanOperation
...
-        :id,
-        :name,
-        :parent_id,
-        :resource,
-        :service,
-        :start_time,
-        :trace_id,
-        :type
+      attr_accessor :id, :trace_id
+      attr_reader :end_time, :name, :parent_id, :resource, :service, :start_time, :type
```
marcotc, line 29: I don't think this change is necessary anymore.

lib/datadog/tracing/tracer.rb
```diff
@@ -113,7 +113,7 @@ def initialize(
       # @param [String] service the service name for this span.
       # @param [Time] start_time time which the span should have started.
       # @param [Hash<String,String>] tags extra tags which should be added to the span.
-      # @param [String] type the type of the span. See {Datadog::Tracing::Metadata::Ext::AppTypes}.
+      # @param [Integer] the id of the new span.
```
marcotc, line 116: It seems like you deleted `type` by mistake 😅

marcotc (approved):
Looks great! Just two small changes to suggested.

marcotc:
> [Profiling tests](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/15051/workflows/38cb8d43-e9e9-45db-bcfc-05fbe8a0723e/jobs/553568) are failing due to 64bit unsigned span_ids being converted to signed integers.

I don't see any changes in your PR that should affect it, can you figure out what line exactly in your PR is triggering the issue?

lib/datadog/tracing/span_operation.rb
mabdinur, line 29:
```suggestion
      attr_reader \
        :end_time,
        :id,
        :name,
        :parent_id,
        :resource,
        :service,
        :start_time,
        :trace_id,
        :type
```

lib/datadog/tracing/tracer.rb
mabdinur, line 116:
```suggestion
      # @param [String] type the type of the span. See {Datadog::Tracing::Metadata::Ext::AppTypes}.
      # @param [Integer] the id of the new span.
```

mabdinur:
> > [Profiling tests](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/15051/workflows/38cb8d43-e9e9-45db-bcfc-05fbe8a0723e/jobs/553568) are failing due to 64bit unsigned span_ids being converted to signed integers.
> 
> I don't see any changes in your PR that should affect it, can you figure out what line exactly in your PR is triggering the issue?


In the [failing test](https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/15051/workflows/38cb8d43-e9e9-45db-bcfc-05fbe8a0723e/jobs/553560) the span_id of `13756630617282998475` is encoded as `-4690113456426553141`. Both numbers have the same 2s complement  representation `1011111011101001010111101010100100100001011010100111000011001011`. 

It seems like the profiler is using signed 64bit integers to correlate spans to threads. However I can't find where this occurs. Looking at the thread collector's implementation it seems like the span_id should be an unsigned integer: https://github.com/DataDog/dd-trace-rb/blob/master/ext/datadog_profiling_native_extension/collectors_thread_context.c#L1145C39-L1145C46. The ruby otel tracer generates span_ids up to 2^64 while the datadog ruby tracer generates span ids up to 2^63 so this wasn't an issue in the past. 

cc: @ivoanjo

ivoanjo:
Taking a look, thanks for the ping!

ivoanjo:
I've pushed a fix in a99f0e51af09d45e57a851640c7407fdb2aa7f80 . Sorry for breaking your PR! I remember thinking "I should correct for the unsigned thing in our tests..." when working on #3466 but clearly ended up not doing it.

As I stated in the commit message -- both the profiler code and the backend know to treat these values as unsigned, and it was really only the test code that also decodes profiles to match on them that wasn't doing the same adjustment.

LMK if I can provide any more help :)

mabdinur:
@marcotc This PR should be good to go

PR #3723, opened 2024-06-18
Add post_install_message in gemspec

TonyCTHsu:
**What does this PR do?**

1. Add post_install_message in gemspec
2. Update unrelease comparison in changelog

### DIFF
--- a/ddtrace.gemspec
+++ b/ddtrace.gemspec
@@ -77,13 +77,12 @@ Gem::Specification.new do |spec|
   spec.extensions = ['ext/datadog_profiling_native_extension/extconf.rb', 'ext/datadog_profiling_loader/extconf.rb']
 
   spec.post_install_message = <<-MSG
-    Thank you for installing ddtrace.
+    Thank you for installing ddtrace. We have released our next major version!
 
-    As of version 2, the `ddtrace` gem has been renamed to `datadog` to better represent the full suite of Datadog products.
+    As of version 2, `ddtrace` gem has been renamed to `datadog`.
     The 1.x series will now only receive maintenance updates for security and critical bug fixes.
 
-    To upgrade, please update your Gemfile to replace `ddtrace` with `datadog`.
-
-    For detailed instructions on migration, see: https://github.com/DataDog/dd-trace-rb/blob/master/docs/UpgradeGuide2.md
+    To upgrade, please replace gem `ddtrace` with gem `datadog`.
+    For detailed instructions on migration, see: https://dtdg.co/ruby-v2-upgrade
   MSG
 end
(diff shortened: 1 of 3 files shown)


ddtrace.gemspec
```diff
@@ -75,4 +75,15 @@ Gem::Specification.new do |spec|
...
+
+  spec.post_install_message = <<-MSG
+    Thank you for installing ddtrace.
+
+    As of version 2, the `ddtrace` gem has been renamed to `datadog` to better represent the full suite of Datadog products.
+    The 1.x series will now only receive maintenance updates for security and critical bug fixes.
+
+    To upgrade, please update your Gemfile to replace `ddtrace` with `datadog`.
+
+    For detailed instructions on migration, see: https://github.com/DataDog/dd-trace-rb/blob/master/docs/UpgradeGuide2.md
```
marcotc, line 87: We should link this to a stable URL (and not `master` branch), because this message will be out in the wild for a long time.

I'm not sure what's the best git reference to pin this to, even the upgrade guide likely evolves overtime.

Or maybe he doesn't change too much that we can just link the specific comment? I'm not super sure.

ddtrace.gemspec
```diff
@@ -75,4 +75,15 @@ Gem::Specification.new do |spec|
...
 
   spec.extensions = ['ext/datadog_profiling_native_extension/extconf.rb', 'ext/datadog_profiling_loader/extconf.rb']
+
+  spec.post_install_message = <<-MSG
+    Thank you for installing ddtrace.
+
+    As of version 2, the `ddtrace` gem has been renamed to `datadog` to better represent the full suite of Datadog products.
+    The 1.x series will now only receive maintenance updates for security and critical bug fixes.
+
+    To upgrade, please update your Gemfile to replace `ddtrace` with `datadog`.
```
marcotc, line 85:
```suggestion
    To upgrade, please use move to the gem `datadog`, instead of `ddtrace`.
```

Do we want to keep the language bundler neutral? In case people are not actually using a Gemfile.

ivoanjo, line 85: I have a suggestion that goes in a bit of another direction from Marco's suggestion. Since as Marco pointed out very correctly "this message will be out in the wild for a long time", I was thinking we could have a very minimalist message, so that people wouldn't get very annoyed by it.

E.g. Yes I'm usually the person that would suggest a nice big message like the one we're using, but I think this situation may deserve a special consideration.

Specifically, I was thinking something like:

> The `ddtrace` gem has been renamed to `datadog`. We strongly recommend upgrading. https://dtdg.co/dd-gem

(The url shortener even allows us to point the message at a different place in the future ).

ivoanjo: The idea here being, you still see the message, but it's not on your face/annoying every time you run bundle install.

ivoanjo: I still think the short-and-sweet message with the redirect link would be the best solution here :)

marcotc (approved):

PR #3725, opened 2024-06-18
Populate ActiveSupport Notification spans as early as possible 

### LINKED ISSUE #2846: ActiveSupport events cause child spans to be mislabeled as service-entry
**Current behaviour**
Spans created by the [Datadog::Tracing::Contrib::ActiveSupport::Notifications::Event](https://github.com/DataDog/dd-trace-rb/blob/7303a3fe39f09768d0f529446326331f15e93465/lib/datadog/tracing/contrib/active_support/notifications/event.rb#L14) module receive their service name, resource name, and tags as the span is finishing.

This means that all child span of the ActiveSupport Notification event have already finished.

This causes service-entry span detection to fail at child spans as it depends on the parent span's service being populated by the the time the children have finished, so that service names can be compared: https://github.com/DataDog/dd-trace-rb/blob/7303a3fe39f09768d0f529446326331f15e93465/lib/datadog/tracing/span_operation.rb#L469

This causes non-service-entry spans to be marked as service-entry in the Datadog UI, making Trace Search return false positives when searching for "Service Entry Spans".

**Expected behaviour**

Service-entry detection should be reliable: child spans should not be marked as service-entry if their service name is the same as their parent, using the best information available.

**Steps to reproduce**

1. Configure `c.tracing.instrument :active_record, service_name: 'ar-service'` (or any custom service name).
1. Create an ActiveRecord trace.
    1. Two spans will be generated: a `#{adapter_name}.query` span, and a `active_record.instantiation`.
1. Inspect that the `active_record.instantiation` span is tagged with the metric `_dd.top_level: 1.0`, despite having the same service name as `#{adapter_name}.query`.

**Environment**

* **ddtrace version: 1.11.1**
* **Ruby version: Any**
* **Operating system: Any**

marcotc:
Fixes #2846

This PR populates spans create it from ActiveSupport Notification spans as early as possible.

Before this PR, a span was started [when the ActiveSupport Notification event started](https://github.com/DataDog/dd-trace-rb/blob/4b1901101c0f4d1033558bd9c0ff2e1c2b963bc7/lib/datadog/tracing/contrib/active_support/notifications/subscription.rb#L80) but the notification callback was only [called when the event ended](https://github.com/DataDog/dd-trace-rb/blob/4b1901101c0f4d1033558bd9c0ff2e1c2b963bc7/lib/datadog/tracing/contrib/active_support/notifications/subscription.rb#L93). This meant that during the event execution, the active span would not have all its fields populated.

This causes issues with children span's service entry detection and service name propagation.

**How to test the change?**
All changes are covered by tests.


lib/datadog/tracing/contrib/active_support/notifications/subscription.rb
```diff
@@ -109,18 +121,20 @@ def subscribers
...
               attr_reader :block
 
-              def initialize(&block)
+              def initialize(block)
                 @block = block
               end
 
               def run(span, name, id, payload)
-                run!(span, name, id, payload)
+                run!(span, name, id, payload) if @block
```
TonyCTHsu, line 129: I think it is more explicit to remove private method `#run!` and change.

```suggestion
                @block.call((span, name, id, payload) if @block
```

lib/datadog/tracing/contrib/active_support/notifications/subscription.rb
```diff
@@ -11,12 +11,19 @@ class Subscription
...
 
-            def initialize(span_name, options, &block)
-              raise ArgumentError, 'Must be given a block!' unless block
+            # @param span_name [String] the operation name for the span
+            # @param options [Hash] options to pass during span creation
+            # @param on_start [Proc] a block to run when the event is fired,
+            #   might not include all required information in the `payload` argument.
+            # @param on_finish [Proc] a block to run when the event has finished processing,
+            #   possibly including more information in the `payload` argument.
+            def initialize(span_name, options, on_start: nil, on_finish: nil)
```
TonyCTHsu, line 20: Would you agree that the interface is more streamlined with `on_start` and `on_finish` being the keys from the `options` hash? 

Oh, this comment also applies to other methods that accept `on_start`, `on_finish` and `options`.

lib/datadog/tracing/contrib/active_support/notifications/event.rb
```diff
@@ -62,6 +64,21 @@ def payload_exception(payload)
                 payload[:exception_object] ||
                   payload[:exception] # Fallback for ActiveSupport < 5.0
               end
+
+              def on_start(_span, _event, _id, _payload); end
+
+              def on_finish(span, _event, _id, payload)
+                record_exception(span, payload)
+              end
```
TonyCTHsu, line 72: 👍 These are making sure `method(:on_start)` and `method(:on_finish)` won't raise exception

lib/datadog/tracing/contrib/action_mailer/events/deliver.rb
```diff
@@ -30,15 +30,20 @@ def span_type
...
               super
 
-              span.type = span_type
+              span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_DELIVER)
+            end
+
+            def on_finish(span, event, _id, payload)
+              super
+
+              span.resource = payload[:mailer]
```
TonyCTHsu, line 42: Is there a reason setting the resource in `on_finish` instead of `on_start` ? Would it because of the data availability?

lib/datadog/tracing/contrib/active_support/notifications/subscription.rb
marcotc: Done!

lib/datadog/tracing/contrib/active_support/notifications/subscription.rb
marcotc: `options` are span options, and passed to `span = Tracing.trace(@span_name, **@options)`, we can't pass `on_start` to `Tracing.trace`.

I can rename options to `span_options` to make it clear.

marcotc: I renamed it to `span_options`.

lib/datadog/tracing/contrib/action_mailer/events/deliver.rb
marcotc: Correct! The data is not available on `on_start`. I'll document that.

TonyCTHsu (approved):

ivoanjo:
:heart: Adding info to spans as early as possible also opens up possibilities for the profiler to do more with this data so thanks for all this work :)

PR #3730, opened 2024-06-21
HTTP propagator: Better handling of invalid data

marcotc:
This PR improve the handling of trace propagation when the parent span_id is not present. In such case, the trace is still propagated without a span_id, which is a valid trace propagation payload.
The distributed tracing will have the correct correlation between the current service with the downstream service, but the downstream service parenting will likely be incorrect. This is still more desirable than completely losing trace correlation.

Also in this PR, there is the better handling of the scenario when a `TraceDigest` is not populated with a trace_id, which cannot happen in a digest created by the tracer, but can happen with users are manually testing the public method `Datadog::Tracing::Contrib::HTTP.inject(digest, env)`.
In such cases, we were printing an error message, as this was an unhandled case that raised a legitimate Ruby exception:
```
[2024-06-06T10:47:09.334345 #42] ERROR -- ddtrace: [ddtrace] (/usr/src/app/vendor/bundle/ruby/3.3.0/gems/ddtrace-1.19.0/lib/datadog/tracing/distributed/propagation.rb:65:in `rescue in block in inject!') Error injecting distributed trace data. Cause: can't convert nil into Integer Location: /usr/src/app/vendor/bundle/ruby/3.3.0/gems/ddtrace-1.19.0/lib/datadog/tracing/distributed/trace_context.rb:124:in `format'
```
 This is now handled early in code, and logged as a debug message.


TonyCTHsu (approved):

bernardobarreto (commented):
Thank you!

PR #3734, opened 2024-06-25
[SDTEST-409] Telemetry metrics data model

anmarchenko:
**What does this PR do?**
This PR is a first step towards full metrics support for internal Datadog telemetry. 

It handles data model for the future metrics implementation:
- Adds `generate-metrics` and `distributions` events to telemetry
- Adds Ruby classes for every metric type we are going to support

This PR does not handle (subject for subsequent PRs):
- metrics buffering in memory
- multi-threaded synchronization
- sending metrics events to the backend

**Motivation:**
Add telemetry metrics support in bite-sized and digestible fashion.

**Additional Notes:**
Pay attention to `#track` method in each of the metrics implementations: this is the aggregations logic that is different for every metric type.

**How to test the change?**
Only unit tests currently.

### DIFF
--- a/lib/datadog/core/telemetry/event.rb
+++ b/lib/datadog/core/telemetry/event.rb
@@ -286,6 +286,26 @@ module Datadog
             'app-closing'
           end
         end
+
+        # Telemetry class for the 'generate-metrics' event
+        class GenerateMetrics < Base
+          def type
+            'generate-metrics'
+          end
+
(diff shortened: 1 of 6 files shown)


p-datadog (commented):
I read the diff and I have no objections, though I am not familiar with the product enough to leave a checkmark at this time.

marcotc (approved):

PR #3742, opened 2024-06-27
[SDTEST-409] Add metrics management capabilities

anmarchenko:
**What does this PR do?**
Adds entities to collect metrics and flush them to telemetry events:
- `MetricsCollection` aggregates metrics per **namespace** and on demand flushes them to a queue passed as a param - this is going to be `Telemetry::Worker`. At most 2 events are created on flush: one for metrics, one for distributions. The methods of `MetricsCollection` are synchronized, protecting access to at most one thread at a time.
- `MetricsManager` is a central piece of metrics aggregation: it manages collections, one per namespace. It synchronizes access to a hashmap with all collections, but allows concurrent access to a collection after it's fetched from the map. The method `flush!` locks access to the collections map and flushes each collection in order.

The concerns of managing metrics and managing namespaces are separated for the following reasons:
- makes it easier to reason about concurrency and thread-safety of each component in isolation
- allows concurrent access to several metrics namespaces at once, i.e. "civisibility" and "tracing" workers can add metrics without being blocked by each other.

**Motivation:**
Next small step to achieving metrics support for internal DD telemetry.

**How to test the change?**
Only unit tests for now (including several concurrency tests).

### DIFF
--- a/lib/datadog/core/telemetry/metric.rb
+++ b/lib/datadog/core/telemetry/metric.rb
@@ -7,39 +7,38 @@ module Datadog
       module Metric
         # Base class for all metric types
         class Base
-          attr_reader :name, :tags, :values, :common, :interval
+          attr_reader :name, :tags, :values, :common
 
           # @param name [String] metric name
           # @param tags [Array<String>|Hash{String=>String}] metric tags as hash of array of "tag:val" strings
           # @param common [Boolean] true if the metric is common for all languages, false for Ruby-specific metric
-          # @param interval [Integer] metrics aggregation interval in seconds
-          def initialize(name, tags: {}, common: true, interval: nil)
+          def initialize(name, tags: {}, common: true)
             @name = name
             @values = []
             @tags = tags_to_array(tags)
             @common = common
-            @interval = interval
           end
 
           def id
             @id ||= "#{type}::#{name}::#{tags.join(',')}"
           end
 
-          def track(value); end
+          def track(value)
+            raise NotImplementedError, 'method must be implemented in subclasses'
+          end
 
-          def type; end
+          def type
+            raise NotImplementedError, 'method must be implemented in subclasses'
+          end
 
           def to_h
-            # @type var res: Hash[Symbol, untyped]
-            res = {
+            {
               metric: name,
               points: values,
               type: type,
               tags: tags,
               common: common
             }
-            res[:interval] = interval if interval
-            res
           end
 
           private
@@ -48,6 +47,29 @@ module Datadog
             return tags if tags.is_a?(Array)
 
             tags.map { |k, v| "#{k}:#{v}" }
+          end
+        end
+
+        # Base class for metrics that require aggregation interval
+        class IntervalMetric < Base
+          attr_reader :interval
+
+          # @param name [String] metric name
+          # @param tags [Array<String>|Hash{String=>String}] metric tags as hash of array of "tag:val" strings
+          # @param common [Boolean] true if the metric is common for all languages, false for Ruby-specific metric
+          # @param interval [Integer] metrics aggregation interval in seconds
+          def initialize(name, interval:, tags: {}, common: true)
+            raise ArgumentError, 'interval must be a positive number' if interval.nil? || interval <= 0
+
+            super(name, tags: tags, common: common)
+
+            @interval = interval
+          end
+
+          def to_h
+            res = super
+            res[:interval] = interval
+            res
           end
         end
 
@@ -84,7 +106,7 @@ module Datadog
         # A gauge type takes the last value reported during the interval. This type would make sense for tracking RAM or
         # CPU usage, where taking the last value provides a representative picture of the host’s behavior during the time
         # interval.
-        class Gauge < Base
+        class Gauge < IntervalMetric
           TYPE = 'gauge'
 
           def type
@@ -104,10 +126,10 @@ module Datadog
 
         # The rate type takes the count and divides it by the length of the time interval. This is useful if you’re
         # interested in the number of hits per second.
-        class Rate < Base
+        class Rate < IntervalMetric
           TYPE = 'rate'
 
-          def initialize(name, tags: {}, common: true, interval: nil)
+          def initialize(name, interval:, tags: {}, common: true)
             super
 
             @value = 0.0
@@ -119,15 +141,7 @@ module Datadog
 
           def track(value = 1.0)
             @value += value
-
-            rate =
-              if interval && interval.positive?
-                @value / interval
-              else
-                0.0
-              end
-
-            @values = [[Time.now.to_i, rate]]
+            @values = [[Time.now.to_i, @value / interval]]
             nil
           end
         end
--- a/lib/datadog/core/telemetry/metric.rb
+++ b/lib/datadog/core/telemetry/metric.rb
@@ -5,10 +5,6 @@ module Datadog
     module Telemetry
       # Telemetry metrics data model (internal Datadog metrics for client libraries)
       module Metric
-        def self.metric_id(type, name, tags = [])
-          "#{type}::#{name}::#{tags.join(',')}"
-        end
-
         # Base class for all metric types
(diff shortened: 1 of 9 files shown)


lib/datadog/core/telemetry/metrics_collection.rb
```diff
@@ -0,0 +1,92 @@
...
+
+          @metrics = {}
+          @distributions = {}
+        end
+
+        def inc(metric_name, value, tags: {}, common: true)
+          metric = Metric::Count.new(metric_name, tags: tags, common: common)
+
+          @mutex.synchronize do
+            metric = fetch_or_add_metric(metric, @metrics)
```
TonyCTHsu, line 27: Nitpick: It looks weird to me passing instance variable to its private method, which already has access to the state of the object. My guess is that you want to make `fetch_or_add_metric` generic for both `@metrics` and `distributions` 

However, I would actually prefer separating these two.

```ruby
 @mutex.synchronize do
  (@metrics[metric.id] ||= metric).tap { |m| m.inc(value) }
end
```
It also put them closer to `@mutex` preventing usage without it.

I would also be fine if you pass instance variable to the private method and refactor `fetch_or_add_metric` to include the `@mutex.sychronization`

lib/datadog/core/telemetry/metrics_manager.rb
```diff
@@ -0,0 +1,78 @@
...
+          @mutex = Mutex.new
+
+          @collections = {}
+        end
+
+        def inc(namespace, metric_name, value, tags: {}, common: true)
+          return unless @enabled
+
+          collection = fetch_or_create_collection(namespace)
+          collection.inc(metric_name, value, tags: tags, common: common)
```
TonyCTHsu, line 24: Does `collection.inc` need to be guarded by mutex?

lib/datadog/core/telemetry/metrics_collection.rb
anmarchenko: I don't have strong opinion on this, your approach works for me too

lib/datadog/core/telemetry/metrics_manager.rb
anmarchenko: The mutable state of the given collection is protected by collection's mutex inside the method.

ivoanjo (commented):
Two quick suggestions, I plan to do a bigger pass in a few minutes :)

lib/datadog/core/telemetry/metric.rb
```diff
@@ -25,6 +21,10 @@ def initialize(name, tags: {}, common: true, interval: nil)
             @interval = interval
           end
 
+          def id
+            @id ||= "#{type}::#{name}::#{tags.join(',')}"
+          end
+
           def track(value); end
```
ivoanjo, line 28: Minor: Not that it has mattered a lot so far, but maybe we should add a `raise 'Must to be overridden by subclasses'` here?

lib/datadog/core/telemetry/metric.rb
```diff
@@ -116,8 +120,15 @@ def type
...
+            rate =
+              if interval && interval.positive?
+                @value / interval
+              else
+                0.0
+              end
+
             @values = [[Time.now.to_i, rate]]
+            nil
           end
```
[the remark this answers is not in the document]
anmarchenko: this is true: I think rate metric could not be used without interval

lib/datadog/core/telemetry/metrics_manager.rb
```diff
@@ -0,0 +1,78 @@
...
+          collection = fetch_or_create_collection(namespace)
+          collection.distribution(metric_name, value, tags: tags, common: common)
+        end
+
+        def flush!(queue)
+          return unless @enabled
+
+          @mutex.synchronize do
+            @collections.each_value { |col| col.flush!(queue) }
+          end
```
ivoanjo, line 60: Minor: I wonder if it would be worth doing something like:

```ruby
collections = @mutex.synchronize { @collections.values }
collections.each_value { ... }
```

this would allow most threads to continue to report metrics, concurrently with flushing.

lib/datadog/core/telemetry/metrics_collection.rb
anmarchenko: this one looks very clean, I will try it

spec/datadog/core/telemetry/metrics_collection_spec.rb
```diff
@@ -0,0 +1,323 @@
...
+        values = payload[:series].map { |s| s[:points] }.sort
+        expect(values).to eq([[5, 6], [5, 7]])
+      end.once
+
+      collection.flush!(queue)
+
+      expect(distributions.size).to eq(0)
+    end
+
+    it 'does not loose metrics when running in multiple threads' do
```
ivoanjo, line 295: Minor:

```suggestion
    it 'does not lose metrics when running in multiple threads' do
```

ivoanjo (approved):
So I kinda wonder if there's some simplifications we can to this -- e.g. I understand on purpose this is a very generic thing, but if this very generic thing will only be used in a more specific way, it may be worth the lower overhead/complexity of making it a bit less generic.

With that said, even in its current state I think it's in reasonable shape, so I'll leave it to you to decide how much to iterate on vs just going with it.

lib/datadog/core/telemetry/metrics_collection.rb
ivoanjo: Ah wait, there's a bug in my code! It should be

```ruby
def fetch_or_add_metric(metric)
  @mutex.synchronize do
    yield (collection[metric.id] ||= metric)
  end
  nil
end

File: lib/datadog/core/telemetry/metrics_manager.rb [hunk 7]
```diff
@@ -0,0 +1,78 @@
...
+      class MetricsManager
+        attr_reader :enabled
+
+        def initialize(aggregation_interval:, enabled:)
+          @interval = aggregation_interval
+          @enabled = enabled
+          @mutex = Mutex.new
+
+          @collections = {}
+        end
```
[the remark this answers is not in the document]
anmarchenko: We don't create namespaces dynamically, but also they are managed outside of this library in internal telemetry metrics definition file: 

https://github.com/DataDog/dd-go/blob/prod/trace/apps/tracer-telemetry-intake/telemetry-metrics/static/common_metrics.json

The namespaces examples are: 
- general
- tracers
- civisibility

I did not want to duplicate this information in this library, so I went with a way to just start reporting metrics for any new namespace that might be created.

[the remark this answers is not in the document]
ivoanjo: I see. :thinking: seeing there's only 3 makes me think we could list them on our side as well and be done with it -- thus avoiding the locking on the manager class (and the overhead of it, including in the testing with multiple threads).

We could add a `raise` if a namespace is unknown, so any lib developers would immediately know they need to update the list if they want to start submitting to the new namespace.

Having said that, I'll stand by what I said on my approval note -- I'd go with the simplification, but I leave the final decision to you!

lib/datadog/core/telemetry/metrics_collection.rb
TonyCTHsu: >...maybe this was what Tony had in mind?

You read my mind.

lib/datadog/core/telemetry/metrics_manager.rb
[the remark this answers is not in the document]
anmarchenko: no, sorry, there are more than 3, I was just lazy to copy paste them all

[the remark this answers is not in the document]
anmarchenko: there is also:
- appsec
- iast
- profilers

So 6 overall, also not many. I would like to keep it as is, because js and python also use similar approach and I really don't want to support the list of namespaces on our side.

TonyCTHsu (approved):

PR #3735, opened 2024-06-25
Run specs on GitHub Actions

lloeki:
**What does this PR do?**

Guess.

**Motivation:**

Hold my beer.

**Additional Notes:**

Trying stuff. May or may not work. But it allows me to test https://github.com/DataDog/images-rb without breaking CircleCI and something may or may not come out of it.

**How to test the change?**

Wait for it.

PR #3744, opened 2024-06-28
Extract matrix from Rakefile

lloeki:
**What does this PR do?**

Extract the test matrix from `Rakefile`

**Motivation:**

- This makes the matrix usable by external tools.
- We may evolve this descriptive matrix format in any way we see fit, including it not being Ruby

**Additional Notes:**

- An example tool is [this `matrix.rb` script](https://github.com/DataDog/dd-trace-rb/blob/428d36bf4a60b9ec51c162893469f8c6f03c3466/matrix.rb) which allows querying and transforming the matrix.
- `TEST_METADATA` should probably not be a top-level constant but task-scoped, and actually tasks should be extracted from this `Rakefile` into their own `tasks/foo.rake` but that's WAY out of scope. This is the simplest, drop-in change.

**How to test the change?**

CI should still pass as it did before.


TonyCTHsu (approved):

lloeki:
That one is really odd, I don't get why it insists on opening it as `US-ASCII`.

```
rake aborted!
SyntaxError: (eval):6: invalid multibyte char (US-ASCII)
```

It's also environment dependent in some way, as it doesn't happen locally.

ivoanjo:
Should we maybe move this into `tasks/` or `appraisal/`? We already have a lot of stuff on the root of the repo, and "Matrixfile" is not a standard so it adds a bit of confusion about what it is and what it's used for.

lloeki:
@ivoanjo Sure I'll find some place (although neither `tasks` nor `appraisal` fit the bill in my mind). A sI mentioned there are certainly deeper refactorings that should be done and it would be a better place to fold such a change.

I did not want this to feature-creep so this was a rapid cherrypick from https://github.com/DataDog/dd-trace-rb/pull/3735 to conceptually get it extracted.

ivoanjo:
Can I convince you to move it as-is (possibly adding a `.rb` as well) before other future refactorings?

PR #3749, opened 2024-07-01
[SDTEST-409] Send telemetry events in batches

anmarchenko:
**What does this PR do?**
Optimises telemetry events sending by sending them in batches using `message-batch` event type.

**Motivation:**
I expect the number of telemetry events to grow as soon as products start using metrics. We can save time by batching the list of events into a single event.

**How to test the change?**
System tests pass for telemetry events.

Logs from rails app with DD_TRACE_DEBUG=1:

```bash
D, [2024-07-01T16:08:58.970222 #20478] DEBUG -- datadog: [datadog] (/Users/andrey.marchenko/p/dd-trace-rb/lib/datadog/core/telemetry/worker.rb:88:in `flush_events') Sending 2 telemetry events
D, [2024-07-01T16:08:59.137402 #20478] DEBUG -- datadog: [datadog] (/Users/andrey.marchenko/p/dd-trace-rb/lib/datadog/core/telemetry/emitter.rb:29:in `request') Telemetry sent for event `message-batch` (code: 202)
```

### DIFF
--- a/lib/datadog/core/telemetry/event.rb
+++ b/lib/datadog/core/telemetry/event.rb
@@ -316,6 +316,8 @@ module Datadog
 
         # Telemetry class for the 'message-batch' event
         class MessageBatch
+          attr_reader :events
+
           def type
             'message-batch'
           end
--- a/lib/datadog/core/telemetry/event.rb
+++ b/lib/datadog/core/telemetry/event.rb
@@ -313,6 +313,26 @@ module Datadog
             'distributions'
           end
         end
+
+        # Telemetry class for the 'message-batch' event
+        class MessageBatch
+          def type
+            'message-batch'
+          end
+
+          def initialize(events)
+            @events = events
+          end
+
+          def payload(seq_id)
+            @events.map do |event|
+              {
+                request_type: event.type,
+                payload: event.payload(seq_id),
+              }
+            end
+          end
+        end
       end
     end
   end
--- a/lib/datadog/core/telemetry/event.rb
+++ b/lib/datadog/core/telemetry/event.rb
@@ -3,7 +3,16 @@
 module Datadog
   module Core
     module Telemetry
+      # Collection of telemetry events
       class Event
+        extend Core::Utils::Forking
+
+        # returns sequence that increments every time the configuration changes
+        def self.configuration_sequence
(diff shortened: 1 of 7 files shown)


lib/datadog/core/telemetry/event.rb
```diff
@@ -313,6 +313,28 @@ def type
...
+
+          def initialize(events)
+            @events = events
+          end
+
+          def payload(seq_id)
+            @events.map do |event|
+              {
+                request_type: event.type,
+                payload: event.payload(seq_id),
```
p-datadog, line 333: Is it OK that seq id is the same for all of the events?

anmarchenko: Great question, I need to look it up in the spec.

anmarchenko: The spec says nothing about it and also it does not mention that `seq_id` should be part of telemetry events payloads at all (it is supposed to be part of common headers)! I will investigate why our payload generation code even depends on seq_id...

anmarchenko: I found where this seq_id is defined in specs: https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v2/SchemaDocumentation/Schemas/conf_key_value.md
```
This field should incremented each time a new configuration key-value pair is applied in order to track the active set of configurations. It can be a global value across all configurations keys or local to each key..
```

This seq_id is in fact different from the seq_id that is used for events:
https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v2/development.md#required-top-level-json-keys

```
Counter that should be auto incremented every time an API call is being made (our pipeline may not always work in a sequential order.This field will help us identify situations where some messages were being lost. seq_id should start at 1 so that the go backend can detect between non set fields.
```

I will see how to decouple them from each other in our implementation

anmarchenko: Solved this by introducing another sequence in Core::Telemetry::Event class for the configuration changes

TonyCTHsu (approved):

PR #3754, opened 2024-07-03
Add quick link to support in issue creation form

ivoanjo:
**What does this PR do?**

This PR adds a quick link to talk to Datadog support as part of the issue creation flow. Of course, folks still can (and are very much encouraged to) file tickets as needed, but often support will be faster and can delve into account specifics to provide better help.

I borrowed this idea from https://github.com/DataDog/test-template/issues/new/choose and you can see it working live in there.

**Motivation:**

Improve support experience for Ruby users.

**Additional Notes:**

I've also added some cool emoji to the existing issue types.

**How to test the change?**

Other than looking at the template repo, I'm not sure if there's a way to see these changes live before we merge them.

But, on the other hand, it's very easy to check and revert, so the risk is very low.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,5 +1,5 @@
 ---
-name: Bug report
+name: 🐛 Bug report
 about: File a bug report
 title: ''
 labels: community, bug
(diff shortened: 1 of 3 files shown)


anmarchenko (approved):

ivoanjo:
It's live:

> [image](https://github.com/DataDog/dd-trace-rb/issues/new/choose)

PR #3756, opened 2024-07-03
Use minimalistic post-install upgrade message

ivoanjo:
**What does this PR do?**

This PR is a follow-up on #3723.

For the 1.23.3 dd-trace-rb release, we included a quite detailed post-install message to tell customers that the new release was out:

```
Thank you for installing ddtrace. We have released our next major version!

As of version 2, `ddtrace` gem has been renamed to `datadog`.
The 1.x series will now only receive maintenance updates for security and critical bug fixes.

To upgrade, please replace gem `ddtrace` with gem `datadog`.
For detailed instructions on migration, see: https://dtdg.co/ruby-v2-upgrade
```

My thinking is -- some customers may not be able to move to 2.x as quickly as they wanted, and thus seeing that message again and again may start to be slightly annoying.

Thus, this PR replaces with with the minimalistic

```
The ddtrace gem has been renamed to datadog. We recommend upgrading: https://dtdg.co/ruby-v2-upgrade
```

**Motivation:**

Continue suggesting to customers that the should upgrade, while at the same time not being too annoying about it.

**Additional Notes:**

It actually may not a half-bad idea that we released a version with the big message, and then shortened it.

Specifically, customers that mostly keep up-to-date will probably see that message quite quickly, and thus upgrade.

For the long tail of customers that may need to remain on 1.x, the more minimalistic message that will go out in a few weeks/months on a future 1.x maintenance release will continue to remind them.

**How to test the change?**

```bash
$ bundle exec rake build
 # ...

ddtrace 1.23.3 built to pkg/ddtrace-1.23.3.gem.

$ gem install pkg/ddtrace-1.23.3.gem
Building native extensions. This could take a while...
    The ddtrace gem has been renamed to datadog. We recommend upgrading: https://dtdg.co/ruby-v2-upgrade
```

### DIFF
--- a/ddtrace.gemspec
+++ b/ddtrace.gemspec
@@ -77,12 +77,6 @@ Gem::Specification.new do |spec|
   spec.extensions = ['ext/datadog_profiling_native_extension/extconf.rb', 'ext/datadog_profiling_loader/extconf.rb']
 
   spec.post_install_message = <<-MSG
-    Thank you for installing ddtrace. We have released our next major version!
-
-    As of version 2, `ddtrace` gem has been renamed to `datadog`.
-    The 1.x series will now only receive maintenance updates for security and critical bug fixes.
-
-    To upgrade, please replace gem `ddtrace` with gem `datadog`.
-    For detailed instructions on migration, see: https://dtdg.co/ruby-v2-upgrade
+    The ddtrace gem has been renamed to datadog. We recommend upgrading: https://dtdg.co/ruby-v2-upgrade
   MSG
 end
--- a/ddtrace.gemspec
+++ b/ddtrace.gemspec
@@ -77,6 +77,7 @@ Gem::Specification.new do |spec|
   spec.extensions = ['ext/datadog_profiling_native_extension/extconf.rb', 'ext/datadog_profiling_loader/extconf.rb']
 
   spec.post_install_message = <<-MSG
-    The ddtrace gem has been renamed to datadog. We recommend upgrading: https://dtdg.co/ruby-v2-upgrade
+    The ddtrace gem has been renamed to datadog in version 2. Upgrading is easy: https://dtdg.co/ruby-v2-upgrade
+    ddtrace 1.x will only receive security updates and critical bug fixes.
   MSG
 end


ddtrace.gemspec
```diff
@@ -77,12 +77,6 @@ Gem::Specification.new do |spec|
...
 
   spec.post_install_message = <<-MSG
-    Thank you for installing ddtrace. We have released our next major version!
-
-    As of version 2, `ddtrace` gem has been renamed to `datadog`.
-    The 1.x series will now only receive maintenance updates for security and critical bug fixes.
-
-    To upgrade, please replace gem `ddtrace` with gem `datadog`.
-    For detailed instructions on migration, see: https://dtdg.co/ruby-v2-upgrade
+    The ddtrace gem has been renamed to datadog. We recommend upgrading: https://dtdg.co/ruby-v2-upgrade
```
marcotc, line 80: I like the original message more, because it captures the "what" in a contextual way (it is tailored for someone installing this version of ddtrace 1.x) and it explains what the risks are immediately ("1.x series will now only receive maintenance updates").
IMO, the original message allows me to make a decision without the indirection. The user is then invited to follow the link if they deem that an upgrade is warranted.

The motivation
> Continue suggesting to customers that the should upgrade, while at the same time not being too annoying about it.

makes sense, I'm just not sure if trumps having a well-described message.

TonyCTHsu: > but I also don't want to start a bikeshed on this topic

Nobody enjoy bikeshedding.

I prefer the original message, but I don't mind improving the brevity as long as the context is captured.

Context: 

1. Major version released
2. Gem name changed
3. The risk of not upgrading: Explain what maintenance mode means.
4. Info about upgrading: Link `https://dtdg.co/ruby-v2-upgrade`

> What do you think of this middle ground:

I would like to see the proposed message capturing (1)

No matter how short the message become, [user can still be annoyed with a one-liner post install message,](https://github.com/tpope/gem-shut-the-fuck-up) consider using bundler `ignore_messages` configuration.

ivoanjo: > I would like to see the proposed message capturing (1)

Maybe

```
Version 2 of the ddtrace gem has been renamed to datadog. Upgrading is easy: https://dtdg.co/ruby-v2-upgrade
The old 1.x ddtrace gem will only receive maintenance updates for security and critical bug fixes.
```

?

> No matter how short the message become, [user can still be annoyed with a one-liner post install message,](https://github.com/tpope/gem-shut-the-fuck-up) consider using bundler ignore_messages configuration.

It's a good point! I didn't know you could disable these. On the other hand, I was kinda trying to avoid annoying people in the first place, rather than annoying them enough that they google how to get rid of the message :rofl: 

I guess this raises one possibility: we could keep the current more detailed message and add a footer to tell customers something like

```
(To disable this message, use `bundle config ignore_messages.ddtrace true`)
```

Although I prefer the minimalist version, I think this would also solve the problem I'm trying to address.

TonyCTHsu: This is my take trying to shorten it. 

```
The next major version (2.x) of `ddtrace` has been released and the gem has been renamed to `datadog`. 
The `ddtrace` gem will only receive maintenance updates for security and critical bug.
We recommend upgrading: https://dtdg.co/ruby-v2-upgrade
```

I will leave it to others to decide whether to include the footer for ignoring message from bundler.

ivoanjo: I was aiming for 1 line, with 2 line being the compromise. I'm not entirely a fan of spilling into 3 lines, at that point I'm not sure we're really making it "minimalistic".

TonyCTHsu: I merged the second and third line.

```
The next major version (2.x) of `ddtrace` has been released and the gem has been renamed to `datadog`. 
The `ddtrace` gem will only receive maintenance update for security and critical bugs. We recommend upgrading: https://dtdg.co/ruby-v2-upgrade
```

marcotc: I think this suggestion captures all the key points and is relatively short (this was suggested by Ivo in this thread, I tweak it just a bit to be shorter):
```
The ddtrace gem has been renamed to datadog in version 2. Upgrading is easy: https://dtdg.co/ruby-v2-upgrade
ddtrace 1.x will only receive security updates and critical bug fixes.
```

marcotc (approved):
Give the input from @TonyCTHsu, @ivoanjo, and myself, I trust you (@ivoanjo) to merge whatever version of the message you see fit, I don't personally see the need to dwell on this for any longer.

p-datadog (approved):
I agree with the sentiment of giving more detail than the 1 line and approve the 2 line version that you all decide to use.

PR #3759, opened 2024-07-05
[PROF-10112] Simplify method names from actionview templates in profiler

ivoanjo:
**What does this PR do?**

Rails's ActionView likes to dynamically generate method names with suffixed hashes/ids, resulting in methods with names such as `_app_views_layouts_explore_html_haml__2304485752546535910_211320`.

This makes these stacks not aggregate well, as well as being not-very-useful data.

This PR modifies the stack collector to detect such methods and simplify them by slicing the hash/id suffix.

**Motivation:**

Improve quality of aggregated flamegraph.

**Additional Notes:**

N/A

**How to test the change?**

Change includes test coverage. Also, all of our internal Rails test apps use templates, so by searching by methods with `erb` or `haml` we'll be able to validate they no longer end with the ids.

### DIFF
--- a/ext/datadog_profiling_native_extension/collectors_stack.c
+++ b/ext/datadog_profiling_native_extension/collectors_stack.c
@@ -34,6 +34,7 @@ static VALUE _native_sample(
 );
 static void maybe_add_placeholder_frames_omitted(VALUE thread, sampling_buffer* buffer, char *frames_omitted_message, int frames_omitted_message_size);
 static void record_placeholder_stack_in_native_code(sampling_buffer* buffer, VALUE recorder_instance, sample_values values, sample_labels labels);
+static void maybe_trim_template_random_ids(ddog_CharSlice *name_slice, ddog_CharSlice *filename_slice);
 
 void collectors_stack_init(VALUE profiling_module) {
   VALUE collectors_module = rb_define_module_under(profiling_module, "Collectors");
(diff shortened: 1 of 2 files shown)


p-datadog (approved):
Do you expect any string without `_app_` prefix to be altered by this logic? If this is specifically for Rails maybe check for this prefix?

ivoanjo:
> Do you expect any string without _app_ prefix to be altered by this logic? If this is specifically for Rails maybe check for this prefix?

This is a good point! I double-checked and the answer is actually that the prefix is the path to the template, .e.g `_app_views_layouts_explore_html_haml__2304485752546535910_211320` is `app/views/layouts/explore.html.haml`.

And because it's a path, it's actually possible to change it. By using `Rails.application.config.paths['app/views'].unshift(Rails.root.join('testcustom', 'views'))` and then moving the file I got a frame named `_testcustom_views_pages_about_html_erb` for a test app I spun up.

So while I don't think it'll be very common that the template will come from outside the `app` folder, it can in rare situations, so I'm leaning towards keeping the more generic version.

I also don't think it would particularly be faster -- since we already have the file name filter (not ".rb") + most methods don't end with numbers, this should be enough to filter out almost all methods which are not templates.

ivoanjo:
I've gone ahead and merged the PR, but LMK if you think I should give another look at focusing only on the `_app_` prefix :pray: :bow:

p-datadog:
My concern was that if someone has their methods named with numeric suffixes that happen to match Rails' generated method names that this PR is considering, they would have their methods unexpectedly renamed. But I wouldn't expect such naming to be very common either.

ivoanjo:
> My concern was that if someone has their methods named with numeric suffixes that happen to match Rails' generated method names that this PR is considering, they would have their methods unexpectedly renamed. But I wouldn't expect such naming to be very common either.

Yeah, they'd need to be quite unlucky:
* Ends with `__` + number + `_` + number
* Filename is not `.rb`

Hopefully the current version is good enough; if someone ever triggers this by accident we'll have examples on how to tighten the pattern to not match the accidental one as well :)

PR #3762, opened 2024-07-05
[Docs] Clarify description for propagation_extract_first

brett0000FF:
**What does this PR do?**

Update description for `tracing.distributed_tracing.propagation_extract_first`.

**Motivation:**

Feedback that "exiting" is confusing.

**Additional Notes:**

**How to test the change?**


maycmlee (approved):

p-datadog (approved):
The change proposed in this PR makes sense to me, I would additionally mention what would happen if the "extract first" setting is not set (which format would be used if there are multiple ones found).

TonyCTHsu (approved):

brett0000FF:
> The change proposed in this PR makes sense to me, I would additionally mention what would happen if the "extract first" setting is not set (which format would be used if there are multiple ones found).

@p-datadog - Good feedback! I agree it'd be helpful to include information for when this option is set to false. In that case, would it use the last valid propagation format it finds? I'm not sure, so any clarification would be helpful. Thanks!

p-datadog:
@brett0000FF I actually don't know which is in part why I mentioned this (I do not work on tracer).

marcotc (approved):

PR #3768, opened 2024-07-09
[SDTEST-409] Telemetry metrics support

anmarchenko:
**What does this PR do?**
This is the final PR to provide metrics support to internal library telemetry.

Notable changes include:
- `Telemetry::Component` now has a public API to track metrics (`#inc`, `#dec`, `#gauge`, `#rate`, `#distribution`)
- `Telemetry::Worker` and `Telemetry::Component` now have shared access to `MetricsManager`
- new settings are added: `DD_TELEMETRY_METRICS_ENABLED` to be able to disable sending metrics separately from telemetry, `DD_TELEMETRY_METRICS_AGGREGATION_INTERVAL` is internal-only setting to change interval to aggregate and send out metrics (10 seconds by default as defined by spec)
- `Telemetry::Worker` now maintains heartbeat ticker: we aggregate metrics every 10 seconds but we send heartbeat events every 60 seconds, so it takes `(heartbeat_interval / metrics_aggregation_interval)` ticks for every heartbeat
- Flushing metrics logic is changed: we actually need to flush metric events to an array to be able to send them out in the same pass of `Worker#perform` - this is needed to be able to flush everything on shutdown

**Additional Notes:**
Usage in your products:

```ruby
telemetry = Datadog.send(:components)
telemetry.inc('tracers', 'spans_created', 10, tags: { 'integration_name' => 'pg' })
```

**How to test the change?**
Unit tests are provided. Please tell me if we have any product that would like to use it to submit metrics.


marcotc (approved):
Awesome work, @anmarchenko!

PR #3776, opened 2024-07-10
feat(otel): add support for span events

mabdinur:
**What does this PR do?**

- Ensures span events generated via the OpenTelemetry SDK are copied from OpenTelemetry Spans to the underlying Datadog span. With change opentelemetry span events are no longer ignored. These events will be sent to the Datadog Agent.
- Implements `Span.record_exception`. This ensures the status of the underlying  Datadog span is updated when an exception is [recorded](https://github.com/open-telemetry/opentelemetry-ruby/blob/35abeac8653359aebd0231c94a23587ad4ee372e/sdk/lib/opentelemetry/sdk/trace/span.rb#L164) on an OpenTelemetry Span.

**Motivation:**

The OpenTelemetry API supports setting Events on spans. Currently these events are ignored by the ddtrace library. To fully support the OpenTelemetry API the ddtrace library must support sending span events to the Datadog Agent.

**Additional Notes:**

**How to test the change?**

Unit tests + https://github.com/DataDog/system-tests/pull/2735


lib/datadog/tracing/span_event.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'time'
+
+# frozen_string_literal: true
+
```
mabdinur, line 6:
```suggestion

```

lib/datadog/tracing/span_event.rb
```diff
@@ -0,0 +1,39 @@
...
+        attributes: nil,
+        time_unix_nano: nil
+      )
+        @name = name
+        @attributes = attributes&.map { |key, val| [key, val.to_s] }&.to_h || {}
+        @time_unix_nano = time_unix_nano || Core::Utils::Time.now.to_f * 1e9
+      end
+
+      def to_hash
+        h = { :name => @name, :time_unix_nano => @time_unix_nano }
```
mabdinur, line 33:
```suggestion
        h = { name: @name, time_unix_nano: @time_unix_nano }
```

spec/datadog/opentelemetry_spec.rb
```diff
@@ -626,6 +626,41 @@
...
+        start_span.add_event('Exception was raised!', **event_options)
+        start_span.finish
+      end
+
+      let(:start_span) { otel_tracer.start_span('start-span', **span_options) }
+      let(:active_span) { Datadog::Tracing.active_span }
+
+      context 'with name, attributes and timestamp' do
+        let(:event_options) do
+          { attributes: { 'raised' => false, 'handler' => 'default', 'count' => 1 }, timestamp: 17206369349 }
```
mabdinur, line 641:
```suggestion
          { attributes: { raised: false, handler: 'default', count: 1 }, timestamp: 17206369349 }
```

spec/datadog/opentelemetry_spec.rb
```diff
@@ -626,6 +626,41 @@
...
+        end
+
+        it 'adds one event to the span' do
+          expect(span.events.count).to eq(1)
+          expect(span.events[0].name).to eq('Exception was raised!')
+          expect(span.events[0].time_unix_nano).to eq(17206369349000000000)
+          expect(span.events[0].attributes).to eq({ 'raised' => 'false', 'handler' => 'default', 'count' => '1' })
```
mabdinur, line 648:
```suggestion
          expect(span.events[0].attributes).to eq({ raised: false, handler: 'default', count: 1 })
```

spec/datadog/tracing/transport/serializable_trace_spec.rb
```diff
@@ -128,6 +129,44 @@
...
+            'dummy',
+            events: [
+              Datadog::Tracing::SpanEvent.new(
+                'First Event',
+                time_unix_nano: 1_000_000_000
+              ),
+              Datadog::Tracing::SpanEvent.new(
+                "Another Event #{i}!",
+                time_unix_nano: 2_000_000_000,
+                attributes: { 'id' => i, 'required' => i == 1 },
```
mabdinur, line 148:
```suggestion
                attributes: { id: i, required: => i == 1 },
```

spec/datadog/opentelemetry_spec.rb
```diff
@@ -626,6 +671,41 @@
...
+        start_span
+        start_span.add_event('Exception was raised!', attributes: attributes, timestamp: timestamp)
+        start_span.finish
+      end
+
+      let(:start_span) { otel_tracer.start_span('start-span', **span_options) }
+      let(:active_span) { Datadog::Tracing.active_span }
+
+      context 'with name, attributes and timestamp' do
+        let(:attributes) { { 'raised' => false, 'handler' => 'default', 'count' => 1 } }
```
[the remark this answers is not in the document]
mabdinur: Opentelemetry API does not support setting symbols as attributes. 

@TonyCTHsu How do I ignore this linting error?

lib/datadog/opentelemetry/sdk/span_processor.rb
```diff
@@ -31,6 +32,15 @@ def on_start(span, parent_context)
         #
         # @param [Span] span the {Span} that just ended.
         def on_finish(span)
+          unless span.events.nil?
+            span.datadog_span.span_events = span.events.map do |event|
```
mabdinur, line 36: The Opentelemetry SDK does not return the Event added via `Span.add_event` method. Overriding Span.add_event and accessing the last event in the list can introduce race conditions: https://github.com/open-telemetry/opentelemetry-ruby/blob/406cabe54f8fb4937d0f6b73fe830a51c04fdb92/sdk/lib/opentelemetry/sdk/trace/span.rb#L173.

The simplest way to convert OpenTelemetry Span Events to Datadog Span Events is on span finish.

lib/datadog/opentelemetry/sdk/trace/span.rb
```diff
@@ -15,6 +15,20 @@ def set_attribute(key, value)
           res
         end
 
+        def record_exception(exception, attributes: nil)
+          res = super
+          if (span = datadog_span)
+            span.set_error(
```
mabdinur, line 21: According to the Datadog specification when an exception is recorded the error status, message, and stacktrace must be set on a span. This is to maintain compatibility with the error tracking product. If multiple exceptions are recorded on a span then the last exception wins.

lib/datadog/tracing/span_event.rb
```diff
@@ -0,0 +1,39 @@
...
+        time_unix_nano: nil
+      )
+        @name = name
+        @attributes = attributes || {}
+        @time_unix_nano = time_unix_nano || Core::Utils::Time.now.to_f * 1e9
+      end
+
+      def to_hash
+        h = { name: @name, time_unix_nano: @time_unix_nano }
+        h[:attributes] = attributes unless @attributes.empty?
```
mabdinur, line 34: Do not serialize the attributes field if it is empty

lib/datadog/tracing/transport/serializable_trace.rb
```diff
@@ -72,6 +72,9 @@ def to_msgpack(packer = nil)
             packer.write_map_header(number_of_elements_to_write) # Set header with how many elements in the map
           end
 
+          # serialize span events as meta tags
+          span.set_tag('events', span.events.map(&:to_hash).to_json) if span.events.count > 0
```
mabdinur, line 76: Do not serialize the events tag if it is empty

marcotc (approved):

lib/datadog/tracing/span_operation.rb
```diff
@@ -35,7 +35,7 @@ class SpanOperation
         :start_time,
         :trace_id,
         :type
-      attr_accessor :links, :status
+      attr_accessor :links, :status, :span_events
```
TonyCTHsu, line 38: It is a bit unfortunate that we have both `events` and `span_events` for this object.

lib/datadog/tracing/span_event.rb
```diff
@@ -0,0 +1,39 @@
...
+      attr_reader :attributes
+
+      # @!attribute [r] time_unix_nano
+      #   @return [Integer]
+      attr_reader :time_unix_nano
+
+      def initialize(
+        name,
+        attributes: nil,
+        time_unix_nano: nil
```
TonyCTHsu, line 25: Is there any usage, that caller does not provide a timestamp?

lib/datadog/opentelemetry/sdk/trace/span.rb
```diff
@@ -15,6 +15,20 @@ def set_attribute(key, value)
...
+              [
+                attributes&.fetch('exception.type', nil) || exception.class.to_s,
```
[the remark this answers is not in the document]
mabdinur: I added test coverage here: https://github.com/DataDog/dd-trace-rb/blob/e46997b66e56b641d207123a3fc2bbe830bdb94a/spec/datadog/opentelemetry_spec.rb#L607.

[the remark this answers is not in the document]
mabdinur: > When does attributes is nil? Can the logic simply become

This method overrides the implementation in the otel-sdk ([Span.record_exception](https://github.com/open-telemetry/opentelemetry-ruby/blob/35abeac8653359aebd0231c94a23587ad4ee372e/sdk/lib/opentelemetry/sdk/trace/span.rb#L164)). In otel the default value of attributes is nil so I kept that convention here.

The logic we need is:
 - If attributes is a hash AND 'exception.type' in attributes
     -  use attributes['exception.type']
 - else
     -  use  exception.class.to_s

I can modify this code to use an `if` statement. That should be easier to read.

lib/datadog/tracing/span_operation.rb
mabdinur: Yup 😞

lib/datadog/tracing/span_event.rb
mabdinur: In the OpenTelemetry API [Span.record_exception](https://github.com/open-telemetry/opentelemetry-ruby/blob/35abeac8653359aebd0231c94a23587ad4ee372e/sdk/lib/opentelemetry/sdk/trace/span.rb#L139) can be called without a timestamp. We can force the caller to always pass a timestamp and convert the time from seconds to nanoseconds but I think it's better to centralize that logic in `SpanEvent.new(...)`

lib/datadog/opentelemetry/sdk/trace/span.rb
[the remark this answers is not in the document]
TonyCTHsu: I try to refactor to make to logic more streamlined.
```ruby
        def record_exception(exception, attributes: nil)
          res = super
          if (span = datadog_span)
            if attributes
              span.set_error([
                attributes['exception.type'],
                attributes['exception.message'],
                attributes['exception.stacktrace']
              ])
            else
              span.set_error([
                exception.class.to_s,
                exception.message,
                exception.backtrace
              ])
            end
          end
          res
        end
```

>I added test coverage here:

Please add a test case when the keys are absent from hash.

p-datadog, line 23: If you mean to get the class name, `exception.class.name` ?

TonyCTHsu (approved):

lib/datadog/opentelemetry/sdk/trace/span.rb
```diff
@@ -15,6 +15,20 @@ def set_attribute(key, value)
           res
         end
 
+        def record_exception(exception, attributes: nil)
```
p-datadog, line 18: Can the arguments of this method be documented? It is unclear why data in `attributes` is overriding the attributes of `exception`, i.e. what the relationship should be between these arguments.

lib/datadog/tracing/span.rb
```diff
@@ -60,7 +61,8 @@ def initialize(
         type: nil,
         trace_id: nil,
         service_entry: nil,
-        links: nil
+        links: nil,
+        events: nil
```
p-datadog, line 65: Can the added parameters be documented in the docstring, specifically their expected types?

lib/datadog/tracing/span_event.rb
```diff
@@ -0,0 +1,39 @@
...
+require 'time'
+
+module Datadog
+  module Tracing
+    # SpanEvent represents an annotation on a span.
+    # @public_api
+    class SpanEvent
+      # @!attribute [r] name
+      #   @return [Integer]
+      attr_reader :name
```
p-datadog, line 12: It is uncommon for `name` to be an integer, can the docstring be expanded to provide more information on this? Is the integer coming from some other system and therefore is expected to mean something specific?

lib/datadog/tracing/span_event.rb
p-datadog, line 25: I see that the time is eventually exported in nanoseconds, but in Ruby most of the times are kept in (floating-point) seconds. If it is desired to keep the time in nanoseconds it would be helpful to have a note explaining the reasoning for this decision.

lib/datadog/tracing/span_event.rb
```diff
@@ -0,0 +1,39 @@
...
+      attr_reader :time_unix_nano
+
+      def initialize(
+        name,
+        attributes: nil,
+        time_unix_nano: nil
+      )
+        @name = name
+        @attributes = attributes || {}
+        @time_unix_nano = time_unix_nano || Core::Utils::Time.now.to_f * 1e9
```
p-datadog, line 29: This writes a floating-point time into `@time_unix_nano`; I assume the time received from otel will be an integer. The expected type for the field should be documented somewhere, and given that Ruby's times do not exceed nanosecond precision, the current time in nanoseconds can be exactly obtained by
```
(Time.now.to_r * 1_000_000_000).to_i
```

lib/datadog/tracing/span_operation.rb
p-datadog: Should some documentation be added explaining what `events` and `span_events` are to prevent confusion?

p-datadog: Should some documentation be added explaining what `events` and `span_events` are to prevent confusion?

lib/datadog/tracing/transport/serializable_trace.rb
p-datadog, line 76: Suggest `span.events.any?` or `!span.events.empty?` instead of `count > 0`.

lib/datadog/opentelemetry/sdk/trace/span.rb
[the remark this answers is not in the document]
mabdinur: attributes can contain a subset of keys (so `exception.type` but not `exception.message`). We need to check each the existence of each key separately. We will need something like this:

```
if (span = datadog_span)
  if attributes&.key?('exception.type')
    exception_type = attributes['exception.type']
  else
    exception_type = exception.class.to_s
  end

  if attributes&.key?('exception.message')
    exception_message = attributes['exception.message']
  else
    exception_message = exception.message
  end

  if attributes&.key?('exception.stacktrace')
    exception_stacktrace = attributes['exception.stacktrace']
  else
    exception_stacktrace = exception.backtrace
  end

  span.set_error_tags([exception_type, exception_message, exception_stacktrace])
end
```

Something like this is a bit more compact and easier to read (I could also use a ternary operators instead):

```
if (span = datadog_span)
  span.set_error_tags([
    if attributes&.key?('exception.type') then attributes['exception.type'] else exception.class.to_s end,
    if attributes&.key?('exception.message') then attributes['exception.message'] else exception.message end,
    if attributes&.key?('exception.stacktrace') then attributes['exception.stacktrace'] else exception.backtrace end,
  ])
end
```

mabdinur: I am using the same convention as the otel sdk: https://github.com/open-telemetry/opentelemetry-ruby/blob/35abeac8653359aebd0231c94a23587ad4ee372e/sdk/lib/opentelemetry/sdk/trace/span.rb#L166. I am not sure why they used ` exception.class.to_s` but I think it's best to keep things consistent.

lib/datadog/opentelemetry/sdk/trace/span.rb
mabdinur: This convention comes from how opentelemetry manages Exeception Events. The attribute map can be used to override the error message, traceback, and error type associated with an exception object: https://github.com/open-telemetry/opentelemetry-ruby/blob/35abeac8653359aebd0231c94a23587ad4ee372e/sdk/lib/opentelemetry/sdk/trace/span.rb#L170. I just copied this logic to map to the Span's error fields.

lib/datadog/tracing/span_event.rb
mabdinur: Oh that's a miss on my part. It should be a string

lib/datadog/tracing/span_event.rb
mabdinur: Nice catch. The current approach has some percission errors. This should improve it

lib/datadog/tracing/span_event.rb
mabdinur: Initially this was done for simplicy. Timestamps for span events are serialized in nanoseconds. There's no need to store them in seconds. 

I believe the start time and duration of a span is stored in seconds. These timestamps are converted to nanoseconds when a span is serialized and sent to the agent. I can use the same approach here.

lib/datadog/opentelemetry/sdk/trace/span.rb
marcotc: I think it's worth adding to do documentation of this method that it has the same interface as:
https://github.com/open-telemetry/opentelemetry-ruby/blob/406cabe54f8fb4937d0f6b73fe830a51c04fdb92/api/lib/opentelemetry/trace/span.rb#L126-L136

spec/datadog/opentelemetry_spec.rb
```diff
@@ -604,6 +604,51 @@
...
+
+          expect(span).to have_error
+        end
+      end
+
+      context 'with attributes' do
+        let(:attributes) do
+          { 'exception.stacktrace' => 'funny_stack', 'exception.type' => 'CustomError', 'exception.message' => 'NewError',
+            'candy' => true }
+        end
```
marcotc, line 639: I suggest adding test where `attributes` is an empty Hash, so we can see it falling back to the defaults.

PR #3779, opened 2024-07-11
[SDTEST-184] Agentless telemetry

anmarchenko:
**What does this PR do?**
Adds configurable agentless mode and configurable shutdown timeout to telemetry

**Motivation:**
Test visibility library runs in agentless mode more often than with Datadog agent, so we need agentless mode for telemetry to submit metrics.

Configurable shutdown timeout will help us to make sure that all metrics will be flushed when test suite finishes.

**How to test the change?**
Unit tests are provided. In parallel to review process I am going to work on POC for sending telemetry from test visibility instrumentation to test this end2end.


lib/datadog/core/configuration/settings.rb
```diff
@@ -663,6 +663,14 @@ def initialize(*_)
             o.type :bool
           end
 
+          # Enable agentless mode for telemetry: submit telemetry events directly to the intake without Datadog Agent.
+          #
+          # @return [Boolean]
+          option :agentless_enabled do |o|
```
ivoanjo, line 669: This description is very internal-oriented, but this is a user-visible setting. Do we think customers will be enabling this themselves?

If so, I would suggest tweaking the comment to better explain why/when you'd want this. If we don't think they'll use this directly, perhaps add a "# We do not recommend tweaking this value." or something to make it clear that this setting is for "Datadog stuff" and usually you don't need to care about it.

lib/datadog/core/telemetry/http/transport.rb
```diff
@@ -13,18 +14,47 @@ module Http
...
+              path: Http::Ext::AGENT_ENDPOINT
+            )
+          end
+
+          def self.build_agentless_transport(api_key:, dd_site:)
+            host =
+              if dd_site == Environment::Ext::DD_SITE_STAGING
+                # special case for staging - the url is constructed differently
+                'all-http-intake.logs.datad0g.com'
+              else
```
ivoanjo, line 32: Hmm... I'm not the hugest fan of having internal-only-exceptions in the production code for the library. Since this is only for testing, may I suggest adding a kind of an override somewhere where you can specify the full host, so we can use it here?

That way this code wouldn't need to know about staging stuff, just that optionally you can override the entire url. (I think we could even add such an override directly to the `AgentSettingsResolver` even? -- add arguments for agentless telemetry config)

lib/datadog/core/environment/ext.rb
```diff
@@ -40,6 +40,8 @@ module Ext
         TAG_SERVICE = 'service'
         TAG_VERSION = 'version'
 
+        DD_SITE_STAGING = 'datad0g.com'
+
```
ivoanjo, line 44: I left a suggestion below to avoid this (and my reasoning). Even if we do keep the code as-is, I don't think we should have this constant here -- we should embed it in the code that uses it, as only that bit of code should ever know about this info.

lib/datadog/core/telemetry/http/transport.rb
```diff
@@ -13,18 +14,47 @@ module Http
...
         class Transport
+          def self.build_agent_transport
+            agent_settings = Configuration::AgentSettingsResolver.call(Datadog.configuration)
+
+            Transport.new(
+              host: agent_settings.hostname,
+              port: agent_settings.port,
+              path: Http::Ext::AGENT_ENDPOINT
+            )
+          end
```
ivoanjo, line 25: Minor: This is one of those "I know this was already like this but...".

So we really missed doing this at the time, but after a bunch of refactoring, we basically have the `build_telemetry` method in components receiving an `agent_settings`, and then not making it available to `Transport`, which ends up needing to recompute stuff _again_. (This will even cause some warnings to be printed more than once).

Aka -- can I convince you to pass in the `agent_settings` from the component here, and use it directly? ;)

lib/datadog/core/configuration/components.rb
```diff
@@ -57,17 +57,35 @@ def build_runtime_metrics_worker(settings)
...
+            end
+
+            transport = if agentless_enabled
+                          Datadog::Core::Telemetry::Http::Transport.build_agentless_transport(
+                            api_key: settings.api_key,
+                            dd_site: settings.site
+                          )
+                        else
+                          Datadog::Core::Telemetry::Http::Transport.build_agent_transport
+                        end
```
ivoanjo, line 79: Minor: There's a lot of telemetry-specific logic in here. I wonder if it would make sense to embed it directly on the telemetry transport or components. E.g. It's a bit "surprising" how much the core configuration components needs to know about these things right now.

(But I do consider it minor, if you don't really want to touch it :D )

sig/datadog/core/environment/ext.rbs
```diff
@@ -2,21 +2,23 @@ module Datadog
   module Core
```
ivoanjo, line 2: Thanks for updating the signatures! But do remember to drop the `ignore` from the `Steepfile` otherwise this isn't really getting checked :)

(Same for other files that you touched :P )

sig/datadog/core/telemetry/component.rbs
```diff
@@ -11,7 +11,7 @@ module Datadog
 
         include Core::Utils::Forking
 
-        def initialize: (heartbeat_interval_seconds: Float, metrics_aggregation_interval_seconds: Float, dependency_collection: bool, ?enabled: bool, ?metrics_enabled: bool) -> void
+        def initialize: (http_transport: Datadog::Core::Telemetry::Http::Transport, heartbeat_interval_seconds: Float, metrics_aggregation_interval_seconds: Float, dependency_collection: bool, ?enabled: bool, ?metrics_enabled: bool, shutdown_timeout_seconds: Float) -> void
```
ivoanjo, line 14: Minor:

```suggestion
        def initialize: (http_transport: Datadog::Core::Telemetry::Http::Transport, heartbeat_interval_seconds: Float, metrics_aggregation_interval_seconds: Float, dependency_collection: bool, ?enabled: bool, ?metrics_enabled: bool, shutdown_timeout_seconds: Numeric) -> void
```

(To allow integers as well?)

sig/datadog/core/telemetry/emitter.rbs
```diff
@@ -7,7 +7,7 @@ module Datadog
         attr_reader http_transport: untyped
 
         extend Core::Utils::Forking
-        def initialize: (?http_transport: untyped) -> void
+        def initialize: (http_transport: untyped) -> void
```
ivoanjo, line 10: Minor: I think this could be pointed at `Datadog::Core::Telemetry::Http::Transport`?

sig/datadog/core/telemetry/worker.rbs
```diff
@@ -21,7 +21,7 @@ module Datadog
         @ticks_per_heartbeat: Integer
         @current_ticks: Integer
 
-        def initialize: (?enabled: bool, heartbeat_interval_seconds: Float, metrics_aggregation_interval_seconds: Float, emitter: Emitter, metrics_manager: MetricsManager, ?shutdown_timeout: Integer, ?buffer_size: Integer, dependency_collection: bool) -> void
+        def initialize: (?enabled: bool, heartbeat_interval_seconds: Float, metrics_aggregation_interval_seconds: Float, emitter: Emitter, metrics_manager: MetricsManager, ?shutdown_timeout: Float, ?buffer_size: Integer, dependency_collection: bool) -> void
```
ivoanjo, line 24: Minor:
```suggestion
        def initialize: (?enabled: bool, heartbeat_interval_seconds: Float, metrics_aggregation_interval_seconds: Float, emitter: Emitter, metrics_manager: MetricsManager, ?shutdown_timeout: Numeric, ?buffer_size: Integer, dependency_collection: bool) -> void
```

spec/datadog/core/configuration/components_spec.rb
```diff
@@ -263,6 +270,41 @@
...
+              metrics_enabled: false, heartbeat_interval_seconds: heartbeat_interval_seconds,
+              metrics_aggregation_interval_seconds: metrics_aggregation_interval_seconds,
+              dependency_collection: dependency_collection, shutdown_timeout_seconds: shutdown_timeout_seconds }
+          end
+
+          it 'does not enable telemetry when agentless mode requested but api key is not present' do
+            expect(logger).to receive(:debug)
+            is_expected.to be(telemetry)
+          end
+        end
```
ivoanjo, line 306: Minor: So this made me think the current logic is kinda confusing. If telemetry gets disabled, why are we still creating transports and the telemetry component at all? :eyes:

ivoanjo (approved):
Left a bunch of notes and suggestions, but overall it LGTM :+1:

lib/datadog/core/configuration/settings.rb
anmarchenko: Oops, I forgot to mark it internal only. Customers are not supposed to do it by themselves.

lib/datadog/core/telemetry/http/transport.rb
anmarchenko: Yes, sure!

sig/datadog/core/environment/ext.rbs
anmarchenko: I wasn't aware that this is ignored... I dropped all ignores from Telemetry signatures a while ago, will do it for this file too

sig/datadog/core/telemetry/component.rbs
anmarchenko: I need to see if this works: in Ruby Numeric does not have any arithmetic operators, so if any are used down the line, numeric won't work...

spec/datadog/core/configuration/components_spec.rb
anmarchenko: To avoid null checks all over the code I would say! we create transport so that we could tell typechecker that transport is always there

sig/datadog/core/telemetry/component.rbs
ivoanjo: Interesting... I guess `Integer | Float` maybe? Although I do admit is a bit of a pain to use.

spec/datadog/core/configuration/components_spec.rb
ivoanjo: I see. I guess I'm not a big fan of the way we've implemented the pattern (and we do this in a few other places) -- I think if we need a "null object", it's probably better to introduce an actual empty implementation, rather than building the whole thing, and then having `enabled` just make it not "go" when it gets called.

(Out of curiosity, the profiler component builder does return `nil` when profiling is disabled)

ivoanjo: Fair enough :sweat_smile: xD

marcotc (approved):
Looks great. I agree with @ivoanjo's points, as well, but otherwise 🚢

lib/datadog/core/configuration/components.rb
anmarchenko: moved to a factory method in Telemetry::Component

PR #3786, opened 2024-07-16
[NO-TICKET] Fixes for rare issues in the profiler + small optimization

ivoanjo:
**What does this PR do?**

This PR collects a few fixes/optimizations for rare issues in the profiler that I collected this week. I recommend reviewing commit-by-commit.

**Motivation:**

Fix these potential issues.

**Additional Notes:**

I've never seen any of the fixed issues happen in a production release of our library; I don't believe our customers were ever affected.

**How to test the change?**

It's really hard to add test coverage for these, as they need a really specific timing + set of  conditions to trigger. Such testing would be very complex, so for now we rely on our other integration tests/dogfooding to catch these things.


anmarchenko (approved):
This is the first time when I understand your PR well enough to approve it!

Issue #3782, opened 2024-07-15
AppSec crashes when parsing integer HTTP headers (undefined method `tr' for nil:NilClass (NoMethodError))


TomNaessens:
Hi all, 👋 

We often have bots hit our systems with malformed HTTP headers. One of the malformed HTTP headers they apparently try (a lot), is a plain number (e.g. `HTTP_2`). This causes the following line to throw an "undefined method `tr' for nil:NilClass (NoMethodError)": https://github.com/DataDog/dd-trace-rb/blob/f867bd9f13d26a6662a3d409510c5c55bf5cd5f5/lib/datadog/appsec/contrib/rack/gateway/request.rb#L44

`k =~ /^HTTP_/` is matched, `k.gsub(/^HTTP_/, '')` results into `"2"`, which becomes `nil` through `downcase!`:

```ruby
[527]dev(main):0> "2".downcase!
=> nil
```

I'm not exactly sure what the right solution is, given that we're already working with a new object after the `gsub`, a regular `downcase` (without `!`) might work (although that will create another object):

```ruby
[526]dev(main):0> "2".downcase
=> "2"
```

To reproduce, spin up a Rails app with dd trace, enable appsec, instrument Rack, and send a request with an integer as HTTP header:

```
 ~  curl http://localhost:3001 -H "2: 3"
Puma caught this error: undefined method `tr' for nil:NilClass (NoMethodError)
```

lloeki:
Good catch! Thanks for the report.

> I'm not exactly sure what the right solution is

Probably something like:

```
h[k.gsub(/^HTTP_/, '').tap(&:downcase!).tr('_', '-')] = v if k =~ /^HTTP_/ 
```

And to save another allocation:

```
h[k.gsub(/^HTTP_/, '').tap(&:downcase!).tap { |s| s.tr('_', '-') }] = v if k =~ /^HTTP_/ 
```

TomNaessens:
I'm wanting to open a PR and write some tests for this, one last question: what's the expected behaviour when such a header is sent? With the `.tap`s, "malformed" headers not containing text will end up in the hash as `nil => value`. Is that intended, or should they just not end up in the HeaderCollection at all?

TonyCTHsu:
👋 @TomNaessens We've released a `2.3.0`, give it a try!

TomNaessens:
Awesome, thank you!

PR #3790, opened 2024-07-17
Fix AppSec crash when parsing integer http headers

vpellan:
**What does this PR do?**

This fixes a crash when malformed headers are sent (e.g. headers with numbers)

**Motivation:**

This fixes issue #3782 

**Additional Notes:**

**How to test the change?**

`bundle exec appraisal ruby-x.x-rack-x rake spec:appsec:rack`

### DIFF
--- a/lib/datadog/appsec/contrib/rack/gateway/request.rb
+++ b/lib/datadog/appsec/contrib/rack/gateway/request.rb
@@ -44,10 +44,11 @@ module Datadog
                 # When multiple headers with the same name are present, they are concatenated with a comma
                 # https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2
                 # Because headers are case insensitive, HTTP_FOO and HTTP_Foo is the same, and should be merged
-                if k =~ /^HTTP_/
-                  key = k.gsub(/^HTTP_/, '').tap(&:downcase!).tap { |s| s.tr!('_', '-') }
-                  h[key] = h[key].nil? ? v : "#{h[key]}, #{v}"
-                end
+                next unless k.start_with?('HTTP_')
+
+                key = k.delete_prefix('HTTP_').tap(&:downcase!).tap { |s| s.tr!('_', '-') }
+                current_val = h[key]
+                h[key] = current_val.nil? ? v : "#{current_val}, #{v}"
               end
 
               result['content-type'] = request.content_type if request.content_type
--- a/lib/datadog/appsec/contrib/rack/gateway/request.rb
+++ b/lib/datadog/appsec/contrib/rack/gateway/request.rb
@@ -41,14 +41,7 @@ module Datadog
 
             def headers
               result = request.env.each_with_object({}) do |(k, v), h|
-                # When multiple headers with the same name are present, they are concatenated with a comma
-                # https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2
-                # Because headers are case insensitive, HTTP_FOO and HTTP_Foo is the same, and should be merged
-                next unless k.start_with?('HTTP_')
-
-                key = k.delete_prefix('HTTP_').tap(&:downcase!).tap { |s| s.tr!('_', '-') }
-                current_val = h[key]
-                h[key] = current_val.nil? ? v : "#{current_val}, #{v}"
+                h[k.delete_prefix('HTTP_').tap(&:downcase!).tap { |s| s.tr!('_', '-') }] = v if k.start_with?('HTTP_')
               end
 
               result['content-type'] = request.content_type if request.content_type
--- a/lib/datadog/appsec/contrib/rack/gateway/request.rb
+++ b/lib/datadog/appsec/contrib/rack/gateway/request.rb
@@ -41,7 +41,7 @@ module Datadog
 
             def headers
               result = request.env.each_with_object({}) do |(k, v), h|
-                h[k.gsub(/^HTTP_/, '').downcase!.tr('_', '-')] = v if k =~ /^HTTP_/
+                h[k.gsub(/^HTTP_/, '').tap(&:downcase!).tap { |s| s.tr!('_', '-') }] = v if k =~ /^HTTP_/
               end
 
               result['content-type'] = request.content_type if request.content_type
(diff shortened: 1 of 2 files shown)


lib/datadog/appsec/contrib/rack/gateway/request.rb
```diff
@@ -41,7 +41,7 @@ def method
 
             def headers
               result = request.env.each_with_object({}) do |(k, v), h|
-                h[k.gsub(/^HTTP_/, '').downcase!.tr('_', '-')] = v if k =~ /^HTTP_/
+                h[k.gsub(/^HTTP_/, '').tap(&:downcase!).tap { |s| s.tr!('_', '-') }] = v if k =~ /^HTTP_/
```
p-datadog, line 44: The test indicates that you are keeping all of the values, therefore in my opinion:

1. There should also be a test for mixed-case string input, e.g. HTTP_Foo
2. I think the logic as currently proposed will result in the mixed case being preserved which I think is not the right behavior

p-datadog:
Given the original report was in reference to invalid headers being sent to the application by external clients, would it also make sense to check for *duplicate* headers being sent? Including in varying case? E.g. HTTP_FOO and HTTP_Foo in the same request.

lib/datadog/appsec/contrib/rack/gateway/request.rb
vpellan: As you pointed in your second comment, HTTP header names are uppercased by Rack (as HTTP header names are case-insensitive according to RFC 2616) and Rack will also drop empty headers so 'HTTP_' should not be possible either. But there's the case of someone adding a middleware to Rack that changes the headers for exemple. I believe that in these case, we still want to send everything to the WAF, as an attack could still be present in the value of these headers.

vpellan:
> Given the original report was in reference to invalid headers being sent to the application by external clients, would it also make sense to check for _duplicate_ headers being sent? Including in varying case? E.g. HTTP_FOO and HTTP_Foo in the same request.

I added support for duplicate headers, including in varying case, as stated by [RFC 2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2), header keys are case insensitive, so HTTP_FOO and HTTP_Foo are the same header key, and thus it is semantically equal to HTTP_FOO: "value1, value2" (for HTTP_FOO: value1 and HTTP_Foo: value2). It is also possible to send the values as an array to the WAF but that would mean replacing all the strings by arrays in tests.

> And I'm thinking HTTP_FOO_2 and HTTP_2_FOO are also interesting cases to check/add to the test suite, and these should be actually obtainable in real world.

Thank you for that suggestion ! I added these tests too

lib/datadog/appsec/contrib/rack/gateway/request.rb
```diff
@@ -41,7 +41,13 @@ def method
 
             def headers
               result = request.env.each_with_object({}) do |(k, v), h|
-                h[k.gsub(/^HTTP_/, '').tap(&:downcase!).tap { |s| s.tr!('_', '-') }] = v if k =~ /^HTTP_/
+                # When multiple headers with the same name are present, they are concatenated with a comma
+                # https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2
+                # Because headers are case insensitive, HTTP_FOO and HTTP_Foo is the same, and should be merged
+                if k =~ /^HTTP_/
```
p-datadog, line 47: Can use `start_with?`

lib/datadog/appsec/contrib/rack/gateway/request.rb
```diff
@@ -41,7 +41,13 @@ def method
...
+                  key = k.gsub(/^HTTP_/, '').tap(&:downcase!).tap { |s| s.tr!('_', '-') }
+                  h[key] = h[key].nil? ? v : "#{h[key]}, #{v}"
```
p-datadog, line 49: Assigning `h[key]` to a variable will save one hash access

p-datadog (approved):
I see several optimization opportunities, don't know how performance-critical you think this code is. Functionally I think it's good to go.

lib/datadog/appsec/contrib/rack/gateway/request.rb
```diff
@@ -41,7 +41,14 @@ def method
 
             def headers
               result = request.env.each_with_object({}) do |(k, v), h|
-                h[k.gsub(/^HTTP_/, '').downcase!.tr('_', '-')] = v if k =~ /^HTTP_/
+                # When multiple headers with the same name are present, they are concatenated with a comma
+                # https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2
+                # Because headers are case insensitive, HTTP_FOO and HTTP_Foo is the same, and should be merged
```
lloeki, line 46: When could this happen?

- The Rack means upcasing HTTP headers into the `HTTP_UPCASED_HEADER` variant
- Given that `request.env.each_with_object` iterates on the rack env (which is a hash) there can be no duplicates except for mixed case, which cannot happen when Rack compliant due to the above.
- The Rack env does not concatenate multiple identical headers into a single hash key, instead making use of a single one (can't recall if first or last).

It follows that the only thing that could fill in some `HTTP_MiXeD_CaSe` is some non-Rack compliant middleware or monkeypatch that  inject pseudo headers that actually don't exist on the HTTP request AND would execute before Datadog's middlewares, which are designed to be run first at the top of the Rack stack.

Note: According to the HTTP spec concatenation MAY happen but MUST only happen for headers that are lists. As is this code does so for ALL headers so it's very much incorrect, and possibly prone to injections or opportunities for a WAF bypass (e.g `Authorization` headers being concatenated in a way that make it valid for libddwaf rules but would in isolation - as seen by the application - be dangerous)

vpellan: I agree that this is an extremely rare edge case, that would be introduced by some sort of weird middleware, and I believe that we should not be responsible for a client modifying the core behaviour of Rack, especially when it is only hypothetical. And a quick benchmark shows that this is 50x slower than your first suggestion with a bit of modification (replacing gsub by delete_prefix and regex in condition by start_with), which I believe is more important for our clients.

lloeki (approved):

PR #3792, opened 2024-07-18
[PROF-10123] Add placeholder for skipped samples in allocation profiler, second try

ivoanjo:
**What does this PR do?**

This PR fixes a corner case issue with the allocation profiler: if too many objects get allocated, we "clamp" the weight of every individual sample (to avoid introducing bias) to a maximum value.

But we previously did not in any way represent the clamping on the profile itself, which meant that the total number of allocations reported by the profiler would be missing any object counts that were clamped.

In practice, we don't expect this to happen very often, as we have a high clamping value `MAX_ALLOC_WEIGHT = 65535` (it's in `collectors_cpu_and_wall_time_worker.c`).

This PR then introduces an extra placeholder sample, when needed, to represent allocations that went over the clamped value.

Thus the total amount of objects allocated reported by the profiler now becomes accurate, even in the presence of clamping.

**This is the second try at this feature. The first attempt was #3760 which due to a bug we ended up reverting in #3775**. The revert was because of a bug in the interaction between allocation profiling and heap profiling. 5c7bd5f22df26aeb6edf3c0dcf34a52075cfd8ad is the fix for the issue that led to the revert, and that commit's description includes an explanation of why the first PR was incorrect.

**Motivation:**

We're considering lowering the maximum clamping value, which would mean it would be hit more often, and thus it's important to fix this issue before we do that.

**Additional Notes:**

N/A

**How to test the change?**

I've added test coverage for this change. I also used the following snippet to quickly experiment while I was developing this PR:

```ruby
allocator = Thread.new { loop { BasicObject.new } }
sleep 10
allocator.kill; allocator.join
puts "Allocated #{GC.stat(:total_allocated_objects)} objects total"
```

Here's how it looks (it's the teal frame on the left):
> image

You can observe that in this testcase with a massive amount of allocations we would be missing around ~10% of allocation counts, which we now track correctly.

### DIFF
--- a/ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -1144,9 +1144,15 @@ static VALUE rescued_sample_allocation(VALUE tracepoint_data) {
     discrete_dynamic_sampler_events_since_last_sample(&state->allocation_sampler) :
     // if we aren't, then we're sampling every event
     1;
-  // TODO: Signal in the profile that clamping happened?
+
+  // To control bias from sampling, we clamp the maximum weight attributed to a single allocation sample. This avoids
+  // assigning a very large number to a sample, if for instance the dynamic sampling mechanism chose a really big interval.
   unsigned int weight = allocations_since_last_sample > MAX_ALLOC_WEIGHT ? MAX_ALLOC_WEIGHT : (unsigned int) allocations_since_last_sample;
   thread_context_collector_sample_allocation(state->thread_context_collector_instance, weight, new_object);
+  // ...but we still represent the skipped samples in the profile, thus the data will account for all allocations.
+  if (weight < allocations_since_last_sample) {
+    thread_context_collector_sample_skipped_allocation_samples(state->thread_context_collector_instance, allocations_since_last_sample - weight);
+  }
 
   // Return a dummy VALUE because we're called from rb_rescue2 which requires it
   return Qnil;
--- a/ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -316,6 +316,8 @@ static const rb_data_type_t cpu_and_wall_time_worker_typed_data = {
 };
 
 static VALUE _native_new(VALUE klass) {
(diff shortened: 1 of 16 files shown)


marcotc:
> We're considering lowering the maximum clamping value, which would mean it would be hit more often, and thus it's important to fix this issue before we do that.

Can you explain why this is beneficial?

anmarchenko (approved):

ivoanjo:
> > We're considering lowering the maximum clamping value, which would mean it would be hit more often, and thus it's important to fix this issue before we do that.
>
> Can you explain why this is beneficial?

1. The placeholder

The placeholder is beneficial because prior to this PR we didn't "flag" in any way in the profile the missing data.

Thus, if an app allocated 1M objects and the profiler chose to sample less often than every `MAX_ALLOC_WEIGHT` objects, then prior to this PR you'd see a lower number than 1M in the profiler/profiler metrics.

For instance, taking my image above as an example

> image

...this example allocated ~96M objects, but the profiler would not report the ~9M that were skipped, so you'd only see 86M objects reported.

2. Lowering the `MAX_ALLOC_WEIGHT` value

This is now done in https://github.com/DataDog/dd-trace-rb/pull/3793 .

Reusing some of the explanation I also used in slack:

The base idea with our current sampling approach is that we sample periodically and assign the spent resource during that period to the stack we chose to sample. E.g.:

* If we sample every 10ms of time, we attribute 10ms to a given stack (for wall-time profiling)
* If we sample every 10 objects, we attribute 10 objects (for allocation profiling)

But allocation profiling poses an issue (more than cpu, for instance): if an application is creating objects like crazy, we may not want to sample as often, to avoid imposing a lot of overhead on the application (every 10 allocations in millions of objects is a lot).

So the profiler adjusts this dynamically based on overhead and number of objects. But this poses an issue: what if the profiler decides to sample really rarely? E.g. let's say the profiler samples every 100k objects. Attributing 100k "objects" to a single sample would be quite extreme.

Thus we introduced the MAX_ALLOC_WEIGHT that says "you can't attribute more than this value to a single allocation sample". Lowering `MAX_ALLOC_WEIGHT` reduces the potential for bias as it lowers the maximum that any single unlucky sample can be "blamed" for -- even if the profiler chooses to sample every 100k or more objects anyway.

ivoanjo:
Thanks for the reviews!

@marcotc hopefully it's OK with you, I'm going to go ahead and merge this one, please do comment further with any feedback, as I really value it and I'm happy to open PRs to address it :)

PR #3801, opened 2024-07-22
[NO-TICKET] RFC: Adopt standardrb for Profiler

ivoanjo:
**What does this PR do?**

This PR replaces the current rubocop style configuration with the [standardrb](https://github.com/standardrb/standard) linter **for the profiler only**.

This PR has several objectives:

1. Experiment with a path for incremental migration of the gem to standardrb
2. Validate co-existence with Rubocop during the migration
3. Serve as an example for future migrations

And in particular it shows:

* How to tweak Rubocop to ignore the files and folders to be managed by standardrb
* How to configure standardrb to operate only on a subset of files
* How to set up `.git-blame-ignore-revs` to improve blame-ability of updated files
* How to integrate standardrb on our usual CI setup

**Motivation:**

We've been adopting standardrb for our new codebases/gems, so we've recently discussed, and I had an action item to experiment with the profiler adopting standardrb.

**Additional Notes:**

You'll note that I've added a `.standard_todo.yml` file rather than fixing every file. 
Specifically, I've yet to fix the following cops:

  - Layout/SpaceInsideHashLiteralBraces
  - Style/QuotedSymbols
  - Style/RedundantBegin
  - Style/StringLiterals
  - Style/TrailingCommaInArguments

This is because fixing each of these cops will create a LOT of diff noise. Thus, I've explicitly not included them in this PR, and plan to fix them separately, in smaller PRs, so we can still sanely review the diff on this PR.

**How to test the change?**

No behavior changes are expected, so our test coverage should be enough.


Gemfile
```diff
@@ -82,6 +82,7 @@ group :check do
     gem 'rbs', '~> 3.2.0', require: false
     gem 'steep', '~> 1.6.0', require: false
   end
+  gem 'standard'
```
lloeki, line 85:
```
gem 'standard', require: false
```

(For some reason I can't make a suggestion).

ivoanjo: Thanks for catching it! Fixed in 252713bf1f7a21015c15db0ee37ba095a336697e

Rakefile
```diff
@@ -415,8 +415,16 @@ namespace :native_dev do
   CLEAN.concat(NATIVE_CLEAN)
 end
 
+task :standard_profiling do
```
lloeki, line 418: What about this?

```
namespace :standard do
  task :profiling do
  
task :standard => [:'standard:profiling']
```

ivoanjo: My thinking was that `standard` already includes their own rake tasks, so I wanted to avoid clashing with the built-in ones.

By using `standard_profiling`, I was hoping to make it clear that this is our custom config for profiling only, and the idea would be that we would remove it once we're happy with using standard everywhere and go back to the built-in standard rake tasks.

Gemfile
lloeki:
```
gem 'standard', require: false
```

Rakefile
lloeki: I ventured as much, but I don't find it very clear that it's our task. It'd be clearer if it were `namespace`d `:datadog`. Also having a `desc` can make that clear, plus make it discoverable through `rake -T` ;)

My rationale is that since the standard rake task is not there yet and it's due for removal once it is then it's totally OK.

lloeki: Also then you can have:

```
task default: ['rubocop', 'standard', 'typecheck', 'spec:main']
```

And when we pivot to the standard task it's a drop-in replacement.

Which makes me think that unless I missed something, as is profiling is not lint-checked in CI anymore? ([see here](https://github.com/DataDog/dd-trace-rb/blob/7de53929e489ff6025855eb263bf79d82be48f43/.circleci/config.yml#L161))

.standard_todo.yml
```diff
@@ -0,0 +1,169 @@
+# Auto generated files with errors to ignore.
+# Remove from this list as you refactor files.
+---
+ignore:
```
lloeki, line 4: What if we add blanket ignores for everything but profiling and use the `standard` rake task?

Thinking of it, it seems like it is the way to go as as is using standard on any other file (e.g my editor is configured to automatically run and fix using standard if it is available) will have standard lint it.

Rakefile
ivoanjo: Fixed in a3588e19905b3a7d272a2e0cc69997e0a55247ee and abc3ef4fa85a7e80f7538337800f79dcb0af40e7 .

I've moved standardrb to a separate namespace + added descriptions and also updated the CI configuration. The CI part in particular I was under the (mistaken) expectation that CI also used the default task, thanks for pointing out it doesn't.

.standard_todo.yml
ivoanjo: It took me some fighting with standardrb but yeah I like this idea, so I've redone the integration using this approach in 6849df52f863e3a41240503fbf9289cd378ebba5 .

lloeki (approved):

ext/datadog_profiling_native_extension/native_extension_helpers.rb
```diff
@@ -104,7 +104,7 @@ def self.libdatadog_folder_relative_to_ruby_extensions_folders(
         # the Gem.dir.
         expected_ruby_extensions_folders = [
           "#{gem_dir}/extensions/platform/extension_api_version/datadog_version/",
-          "#{gem_dir}/bundler/gems/extensions/platform/extension_api_version/datadog_version/",
+          "#{gem_dir}/bundler/gems/extensions/platform/extension_api_version/datadog_version/"
```
marcotc, line 107: I think all the cop changes looks super reasonable!
Except for removing the trailing comma from multi-line lists, like this one.

This guarantees that diffs for this list have a useless diff line on the trailing element, just adding a comma.

It's a small thing, but you always makes me slightly confused when I'm reviewing changes and I don't know if someone changed the last element of the list or not. Specially, for long lines, like `"#{gem_dir}/bundler/gems/extensions/platform/extension_api_version/datadog_version/"`, I have to inspect the whole string, instead of it simply never appearing in the diff, because in fact it did not change at all.

marcotc: > accept it as-is

I want to yield such power!

marcotc: I see this was already reported: https://github.com/standardrb/standard/issues/611

.standard.yml
```diff
@@ -0,0 +1,3 @@
+# For available configuration options, see:
+#   https://github.com/testdouble/standard
+ruby_version: 2.5
```
marcotc, line 3:
```suggestion
ruby_version: 2.5

ignore:
  - '**/*':
      - Style/TrailingCommaInArguments
      - Style/TrailingCommaInArrayLiteral
      - Style/TrailingCommaInHashLiteral
```
I think this is the closest we can get to it 😝

ivoanjo: Interesting... I'd be happy to go with this also, I do like leaving the trailing commas.

I guess let's see what the rest of the team thinks? I'm happy to apply the change, but want to see if we can get consensus :)

ext/datadog_profiling_native_extension/native_extension_helpers.rb
lloeki: Yeah the comma thing is the most annoying one ever (with the "double-quote strings everywhere" as a close second). I mean beyond diffs, in editors "duplicate line" is a very useful thing.

I really don't understand it, I recall their argument being mostly "well you shouldn't care because it's auto formatted on save".

A workaround is:

- disable the rule at the standard level (standard.yml)
- enforce the chosen rule at the rubocop level

ivoanjo: So it's clear there are ways of bending standardrb to our will on this one.

My question is: do we want to?

I'm leaning towards soft-no just because it goes ~slightly against the spirit of standardrb, but being a soft-no, I'm happy to go in the other direction.

p-datadog (approved):
It is my opinion that different code fragments call for different formatting in order to most clearly express the intent and minimize confusion and likelihood of errors. A competent programmer can intelligently choose formatting and style on a case-by-case basis to make their code clear to the reader. I find that at least in Ruby, any programmatically enforced code style produces inferior readability and maintainability compared to an intelligent programmer writing their code with care, because any chosen style is worse for some code fragments and the tool demands the selected formatting all the time including when it makes the code harder to read or harder to maintain.

Specifically for this PR, I am a strong believer in trailing commas to reduce diff noise and my biggest issue with the existing rubocop rule set is that it demands inversion of flow in conditionals by demanding the `action if condition` pattern for single-statement actions, which appears to be retained in this PR (I don't know if standard also demands this inversion of flow or permits remaining inverted conditions to remain, but would also permit normal flow for single-statement actions.)

I think standard is inferior to rubocop due to trailing comma deletion, and they are either tied for conditional flow inversion or possibly standard is better if it does not force flow inversion. Code as I type it does not satisfy any programmatically defined style, therefore whether rubocop or standard is used I would have them reformat everything.  As such, the two tools seem to be to be approximately equal. Therefore I do not object to replacing one with the other.

ext/datadog_profiling_native_extension/native_extension_helpers.rb
lloeki: There's value in not quibbling. See, this PR seems blocked because of this? ;)

We can revisit later if we find it very annoying.

lloeki:
Seems like we by and large don't like the comma thing, so I'm OK to globally disable that rule at the standard level.

>  find that at least in Ruby, any programmatically enforced code style produces inferior readability and maintainability compared to an intelligent programmer writing their code with care

@p-datadog I agree that intentional formatting helps conveying intent. In practice I've found only a few cases where standard made choices that annoyed me:

- high irritation: last-item comma, adding items by duping lines is annoying (although autocorrected on save) and diffs become an eyesore
- medium irritation: `foo = if / then / else / end` and `foo = case / when / else / end` alignment, which visually obscures the assignment
- low irritation: single versus double quotes for strings. I really like the single quote => no interpolation nor escape signalling without having to visually scan the string.

Otherwise the remainder is really helpful and not too intrusive, the set of enforced rules is much smaller than Rubocop defaults, and I've been able to format code and convey intent as usual, in ways that Rubocop rule defaults would prevent and be much more annoying.

lloeki:
BTW I chimed in: https://github.com/standardrb/standard/issues/611#issuecomment-2270941861

ivoanjo:
Thanks everyone! I've excluded the cops that check for trailing commas, so we can continue to use them, and I believe annoyed everyone about this topic so I think we're good to go!

I'll press the green button. As usual, this is still open for revisiting, and please do share any issues/concerns :)

PR #3809, opened 2024-07-25
DBM: Add option to append SQL comment propagation

### LINKED ISSUE #3601: Allow to append (instead of prepend) `Propagation::SqlComment` comment
Hello 👋

When database monitoring is enable, and propagation mode is set to `full` or `service`, `ddtrace` is prepending a comment to each SQL query.

https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/propagation/sql_comment.rb

While we understand the value of having them prefixed, this comes with the drawbacks of breaking query preview of other monitoring tool like [AWS Performance Insights](https://aws.amazon.com/rds/performance-insights/).

Would it be possible to expose a configuration in order to opt for suffixing instead of prefixing?
Note that we maintain a patch on our side to handle this, and Database Monitoring still behaves correctly.

marcotc:
Fixes #3601

Add an option (defaults to `false`) to append the DBM trace propagation comment instead, of prepending it.
This allow users using database monitor tools to see the SQL query itself first, instead of the Datadog trace propagation comment first.

**Motivation:**
To [correlate Database Monitoring and Traces](https://docs.datadoghq.com/database_monitoring/connect_dbm_and_apm/?tab=ruby), the tracer add a SQL comments to instrumented SQL queries.

Currently, the comments are prepended to the query, to prevent any possible truncation from removing the propagation data.
But this means that, in any external tools use to monitor the database, the queries will have the Datadog propagation comment first, and not the query itself. This makes inspecting the query quite inconvenient in many tools.

**How to test the change?**
There are tests covering all the changes.


cswatt (approved):

lib/datadog/tracing/contrib/mysql2/instrumentation.rb
```diff
@@ -57,22 +57,27 @@ def query(sql, options = {})
...
 
+            def inject_propagation(span, sql, trace_op)
+              propagation_mode = Contrib::Propagation::SqlComment::Mode.new(datadog_configuration[:comment_propagation])
+
+              Contrib::Propagation::SqlComment.annotate!(span, propagation_mode)
+              Contrib::Propagation::SqlComment.prepend_comment(
+                sql,
+                span,
+                trace_op,
+                propagation_mode,
```
TonyCTHsu, line 76: Instead of adding another parameter to `prepend_comment`, I suggest refactor and extend `SqlComment::Mode` by injecting the configuration.

```ruby
Contrib::Propagation::SqlComment::Mode.new(datadog_configuration)
```

marcotc: Done!

TonyCTHsu (approved):

docs/GettingStarted.md
```diff
@@ -1171,6 +1171,7 @@ client.query("SELECT * FROM users WHERE group='x'")
...
 | `comment_propagation` | `DD_DBM_PROPAGATION_MODE`      | `String` | SQL comment propagation mode for database monitoring. <br />(example: `disabled` \| `service`\| `full`). <br /><br />**Important**: _Note that enabling SQL comment propagation results in potentially confidential data (service names) being stored in the databases which can then be accessed by other third parties that have been granted access to the database._ | `'disabled'`                                                      |
+| `append_propagation` | | `Bool` | Appends the SQL comment propagation to the query string. Prepends the comment if `false`. For long query strings, the appended propagation comment might be truncated, causing loss of correlation between the query and trace. | `false` |
```
TonyCTHsu, line 1174: Perhaps rename to `append_comment` ?

lib/datadog/tracing/contrib/propagation/sql_comment/mode.rb
```diff
@@ -20,6 +20,14 @@ def service?
             def full?
               mode == Ext::FULL
             end
+
+            def prepend?
```
p-datadog, line 24: What is the purpose of having both `prepend?` and `append?`? Is this to satisfy an externally defined API? Because this PR only calls one of the two (`append?`).

docs/GettingStarted.md
p-datadog, line 1174: Is the option called `append` in the code?

docs/GettingStarted.md
```diff
@@ -1990,6 +1992,9 @@ client.query("SELECT * FROM users WHERE group='x'")
...
+| `comment_propagation` | `DD_DBM_PROPAGATION_MODE` | `String` | SQL comment propagation mode for database monitoring. <br />(example: `disabled` \| `service`\| `full`). <br /><br />**Important**: _Note that enabling SQL comment propagation results in potentially confidential data (service names) being stored in the databases which can then be accessed by other third parties that have been granted access to the database._ | `'disabled'` |
+| `append_propagation` | | `Bool` | Appends the SQL comment propagation to the query string. Prepends the comment if `false`. For long query strings, the appended propagation comment might be truncated, causing loss of correlation between the query and trace. | `false` |
+| `on_error` | | `Proc` | Custom error handler invoked when MySQL raises an error. Provided `span` and `error` as arguments. Sets error on the span by default. Useful for ignoring errors that are handled at the application level. | `proc { \|span, error\| span.set_error(error) unless span.nil? }` |
```
p-datadog, line 1997: Is this documenting an existing option that wasn't documented? Would be clearer to have this addition in its own PR.

docs/GettingStarted.md
TonyCTHsu: Oops, nice catch!

marcotc: Done!

lib/datadog/tracing/contrib/propagation/sql_comment/mode.rb
marcotc: Yeah, just API symmetry. I'll remove it.

docs/GettingStarted.md
marcotc: > Is this documenting an existing option that wasn't documented? 

Yes.

Can I keep it here, please? I will save me like 20 minutes and some stress.

PR #3810, opened 2024-07-25
DEBUG-2647 Run multiple Ruby micro-benchmark files

p-datadog:
**What does this PR do?**

This PR runs the profiling benchmarks, which already exist in the source tree, in addition to the tracing benchmarks.

The gem loading benchmark is moved to its own file, out of tracing file, because it applies to the entire library, not just to tracing.

This PR requires the benchmarking-platform tooling changes in https://github.com/DataDog/benchmarking-platform/pull/86.

**Motivation:**

I wrote some micro-benchmarks for dynamic instrumentation, but the existing benchmark runner hardcodes tracing as the only benchmark it's executing.

Changing the runner to run more than one benchmark permits executing existing profiling benchmarks.

**Additional Notes:**

**How to test the change?**

Current benchmark results: https://benchmarking.us1.prod.dog/benchmarks?projectId=5&page=1&ciJobDateStart=1719776035187&ciJobDateEnd=1722368035187&benchmarkGroupPipelineId=40500915&benchmarkGroupSha=191b99d7ee22245ead9d9b5292bfb0bab6f7d09a&benchmarkId=4574549


benchmarks/run_all.sh
```diff
@@ -0,0 +1,14 @@
...
+
+set -ex
+
+for file in \
+  `dirname "$0"`/tracing_trace.rb \
+  `dirname "$0"`/gem_loading.rb \
+  `dirname "$0"`/profiler_*.rb;
+do
+  bundle exec ruby "$file"
+done
```
ivoanjo, line 14: So I don't think it's blocking for this PR, but I wonder if we should try to put in some kind of check for files that aren't run.

E.g. [something along the lines of this check](https://github.com/DataDog/dd-trace-rb/blob/master/spec/datadog/profiling/validate_benchmarks_spec.rb#L29)?

Otherwise, it can be a bit confusing that if anyone adds a new profiler benchmark, it gets automatically picked up, but if you add any other kind of benchmark, it doesn't?

Alternatively, should we just run every `.rb` file on the `benchmark/` folder, and move anything that's not a benchmark to a `lib/` subdirectory or something like that? (Or maybe list here the files that shouldn't be run instead of the ones that should?)

benchmarks/gem_loading.rb
```diff
@@ -0,0 +1,52 @@
...
+
+    raise "Benchmark failed with status #{status}: #{output}" unless status.success?
+  end
+end
+
+puts "Current pid is #{Process.pid}"
+
+GemLoadingBenchmark.new.instance_exec do
+  benchmark_gem_loading
+end
```
ivoanjo, line 52: Don't forget to update https://github.com/DataDog/dd-trace-rb/blob/master/spec/datadog/tracing/validate_benchmarks_spec.rb#L12 :)

(Technically this is not a tracing-specific thing but I think it's fine to keep it there for now? -- as long as we test it _somewhere_)

ivoanjo (approved):
Left some notes but it overall LGTM :+1:

benchmarks/run_all.sh
p-datadog: I was actually not sure what the right approach to invoking "all" benchmarks was. Let me enumerate the files explicitly in this PR and try to improve the situation later.

p-datadog: Move of supporting files to `lib` is actually part of https://github.com/DataDog/dd-trace-rb/pull/3787 which will be on top of this PR.

benchmarks/gem_loading.rb
p-datadog: I added a readme to the benchmarks directory with this point as a reminder.

p-datadog:
The benchmarks are now tested in CI but the profiling ones are failing due to what looks like the C extension(s) not being built. Does that file need to run in a particular configuration?

benchmarks/run_all.sh
ivoanjo: Sounds reasonable -- now there's less magic, but it's a bit more consistent :+1:

ivoanjo:
> The benchmarks are now tested in CI but the profiling ones are failing due to what looks like the C extension(s) not being built. Does that file need to run in a particular configuration?

Yes:

* The profiler isn't supported on macOS (*you can build and use it for our own development, but needs some manual steps) so this is why the GitHub macOS builds don't work

* The profiler isn't compiled by default when running the test suite. You can see this in the `Rakefile` -- `compile_native_extensions` is only added as a dependency to `profiling` stuff. This is for the same reason as above in a way -- most of the team works on macOS so I didn't want to break their testing.

TBH my suggestion would be to keep the `validate_benchmarks_spec.rb` for profiling specs separate, and inside the profiling folder, so you don't need to care about any of the two points above ;)

ivoanjo (approved):
:+1: LGTM, thanks again!

p-datadog:
I found the profiler benchmark validator and have done I think what you initially requested, which was to add a validator for the library benchmarks that I split out of tracing. I also added some text to the readme to mention the various validators that need to be updated.

Issue #3804, opened 2024-07-23
GraphQL unified tracer not autoloaded


WoutDev:
**Current behaviour**
Following the [documentation](https://docs.datadoghq.com/tracing/trace_collection/automatic_instrumentation/dd_libraries/ruby/#graphql), I'm getting a `NameError` when attempting to use the new GraphQL unified tracer:
```
NameError:
  uninitialized constant Datadog::Tracing::Contrib::GraphQL::UnifiedTrace
```

This is solved by requiring the file manually:
```ruby
require 'datadog/tracing/contrib/graphql/unified_trace'
```

So not sure if this is a documentation issue, or it should be autoloaded while it is not.

**Expected behaviour**
Looking at the documentation, it looks like a drop-in replacement for `GraphQL::Tracing::DataDogTrace` and 'just work' instead of raising a `NameError`.

**Steps to reproduce**
Use a custom tracer based on UnifiedTrace:
```ruby
# frozen_string_literal: true

module Util
  module DatadogTracer
    include Datadog::Tracing::Contrib::GraphQL::UnifiedTrace

    def prepare_span(key, data, span)
      # REDACTED: add some tags
    end
  end
end
```

and `trace_with` the custom tracer like this:
```ruby
# frozen_string_literal: true

class SomeSchema < GraphQL::Schema
  trace_with Util::DatadogTracer
end
```

**How does `datadog` help you?**
Some great tracing and profiling :)

**Environment**

* **datadog version: 2.2.0**
* **Configuration block (`Datadog.configure ...`):**
```ruby
Datadog.configure do |c|
  c.runtime_metrics.enabled = true

  # REDACTED: some non-graphql instrumentation here

  c.tracing.report_hostname = true

  c.profiling.allocation_enabled = true
end
```
* **Ruby version:** 3.3.0
* **Operating system:** Windows 11 WSL2 Ubuntu 22.04
* **Relevant library versions:** N/A

vpellan:
Hello ! A fix will be available in the next release, thanks for your patience !

vpellan:
Hello! `2.3.0` just released, give it a try!

PR #3813, opened 2024-07-29
Fix GraphQL patchers 'require' issue

vpellan:
**What does this PR do?**

Making a custom tracer based on unified_tracer should now not need to require explicitly the unified_tracer.rb file anymore. This also fixes the problem with currently unreleased appsec_tracer.

**Motivation:**

Fixes issue #3804 

**Additional Notes:**

**How to test the change?**

Create a Rails + GraphQL app with a custom tracer that includes unified_tracer. GraphQL AppSec test app is using this so if system-tests passes, it works.

### DIFF
--- a/lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb
+++ b/lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require 'graphql/language/nodes'
+require 'graphql'
 
 require_relative '../../../instrumentation/gateway/argument'
 
--- a/lib/datadog/appsec/contrib/graphql/patcher.rb
+++ b/lib/datadog/appsec/contrib/graphql/patcher.rb
@@ -2,6 +2,7 @@
 
 require_relative '../patcher'
 require_relative 'gateway/watcher'
+require_relative 'appsec_trace' if Gem.loaded_specs['graphql'] && Gem.loaded_specs['graphql'].version >= Gem::Version.new('2.0.19')
 
 module Datadog
   module AppSec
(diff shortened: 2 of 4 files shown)


marcotc (approved):

p-datadog:
Is it supported to load datadog first and graphql second, as well as the other way around?

We already have tests like https://github.com/DataDog/dd-trace-rb/blob/master/spec/loading_spec.rb that start with a clean environment, given that this was a customer-reported issue I would consider adding a test that verifies that the reported case works correctly going forward.

vpellan:
> Is it supported to load datadog first and graphql second, as well as the other way around?
> 
> We already have tests like [`master`/spec/loading_spec.rb](https://github.com/DataDog/dd-trace-rb/blob/master/spec/loading_spec.rb?rgh-link-date=2024-07-30T17%3A50%3A46Z) that start with a clean environment, given that this was a customer-reported issue I would consider adding a test that verifies that the reported case works correctly going forward.

I've added a test that will require datadog first, then graphql, and the other way around too.
I've also added a test for trace_with.

spec/datadog/tracing/contrib/graphql/loading_spec.rb
```diff
@@ -0,0 +1,31 @@
+RSpec.describe 'loading graphql' do
```
p-datadog, line 1: I would think that adding `require 'shellwords'` in this file is necessary so that it can be executed by itself? The only other place that requires this library is the top-level loading test that I can see.

p-datadog (approved):

spec/datadog/tracing/contrib/graphql/loading_spec.rb
vpellan: Thanks for the suggestion, I've added it !

PR #3816, opened 2024-07-31
[PROF-10275] Enable crashtracking telemetry by default when profiler is enabled

ivoanjo:
**What does this PR do?**

This PR enables the crashtracking telemetry feature by default when the profiler is enabled.

This feature can be disabled using the `DD_PROFILING_EXPERIMENTAL_CRASH_TRACKING_ENABLED` environment variable or the `c.profiling.advanced.experimental_crash_tracking_enabled`  setting via code.

This setting is not very well-named, as the feature is no longer experimental. But as we're actually considering separating crashtracking from profiling, I decided to leave the current setting in place until we can decide where this is going to live, since that's going to influence where the new setting without "experimental" on its name should land.

**Motivation:**

Crashtracking allows us to identify issues that may otherwise go unnoticed. We've been deploying it on other Datadog libraries with great success so far.

**Additional Notes:**

We've had configurations running with crashtracking enabled on our usual validation environments + almost all (1 remaining!) internal Ruby apps also has been running with this feature.

We've not seen any noticeable impact for enabling this feature.

In fact, now that it's going to be the default, I'll go ahead and remove the separate configurations we had for testing it.

**How to test the change?**

The following snippet starts a Ruby instance with profiling, and crashes it immediately:

```bash
DD_PROFILING_ENABLED=true bundle exec ddprofrb exec ruby -e 'Process.kill("SEGV", Process.pid)'
```

You should see an error report submitted to the agent.

### DIFF
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -109,13 +109,6 @@ only-profiling-heap:
     DD_PROFILING_ALLOCATION_ENABLED: "true"
     DD_PROFILING_EXPERIMENTAL_HEAP_ENABLED: "true"
 
-only-profiling-crashtracking:
-  extends: .benchmarks
-  variables:
-    DD_BENCHMARKS_CONFIGURATION: only-profiling
-    DD_PROFILING_ENABLED: "true"
-    DD_PROFILING_EXPERIMENTAL_CRASH_TRACKING_ENABLED: "true"
-
 profiling-and-tracing:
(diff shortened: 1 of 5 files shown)


p-datadog (approved):

ivoanjo:
The one CI failure is being tracked separately and is unrelated to this PR (it's affecting all PRs for the past ~2 days).

I'm going to go ahead and merge this one in!

PR #3818, opened 2024-07-31
[PROF-10128] Fix profiling benchmarking configuration still having auto_instrument overhead

ivoanjo:
**What does this PR do?**

This PR tweaks the `ADD_TO_GEMFILE` environment variable for benchmarking configurations where only the profiler should be active to not include `require: 'datadog/auto_instrument'`.

It turns out this change was added in #3614 when we fixed the benchmarks not running. Loading the instrumentation code and then disabling tracing has a measurable performance overhead vs not loading it at all (20% regression of median throughput in high load gitlab tests).

**Motivation:**

I expect that with this change, the `only-profiling` configuration will reflect historical data we had before the change in #3614.

This is relevant because we use this data as basis for comparison with other profiling features (e.g. allocation) and thus it's important to have a correct baseline.

**Additional Notes:**

I noticed a similar-ish configuration issue in our reliability environment. I'll open a separate PR for that one as well.

**How to test the change?**

I've tested a similar configuration using `bp-runner` locally and confirmed that overhead is reduced. Once we merge this in, I'll validate that the nightly runs show performance is back to where it was on the 18th of April.

### DIFF
--- a/.gitlab/benchmarks.yml
+++ b/.gitlab/benchmarks.yml
@@ -85,6 +85,7 @@ only-profiling:
   variables:
     DD_BENCHMARKS_CONFIGURATION: only-profiling
     DD_PROFILING_ENABLED: "true"
+    ADD_TO_GEMFILE: "gem 'datadog', github: 'datadog/dd-trace-rb', ref: '$CI_COMMIT_SHA'"
 
 only-profiling-alloc:
   extends: .benchmarks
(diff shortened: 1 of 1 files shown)


ivoanjo:
Here's a screenshot showing how tracing was still getting loaded even in the `only-profiling` configuration:
> image

ivoanjo:
And here's how my run with `bp-runner` looks showing there's no stack traces from the `datadog` gem instrumenting rack:

image

TonyCTHsu (approved):

PR #3823, opened 2024-08-01
Sidekiq:Add trace correlation for internal logs

### LINKED ISSUE #1068: Datadog.tracer.active_correlation not available in Sidekiq logging
With Sidekiq 5 (released in 2017), job logging is no longer a middleware, and now sits above all middleware in the stack.

I think this is why `Datadog.tracer.active_correlation` is not available in the new JobLogger, because the correlation is flushed in the `Sidekiq::ServerTracer`  middleware.

Seems crazy that no one has noticed this before, so maybe I have it wrong?

My setup is like

```
Sidekiq.configure_server do |config|
  config.log_formatter = Logging::Formatters::Json.new
  config.options[:job_logger] = Logging::Sidekiq::JobLogger

    config.server_middleware do |chain|
    chain.add Logging::Sidekiq::Server::JobLogger
  end
end
```

`Datadog.tracer.active_correlation` is always 0 for space_id and trace_id in the `JobLogger`

`Datadog.tracer.active_correlation` in my `Logging::Sidekiq::Server::JobLogger` middleware has a span_Id, but it seems to be one of the inner spans, e.g. a postgres query. 

I suppose I could grab the root span in my custom middleware,  store it in `Thread.current[:dd_trace] = Datadog.tracer.active_root_span` and then retrieve it in the `JobLogger` class for logging, (with an `ensure` block to flush it) ? But since other people must have this problem and I can't find a documented solution, thought I'd bring it up here.

marcotc:
Fixes #1068 

This PR ensures that a trace is active so that Sidekiq's logger instance can emit correlated logs before and after a job is executed.

**Motivation:**

By default, Sidekiq logs a [few messages](https://github.com/sidekiq/sidekiq/blob/f14749c99af163f89d1de66c46d25baac0ec8f31/lib/sidekiq/job_logger.rb#L23-L31) before and after a job is executed in the Sidekiq server process.
These messages happen outside the context of Sidekiq server's middlewares. And because APM traces are created from a servers middleware, these log entries are not correlated with an APM trace.

These logs are important because they are emitted by default, and not having them correlated with traces creates a bad experience out-of-the-box when inspecting the logs from a new Sidekiq server.

**How to test the change?**
There are new integration tests to run a realistic Sidekiq server.

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -1847,6 +1847,21 @@ end
 | `on_error`            | | `Proc` | Custom error handler invoked when a job raises an error. Provided `span` and `error` as arguments. Sets error on the span by default. Useful for ignoring transient errors.                                                                                                                                                                                                                                                                           | `proc { \|span, error\| span.set_error(error) unless span.nil? }` |
 | `quantize`            | | `Hash` | Hash containing options for quantization of job arguments.                                                                                                                                                                                                                                                                                                                                                                                            | `{}`                                                              |
 
+#### Log Correlation
+
+To correlate Sidekiq logs with traces, you can configure [Sidekiq's logger](https://github.com/sidekiq/sidekiq/wiki/Logging)
+to either use an [existing supported logger instance](#for-logging-in-rails-applications) or use Sidekiq's JSON logger formatter:
+
+```ruby
+Sidekiq.configure_client do |config|
+   config.logger.formatter = Sidekiq::Logger::Formatters::JSON.new
+end
+
+Sidekiq.configure_server do |config|
+  config.logger.formatter = Sidekiq::Logger::Formatters::JSON.new
+end
+```
+
 ### Sinatra
 
 The Sinatra integration traces requests and template rendering.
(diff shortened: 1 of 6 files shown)


alexevanczuk (approved):
Woo hoo! Thanks for doing this, very excited to try this out.

urseberry (approved):

anmarchenko (approved):

marcotc:
To get all test scenarios to pass, I'll try a different approach of focusing the code changes on the JobLogger.

alexevanczuk:
Any more luck on this one @marcotc ? Very eager to get correlated logs for sidekiq 🙌🏼

marcotc:
It's way trickier that we though, sorry @alexevanczuk, we are still trying to figure out the best approach.

alexevanczuk:
Okay @marcotc , thanks for your perseverance!!

nqlongdl:
Will this be continue @marcotc? 👀 

If I understand correctly, after this PR, we just need to setup the necessary config for Datadog when initialize Sidekiq, and then the logs can be automatically attached with fields for tracing correlation like trace_id, etc.?

Drowze:
Very much looking forward to this! Sidekiq logs not getting correlated is something that has bothered our devs for a long time.

FiloSpaTeam:
Any news for this?

Drowze:
It's been a while since this PR was opened and the original issue still remains (sidekiq's before/after logs AND exception logs aren't co-related to their respective span).

@marcotc can we prioritize this fix somehow? 🙇 

---

_Screenshot of a Sidekiq log - note the missing Trace ID_
Screenshot 2026-06-03 at 14 35 04

PR #3826, opened 2024-08-06
Enable crashtracking without profiler

TonyCTHsu:
**What does this PR do?**

This PR extracts crash tracking capabilities from profiler. This means crash tracking could be enabled without profiler.


lib/datadog/core/crashtracking/agent_base_url.rb
```diff
@@ -0,0 +1,25 @@
...
+
+require_relative '../configuration/ext'
+
+module Datadog
+  module Core
+    module Crashtracking
+      module AgentBaseUrl
+        module_function
+
+        def resolve(agent_settings)
```
ivoanjo, line 11: Minor: I suggest not using `module_function` as it also makes this module something we can include but we have no plans of including it?

```suggestion
        def self.resolve(agent_settings)
```

lib/datadog/core/crashtracking/component.rb
```diff
@@ -0,0 +1,101 @@
...
+          rescue LoadError => e
+            e.message
+          end
+
+        def self.build(settings, agent_settings)
+          tags = TagBuilder.call(settings)
+          agent_base_url = AgentBaseUrl.resolve(agent_settings)
+          unless agent_base_url
+            Datadog.logger.warn('Missing agent base URL; cannot enable crash tracking')
+          end
```
ivoanjo, line 32: Minor: Consider moving this error message inside the `resolve`? Right now if we fail this part there's actually two warning-level messages about the same error.

ivoanjo, line 32: Minor: Consider moving this error message inside the `resolve`? Right now if we fail this part there's actually two warning-level messages about the same error.

lib/datadog/core/crashtracking/tag_builder.rb
```diff
@@ -0,0 +1,44 @@
...
+module Datadog
+  module Core
+    module Crashtracking
+      # Builds a hash of default plus user tags to be included in a profile
+      module TagBuilder
+        module_function
+
+        def call(settings)
+          # When changing or adding these, make sure they are kept in sync with
+          # https://docs.google.com/spreadsheets/d/1LOGMf4c4Avbtn36uZ2SWvhIGKRPLM1BoWkUP4JYj7hA/ (Datadog internal link)
```
ivoanjo, line 17: This doc is for profiling so I think the comment doesn't really make sense for the crashtracker.

lib/datadog/core/crashtracking/tag_builder.rb
```diff
@@ -0,0 +1,44 @@
...
+
+require_relative '../utils'
+require_relative '../environment/socket'
+require_relative '../environment/identity'
+require_relative '../environment/git'
+
+module Datadog
+  module Core
+    module Crashtracking
+      # Builds a hash of default plus user tags to be included in a profile
```
ivoanjo, line 11: Minor: Still mentions profiler

lib/datadog/core/crashtracking/tag_builder.rb
```diff
@@ -0,0 +1,44 @@
...
+        module_function
+
+        def call(settings)
+          # When changing or adding these, make sure they are kept in sync with
+          # https://docs.google.com/spreadsheets/d/1LOGMf4c4Avbtn36uZ2SWvhIGKRPLM1BoWkUP4JYj7hA/ (Datadog internal link)
+          tags = {
+            'host' => Environment::Socket.hostname,
+            'language' => Environment::Identity.lang,
+            'process_id' => Process.pid.to_s,
+            'profiler_version' => Environment::Identity.gem_datadog_version,
```
ivoanjo, line 22: Minor: still mentions profiler

lib/datadog/core/crashtracking/tag_builder.rb
```diff
@@ -0,0 +1,44 @@
...
+            'runtime_version' => Environment::Identity.lang_version,
+          }
+
+          tags['env'] = settings.env if settings.env
+          tags['service'] = settings.service if settings.service
+          tags['version'] = settings.version if settings.version
+          tags['git.repository_url'] = Environment::Git.git_repository_url if Environment::Git.git_repository_url
+          tags['git.commit.sha'] = Environment::Git.git_commit_sha if Environment::Git.git_commit_sha
+
+          # Make sure everything is an utf-8 string, to avoid encoding issues in native code/libddprof/further downstream
```
ivoanjo, line 36: Minor:
```suggestion
          # Make sure everything is an utf-8 string, to avoid encoding issues in native code/further downstream
```

lib/datadog/core/configuration/settings.rb
```diff
@@ -450,19 +450,6 @@ def initialize(*_)
...
-            # properly-named one soon.
-            #
-            # @default `DD_PROFILING_EXPERIMENTAL_CRASH_TRACKING_ENABLED` environment variable as a boolean,
-            # otherwise `true`
-            option :experimental_crash_tracking_enabled do |o|
-              o.type :bool
-              o.env 'DD_PROFILING_EXPERIMENTAL_CRASH_TRACKING_ENABLED'
-              o.default true
-            end
           end
```
ivoanjo, line 453: While this option is experimental, removing it is still a breaking change. I suggest replacing it with a deprecation message (similar to the ones we did in the 1.x-stable) branch.

ivoanjo (commented):
Left a few quick notes!

lib/datadog/core/crashtracking/tag_builder.rb
TonyCTHsu: Will change, still looking for a good name

lib/datadog/core/crashtracking/component.rb
```diff
@@ -0,0 +1,109 @@
...
+          @logger = logger
+        end
+
+        def start
+          Utils::AtForkMonkeyPatch.apply!
+
+          start_or_update_on_fork(action: :start)
+          ONLY_ONCE.run do
+            Utils::AtForkMonkeyPatch.at_fork(:child) do
+              # Must NOT reference `self` here, as it become stale after fork
```
ivoanjo, line 71: It's actually not made stale by the fork but by multiple calls to `Datadog.configure`, so here's a suggestion:

```suggestion
              # Must NOT reference `self` here, as only the first instance will
              # be captured by the ONLY_ONCE and we want to pick the latest active one
              # (which may have different tags or agent config)
```

lib/datadog/core/crashtracking/component.rb
```diff
@@ -0,0 +1,109 @@
...
+        rescue => e
+          logger.error("Failed to stop crash tracking: #{e.message}")
+        end
+
+        private
+
+        attr_reader :tags, :agent_base_url, :ld_library_path, :path_to_crashtracking_receiver_binary, :logger
+
+        def start_or_update_on_fork(action:)
+          logger.debug("Crash tracking #{action}...")
```
ivoanjo, line 93: Minor: Since in some cases we ask customers to enable debug logging, I suggest avoiding having a bit too much logging:

```suggestion
```

lib/datadog/core/crashtracking/component.rb
```diff
@@ -0,0 +1,109 @@
...
+          self.class._native_start_or_update_on_fork(
+            action: action,
+            exporter_configuration: [:agent, agent_base_url],
+            path_to_crashtracking_receiver_binary: path_to_crashtracking_receiver_binary,
+            ld_library_path: ld_library_path,
+            tags_as_array: tags.to_a,
+            upload_timeout_seconds: 1
```
ivoanjo, line 100: Did you end up discussing this with @danielsn ? A timeout of 1 second seems really small for this data -- previously we were using profiler's default upload timeout which is 30 seconds.

lib/datadog/core/crashtracking/tag_builder.rb
```diff
@@ -0,0 +1,39 @@
...
+  module Core
+    module Crashtracking
+      # This module builds a hash of tags
+      module TagBuilder
+        def self.call(settings)
+          hash = {
+            'host' => Environment::Socket.hostname,
+            'language' => Environment::Identity.lang,
+            'process_id' => Process.pid.to_s,
+            'library_version' => Environment::Identity.gem_datadog_version,
```
ivoanjo, line 18: I don't think this is needed, as the crashtracker API already gets this as a direct argument (in the C code)

```suggestion
```

lib/datadog/core/crashtracking/tag_builder.rb
```diff
@@ -0,0 +1,39 @@
...
+
+module Datadog
+  module Core
+    module Crashtracking
+      # This module builds a hash of tags
+      module TagBuilder
+        def self.call(settings)
+          hash = {
+            'host' => Environment::Socket.hostname,
+            'language' => Environment::Identity.lang,
```
ivoanjo, line 16: I don't think this is needed, as the crashtracker API already gets this as a direct argument (in the C code)

```suggestion
```

lib/datadog/core/crashtracking/tag_builder.rb
```diff
@@ -0,0 +1,39 @@
...
+            'process_id' => Process.pid.to_s,
+            'library_version' => Environment::Identity.gem_datadog_version,
+            'runtime' => Environment::Identity.lang, # This is known to be repeated from language, above
```
ivoanjo, line 19: I don't think this is needed, as the crashtracker API already gets this as a direct argument (in the C code)

```suggestion
```

lib/datadog/profiling/component.rb
```diff
@@ -73,8 +73,10 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
         exporter = build_profiler_exporter(settings, recorder, worker, internal_metadata: internal_metadata)
         transport = build_profiler_transport(settings, agent_settings)
         scheduler = Profiling::Scheduler.new(exporter: exporter, transport: transport, interval: upload_period_seconds)
-        crashtracker = build_crashtracker(settings, transport)
-        profiler = Profiling::Profiler.new(worker: worker, scheduler: scheduler, optional_crashtracker: crashtracker)
+        profiler = Profiling::Profiler.new(
+          worker: worker,
+          scheduler: scheduler
+        )
```
ivoanjo, line 79: Minor:

```suggestion
        profiler = Profiling::Profiler.new(worker: worker, scheduler: scheduler)
```

ivoanjo (commented):
I ran out of time for now, here's 1/3 of a review, I'll come back for the other files soon :)

.gitlab-ci.yml
```diff
@@ -153,4 +153,4 @@ vaccine:
       -H "Accept: application/vnd.github.v3+json" \
       -H "Authorization: token $GH_VACCINE_PAT" \
       https://api.github.com/repos/TonyCTHsu/vaccine/actions/workflows/vaccine.yml/dispatches \
-      -d '{"ref":"master", "inputs": {"commit_sha": "'$CI_COMMIT_SHA'"}}'
+      -d '{"ref":"tonycthsu/spec-for-crash", "inputs": {"commit_sha": "'$CI_COMMIT_SHA'"}}'
```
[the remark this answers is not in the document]
TonyCTHsu: Currently, no.

lib/datadog/core/configuration/settings.rb
```diff
@@ -833,6 +832,15 @@ def initialize(*_)
           option :service
         end
 
+        settings :crashtracking do
+          # Enables reporting of information when Ruby VM crashes.
+          option :enabled do |o|
+            o.type :bool
+            o.default true
+            o.env 'DD_CRASHTRACKING_ENABLED'
```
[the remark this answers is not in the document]
TonyCTHsu: This is from the discussion of @sanchda @kevingosse

lib/datadog/core/crashtracking/component.rb
TonyCTHsu: According to the discussion with @danielsn , there is no recommended value (Libdatadog's default is 3, but it is just copy and paste from the previous code). 

@danielsn mentioned that during a crash event, 1 second timeout is quite accommodating and make sense as an educated guess. (something between 1 to 3 sounds like a good start)

.gitlab-ci.yml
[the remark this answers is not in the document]
ivoanjo: Is there any documentation about this? Why isn't it under the datadog organization?

lib/datadog/core/configuration/settings.rb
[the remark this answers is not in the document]
ivoanjo: Got it, thanks for confirming :pray:

lib/datadog/core/crashtracking/component.rb
ivoanjo: It seems a bit low to me, but I'm happy to go with Daniel's proposal.

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+require 'webrick'
+require 'fiddle'
+
+RSpec.describe Datadog::Core::Crashtracking::Component,
+  skip: Datadog::Core::Crashtracking::Component::LIBDATADOG_API_FAILURE do
+    describe '.build' do
+      let(:settings) { Datadog::Core::Configuration::Settings.new }
+      let(:agent_settings) { double('agent_settings') }
+      let(:logger) { Logger.new($stdout) }
+      let(:tags) { {} }
```
ivoanjo, line 13: Minor: Consider adding some fake tags here, as it's easier for an empty hash to make the tests pass accidentally on a future refactoring.

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+            .and_return(tags)
+          expect(Datadog::Core::Crashtracking::AgentBaseUrl).to receive(:resolve).with(agent_settings)
+            .and_return(agent_base_url)
+          expect(::Libdatadog).to receive(:ld_library_path)
+            .and_return(ld_library_path)
+          expect(::Libdatadog).to receive(:path_to_crashtracking_receiver_binary)
+            .and_return(path_to_crashtracking_receiver_binary)
+
+          expect(described_class.build(settings, agent_settings, logger: logger)).to be_nil
+        end
```
ivoanjo, line 92: Consider adding a test for the `warn` that gets emitted when the required parameters are `nil`.

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+
+            expect(described_class).to receive(:_native_start_or_update_on_fork) { raise 'Test failure' }
+            expect(logger).to receive(:error).with(/Failed to start crash tracking: Test failure/)
+
+            crashtracker.start
+          end
+        end
+
+        it 'starts the crash tracker' do
+          crashtracker = build_crashtracker
```
ivoanjo, line 118: Consider extracting this into a `subject`

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+      around do |example|
+        wait_for { `pgrep -f libdatadog-crashtracking-receiver` }.to be_empty
+        example.run
+        wait_for { `pgrep -f libdatadog-crashtracking-receiver` }.to be_empty
+      end
+
+      describe '#start' do
+        context 'when _native_start_or_update_on_fork raises an exception' do
+          it 'logs the exception' do
+            logger = Logger.new($stdout)
```
ivoanjo, line 107: Consider extracting this into a `let` (it can even be shared with the `let` from `.build`, above)

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+        end
+
+        it 'starts the crash tracker' do
+          crashtracker = build_crashtracker
+
+          crashtracker.start
+
+          wait_for { `pgrep -f libdatadog-crashtracking-receiver` }.to_not be_empty
+
+          tear_down!
```
ivoanjo, line 124: Consider using `crashtracker.stop` here?

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+            crashtracker = build_crashtracker
+            crashtracker.start
+
+            another_crashtracker = build_crashtracker
+            another_crashtracker.update_on_fork
+
+            wait_for { `pgrep -f libdatadog-crashtracking-receiver`.lines.size }.to be 1
+
+            tear_down!
+          end
```
ivoanjo, line 225: Given we have test coverage that:

1. If you have multiple crashtracker instances, only one process gets started
2. When you update on fork, the existing process gets updated

This "multiple instances test" seems to be 1+2, but it doesn't seem to add coverage that we already didn't get with 1 and 2?

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -0,0 +1,351 @@
...
+              family: 'ruby',
+              tags: ['tag1:value1', 'tag2:value2'],
+            )
+            expect(crash_report_message[:files][:'/proc/self/maps']).to_not be_empty
+            expect(crash_report_message[:os_info]).to_not be_empty
+          end
+        end
+
+        context 'when forked' do
+          it 'ensures the latest configuration applied' do
```
ivoanjo, line 314: Since this is a very non-trivial issue (e.g. it took us multiple minutes to discuss and understand it), I think this would benefit from a comment explaining why exactly this test is here and what's the subtle problem being tested.

It's true that we have test coverage for the issue now, and that's great because this test will fail if someone undoes the very subtle change we introduced to fix this problem.

BUT once this test starts failing, without a good comment, whoever makes it fail will need to "reverse engineer" why this test is here and what's it doing. (They may also arrive to the conclusion that the test doesn't make any sense and choose to remove it instead, which would be doubly bad)

.gitlab/install_datadog_deps.rb
```diff
@@ -63,8 +63,21 @@
...
+}
+
+[
+  'debase-ruby_core_source',
+  'ffi',
+  'libddwaf',
+  'msgpack',
+  'libdatadog', # libdatadog MUST be installed before datadog
+  'datadog',
+].each do |gem|
```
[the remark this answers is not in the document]
TonyCTHsu: Do you think it still is necessary that there is test tracking dependency changes?

https://github.com/DataDog/dd-trace-rb/blob/a589a98952f46b67de4c828f6d2073c664cf3a9c/spec/datadog/release_gem_spec.rb#L75-L107

Rakefile
```diff
@@ -74,7 +74,7 @@ namespace :spec do
   task all: [:main, :benchmark,
              :rails, :railsredis, :railsredis_activesupport, :railsactivejob,
              :elasticsearch, :http, :redis, :sidekiq, :sinatra, :hanami, :hanami_autoinstrument,
-             :profiling]
+             :profiling, :crashtracking]
 
   desc '' # "Explicitly hiding from `rake -T`"
   RSpec::Core::RakeTask.new(:main) do |t, args|
```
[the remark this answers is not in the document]
TonyCTHsu: `:crashtracking` is considered to be a dependency for `:all`, not `:main` ?

sig/datadog/core/utils/at_fork_monkey_patch.rbs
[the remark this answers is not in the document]
TonyCTHsu: My editor is complaining about this wherever there is code invoking `AtForkMonkeyPatch`. I added the type to reduce the noise. 

I added a comment on the rbs since I was not able to fix 
```
lib/datadog/core/utils/at_fork_monkey_patch.rb:63:54: [warning] Cannot pass a value of type `::Proc` as a block-pass-argument of type `^() -> void`
│   ::Proc <: ^() -> void
│
│ Diagnostic ID: Ruby::BlockTypeMismatch
│
└             result = child_block.nil? ? super : super(&child_block)
                                                        ~~~~~~~~~~~~
```

.gitlab-ci.yml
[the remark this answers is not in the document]
TonyCTHsu: It is my personal development tool, not yet ready for sharing.

sig/datadog/core/configuration/settings.rbs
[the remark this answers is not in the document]
TonyCTHsu: Similar to https://github.com/DataDog/dd-trace-rb/pull/3826#discussion_r1716880330

lib/datadog/core/configuration/components.rb
```diff
@@ -58,6 +59,17 @@ def build_runtime_metrics_worker(settings)
...
+
+          def build_crashtracker(settings, agent_settings, logger:)
+            return unless settings.crashtracking.enabled
+
+            if (libdatadog_api_failure = Datadog::Core::Crashtracking::Component::LIBDATADOG_API_FAILURE)
+              logger.debug("Cannot enable crashtracking: #{libdatadog_api_failure}")
+              return
+            end
+
+            Datadog::Core::Crashtracking::Component.build(settings, agent_settings, logger: logger)
```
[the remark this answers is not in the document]
TonyCTHsu: > One thing I noticed (not sure if we want to do anything about it)

I understand there is no harm at this moment. The current crashtracking component is implemented without additonal phases or internal state of enabled/disabled. When enabled, it creates an enabled instance, this is straightforward and easy to manage.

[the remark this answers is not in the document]
ivoanjo: Ok, I don't mind leaving it as-is either, I'll mark as resolved then.

spec/datadog/core/crashtracking/component_spec.rb
TonyCTHsu: I am adopting:  https://thoughtbot.com/blog/lets-not

In essence: Reducing mysterious guest by making the test example self contained and avoid indirection with rspec's caching DSL. Not DRYing with extraction has the the tradeoff of increase verbosity.

spec/datadog/core/crashtracking/component_spec.rb
TonyCTHsu: https://github.com/DataDog/dd-trace-rb/pull/3826#discussion_r1717158258

spec/datadog/core/crashtracking/component_spec.rb
TonyCTHsu: I prefer to use the `tear_down!` with native stop, since it would be disambiguous in the situation where multiple instances are in place.

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -0,0 +1,361 @@
+require 'spec_helper'
+require 'datadog/core/crashtracking/component'
+
+require 'webrick'
+require 'fiddle'
+
+RSpec.describe Datadog::Core::Crashtracking::Component,
+  skip: !CrashtrackingHelpers.supported? do
```
ivoanjo, line 8: Minor: Small suggestion -- consider turning this into a regular `before { skip (...) }`

1. Adding a new helper to spec_helper means *every* run of rspec, even when testing locally on completely different files will now have to load a helper that's only ever used by this one file

2. It's kinda weird that this whole file has one level of indentation extra just because of this linebreak.

spec/support/crashtracking_helpers.rb
```diff
@@ -0,0 +1,17 @@
+require 'datadog/core/crashtracking/component'
+require 'support/platform_helpers'
+
+module CrashtrackingHelpers
+  def self.supported?
+    # Only works with MRI on Linux
+    if PlatformHelpers.mri? && PlatformHelpers.linux?
```
ivoanjo, line 7: Fingers crossed that we'll be able to add macOS to this list soon ;)

lib/datadog/core/crashtracking/component.rb
```diff
@@ -0,0 +1,110 @@
...
+require_relative 'agent_base_url'
+require_relative '../utils/only_once'
+require_relative '../utils/at_fork_monkey_patch'
+
+module Datadog
+  module Core
+    module Crashtracking
+      # Used to report Ruby VM crashes.
+      #
+      # NOTE: The crashtracker native state is a singleton; so even if you create multiple instances of `Crashtracker`
```
ivoanjo, line 15: Minor:

```suggestion
      # NOTE: The crashtracker native state is a singleton; so even if you create multiple instances of
      # `Crashtracking::Component`
```

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -0,0 +1,361 @@
...
+            .and_return(tags)
+          expect(Datadog::Core::Crashtracking::AgentBaseUrl).to receive(:resolve).with(agent_settings)
+            .and_return(agent_base_url)
+          expect(::Libdatadog).to receive(:ld_library_path)
+            .and_return(ld_library_path)
+          expect(::Libdatadog).to receive(:path_to_crashtracking_receiver_binary)
+            .and_return(path_to_crashtracking_receiver_binary)
+          expect(logger).to_not receive(:warn)
+
+          component = double(instance_double(described_class))
```
ivoanjo, line 30: I'm not even quite sure what `double(instance_double(...))` does, but I suspect it may not be the right thing ;)

```suggestion
          component = instance_double(described_class)
```

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -0,0 +1,361 @@
...
+        context 'when forked' do
+          # This integration test coverages the case that
+          # the callback registered with `Utils::AtForkMonkeyPatch.at_fork`
+          # does not contain a stale instance of the crashtracker component.
+          it 'ensures the latest configuration applied' do
+            allow(described_class).to receive(:_native_start_or_update_on_fork)
+
+            # `Datadog.configure` to trigger crashtracking component reinstantiation,
+            #  a callback is first registered with `Utils::AtForkMonkeyPatch.at_fork`,
+            #  but not with the second `Datadog.configure` invokation.
```
ivoanjo, line 326: Nice description :+1:

ivoanjo (approved):
:+1: LGTM

.gitlab/install_datadog_deps.rb
```diff
@@ -99,6 +106,14 @@
   end
 end
 
+raise "#{gem_version_mapping.keys.join(',')} are not installed." if gem_version_mapping.any?
+
+datadog_gem_path = versioned_path.join("gems/datadog-#{ENV.fetch('RUBY_PACKAGE_VERSION')}")
+libdatadog_so_file = "libdatadog_api.#{RUBY_VERSION[/\d+.\d+/]}_#{RUBY_PLATFORM}.so"
+unless File.exist?("#{datadog_gem_path}/lib/#{libdatadog_so_file}")
+  raise "Missing #{libdatadog_so_file} in #{datadog_gem_path}."
+end
```
ivoanjo, line 115: You can also get this check by setting the `DD_FAIL_INSTALL_IF_MISSING_EXTENSION` env variable to `"true"` when calling `gem install` :)

lib/datadog/core/crashtracking/tag_builder.rb
```diff
@@ -0,0 +1,37 @@
...
+            'runtime_engine' => Environment::Identity.lang_engine,
+            'runtime-id' => Environment::Identity.id,
+            'runtime_platform' => Environment::Identity.lang_platform,
+            'runtime_version' => Environment::Identity.lang_version,
+            'env' => settings.env,
+            'service' => settings.service,
+            'version' => settings.version,
+            'git.repository_url' => Environment::Git.git_repository_url,
+            'git.commit.sha' => Environment::Git.git_commit_sha,
+            'is_crash' => true
```
ivoanjo, line 26: Nice catch! I was talking with @sanchda about this yesterday and he mentioned libdatadog wasn't adding these automatically.

In Python we also add `severity:crash` (see https://github.com/DataDog/dd-trace-py/pull/10199 ). I was planning on doing a PR to add this one as well, but in case you want to go ahead ;)

ivoanjo (approved):
:+1: LGTM

PR #3829, opened 2024-08-07
[PROF-10241] Extract profiler `at_fork` monkey patch to utils

ivoanjo:
**What does this PR do?**

This PR extracts the profiler `Datadog::Profiling::Ext::Forking` monkey patch to now live under `Datadog::Core::Utils::AtForkMonkeyPatch`.

It also does a clean up pass on some of the comments and specs, as well as rename some of the modules used for the actual monkey patching.

**Motivation:**

This monkey patch provides a `datadog_at_fork` mechanism that can be used to run code just after a process gets forked.

Here's a quick example:

```ruby
$ bundle exec pry
[1] pry(main)> require 'datadog/core/utils/at_fork_monkey_patch'
=> true
[2] pry(main)> Datadog::Core::Utils::AtForkMonkeyPatch.apply!
=> true
[3] pry(main)> Process.datadog_at_fork(:child) { puts "This code is running in a child process!" }
=> nil
[4] pry(main)> fork { puts "Hello from fork" }
This code is running in a child process!
Hello from fork
=> 447306
```

Up until now, only the profiler needed this behavior of getting auto-started right after a process forks.

Every other component of the datadog gem lazily waited until it got called again (e.g. the tracing bits would not get re-initialized until you tried to call the tracer).

With the introduction of the crashtracker, we also want the crashtracker to auto-start after a fork, so I'm extracting the monkey patch so we can use it on that component as well.

**Additional Notes:**

I'm preparing a follow-up PR to add support for `Process._fork`. 
It's been a long time coming, and it's not that big of a change.

**How to test the change?**

Our existing specs already cover this change. Also see the example above.


lib/datadog/core/utils/at_fork_monkey_patch.rb
```diff
@@ -50,40 +51,42 @@ def fork
...
           module_function
 
           def datadog_at_fork_blocks
-            # Blocks should be shared across all users of this module,
+            # Blocks are shared across all users of this module,
             # e.g. Process#fork, Kernel#fork, etc. should all invoke the same callbacks.
-            # rubocop:disable Style/ClassVars
-            @@datadog_at_fork_blocks ||= {}
-            # rubocop:enable Style/ClassVars
+            @@datadog_at_fork_blocks ||= {} # rubocop:disable Style/ClassVars
```
p-datadog, line 79: I feel like this should be protected by a mutex?

p-datadog (approved):

TonyCTHsu (approved):

lib/datadog/core/utils/at_fork_monkey_patch.rb
ivoanjo: Yeah, it's an excellent point. When this code runs the first time we shouldn't see concurrent forking, but "shouldn't" is not "won't".

Since we want to ship some of these changes as soon as possible, I've gone ahead and created https://github.com/DataDog/dd-trace-rb/issues/3831 and I'll come back to refactor this in a few days :)

PR #3830, opened 2024-08-07
[PROF-10241] Use Process._fork hook in `at_fork` monkey patch on Ruby 3.1+

ivoanjo:
**What does this PR do?**

This PR solves the long-standing "TODO" we had on the `at_fork` monkey patch where on Ruby 3.1+ there's a VM hook for instrumenting fork operations that avoids us needing to monkey patch `Kernel` and `Object`.

To use the new hook we need to monkey patch `Process._fork` instead (see https://github.com/ruby/ruby/pull/5017 and
https://bugs.ruby-lang.org/issues/17795).

Thus, I've added this monkey patch and disabled patching of `Kernel` and `Object` on Ruby 3.1+.

**Motivation:**

Avoid monkey patching core classes that no longer need to.

**Additional Notes:**

I'm not quite happy with either the AtForkMonkeyPatch or its specs, but I decided to not do a full revamp for now.

TIP: I recommend reviewing this PR with no whitespace in the diff.

**How to test the change?**

Here's a tiny test app that also exercises this feature:

```ruby
puts RUBY_DESCRIPTION

require 'datadog/core/utils/at_fork_monkey_patch'

Datadog::Core::Utils::AtForkMonkeyPatch.apply!

Process.datadog_at_fork(:child) { puts "Hello from child process: #{Process.pid} "}

puts "Parent pid: #{Process.pid}"

puts "fork { }"
fork { }
Process.wait

puts "Kernel.fork { }"
Kernel.fork { }
Process.wait

puts "Process.fork { }"
Process.fork { }
Process.wait

puts "Foo.new.call"
class Foo
  def call
    fork { }
  end

  def self.do_fork
    fork {  }
  end
end
Foo.new.call
Process.wait

puts "Foo.do_fork"
Foo.do_fork
Process.wait

puts "Fork from outside"
Foo.new.send(:fork) { }
Process.wait

class BasicFoo < BasicObject
  include ::Kernel

  def call
    fork { }
  end

  def self.do_fork
    fork { }
  end
end

puts "BasicFoo.new.call"
BasicFoo.new.call
Process.wait

puts "BasicObject:"
Class.new(BasicObject) { include(::Kernel); def call; fork { }; end }.new.call
Process.wait

puts "BasicFork.do_fork"
BasicFoo.do_fork
Process.wait
```

P.s.: You may notice when you run this example on < 3.1 that `BasicFoo.new.call` and `BasicObject` are not instrumented!

This was actually a known-gap from the old instrumentation.

### DIFF
--- a/lib/datadog/core/utils/at_fork_monkey_patch.rb
+++ b/lib/datadog/core/utils/at_fork_monkey_patch.rb
@@ -73,7 +73,7 @@ module Datadog
 
             pid = super
 
-            datadog_at_fork_blocks[:child].each(&:call) if datadog_at_fork_blocks.key?(:child) && pid == 0
+            datadog_at_fork_blocks[:child].each(&:call) if pid == 0 && datadog_at_fork_blocks.key?(:child)
 
             pid
           end
--- a/lib/datadog/core/utils/at_fork_monkey_patch.rb
+++ b/lib/datadog/core/utils/at_fork_monkey_patch.rb
@@ -3,15 +3,8 @@
 module Datadog
   module Core
     module Utils
-      # Monkey patches `Kernel#fork` and similar functions, adding a `Kernel#datadog_at_fork` callback mechanism which
+      # Monkey patches `Kernel#fork` and similar functions, adding a `Process#datadog_at_fork` callback mechanism which
       # is used to restart observability after the VM forks (e.g. in multiprocess Ruby apps).
-      #
-      # TODO: Use `Process._fork` on Ruby 3.1+, see
-      #       https://github.com/ruby/ruby/pull/5017 and https://bugs.ruby-lang.org/issues/17795
-      #
-      # Known limitations: Does not handle `BasicObject`s that include `Kernel` directly; e.g.
-      # `Class.new(BasicObject) { include(::Kernel); def call; fork { }; end }.new.call`.
-      # This will be fixed once we move to hooking into `Process._fork`
       module AtForkMonkeyPatch
         def self.supported?
           Process.respond_to?(:fork)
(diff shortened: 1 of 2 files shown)


TonyCTHsu (approved):

lib/datadog/core/utils/at_fork_monkey_patch.rb
```diff
@@ -80,11 +64,23 @@ def datadog_at_fork_blocks
...
+        # Adds `datadog_at_fork` behavior; see parent module for details.
         module ProcessDaemonMonkeyPatch
+          # Hook provided by Ruby 3.1+ for observability libraries that want to know about fork, see
+          # https://github.com/ruby/ruby/pull/5017 and https://bugs.ruby-lang.org/issues/17795
+          def _fork
+            datadog_at_fork_blocks = Datadog::Core::Utils::AtForkMonkeyPatch::KernelMonkeyPatch.datadog_at_fork_blocks
+
+            pid = super
+
+            datadog_at_fork_blocks[:child].each(&:call) if datadog_at_fork_blocks.key?(:child) && pid == 0
```
p-datadog, line 76: I would check `pid` before `key?` for efficiency.

lib/datadog/core/utils/at_fork_monkey_patch.rb
```diff
@@ -80,11 +64,23 @@ def datadog_at_fork_blocks
           end
         end
 
-        # A call to Process.daemon ( https://rubyapi.org/3.1/o/process#method-c-daemon ) forks the current process and
-        # keeps executing code in the child process, killing off the parent, thus effectively replacing it.
-        #
-        # This monkey patch makes the `Kernel#datadog_at_fork` mechanism defined above also work in this situation.
+        # Adds `datadog_at_fork` behavior; see parent module for details.
         module ProcessDaemonMonkeyPatch
```
p-datadog, line 68: Should the module now be called just `ProcessMoneyPatch`?

p-datadog (approved):

p-datadog (approved):

lib/datadog/core/utils/at_fork_monkey_patch.rb
ivoanjo: Process :moneybag: Patch :rofl: 

Fixed in 0c719dd506063a233c8b436b0217e3a9e057e3dc

ivoanjo: Process :moneybag: Patch :rofl: 

Fixed in 0c719dd506063a233c8b436b0217e3a9e057e3dc

lib/datadog/core/utils/at_fork_monkey_patch.rb
ivoanjo: Fixed in 9cf5444505bac53e007780f8bbbe8b6752f963c4

p-datadog (approved):
:+1:

PR #3833, opened 2024-08-08
[NO-TICKET] Don't put results in benchmarking folder directly

ivoanjo:
**What does this PR do?**

This PR tweaks the configuration for benchmark output results to place result files in the current directory the benchmark gets executed from, instead of placing the results in the benchmarks directory.

**Motivation:**

In #3810 we standardized the results output file name in benchmarks to match the benchmark name.

In that PR, we used `__FILE__` as a prefix. This changed where results are placed: where previously they were placed in the current folder where the benchmarks were run from (often the root of the repo), with this PR, they started getting placed in the benchmarks directory.

This clashes with our `validate_benchmarks_spec.rb` that look for files in those directories, e.g. running a benchmark and then running the test suite will make the test suite fail which is somewhat annoying.

While I could've changed the tests to filter out results files, I also find it useful to place the results where I'm executing the benchmarks from, as it makes organization easier (you just run the benchmark from where you want and you get the result there ).

**Additional Notes:**

N/A

**How to test the change?**

Run the benchmarks and confirm the results file gets placed in the current folder! :)

### DIFF
--- a/benchmarks/library_gem_loading.rb
+++ b/benchmarks/library_gem_loading.rb
@@ -34,7 +34,7 @@ class GemLoadingBenchmark
           raise unless status.success?
         end
 
-        x.save! "#{__FILE__}-results.json" unless VALIDATE_BENCHMARK_MODE
+        x.save! "#{File.basename(__FILE__)}-results.json" unless VALIDATE_BENCHMARK_MODE
         x.compare!
       end
     RUBY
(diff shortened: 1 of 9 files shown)


marcotc (approved):

ivoanjo:
I need to tweak our test execution harness before merging this, it's failing with

```
+ cp 'benchmarks/*-results.json' /go/src/github.com/DataDog/apm-reliability/dd-trace-rb/artifacts/candidate
cp: cannot stat 'benchmarks/*-results.json': No such file or directory
```

ivoanjo:
This is currently blocked by https://github.com/DataDog/benchmarking-platform/pull/92

ddyurchenko (approved):

PR #3836, opened 2024-08-08
Suppress startup logs in test and development environments

### LINKED ISSUE #3807: suppress logs by default when running minitest tests
when running tests, we get a lot of startup diagnostic logging that is unimportant/noisy. because the gem gets required before rails initializers are loaded, we can't seem to set `Datadog.configuration.diagnostics.startup_logs.enabled` to false in time before the output occurs.

when [looking at the source](https://github.com/DataDog/dd-trace-rb/blob/2f591a149505a95dcff50ce8378c9eb101eee544/lib/datadog/core/diagnostics/environment_logger.rb#L34), it seems what i'm desiring happens when running a program name ending in `rspec`.

checking `$PROGRAM_NAME` doesn't exactly work though for running rails tests the built-in way. when running `bin/rails t`, the program name is `bin/rails`. but you can also run an individual minitest with `bundle exec ruby ...`, in which case program name would be `ruby`.

would there be any gotchas in just checking `ENV["RAILS_ENV"] == "test"`?

marcotc:
Fixes #3807

This PR aligns the development environment detection for remote configuration, telemetry and, now, startup logs.

The development environment detection for startup logs now includes test environments and local development environments.

### DIFF
--- a/lib/datadog/core/diagnostics/environment_logger.rb
+++ b/lib/datadog/core/diagnostics/environment_logger.rb
@@ -8,6 +8,8 @@ module Datadog
   module Core
     module Diagnostics
       # Base class for EnvironmentLoggers - should allow for easy reporting by users to Datadog support.
+      #
+      # The EnvironmentLogger should not pollute the logs in a development environment.
       module EnvironmentLogging
         def log_configuration!(prefix, data)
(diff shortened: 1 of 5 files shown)


anmarchenko (approved):
🎉

ivoanjo (approved):
:+1: LGTM, thanks for picking this one up!

PR #3837, opened 2024-08-09
[PROF-10128] Refactor and speed up profiler stack sampling

ivoanjo:
**What does this PR do?**

This PR refactors the way the profiler samples the Ruby stack, and then builds on that refactoring to speed up sampling.

It also opens the door to future speedups once libdatadog allows us to skip re-interning strings, frames and stacks that we know have been sampled before.

Specifically this PR does the following:
1. It tweaks the data that `ddtrace_rb_profile_frames` returns from sampling the Ruby stack so that the stack collector can use cheaper APIs to get the same information as before.
2. It changes the sampling buffer we put stack traces in to be per-thread. This allows us to reuse some of the work we do from sample to sample, in cases where frames on the stack are the same. Thus sampling the same stack over and over will be slightly cheaper.
3. It adds a few more micro-optimizations that I spotted from profiling the profiler :)
4. When we sample stacks, we expose a flag that can be used to determine if a frame was unchanged from the last sample. This is what opens the door for future optimizations using libdatadog.

Here's the results of running benchmarks/profiler_sample_loop_v2.rb with these changes on my local machine:

```
stack collector improved-sampling
                       783.000 i/100ms
stack collector improved-sampling
                          7.828k (± 3.3%) i/s -    234.900k in  30.042969s

stack collector master
                       746.000 i/100ms
stack collector master
                          7.057k (± 3.3%) i/s -    211.864k in  30.058848s

Comparison:
stack collector improved-sampling:     7827.8 i/s
stack collector master:     7056.8 i/s - 1.11x  slower
```

**Motivation:**

This PR further reduces the overhead of profiling, especially more complex/deep stacks. I started looking into this with the goal of improving allocation profiling, but stack sampling is shared with cpu/wall-time profiling as well, so those will also be able to take advantage of the improvement.

**Additional Notes:**

There's a lot of small parts moving around in this PR. I recommend reviewing commit-by-commit, as I tried to break every change into a small standalone step that can be reviewed in a much nicer way.

**How to test the change?**

Our existing tests cover these changes.

### DIFF
--- a/ext/datadog_profiling_native_extension/collectors_stack.c
+++ b/ext/datadog_profiling_native_extension/collectors_stack.c
@@ -16,10 +16,10 @@ static VALUE missing_string = Qnil;
 // Used as scratch space during sampling
 struct sampling_buffer {
   unsigned int max_frames;
+  ddog_prof_Location *locations;
   VALUE *stack_buffer;
   int *lines_buffer;
   bool *is_ruby_frame;
-  ddog_prof_Location *locations;
 }; // Note: typedef'd in the header to sampling_buffer
 
 static VALUE _native_sample(
@@ -105,7 +105,8 @@ static VALUE _native_sample(
   int max_frames_requested = NUM2INT(max_frames);
   if (max_frames_requested < 0) rb_raise(rb_eArgError, "Invalid max_frames: value must not be negative");
 
-  sampling_buffer *buffer = sampling_buffer_new(max_frames_requested);
+  ddog_prof_Location *locations = ruby_xcalloc(max_frames_requested, sizeof(ddog_prof_Location));
+  sampling_buffer *buffer = sampling_buffer_new(max_frames_requested, locations);
 
   ddog_prof_Slice_Label slice_labels = {.ptr = labels, .len = labels_count};
 
@@ -126,6 +127,7 @@ static VALUE _native_sample(
     );
   }
 
+  ruby_xfree(locations);
   sampling_buffer_free(buffer);
 
   return Qtrue;
@@ -391,18 +393,17 @@ uint16_t sampling_buffer_check_max_frames(int max_frames) {
   return max_frames;
 }
 
-sampling_buffer *sampling_buffer_new(uint16_t max_frames) {
+sampling_buffer *sampling_buffer_new(uint16_t max_frames, ddog_prof_Location *locations) {
   sampling_buffer_check_max_frames(max_frames);
 
   // Note: never returns NULL; if out of memory, it calls the Ruby out-of-memory handlers
   sampling_buffer* buffer = ruby_xcalloc(1, sizeof(sampling_buffer));
 
   buffer->max_frames = max_frames;
-
+  buffer->locations = locations;
   buffer->stack_buffer  = ruby_xcalloc(max_frames, sizeof(VALUE));
   buffer->lines_buffer  = ruby_xcalloc(max_frames, sizeof(int));
   buffer->is_ruby_frame = ruby_xcalloc(max_frames, sizeof(bool));
-  buffer->locations     = ruby_xcalloc(max_frames, sizeof(ddog_prof_Location));
 
   return buffer;
 }
@@ -410,10 +411,10 @@ sampling_buffer *sampling_buffer_new(uint16_t max_frames) {
 void sampling_buffer_free(sampling_buffer *buffer) {
   if (buffer == NULL) rb_raise(rb_eArgError, "sampling_buffer_free called with NULL buffer");
 
+  // buffer->locations are owned by whoever called sampling_buffer_new, not us
   ruby_xfree(buffer->stack_buffer);
   ruby_xfree(buffer->lines_buffer);
   ruby_xfree(buffer->is_ruby_frame);
-  ruby_xfree(buffer->locations);
 
   ruby_xfree(buffer);
 }
--- a/ext/datadog_profiling_native_extension/collectors_stack.c
+++ b/ext/datadog_profiling_native_extension/collectors_stack.c
@@ -161,6 +161,14 @@ void sample_thread(
     return;
   }
 
+  // if (captured_frames > 0) {
+  //   int cache_hits = 0;
+  //   for (int i = 0; i < captured_frames; i++) {
+  //     if (buffer->stack_buffer[i].same_frame) cache_hits++;
+  //   }
+  //   fprintf(stderr, "Sampling cache hits: %f\n", ((double) cache_hits / captured_frames) * 100);
+  // }
+
   // Ruby does not give us path and line number for methods implemented using native code.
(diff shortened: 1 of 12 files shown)


ext/datadog_profiling_native_extension/collectors_stack.c
```diff
@@ -370,47 +378,48 @@ static void record_placeholder_stack_in_native_code(
...
   sampling_buffer* buffer = ruby_xcalloc(1, sizeof(sampling_buffer));
 
   buffer->max_frames = max_frames;
-
-  buffer->stack_buffer  = ruby_xcalloc(max_frames, sizeof(VALUE));
-  buffer->lines_buffer  = ruby_xcalloc(max_frames, sizeof(int));
-  buffer->is_ruby_frame = ruby_xcalloc(max_frames, sizeof(bool));
-  buffer->locations     = ruby_xcalloc(max_frames, sizeof(ddog_prof_Location));
+  buffer->locations = locations;
+  buffer->stack_buffer = ruby_xcalloc(max_frames, sizeof(frame_info));
```
marcotc, line 413: Nice consolidation here!

ext/datadog_profiling_native_extension/private_vm_api_access.c
```diff
@@ -479,164 +480,79 @@ int ddtrace_rb_profile_frames(VALUE thread, int start, int limit, VALUE *buff, i
...
+            // In practice, they are usually the same; the difference is that when you have e.g. block, one gets you a
+            // reference to the block, and the other to the method containing the block.
+            // This would be important if we used `rb_profile_frame_label` and wanted the "block in foo" label instead
+            // of just "foo". But we're currently using `rb_profile_frame_base_label` which I believe is always the same
+            // between the rb_callable_method_entry_t and the iseq. Thus, to simplify a bit our logic and reduce a bit
+            // the overhead, we always use the iseq here.
+            //
+            // cme = rb_vm_frame_method_entry(cfp);
+
+            // if (cme && cme->def->type == VM_METHOD_TYPE_ISEQ &&
```
marcotc, line 504: Can you add a comment here explicitly explaining what this commented out code section means (so we know when to restore it if needed, or remove it)

ext/libdatadog_api/extconf.rb
```diff
@@ -67,7 +67,7 @@ def skip_building_extension!(reason)
 
 if ENV['DDTRACE_DEBUG'] == 'true'
   $defs << '-DDD_DEBUG'
-  CONFIG['optflags'] = '-O0'
```
marcotc, line 70: Is `O0` not necessary to get the debug information we need?

marcotc (approved):
Looks great overall!

ext/libdatadog_api/extconf.rb
ivoanjo: I've used "-O1" with gdb and didn't see any issue/missing info, so I think it's fine, and it's easy to tweak locally -- I just wanted to avoid a default config that was broken (at least on my Ubuntu/gcc version?).

ext/datadog_profiling_native_extension/private_vm_api_access.c
ivoanjo: Good idea, left note in e6c912258c4e24e540ea23fea95545752126ecc1

ext/datadog_profiling_native_extension/private_vm_api_access.c
```diff
@@ -479,6 +480,16 @@ int ddtrace_rb_profile_frames(VALUE thread, int start, int limit, frame_info *st
                 continue;
             }
```
AlexJF, line 481: So in the past, the same cfp would eat all the "start" whereas now it'll skip just one? Not sure if this was ever something we relied on but just highlighting this slight change in behaviour.

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -1027,6 +1027,17 @@ def sample_allocation(obj)
...
 
+        # @ivoanjo: For some weird reason, the last object sampled in the "dead_heap_samples" does not always
+        # get collected the first time, leading to a flaky spec.
+        # I was able to reproduce it with `rspec spec/datadog/profiling --seed 48141 --fail-fast` and it's
+        # kinda bizarre since e.g. if you add one more `Object.new` it also stops flaking, so is it perhaps
+        # related to Ruby's conservative GC?
+        # I've bisected this and undoing 3d4b7fcf30b529b191ca737ae13629eb27b8ab63 also makes the flakiness
+        # go away, but again, that change doesn't seem to have anything to do with GC.
+        # As the weird behavior is transitory, e.g. it provably goes away on the next GC, I'll go with this
+        # workaround for now.
```
AlexJF, line 1038: Yeah I seem to remember Ruby taking into account register values in its heuristic to detect usages of objects. May just be that with the extra caching of things like iseq and what not we make it slightly more likely to hit these edge cases?

AlexJF (approved):

ext/datadog_profiling_native_extension/private_vm_api_access.h
```diff
@@ -18,6 +18,19 @@ typedef struct {
...
+typedef struct frame_info {
+  union {
+    struct {
+      VALUE iseq;
+      int line;
+    } ruby_frame;
+    struct {
+      ID method_id;
+    } native_frame;
+  } as;
```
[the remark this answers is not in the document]
ivoanjo: Something something [great artists copy](https://github.com/search?q=repo%3Aruby%2Fruby+%22%7D+as%3B%22&type=code).

ext/datadog_profiling_native_extension/private_vm_api_access.c
ivoanjo: This is a really great catch!

I completely glossed over it because a) we don't use `start` and b) I remembered some discussion somewhere mentioning it was actually broken.

It's actually fixed nowadays https://bugs.ruby-lang.org/issues/14607 // https://github.com/ruby/ruby/pull/8280 . I had spotted that upstream had moved the `cfp = ...` to the for header as well, but I hadn't spotted it was to fix `start`.

I did spot that upstream also added the start skip logic to the `if (cme && cme->def->type == VM_METHOD_TYPE_CFUNC)` branch below. I'll push a commit to add this one as well, so we keep closer to upstream in case we ever want to use `start`.

spec/datadog/profiling/stack_recorder_spec.rb
ivoanjo: Yeah this is my suspicion -- Ruby's GC is conservative, in that if anything "looks" like a `VALUE`, it assumes it is. Actually debugging this would mean going into the GC marking code and understand why exactly it's making this decision which is somewhat of a time investment which is why I took the slightly-easier way out 😅.

PR #3839, opened 2024-08-09
 Emit log message for incompatible Lograge setup 

### LINKED ISSUE #2012: Lograge in JSON mode in Rails causes double addition of log correlation information and breaks log parsing
The current implementation of the Rails integration causes log correlation information to be added twice to the logger when using Lograge.  This breaks log parsing in Datadog if Lograge is using the JSON formatter.

I'm currently attempting to work around this by setting `Rails.configuration.log_tags = []` in `initializers/datadog.rb`.

As far as I can tell, this is likely going to be a problem in any Rails project that doesn't override the default logger when using Lograge (which doesn't modify the default logger) and using the JSON formatter.

I'm not sure if it's possible to simply delay modifying `Rails.configuration.log_tags` until `after_initialize` and still have everything function correctly.  If nothing else, a mention in the README about this potential issue would save other people the head-scratching.

marcotc:
Closes #2012

This PR emits an error-level log message when the `lograge` gem is used with Rails tagged logging, which is configured by default in Rails since Rails 5.

We can't practically fix the issue here in `dd-trace-rb`, since the issue is a log formatting conflict between `lograge` and `Rails`.
The reason this affects APM Tracing is because it breaks trace-log correlation.

The only alternatives are to either: disable Rails tagged logging; or not use `lograge`.

The investigation and suggest alternatives are well documented here: https://github.com/DataDog/dd-trace-rb/issues/2012#issuecomment-2038438494

### DIFF
--- a/lib/datadog/tracing/contrib/lograge/patcher.rb
+++ b/lib/datadog/tracing/contrib/lograge/patcher.rb
@@ -20,6 +20,22 @@ module Datadog
 
           # patch applies our patch
           def patch
+            # ActiveSupport::TaggedLogging is the default Rails logger since Rails 5
+            if defined?(::ActiveSupport::TaggedLogging::Formatter) &&
+                ::Lograge::LogSubscribers::ActionController
+                    .logger&.formatter.is_a?(::ActiveSupport::TaggedLogging::Formatter)
+
+              Datadog.logger.error(
+                'Lograge and ActiveSupport::TaggedLogging (the default Rails log formatter) are not compatible: ' \
+                  'Lograge does not account for Rails log tags, creating polluted logs and breaking log formatting. ' \
+                  'Traces and Logs correlation may not work. ' \
+                  'Either: 1. Disable tagged logging in your Rails configuration ' \
+                  '`config.logger = ActiveSupport::Logger.new(STDOUT); ' \
+                  'config.active_job.logger = ActiveSupport::Logger.new(STDOUT)` ' \
+                  'or 2. Use the `semantic_logger` gem instead of `lograge`.'
+              )
+            end
+
             ::Lograge::LogSubscribers::Base.include(Instrumentation)
           end
         end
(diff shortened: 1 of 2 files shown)


TonyCTHsu (approved):

armstrjare:
We've started getting this error message since this update.

However, we already account for this with customisations to the log formatter to output in JSON *and* include Datadog tags:
Part of our implementation for this is made public here: https://github.com/armstrjare/rails-json-tagged-logging

Can we add a way to disable this warning to stop polluting our startup logs?

@marcotc @TonyCTHsu

alexevanczuk:
Hey @marcotc just wanted to +1 for a way to disable this warning message.

I'm also confused about the exact behavior that's broken. We set up some log remapping that ensures the datadog trace IDs are properly read as trace IDs and we still have log correlation working. Is there anything else that might be missing with `lograge` that I'm not seeing?

Thank you!

marcotc:
Hey @armstrjare and @alexevanczuk, thank you for reporting this issue. It makes total sense.

I opened #4268 so we can track the work of adding an option to disable this log message.

alexevanczuk:
Thanks @marcotc !! Also let me know if there's anything that lograge makes not possible besides log correlation (which we 'roll ourselves' by putting the trace ID into the lograge logged message)! 

Thanks again!

tcannonfodder:
> We've started getting this error message since this update.
> 
> However, we already account for this with customisations to the log formatter to output in JSON _and_ include Datadog tags: Part of our implementation for this is made public here: https://github.com/armstrjare/rails-json-tagged-logging
> 
> Can we add a way to disable this warning to stop polluting our startup logs?
> 
> @marcotc @TonyCTHsu

**Warning**: If upgrading from Rails 7.1 => 7.2 (even if you have a working JSON tagged logging setup using `lograge` + `rails-json-tagged-logging`), **your logging setup will break, which seems to also cause a memory leak.** I wasn't able to debug exactly what broke; I instead switched everything over to `rails_semantic_logger`

armstrjare:
> **Warning**: If upgrading from Rails 7.1 => 7.2 (even if you have a working JSON tagged logging setup using `lograge` + `rails-json-tagged-logging`), **your logging setup will break, which seems to also cause a memory leak.** I wasn't able to debug exactly what broke; I instead switched everything over to `rails_semantic_logger`

@tcannonfodder I’m still using this in production without issue with a decent request volume - and also was on 7.2 when it was released. Happy to help if you throw an issue over at my repo, but sounds like you were happy to just move to semantic logger anyway 👍

PR #3844, opened 2024-08-14
[NO-TICKET] Apply more of standardrb autoformatter to profiler files

ivoanjo:
**What does this PR do?**

In #3801 we adopted standardrb for the profiler. At the time, I left a few cops disabled:

> This is because fixing each of these cops will create a LOT of diff noise. Thus, I've explicitly not included them in this PR, and plan to fix them separately, in smaller PRs, so we can still sanely review the diff on this PR.

This is the first of those smaller PRs that can be reviewed sanely.

**Motivation:**

Fully adopt standardrb for profiling.

**Additional Notes:**

The autofix to remove begin/end creates a lot of diff noise. I recommend reviewing this PR without whitespace as it's easier to spot the change.

**How to test the change?**

This PR does not change any behaviors, so existing test coverage being green is the thing to aim for :)


AlexJF (approved):
LGTM! The quote changes in `cpu_and_wall_time_worker_spec.rb` look partially applied (there are still `'` around) and unrelated to the non-ignored rules but since it already adheres to the style we plan to introduce in a subsequent PR I vote we let it be 😄

ivoanjo:
> LGTM! The quote changes in cpu_and_wall_time_worker_spec.rb look partially applied (there are still ' around) and unrelated to the non-ignored rules but since it already adheres to the style we plan to introduce in a subsequent PR I vote we let it be 😄

Yes: The autofix was only for quotes in symbols not strings, I'll open up the bigger "make all strings consistent" next :)

PR #3845, opened 2024-08-14
[NO-TICKET] Finish autoformatting profiler with standardrb

ivoanjo:
**What does this PR do?**

In #3801 we adopted standardrb for the profiler. At the time, I left a few cops disabled:

> This is because fixing each of these cops will create a LOT of diff noise. Thus, I've explicitly not included them in this PR, and plan to fix them separately, in smaller PRs, so we can still sanely review the diff on this PR.

This PR contains one of the most noisiest changes, which is using `"` for strings instead of `'`.

After this PR lands, all of the profiler is matching standardrb's expected style (except trailing commas, see discussion on #3801).

**Motivation:**

Fully adopt standardrb for profiling.

**Additional Notes:**

N/A

**How to test the change?**

This PR does not change any behaviors, so existing test coverage being green is the thing to aim for :)


AlexJF (approved):
Didn't spot any sneaky cryptomining code so LGTM!

AlexJF:
> Found 0 performance improvements and 2 performance regressions! Performance is the same for 21 metrics, 2 unstable metrics.
> scenario:profiler - hold / resume
> 🟥 throughput [-73367.867op/s; -71529.071op/s] or [-3.510%; -3.422%]

Oh this won't do @ivoanjo can we switch back to single quotes? :trollface:

ivoanjo:
> > Found 0 performance improvements and 2 performance regressions! Performance is the same for 21 metrics, 2 unstable metrics.
> > scenario:profiler - hold / resume
> > 🟥 throughput [-73367.867op/s; -71529.071op/s] or [-3.510%; -3.422%]

> Oh this won't do @ivoanjo can we switch back to single quotes? :trollface: 

Oh no, single quotes are faster confirmed! Fire up the youtube channels!

Running the profiling benchmarks for PRs is a really new change (#3810) so I haven't had time to look into it, but seems that some of the benchmarks are not very stable.

Maybe we just need to run them for longer, or maybe they need some tweaking to not provide false-positives? :thinking:

PR #3849, opened 2024-08-19
Add http.route tag to rails, grape, and sinatra integrations

y9v:
**What does this PR do?**
This PR adds `http.route` tag to rack spans for Rails (including Rails Engines), Grape, and Sinatra applications.

**Motivation:**
Having `http.route` tag on spans helps grouping requests.

**Additional Notes:**
In Rails version 7.1 internal `ActionDispatch::Journey::Router#find_routes` method [was changed](https://github.com/rails/rails/commit/35b280fcc2d5d474f9f2be3aca3ae7aa6bba66eb), which was the reason why the [previous implementation](https://github.com/DataDog/dd-trace-rb/pull/3345) had to be [rolled back](https://github.com/DataDog/dd-trace-rb/pull/3539).

This PR fixes this issue by checking action_pack version, and either patching `ActionDispatch::Journey::Router#find_routes` for Rails versions prior to 7.1, or patching `ActionDispatch::Journey::Router#serve` and using `#route_uri_pattern` that gets set on the request object.

**How to test the change?**
Screenshots from local testing:
Screenshot 2024-08-19 at 16 24 10


y9v:
deleted bunch of duplicated comments from the static code analyzer. [This commit](https://github.com/DataDog/dd-trace-rb/pull/3849/commits/234cfe3d16dde297e47d0a667ee495fab4aef7eb) fixes the warning

y9v:
I will add Rails 7.2 to system tests repo to increase coverage.

spec/datadog/tracing/contrib/rails/support/rails7.rb
```diff
@@ -0,0 +1,213 @@
...
+
+    # TODO: Remove this side-effect on missing log entries
+    Lograge.remove_existing_log_subscriptions if defined?(::Lograge)
+
+    reset_class_variable(ActiveRecord::Railtie::Configuration, :@@options) if Module.const_defined?(:ActiveRecord)
+
+    ActiveSupport::Dependencies.autoload_paths = []
+    ActiveSupport::Dependencies.autoload_once_paths = []
+    ActiveSupport::Dependencies._eager_load_paths = Set.new
+    ActiveSupport::Dependencies._autoloaded_tracked_classes = Set.new
```
vpellan, line 191: Should we also reset this class variable in rails6.rb ?

vpellan (approved):
LGTM !

spec/datadog/tracing/contrib/rails/support/rails7.rb
y9v: yes, we can added, I just wanted to avoid unrelated changes in this PR, since it's pretty big already

y9v: and I was actually wondering why tests were not failing before without this file

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
```diff
@@ -0,0 +1,88 @@
...
+                    # This array is [match_data, path_parameters, route].
+                    # We need the route object, since it has a path with route specification.
+                    current_route = result.last&.last&.path&.spec
+                    return result unless current_route
+
+                    # When Rails is serving requests to Rails Engine routes, this function is called
+                    # twice: first time for the route on which the engine is mounted, and second
+                    # time for the internal engine route.
+                    last_route = Tracing.active_span&.get_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE)
+                    Instrumentation.set_http_route_tag(last_route.to_s + current_route.to_s)
```
ivoanjo, line 47: Is it still intended/ok to call `set_http_route_tag` when `active_span` is nil?

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
```diff
@@ -0,0 +1,88 @@
...
+  module Tracing
+    module Contrib
+      module ActionPack
+        module ActionDispatch
+          # Instrumentation for ActionDispatch components
+          module Instrumentation
+            module_function
+
+            def set_http_route_tag(http_route)
+              return if http_route.nil?
```
ivoanjo, line 15: Most callers of `set_http_route_tag` seem to call it with a string, so I'm not sure `http_route` will ever be nil. Should this check perhaps be for an empty string, rather than nil?

ivoanjo (commented):
Left a few notes!

It wasn't quite clear to me -- do we already have testing for both Rails 7.0 and Rails 7.1?

Also, given their previous work, it may be worth pinging @zarirhamza or @lloeki for a review?

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
y9v: good point, I will add an early return instead of a safe operator

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
y9v: yes, I think it will be better

y9v:
@ivoanjo thank you for the review!

Regarding Rails 7 tests: we don't have them yet ([except system tests for Rails 7.1](https://github.com/DataDog/system-tests/commit/9e0e018c9502ab1598b6c057fb8f7ab8e37d1d38). The reason for adding rails 7 spec helper was that we are testing `action_pack` with [ActiveSupport version 7.0.8](https://github.com/DataDog/dd-trace-rb/blob/master/gemfiles/ruby_3.3_activesupport.gemfile.lock#L23-L25) and I needed a test rails application to test the router within a rack span.

The fact that we are not testing against version 7.1 and 7.2 yet is the reason why the test coverage is low for this patch.

spec/datadog/tracing/contrib/rails/support/rails7.rb
y9v: apparently this is a new property in rails 7: https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/15860/workflows/e14401f3-8207-4685-ac14-804236cecf0f/jobs/575944

lib/datadog/tracing/contrib/action_pack/action_dispatch/patcher.rb
```diff
@@ -0,0 +1,33 @@
...
+            include Contrib::Patcher
+
+            module_function
+
+            def target_version
+              Integration.version
+            end
+
+            def patch
+              if ::ActionPack.gem_version >= Gem::Version.new(7.1)
```
p-datadog, line 22: Would `'7.1'` not be more correct usage?

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
```diff
@@ -0,0 +1,88 @@
...
+
+              active_span = Tracing.active_span
+              return unless active_span
+
+              active_span.set_tag(
+                Tracing::Metadata::Ext::HTTP::TAG_ROUTE,
+                http_route.gsub(/\(.:format\)$/, '')
```
p-datadog, line 22: In Ruby the end of string is supposed to be `\z`, `$` means end of line.

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
```diff
@@ -0,0 +1,88 @@
...
+
+            # Instrumentation for ActionDispatch::Journey components
+            module Journey
+              # Instrumentation for ActionDispatch::Journey::Router
+              # for Rails versions older than 7.1
+              module Router
+                def find_routes(req)
+                  result = super(req)
+
+                  return response unless Tracing.enabled?
```
p-datadog, line 34: `return result` ?

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
```diff
@@ -0,0 +1,88 @@
...
+              )
+            end
+
+            # Instrumentation for ActionDispatch::Journey components
+            module Journey
+              # Instrumentation for ActionDispatch::Journey::Router
+              # for Rails versions older than 7.1
+              module Router
+                def find_routes(req)
+                  result = super(req)
```
p-datadog, line 32: This should be equivalent to the no-argument `super`, i.e. `result = super` (which looks like the usage you have in the other path).

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
p-datadog, line 47: Is this concatenating two routes or parts of the same route? If two routes, don't you need some kind of a separator?

lib/datadog/tracing/contrib/sinatra/tracer_middleware.rb
```diff
@@ -65,6 +66,7 @@ def call(env)
                 # since the latter is unaware of what the resource might be
                 # and would fallback to a generic resource name when unset
                 rack_request_span.resource ||= span.resource if rack_request_span
+                rack_request_span.set_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE, datadog_route) if datadog_route
```
p-datadog, line 69: Line 53 above writes `datadog_route` into `TAG_ROUTE_PATH`, are both of the keys supposed to be set?

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
p-datadog: And it seems like there perhaps isn't test coverage for this path.

y9v: thank you for finding this! Yes, there is none, I am working to add appraisals for different action pack versions. I am a bit afraid to merge such changes without coverage (especially after this find).

lib/datadog/tracing/contrib/sinatra/tracer_middleware.rb
y9v: tbh I don't know, `TAG_ROUTE_PATH` [sets](https://github.com/DataDog/dd-trace-rb/blob/b7b99b2c72ca97cb42b7c8b72accb5571e8c4f3f/lib/datadog/tracing/contrib/sinatra/ext.rb#L23) `sinatra.route.path`.

I have the same situation with Grape, it also [sets](https://github.com/DataDog/dd-trace-rb/blob/b7b99b2c72ca97cb42b7c8b72accb5571e8c4f3f/lib/datadog/tracing/contrib/grape/ext.rb#L24) `grape.route.path`, and I added setting of `http.route` to it as well.

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
y9v: In my local testing, all the routes had a `/` in the beginning (both rails routes and rails engine routes). I could not find a case where there is no slash at the beginning of the route.

lib/datadog/tracing/contrib/action_pack/action_dispatch/patcher.rb
y9v: you are right, it is supposed to work with a string

lib/datadog/tracing/contrib/sinatra/tracer_middleware.rb
y9v: I am not entirely sure, but I think we need `http.route` to be set for some appsec features, so probably it is worth adding it.

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
y9v: Added appraisals for different action_pack versions here [59b5290](https://github.com/DataDog/dd-trace-rb/pull/3849/commits/59b52908dfe533ff6d93e80270b7eba709e4413f)

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
y9v: These are 2 parts of the same route: one for the route on which the rails engine is mounted, and the second one is the internal rails engine route.

```ruby
  namespace :api do
    mount AuthEngine::Engine => "/auth"
  end
```

In this example we will have 2 parts:  `/api/auth` and `/sign-in(/:expires_at)`.

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
y9v: I removed this method for setting the route and replaced it with a simple method for route formatting here https://github.com/DataDog/dd-trace-rb/pull/3849/commits/137d110e3e90313c4d67554905d6a48893027e9f

Matrixfile
```diff
@@ -27,7 +27,10 @@
     'opentelemetry' => '❌ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ❌ jruby'
   },
   'action_pack' => {
-    'activesupport' => '✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ jruby',
+    'actionpack-5.0' => '✅ 2.5 / ✅ 2.6 / ❌ 2.7 / ❌ 3.0 / ❌ 3.1 / ❌ 3.2 / ❌ 3.3 / ❌ 3.4 / ✅ jruby',
+    'actionpack-6.0' => '❌ 2.5 / ❌ 2.6 / ✅ 2.7 / ❌ 3.0 / ❌ 3.1 / ❌ 3.2 / ❌ 3.3 / ❌ 3.4 / ✅ jruby',
+    'actionpack-7.0' => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ❌ jruby',
+    'actionpack-7.1' => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ❌ jruby',
```
ivoanjo, line 33: As you might've noticed, each separate `appraise` that gets added results in an explosion of files to maintain/update.

Since the main dependency needed for these tests is rails, have you considered reusing some of the existing appraisal groups when possible?

E.g. For instance using `rails5-mysql2` / `rails6-mysql2` and then adding `rails7` and `rails71`?

(Our whole setup for appraisals does help us get coverage but it has a lot of sharp edges and most of us are dreaming of something better lol)

y9v: yes, adding appraisals for action_pack was far from being a perfect solution :)

As for reusing existing appraisals - I was not sure if I could just add `rails7` or something similar, since all of our existing rails appraisals are bundled together with a database adapter. How do you think, does it makes sense to add appraisals for bare-bones rails?

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
y9v: fixed

lib/datadog/tracing/contrib/action_pack/action_dispatch/patcher.rb
y9v: fixed

Matrixfile
y9v: removed `actionpack` appraisal and added appraisals for `rails7` and `rails71` here [038afb6](https://github.com/DataDog/dd-trace-rb/pull/3849/commits/038afb6bc712d8cc9036a0a6f96597a2b9b46f5f)

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
y9v: In the codebase rails also [uses a term "route" for those](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/journey/router.rb#L126). I think the problem here is that if you have a rails engine mounted at some route, this `#find_routes` will be called twice: once for the path on which the engine is mounted, and the second time for the internal engine route.

So we only have 2 "routes" here when using rails engines.

What do you think, should I then just keep the term "route" here, or rename it to "route fragment" so that the code is more readable?

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
y9v: Appraisals are fixed in this comment: https://github.com/DataDog/dd-trace-rb/pull/3849/commits/5ba0feff11e20c684148c18eb103aaa41d79b2dc

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
```diff
@@ -0,0 +1,93 @@
...
+                  result
+                end
+              end
+
+              # Since Rails 7.1 `Router#serve` adds `#route_uri_pattern` attribute to the request,
+              # and the `Router#find_routes` now takes a block as an argument to make the route computation lazy
+              # https://github.com/rails/rails/commit/35b280fcc2d5d474f9f2be3aca3ae7aa6bba66eb
+              module LazyRouter
+                def serve(req)
+                  response = super
```
lloeki, line 60: This makes the route resolved to *after* the request is processed by whatever is behind `super`, which means:

- route information is not available to other parts until after the fact
- for ASM it is not blockable based on route information (or rather it could block but  way too late when effects have been applied, possibly with disastrous consequences).

Here's the patch I had for Rails 7.1+ that gets the information at exactly the right time by hooking into the block†:

```diff
commit 1bc8c3fd94bd32abe79c0c29d9920bc193c7d9f2
Merge: 14b02858e f92f6b430
Author: Loic Nageleisen <loic.nageleisen@datadoghq.com>
Date:   Tue Mar 19 16:08:42 2024 +0100

    WIP on patch-sinatra-once: 14b02858e Move configuration of Rack integration within initial Datadog.configure

diff --cc lib/datadog/tracing/contrib/rails/patcher.rb
index e2d147cd0,e2d147cd0..07d4b4a7f
--- a/lib/datadog/tracing/contrib/rails/patcher.rb
+++ b/lib/datadog/tracing/contrib/rails/patcher.rb
@@@ -13,13 -13,13 +13,34 @@@ module Datado
        module Rails
          # Patcher to trace rails routing done by JourneyRouter
          module JourneyRouterPatch
--          def find_routes(*args)
--            result = super
--            integration_route = result.first[2].path.spec.to_s.gsub(/\(\.:format\)/, '') if result.any?
--            request_trace = Tracing.active_trace || TraceOperation.new
--            request_trace.set_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE, integration_route)
--            request_trace.set_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE_PATH, args.first.env['SCRIPT_NAME'])
--            result
++          def serve(req)
++            r = super
++          ensure
++            binding.irb
++          end
++
++          def find_routes(req, &block)
++            binding.irb
++            result = []
++            ret = super(req) do |match, parameters, route|
++              binding.irb
++              result << [match, parameters, route]
++              block.call(match, parameters, route)
++            ensure
++              binding.irb
++              p ret.nil?
++              p result.any?
++              p ret.nil? && result.any?
++              if ret.nil? && result.any?
++                binding.irb
++                integration_route = result.last[2].path.spec.to_s.gsub(/\(\.:format\)/, '')
++                request_trace = Tracing.active_trace || TraceOperation.new
++                request_trace.set_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE, integration_route)
++                request_trace.set_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE_PATH, req.env['SCRIPT_NAME'])
++              end
++            end
++          ensure
++            binding.irb
            end
          end
  
```

Note that the previous implementation which supported rails 7.0 and down worked out of a lucky `result.first` guess that is _usually_ correct (there's only one result) but not _guaranteed_ correct as multiple routes could theoretically be matched (the return value is an `Array`/`Enumerable`).

It would be interesting to have a test case where multiple routes are matched by the route resolver (IIRC in that case Rails takes the first one) to ensure that we have a correct understanding of this case + that our implementation is correct in this case as well.

† This kind of hooking will be made easier by https://github.com/DataDog/graft-rb including guaranteed ordering between `Tracing` and `AppSec`.

lloeki (changes requested):

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
y9v: Thank you for the review!

I changed patching of rails 7.1 [079872b](https://github.com/DataDog/dd-trace-rb/pull/3849/commits/079872b319334a9804ee42edb40318dd5ef11247) according to your suggestion.

And I also added [d5bbc94](https://github.com/DataDog/dd-trace-rb/pull/3849/commits/d5bbc947e8cad116c52cb4f94c62408e7cf64674) as you suggested.

y9v: and thank you again for mentioning the importance of handling the route and script name separately, this helped to correctly handle nested Grape application routes

spec/datadog/tracing/contrib/rack/http_route_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  before do
+    Datadog.configure do |c|
+      c.tracing.instrument :rack
+    end
+  end
+
+  after do
+    Datadog.configuration.tracing[:rack].reset!
+  end
```
Strech, line 19: Not sure that I have all the context, but maybe it makes sense to wrap the configuration change (which is shared) in the `ensure` section to be sure that tests are not leaking in case of the exception?

Sorry if it's not relevant, but just to be sure

```suggestion
  around do |example|
    Datadog.configure { |c| c.tracing.instrument :rack }
    example.run
  ensure
    Datadog.configuration.tracing[:rack].reset!
  end
```

spec/datadog/tracing/contrib/rails/support/rails7.rb
```diff
@@ -0,0 +1,213 @@
...
+  end
+
+  after do
+    reset_rails_configuration!
+
+    # Push this to base when Rails 3 removed
+    # Reset references stored in the Rails class
+    Rails.app_class = nil
+    Rails.cache = nil
+  end
```
Strech, line 148: I think same idea applied here regarding safe-guarding the reset.

Strech (commented):
It's just my small input regarding some testing practices I think matter, but could be irrelevant

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
```diff
@@ -0,0 +1,80 @@
...
+
+            def set_http_route_tags(route_spec, script_name)
+              return unless Tracing.enabled?
+
+              return unless route_spec
+
+              request_trace = Tracing.active_trace
+              return unless request_trace
+
+              request_trace.set_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE, route_spec.to_s.gsub(/\(.:format\)\z/, ''))
```
p-datadog, line 22: Maybe add a note here why the format is removed from route? Because there can be other parameters in the rest of the route?

lib/datadog/tracing/contrib/grape/endpoint.rb
```diff
@@ -61,6 +62,15 @@ def endpoint_start_process(_name, _start, _finish, _id, payload)
               span.set_tag(Tracing::Metadata::Ext::TAG_COMPONENT, Ext::TAG_COMPONENT)
               span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_ENDPOINT_RUN)
 
+              if (grape_route = env['grape.routing_args'] && env['grape.routing_args'][:route_info])
+                trace.set_tag(
+                  Tracing::Metadata::Ext::HTTP::TAG_ROUTE,
+                  grape_route.path&.gsub(/\(\.{1}:?\w+\)\z/, '')
```
p-datadog, line 68: `{1}` is redundant, no?

lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -138,6 +138,29 @@ def set_request_tags!(trace, request_span, env, status, headers, response, origi
             request_span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_REQUEST)
             request_span.set_tag(Tracing::Metadata::Ext::TAG_KIND, Tracing::Metadata::Ext::SpanKind::TAG_SERVER)
 
+            if status != 404 && (last_route = trace.get_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE))
+              last_script_name = trace.get_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE_PATH).to_s
```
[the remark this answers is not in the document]
p-datadog: If you are coercing `nil` to `''` I would perform that in the comparison below (on line 150), i.e. `last_script_name || '' == ''` which I think conveys the intent clearer. Or even default that here instead of using `to_s`.

spec/datadog/tracing/contrib/action_pack/action_dispatch/journey/router_spec.rb
```diff
@@ -0,0 +1,233 @@
...
+      end
+
+      it 'sets http.route and http.route.path for rails engine routes' do
+        get '/api/auth/sign-in'
+
+        request_span = spans.first
+
+        expect(last_response).to be_ok
+        expect(request_span.name).to eq('rack.request')
+        expect(request_span.tags.fetch('http.route')).to eq('/api/auth/sign-in(/:expires_in)')
```
p-datadog, line 190: This looks to test an optional route parameter which is a great test to have, I would suggest explicitly declaring at least one of the tests as testing the optional parameter so that wouldn't be accidentally removed in the future.

lib/datadog/tracing/contrib/grape/endpoint.rb
y9v: yes, thanks!

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
y9v: I am actually not sure whether we want to remove it. A fully qualified rails route will have `(.:format)` at the end. The reason why I added this formatting is because it was in the [first attempt](https://github.com/DataDog/dd-trace-rb/pull/3345) to implement this feature.

p-datadog: I think this decision is not mine to make therefore I will resolve this comment.

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
```diff
@@ -0,0 +1,80 @@
...
+
+              if script_name && !script_name.empty?
+                request_trace.set_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE_PATH, script_name)
+              end
+            rescue StandardError => e
```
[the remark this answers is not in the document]
y9v: I think you are right, I have checks here to ensure that the active trace is present, hard to imagine that method like `#set_tag` will raise something

spec/datadog/tracing/contrib/action_pack/action_dispatch/journey/router_spec.rb
y9v: fixed [3207b16](https://github.com/DataDog/dd-trace-rb/pull/3849/commits/3207b16ea39d9a5b7af0d7a1ae392058d2c39656)

lib/datadog/tracing/contrib/rack/middlewares.rb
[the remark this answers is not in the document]
y9v: fixed [3207b16](https://github.com/DataDog/dd-trace-rb/pull/3849/commits/3207b16ea39d9a5b7af0d7a1ae392058d2c39656)

lib/datadog/tracing/contrib/grape/endpoint.rb
y9v: fixed [3207b16](https://github.com/DataDog/dd-trace-rb/pull/3849/commits/3207b16ea39d9a5b7af0d7a1ae392058d2c39656)

lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb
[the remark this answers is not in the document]
y9v: fixed [3207b16](https://github.com/DataDog/dd-trace-rb/pull/3849/commits/3207b16ea39d9a5b7af0d7a1ae392058d2c39656)

spec/datadog/tracing/contrib/rack/http_route_spec.rb
y9v: fixed [cae659f](https://github.com/DataDog/dd-trace-rb/pull/3849/commits/cae659f4f62b2b976356afe212ca899fcba3041c)

spec/datadog/tracing/contrib/action_pack/action_dispatch/journey/router_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+
+    let(:users_controller) do
+      stub_const(
+        'Api::UsersController',
+        Class.new(ActionController::Base) do
+          def show
+            head :ok
+          end
+        end
+      )
```
Strech, line 70: I'm still new, but is there a reason we don't use `class_double` here?

https://www.rubydoc.info/github/rspec/rspec-mocks/RSpec%2FMocks%2FExampleMethods:class_double

It will allow us to control stabbing of non-existent methods, but sometimes it could work bad when class has dynamic methods.

Strech (approved):
From my point of view (still lacking context) it's good (with a few questions)

spec/datadog/tracing/contrib/action_pack/action_dispatch/journey/router_spec.rb
y9v: I don't know, this is what we used in other places for creating temporary controller classes. I guess using `class_double` on `ActionController::Base` won't allow us stubbing action methods, since they don't exist on `ActionController::Base`?

lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -139,7 +139,7 @@ def set_request_tags!(trace, request_span, env, status, headers, response, origi
             request_span.set_tag(Tracing::Metadata::Ext::TAG_KIND, Tracing::Metadata::Ext::SpanKind::TAG_SERVER)
 
             if status != 404 && (last_route = trace.get_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE))
-              last_script_name = trace.get_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE_PATH).to_s
+              last_script_name = trace.get_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE_PATH).to_s || ''
```
p-datadog, line 142: to_s will always be truthy I think so the `|| ''` wouldn't do anything?

lib/datadog/tracing/contrib/rack/middlewares.rb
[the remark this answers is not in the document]
p-datadog: Looking at this again the .to_s should already convert to a non-nil value so the defaulting shouldn't be needed?

p-datadog (approved):

lib/datadog/tracing/contrib/grape/endpoint.rb
```diff
@@ -62,6 +63,15 @@ def endpoint_start_process(_name, _start, _finish, _id, payload)
               span.set_tag(Tracing::Metadata::Ext::TAG_COMPONENT, Ext::TAG_COMPONENT)
               span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_ENDPOINT_RUN)
 
+              if (grape_route = env['grape.routing_args'] && env['grape.routing_args'][:route_info])
```
marcotc, line 66: Can you add parenthesis here to improve reading clarity in this `if` condition?
```suggestion
              if ((grape_route = env['grape.routing_args']) && grape_route[:route_info])
```

lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -138,6 +138,29 @@ def set_request_tags!(trace, request_span, env, status, headers, response, origi
...
+            if status != 404 && (last_route = trace.get_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE))
+              last_script_name = trace.get_tag(Tracing::Metadata::Ext::HTTP::TAG_ROUTE_PATH).to_s || ''
+
+              # If the last_script_name is empty but the env['SCRIPT_NAME'] is NOT empty
+              # then the current rack request was not routed and must be accounted for
+              # which only happens in pure nested rack requests i.e /rack/rack/hello/world
+              #
+              # To account for the unaccounted nested rack requests of /rack/hello/world,
+              # we use 'PATH_INFO knowing that rack cannot have named parameters
+              if last_script_name == '' && env['SCRIPT_NAME'] != ''
```
marcotc, line 150: Can `env['SCRIPT_NAME']` be `nil` here, and thus triggering `env['SCRIPT_NAME'] != ''` to be true.

spec/datadog/tracing/contrib/rails/support/rails7.rb
```diff
@@ -0,0 +1,213 @@
+# Loaded by the `bin/rails` script in a real Rails application
```
marcotc, line 1: Thank you so much for adding Rails 7 to the test matrix!

marcotc (approved):
I left a few minor comments, but it looks great overall!

spec/datadog/tracing/contrib/action_pack/action_dispatch/journey/router_spec.rb
Strech: Yes you can't stub non-existent methods, but you should not mock the `Base` class, but `UsersController` instead.

But I think now I see the reason, we will not be able to return `head(:ok)` response because we will not have the context in the RSpec file (we need to be "inside" the Base class)

Sorry for the noise 🥺

lib/datadog/tracing/contrib/rack/middlewares.rb
y9v: fixed: [7317018](https://github.com/DataDog/dd-trace-rb/pull/3849/commits/7317018fe09188035e5f7799603c0d46b5a95912)

lib/datadog/tracing/contrib/rack/middlewares.rb
y9v: yeah, I think it actually can. I will add handling for nil here too

lib/datadog/tracing/contrib/grape/endpoint.rb
```diff
@@ -62,6 +63,21 @@ def endpoint_start_process(_name, _start, _finish, _id, payload)
               span.set_tag(Tracing::Metadata::Ext::TAG_COMPONENT, Ext::TAG_COMPONENT)
               span.set_tag(Tracing::Metadata::Ext::TAG_OPERATION, Ext::TAG_OPERATION_ENDPOINT_RUN)
 
+              if (grape_route = env['grape.routing_args']) && grape_route[:route_info]
+                puts '=' * 80
+                puts grape_route[:route_info].path
+                puts '=' * 80
```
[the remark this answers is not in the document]
y9v: oops, thanks! [deaba02](https://github.com/DataDog/dd-trace-rb/pull/3849/commits/deaba021ee07434ab2bd99da250ed5ba12f8cf19)

lloeki (approved):

PR #3850, opened 2024-08-19
Implement telemetry log

TonyCTHsu:
**What does this PR do?**

This PR enables developer to send telemetry logs for internal observability.

### DIFF
--- a/lib/datadog/core/telemetry/event.rb
+++ b/lib/datadog/core/telemetry/event.rb
@@ -334,6 +334,33 @@ module Datadog
           end
         end
 
+        # Telemetry class for the 'logs' event
+        class Log < Base
+          LEVELS = {
+            error: 'ERROR',
+            debug: 'DEBUG',
+            warn: 'WARN',
+          }.freeze
+
+          def type
+            'logs'
+          end
+
+          def initialize(message:, level:)
+            super()
+            @message = message
+            @level = LEVELS.fetch(level) { |k| raise ArgumentError, "Invalid log level :#{k}"}
+          end
+
+          def payload
+            {
+              message: @message,
+              level: @level,
+              # More optional fields to be added here...
+            }
+          end
+        end
+
         # Telemetry class for the 'distributions' event
         class Distributions < GenerateMetrics
           def type
--- a/lib/datadog/core/telemetry/event.rb
+++ b/lib/datadog/core/telemetry/event.rb
@@ -349,7 +349,7 @@ module Datadog
           def initialize(message:, level:)
             super()
             @message = message
-            @level = LEVELS.fetch(level) { |k| raise ArgumentError, "Invalid log level :#{k}"}
+            @level = LEVELS.fetch(level) { |k| raise ArgumentError, "Invalid log level :#{k}" }
           end
 
           def payload
--- a/lib/datadog/core/telemetry/event.rb
+++ b/lib/datadog/core/telemetry/event.rb
@@ -354,9 +354,11 @@ module Datadog
 
           def payload
             {
-              message: @message,
-              level: @level,
-              # More optional fields to be added here...
+              logs: [{
+                message: @message,
+                level: @level,
+                # More optional fields to be added here...
+              }]
             }
           end
         end
--- a/lib/datadog/core/telemetry/logging.rb
+++ b/lib/datadog/core/telemetry/logging.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require_relative 'event'
+
+module Datadog
+  module Core
+    module Telemetry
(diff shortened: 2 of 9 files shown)


lib/datadog/core/telemetry/event.rb
```diff
@@ -334,6 +334,33 @@ def payload
...
+          }.freeze
+
+          def type
+            'logs'
+          end
+
+          def initialize(message:, level:)
+            super()
+            @message = message
+            @level = LEVELS.fetch(level)
```
y9v, line 352: What do you think, does it make sense to throw an `ArgumentError` here instead of a `KeyError`?

```ruby
LEVELS.fetch(level) { |k| raise ArgumentError, "Invalid log level :#{k}"}
```

lib/datadog/core/telemetry/logging.rb
```diff
@@ -0,0 +1,30 @@
...
+          event = Event::Log.new(
+            message: message,
+            level: level
+          )
+
+          if (telemetry = Datadog.send(:components).telemetry)
+            telemetry.log!(event)
+          else
+            Datadog.logger.debug do
+              "Sending telemetry log about #{message} before telemetry component is ready"
```
p-datadog, line 23: I find "sending" to be misleading because nothing is actually sent. Suggest either "attempting to send" or "requested to send" or similar. Also, I would put the message at the end so that it's clear where the diagnostics of this component ends and where the forwarded message is. E.g. "Requested to send telemetry log when telemetry component is not ready: #{message}". Is it possible for telemetry to be never turned on, i.e. it would actually not ever become ready?

marcotc (approved):
Looks good!
I believe @y9v's and @p-datadog's comments are relevant and should be responded to or addressed.

lib/datadog/core/telemetry/logging.rb
marcotc: The telemetry object is always created, even if telemetry is disabled https://github.com/DataDog/dd-trace-rb/blob/b7b99b2c72ca97cb42b7c8b72accb5571e8c4f3f/lib/datadog/core/telemetry/component.rb#L45-L47
so it will eventually become no-nil.

I also like that this message is at `debug` level, so people have to manually enable debug-level logs to see it.

I do agree that the messaging can be more precise.

ivoanjo: Minor: Should we perhaps print the whole `event` and not just the `message` when logging

spec/datadog/core/telemetry/component_spec.rb
```diff
@@ -219,6 +219,55 @@
     end
   end
 
+  describe '#log1' do
```
ivoanjo, line 222: Minor:

```suggestion
  describe '#log!' do
```

spec/datadog/core/telemetry/logging_spec.rb
```diff
@@ -0,0 +1,60 @@
+require 'spec_helper'
+
+require 'datadog/core/telemetry/logging'
+
+RSpec.describe Datadog::Core::Telemetry::Logging do
+  describe '.report' do
+    context 'with named exception' do
+      it 'sends a log event to via telemetry' do
+        telemetry = double('telemetry', log: nil)
```
ivoanjo, line 9: I recommend using an `instance_double` here; regular doubles are quite dangerous since there's nothing that ensures that the double matches the actual class being tested.

lib/datadog/core/telemetry/logging.rb
TonyCTHsu: I don't think printing the whole event is necessary, since the issue is not caused by event itself.

spec/datadog/core/telemetry/logging_spec.rb
TonyCTHsu: I am trying to NOT be religious about `instance_double` is safer with 'double'. 

Since they are both not the `Datadog::Core::Telemetry::Component` instance,  I would prefer a mock object with minimized footprint (I can also replaced it with an `OpenStruct`).

lib/datadog/core/telemetry/logging.rb
ivoanjo: My thinking for this suggestion is, if later we add more information to the event (e.g. stack), we'll need to remember to tweak the debug message, whereas if we print the whole event, then we won't need to touch it again.

spec/datadog/core/telemetry/logging_spec.rb
ivoanjo: I'm curious to know why you don't like `instance_double`; I've sent you a message, as I'd like to chat about this.

y9v (approved):

PR #3851, opened 2024-08-20
Fix confusion in the SpanFilter docstring

sulami:
**What does this PR do?**
The docstring was claiming that a truthy value from the filter block keeps the span, but in fact it's the other way around, a truthy value drops the span.

**Motivation:**
Found this by accident, and given that "truthy to filter" is already unintuitive enough, having the docs lie doesn't exactly help :D

**Additional Notes:**
Nope.

**How to test the change?**
Visual review should suffice.


p-datadog (approved):

p-datadog:
Thank you very much for the PR @sulami

PR #3852, opened 2024-08-20
[NO-TICKET] Add memory leak testing for profiling + fix (small) memory leak in profiler

ivoanjo:
**What does this PR do?**

The [`ruby_memcheck`](https://github.com/Shopify/ruby_memcheck) gem provides a Ruby-specific user-friendly wrapper around valgrind.

This PR sets up this gem to run the profiling test suite in CI with this gem, thus helping us catch memory leaks during development.

It also fixes a few memory leaks:
* We were leaking a small amount of memory when the Ruby process forked with the profiler enabled (we cleaned up the per-thread context hashtable, but did not call free on the things it contained)
* The stack collector native test code was a bit YOLO with exceptions and would not release memory in a few cases when exceptions were raised. This code was only used for running specs, so it did not impact actual usage, but solving it allows us to have a 100% clean ruby_memcheck run

**Motivation:**

Improve validation of our native bits.

**Additional Notes:**

I've chosen to use GitHub Actions for the new CI job, rather than CircleCI. I don't have a very strong preference for either, it's just that it was easier to use the latest valgrind version by picking the latest Ubuntu image on GHA.

In some cases, some of our tests were hanging or timing out due to how valgrind runs code in a sequential manner. I've added a new tag to be able to skip such tests.

I also explored using valgrind's `fair-sched` option (see https://github.com/Shopify/ruby_memcheck/issues/51) but ran into issues with its output being incomplete, so decided to go the skip route instead.

Initially I started testing `ruby_memcheck` with Ruby 3.3, but on that version the filtering of valgrind output is too aggressive for our setup, and it did not report anything [(see here for more details)](https://github.com/Shopify/ruby_memcheck#:~:text=use_only_ruby_free_at_exit). Thus, I moved my experiments to Ruby 3.4 and it worked much better.

**How to test the change?**

You can run the new job with `bundle exec rake spec:profiling:memcheck`.

Here's a successful run of the new CI job: https://github.com/DataDog/dd-trace-rb/actions/runs/10468881841/job/28990709088

Also, if you're curious how it looks, comment freeing any libdatadog resource or other profiler native resource, and you will see valgrind complaining. Here's what happened when I did that: https://github.com/DataDog/dd-trace-rb/actions/runs/10469151578/job/28991521438

### DIFF
--- a/ext/datadog_profiling_native_extension/collectors_thread_context.c
+++ b/ext/datadog_profiling_native_extension/collectors_thread_context.c
@@ -182,7 +182,7 @@ static VALUE _native_initialize(
 static VALUE _native_sample(VALUE self, VALUE collector_instance, VALUE profiler_overhead_stack_thread);
 static VALUE _native_on_gc_start(VALUE self, VALUE collector_instance);
 static VALUE _native_on_gc_finish(VALUE self, VALUE collector_instance);
-static VALUE _native_sample_after_gc(DDTRACE_UNUSED VALUE self, VALUE collector_instance);
+static VALUE _native_sample_after_gc(DDTRACE_UNUSED VALUE self, VALUE collector_instance, VALUE reset_monotonic_to_system_state);
 void update_metrics_and_sample(
   struct thread_context_collector_state *state,
   VALUE thread_being_sampled,
@@ -261,7 +261,7 @@ void collectors_thread_context_init(VALUE profiling_module) {
   rb_define_singleton_method(testing_module, "_native_sample_allocation", _native_sample_allocation, 3);
   rb_define_singleton_method(testing_module, "_native_on_gc_start", _native_on_gc_start, 1);
   rb_define_singleton_method(testing_module, "_native_on_gc_finish", _native_on_gc_finish, 1);
-  rb_define_singleton_method(testing_module, "_native_sample_after_gc", _native_sample_after_gc, 1);
+  rb_define_singleton_method(testing_module, "_native_sample_after_gc", _native_sample_after_gc, 2);
   rb_define_singleton_method(testing_module, "_native_thread_list", _native_thread_list, 0);
   rb_define_singleton_method(testing_module, "_native_per_thread_context", _native_per_thread_context, 1);
   rb_define_singleton_method(testing_module, "_native_stats", _native_stats, 1);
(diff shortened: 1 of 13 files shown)


p-datadog (approved):

ivoanjo:
The failure on `dd-gitlab/benchmarks` is expected. Specifically, this is the failure:

```
+ bundle exec ruby benchmarks/profiler_gc.rb
Current pid is 3104
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [x86_64-linux]
profiler - profiler gc
                    benchmarks/profiler_gc.rb:41:in `_native_sample_after_gc': wrong number of arguments (given 2, expected 1) (ArgumentError)
        Datadog::Profiling::Collectors::ThreadContext::Testing._native_sample_after_gc(@collector, false)
                                                                                       ^^^^^^^^^^^^^^^^^
	from benchmarks/profiler_gc.rb:41:in `block (2 levels) in run_benchmark'
```

This failure is happening because I've added an extra argument to `_native_sample_after_gc` on this branch BUT the benchmarking platform runs the new version of the benchmark code (which passes 2 arguments) with the code in master (which expects only one).

Once we merge this PR the issue will go away. I've considered a workaround, but changing the benchmark/or the method/or the benchmarking platform just for this rare occasion does not seem to be worth the trouble (IMO -- happy to discuss if other people aren't convinced).

ext/datadog_profiling_native_extension/collectors_thread_context.c
```diff
@@ -439,7 +439,16 @@ static VALUE _native_on_gc_finish(DDTRACE_UNUSED VALUE self, VALUE collector_ins
...
-static VALUE _native_sample_after_gc(DDTRACE_UNUSED VALUE self, VALUE collector_instance) {
+static VALUE _native_sample_after_gc(DDTRACE_UNUSED VALUE self, VALUE collector_instance, VALUE reset_monotonic_to_system_state) {
+  ENFORCE_BOOLEAN(reset_monotonic_to_system_state);
+
+  struct thread_context_collector_state *state;
+  TypedData_Get_Struct(collector_instance, struct thread_context_collector_state, &thread_context_collector_typed_data, state);
+
+  if (reset_monotonic_to_system_state == Qtrue) {
+    state->time_converter_state = (monotonic_to_system_epoch_state) MONOTONIC_TO_SYSTEM_EPOCH_INITIALIZER;
+  }
```
AlexJF, line 450: Can't seem to find any code overriding this state. Why do we need to do this?

ivoanjo: See commit description for c8f2338b201230849c6cc255c565eed06dc5d214 -- it's my attempt at fixing a flaky spec that showed up.

AlexJF: Ack! (not used to seeing useful commit descriptions 😂 )

AlexJF (approved):

ext/datadog_profiling_native_extension/collectors_thread_context.c
ivoanjo: > Ack! (not used to seeing useful commit descriptions 😂 )

🔥🔥🔥

PR #3854, opened 2024-08-21
Cleanup docker compose

TonyCTHsu:
**What does this PR do?**

1. Fix `TEST_PRESTO_HOST` from `.env` to use `127.0.0.1` 
2. Remove obsolete `common-environment-3x` anchor (They are identical to `common-environment`, except `presto` host which is fixed from `.env`)
3. Extract `common-depends-on` anchor
4. Remove obsolete `version` field from `docker-compose.yml`

**Why** 

I stumble on this overwriting complexity (with layers of overwriting environment) when testing with telemetry.


docker-compose.yml
```diff
@@ -28,7 +27,6 @@ services:
       TEST_OPENSEARCH_PORT: 9200
       TEST_POSTGRES_HOST: postgres
       TEST_PRESTO_HOST: presto
-      TEST_PRESTO_PORT: 8080
```
y9v, line 31: should this line be removed? The env var is still referenced [here](https://github.com/DataDog/dd-trace-rb/blob/7622101888c86f1ab49cb95a714af19e5c44a506/docker-compose.yml#L290).

TonyCTHsu: I think so, it is provided by `.env` file. 

For each service, we are applying environment variables by providing `env_file` and overwrite with environment. 

For example:

```
    env_file: ./.env
    environment: 
      -  ...
```

y9v: thanks! I missed the `.env` file

PR #3855, opened 2024-08-21
Add specs that were not executed to `test:contrib` task

y9v:
**What does this PR do?**
Some specs in `spec/datadog/tracing/contrib` folder were not executed with `test:contrib` rake task.
This commit enables them to see if they fail.

**Motivation:**
When working on adding [http.route](https://github.com/DataDog/dd-trace-rb/pull/3849) to tracing, I noticed a one file - `spec/datadog/tracing/contrib/http_route_spec.rb` that should have been failing in CI.

After doing a small investigation with help of @anmarchenko it was clear that this file is not being executed, along with some other spec files.

This is the result we got when running `rake test:contrib` with a breakpoint in [RSpec::Core::World](https://github.com/rspec/rspec-core/blob/main/lib/rspec/core/world.rb#L52):

```ruby
Dir.glob('./spec/datadog/tracing/contrib/*_spec.rb') - @example_groups.map(&:file_path)
=>
["./spec/datadog/tracing/contrib/component_spec.rb",
 "./spec/datadog/tracing/contrib/grpc_spec.rb",
 "./spec/datadog/tracing/contrib/http_route_spec.rb",
 "./spec/datadog/tracing/contrib/http_spec.rb",
 "./spec/datadog/tracing/contrib/span_attribute_schema_spec.rb",
 "./spec/datadog/tracing/contrib/status_range_env_parser_spec.rb",
 "./spec/datadog/tracing/contrib/status_range_matcher_spec.rb"]
```

**Additional Notes:**
Most likely some of added specs will fail, this is expected.

**How to test the change?**
It will be tested in CI pipeline.


ivoanjo:
> Some specs in spec/datadog/tracing/contrib folder were not executed with test:contrib rake task.
> This commit enables them to see if they fail.

Really nice find!

> This PR is for opening a discussion - should we delete those files or fix them if they fail?

My suggestion on this one would be to add `pending`s with TODOs for individual specs that fail. That way we could incrementally fix/review such specs.

I'm spotting that CI is failing due to missing `grape`, so I think at least some of the tests need to be run under the grape appraisal?

> Another question is whether it makes sense to change test:contrib task definition to execute all specs from contrib folder. Most likely there was some good reason to not do so.

I suspect it may be due to some of them needing extra dependencies.

p-datadog:
What is the situation with the removed http route test? Is the corresponding code covered by other tests, is there a plan to reinstate the removed test?

y9v:
@p-datadog this http route test was actually introduced in the [first attempt of http.route](https://github.com/DataDog/dd-trace-rb/pull/3345) implementation that was [rolled back](https://github.com/DataDog/dd-trace-rb/pull/3539) (due to incompatibility with rails 7.1). I am not exactly sure why this file still exists in the repo, maybe due to incorrect rebase.
I was not sure if I should add it to my [PR](https://github.com/DataDog/dd-trace-rb/pull/3849), since I think we have no appraisals that include both `grape` and `sinatra`. I think I will try to add it to my PR and see if I can make it pass.

anmarchenko (approved):
I propose to merge this now, it executes more tests than before anyway. 

`'*'` in `contrib_paths` works better than what we had before: now newly added tests will be executed by default, not ignored by default as before. Let's handle http route testing in its own PR. Thanks for handling this, @y9v

p-datadog:
:+1:

PR #3863, opened 2024-08-23
Telemetry log: Add `stack_trace` field when given an exception

TonyCTHsu:
**What does this PR do?**

Providing a class name of the exception is not very useful. Sending stack_trace help us identified the location is helpful

1. Add `stack_trace` field with reporting an exception
2. Obfuscate backtrace
3. Remove redundant `debug` level (The enriched data for debugging can be a risk and high data cardinality)

### DIFF
--- a/lib/datadog/core/telemetry/event.rb
+++ b/lib/datadog/core/telemetry/event.rb
@@ -356,10 +356,8 @@ module Datadog
             {
               logs: [
                 {
-                  # Required fields
                   message: @message,
                   level: @level,
-                  # More optional fields to be added here...
                   stack_trace: @stack_trace,
                 }.compact
               ]
--- a/lib/datadog/core/telemetry/event.rb
+++ b/lib/datadog/core/telemetry/event.rb
@@ -346,19 +346,23 @@ module Datadog
             'logs'
           end
 
-          def initialize(message:, level:)
+          def initialize(message:, level:, stack_trace: nil)
             super()
             @message = message
+            @stack_trace = stack_trace
             @level = LEVELS.fetch(level) { |k| raise ArgumentError, "Invalid log level :#{k}" }
           end
 
           def payload
             {
-              logs: [{
-                message: @message,
-                level: @level,
-                # More optional fields to be added here...
-              }]
+              logs: [
+                {
+                  message: @message,
+                  level: @level,
+                  # More optional fields to be added here...
+                  stack_trace: @stack_trace,
+                }.compact
+              ]
             }
           end
         end
--- a/lib/datadog/core/telemetry/event.rb
+++ b/lib/datadog/core/telemetry/event.rb
@@ -338,7 +338,6 @@ module Datadog
         class Log < Base
           LEVELS = {
             error: 'ERROR',
-            debug: 'DEBUG',
             warn: 'WARN',
           }.freeze
 
@@ -357,6 +356,7 @@ module Datadog
             {
               logs: [
                 {
+                  # Required fields
                   message: @message,
                   level: @level,
                   # More optional fields to be added here...
--- a/lib/datadog/core/telemetry/logging.rb
+++ b/lib/datadog/core/telemetry/logging.rb
@@ -14,14 +14,20 @@ module Datadog
 
         # Extract datadog stack trace from the exception
         module DatadogStackTrace
-          REGEX = %r{datadog-.*?/lib/datadog/}.freeze
+          # Typically, `lib` is under `#{gem_name}-#{version}/`
+          # but not the case when providing a bundler custom path in `Gemfile`
+          REGEX = %r{/lib/datadog/}.freeze
 
           def self.from(exception)
-            return unless exception.backtrace
+            backtrace = exception.backtrace
+
+            return unless backtrace
+            return if backtrace.empty?
 
             stack_trace = +''
-            (exception.backtrace || []).each do |line|
+            backtrace.each do |line|
               stack_trace << if line.match?(REGEX)
+                               # Removing host related information
                                line.sub(/^.*?(#{REGEX})/o, '\1') << ','
                              else
                                'REDACTED,'
--- a/lib/datadog/core/telemetry/logging.rb
+++ b/lib/datadog/core/telemetry/logging.rb
@@ -12,6 +12,26 @@ module Datadog
       module Logging
         extend self
 
+        # Extract datadog stack trace from the exception
+        module DatadogStackTrace
+          REGEX = %r{datadog-.*?/lib/datadog/}.freeze
+
+          def self.from(exception)
+            return unless exception.backtrace
+
+            stack_trace = +''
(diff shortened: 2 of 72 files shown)


lib/datadog/core/telemetry/logging.rb
```diff
@@ -5,13 +5,48 @@
...
+
+            stack_trace = +''
+            backtrace.each do |line|
+              stack_trace << if line.match?(REGEX)
+                               # Removing host related information
+                               line.sub(/^.*?(#{REGEX})/o, '\1') << ','
+                             else
+                               'REDACTED,'
+                             end
+            end
```
ivoanjo, line 44: So while this approach works, I'm not the biggest fan as:

* It matches on every `lib/datadog`, not just ours (e.g. the customer may have a datadog folder where they put some datadog setup they use)
* There's a bit of overhead in doing regexp matching and replacement

Have you considered prefix matching? E.g. instead of `REGEX = ...` you can get the root of the gem folder by starting from the current file `GEM_ROOT = Pathname.new("#{__dir__}/../../").cleanpath.to_s`

Then, you can match using `.start_with?` and use string slicing to remove the prefix.

lib/datadog/core/telemetry/event.rb
```diff
@@ -338,27 +338,31 @@ def payload
...
-                message: @message,
-                level: @level,
-                # More optional fields to be added here...
-              }]
+              logs: [
+                {
+                  # Required fields
+                  message: @message,
+                  level: @level,
+                  # More optional fields to be added here...
```
ivoanjo, line 362: Should we perhaps add a message about consulting some documentation / with some team before adding elements here? E.g. to avoid inconsistencies with other libraries.

lib/datadog/core/telemetry/logging.rb
```diff
@@ -5,13 +5,48 @@
...
+      # - How severe/critical is this error? (ie. error, warning, fatal)
+      # - What information needed to make it actionable?
       #
-      # Reporting internal error so that we can fix them.
-      # IMPORTANT: Make sure to not log any sensitive information.
       module Logging
         extend self
 
+        # Extract datadog stack trace from the exception
+        module DatadogStackTrace
```
ivoanjo, line 25: I possibly not for this PR, but it would be cool to perhaps leave stack entries from:

* The Ruby standard library -- paths starting with `RbConfig::CONFIG.fetch("rubylibdir")` (The profiler uses this to categorize Ruby code automatically as being from the standard library!)
* dogstatsd-ruby -- by perhaps asking rubygems where dogstatsd-ruby is, when available

ivoanjo (commented):
This is cool! Left a few notes

spec/datadog/core/telemetry/logging_spec.rb
```diff
@@ -100,3 +109,47 @@
...
+      result = described_class.from(exception)
+
+      expect(result).to eq(
+        [
+          '/lib/datadog/core/telemetry/logging.rb:1 in `report`',
+          'REDACTED',
+          'REDACTED',
+          'REDACTED'
+        ].join(',')
+      )
```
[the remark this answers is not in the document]
TonyCTHsu: > does this format comply with the format that the other libraries are using?

There is no specification about the format, but it is behaving as how DotNet is sending.

spec/datadog/core/telemetry/logging_spec.rb
```diff
@@ -100,3 +109,49 @@
...
+
+      exception = StandardError.new('Yo!')
+      exception.set_backtrace(
+        [
+          "#{gem_root}/lib/datadog/core/telemetry/logging.rb:1 in `report`",
+          '/foo/bar/baz.rb:1 in `baz`',
+          '/foo/bar.rb:1 in `bar`',
+          '/foo.rb:1 in `foo`',
+        ]
+      )
```
ivoanjo, line 143: Hmmm.... This test makes me a bit suspicious, because it just "assumes" the contents of `GEM_ROOT` are correct.

In particular, if I do this:

```ruby
          GEM_ROOT = Pathname.new("/potato").cleanpath.to_s
# or
          GEM_ROOT = Pathname.new("#{__dir__}/..").cleanpath.to_s
```

the test still passes.

The latter one is especially dangerous: it means that if someone decides to move all of telemetry to a different folder, and re-runs the tests, the test suite will pass but the `GEM_ROOT` will be incorrect (because it relies on the exact "depth" the file is in the folder structure).

Here's my suggestion: use the filtering on a real exception, and see if you get the right result:

```
    it 'returns redacted stack trace' do
      realistic_exception =
        begin
          Datadog::Core::Telemetry::Event::Base.new.type
          nil
        rescue NotImplementedError => e
          e
        end
      expect(realistic_exception).to_not be nil

      result = described_class.from(realistic_exception)

      expect(result).to start_with('/lib/datadog/core/telemetry/event.rb')
      expect(result).to end_with(',REDACTED')
    end
```

Although a bit weirder, this test will correctly fail in both of the situations I presented above.

lib/datadog/core/telemetry/event.rb
ivoanjo: Ping? :arrow_up:

ivoanjo (approved):
Left a recommendation for improving the test (I think the current one is too "mocky"), but otherwise it LGTM :+1:

lib/datadog/core/telemetry/logging.rb
```diff
@@ -2,16 +2,51 @@
...
+                               line[GEM_ROOT.length..-1] || ''
+                             else
+                               'REDACTED'
+                             end
+              stack_trace << ','
+            end
+
+            stack_trace.chomp(',')
+          end
+        end
```
ivoanjo, line 48: It occurred to me that if in the future we want to start including stack traces from other datadog libraries (datadog-ci, dogstatsd-ruby), that we don't have any "namespacing" on the current stacks.

E.g. our current stack traces look like

```
/lib/some/file.rb: ...
```

which is fine if we're only allowing files from dd-trace-rb in the stacktrace.

But if we later allow other datadog libraries, how will this look?

```
/lib/some/file.rb: ... # is this file from dd-trace-rb? or not?
/lib/some/other_file.rb: ... # is this file from dd-trace-rb? or not?
```

TL;DR should we prefix the gem name in the stacks we generate?

e.g.

```
datadog/lib/some/file.rb
datadog-ci/lib/some/other_file.rb
dogstatsd-ruby/lib/yet_another/file.rb
```

This is not a problem on this PR, but does mean that if we go as-is, later we'll "invalidate" our stacks if we need to change, so perhaps it's worth considering already prepending `datadog` on this PR?

TonyCTHsu: As far as I know, we might need to able to accept more roots to filter, but the our file system convention provides namespaces under `/lib/datadog/..`. 

For example: 
- `/lib/datadog/tracing`
- `/lib/datadog/profiling`

PR #3866, opened 2024-08-26
sample before propagation, context switching, and writing

ZStriker19:
**What does this PR do?**
This change is so that sampling will be performed as late as possible for a trace. This gives us the greatest amount of information we can have on a trace available (namely tags and resource name) to choose if that trace should be sampled or not.

1. Modifies the sampler so that it can evaluate right before propagation, forking, or writing occurs.
2. Samples right before the above scenarios
3. Removes sampling on span creation.

**Motivation:**
Bring Ruby tracer in line with other tracers and allow further user fine grain control over sampling.

**How to test the change?**

### DIFF
--- a/lib/datadog/tracing/distributed/propagation.rb
+++ b/lib/datadog/tracing/distributed/propagation.rb
@@ -62,6 +62,12 @@ module Datadog
           end
 
           result = false
+          # Probably sample here
+          # Do we need to update tags on the parent span here first?
+          # That should probably be done in the tracer, not here.
+          # Can we sample off of digest or do we need to grab the parent span?
+          # Looking at how we do span sampling would be helpful to answer some of this.
+          components.sampler.sample!(digest)
 
           # Inject all configured propagation styles
           @propagation_style_inject.each do |propagator|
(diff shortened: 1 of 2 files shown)


lib/datadog/tracing/distributed/propagation.rb
```diff
@@ -62,6 +62,12 @@ def inject!(digest, data)
           end
 
           result = false
+          # Probably sample here
```
ZStriker19, line 65: Line 57 we try to convert a span to a digest anyways.

ZStriker19: However we could possibly get the current span with the tracer object?

ZStriker19: The issue here is that the digest is already created at this point, so even if we run sampling on the trace, the digest won't change.

ivoanjo:
Hey @ZStriker19 is still relevant?

ivoanjo:
This has been stale for a long time, so I'm going to go ahead and close it; please do reopen if this becomes relevant again!

PR #3867, opened 2024-08-27
Avoid tracking logins from Devise::Strategies::Rememberable

y9v:
**What does this PR do?**
In Devise integration [`Devise::Strategies::Authenticatable#validate`](https://github.com/heartcombo/devise/blob/main/lib/devise/strategies/authenticatable.rb#L37) is patched to track login successes and failures.

This method is not only used by [`Devise::Strategies::DatabaseAuthenticatable`](https://github.com/heartcombo/devise/blob/main/lib/devise/strategies/database_authenticatable.rb#L13), but also by [`Devise::Strategies::Rememberable`](https://github.com/heartcombo/devise/blob/main/lib/devise/strategies/rememberable.rb#L29) that is used for signing in the user using a remember token cookie.

**Motivation:**
This should reduce the amount of false-positive Login Success events for Devise AppSec integration.

**Additional Notes:**
I added an alias method for `validate` to avoid calling `super` from `Rememberable` strategy, since the super method is the `validate` from prepended `AuthenticatablePatch`.

**How to test the change?**
I tested this in a simple rails application with devise.
To test that regular login tracking works it's enough to log in via the sign-in form and observe a `users.login.success` tag on a trace.

To test that logins from `Devise::Strategies::Rememberable` strategy are not tracked:
- Log in via the sign-in form with checked "Remember Me" checkbox
- Go to any page that requires user authentication and delete the session cookie from Developer Tools
- Reload the page, the user will still be logged in, and the session cookie should reappear
- Observe that the trace was not tagged with `users.login.success`


ivoanjo:
Feel free to ignore the failing memory leaks test, fix is in https://github.com/DataDog/dd-trace-rb/pull/3868

lib/datadog/appsec/contrib/devise/patcher.rb
```diff
@@ -23,16 +24,26 @@ def target_version
...
+            patch_authenticatable_strategy
+            patch_rememberable_strategy
             patch_registration_controller
 
             Patcher.instance_variable_set(:@patched, true)
           end
 
-          def patch_authenticable_strategy
+          def patch_authenticatable_strategy
+            ::Devise::Strategies::Authenticatable.alias_method(:__validate, :validate)
```
p-datadog, line 35: I suggest using a prefix/postfix that indicates the purpose and owner of this alias, e.g. something along the lines of `__validate_datadog` or `__validate_datadog_authenticatable`. With just a double-underscore prefix there is potential for collisions with other code and it's not clear what the purpose of the method is to someone who might be looking at it in their application for whatever reason (and has no knowledge of dd-trace-rb internals).

p-datadog (approved):

TonyCTHsu (approved):

lib/datadog/appsec/contrib/devise/patcher.rb
```diff
@@ -23,16 +24,26 @@ def target_version
...
+          def patch_authenticatable_strategy
+            ::Devise::Strategies::Authenticatable.alias_method(:__validate_datadog_authenticatable, :validate)
             ::Devise::Strategies::Authenticatable.prepend(AuthenticatablePatch)
           end
 
+          def patch_rememberable_strategy
+            return unless ::Devise::STRATEGIES.include?(:rememberable)
+
+            # Rememberable strategy is required in autoloaded Rememberable model
+            ::Devise::Models::Rememberable # rubocop:disable Lint/Void
```
lloeki, line 43: Oh my, I hate autoloading so much :(

lib/datadog/appsec/contrib/devise/patcher.rb
```diff
@@ -23,16 +24,26 @@ def target_version
...
+            patch_authenticatable_strategy
+            patch_rememberable_strategy
             patch_registration_controller
 
             Patcher.instance_variable_set(:@patched, true)
           end
 
-          def patch_authenticable_strategy
+          def patch_authenticatable_strategy
+            ::Devise::Strategies::Authenticatable.alias_method(:__validate_datadog_authenticatable, :validate)
```
lloeki, line 35: This looks dangerous, as mixing aliasing and prepending can easily lead to obscure issues such as recursive loops.

AIUI the reason this was introduced is that the call stack is:

```
Foo#validate    # some app class that included ::Devise::Strategies::Rememberable and may implement validate that calls `super` or simply rely on it being provided by a parent
::Devise::Strategies::Rememberable#validate                                   # hypothetical, not there upstream as of yet
::Datadog::AppSec::Contrib::Devise::Patcher::AuthenticatablePatch#validate
::Devise::Strategies::Authenticatable#validate
```

But just prepending a patch results in this:

```
Foo#validate    # some app class that included ::Devise::Strategies::Rememberable and may implement validate that calls `super` or simply rely on it being provided by a parent
::Datadog::AppSec::Contrib::Devise::Patcher::RememberablePatch#validate
::Devise::Strategies::Rememberable#validate                                   # hypothetical, not there upstream as of yet
::Datadog::AppSec::Contrib::Devise::Patcher::AuthenticatablePatch#validate
::Devise::Strategies::Authenticatable#validate
```

But we want to skip some/all of `::Datadog::AppSec::Contrib::Devise::Patcher::AuthenticatablePatch#validate`, so this attempts to walk around it with:

```
Foo#validate    # some app class that included ::Devise::Strategies::Rememberable and may implement validate that calls `super` or simply rely on it being provided by a parent
::Datadog::AppSec::Contrib::Devise::Patcher::RememberablePatch#validate
::Devise::Strategies::Authenticatable#__validate_datadog_authenticatable
```

The problems are:

- suddenly `::Devise::Strategies::Authenticatable#validate` is not called at all (the method body is, but not the name). This can cause issues: a third party instrumenting `::Devise::Strategies::Authenticatable#validate` as well would never be called, or if `::Devise::Strategies::Authenticatable#validate`'s source code calls `super` it will actually look for `__validate_datadog_authenticatable` in a parent.
- Between `::Datadog::AppSec::Contrib::Devise::Patcher::RememberablePatch#validate` and `::Devise::Strategies::Authenticatable#__validate_datadog_authenticatable` the ancestor chain of calls is broken: Ruby "reenters" the instance and starts a new ancestor chain of calls, which creates the opportunity for infinite recursive looping.

I would favour a method where the patches collaborate through context passing: `RememberablePatch` sets a flag in a context, and `AuthenticatablePatch` checks for that flag.

lib/datadog/appsec/contrib/devise/patcher.rb
y9v: yeah, especially when autoloaded classes require other classes/modules

lib/datadog/appsec/contrib/devise/patcher.rb
y9v: Thank you! It is indeed better to use some variable for context. Fixed in [c65e4b0](https://github.com/DataDog/dd-trace-rb/pull/3867/commits/c65e4b03ea961d2a1fd126e700e30e1999efeaad)

y9v:
Failing system tests should be fixed by [this PR](https://github.com/DataDog/dd-trace-rb/pull/3886#pullrequestreview-2285421050)

Strech (approved):
Like it!

anmarchenko (approved):

PR #3869, opened 2024-08-27
[NO-TICKET] Raise benchmark default durations

ivoanjo:
**What does this PR do?**

This PR goes through our existing benchmarks, and for those that used a default duration of 10 or 12 seconds, raises the duration to 30s.

**Motivation:**

We've observed most of these benchmarks having flaky regressions/improvements on PRs that shouldn't affect them.

I suspect that the running the benchmarks for such short time may be contributing to flakiness. For instance, one of the benchmarks that flakes the most often is `profiler_sample_serialize.rb`, and concidentally that's a benchmark where each iteration does a lot of work, and thus in a typical 10 second run we may only see around 70 iterations.

Hopefully by running the benchmarks for slightly longer we'll have more consistent results.

**Additional Notes:**

There is a downside to this -- because our benchmarks are currently executed sequentially in CI, this will make the benchmark run took quite longer than it used to.

Hopefully this trade-off is reasonable; if not, we can re-evaluate.

**How to test the change?**

Check the latest benchmarks CI run, and confirm the tests are running for 30 seconds, rather than 10/12.

### DIFF
--- a/benchmarks/profiler_hold_resume_interruptions.rb
+++ b/benchmarks/profiler_hold_resume_interruptions.rb
@@ -16,7 +16,7 @@ class ProfilerHoldResumeInterruptions
 
   def run_benchmark
     Benchmark.ips do |x|
-      benchmark_time = VALIDATE_BENCHMARK_MODE ? { time: 0.01, warmup: 0 } : { time: 10, warmup: 2 }
+      benchmark_time = VALIDATE_BENCHMARK_MODE ? { time: 0.01, warmup: 0 } : { time: 30, warmup: 2 }
       x.config(
         **benchmark_time,
       )
--- a/benchmarks/profiler_hold_resume_interruptions.rb
+++ b/benchmarks/profiler_hold_resume_interruptions.rb
@@ -16,7 +16,7 @@ class ProfilerHoldResumeInterruptions
 
   def run_benchmark
     Benchmark.ips do |x|
-      benchmark_time = VALIDATE_BENCHMARK_MODE ? { time: 0.01, warmup: 0 } : { time: 30, warmup: 2 }
+      benchmark_time = VALIDATE_BENCHMARK_MODE ? { time: 0.01, warmup: 0 } : { time: 60, warmup: 2 }
       x.config(
         **benchmark_time,
       )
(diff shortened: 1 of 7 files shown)


benchmarks/profiler_hold_resume_interruptions.rb
```diff
@@ -16,7 +16,7 @@ def create_profiler
 
   def run_benchmark
     Benchmark.ips do |x|
-      benchmark_time = VALIDATE_BENCHMARK_MODE ? { time: 0.01, warmup: 0 } : { time: 10, warmup: 2 }
+      benchmark_time = VALIDATE_BENCHMARK_MODE ? { time: 0.01, warmup: 0 } : { time: 60, warmup: 2 }
```
AlexJF, line 19: Is this 60 intentional?

AlexJF (approved):

benchmarks/profiler_hold_resume_interruptions.rb
ivoanjo: It is! I pushed 01b37f8954cc30c27ba1609fb3d6851b6be7b94c as a last-ditch attempt after I saw a false performance change in this PR...

And now I'm seeing another one coming from the tracing benchmarks. I'm going to try tweaking them to 60 seconds, and if that doesn't do it, I'll throw this PR away :/

AlexJF (approved):

p-datadog (approved):
I support the changes if they make the benchmarks less flaky, but the runtime is now 45 minutes for this configuration? Should we consider splitting it?

ivoanjo:
> I support the changes if they make the benchmarks less flaky, but the runtime is now 45 minutes for this configuration? Should we consider splitting it?

Good point. To be honest, I'm still seeing too much variation on the benchmarks, so as a last desperate measure I'll try raising every one to 1 minute.

If one minute is reasonable (I'm starting to suspect it's not going to be), I think it's worth looking into the splitting...

ivoanjo:
Ok, it seems like even with 60 seconds duration, we're still seeing bogus improvements/regressions being reported. Let me move this PR to draft for now while I follow-up on other solutions.

ivoanjo:
We ended up tweaking the thresholds on the benchmarking platform and it reduced the variance a lot. There's still a few tests that seem to show wider variance -- I'll open up separate PRs for tweaking their duration as needed.

Closing this one for now.

PR #3875, opened 2024-08-29
Expand stripe matrix

TonyCTHsu:
**What does this PR do?**

Expand stripe test matrix.

Our stripe instrumentation is very stable, and I am using it as a demonstration for improving coverage.


y9v (approved):

tasks/appraisal.rake
```diff
@@ -202,17 +202,13 @@ TRACER_VERSIONS = [
   '3.2',
   '3.3',
   '3.4',
-  # ADD NEW RUBIES HERE
```
ivoanjo, line 205: This note is important! We've spread these notes in our codebase to mark all places that need to be updated when a new Ruby comes out. I do think it's important to leave it :/

TonyCTHsu: https://github.com/DataDog/dd-trace-rb/pull/3876/commits/1c058c4ab167cd907b43b12881a7e3aa7dbced29

PR #3877, opened 2024-08-30
Backport disable persist credentials

TonyCTHsu:
**What does this PR do?**

Backport: https://github.com/DataDog/dd-trace-rb/pull/3860

### DIFF
--- a/.github/workflows/lib-injection.yml
+++ b/.github/workflows/lib-injection.yml
@@ -50,8 +50,8 @@ jobs:
     env:
       TEST_LIBRARY: ruby
       WEBLOG_VARIANT: ${{ matrix.weblog-variant }}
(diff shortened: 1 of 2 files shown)


.github/workflows/lib-injection.yml
```diff
@@ -58,9 +58,10 @@ jobs:
       uses: actions/checkout@v4
       with:
           repository: 'DataDog/system-tests'
+          persist-credentials: false
```
ivoanjo, line 61: :arrow_down: I think we're missing the same change in test-negative, below.

ivoanjo (approved):
There's one missing change, but otherwise it LGTM :+1:

cbeauchesne (approved):

.github/workflows/lib-injection.yml
TonyCTHsu: Good catch! Updated

PR #3881, opened 2024-09-03
[SDTEST-756] add +ci-X.Y.Z to the tracer_version reported to internal telemetry if datadog-ci is present and CI mode is enabled

anmarchenko:
**What does this PR do?**
As discussed before, we will extend tracer_version reported for internal telemetry with datadog-ci gem version if tracing is running in CI mode.

The monkey patch should be removed in the next release: https://github.com/DataDog/datadog-ci-rb/pull/228

**Motivation:**
We need a better solution to distinguish datadog-ci versions when examining telemetry metrics emitted from the datadog-ci gem.

**Additional Notes:**
The original idea with `+ci-X.Y.Z` doesn't work perfect - it gets replaced by `_`:
image

`-ci-X.Y.Z` works fine though:
image

**How to test the change?**
Tested using staging (see reported tracer versions above). I didn't add unit tests because it would require bringing datadog-ci in appraisals which would be an overkill.

### DIFF
--- a/lib/datadog/core/telemetry/request.rb
+++ b/lib/datadog/core/telemetry/request.rb
@@ -33,8 +33,8 @@ module Datadog
             config = Datadog.configuration
 
             tracer_version = Core::Environment::Identity.gem_datadog_version_semver2
-            if config.respond_to?(:ci) && config.ci.enabled && defined?(::Datadog::CI::VERSION)
-              tracer_version = "#{tracer_version}+ci-#{::Datadog::CI::VERSION}"
+            if defined?(::Datadog::CI::VERSION) && config.respond_to?(:ci) && config.ci.enabled
+              tracer_version = "#{tracer_version}-ci-#{::Datadog::CI::VERSION::STRING}"
             end
 
             {
--- a/lib/datadog/core/telemetry/request.rb
+++ b/lib/datadog/core/telemetry/request.rb
@@ -31,6 +31,12 @@ module Datadog
 
           def application
             config = Datadog.configuration
+
+            tracer_version = Core::Environment::Identity.gem_datadog_version_semver2
+            if config.respond_to?(:ci) && config.ci.enabled && defined?(::Datadog::CI::VERSION)
+              tracer_version = "#{tracer_version}+ci-#{::Datadog::CI::VERSION}"
+            end
+
             {
               env: config.env,
               language_name: Core::Environment::Ext::LANG,
@@ -39,7 +45,7 @@ module Datadog
               runtime_version: Core::Environment::Ext::ENGINE_VERSION,
               service_name: config.service,
               service_version: config.version,
-              tracer_version: Core::Environment::Identity.gem_datadog_version_semver2
+              tracer_version: tracer_version
             }
           end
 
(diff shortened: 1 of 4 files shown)


lib/datadog/core/telemetry/request.rb
```diff
@@ -31,6 +31,12 @@ def build_payload(event, seq_id)
 
           def application
             config = Datadog.configuration
+
+            tracer_version = Core::Environment::Identity.gem_datadog_version_semver2
+            if config.respond_to?(:ci) && config.ci.enabled && defined?(::Datadog::CI::VERSION)
+              tracer_version = "#{tracer_version}+ci-#{::Datadog::CI::VERSION}"
+            end
```
ivoanjo, line 38: Really minor: Consider starting with the `defined?` first? I suspect that one is cheaper than any of the other checks, and will allows us to bail out earlier.

p-datadog:
Does this PR not need any test coverage?

lib/datadog/core/telemetry/request.rb
```diff
@@ -39,7 +45,7 @@ def application
               runtime_version: Core::Environment::Ext::ENGINE_VERSION,
               service_name: config.service,
               service_version: config.version,
-              tracer_version: Core::Environment::Identity.gem_datadog_version_semver2
+              tracer_version: tracer_version
```
ivoanjo, line 48: The previous monkey patching approach affected a lot more than just the telemetry request (e.g. effectively anything that used the `gem_datadog_version` helpers.

I'm assuming you're ok with this new reporting only affecting telemetry requests; is that correct?

lib/datadog/core/telemetry/request.rb
```diff
@@ -31,6 +31,12 @@ def build_payload(event, seq_id)
 
           def application
             config = Datadog.configuration
+
+            tracer_version = Core::Environment::Identity.gem_datadog_version_semver2
+            if defined?(::Datadog::CI::VERSION) && config.respond_to?(:ci) && config.ci.enabled
+              tracer_version = "#{tracer_version}-ci-#{::Datadog::CI::VERSION::STRING}"
+            end
```
ivoanjo, line 38: It may be worth adding here a bit of context to why we're doing this. Something closer to the commit description?

Since this is "suspicious", I think it's the kind of thing that deserves a bit of a note acknowledging that.

lib/datadog/core/telemetry/request.rb
ivoanjo, line 1: > Tested using staging (see reported tracer versions above). I didn't add unit tests because it would require bringing datadog-ci in appraisals which would be an overkill.

Actually, I think appraisals is probably the solution to use here. Otherwise, maybe at least a unit test that mocks the CI stuff?

It's weird to think that I can remove this change tomorrow and nothing will tell me something went wrong, until some time later we realize our dashboards are missing info for the CI gem.

ivoanjo (commented):
Left a few notes!

lib/datadog/core/telemetry/request.rb
anmarchenko: yes, I specifically checked that gem_datadog_version wasn't used anywhere else except telemetry

lib/datadog/core/telemetry/request.rb
anmarchenko: yeah, makes sense, I will try to mimic CI in a test

ivoanjo (approved):
:+1: LGTM

y9v (approved):

PR #3883, opened 2024-09-03
Fix system tests

p-datadog:
Add build-harness and build-apps to the list of needs for aggregate job so it won't run if a build job fails


cbeauchesne (approved):

p-datadog:
I will merge this and it looks like https://github.com/DataDog/dd-trace-rb/pull/3886 is attempting to repair the ASM test failures.

PR #3885, opened 2024-09-04
[PROF-10477] Upgrade to libdatadog 12

AlexJF:
**What does this PR do?**

This PR upgrades the datadog gem to use libdatadog 12.

Drop-in upgrade due to no API changes.

**Motivation:**

Staying on the latest libdatadog enables us to pick up the latest improvements and also makes it easier to develop new features that need libdatadog changes.

**How to test the change?**

Our existing test coverage includes libdatadog testing, so a green CI is good here :)


ext/libdatadog_api/datadog_ruby_common.c
```diff
@@ -29,7 +29,7 @@ VALUE datadog_gem_version(void) {
 }
 
 __attribute__((warn_unused_result))
-ddog_prof_Endpoint endpoint_from(VALUE exporter_configuration) {
+ddog_Endpoint* endpoint_from(VALUE exporter_configuration) {
   ENFORCE_TYPE(exporter_configuration, T_ARRAY);
```
ivoanjo, line 33: I recently introduced this `datadog_ruby_common.h|.c` because it turns out to be surprisingly hard to get extconf.rb/mkmf to allow us to use headers and c files from a different folder.

(Specifically, to not treat them as an outside library, and e.g. include them in the makefile so that they trigger rebuilds and whatnot).

So I created this .h/.c to contain things that both `libdatadog_api` and `datadog_profiling_native_extension` need. And there's a copy-pasted copy on each folder.

I guess since we need to diverge for this function, my suggestion is to remove it entirely from the common .h/.c: move the libdatadog_api copy to crashtracker.c and the profiler copy to ... libdatadog_helpers.c? http_transport.c?

That way we can keep the common file in-sync with copy-paste :smile:

ext/libdatadog_api/datadog_ruby_common.c
```diff
@@ -47,11 +47,20 @@ ddog_prof_Endpoint endpoint_from(VALUE exporter_configuration) {
...
+    ddog_Endpoint *endpoint = NULL;
+    ddog_Error *error = ddog_endpoint_from_api_key_and_site(
+        char_slice_from_ruby_string(site), char_slice_from_ruby_string(api_key),
+        &endpoint);
+    if (error != NULL) {
+      rb_raise(rb_eRuntimeError,
+               "Failed to initialize agentless endpoint: %" PRIsVALUE,
+               get_error_details_and_drop(error));
+    }
+    return endpoint;
```
ivoanjo, line 59: Minor: TBH I'm not sure if we care about agentless error tracking (E.g. if we're going to have a version of this function for crashtracking only, maybe we can just simplify and drop the agentless support there)

ivoanjo (approved):
Other than the suggestion to move stuff away from datadog_ruby_common, looks good to me, thanks for looking into this :+1:

ext/libdatadog_api/datadog_ruby_common.c
AlexJF: Done!

ext/libdatadog_api/datadog_ruby_common.c
AlexJF: Done!

lib/datadog/core/crashtracking/component.rb
```diff
@@ -95,7 +95,7 @@ def stop
         def start_or_update_on_fork(action:)
           self.class._native_start_or_update_on_fork(
             action: action,
-            exporter_configuration: [:agent, agent_base_url],
+            agent_base_url: agent_base_url,
             path_to_crashtracking_receiver_binary: path_to_crashtracking_receiver_binary,
             ld_library_path: ld_library_path,
             tags_as_array: tags.to_a,
```
ivoanjo, line 101: Oh, I like this simplification! :)

ivoanjo (approved):
:+1: LGTM

Issue #3874, opened 2024-08-29
Upgrade from 2.2.0 to 2.3.0 breaks app


nesrual:
**Current behaviour**
Logging works as expected with 2.2.0 

**Expected behaviour**
The app crashes with this error message:

```
undefined method `arguments' for an instance of GraphQL::Language::Nodes::FragmentSpread (NoMethodError)

                  set_hash_with_variables(resolver_args, selection.arguments, query.provided_variables)
                                                                  ^^^^^^^^^^
```

This is when GraphQL queries are performed:

```
result = DaBackendSchema.execute(query, variables:, context:, operation_name:)
```

```
Crashed in non-app: datadog (2.3.0) lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb in block in create_arguments_hash
```

y9v:
hey @nesrual, thank you for reaching out!

I will look into this.

y9v:
@nesrual could you tell me your `graphql` gem version, and maybe also something about how I can reproduce this issue?

Is this some particular query that is failing?

nesrual:
It seems to be all queries. I have not been able to reproduce the error in test/dev yet. I will test further tomorrow and let you know my findings.

We use the following GraphQL gems:

```
graphql-pro (1.29.0)
graphql (2.3.14)
      base64
      fiber-storage
    graphql-batch (0.6.0)
      graphql (>= 1.12.18, < 3)
      promise.rb (~> 0.7.2)
    graphql-c_parser (1.1.1)
      graphql (>= 2.2.10)
    graphql-persisted_queries (1.8.1)
      graphql (>= 1.12)
```

nesrual:
I have done some more digging around and I found that this triggers the error:

This query will result in a 500 server error:

```
query {
  accounts {
    ...AccountDetails
    __typename
  }
}

fragment AccountDetails on Account {
  id
  name
  accountType
  useBudget
  parent {
    id
    __typename
  }
  __typename
}
```

whereas this works fine:

```
query {
  accounts {
  id
  name
  accountType
  useBudget
  parent {
    id
    __typename
  }
    __typename
  }
}
```

So I think it could have something to do with using `fragments` in the query.

The error returned is:

```
web-1  | App 321 output: NoMethodError (undefined method `arguments' for an instance of GraphQL::Language::Nodes::FragmentSpread):
web-1  | App 321 output:
[...]

PR #3887, opened 2024-09-04
Fix building of GraphQL queries arguments hash

y9v:
Fixes https://github.com/DataDog/dd-trace-rb/issues/3874

**What does this PR do?**
This PR changes the way we extract arguments from GraphQL queries.
It fixes following things:
- handling of fragments
- handling for field aliases
- handling of default values
- handling of mutation input object

**Motivation:**
[This issue](https://github.com/DataDog/dd-trace-rb/issues/3874)

**Additional Notes:**
This PR also changes the keys for arguments, they were query names before, but according to our RFC it should be resolver names.

**How to test the change?**
Unit tests are included. I also set up a local test GraphQL application for testing which I can share.


spec/datadog/appsec/contrib/graphql/gateway/multiplex_spec.rb
```diff
@@ -1,31 +1,382 @@
...
+    # we are only testing how arguments are extracted from the queries,
+    # therefore we don't need a real schema here
+    stub_const('TestSchema', Class.new(::GraphQL::Schema))
+  end
+
+  let(:multiplex) do
+    ::GraphQL::Execution::Multiplex.new(
+      schema: schema,
+      queries: queries,
+      context: { :dataloader => GraphQL::Dataloader.new(nonblocking: nil) },
```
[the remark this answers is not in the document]
y9v: fixed in [55606b3](https://github.com/DataDog/dd-trace-rb/pull/3887/commits/55606b358034016499ea3a96e8f00da9e0896b8e)

lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb
```diff
@@ -28,41 +28,74 @@ def queries
...
+            def arguments_from_directives(directives, query_variables)
+              directives.map do |directive|
+                next unless directive.is_a?(::GraphQL::Language::Nodes::Directive)
+
+                [directive.name, arguments_hash(directive.arguments, query_variables)]
+              end.to_h
+            end
+
+            def arguments_hash(arguments, query_variables)
+              arguments.map do |argument|
```
anmarchenko, line 83: if we use more stupid solution

```ruby
arguments.each_with_object({})  do |argument, hash|
  hash[argument.name] = argument_value(argument, query_variables)
end
```

it will save N+1 array allocations in memory, which on a scale [could be a lot](https://mensfeld.pl/2024/08/under-the-hood-enhancing-karafkas-cpu-and-memory-efficiency)

lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb
```diff
@@ -28,41 +28,74 @@ def queries
...
-            def set_hash_with_variables(resolver_hash, arguments, provided_variables)
-              arguments.each do |arg|
-                resolver_hash[arg.name] =
-                  if arg.value.is_a?(::GraphQL::Language::Nodes::VariableIdentifier)
-                    provided_variables[arg.value.name]
-                  else
-                    arg.value
-                  end
+            def arguments_from_directives(directives, query_variables)
+              directives.map do |directive|
```
anmarchenko, line 75: same comment as for the line 83

lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb
```diff
@@ -28,41 +28,74 @@ def queries
...
+            def arguments_from_selections(selections, query_variables, args_hash)
+              selections.each do |selection|
+                next unless selection.is_a?(::GraphQL::Language::Nodes::Field)
+
+                selection_name = selection.alias || selection.name
+
+                if selection.arguments.any? || selection.directives.any?
+                  args_hash[selection_name] ||= []
+                  args_hash[selection_name] <<
+                    arguments_hash(selection.arguments, query_variables).merge(
```
anmarchenko, line 65: if we use `merge!`, it saves one hash allocation

lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb
```diff
@@ -28,41 +28,74 @@ def queries
...
+              end
+            end
+
+            def arguments_from_selections(selections, query_variables, args_hash)
+              selections.each do |selection|
+                next unless selection.is_a?(::GraphQL::Language::Nodes::Field)
+
+                selection_name = selection.alias || selection.name
+
+                if selection.arguments.any? || selection.directives.any?
```
anmarchenko, line 62: Do you use `any?` here because we actually need to check if there are some non-nil elements in the array or would checking if array is empty enough?

`any?` implicitly [constructs a block and iterates the enumerable](https://apidock.com/ruby/Enumerable/any%3F), `empty?` only checks the array length (so it is almost a single instruction operation)

lib/datadog/appsec/contrib/graphql/gateway/watcher.rb
```diff
@@ -24,27 +24,30 @@ def watch_multiplex(gateway = Instrumentation.gateway)
...
-                  AppSec::Reactive::Operation.new('graphql.multiplex') do |op|
-                    GraphQL::Reactive::Multiplex.subscribe(op, scope.processor_context) do |result|
-                      event = {
-                        waf_result: result,
-                        trace: scope.trace,
-                        span: scope.service_entry_span,
-                        multiplex: gateway_multiplex,
-                        actions: result.actions
-                      }
+                  if scope
```
anmarchenko, line 30: is it possible to use `return nil unless scope` here instead of wrapping it with `if`

lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb
```diff
@@ -28,41 +28,74 @@ def queries
...
+              end.to_h
+            end
+
+            def argument_value(argument, query_variables)
+              case argument.value
+              when ::GraphQL::Language::Nodes::VariableIdentifier
+                # we need to pass query.variables here instead of query.provided_variables,
+                # since #provided_variables don't know anything about variable default value
+                var_name = argument.value.name
+                query_variables.fetch(var_name) if query_variables.key?(var_name)
```
anmarchenko, line 94: I think this line does basically the same as `query_variables[var_name]` because it still returns nil if var_name is not in a hash

lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb
y9v: it should be enough to check if it is empty actually, thanks!

lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb
y9v: Fixed here: [d51cc17](https://github.com/DataDog/dd-trace-rb/pull/3887/commits/d51cc170da1a913520719904697ed343f43d51ad)

lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb
y9v: Fixed: [d51cc17](https://github.com/DataDog/dd-trace-rb/pull/3887/commits/d51cc170da1a913520719904697ed343f43d51ad)

lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb
y9v: Fixed [d51cc17](https://github.com/DataDog/dd-trace-rb/pull/3887/commits/d51cc170da1a913520719904697ed343f43d51ad)

lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb
y9v: fixed [d620229](https://github.com/DataDog/dd-trace-rb/pull/3887/commits/d62022980f08762cd55afb8189025e723b9dd5ed)

lib/datadog/appsec/contrib/graphql/gateway/watcher.rb
y9v: yes, this would actually be better. Fixed here: [31da3fb](https://github.com/DataDog/dd-trace-rb/pull/3887/commits/31da3fba8f87e39277387726e4b4d1526d02bf35)

anmarchenko: now I'm having second thoughts about my comment 😓 

Is it ok that when we use `return` the following will not be called:
```ruby
                  next [nil, [[:block, event]]] if block
                  ret, res = stack.call(gateway_multiplex.arguments)
                  if event
                    res ||= []
                    res << [:monitor, event]
                  end
                  [ret, res]
```

I might have proposed some premature optimisation

anmarchenko (approved):
Looks good, thank you 👍🏻

lib/datadog/appsec/contrib/graphql/gateway/watcher.rb
y9v: I think this is fine considering that `event` is set within [`GraphQL::Reactive::Multiplex.subscribe` block](https://github.com/DataDog/dd-trace-rb/blob/35f8ccd7703337dfbe333d9af038a7f68c3bf4f0/lib/datadog/appsec/contrib/graphql/gateway/watcher.rb#L32) that relies on the active scope being present.

anmarchenko: no, actually this is a block that is passed to `watch`, so we should use `next` with appropriate arguments, not `return`

I confused you previously, I think the version with `if scope` was actually more correct

y9v: yes, you are right, I removed my commit

TonyCTHsu:
We've talked about being slightly more defensive:

1. Rescuing unexpected error and return a benign value 
2. Validate the presence of constant which the implementation depends on before applying the patch

lib/datadog/appsec/contrib/graphql/integration.rb
```diff
@@ -24,13 +31,19 @@ def self.loaded?
...
           def self.compatible?
-            super && version >= MINIMUM_VERSION
+            super && version >= MINIMUM_VERSION && ast_node_classes_defined?
           end
 
           def self.auto_instrument?
             true
           end
 
+          def self.ast_node_classes_defined?
```
TonyCTHsu, line 41: Adding a couple of test cases 

1. Test positive with our graphql appraisal groups.
2. Test negative with `hide_const`

TonyCTHsu (commented):
Approve!

lib/datadog/appsec/contrib/graphql/integration.rb
y9v: added here: [9522f70](https://github.com/DataDog/dd-trace-rb/pull/3887/commits/9522f70bd948866ef9de043a4b9ed10a3a32aee1)

Strech (commented):
Can't comment much on the correctness of the code, but here are my 50c with the tests.

spec/datadog/appsec/contrib/graphql/gateway/multiplex_spec.rb
```diff
@@ -1,31 +1,380 @@
...
+            post(slug: $postSlug) { title }
+          }
+        END_OF_QUERY
+      )
+    end
+
+    let(:queries) { [query] }
+
+    it 'returns queries' do
+      expect(dd_multiplex.queries).to eq(queries)
```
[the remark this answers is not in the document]
y9v: thanks! fixed in [3562af9](https://github.com/DataDog/dd-trace-rb/pull/3887/commits/3562af94bfa9a25fa9ca108cb53e5d77d9972cbc)

spec/datadog/appsec/contrib/graphql/integration_spec.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+require 'datadog/appsec/spec_helper'
+require 'datadog/appsec/contrib/graphql/integration'
+
+RSpec.describe Datadog::AppSec::Contrib::GraphQL::Integration do
+  describe '.ast_node_classes_defined?' do
+    it 'returns true when all AST node classes are defined' do
+      expect(described_class.ast_node_classes_defined?).to eq(true)
```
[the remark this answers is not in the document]
y9v: it is the same actually, I just thought that `be(true)` would be a bit more readable

lloeki (approved):

lloeki:
Approved, although. I can't say much about the more intricate technical details here.

lib/datadog/appsec/contrib/graphql/gateway/multiplex.rb
```diff
@@ -55,7 +55,9 @@ def build_arguments_hash
 
             def arguments_from_selections(selections, query_variables, args_hash)
               selections.each do |selection|
-                next unless selection.is_a?(::GraphQL::Language::Nodes::Field)
+                # rubocop:disable Style/ClassEqualityComparison
+                next unless selection.class.to_s == Integration::AST_NODE_CLASS_NAMES[:field]
```
p-datadog, line 59: If this is intending to get the class name, why not use `selection.class.name`?

y9v: oh, good catch, sorry, I should use Copilot more carefully

lib/datadog/appsec/contrib/graphql/appsec_trace.rb
```diff
@@ -27,20 +27,10 @@ def execute_multiplex(multiplex:)
...
-            Datadog::Tracing.active_trace
-          end
-
-          def active_span
-            return unless defined?(Datadog::Tracing)
+          rescue StandardError => e
+            Datadog.logger.error(e.message)
 
-            Datadog::Tracing.active_span
+            multiplex_return || super
```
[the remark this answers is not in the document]
y9v: here the `multiplex_return` is actually the return value of previous `super`. I am now thinking whether there may be cases when it returns a falsy value

lib/datadog/appsec/contrib/graphql/appsec_trace.rb
```diff
@@ -27,20 +27,10 @@ def execute_multiplex(multiplex:)
...
-
-          def active_trace
-            return unless defined?(Datadog::Tracing)
-
-            Datadog::Tracing.active_trace
-          end
-
-          def active_span
-            return unless defined?(Datadog::Tracing)
+          rescue StandardError => e
```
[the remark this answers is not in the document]
y9v: you are actually right, this will also hide exceptions that could be raised in the code of the customer. I am thinking to fully remove this, and add a task to address this [TODO comment](https://github.com/DataDog/dd-trace-rb/blob/06aa9aceb804a42435f2ff4ad9970f55688713e4/lib/datadog/appsec/instrumentation/gateway.rb#L40).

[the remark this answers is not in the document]
y9v: @TonyCTHsu what do you think, is it ok to leave this exception handling out of this PR?

spec/datadog/appsec/contrib/graphql/integration_spec.rb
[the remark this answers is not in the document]
p-datadog: I personally type `expect(xxx).to be true` because it saves me two shift presses with pinky but as far as reading the code I don't see any difference between `be true` and `eq(true)`. Semantically I believe the two are identical.

y9v:
Merging this, the failing system test will be fixed by https://github.com/DataDog/system-tests/pull/2988

PR #3890, opened 2024-09-04
Add ostruct to Gemfile for Ruby >= 3.3.5

y9v:
This addresses the deprecation warning that `ostruct` will not be part of standard library since ruby 3.5

**What does this PR do?**
This PR adds `ostruct` gem to Gemfile for Ruby version 3.3.5 and higher.

**Motivation:**
Failing tests due to a deprecation warning:
https://github.com/DataDog/dd-trace-rb/actions/runs/10703793866/job/29675192240?pr=3887

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -62,6 +62,8 @@ if RUBY_VERSION >= '2.6.0'
   gem 'rubocop-rspec', ['~> 2.20', '< 2.21'], require: false
 end
 
+gem 'ostruct' if RUBY_VERSION >= '3.3'
+
 # Optional extensions
 # TODO: Move this to Appraisals?
 # dogstatsd v5, but lower than 5.2, has possible memory leak with datadog.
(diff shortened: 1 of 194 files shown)


Gemfile
```diff
@@ -62,6 +62,8 @@ if RUBY_VERSION >= '2.6.0'
   gem 'rubocop-rspec', ['~> 2.20', '< 2.21'], require: false
 end
 
+gem 'ostruct' if RUBY_VERSION >= '3.3.5'
```
ivoanjo, line 65: Minor: I suggest maybe doing `>= 3.3` instead of `3.3.5`.

My thinking is that we're not very consistent in upgrading to point releases in our different test/development environments, and on the off chance we have some weird issue from this, it may be a bit harder to diagnose if only some of the 3.3 series gets this new version.

y9v: fixed here [261b590](https://github.com/DataDog/dd-trace-rb/pull/3890/commits/261b59017ec0a670e3aea1e99a901b4ced0bdda0) (force pushed)

y9v:
@ivoanjo thanks! I haven't thought about this.

ivoanjo (approved):
:+1: LGTM

PR #3902, opened 2024-09-11
[NO-TICKET] Add integration testing for crashtracker component reporting via uds

ivoanjo:
**What does this PR do?**

This PR adds crash report testing via uds. I've borrowed from the profiler's `http_transport.spec` which already had uds testing.

**Motivation:**

Improve test coverage.

**Additional Notes:**

I chose to copy-paste a bit of the existing crash test code, rather than trying to be very clever in factoring it out.

**How to test the change?**

Validate CI is green!

### DIFF
--- a/spec/datadog/core/crashtracking/component_spec.rb
+++ b/spec/datadog/core/crashtracking/component_spec.rb
@@ -329,13 +329,14 @@ RSpec.describe Datadog::Core::Crashtracking::Component, skip: !CrashtrackingHelp
 
           Datadog.configure do |c|
             c.agent.host = 'google.com'
+            c.agent.port = '12345'
           end
 
           expect_in_fork do
             expect(described_class).to have_received(:_native_start_or_update_on_fork).with(
               hash_including(
                 action: :update_on_fork,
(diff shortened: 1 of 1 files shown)


spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -329,13 +377,14 @@
 
           Datadog.configure do |c|
             c.agent.host = 'google.com'
+            c.agent.port = '12345'
```
TonyCTHsu, line 380:
```suggestion
            c.agent.port = 12345
```

TonyCTHsu (approved):

spec/datadog/core/crashtracking/component_spec.rb
ivoanjo: I also spotted that it's documented as being a string (https://github.com/DataDog/dd-trace-rb/blob/bc841812cbc73fdba7b798bfcf74140d6df3da85/lib/datadog/core/configuration/settings.rb#L63) although the agent settings resolver accepts both ints and strings.

I'll apply the suggestion anyway :)

PR #3905, opened 2024-09-11
DEBUG-2334 Dynamic Instrumentation Serializer component

p-datadog:
**What does this PR do?**

Adds the Serializer component which is responsible for converting arbitrary Ruby objects to object trees made up of basic types.

**Motivation:**

Ruby DI implementation

**Additional Notes:**

**How to test the change?**
Unit tests only in this PR.

### DIFF
--- a/lib/datadog/di/serializer.rb
+++ b/lib/datadog/di/serializer.rb
@@ -1,0 +1,168 @@
+# frozen_string_literal: true
+
+require_relative "redactor"
+
+module Datadog
+  module DI
+    # Serializes captured snapshot to primitive types, which are subsequently
+    # serialized to JSON and sent to the backend.
+    #
+    # This class performs actual removal of sensitive values from the
+    # snapshots. It uses Redactor to determine which values are sensitive
+    # and need to be removed.
+    #
+    # Serializer normally ought not to invoke user (application) code,
+    # to guarantee predictable performance. However, objects like ActiveRecord
+    # models cannot be usefully serialized into primitive types without
+    # custom logic (for example, the attributes are more than 3 levels
+    # down from the top-level object which is the default capture depth,
+    # thus they won't be captured at all). To accommodate complex objects,
+    # there is an extension mechanism implemented permitting registration
+    # of serializer callbacks for arbitrary types. Applications and libraries
+    # definining such serializer callbacks should be very careful to
+    # have predictable performance and avoid exceptions and infinite loops
+    # and other such issues.
+    #
+    # All serialization methods take the names of the variables being
+    # serialized in order to be able to redact values.
+    #
+    # @api private
+    class Serializer
+      def initialize(settings, redactor)
+        @settings = settings
+        @redactor = redactor
+      end
+
+      attr_reader :settings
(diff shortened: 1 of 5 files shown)


lib/datadog/di/serializer.rb
```diff
@@ -0,0 +1,168 @@
...
+            serialized.update(entries: entries)
+          end
+        else
+          if depth < 0
+            serialized.update(notCapturedReason: "depth")
+          else
+            fields = {}
+            max = settings.dynamic_instrumentation.max_capture_attribute_count
+            cur = 0
+            value.instance_variables.each do |ivar|
```
y9v, line 142: just out of curiosity: we are not serializing class variables since they are not frequently used?

y9v (approved):

lib/datadog/di/serializer.rb
```diff
@@ -0,0 +1,168 @@
...
+      # arguments to keyword arguments ("arg1", "arg2", ...) and ensure
+      # the positional arguments are listed first.
+      def serialize_args(args, kwargs)
+        counter = 0
+        combined = args.each_with_object({}) do |value, c|
+          counter += 1
+          # Conversion to symbol is needed here to put args ahead of
+          # kwargs when they are merged below.
+          c[:"arg#{counter}"] = value
+        end.update(kwargs)
```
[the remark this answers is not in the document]
p-datadog: Personal preference, I suppose. I always thought `merge!` was a Rails extension for quite a while but it appears to have been added to core language in 1.8.0. `update` also does not require the use of pinky to hit the shift key.

lib/datadog/di/serializer.rb
```diff
@@ -0,0 +1,168 @@
...
+        serialize_vars(combined)
+      end
+
+      # Serializes variables captured by a line probe.
+      #
+      # These are normally local variables that exist on a particular line
+      # of executed code.
+      def serialize_vars(vars)
+        vars.map do |k, v|
```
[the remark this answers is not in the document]
p-datadog: Changed, is this what you had in mind?

lib/datadog/di/serializer.rb
y9v: no, I mean class variables that start with `@@` (I don't see them being used often though, due to thread-safety issues).

Rails uses them in some places:
https://github.com/rails/rails/blob/95e390156142f94649f11f803eb05cd2bb558e7d/actionview/lib/action_view/digestor.rb#L7
https://github.com/rails/rails/blob/95e390156142f94649f11f803eb05cd2bb558e7d/railties/lib/rails/railtie/configuration.rb#L9
https://github.com/rails/rails/blob/95e390156142f94649f11f803eb05cd2bb558e7d/railties/lib/rails/railtie.rb#L207

p-datadog: Right. What I was saying was I think it makes sense to serialize them if and only if the value being serialized is a `Class`.

y9v: yes, you are right. Thanks! This is all quite amazing stuff really

PR #3910, opened 2024-09-13
Prevent telemetry deadlock with no-op

TonyCTHsu:
**What does this PR do?**

It is still far too dangerous for `Telemetry::Logger` to hit a deadlock. This is the risk should be mitigated especially for telemetry which is considered to be a kind of best effort utility.

### DIFF
--- a/lib/datadog/core/telemetry/logger.rb
+++ b/lib/datadog/core/telemetry/logger.rb
@@ -31,7 +31,7 @@ module Datadog
             if components && components.telemetry
               components.telemetry
             else
-              Datadog.logger.error(
+              Datadog.logger.warn(
                 'Fail to send telemetry log before components initialization or within components lifecycle'
               )
               nil
--- a/lib/datadog/core/telemetry/logger.rb
+++ b/lib/datadog/core/telemetry/logger.rb
@@ -25,7 +25,8 @@ module Datadog
           private
 
           def instance
-            # `allow_initialization: false` to prevent deadlock from components lifecycle
+            # `allow_initialization: false` would avoid referencing the components via `safely_synchronize` (mutex)
+            # which could cause deadlock during components initialization.
             components = Datadog.send(:components, allow_initialization: false)
 
             if components && components.telemetry
--- a/lib/datadog/core/telemetry/logger.rb
+++ b/lib/datadog/core/telemetry/logger.rb
@@ -25,8 +25,14 @@ module Datadog
           private
 
           def instance
-            # `allow_initialization: false` would avoid referencing the components via `safely_synchronize` (mutex)
-            # which could cause deadlock during components initialization.
+            # Component initialization uses a mutex to avoid having concurrent initialization.
+            # Trying to access the telemetry component during initialization (specifically:
+            # from the thread that's actually doing the initialization) would cause a deadlock,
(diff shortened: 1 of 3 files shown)


lib/datadog/core/telemetry/logger.rb
```diff
@@ -8,7 +8,7 @@ module Telemetry
       # Report telemetry logs via delegating to the telemetry component instance via mutex.
```
ivoanjo, line 8: Minor: :arrow_up: There's a typo on "INTRENAL"

lib/datadog/core/telemetry/logger.rb
```diff
@@ -8,7 +8,7 @@ module Telemetry
...
       #
       # IMPORTANT: Invoking this method during the lifecycle of component initialization will
-      # cause a non-recoverable deadlock
+      # be no-op instead.
```
ivoanjo, line 11: While causing a deadlock is really bad, I think we should plan to fix the underlying problem. E.g. I think the core should provide to components the service of making telemetry available very early on; it seems reasonable for components to want to report data during initialization.

lib/datadog/core/telemetry/logger.rb
```diff
@@ -25,7 +25,17 @@ def error(description)
           private
 
           def instance
-            Datadog.send(:components).telemetry
+            # `allow_initialization: false` to prevent deadlock from components lifecycle
+            components = Datadog.send(:components, allow_initialization: false)
```
ivoanjo, line 29: The reason for this deadlock is quite subtle; I would suggest writing a bit more here to explain what's going on and why we're doing this (so in the future other folks don't need to "reverse engineer" what happened here)

anmarchenko (approved):
a very nice idea, thanks!

lib/datadog/core/telemetry/logger.rb
```diff
@@ -25,7 +25,17 @@ def error(description)
...
+
+            if components && components.telemetry
+              components.telemetry
+            else
+              Datadog.logger.error(
+                'Fail to send telemetry log before components initialization or within components lifecycle'
+              )
```
ivoanjo, line 36: This will be seen by customers -- do we want to perhaps lower this to warn or even lower? It's not like they'll be able to do anything about this issue, since this is a bug on our side

ivoanjo (commented):
Left a few notes!

lib/datadog/core/telemetry/logger.rb
```diff
@@ -25,7 +25,18 @@ def error(description)
           private
 
           def instance
-            Datadog.send(:components).telemetry
+            # `allow_initialization: false` would avoid referencing the components via `safely_synchronize` (mutex)
+            # which could cause deadlock during components initialization.
+            components = Datadog.send(:components, allow_initialization: false)
```
ivoanjo, line 30: I still suspect the current explanation requires quite a bit of context to understand. Perhaps something like

"Component initialization uses a mutex to avoid having concurrent initialization. Trying to access the telemetry component during initialization (specifically: from the thread that's actually doing the initialization) would cause a deadlock, since accessing the components would try to recursively lock the mutex.

To work around this, we use allow_initialization: false to avoid triggering this issue.

The downside is: this leaves us unable to report telemetry during component initialization."

Something like this?

PR #3912, opened 2024-09-16
DEBUG-2334 dynamic instrumentation Probe class

p-datadog:
**What does this PR do?**

Adds the Probe class which encapsulates probe data and general state such as the rate limiter. Also adds exception classes (all of them that I have currently defined to avoid conflicts during rebasing).

**Motivation:**

Dynamic instrumentation implementation.

**Additional Notes:**

**How to test the change?**

Unit tests at this time.

### DIFF
--- a/lib/datadog/di/probe.rb
+++ b/lib/datadog/di/probe.rb
@@ -1,0 +1,128 @@
+# frozen_string_literal: true
+
+require_relative "error"
+require_relative "../core/rate_limiter"
+
+module Datadog
+  module DI
+    # Encapsulates probe information (as received via remote config)
+    # and state (e.g. whether the probe was installed, or executed).
+    #
+    # It is possible that remote configuration will specify an unsupported
+    # probe type or attribute, due to new DI functionality being added
+    # over time. We want to have predictable behavior in such cases, and
+    # since we can't guarantee that there will be enough information in
+    # a remote config payload to construct a functional probe, ProbeBuilder
+    # and remote config code must be prepared to deal with exceptions
+    # raised by Probe constructor in particular. Therefore, Probe constructor
+    # will raise an exception if it determines that there is not enough
+    # information (or confilcting information) in the arguments to create a
+    # functional probe, and upstream code is tasked with not spamming logs
+    # with notifications of such errors (and potentially limiting the
+    # attempts to construct probe from a given payload).
+    #
+    # Note that, while remote configuration provides line numbers as an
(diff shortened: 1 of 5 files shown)


lib/datadog/di/probe.rb
```diff
@@ -0,0 +1,128 @@
...
+    # Note that, while remote configuration provides line numbers as an
+    # array, the only supported line number configuration is a single line
+    # (this is the case for all languages currently). Therefore Probe
+    # only supports one line number, and ProbeBuilder is responsible for
+    # extracting that one line number out of the array received from RC.
+    #
+    # Note: only some of the parameter/attribute values are currently validated.
+    #
+    # @api private
+    class Probe
```
TonyCTHsu, line 33: I am interested to understand how the caller is going to handle various exception from this object.

lib/datadog/di/probe.rb
```diff
@@ -0,0 +1,128 @@
...
+      attr_reader :template
+
+      # For internal DI use only
+      attr_reader :rate_limiter
+
+      def capture_snapshot?
+        @capture_snapshot
+      end
+
+      # Returns whether the probe is a line probe.
```
TonyCTHsu, line 86: There is a code smell of polymorphism (line probe and method probe). 

Perhaps the constructor can have distinguish from the input and build different probes accordingly.

TonyCTHsu (approved):

y9v (approved):

lib/datadog/di/probe.rb
p-datadog: The remote config processor and the callbacks invoked as a result of execution of instrumented code will have blanket rescues of all exceptions. Probe construction exceptions will be handled by the rescue in remote config processor. I also added "exception boundaries" section to the RFC describing this.

lib/datadog/di/probe.rb
p-datadog: There are many dimensions of (sub)types of probes. The distinction between line and method probes is codified in this PR because I implement both types, but there are also log/metric/span decoration/trigger probes (not broken out because for initial implementation I will only support log probes), enriched/non-enriched probes that vary in what they capture, for method probes there is further distinction of capturing values on entry or exit. Trigger probes are an actively-discussed/prototyped feature in other languages. Creating multiple probe classes would not only require a non-trivial amount of effort to map existing probe functionality into a new domain, it would also cause continuous additional effort to fit new (and currently unknown) probe types and DI functionality into this domain.

PR #3918, opened 2024-09-17
Fix Yard Github action

TonyCTHsu:
**What does this PR do?**

The PR fixes the Github action for generating yard doc and uploaded to Github Pages.

It was broken due to: https://github.com/DataDog/dd-trace-rb/pull/3710, which I was trying to reduce the size of gemfile by removing `redcarpet`

To address this: I removed `yard` gem entirely from our Gemfile and migrate to a minimized Gemfile (`yard.gemfile`) under `tools` and updated the Github action with this dependency.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,7 +4,6 @@ require 'rubocop/rake_task' if Gem.loaded_specs.key? 'rubocop'
 require 'standard/rake' if Gem.loaded_specs.key? 'standard'
 require 'rspec/core/rake_task'
 require 'rake/extensiontask'
-require 'yard'
 require 'os'
 if Gem.loaded_specs.key? 'ruby_memcheck'
   require 'ruby_memcheck'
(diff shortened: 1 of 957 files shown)


bouwkast (approved):
Nice! I'm not a ruby guild member so my review ain't worth much but this looks good to me thanks for fixing it @TonyCTHsu!

y9v (approved):

Rakefile
```diff
@@ -350,17 +349,6 @@ if defined?(RuboCop::RakeTask)
...
 
-YARD::Rake::YardocTask.new(:docs) do |t|
-  # Options defined in `.yardopts` are read first, then merged with
-  # options defined here.
-  #
-  # It's recommended to define options in `.yardopts` instead of here,
-  # as `.yardopts` can be read by external YARD tools, like the
-  # hot-reload YARD server `yard server --reload`.
-
-  t.options += ['--title', "datadog #{Datadog::VERSION::STRING} documentation"]
```
marcotc, line 361: Looks like we removed this one, but didn't add it anywhere else.

This part is responsible for the title at the top of this page https://datadoghq.dev/dd-trace-rb/_index.html.

We can try moving it to the yard rakefile instead.

TonyCTHsu: It has been move to https://github.com/DataDog/dd-trace-rb/pull/3918/files#diff-738cc9f696f5a20181d325f04bc2449722c513cbdb6cda88746d6bc831c73514

Hence, calling https://github.com/DataDog/dd-trace-rb/pull/3918/files#diff-0fb7d6cd1084c2d17c22be8545371174ae11bae55b94b86033279e6bcf811060R39

PR #3927, opened 2024-09-19
Add auto-update gemfiles workflow

quinna-h:
**What does this PR do?**

Automates updating all gemfiles with latest versions of gems, a rake task created by @TonyCTHsu in https://github.com/DataDog/dd-trace-rb/pull/3916.
This creates the workflow for automatically generated PRs.

Generated PR: https://github.com/DataDog/dd-trace-rb/pull/3923

Future items:

Expand the matrix in Matrixfile to include latest for all gems, beyond stripe, opensearch, elasticsearch.

**Motivation:**

**Additional Notes:**

**How to test the change?**


.github/workflows/auto-update-gemfiles.yml
```diff
@@ -0,0 +1,169 @@
...
+# This action cannot be skipped altogether because it is a mandatory status check.
+# Instead we conditionally skip it at job level, instead of workflow level.
+on:
+  schedule:
+    - cron: '0 0 * * 0' # Every Sunday at midnight
+  workflow_dispatch:
+
+# Ensure obsolete job is cancelled if another commit is pushed to the same branch.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
y9v, line 12: how does `||` operator work here? can `github.event.pull_request.number` be falsey?

y9v (approved):

.github/workflows/auto-update-gemfiles.yml
TonyCTHsu: The workflow cannot be triggered by `pull_request`

.github/workflows/auto-update-gemfiles.yml
```diff
@@ -0,0 +1,169 @@
...
+  cancel-in-progress: true
+
+jobs:
+  build:
+    runs-on: ubuntu-22.04
+    permissions:
```
TonyCTHsu, line 18: Do we need those permissions?

.github/workflows/auto-update-gemfiles.yml
```diff
@@ -0,0 +1,169 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read # read secrets
+      pull-requests: write
+      contents: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Download artifacts for Ruby 2.5
```
TonyCTHsu, line 81: Refactor with `merge-multiple` and `pattern` 

https://github.com/actions/download-artifact/tree/v4/?tab=readme-ov-file#download-multiple-filtered-artifacts-to-the-same-directory

.github/workflows/auto-update-gemfiles.yml
```diff
@@ -0,0 +1,169 @@
...
+          path: gemfiles
+      
+      # ADD NEW RUBIES HERE
+      
+      - run: git diff
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{secrets.GITHUB_TOKEN}}
```
TonyCTHsu, line 154: `token` seems to be redundant.

.github/workflows/auto-update-gemfiles.yml
```diff
@@ -0,0 +1,169 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          name: 'datadog-gem-jruby-9.4-gha${{ github.run_id }}'
+          path: gemfiles
+      
+      # ADD NEW RUBIES HERE
+      
+      - run: git diff
+
+      - name: Create Pull Request
```
TonyCTHsu, line 151: Since we are having a fixed branch name, it would be helpful to clarify with a comment

> The default behaviour is to create a pull request that will be continually updated with new changes until it is merged or closed. 

Otherwise, the behaviour becomes unclear after one week from now without merging the PR.

.github/workflows/auto-update-gemfiles.yml
```diff
@@ -0,0 +1,169 @@
...
+        uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{secrets.GITHUB_TOKEN}}
+          branch: update-latest-gemfiles
+          commit-message: "update latest gemfiles"
```
TonyCTHsu, line 156: Add `${{github.run_id}}` into commit message to make it traceable

.github/workflows/auto-update-gemfiles.yml
```diff
@@ -0,0 +1,169 @@
...
+          delete-branch: true
+          base: master
+          title: 'Update Latest Gemfiles'
+          body: |
```
TonyCTHsu, line 160: No need to follow the template structure, refining the body: 

Describing itself and origin, for example:
> This is an Autogenerated PR from blablabla workflow

Explaining the intention, for example:
> This PR updates our integration dependencies being tested so we can keep up with the community development

Call to action, for example
> Merge it if green, if not .....


In addition: apply the correct label and staff to review the PR

TonyCTHsu (changes requested):

.github/workflows/auto-update-gemfiles.yml
TonyCTHsu: yeah, the workflow cannot triggered by pull request, the code seems to be copy & pasted without changes.

.github/workflows/auto-update-gemfiles.yml
bouwkast: Hmm _maybe_ nice catch @TonyCTHsu : "The token that the action will use to create and update the pull request." vs "The token input defaults to the repository's GITHUB_TOKEN."

The description of `token` clashes with the further description in the docs: https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#token

@quinna-h could we try removing this and seeing what happens we may be able to remove this

.github/workflows/auto-update-gemfiles.yml
bouwkast: Nice!

.github/workflows/auto-update-gemfiles.yml
bouwkast: Thanks Tony, I _think_ it needs `contents: write` for artifact uploading, but I may be wrong. @quinna-h could we try removing both `actions: read` and `contents: write` and seeing if it still works?

TonyCTHsu: I was able to upload artifacts without them: https://github.com/DataDog/dd-trace-rb/pull/3926/files#diff-50517d6517bd2924edda04f5d4e68e631ddc36d92cff5720e54fbde4f2a46c50R59

quinna-h: It seems to work without `actions: read` and `contents: write` , but I think we still need `contents: read`

TonyCTHsu (changes requested):

TonyCTHsu (approved):

PR #3931, opened 2024-09-20
Split single step instrument entry point

TonyCTHsu:
**What does this PR do?**

For single step instrumentation, use a different entry point that would rescue `LoadError` and not crash the process.

### DIFF
--- a/lib/datadog/single_step_instrument.rb
+++ b/lib/datadog/single_step_instrument.rb
@@ -8,5 +8,5 @@
 begin
   require_relative 'auto_instrument'
 rescue StandardError, LoadError => e
-  warn "Single step instrumentation failed: #{e.class}, #{e.message}"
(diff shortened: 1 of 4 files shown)


lib/datadog/single_step_instrument.rb
```diff
@@ -0,0 +1,12 @@
...
+
+#
+# Entrypoint file for single step instrumentation.
+#
+# This file's path is private. Do not reference this file.
+#
+begin
+  require_relative 'auto_instrument'
+rescue StandardError, LoadError => e
+  warn "Single step instrumentation failed: #{e.class}, #{e.message}"
```
marcotc, line 11: I think we need the stack trace too to help us debug the issue.

marcotc (approved):

PR #3937, opened 2024-09-20
Add missing GraphQL and Hanami tests to CI

marcotc:
This PR adds tests that were added as Rake tasks, but not to the Matrixfile.

It also updates the `docs/DevelopmentGuide.md` to improve a bit on how to add this information when writing new tests.

### DIFF
--- a/spec/datadog/tracing/contrib/graphql/tracing_patcher_spec.rb
+++ b/spec/datadog/tracing/contrib/graphql/tracing_patcher_spec.rb
@@ -35,17 +35,5 @@ RSpec.describe Datadog::Tracing::Contrib::GraphQL::TracingPatcher do
         end
       end
     end
-
-    context 'when given something else' do
-      before { remove_patch!(:graphql) }
-
-      it do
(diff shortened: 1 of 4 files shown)


spec/datadog/tracing/contrib/graphql/tracing_patcher_spec.rb
```diff
@@ -35,17 +35,5 @@
...
-      before { remove_patch!(:graphql) }
-
-      it do
-        expect_any_instance_of(Datadog::Core::Logger).to receive(:warn).with(/Unable to patch/)
-
-        Datadog.configure do |c|
-          c.tracing.instrument :graphql, with_deprecated_tracer: true, schemas: [OpenStruct.new]
-        end
-      end
-    end
```
marcotc, line 49: I've removed this one GraphQL test because it adds little value (testing a warning), but to make it not flaky I need to split it out into it's own test file, Rake task, and execution group, because it pollutes the current application with a failed instrumentation.

TonyCTHsu (approved):

PR #3939, opened 2024-09-23
[PROF-10588] Support GVL profiling preview on Ruby 3.2

ivoanjo:
**What does this PR do?**

This PR builds atop #3929 that added support for GVL profiling for Ruby 3.3+ and makes GVL profiling also work for Ruby 3.2 .

Supporting GVL profiling on Ruby 3.2 needed special additional work. That's because while in Ruby 3.2 we have the GVL instrumentation API giving us the events we need to profile the GVL, we're missing:

1. Getting the Ruby thread `VALUE` as an argument in GVL instrumentation API events
2. The `rb_internal_thread_specific` API that allows us to attach in a thread-safe way data to Ruby thread objects

Both 1 and 2 were only introduced in Ruby 3.3, and our implementation of GVL profiling relies/relied on them.

This PR... reimplements 1 & 2 in an alternative way, allowing us to keep our existing design for 3.3+, while also supporting the older Ruby version.

I've split it into two commits:

 i. Abstracting access and management of 1 & 2 into a new set of files (`gvl_profiling_helper.c`/`gvl_profiling_helper.h`). These new files are zero-overhead abstractions for most situations.

ii. Implementing 1 & 2 for Ruby 3.2.

**Motivation:**

We believe GVL profiling is quite an important observability feature for Ruby, and thus we want to support it on all Ruby versions that provide the GVL instrumentation API.

**Additional Notes:**

To solve 1, we're using native level thread-locals (GCC's `__thread`) to keep a pointer to the underlying Ruby `rb_thread_t` structure.

This is more complex than than "just keep it on a thread-local" because:

a) Ruby reuses native threads. When a Ruby thread dies, Ruby keeps the underlying native thread around for a bit, and if another Ruby thread is born very quickly after the previous one, Ruby will reuse the native thread and attach it to the new Ruby thread.

   To avoid incorrectly reusing the thread-locals, we install an event hook on Ruby thread start, and make sure to clean any native thread-locals when a new thread stats.

b) Some of the GVL instrumentation API events are emitted while the thread does not have the GVL and so we need to be careful when we can and cannot read VM information.

   Thus, we only initialize the thread-local during the `RUBY_INTERNAL_THREAD_EVENT_RESUMED` which is emitted while the thread owns the GVL.

c) Since we don't get the current thread in events, we need to get a bit... creative. Thus, what we do is in `RUBY_INTERNAL_THREAD_EVENT_RESUMED`, because we know the current thread MUST own the GVL, we read from the internal Ruby VM state which thread is the GVL owner to find the info we need.

With a + b + c together we are able to keep a pointer to the underlying `rb_thread_t` up-to-date in a native thread local, thus replacing the need to get a `VALUE thread` as an argument.

To solve 2, we rely on an important observation: there's a `VALUE stat_insn_usage` field inside `rb_thread_t` that's unused and seems to have effectively been forgotten about.

There's nowhere in the VM code that's writing or reading it (other than marking it for GC), and not even git history reveals a time where this field was used. I could not find any other references to this field anywhere else. Thus, we make use of this field to store the information we need, as a replacement for `rb_internal_thread_specific`.

Since presumably Ruby 3.2 will never see this field either removed or used during its remaining maintenance release period this should work fine, and we have a nice clean solution for 3.3+.

**How to test the change?**

Happily, with the changes on this PR, the existing test coverage we added for GVL profiling on 3.3 is also green on 3.2! :)

And here's a few screenshots of the same test scripts I've included in #3929, now with Ruby 3.2:

* [https://github.com/ivoanjo/gvl-tracing/blob/master/examples/example1.rb](https://github.com/ivoanjo/gvl-tracing/blob/master/examples/example1.rb?rgh-link-date=2024-09-19T10%3A27%3A39Z): image

* [https://github.com/ivoanjo/gvl-tracing/blob/master/examples/example4.rb](https://github.com/ivoanjo/gvl-tracing/blob/master/examples/example4.rb?rgh-link-date=2024-09-19T10%3A27%3A39Z): image

* [https://github.com/ivoanjo/gvl-tracing/blob/master/examples/rubykaigi2023/rk-example2.rb](https://github.com/ivoanjo/gvl-tracing/blob/master/examples/rubykaigi2023/rk-example2.rb?rgh-link-date=2024-09-19T10%3A27%3A39Z): image

* [https://github.com/ivoanjo/gvl-tracing/blob/master/examples/rubykaigi2023/rk-example4.rb](https://github.com/ivoanjo/gvl-tracing/blob/master/examples/rubykaigi2023/rk-example4.rb?rgh-link-date=2024-09-19T10%3A27%3A39Z): image

* [https://github.com/ivoanjo/gvl-tracing/blob/master/examples/rubykaigi2023/rk-example6.rb](https://github.com/ivoanjo/gvl-tracing/blob/master/examples/rubykaigi2023/rk-example6.rb?rgh-link-date=2024-09-19T10%3A27%3A39Z): image

### DIFF
--- a/docs/ProfilingDevelopment.md
+++ b/docs/ProfilingDevelopment.md
@@ -151,8 +151,7 @@ See comments at the top of `collectors_thread_context.c` for an explanation on h
 
 Profiling the Ruby Global VM Lock (GVL) works by using the [GVL instrumentation API](https://github.com/ruby/ruby/pull/5500).
 
-This API currently only works on Ruby 3.2+, although there were a few changes and refinements on Ruby 3.3+ and as of this writing
-(September 2024) we _only_ support Ruby 3.3+.
+This API currently only works on Ruby 3.2+.
 
 These blog posts are good starting points to understand this API:
 
@@ -240,7 +239,7 @@ The weirdest piece of this puzzle is done in the `ThreadContext` collector. Beca
 
 Instead, we ask Ruby to hold the data we need for us using the `rb_internal_thread_specific` API. This API provides
 an extremely low-level and limited thread-local storage mechanism, but one that's thread-safe and thus a good match
-for our needs. (This API is only available on Ruby 3.3+; to support Ruby 3.2 we'll need to figure out an alternative.)
+for our needs. (This API is only available on Ruby 3.3+; to support Ruby 3.2 we use an alternative implementation.)
 
 So, here's how this works: the first time the `ThreadContext` collector sees a thread (e.g. when it creates a context
 for it), it uses this API to tag this thread as a thread we want to know about:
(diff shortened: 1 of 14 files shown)


marcotc (approved):
🔥

ivoanjo:
Update: I'm still finishing up some testing, and if all goes well I plan to merge this in next week.

p-datadog (approved):
https://github.com/DataDog/dd-trace-rb/pull/3929 has markdown docs describing limitations of 3.2, should that be updated in this PR now that 3.2 is usable?

ivoanjo:
Good catch, I've updated the development docs in 97751d7a684edc7ee7ea956280155475dfeb9462 to reflect that we now support 3.2 (and how), based on the description of this PR.

p-datadog (approved):

PR #3942, opened 2024-09-24
DEBUG-2334 Dynamic Instrumentation code tracker component

p-datadog:
**What does this PR do?**

Adds the code tracker component. This is responsible for tracking the mapping from source file path to RubyVM::InstructionSequence object used for setting targeted trace points.

**Motivation:**
Efficient instrumentation of lines.

**Additional Notes:**
There will be further functionality added to CodeTracker later to instrument loaded code (requires the instrumentation component that hasn't been PR'ed yet).

**How to test the change?**
Unit tests at this time.


lib/datadog/di/code_tracker.rb
```diff
@@ -0,0 +1,152 @@
...
+    #
+    # @api private
+    class CodeTracker
+      def initialize
+        @registry = Hash.new
+        @trace_point_lock = Mutex.new
+        @registry_lock = Hash.new
+      end
+
+      def start
```
marcotc, line 25: If this method is not called at a high frequency (which it doesn't seem like it is), I believe to make it truly tread safe we have to wrap the whole execution in a mutex.

lib/datadog/di/code_tracker.rb
```diff
@@ -0,0 +1,152 @@
...
+    # recreated when the DI component is created.
+    #
+    # @api private
+    class CodeTracker
+      def initialize
+        @registry = Hash.new
+        @trace_point_lock = Mutex.new
+        @registry_lock = Hash.new
+      end
+
```
marcotc, line 24: I suggest adding, even if a very simple, method level documentation to the `start` method.

p-datadog: Yes, i will type up a comment.

lib/datadog/di/code_tracker.rb
p-datadog: I already have a lock taken for all reads and writes of instance variables and  I believe  I account for e.g. state changing during constructor execution. Can you elaborate on what situation you think is not properly covered by existing locking?

spec/datadog/di/code_tracker_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+
+  describe "#iseqs_for_path" do
+    around do |example|
+      tracker.start
+
+      load File.join(File.dirname(__FILE__), "code_tracker_test_class_1.rb")
+      load File.join(File.dirname(__FILE__), "code_tracker_test_class_2.rb")
+      load File.join(File.dirname(__FILE__), "code_tracker_test_class_3.rb")
+      load File.join(File.dirname(__FILE__), "code_tracker_test_classes", "code_tracker_test_class_1.rb")
+      expect(tracker.send(:registry).each.to_a.length).to eq(4)
```
y9v, line 77: I think this could be simplified to `tracker.send(:registry).length`?

y9v (approved):

spec/datadog/di/code_tracker_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+RSpec.describe Datadog::DI::CodeTracker do
+  let(:tracker) do
+    described_class.new
+  end
+
+  describe ".new" do
+    it "creates an instance" do
+      expect(tracker).to be_a(described_class)
+    end
+  end
```
Strech, line 12: Sorry for asking, but what a point of this test? Technically we are testing here that Ruby is going to instantiate a class if we call new, but do we really challenge it?

spec/datadog/di/code_tracker_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+    it "tracks loaded files" do
+      # The expectations appear to be lazy-loaded, therefore
+      # we need to invoke the same expectation before starting
+      # code tracking as we'll be using later in the test.
+      expect(tracker.send(:registry)).to be_empty
+      tracker.start
+      # Should still be empty here.
+      expect(tracker.send(:registry)).to be_empty
+      load File.join(File.dirname(__FILE__), "code_tracker_test_class_1.rb")
+      expect(tracker.send(:registry).each.to_a.length).to eq(1)
```
Strech, line 28: Can we do just length? Since we are looking into guts, we assume that it behave like "something enumerable", do we need to wrap it as one and make it also array?

```suggestion
      expect(tracker.send(:registry).length).to eq(1)
```

spec/datadog/di/code_tracker_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+
+      path = tracker.send(:registry).each.to_a.first.first
```
Strech, line 30: I think `to_a` should be enough to convert it into pairs

```suggestion
      path = tracker.send(:registry).to_a.first.first
```

lib/datadog/di/code_tracker.rb
```diff
@@ -0,0 +1,152 @@
...
+      # Otherwise all known paths that end in the suffix are returned.
+      # If no paths match, an empty array is returned.
+      def iseqs_for_path(suffix)
+        registry_lock.synchronize do
+          exact = registry[suffix]
+          if exact
+            return [exact]
+          end
+          inexact = []
+          registry.each do |path, iseq|
```
Strech, line 100: May I suggest to make a guard-clause more visual or there are some constraints?

```suggestion
          exact = registry[suffix]
          return [exact] if exact

          inexact = []
          registry.each do |path, iseq|
```

Strech (commented):
I have a small feedback, but nothing critical to say

lib/datadog/di/code_tracker.rb
```diff
@@ -0,0 +1,152 @@
...
+            # meaning either the first character of the suffix is a slash
+            # or the previous character in the path is a slash.
+            # For now only check for forward slashes for Unix-like OSes;
+            # backslash is a legitimate character of a file name in Unix
+            # therefore simply permitting forward or back slash is not
+            # sufficient, we need to perform an OS check to know which
+            # path separator to use.
+            if path.length > suffix.length && (
+              path[path.length - suffix.length - 1] == "/" ||
+              suffix[0] == "/"
```
Strech, line 112: May I ask – can we improve split the if to improve the readability and maybe use variable to give a name to our intents? We anyway is inside the mutex section and performance is probably not highly impacted by extra branching? WDYT?

spec/datadog/di/code_tracker_spec.rb
p-datadog: Concurrent::Map apparently doesn't have a `length` and requires this incantation, but now that I switched to a plain Hash I will get rid of it. Thanks for spotting.

spec/datadog/di/code_tracker_spec.rb
p-datadog: Concurrent::Map apparently doesn't have a length and requires this incantation, but now that I switched to a plain Hash I will get rid of it. Thanks for spotting.

lib/datadog/di/code_tracker.rb
```diff
@@ -0,0 +1,152 @@
...
+          #   not where the loaded code is; use .path on the instruction sequence
+          #   to obtain the location of the compiled code)
+          # .eval_script
+          #
+          # For now just map the path to the instruction sequence.
+          path = tp.instruction_sequence.path
+          registry_lock.synchronize do
+            registry[path] = tp.instruction_sequence
+          end
+        end
```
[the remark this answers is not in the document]
p-datadog: No, it should be emitted once per file.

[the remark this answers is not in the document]
ivoanjo: Wait, in that case, how do we know the correct iseq to target, if 1 file has N iseqs? (Or I may be misunderstanding how this works?)

spec/datadog/di/code_tracker_spec.rb
p-datadog: This test only exercises the constructor. If you object to it I can remove it.

lib/datadog/di/code_tracker.rb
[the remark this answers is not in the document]
p-datadog: There is one iseq per file.

lib/datadog/di/code_tracker.rb
p-datadog: I am not sure what you are asking. The logic of this code fragment is explained in the comment above. What are you suggesting to split exactly?

lib/datadog/di/code_tracker.rb
```diff
@@ -0,0 +1,152 @@
...
+            # backslash is a legitimate character of a file name in Unix
+            # therefore simply permitting forward or back slash is not
+            # sufficient, we need to perform an OS check to know which
+            # path separator to use.
+            if path.length > suffix.length && (
+              path[path.length - suffix.length - 1] == "/" ||
+              suffix[0] == "/"
+            ) && path.end_with?(suffix)
+              inexact << iseq
+            end
```
Strech, line 111: I think something like this. Not sure where do we stand in "comments vs code", but maybe this may be it?

```suggestion
            if path.length > suffix.length
              previous_char = path[path.length - suffix.length - 1]
              inexact << iseq if previous_char == "/" && path.end_with?(suffix)
            end
```

p-datadog: It looks to me like you lost the `suffix[0] == "/"` condition along the way and I personally find your proposed version more difficult to follow because it lacks symmetry where there is symmetry in the logic and inverts cause and effect with the postfix if, thus also putting the action in the middle of the conditions, but I'm happy to commit a rendering you are happy with as long as it is equivalent functionally to what is currently in the diff.

lib/datadog/di/code_tracker.rb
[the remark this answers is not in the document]
ivoanjo: Interesting... I think my mental model was slightly off for this one.

As you pointed out taking for instance

```ruby
# test.rb

def a
  puts "a!"
end

def b
  puts "b!"
end
```

and doing

```
[1] pry(main)> iseq = RubyVM::InstructionSequence.compile(File.read("test.rb"))
[3] pry(main)> puts RubyVM::InstructionSequence.disasm(iseq)
== disasm: #<ISeq:<compiled>@<compiled>:1 (1,0)-(8,3)> (catch: FALSE)
0000 definemethod                           :a, a                     (   2)[Li]
0003 definemethod                           :b, b                     (   6)[Li]
0006 putobject                              :b
0008 leave

== disasm: #<ISeq:a@<compiled>:2 (2,0)-(4,3)> (catch: FALSE)
0000 putself                                                          (   3)[LiCa]
0001 putstring                              "a!"
0003 opt_send_without_block                 <calldata!mid:puts, argc:1, FCALL|ARGS_SIMPLE>
0005 leave                                                            (   4)[Re]

== disasm: #<ISeq:b@<compiled>:6 (6,0)-(8,3)> (catch: FALSE)
0000 putself                                                          (   7)[LiCa]
0001 putstring                              "b!"
0003 opt_send_without_block                 <calldata!mid:puts, argc:1, FCALL|ARGS_SIMPLE>
0005 leave                                                            (   8)[Re]
=> nil
```

e.g. will mean we get it all in one go.

Yet... it seems there's still separate objects for the different iseqs -- there's a `RubyVM::InstructionSequence#each_child` and the object ids of the objects it returns are different from the top-level iseq so they seem like separate objects arranged in a tree, not one object that's presenting different views of itself.

That said, for the usage we'll be making in DI, maybe this extra distinction doesn't matter very much? And I learned something new :)

lib/datadog/di/code_tracker.rb
```diff
@@ -0,0 +1,152 @@
...
+          end
+        end
+
+        trace_point_lock.synchronize do
+          # Since trace point creation itself is not under a lock, see if
+          # another thread created the trace point, in which case we can
+          # disable our trace point and do nothing.
+          if @compiled_trace_point
+            # Disable the local variable, leave instance variable as it is.
+            compiled_trace_point.disable
```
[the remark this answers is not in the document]
ivoanjo: Aaaaah that's super subtle, missed it! May be worth adding a comment or using an explicit enable to make it easier on the readers? (but just a suggestion)

lib/datadog/di/code_tracker.rb
Strech: Sorry, diff was written under high distraction, you're right, added `suffix` check.

```suggestion
            if path.length > suffix.length && path.end_with?(suffix)
              previous_char = path[path.length - suffix.length - 1]

              inexact << iseq if (previous_char == "/" || suffix[0] == "/")
            end
```

How about this variant. I've grouped the path conditions which were guarding the whole `if` statement anyway. I understand that it might exit later than your statement, but now I think it's more readable (I know that's a matter of taste and perception sometimes, so no hard feelings)

spec/datadog/di/code_tracker_spec.rb
Strech: I think it makes sense to keep the test suite slim. I think we don't need to challenge ruby internals considering we have no additional logic to it

p-datadog: Deleted.

p-datadog:
I don't know why coverage for the 3 lines is missing. Those lines are exercised by the tests but perhaps they are not counted due to being under a trace point.

lib/datadog/di/code_tracker.rb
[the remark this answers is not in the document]
p-datadog: Added a note.

lib/datadog/di/code_tracker.rb
```diff
@@ -0,0 +1,147 @@
...
+            # .path (refers to the code location that called the require/eval/etc.,
+            #   not where the loaded code is; use .path on the instruction sequence
+            #   to obtain the location of the compiled code)
+            # .eval_script
+            #
+            # For now just map the path to the instruction sequence.
+            path = tp.instruction_sequence.path
+            registry_lock.synchronize do
+              registry[path] = tp.instruction_sequence
+            end
```
ivoanjo, line 58: I wonder if there's situations where the path will be `nil`? e.g. does this tracepoint only fire on actual files, or would it catch evals and other things where the path may not exist? :thinking:

lib/datadog/di/code_tracker.rb
```diff
@@ -0,0 +1,147 @@
...
+    #
+    # The loaded code is used to target line trace points when installing
+    # line probes which dramatically improves efficiency of line trace points.
+    #
+    # Note that, since most files will only be loaded one time (via the
+    # "require" mechanism), the code tracker needs to be global and not be
+    # recreated when the DI component is created.
+    #
+    # @api private
+    class CodeTracker
```
ivoanjo, line 18: Might be worth documenting what parts of this class are concurrent and why.

e.g.If I understand it well (and I'm speculating/reverse engineering), the concurrency in `registry` is because `iseqs_for_path` may be concurrent with the tracepoint, but usually there will be no concurrency between multiple invocations for `iseqs_for_path` (presumably because remote configuration apply is sequential).

(I don't quite understand when there can be concurrency in `start`/`stop`; given that usually the dd-trace-rb components system takes care of enforcing that starting and stopping things is a sequential operation as well)

ivoanjo (approved):
:+1: Left a few final notes/suggestions/questions, but overall it LGTM

Strech (approved):
well done, my points are resolved 👍🏼

lib/datadog/di/code_tracker.rb
p-datadog: start & stop should not be running at the same time. start/stop could (I suppose) be running while the trace point is invoked. Since starting and stopping should happen once and never in normal usage, I reused the mutexes.

lib/datadog/di/code_tracker.rb
p-datadog: The trace point is indeed called for eval'd code, but `path` will not be nil here because it is synthesized to be `(eval at <file>:<line>)` in these cases. There is also `.absolute_path` which is in fact nil for eval'd code, I changed to using this and filter out eval'd iseqs since I do not see how DI will be able to target such code by file & line.

lib/datadog/di/code_tracker.rb
ivoanjo: I'll admit that if you confirm that start/stop should not be concurrent then I don't quite understand what `trace_point_lock` is protecting :thinking: 

I saw you setup a meeting to discuss this, so let's continue the discussion there. In any case, I don't think the discussion should be a blocker to merging this PR, as we can always tweak this in a later PR.

lib/datadog/di/code_tracker.rb
ivoanjo: :+1: nice, looks good!

PR #3944, opened 2024-09-24
fix: sample on TraceOperation and root span metrics and tags

ZStriker19:
**What does this PR do?**
By changing what the trace operation returns for its metadata by checking the root span as well for it, we can easily sample on all values both in the trace operation (which will be added to the root span before flush) and the root span. 

**Motivation:**
Previously Ruby was not up to spec because only metadata attached to the trace operation was able to be sampled on. This meant that users could possibly see tag or metric values on the root span in the UI but not be able actually sample on them.

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/lib/datadog/tracing/trace_operation.rb
+++ b/lib/datadog/tracing/trace_operation.rb
@@ -166,7 +166,6 @@ module Datadog
         super || (root_span && root_span.get_tag(key))
       end
 
-
       def get_metric(key)
         @metric[key?] || (root_span && root_span.get_metric(key))
       end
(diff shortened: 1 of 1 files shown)


lib/datadog/tracing/trace_operation.rb
```diff
@@ -161,6 +161,23 @@ def resource
         @resource || (root_span && root_span.resource)
       end
 
+      # When retrieving tags or metrics we need to include root span tags for sampling purposes
+      def get_tag(key)
+        super || (root_span && root_span.get_tag(key))
+      end
+
+      def get_metric(key)
+        @metric[key?] || (root_span && root_span.get_metric(key))
```
ZStriker19, line 170:
```suggestion
        super || (root_span && root_span.get_metric(key))
```

PR #3945, opened 2024-09-25
chore: sample before digest

ZStriker19:
**What does this PR do?**
This PR causes a trace to be sampled upon digest creation with the intent of making sampling lazy. The benefit of this approach is that we only create digests right before either propagation or context changes like forking or threading. Therefore it covers multiple exits. The downside is having sampling run as a side effect of creating a digest.

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/lib/datadog/tracing/trace_operation.rb
+++ b/lib/datadog/tracing/trace_operation.rb
@@ -2,7 +2,7 @@
 
 require_relative '../core/environment/identity'
 require_relative '../core/utils'
-
+require_relative 'tracer'
 require_relative 'event'
 require_relative 'metadata/tagging'
 require_relative 'sampling/ext'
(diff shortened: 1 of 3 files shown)


lib/datadog/tracing/trace_operation.rb
```diff
@@ -284,10 +284,14 @@ def flush!
...
       # Used for propagation across execution contexts.
       # Data should reflect the active state of the trace.
+      # DEV-3.0: Sampling is a side effect of generating the digest.
+      # We should move the sample call to inject and right before moving to new contexts(threads, forking etc.)
       def to_digest
         # Resolve current span ID
         span_id = @active_span && @active_span.id
         span_id ||= @parent_span_id unless finished?
+        # sample the trace_operation with the tracer
+        tracer.sample_trace(self) unless sampled?
```
ZStriker19, line 294:
```suggestion
        tracer.sample_trace(self) unless priority_sampled?
```

PR #3946, opened 2024-09-25
chore: sample trace before flush unless already sampled

ZStriker19:
**What does this PR do?**
This PR causes a trace to be sampled during flush with the intent of making sampling lazy. 
**Motivation:**
Lazy sampling allows sampling on tags and resource

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/lib/datadog/tracing/tracer.rb
+++ b/lib/datadog/tracing/tracer.rb
@@ -494,7 +494,11 @@ module Datadog
       def flush_trace(trace_op)
         begin
           trace = @trace_flush.consume!(trace_op)
-          write(trace) if trace && !trace.empty?
+          if trace && !trace.empty?
+            # check if trace is not sampled
+            sample_trace(trace) unless trace.sampled?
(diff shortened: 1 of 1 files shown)


lib/datadog/tracing/tracer.rb
```diff
@@ -494,7 +494,11 @@ def sample_span(trace_op, span)
       def flush_trace(trace_op)
         begin
           trace = @trace_flush.consume!(trace_op)
-          write(trace) if trace && !trace.empty?
+          if trace && !trace.empty?
+            # check if trace is not sampled
+            sample_trace(trace) unless trace.sampled?
```
ZStriker19, line 499:
```suggestion
            sample_trace(trace)
```
sample_trace method checks if the sampling priority is already set

ZStriker19, line 499:
```suggestion
            sample_trace(trace) unless trace.priority_sampled?
```

PR #1224, opened 2020-10-29
Add time_now_provider configuration option

ericmustin:
This PR adds a `time_now_provider` configuration option which allows users to specify a custom current time provider. This option can be set so that a Span's `start_time` and `end_time` accurately reflect wall clock when using the Tracer in environments that change the default time provider, like `Timecop`. 

`Timecop` does monkeypatching of the [base `Time` class ](https://github.com/travisjeffery/timecop/blob/60d1f4a57b0cef9dd4b4d550829535305c9f78c1/lib/timecop/time_extensions.rb#L11) and alias's the `now` method, supplanting it with it's own "mocked" `now`. While most our Span `duration` calculation's rely on Monotonic Clock and so are not effected, by using the "mock"'d `start_time` and `end_time` our traces will not accurately reflect the execution timing details of the underlying span's operations, and they'll appear inaccurate in the UI.

#### Usage

```
Datadog.configure do |c|
  c.time_now_provider = -> { Timecop.now_without_mock_time }
end

# ...

Timecop.freeze(my_custom_time) do
  tracer.trace('web.request', service: 'my-web-site') do
  # Your work here
  end
end
```

In the above example the duration would be the time "# Your work here" takes, and  start_time and end_time would be `my_custom_time`.

PR #3948, opened 2024-09-25
[SDTEST-482] Add get_time_provider setting to avoid timecop mocking the clock_gettime

anmarchenko:
**What does this PR do?**
Adds `get_time_provider` setting similar to `time_now_provider` that provides a workaround for the situation when timecop mocks Process.clock_gettime.

**Motivation:**
Timecop v0.9.9 allows to [mock monotonic clock](https://github.com/travisjeffery/timecop/pull/419), which might affect tests tracing. See the example of this failure [here](https://github.com/DataDog/datadog-ci-rb/pull/242).

**How to test the change?**
Unit tests are provided


TonyCTHsu:
Is this an internal usage from CI gem? Perhaps we should avoid making it public

anmarchenko:
@TonyCTHsu the option to set Time.now provider is public and documented for customers here: 
https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#setting-the-time-provider

In this case, I see no reason to hide this option from public usage: there are customers that used `time_now_provider`, so they could use `get_time_provider` as well

ivoanjo:
`time_now_provider` was added in https://github.com/DataDog/dd-trace-rb/pull/1224 . I guess the intent was to allow better control over span start/end times?

lib/datadog/core/configuration/settings.rb
```diff
@@ -654,6 +654,33 @@ def initialize(*_)
           end
         end
 
+        # The monotonic clock time provider used by Datadog.
```
TonyCTHsu, line 657: I think it is worth adding a comment to mention that `datadog-ci` is using this option to generate traces, not skewed by `Timecop` and we consider this internal (not public).

TonyCTHsu (approved):

delner (commented):
I like the idea of configuring in a cooperative way over forcing a mock. My question is more about the name.

`time_provider` suggests that all time in the library will be provided through this provider? Do we have a means of really enforcing this consistent use? If we use `Time.now` somewhere instead of this provider, we may end up with broken expectations around behavior.

Off the top of my head, I feel like this could be either:

1. Commit to using this more broadly
2. OR rename this to `trace_time_provider` and more narrowly bound its use. But then it doesn't belong in `Core`.

WDYT? (I'm not quite sure what to do about this.)

anmarchenko:
@delner I don't have any preference on this topic, so I think whatever works for tracing/core is fine for me

p-datadog (approved):
Diff LGTM, I have no opinion on the naming/scope.

anmarchenko:
@delner I suggest let's merge it now because it solves a problem for the users, I will leave the question of whether it belongs to Core or Tracing open for now

PR #3950, opened 2024-09-25
DEBUG-2334 upgrade steep & rbs

p-datadog:
**What does this PR do?**

This PR upgrades steep and rbs to current versions.

For rbs, current version is somehow a 3.6 preview - I constrained it to a 3.5 release instead of using the preview. Can upgrade to the preview if this is more desirable.

**Motivation:**

To type check the code tracking component (https://github.com/DataDog/dd-trace-rb/pull/3942) I need type definitions for RubyVM::InstructionSequence which I added in https://github.com/ruby/rbs/pull/2027 to upstream. When I tried to use master of rbs in dd-trace-rb I received several type check errors. This PR repairs the errors to permit usage of current steep & rbs.

**Additional Notes:**

- `defined?` returns nil or a string, not a bool. `!!` was added in several places to cast the return value to a boolean.
- "anything" in Ruby is apparently not an Object (I guess there is BasicObject as well?). event.rbs was changed from Object to untyped for "any value", I think "untyped" means anything but correct me if this is not the right type to use.
- The code in tie.rb runs into https://github.com/soutaro/steep/issues/1231 where steep does not assign correct type to a variable when its value is overwritten from an inner scope. I don't know how to fix this error on our side (hoping to receive some guidance in the upstream issue) and in the mean time I added tie.rb to excluded files. I checked steep 1.6 that is currently being used in master and in my testing it seems to have the same behavior/issue, thus I don't know why we currently aren't receiving this error in master. My theory is the file somehow isn't actually checked (maybe surrounding types aren't resolved as well as with steep 1.7 and as a result steep doesn't attempt to do anything on the Boot return).
- The change from `self.enabled` to `enabled` in appsec fixes the test suite but I don't know why it was `self.enabled` before.

**How to test the change?**

### DIFF
--- a/sig/datadog/appsec/configuration/settings.rbs
+++ b/sig/datadog/appsec/configuration/settings.rbs
@@ -17,7 +17,7 @@ module Datadog
 
         def self.add_settings!: (untyped base) -> untyped
 
-        def self.enabled:  -> bool
+        def enabled:  -> bool
       end
     end
   end
--- a/sig/datadog/appsec/configuration/settings.rbs
+++ b/sig/datadog/appsec/configuration/settings.rbs
@@ -17,7 +17,7 @@ module Datadog
 
         def self.add_settings!: (untyped base) -> untyped
 
-        def enabled:  -> bool
+        def self.enabled:  -> bool
       end
     end
   end
--- a/Gemfile
+++ b/Gemfile
@@ -82,8 +82,8 @@ end
 
 group :check do
   if RUBY_VERSION >= '3.0.0' && RUBY_PLATFORM != 'java'
-    gem 'rbs', '~> 3.2.0', require: false
-    gem 'steep', '~> 1.6.0', require: false
+    gem 'rbs', '~> 3.5.0', require: false
+    gem 'steep', '~> 1.7.0', require: false
   end
   gem 'ruby_memcheck', '>= 3' if RUBY_VERSION >= '3.4.0' && RUBY_PLATFORM != 'java'
   gem 'standard', require: false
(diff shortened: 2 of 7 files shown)


sig/datadog/appsec/configuration/settings.rbs
```diff
@@ -17,7 +17,7 @@ module Datadog
 
         def self.add_settings!: (untyped base) -> untyped
 
-        def self.enabled:  -> bool
+        def enabled:  -> bool
```
ivoanjo, line 20: This doesn't look correct :thinking:. Including this module in a class does not seem to cause a class to have a `#enabled` method?

ivoanjo (approved):
:+1: LGTM. The appsec signature change doesn't look correct, but it seems OK to me to merge this PR and then drop a note to the appsec folks to decide what to do with it.

ivoanjo (approved):
:+1: LGTM. The appsec signature change doesn't look correct, but it seems OK to me to merge this PR and then drop a note to the appsec folks to decide what to do with it.

sig/datadog/appsec/configuration/settings.rbs
p-datadog: I think this is a case of https://github.com/soutaro/steep/issues/1232.

p-datadog:
Thinking about the appsec settings situation and https://github.com/soutaro/steep/issues/1232, my impression is that steep would not produce anything useful for that entire file anyway given it has a totally wrong understanding of what `self` for the vast majority of the file. Therefore I reverted the code change back to what it was (`self.enabled`), and excluded the file from steep.

PR #3953, opened 2024-09-25
Retry apt-get install for memory-leaks config

p-datadog:
**What does this PR do?**

This configuration is frequently failing due to HTTP 503 error when installing valgrind. Retry the installation one time after a short break to see if this would make the configuration less flaky.

**Motivation:**
Repeated test failures

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/.github/workflows/test-memory-leaks.yaml
+++ b/.github/workflows/test-memory-leaks.yaml
@@ -11,7 +11,7 @@ jobs:
           bundler-cache: true # runs 'bundle install' and caches installed gems automatically
           bundler: latest
           cache-version: v1 # bump this to invalidate cache
-      - run: sudo apt update && sudo apt install -y valgrind && valgrind --version
+      - run: sudo apt-get update && (sudo apt-get install -y valgrind || sleep 5 && sudo apt-get install -y valgrind) && valgrind --version
       - run: bundle exec rake compile spec:profiling:memcheck
   test-asan:
     # Temporarily disabled on 2024-09-17 until ruby-asan builds are available again on


.github/workflows/test-memory-leaks.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
           bundler-cache: true # runs 'bundle install' and caches installed gems automatically
           bundler: latest
           cache-version: v1 # bump this to invalidate cache
-      - run: sudo apt update && sudo apt install -y valgrind && valgrind --version
+      - run: sudo apt-get update && (sudo apt-get install -y valgrind || sleep 5 && sudo apt-get install -y valgrind) && valgrind --version
```
ivoanjo, line 14: Minor: Modern versions of apt don't need `apt-get` (they're mostly aliases)

```suggestion
      - run: sudo apt update && (sudo apt install -y valgrind || sleep 5 && sudo apt install -y valgrind) && valgrind --version
```

ivoanjo (approved):
:+1: I'm not sure if this will fix it, but it's cheap to try

.github/workflows/test-memory-leaks.yaml
p-datadog: While apt and apt-get do overlap in functionality, apt is not intended for script usage and makes fewer/no interface compatibility guarantees compared to underlying tools like apt-get (https://unix.stackexchange.com/questions/590699/should-i-use-apt-or-apt-get-in-shell-scripting, https://serverfault.com/questions/958003/how-to-disable-warning-apt-does-not-have-a-stable-cli-interface, https://manpages.ubuntu.com/manpages/xenial/man8/apt.8.html).

ivoanjo: Interesting, thanks for sharing! I definitely did not know this, and I guess have been making the mistake of using `apt` in scripts for a while now. Oops :sweat_smile:

PR #3956, opened 2024-09-25
feat: delay sampling decision to allow sampling on tags and resource

ZStriker19:
**What does this PR do?**
This PR enables lazy sampling with the feature being the ability to sample on tags and resource values for root spans. It does this by:
1. Removing sampling at span start
2. Change the metadata values returned by `TraceOperation` so that values from the root span will be returned if available, allowing sampling on all tags that will end up attached to the root span.
3. Running sampling in `TraceOperation.to_digest` which runs before context propagation
4. Running sampling in `Tracer.flush` which runs before spans are written to the agent.

This PR also fixes wildcard rule sampling when multiple `*` characters are repeated (e.g. `***`), which are now considered equivalent to `*`, meaning "match all".

**Motivation:**
The ability to sample on tags and resource will be helpful to users trying to reduce costs in a precise way as well as keep those spans that are truly useful to them.

**How to test the change?**
Added tests for new behaviors and passes system-tests trace_sampling.py


lib/datadog/tracing/trace_operation.rb
```diff
@@ -161,6 +164,23 @@ def resource
...
+        super || (root_span && root_span.get_metric(key))
+      end
+
+      def tags
+        all_tags = {}
+        all_tags.merge(root_span.tags) if root_span
+        all_tags.merge(@tags)
+        all_tags.merge(@metrics)
+        all_tags
+      end
```
marcotc, line 182: I think we should add unit tests for these new methods.

marcotc (approved):
Awesome work, @ZStriker19!

PR #3961, opened 2024-09-30
DEBUG-2334 Add integration_test helper

p-datadog:
DRY the logic to skip integration tests into a class-level helper

**What does this PR do?**

Adds `integration_test` helper method to mark test cases that require a running agent.

**Motivation:**

Testing of DI network clients.

**Additional Notes:**

**How to test the change?**
No code changes in this PR.

### DIFF
--- a/spec/support/core_helpers.rb
+++ b/spec/support/core_helpers.rb
@@ -49,7 +49,7 @@ module CoreHelpers
     def integration_test
       unless ENV['TEST_DATADOG_INTEGRATION']
         before(:all) do
-          skip "Set TEST_DATADOG_INTEGRATION=1 in environment to run this test"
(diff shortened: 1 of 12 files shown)


spec/support/core_helpers.rb
```diff
@@ -44,4 +44,18 @@ def supports_value_expectations?
       false
     end
   end
+
+  module ClassMethods
+    def integration_test
```
ivoanjo, line 49: `integration_test` imho is not very clear what it does at the top of a test. May I suggest something along the lines of `skip_unless_integration_testing_enabled` or something along does lines?

ivoanjo (approved):
:+1: LGTM

spec/support/core_helpers.rb
p-datadog: It's declarative instead of imperative :)

Renamed.

PR #3965, opened 2024-10-01
ASM Standalone Billing

vpellan:
**What does this PR do?**

This PR add the DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED feature flag to enable ASM standalone billing (or disable APM by limiting the number of traces as much as possible, including distributed tracing)

**How to test the change?**

There are system-tests specific to APPSEC_STANDALONE for each weblog (Rack, Rails 4.2-7.1, Sinatra 1.4-4.0)


lib/datadog/tracing/sampling/rule_sampler.rb
```diff
@@ -30,15 +30,20 @@ def initialize(
           default_sampler: nil
         )
           @rules = rules
-          @rate_limiter = if rate_limiter
+          @rate_limiter = if Datadog.configuration.appsec.standalone.enabled
+                            # 0.0167 ~ 1 trace per minute
+                            Core::TokenBucket.new(0.0167, 1.0)
```
y9v, line 35: for readability I would propose to use `1/60` instead of `0.0167` here

lib/datadog/tracing/contrib/http/circuit_breaker.rb
```diff
@@ -29,6 +29,14 @@ def internal_request?(request)
           end
 
           def should_skip_distributed_tracing?(client_config)
+            if Datadog.configuration.appsec.standalone.enabled
+              active_trace = Tracing.active_trace
+
+              # return true if absence of upstream ASM event (_dd.p.appsec:1)
+              # and no local security event (which sets _dd.p.appsec:1 locally)
+              return true if active_trace.nil? || active_trace.get_tag(Datadog::AppSec::Ext::TAG_APPSEC_EVENT) != '1'
```
Strech, line 37: Will it be possible to cut 1 namespace? (not sure)

```suggestion
              return true if active_trace.nil? || active_trace.get_tag(AppSec::Ext::TAG_APPSEC_EVENT) != '1'
```

spec/datadog/appsec/configuration/settings_spec.rb
```diff
@@ -757,5 +757,45 @@ def patcher
...
+
+        [true, false].each do |value|
+          context "when given #{value}" do
+            let(:appsec_standalone_enabled) { value }
+
+            before { set_appsec_standalone_enabled }
+
+            it { expect(settings.appsec.standalone.enabled).to eq(value) }
+          end
+        end
```
Strech, line 797: It might be personal opinion, but I think DRY in tests makes them harder to read. WDYT?

spec/datadog/appsec/contrib/support/integration/shared_examples.rb
```diff
@@ -165,3 +167,21 @@
...
   end
 end
+
+RSpec.shared_examples 'a trace with ASM Standalone tags' do
+  context 'with appsec enabled' do
+    let(:appsec_enabled) { true }
+    it do
+      expect(service_span.send(:metrics)['_dd.apm.enabled']).to eq(0)
+      expect(service_span.send(:metrics)['_dd.appsec.enabled']).to eq(1.0)
+    end
```
Strech, line 177:
```suggestion
    let(:appsec_enabled) { true }

    it do
      expect(service_span.send(:metrics)['_dd.apm.enabled']).to eq(0)
      expect(service_span.send(:metrics)['_dd.appsec.enabled']).to eq(1.0)
    end
```

spec/datadog/appsec/contrib/support/integration/shared_examples.rb
```diff
@@ -165,3 +167,21 @@
...
+  end
+
+  context 'with appsec disabled' do
+    let(:appsec_enabled) { false }
+    it do
+      expect(service_span.send(:metrics)['_dd.apm.enabled']).to be_nil
+      expect(service_span.send(:metrics)['_dd.appsec.enabled']).to be_nil
+    end
```
Strech, line 185:
```suggestion
    let(:appsec_enabled) { false }

    it do
      expect(service_span.send(:metrics)['_dd.apm.enabled']).to be_nil
      expect(service_span.send(:metrics)['_dd.appsec.enabled']).to be_nil
    end
```

lib/datadog/tracing/contrib/http/circuit_breaker.rb
Strech, line 37: Since this is a very-special case and we put some comments, we also might simplify reading with a good naming (not saying this one is good, but would like to show an example)

```suggestion
              absent_trace_or_no_local_appsec_event = active_trace.nil? || active_trace.get_tag(Datadog::AppSec::Ext::TAG_APPSEC_EVENT) != '1'
              return true if absent_trace_or_no_local_appsec_event
```

lib/datadog/tracing/contrib/http/instrumentation.rb
```diff
@@ -35,6 +35,11 @@ def request(req, body = nil, &block)
                 span.type = Tracing::Metadata::Ext::HTTP::TYPE_OUTBOUND
                 span.resource = req.method
 
+                if Datadog.configuration.appsec.standalone.enabled &&
+                    (trace.nil? || trace.get_tag(Datadog::AppSec::Ext::TAG_APPSEC_EVENT) != '1')
```
Strech, line 39: WDYT about the same approach as ⬆️ ?

Strech (commented):
Great work, I sprinkled a bit of suggestions here and there in attempt to improve this special-case code

lib/datadog/tracing/contrib/http/circuit_breaker.rb
vpellan: It does work but it seems that there's a convention in the codebase to still add the parent namespace (I suppose for readability). You can see another example at l.27-28 of the same file.

spec/datadog/appsec/configuration/settings_spec.rb
vpellan: I agree for more complex cases, but I think that boolean are simple enough that it does not lower the readability too much. `[true, false].each` is used by APM team (e.g. in `spec/datadog/core/configuration/settings_spec.rb`) so I think it's better to use it for cohesion.

ivoanjo: To be honest, this spec is such a massive copy-paste with these boolean checks that I kinda wish we had a `behaves_like_a_binary_setting` or something like that helper rather than repeating it every time.

Not to say we need to fix it in this PR, but yeah, it's been bugging me for a while and it seemed like a good time to share ;)

lib/datadog/tracing/sampling/rule_sampler.rb
vpellan: Fixed!

lib/datadog/appsec/contrib/rack/request_middleware.rb
```diff
@@ -150,7 +150,8 @@ def add_appsec_tags(processor, scope)
 
             return unless trace && span
 
-            span.set_tag('_dd.appsec.enabled', 1)
+            span.set_tag(Datadog::AppSec::Ext::TAG_APM_ENABLED, 0) if Datadog.configuration.appsec.standalone.enabled
+            span.set_tag(Datadog::AppSec::Ext::TAG_APPSEC_ENABLED, 1)
```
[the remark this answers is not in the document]
vpellan: Fixed!

lib/datadog/appsec/event.rb
```diff
@@ -137,6 +137,24 @@ def build_service_entry_tags(event_group)
...
+          # Propagate to downstream services the information that the current distributed trace is
+          # containing at least one ASM security event
+          if scope.trace
+            scope.trace.keep!
+            scope.trace.set_tag(
+              Datadog::Tracing::Metadata::Ext::Distributed::TAG_DECISION_MAKER,
+              Datadog::Tracing::Sampling::Ext::Decision::ASM
+            )
+            scope.trace.set_tag(Datadog::AppSec::Ext::TAG_APPSEC_EVENT, '1')
+          end
```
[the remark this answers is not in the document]
vpellan: Fixed!

lib/datadog/tracing/contrib/http/circuit_breaker.rb
vpellan: Fixed!

lib/datadog/tracing/contrib/http/instrumentation.rb
vpellan: Fixed!

spec/datadog/appsec/configuration/settings_spec.rb
Strech: Yeah, shared behavior sounds good. Agree that maybe not right now, but it's good to share the pain 😉

y9v: what are we actually testing here? If we want to test that the value is being assigned, it is enough to test once with one value. If we want to test that this setting could be only set to a boolean value, we also need a test case when a non-boolean value is being passed to this setting.

lib/datadog/appsec/contrib/rack/request_middleware.rb
```diff
@@ -150,7 +150,8 @@ def add_appsec_tags(processor, scope)
 
             return unless trace && span
 
-            span.set_tag('_dd.appsec.enabled', 1)
+            span.set_tag(Datadog::AppSec::Ext::TAG_APPSEC_ENABLED, 1)
+            span.set_tag(Datadog::AppSec::Ext::TAG_APM_ENABLED, 0) if Datadog.configuration.appsec.standalone.enabled
```
y9v, line 154: not sure, but does this need a comment?

lib/datadog/appsec/event.rb
```diff
@@ -137,6 +137,24 @@ def build_service_entry_tags(event_group)
         end
         # rubocop:enable Metrics/MethodLength
 
+        def add_event_tags(scope, waf_result)
```
y9v, line 140: Since the class is already named `Event`, should we rename this function to `add_tags`?

lib/datadog/appsec/event.rb
```diff
@@ -137,6 +137,24 @@ def build_service_entry_tags(event_group)
...
+          if scope.service_entry_span
+            scope.service_entry_span.set_tag('appsec.blocked', 'true') if waf_result.actions.include?('block')
+            scope.service_entry_span.set_tag('appsec.event', 'true')
+          end
+
+          # Propagate to downstream services the information that the current distributed trace is
+          # containing at least one ASM security event
```
y9v, line 147: I would put this comment before `scope.trace.keep!`, it's more relevant there

lib/datadog/appsec/event.rb
```diff
@@ -137,6 +137,24 @@ def build_service_entry_tags(event_group)
...
+          return unless scope.trace
+
+          scope.trace.keep!
+          scope.trace.set_tag(
+            Datadog::Tracing::Metadata::Ext::Distributed::TAG_DECISION_MAKER,
+            Datadog::Tracing::Sampling::Ext::Decision::ASM
+          )
+          scope.trace.set_tag(Datadog::AppSec::Ext::TAG_APPSEC_EVENT, '1')
```
y9v, line 155: Do we want to execute this code only if the standalone appsec is enabled?

lib/datadog/appsec/ext.rb
```diff
@@ -5,6 +5,10 @@ module AppSec
     module Ext
       INTERRUPT = :datadog_appsec_interrupt
       SCOPE_KEY = 'datadog.appsec.scope'
+
+      TAG_APPSEC_ENABLED = '_dd.appsec.enabled'
+      TAG_APM_ENABLED = '_dd.apm.enabled'
```
y9v, line 10: does this constant really belong to `Appsec` module?

lib/datadog/core/remote/transport/http.rb
```diff
@@ -120,6 +120,10 @@ def default_headers
               # Add container ID, if present.
               container_id = Datadog::Core::Environment::Container.container_id
               headers[Datadog::Core::Transport::Ext::HTTP::HEADER_CONTAINER_ID] = container_id unless container_id.nil?
+              # Pretend that stats computation are already done by the client
```
y9v, line 123: I don't have much background here, but I think it would be helpful to extend this comment and add an explanation on what this header is used for, and why do we set it to `yes` manually.

lib/datadog/tracing/contrib/http/circuit_breaker.rb
```diff
@@ -29,6 +29,16 @@ def internal_request?(request)
...
 
           def should_skip_distributed_tracing?(client_config)
+            if Datadog.configuration.appsec.standalone.enabled
+              active_trace = Tracing.active_trace
+
+              # return true if absence of upstream ASM event (_dd.p.appsec:1)
+              # and no local security event (which sets _dd.p.appsec:1 locally)
+              no_upstream_or_local_event = active_trace.nil? ||
+                active_trace.get_tag(Datadog::AppSec::Ext::TAG_APPSEC_EVENT) != '1'
+              return true if no_upstream_or_local_event
```
y9v, line 39: how about this?

```ruby
return true unless Tracing.active_trace

Tracing.active_trace.get_tag(Datadog::AppSec::Ext::TAG_APPSEC_EVENT) != '1'
```

lib/datadog/tracing/contrib/http/circuit_breaker.rb
```diff
@@ -29,6 +29,16 @@ def internal_request?(request)
           end
 
           def should_skip_distributed_tracing?(client_config)
+            if Datadog.configuration.appsec.standalone.enabled
+              active_trace = Tracing.active_trace
+
+              # return true if absence of upstream ASM event (_dd.p.appsec:1)
+              # and no local security event (which sets _dd.p.appsec:1 locally)
```
y9v, line 36: again, I think it makes sense to explain here in which case `_dd.p.appsec` is being set to `1` and why we want to skip distributed tracing in such case

lib/datadog/tracing/contrib/http/instrumentation.rb
```diff
@@ -35,6 +35,13 @@ def request(req, body = nil, &block)
                 span.type = Tracing::Metadata::Ext::HTTP::TYPE_OUTBOUND
                 span.resource = req.method
 
+                asm_standalone_no_upstream_or_local_event = Datadog.configuration.appsec.standalone.enabled &&
+                  (trace.nil? || trace.get_tag(Datadog::AppSec::Ext::TAG_APPSEC_EVENT) != '1')
```
y9v, line 39: should this be a method on `Datadog::Trace`?

lib/datadog/tracing/sampling/rule_sampler.rb
```diff
@@ -30,15 +30,20 @@ def initialize(
...
                             rate_limiter
                           elsif rate_limit
                             Core::TokenBucket.new(rate_limit)
                           else
                             Core::UnlimitedLimiter.new
                           end
-
-          @default_sampler = if default_sampler
+          @default_sampler = if Datadog.configuration.appsec.standalone.enabled
+                               @rules = [SimpleRule.new(sample_rate: 1.0)]
```
y9v, line 44: I think it's quite confusing that we are setting `@rules` when assigning `@default_sampler` value. Should we move this out from here?

y9v (changes requested):

.github/workflows/system-tests.yml
```diff
@@ -11,9 +11,9 @@ on:
 env:
   REGISTRY: ghcr.io
   REPO: ghcr.io/datadog/dd-trace-rb
-  ST_REF: main
-  FORCE_TESTS: -F tests/appsec/waf/test_addresses.py::Test_GraphQL -F tests/appsec/test_blocking_addresses.py::Test_BlockingGraphqlResolvers
-  FORCE_TESTS_SCENARIO: GRAPHQL_APPSEC
```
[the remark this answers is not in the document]
vpellan: This line must be cleaned after the release of the specified version in the manifest. In the case of GraphQL Appsec, it should have been cleaned after 2.3.0 release, but it does not change anything after the release, as we are forcing the execution of a test that is already executed. Charles specified in the doc that "from time to time" you should removes all the -F from the CI (https://github.com/DataDog/system-tests/blob/main/docs/edit/egg-chicken-changes.md?plain=1#L34), so this is why I did not considered it a priority.
I guess that a clean way to do it would be to add a step in fast_castle.
We can do a separate PR but the way FORCE_TESTS_SCENARIO is implemented does not enable to force tests execution on multiple scenarios, so that would mean that I cannot force execute APPSEC_STANDALONE on this PR

lib/datadog/appsec/event.rb
vpellan: > Do we want to execute this code only if the standalone appsec is enabled?

We want to execute this code each time there is an Appsec event, it is not specific to Appsec standalone

I wrote this method to factorise `watch_xxx` methods in each integration's Gateway Watcher so that it would comply to our Rubocop spec.

> and I would actually move this code to another function

Would you move it to another method because it is acting on `scope.trace` and not `scope.service_entry_span` anymore ?

lib/datadog/appsec/event.rb
vpellan: Fixed!

lib/datadog/appsec/ext.rb
vpellan: I believe so as the only time it is added to a trace is when ASM standalone is activated. Usually, we don't want disable APM, this is for now the only case when it is disabled.

lib/datadog/appsec/event.rb
vpellan: Fixed!

lib/datadog/core/remote/transport/http.rb
vpellan: Thanks, I added more context in the comment

lib/datadog/tracing/contrib/http/circuit_breaker.rb
vpellan: I also added more context to this one

lib/datadog/tracing/contrib/http/circuit_breaker.rb
vpellan: I did something similar (that would also satisfy Rubocop) thanks !

lib/datadog/tracing/sampling/rule_sampler.rb
vpellan: I agree, I separated rules and default_sampler initializers, thanks!

lib/datadog/appsec/event.rb
y9v: yes, I think it does a bit more then setting tags - if I understood correctly it's more about making sure that the trace is preserved?

lib/datadog/appsec/contrib/rack/request_middleware.rb
y9v: I would add somethink like "we want to make sure the customer will not get billed for APM when standalone ASM is enabled". But again, it's probably fine to leave it as it is

p-datadog: I read PR description but did not understand that this is what the PR is trying to achieve. So probably would be helpful to include this bit *somewhere*.

lib/datadog/tracing/sampling/rule_sampler.rb
```diff
@@ -29,21 +29,29 @@ def initialize(
           default_sample_rate: Datadog.configuration.tracing.sampling.default_rate,
           default_sampler: nil
         )
```
p-datadog, line 31: I think this section of the code will benefit from a comment explaining appsec & tracing interaction. Especially for someone who is new to both appsec and tracing.

lib/datadog/appsec/contrib/rack/request_middleware.rb
vpellan: I added a comment, thanks!

lib/datadog/tracing/sampling/rule_sampler.rb
vpellan: Thanks, I added a comment to explain it a bit more, let me know if you think that it is not comprehensive enough !

spec/datadog/appsec/configuration/settings_spec.rb
vpellan: We are testing that the value is being assigned. I agree that these tests should be reworked (on another PR)

lib/datadog/tracing/contrib/http/instrumentation.rb
```diff
@@ -35,6 +35,10 @@ def request(req, body = nil, &block)
                 span.type = Tracing::Metadata::Ext::HTTP::TYPE_OUTBOUND
                 span.resource = req.method
 
+                if Tracing::TraceOperation.appsec_standalone_reject?(trace)
+                  trace.sampling_priority = Tracing::Sampling::Ext::Priority::AUTO_REJECT
```
marcotc, line 39: This will override distributed tracing sampling priority, is that expected?

lib/datadog/tracing/sampling/rule_sampler.rb
```diff
@@ -29,21 +29,34 @@ def initialize(
...
                             Core::UnlimitedLimiter.new
                           end
-
-          @default_sampler = if default_sampler
-                               default_sampler
-                             elsif default_sample_rate
-                               # Add to the end of the rule list a rule always matches any trace
-                               @rules << SimpleRule.new(sample_rate: default_sample_rate)
+          @default_sampler = if Datadog.configuration.appsec.standalone.enabled ||
+              (default_sample_rate && !default_sampler)
```
marcotc, line 56: I suggest moving this logic to the initialization of the `RuleSampler`, instead of inside of it:
https://github.com/DataDog/dd-trace-rb/blob/d3d7fde89f4a5bba1032d60867c00ddc669095f7/lib/datadog/tracing/component.rb#L70-L96

It looks like this logic can manipulate the parameters passed to `RuleSampler` to accomplish the same goal, I think it would be: `RuleSampler.new([], rate_limiter: Core::TokenBucket.new(1.0 / 60, 1.0), default_sample_rate: 1.0)`.

lib/datadog/appsec/event.rb
```diff
@@ -165,6 +174,18 @@ def gzip(value)
...
+        # Propagate to downstream services the information that the current distributed trace is
+        # containing at least one ASM security event.
+        def add_distributed_tags(trace)
+          return unless trace
+
+          trace.set_tag(
+            Datadog::Tracing::Metadata::Ext::Distributed::TAG_DECISION_MAKER,
+            Datadog::Tracing::Sampling::Ext::Decision::ASM
+          )
+          trace.set_tag(Datadog::AppSec::Ext::TAG_DISTRIBUTED_APPSEC_EVENT, '1')
```
marcotc, line 187: Is `Datadog::AppSec::Ext::TAG_DISTRIBUTED_APPSEC_EVENT` supposed to be a string `1` or number `1`?

lib/datadog/tracing/contrib/http/instrumentation.rb
marcotc: Another thing, this likely has to be done for all places where `inject(trace, req)` happens.

marcotc: > Another thing, this likely has to be done for all places where inject(trace, req) happens.

If this is true, I suggest adding this logic to the shared `inject!` logic here: https://github.com/DataDog/dd-trace-rb/blob/7cd16a58f820845ce812877d32fa2a3129f9d291/lib/datadog/tracing/distributed/propagation.rb#L53

marcotc: Hmm, this logic here, from `lib/datadog/tracing/contrib/http/circuit_breaker.rb`, can probably also be moved to `inject!`:
```ruby
            if Datadog.configuration.appsec.standalone.enabled
              # Skip distributed tracing so that we don't bill distributed traces in case of absence of
              # upstream ASM event (_dd.p.appsec:1) and no local security event (which sets _dd.p.appsec:1 locally).
              # If there is an ASM event, we still have to check if distributed tracing is enabled or not
              return true unless Tracing.active_trace

              return true if Tracing.active_trace.get_tag(Datadog::AppSec::Ext::TAG_DISTRIBUTED_APPSEC_EVENT) != '1'
            end
```

lib/datadog/appsec/contrib/graphql/gateway/watcher.rb
```diff
@@ -38,11 +38,9 @@ def watch_multiplex(gateway = Instrumentation.gateway)
...
                         }
 
-                        if scope.service_entry_span
-                          scope.service_entry_span.set_tag('appsec.blocked', 'true') if result.actions.include?('block')
-                          scope.service_entry_span.set_tag('appsec.event', 'true')
-                        end
-
+                        # We want to keep the trace in case of security event
+                        scope.trace.keep! if scope.trace
+                        Datadog::AppSec::Event.add_tags(scope, result)
```
marcotc, line 43: Does it make sense to always bundle these two lines together?

lib/datadog/tracing/trace_operation.rb
```diff
@@ -373,6 +373,11 @@ def fork_clone
         )
       end
 
+      def self.appsec_standalone_reject?(trace)
+        Datadog.configuration.appsec.standalone.enabled &&
+          (trace.nil? || trace.get_tag(Datadog::AppSec::Ext::TAG_DISTRIBUTED_APPSEC_EVENT) != '1')
+      end
+
```
marcotc, line 380: I suggest moving this method to the `AppSec` namespace, since it doesn't really belong to the trace operation, and is a "static" method.

sig/datadog/appsec/event.rbs
```diff
@@ -13,14 +13,18 @@ module Datadog
...
+                       
       private
 
       def self.compressed_and_base64_encoded: (untyped value) -> untyped
 
       def self.json_parse: (untyped value) -> untyped
 
       def self.gzip: (untyped value) -> untyped
+                   
+      def self.add_distributed_tags: (untyped trace) -> untyped
```
Strech, line 27: May I ask to type it? (Not sure about `trace` type, but I think it could be defined too.

```suggestion
      def self.add_distributed_tags: (untyped trace) -> void
```

sig/datadog/appsec/event.rbs
```diff
@@ -13,14 +13,18 @@ module Datadog
       def self.record_via_span: (Datadog::Tracing::SpanOperation, *untyped events) -> untyped
 
       def self.build_service_entry_tags: (Array[Hash[::Symbol, untyped]] event_group) -> Hash[::String, untyped]
-
+                                       
+      def self.add_tags: (Datadog::AppSec::Scope scope, Datadog::AppSec::WAF::Result waf_result) -> untyped
```
Strech, line 17:
```suggestion
      def self.add_tags: (Datadog::AppSec::Scope scope, Datadog::AppSec::WAF::Result waf_result) -> void
```

Strech (approved):
I don't have much to add to existing suggestions by others, great job 👏🏼

lib/datadog/appsec/contrib/rack/request_middleware.rb
```diff
@@ -150,7 +150,9 @@ def add_appsec_tags(processor, scope)
 
             return unless trace && span
 
-            span.set_tag('_dd.appsec.enabled', 1)
+            span.set_tag(Datadog::AppSec::Ext::TAG_APPSEC_ENABLED, 1)
```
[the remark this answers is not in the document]
vpellan: According to [these system-tests](https://github.com/DataDog/system-tests/blob/main/tests/appsec/test_asm_standalone.py#L92), `_dd.apm.enabled` should be a metric, so I believe `_dd.appsec.enabled` too.

I will replace all the `set_tag` that actually sets metrics with `set_metric`, thanks!

lib/datadog/appsec/event.rb
vpellan: It is supposed to be a string according to [these system-tests](https://github.com/DataDog/system-tests/blob/main/tests/appsec/test_asm_standalone.py#L82)

lib/datadog/appsec/contrib/graphql/gateway/watcher.rb
vpellan: If we create a new method to bundle these two line together, I think that `add_tags_and_keep!` or `tag_and_keep!` would be a better name than just `tag!` as `keep!` does more than just setting a tag (see this [review](https://github.com/DataDog/dd-trace-rb/pull/3965#discussion_r1791328487))

lib/datadog/tracing/contrib/http/circuit_breaker.rb
```diff
@@ -29,6 +29,15 @@ def internal_request?(request)
           end
 
           def should_skip_distributed_tracing?(client_config)
+            if Datadog.configuration.appsec.standalone.enabled
```
anmarchenko, line 32: I see that we add a bunch of `Datadog.configuration.appsec.standalone.enabled` all over the tracing component which imposes a hard dependency between Tracing and AppSec internal details. What's especially worrying, is that in this case we reverse the dependency and now Tracing knows about AppSec internals (even though AppSec is supposed to be a client of Tracing).

So my question is for the Tracing team: do you think it would make sense to introduce some tracing settings to allow client products to tweak tracing behaviour even more? Like we do between CI and Tracing modules: Tracing exposes settings, CI configures them. 

@marcotc what do you think?

lib/datadog/tracing/contrib/http/instrumentation.rb
vpellan: > Another thing, this likely has to be done for all places where `inject(trace, req)` happens.

I agree that this should most likely be done in all the places where there is `HTTP.inject`, however I don't think we should add it to `inject!`, as we must add `_sampling_priority_v1` to the trace even though it is not distributed. If we add it to inject and distributed tracing is disabled in the client config, it will not force the sampling priority to be rejected.

lib/datadog/tracing/sampling/rule_sampler.rb
vpellan: It is also initialised in here:
https://github.com/DataDog/dd-trace-rb/blob/60febf8012de92758b85a7de0287342475ef0fb0/lib/datadog/tracing/tracer.rb#L50-L62

Do you think I should also move the logic there ? Imo it would mean duplicated code

sig/datadog/appsec/event.rbs
vpellan: Fixed in 22f18a8

sig/datadog/appsec/event.rbs
vpellan: Fixed in 22f18a8

lib/datadog/tracing/trace_operation.rb
vpellan: Fixed in e24310ef

lib/datadog/appsec/contrib/graphql/gateway/watcher.rb
vpellan: Fixed in 73f8af2cd11e7e37a9d91a6b0cb1e72e1edfae02

lib/datadog/tracing/contrib/http/instrumentation.rb
vpellan: > Another thing, this likely has to be done for all places where `inject(trace, req)` happens.

Fixed in e24310efa3ba270c6a90be6bd8e779b7d445a178

lib/datadog/appsec/contrib/rack/request_middleware.rb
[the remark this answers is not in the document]
vpellan: Fixed in 4a2bf673504f46b2d20708b218d4988643af3084

lib/datadog/tracing/sampling/rule_sampler.rb
marcotc: You can ignore this one in `dd-trace-rb/lib/datadog/tracing/tracer.rb`. It's not used anymore, just a relic from when the Tracer class itself was public. You can always assume well go through the `Component` initialization.

marcotc (approved):
Looks good!

y9v (approved):
:+1:

lib/datadog/tracing/contrib/http/circuit_breaker.rb
vpellan: To add more context: appsec standalone is still experimental and the env var/configuration might change name in the future, and the 'appsec' part of this setting name may disappear. What it do on the tracing part is disabling distributed tracing, adding a header to the payload sent to the agent, creating a 1 trace per minute rate limiter, and setting the sampling priority to 'reject' if there is an appsec event detected. Although the first 3 points aren't directly tied to appsec, the last one is, but it is still acting only on the traces. If you think this is not a blocker, I will merge this PR this evening

PR #3969, opened 2024-10-02
Add a force disable of appsec when using Ruby >= 3.3 with old ffi

y9v:
Ruby 3.3.0 and above has a memory leak when using `libddwaf-rb` with `ffi` gem older than 1.16.0.

We want to avoid forcing our customers to update `ffi` gem by setting the minimum version in `libddwaf-rb`, since prebuilt `ffi` extension is incompatible with rubygems version supplied with ruby 2.7 and below.

The memory leak happens because `ffi` gem used untyped DATA API prior to this commit (which is part of 1.16.0 release):
https://github.com/ffi/ffi/commit/ca2744aed44da7486b3839b879388af3f6307a88

**Motivation:**
Memory leak that occurs when enabling `appsec` for applications running on ruby 3.3.x and an old version of `ffi` gem locked in Gemfile.lock (older than `1.16.0`).

### DIFF
--- a/lib/datadog/appsec/component.rb
+++ b/lib/datadog/appsec/component.rb
@@ -11,7 +11,8 @@ module Datadog
     class Component
       class << self
         def build_appsec_component(settings, telemetry:)
-          return unless settings.respond_to?(:appsec) && settings.appsec.enabled
+          return if !settings.respond_to?(:appsec) || !settings.appsec.enabled
+          return if incompatible_ffi_version?
 
           processor = create_processor(settings, telemetry)
 
@@ -27,6 +28,21 @@ module Datadog
         end
 
         private
+
+        def incompatible_ffi_version?
+          ffi_version = Gem.loaded_specs['ffi'] && Gem.loaded_specs['ffi'].version
+          if RUBY_VERSION >= '3.3.0' && ffi_version < '1.16.0'
+            Datadog.logger.warn(
+              'AppSec is not supported in Ruby versions above 3.3.0 when using `ffi` versions older than 1.16.0, ' \
+              'and will be forcibly disabled due to a memory leak in `ffi`. ' \
+              'Please upgrade your `ffi` version to 1.16.0 or higher.'
+            )
+
+            true
+          else
+            false
+          end
+        end
 
         def create_processor(settings, telemetry)
           rules = AppSec::Processor::RuleLoader.load_rules(
--- a/lib/datadog/appsec/component.rb
+++ b/lib/datadog/appsec/component.rb
@@ -31,7 +31,8 @@ module Datadog
 
         def incompatible_ffi_version?
           ffi_version = Gem.loaded_specs['ffi'] && Gem.loaded_specs['ffi'].version
-          return false unless RUBY_VERSION >= '3.3.0' && ffi_version < '1.16.0'
+          return false unless Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.3') &&
+            ffi_version < Gem::Version.new('1.16.0')
 
           Datadog.logger.warn(
             'AppSec is not supported in Ruby versions above 3.3.0 when using `ffi` versions older than 1.16.0, ' \
--- a/lib/datadog/appsec/component.rb
+++ b/lib/datadog/appsec/component.rb
@@ -31,17 +31,15 @@ module Datadog
 
         def incompatible_ffi_version?
           ffi_version = Gem.loaded_specs['ffi'] && Gem.loaded_specs['ffi'].version
-          if RUBY_VERSION >= '3.3.0' && ffi_version < '1.16.0'
-            Datadog.logger.warn(
-              'AppSec is not supported in Ruby versions above 3.3.0 when using `ffi` versions older than 1.16.0, ' \
-              'and will be forcibly disabled due to a memory leak in `ffi`. ' \
-              'Please upgrade your `ffi` version to 1.16.0 or higher.'
-            )
+          return false unless RUBY_VERSION >= '3.3.0' && ffi_version < '1.16.0'
 
-            true
-          else
-            false
-          end
+          Datadog.logger.warn(
+            'AppSec is not supported in Ruby versions above 3.3.0 when using `ffi` versions older than 1.16.0, ' \
+            'and will be forcibly disabled due to a memory leak in `ffi`. ' \
+            'Please upgrade your `ffi` version to 1.16.0 or higher.'
+          )
+
+          true
         end
 
         def create_processor(settings, telemetry)
--- a/spec/datadog/appsec/component_spec.rb
+++ b/spec/datadog/appsec/component_spec.rb
@@ -19,6 +19,20 @@ RSpec.describe Datadog::AppSec::Component do
         expect(component).to be_a(described_class)
       end
 
+      context 'when using old ffi version with Ruby 3.3.x' do
+        before do
+          stub_const('RUBY_VERSION', '3.3.0')
+          allow(Gem).to receive(:loaded_specs).and_return('ffi' => double(version: Gem::Version.new('1.15.4')))
+        end
+
+        it 'returns a Datadog::AppSec::Component instance with a nil processor' do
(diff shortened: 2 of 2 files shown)


lib/datadog/appsec/component.rb
```diff
@@ -28,6 +29,22 @@ def build_appsec_component(settings, telemetry:)
...
 
+        def enable_appsec?(settings)
+          return false unless settings.respond_to?(:appsec) && settings.appsec.enabled
+
+          ffi_version = Gem.loaded_specs['ffi'] && Gem.loaded_specs['ffi'].version
+          if RUBY_VERSION >= '3.3.0' && ffi_version < '1.16.0'
+            Datadog.logger.warn(
+              'Appsec is not supported in Ruby versions above 3.3.0 when using `ffi` versions older than 1.16.0, ' \
+              'and will be forcibly disabled. This is due to a memory leak in `ffi` versions lower than 1.16.0.' \
+              'Please upgrade your `ffi` version to 1.16.0 or higher.'
```
Strech, line 40: WDYT if we join the reason to the main sentence?

```suggestion
              'AppSec is not supported in Ruby versions above 3.3.0 when using `ffi` versions older than 1.16.0, ' \
              'and will be forcibly disabled due to a memory leak in `ffi`. ' \
              'Please upgrade your `ffi` version to 1.16.0 or higher.'
```

lib/datadog/appsec/component.rb
```diff
@@ -28,6 +29,22 @@ def build_appsec_component(settings, telemetry:)
 
         private
 
+        def enable_appsec?(settings)
+          return false unless settings.respond_to?(:appsec) && settings.appsec.enabled
```
Strech, line 33: Do you think inverting `unless &&` makes sense here? I find it hard to understand (negate logic with addition to `&&`)

```suggestion
          return false if !settings.respond_to?(:appsec) || !settings.appsec.enabled
```

lib/datadog/appsec/component.rb
```diff
@@ -12,6 +12,7 @@ class Component
       class << self
         def build_appsec_component(settings, telemetry:)
           return unless settings.respond_to?(:appsec) && settings.appsec.enabled
+          return unless enable_appsec?(settings)
```
Strech, line 15: I think the method with `?` is doing more than 1 thing. WDYT on splitting it on something like this

```suggestion
          return unless appsec_enabled?(settings)
          return if incompatible_ffi_version? 
```

lib/datadog/appsec/component.rb
y9v: yes, this is definitely better, thanks!

lib/datadog/appsec/component.rb
y9v: I agree, this will improve readability

lib/datadog/appsec/component.rb
y9v: fixed: [b7ca70a](https://github.com/DataDog/dd-trace-rb/pull/3969/commits/b7ca70ab82e390bb76527f9356162930e706b830)

lib/datadog/appsec/component.rb
y9v: I removed `appsec_enabled?` method and added `incompatible_ffi_version?` here: [51a803f](https://github.com/DataDog/dd-trace-rb/pull/3969/commits/51a803f2b68f70da7b2eb9d49c8173501cb3d59d)

lib/datadog/appsec/component.rb
```diff
@@ -28,6 +29,19 @@ def build_appsec_component(settings, telemetry:)
 
         private
 
+        def incompatible_ffi_version?
+          ffi_version = Gem.loaded_specs['ffi'] && Gem.loaded_specs['ffi'].version
+          return false unless RUBY_VERSION >= '3.3.0' && ffi_version < '1.16.0'
```
p-datadog, line 34: I think you would need to parse the FFI version correctly, string comparison works for Ruby versions only because their components are kept to single digits.

I would also check against 3.3 not 3.3.0.

spec/datadog/appsec/component_spec.rb
```diff
@@ -19,6 +19,20 @@
         expect(component).to be_a(described_class)
       end
 
+      context 'when using old ffi version with Ruby 3.3.x' do
+        before do
+          stub_const('RUBY_VERSION', '3.3.0')
+          allow(Gem).to receive(:loaded_specs).and_return('ffi' => double(version: Gem::Version.new('1.15.4')))
```
p-datadog, line 25: If you use 1.9.0 for the FFI version the test will fail right?

lib/datadog/appsec/component.rb
p-datadog: For the record I think the `unless` & `&&` version was easier to follow.

spec/datadog/appsec/component_spec.rb
y9v: no, it is still green:

```ruby
Gem::Version.new('1.9.0') < '1.16.0'
=> true
```

I think this is because we are comparing `Gem::Version` with a `String` (not sure this will work correctly in older ruby versions though)

lib/datadog/appsec/component.rb
y9v: Added `Gem::Version` for all version comparisons here: [03f3bc9](https://github.com/DataDog/dd-trace-rb/pull/3969/commits/03f3bc90a34dcc748370d295a0635157526f58a9)

spec/datadog/appsec/component_spec.rb
Strech: We can dismantle version on major, minor and test piece-by-piece (just in case)

lib/datadog/appsec/component.rb
```diff
@@ -28,6 +29,20 @@ def build_appsec_component(settings, telemetry:)
 
         private
 
+        def incompatible_ffi_version?
+          ffi_version = Gem.loaded_specs['ffi'] && Gem.loaded_specs['ffi'].version
+          return false unless Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.3') &&
+            ffi_version < Gem::Version.new('1.16.0')
```
Strech, line 35: Not sure how it could be relevant and don't want to over complicate it, but if we check that `Gem.loaded_specs['ffi']` not `nil`, what if it's `nil`?

Then entire `ffi_version` is `nil`, right? And it could fail next check

```
(irb):3:in `<main>': undefined method `<' for nil (NoMethodError)

            ffi_version < Gem::Version.new('1.16.0')
```

But I made it up, maybe it never happen

Strech (approved):
I think it's good, maybe with a few adjustments (if needed)

lib/datadog/appsec/component.rb
y9v: I will still add a guard clause for this, just in case

y9v: although I'm not even sure what to do in this case - we don't want to create a processor, and we probably don't want to use the same warning message. Do we want to silently return true in such case?

spec/datadog/appsec/component_spec.rb
p-datadog: Yep, https://github.com/rubygems/rubygems/blob/master/lib/rubygems/version.rb#L358, added in https://github.com/rubygems/rubygems/commit/7e0dbb79f2c351532ae6c8cf12d5b39579dc4c54 2 years ago, which sounds like this functionality is probably not going to exist in Ruby 2.5?

p-datadog: https://github.com/rails/rails/pull/47480 conveniently says that the threshold for this feature is Ruby 3.1.

p-datadog (approved):

lib/datadog/appsec/component.rb
p-datadog: I think it would be fine to just not warn in this situation and carry on to subsequent code.

Strech: I think we can consider it to be "not there" ...

Issue #3954, opened 2024-09-25
Crash tracker spawning long-running child processes?


luc-financeit:
**Current behaviour**

Upgrading the `datadog` gem from 2.2.0 to 2.3.0 causes a spec in our test suite to hang. The code being tested does something like

```rb
6.times do
  Process.fork { ... }
end

Process.waitall # Hangs
```

Looking at `ps` output, there's a `libdatadog-crashtracking-receiver` child process which our `waitall` is hung up on — manually killing the crashtracker process, or disabling the crashtracker entirely (via `DD_CRASHTRACKING_ENABLED`) allows the test to complete successfully.

```
F   UID     PID    PPID PRI  NI    VSZ   RSS WCHAN  STAT TTY        TIME COMMAND
4     0     328       0  20   0  16460  7680 -      Ss   ?          0:00 bash -l
4     0     557     328  20   0 190964 29656 -      Sl   ?          0:00  \_ ruby /usr/local/rvm/gems/ruby-3.1.6/bin/parallel_rspec --verbose spec
4     0     559     557  20   0 1414052 518752 -    Sl   ?          1:18      \_ ruby bin/rspec -O .rspec_parallel spec/the_hanging_spec.rb
0     0     560     559  20   0  24128  3844 -      S    ?          0:00          \_ /usr/local/rvm/gems/ruby-3.1.6/gems/libdatadog-11.0.0.1.0-x86_64-linux/vendor/libdatadog-11.0.0/x86_64-linux/libdatadog-x86_64-unknown-linux-gnu/bin/libdatadog-crashtracking-receiver 
```

**Expected behaviour**

No hang.

**Steps to reproduce**

I'm afraid I don't have a stand-alone reproduction — the actual code is significantly more complicated, and I haven't tried slimming it down to see if there's additional triggers. I can dig in to it more if the above isn't enough to go on.

**Environment**

* **datadog version:** 2.3.0
* **Configuration block (`Datadog.configure ...`):** See [gist](https://gist.github.com/luc-financeit/42c5250672b6d6cc9aae5d0edd99d300).
* **Ruby version:** 3.1.6
* **Operating system:** Amazon Linux 2
* **Relevant library versions:** N/A

anmarchenko:
Hi @luc-financeit! Thanks for reporting this and sorry you've run into this issue.

[...]

PR #3970, opened 2024-10-03
Disable crashtracking by default

lloeki:
**What does this PR do?**

Disable by default as a cautionary measure.

**Motivation:**

Crashtracking has been identified as causing issues with `ECHILD` and `waitpid`, and possibly anything involving `SIGCHLD` with obscure failure modes of hard to anticipate consequences.

**Additional Notes:**

- Mitigates #3954.
- Attempted fix for `Process.wait(pid)` at #3959, but the problem is more intricate.
- Deeper fix `libdatadog` pending.

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/core/configuration/settings.rb
+++ b/lib/datadog/core/configuration/settings.rb
@@ -886,7 +886,7 @@ module Datadog
           # Enables reporting of information when Ruby VM crashes.
           option :enabled do |o|
             o.type :bool
-            o.default true
+            o.default false
             o.env 'DD_CRASHTRACKING_ENABLED'
           end
         end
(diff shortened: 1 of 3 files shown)


ivoanjo (approved):
:+1: Change seems reasonable, although the crashtracker specs are failing in CI, I guess they need a bit of fixing given the new default.

lloeki:
> I guess they need a bit of fixing given the new default.

Yes that's what I gathered. The errors seem weird though, and the `pgrep` check is a bit naive (on top of requiring `pgrep`): it would pass if a matching process already existed.

y9v (approved):

PR #3975, opened 2024-10-07
[APPSEC-10303] Replace AppSec rate limiter with core rate limiter

Strech:
**What does this PR do?**

This allows us to be more precise in throttling of outgoing AppSec traces and removes additional custom logic for rate limiting.

**Motivation:**

We can re-use recently moved to core `BucketLimiter`. It will eliminate of logic implementation and improve rate limiting. Existing limiter is too aggressive (correct number is `10` traces per test run):

```console
[15:19:31] AppSec/system-tests main
❯❯❯ while true; do docker logs -f system-tests-weblog -f 2>/dev/null | rg -o '_sampling_priority_v1\W:2' | wc -l; sleep 1; done
3
3
3
3
3
3
3
```

with this PR

```console
[15:25:11] AppSec/system-tests appsec-10303-fix-rate-limiter
❯❯❯ while true; do docker logs -f system-tests-weblog -f 2>/dev/null | rg -o '_sampling_priority_v1\W:2' | wc -l; sleep 1; done
10
10
10
10
10
10
10
```

**Additional Notes:**

I've corrected some existing tests and add a bit of a new logic under tests (since we rate limiting per thread).

**IMPORTANT** Unfortunately, it's impossible to guarantee that we will receive exactly `N` amount of traces after limiting due to execution speed of the block inside the `Datadog::AppSec::RateLimiter#limit`. If it take a little longer the fractional part of the token will accumulate and give us an extra spare token to spend. Because of that tests were adjusted to allow this drift of a single token.

**How to test the change?**

Set the AppSec rate limiter on 1/TPS and shoot a bunch of requests (hopefully soon test generator will come in play)

### DIFF
--- a/lib/datadog/appsec/event.rb
+++ b/lib/datadog/appsec/event.rb
@@ -52,7 +52,7 @@ module Datadog
           # ensure rate limiter is called only when there are events to record
           return if events.empty? || span.nil?
 
-          Datadog::AppSec::RateLimiter.limit(:traces) do
+          Datadog::AppSec::RateLimiter.thread_local.limit do
             record_via_span(span, *events)
           end
         end
--- a/lib/datadog/appsec/rate_limiter.rb
+++ b/lib/datadog/appsec/rate_limiter.rb
@@ -1,59 +1,44 @@
 # frozen_string_literal: true
 
+require_relative '../core/rate_limiter'
+
 module Datadog
   module AppSec
-    # Simple per-thread rate limiter
-    # Since AppSec marks sampling to keep on a security event, this limits the flood of egress traces involving AppSec
+    # Per-thread rate limiter based on token bucket rate limiter.
+    #
+    # Since AppSec marks sampling to keep on a security event, this limits
+    # the flood of egress traces involving AppSec
     class RateLimiter
-      def initialize(rate)
-        @rate = rate
-        @timestamps = []
-      end
-
-      def limit
-        now = Time.now.to_f
-
-        loop do
-          oldest = @timestamps.first
-
-          break if oldest.nil? || now - oldest < 1
-
-          @timestamps.shift
-        end
-
-        @timestamps << now
-
-        if (count = @timestamps.count) <= @rate
-          yield
-        else
-          Datadog.logger.debug { "Rate limit hit: #{count}/#{@rate} AppSec traces/second" }
-        end
-      end
+      THREAD_KEY = :datadog_security_appsec_rate_limiter
 
       class << self
-        def limit(name, &block)
-          rate_limiter(name).limit(&block)
+        def thread_local
+          rate_limiter = Thread.current.thread_variable_get(THREAD_KEY)
+          return rate_limiter unless rate_limiter.nil?
+
+          Thread.current.thread_variable_set(THREAD_KEY, new(trace_rate_limit))
         end
 
         # reset a rate limiter: used for testing
-        def reset!(name)
-          Thread.current[:datadog_security_trace_rate_limiter] = nil
+        def reset!
+          Thread.current.thread_variable_set(THREAD_KEY, nil)
         end
 
-        protected
-
-        def rate_limiter(name)
-          case name
-          when :traces
-            Thread.current[:datadog_security_trace_rate_limiter] ||= RateLimiter.new(trace_rate_limit)
-          else
-            raise "unsupported rate limiter: #{name.inspect}"
-          end
-        end
+        private
 
         def trace_rate_limit
           Datadog.configuration.appsec.trace_rate_limit
         end
+      end
+
+      def initialize(rate)
+        @rate_limiter = Core::TokenBucket.new(rate)
+      end
+
+      def limit
+        return yield if @rate_limiter.allow?
+
+        Datadog.logger.debug { "Rate limit hit: #{@rate_limiter.current_window_rate} AppSec traces/second" }
       end
     end
   end
--- a/sig/datadog/core/rate_limiter.rbs
+++ b/sig/datadog/core/rate_limiter.rbs
@@ -1,34 +1,72 @@
 module Datadog
   module Core
     class RateLimiter
-      def allow?: (untyped size) -> nil
-      def effective_rate: () -> nil
-    end
-    class TokenBucket < RateLimiter
-      attr_reader rate: untyped
+      def allow?: (?::Integer size) -> void
 
-      attr_reader max_tokens: untyped
-      def initialize: (untyped rate, ?untyped max_tokens) -> void
-      def allow?: (untyped size) -> untyped
-      def effective_rate: () -> (::Float | untyped)
-      def current_window_rate: () -> (::Float | untyped)
-      def available_tokens: () -> untyped
+      def effective_rate: () -> void
+    end
+
+    class TokenBucket < RateLimiter
+      # This should be `::Numeric`, but it's used with `*` method on another
+      # `Numeric` which makes steep fail with an error `Ruby::UnresolvedOverloading`
+      # "Cannot find compatible overloading of method"
+      @rate: ::Float
+
+      @max_tokens: ::Numeric
+
+      @tokens: ::Numeric
+
+      @total_messages: ::Integer
+
+      @conforming_messages: ::Integer
+
+      # This should be `::Integer?` but steep can't see conditional branching
+      # which safe-guards us from errors like "undefined method for NilClass"
+      @prev_conforming_messages: ::Integer
+
+      # This should be `::Integer?` but steep can't see conditional branching
+      # which safe-guards us from errors like "undefined method for NilClass"
+      @prev_total_messages: ::Integer
+
+      # This should be `::Integer?`, but steep can't see conditional branching
+      # which safe-guards us from errors like "undefined method for NilClass"
+      @current_window: ::Numeric
+
+      @last_refill: ::Numeric
+
+      # This should be `::Numeric`, but has to follow `@rate` type definition
+      attr_reader rate: ::Float
+
+      attr_reader max_tokens: ::Numeric
+
+      def initialize: (::Numeric rate, ?::Numeric max_tokens) -> void
+
+      def allow?: (?::Integer size) -> bool
+
+      def effective_rate: () -> ::Float
+
+      def current_window_rate: () -> ::Float
+
+      def available_tokens: () -> ::Numeric
 
       private
 
-      def refill_since_last_message: () -> untyped
+      def refill_since_last_message: () -> void
 
-      def refill_tokens: (untyped size) -> untyped
+      def refill_tokens: (::Numeric size) -> void
 
-      def increment_total_count: () -> untyped
+      def increment_total_count: () -> void
 
-      def increment_conforming_count: () -> untyped
+      def increment_conforming_count: () -> void
 
-      def should_allow?: (untyped size) -> (false | true)
-      def update_rate_counts: (untyped allowed) -> untyped
+      def should_allow?: (?::Integer size) -> bool
+
+      def update_rate_counts: (bool allowed) -> void
     end
+
     class UnlimitedLimiter < RateLimiter
-      def allow?: (untyped _) -> true
+      def allow?: (?::Integer _) -> true
+
       def effective_rate: () -> ::Float
     end
   end
--- a/spec/datadog/appsec/event_spec.rb
+++ b/spec/datadog/appsec/event_spec.rb
@@ -45,7 +45,7 @@ RSpec.describe Datadog::AppSec::Event do
   describe '.record' do
     before do
       # prevent rate limiter to bias tests
-      Datadog::AppSec::RateLimiter.reset!(:traces)
+      Datadog::AppSec::RateLimiter.reset!
     end
 
     let(:options) { {} }
@@ -309,7 +309,7 @@ RSpec.describe Datadog::AppSec::Event do
       end
 
       it 'does not call the rate limiter' do
-        expect(Datadog::AppSec::RateLimiter).to_not receive(:limit)
+        expect_any_instance_of(Datadog::AppSec::RateLimiter).to_not receive(:limit)
 
         expect(trace).to_not be nil
       end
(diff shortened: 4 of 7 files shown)


spec/datadog/appsec/event_spec.rb
```diff
@@ -349,9 +353,11 @@
         end
       end
 
+      # NOTE: Allows 10% excessive traces due to differece in execution time
+      #       of the limited block of code on different machines
       it 'rate limits event recording' do
-        expect(described_class).to receive(:record_via_span).exactly(rate_limit).times.and_call_original
-
+        expect(described_class).to receive(:record_via_span)
+          .at_least(rate_limit).at_most(rate_limit + 5).times.and_call_original
```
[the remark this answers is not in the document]
p-datadog: I think the allowance is OK, I would just use `rate_limit * 1.1` (or `(rate_limit * 1.1).to_i` if an integer argument is required) to match the logic you described.

spec/datadog/appsec/rate_limiter_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+      let(:rate_limit) { 50 }
+      let(:spy_1) { spy('thread-1 workload') }
+      let(:spy_2) { spy('thread-2 workload') }
+
+      it 'creates separate rate limiter per thread' do
+        thread_1 = Thread.new do
+          Array.new(rate_limit * 2) do
+            described_class.limit(:traces) { spy_1.call }
+          end
+        end
```
p-datadog, line 22: Is it possible that thread_1 executes completely before thread_2 starts, given that MRI is not necessarily concurrent fairly? What do you think about adding some `sleep 0` in the workload?

p-datadog (approved):

spec/datadog/appsec/rate_limiter_spec.rb
Strech: Yes, it is, but the thing is even so, we will see limiting in second thread if they share the limiter. And the point is to show the split of limiter per thread.

y9v (approved):

spec/datadog/appsec/rate_limiter_spec.rb
ivoanjo: So I was summoned to this thread by seeing the idea of `sleep` in specs :rofl: 

`sleep`s tend to cause a lot of flaky specs in our test suite.

In the specific case we're discussing of `sleep 0`, I think you may want `Thread.pass` instead, since it's a stronger hint to the VM to "let other things run".

But to be honest, I think instead of this, you may want to go a bit more "white-box" testing, e.g. instead of testing that two threads can both use the rate limit, consider instead:
* Checking that two rate limit objects get created + (one per thread)
* Checking that each thread only plays with its own rate limit object

There's multiple ways of doing this, happy to help if it's not something you've done before :)

lib/datadog/appsec/rate_limiter.rb
```diff
@@ -1,35 +1,14 @@
...
+require_relative '../core/rate_limiter'
+
 module Datadog
   module AppSec
-    # Simple per-thread rate limiter
-    # Since AppSec marks sampling to keep on a security event, this limits the flood of egress traces involving AppSec
+    # Per-thread rate limiter based on token bucket rate limiter.
+    #
+    # Since AppSec marks sampling to keep on a security event, this limits
+    # the flood of egress traces involving AppSec
```
ivoanjo, line 10: Also, for what it's worth, note that `Thread#[]` in Ruby is not thread-local, or more specifically, it is fiber-local, so you can have N values for a single thread, if multiple fibers are in use. (But if no fibers are in use, you'll have 1 value:1 thread, e.g. thread-local-like behavior).

There's a different API in Ruby for actual thread-locals that are always 1:1 with threads, even in the presence of multiple fibers.

spec/datadog/appsec/rate_limiter_spec.rb
p-datadog: `Thread.pass` is what I should have said, I keep forgetting it exists.

Strech: @ivoanjo In 99.9% of the cases I'm against `sleep` in tests. And this test do exactly like you've described, Threads are using `spy` object and run more than allowed calls (x4 time in total) and I'm expecting both `spy` objects to receive similar amount of calls.

Because if they share the limiter, one of them will get less or non calls allowed.

lib/datadog/appsec/rate_limiter.rb
Strech: My bad, going to update the code to use Thread-local variable set/get

Strech: Fixed https://github.com/DataDog/dd-trace-rb/pull/3975/commits/e59589647bbbffb56e373df037596f084153cce2

spec/datadog/appsec/rate_limiter_spec.rb
Strech: Gotcha, let me try to do so.

Strech: Fixed https://github.com/DataDog/dd-trace-rb/pull/3975/commits/6ef9fd2ee66c4cdcf688c0a626f7bc45e986b9ba

Strech:
@ivoanjo I've adjusted the code to consider Thread locality instead of Fiber locality of the variable. Unfortunately, in 2.5.9 those functions lack a bit of behavior which is possible to find in 3.x version, like `thread_variable?` reacts different on `nil` value (in 2.5 it's considered as set and in 3.x it's considered as unset 🤣)

The test is also adjusted to be a white-box instead of the black-box.

If you have time, could you please take another look. Thanks 🙌🏼

spec/datadog/appsec/event_spec.rb
```diff
@@ -349,9 +353,11 @@
         end
       end
 
+      # NOTE: Allows 10% excessive traces due to differece in execution time
+      #       of the limited block of code on different machines
       it 'rate limits event recording' do
-        expect(described_class).to receive(:record_via_span).exactly(rate_limit).times.and_call_original
-
+        expect(described_class).to receive(:record_via_span)
+          .at_least(rate_limit).at_most(rate_limit * 1.1).times.and_call_original
```
Strech, line 360: I'm going to find a way to rework it, it brings flakiness 😞

Strech: Fixed with time freezing and emulating 1s processing time

lib/datadog/appsec/rate_limiter.rb
```diff
@@ -55,6 +39,16 @@ def trace_rate_limit
...
       end
+
+      def initialize(rate)
+        @rate_limiter = Core::TokenBucket.new(rate)
+      end
+
+      def limit
+        return yield if @rate_limiter.allow?
+
+        Datadog.logger.debug { "Rate limit hit: #{@rate_limiter.current_window_rate} AppSec traces/second" }
```
ivoanjo, line 50: Minor: This debug message seems like it can happen reasonably frequently. Do we need it around?

My (small) concern is that we don't currently have a fine-grained way of disabling some messages + sometimes to debug issues we need to ask customers to enable debug-level logging. Thus, having messages that are very noisy is somewhat annoying in that situation.

I do understand if you find this message quite important and want to keep it; I'm more explaining the trade-off we have here. (The better solution would be for our logging to not be as coarse-grained...)

lib/datadog/appsec/rate_limiter.rb
```diff
@@ -1,34 +1,15 @@
...
-        if (count = @timestamps.count) <= @rate
-          yield
-        else
-          Datadog.logger.debug { "Rate limit hit: #{count}/#{@rate} AppSec traces/second" }
-        end
-      end
+      TRACES_KEY = :datadog_security_traces_rate_limiter
 
       class << self
         def limit(name, &block)
```
ivoanjo, line 15: Minor: Should we get rid of `name`? It seems it's been around for a while and we never used it, so I think it may be worth dropping it since we can always re-add when we actually need it?

spec/datadog/appsec/rate_limiter_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+          described_class.limit(:traces) { 'workload' }
+        end
+
+        thread_1.join
+        thread_2.join
+
+        rate_limiter_1 = thread_1.thread_variable_get(:__spec_traces)
+        rate_limiter_2 = thread_2.thread_variable_get(:__spec_traces)
+
+        expect(rate_limiter_1.object_id).not_to eq(rate_limiter_2.object_id)
```
ivoanjo, line 26: Minor: I think this is equivalent?

```suggestion
        expect(rate_limiter_1).not_to be(rate_limiter_2)
```

sig/datadog/core/rate_limiter.rbs
ivoanjo, line 1: Minor: Note that this file is not being typechecked right now -- it's listed as ignore in the `Steepfile` so this is a bit of a no-op change (and may actually be silently wrong)

sig/datadog/core/rate_limiter.rbs
```diff
@@ -1,17 +1,42 @@
 module Datadog
   module Core
     class RateLimiter
-      def allow?: (untyped size) -> nil
+      def allow?: (?::Integer size) -> (false | true)
```
ivoanjo, line 4: Minor:

```suggestion
      def allow?: (?::Integer size) -> bool
```

sig/datadog/core/rate_limiter.rbs
```diff
@@ -1,17 +1,42 @@
...
+
       def effective_rate: () -> nil
```
ivoanjo, line 6: Minor: This is an interesting one... `nil` is more of a "what it is right now" because this is a bit of an abstract method. Should we perhaps use `-> void` to mark that this shouldn't be used by itself? Or `Float?`

sig/datadog/core/rate_limiter.rbs
```diff
@@ -1,17 +1,42 @@
...
+      @last_refill: untyped
+
       attr_reader rate: untyped
 
       attr_reader max_tokens: untyped
+
       def initialize: (untyped rate, ?untyped max_tokens) -> void
-      def allow?: (untyped size) -> untyped
+
+      def allow?: (?::Integer size) -> (false | true)
```
ivoanjo, line 34: Minor:

```suggestion
      def allow?: (?::Integer size) -> bool
```

sig/datadog/core/rate_limiter.rbs
```diff
@@ -1,17 +1,42 @@
...
+
       def effective_rate: () -> (::Float | untyped)
+
       def current_window_rate: () -> (::Float | untyped)
```
ivoanjo, line 38: Minor: I think this one will always be Float?

```suggestion
      def current_window_rate: () -> ::Float
```

sig/datadog/core/rate_limiter.rbs
```diff
@@ -24,11 +49,12 @@ module Datadog
 
       def increment_conforming_count: () -> untyped
 
-      def should_allow?: (untyped size) -> (false | true)
-      def update_rate_counts: (untyped allowed) -> untyped
+      def should_allow?: (?::Integer size) -> (false | true)
```
ivoanjo, line 52: Minor:

```suggestion
      def should_allow?: (?::Integer size) -> bool
```

sig/datadog/core/rate_limiter.rbs
```diff
@@ -24,11 +49,12 @@ module Datadog
 
       def increment_conforming_count: () -> untyped
 
-      def should_allow?: (untyped size) -> (false | true)
-      def update_rate_counts: (untyped allowed) -> untyped
```
ivoanjo, line 28: Minor: Wait, doesn't this method still exist? Why remove it?

ivoanjo (approved):
I had not yet given a pass on other files, so here's a smattering of comments on them >_>

Just a bunch of small stuff, nothing blocking :+1: LGTM otherwise

sig/datadog/core/rate_limiter.rbs
Strech: 🤔 I ran re-generating via and thought it will generate it well enough 

```
bundle exec rbs prototype rb lib/datadog/core/rate_limiter.rb > sig/datadog/core/rate_limiter.rbs 
```

lib/datadog/appsec/rate_limiter.rb
Strech: Since I inherit it, I wasn't sure about the whole rewrite because I also don't like an idea of "programming for the future", hence I didn't go full power. If you're saying that it makes sense, then I will shrink the class even more.

sig/datadog/core/rate_limiter.rbs
ivoanjo: Hmm... That's weird. What version of `rbs` are you running? I just tried it and it kept the function for me... :eyes: 

```
$ bundle exec rbs --version
rbs 3.6.1
```

lib/datadog/appsec/rate_limiter.rb
ivoanjo: Yeah... I don't have more context on why this was added initially than you may have, just more of an observation of "this code has been around long enough that we're even refactoring it now, and the future extensibility wasn't needed so far".

Usually I mark as "minor" things I leave up to you the final decision if you want to do anything about it or not, so I can understand if you want to close this one as-is and maybe revisit this later (/never) :)

sig/datadog/core/rate_limiter.rbs
Strech: Same version. I will spend a bit more time with RBS then.

lib/datadog/appsec/event.rb
```diff
@@ -52,7 +52,7 @@ def record(span, *events)
           # ensure rate limiter is called only when there are events to record
           return if events.empty? || span.nil?
 
-          Datadog::AppSec::RateLimiter.limit(:traces) do
+          Datadog::AppSec::RateLimiter.instance.limit do
```
p-datadog, line 55: This line matches Singleton usage and I thought `instance` referred to the only global instance. I suggest another name for it (maybe `local_instance` if you don't have better ideas).

p-datadog (commented):
Since there were types previously in the sig files changed by this PR, I would prefer to see at least those carried over to the new tree.

lib/datadog/appsec/event.rb
Strech: Agreed, still working on the name. I don't want to have bypass method `limit` and rather provide builder/access method to the configured instance.

Strech: I choose the `thread_local` name to reflect locality of the limiter to the client.

```ruby
Datadog::AppSec::RateLimiter.thread_local.limit do
  record_via_span(span, *events)
end

```

Strech:
@p-datadog @ivoanjo Thanks for the feedback 🙏🏼.

I've updated the sig files properly and also enable the `Core::RateLimiter` to be checked by Steep. I don't think it makes sense to move these changes in the separate PR because they definitely belongs here (as I'm updating the code)

I don't know what to do with debug message, but I get the point and definitely will try to come up with an idea, but maybe later.

If there is no objection, I would like to merge it and 🚢 it!

ivoanjo:
No concerns from my side! :bow:

PR #3976, opened 2024-10-07
Permit single-quoted strings in standard.rb

p-datadog:
**What does this PR do?**

This PR changes standard.rb rules to not enforce particular string quoting style.

**Motivation:**

I usually use single-quoted strings because they are easier to type (one use of pinky per quote instead of two). standard.rb is currently used by profiler and dynamic instrumentation and @ivoanjo is OK with string quote style being left alone.

**Additional Notes:**

None

**How to test the change?**
Not applicable

### DIFF
--- a/.standard_todo.yml
+++ b/.standard_todo.yml
@@ -9,6 +9,7 @@ ignore:
   - Style/TrailingCommaInArguments
   - Style/TrailingCommaInArrayLiteral
   - Style/TrailingCommaInHashLiteral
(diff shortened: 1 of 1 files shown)


ivoanjo (approved):
:+1: LGTM

No concerns from my side. In particular, this is a very "two-way door decision" -- at any point we can revisit it and then ask standardrb to reformat the code, so I think it's fine to not enforce this for now.

PR #3977, opened 2024-10-07
Use current released version of simplecov for MRI

p-datadog:
**What does this PR do?**

This PR changes to the most recent simplecov version for MRI. JRuby retains the patched simplecov tree.

**Motivation:**

It is annoying having to `bundle install` all the time because "simplecov version is not checked out".

**Additional Notes:**

The original patch was made for Ruby 2.3 (and 2.4?) and JRuby. It was submitted upstream in https://github.com/simplecov-ruby/simplecov/pull/972 but the focus there was on MRI 2.3 and 2.4, and the patch has not been accepted so far (3 years) due to targeting old Ruby versions. However, it is also needed for JRuby. 

I restricted the patch to JRubies only and made MRI use a regular simplecov release.

**How to test the change?**
I don't know if there are any requirements besides CI being green. Without the patch in place JRuby 9.3 and 9.4 configurations fail CI.


ivoanjo:
(I've triggered a re-run of the failed specs to skip that one)

p-datadog:
JRuby 9.3 and 9.4 are failing with flat_map call on nil.

ivoanjo:
(Btw I've opened https://github.com/DataDog/dd-trace-rb/pull/3980 do add a bit more logging to the flaky spec, and hopefully I'll be able to debug the issue with that extra info)

p-datadog:
The coverage task is still failing, I think what is happening is since JRuby doesn't generate branch coverage, attempting to merge JRuby and MRI results from MRI will fail without the patch in https://github.com/simplecov-ruby/simplecov/pull/972. Which means we always need that patch even on MRI.

ivoanjo:
To be honest, our patch is so small, perhaps we could carry our patch in-tree and load it as a monkey patch instead?

E.g. right now we're using an older version of simplecov from git just to be able to have that change -- with the monkey patch we could probably use the latest version instead.

p-datadog:
I support an overlay patch instead of using a forked repo.

PR #3981, opened 2024-10-07
DEBUG-2334 Dynamic instrumentation transport component

p-datadog:
**What does this PR do?**

This PR adds the transport component for dynamic instrumentation, which is a facade on top of the core HTTP transport. The DI transport is responsible for properly formatting the payloads for submission to the debugger agent endpoints and delegates the rest of work including actual network I/O to the core transport.

**Motivation:**

DI implementation for Ruby.

**Additional Notes:**

There may be further opportunities to DRY transport code later on. I originally started with something closer to direct Net::HTTP usage and the current version uses core transport to respect agent settings (host, port, timeout and TLS flag) and has relatively little code.

**How to test the change?**

Unit tests at this time.


lib/datadog/di/transport.rb
```diff
@@ -0,0 +1,67 @@
...
+      def send_request(desc, path, payload, headers: {})
+        # steep:ignore:start
+        env = OpenStruct.new(
+          path: path,
+          form: payload,
+          headers: headers,
+        )
+        # steep:ignore:end
+        response = client.post(env)
+        unless (200..299).cover?(response.code)
```
Strech, line 59: There is a method in [Response class](https://github.com/DataDog/dd-trace-rb/blob/f0dd28e5640dfa3b3c8666fa5d6910651eaa7478/lib/datadog/core/transport/http/adapters/net.rb#L120)

```suggestion
        unless response.ok?
```

spec/datadog/di/transport_spec.rb
```diff
@@ -0,0 +1,96 @@
+require "datadog/di/spec_helper"
+require "datadog/di/transport"
+
+RSpec.describe Datadog::DI::Transport do
+  let(:agent_settings) do
+    double('agent settings')
```
Strech, line 6: I think it makes sense to replace it with `instance_double` to avoid over stabbing or introducing non existing methods

spec/datadog/di/transport_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+  describe '.new' do
+    it 'creates an instance using agent settings' do
+      expect(agent_settings).to receive(:hostname).and_return('localhost')
+      expect(agent_settings).to receive(:port).and_return(8126)
+      expect(agent_settings).to receive(:timeout_seconds).and_return(1)
+      expect(agent_settings).to receive(:ssl).and_return(false)
+
+      expect(described_class.new(agent_settings)).to be_a(described_class)
+    end
+  end
```
Strech, line 18: Could we avoid this test because it's not testing much.

spec/datadog/di/transport_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+    describe '.send_diagnostics' do
+      let(:payload) do
+        {}
+      end
+
+      it 'does not raise exceptions' do
+        expect do
+          client.send_diagnostics(payload)
+        end.not_to raise_exception
+      end
```
Strech, line 52: There is not much value in the `let` statement, so we can shrink entire test to this:

```suggestion
      it 'does not raise exceptions' do
        expect { client.send_diagnostics({}) }.not_to raise_exception
      end
```

spec/datadog/di/transport_spec.rb
```diff
@@ -0,0 +1,96 @@
...
+      it 'does not raise exceptions' do
+        expect do
+          client.send_input(payload)
+        end.not_to raise_exception
+      end
+    end
+
+    context 'when agent is not listening' do
+      # Use a bogus port
+      let(:port) { 99999 }
```
Strech, line 69: If we use instance_double, we can re-stub it in a nice way in before block

```ruby
before { allow(setting).to receive(:port).and_return(99999) }
```

lib/datadog/di/transport.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      def initialize(agent_settings)
+        # Note that this uses host, port, timeout and TLS flag from
+        # agent settings.
+        @client = Datadog::Core::Transport::HTTP::Adapters::Net.new(agent_settings)
+      end
+
+      def send_diagnostics(payload)
+        event_payload = Datadog::Core::Vendor::Multipart::Post::UploadIO.new(
+          StringIO.new(JSON.dump(payload)), 'application/json', 'event.json'
```
Strech, line 35: Just to get some insights, how deep payload could go?

p-datadog: If you are referring to the size limit, I believe the limit on the receiving end is 1 MB but there aren't any limits as to what can be generated by the upstream code.

spec/datadog/di/transport_spec.rb
p-datadog: This is actually testing that the agent settings are used for the transport.

lib/datadog/di/transport.rb
Strech: Ok, so we don't limit nesting here, just worried about how big could be payload.

p-datadog: I added a bullet point to the future work section in the RFC covering this.

spec/datadog/di/transport_spec.rb
p-datadog: I want it to be clear what the payload is, hence it is declared separately.

spec/datadog/di/transport_spec.rb
p-datadog: Done.

spec/datadog/di/transport_spec.rb
p-datadog: I find a let block for the port to be clear in its intent. Your suggested line is not clear in that it is changing the port.

p-datadog:
As far as I can tell CI does not have a configuration that has a working datadog agent (that is not a system test). The non-system test configurations use "dd-apm-agent" which does not implement debugger endpoints needed for DI. As such the semi-integration test in this PR can be run locally with the environment variables set appropriately, and the functionality will be eventually tested by system tests.

lib/datadog/di/transport.rb
```diff
@@ -0,0 +1,67 @@
...
+          headers: {'content-type' => 'application/json'},)
+      end
+
+      private
+
+      attr_reader :client
+
+      def send_request(desc, path, payload, headers: {})
+        # steep:ignore:start
+        env = OpenStruct.new(
```
marcotc, line 52: I recommend not using `OpenStruct` here because it busts the method cache in CRuby when initialized. A Struct or Hash can be good alternatives.
Also, you can use existing transport classes in the `core` directory if they work for you.

marcotc (approved):
Overall it looks good! Just one comment about `OpenStruct`.

lib/datadog/di/transport.rb
p-datadog: I this as a follow-up item in the RFC but I have now moved it to be a required work item before initial release. I don't see how Hash will work, Struct I imagine should work and something from core I imagine ought to work as well but would require research to figure out how to use.

PR #3983, opened 2024-10-09
Fix crashtracking specs

lloeki:
**What does this PR do?**

Reinstate crashtracking specs by fixing them. 

**Motivation:**

- #3970 identified these specs as being unreliable and/or leaking threads.
- Coverage is both limited and relies on internal behaviours.
- Crashtracking changes to address #3954 will make these same specs fail.

**Additional Notes:**

**How to test the change?**

Well, specs!


ivoanjo:
I believe this is no longer relevant, since crashtracking now does not spawn a separate process until there's a crash (and this specs were testing the old separate process behavior).

I'm going to go ahead and close this one.

PR #3984, opened 2024-10-09
[PROF-10679] Add preview support for correlating profiling with otel ruby gem

ivoanjo:
**What does this PR do?**

This PR picks up from where #3510 left off and adds preview support for correlating profiling data with traces/spans being tracked by the otel ruby gem (including endpoint profiling). (And specifically this PR just adds a few things on top of #3510, but does not otherwise change it significantly).

Note that this differs from the recommended way of using opentelemetry/otel with dd-trace-rb, which is to follow the instructions from https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/otel_instrumentation/ruby/ .

The key difference is -- this PR makes code hotspots work even for setups that opt to not use `require 'datadog/opentelemetry'` (which we still recommend as the easier path).

The approach taken here is similar to #2342 and #3466: we peek inside the implementation of the otel gem to extract the information we need (namely the span id, local root span id, trace type, and trace endpoint). This approach is potentially brittle, which is why the code is written very defensively, with the aim of never breaking the application (or profiling) if something is off -- it just won't collect code hotspots.

This feature is off by default. This PR adds a new setting to control this feature:
* Via an environment variable (`DD_PROFILING_PREVIEW_OTEL_CONTEXT_ENABLED` -> set to `only`/`both`/`false`/`true`)
* Via code `c.profiling.advanced.preview_otel_context_enabled =` -> set to `"only"`/`"both"`/`true`/`false`

Setting this feature to `both` means that the profiler will automatically pick up context from either dd-trace-rb tracing or otel library; whereas setting it to `only` means that the profiler will skip dd-trace-rb tracing entirely and _only_ read context from the otel library.

(This was suggested by @alexjf in https://github.com/DataDog/dd-trace-rb/pull/3510#discussion_r1514692905, and also doubles as a way to raise confidence in releasing this since it allows us to have it off by default)

**Motivation:**

We have a customer interested in running this setup, so hopefully they'll be able to start using this feature.

**Additional Notes:**

**For this feature to work, otel traces need to be tagged with the dd-trace-rb runtime-id**. See below for an example of how to do this.

If the runtime-id is not set up, it will not be possible to see profiles scoped to individual requests in the datadog UX.

**How to test the change?**

On top of the added test coverage, I was able to see code hotspots working for the following sinatra example app:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'rackup'
  gem 'dogstatsd-ruby'
  gem 'datadog', git: 'https://github.com/datadog/dd-trace-rb', branch: 'ivoanjo/prof-9263-otlp-ruby-code-hotspots'
  gem 'sinatra'
  gem 'opentelemetry-api'
  gem 'opentelemetry-sdk'
  gem 'opentelemetry-instrumentation-sinatra'
  gem 'opentelemetry-exporter-otlp'
  gem 'pry'
end

require 'sinatra/base'
require 'opentelemetry/sdk'
require 'pry'

Datadog.configure do |c|
  c.service = 'ivoanjo-testing-opentelemetry-test'
  c.profiling.enabled = true
  c.profiling.advanced.preview_otel_context_enabled = :only
end

# Configure OpenTelemetry
OpenTelemetry::SDK.configure do |c|
  c.service_name = 'ivoanjo-testing-opentelemetry-test'
  c.use 'OpenTelemetry::Instrumentation::Sinatra'
end

class MyApp < Sinatra::Base
  get '/' do
    OpenTelemetry::Trace.current_span.add_attributes({'runtime-id' => Datadog::Core::Environment::Identity.id})
    sleep 1
    'Hello, OpenTelemetry!'
  end
end

MyApp.run!
```

And here's how it looks:

> image

and 

> image


spec/datadog/profiling/collectors/thread_context_spec.rb
```diff
@@ -414,8 +417,10 @@ def another_way_of_calling_sample(profiler_overhead_stack_thread: Thread.current
           end
 
           before do
-            expect(@t1_span_id.to_i).to be > 0
-            expect(@t1_local_root_span_id.to_i).to be > 0
+            unless allow_invalid_ids
+              expect(@t1_span_id.to_i).to be > 0
+              expect(@t1_local_root_span_id.to_i).to be > 0
```
y9v, line 422: matter of personal preference probably, but this would also work:
```ruby
expect(@t1_span_id.to_i).to be_positive
```

y9v (approved):

ext/datadog_profiling_native_extension/collectors_thread_context.c
```diff
@@ -1551,6 +1613,124 @@ static VALUE _native_sample_skipped_allocation_samples(DDTRACE_UNUSED VALUE self
...
+  VALUE thread,
+  struct trace_identifiers *trace_identifiers_result
+) {
+  VALUE context_storage = rb_thread_local_aref(thread, otel_context_storage_id /* __opentelemetry_context_storage__ */);
+
+  // If it exists, context_storage is expected to be an Array[OpenTelemetry::Context]
+  if (context_storage == Qnil || !RB_TYPE_P(context_storage, T_ARRAY)) return;
+
+  VALUE otel_current_span_key = get_otel_current_span_key(state);
+  if (otel_current_span_key == Qfalse) return;
```
AlexJF, line 1648: Nit: Noting that all the other short-circuiting ifs are essentially of the form `if (xxx == Qnil)`. Only `otel_current_span_key` behaves differently. I can imagine some time in the future when this code is less fresh in our minds this could bite us back. Granted tests hopefully catch it but perhaps not a wild idea to flip the values we use as not-found and not-resolved markers so all short-circuiting ifs can be of the `== Qnil` form?

ext/datadog_profiling_native_extension/collectors_thread_context.c
```diff
@@ -1551,6 +1613,124 @@ static VALUE _native_sample_skipped_allocation_samples(DDTRACE_UNUSED VALUE self
...
+
+  // Now find the oldest span starting from the active span that still has the same trace id as the active span
+  for (int i = active_context_index - 1; i >= 0; i--) {
+    struct otel_span span_and_context = otel_span_and_context_from(rb_ary_entry(context_storage, i), otel_current_span_key);
+    VALUE span_context = span_and_context.span_context;
+    if (span_context == Qnil) return;
+
+    VALUE span_id = rb_ivar_get(span_context, at_span_id_id /* @span_id */);
+    VALUE span_trace_id = rb_ivar_get(span_context, at_trace_id_id /* @trace_id */);
+    if (span_id == Qnil || span_trace_id == Qnil || !RB_TYPE_P(span_id, T_STRING) || !RB_TYPE_P(span_trace_id, T_STRING)) return;
```
AlexJF, line 1674: Every time you call `otel_span_and_context_from` you end up doing these things (refer to lines 1655-1661 above).

Could we not include all of these into `otel_span_and_context_from`? It seems like it'd improve readability and maintainability a fair bit? Would also lead `otel_span` to be something like `{ .span_id, .trace_id, .span }` which also feels like a simpler API?

AlexJF, line 1674: Minor: Giving up on the whole thing at this point seems wasteful. 

I realize this is part of your strategy of assuming any unexpected thing meaning we may not be operating under the same constraints we considered when we first added this. But at this point we know we have a valid active_span and may have already found a chain of multiple valid spans.

Would it be that dangerous to `break` rather than `return` here and just carry on, taking the last valid span in the chain as the root span?

lib/datadog/profiling/component.rb
```diff
@@ -84,13 +84,24 @@ def self.build_profiler_component(settings:, agent_settings:, optional_tracer:)
...
       private_class_method def self.build_thread_context_collector(settings, recorder, optional_tracer, timeline_enabled)
+        otel_context_enabled =
+          case settings.profiling.advanced.preview_otel_context_enabled
+          when 'both'
+            :both
+          when 'only'
+            :only
+          else
+            false
+          end
```
AlexJF, line 95: Q: I was wondering if this was a style choice or if there's some limitation of the settings system that prevents us from using `:both`, `:only` and `false` directly as the settings normalized values?

AlexJF (approved):

ivoanjo:
Thanks folks! As we want to get this out asap, I'll merge the PR as-is and open a follow-up PR to address your feedback.

spec/datadog/profiling/collectors/thread_context_spec.rb
ivoanjo: It's not a very strong reason, but I like the > 0 as being more explicit if zero is included or not.

In Ruby, 0 is neither positive nor negative...  But you can still get a few weirdnesses:

```
[14] pry(main)> 0.0
=> 0.0
[15] pry(main)> -0.0
=> -0.0
[16] pry(main)> (0.0).object_id
=> -9223372036854775806
[17] pry(main)> (-0.0).object_id
=> 2660
[19] pry(main)> [0.0, -0.0].map(&:positive?)
=> [false, false]
[20] pry(main)> [0.0, -0.0].map(&:negative?)
=> [false, false]
```

so IDK, I feel like > 0 is ultra-clear while positive may leave you wondering :)

y9v: you are right, I didn't know about `-0.0.negative?` thing

ivoanjo: I guess it makes sense for `-0.0` have `negative? => false` because `-0.0 == 0.0` but yea... computers :rofl:

ext/datadog_profiling_native_extension/collectors_thread_context.c
ivoanjo: Fixed in 2b99cb4473d33ea39b89392ac87456c512ef284c (PR #3987)

ext/datadog_profiling_native_extension/collectors_thread_context.c
ivoanjo: Yup, it makes for a nice simplification in 0c124ec496fc15c77ce85d44b18d435fcfc827e0 (PR #3987)

ivoanjo: Hmmm it's a reasonable point.

My thinking is twofold:

* If we get the wrong local root span id then I think we'd be sending incorrect data downstream. I can imagine us trying to investigate "hey, why does this span not show up correctly when 'span and children' is selected, but if I select it directly it works?" and how hard it would be to track it down to us being lenient here.

* If this class gets changed, then potentially we won't be able to extract data from any. It seems unlikely (but not impossible, I'll admit), that we'll have some instances with the data, but not others. So I'm not sure if in practice this "mixed situation would happen"

...so I'm not entirely convinced it's worth trying to be lenient here :thinking:

lib/datadog/profiling/component.rb
ivoanjo: Fixed in 222b4d4b9ea60e48c6ad8319e0aa85cf5a747b31 (PR #3987). To be honest I wasn't sure about some of the details in the settings system, but on a second pass it works quite nicely.

PR #3985, opened 2024-10-09
Update integration test READMEs

sarahchen6:
**What does this PR do?**
This PR updates the integration test READMEs to address issues when running the tests on aarch64 / arm64 architecture machines (e.g. mac). 

The images by default build on x86_64 / amd64 architecture, so those not running a compatible machine need to specify the aarch64 / arm64 platform when building the image. When running the test scripts, macOS users may also need to disable the automatic use of the buildx docker command, so a fix for this is noted in the updated READMEs as well.

**Motivation:**
I was trying to run integration tests on my local mac, and I encountered issues with image architecture build incompatibilities. Notably, I found that with my macOS (aarch64 / arm64 architecture), even though I had a local image with the same name, the scripts were unable to "find" that image locally because it was built on an incompatible architecture. I was able to run the tests by specifying the platform when building the images (i.e. prefix with `DOCKER_DEFAULT_PLATFORM=linux/arm64`) and disabling the automatic use of the buildx docker command on macOS (i.e. prefix with `DOCKER_BUILDKIT=0`). Since macOS and generally aarch64 / arm64 architecture is widely used, I thought that notes for this fix could be helpful to add to our READMEs.

**Additional Notes:**

**How to test the change?**
This change only modifies README documents; however, you can test the newly added advice locally on a mac. Running the original README commands (e.g. `./script/build-images`) produces an error along the lines of `InvalidBaseImagePlatform: Base image datadog/dd-apm-demo:ruby-2.7 was pulled with platform "linux/amd64", expected "linux/arm64" for current build`. After adding the `DOCKER_DEFAULT_PLATFORM=linux/arm64` prefix, the image should successfully build.

### DIFF
--- a/integration/apps/rspec/README.md
+++ b/integration/apps/rspec/README.md
@@ -10,7 +10,7 @@ Install [direnv](https://github.com/direnv/direnv) for applying local settings.
 
 1. `cp .envrc.sample .envrc` and add your Datadog API key.
 2. `direnv allow` to load the env var.
-4. `docker-compose run --rm app bin/setup`
+3. `docker-compose run --rm app bin/setup`
 
 ## Running the application
 
@@ -78,7 +78,7 @@ e.g. `DD_DEMO_ENV_FEATURES=tracing,profiling`
 
 ### Running integration tests
 
-You can run integration tests using the following and substituting for the Ruby major and minor version (e.g. `2.7`)
+You can run integration tests using the following and substituting for the Ruby major and minor version (e.g. `2.7`). Currently, this integration test has nothing to do in CI, and there are no RSpec tests in `/bin/test`.
 
 ```sh
 ./script/build-images -v <RUBY_VERSION>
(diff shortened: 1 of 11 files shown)


integration/apps/rspec/README.md
```diff
@@ -78,7 +78,7 @@ e.g. `DD_DEMO_ENV_FEATURES=tracing,profiling`
 
 ### Running integration tests
 
-You can run integration tests using the following and substituting for the Ruby major and minor version (e.g. `2.7`)
+You can run integration tests using the following and substituting for the Ruby major and minor version (e.g. `2.7`). Currently, this integration test has nothing to do in CI, and there are no RSpec tests in `/bin/test`.
```
anmarchenko, line 81: This one is a great find, these tests are leftovers of ddtrace => datadog-ci migration, we need to remove this folder completely. I will do it in a separate PR.

marcotc (approved):
Thank you so much for the clean up!

p-datadog (approved):

PR #3986, opened 2024-10-10
[PROF-10589] Fix GVL profiling bug that was causing test flakiness

ivoanjo:
**What does this PR do?**

This PR fixes a small bug in GVL profiling that was causing test flakiness such as
https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/16739/workflows/fe009eb1-d942-4432-bc39-713fa9a3b751/jobs/600523

The TL;DR is that it's possible for the "Waiting for GVL" state to be entered in parallel with a sample. This means that the wall-time timestamp used by the profiler trying to take a sample is actually < the wall-time timestamp where the "Waiting for GVL" started.

This was causing the profiler to generate samples with wall-time of zero, which means:
1. Samples that are not useful are created
2. These samples, which were "Waiting for GVL" samples, were not showing up as such because the stack collector did not apply states to samples with wall-time of zero.

These two things together were confusing our
"Datadog::Profiling::Collectors::CpuAndWallTimeWorker#start when main thread is sleeping but a background thread is working when GVL profiling is enabled records Waiting for GVL samples"
spec, because that spec assumed that all samples for the `background_thread_affected_by_gvl_contention` were either "Waiting for GVL" or "on cpu", but we were getting a bunch of "unknown" state samples.

**Motivation:**

This avoids producing samples which are not useful and also gets rid of a source of flakiness for a test.

**Additional Notes:**

It's a bit hard to test some of this behavior using unit tests, so I'm relying on the integration-style checks to catch it.

**How to test the change?**

The main test for this behavior is the spec being re-activated.

### DIFF
--- a/ext/datadog_profiling_native_extension/collectors_stack.c
+++ b/ext/datadog_profiling_native_extension/collectors_stack.c
@@ -217,7 +217,10 @@ void sample_thread(
   ddog_prof_Label *state_label = labels.state_label;
   bool cpu_or_wall_sample = values.cpu_or_wall_samples > 0;
   bool has_cpu_time = cpu_or_wall_sample && values.cpu_time_ns > 0;
-  bool only_wall_time = cpu_or_wall_sample && values.cpu_time_ns == 0 && values.wall_time_ns > 0;
+  // Note: In theory, a cpu_or_wall_sample should always have some wall-time. In practice, the first sample for a thread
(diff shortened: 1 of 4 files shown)


AlexJF (approved):

ivoanjo:
...and it flaked again! Guess I need to spend some more time on this one :)

AlexJF (approved):

PR #3661, opened 2024-05-23
Support stats "yjit_alloc_size"

MITSUBOSHI:
**2.0 Upgrade Guide notes**

**What does this PR do?**

**Motivation:**

**Additional Notes:**

**How to test the change?**

Unsure? Have a question? Request a review!

PR #3991, opened 2024-10-10
Support stats "ratio_in_yjit"

jethrodaniel:
**What does this PR do?**

When ruby's extended YJIT stats are available (`--yjit-stat`), then we have access to numerous additional stats.

This adds `:ratio_in_yjit`, which is only present when additional stats are enabled.

Example:

```shell
RUBYOPT='--yjit --yjit-stats=quiet' ruby -e '
  def foo = "a" * 10
  300.times{foo}
  puts "ratio_in_yjit: #{::RubyVM::YJIT.runtime_stats[:ratio_in_yjit]}"
'
ratio_in_yjit: 0.9550806290358625
```

**Motivation:**

This should allow for a better understanding of how to tune `--yjit-exec-mem-size`.

From https://github.com/ruby/ruby/blob/ef084cc8f4958c1b6e4ead99136631bef6d8ddba/doc/yjit/yjit.md?plain=1#L213-L223:

> If you start Ruby with `--yjit-stats`, e.g. using an environment variable `RUBYOPT=--yjit-stats`,
> `RubyVM::YJIT.runtime_stats[:ratio_in_yjit]` shows the ratio of YJIT-executed instructions in %.
> Ideally, `ratio_in_yjit` should be as large as 99%, and increasing `--yjit-exec-mem-size` often
> helps improving `ratio_in_yjit`.

**Additional Notes:**

Fwiw, I directly based this on the previous YJIT PRs, which were very helpful here:
- https://github.com/DataDog/dd-trace-rb/pull/2711
- https://github.com/DataDog/dd-trace-rb/pull/3661

`ratio_in_yjit` has been available sine 3.3.0: https://github.com/ruby/ruby/blob/caf0d2058aa223515897401ff94e11e1c0671ce0/doc/NEWS/NEWS-3.3.0.md?plain=1#L436

**How to test the change?**

```
RUBYOPT="--yjit --yjit-stats=quiet" bundle exec rake spec:yjit

bundle exec rake spec:yjit # skipped
```

### DIFF
--- a/.github/workflows/test-yjit.yaml
+++ b/.github/workflows/test-yjit.yaml
@@ -11,10 +11,40 @@ jobs:
           - '3.2'
           - '3.3'
           # ADD NEW RUBIES HERE
-    name: Test (${{ matrix.os }}, ${{ matrix.ruby }})
+    name: Test YJIT (${{ matrix.os }}, ${{ matrix.ruby }})
     runs-on: ${{ matrix.os }}
     env:
       RUBYOPT: "--yjit"
+      SKIP_SIMPLECOV: 1
+      DD_INSTRUMENTATION_TELEMETRY_ENABLED: false
+      DD_REMOTE_CONFIGURATION_ENABLED: false
+    steps:
+      - uses: actions/checkout@v4
+      # bundler appears to match both prerelease and release rubies when we
+      # want the former only. relax the constraint to allow any version for
+      # head rubies
+      - if: ${{ matrix.ruby == 'head' }}
+        run: sed -i~ -e '/spec\.required_ruby_version/d' datadog.gemspec
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true # runs 'bundle install' and caches installed gems automatically
+          bundler: latest # needed to fix issue with steep on Ruby 3.0/3.1
+          cache-version: v2 # bump this to invalidate cache
+      - run: bundle exec rake spec:yjit
+  test-yjit-stats:
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+        ruby:
+          - '3.3'
+          # ADD NEW RUBIES HERE
+    name: Test YJIT Stats (${{ matrix.os }}, ${{ matrix.ruby }})
+    runs-on: ${{ matrix.os }}
+    env:
+      RUBYOPT: "--yjit --yjit-stats=quiet"
       SKIP_SIMPLECOV: 1
       DD_INSTRUMENTATION_TELEMETRY_ENABLED: false
       DD_REMOTE_CONFIGURATION_ENABLED: false
--- a/.github/workflows/test-yjit.yaml
+++ b/.github/workflows/test-yjit.yaml
@@ -10,41 +10,14 @@ jobs:
         ruby:
           - '3.2'
           - '3.3'
+        rubyopt:
+          - '--yjit'
+          - '--yjit --yjit-stats=quiet'
           # ADD NEW RUBIES HERE
-    name: Test YJIT (${{ matrix.os }}, ${{ matrix.ruby }})
+    name: Test YJIT (${{ matrix.os }}, ${{ matrix.ruby }} ${{ matrix.rubyopt }})
     runs-on: ${{ matrix.os }}
     env:
-      RUBYOPT: "--yjit"
-      SKIP_SIMPLECOV: 1
-      DD_INSTRUMENTATION_TELEMETRY_ENABLED: false
-      DD_REMOTE_CONFIGURATION_ENABLED: false
-    steps:
-      - uses: actions/checkout@v4
-      # bundler appears to match both prerelease and release rubies when we
-      # want the former only. relax the constraint to allow any version for
-      # head rubies
-      - if: ${{ matrix.ruby == 'head' }}
-        run: sed -i~ -e '/spec\.required_ruby_version/d' datadog.gemspec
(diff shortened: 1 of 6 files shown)


lib/datadog/core/environment/yjit.rb
```diff
@@ -52,11 +52,20 @@ def yjit_alloc_size
...
+
         def available?
           defined?(::RubyVM::YJIT) \
             && ::RubyVM::YJIT.enabled? \
             && ::RubyVM::YJIT.respond_to?(:runtime_stats)
         end
+
+        def stats_available?
+          available? && ::RubyVM::YJIT.respond_to?(:stats_enabled?)
+        end
```
ivoanjo, line 68: Minor: Since this is only for the tests to use, I think it's better to keep it there, otherwise it looks like a forgotten/abandoned method (e.g. "why doesn't `available?` call `stats_available?` comes to mind as a question when looking at this without context).

.github/workflows/test-yjit.yaml
```diff
@@ -32,3 +32,33 @@ jobs:
...
+      matrix:
+        os:
+          - ubuntu-latest
+        ruby:
+          - '3.3'
+          # ADD NEW RUBIES HERE
+    name: Test YJIT Stats (${{ matrix.os }}, ${{ matrix.ruby }})
+    runs-on: ${{ matrix.os }}
+    env:
+      RUBYOPT: "--yjit --yjit-stats=quiet"
```
jethrodaniel, line 47: I just copied the job, seemed easier/more maintainable than trying to fit this logic into the matrix using a single job.

Also, I don't see these actions in https://github.com/DataDog/dd-trace-rb/actions, so I'm guessing they need to be approved by a maintainer first.

.github/workflows/test-yjit.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
           - '3.2'
           - '3.3'
           # ADD NEW RUBIES HERE
-    name: Test (${{ matrix.os }}, ${{ matrix.ruby }})
+    name: Test YJIT (${{ matrix.os }}, ${{ matrix.ruby }})
     runs-on: ${{ matrix.os }}
     env:
       RUBYOPT: "--yjit"
```
ivoanjo, line 17:
```suggestion
      RUBYOPT: ${{ matrix.rubyopt }}
```

.github/workflows/test-yjit.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
           - '3.2'
           - '3.3'
           # ADD NEW RUBIES HERE
-    name: Test (${{ matrix.os }}, ${{ matrix.ruby }})
+    name: Test YJIT (${{ matrix.os }}, ${{ matrix.ruby }})
```
ivoanjo, line 14:
```suggestion
    name: Test YJIT (${{ matrix.os }}, ${{ matrix.ruby }} ${{ matrix.rubyopt }})
```

.github/workflows/test-yjit.yaml
ivoanjo: Yeah, we need to allow it on our side! I'll press the magic button :)

That said, I've suggested a matrix solution for this -- while I don't think it's the best solution every time, it does seem relatively OK on this case (vs having to maintain this whole other config?)

spec/datadog/core/runtime/metrics_spec.rb
ivoanjo, line 1: Looking again at line 204

```ruby
              skip('Test only runs with YJIT enabled and RubyVM::YJIT.stats_enabled? false')
```

...that doesn't look right -- we can report these metrics regardless of `stats_enabled?` being enabled or not.

Since we're extending the spec, do you mind updating that to remove the `stats_enabled?` reference?

ivoanjo (commented):
I've triggered the CI run and left a few notes. TBH I think this could go as-is, but I've suggested a few improvements ;)

.github/workflows/test-yjit.yaml
```diff
@@ -11,7 +11,7 @@ jobs:
           - '3.2'
           - '3.3'
           # ADD NEW RUBIES HERE
```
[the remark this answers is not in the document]
jethrodaniel: Ooh, I like this solution a lot more, many thanks.

ivoanjo (approved):
:+1: LGTM. I realized my suggestion for the matrix did not address the part where 3.2 doesn't support `--yjit-stats=quiet` so I pushed a tiny change to address that.

I'm just waiting for CI to settle and I'll merge this in. Thanks for the nice contribution! :pray:

.github/workflows/test-yjit.yaml
```diff
@@ -11,10 +11,16 @@ jobs:
...
-    name: Test (${{ matrix.os }}, ${{ matrix.ruby }})
+        rubyopt:
+          - '--yjit'
+          - '--yjit --yjit-stats=quiet'
+        exclude:
+          - ruby: '3.2'
+            rubyopt: '--yjit --yjit-stats=quiet'
```
p-datadog, line 19: Why is this configuration excluded?

jethrodaniel: `--yjit-stats` exists on ruby 3.2, but it dumps all the stat info after the interpreter exits, and it doesn't include the new metric added (`ratio_in_yjit`, only in 3.3+), so having that for 3.2 isn't valuable here.

`--yjit-stats=quiet` was introduced in 3.3, and _doesn't_ dump that info to stdout.

I can put up a PR with a comment for clarity.

jethrodaniel: for your consideration: https://github.com/DataDog/dd-trace-rb/pull/4010

PR #3999, opened 2024-10-14
Remove Rails version check from active_record/app.rb

RicardoTrindade:
**What does this PR do?**
Remove a Rails gem version check in `spec/datadog/tracing/contrib/active_record/app.rb`

**Motivation:**
Since dd-trace-rb v2.0+ requires Rails 4+ then this check can be removed. I'm not sure if there's any additional cleanup that can be performed here (or even some gem cleanup?)

**How to test the change?**
Didn't see any tests covering this feature but since the test matrix does not install Rails below 4.2 then I would assume all tests pass. Also removed a `skip` condition that does not make sense anymore given test matrix is Rails 4.2+

### DIFF
--- a/spec/datadog/tracing/contrib/active_record/instantiation_spec.rb
+++ b/spec/datadog/tracing/contrib/active_record/instantiation_spec.rb
@@ -12,9 +12,6 @@ RSpec.describe 'ActiveRecord instantiation instrumentation' do
   let(:article) { Article.first }
 
   before do
-    if Gem::Version.new(ActiveRecord::VERSION::STRING) < Gem::Version.new('4.2')
-      skip 'ActiveRecord instantiation events were added in Rails 4.2'
-    end
     Article.create!(title: 'test')
 
     # Reset options (that might linger from other tests)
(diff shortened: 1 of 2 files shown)


spec/datadog/tracing/contrib/active_record/instantiation_spec.rb
```diff
@@ -12,9 +12,6 @@
   let(:article) { Article.first }
 
   before do
-    if Gem::Version.new(ActiveRecord::VERSION::STRING) < Gem::Version.new('4.2')
-      skip 'ActiveRecord instantiation events were added in Rails 4.2'
-    end
```
ivoanjo, line 17: I did spot that the public documentation lists Rails 4.0, but we're only testing with 4.2 and so indeed this is not needed.

@marcotc I think we should update the public docs to list 4.2 instead of 4.0?

ivoanjo (approved):
:+1: LGTM! Thanks for the nice clean-up; I've triggered CI and will merge once it's all green :)

PR #4001, opened 2024-10-15
Add lib injection support for Ruby 3.3

marcotc:
Support for Ruby 3.3 is present in the gem, the only thing needed for injection is to change the installation script to allow it to run with that version of Ruby.

Because the Ruby API does not change in incompatible ways in between patch versions, we declare our compatibility with the API version "3.3.0", thus supporting all 3.3.x versions.

**How to test the change?**

### DIFF
--- a/lib-injection/host_inject.rb
+++ b/lib-injection/host_inject.rb
@@ -83,11 +83,14 @@ else
           Bundler::SharedHelpers.in_bundle?
         end
 
+        # CRuby does *not* break API or ABI compatibility in TEENY version bumps.
+        # The ABI always sets the TEENY version to zero to reflect that: {MAJOR}.{MINOR}.0.
+        # https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/
         def runtime_supported?
-          major, minor, = RUBY_VERSION.split('.')
+          major, minor, = RUBY_VERSION.split('.') # {MAJOR}_{MINOR}_{TEENY}
           ruby_api_version = "#{major}.#{minor}.0"
 
-          supported_ruby_api_versions = ['2.7.0', '3.0.0', '3.1.0', '3.2.0'].freeze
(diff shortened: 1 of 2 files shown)


lib-injection/host_inject.rb
```diff
@@ -83,11 +83,14 @@ def in_bundle?
           Bundler::SharedHelpers.in_bundle?
         end
 
+        # CRuby does *not* break API or ABI compatibility in TEENY version bumps.
+        # The ABI always sets the TEENY version to zero to reflect that: {MAJOR}.{MINOR}.0.
+        # https://www.ruby-lang.org/en/news/2013/12/21/ruby-version-policy-changes-with-2-1-0/
         def runtime_supported?
-          major, minor, = RUBY_VERSION.split('.')
+          major, minor, = RUBY_VERSION.split('.') # {MAJOR}_{MINOR}_{TEENY}
           ruby_api_version = "#{major}.#{minor}.0"
```
p-datadog, line 91: What is the purpose of `.0` here? If only two components are used by the logic why not just have two components here?

p-datadog (approved):

TonyCTHsu (changes requested):
What is blocking?

- Need a custom Ruby 3.3 image to build gems in a static way
- We are currently dealing with package size constraint, which is a prerequisite of this PR.

marcotc:
@TonyCTHsu
> Need a custom Ruby 3.3 image to build gems in a static way
Where are these images stored?

ivoanjo:
> Where are these images stored?

Afaik the dockerfiles are those in https://github.com/DataDog/dd-trace-rb/tree/master/.gitlab . I'm not sure how the rest of the pipeline is set up.

It may be worth documenting this somewhere? E.g. so it's clear how we can every year add new Ruby versions for the injection.

TonyCTHsu:
Obsolete by: https://github.com/DataDog/dd-trace-rb/pull/4137

PR #4002, opened 2024-10-15
Remove Ruby 2.3 version checks in Configuration and workers

RicardoTrindade:
Similar to https://github.com/DataDog/dd-trace-rb/pull/3998
Removing a couple more mentions to Ruby 2.3 now that the gem requires 2.5+
  - lib/datadog/core/configuration.rb
  - lib/datadog/tracing/workers.rb
  - lib/datadog/core/remote/worker.rb

**How to test the change?**
I've updated the corresponding specs to also remove the version check as the test matrix is also not running ruby 2.3 anymore

There are still a couple more places that I'd like to cleanup, but I'm keeping the PRs short 🤞

### DIFF
--- a/lib/datadog/core/configuration.rb
+++ b/lib/datadog/core/configuration.rb
@@ -278,9 +278,7 @@ module Datadog
       def handle_interrupt_shutdown!
         logger = Datadog.logger
         shutdown_thread = Thread.new { shutdown! }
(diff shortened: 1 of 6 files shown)


p-datadog (approved):

p-datadog:
Thank you very much for the PR @RicardoTrindade

marcotc (approved):

PR #4007, opened 2024-10-16
Remove Ruby version checks from Gemfiles

RicardoTrindade:
Remove Ruby 2.3 version checks from Gemfiles as `dd-trace-rb` requires Ruby 2.5+.

- The issue mentioned in the redis gem version restriction has been closed so maybe it's worth looking into it in the future ?
- From what I understand the `Gemfile2-x` `Gemfile3-x` do not need to be updated

EDIT: Need to look into the protobuf failures, maybe missed something

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -73,8 +73,6 @@ gem 'dogstatsd-ruby', '>= 3.3.0', '!= 5.0.0', '!= 5.0.1', '!= 5.1.0'
 if RUBY_PLATFORM != 'java'
   if RUBY_VERSION >= '2.7.0' # Bundler 1.x fails to find that versions >= 3.8.0 are not compatible because of binary gems
     gem 'google-protobuf', ['~> 3.0', '!= 3.7.0', '!= 3.7.1']
-  elsif RUBY_VERSION >= '2.3.0'
-    gem 'google-protobuf', ['~> 3.0', '!= 3.7.0', '!= 3.7.1', '< 3.19.2']
   else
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -73,8 +73,6 @@ gem 'dogstatsd-ruby', '>= 3.3.0', '!= 5.0.0', '!= 5.0.1', '!= 5.1.0'
 if RUBY_PLATFORM != 'java'
   if RUBY_VERSION >= '2.7.0' # Bundler 1.x fails to find that versions >= 3.8.0 are not compatible because of binary gems
     gem 'google-protobuf', ['~> 3.0', '!= 3.7.0', '!= 3.7.1']
-  elsif RUBY_VERSION >= '2.3.0'
-    gem 'google-protobuf', ['~> 3.0', '!= 3.7.0', '!= 3.7.1', '< 3.19.2']
   else
     gem 'google-protobuf', ['~> 3.0', '!= 3.7.0', '!= 3.7.1', '< 3.8.0']
```
ivoanjo, line 77: > EDIT: Need to look into the protobuf failures, maybe missed something

You did! Haha :P

This logic reads from top to bottom, e.g.

* on modern rubies...
* on less modern rubies...
* on really old rubies 

and this PR removes the mid tier, not the bottom one ;)

ivoanjo (changes requested):

Gemfile
RicardoTrindade: Yep sorry was already a bit late for 😅 
Should be ok now 🤞  thanks for your help!

ivoanjo (approved):
:+1: LGTM, thanks for the cleanup

PR #4009, opened 2024-10-17
DEBUG-2334 duplicate mutable values when serializing for dynamic inst…

p-datadog:
…rumentation

**Change log entry**
None needed, DI is not usable by customers yet.

**What does this PR do?**

When serializing parameter values at method entry, the resulting snapshot should not change if the parameter is mutated during method execution. This currently only applies to string values as other values used as serializer output are not mutable.

This PR  duplicates the string values which achieves immutability with no additional API complexity but at the cost of duplicating all strings, even if they are not entry parameters and thus can be stored without duplication.

**Motivation:**
Correct capture of method arguments at method entry.

**Additional Notes:**

**How to test the change?**
Unit tests at this time.

### DIFF
--- a/lib/datadog/di/serializer.rb
+++ b/lib/datadog/di/serializer.rb
@@ -103,8 +103,7 @@ module Datadog
           serialized.update(value: value.to_s)
         when String, Symbol
           need_dup = false
-          value = case value
-          when String
+          value = if String === value
             # This is the only place where we duplicate the value, currently.
             # All other values are immutable primitives (e.g. numbers).
             # However, do not duplicate if the string is frozen, or if
--- a/lib/datadog/di/serializer.rb
+++ b/lib/datadog/di/serializer.rb
@@ -102,11 +102,15 @@ module Datadog
         when Integer, Float, TrueClass, FalseClass
           serialized.update(value: value.to_s)
         when String, Symbol
+          need_dup = false
           value = case value
           when String
             # This is the only place where we duplicate the value, currently.
             # All other values are immutable primitives (e.g. numbers).
-            value.dup
+            # However, do not duplicate if the string is frozen, or if
+            # it is later truncated.
+            need_dup = !value.frozen?
+            value
           else
             value.to_s
           end
@@ -114,7 +118,9 @@ module Datadog
           if value.length > max
             serialized.update(truncated: true, size: value.length)
             value = value[0...max]
+            need_dup = false
           end
+          value = value.dup if need_dup
           serialized.update(value: value)
         when Array
           if depth < 0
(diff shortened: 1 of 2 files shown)


lib/datadog/di/serializer.rb
```diff
@@ -92,7 +102,14 @@ def serialize_value(value, name: nil, depth: settings.dynamic_instrumentation.ma
...
+          when String
+            # This is the only place where we duplicate the value, currently.
+            # All other values are immutable primitives (e.g. numbers).
+            value.dup
+          else
+            value.to_s
+          end
           max = settings.dynamic_instrumentation.max_capture_string_length
           if value.length > max
             serialized.update(truncated: true, size: value.length)
```
ivoanjo, line 115: May be worth using https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/core/utils/safe_dup.rb or `-v` directly? :thinking: 

Also, it may be worth checking if truncation needs to happen first, otherwise we'll be creating yet another copy for the truncation :eyes:

p-datadog: Good point, if the string is to be truncated there is no need to dup it, and for frozen strings there is also no need to dup since they can't be mutated.

p-datadog: Done.

lib/datadog/di/serializer.rb
```diff
@@ -92,12 +102,25 @@ def serialize_value(value, name: nil, depth: settings.dynamic_instrumentation.ma
...
-          value = value.to_s
+          need_dup = false
+          value = case value
+          when String
+            # This is the only place where we duplicate the value, currently.
+            # All other values are immutable primitives (e.g. numbers).
+            # However, do not duplicate if the string is frozen, or if
+            # it is later truncated.
+            need_dup = !value.frozen?
+            value
```
marcotc, line 113: Is the util `Datadog::Core::Utils::SafeDup.frozen_or_dup` what you are looking for here?

marcotc (approved):
I left a comment in case we can reuse the existing util.

lib/datadog/di/serializer.rb
```diff
@@ -92,12 +102,25 @@ def serialize_value(value, name: nil, depth: settings.dynamic_instrumentation.ma
...
+          else
+            value.to_s
+          end
```
Strech, line 116: We can turn `case` statement into `if` because it's just it

```suggestion
          value = if value.is_a?(String)
            # This is the only place where we duplicate the value, currently.
            # All other values are immutable primitives (e.g. numbers).
            # However, do not duplicate if the string is frozen, or if
            # it is later truncated.
            need_dup = !value.frozen?
            value
          else
            value.to_s
          end
```

Strech (approved):
👍🏼

lib/datadog/di/serializer.rb
p-datadog: Replaced case with if.

PR #4011, opened 2024-10-17
DEBUG-2334 dynamic instrumentation probe notification builder

p-datadog:
**Change log entry**

None, DI is not customer-visible at this time.

**What does this PR do?**

This PR adds a component to create status and snapshot payloads for probes.

**Motivation:**

Initial DI implementation in Ruby.

**Additional Notes:**

Most of the test coverage is in integration tests for DI overall. This PR has minimal test coverage.

Correctness of generated payloads is meant to ultimately be verified by system tests.

**How to test the change?**


lib/datadog/di/probe_notification_builder.rb
```diff
@@ -0,0 +1,216 @@
...
+          value = binding.local_variable_get(name)
+          map[name] = value
+        end
+      end
+
+      def thread_id
+        thread = Thread.current
+        if thread.respond_to?(:native_thread_id)
+          # Ruby 3.1+
+          thread.native_thread_id
```
ivoanjo, line 209: Be careful with using `native_thread_id` here -- in Ruby 3.3+ ruby threads can move across OS threads (see [M:N announcement](https://www.ruby-lang.org/en/news/2023/12/25/ruby-3-3-0-released/#:~:text=of%20RubyKaigi.-,M%3AN%20thread%20scheduler,-M%3AN%20thread)). This behavior change is off by default in the main ractor but that may change...

lib/datadog/di/probe_notification_builder.rb
```diff
@@ -0,0 +1,216 @@
...
+          host: nil,
+          logger: {
+            name: probe.file,
+            method: probe.method_name || 'no_method',
+            thread_name: Thread.current.name,
+            thread_id: thread_id,
+            version: 2,
+          },
+          "dd.trace_id": 136035165280417366521542318182735500431,
+          "dd.span_id": 17576285113343575026,
```
ivoanjo, line 138: These look like placeholders -- should there be a FIXME or something here?

lib/datadog/di/probe_notification_builder.rb
p-datadog: I don't use it myself for anything and I made a note to inquire with my team what it is supposed to be.

lib/datadog/di/probe_notification_builder.rb
p-datadog: Good catch, I forgot about these completely - they probably need to be filled out somehow.

p-datadog: But, I  can do that in a separate PR.

lib/datadog/di/probe_notification_builder.rb
```diff
@@ -0,0 +1,212 @@
...
+        location = if probe.line?
+          actual_file = if probe.file
+            # Normally caller_locations should always be filled for a line probe
+            # but in the test suite we don't always provide all arguments.
+            actual_file_basename = File.basename(probe.file)
+            caller_locations&.detect do |loc|
+              # TODO record actual path that probe was installed into,
+              # perform exact match here against that path.
+              File.basename(loc.path) == actual_file_basename
+            end&.path || probe.file
```
ivoanjo, line 95: This seems like a slightly expensive thing to do for just the tests :thinking:. Or can there be situations in user code where this gets triggered?

lib/datadog/di/probe_notification_builder.rb
```diff
@@ -0,0 +1,212 @@
...
+            thread_name: Thread.current.name,
+            thread_id: thread_id,
+            version: 2,
+          },
+          # TODO add tests that the trace/span id is correctly propagated
+          "dd.trace_id": Datadog::Tracing.active_trace&.id,
+          "dd.span_id": Datadog::Tracing.active_span&.id,
+          ddsource: 'dd_debugger',
+          message: probe.template && evaluate_template(probe.template,
+            duration: duration ? duration * 1000 : nil),
```
ivoanjo, line 145: Minor: It may be worth encoding the units of `duration` somewhere in the variable names (`duration_ms`?) or using an explicit type, since units are getting mixed and converted a bit and it seems easy to make a mistake...

lib/datadog/di/probe_notification_builder.rb
```diff
@@ -0,0 +1,212 @@
...
+        caller_locations.map do |loc|
+          {fileName: loc.path, function: loc.label, lineNumber: loc.lineno}
+        end
+      end
+
+      def evaluate_template(template, **vars)
+        message = template.dup
+        vars.each do |key, value|
+          message.gsub!("{@#{key}}", value.to_s)
+        end
```
ivoanjo, line 178: Minor: Not sure if worth it, but occurred to me that we could turn the template into an input for [sprintf](https://ruby-doc.org/3.3.5/format_specifications_rdoc.html) or [`String#%`](https://ruby-doc.org/3.3.5/String.html#method-i-25) and thus we'd be able to template all vars at one.

sig/datadog/di/probe_notification_builder.rbs
```diff
@@ -0,0 +1,36 @@
...
+
+      def build_received: (Probe probe) -> untyped
+
+      def build_installed: (Probe probe) -> untyped
+
+      def build_emitting: (Probe probe) -> untyped
+
+      def build_executed: (Probe probe, ?trace_point: untyped?, ?rv: untyped?, ?duration: untyped?, ?callers: untyped?, ?args: untyped?, ?kwargs: untyped?, ?serialized_entry_args: untyped?) -> untyped
+
+      def build_snapshot: (Probe probe, ?rv: untyped?, ?snapshot: untyped?, ?duration: untyped?, ?callers: untyped?, ?args: untyped?, ?kwargs: untyped?, ?serialized_entry_args: untyped?) -> { service: untyped, :"debugger.snapshot" => { id: untyped, timestamp: untyped, evaluationErrors: ::Array[untyped], probe: { id: untyped, version: 0, location: untyped }, language: "ruby", stack: untyped, captures: untyped }, duration: untyped, host: nil, logger: { name: untyped, method: untyped, thread_name: untyped, thread_id: untyped, version: 2 }, :"dd.trace_id" => 136035165280417366521542318182735500431, :"dd.span_id" => 17576285113343575026, ddsource: "dd_debugger", message: untyped, timestamp: untyped }
```
ivoanjo, line 19: Minor: This return looks quite hard to maintain as-is. Perhaps worth either generalizing (into just `Hash`) or introducing a class for this?

lib/datadog/di/probe_notification_builder.rb
p-datadog: For the message, there is actually an AST that is supposed to be evaluated which contains complex expressions. This evaluation will not be implemented for initial DI implementation. I have the above substitution as a "quick demo" of variable evaluation, but this entire implementation is temporary.

I made a note to revisit this part of the code but I think there won't be a large gain realized here with % because it'll still require some parsing or massaging and a straightforward string replacement should be very fast (and users should only have a couple of variables at most in their log message, normally).

lib/datadog/di/probe_notification_builder.rb
p-datadog: The convention I use is when units are not specified, time is measured in seconds. In this particular case, DI specifies that  duration in the template should be measured in milliseconds, but the variable is called `duration`. Thus I am not sure how to add a suffix here.

lib/datadog/di/probe_notification_builder.rb
p-datadog: We need the path to the file in the payload, but the path should be determined during instrumentation as per the comment just above. I've hacked path matching and processing by doing basename comparisons but this isn't sufficient, for v1 I will have the actual path stored in the probe after instrumentation and nothing should be calculated here.

lib/datadog/di/probe_notification_builder.rb
ivoanjo: Even if the key itself in the data is duration, maybe our variable can be `duration_s` or `duration_secs`?

lib/datadog/di/probe_notification_builder.rb
p-datadog: I removed thread_id sending because currently DI does not need it.

lib/datadog/di/probe_notification_builder.rb
```diff
@@ -0,0 +1,217 @@
...
+        }
+      end
+
+      def format_callers(callers)
+        callers.map do |caller|
+          if caller =~ /\A([^:]+):(\d+):in `([^']+)'\z/
+            {
+              fileName: $1, function: $3, lineNumber: Integer($2),
+            }
+          else
```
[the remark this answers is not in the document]
p-datadog: Done, thank you for the pointer.

lib/datadog/di/probe_notification_builder.rb
p-datadog: I added a comment stating the units of duration for the method parameter. I personally have the convention that time units without an explicitly stated unit are in seconds, which I believe is how everything in Ruby standard library works, and to me adding a suffix for seconds is redundant. But if you want a suffix let's go with whichever one you prefer.

sig/datadog/di/probe_notification_builder.rbs
p-datadog: Repaired.

lib/datadog/di/probe_notification_builder.rb
Strech: I totally agree with @ivoanjo regarding naming with postfix values with different scale. Comments are nice, but code is better. Proper name will prevent from storing a wrong value scale in it, comment - not.

p-datadog: As mentioned in https://github.com/DataDog/dd-trace-rb/pull/4028 I would like to leave the time units alone until there is team consensus on how they should be indicated and all of the other pending DI code is merged.

Strech (approved):

PR #4012, opened 2024-10-17
DEBUG-2334 dynamic instrumentation instrumenter component

p-datadog:
**Change log entry**

None, DI is not yet customer-visible.

**What does this PR do?**

This component instruments target code, meaning when target method or target line is invoked, the specified callback will be called with information about the program state at the target point (e.g. method parameters for method probe, local variables for line probe).

**Motivation:**
Initial implementation of Ruby dynamic instrumentation.

**Additional Notes:**
The instrumenter itself is under 300 lines. Most of the diff is in the test suite.

**How to test the change?**
This PR includes unit tests, but a large part of instrumentation is dealing with code that is loaded after the instrumentation is installed (since much Ruby code is lazily loaded) and test coverage for this will be added later since it requires more DI components.


p-datadog:
This PR has some `xit` tests, I am planning on reviewing them to either make them work or delete them from the tree.

lib/datadog/di.rb
```diff
@@ -10,5 +16,42 @@ module Datadog
...
+      def deactivate_tracking!
+        code_tracker&.stop
+      end
+
+      # Returns whether code tracking is available.
+      # This method should be used instead of querying #code_tracker
+      # because the latter one may be nil.
+      def code_tracking_active?
+        code_tracker&.active? || false
+      end
```
Strech, line 54: I think steep complains because all 3 methods are order dependent. Without calling `activate_tracking!` instance variable will not be set. And because of that it could be `nil`.

Maybe we can follow common practice with reader like this

```ruby
def code_tracker
  @code_tracker ||= CodeTracker.new
end
```

I think after that Steep should be quite (and it will allow to untangle the order dependency)

lib/datadog/di/utils.rb
```diff
@@ -0,0 +1,39 @@
...
+          # Exact match is not possible here, meaning any matching path
+          # has to be longer than the suffix. Require full component matches,
+          # meaning either the first character of the suffix is a slash
+          # or the previous character in the path is a slash.
+          # For now only check for forward slashes for Unix-like OSes;
+          # backslash is a legitimate character of a file name in Unix
+          # therefore simply permitting forward or back slash is not
+          # sufficient, we need to perform an OS check to know which
+          # path separator to use.
+          !!
```
Strech, line 27: Is it needed?

sig/datadog/di.rbs
```diff
@@ -1,4 +1,5 @@
 module Datadog
   module DI
+    def self.code_tracker: () -> CodeTracker?
```
Strech, line 3: This is the reason of Steep complains, it's a nil-able return

Strech (approved):
I have a few minor notes regarding code checker variable, but despite that 🟢

lib/datadog/di.rb
```diff
@@ -10,5 +16,42 @@ module Datadog
...
+      # are loaded. If you definitely will not be instrumenting
+      # third-party libraries, activating tracking after third-party libraries
+      # have been loaded may improve lookup performance.
+      #
+      # TODO test that activating tracker multiple times preserves
+      # existing mappings in the registry
+      def activate_tracking!
+        @code_tracker ||= CodeTracker.new
+        # & is demanded by steep, code tracker is always not nil here.
+        code_tracker&.start
```
ivoanjo, line 35: Minor: I wonder if steep would be happy with

```suggestion
        (@code_tracker ||= CodeTracker.new).start
```

lib/datadog/di/instrumenter.rb
```diff
@@ -0,0 +1,301 @@
...
+            if rate_limiter.nil? || rate_limiter.allow?
+              # Arguments may be mutated by the method, therefore
+              # they need to be serialized prior to method invocation.
+              entry_args = if probe.capture_snapshot?
+                serializer.serialize_args(args, kwargs)
+              end
+              rv = nil
+              duration = Benchmark.realtime do # steep:ignore
+                rv = super(*args, **kwargs)
+              end
```
ivoanjo, line 108: I'm not sure this is delegation is correct -- there may be corner cases with args/keyword args; see https://eregon.me/blog/2019/11/10/the-delegation-challenge-of-ruby27.html for details. I think this is a quite sensitive part since getting it wrong means installing a probe can break the app.

(Also @lloeki quite likes this topic so I think he's the person to ping for a quick check on this bit ;D )

lib/datadog/di/utils.rb
```diff
@@ -0,0 +1,39 @@
...
+          if path.length > suffix.length && path.end_with?(suffix)
```
y9v, line 28: do we need a length check here? I thought `end_with?` would just return `false` when `path` is shorter than `suffix`?

spec/datadog/di/instrumenter_spec.rb
```diff
@@ -0,0 +1,641 @@
...
+require_relative 'hook_line'
+require_relative 'hook_method'
+
+RSpec.describe Datadog::DI::Instrumenter do
+  di_test
+
+  let(:observed_calls) { [] }
+
+  let(:settings) do
+    double('settings').tap do |settings|
```
y9v, line 12: what do you think about using `instance_double` here instead? It has a benefit that the test will fail if stubbed methods are not available in the specified class.

p-datadog: I think this is a good suggestion and after all of the pending DI code is merged I intend to go through the tests and replace `double` with `instance_double`. However, for settings,  I don't know what type to use for non-top-level settings since those use unnamed classes.

lib/datadog/di.rb
p-datadog: Yes, it seems to do the trick.

lib/datadog/di.rb
p-datadog: Code tracker should only exist if code tracking is active. Checking whether code tracker exists shouldn't create it.

lib/datadog/di/instrumenter.rb
p-datadog: This is one of my items to address before initial release, but the "shortcut" is I can just mandate Ruby 3.0+ for DI if this problem is difficult to solve.

lib/datadog/di/utils.rb
p-datadog: Yes, because the `if` doesn't have an `else` branch.

lib/datadog/di/utils.rb
p-datadog: The length check is to prevent suffix from matching the path exactly when (neither) starts with a slash. It filters out the case of equal length, not suffix being longer.

y9v (approved):

PR #4013, opened 2024-10-17
DEBUG-2334 enforce probe type validity

p-datadog:
**Change log entry**

None, DI is not yet shipped to customers.

**What does this PR do?**

Dynamic instrumentation currently supports 4 probe types (log, metric, span and span decoration). Initial implementation of Ruby DI will only support log probes.
To avoid installing the other probe types, potentially with incorrect results, add a validation of probe type and require that the probe is of the supported type
(currently, just the log probe).

**Motivation:**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/lib/datadog/di/probe.rb
+++ b/lib/datadog/di/probe.rb
@@ -31,11 +31,17 @@ module Datadog
     #
     # @api private
     class Probe
+      KNOWN_TYPES = %i[log].freeze
+
       def initialize(id:, type:,
         file: nil, line_no: nil, type_name: nil, method_name: nil,
         template: nil, capture_snapshot: false, max_capture_depth: nil, rate_limit: nil)
         # Perform some sanity checks here to detect unexpected attribute
         # combinations, in order to not do them in subsequent code.
+        unless KNOWN_TYPES.include?(type)
+          raise ArgumentError, "Unknown probe type: #{type}"
+        end
+
         if line_no && method_name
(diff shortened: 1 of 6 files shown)


lib/datadog/di/probe.rb
```diff
@@ -31,11 +31,17 @@ module DI
...
         template: nil, capture_snapshot: false, max_capture_depth: nil, rate_limit: nil)
         # Perform some sanity checks here to detect unexpected attribute
         # combinations, in order to not do them in subsequent code.
+        unless KNOWN_TYPES.include?(type)
+          raise ArgumentError, "Unknown probe type: #{type}"
+        end
+
         if line_no && method_name
           raise ArgumentError, "Probe contains both line number and method name: #{id}"
         end
```
ivoanjo, line 47: Out of curiosity, what happens if this error gets triggered? E.g. does the component that is applying probe configuration deal correctly with these exceptions?

(And here "correctly" is doing a lot of work -- I guess "correctly" would be to apply all the probes it can, ignore the ones it can't, and maybe not fill up the logs too much? Also, not try to apply more than once probes that were rejected?)

p-datadog: Exception boundaries are listed in the RFC, the one applicable to this exception is in the remote config processor when it is processing probe specifications to install.

For not trying repeatedly, this is an interesting question. I added it to my list of questions to research.

marcotc (approved):

PR #3906, opened 2024-09-12
[PROF-9470] Cleanup heap recorder after GC, time based

AlexJF:
This reverts commit b8cebb4.

**What does this PR do?**

**Motivation:**

**Additional Notes:**

**How to test the change?**

Unsure? Have a question? Request a review!
[PROF-9470] Remove unneded include

PR #4020, opened 2024-10-23
[PROF-9470] Align heap recorder cleanup with GC activity (second try)

ivoanjo:
**Change log entry**

Add setting to lower heap profiling memory use/latency by cleaning up young objects after Ruby GC

**What does this PR do?**

This PR adds a new mechanism to lower heap profiling overhead (memory and latency): the ability to clean young objects being tracked by the heap profiler after a Ruby GC cycle runs, rather than waiting for serialization time.

This mechanism is currently off by default and controlled by the setting `c.profiling.advanced.heap_clean_after_gc_enabled` / `DD_PROFILING_HEAP_CLEAN_AFTER_GC_ENABLED`

We plan to run a bit more validation before enabling it by default.

**Motivation:**

By doing these cleanups, we lower memory usage (we don't need to wait until until next serialization to notice an object was already collected), and latency, because there's less work to be done at serialization time.

**Additional Notes:**

This PR started from https://github.com/DataDog/dd-trace-rb/pull/3906; the main change from that earlier PR is that the heap recorder clean after GC behavior has been changed to be a best-effort mechanism (e.g. it's optional and not required for correctness).

**How to test the change?**

This change includes test coverage. I've also added a new benchmarking configuration to evaluate this new setting using our usual harness.

### DIFF
--- a/ext/datadog_profiling_native_extension/heap_recorder.c
+++ b/ext/datadog_profiling_native_extension/heap_recorder.c
@@ -156,11 +156,6 @@ struct heap_recorder {
   // mutation of the data so iteration can occur without acquiring a lock.
   // NOTE: Contrary to object_records, this table has no ownership of its data.
   st_table *object_records_snapshot;
-  // Is there a pending update due to a conflict with an ongoing iteration?
-  // If there is, we'll want to re-issue it after we finish the iteration.
-  bool pending_update;
-  // Was that pending update forced to be a full update?
-  bool pending_update_forced_full;
   // Are we currently updating or not?
   bool updating;
   // The GC gen/epoch/count in which we are updating (or last updated if not currently updating).
@@ -442,13 +437,11 @@ void heap_recorder_update(heap_recorder *heap_recorder, bool force_full_update) 
   }
 
   if (heap_recorder->object_records_snapshot != NULL) {
-    // If we're currently iterating, don't do the update now, we'll do it when we finish the iteration.
+    // While serialization is happening, it runs without the GVL and uses the object_records_snapshot.
     // Although we iterate on a snapshot of object_records, these records point to other data that has not been
-    // snapshotted for efficiency reasons (e.g. heap_records). Since there's a chance that updating may invalidate
-    // some of that non-snapshotted data, lets refrain from doing updates during iteration. This also enforces the
+    // snapshotted for efficiency reasons (e.g. heap_records). Since updating may invalidate
+    // some of that non-snapshotted data, let's refrain from doing updates during iteration. This also enforces the
     // semantic that iteration will operate as a point-in-time snapshot.
-    heap_recorder->pending_update = true;
-    heap_recorder->pending_update_forced_full = force_full_update;
     return;
   }
 
@@ -538,10 +531,6 @@ void heap_recorder_finish_iteration(heap_recorder *heap_recorder) {
 
   st_free_table(heap_recorder->object_records_snapshot);
   heap_recorder->object_records_snapshot = NULL;
-
-  if (heap_recorder->pending_update) {
-    heap_recorder_update(heap_recorder, heap_recorder->pending_update_forced_full);
-  }
 }
 
 // Internal data we need while performing iteration over live objects.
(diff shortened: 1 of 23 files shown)


ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -394,23 +428,91 @@ static VALUE end_heap_allocation_recording(VALUE end_heap_allocation_args) {
...
 
   st_foreach(heap_recorder->object_records, st_object_record_update, (st_data_t) heap_recorder);
 
+  heap_recorder->last_update_ns = now_ns;
+  heap_recorder->stats_lifetime.updates_successful++;
+
+  // Lifetime stats updating
+  // TODO: Discuss with Alex -- should we separate these out between young objects only and full updates?
+  // Tracking them together in this way seems to be muddying the waters -- young object updates will have more objects
+  // skipped, and different mixes of alive/dead
```
AlexJF, line 496: Yes, good point.

ext/datadog_profiling_native_extension/stack_recorder.c
```diff
@@ -330,6 +327,8 @@ static VALUE _native_new(VALUE klass) {
   // Note: Any exceptions raised from this note until the TypedData_Wrap_Struct call will lead to the state memory
   // being leaked.
 
+  state->heap_clean_after_gc_enabled = true;
```
AlexJF, line 330: Given the current status of this is default false, why are we setting to `true` here? I know it'll get overriden by `_native_initiatlize` but seems unnecessary?

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
```diff
@@ -897,6 +914,78 @@
...
+
+          # Let's replace the test_object reference with another object, so that the original one can be GC'd
+          test_object = Object.new # rubocop:disable Lint/UselessAssignment
+
+          # Force an update to happen on the next GC
+          Datadog::Profiling::StackRecorder::Testing._native_heap_recorder_reset_last_update(recorder)
+
+          GC.start
+
+          test_object_id
```
AlexJF, line 944: From previous flakes we may need to redo this a bunch of times just in case we have a ref in a register somewhere keeping it alive?

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -812,6 +822,107 @@ def create_obj_in_recycled_slot(should_sample_original: false)
...
         end
+
+        describe "#recorder_after_gc_step" do
+          def sample_and_clear
+            test_object = Object.new
+            test_object_id = test_object.object_id
+            sample_allocation(test_object)
+            # Let's replace the test_object reference with another object, so that the original one can be GC'd
+            test_object = Object.new # rubocop:disable Lint/UselessAssignment
+            GC.start
```
AlexJF, line 833: Similar flake potential as above?

AlexJF (approved):
LGTM! Nice test suite 👍

ivoanjo:
Note: The `dd-gitlab/benchmarks` step is failing because I've refactored the benchmarks to use `StackRecorder.for_testing` and thus running the benchmarks from this PR against the master branch does not work.

Once this PR is merged, the new `for_testing` method will also be available in the master branch so the benchmarks will start working again.

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
ivoanjo: Yeah, I wondered about that as well.

In this case I'm explicitly allocating a new object + changing the local variable + and this is in a separate method which was somewhat deliberate as "a bunch of random actions to hopefully flush any hidden/leftover references".

I guess we'll see if we get lucky or not?

spec/datadog/profiling/stack_recorder_spec.rb
ivoanjo: (See my note on the other one :smile: )

ext/datadog_profiling_native_extension/stack_recorder.c
ivoanjo: Yeah, you're right, in hindsight this may be a bit misleading. Changed in 6494eee31afbe68e9008804dde825bc3e79556c6 .

ext/datadog_profiling_native_extension/heap_recorder.c
ivoanjo: Split the stats in dd653cea919ed729daa8848d86b79dda36a524cc

AlexJF (approved):
LEBTBTM (Looks Even Better Than Before To Me)

PR #4023, opened 2024-10-24
[APPSEC-55378] Extract processor context into separate file

Strech:
**Change log entry**

Not needed.

**What does this PR do?**

This is a preparation PR which extracts `Datadog::AppSec::Processor::Context` into a separate file. **No changes** (yet 😏) to the logic or implementation were done.

**Motivation:**

We are going to work on those items soon and it makes sense to separate files.

**Additional Notes:**

None.

**How to test the change?**

Just CI.


sig/datadog/appsec/processor/context.rbs
```diff
@@ -0,0 +1,27 @@
+module Datadog
+  module AppSec
+    class Processor
+      class Context
+        type event = untyped
```
vpellan, line 5: I think `event` is a ::Hash[Symbol, untyped] according to where it's used (e.g.: https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/appsec/contrib/rack/gateway/watcher.rb#L47), I'm not sure it is more useful than declaring it as untyped though
```suggestion
        type event = ::Hash[Symbol, untyped]
```

vpellan (approved):
LGTM!

sig/datadog/appsec/processor/context.rbs
Strech: Maybe, I'm not sure that I fully understand the code to change it now, we have events from the `WAF` module too and so far they are always empty.

I think we can definitely change it as soon as we 100% sure.

PR #4024, opened 2024-10-24
fix: ensure uds always takes precedence over http if both configurations are defined

mabdinur:
**What does this PR do?**

Ensures uds configurations are parsed in a manner that is consistent with other libraries. 

**Motivation:**

- Ensures `DD_TRACE_AGENT_URL` is applied in a consistent manner.

**Change log entry**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/lib/datadog/core/configuration/agent_settings_resolver.rb
+++ b/lib/datadog/core/configuration/agent_settings_resolver.rb
@@ -234,6 +234,11 @@ module Datadog
         def should_use_uds?
           # When we have mixed settings for http/https and uds, we print a warning
           # and use the uds settings.
+          mixed_http_and_uds
+          can_use_uds?
+        end
+
+        def mixed_http_and_uds
           unless defined?(@mixed_http_and_uds)
             @mixed_http_and_uds = (configured_hostname || configured_port) && can_use_uds?
             if @mixed_http_and_uds
@@ -251,7 +256,6 @@ module Datadog
               )
             end
           end
-          can_use_uds?
         end
 
         def can_use_uds?
--- a/lib/datadog/core/configuration/agent_settings_resolver.rb
+++ b/lib/datadog/core/configuration/agent_settings_resolver.rb
@@ -240,7 +240,7 @@ module Datadog
 
         def mixed_http_and_uds
           return @mixed_http_and_uds if defined?(@mixed_http_and_uds)
-          
+
           @mixed_http_and_uds = (configured_hostname || configured_port) && can_use_uds?
           if @mixed_http_and_uds
             warn_if_configuration_mismatch(
@@ -256,7 +256,7 @@ module Datadog
               ]
             )
           end
-          
+
           @mixed_http_and_uds
         end
 
--- a/lib/datadog/core/configuration/agent_settings_resolver.rb
+++ b/lib/datadog/core/configuration/agent_settings_resolver.rb
@@ -232,7 +232,26 @@ module Datadog
         end
 
         def should_use_uds?
-          can_use_uds? && !mixed_http_and_uds?
+          # When we have mixed settings for http/https and uds, we print a warning
+          # and use the uds settings.
+          unless defined?(@mixed_http_and_uds)
+            @mixed_http_and_uds = (configured_hostname || configured_port) && can_use_uds?
+            if @mixed_http_and_uds
+              warn_if_configuration_mismatch(
+                [
+                  DetectedConfiguration.new(
+                    friendly_name: 'configuration of hostname/port for http/https use',
+                    value: "hostname: '#{hostname}', port: '#{port}'",
+                  ),
+                  DetectedConfiguration.new(
+                    friendly_name: 'configuration for unix domain socket',
+                    value: parsed_url.to_s,
+                  ),
+                ]
+              )
+            end
+          end
+          can_use_uds?
         end
 
         def can_use_uds?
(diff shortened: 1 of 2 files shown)


lib/datadog/core/configuration/agent_settings_resolver.rb
```diff
@@ -232,7 +232,26 @@ def uds_fallback
...
+                  ),
+                  DetectedConfiguration.new(
+                    friendly_name: 'configuration of hostname/port for http/https use',
+                    value: "hostname: '#{hostname}', port: '#{port}'",
+                  ),
+                ]
+              )
+            end
+          end
+          can_use_uds?
```
Strech, line 254: With these changes we violate SRP. Question-mark methods doesn't suppose to set any variables, but provide only true/false checks.

Solution: Rename existing method `mixed_http_and_uds?` into `mixed_http_and_uds` and adjust `should_use_uds?` accordingly.

Strech (changes requested):

ivoanjo:
Thanks for picking this up! I assume from

> Ensures uds configurations are parsed in a manner that is consistent with other libraries.

That other dd libraries behave in the way we're changing Ruby to adhere to.

Do you mind if I ask for some references? Is there perhaps a system-test for this that we can use to ensure we don't regress across libraries?

marcotc (approved):

spec/datadog/core/configuration/agent_settings_resolver_spec.rb
```diff
@@ -257,16 +248,13 @@
...
             end
 
-            it 'logs a warning including the hostname and port' do
+            it 'logs a warning including the uds configuration' do
               expect(logger).to receive(:warn)
-                .with(/
-                  Configuration\ mismatch:\ values\ differ\ between\ configuration.*
-                  Using\ "hostname:\ 'custom-hostname',\ port:\ '5678'".*
-                /x)
+                .with(%r{Configuration mismatch.*Using "unix:[/|]some/path"})
```
mabdinur, line 257:
```suggestion
                .with(%r{Configuration mismatch.*Using "unix:[\/]{1,3}some/path"})
```

spec/datadog/core/configuration/agent_settings_resolver_spec.rb
```diff
@@ -204,51 +201,45 @@
...
             end
 
-            it 'logs a warning including the default hostname and port' do
+            it 'logs a warning including the uds configuration' do
               expect(logger).to receive(:warn)
-                .with(/
-                  Configuration\ mismatch:\ values\ differ\ between\ configuration.*
-                  Using\ "hostname:\ '127.0.0.1',\ port:\ '5678'".*
-                /x)
+                .with(%r{Configuration mismatch.*Using "unix:[/|]some/path"})
```
mabdinur, line 242:
```suggestion
                .with(%r{Configuration mismatch.*Using "unix:[\/]{1,3}some/path"})
```

spec/datadog/core/configuration/agent_settings_resolver_spec.rb
```diff
@@ -170,32 +170,29 @@
...
             end
 
-            it 'logs a warning including the hostname and default port' do
+            it 'logs a warning including the uds path' do
               expect(logger).to receive(:warn)
-                .with(/
-                  Configuration\ mismatch:\ values\ differ\ between\ configuration.*
-                  Using\ "hostname:\ 'custom-hostname',\ port:\ '8126'".*
-                /x)
+                .with(%r{Configuration mismatch.*Using "unix:[/|]some/path"})
```
mabdinur, line 195:
```suggestion
                .with(%r{Configuration mismatch.*Using "unix:[\/]{1,3}some/path"})
```

spec/datadog/core/configuration/agent_settings_resolver_spec.rb
```diff
@@ -204,51 +201,45 @@
...
             end
 
-            it 'logs a warning including the hostname and port' do
+            it 'logs a warning including the uds configuration' do
               expect(logger).to receive(:warn)
-                .with(/
-                  Configuration\ mismatch:\ values\ differ\ between\ configuration.*
-                  Using\ "hostname:\ 'custom-hostname',\ port:\ '1234'".*
-                /x)
+                .with(%r{Configuration mismatch.*Using "unix:[/|]some/path"})
```
mabdinur, line 210:
```suggestion
                .with(%r{Configuration mismatch.*Using "unix:[\/]{1,3}some/path"})
```

ivoanjo:
> We have this system-test: https://github.com/DataDog/system-tests/blob/5e90616079fc11656348026f5f771aea8bafa748/tests/parametric/test_config_consistency.py#L145.

Amazing, thanks for sharing :heart:

lib/datadog/core/configuration/agent_settings_resolver.rb
```diff
@@ -232,7 +232,30 @@ def uds_fallback
...
+                    value: parsed_url.to_s,
+                  ),
+                  DetectedConfiguration.new(
+                    friendly_name: 'configuration of hostname/port for http/https use',
+                    value: "hostname: '#{hostname}', port: '#{port}'",
+                  ),
+                ]
+              )
+            end
+          end
```
Strech, line 258: Don't you mind if I offer to slightly redo the method because it's not returning `@mixed_http_and_uds` and we can spare nesting with guard statements:

```suggestion
          return @mixed_http_and_uds if defined?(@mixed_http_and_uds)
          
          @mixed_http_and_uds = (configured_hostname || configured_port) && can_use_uds?
          if @mixed_http_and_uds
            warn_if_configuration_mismatch(
              [
                DetectedConfiguration.new(
                  friendly_name: 'configuration for unix domain socket',
                  value: parsed_url.to_s,
                ),
                DetectedConfiguration.new(
                  friendly_name: 'configuration of hostname/port for http/https use',
                  value: "hostname: '#{hostname}', port: '#{port}'",
                ),
              ]
            )
          end
          
          @mixed_http_and_uds
```

Strech (approved):
I think it's LGTM if we fix reader method to return value.

lib/datadog/core/configuration/agent_settings_resolver.rb
p-datadog: The variable being set is just a cache for the value being returned or am I missing something?

ivoanjo: Memoizing values plays an important part when combined with the warnings: it means we don't repeat warnings just because some helper function queried the same info again.

PR #4028, opened 2024-10-24
DEBUG-2334 Probe Notifier Worker component

p-datadog:
**What does this PR do?**
This is a background thread that notification payloads (probe
status and probe snapshots) can be submitted to.
The payloads will be batched into groups if possible, and
sent to the local agent asynchronously.

**Motivation:**

Initial DI implementation.

**Change log entry**
None

**Additional Notes:**

**How to test the change?**
Unit tests in this PR


lib/datadog/di/probe_notifier_worker.rb
```diff
@@ -0,0 +1,242 @@
...
+            rescue => exc
+              raise if settings.dynamic_instrumentation.propagate_all_exceptions
+
+              warn "Error in probe notifier worker: #{exc.class}: #{exc} (at #{exc.backtrace.first})"
+            end
+            @lock.synchronize do
+              @wake_scheduled = more
+            end
+            wake.wait(more ? MIN_SEND_INTERVAL : nil)
+          end
```
Strech, line 81:
```suggestion
            begin
              more = maybe_send
            rescue => exc
              raise if settings.dynamic_instrumentation.propagate_all_exceptions

              warn "Error in probe notifier worker: #{exc.class}: #{exc} (at #{exc.backtrace.first})"
            end

            @lock.synchronize { @wake_scheduled = more }
            wake.wait(more ? MIN_SEND_INTERVAL : nil)
          end
```

lib/datadog/di/probe_notifier_worker.rb
```diff
@@ -0,0 +1,242 @@
...
+      # Stops the background thread.
+      #
+      # Attempts a graceful stop with the specified timeout, then falls back
+      # to killing the thread using Thread#kill.
+      def stop(timeout = 1)
+        @stop_requested = true
+        wake.signal
+        unless thread&.join(timeout)
+          thread.kill
+        end
```
Strech, line 94:
```suggestion
        thread.kill unless thread&.join(timeout)
```

lib/datadog/di/probe_notifier_worker.rb
```diff
@@ -0,0 +1,242 @@
...
+      # This method waits for the notification queue to become empty
+      # rather than for a particular set of notifications to be sent out,
+      # therefore, it should only be called when there is no parallel
+      # activity (in another thread) that causes more notifications
+      # to be generated.
+      def flush
+        loop do
+          if @thread.nil? || !@thread.alive?
+            return
+          end
```
Strech, line 109:
```suggestion
          return if @thread.nil? || !@thread.alive?
```

lib/datadog/di/probe_notifier_worker.rb
```diff
@@ -0,0 +1,242 @@
...
+          if io_in_progress
+            # If we just call Thread.pass we could be in a busy loop -
+            # add a sleep.
+            sleep 0.25
+            next
+          elsif queues_empty
+            break
+          else
+            sleep 0.25
+            next
```
Strech, line 124:
```suggestion
            next sleep(0.25)
```

lib/datadog/di/probe_notifier_worker.rb
```diff
@@ -0,0 +1,242 @@
...
+      def io_in_progress?
+        @io_in_progress
+      end
+
+      attr_reader :last_sent
+
+      [
+        [:status, 'probe status'],
+        [:snapshot, 'snapshot'],
+      ].each do |(event_type, event_name)|
```
Strech, line 145:
```suggestion
      {status: 'probe status', snapshot: 'snapshot'}.each do |event_type, event_name|
```

lib/datadog/di/probe_notifier_worker.rb
```diff
@@ -0,0 +1,242 @@
...
+        @io_in_progress = false
+        @sleep_remaining = nil
+        @wake_scheduled = false
+      end
+
+      attr_reader :settings
+
+      def start
+        return if defined?(@thread) && @thread
+        @thread = Thread.new do
```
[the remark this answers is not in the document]
ivoanjo: Actually, consider setting `@thread` to `nil` in the constructor -- this ensures that all instances of the class share the same shape. (and thus are easier for the VM to optimize)

lib/datadog/di/probe_notifier_worker.rb
ivoanjo, line 94: :eyes: This will fail if `thread` is `nil`:

```
[3] pry(main)> thread = nil
=> nil
[4] pry(main)> unless thread&.join(123)
[4] pry(main)*   thread.kill
[4] pry(main)* end  
NoMethodError: undefined method `kill' for nil:NilClass
from (pry):7:in `__pry__
```

lib/datadog/di/probe_notifier_worker.rb
ivoanjo, line 124: It's possible to avoid the sleeping by using a condition variable to flag when the queue is empty

spec/datadog/di/probe_notifier_worker_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+
+          sleep 0.4
+          # Still within the cooldown period
+          expect(worker.send(:snapshot_queue)).to eq([snapshot, snapshot])
+
+          expect(transport).to receive(:send_snapshot).once.with([snapshot, snapshot])
+
+          sleep 0.5
+          expect(worker.send(:snapshot_queue)).to eq([])
+        end
```
ivoanjo, line 112: If possible, avoid using `sleep`s in tests -- they make the test suite both slower and flakier >_>

lib/datadog/di/probe_notifier_worker.rb
```diff
@@ -0,0 +1,242 @@
...
+        @settings = settings
+        @status_queue = []
+        @snapshot_queue = []
+        @transport = transport
+        @lock = Mutex.new
+        @wake = Core::Semaphore.new
+        @io_in_progress = false
+        @sleep_remaining = nil
+        @wake_scheduled = false
+      end
```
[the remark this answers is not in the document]
p-datadog: They are now consumed by the transport, I removed agent settings from probe notifier worker.

lib/datadog/di/probe_notifier_worker.rb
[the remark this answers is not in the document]
p-datadog: I set it to nil in constructor.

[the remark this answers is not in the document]
ivoanjo: What, did I miss it? (I don't see it -- and wouldn't that make the `defined?(@thread)` not needed?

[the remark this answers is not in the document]
ivoanjo: Ah I think I saw your comment before 0b08177d5c1a0d10e580de2205fdcfe147fd2968 was pushed ;D

(Concurrency joke goes here)

lib/datadog/di/probe_notifier_worker.rb
p-datadog: Thank you, repaired.

lib/datadog/di/probe_notifier_worker.rb
p-datadog: I added a note to investigate this.

spec/datadog/di/probe_notifier_worker_spec.rb
ivoanjo: Note that slow specs kinda affect all CI runs so yeah, try not to add too much to that fire >_>

Strech (approved):
I put a few cents here, but I think it's ok

lib/datadog/di/probe_notifier_worker.rb
```diff
@@ -0,0 +1,244 @@
...
+        # This method should ideally only be called when there are actually
+        # events to send, but it can be called when there is nothing to do.
+        # Currently we only have one wake-up signaling object and two
+        # types of events. Therefore on most wake-ups we expect to only
+        # send one type of events.
+        define_method("maybe_send_#{event_type}") do
+          batch = nil
+          @lock.synchronize do
+            batch = instance_variable_get("@#{event_type}_queue")
+            instance_variable_set("@#{event_type}_queue", [])
```
ivoanjo, line 210: Minor: It may be worth adding an `attr_accessor` and then use `send(...)` to access these variables.

p-datadog: The reading side is done with an attribute, are you concerned that there maybe a spelling mistake here and the wrong variable will be written to?

ivoanjo: I was mostly thinking that `instance_variable_set` and `instance_variable_get` are very sharp weapons so having the attr_accessor seems a bit easier to avoid bugs.

On the other hand, it is true that we can misspell the creation of the attr_accessor as well, so it's not like there's no potential for bugs there either.

p-datadog: Let me leave it this way for now but if the matter comes up again I can redo as attr_accessor.

PR #4031, opened 2024-10-25
[NO-TICKET] Fix crashtracking sending parent pid/runtime-id when child crashes

ivoanjo:
**What does this PR do?**

This PR fixes a bug in the crashtracking component: because it cached all of the tags from creation, if a process forked, it would use the tags from the parent, which included a (now-incorrect) pid and runtime-id.

**Motivation:**

This makes sure we have accurate metadata in crashtracking reports.

**Change log entry**

(Internal change, not relevant for changelog)

**Additional Notes:**

N/A

**How to test the change?**

I've added test coverage for this change. It can also be observed by using

> `$ DD_CRASHTRACKING_ENABLED=true bundle exec ruby -e 'require "datadog/auto_instrument"; puts "Parent: #{Process.pid} #{Datadog::Core::Environment::Identity.id}"; fork { puts "Child: #{Process.pid} #{Datadog::Core::Environment::Identity.id}"; Process.kill("SEGV", Process.pid) }'`

and validating that the correct pid/runtime-id from the child is sent in the metadata.

### DIFF
--- a/spec/datadog/core/crashtracking/component_spec.rb
+++ b/spec/datadog/core/crashtracking/component_spec.rb
@@ -125,6 +125,8 @@ RSpec.describe Datadog::Core::Crashtracking::Component, skip: !CrashtrackingHelp
     end
 
     describe '#update_on_fork' do
+      before { allow(logger).to receive(:debug) }
+
       context 'when _native_stop_crashtracker raises an exception' do
         it 'logs the exception' do
           crashtracker = build_crashtracker(logger: logger)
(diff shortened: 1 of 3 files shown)


spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -138,12 +138,24 @@
...
 
-        crashtracker = build_crashtracker
+        crashtracker = build_crashtracker(logger: logger)
+
+        crashtracker.update_on_fork
+      end
+
+      it 'refreshes the latest settings' do
+        expect(Datadog).to receive(:configuration).and_return(:latest_settings)
+        expect(Datadog::Core::Crashtracking::TagBuilder).to receive(:call).with(:latest_settings).and_return([:latest_tags])
```
Strech, line 153: I think it's very minor thing, but technically this is not an expectation, but rather allowance. Not critical, but I would split expectations from the setup, like this

```suggestion
        allow(Datadog).to receive(:configuration).and_return(:latest_settings)
        allow(Datadog::Core::Crashtracking::TagBuilder).to receive(:call).with(:latest_settings).and_return([:latest_tags])

```

ivoanjo: Yeah -- I'll admit I'm sometimes slightly inconsistent about remembering to do this, although I agree with your reasoning. Applied in 06dfc27b6001ea1d621c4e87331de29297ff087f

PR #4033, opened 2024-10-27
Fix resource name when Grape paths contain a leading slash

jethrodaniel:
**What does this PR do?**

Fix an issue where a [Grape](https://github.com/ruby-grape/grape) resource like this

```ruby
class TestingAPI < Grape::API
  get '/path/with/leading/slash' do
    'leading-slash'
  end
end
```

results in a resource `TestingAPI GET //path/with/leading/slash`, when it _should_ result in `TestingAPI GET /path/with/leading/slash`.

**Motivation:**

Seeing the resource name have an unnecessary leading slash.

**Change log entry**

Fix resource name when Grape paths contain a leading slash

**Additional Notes:**

**How to test the change?**

```
rake test:grape
```


lib/datadog/tracing/contrib/grape/endpoint.rb
```diff
@@ -268,7 +268,7 @@ def endpoint_expand_path(endpoint)
               namespace = endpoint.routes.first && endpoint.routes.first.namespace || ''
 
               parts = (namespace.split('/') + route_path).reject { |p| p.blank? || p.eql?('/') }
-              parts.join('/').prepend('/')
+              parts.join('/').delete_prefix('/').prepend('/')
```
Strech, line 271: Hey @jethrodaniel thanks for your contribution 👍🏼 

I think given that we already have a parts with a first part to be prefixed, adding and removing `/` looks excessive. WDYT if we just check before prepending if the first char is a `/`?

Something like:

```ruby
path = (namespace.split('/') + route_path).reject { |p| p.blank? || p.eql?('/') }.join('/')
path.prepend('/') if path[0] != '/'
```

jethrodaniel: +1 for the more efficient approach - will update when I get time. Thanks!

Strech (approved):
LGTM 👍🏼 

Thanks for your time 💟

p-datadog (approved):

p-datadog:
We need to verify that `http.route` is the same as `resource_name` with respect to the leading slashes.

Strech:
Hey @jethrodaniel thanks for your contribution. This change was released as a part of [v2.5.0](https://github.com/DataDog/dd-trace-rb/releases/tag/v2.5.0) release. 🎉

PR #4035, opened 2024-10-28
[APPSEC-55378] Rework Processor::Context logging and telemetry

Strech:
**What does this PR do?**

This PR centralizes **debug** handling of the WAF execution into a single place `Processor::Context`

**Motivation:**

During upgrades we would like to know when we've made a mistake and be alerted about it upfront. We double-checked existing result codes of the WAF execution and removed outdated. At the same time if we use Telemetry error reporting to deliver unhandled errors such as:

* `DDWAF_ERR_INTERNAL` (ruby: `:err_internal`)
* `DDWAF_ERR_INVALID_OBJECT` (ruby: `:err_invalid_object`)
* `DDWAF_ERR_INVALID_ARGUMENT` (ruby: `:err_invalid_argument`)

See [full list](https://github.com/DataDog/libddwaf/blob/10e3a1dfc7bc9bb8ab11a09a9f8b6b339eaf3271/include/ddwaf.h#L64-L68)

**Change log entry**

Not needed.

**Additional Notes:**

This PR could be released without `libddwaf-rb` upgrade

**How to test the change?**

CI is enough.


lib/datadog/appsec/processor/context.rb
```diff
@@ -24,24 +26,22 @@ def run(input, timeout = WAF::LibDDWAF::DDWAF_RUN_TIMEOUT)
           start_ns = Core::Utils::Time.get_time(:nanosecond)
 
           input.reject! do |_, v|
-            case v
-            when TrueClass, FalseClass
-              false
-            else
-              v.nil? ? true : v.empty?
-            end
+            next false if v.is_a?(TrueClass) || v.is_a?(FalseClass)
```
y9v, line 29: I think we actually don't need this line, and this whole `reject!` could be changed to:
```ruby
input.compact!.reject(&:empty?)
```

y9v: and it's also fine to leave this block like it is, to reduce amount of changes in this PR

y9v (approved):

lib/datadog/appsec/processor/context.rb
```diff
@@ -66,6 +67,19 @@ def finalize
 
         private
 
+        def report_execution(result)
+          Datadog.logger.debug { "libddwaf execution timed out: #{result.inspect}" } if result.timeout
+
+          if LIBDDWAF_SUCCESSFUL_EXECUTION_CODES.include?(result.status)
+            Datadog.logger.debug { "libddwaf execution result: #{result.inspect}" }
+          else
+            message = "libddwaf execution error: #{result.status.inspect} with #{result.inspect}"
```
Strech, line 76: After discussion with @TonyCTHsu we agreed to change it into the following:

```
message = "libddwaf:1.14.0 execution error: #{result.status.inspect}"
```

we will drop `result.inspect` to reduce cardinality and also it doesn't bring much value.

y9v (approved):

PR #4039, opened 2024-10-28
Add Ruby 3.4.0-preview2 support for github workflow tests

sarahchen6:
**What does this PR do?**
This PR adds Ruby 3.4.0-preview2 support for github workflow tests, specifically macos and yjit, by adding a "3.4" version to the `test-yjit.yaml` and `test-macos.yml` files under `.github`. It also updates the Dockerfile-3.4.0 base image to `ruby:3.4.0-preview2` and adjusts a few of the spec tests, given new Ruby output changes (see [here](https://www.ruby-lang.org/en/news/2024/10/07/ruby-3-4-0-preview2-released/)).

**Motivation:**
Though the official Ruby 3.4 is not yet released, we can temporarily add Ruby 3.4.0-preview2. This allows us to start testing with Ruby 3.4 in our CI and spot -> address issues early. After Ruby 3.4 is officially released post-Dec 25, 2024, we can replace the preview with it.

**Change log entry**
No change log entry for now. Thinking to have a "Ruby 3.4.0-preview2 support added" update after it is fully supported.

**Additional Notes:**
More Ruby 3.4.0-preview2 support is needed. This PR is one step toward full support and only addresses github workflow tests. It acts alongside #4038 .

**How to test the change?**
This change can be tested in CI.

### DIFF
--- a/.github/workflows/test-macos.yaml
+++ b/.github/workflows/test-macos.yaml
@@ -21,6 +21,7 @@ jobs:
           - '3.1'
           - '3.2'
           - '3.3'
+          - '3.4'
           # ADD NEW RUBIES HERE
     name: Test (${{ matrix.os }}, ${{ matrix.ruby }})
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/test-macos.yaml
+++ b/.github/workflows/test-macos.yaml
@@ -41,11 +41,9 @@ jobs:
           ruby-version: ${{ matrix.ruby }}
           rubygems: 3.3.26
           bundler: 2.3.26 # needed to fix issue with steep on Ruby 3.0/3.1
-      - if: ${{ matrix.ruby == '3.3' }}
+      # Specify gem version for 3.4 because default version (3.6.0.dev)
+      # leads to an incorrect gem root path
+      - if: ${{ matrix.ruby == '3.3' || matrix.ruby == '3.4' }}
         run: gem update --system 3.5.21
-      - run: |
-          ruby -v
-          gem -v
-          bundler -v
       - run: bundle install
       - run: bundle exec rake spec:main
--- a/spec/datadog/core/environment/execution_spec.rb
+++ b/spec/datadog/core/environment/execution_spec.rb
@@ -191,7 +191,8 @@ RSpec.describe Datadog::Core::Environment::Execution do
                 Open3.capture3('ruby', stdin_data: script)
               end
 
-              skip('DEBUG')
+              # Ruby 3.4 outputs an exception instead of the information to be asserted because of the forked process.
+              pending('Pending for Ruby 3.4.') if RUBY_VERSION.start_with?('3.4.')
               expect(err).to include('ACTUAL:true')
             end
           end
--- a/spec/datadog/core/environment/execution_spec.rb
+++ b/spec/datadog/core/environment/execution_spec.rb
@@ -191,6 +191,7 @@ RSpec.describe Datadog::Core::Environment::Execution do
                 Open3.capture3('ruby', stdin_data: script)
               end
 
+              skip('DEBUG')
               expect(err).to include('ACTUAL:true')
             end
           end
--- a/spec/datadog/core/error_spec.rb
+++ b/spec/datadog/core/error_spec.rb
@@ -90,11 +90,11 @@ RSpec.describe Datadog::Core::Error do
 
           # Outer-most error first, inner-most last
           wrapper_error_message = /in.*wrapper': wrapper layer \(RuntimeError\)/
-          wrapper_caller = /from.*in `call'/
+          wrapper_caller = /from.*in ['`]call'/
           middle_error_message = /in.*middle': middle cause \(RuntimeError\)/
-          middle_caller = /from.*in `wrapper'/
-          root_error_message = /in `root': root cause \(RuntimeError\)/
-          root_caller = /from.*in `middle'/
+          middle_caller = /from.*in ['`]wrapper'/
+          root_error_message = /in ['`]root': root cause \(RuntimeError\)/
+          root_caller = /from.*in ['`]middle'/
 
           expect(error.backtrace)
             .to match(
--- a/spec/datadog/core/error_spec.rb
+++ b/spec/datadog/core/error_spec.rb
@@ -89,6 +89,8 @@ RSpec.describe Datadog::Core::Error do
           expect(error.message).to eq('wrapper layer')
 
           # Outer-most error first, inner-most last
+          # Ruby 3.4 adjusts the format of error messages and Hash#inspect renderings
+          # https://www.ruby-lang.org/en/news/2024/10/07/ruby-3-4-0-preview2-released/
           wrapper_error_message = /in.*wrapper': wrapper layer \(RuntimeError\)/
           wrapper_caller = /from.*in ['`]call'/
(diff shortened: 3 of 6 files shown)


.github/workflows/test-macos.yaml
```diff
@@ -41,4 +42,6 @@ jobs:
           rubygems: 3.3.26
           bundler: 2.3.26 # needed to fix issue with steep on Ruby 3.0/3.1
       - run: bundle install
+      - if: ${{ matrix.ruby == '3.4' }}
+        run: gem update --system 3.5.21
```
sarahchen6, line 46: The latest version of `rubygem` is 3.5.22. However, version 3.5.22 changes macos files ([see here](https://github.com/rubygems/rubygems/blob/master/CHANGELOG.md#3522--2024-10-16)) and causes the rspec tests to fail. Specifically, `/spec/datadog/tracing/contrib_spec.rb:58` is unable to find the correct file paths because the root gem path has changed.

I'm unsure that hard-coding rubygem v3.5.21 is the best solution, so very open to any suggestions!

Gemfile
```diff
@@ -36,7 +36,12 @@ gem 'rspec_junit_formatter', '>= 0.5.1'
 # Merging branch coverage results does not work for old, unsupported rubies and JRuby
 # We have a fix up for review, https://github.com/simplecov-ruby/simplecov/pull/972,
 # but given it only affects unsupported version of Ruby, it might not get merged.
-gem 'simplecov', git: 'https://github.com/DataDog/simplecov', ref: '3bb6b7ee58bf4b1954ca205f50dd44d6f41c57db'
+if RUBY_VERSION.start_with?('3.4.')
```
sarahchen6, line 39: Seems like `https://github.com/DataDog/simplecov` has been archived, and Ruby 3.4 requires a newer existing version of `simplecov`. Should more versions than just 3.4 be switched to `gem 'simplecov'`?

Gemfile
```diff
@@ -96,3 +101,8 @@ end
 #
 # TODO: Remove this once the issue is resolved: https://github.com/ffi/ffi/issues/1107
 gem 'ffi', '~> 1.16.3', require: false
+
+# Ruby 3.4 should be supported by strscan v3.1.1. However, the latest release of strscan is v3.1.0.
+#
+# TODO: Remove this v3.1.0 requirement once v3.1.1 is released.
+gem 'strscan', '3.1.0' if RUBY_VERSION.start_with?('3.4.')
```
sarahchen6, line 108: Ruby 3.4 automatically uses v3.1.1 of `gem 'strscan'` ([see here](https://stdgems.org/strscan/)), but v3.1.0 is the latest release of `strscan` ([see here](https://rubygems.org/gems/strscan)). After the release of v3.1.1, I assume this version requirement can be removed.

spec/datadog/core/environment/execution_spec.rb
```diff
@@ -191,6 +191,7 @@ def test_it_does_something_useful
                 Open3.capture3('ruby', stdin_data: script)
               end
 
+              skip('DEBUG')
```
sarahchen6, line 194: After talking with Tony, it seems that Ruby 3.4 outputs an exception here instead of the information to be asserted because of the forked process. I've skipped this test for now, but open to any other fixes!

spec/datadog/core/error_spec.rb
```diff
@@ -90,11 +90,11 @@ def call
 
           # Outer-most error first, inner-most last
           wrapper_error_message = /in.*wrapper': wrapper layer \(RuntimeError\)/
-          wrapper_caller = /from.*in `call'/
+          wrapper_caller = /from.*in ['`]call'/
```
sarahchen6, line 93: Ruby 3.4.0-preview2 changes the format of error messages ([see here](https://www.ruby-lang.org/en/news/2024/10/07/ruby-3-4-0-preview2-released/)), so I adjusted the syntax of these rspec tests to consider the new formatting.

spec/datadog/core/environment/execution_spec.rb
TonyCTHsu: We can mark this test `pending` conditionally with ruby version.

spec/datadog/core/error_spec.rb
TonyCTHsu: Lets write a code comment about this.

Gemfile
TonyCTHsu: `simplecov` is used in CircleCI to collect coverage data in CircleCI. This fork contains the code we need to report all the ruby versions. Meaning this change need to be validated in CircleCI changes in: https://github.com/DataDog/dd-trace-rb/pull/4040

The source code we have been using has not changed for a long time. Perhaps its a good idea to follow the path of `datadog-ruby_core_source` to release its own gem. This can also avoid smashing Github very hard on this.

Gemfile
vpellan: Maybe instead of forcing the usage of 3.1.0, we could use the git repository, so if Ruby 3.4 is using any new feature or bugfix that has been added since the release of 3.1.0, it should still work

 ```suggestion
gem 'strscan', git: 'https://github.com/ruby/strscan' if RUBY_VERSION.start_with?('3.4.')
```

File: Gemfile (hunk 2 above)
Comment (ivoanjo) [reply Gemfile:39]: Looking at https://github.com/simplecov-ruby/simplecov/pull/972/files, our change is a one-liner. I think we could try applying this change as a monkey-patch when we load rspec perhaps?

(I'd advise against getting too fork-happy, because the more things we fork, the more things we need to maintain, and we're already struggling to maintain one gem >_>)

File: Gemfile (hunk 3 above)
Comment (sarahchen6) [reply Gemfile:108]: Good idea! Will do this

File: spec/datadog/core/environment/execution_spec.rb (hunk 4 above)
Comment (sarahchen6) [reply spec/datadog/core/environment/execution_spec.rb:194]: Sounds good!

File: Gemfile (hunk 2 above)
Comment (lloeki) [reply Gemfile:39]: Monkeypatching (actually, vendoring that file, modifying it to our end) seems like the way to go here, plus it would remove that pesky git repo reference.

Comment (sarahchen6) [reply Gemfile]: After meeting with Oleg, I realized the existing `simplecov` specification works. No change is needed. Thanks for your help!

File: .github/workflows/test-macos.yaml (hunk 1 above)
Comment (p-datadog) [reply .github/workflows/test-macos.yaml:46]: I would suggest adding a note to this effect to the source for future readers.

**Review by vpellan — APPROVED** [review]

File: .github/workflows/test-macos.yaml [hunk 6]
```diff
@@ -41,4 +42,6 @@ jobs:
           rubygems: 3.3.26
           bundler: 2.3.26 # needed to fix issue with steep on Ruby 3.0/3.1
       - run: bundle install
+      - if: ${{ matrix.ruby == '3.4' }}
+        run: gem update --system 3.5.21 # the latest gem causes 3.4 macos failues
```
marcotc, line 46: Can you link to the issue that made us lock this (maybe it's in [this page](https://github.com/rubygems/rubygems/issues?q=is%3Aissue+mac+is%3Aclosed)), so we know when to unlock it?

marcotc (approved):

p-datadog (approved):

TonyCTHsu (approved):

PR #4041, opened 2024-10-28
[APPSEC-55194] Handle low-level libddwaf exceptions

Strech:
**What does this PR do?**

Wrap `Processor::Context#run` with an error handling and telemetry reporting.

**Motivation:**

Is some specific situations, our `libddwaf-rb` is going to throw an exception when it fails to convert C-object into Ruby-object and some other low-level occasions.

**Change log entry**

Not needed.

**Additional Notes:**

Likely we would change the interface of libddwaf-rb soon™️, but in foreseen future we need to be able to land on feet when something wrong happen with our FFI binding.

**How to test the change?**

CI is enough.

### DIFF
--- a/lib/datadog/appsec/processor/context.rb
+++ b/lib/datadog/appsec/processor/context.rb
@@ -21,7 +21,7 @@ module Datadog
           @events = []
           @run_mutex = Mutex.new
 
-          @libddwaf_tag = "libddwaf:#{WAF::VERSION::STRING}"
+          @libddwaf_debug_tag = "libddwaf:#{WAF::VERSION::STRING}"
         end
 
         def run(input, timeout = WAF::LibDDWAF::DDWAF_RUN_TIMEOUT)
@@ -35,7 +35,7 @@ module Datadog
             v.nil? ? true : v.empty?
           end
 
-          _code, result = @context.run(input, timeout)
+          _code, result = try_run(input, timeout)
 
           stop_ns = Core::Utils::Time.get_time(:nanosecond)
 
@@ -59,7 +59,7 @@ module Datadog
             }
           }
 
-          _code, result = @context.run(input, WAF::LibDDWAF::DDWAF_RUN_TIMEOUT)
+          _code, result = try_run(input, WAF::LibDDWAF::DDWAF_RUN_TIMEOUT)
 
           report_execution(result)
           result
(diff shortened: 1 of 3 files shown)


lib/datadog/appsec/processor/context.rb
```diff
@@ -22,6 +22,7 @@ def initialize(handle, telemetry:)
           @run_mutex = Mutex.new
 
           @libddwaf_tag = "libddwaf:#{WAF::VERSION::STRING}"
+          @libddwaf_fallback_result = [:ok, WAF::Result.new(:ok, [], 0.0, false, [], [])]
```
Strech, line 25: Unfortunately, we can't define it as a constant due to lazy-load of the component and `libddwaf` require statement in `Processor`.

lib/datadog/appsec/processor/context.rb
```diff
@@ -71,6 +72,15 @@ def finalize
 
         private
 
+        def try_run(input, timeout)
+          @context.run(input, timeout)
+        rescue WAF::LibDDWAF::Error => e
+          Datadog.logger.debug { "#{@libddwaf_tag} execution error: #{e} backtrace: #{e.backtrace&.first(3)}" }
```
Strech, line 78: According to Steep, backtrace could be potentially `nil`, hence safe-navigation

lib/datadog/appsec/processor/context.rb
Strech: I'm going to rewrite it 😞

spec/datadog/appsec/processor/context_spec.rb
```diff
@@ -255,6 +255,23 @@
...
+      before do
+        allow(context.instance_variable_get(:@context)).to receive(:run).with(input, timeout)
+          .and_raise(Datadog::AppSec::WAF::LibDDWAF::Error, 'Could not convert persistent data')
+      end
+
+      let(:result) { context.run(input, timeout) }
+
+      it 'sends telemetry report' do
+        expect(telemetry).to receive(:error).with(/libddwaf:[\d.]+ execution error: :err_internal/)
+        expect(telemetry).to receive(:report)
```
Strech, line 269: This suppose to be very unfortunate case, we will start with this approach and refactor it next (to report once?)

PR #4060, opened 2024-11-01
Revert Datadog testagent version from 1.18.0 to 1.12.0

sarahchen6:
**What does this PR do?**
This PR changes the Datadog testagent version from 1.18.0 back to 1.12.0 in the `docker-compose.yml` and `.circleci/config.yml`.

**Motivation:**
Testagent v1.18.0 causes local tracing parsing errors, such as: `ERROR -- datadog: [datadog] (/app/lib/datadog/tracing/transport/http/client.rb:42:in `rescue in send_request') Internal error during Datadog::Tracing::Transport::HTTP::Client request. Cause: JSON::ParserError unexpected token at '{"rate_by_service": {"service:,e' Location: /usr/local/bundle/gems/json-2.7.5/lib/json/common.rb:220:in `parse'`. No particular functionality of v1.18.0 was needed, and reverting the version to 1.12.0 fixes these errors.

**Change log entry**
No change log entry needed.

**Additional Notes:**

**How to test the change?**
To test this change, run `docker compose run --rm -it tracer-<version number> bash -c "rm *lock && bundle install && bundle exec rake spec:main"` locally. All tests should pass, whereas previously, you would see many tracing errors.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -107,7 +107,7 @@ test_containers:
     image: memcached:1.5-alpine
   - &memcached_port 11211
   - &container_testagent
-    image: ghcr.io/datadog/dd-apm-test-agent/ddapm-test-agent:v1.18.0
+    image: ghcr.io/datadog/dd-apm-test-agent/ddapm-test-agent:v1.12.0
     name: testagent
     environment:
       - LOG_LEVEL=DEBUG
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -203,7 +203,7 @@ services:
       - ddagent_var_run:/var/run/datadog
 
   testagent:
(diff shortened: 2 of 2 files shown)


Strech (approved):
All right!

ivoanjo:
I'm guessing this is the "stop the bleeding PR" which makes sense; but we should also make sure to ping the folks that maintain the test agent to let them know that 1.18.0 appears broken, as permanently staying on an old version doesn't seem great.

p-datadog (approved):

sarahchen6:
> I'm guessing this is the "stop the bleeding PR" which makes sense; but we should also make sure to ping the folks that maintain the test agent to let them know that 1.18.0 appears broken, as permanently staying on an old version doesn't seem great.

I agree. Will do!

PR #4062, opened 2024-11-04
[NO-TICKET] Fix rubocop lint

ivoanjo:
**What does this PR do?**

This PR moves the `start_worker` and `stop_worker` to the `private` section of `writer.rb`.

It does this instead of using
`private :start_worker, :stop_worker` to make rubocop happy.

**Motivation:**

Unbreak rubocop in master.

**Change log entry**
(Not relevant)

**Additional Notes:**

I thought we had pinned down the version of rubocop in CI, but it appears we have a gap. I've not investigated further.

**How to test the change?**

Validate that the lint check in CI is back to green.

### DIFF
--- a/lib/datadog/tracing/writer.rb
+++ b/lib/datadog/tracing/writer.rb
@@ -68,21 +68,6 @@ module Datadog
         end
       end
 
-      # spawns a worker for spans; they share the same transport which is thread-safe
-      # @!visibility private
-      def start_worker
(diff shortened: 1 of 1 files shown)


Strech (approved):
👏🏼

bouwkast (approved):
LGTM (but am not Ruby dev 😄 )

TonyCTHsu (approved):

zarirhamza (approved):

PR #4063, opened 2024-11-04
DEBUG-2334 Remaining dynamic instrumentation code

p-datadog:
**What does this PR do?**

Adds remaining dynamic instrumentation code, tests and exposes the feature to users.

- Adds serializers for Time, Date, DateTime (used by ActiveRecord timestamps)
- Adds a mechanism to register custom serializers (used by the ActiveRecord::Base serializer included)
- Adds a custom serializer for ActiveRecord::Base objects, currently this just exports the attributes (thus *not* exporting all of the other fields of AR objects that usually are not interesting) but could also export new_record? flag and possibly some other fields
- Internal DI settings moved to under `internal` namespace
- Internal setting added to enable DI in development environments, this is needed for the test suite to work and I am using this for internal troubleshooting/testing
- Minimum send interval for the network submission worker moved to configuration setting from a constant
- Proper logging implemented
- Telemetry is injected and unhandled exceptions are reported to it
- Line probes now require a single match of the provided path suffix, if multiple paths match an exception is raised (this will be reported via UI to the user, on the assumption that user wanted a specific file instrumented and we don't know which of the multiple matching files is the correct one)
- When probe installation fails for reasons other than the referenced target being missing (presumably not yet loaded), the failure is noted and that probe is not attempted again in the future
- **DI code tracking is enabled by default when DI loads**. This is needed to instrument third-party and customer code via line probes.
- **DI is loaded by default** 
- DI component added
- Remote client updated to invoke DI to process DI configs
- Probe manager component added - this stores the probes received from remote config, tracks their status (pending/installed), and drives the rest of DI to instrument code, send notifications/events, etc.
- Code tracker invokes DI when a file is loaded to install any pending line probes in that file
- Notifications/events now use actual path of file where the code was invoked, previously this was hacked via basename comparisons

**Motivation:**
Initial implementation of DI for Ruby.

**Change log entry**
Dynamic instrumentation is now available in Ruby. Currently only log probes are implemented, but they can be set on both methods and lines.

**Additional Notes:**

Apologies for the diff size. `lib` diff is only about 1/3 of the total size, the rest is in `spec`.

Still remaining:

- Go through all DI code and check all outstanding TODO comments for what needs to be addressed prior to initial release
- Various additional tests, I plan to write them in parallel with working on system tests, documentation and reliability environment

**How to test the change?**
This PR includes fairly extensive integration tests for DI, however true end-to-end testing is to be done via system tests which is my current task.


lib/datadog/core/configuration/components.rb
```diff
@@ -126,6 +129,11 @@ def startup!(settings)
             end
           end
 
+          if settings.remote.enabled && old_state&.[](:remote)
+            # remote should be defined here
+            remote.start
+          end
```
marcotc, line 135: Why is this change needed?

lib/datadog/di.rb
```diff
@@ -1,21 +1,45 @@
...
+if defined?(ActiveRecord::Base)
+  # The third-party library integrations need to be loaded after the
+  # third-party libraries are loaded. Tracing and appsec use Railtie
+  # to delay integrations until all of the application's dependencies
+  # are loaded, when running under Rails. We should do the same here in
+  # principle, however DI currently only has an ActiveRecord integration
+  # and AR should be loaded before any application code is loaded, being
+  # part of Rails, therefore for now we should be OK to just require the
+  # AR integration from here.
+  require_relative 'di/contrib/active_record'
```
marcotc, line 29: Does this work in a regular Rails app (meaning, `ActiveRecord::Base` is loaded when `dd-trace-rb` is loaded)? If so, we add to the comment that "it works in practice blah blah".

lib/datadog/di/code_tracker.rb
```diff
@@ -44,37 +44,54 @@ def start
...
+              # without checking that it is not nil, so here, maybe there is
+              # some situation where path would in fact be nil and
+              # steep would end up saving the day.
+              if path && !tp.eval_script
+                registry_lock.synchronize do
+                  registry[path] = tp.instruction_sequence
+                end
+              end
+
+              # TODO this line is not being exercised
```
marcotc, line 82: This comment doesn't help me understand why the line is not being executed, but if it's clear to you, it's ok!

lib/datadog/di/code_tracker.rb
```diff
@@ -44,37 +44,54 @@ def start
...
+              DI.component&.probe_manager&.install_pending_line_probes(path)
+            rescue => exc
+              if component = DI.component
```
marcotc, line 85: Can you document when this error path happens? It looks like the log lines below give a hint, but it would be nice to have a description here, even if short.

lib/datadog/di/contrib/active_record.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Datadog::DI::Serializer.register(condition: lambda { |value| ActiveRecord::Base === value }) \
+do |serializer, value, name:, depth:|
+  value_to_serialize = {
+    attributes: value.attributes,
+  }
+  serializer.serialize_value(value_to_serialize, depth: depth ? depth - 1 : nil, type: value.class)
```
marcotc, line 8: Why the `- 1` in `depth - 1`?

lib/datadog/di/contrib/active_record.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+Datadog::DI::Serializer.register(condition: lambda { |value| ActiveRecord::Base === value }) \
```
marcotc, line 3: Maybe a stupid question, but why not make this condition 
```suggestion
Datadog::DI::Serializer.register(condition: lambda { |value| defined?(ActiveRecord::Base) && ActiveRecord::Base === value }) \
```
so that it can always be loaded, even if ActiveRecord is not loaded?

lib/datadog/di/probe_manager.rb
```diff
@@ -0,0 +1,182 @@
+# frozen_string_literal: true
+
+module Datadog
+  module DI
+    # Stores probes received from remote config (that we can parse, in other
+    # wordrs, whose type/attributes we support), requests needed instrumentation
```
marcotc, line 6:
```suggestion
    # words, whose type/attributes we support), requests needed instrumentation
```

lib/datadog/di/probe_manager.rb
```diff
@@ -0,0 +1,182 @@
...
+      attr_reader :pending_probes
+
+      # Probes that failed to instrument for reasons other than the target is
+      # not yet loaded are added to this collection, so that we do not try
+      # to instrument them every time remote configuration is processed.
+      attr_reader :failed_probes
+
+      # config is one probe info
+      def add_probe(probe)
+        # TODO lock here?
```
marcotc, line 70: If I understand correctly, adding a probe is normally done by a "slow" process (user action, application initialization, remote configuration) **and** if adding two conflicting probes concurrently can cause big issues, I think it's reasonable to add a lock here.

lib/datadog/di/probe_manager.rb
```diff
@@ -0,0 +1,182 @@
...
+      attr_reader :probe_notifier_worker
+      attr_reader :installed_probes
+      attr_reader :pending_probes
+
+      # Probes that failed to instrument for reasons other than the target is
+      # not yet loaded are added to this collection, so that we do not try
+      # to instrument them every time remote configuration is processed.
+      attr_reader :failed_probes
+
+      # config is one probe info
```
marcotc, line 68: I think this comment is probably not accurate anymore to describe this function.

lib/datadog/di/probe_manager.rb
```diff
@@ -0,0 +1,182 @@
...
+        # install it again.
+
+        # TODO add top stack frame to message
+        failed_probes[probe.id] = "#{exc.class}: #{exc}"
+
+        raise
+      end
+
+      def remove_other_probes(probe_ids)
+        pending_probes.values.each do |probe|
```
marcotc, line 110:
```suggestion
        pending_probes.each_value do |probe|
```

lib/datadog/di/probe_manager.rb
```diff
@@ -0,0 +1,182 @@
...
+              # successfully de-instrumented. Active probes do incur overhead
+              # for the running application, and if the error is ephemeral
+              # we want to try removing the probe again at the next opportunity.
+              #
+              # TODO give up after some time?
+              installed_probes.delete(probe.id)
+            rescue => exc
+              raise if settings.dynamic_instrumentation.internal.propagate_all_exceptions
+              # Silence all exceptions?
+              # TODO should we propagate here and rescue upstream?
```
marcotc, line 129: Trying to answer this comment: in my read of this PR, it makes sense to try to `unhook` all probes, even if one of them fail to unhook. Which would mean that propagating the error early will prevent other probes from being propagated. So it looks to me that we should not propagate the error immediately here.

lib/datadog/di/probe_manager.rb
```diff
@@ -0,0 +1,182 @@
...
+
+      private def install_pending_method_probes(cls)
+        # TODO search more efficiently than linearly
+        pending_probes.each do |probe_id, probe|
+          if probe.method?
+            # TODO move this stringification elsewhere
+            if probe.type_name == cls.name
+              # TODO is it OK to hook from trace point handler?
+              # TODO the class is now defined, but can hooking still fail?
+              # TODO pass rate_limiter here, need to get it from somewhere
```
marcotc, line 145: Isn't the rate limiter available here, in `probe.rate_limiter`?

lib/datadog/di/probe_manager.rb
```diff
@@ -0,0 +1,182 @@
...
+        end
+
+        payload = probe_notification_builder.build_executed(probe, **opts)
+        probe_notifier_worker.add_snapshot(payload)
+      end
+
+      # Class/module definition trace point (:end type).
+      # Used to install hooks when the target classes/modules aren't yet
+      # defined when the hook request is received.
+      attr_reader :definition_trace_point
```
marcotc, line 179: Any reason this `attr_reader` is declared at the end of the class, instead of together with the other `attr_reader`s?

lib/datadog/core/configuration/components.rb
p-datadog: I use it for testing to avoid having to hit the app before it instruments code but it does not belong in this PR, I will revert it.

lib/datadog/di/configuration/settings.rb
```diff
@@ -154,6 +110,72 @@ def self.add_settings!(base)
                 o.type :int
                 o.default 20
               end
+
+              settings :internal do
```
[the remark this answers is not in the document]
p-datadog: Will do. The group is internal because it is not for use by customers - the options are meant for (my) development and testing only.

lib/datadog/di/contrib/active_record.rb
p-datadog: We went from the object to its attributes, this is one level in. I will double-check against other serialization code.

p-datadog: We went from the object to its attributes, this is one level in. I will double-check against other serialization code.

lib/datadog/di/contrib/active_record.rb
p-datadog: It is a genius idea actually. The reason not to do it is for performance when AR is not loaded (since otherwise every object will have to evaluate this condition) but perhaps it's worthwhile to take the performance hit to permit loading Rails after datadog.

lib/datadog/di/probe_manager.rb
p-datadog: Yes, deleted the comment.

lib/datadog/di/configuration/settings.rb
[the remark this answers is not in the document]
p-datadog: Done in #4089.

lib/datadog/di/code_tracker.rb
p-datadog: Done in #4089

lib/datadog/di/configuration/settings.rb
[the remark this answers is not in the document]
p-datadog: Done in #4089.

lib/datadog/di/probe_manager.rb
p-datadog: Added in https://github.com/DataDog/dd-trace-rb/pull/4097.

lib/datadog/di/probe_manager.rb
p-datadog: Removed in https://github.com/DataDog/dd-trace-rb/pull/4097.

lib/datadog/di/probe_manager.rb
p-datadog: The way I had it was intentional because I am modifying the container during iteration.

lib/datadog/di/probe_manager.rb
p-datadog: It's an internal attribute, even though it is set from outside of the class.

p-datadog:
Replaced by the PRs listed above.

PR #4064, opened 2024-11-05
Initial

sanchda:
**What does this PR do?**
libdatadog v14.0.0 ships a new release of crashtracking, which makes the component safer for a wider variety of applications.  At a high level, the main design difference is that crashtracking now waits for a crash to spawn a receiver process, and it tries very hard to ensure that process is reaped by the crashing process.

* Change the top-level crashtracking timeout from seconds to milliseconds (and improve the implementation of that timeout to be more strict)
* Remove init functions--now there's just one
* Remove "wait for receiver" as this is now implied
* Decouple altstack creation from altstack usage

**Motivation:**
Hopefully this fix makes crashtracking better for Ruby


ivoanjo:
Thanks @sanchda for starting the libdatadog 14 upgrade! It got finished in #4065, so closing this one for now :)

PR #4065, opened 2024-11-05
[NO-TICKET] Upgrade to libdatadog 14.0

ivoanjo:
**What does this PR do?**

This PR upgrades the libdatadog dependency to the latest 14.0 release.

It also updates our API usage to match some 13.1 -> 14.0 changes.

**Motivation:**

The libdatadog 14.0 release includes a number of improvements to the crashtracking implementation, and we want to pick them up.

**Change log entry**
Upgraded libdatadog dependency to version 14.0

**Additional Notes:**

As a first pass, I've disabled the creation and use of the alt stack, to match our existing setup.

I did not yet experiment with `use_alt_stack = true`.

**How to test the change?**

Existing test coverage should be enough to cover our libdatadog API usage.

### DIFF
--- a/datadog.gemspec
+++ b/datadog.gemspec
@@ -69,7 +69,7 @@ Gem::Specification.new do |spec|
 
   # When updating the version here, please also update the version in `libdatadog_extconf_helpers.rb`
   # (and yes we have a test for it)
-  spec.add_dependency 'libdatadog', '~> 13.1.0.1.0'
+  spec.add_dependency 'libdatadog', '~> 14.0.0.1.0'
 
   spec.extensions = [
     'ext/datadog_profiling_native_extension/extconf.rb',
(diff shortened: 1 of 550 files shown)


Strech (approved):
I hope no things forgotten to be set 🙈

ivoanjo:
> I hope no things forgotten to be set 🙈

I've dropped a note to @sanchda which has been doing a lot of working on libdatadog crashtracking so he can do a quick pass on that part as well :)

sanchda (approved):
The only issue, which we discussed offline, is whether or not it's appropriate to use the altstack even if it isn't created.  I trust @ivoanjo to make whatever decision is appropriate for the runtime, so approving.

👍 🚀 🚀

p-datadog (approved):

p-datadog (approved):

ivoanjo:
I spent some time looking into `use_alt_stack` setting and I think there's a bug in libdatadog and that setting doesn't seem to be doing anything at this point.

Since the Ruby VM itself sets an altstack, this leaves us:

* In libdatadog 13.1, the crashtracker code always used the altstack (e.g. equivalent of `use_alt_stack` was hardcoded to `true`) => in libdatadog 13.1 we used the Ruby altstack
* In libdatadog 14.0, the crashtracker code chooses to use the altstack based on the value of `create_alt_stack` => libdatadog 14.0 we can't use the Ruby altstack

Since the crashtracker is currently off by default, I don't think this should be a blocker for upgrading to libdatadog 14.

In preparation for `use_alt_stack` to get fixed in a future libdatadog release, I'll push a commit to change it to true, although, again, right now it's a no-op.

AlexJF (approved):

PR #4014, opened 2024-10-18
Replace 'debase' with 'datadog'-ruby_core_source

sarahchen6:
**Change log entry**
Replace the `debase-ruby_core_source` gem with the `datadog-ruby_core_source`.

**What does this PR do?**
This PR removes the `debase-ruby_core_source` gem from the gemspec and adds `datadog-ruby_core_source` to the Gemfile. It replaces `debase-ruby_core_source` with `datadog-ruby_core_source` but leaves the spec requirements commented out, as `datadog-ruby_core_source` cannot be added to the gemspecs yet.

**Motivation:**
The goal of this PR is to validate `datadog-ruby_core_source` before it is officially published as a Ruby gem. Specifically, we want to ensure that profiling support is maintained.

**Additional Notes:**
This PR is a step toward the overall project goal of creating a new `datadog-ruby_core_source` gem. See [here](https://docs.google.com/document/d/1lbehlvLADt_ztblhlEnnTFKQ8EVxxOWHdgFzTSqD9Qc/edit?tab=t.0#task=YxycgZORWQhz-chI) for more info on the project.

**How to test the change?**
TBD

Unsure? Have a question? Request a review!

PR #3997, opened 2024-10-14
[PROF-10603] Upgrade to libdatadog 13.1

ivoanjo:
**What does this PR do?**

This PR upgrades the datadog gem to use libdatadog 13.1.

There were no needed API changes, it's a drop-in upgrade.

**Motivation:**

Staying on the latest libdatadog enables us to pick up the latest improvements and also makes it easier to develop new features that need libdatadog changes.

**Additional Notes:**

As usual, I'm opening this PR as a draft as libdatadog 13.1 is not yet available on rubygems.org, and I'll come back to re-trigger CI and mark this as non-draft once it is.

**How to test the change?**

Our existing test coverage includes libdatadog testing, so a green CI is good here :)

PR #3996, opened 2024-10-11
Gracefully handle Rails runner patching whenever spans aren't instrumented

tylerwillingham:
**What does this PR do?**

- Resolves #3995 

**Motivation:**

This avoids an `undefined method` error on `nilClass` whenever when attempting `span.name=`. The spec correctly captures the scenario we're seeing during deploys in our Rails app.

**Additional Notes:**

**How to test the change?**

Unsure? Have a question? Request a review!

PR #4067, opened 2024-11-05
Bump to version 2.5.0

Strech:
### Added

* Performance: Profiling: Add setting to lower heap profiling memory use/latency by cleaning up young objects after Ruby GC (#4020)

### Changed

* Core: Replace the `debase-ruby_core_source` gem with the `datadog-ruby_core_source` (#4014)
* Core: Upgrade to `libdatadog` 13.1 (#3997)

### Fixed

* Fix `undefined method` error for Rails runner (#3996)
* Apply version tag only to spans that use the global/default service name (#4027)
* Ensure UDS takes precedence over HTTP when both Agent configurations defined (#4024)


ivoanjo (approved):
:+1: LGTM

PR #4040, opened 2024-10-28
Add Ruby 3.4.0-preview2 support for CircleCI tests

sarahchen6:
**What does this PR do?**
This PR adds "3.4" versions to the `config.yml` file under `.circleci` everywhere there is a comment to `# ADD NEW RUBIES HERE`.

**Motivation:**
Though the official Ruby 3.4 is not yet released, we can temporarily add Ruby 3.4.0-preview2. This allows us to start testing with Ruby 3.4 in our CI and spot -> address issues early. After Ruby 3.4 is officially released post-Dec 25, 2024, we can replace the preview with it.

**Change log entry**
No change log entry for now. Thinking to have a "Ruby 3.4.0-preview2 support added" update after it is fully supported.

**Additional Notes:**
This PR is one step toward full support and only addresses CircleCI workflow tests. It acts alongside #4038 and #4039 .

**How to test the change?**
This change can be tested in CI.

PR #4069, opened 2024-11-05
Use BUNDLE_GEMFILE directly instead of appraisal binary

p-datadog:
**What does this PR do?**
appraisal binary appears to be incompatible with bundler 2.4+ that set BUNDLER_SETUP environment variable.
In some cases developers are already using BUNDLE_GEMFILE manually to invoke appraisal configurations, probably those using ruby 3.3 and newer.

This PR changes the rakefile to always use BUNDLE_GEMFILE approach, making it work for all Ruby versions, and adds documentation stating that appraisal binary does not work and to use the BUNDLE_GEMFILE manual approach.

**Motivation:**

Fixing CI in #4040

**Change log entry**

None

**Additional Notes:**

**How to test the change?**
Tested in #4059

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -59,7 +59,7 @@ namespace :test do
         command = if appraisal_group.empty?
                     "bundle exec rake #{spec_task}"
                   else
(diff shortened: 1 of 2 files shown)


p-datadog:
This PR should repair the need to pin to old versions of bundler as was done in https://github.com/DataDog/dd-trace-rb/pull/4054.

PR #4091, opened 2024-11-08
Add denied commands

lloeki:
**What does this PR do?**

Add denied commands

**Motivation:**

Some commands are problematic or provide no value but high risk.

**Change log entry**

Disable SSI for certain Ruby commands.

**Additional Notes:**

There may be more missing.

**JIRA**

- [INPLAT-100](https://datadoghq.atlassian.net/browse/INPLAT-100)

**How to test the change?**

[INPLAT-100]: https://datadoghq.atlassian.net/browse/INPLAT-100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib-injection/requirements.json
+++ b/lib-injection/requirements.json
@@ -66,6 +66,88 @@
       "envars": null
     },
     {
+      "id": "bundle_install",
+      "description": "Ignore bundle install",
+      "os": null,
+      "cmds": [
+        "**/ruby"
+      ],
+      "args": [
+        {
+          "args": [
+            "*/bundle",
+            "*",
+            "install"
+          ],
+          "position": 1
+        }
+      ],
+      "envars": null
+    },
+    {
+      "id": "bundle_install",
+      "description": "Ignore bundle install",
+      "os": null,
+      "cmds": [
+        "**/ruby"
+      ],
+      "args": [
+        {
+          "args": [
+            "*/bundle",
+            "*",
+            "*",
+            "install"
+          ],
+          "position": 1
+        }
+      ],
+      "envars": null
+    },
+    {
+      "id": "bundle_install",
+      "description": "Ignore bundle install",
+      "os": null,
+      "cmds": [
+        "**/ruby"
+      ],
+      "args": [
+        {
+          "args": [
+            "*/bundle",
+            "*",
+            "*",
+            "*",
+            "install"
+          ],
+          "position": 1
+        }
+      ],
+      "envars": null
+    },
+    {
+      "id": "bundle_install",
+      "description": "Ignore bundle install",
+      "os": null,
+      "cmds": [
+        "**/ruby"
+      ],
+      "args": [
+        {
+          "args": [
+            "*/bundle",
+            "*",
+            "*",
+            "*",
+            "*",
+            "install"
+          ],
+          "position": 1
+        }
+      ],
+      "envars": null
+    },
+    {
       "id": "bundle_update",
       "description": "Ignore bundle update",
       "os": null,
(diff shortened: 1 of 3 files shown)


lib-injection/requirements.json
```diff
@@ -29,5 +29,1281 @@
...
+      "id": "gem",
+      "description": "Ignore the rubygems CLI",
+      "os": null,
+      "cmds": [
+        "**/ruby"
+      ],
+      "args": [
+        {
+          "args": [
+            "**/gem"
```
andrewlock, line 43: FYI, the `args` only use simple wildcard, not glob patterns, so this is functionally equivalent to `"*/gem"`

lib-injection/requirements.json
```diff
@@ -29,5 +29,1281 @@
...
+      "id": "bundle_install",
+      "description": "Ignore bundle install",
+      "os": null,
+      "cmds": [
+        "**/ruby"
+      ],
+      "args": [
+        {
+          "args": [
+            "**/bundle",
```
andrewlock, line 60: Is it possible to have args _between_ the bundle and `install` args? 🤔

andrewlock (approved):
Nice :+1

lib-injection/requirements.rb
```diff
@@ -0,0 +1,233 @@
...
+    'description' => 'Ignore the rubygems CLI',
+    'os' => nil,
+    'cmds' => [
+      '**/ruby'
+    ],
+    'args' => [{ 'args' => ['**/gem'], 'position' => 1 }],
+    'envars' => nil,
+  }
+
+  # `bundle exec` is the only command we want to inject into.
```
marcotc, line 53: Should we account for `bundle_version_ cmd`?

marcotc (approved):
In general, do we just want to exclude everything under `RbConfig::CONFIG["bindir"]`, except `ruby`?

lib-injection/requirements.json
lloeki: Yes it is (I wanted to chime in on that and a few other things):

```
bundle --gemfile=/path/to/Gemfile.somethingsomething install
```

Unfortunately I don't thing `args` supports this kind of thing:

```
"args": [
  "*/bundle",
  "**" // any number of args, where "*" would be any arg but just one
  "install"
]
```

And there's also this situation which would be matched above yet we might or might not want to inject but really we don't know:

```
"args": [
  "*/bundle",
  "exec"
  "somethingwedontknowabout"
  "install"
]
```

lib-injection/requirements.json
lloeki: Huh I could have sworn swear I changed that to use `*`.

Note: that inconsistency is confusing AF.

lloeki:
> In general, do we just want to exclude everything under RbConfig::CONFIG["bindir"], except ruby?

Yes, that's basically what I did. We can't do it dynamically though (it'd be runtime-dependent, as in the runtime that executes `requirements.rb`) so I hardcoded them.

lib-injection/requirements.rb
```diff
@@ -0,0 +1,233 @@
...
+      lock
+      viz
+      init
+      gem
+      platform
+      clean
+      doctor
+      remove
+      plugin
+      version
```
lloeki, line 79: @andrewlock This is supremely annoying, basically we want "deny everything but `bundle exec`".

This could be achieved with an `allow:` key having higher precedence than `deny:`, defaulting to `allow: []`, thus:

- anything allowed would not go through `deny:`
- is backwards compatible

```
  "allow": [
    {
      "id": "bundle_exec",
      "description": "Allow bundle exec",
      "os": null,
      "cmds": [
        "**/ruby"
      ],
      "args": [
        {
          "args": [
            "**/bundle",
            "exec"
          ],
          "position": 1
        }
      ],
      "envars": null
    }
  ],
  "deny": [
    {
      "id": "bundle_not_exec",
      "description": "Deny everything but bundle exec",
      "os": null,
      "cmds": [
        "**/ruby"
      ],
      "args": [
        {
          "args": [
            "**/bundle"
          ],
          "position": 1
        }
      ],
      "envars": null
    }
  ],
```

lib-injection/requirements.json
andrewlock: Correct, it does not support that, it's a known limitation unfortunately

lib-injection/requirements.rb
Strech: Not only annoying, but with new command added we will chase the edge all the time. What you've suggested is way better (and safer)

TonyCTHsu: I totally agree with this.

TonyCTHsu (approved):

lib-injection/requirements.json
lloeki: So the closest thing I can do is:

```
"args": [
  "*/bundle",
  "install"
]
"args": [
  "*/bundle",
  "*"
  "install"
]
"args": [
  "*/bundle",
  "*"
  "*"
  "install"
]
```

and hope nobody is going to use more than N args.

PR #4092, opened 2024-11-08
Changed forcing system-tests from env var in .yml to .json file

vpellan:
**What does this PR do?**

This PR change the forcing of system-tests from environment variable in system-tests.yml to a .json file. This enables the forcing of system-tests on multiple scenarios at the same time. This will also be easier to clean in the post release steps of fast_castle.

**Motivation:**

fast_castle post release cleanup step currently contain bugs, and the current force-executing of tests feature is pretty limited.

**Change log entry**

None

**Additional Notes:**

Github Actions is very limited and I believe poorly documented, I'd really enjoy any advice on what/how it could be done better

**How to test the change?**

Add a test you want to force-execute in the .github/forced-tests-list.json file by following the documentation template


.github/workflows/system-tests.yml
```diff
@@ -11,9 +11,7 @@ on:
 env:
   REGISTRY: ghcr.io
   REPO: ghcr.io/datadog/dd-trace-rb
-  ST_REF: main
-  FORCE_TESTS:
-  FORCE_TESTS_SCENARIO:
+  ST_REF: main # This must always be set to `main` on dd-trace-rb's master branch
```
Strech, line 14: If we control the naming, I would suggest `SYSTEM_TESTS_REF` or so.

.github/workflows/system-tests.yml
```diff
@@ -111,6 +109,8 @@ jobs:
     name: Build (${{ matrix.app }})
     permissions:
       packages: write
+    outputs:
+      FORCED_TESTS_LIST_JSON: ${{steps.read_forced_tests_list.outputs.FORCED_TESTS_LIST_JSON}}
```
Strech, line 113: As a good practice - don't include the data structure in the name, because if structure change it will force the name change.

```suggestion
      FORCED_SYSTEM_TESTS: ${{steps.read_forced_tests_list.outputs.FORCED_TESTS_LIST_JSON}}
```

Strech (changes requested):
I think the whole approach is correct 👍🏼 I would suggest to improve the naming for rest without good context of what's going on.

.github/workflows/system-tests.yml
vpellan: Fixed in f7dc462f057c266beba9f44ed93b7d3fc38930a5

.github/workflows/system-tests.yml
vpellan: Fixed in ad349271a55222d69c2b9b5d40d8ef36f52d5ee9

p-datadog (approved):

p-datadog (approved):

Strech (approved):

PR #4095, opened 2024-11-08
Update Gemfile for Ruby 3.4

sarahchen6:
**What does this PR do?**
This PR updates the Gemfile to support Ruby 3.4

**Motivation:**
Ruby 3.4 no longer includes a few necessary default gems (see [here](https://stdgems.org/new-in/3.4/)). The newest version of the `webrick` gem  has been released, and the `strscan` gem version that supports Ruby 3.4 has not been released. This PR addresses these points.

**Change log entry**
No change log necessary.

**Additional Notes:**

**How to test the change?**
Green CI.


Gemfile
```diff
@@ -90,3 +86,23 @@ end
...
+# TODO: Remove strscan specification once v3.1.1 is released.
+if RUBY_VERSION.start_with?('3.4.')
+  gem 'strscan',
+    git: 'https://github.com/ruby/strscan',
+    ref: '041b15df4ccc067deabd85fd489b2c15961d0e2f'
+  # No longer bundled by default since Ruby 3.4
+  gem 'base64'
+  gem 'bigdecimal'
+  gem 'mutex_m'
+end
```
ivoanjo, line 108: It may be worth checking if these dependencies are actually used in production or only for tests; that is -- if our gem needs these things to work on Ruby 3.4, we may need to consider adding them as real dependencies (on our `.gemspec`) or at least documenting it somewhere.

sarahchen6: Sounds good!

sarahchen6:
This PR' changes (along with #4111 ) have been combined in #4114.

PR #4098, opened 2024-11-11
DEBUG-2334 remaining dynamic instrumentation code

p-datadog:
**What does this PR do?**

- Adds DI Remote and integrates DI into remote config processing
- Integration test starting from remote config parsing, through instrumentation to snapshot sending to backend
- **DI is now enabled by default**
- **DI code tracking is now enabled by default (script_compiled trace point) if DD_DYNAMIC_INSTRUMENTATION_ENABLED` is set in environment`**

**Motivation:**

Initial DI implementation in Ruby

**Change log entry**
Dynamic instrumentation is now available in Ruby as a Preview. Currently only log probes are implemented, they can be set on both methods and lines. If you are interested in trying out Dynamic Instrumentation in Ruby, please contact Datadog to be added to the Preview program.

**Additional Notes:**

Replaces https://github.com/DataDog/dd-trace-rb/pull/4063

**How to test the change?**
Additional unit and integration tests are included + system tests in https://github.com/DataDog/system-tests/pull/3516


lib/datadog/di.rb
```diff
@@ -58,6 +59,32 @@ def activate_tracking!
...
+          rescue => exc
+            if defined?(Datadog.logger)
+              Datadog.logger.warn("Failed to activate code tracking for DI: #{exc.class}: #{exc}")
+            else
+              # We do not have Datadog logger potentially because DI code tracker is
+              # being loaded early in application boot process and the rest of datadog
+              # wasn't loaded yet. Output to standard error.
+              warn("Failed to activate code tracking for DI: #{exc.class}: #{exc}")
+            end
+          end
```
marcotc, line 84: Small nit:
```suggestion
        return if RUBY_VERSION >= '2.6'

        # Activate code tracking by default because line trace points will not work
        # without it.
        Datadog::DI.activate_tracking!
        rescue => exc
          if defined?(Datadog.logger)
            Datadog.logger.warn("Failed to activate code tracking for DI: #{exc.class}: #{exc}")
          else
            # We do not have Datadog logger potentially because DI code tracker is
            # being loaded early in application boot process and the rest of datadog
            # wasn't loaded yet. Output to standard error.
            warn("Failed to activate code tracking for DI: #{exc.class}: #{exc}")
        end
```

lib/datadog/di/remote.rb
```diff
@@ -0,0 +1,122 @@
...
+          receiver do |repository, _changes|
+            # DEV: Filter our by product. Given it will be very common
+            # DEV: we can filter this out before we receive the data in this method.
+            # DEV: Apply this refactor to AppSec as well if implemented.
+
+            component = DI.component
+            # TODO when would this happen? Do we request DI RC when DI is not on?
+            # Or we always get RC even if DI is not turned on?
+            # -- in dev. env when component is not built but config is still requested?
+            # TODO log something?
```
marcotc, line 38: I don't think you should have this remote configuration class receiving data if DI's component was never initialize.
I'd say we can assume that this is an impossible case. It makes sense to check if DI's component is always initialized early (which I think it is).

lib/datadog/di.rb
p-datadog: I think you meant to reverse the sign on comparison which I did.

lib/datadog/di.rb
```diff
@@ -77,29 +104,10 @@ def code_tracking_active?
...
-    else
-      # We do not have Datadog logger potentially because DI code tracker is
-      # being loaded early in application boot process and the rest of datadog
-      # wasn't loaded yet. Output to standard error.
-      warn("Failed to activate code tracking for DI: #{exc.class}: #{exc}")
-    end
-  end
-end
-=end
+Datadog::DI.activate_tracking
```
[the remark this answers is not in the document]
p-datadog: Added a comment and this is now conditional on DI environment variable being set.

lib/datadog/di/remote.rb
p-datadog: Adjusted the comment and maybe we can inject the DI component into here at some point in the future instead of having to reference a global. I think this is the only place besides code tracker (which necessarily must reference a global, since normally it's created before DI component exists at all) that doesn't have all of its dependencies injected.

lib/datadog/core/configuration.rb
```diff
@@ -258,8 +258,16 @@ def build_components(settings)
       def replace_components!(settings, old)
         components = Components.new(settings)
 
+        # Carry over state from existing components to the new ones.
+        # Currently, if we already started the remote component (which
+        # happens after a request goes through installed Rack middleware),
+        # we will start the new remote component as well.
+        old_state = {
+          remote_started: old.remote&.started?,
+        }
```
p-datadog, line 267: @marcotc I added a note here per your comment in the other PR, does it sound OK to you?

PR #4100, opened 2024-11-11
Remove static dependency task for each group

TonyCTHsu:
**Motivation:**

Providing dependency task for each group is not ideal, because

- `Dir.glob(AppraisalConversion.gemfile_pattern)` is cached for tasks definition.
- Adding/removing gemfiles makes the cache obsolete. 

**What does this PR do?**

Glob gemfile at runtime instead. 

PS. There will be follow up PR to replace `appraisal`

### DIFF
--- a/tasks/dependency.rake
+++ b/tasks/dependency.rake
@@ -3,11 +3,13 @@ require 'open3'
 require_relative 'appraisal_conversion'
 
 namespace :dependency do
-  gemfiles = Dir.glob(AppraisalConversion.gemfile_pattern)
-
   # Replacement for `bundle exec appraisal list`
   desc "List dependencies for #{AppraisalConversion.runtime_identifier}"
-  task :list do
+  task :list do |t, args|
+    pattern = args.extras.any? ? args.extras : AppraisalConversion.gemfile_pattern
+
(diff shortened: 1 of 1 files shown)


tasks/dependency.rake
```diff
@@ -2,13 +2,32 @@ require 'open3'
...
+
+    gemfiles = Dir.glob(pattern, base: AppraisalConversion.root_path)
+
+    puts "Ahoy! Here is a list of gemfiles you are looking for:\n\n"
+
+    puts "========================================\n"
+    puts gemfiles
+    puts "========================================\n"
+
+    puts "You can do a bunch of cool stuff by assign it to the BUNDLE_GEMFILE environment variable, like:\n"
```
p-datadog, line 19: "assigning a gemfile path"

p-datadog (approved):
Does this change the recommended appraisal workflow as described in https://github.com/DataDog/dd-trace-rb/blob/master/docs/DevelopmentGuide.md ?

TonyCTHsu:
@p-datadog I am still working removing the appraisal command line usage. I plan to update our development guide once it is finish. So far it should remain compatible.

PR #4101, opened 2024-11-11
Use image contains updated gem

TonyCTHsu:
**Why**

I published docker images that contains rubygems 3.5.21 (containing 2.5.21 Bundler). Bundler 2.3.26 is still installed.  

**What does this PR do?**

1. Bump `binary_version` to purge image cache from `images-rb`
2. `"bundle check || bundle install && bundle exec rake #{spec_task}"` To install, when missing dependencies from stale cache
3. Use `Bundler.with_unbundled_env` to deactivate gems when conflicting


.circleci/config.yml
```diff
@@ -164,6 +164,12 @@ orbs:
         <<: *test_job_default
         steps:
           - checkout
+          - run:
+              name: Know the environment
+              command: |
+                ruby -v
+                gem -v
+                bundle -v
```
sarahchen6, line 172: Should this section be kept?

sarahchen6 (approved):

sarahchen6 (approved):
Thanks for addressing this!

.circleci/config.yml
TonyCTHsu: I like to keep those debug info handy, and they are cheap to keep

PR #4102, opened 2024-11-12
CircleCI cache fix

TonyCTHsu:
**Motivation:**

CircleCI cache is corrupted. Running `bundle exec appraisal install` would return a false negative result that dependencies are still missing. 

For example: https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/17478/workflows/d773b359-cf13-4438-821d-150202473ba3

The bundle cache is around 50 mb and the missing dependencies are install with a fallback when running in the task.

**What does this PR do?**

1. Replace the broken `bundle exec appraisal install` with `bundle exec rake dependency:install` 
2. Adding `gemfiles/*.gemfile` into bundle checksum 
3. Other minor improvements


Strech (approved):
LGTM 👍🏼

y9v (approved):

PR #4105, opened 2024-11-12
DEBUG-2334 require file presence for line probe

p-datadog:
**What does this PR do?**

Adds checks that a line probe contains a file, which is required to instrument such a probe.

**Motivation:**
While testing manually I encountered an exception due to a line probe missing the file. This should not happen and such a configuration is not usable since we don't know which file to instrument. Add extra guards and a unit test to handle this case.

**Change log entry**
None

**Additional Notes:**

**How to test the change?**
Unit tests


sarahchen6 (approved):
Looks good!

PR #4120, opened 2024-11-15
DEBUG-2334 Ruby 2 compatibility for dynamic instrumentation

p-datadog:
**What does this PR do?**
Adds special handling for Ruby 2 to dynamic instrumentation. Specifically,  Ruby 2.6 is not able to execute `super(*args, **kwargs)` when the target method has no arguments.

**Motivation:**
Ruby 2 support was on the roadmap for DI, however the lack of support was exposed by tests added in https://github.com/DataDog/dd-trace-rb/pull/4106.

**Change log entry**
None

**Additional Notes:**
More tests need to be added for Ruby 2  - the folding of keyword arguments into positional `options` argument at the end of a method.

**How to test the change?**
Unit tests are included.

### DIFF
--- a/lib/datadog/di/instrumenter.rb
+++ b/lib/datadog/di/instrumenter.rb
@@ -105,7 +105,7 @@ module Datadog
                 serializer.serialize_args(args, kwargs)
               end
               rv = nil
-              # Under Ruby 2.x we cannot just call super(*args, **kwargs)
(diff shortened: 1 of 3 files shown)


TonyCTHsu (approved):
Wow, nice that the test suite catch this!

PR #4128, opened 2024-11-18
Fix execution spec with cucumber on ruby 3.4

p-datadog:
**What does this PR do?**
Cucumber is broken on ruby 3.4 currently, requires the fix in https://github.com/cucumber/cucumber-ruby/pull/1757 (not released yet).

Use their master from git on ruby 3.4 and unpend the test.

**Motivation:**
Investigating intermittent failure in that test on the other Ruby versions.

**Change log entry**
None

**Additional Notes:**

**How to test the change?**

Existing test is unpended.


sarahchen6 (approved):
👍👍

PR #4134, opened 2024-11-20
DEBUG-2334 Set duration to 0 for DI line probes

p-datadog:
**What does this PR do?**
This PR  changes the probe notification builder to set duration to 0 in the generated payload if it was not supplied by the caller.

**Motivation:**
Line probes do not have a meaningful duration, however schema validation in system tests demands a numeric value.

Follow the behavior of other tracers and send 0 instead of nil in this case.

**Change log entry**
None

**Additional Notes:**

This PR also includes additional assertions in the unit test for probe notification builder to make that test in general way more useful. Some of these assertions already exist in integration tests but here they are closer to the code.

**How to test the change?**
Tested by system tests (to be committed in a later PR) and by the unit tests in this PR.

### DIFF
--- a/lib/datadog/di/probe_notification_builder.rb
+++ b/lib/datadog/di/probe_notification_builder.rb
@@ -121,7 +121,7 @@ module Datadog
           },
           # In python tracer duration is under debugger.snapshot,
           # but UI appears to expect it here at top level.
-          duration: duration ? (duration * 10**9).to_i : nil,
+          duration: duration ? (duration * 10**9).to_i : 0,
           host: nil,
           logger: {
             name: probe.file,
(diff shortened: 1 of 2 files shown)


TonyCTHsu (approved):
Default 0 make sense to me. Can remove other nil check from the caller. 

I assumed to input duration can be `nil` or `Float`, otherwise, `nil.to_i` can convert to 0 as well.

p-datadog:
It is a `number` which I'm assuming is a Float? https://github.com/DataDog/system-tests/blob/main/utils/interfaces/schemas/library/debugger/v1/input-request.json#L9

PR #4140, opened 2024-11-21
Guard injection against old rubies

lloeki:
**What does this PR do?**

Add a known-processable frontend script to the main injection script.

**Motivation:**

Main injection script may be attempted to be loaded via `RUBYOPT=-r'/path/to/host_inject.rb'` but be unprocessable.

**Change log entry**

Add safe NOOP injection script for very old rubies

**Additional Notes:**

**How to test the change?**


lib-injection/host_inject.rb
```diff
@@ -1,249 +1,7 @@
...
-      ENV['BUNDLE_GEMFILE'] = gemfile
-      utils.debug "Fork success: Using Gemfile `#{gemfile}`"
-    else
-      utils.debug 'Fork abort'
-    end
-  else
-    utils.debug 'Fork not supported... skipping injection'
-    telemetry.emit(pid, utils.version, [{ name: 'library_entrypoint.abort', tags: ['reason:fork_not_supported'] }])
-  end
+if RUBY_VERSION.start_with?('2.5.')
```
p-datadog, line 5: Should this be `>=` ?

TonyCTHsu: @p-datadog is right!

[Link to red vaccine ](https://github.com/TonyCTHsu/vaccine/actions/runs/11956259855/job/33330597924 )

TonyCTHsu (changes requested):
Since the PR adds a new file, this script needs to be updated.
https://github.com/DataDog/dd-trace-rb/blob/0be95cb6681e472065ab137755982ab0283dd090/.gitlab/prepare-oci-package.sh#L7

lib-injection/host_inject.rb
lloeki: Oh my 🤦. I guess that's what you get when writing code 5min before signing off for the day xD

lib-injection/host_inject.rb
```diff
@@ -0,0 +1,7 @@
+# This file's intent is to be parseable and executable by all ruby versions
+# to call into the main one only for versions for which it is known-compatible
+# with at the language level.
+
+if RUBY_VERSION >= '2.5.'
```
lloeki, line 5: @TonyCTHsu since Vaccine is testing down to 2.3, should we lower this to 2.3?

TonyCTHsu: We can do that, but vaccine is testing the last patch version of 2.3

```
% docker run --rm ghcr.io/datadog/images-rb/engines/ruby:2.3 bash -c 'ruby -e "puts RUBY_VERSION"'
2.3.8
```

TonyCTHsu (approved):

lib-injection/host_inject.rb
lloeki: Technically true, although as far as parsing/processing goes 2.3.0 to 2.3.8 are the same, so I expect it to be OK.

PR #4141, opened 2024-11-21
Move `_dd.apm.enabled` tag in tracing Rack middleware

vpellan:
**What does this PR do?**
It moves the addition of `_dd.apm.enabled` tag to spans in tracing's Rack middleware instead of Appsec middleware.

**Motivation:**
During the review of Standalone SCA system-tests, I found out that enabling 'Standalone Appsec Billing' (`DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED=true`) and disabling 'Appsec Threats' ( `DD_APPSEC_ENABLED=false` ) would make the tests fails as `_dd.apm.enabled` is set in appsec Rack middleware.

This is not correct as this tag is suppose to communicate to the WAF that we should not bill for APM traces, and Standalone Appsec Billing will still limit the number of traces to 1 per minute to keep services alive on the backend side even if Appsec Threats is not enabled, thus the traces must contain that tag.

**Change log entry**

None.

**How to test the change?**

Standalone SCA system-tests (./run.sh SCA_STANDALONE) should XPASS (or pass if force executed)


lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -134,6 +134,11 @@ def set_request_tags!(trace, request_span, env, status, headers, response, origi
             # Otherwise, the getter method would delegate to its root span
             trace.resource = request_span.resource unless trace.resource_override?
 
+            # We add this metric when ASM standalone is enabled to make sure we don't bill APM
+            if Datadog.configuration.appsec.standalone.enabled
+              request_span.set_metric(Tracing::Metadata::Ext::TAG_APM_ENABLED, 0)
+            end
```
p-datadog, line 140: In https://github.com/DataDog/dd-trace-rb/pull/3965 @anmarchenko mentioned tracing gaining a dependency on appsec and that pattern appears to continue here. I my opinion it would be better for tracing to provide hooks that appsec will hook into to avoid tracing referencing appsec in code, but I am not developing either part of the library.

I do also find it hard to ascertain whether the code behaves correctly - this PR sets APM enabled to 0 from tracing, but it is not clear from the diff where/under what conditions APM enabled is set to 1 (or if that happens at all).

p-datadog (approved):

lib/datadog/tracing/contrib/rack/middlewares.rb
TonyCTHsu: Can you elaborate the reason about tagging a `rack` span specifically?

vpellan: The env var name `DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED` is temporary, and what Standalone Appsec actually does is disabling tracing (or more precisely rate-limiting it at 1 trace per minute), and when 'Appsec' is activated along 'Standalone Appsec' (as Standalone Appsec does not in fact activate Appsec by itself), it force-keep traces that contains Appsec events.

There is a RFC currently being reviewed to find the final name for this env var, which will most likely be `DD_TRACE_APM_ENABLED`. When it will be approved, I will also change the configuration from `Datadog.configuration.appsec.standalone.enabled` to something like `Datadog.configuration.tracing.apm.enabled`, which will make more sense and show that there is no dependency from appsec in tracing (but the other way will be true, which already is anyway)

vpellan: > Can you elaborate the reason about tagging a rack span specifically?

For now, the only product other than tracing that uses traces is Appsec, and in Appsec we only instrument rack-based apps (Rails, Sinatra, Rack itself...)
But maybe we could tag somewhere else, like in http frameworks, now that we have moved this part in tracing.

TonyCTHsu: I understand this is experimental feature but similar concern with shotgun surgery and product coupling has been raised in the [previous pull request code review](https://github.com/DataDog/dd-trace-rb/pull/3965#discussion_r1804435163), that signals the severity of design issue. 

Currently, 
1. It is hard to make sense of the convention and logic using `Datadog.configuration.tracing.apm.enabled`, which APM includes multiple products including `profiling`. 
2. I can see similar approach like `rack`, that AppSec contains the 3rd party patches that depends on TraceOperation.
3. Or maybe other kinds of dependency injection mechanism

Overall, the requirements (some kinds of hook systems, perhaps could be encompassed by @lloeki 's Instrumentation API ?) are not clear and I am afraid that if we are indifferent now could cast a long shadow in the future.

.github/workflows/system-tests.yml
```diff
@@ -11,7 +11,7 @@ on:
 env:
   REGISTRY: ghcr.io
   REPO: ghcr.io/datadog/dd-trace-rb
-  SYSTEM_TESTS_REF: main # This must always be set to `main` on dd-trace-rb's master branch
+  SYSTEM_TESTS_REF: igor/standalone-sca # This must always be set to `main` on dd-trace-rb's master branch
```
y9v, line 14: reminder: please change this branch to master before merging

lib/datadog/tracing/contrib/rack/middlewares.rb
y9v: does it make sense to find the root span here and not the rack span?

vpellan: Yes, this is changed in e0c503a06a8d14b82701e51ca0a0cef8dc2167d2

Note that "non-billing" naming is temporary.

vpellan:
As of 27th. of November, the scope of this PR has been changed to moving all "non-billing" mode code to Tracing part.
("Non billing" mode is what was previously referred as "ASM Standalone" mode, that is now referred as "APM Tracing disabled" mode, but since we have already have a different way to completely disable tracing, I refer to it as non-billing mode in this PR to differentiate both, but please note that this is not the official naming for this feature)
This is still a WIP so I changed this PR back to Draft

lib/datadog/tracing/trace_operation.rb
```diff
@@ -373,6 +375,14 @@ def fork_clone
         )
       end
 
+      # When APM is running in 'non-billing' mode, we should skip distributed tracing
+      # unless specific distributed tags are present. These tags can come from upstream
+      # services, or from the service itself.
+      # For now, only '_dd.p.appsec=1' is supported.
+      def non_billing_reject?
```
p-datadog, line 382: This method name contains two negatives. I would prefer something like `should_send?` for example but before renaming let's figure out the overall name for the setting.

lib/datadog/tracing/configuration/ext.rb
```diff
@@ -12,6 +12,7 @@ module Ext
         ENV_OTEL_TRACES_EXPORTER = 'OTEL_TRACES_EXPORTER'
         ENV_HEADER_TAGS = 'DD_TRACE_HEADER_TAGS'
         ENV_TRACE_ID_128_BIT_GENERATION_ENABLED = 'DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED'
+        ENV_APM_ENABLED = 'DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED' # Temporary until RFC decides final name
```
Strech, line 15: Could we highlight that "temporary" nature in the name? It's a big ask to name it `ENV_APM_ENABLED`, but actually check for standalone feature

_P.S let's do it only if it makes sense (merged before RFC is in final state)_

PR #4130, opened 2024-11-18
Run the cucumber test in unbundled environment

p-datadog:
**What does this PR do?**
This PR tries to fix intermittent failures in cucumber `development?` spec by running it in the unbundled environment.

**Motivation:**
The test is flaky.

**Change log entry**
None

**Additional Notes:**
I am not sure that this is the correct fix, but when the test is failing, it appears to be loading `psych` from two locations (ruby base runtime and gem path). Since the test does not clear its environment, it might be inheriting some paths from the test suite, but then tries to load bundler and its own set of gems, which is problematic?

**How to test the change?**

CI consistently passing

PR #4144, opened 2024-11-21
Address psych flaky test

sarahchen6:
**What does this PR do?**
This PR adds the `psych` gem to the cucumber loading script and uninstalls `psych` before the script runs.

**Motivation:**
The goal of this PR is to address a flaky test caused by the `psych` gem double loading. @p-datadog includes details in #4130. Per Oleg's findings, the built-in `psych` gem attempts to override the local/bundle gem. The error is caused when `lib/psych/class_loader.rb` is loaded for the second time. This solution first uninstalls `psych` before reinstalling it in the cucumber loading script in order to avoid the double-load error; however, it does not address the original gem load.

**Change log entry**
No change log entry.

**Additional Notes:**
This PR addresses the @DataDog/ruby-guild issue [here](https://github.com/DataDog/ruby-guild/issues/185).

**How to test the change?**
Given the nature of flakiness, I am testing the change through a Green CI and following up to confirm that no more flaky tests related to the `psych` gem double-loading occur throughout the repo.

### DIFF
--- a/spec/datadog/core/environment/execution_spec.rb
+++ b/spec/datadog/core/environment/execution_spec.rb
@@ -162,7 +162,9 @@ RSpec.describe Datadog::Core::Environment::Execution do
 
       context 'for Cucumber' do
         before do
-          unless PlatformHelpers.ci? || Gem.loaded_specs['cucumber']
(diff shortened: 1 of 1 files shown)


sarahchen6:
Improved by #4145!

PR #4145, opened 2024-11-21
Attempt to fix psych test flakiness

p-datadog:
**What does this PR do?**
This PR tries to fix intermittent failures in cucumber development? spec by explicitly loading *some* `psych` gem in the cucumber launcher.

**Motivation:**
The cucumber test is intermittently failing on Ruby 3.0 through 3.2. When it fails, there are two copies of psych being loaded, and the second copy fails during loading (psych can't overwrite itself).

We haven't figured out what is triggering the second load, but it seems to happen during package installation (i.e., cucumber-*) which requires YAML to parse gem metadata.

https://github.com/DataDog/dd-trace-rb/pull/4130 has additional investigation.

This PR offers a workaround for the issue by always configuring psych to be loaded from (some) gem, thus avoiding the error case where it's loaded from a gem and then is attempted to be loaded from standard library.

**Change log entry**
None

**Additional Notes:**

**How to test the change?**

This PR passed full CI runs 3 times.

### DIFF
--- a/spec/datadog/core/environment/execution_spec.rb
+++ b/spec/datadog/core/environment/execution_spec.rb
@@ -112,6 +112,15 @@ RSpec.describe Datadog::Core::Environment::Execution do
 
         let(:script) do
           <<-RUBY
+            # Under Ruby 3.0 through 3.2 there is a weird error that occurs
+            # in CI where two copies of psych get loaded in the same process,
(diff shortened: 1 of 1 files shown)


sarahchen6 (approved):

p-datadog:
This is still not working (on master).

Issue #1660, opened 2021-08-24
Distributed tracing support through Kafka integration


tak1n:
From what I can tell the Java and NodeJS integration for kafka support distributed tracing:

https://docs.datadoghq.com/tracing/setup_overview/compatibility_requirements/java/#networking-framework-compatibility
```
Datadog’s Kafka integration works with Kafka version 0.11+, which supports the Header API. This API is used to inject and extract trace context.
```

https://github.com/DataDog/dd-trace-js/blob/master/packages/datadog-plugin-kafkajs/src/index.js#L64

Is this a wanted feature and doable? If yes I could try to craft a PR with some guidance.

tak1n:
Not sure how this should work, the operation for consuming is called `kafka.consumer.process_batch` and from the traces I have in front of me it seems 1 trace represents multiple calls of `Consumer#consume`.

We are using https://github.com/karafka/karafka and it seems one can decide whether todo batch consuming or single message consuming https://github.com/karafka/karafka/wiki/Consuming-messages#single-message-consuming.

Means distributed tracing in theory could only work in single message consuming mode.

marcotc:
👋 @tak1n, we have a large effort at Datadog around improving tracing for distributed payloads, Kafka being the most popular system representing such payloads today.

This effort is being championed by the Java team and thus will cascade down to other languages after the groundwork in both the Java tracer and backend/UI have been flushed out.

`ddtrace` will follow suit when the internal specs are more flushed out and Kafka will definitely be one of the supported use cases.

At this moment, I would not recommend you tackle any work regarding adding such support do `ddtrace`, as our data modelling and presentation for traces today is insufficient to captured the complexity of all possible Kafka scenarios.

tak1n:
@marcotc thx, for getting back to me and clarifying it.
This sounds great and I'm really excited how it will shape up 😃

jeffscelza:
Has there been any movement on this?

tostroski:
This is exactly what I am looking for.  Is there any information about where this stands?

delner:
Hey folks, sorry for the delayed response on this. The Ruby team is spread a bit thin, and other teams at Datadog have been doing quite a bit on this topic; we're trying to get a full picture of how Ruby support should fit in with it.

Current expectation is that we should implement some headers within our Kafka instrumentation (in this library) to support distributed tracing with Kafka. At the moment, the details of the tech specs still escape me... trying to get more clarity on that from other teams.

[...]

PR #4147, opened 2024-11-22
Add Karafka integration

nvh0412:
**What does this PR do?**

Fixed https://github.com/DataDog/dd-trace-rb/issues/1660

In this PR, we introduce the Kafka integration for the Kafka gem. Which includes:
1. Distributed tracing by utilizing message.metadata.headers
2. Traces for worker.process and each message executor inside the batch, so we can link them to the origin trace if the distributed tracing is on

**Motivation:**

We’re integrating Karafka to implement proper distributed tracing in our system with Datadog, as it lacks an official integration. This integration will also enable distributed tracing if the message headers include distributed tracing data.

Distributed tracing will help us create a proper service map, connecting Kafka producers and consumers.

**Change log entry**

**Additional Notes:**

**How to test the change?**

Screenshot 2024-11-22 at 3 43 09 PM


docs/GettingStarted.md
```diff
@@ -1093,9 +1093,31 @@ end
...
 `options` are the following keyword arguments:
 
-| Key       | Env Var                         | Type   | Description                                  | Default |
-| --------- | ------------------------------- | ------ | -------------------------------------------- | ------- |
-| `enabled` | `DD_TRACE_KAFKA_ENABLED` | `Bool` | Whether the integration should create spans. | `true` |
+| Key       | Env Var                  | Type   | Description                                  | Default |
+| --------- | ------------------------ | ------ | -------------------------------------------- | ------- |
+| `enabled` | `DD_TRACE_KAFKA_ENABLED` | `Bool` | Whether the integration should create spans. | `true`  |
+
+### Karafka
```
marcotc, line 1100: Can you remove the unrelated changes in the `GettingStarted.md`, and leave the newly introduced Karafka changes only?

lib/datadog/tracing/contrib/karafka/configuration/settings.rb
```diff
@@ -0,0 +1,27 @@
...
+          class Settings < Contrib::Configuration::Settings
+            option :enabled do |o|
+              o.type :bool
+              o.env Ext::ENV_ENABLED
+              o.default true
+            end
+
+            option :service_name
+
+            option :distributed_tracing, default: false, type: :bool
```
marcotc, line 21: I'm curious, for your use case, wouldn't you prefer that `distributed_tracing` is enabled by default?

drichards-87:
Created a Jira card for Docs Team editorial review.

lib/datadog/tracing/contrib/karafka/events/worker/process.rb
```diff
@@ -0,0 +1,77 @@
...
+                    span.set_tag(Ext::TAG_PARTITION, job.executor.partition)
+                    span.set_tag(Ext::TAG_OFFSET, job.messages.first.metadata.offset)
+                  end
+
+                  span
+                end
+
+                ::Karafka.monitor.subscribe 'worker.completed' do |event|
+                  Tracing.active_span&.finish
+                end
```
marcotc, line 60: Having separate locations for span creation `Tracing.trace` and span conclusion `Tracing.active_span&.finish` is always a possible source of hard-to-debug errors and span leakage (and thus memory leaks). We normally only do it when it's impossible to use `Tracing.trace { do_work_here }`.

Also, `Tracing.trace { do_work_here }` takes care of error handling, properly tagging the current span with error information.

In this case, we have the event `worker.processed` that looks like it's just what we need:
https://github.com/karafka/karafka/blob/ab4f9bcd3620f46adb8c0d158b5396b245619ed3/lib/karafka/processing/worker.rb#L58-L78

_**Except**_ that it doesn't call the event listeners when an error is raised by the job. There are no error handlers in this method that call `assigned_listeners`: https://github.com/karafka/karafka-core/blob/a1425725d275796673424c1cd9be517d06518ec9/lib/karafka/core/monitoring/notifications.rb#L120

I opened a PR to Karafka to address this, but even if it's approved, it won't affect users of older versions of the library: https://github.com/karafka/karafka-core/pull/145

The being said, I still lean towards using a single `Tracing.trace { do_work_here }` only because the safety of not having to worry about leaky spans is too advantageous.

docs/GettingStarted.md
nvh0412: ahh, my auto correction setting, thanks @marcotc, will discard them

lib/datadog/tracing/contrib/karafka/events/worker/process.rb
nvh0412: That’s precisely what I’d like to hear from the Datadog team regarding this. My initial intuition when using `Tracing.active_span` was that I couldn’t be certain whether the span I’m currently working on is the one I want to complete. Considering this and your insights, would it be feasible to remove this tracing event? At the moment, I prefer to retain the `Ext::SPAN_MESSAGE_CONSUME` event only for this integration, as it effectively meets my requirement: enabling distributed tracing, ensuring that each message span is linked to the root trace in the distributed tracing system.

appraisal/ruby-2.7.rb
```diff
@@ -193,6 +193,7 @@
   gem 'concurrent-ruby'
   gem 'dalli', '>= 3.0.0'
   gem 'grpc'
+  gem 'karafka'
```
mensfeld, line 196: karafka no longer supports 2.7 based on: https://karafka.io/docs/Versions-Lifecycle-and-EOL/

appraisal/ruby-3.0.rb
```diff
@@ -115,6 +115,7 @@
   gem 'concurrent-ruby'
   gem 'dalli', '>= 3.0.0'
   gem 'grpc', '>= 1.38.0', platform: :ruby # Minimum version with Ruby 3.0 support
+  gem 'karafka'
```
mensfeld, line 118: karafka no longer supports 3.0 based on: https://karafka.io/docs/Versions-Lifecycle-and-EOL/

lib/datadog/tracing/contrib/karafka/events/worker/process.rb
```diff
@@ -0,0 +1,77 @@
...
+  module Tracing
+    module Contrib
+      module Karafka
+        module Events
+          module Worker
+            module Process
+              include Karafka::Event
+
+              def self.subscribe!
+                ::Karafka.monitor.subscribe 'worker.process' do |event|
```
mensfeld, line 16: I would not recommend using this layer for instrumentation of that type and would advise you to reconsider something "closer" to the actual execution of work.

lib/datadog/tracing/contrib/karafka/events/worker/process.rb
mensfeld: > I opened a PR to Karafka to address this, but even if it's approved, it won't affect users of older versions of the library

This PR is not needed to achieve the expected wrapping because it is already done for example for OpenTelemetry:

ref1: https://karafka.io/docs/Monitoring-and-Logging/#opentelemetry
ref2: https://karafka.io/docs/Monitoring-and-Logging/#monitor-wrapping-and-replacement

lib/datadog/tracing/contrib/karafka/integration.rb
```diff
@@ -0,0 +1,42 @@
...
+          def self.version
+            Gem.loaded_specs['karafka'] && Gem.loaded_specs['karafka'].version
+          end
+
+          def self.loaded?
+            !defined?(::Karafka).nil?
+          end
+
+          def self.compatible?
+            super && version >= MINIMUM_VERSION
```
mensfeld, line 28: please note how I do versioning: https://karafka.io/docs/Versions-Lifecycle-and-EOL/#versioning-strategy

nvh0412: I don’t think simply picking the latest or an updated version should be the goal of this compatibility check. Ideally, the `minimum_version` should be the version that supports the specific events we want to subscribe to. Could you suggest one that fits this criteria? I can use 2.0.x, but those versions seem too outdated and 2.1.0 seems fit to me.

lib/datadog/tracing/contrib/karafka/events/worker/process.rb
nvh0412: Would it be possible for you to explain what "closer" means here :)? I utilised what you did on instrumentation/vendors/datadog/logger_listener.rb here, but what I'm reconsidering here that we don't actually need to subscribe to this event anymore, because our goal is to have message traces for each message inside messages enumerator, so we can link them to the distributed traces. So I can remove this even if everything is getting our of control here.

lib/datadog/tracing/contrib/karafka/ext.rb
```diff
@@ -0,0 +1,25 @@
...
+  module Tracing
+    module Contrib
+      module Karafka
+        module Ext
+          ENV_ENABLED = 'DD_TRACE_KARAFKA_ENABLED'
+
+          SPAN_MESSAGE_CONSUME = 'karafka.consume'
+          SPAN_WORKER_PROCESS = 'worker.process'
+
+          TAG_TOPIC = 'kafka.topic'
```
mensfeld, line 13: this is a side note because maybe it is done: please keep in mind that karafka supports multi-consumer group operations in one karafka process thus CG (consumer group) always needs to be reported alongside metrics.

lib/datadog/tracing/contrib/karafka/patcher.rb
```diff
@@ -0,0 +1,59 @@
...
+          def configuration
+            Datadog.configuration.tracing[:sidekiq]
+          end
+
+          def propagation
+            @propagation ||= Contrib::Sidekiq::Distributed::Propagation.new
+          end
+
+          def each(&block)
+            @messages_array.each do |message|
```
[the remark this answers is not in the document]
nvh0412: Regarding what you wrote here @mensfeld https://karafka.io/docs/Consuming-Messages/#in-batches, this messages enumerator is supposed to be the only API interface that our users use for batch operations, am I right? Seems I can't see any alternative ways in this doc, except you're saying these high-scale users delegate the whole messages to another process by saving them to a storage (e.g., like Event.insert_all messages) without calling .each ?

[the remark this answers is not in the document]
mensfeld: > this messages enumerator is supposed to be the only API interface that our users use for batch operations, am I right?

No.

> except you're saying these high-scale users delegate the whole messages to another process by saving them to a storage (e.g., like Event.insert_all messages) without calling .each ?

Yes. Alongside you can use `Messages#payloads` directly omitting the headers when not needed. `Messages#each` is a popular use-case but not the only one. While Karafka provides primitives for both, it my itself makes no assumptions about that nature of the processing, that's why `#each` is not instrumented on my side. The moment I `yield` control to the user, it is on the user to define the nature of the processing.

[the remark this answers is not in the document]
nvh0412: Thanks @mensfeld, I see your point, but honestly, this seems like the most suitable place for me to patch and enable distributed tracing, if the users (who also use Datadog like me) want to use #payloads to do their batching implementation, then loosing distributed tracing, in that case, would be reasonable. 

I understand that the `#each` method in the Karafka gem isn’t instrumented since neither you nor the gem itself needs to expose it. However, this distributed tracing is specific to the `datadog` gem and aligns perfectly with my motivation (in my PR description). So would love to hear thoughts on this from the Datadog team as well. @marcotc 

willing to address if there is another place that we should patch to have our distributed tracing instead.

lib/datadog/tracing/contrib/karafka/integration.rb
nvh0412: happy to start with 2.2

mensfeld: Good with me!

lib/datadog/tracing/contrib/karafka/patcher.rb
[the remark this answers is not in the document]
mensfeld: > willing to address if there is another place that we should patch to have our distributed tracing instead.

The only thing here is a philosophy of tracing. Ideally if we could have both and allow users to go with the approach they are taking for processing their data, then it would be ideal.

FYI I'm not sure if I mentioned this but feel free to ping me if you need any more help. Also happy to help with this work via Slack/etc if needed.

docs/GettingStarted.md
```diff
@@ -1097,6 +1097,27 @@ end
...
+  c.tracing.instrument :karafka, **options
+end
+
+```
+`options` are the following keyword arguments:
+
+| Key                   | Env Var                  | Type   | Description                                         | Default |
+| --------------------- | ------------------------ | ------ | --------------------------------------------------- | ------- |
+| `enabled`             | `DD_TRACE_KARAFKA_ENABLED` | `Bool` | Whether the integration should create spans.        | `true`  |
+| `distributed_tracing` |                          | `Bool` | Enables [distributed tracing](#distributed-tracing) | `false` |
```
drichards-87, line 1119:
```suggestion
| `distributed_tracing` |                          | `Bool` | Enables [distributed tracing](#distributed-tracing). | `false` |
```

docs/GettingStarted.md
```diff
@@ -1097,6 +1097,27 @@ end
...
+Datadog.configure do |c|
+  c.tracing.instrument :karafka, **options
+end
+
+```
+`options` are the following keyword arguments:
+
+| Key                   | Env Var                  | Type   | Description                                         | Default |
+| --------------------- | ------------------------ | ------ | --------------------------------------------------- | ------- |
+| `enabled`             | `DD_TRACE_KARAFKA_ENABLED` | `Bool` | Whether the integration should create spans.        | `true`  |
```
drichards-87, line 1118:
```suggestion
| `enabled`             | `DD_TRACE_KARAFKA_ENABLED` | `Bool` | Specifies whether the integration should create spans.        | `true`  |
```

drichards-87 (approved):
Left a couple of very small suggestions from Docs and approved the PR.

nvh0412:
Hi team and @marcotc 

Let's settle this. My gut feeling about this PR is that we do have some limitations in how tracing is integrated into the Karafka gem. I’ve switched to using `Tracing.trace` with a block rather than starting and finishing traces manually. However, the core of this PR is the distributed tracing with the message block.

This implementation fits perfectly with my system and has been working well with an internal patch. That said, let me know your thoughts on this integration. I’m fine if we can't settle on it, as I can always maintain the internal patch if needed.

marcotc:
Sorry for the delay, I'll circle back here next week to get this PR unblocked and hopefully merged.

mensfeld:
@marcotc feel free to ping me as well prior to merging/deciding what to do with it. I will give it a second look.

lib/datadog/tracing/contrib/karafka/configuration/settings.rb
marcotc: This is a fair point. The `distributed_tracing` implementation is directly changing the parent/child relationship of the trace, and not using [Span Links](https://docs.datadoghq.com/tracing/trace_collection/span_links/) so I think it makes sense to leave it disabled by default.

lib/datadog/tracing/contrib/karafka/monitor.rb
```diff
@@ -0,0 +1,68 @@
...
+require_relative 'ext'
+
+module Datadog
+  module Tracing
+    module Contrib
+      module Karafka
+        # Custom monitor for Karafka
+        class Monitor < ::Karafka::Instrumentation::Monitor
+          TRACEABLE_EVENTS = %w[
+            worker.processed
```
marcotc, line 12: Could we subscribe to the event `worker.processed` (e.g. `Karafka.monitor.subscribe 'worker.processed' do |event|`), instead of creating a custom `::Karafka::Instrumentation::Monitor`?

I ask because in `Patcher::patch` we have `::Karafka::App.config.monitor = Monitor.new`, and I'm thinking that this overrides whatever other monitors the application already had (e.g. Sentry error monitoring). I'm wondering if `Karafka.monitor.subscribe 'worker.processed'` would be less destructive.

marcotc (commented):
Awesome work @nvh0412! And thank you so much for the insightful review @mensfeld!

lib/datadog/tracing/contrib/karafka/monitor.rb
mensfeld: you can do that yes but you wont have ability to `yield` around to instrument with wrapping.

lib/datadog/tracing/contrib/karafka/monitor.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+require_relative 'ext'
+
+module Datadog
+  module Tracing
+    module Contrib
+      module Karafka
+        # Custom monitor for Karafka
```
marcotc, line 9:
```suggestion
        # Custom monitor for Karafka.
        # Creating a custom monitor, instead of subscribing to an event (e.g. `Karafka.monitor.subscribe 'worker.processed'`),
        # is required because event subscriptions cannot wrap the event execution (`yield`).
```

lib/datadog/tracing/contrib/karafka/patcher.rb
```diff
@@ -0,0 +1,60 @@
...
+        module MessagesPatch
+          def configuration
+            Datadog.configuration.tracing[:karafka]
+          end
+
+          def propagation
+            @propagation ||= Contrib::Karafka::Distributed::Propagation.new
+          end
+
+          def each(&block)
```
marcotc, line 21:
```suggestion
          # `each` is the most popular access point to Karafka messages, but not the only one
          #  Other access patterns do not have a straightforward tracing avenue (e.g. `my_batch_operation messages.payloads`)
          # @see https://github.com/karafka/karafka/blob/b06d1f7c17818e1605f80c2bb573454a33376b40/README.md?plain=1#L29-L35
          def each(&block)
```

docs/GettingStarted.md
```diff
@@ -1098,6 +1098,27 @@ end
 | --------- | ------------------------------- | ------ | -------------------------------------------- | ------- |
 | `enabled` | `DD_TRACE_KAFKA_ENABLED` | `Bool` | Whether the integration should create spans. | `true` |
 
+### Karafka
+
+The karafka integration provides tracing of the `karafka` gem:
```
marcotc, line 1103:
```suggestion
The karafka integration provides tracing of the `karafka` gem.
```

lib/datadog/tracing/contrib/karafka/patcher.rb
```diff
@@ -0,0 +1,63 @@
...
+            @messages_array.each do |message|
+              if configuration[:distributed_tracing]
+                trace_digest = Karafka.extract(message.metadata.headers)
+                Datadog::Tracing.continue_trace!(trace_digest) if trace_digest
+              end
+
+              Tracing.trace(Ext::SPAN_MESSAGE_CONSUME) do |span|
+                span.set_tag(Ext::TAG_OFFSET, message.metadata.offset)
+                span.set_tag(Ext::TAG_TOPIC, message.topic)
```
marcotc, line 33: We should add a few standard tags for any messaging system spans, as these enable richer product integration for messaging system observability:
```suggestion
                span.set_tag(Ext::TAG_OFFSET, message.metadata.offset)
                span.set_tag('messaging.destination', message.topic) # Please create a constant for this one, close to `Contrib::Ext::Messaging::TAG_SYSTEM`
                span.set_tag(Contrib::Ext::Messaging::TAG_SYSTEM, 'kafka')
```

marcotc: The same should be done for the span `Datadog::Tracing.trace(Ext::SPAN_WORKER_PROCESS) do |span|`.

marcotc: I think that's the last of it! Thank you so much, @nvh0412!

nvh0412: Thanks @marcotc, let me know if the standard tags I’ve added are correct.

nvh0412:
Hey @marcotc , do you know how to have these 2 envs?

> Neither DATADOG_API_KEY nor DD_API_KEY is in your environment.

re: https://app.circleci.com/pipelines/github/DataDog/dd-trace-rb/19004/workflows/45a01e1d-a2ee-441b-8443-775013999391/jobs/665625

marcotc (approved):
👋 @nvh0412, thank you so much for your patience.
We are in the middle of a CI migration, so the test runs are a bit confusing now.

One thing I noticed missing is the entry for `karafka` in the Rakefile, which means tests are not running in CI. It should be added here: https://github.com/DataDog/dd-trace-rb/blob/30f6e2150af44a2039078d8c8f8a6bbc24a437d7/Rakefile#L222-L269

As well as in the Matrixfile, which you probably copy the current entry for `kafka` to use for your tests: https://github.com/DataDog/dd-trace-rb/blob/30f6e2150af44a2039078d8c8f8a6bbc24a437d7/Matrixfile#L109-L111

nvh0412:
> Sacrifice for the linting gods ☠️

This commit msg actually made my day

taltcher:
Hi - is there an ETA when this feature could be released? this is something we are really waiting for :-)

lib/datadog/tracing/contrib/karafka/integration.rb
```diff
@@ -0,0 +1,45 @@
...
+
+          # Minimum version of the Karafka library that we support
+          # https://karafka.io/docs/Versions-Lifecycle-and-EOL/#versioning-strategy
+          MINIMUM_VERSION = Gem::Version.new('2.0.41')
+
+          # @public_api Changing the integration name or integration options can cause breaking changes
+          register_as :karafka, auto_patch: false
+
+          def self.version
+            Gem.loaded_specs['karafka'] && Gem.loaded_specs['karafka'].version
```
p-datadog, line 23: Can be simplified

lib/datadog/tracing/contrib/karafka/patcher.rb
```diff
@@ -0,0 +1,69 @@
...
+          module_function
+
+          def target_version
+            Integration.version
+          end
+
+          def patch
+            require_relative 'monitor'
+
+            ::Karafka::App.config.monitor = Monitor.new
```
p-datadog, line 62: Could this be replacing another monitor added by the application?

p-datadog: https://github.com/karafka/karafka/wiki/Monitoring-and-logging appears to suggest a different pattern

mensfeld: @p-datadog not really: https://karafka.io/docs/Monitoring-and-Logging/#monitor-wrapping-and-replacement

both are supported and in this case there has to be a custom monitor to handle the tracing as far as I understand (karafka author here)

p-datadog: @mensfeld Thank you for the response. Can you clarify what happens when the monitor is replaced - is the original monitor going to continue receiving events? 

I made a small change that patches the Karafka Monitor to invoke dd-trace-rb instrumentation which should work with any installed monitor. What is your opininion of this change?

p-datadog:
karafka 2.0.41 (min version we are testing) uses Concurrent::Array without declaring a dependency on or requiring concurrent-ruby, which karafka-core 2.3.0 (that is being used currently by CI) does also not use in any way. I think the solution here is to pin to and older karafka-core version for karafka 2.0.41.

mensfeld:
> karafka 2.0.41 (min version we are testing) uses Concurrent::Array without declaring a dependency on or requiring concurrent-ruby, which karafka-core 2.3.0 (that is being used currently by CI) does also not use in any way. I think the solution here is to pin to and older karafka-core version for karafka 2.0.41.

2.0.41 is totally EOL and making any decisions based on it does not make any sense: https://karafka.io/docs/Versions-Lifecycle-and-EOL/#karafka-components-support

this dependency reference was fixed later on. Concurrent ruby was completely removed in 2.3.

lib/datadog/tracing/contrib/karafka/patcher.rb
mensfeld: @p-datadog when monitor is replaced it is replaced. You can use a monitor that wraps a different one and document it. 98-99% of people stick with the default one so it is in my opinion an edge case.

p-datadog:
What minimum karafka version would you suggest we support? 2.3?

mensfeld:
Yes. 2.3 is already EOL but it has all the dependency issues solved and a stable API.

p-datadog:
I tested manually and I see a producer and a consumer trace in datadog UI. Given that the PR was already approved by @marco I am going to merge it now.

p-datadog (approved):

PR #4153, opened 2024-11-22
Telemetry: Add environment variable to disable logs

marcotc:
Adds the `DD_TELEMETRY_LOG_COLLECTION_ENABLED` environment variable to disable telemetry log collection, which is enabled by default.


lib/datadog/core/telemetry/component.rb
```diff
@@ -114,7 +119,7 @@ def integrations_change!
         end
 
         def log!(event)
-          return unless @enabled || forked?
+          return if !@enabled || forked? || !@log_collection_enabled
```
marcotc, line 122: I inverted this condition for the `forked?` boolean value, meaning it does not send logs on forked processes. This is the case for all telemetry signals: they don't send data on forks.

TonyCTHsu: I am pretty sure, telemetry logs are not sent on forks, due to the worker (the workers are not enabled after fork)

p-datadog: I would put this note in the code, either here or in the setting description added earlier in the diff. Future readers of the code are unlikely to think to check PR comments for an explanation of why logging is not happening in a fork.

p-datadog (approved):

ivoanjo:
> Also, this PR disables telemetry logs in forked processes, to align telemetry logs with all other telemetry signals, which do not send on forks.

Interesting -- this was surprising to me. Is the intent here to reduce the amount of data sent?

lib/datadog/core/telemetry/component.rb
marcotc: I added a class-level comment to document that none telemetry is sent from forked processes.

marcotc:
> Interesting -- this was surprising to me. Is the intent here to reduce the amount of data sent?

@ivoanjo, yes. It's a cautionary setting in case log generations suddenly starts spamming logs due to unforeseen changes. This toggle was supposed to be here from the start, when we first introduced telemetry logs, in order to have a safer rollout.

TonyCTHsu (approved):

lib/datadog/core/telemetry/component.rb
p-datadog: :+1:

PR #4154, opened 2024-11-23
Telemetry: Deduplicate log entries

marcotc:
This PR removes duplicate telemetry logs sent within the same batch (currently 10s). The number of occurrences is captured in the new `count` field for the `Log` telemetry event, so we can know how many events were removed as duplicated.

**Change log entry**

None


lib/datadog/core/telemetry/metric.rb
```diff
@@ -41,6 +41,18 @@ def to_h
             }
           end
```
marcotc, line 43: In implementing equality for all the `Datadog::Core::Telemetry::Event` in `event.rb`, the `Datadog::Core::Telemetry::Metric` objects also needed their equality implemented since some events contain metrics, which need to be compared.

sarahchen6 (approved):

p-datadog (approved):
Should the worker spec contain a test for metrics?

Strech (approved):

PR #4159, opened 2024-11-25
Update Action with JRuby9.2, Rails 4 exception

TonyCTHsu:
https://github.com/DataDog/dd-trace-rb/actions/runs/12011727682/job/33481379194

**What does this PR do?**

Update with an exception case which was absent from matrix

### DIFF
--- a/Matrixfile
+++ b/Matrixfile
@@ -191,8 +191,8 @@
     'relational_db' => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ❌ jruby'
   },
   'rails' => {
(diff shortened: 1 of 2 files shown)


tasks/edge.rake
```diff
@@ -7,10 +7,17 @@ require_relative 'runtime_matcher'
...
+    ruby_runtime = "#{RUBY_ENGINE}-#{major}.#{minor}"
+
     candidates = Set.new
 
     TEST_METADATA.each do |_, metadata|
       metadata.each do |group, rubies|
+        # Exceptions:
+        # Rails 4.x is not supported on JRuby 9.2 (which is RUBY_VERSION 2.5)
+        next false if ruby_runtime == 'jruby-9.2' && group.start_with?('rails4')
+
```
ivoanjo, line 20: Hmm I wonder if this is the right place to do this... E.g. it looks like we're reinventing a bit the logic we already have in the `Matrixfile`.

E.g. Maybe we could flip the Rails 4 tests to `❌ jruby`?

ivoanjo (approved):
:+1: Looks reasonable!

PR #4164, opened 2024-11-26
Update libddwaf to 1.18.0.0.0

y9v:
Libddwaf 1.18.0.0.0 had some breaking changes - `actions` in the result is now an array of objects, instead of an array of strings.

**Please don't merge until 2.8.0 is released**

**What does this PR do?**
This PR updates `libddwaf` dependency to 1.18.0.0.0 and fixes specs that were broken with this update.

**Motivation:**
This version of `libddwaf` is required for detecting SQL injections.

**Change log entry**
None

**Additional Notes:**
None

**How to test the change?**
CI is enough

### DIFF
--- a/datadog.gemspec
+++ b/datadog.gemspec
@@ -65,7 +65,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'datadog-ruby_core_source', '~> 3.3'
 
   # Used by appsec
-  spec.add_dependency 'libddwaf', '~> 1.15.0.0.0'
+  spec.add_dependency 'libddwaf', '~> 1.18.0.0.0'
 
   # When updating the version here, please also update the version in `libdatadog_extconf_helpers.rb`
   # (and yes we have a test for it)
--- a/lib/datadog/appsec/event.rb
+++ b/lib/datadog/appsec/event.rb
@@ -142,7 +142,7 @@ module Datadog
           scope.trace.keep! if scope.trace
 
           if scope.service_entry_span
-            scope.service_entry_span.set_tag('appsec.blocked', 'true') if waf_result.actions.include?('block')
+            scope.service_entry_span.set_tag('appsec.blocked', 'true') if waf_result.actions.include?('block_request')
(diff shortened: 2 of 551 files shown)


lib/datadog/appsec/event.rb
```diff
@@ -142,7 +142,7 @@ def tag_and_keep!(scope, waf_result)
           scope.trace.keep! if scope.trace
 
           if scope.service_entry_span
-            scope.service_entry_span.set_tag('appsec.blocked', 'true') if waf_result.actions.include?('block')
+            scope.service_entry_span.set_tag('appsec.blocked', 'true') if waf_result.actions.key?('block_request')
             scope.service_entry_span.set_tag('appsec.event', 'true')
           end
```
ivoanjo, line 147: Quick question: This does match the changes done to the unit specs. Do we have integration specs that are covering this as well? E.g. if the waf library actually used `block___request`, would our tests fail?

I'm assuming the question is yes, but asking just in case.

datadog.gemspec
```diff
@@ -65,7 +65,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'datadog-ruby_core_source', '~> 3.3'
 
   # Used by appsec
-  spec.add_dependency 'libddwaf', '~> 1.15.0.0.0'
+  spec.add_dependency 'libddwaf', '~> 1.18.0.0.0'
```
ivoanjo, line 68: I believe we're planning a release for later today, so reminder that if you press that merge button, this is going to go out today. (You're probably aware, but just in case!)

ivoanjo (approved):
:+1: Left two "just-in-case-check/questions", otherwise it LGTM

lib/datadog/appsec/event.rb
y9v: Yes, we have such tests in `system-tests`. No integration tests in `dd-trace-rb` though.
We aim to work on action handling soon and will address this problem also.

Strech (approved):
Awesome 👏🏼

PR #4167, opened 2024-11-27
Add ActiveRecord instrumentation to detect SQLi in AppSec

y9v:
**What does this PR do?**
This PR adds ActiveRecord instrumentation to AppSec for detecting SQL injections.

**Motivation:**
We want to enable SQL injection detection for users using AppSec. Libddwaf already supports this, so this is just about providing required data to libddwaf.

**Change log entry**
Add SQL injection detection for ActiveRecord for following adapters: `mysql2`, `postgresql`, and `sqlite3`

**Additional Notes:**
This is the initial implementation - without blocking, and without passing the SQL span ID with the appsec event.
Passing the span ID of the SQL span is currently not possible without major changes, since tracing creates the SQL span at a later time point than we do our instrumentation.

**How to test the change?**
This feature can be tested using https://github.com/DataDog/appsec-app-generator (`activerecord-sqli` variant):
```bash
bin/run activerecord-sqli -R 7.2.0 -d postgresql
```


lib/datadog/appsec/contrib/active_record/instrumentation.rb
```diff
@@ -0,0 +1,68 @@
...
+
+module Datadog
+  module AppSec
+    module Contrib
+      module ActiveRecord
+        # AppSec module that will be prepended to ActiveRecord adapter
+        module Instrumentation
+          module_function
+
+          def detect_sqli(sql, adapter_name)
```
anmarchenko, line 11: would you be open to name it `detect_sql_injection` for people outside of ASM because I spent a lot of time trying to guess what's sqli is? 😅

y9v: yes, that's a great point, thanks! I got so used to it, that I don't notice anymore that it's actually an abbreviation

y9v: [daa831f](https://github.com/DataDog/dd-trace-rb/pull/4167/commits/daa831f08b370a4ea965a6f66eb45d1b5a7caebc)

lib/datadog/appsec/contrib/active_record/patcher.rb
```diff
@@ -0,0 +1,53 @@
...
+
+          def prepended_class_name(adapter_name)
+            if ::ActiveRecord.gem_version >= Gem::Version.new('7.1')
+              Instrumentation::InternalExecQueryAdapterPatch
+            elsif adapter_name == :postgresql && !defined?(::ActiveRecord::ConnectionAdapters::JdbcAdapter)
+              Instrumentation::ExecuteAndClearAdapterPatch
+            else
+              Instrumentation::ExecQueryAdapterPatch
+            end
+          end
```
Strech, line 48: I think this method creates confusion. We need an adapter name only for `postgres`, for the rest simple version check is working.

What if we extract the version check logic into just normal assignment at the top of the `patch` method

```ruby
def patch
  instrumentation_module = if ::ActiveRecord.gem_version >= Gem::Version.new('7.1')
    Instrumentation::InternalExecQueryAdapterPatch
  else
    Instrumentation::ExecQueryAdapterPatch
  end
  
  # ... omitted ...
end
```

and then for `postgres` we can use additional check right in place

```ruby
if defined?(::ActiveRecord::ConnectionAdapters::PostgreSQLAdapter)
 instrumentation_module = Instrumentation::ExecuteAndClearAdapterPatch unless defined?(::ActiveRecord::ConnectionAdapters::JdbcAdapter)
 ::ActiveRecord::ConnectionAdapters::PostgreSQLAdapter.prepend(instrumentation_module)
end
```

sig/datadog/appsec/contrib/active_record/instrumentation.rbs
```diff
@@ -0,0 +1,23 @@
+module Datadog
+  module AppSec
+    module Contrib
+      module ActiveRecord
+        module Instrumentation
+          def self?.detect_sqli: (String sql, String adapter_name) -> void
```
Strech, line 6:
```suggestion
          def self?.detect_sql_injection: (String sql, String adapter_name) -> void
```

spec/datadog/appsec/contrib/active_record/patcher_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+    context 'when ActiveRecord version is lower than 7.1' do
+      before do
+        stub_const(
+          '::ActiveRecord',
+          Module.new do
+            module_function def gem_version
+              Gem::Version.new('7.0')
+            end
+          end
+        )
```
Strech, line 36: Maybe a matter of preference, but I find it easier to read

```suggestion
        stub_const('::ActiveRecord', Module.new do
          module_function def gem_version
            Gem::Version.new('7.0')
          end
        end)
```

spec/datadog/appsec/contrib/active_record/patcher_spec.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+require 'datadog/appsec/spec_helper'
+require 'datadog/appsec/contrib/active_record/patcher'
+
+RSpec.describe Datadog::AppSec::Contrib::ActiveRecord::Patcher do
+  describe '#prepended_class_name' do
```
Strech, line 7: Technically this is an internal method (which I hope can just go) and should not be tested directly because it repeats the next section of tests

Strech (changes requested):
🔎 Woah, lots of the digging into internals, congrats on our first RAS(P).

I think we can slightly simplify few bits and pieces on a way.

lib/datadog/appsec/contrib/active_record/patcher.rb
y9v: it's a bit more complicated - we need this adapter only for postgres, but for ActiveRecord versions < 7.1 and not for JRuby

lib/datadog/appsec/contrib/active_record/integration.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require_relative '../integration'
+require_relative 'patcher'
+
+module Datadog
+  module AppSec
+    module Contrib
+      module ActiveRecord
+        # Description of ActiveRecord integration
```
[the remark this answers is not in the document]
y9v: this is actually the same comment as we have for other integration classes, but I agree that I shouldn't copy it :)

lib/datadog/appsec.rb
```diff
@@ -53,6 +53,7 @@ def components
 end
 
 # Integrations
+require_relative 'appsec/contrib/active_record/integration'
```
Strech, line 56: Not a deal-breaker, but I would stick it close to the rails

Strech (approved):
LGTM 👍🏼

PR #4171, opened 2024-11-28
ci: adds APPSEC_BLOCKING scenario in CI

cbeauchesne:
**What does this PR do?**
Adds system-tests scenario APPSEC_BLOCKING in the CI

**Motivation:**
A change (probably #4163) broke blocking feature. this scenario will prevent any future bug

**Change log entry**

**Additional Notes:**
This PR can't be merged before the feature is fixed, or the test adapted

**How to test the change?**
CI is green.

PR #4173, opened 2024-11-28
Fix AppSec libddwaf actions handling after update to 1.17.0.0.0

y9v:
In version 1.17.0.0.0 libddwaf changes the format of the actions that are returned from an array of actions IDs to an array of actions objects. This means that we no longer have to look up actions in the ruleset and can just use actions that libddwaf returns.

https://github.com/DataDog/libddwaf/blob/master/UPGRADING.md#action-semantics

**What does this PR do?**
This PR removes `AppSec::Actions` class, since we don't need to hold and lookup ruleset actions now, libddwaf does it for us. It also fixes custom action handling in `Datadog::AppSec::Response`.

**Motivation:**
Failing system tests that test blocking using custom actions:
https://github.com/DataDog/dd-trace-rb/pull/4171

**Change log entry**
None

**Additional Notes:**
None

**How to test the change?**
CI is enough (system tests have to be run from `cbeauchesne/add-system-tests-scenario` branch)


Strech (approved):
Awesome 🟢

marcotc (approved):

PR #4174, opened 2024-11-28
Update master with 2.7.1 release

TonyCTHsu:
**What does this PR do?**

Update master with patch release


ivoanjo:
I've clicked the button to merge master into this branch so the diff is correct (e.g. the diff is really the differences between `master` and `2.7-stable`; otherwise the diff is really weird.

This was done under the assumption we won't do other releases in the 2.7.x series; if we do need to do that, we can force-push the branch back to [94f50fe](https://github.com/DataDog/dd-trace-rb/pull/4174/commits/94f50fe43d6274aa781b3f0be466c2956ad8c481) and resume from there.

.github/workflows/system-tests.yml
```diff
@@ -456,3 +456,29 @@ jobs:
...
+  parametric:
+    needs:
+      - build-artifacts
+    uses: DataDog/system-tests/.github/workflows/run-parametric.yml@main
+    secrets: inherit
+    with:
+      library: ruby
+      binaries_artifact: system_tests_binaries
+      _experimental_job_count: 8
+      _experimental_job_matrix: '[1,2,3,4,5,6,7,8]'
```
ivoanjo, line 484: I'm not entirely sure if these changes should go to master; I've pinged folks about it internally.

ivoanjo (commented):
:+1: This LGTM. I suggest holding on merging before we get a clarification about the `system-tests.yml` changes, but otherwise it's good to go.

.github/workflows/system-tests.yml
lloeki: Yes, we should run parametric tests. This may have caught the regression (or may not, because parametric test CI did pass on system tests repository, but you get the idea about coverage within our CI)

ivoanjo (approved):
:+1: LGTM

ivoanjo:
Hopefully CI will be clear now and we can merge this!

PR #4177, opened 2024-11-30
GraphQL: Report multiple query errors

marcotc:
Captures GraphQL error information as span events: 
Screenshot 2025-01-17 at 3 59 23 PM

This is necessary because each query can have multiple errors ([GraphQL spec for the "errors" field](https://spec.graphql.org/October2021/#sec-Errors.Error-result-format)), which cannot reported using span tags (span tags only support one error per span).

**Change log entry**
GraphQL query errors are now reported as Span Events. This includes support for multiple errors, if present.

**How to test the change?**
All changes have unit tests and system-tests: https://github.com/DataDog/system-tests/pull/3840


.github/workflows/system-tests.yml
```diff
@@ -11,7 +11,7 @@ on:
 env:
   REGISTRY: ghcr.io
   REPO: ghcr.io/datadog/dd-trace-rb
-  SYSTEM_TESTS_REF: main # This must always be set to `main` on dd-trace-rb's master branch
+  SYSTEM_TESTS_REF: marcotc/graphql-error-events # This must always be set to `main` on dd-trace-rb's master branch
```
marcotc, line 14: Reminder: remove this before merging.
Leaving this here during code review because it proves that this PR passes system-test changes.

brett0000FF (approved):
Approving docs change. Thanks!

.github/forced-tests-list.json
ivoanjo, line 1: I was curious what this did exactly and there's excellent docs to explain it! https://github.com/DataDog/dd-trace-rb/blob/502c4da38bffc97097df024e97047277346fd06a/docs/ForcingSystemTests.md :D

docs/GettingStarted.md
```diff
@@ -880,7 +880,7 @@ The `instrument :graphql` method accepts the following parameters. Additional op
...
-| `with_unified_tracer`    |                            | `Bool`   | (Recommended) Enable to instrument with `UnifiedTrace` tracer for `graphql` >= v2.2, **enabling support for Endpoints list** in the Service Catalog. `with_deprecated_tracer` has priority over this. Default is `false`, using `GraphQL::Tracing::DataDogTrace` instead | `false` |
+| `with_unified_tracer`    | `DD_TRACE_GRAPHQL_WITH_UNIFIED_TRACER` | `Bool`   | (Recommended) Enable to instrument with `UnifiedTrace` tracer for `graphql` >= v2.2, **enabling support for Endpoints list** in the Service Catalog. `with_deprecated_tracer` has priority over this. Default is `false`, using `GraphQL::Tracing::DataDogTrace` instead | `false` |
 | `with_deprecated_tracer` |                            | `Bool`   | Enable to instrument with deprecated `GraphQL::Tracing::DataDogTracing`. This has priority over `with_unified_tracer`. Default is `false`, using `GraphQL::Tracing::DataDogTrace` instead | `false` |
```
ivoanjo, line 884: Minor: I'm assuming we haven't flipped the default for backwards-compatibility. But maybe we could be a bit clearer in here that that's the reason? E.g. something like "Due to backwards compatibility, this is not the default, but we strongly suggest enabling this if possible" or something like that?

(Or we could say, "this will be the default for dd-trace-rb 3.x"...)

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -144,11 +165,17 @@ def trace(callable, trace_key, resource, **kwargs)
                 Contrib::Analytics.set_sample_rate(span, config[:analytics_sample_rate])
               end
 
-              yield(span) if block_given?
+              if (before_callable = before || before_block)
+                before_callable.call(span)
+              end
```
ivoanjo, line 170: Worth perhaps raising if there's a `before && before_block`, just to make sure we don't accidentally introduce bugs in the future?

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -163,6 +190,44 @@ def multiplex_resource(multiplex)
...
+          # given their event name `dd.graphql.query.error`.
+          def add_query_error_events(span, errors)
+            errors.each do |error|
+              e = Core::Error.build_from(error)
+              err = error.to_h
+
+              span.span_events << Datadog::Tracing::SpanEvent.new(
+                Ext::EVENT_QUERY_ERROR,
+                attributes: {
+                  message: err['message'],
```
ivoanjo, line 206: Minor: Can this be `e.message`?

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -163,6 +190,44 @@ def multiplex_resource(multiplex)
...
+          #    {"line" => 3, "column" => 10},
+          #    {"line" => 7, "column" => 8},
+          #   ]
+          # is serialized as:
+          #   ["3:10", "7:8"]
+          def serialize_error_locations(locations)
+            locations.map do |location|
+              "#{location['line']}:#{location['column']}"
+            end
+          end
```
ivoanjo, line 230: Are these locations a GraphQL thing? They're not backtracelocations, right?

sig/datadog/tracing/contrib/graphql/ext.rbs
```diff
@@ -9,6 +9,8 @@ module Datadog
 
           ENV_ANALYTICS_SAMPLE_RATE: "DD_TRACE_GRAPHQL_ANALYTICS_SAMPLE_RATE"
 
+          ENV_WITH_UNIFIED_TRACER: string
+          EVENT_QUERY_ERROR: String
```
ivoanjo, line 13: Minor: I like it when we inline the actual strings here because we usually use the constants everywhere and it seems slightly weird to me if there isn't a test that fails if we rename the constants. Having the values here is equivalent to having a test, and means we need to change them in two places and thus that seems like a deliberate, rather than an accidental thing :)

```suggestion
          ENV_WITH_UNIFIED_TRACER: "DD_TRACE_GRAPHQL_WITH_UNIFIED_TRACER"
          EVENT_QUERY_ERROR: "dd.graphql.query.error"
```

spec/datadog/tracing/contrib/graphql/test_schema_examples.rb
```diff
@@ -134,4 +151,37 @@ def unload_test_schema(prefix: '')
...
+            'message' => 'GraphQL error',
+            'type' => 'GraphQL::ExecutionError',
+            'stacktrace' => include(__FILE__),
+            'locations' => ['1:14'],
+            'path' => ['graphqlError'],
+          }
+        )
+      )
+    end
+  end
```
ivoanjo, line 186: Maybe worth perhaps adding a test for multiple errors? Since that seems a big part of the feature

vendor/rbs/graphql/0/graphql.rbs
```diff
@@ -38,4 +38,8 @@ module GraphQL
     class Multiplex
     end
   end
+
+  class Error
+    def to_h: -> Hash[String, untyped]
+  end
```
ivoanjo, line 44: I'm curious -- why do we need to turn the object into a hash first? Is it harder to get the values we want without the additional intermediate hash?

ivoanjo (approved):
:+1: LGTM! Looks pretty slick!

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -47,14 +47,26 @@ def execute_multiplex(*args, multiplex:, **kwargs)
...
           def execute_query(*args, query:, **kwargs)
-            trace(proc { super }, 'execute', query.selected_operation_name, query: query) do |span|
-              span.set_tag('graphql.source', query.query_string)
-              span.set_tag('graphql.operation.type', query.selected_operation.operation_type)
-              span.set_tag('graphql.operation.name', query.selected_operation_name) if query.selected_operation_name
-              query.variables.instance_variable_get(:@storage).each do |key, value|
-                span.set_tag("graphql.variables.#{key}", value)
-              end
-            end
+            trace(
```
TonyCTHsu, line 50: Nitpick: Can we extract those callable lambda and procs? 

I assume they are static and they looked weird to me that we are creating new ones in the memory for every invocation.

TonyCTHsu (approved):

lib/datadog/tracing/contrib/graphql/unified_trace.rb
marcotc: A graphql thing!

vendor/rbs/graphql/0/graphql.rbs
marcotc: I added a comment capturing this:
```ruby
# {::GraphQL::Error#to_h} returns the error formatted in compliance with the GraphQL spec.
# This is an unwritten contract in the `graphql` library.
# See for an example: https://github.com/rmosolgo/graphql-ruby/blob/0afa241775e5a113863766cce126214dee093464/lib/graphql/execution_error.rb#L32
```

sig/datadog/tracing/contrib/graphql/ext.rbs
marcotc: I never though of having rbs as some sort of test for constants...
I'll update the signatures.

lib/datadog/tracing/contrib/graphql/unified_trace.rb
marcotc: 👍

docs/GettingStarted.md
marcotc: I tweaked the language to add this suggestion. Thank you for the feedback!

lib/datadog/tracing/contrib/graphql/unified_trace.rb
marcotc: It's the (possibly multiple) locations inside the original GraphQL query where the error happened.

lib/datadog/tracing/contrib/graphql/unified_trace.rb
marcotc: I tried to extract them, but they fail with `undefined local variable or method 'query'`.
I forgot that they are closures that capture the value of `query` from the parent's context in order to run, so I can't extract them easily.

zarirhamza (approved):

spec/datadog/tracing/contrib/graphql/test_schema_examples.rb
marcotc: Added!

Issue #4149, opened 2024-11-22
How do I totally turn off tracing in local development? Why is datadog still trying to make network calls?


alexevanczuk:
When I run my Ruby tests, I am getting this:
```
E, [2023-02-21T00:00:01.175004 #34286] ERROR -- ddtrace: [ddtrace] (/Users/alexevanczuk/.rbenv/versions/3.3.5/lib/ruby/gems/3.3.0/gems/ddtrace-1.23.3/lib/datadog/tracing/transport/http/client.rb:41:in `rescue in send_request') Internal error during Datadog::Tracing::Transport::HTTP::Client request. Cause: Errno::ECONNREFUSED Failed to open TCP connection to 127.0.0.1:8126 (Connection refused - connect(2) for "127.0.0.1" port 8126) Location: /Users/alexevanczuk/.rbenv/versions/3.3.5/lib/ruby/3.3.0/net/http.rb:1603:in `initialize'
```

However, I have set this configuration:
```ruby
require 'ddtrace/auto_instrument'
require 'version_tracking'
require 'datadog/appsec'

is_test = ENV['CI'] || Rails.env.test?
Datadog.configure do |c|
  if !is_test
    # ... sets up tracing
  elsif is_test
    c.tracing.test_mode.enabled = true
  end

  c.env = Rails.env
  c.service = DatadogHelper.service_name
  c.version = git_sha

  c.tags = {...}
end
```

Thank you for your help!

Decker87:
+1 on this request... below is a screenshot showing how overwhelming the datadog output can be.

![Image](https://github.com/user-attachments/assets/32718e09-a1a7-43ac-a927-e44c437b2955)

alexevanczuk:
I noticed in this file `lib/datadog/core/diagnostics/environment_logger.rb` there is this code snippet:
```

        # Are we logging the environment data?
        def log?
          startup_logs_enabled = Datadog.configuration.diagnostics.startup_logs.enabled
          if startup_logs_enabled.nil?
            # Do not pollute the logs in a development environment.
            !Datadog::Core::Environment::Execution.development?
          else
            startup_logs_enabled
          end
        end
      end
```

I also see a setting:

```
diagnostics.startup_logs.enabled
```

However, despite setting this, I'm still seeing the startup logs in development 🤔

ivoanjo:
Hey folks! As usual, thanks for reaching out, and sorry for the annoyance.

[...]

PR #4182, opened 2024-12-03
[NO-TICKET] Update documentation to clarify what test mode does

ivoanjo:
**What does this PR do?**

This PR updates our documentation to clarify what test mode does:

1. I've tweaked the wording to make it clear that this feature is used for testing distributed tracing behavior

2. I've explicitly mentioned that traces will still be sent; the previous version of the docs erroneously claimed a no-op transport would be applied, which is not true

3. I've left hints for folks that search for "test" while trying to disable dd-trace-rb in their test suites, that they should look at `tracing.enabled` instead

4. I've removed the bigger code examples for this feature, as it's a very niche thing that almost nobody will want to use, and having the code examples invites copy-paste of something that is probably not what the customer wants

**Motivation:**

The fact this doc is confusing came up in #4149 so it seemed relevant to fix our docs to avoid more confusion in the future.

**Change log entry**

(Not relevant to changelog)

**Additional Notes:**

N/A

**How to test the change?**

Docs-only change :)

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -108,7 +108,7 @@ OR
 
 #### Other Ruby applications
 
-If your application does not use the supported gems (Rails or Hanami) above, you can set it up as follows:
+If your application does not use the above mentioned gems (Rails or Hanami), you can set it up as follows:
 
 1. Add the `datadog` gem to your Gemfile:
 
@@ -2021,7 +2021,6 @@ To change the default behavior of `datadog`, you can use, in order of priority, 
      c.env = ENV['RACK_ENV']
 
      c.tracing.report_hostname = true
-     c.tracing.test_mode.enabled = (ENV['RACK_ENV'] == 'test')
    end
    ```
 
@@ -2067,7 +2066,7 @@ For example, if `tracing.sampling.default_rate` is configured by [Remote Configu
 | `tracing.sampling.span_rules`                          | `DD_SPAN_SAMPLING_RULES`,`ENV_SPAN_SAMPLING_RULES_FILE` | `String`                              | `nil`                        | Sets [Single Span Sampling](#single-span-sampling) rules. These rules allow you to keep spans even when their respective traces are dropped.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           |
 | `tracing.trace_id_128_bit_generation_enabled`          | `DD_TRACE_128_BIT_TRACEID_GENERATION_ENABLED`           | `Bool`                                | `true`                       | `true` to generate 128 bits trace ID and `false` to generate 64 bits trace ID                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
 | `tracing.report_hostname`                              | `DD_TRACE_REPORT_HOSTNAME`                              | `Bool`                                | `false`                      | Adds hostname tag to traces.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           |
-| `tracing.test_mode.enabled`                            | `DD_TRACE_TEST_MODE_ENABLED`                            | `Bool`                                | `false`                      | Enables or disables test mode, for use of tracing in test suites.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                      |
+| `tracing.test_mode.enabled`                            | `DD_TRACE_TEST_MODE_ENABLED`                            | `Bool`                                | `false`                      | Enables or disables "test mode", used for testing distributed tracing behavior. Note that traces will still be sent; consider using `tracing.enabled` instead to disable sending of traces entirely. |
 | `tracing.test_mode.trace_flush`                        |                                                         | `Datadog::Tracing::TraceFlush`        | `nil`                        | Object that determines trace flushing behavior.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        |
 
 #### Custom logging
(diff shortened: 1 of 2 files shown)


anmarchenko (approved):

docs/GettingStarted.md
```diff
@@ -2637,15 +2636,12 @@ or `c.agent.port`.
...
+span will be sent (no sampling is applied) and the transport will send data synchronously.
+It is configured by setting `c.tracing.test_mode.enabled` to true.
 
-```ruby
-Datadog.configure do |c|
-  c.tracing.test_mode.enabled = true
-end
-```
+If you're looking to disable tracing while your test suite is running, consider using the `tracing.enabled` setting
+to disable sending of traces.
```
Strech, line 2644: A minor thing, but it would be cool to give an example on how-to, even if it's obvious

```suggestion
If you're looking to disable tracing while your test suite is running, consider using the `tracing.enabled` setting
to disable sending of traces.

```ruby
Datadog.configure do |c|
  c.tracing.test_mode.enabled = false
end
```
```

**Review by Strech — APPROVED** [review]
LGTM

File: docs/GettingStarted.md (hunk 1 above)
Comment (ivoanjo) [reply docs/GettingStarted.md:2644]: Wait, I'm somewhat confused with this suggestion -- did you mean to suggest an example of using `tracing.enable`?

I explicitly removed the example with `test_mode.enabled` and stated why (item 4) so I'm not sure I understand your suggestion :eyes:

**Review by p-datadog — APPROVED** [review]
Is test mode an internal feature not meant for use by customers at all?

**Comment by ivoanjo** [comment]
> Is test mode an internal feature not meant for use by customers at all?

That's a good question. I've done some digging and this was added by @delner back in https://github.com/DataDog/dd-trace-rb/pull/1504 . So it seems like a vestige of a feature that was added to make CI work.

Perhaps something that we should look to deprecate for 3.x?

**Comment by anmarchenko** [comment]
@p-datadog @ivoanjo test mode is actually used by `datadog-ci` library [like that](https://github.com/DataDog/datadog-ci-rb/blob/e8a361dd0098a4f7b82199fbdeb3726ea2920219/lib/datadog/ci/configuration/components.rb#L93-L96):

```ruby
          settings.tracing.test_mode.enabled = true
          settings.tracing.test_mode.async = true
          settings.tracing.test_mode.trace_flush = settings.ci.trace_flush || CI::TestVisibility::Flush::Partial.new
```

This is not supposed to be used by customers, so at some point we better remove it from the documentation completely.

ivoanjo:
Any thoughts on removing it from the doc in this PR? E.g. it's not like omitting docs is a breaking change ;)

anmarchenko:
@ivoanjo I am not against it, probably worth asking @marcotc about that

docs/GettingStarted.md
```diff
@@ -2637,15 +2636,12 @@ or `c.agent.port`.
 
 #### Transporting in Test Mode
 
-When test mode is enabled, the tracer uses a `Test` adapter for no-op transport that can optionally buffer requests in
-test suites or other non-production environments. It is configured by setting `c.tracing.test_mode.enabled` to true.
-This mode only works for tracing.
+The test mode can be used for testing distributed tracing behavior. Enabling this setting will mean every
+span will be sent (no sampling is applied) and the transport will send data synchronously.
+It is configured by setting `c.tracing.test_mode.enabled` to true.
```
hestonhoffman, line 2641:
```suggestion
Use the test mode for testing distributed tracing behavior. Enabling this setting causes every
span to be sent (no sampling is applied) and causes the transport to send data synchronously.
To configure test mode, set `c.tracing.test_mode.enabled` to `true`.
```

hestonhoffman (commented):
Some minor style suggestions

p-datadog:
I like putting the internal settings under an `internal` namespace (same as what I have done for DI based on someone else's suggestion). Given that the setting is already implemented I would suggest just adding a "for internal DataDog use only" to the docs (in this PR).

marcotc (approved):

hestonhoffman (approved):

marcotc:
> Any thoughts on removing it from the doc in this PR? E.g. it's not like omitting docs is a breaking change ;)

I agree here. Removing it from the public docs, then adding a comment in code on why we removed it (and that we should not expose it in the next major release).

ivoanjo:
Thanks folks! Here's what I've done:

1. Completely removed any references to `test_mode` from `GettingStarted.md`
2. Added a new section to `GettingStarted.md` to tell customers quickly how to disable tracing, and also merged it with the link to the CI gem
3. Removed user docs from the `settings.rb` definition, leaving only a note saying it's for internal use and that we may change it in the future

docs/GettingStarted.md
```diff
@@ -167,6 +167,24 @@ If your Agent runs on a different host or container than your application, or yo
...
+
+If you'd like to disable tracing when running your tests/specs, you can set the `DD_TRACE_ENABLED` environment variable to `false` or via code:
+
+```ruby
+Datadog.configure do |c|
+  c.tracing.enabled = false
+end
+```
+
+Furthermore, if you're trying to get more visibility into your tests or you're struggling with slow or flaky test suites, consider looking at Datadog's [Test optimization](https://docs.datadoghq.com/tests/) via the [`datadog-ci`](https://github.com/DataDog/datadog-ci-rb?tab=readme-ov-file#datadog-test-optimization-for-ruby) gem.
```
anmarchenko, line 186: thanks for mentioning the test optimization tool!

ivoanjo:
Going to go ahead and merge this one in, thanks everyone for the great feedback -- I think it's one less sharp edge that we just sanded off ;)

docs/GettingStarted.md
ivoanjo: Seemed very relevant to mention is folks are searching for "spec"/"test" in our readme :)

PR #4190, opened 2024-12-04
Replace `ubuntu-latest` with pinned version

TonyCTHsu:
**Motivation:**

https://github.blog/changelog/2024-11-05-notice-of-breaking-changes-for-github-actions/#ubuntu-latest-upcoming-breaking-changes

>We will migrate the ubuntu-latest label to ubuntu 24 starting on December 5, 2024 and ending on January 17, 2025. The ubuntu 24 image has a different set of tools and packages than ubuntu 22. We have made cuts to the list of packages so that we can maintain our SLA for free disk space. This may break your workflows if you depend on certain packages that have been removed. Please review this [list](https://github.com/actions/runner-images/issues/10636) to see if you are using any affected packages.

**What does this PR do?**

Pin to ubuntu 22 instead of `latest`

### DIFF
--- a/.github/workflows/add-milestone-to-pull-requests.yml
+++ b/.github/workflows/add-milestone-to-pull-requests.yml
@@ -9,7 +9,7 @@ jobs:
       contents: read
       issues: write
       pull-requests: write
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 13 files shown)


Strech (approved):
Totally makes sense 👍🏼

PR #4193, opened 2024-12-04
create propagate span event and publish it in to_digest

ZStriker19:
**What does this PR do?**

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/lib/datadog/tracing/trace_operation.rb
+++ b/lib/datadog/tracing/trace_operation.rb
@@ -311,7 +311,7 @@ module Datadog
         span_id = @active_span && @active_span.id
         span_id ||= @parent_span_id unless finished?
         # sample the trace_operation with the tracer
-        @tracer&.sample_trace(self) unless sampling_priority
+        events.propagate.publish(span, self)
 
         TraceDigest.new(
           span_id: span_id,
@@ -380,12 +380,14 @@ module Datadog
         attr_reader \
           :span_before_start,
           :span_finished,
-          :trace_finished
+          :trace_finished,
+          :propagate
 
         def initialize
           @span_before_start = SpanBeforeStart.new
           @span_finished = SpanFinished.new
           @trace_finished = TraceFinished.new
+          @propagate = Propagate.new
         end
 
         # Triggered before a span starts.
@@ -399,6 +401,13 @@ module Datadog
         class SpanFinished < Tracing::Event
           def initialize
             super(:span_finished)
+          end
+        end
+
+        #  Triggered when trace is being propagated between applications or contexts
+        class Propagate < Tracing::Event
+          def initialize
+            super(:propagate)
           end
         end
 
--- a/lib/datadog/tracing/tracer.rb
+++ b/lib/datadog/tracing/tracer.rb
@@ -362,6 +362,11 @@ module Datadog
           event_span_op.service ||= @default_service
         end
 
+        events.propagate.subscribe do |event_span, event_trace_op|
+          sample_span(event_trace_op, event_span)
+          sample_trace(event_trace_op) unless event_trace_op.sampling_priority
+        end
+
         events.span_finished.subscribe do |event_span, event_trace_op|
           sample_span(event_trace_op, event_span)
           flush_trace(event_trace_op)


lib/datadog/tracing/trace_operation.rb
```diff
@@ -402,6 +404,13 @@ def initialize
           end
         end
 
+        #  Triggered when trace is being propagated between applications or contexts
+        class Propagate < Tracing::Event
```
delner, line 408: I would like to see this name to reflect the convention established with the other events ("noun-verb") so that (1) its clear what is being propagated and (2) its easier to group events categorically (e.g. by span or by trace)

lib/datadog/tracing/tracer.rb
```diff
@@ -362,7 +362,12 @@ def bind_trace_events!(trace_op)
           event_span_op.service ||= @default_service
         end
 
+        events.propagate.subscribe do |_event_span, event_trace_op|
```
delner, line 365: Curious; why does this even provide an `event_span` parameter? What would that mean? The span that was active when the propagation occurred?

Is this an artifact of a promotion of an `event_span` parameter to an event base class? I'd like to affirm this design before we commit to a particular API for propagation.

lib/datadog/tracing/tracer.rb
```diff
@@ -362,7 +362,12 @@ def bind_trace_events!(trace_op)
...
+          sample_trace(event_trace_op) unless event_trace_op.sampling_priority
+        end
+
         events.span_finished.subscribe do |event_span, event_trace_op|
+          sample_trace(trace_op) unless trace_op.sampling_priority
```
delner, line 370: `sample_trace` is now (possibly) triggered twice or more during the span lifecycle. Do we have short-circuits/guards that prevent sampling twice or help us save on performance?

lib/datadog/tracing/trace_operation.rb
```diff
@@ -311,7 +311,7 @@ def to_digest
         span_id = @active_span && @active_span.id
         span_id ||= @parent_span_id unless finished?
         # sample the trace_operation with the tracer
-        @tracer&.sample_trace(self) unless sampling_priority
+        events.propagate.publish(@active_span, self)
```
delner, line 314: It dawned on me, I don't think `to_digest` is just for propagation, it may also be used for log correlations. My concern is that if we print a log message with correlation behavior on, say at the beginning of a web request, then the sampling decision will actually be made very early (based only off the root span in the worst case.)

Should check this isn't the case...

delner (changes requested):
This is very much in the style of what we discussed; off to a great start.

There are some questions about the design (parameters, naming, idempotency, etc) that I'd like to answer.

I'd also like to see a lot more tests around this new behavior, that affirms a good answer to all the questions above, e.g.:

- Unit tests, including but not limited to...
  - Does the trace pass correct parameters to the propagate event block?
  - Does triggering the propagation event with different sampling priority values cause sample to run properly?
- Feature tests, including but not limited to...
  - When HTTP is injected/log correlation is run/trace is manually continued across execution contexts with `continue_from`, does is trigger the propagation event properly?

lib/datadog/tracing/tracer.rb
ZStriker19: I don't see how sample_trace would be triggered twice during the span lifecycle since if it was triggered once, we'd have a sampling_priority and then we wouldn't sample. 

I'm happy to make the check internal to the method as I don't see a case where we'd ever want to run trace_sampling again after a sampling_priority value has been set.

lib/datadog/tracing/trace_operation.rb
ZStriker19: So I couldn't find it being used in log correlation, however we do use to_digest for opentelemetry here: https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/opentelemetry/trace.rb#L20-L33 Which seems like it could be an issue if I'm understanding the code correctly because this basically runs to_digest every time we create an opentelemetry context? Let me know if I'm off here.

ZStriker19:
> * Does triggering the propagation event with different sampling priority values cause sample to run properly?

Not sure I understand what you mean by this one. The propagation event just triggers to_digest to be run, which will trigger the event in tracer.py to run which will run sampling if there's not already a sampling priority set.

rtrieu (approved):

p-datadog:
Marking as draft since there is an open discussion/request for changes and there have been no changes made to this PR in a while.

ivoanjo:

ivoanjo:
This has been open for a long time, so I'll go ahead and close it down for now. Please do re-open if needed!

PR #4197, opened 2024-12-05
Do not load `redis` gem when incompatible

marcotc:
**Change log entry**
Yes. Avoids loading `ActiveSupport::Cache::RedisCacheStore`, which tries to load `redis >= 4.0.1` regardless of the version of Redis the host application has installed.

**Additional Notes:**

**How to test the change?**
There are automated tests in this PR, by running the Rails cache tests with an older Redis.


lib/datadog/tracing/contrib/active_support/cache/redis.rb
```diff
@@ -22,17 +22,29 @@ module Patcher
...
+              # Patches the Rails built-in Redis cache backend `redis_cache_store`, added in Rails 5.2.
+              # We avoid loading the RedisCacheStore class, as it invokes the statement `gem "redis", ">= 4.0.1"` which
+              # fails if the application is using an old version of Redis, or not using Redis at all.
+              # @see https://github.com/rails/rails/blob/d0dcb8fa6073a0c4d42600c15e82e3bb386b27d3/activesupport/lib/active_support/cache/redis_cache_store.rb#L4
+              def patch_redis_cache_store?(meth)
+                Gem.loaded_specs['redis'] &&
+                  # Autoload constants return `constant` for `defined?`, but that doesn't mean they are loaded...
+                  defined?(::ActiveSupport::Cache::RedisCacheStore) &&
+                  # ... to check that we need to call `autoload?` and check if it returns `nil`, meaning it's loaded.
+                  ::ActiveSupport::Cache.autoload?(:RedisCacheStore).nil? &&
```
ivoanjo, line 40: I'm curious -- are we called "late enough" in the Rails boot process that we would almost always see `RedisCacheStore` if it was going to be used?

I guess in that situation we could always tell folks to load it sooner if they want it patched, or to call `Datadog.configure { }` again after loading it.

ivoanjo (approved):
:+1: LGTM -- really nice catch!

lib/datadog/tracing/contrib/active_support/cache/redis.rb
```diff
@@ -22,17 +22,29 @@ module Patcher
               # For Rails >= 5.2 w/o redis-activesupport...
               # ActiveSupport includes a Redis cache store internally, and does not require these overrides.
               # https://github.com/rails/rails/blob/master/activesupport/lib/active_support/cache/redis_cache_store.rb
-              def patch_redis?(meth)
+              def patch_redis_activesupport?(meth)
```
Strech, line 25: I find the naming confusing because later we anyway patching something within `ActiveSupport`, WDYT of this name?

```suggestion
              def patch_redis_store?(meth)
```

It follows the idea to target specific store, and next method is a good example, you get what you've asked, patch either redis store or redis cache store.

TonyCTHsu (approved):

lib/datadog/tracing/contrib/active_support/cache/redis.rb
marcotc: Yes, we are always called on Rails after initialization is done, right before the application starts serving requests: https://github.com/DataDog/dd-trace-rb/blob/9875835e4c64eaf2479508fb07b5fc441b58a5f8/lib/datadog/tracing/contrib/rails/patcher.rb#L69-L70

lib/datadog/tracing/contrib/active_support/cache/redis.rb
marcotc: Done!

zarirhamza (approved):

kayn1:
When this will be released under new tag?

ivoanjo:
Hey @kayn1 this will be part of the next version (2.8.0), which we plan on publishing this week

PR #4206, opened 2024-12-09
DEBUG-2334 DI: method probes: pass block to target method when rate limited

p-datadog:
**What does this PR do?**

Fixes a bug in dynamic instrumentation where once a method was instrumented with a method probe and the rate limit for the probe was reached, the block would not be passed to target method from DI.

**Motivation:**

Code inspection when implementing snapshot benchmarks (https://github.com/DataDog/dd-trace-rb/pull/4207).

**Change log entry**

No.

**Additional Notes:**

*Reason: while this issue would have affected customer applications, DI has not yet been shipped.*

I added many tests but could not yet reproduce the scenario where the separated calls to `super` were necessary as they are in the non-rate limited path. I added the separated calls in case they are indeed required.

**How to test the change?**

Unit tests are included.

### DIFF
--- a/lib/datadog/di/instrumenter.rb
+++ b/lib/datadog/di/instrumenter.rb
@@ -153,7 +153,25 @@ module Datadog
                 serialized_entry_args: entry_args)
               rv
             else
-              super(*args, **kwargs, &target_block)
+              # The necessity to invoke super in each of these specific
+              # ways is very difficult to test.
+              # Existing tests, even though I wrote many, still don't
+              # cause a failure if I replace all of the below with a
+              # simple super(*args, **kwargs, &target_block).
+              # But, let's be safe and go through the motions in case
+              # there is actually a legitimate need for the breakdown.
+              # TODO figure out how to test this properly.
+              if args.any?
+                if kwargs.any?
+                  super(*args, **kwargs, &target_block)
+                else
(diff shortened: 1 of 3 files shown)


lib/datadog/di/instrumenter.rb
```diff
@@ -152,7 +153,28 @@ def hook_method(probe, &block)
                 serialized_entry_args: entry_args)
               rv
             else
-              super(*args, **kwargs)
+              # stop standard from trying to mess up my code
+              _ = 42
```
marcotc, line 157: Are are you able to do `# standard:disable` instead here?

marcotc (approved):

lib/datadog/di/instrumenter.rb
p-datadog: The issue is not on this line - standard is trying to lift the contents of this branch up to the parent conditional, in the process completely ruining the encapsulation of the fallback logic that is currently very clear. It wants to have a huge single conditional instead of two when the second one is nested in the `else` branch of the first one.

p-datadog: It is not even deterred by my giant comment explaining the code that follows - it thinks it's perfectly fine to move that comment to one of the branches that it pertains to (rather than the entire logic block).

PR #4208, opened 2024-12-09
DEBUG-3176 DI: report an error when instrumenting a loaded file that is not in registry

p-datadog:
**What does this PR do?**

Ruby stores the list of loaded files in $LOADED_FEATURES. Use this list to detect line probes that attempt to instrument files that have been loaded prior to code tracking being active, and report error to the user when instrumenting those files (since they won’t ever be instrumented in production).

**Motivation:**

Better diagnostics for DI.

**Change log entry**

Report instrumentation error in dynamic instrumentation when attempting to instrument line probes and the target file is loaded but not in code tracker registry.

**Additional Notes:**

The check can be expensive because it performs path suffix matching for potentially thousands of file paths corresponding to the loaded files.

**How to test the change?**

Unit tests are included

### DIFF
--- a/lib/datadog/di/error.rb
+++ b/lib/datadog/di/error.rb
@@ -27,6 +27,11 @@ module Datadog
       class DITargetNotDefined < Error
       end
 
+      # Attempting to instrument a line and the file containing the line
+      # was loaded prior to code tracking being enabled.
+      class DITargetNotInRegistry < Error
+      end
+
       # Raised when trying to install a probe whose installation failed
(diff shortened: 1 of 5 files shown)


marcotc (approved):

p-datadog:
Note that this PR only applies to files loaded via `require`, not `load`. Files loaded via `load` are not added to `$LOADED_FEATURES` and  thus cannot be identified by the check added in this PR.

PR #4210, opened 2024-12-09
add supported versions workflow

quinna-h:
**What does this PR do?**

- Adds a script `find_gem_version_bounds.rb` that parses `gemfiles` to get the minimum and maximum supported version for integrations, which is run in the workflow [Generate Supported Versions](https://github.com/DataDog/dd-trace-rb/actions/workflows/generate-supported-versions.yml)
- The script `generate_table_versions.rb` outputs the integrations and minimum supported versions in a markdown table.
- Adds the gem names in the integration declarations

Generated PR example: https://github.com/DataDog/dd-trace-rb/pull/4236

**Motivation:**
Part of Autodoc generation for supported versions.

**Change log entry**
No.

**Additional Notes:**

**How to test the change?**


.github/scripts/find_gem_version_bounds.rb
```diff
@@ -0,0 +1,140 @@
...
+      end
+    end
+  end
+
+  [minimum_gems_ruby, minimum_gems_jruby, maximum_gems_ruby, maximum_gems_jruby]
+end
+
+# Helper: Parse a Gemfile-style gem declaration
+# ex. gem 'ruby-kafka', '~> 5.0'
+def parse_gemfile_entry(line)
```
marcotc, line 53: You can parse `Gemfile`s using `bundler`:
```ruby
require 'bundler'

gemfile_path = 'Gemfile'
definition = Bundler::Definition.build(gemfile_path, nil, nil)
pp definition.dependencies
```

.github/scripts/find_gem_version_bounds.rb
```diff
@@ -0,0 +1,140 @@
...
+  if (match = line.match(/^\s*gem\s+["']([^"']+)["']\s*,?\s*["']?([^"']*)["']?/))
+    gem_name, version_constraint = match[1], match[2]
+    version = extract_version(version_constraint)
+    [gem_name, version]
+  end
+end
+
+# Helper: Parse a Gemfile.lock-style entry
+# matches on ex. actionmailer (= 6.0.6)
+def parse_gemfile_lock_entry(line)
```
marcotc, line 63: You can parse `Gemfile.lock`s using `bundler` too:
```ruby
require 'bundler'

lockfile_path = 'Gemfile.lock'
lockfile_contents = File.read(lockfile_path)

parser = Bundler::LockfileParser.new(lockfile_contents)
parser.specs.each do |spec|
  puts "Gem: #{spec.name}, Version: #{spec.version}"
end
```

.github/scripts/find_gem_version_bounds.rb
```diff
@@ -0,0 +1,140 @@
...
+    return []
+  end
+
+  # Get all subdirectories inside the specified directory
+  Dir.children(directory).select do |entry|
+    File.directory?(File.join(directory, entry))
+  end
+end
+
+mapping = {
```
marcotc, line 100:
```suggestion
# TODO: The gem information should reside in the integration declaration instead of here.
mapping = {
```

.github/scripts/find_gem_version_bounds.rb
```diff
@@ -0,0 +1,140 @@
...
+def get_integration_names(directory = 'lib/datadog/tracing/contrib/')
+  unless Dir.exist?(directory)
+    puts "Directory '#{directory}' not found!"
+    return []
+  end
+
+  # Get all subdirectories inside the specified directory
+  Dir.children(directory).select do |entry|
+    File.directory?(File.join(directory, entry))
+  end
```
marcotc, line 97: You need to add this to the top of the file too. It loads the current directory's version of the `datadog` gem to memory:
```ruby
lib = File.expand_path('lib', __dir__)
$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
require 'datadog'
```

```suggestion
  Datadog::Tracing::Contrib::REGISTRY.map{ |i| i.name.to_s }
```

integration_versions.md
```diff
@@ -0,0 +1,52 @@
+# This is a table of supported integration versions generated from gemfiles.
+
+| Integration | Ruby Min |  Ruby Max | JRuby Min | JRuby Max |
+|-------------|----------|-----------|----------|----------|
+| action_cable | 5.2.8.1 | 7.1.3.4 | 5.2.8.1 | 6.1.7.8 |
```
marcotc, line 5: My suggestion for excluded versions (e.g. `!= 3.7.0`) is to add an extra column (Ruby Unsupported & JRuby Unsupported) and add these individually excluded versions in it.

integration_versions.md
```diff
@@ -0,0 +1,52 @@
+# This is a table of supported integration versions generated from gemfiles.
+
+| Integration | Ruby Min |  Ruby Max | JRuby Min | JRuby Max |
```
marcotc, line 3: I think approach in this table answers your question on upper bounds:
> What to do with upper bounds
Given we are reporting the effective `Max` we test againts, we should be already respecting that upper bound. I don't think we need to surface that in the final table.

marcotc:
> Potentially either the existing Update Latest Dependency workflow or a new workflow

I'd say add it to `Update Latest Dependency` for now. It should technically be on every PR that changes dependencies, but would be tricky to coordinate two automated commits at the same time (this and the existing `Lock Dependency` action).

quinna-h:
> > Potentially either the existing Update Latest Dependency workflow or a new workflow
> 
> I'd say add it to `Update Latest Dependency` for now. It should technically be on every PR that changes dependencies, but would be tricky to coordinate two automated commits at the same time (this and the existing `Lock Dependency` action).

I think we are going to end up putting it in a separate workflow, as we want it to be tied to the tracer release process (but for now make it manual, aka triggered on `workflow dispatch`)

integration_versions.md
bouwkast: @marcotc Are there currently versions that need are excluded within supported ranges of gems or is this something that would be a nice to have?

.github/scripts/generate_table_versions.rb
```diff
@@ -0,0 +1,25 @@
...
+
+# Input and output file names
+input_file = 'gem_output.json'
+output_file = 'integration_versions.md'
+
+# Read JSON data from the input file
+data = JSON.parse(File.read(input_file))
+
+# Prepare the Markdown content
+comment = "# This is a table of supported integration versions generated from gemfiles.\n\n"
```
bouwkast, line 11:
```suggestion
comment = "# Integrations\n\n"
```

nitpicky, but maps to the current [Ruby compatiblity](https://docs.datadoghq.com/tracing/trace_collection/compatibility/ruby/#integrations)

.github/scripts/find_gem_version_bounds.rb
```diff
@@ -0,0 +1,133 @@
...
+end
+
+def get_integration_names(directory = 'lib/datadog/tracing/contrib/')
+  Datadog::Tracing::Contrib::REGISTRY.map{ |i| i.name.to_s }
+end
+
+# TODO: The gem information should reside in the integration declaration instead of here.
+
+mapping = {
```
bouwkast, line 93: There are some that I don't see and I don't think have a clear automation path that we could potentially just hardcode in:

`Net/HTTP` and `Makara (via Active Record)` are the only two I see (may have missed one though)


Just went through the list of them here: https://docs.datadoghq.com/tracing/trace_collection/compatibility/ruby/#integrations

bouwkast (approved):
LGTM

We should get a language experts review on this and the output table here to ensure that they are happy with the integrations and associated versions: https://github.com/DataDog/dd-trace-rb/pull/4236/files#diff-f0b9875c282b2a3ae40002cd5ea7085509e744df2ba94d0301f5c0ba313203fd

.github/scripts/find_gem_version_bounds.rb
quinna-h: left comments about these on the generated PR: #4236

.github/workflows/generate-supported-versions.yml
```diff
@@ -0,0 +1,52 @@
+name: "Generate Supported Versions"
+
+on:
+  push:
+    branches:
+      - quinna.halim/add-supported-versions-table
```
quinna-h, line 6: note to self: remove before merge

XLUCARD (approved):

lib/datadog/tracing/contrib/mongodb/integration.rb
```diff
@@ -17,6 +17,9 @@ class Integration
 
           # @public_api Changing the integration name or integration options can cause breaking changes
           register_as :mongo, auto_patch: true
+          def self.gem_name
+            'mongo'
+          end
 
           def self.version
             Gem.loaded_specs['mongo'] && Gem.loaded_specs['mongo'].version
```
ivoanjo, line 25: Minor: Maybe not in scope for this PR, but it seems to me that now that we have a `gem_name`, we could probably implement a default `version` that's just `Gem.loaded_specs[gem_name]&.version` since that seems to cover (almost?) all of them :)

bouwkast: Would the version returned by `Gem.loaded_specs[gem_name]&.version` be a version that is being tested or just a version that is loaded in the environment? If it is pulling the version from the test suite that'd be fine and could be a good improvement for a future PR.

ivoanjo: If I understand your question correctly, `Gem.loaded_specs[gem_name]` is the version that's either loaded OR available to load (it may have not been `require`d yet -- which is why we usually have a separate `loaded?` check for that).

This `version` is usually combined with the `MINIMUM_VERSION` constant to determine the outcome for `compatible?`.

bouwkast: Yeah that answers it thanks (I think 😄)!

The goal that we have with documentation for "supported" integrations cross language is to report the minimum and maximum _tested_ versions only. So I don't think we can safely use this value.

To elaborate for this instance of `mongo`:

We have `MINIMUM_VERSION = Gem::Version.new('2.1.0')`.
If in the integration tests for `mongo` the minimum version tested is `2.5.0` we will report `2.5.0` as the minimum version _supported_ moving forward.

marcotc (approved):

integration_versions.md
marcotc: This was a question @quinna-h asked in the PR description, as an open question to solve.

marcotc (approved):

PR #4221, opened 2024-12-11
Upgrade steep and make it a bit more strict

marcotc:
Upgrade steep to `1.9.1` and rbs to `3.7.0`, as I've [encountered recent steep bugs](https://github.com/DataDog/dd-trace-rb/blob/43498ff628ca2c5b0a852798e1260858416021d7/sig/datadog/tracing/span_event.rbs#L22) that have been fixed in newer versions.

I also played a bit with the default type checking level: we currently use [`default`](https://github.com/soutaro/steep/blob/b8de78bfcbceb4e7b895a2d574ffe178e569e453/lib/steep/diagnostic/ruby.rb#L992) but I've tried `strict` and noticed there are a few very helpful checks.
Most notably:
* `DifferentMethodParameterKind` which checks that your `.rbs` declarations actually match the method parameters.
* `IncompatibleAssignment` which checks if we are assigning incompatible types to fields we've explicitly typed already.

There's one functional change which steep caught in this upgrade: there was one dead code branch in `lib/datadog/appsec/processor/rule_loader.rb`, which I removed. This is actually dead code, since `ip_passlist` comes from a non-nillable configuration option.

**Change log entry**
None


lib/datadog/appsec/processor/rule_loader.rb
```diff
@@ -74,9 +74,6 @@ def passlist_exclusions(ip_passlist) # rubocop:disable Metrics/MethodLength
             when Hash
               pass = ip_passlist[:pass]
               monitor = ip_passlist[:monitor]
-            else
-              pass = []
-              monitor = []
```
marcotc, line 79: This is actually dead code, since `ip_passlist` comes from a non-nillable configuration option: it can be empty, but not `nil`.

sarahchen6 (approved):
👍

Steepfile
```diff
@@ -1,8 +1,48 @@
+# Declare shortcuts for Steep::Signature::Ruby to make this file easier to read
+# as well as facilitating the findability of violation types emitted by the CLI
+# (e.g. the CLI emits `Diagnostic ID: Ruby::UnknownConstant` when finding errors).
+Ruby = Steep::Diagnostic::Ruby
```
Strech, line 4: Not a deal-breaker, but `Diagnostic` might suite better here.

Examples

```ruby
  configure_code_diagnostics(Diagnostic.strict) do |config|
    # ...
    config[Diagnostic::FallbackAny] = :information
  end
```

ruby-3.0.gemfile
```diff
@@ -53,8 +53,8 @@ gem 'webmock', '>= 3.10.0'
 gem 'webrick', '>= 1.7.0'
 
 group :check do
-  gem 'rbs', '~> 3.6', require: false
-  gem 'steep', '~> 1.7.0', require: false
+  gem 'rbs', '~> 3', require: false
+  gem 'steep', '~> 1', require: false
```
Strech, line 57: That type of boundaries a basically - no boundaries, except more or equal version `1` (i.e why `~>`?)

```ruby
gem 'steep', '~> 1'
```

I would rather go with `> 1.7` to highlight that version 1.7 is no fit for us

```suggestion
  gem 'steep', '> 1.7', require: false
```

Strech (approved):
I think my suggestions are non-blocking, but gems versions are kind-a important

ruby-3.0.gemfile
p-datadog: Multiple constraints can be specified, I believe I was the one who typed the `~> 1.7` part and we actually need that minimum version for code in master, therefore I would suggest `~> 1.7, >= 1.8` as the constraint.

ruby-3.0.gemfile
```diff
@@ -53,8 +53,8 @@ gem 'webmock', '>= 3.10.0'
 gem 'webrick', '>= 1.7.0'
 
 group :check do
-  gem 'rbs', '~> 3.6', require: false
-  gem 'steep', '~> 1.7.0', require: false
+  gem 'rbs', '~> 3', require: false
```
p-datadog, line 56: I think this should also be `~> 3.7`.

Steepfile
marcotc: The problem with this one is that it is not `Diagnostic::FallbackAny`, it would be `Diagnostic::Ruby::FallbackAny`. It gets repetitive.

Tbh, this is what Steep does themselves: https://github.com/soutaro/steep/blob/81da4a1172fa61551ec4f0d871fb5ed7cd588d2f/Steepfile#L1

ruby-3.0.gemfile
marcotc: I'll unlock it, tbh, since we commit the lock files and bump them ourselves when needed.

ruby-3.0.gemfile
marcotc: I added multiple constraints, like @p-datadog suggested.

Steepfile
marcotc: I'm ok updating it as a follow up if we feel strongly about it.

PR #4222, opened 2024-12-12
[APPSEC-56128] Replace telemetry error with report

Strech:
**What does this PR do?**

Replace AppSec telemetry error with report upon receiving internal "implicit" error of `libddwaf`.

**Motivation:**

This change is a part-done set of changes in order to have a single monitor to control issues with `libddwaf` calls.

With this change we will be able to see errors both "implicit" and explicit in the same place to build a Monitor around them.

Screenshot 2024-12-12 at 12 58 35
<sup>(example picture for reference - <a href="https://app.datadoghq.com/error-tracking?query=%40lib_language%3Aruby%20%22AppSec%3A%3AWAF%3A%3ALibDDWAF%22&et-viz=sample&fromUser=false&refresh_mode=sliding&source=backend&from_ts=1733912983976&to_ts=1733999383976&live=true">live log</a>)</sup>

We have a task to replace exceptions in `libddwaf-rb` with return status Result (implicit error) which means two things:

1. a single place to record the error - in report execution method
2. a better `Result` object from `libddwaf-rb` (decouple from its origin interface)

**Change log entry**

No.

**Additional Notes:**

In order to deliver log which will appear in ErrorTracking I need a backtrace and a faster backtrace collection appears only in [Ruby 3.4.0](https://github.com/ruby/ruby/pull/10017).

Telemetry to ErrorTracking is not in the perfect state, which means errors will have a generic name instead of a name of exception. But I've added this request to the APM Language team.

**How to test the change?**

CI is enough

### DIFF
--- a/lib/datadog/appsec/processor/context.rb
+++ b/lib/datadog/appsec/processor/context.rb
@@ -79,7 +79,7 @@ module Datadog
           @context.run(persistent_data, ephemeral_data, timeout)
         rescue WAF::LibDDWAF::Error => e
           Datadog.logger.debug { "#{@libddwaf_debug_tag} execution error: #{e} backtrace: #{e.backtrace&.first(3)}" }
-          @telemetry.report(e, description: 'libddwaf-rb internal low-level error')
+          @telemetry.report(e, description: "#{@libddwaf_debug_tag} internal low-level error")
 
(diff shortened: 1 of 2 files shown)


Strech:
After I sleep over with that change, I think it's unnecessary performance impact without actual benefit, hence - closed

PR #4223, opened 2024-12-12
DEBUG-3182 DI: remove hard dependency on tracing

p-datadog:
**What does this PR do?**

Removes hard dependency of DI on Tracing.

- Existing requires of tracing components from DI removed
- DI will check if Datadog::Tracing exists before referencing current_span & current_trace
- Unit tests adjusted to require their dependencies explicitly and to use local CodeTracker instances rather than the global one

**Motivation:**

DI code tracker must be loaded early in application boot process. Tracing must be loaded late to instrument third-party libraries. Currently, DI requires some tracing components which may result in tracing not loading correctly.

**Change log entry**

None - I am currently not aware of specific issues with DI or tracing loading.

**Additional Notes:**

I intend to add a test in a separate PR that loads DI code tracking in a forked process, verifies that the code tracking works correctly, and asserts that none of the other components (most importantly tracing but we could check others too) have been loaded.

**How to test the change?**

I tested the encapsulation manually by removing a bunch of requires pertaining to tracing from spec_helper.rb and other helpers. The main test of the DI code tracker both working correctly and not loading tracing will be in a later PR.

### DIFF
--- a/lib/datadog/di/init.rb
+++ b/lib/datadog/di/init.rb
@@ -4,8 +4,6 @@
 # enable dynamic instrumentation for third-party libraries used by the
 # application.
 
-require_relative '../tracing'
-require_relative '../tracing/contrib'
 require_relative '../di'
 
 # Code tracking is required for line probes to work; see the comments
--- a/lib/datadog/di/probe_notification_builder.rb
+++ b/lib/datadog/di/probe_notification_builder.rb
@@ -141,14 +141,16 @@ module Datadog
             version: 2,
           },
           # TODO add tests that the trace/span id is correctly propagated
-          "dd.trace_id": Datadog::Tracing.active_trace&.id&.to_s,
-          "dd.span_id": Datadog::Tracing.active_span&.id&.to_s,
+          "dd.trace_id": active_trace&.id&.to_s,
+          "dd.span_id": active_span&.id&.to_s,
           ddsource: 'dd_debugger',
           message: probe.template && evaluate_template(probe.template,
             duration: duration ? duration * 1000 : 0),
           timestamp: timestamp,
         }
       end
+
+      private
 
       def build_status(probe, message:, status:)
         {
@@ -198,6 +200,18 @@ module Datadog
         binding.local_variables.each_with_object({}) do |name, map|
           value = binding.local_variable_get(name)
           map[name] = value
+        end
+      end
+
+      def active_trace
+        if defined?(Datadog::Tracing)
+          Datadog::Tracing.active_trace
+        end
+      end
+
+      def active_span
+        if defined?(Datadog::Tracing)
+          Datadog::Tracing.active_span
         end
       end
     end
--- a/lib/datadog/di/component.rb
+++ b/lib/datadog/di/component.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative '../core'
+
 module Datadog
   module DI
     # Component for dynamic instrumentation.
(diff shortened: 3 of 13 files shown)


anmarchenko (approved):

lib/datadog/di/probe_notification_builder.rb
```diff
@@ -200,6 +202,18 @@ def get_local_variables(trace_point)
...
+        if defined?(Datadog::Tracing)
+          Datadog::Tracing.active_trace
+        end
+      end
+
+      def active_span
+        if defined?(Datadog::Tracing)
+          Datadog::Tracing.active_span
+        end
+      end
```
ivoanjo, line 216: Not sure what's the overhead trade-off (maybe it's very minor), but the `defined?(Datadog::Tracing)` check could be made inside the component (or init) and cached.

This is because anyway any change to load/enable a product means that the components will get rebuilt, so you'll always be able to check once and rely on that check from them on.

lib/datadog/di/init.rb
```diff
@@ -4,8 +4,6 @@
 # enable dynamic instrumentation for third-party libraries used by the
 # application.
 
-require_relative '../tracing'
-require_relative '../tracing/contrib'
```
marcotc, line 8: 🥳

marcotc (approved):
I agree with @ivoanjo's point here, but overall awesome work! :shipit:

lib/datadog/di/probe_notification_builder.rb
p-datadog: I wrote a benchmark comparing a `defined?` call with instance variable retrieval (`@var`) and the `defined?` call is 66% of IPS of the instance variable retrieval, which I think is not a significant performance penalty.

I can make notification builder take a flag indicating whether to use tracing or not, which will be passed from a higher level class after it checks whether tracing is defined, for the purpose of explicitly controlling tracer usage within the notification builder (with the downside of needing to re-instantiate the notification builder when tracing is loaded, which practically I don't think will ever happen). But making this change for the purpose of optimization I think is not worthwhile.

What do you think? @marcotc @ivoanjo

ivoanjo: Thanks for checking... I think your findings support the `defined` version in terms of overhead.

This did get me thinking: I think in almost all cases the tracing code will be loaded, so if you are targeting customers that disable tracing, you may want to detect if tracing is enabled or not, rather than loaded or not.

p-datadog: From what I understand, in practice every DI customer will be using tracing. I do not intend to specifically invest effort into supporting Di-without-tracing scenario, the only important part is I need to not load tracing earlier than it otherwise would be loaded due to integrations.

ivoanjo: In that case... I'm actually not sure this change is needed at all. Specifically, I understand the changes regarding the early boot of the library, but the `ProbeNotificationBuilder` only gets called via the components, and well, if the components are running, then tracing has already been required.

So I suspect this part of the change is a bit of a no-op?

p-datadog: I don't want to assume that tracing exists if I (in DI) never required it (assuming that the library overall loaded tracing after it loaded DI).

Ideally the DI unit tests won't be loading tracing also (currently they do by virtue of loading the global spec helper).

ivoanjo: Yeah, I think that's OK, just wanted to clarify this bit since it seemed relevant context to how overall the components and whatnot works today :)

PR #4230, opened 2024-12-16
DEBUG-3210 DI: change logging to be appropriate for customer inspection

p-datadog:
**What does this PR do?**

Failure to enable DI is now a warning (since enabling it must have been requested by the customer).

Remaining logging calls are now emitted at debug level by default, making those entries not visible in customer logs.

An internal DI setting is added to reinstate logging of internal issues at warn level, for developing/debugging DI.

**Motivation:**

DI currently uses the logger for reporting its internal issues. However, the logger normally logs to customer logs, and as such the default logging should be customer-centric, meaning log entries should only be created for matters that the customers understand and can fix.

**Change log entry**

Dynamic instrumentation: logging of internal conditions is now done on debug level, and these log entries will not be visible by default in customer applications.

**Additional Notes:**

**How to test the change?**

Due to the introduction of an additional layer of abstraction for DI logging, end-to-end tests will need to have additional assertions added for log entries being visible (or not). This will be done as part of DEBUG-3210 but is not part of this PR.

### DIFF
--- a/lib/datadog/di/component.rb
+++ b/lib/datadog/di/component.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative 'logging'
+
 module Datadog
   module DI
     # Component for dynamic instrumentation.
@@ -13,12 +15,14 @@ module Datadog
     # intalled tracepoints and so on. Component will clean up all
     # resources and installed tracepoints upon shutdown.
     class Component
+      include Logging
+
       class << self
         def build(settings, agent_settings, telemetry: nil)
           return unless settings.respond_to?(:dynamic_instrumentation) && settings.dynamic_instrumentation.enabled
 
           unless settings.respond_to?(:remote) && settings.remote.enabled
-            Datadog.logger.debug("Dynamic Instrumentation could not be enabled because Remote Configuration Management is not available. To enable Remote Configuration, see https://docs.datadoghq.com/agent/remote_config")
+            Datadog.logger.warn("Dynamic Instrumentation could not be enabled because Remote Configuration Management is not available. To enable Remote Configuration, see https://docs.datadoghq.com/agent/remote_config")
             return
           end
 
@@ -53,12 +57,12 @@ module Datadog
           # TODO add tests?
           unless settings.dynamic_instrumentation.internal.development
             if Datadog::Core::Environment::Execution.development?
-              Datadog.logger.debug("Not enabling dynamic instrumentation because we are in development environment")
+              Datadog.logger.warn("Not enabling dynamic instrumentation because we are in development environment")
               return false
             end
           end
           if RUBY_ENGINE != 'ruby' || RUBY_VERSION < '2.6'
-            Datadog.logger.debug("Not enabling dynamic instrumentation because of unsupported Ruby version")
+            Datadog.logger.warn("Not enabling dynamic instrumentation because of unsupported Ruby version")
             return false
           end
           true
--- a/lib/datadog/di/instrumenter.rb
+++ b/lib/datadog/di/instrumenter.rb
@@ -3,6 +3,7 @@
 # rubocop:disable Lint/AssignmentInCondition
 
 require 'benchmark'
+require_relative 'logging'
 
 module Datadog
   module DI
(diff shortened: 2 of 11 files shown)


lib/datadog/di/logging.rb
```diff
@@ -0,0 +1,30 @@
...
+      end
+      
+      # Logs an internal informational message.
+      # When verbose logging is enabled, the logging happens at info level.
+      # When verbose logging is disabled, nothing is logged.
+      def log_info_internal(msg)
+        if settings.dynamic_instrumentation.internal.verbose_logging
+          logger.info(msg)
+        end
+      end
```
ivoanjo, line 27: Minor: This is a bit unexpected to me; why not log it at debug level as well? E.g. it seems weird that if I enable debug logging with `DD_TRACE_DEBUG=true`, there's still some logging that won't show up?

lib/datadog/di/component.rb
```diff
@@ -53,12 +57,12 @@ def environment_supported?(settings)
           # TODO add tests?
           unless settings.dynamic_instrumentation.internal.development
             if Datadog::Core::Environment::Execution.development?
-              Datadog.logger.debug("Not enabling dynamic instrumentation because we are in development environment")
+              Datadog.logger.warn("Not enabling dynamic instrumentation because we are in development environment")
```
ivoanjo, line 60: Minor: Since this is a customer-visible string, I'd suggest not using the "we" since it's a bit confusing. Here's a potential alternative

```suggestion
              Datadog.logger.warn("Development environment detected; not enabling dynamic instrumentation")
```

lib/datadog/di/component.rb
```diff
@@ -53,12 +57,12 @@ def environment_supported?(settings)
...
               return false
             end
           end
           if RUBY_ENGINE != 'ruby' || RUBY_VERSION < '2.6'
-            Datadog.logger.debug("Not enabling dynamic instrumentation because of unsupported Ruby version")
+            Datadog.logger.warn("Not enabling dynamic instrumentation because of unsupported Ruby version")
```
ivoanjo, line 65: For this one, I suggest being a bit more actionable, e.g. 

```suggestion
            Datadog.logger.warn("Cannot enable dynamic instrumentation: Ruby 2.6+ is required")
```

lib/datadog/di/instrumenter.rb
```diff
@@ -305,13 +308,13 @@ def hook_line(probe, &block)
...
-            logger.warn("Unhandled exception in line trace point: #{exc.class}: #{exc}")
+            log_warn_internal("Unhandled exception in line trace point: #{exc.class}: #{exc}")
             telemetry&.report(exc, description: "Unhandled exception in line trace point")
             # TODO test this path
           end
         rescue => exc
           raise if settings.dynamic_instrumentation.internal.propagate_all_exceptions
-          logger.warn("Unhandled exception in line trace point: #{exc.class}: #{exc}")
+          log_warn_internal("Unhandled exception in line trace point: #{exc.class}: #{exc}")
           telemetry&.report(exc, description: "Unhandled exception in line trace point")
```
ivoanjo, line 318: You may want to expose a block-based version of these methods. E.g. right now the whole string is built, even if the logger is going to throw it away.

In other parts of the codebase we often do `Datadog.logger.debug do ... end` or similar, and that version tends to be cheaper since debug logging is often off so it nopes out without building the string. (This is a common trick for logging libraries)

lib/datadog/di/logging.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Datadog
+  module DI
+    # Logging for DI
+    module Logging
```
ivoanjo, line 6: Minor: Since every class that uses `Logging` ends up needing to get passed a `logger` object anyway, I wonder if it would make sense to turn this module into an actual class that would have a `logger` as an instance variable, and then you'd always call methods on this class instead of the logger directly.

This would also solve the potential (small) sharp edge of "oops I included Logging but this class doesn't already have a `logger`" -- since you know that if you have the object, then you have all that you need.

ivoanjo (commented):
Looks reasonable -- left a few notes!

p-datadog:
Replaced by https://github.com/DataDog/dd-trace-rb/pull/4266

PR #4232, opened 2024-12-17
[PROF-11045] Fix profiling warnings being really hard to silence

### LINKED ISSUE #4231: i cannot stop this message: "WARN -- datadog: [datadog] In all known versions of Ruby 3.x]...
**Current behaviour**
we have a rails app running in ecs. we run in docker container with the dd sidecar.

since we enabled allocation profiling this message has been printing to `STDOUT` multiple times any time we run any sort of `rails` command:
```
# bin/rails console
W, [2024-12-16T21:18:33.543770 #1526]  WARN -- datadog: [datadog] In all known versions of Ruby 3.x, using Ractors may result in allocation profiling unexpectedly stopping (https://bugs.ruby-lang.org/issues/19112). Note that this stop has no impact in your application stability or performance. This does not happen if Ractors are not used.
W, [2024-12-16T21:18:33.923679 #1526]  WARN -- datadog: [datadog] In all known versions of Ruby 3.x, using Ractors may result in allocation profiling unexpectedly stopping (https://bugs.ruby-lang.org/issues/19112). Note that this stop has no impact in your application stability or performance. This does not happen if Ractors are not used.
W, [2024-12-16T21:18:33.975804 #1526]  WARN -- datadog: [datadog] In all known versions of Ruby 3.x, using Ractors may result in allocation profiling unexpectedly stopping (https://bugs.ruby-lang.org/issues/19112). Note that this stop has no impact in your application stability or performance. This does not happen if Ractors are not used.
Loading sandbox environment (Rails 8.0.1)
rails-app(sandbox)>
```

the logging comes from https://github.com/DataDog/dd-trace-rb/blob/4f1c8bd8f4674d46555a79b7c90046ec597ee660/lib/datadog/profiling/component.rb#L192

note that the above log messages are not going to the datadog or rails `logger` because they're not properly formatted (all of our logs are json formatted, including the datadog logs). 

we've tried setting:
 * `DD_TRACE_STARTUP_LOGS=false` in the datadog agent container and our application container
 * `DD_LOG_LEVEL=error` in the datadog agent container and our application container
 * `c.logger.level = Logger::ERROR` in the initializer

but nothing changes. 

this is breaking some of our automation. short of disabling this functionality (which is what we're going to do now) is there anything we can do to quiet this?

**Expected behaviour**
there should be a way to disable these

**Steps to reproduce**

**Environment**

* **datadog version:** datadog (2.8.0)
```ruby
  # see https://github.com/DataDog/dd-trace-rb/issues/2012
  Datadog.configuration.tracing.log_injection = false

  Datadog.configure do |c|
    c.tracing.enabled = true
    # c.profiling.enabled = true

    c.tracing.instrument :action_mailer
    c.tracing.instrument :action_cable
    c.tracing.instrument :rails
    c.tracing.instrument :faraday, split_by_domain: true

    c.logger.level = Logger::ERROR
  end
```

```
DD_PROFILING_ALLOCATION_ENABLED=true
DD_GIT_REPOSITORY_URL=a-git-repo-url
DD_PROFILING_PREVIEW_GVL_ENABLED=true
DD_PROFILING_ENABLED=true
DD_LOG_LEVEL=ERROR
DD_ENV=sandbox
DD_VERSION=a-git-sha
DD_GIT_COMMIT_SHA=a-git-sha
DD_TRACE_STARTUP_LOGS=false
```
* **Ruby version:**
```
ruby 3.3.6 (2024-11-05 revision 75015d4c1f) [arm64-darwin24]
```

* **Operating system:**
```
Linux ip-10-0-87-157.ec2.internal 5.10.228-219.884.amzn2.x86_64 #1 SMP Wed Oct 23 17:17:00 UTC 2024 x86_64 GNU/Linux
```

ivoanjo:
**What does this PR do?**

This PR fixes the report we had in #4231 that the profiling initialization warnings were hard to silence.

In particular, due to the way the profiling component was accessing the dd-trace-rb logger, any configuration changes to change the log level were actually not visible to the profiling component. This issue is now fixed, and enabling profiling + changing the log level in the same `Datadog.configure` block will be respected.

I also included a few more improvements:
* Wrapped the message from #4231 in an `OnlyOnce` so even when it does get logged, it will only get logged at most once
* Lowered the log level of some of our log messages
* Simplified/cleaned up a few of the messages

**Motivation:**

Fixes #4231

**Change log entry**

Yes. Fix profiling warnings being really hard to silence

**Additional Notes:**

There's still one case where it's not possible to avoid the warning: when starting profiling via an environment variable (e.g. `DD_PROFILING_ENABLED=true bundle exec ddprofrb ruby <somescript>`).

This is a result of there not being a way to set the log level of dd-trace-rb via an environment variable. I decided to not tackle this one for now; hopefully the fixes above are enough of an improvement to the status quo ;)

**How to test the change?**

Here's two sample test cases:

1. `DD_PROFILING_ENABLED=true DD_PROFILING_ALLOCATION_ENABLED=true bundle exec ddprofrb exec ruby -e "Datadog.configure { |c| c.logger.level = Logger::ERROR; c.profiling.enabled = true };"`

With this PR prints the "using Ractors may result in allocation profiling stopping" only once, whereas in master it prints it multiple times

2. `DD_PROFILING_ALLOCATION_ENABLED=true bundle exec ruby -e "require 'datadog/auto_instrument'; Datadog.configure { |c| c.logger.level = Logger::ERROR; c.profiling.enabled = true }"`

With this PR does not print the message; whereas in master it prints it once.

### DIFF
--- a/spec/datadog/profiling/component_spec.rb
+++ b/spec/datadog/profiling/component_spec.rb
@@ -360,6 +360,7 @@ RSpec.describe Datadog::Profiling::Component do
                   .with(hash_including(heap_samples_enabled: true, heap_size_enabled: false))
                   .and_call_original
 
+                expect(Datadog.logger).to receive(:info).with(/Ractors.+stopping/)
                 expect(Datadog.logger).to receive(:debug).with(/Enabled allocation profiling/)
                 expect(Datadog.logger).to receive(:warn).with(/experimental heap profiling/)
                 expect(Datadog.logger).not_to receive(:warn).with(/experimental heap size profiling/)
--- a/spec/datadog/profiling/component_spec.rb
+++ b/spec/datadog/profiling/component_spec.rb
@@ -203,6 +203,8 @@ RSpec.describe Datadog::Profiling::Component do
             settings.profiling.allocation_enabled = true
             settings.profiling.advanced.gc_enabled = false # Disable this to avoid any additional warnings coming from it
             stub_const("RUBY_VERSION", testing_version)
+
+            described_class.const_get(:ALLOCATION_WITH_RACTORS_ONLY_ONCE).send(:reset_ran_once_state_for_tests)
           end
 
           context "on Ruby 2.x" do
@@ -337,6 +339,8 @@ RSpec.describe Datadog::Profiling::Component do
           context "and allocation profiling enabled and supported" do
             before do
               settings.profiling.allocation_enabled = true
+
+              described_class.const_get(:ALLOCATION_WITH_RACTORS_ONLY_ONCE).send(:reset_ran_once_state_for_tests)
             end
 
             it "initializes StackRecorder with heap sampling support and warns" do
--- a/spec/datadog/profiling/component_spec.rb
+++ b/spec/datadog/profiling/component_spec.rb
@@ -272,7 +272,7 @@ RSpec.describe Datadog::Profiling::Component do
                   .with(hash_including(alloc_samples_enabled: true))
                   .and_call_original
 
-                expect(Datadog.logger).to receive(:warn).with(/Ractors.+stopping/)
+                expect(Datadog.logger).to receive(:info).with(/Ractors.+stopping/)
                 expect(Datadog.logger).to receive(:debug).with(/Enabled allocation profiling/)
 
                 build_profiler_component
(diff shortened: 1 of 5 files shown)


spec/datadog/profiling/component_spec.rb
```diff
@@ -1006,13 +1013,8 @@
...
-      end
 
-      it "logs an error message mentioning that the invalid value will be ignored" do
-        expect(Datadog.logger).to receive(:error).with(/Ignoring invalid value/)
-
-        no_signals_workaround_enabled?
+        expect(logger).to receive(:warn).with(/Ignoring invalid value/)
       end
 
       include_examples "no_signals_workaround_enabled :auto behavior"
```
ivoanjo, line 1020: I chose to simplify this a bit because the change from logging this info at error to warn was also interacting with a few of the specs from the `include_examples`, so I chose to just call the `expect` in the before rather than doing something weirder (or copy-pasting the specs to change them slightly).

ivoanjo:
> ⌛ 5 Performance Regressions

These regressions are false positives, I'll see what I can do about updating our config to correct this.

p-datadog (approved):

PR #4234, opened 2024-12-17
K8s lib injection tests new scenarios

robertomonteromiguel:
**What does this PR do?**

Adapt the gitlab pipeline to add the new k8s lib injection tests scenarios

**Motivation:**

We need to ensure the scalability of the system-tests k8s lib injection tests.
We added the new scenarios on system-tests. We changed the one pipeline and now on the tracer we need to create the test matrix with the new scenarios

**Change log entry**
No

**Additional Notes:**

**How to test the change?**


.gitlab-ci.yml
```diff
@@ -131,7 +131,9 @@ onboarding_tests_installer:
 onboarding_tests_k8s_injection:
   parallel:
     matrix:
-      - WEBLOG_VARIANT: [dd-lib-ruby-init-test-rails, dd-lib-ruby-init-test-rails-explicit,dd-lib-ruby-init-test-rails-gemsrb]
+    - WEBLOG_VARIANT: [dd-lib-ruby-init-test-rails, dd-lib-ruby-init-test-rails-explicit,dd-lib-ruby-init-test-rails-gemsrb]
+      SCENARIO: [K8S_LIB_INJECTION, K8S_LIB_INJECTION_UDS, K8S_LIB_INJECTION_NO_AC, K8S_LIB_INJECTION_NO_AC_UDS, K8S_LIB_INJECTION_PROFILING_DISABLED, K8S_LIB_INJECTION_PROFILING_ENABLED, K8S_LIB_INJECTION_PROFILING_OVERRIDE]
```
p-datadog, line 135: Does this add the specified scenarios or replace the scenarios that were being executed previously?

robertomonteromiguel: A few weeks ago we were executing all the tests under the same scenario. Now we split the scenarios, and we trust on gitlab runner to execute them in parallel

p-datadog (approved):
I am not familiar with the tests in question, therefore I can add my checkmark on the basis of this PR not obviously affecting the rest of dd-trace-rb in a negative way but I cannot tell whether the changes do test everything correctly.

The tests also look like they are stuck.

robertomonteromiguel:
> The tests also look like they are stuck.

https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-rb/-/pipelines/52523725

I can't merge the PR: "The base branch restricts merging to authorized users. "

ivoanjo:
These 3 tests seem to be completely stuck and I've not seen this in master so they do seem to be caused by this PR:

> image

@robertomonteromiguel can you check what's up there?

robertomonteromiguel:
> These 3 tests seem to be completely stuck and I've not seen this in master so they do seem to be caused by this PR:
> 
> > image
> 
> @robertomonteromiguel can you check what's up there?

@ivoanjo I can't access to your repository settings, but I seems that there are specific requirements for merge a PR. These requirements are pointed to the jobs that not longer exists. You need to change them to the new scenarios/jobs on ruby. Ping me if you need help to do that.

robertomonteromiguel:
> > These 3 tests seem to be completely stuck and I've not seen this in master so they do seem to be caused by this PR:
> > > image
> > 
> > 
> > @robertomonteromiguel can you check what's up there?
> 
> @ivoanjo I can't access to your repository settings, but I seems that there are specific requirements for merge a PR. These requirements are pointed to the jobs that not longer exists. You need to change them to the new scenarios/jobs on ruby. Ping me if you need help to do that.

Here an example of system-tests repository requirements:

ivoanjo:
> @ivoanjo I can't access to your repository settings, but I seems that there are specific requirements for merge a PR. These requirements are pointed to the jobs that not longer exists. You need to change them to the new scenarios/jobs on ruby. Ping me if you need help to do that.

Ahh right, this has happened to me before -- github couples on the exact names of the tests and so if we change it a bit (like we're doing here) then GitHub gets really confused.

Two notes:

1. Since now we have many separate tests, do you have any recommendations on making sure that they are all green? E.g. is there a single step we can rely on that runs after all these or something like that?

> image

2. I've gone ahead and removed these tests from the require list to unblock this PR:

* dd-gitlab/onboarding_tests_k8s_injection: [dd-lib-ruby-init-test-rails-explicit]
* dd-gitlab/onboarding_tests_k8s_injection: [dd-lib-ruby-init-test-rails-gemsrb]
* dd-gitlab/onboarding_tests_k8s_injection: [dd-lib-ruby-init-test-rails]

robertomonteromiguel:
> 1. we can rely on that runs after all the

I still don't have a clear idea of how to manage this. The same thing happens with other tracers and with the other onboarding tests.
I had the same problem a couple of months ago, with other tests in another repository. 
It is painful to rename a test scenario or create a new one. Since you have to add it in all tracers, and also these checks are different in each repository, and I don't have permissions to modify these checks in the repositories either.
Let me discuss it with the team, to try to come up with a definitive solution that applies to all tracers.

ivoanjo:
> Let me discuss it with the team, to try to come up with a definitive solution that applies to all tracers.

Awesome, thank you, looking forward to it :)

PR #4236, opened 2024-12-18
[🤖] Update Supported Versions

TonyCTHsu:
This is a PR to update the table for supported integration versions.
The supported versions markdown is generated from the minimum and maximum tested versions of each integration,
as defined from the `gemfile.lock` gem declarations.

Workflow run: [Generate Supported Versions](https://github.com/DataDog/dd-trace-rb/actions/workflows/generate-supported-versions.yml)

This should be tied to tracer releases, or triggered manually.


integration_versions.md
```diff
@@ -0,0 +1,52 @@
...
+| delayed_job | 4.1.11 | 4.1.13 | 4.1.11 | 4.1.11 |
+| elasticsearch | 7.17.11 | 8.17.0 | 7.17.11 | 8.17.0 |
+| ethon | 0.16.0 | 0.16.0 | 0.14.0 | 0.16.0 |
+| excon | 0.102.0 | 0.110.0 | 0.102.0 | 0.102.0 |
+| faraday | 0.17.0 | 2.12.2 | 0.17.0 | 2.12.2 |
+| grape | 1.7.0 | 2.1.2 | 1.7.0 | 1.8.0 |
+| graphql | 1.13.21 | 2.3.7 | 1.13.21 | 2.3.6 |
+| grpc | 1.48.0 | 1.67.0 | None | None |
+| hanami | 1.3.5 | 1.3.5 | None | None |
+| http | 5.0.1 | 5.2.0 | 4.4.1 | 4.4.1 |
```
quinna-h, line 25: this is `net/http` in the curent docs: https://docs.datadoghq.com/tracing/trace_collection/compatibility/ruby/#integrations

integration_versions.md
```diff
@@ -0,0 +1,52 @@
...
+| httpclient | 2.8.3 | 2.8.3 | 2.8.3 | 2.8.3 |
+| httprb | 5.0.1 | 5.2.0 | 4.4.1 | 4.4.1 |
```
quinna-h, line 27: missing `httpx`

integration_versions.md
```diff
@@ -0,0 +1,52 @@
...
+| kafka | 1.5.0 | 1.5.0 | 1.5.0 | 1.5.0 |
+| lograge | 0.12.0 | 0.14.0 | 0.12.0 | 0.14.0 |
```
quinna-h, line 29: missing `makara` because it is not its own integration, but part of `active_record`
maybe in the `contrib/active_record/integration.rb` need to include `self.gems`

gem_output.json
```diff
@@ -0,0 +1,302 @@
...
+    "1.3.5",
+    null,
+    null
+  ],
+  "http": [
+    "5.0.1",
+    "5.2.0",
+    "4.4.1",
+    "4.4.1"
+  ],
```
bouwkast, line 127: I am unsure if this is correct - I _think_ it is supposed to be `Net::HTTP` which is a library built into Ruby whereas `http.rb` is an actual Ruby gem.

So I think this one we may need to treat differently as it _should_ map to the tested Ruby versions of `Net::HTTP`

Thoughts?

bouwkast (commented):
Just a question on `Net::HTTP` as I think the versions we're reporting are actually `http.rb`

integration_versions.md
```diff
@@ -0,0 +1,54 @@
...
+| faraday | 0.17.0 | 2.12.2 | 0.17.0 | 2.12.2 |
+| grape | 1.7.0 | 2.1.2 | 1.7.0 | 1.8.0 |
+| graphql | 1.13.21 | 2.3.7 | 1.13.21 | 2.3.6 |
+| grpc | 1.48.0 | 1.67.0 | None | None |
+| hanami | 1.3.5 | 1.3.5 | None | None |
+| http | 5.0.1 | 5.2.0 | 4.4.1 | 4.4.1 |
+| httpclient | 2.8.3 | 2.8.3 | 2.8.3 | 2.8.3 |
+| httprb | 5.0.1 | 5.2.0 | 4.4.1 | 4.4.1 |
+| httpx | 0.11 | 0.11 | None | None |
+| kafka | 1.5.0 | 1.5.0 | 1.5.0 | 1.5.0 |
```
quinna-h, line 29: this should be `ruby-kafka`

integration_versions.md
```diff
@@ -0,0 +1,54 @@
...
+| lograge | 0.12.0 | 0.14.0 | 0.12.0 | 0.14.0 |
```
quinna-h, line 30: This isn't in the docs, but we do have it as a contrib

integration_versions.md
```diff
@@ -0,0 +1,54 @@
...
+| dalli | 2.7.11 | 3.2.8 | 2.7.11 | 3.2.3 |
+| delayed_job | 4.1.11 | 4.1.13 | 4.1.11 | 4.1.11 |
+| elasticsearch | 7.17.11 | 8.17.0 | 7.17.11 | 8.17.0 |
+| ethon | 0.16.0 | 0.16.0 | 0.14.0 | 0.16.0 |
+| excon | 0.102.0 | 0.110.0 | 0.102.0 | 0.102.0 |
+| faraday | 0.17.0 | 2.12.2 | 0.17.0 | 2.12.2 |
+| grape | 1.7.0 | 2.1.2 | 1.7.0 | 1.8.0 |
+| graphql | 1.13.21 | 2.3.7 | 1.13.21 | 2.3.6 |
+| grpc | 1.48.0 | 1.67.0 | None | None |
+| hanami | 1.3.5 | 1.3.5 | None | None |
```
quinna-h, line 24: check this for jruby

gem_output.json
bouwkast: Ahh okay, so we essentially then have `http` and `http.rb` duplicated and are missing `net-http` in the table thanks!

integration_versions.md
quinna-h: Checked: did not find `hanami` defined in any `jruby` gemfiles.

bouwkast: Well I guess that is for no tested versions right?

Does it "work" for JRuby just lacks any tests?

integration_versions.md
```diff
@@ -0,0 +1,54 @@
...
+| delayed_job | 4.1.11 | latest | 4.1.11 | latest |
+| elasticsearch | 7.17.11 | latest | 7.17.11 | latest |
+| ethon | 0.16.0 | latest | 0.14.0 | latest |
+| excon | 0.102.0 | latest | 0.102.0 | latest |
+| faraday | 0.17.0 | latest | 0.17.0 | latest |
+| grape | 1.7.0 | latest | 1.7.0 | latest |
+| graphql | 1.13.21 | 2.3.7 | 1.13.21 | 2.3.6 |
+| grpc | 1.48.0 | latest | None | None |
+| hanami | 1.3.5 | 1.3.5 | None | None |
+| http | 5.0.1 | latest | 4.4.1 | 4.4.1 |
```
quinna-h, line 25: check this, our docs say `>= 2.0`

integration_versions.md
quinna-h: note: in the compatibility docs we state that `kafka >= 0.7.10`

In the gemfile we have `>= 0.7.10` https://github.com/DataDog/dd-trace-rb/blob/7fa295e7ddd71202bae53123b9beb65d358c1d29/gemfiles/ruby_3.4_activesupport.gemfile#L39
but in the `Gemfile.lock` we have `1.5.0` https://github.com/DataDog/dd-trace-rb/blob/7fa295e7ddd71202bae53123b9beb65d358c1d29/gemfiles/ruby_3.4_activesupport.gemfile.lock#L211

integration_versions.md
```diff
@@ -0,0 +1,54 @@
...
+| httpclient | 2.8.3 | latest | 2.8.3 | latest |
+| httprb | 5.0.1 | latest | 4.4.1 | 4.4.1 |
+| httpx | 0.11 | None | 0.11 | None |
+| kafka | 1.5.0 | 1.5.0 | 1.5.0 | 1.5.0 |
+| lograge | 0.12.0 | latest | 0.12.0 | latest |
+| makara | 0.3.5 | None | 0.3.5 | None |
+| mongo | 2.14.1 | 2.14.1 | 2.14.1 | 2.14.1 |
```
quinna-h, line 32: This is the version stated in the `gemfile.lock` file, but in the `gemfile` it states
https://github.com/DataDog/dd-trace-rb/blob/7fa295e7ddd71202bae53123b9beb65d358c1d29/gemfiles/ruby_3.4_contrib.gemfile#L34

integration_versions.md
```diff
@@ -0,0 +1,54 @@
...
+| mysql2 | 0.5.5 | 0.5.6 | None | None |
+| opensearch | 2.1.0 | latest | 2.1.0 | latest |
+| pg | 0.21.0 | latest | None | None |
+| presto | 0.6.6 | 0.6.6 | 0.6.6 | 0.6.6 |
```
quinna-h, line 36: similarly, `gemfile.lock` states: https://github.com/DataDog/dd-trace-rb/blob/7fa295e7ddd71202bae53123b9beb65d358c1d29/gemfiles/jruby_9.2_contrib_old.gemfile#L28

integration_versions.md
```diff
@@ -0,0 +1,54 @@
...
+| que | 1.4.1 | 2.3.0 | 1.4.1 | 2.2.0 |
+| racecar | 2.6.0 | 2.11.0 | 2.6.0 | 2.8.2 |
+| rack | 1.6.13 | latest | 1.6.13 | latest |
+| rails | 4.2.11.3 | latest | 5.2.8.1 | 6.1.7.8 |
+| rake | 12.3.3 | 13.2.1 | 12.3.3 | 13.2.1 |
```
quinna-h, line 41: Same here: https://github.com/DataDog/dd-trace-rb/blob/7fa295e7ddd71202bae53123b9beb65d358c1d29/gemfiles/jruby_9.2_contrib_old.gemfile#L16

integration_versions.md
```diff
@@ -0,0 +1,54 @@
...
+| redis | 3.3.5 | 5.2.0 | 3.3.5 | 5.0.6 |
+| resque | 2.4.0 | latest | 2.4.0 | latest |
+| rest_client | 2.1.0 | latest | 2.1.0 | latest |
+| roda | 3.65.0 | 3.81.0 | 3.64.0 | 3.72.0 |
+| semantic_logger | 4.12.0 | 4.15.0 | 4.12.0 | 4.14.0 |
```
quinna-h, line 46: Do we need this? Not in the docs

integration_versions.md
```diff
@@ -0,0 +1,54 @@
...
+| makara | 0.3.5 | None | 0.3.5 | None |
+| mongo | 2.14.1 | 2.14.1 | 2.14.1 | 2.14.1 |
+| mysql2 | 0.5.5 | 0.5.6 | None | None |
+| opensearch | 2.1.0 | latest | 2.1.0 | latest |
+| pg | 0.21.0 | latest | None | None |
+| presto | 0.6.6 | 0.6.6 | 0.6.6 | 0.6.6 |
+| que | 1.4.1 | 2.3.0 | 1.4.1 | 2.2.0 |
+| racecar | 2.6.0 | 2.11.0 | 2.6.0 | 2.8.2 |
+| rack | 1.6.13 | latest | 1.6.13 | latest |
+| rails | 4.2.11.3 | latest | 5.2.8.1 | 6.1.7.8 |
```
quinna-h, line 40: Similarly here, we should check rails support. The docs say `>= 4.0` for both ruby and jruby.

integration_versions.md
marcotc: if it's not in the jruby gemfile locks, then we can set it to no support right now.

integration_versions.md
marcotc: this is good, it's a real ruby gem we support for trace->log correlation.
we probably forgot to add it here and you are now improving it :)

marcotc (approved):

p-datadog (changes requested):
Both of the added files are unclear as to their contents, purpose and how they come to be:

`gem_output.json` provides no hints/explanation as to what it contains or where it came from.

The markdown file, if it's meant for consumption by humans, I think should have some prose in it describing its contents. If it's not meant for consumption by humans, perhaps it should be in some other format.

gem_output.json
```diff
@@ -0,0 +1,302 @@
+{
```
marcotc, line 1: Can we move `gem_output.json` and `integration_versions.md` to the `docs` folder?

integration_versions.md
```diff
@@ -0,0 +1,54 @@
+# Integrations
```
marcotc, line 1: Can we add a markdown comment to the top of the file saying it was generated from `.github/scripts/generate_table_versions.rb`?

gem_output.json
bouwkast: Yeah we just need to be cognizant of not pinging the docs team as a reviewer (yet) as we want to ensure that the language we are presenting is consistent with what we and product want.
Looking at CODEOWNERS we should be able to move these to `docs` without pinging them, so that sounds good to me.

bouwkast:
> Both of the added files are unclear as to their contents, purpose and how they come to be:
> 
> `gem_output.json` provides no hints/explanation as to what it contains or where it came from.
> 
> The markdown file, if it's meant for consumption by humans, I think should have some prose in it describing its contents. If it's not meant for consumption by humans, perhaps it should be in some other format.

@p-datadog thanks!
For how they came to be the workflow run that generates the files is in the PR description, but maybe it'd be clearer to go with @marcotc's suggestion to add a comment at the top of both files about _what_ generates them to make it more obvious.

`gem_output.json` is the "machine"-readable format of the markdown table, _maybe_ it isn't entirely necessary @quinna-h? The level of care though is low there in my opinion, but we've made a machine-readable format of this in other repositories to allow different output formats and allow for some querying.

The markdown file is definitely meant for human consumption. What other format would work better? Note that at the end of the quarter these will be sourced onto the public documentation replacing this table: https://docs.datadoghq.com/tracing/trace_collection/compatibility/ruby#integrations

Note that we will still need to (likely hard code) the configuration options over.

@quinna-h could we add a sentence at the top:

`For a list of available integrations, and their supported version ranges, refer to the following:`

marcotc:
> The markdown file is definitely meant for human consumption. What other format would work better? Note that at the end of the quarter these will be sourced onto the public documentation replacing this table: [docs.datadoghq.com/tracing/trace_collection/compatibility/ruby#integrations](https://docs.datadoghq.com/tracing/trace_collection/compatibility/ruby#integrations)

The public docs page https://docs.datadoghq.com/tracing/trace_collection/compatibility/ruby/ is sourced from https://github.com/DataDog/dd-trace-rb/blob/fe9272b14743e9af065f71cd9ef71181f1086d89/docs/Compatibility.md.
One way we accomplish the goal to have it published by the end of the quarter is to merge the result from `docs/integration_versions.md` into the `# Integrations` portion of `Compatibility.md`:
https://github.com/DataDog/dd-trace-rb/blob/fe9272b14743e9af065f71cd9ef71181f1086d89/docs/Compatibility.md?plain=1#L64

For example, we can rename `Compatibility.md` to `Compatibility.template.md`, and delete the content from the `Integrations` section: 
```diff
## Integrations

- For a list of available integrations, and their configuration options, refer to the following:
+ [GENERATED_INTEGRATION_VERSIONS_PLACEHOLDER]
- ...
```
The GitHub action you wrote would then render the real `Compatibility.md`, replacing `[GENERATED_INTEGRATION_VERSIONS_PLACEHOLDER]` with the real value. Users could then manually update `Compatibility.template.md` if needed, to update other sections.

p-datadog (approved):

bouwkast (approved):
Nice! Thanks!

docs/integration_versions.md
```diff
@@ -0,0 +1,58 @@
+<!--
```
TonyCTHsu, line 1: The comment should also mentioned this file should not be manual edited, its content is generated by automation. 

Here is an example: https://github.com/DataDog/dd-trace-rb/blob/master/.github/workflows/test.yml#L1-L2

quinna-h: done! Thanks for the suggestion.

PR #4237, opened 2024-12-19
APMLP-350 fix crash in crashtracker when agent url is an ipv6 address

p-datadog:
**What does this PR do?**

- Repairs the crash in crashtracker due to insufficient error handling - adds a check for the endpoint construction failing (returning NULL) and raises an exception in this case
- Repairs incorrect agent URL construction from IPv6 addresses - the IPv6 addresses must be bracketed when put into the hostname of a URL

**Motivation:**

Crash reported by system tests

**Change log entry**

Yes: fix a crash in crashtracker when agent hostname is an IPv6 address

**Additional Notes:**

The code that this PR touches may have more missing error handling (of other situations), I didn't check it exhaustively.

**How to test the change?**

Unit tests are added. 
I also wrote https://github.com/DataDog/libdatadog/pull/809 for libdatadog to verify it works OK with our provided input (which it does).

### DIFF
--- a/lib/datadog/core/crashtracking/agent_base_url.rb
+++ b/lib/datadog/core/crashtracking/agent_base_url.rb
@@ -10,15 +10,13 @@ module Datadog
         # IPv6 regular expression from
         # https://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses
         # Does not match IPv4 addresses.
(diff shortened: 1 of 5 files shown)


lib/datadog/core/crashtracking/agent_base_url.rb
```diff
@@ -7,10 +7,17 @@ module Core
     module Crashtracking
       # This module provides a method to resolve the base URL of the agent
       module AgentBaseUrl
+        # IPv6 regular expression from
+        # https://stackoverflow.com/questions/53497/regular-expression-that-matches-valid-ipv6-addresses
```
Strech, line 11: https://ihateregex.io/expr/ipv6/ 😆

Strech (approved):
LGTM

lib/datadog/core/crashtracking/agent_base_url.rb
```diff
@@ -7,10 +7,17 @@ module Core
...
+
         def self.resolve(agent_settings)
           case agent_settings.adapter
           when Datadog::Core::Configuration::Ext::Agent::HTTP::ADAPTER
-            "#{agent_settings.ssl ? 'https' : 'http'}://#{agent_settings.hostname}:#{agent_settings.port}/"
+            hostname = agent_settings.hostname
+            hostname = "[#{hostname}]" if hostname =~ IPV6_REGEXP
+            "#{agent_settings.ssl ? 'https' : 'http'}://#{hostname}:#{agent_settings.port}/"
           when Datadog::Core::Configuration::Ext::Agent::UnixSocket::ADAPTER
             "unix://#{agent_settings.uds_path}"
```
ivoanjo, line 22: Looking at this makes me think that perhaps all of this class is in the wrong location.

What you think of maybe adding this into the `AgentSettingsResolver` as a `agent_settings.url`? That way we'd have a centralized place to generate the correct url, rather than needing to repeat (and potentially get wrong) this code.

lib/datadog/core/crashtracking/agent_base_url.rb
p-datadog: I did actually look at that and it doesn't appear to work right.

lib/datadog/core/crashtracking/agent_base_url.rb
p-datadog: I haven't looked into the details but at surface level I agree with what you are saying here that this logic seems like it should be elsewhere. The approval I had for repairing the crash was for a limited-time engagement, and refactoring working code - while good - wouldn't fall under that approval. Let's make a follow-up ticket to deal with this in the future?

p-datadog: @ivoanjo Since you are on PTO let's connect on this when you are back.

ivoanjo: Yeap, thanks for pushing this along, fixing the issue is higher priority.

ivoanjo: Since I needed to propagate a similar fix to profiling, I ended up taking my own suggestion in https://github.com/DataDog/dd-trace-rb/pull/4252 . Feedback welcome ;)

PR #4239, opened 2024-12-19
DEBUG-3182 Rework DI loading 

p-datadog:
**What does this PR do?**

This PR separates the part of DI that is required for code tracker (which is the code tracker itself and the DI.current_component group of methods, plus exception classes) to be self-contained. Currently in master loading `datadog/di/init` would bring in `datadog/di` as well which is too much (for example, that would cause DI to attempt to load ActiveRecord contrib, which is definitely not the right time for it).

**Motivation:**

Ensuring DI code tracker is loaded early without causing the rest of dd-trace-rb to be loaded early.

**Change log entry**

Yes: Improved early loading mechanism of dynamic instrumentation (`datadog/di/init`).

**Additional Notes:**

While this PR changes the behavior of `datadog/di/init`, this behavior has not yet been documented anywhere and should not be used by customers.

**How to test the change?**

Integration tests specifically for `datadog/di/init` will be added in a follow-up PR. The existing unit tests verify that basic DI functionality continues to operate correctly.

### DIFF
--- a/spec/datadog/di/init_spec.rb
+++ b/spec/datadog/di/init_spec.rb
@@ -4,6 +4,7 @@ require 'open3'
 RSpec.describe 'DI initializer' do
   di_test
 
+  # rubocop:disable Lint/ConstantDefinitionInBlock
   BOOTSTRAP_SCRIPT = <<-SCRIPT
     if defined?(Datadog) && Datadog.constants != %i(VERSION)
       raise "Datadog code loaded too early"
@@ -38,6 +39,7 @@ RSpec.describe 'DI initializer' do
       raise "Too many datadog components loaded at the end of execution: \#{Datadog.constants}"
     end
   SCRIPT
+  # rubocop:enable Lint/ConstantDefinitionInBlock
 
   context 'when loaded initially into a clean process' do
     it 'loads only DI code tracker' do


lib/datadog/di/base.rb
```diff
@@ -0,0 +1,116 @@
...
+          @current_components ||= []
+          @current_components << component
+        end
+      end
+
+      def remove_current_component(component)
+        LOCK.synchronize do
+          @current_components&.delete(component)
+        end
+      end
```
ivoanjo, line 113: If I understand your use-case well, you can maybe replace this with something along the lines of

```ruby
        components = Datadog.send(:components, allow_initialization: false)
        di = components.di if components
```

This allows you to access the components without triggering initialization. This is something [we use in for instance `Datadog::Tracing.correlation`](https://github.com/DataDog/dd-trace-rb/blob/7fd1feb1058822554c23d49f8e5ed03918759f9d/lib/datadog/tracing.rb#L98) to allow the API to be used, but not cause initialization as a side-effect.

ivoanjo:
> While this PR changes the behavior of datadog/di/init, this behavior has not yet been documented anywhere and should not be used by customers.

If I can offer some advice on this, is as much as possible don't expose any API that's expected to be used directly by customers in DI. Specifically, since DI is mostly about "turn it on, then go to the datadog UX to clicky clicky", it's worth taking advantage of this by not allowing too much configuration/fiddling/and manual operations.

This is somewhat similar to profiling: it's supported to be -- turn it on, then go look at the results. There's one or two public APIs but very very little surface, which has made it much easier to move fast and improve things without needing to walk on egg shells (or needing to put extra work or add extra overhead to comply with some specific public APIs)

lib/datadog/di/base.rb
```diff
@@ -0,0 +1,116 @@
...
+# are loaded, and also none of the rest of datadog library which also
+# has contrib code in other products.
+
+require_relative 'code_tracker'
+
+module Datadog
+  # Namespace for Datadog dynamic instrumentation.
+  #
+  # @api private
+  module DI
```
ivoanjo, line 17: Consider actually putting this under `DI::Base` or something like that? Specifically, I always find it mega confusing if I'm looking at `datadog/foo` and I open `datadog/foo.rb` and I'm missing code -- and I might not ever realize it.

I think as much as possible, keeping the 1:1 between namespace and file helps a lot of not getting lost in the codebase.

lib/datadog/di/base.rb
```diff
@@ -0,0 +1,116 @@
...
+
+    class << self
+      attr_reader :code_tracker
+
+      # Activates code tracking. Normally this method should be called
+      # when the application starts. If instrumenting third-party code,
+      # code tracking needs to be enabled before the third-party libraries
+      # are loaded. If you definitely will not be instrumenting
+      # third-party libraries, activating tracking after third-party libraries
+      # have been loaded may improve lookup performance.
```
ivoanjo, line 28: > If you definitely will not be instrumenting third-party libraries...

This seems like like a potential support very sharp edge? E.g. person A onboards app X, then perhaps even leaves the company, then person B tries to use DI with X, and has a bad experience.

I would advise being careful about adding these kinds of sharp edges, at least until you have a very specific customer that really really wants this, and in that case you could still document it in a more "this is probably not something you want to ever enable, as it will break etc etc" kind of scary way.

lib/datadog/di/base.rb
```diff
@@ -0,0 +1,116 @@
...
+          # without it.
+          Datadog::DI.activate_tracking!
+        rescue => exc
+          if defined?(Datadog.logger)
+            Datadog.logger.warn("Failed to activate code tracking for DI: #{exc.class}: #{exc}")
+          else
+            # We do not have Datadog logger potentially because DI code tracker is
+            # being loaded early in application boot process and the rest of datadog
+            # wasn't loaded yet. Output to standard error.
+            warn("Failed to activate code tracking for DI: #{exc.class}: #{exc}")
```
ivoanjo, line 57: Minor: Since in this mode there'll be no extra info in the log line, consider adding a `"[DATADOG]"` prefix or something like that. As a team member on some service, I might not know "what DI" is and that datadog is instrumenting my app, so it's useful to have something to quickly identify where this message is coming from.

lib/datadog/di/base.rb
```diff
@@ -0,0 +1,116 @@
...
+      # which contains a reference to the most recently instantiated
+      # DI::Component. This way, if a DI component hasn't been instantiated,
+      # we do not try to reference Datadog.components.
+      # In other words, this method exists so that we never attempt to call
+      # Datadog.components from the code tracker.
+      def current_component
+        LOCK.synchronize do
+          @current_components&.last
+        end
+      end
```
ivoanjo, line 94: See my note below on how this can be replaced with a simpler approach :)

ivoanjo (commented):
Left a few notes!

lib/datadog/di/base.rb
p-datadog: Check `defined?(Datadog.components)` before proceeding as above

lib/datadog/di/base.rb
p-datadog: Adjusted.

lib/datadog/di/base.rb
p-datadog: Added.

ivoanjo (approved):
:+1:

We spoke via dm and while I'm somewhat concerned about the complexity around tracking the current component (locks, state to be tracked, potential memory leaks if not used correctly), this PR only moves something that already existed, so we can tackle that separately.

PR #4240, opened 2024-12-19
[PROF-11003] Fix allocation profiling + otel tracing causing Ruby crash

ivoanjo:
**What does this PR do?**

This PR is a follow-up to #4195 and fixes a similar issue to that earlier PR.

Specifically, when the profiler is trying to read from the opentelemetry ("otel") context, we would try to look up a value from a Ruby hash using `rb_hash_lookup(...)`. This innocent-looking method would call `#hash` on the key being used, which meant that a Ruby method could start running in the middle of a profiler sample.

When combined with allocation profiling, we had reports of Ruby crashing with:

```
[BUG] unexpected situation - recordd:1 current:0

-- C level backtrace information -------------------------------------------
ruby(rb_print_backtrace+0x11) [0x55ba03ccf90f] vm_dump.c:820
ruby(rb_vm_bugreport) vm_dump.c:1151
ruby(bug_report_end+0x0) [0x55ba03e91607] error.c:1042
ruby(rb_bug_without_die) error.c:1042
ruby(die+0x0) [0x55ba03ac0998] error.c:1050
ruby(rb_bug) error.c:1052
ruby(disallow_reentry+0x0) [0x55ba03ab6dcc] vm_sync.c:226
ruby(rb_ec_vm_lock_rec_check+0x1a) [0x55ba03cb17aa] eval_intern.h:144
ruby(rb_ec_tag_state) eval_intern.h:155
ruby(rb_vm_exec) vm.c:2484
ruby(vm_invoke_proc+0x201) [0x55ba03cb62b1] vm.c:1509
ruby(rb_vm_invoke_proc+0x33) [0x55ba03cb65d3] vm.c:1728
ruby(thread_do_start_proc+0x176) [0x55ba03c63516] thread.c:598
ruby(thread_do_start+0x12) [0x55ba03c648a2] thread.c:615
ruby(thread_start_func_2) thread.c:672
ruby(nt_start+0x107) [0x55ba03c65137] thread_pthread.c:2187
/lib/x86_64-linux-gnu/libpthread.so.0(start_thread+0xd9) [0x7ff360b66609]
/lib/x86_64-linux-gnu/libc.so.6(clone+0x43) [0x7ff360a70353]
```

This PR fixes this by replacing the `rb_hash_lookup` with a native-level linear iteration on the hash contents, which allows us to skip invoking any kind of Ruby methods.

Additionally, this PR:

* Adds a checker mechanism to detect these kinds of problem conditions when executing our test suite. This mechanism was able to flag both the issue in #4195 as well as the one in `rb_hash_lookup` 

*  Adds a tweak to avoid allocation sampling while Ruby is raising an exception, as in my testing this was another way of very easily causing Ruby to crash or misbehave

* Fixes a potential issue with static vs dynamic Ruby symbols when reading from the otel context code.

**Motivation:**

Fix the crash.

**Change log entry**

Yes. Fix allocation profiling + otel tracing causing Ruby crash

**Additional Notes:**

As a reminder, reading from the opentelemetry context is disabled by default, so only customers setting `DD_PROFILING_PREVIEW_OTEL_CONTEXT_ENABLED` to `true/both/only` could be affected by this issue.

Also, the crash was only triggered when allocation profiling was in use.

This PR may make a bit more sense to read commit-by-commit.

**How to test the change?**

This PR adds code coverage + a new checker to find the fixed issues.


ext/datadog_profiling_native_extension/unsafe_api_calls_check.c
```diff
@@ -0,0 +1,47 @@
...
+    rb_postponed_job_register(0, check_for_unsafe_api_calls, NULL);
+  #endif
+}
+
+void debug_leave_unsafe_context(void) {
+  inside_unsafe_context = false;
+}
+
+static void check_for_unsafe_api_calls(DDTRACE_UNUSED void *_unused) {
+  if (inside_unsafe_context) rb_bug(
```
AlexJF, line 44: This is a hard crash right? I see how it would make issues more visible but won't it also lead the profiler to crash client apps more often than what may be caused by some "racy" issues? Are we trading off potential profiling accuracy with increased crashing or would the unsafe calls potentially manifest in issues more dangerous than a simple crash (e.g. a data corruption)?

lib/datadog/core/configuration.rb
```diff
@@ -236,7 +236,7 @@ def safely_synchronize
           rescue ThreadError => e
             logger_without_components.error(
               'Detected deadlock during datadog initialization. ' \
-              'Please report this at https://github.com/DataDog/dd-trace-rb/blob/master/CONTRIBUTING.md#found-a-bug' \
+              'Please report this at https://github.com/datadog/dd-trace-rb/blob/master/CONTRIBUTING.md#found-a-bug' \
```
AlexJF, line 239: Nit: Intended? I mean it works but the official casing if the previous.

AlexJF (approved):

lib/datadog/core/configuration.rb
ivoanjo: I usually try to lower-case whenever I see it linked like this, trying to convince the world it's not DataDog ;)

ext/datadog_profiling_native_extension/unsafe_api_calls_check.c
ivoanjo: Added big scary comment in 7e493241465e0a6e1a817dc0be9b4e5502bb6a9a

AlexJF: >  I've only placed calls to debug_enter_unsafe_context and debug_leave_unsafe_context in methods that are in the testing_module, so this code never gets exercised in the "normal" in-production profiling path.

Ah I did wonder if that was true given the first few places you added it to had the "don't use this anywhere else" comment. But some of the other methods didn't and I was lazy to check where they were registered or called from :shame:

p-datadog:
Is the iteration limit for hash lookup added in this PR customer visible? Seems to me that it would be? If yes should it be noted somewhere?

ivoanjo:
> Is the iteration limit for hash lookup added in this PR customer visible? Seems to me that it would be? If yes should it be noted somewhere?

It's a good call-out but... I'm somewhat hesitant to introduce too many knobs of this kind if we can avoid it.

This whole feature itself is still under preview (and off by default), so my thinking is: let's see if any customer reports any issue with the current hardcoded value for this setting, and in that case I'll work with them do decide what value would be needed. After that, we can decide if we bump the default, or do something else.

Does this sound reasonable, or are you not convinced yet?

ivoanjo:
Ok so the failure in the benchmarks is expected and an unfortunate effect of how the benchmarks are currently ran. In particular, our harness tries to run the benchmarks from this branch with the main branch, which is an interesting idea BUT forces the benchmark code to be backwards-compatible, which is really annoying for some changes such as this one.

The master branch is currently red due to unrelated issues; my current plan for this PR is to wait until master is green again, then merge it into this branch, and then merge it once everything is green other than the benchmarks.

marcotc (approved):

ivoanjo:
Update: After merging to master, CI is green as expected:
image

PR #4243, opened 2024-12-25
Update integrations images build script

p-datadog:
**What does this PR do?**
Remove pre-2.5 image builds commands since the minimum Ruby version we now support is 2.5.

Add 3.4 as a supported Ruby version to the help text (it is already handled by the script).

**Motivation:**
Researching how to perform integration testing of DI.

**Change log entry**
None

**Additional Notes:**

**How to test the change?**
N/A


y9v (approved):

ivoanjo (approved):
:+1: Thanks for the cleanup

PR #4257, opened 2025-01-06
[NO-TICKET] RFC: Add `logger` gem as dependency to prepare for future Ruby versions

ivoanjo:
**What does this PR do?**

This PR declares a dependency from the `datadog` gem to the `logger` gem.

While every PR is an RFC, I explicitly marked this one as such since adding a new dependency should never be done lightly.

**Motivation:**

This is needed to avoid a warning:

> lib/datadog/core/configuration/settings.rb:3: warning: logger was
> loaded from the standard library, but will no longer be part of the
> default gems starting from Ruby 3.5.0.
> You can add logger to your Gemfile or gemspec to silence this warning.

**Change log entry**

Yes. Add `logger` gem as dependency to prepare for future Ruby versions

**Additional Notes:**

`logger` is just the latest in a long line of gems that have been refactored out in this way. See
https://github.com/ruby/ruby/commit/d7e558e3c48c213d0e8bedca4fb547db55613f7c and https://stdgems.org/ (not yet updated for 3.5 as of this writing).

While in the past we've avoided adding extra dependencies due to this extraction work, I think this one does make sense to add:

1. It's supported on every Ruby version we currently support (>= 2.5)
2. It's not a small amount of code to shed/rewrite. And on the other side, I don't see a strong reason for embedding instead of depending on it.

We're getting warnings for two more gems: `ostruct` and `benchmark`.

From our internal discussions, we can probably get rid of our usage of them, rather than promote them to a dependency, so that's why I did not touch those warnings in this PR.

**How to test the change?**

The following script can be used (on Ruby 3.4 and above) to trigger the warning, and to confirm it's gone once `logger` is added as a dependency:

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'datadog', path: '.'
end

puts RUBY_DESCRIPTION
require 'datadog'
puts Datadog::VERSION::STRING
```

Make sure to run with just `ruby example.rb` and not with `bundle exec`.


p-datadog (approved):

marcotc (approved):

ivoanjo:
While I don't expect this to have any fallback, changing dependencies always makes me a bit nervous so I'll mark this as do-not-merge until we put the 2.9.0 release out. 

This way we can merge it in early during the 2.10 cycle and a bit longer baking time.

ivoanjo:
Now that 2.9 is out, I'm going to go ahead and merge this one. Thanks for the help, folks!

PR #4261, opened 2025-01-06
DEBUG-3182 move Rails utils to core

p-datadog:
**What does this PR do?**

Moves Tracing::Contrib::Rails::Utils to Core

**Motivation:**
DI requires a similar loading logic for contribs as Tracing (currently we have one DI contrib which is ActiveRecord). I am working on having a DI railtie that would load the DI AR contrib late in application bootstrap, like tracing is doing with its contribs. To detect whether Railtie functionality is available, I need the Contrib::Rails::Utils methods.

**Change log entry**
None

**Additional Notes:**

**How to test the change?**


lib/datadog/core/contrib/rails/utils.rb
```diff
@@ -1,9 +1,7 @@
 # frozen_string_literal: true
```
ivoanjo, line 1: Minor: It seems a bit overkill to create a `contrib/` folder with a single `rails/` subfolder with a single file just for these tiny utilities :eyes:...

May I suggest putting this instead into `lib/datadog/core/utils/rails.rb` / or at most `lib/datadog/core/contrib/rails.rb`? We can always move it later if we find we have too many utils in the same folder.

ivoanjo (approved):
:+1: LGTM

y9v (approved):

lib/datadog/core/contrib/rails/utils.rb
ivoanjo: Looking at the specs, I don't think they need rails working right now? But yeah if it's getting too annoying, we can always go back to the previous (sigh... I really dislike how/we/get/very/deep/folder/structures/in/our/project :P )

p-datadog: The tests for this module itself requires rails to be loaded, and outside of contrib that is not the case.

marcotc (approved):

PR #4262, opened 2025-01-07
DEBUG-3251 dependency inject logger into DI component

p-datadog:
**What does this PR do?**

Alters DI component initialization to pass the logger into DI as a parameter rather than DI reading it out of Datadog module.

**Motivation:**

Logging should be passed into DI in one way.
Additionally, #4230 makes changes to how the logging is handled within DI, this PR makes the test suite log to mocks in several places and these places would need to be adjusted in addition to the other code in #4230.

**Change log entry**
None

**Additional Notes:**

**How to test the change?**


lib/datadog/di/component.rb
```diff
@@ -16,22 +16,22 @@ module DI
     # resources and installed tracepoints upon shutdown.
     class Component
       class << self
-        def build(settings, agent_settings, telemetry: nil)
+        def build(settings, agent_settings, logger, telemetry: nil)
```
ivoanjo, line 19: Minor: May be worth moving to all keyword arguments? (Although the type checker will complain if we get this wrong since you added signatures -- nice!)

ivoanjo (approved):
:+1: LGTM

PR #4269, opened 2025-01-09
Add stack trace collection support

vpellan:
**What does this PR do?**

This PR adds stack trace collection support using meta_struct

**Motivation:**

Stack trace collection is required for exploit prevention.

**Change log entry**

Yes. Fix stack trace not being with RASP SQLi signals

**Additional Notes:**

**How to test the change?**


lib/datadog/appsec/actions_handler.rb
```diff
@@ -19,7 +21,26 @@ def interrupt_execution(action_params)
         throw(Datadog::AppSec::Ext::INTERRUPT, action_params)
       end
 
-      def generate_stack(_action_params); end
+      def generate_stack(action_params)
+        if Datadog.configuration.appsec.stack_trace.enabled
+          context = AppSec::Context.active
+          return if context.nil? ||
+            ActionsHandler::StackTrace.skip_stack_trace?(context, group: AppSec::Ext::EXPLOIT_PREVENTION_EVENT_CATEGORY)
```
Strech, line 28: This is a sign that responsibilities are mixed. This module is a glue and should know about **when** to collect and **what** to do. On the other hand StackTrace should have 0 knowledge about when to do something, it should know only **how**.

lib/datadog/appsec/actions_handler/stack_trace.rb
```diff
@@ -0,0 +1,76 @@
...
+    module ActionsHandler
+      # Adds stack traces to meta_struct
+      module StackTrace
+        module_function
+
+        def skip_stack_trace?(context, group:)
+          if context.trace.nil? && context.span.nil?
+            Datadog.logger.debug { 'Cannot find trace or service entry span to add stack trace' }
+            return true
+          end
```
Strech, line 19: This doesn't belong here

lib/datadog/appsec/actions_handler/stack_trace.rb
```diff
@@ -0,0 +1,76 @@
...
+        end
+
+        def collect_stack_frames
+          # caller_locations without params always returns an array but steep still thinks it can be nil
+          # So we add || [] but it will never run the second part anyway (either this or steep:ignore)
+          stack_frames = caller_locations || []
+          # Steep thinks that path can still be nil and that include? is not a method of nil
+          # We must add a variable assignment to avoid this
+          stack_frames.reject! do |loc|
+            path = loc.path
```
Strech, line 46: Is this variable needed? What is it saving us?

lib/datadog/appsec/actions_handler/stack_trace/frame.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Datadog
+  module AppSec
+    module ActionsHandler
+      module StackTrace
+        # Formatted stack frame.
+        # This class extends a Struct as it's required by Steep to be able to add a method to it.
+        class Frame < Struct.new(:id, :text, :file, :line, :function, keyword_init: true) # rubocop:disable Style/StructInheritance
```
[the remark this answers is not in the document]
Strech:
```suggestion
        Frame = Struct.new(:id, :text, :file, :line, :function, keyword_init: true) do
```

lib/datadog/appsec/actions_handler/stack_trace/collector.rb
```diff
@@ -0,0 +1,61 @@
...
+
+require_relative 'frame'
+
+module Datadog
+  module AppSec
+    module ActionsHandler
+      module StackTrace
+        # Represent a stack trace with its id and message in message pack
+        module Collector
+          class << self
```
Strech, line 11: That's suspicious 

```suggestion
          module_function
```

Strech (changes requested):
I think the whole structure needs a general overhaul. Responsibilities are mixed, class/modules definitions signaling that something is off.

I think the issue is that somewhere it makes sense to have the class, not module, somewhere it makes sense to have checks and somewhere it needs to know nothing.

I have a feeling that it will require lots of changes to have another round of the review and my suggestion is to go over 1:1 instead

lib/datadog/appsec/actions_handler/stack_trace/collector.rb
```diff
@@ -0,0 +1,61 @@
...
+            def collect(locations)
+              return [] if locations.nil? || locations.empty?
+
+              skip_frames = skip_frames(locations.size)
+              frames = []
```
y9v, line 16: I would suggest the same here and everywhere - let's try to have consistent naming, if we call those `frames`, let's not call them `locations` (or the other way around)

lib/datadog/appsec/actions_handler/stack_trace/collector.rb
```diff
@@ -0,0 +1,61 @@
...
+                  line: location.lineno,
+                  function: function_label(location)
+                )
+              end
+              frames
+            end
+
+            private
+
+            def skip_frames(locations_size)
```
y9v, line 34: from this name I don't get what this function is doing - is it skipping frames?

lib/datadog/appsec/actions_handler/stack_trace/collector.rb
```diff
@@ -0,0 +1,61 @@
...
+        module Collector
+          class << self
+            def collect(locations)
+              return [] if locations.nil? || locations.empty?
+
+              skip_frames = skip_frames(locations.size)
+              frames = []
+
+              locations.each_with_index do |location, index|
+                next if skip_frames.include?(index)
```
y9v, line 19: it's ruby, this code can be way simpler - without a temporary array `frames` that we don't need, without iterating with index - if you know the lower and upper bound, you can just take an array slice and work with it

lib/datadog/appsec/actions_handler/stack_trace/collector.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+require_relative 'frame'
+
+module Datadog
+  module AppSec
+    module ActionsHandler
+      module StackTrace
+        # Represent a stack trace with its id and message in message pack
+        module Collector
```
y9v, line 10: what is the responsibility of this `Collector` module? As I see it is only taking a slice from the stack trace depending on some factors, but it's not collecting anything

lib/datadog/appsec/context.rb
```diff
@@ -60,10 +60,13 @@ def extract_schema
       end
 
       def export_metrics
-        return if @span.nil?
+        # Required to satisfy steep, as @span is defined as nilable.
+        # According to soutaro, this is because instance variable can be changed by other threads.
+        span = @span
+        return if span.nil?
```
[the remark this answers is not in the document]
y9v: I would rather find another way to silence steep (and maybe think whether we really need such tool)

lib/datadog/appsec/actions_handler/stack_trace.rb
```diff
@@ -0,0 +1,76 @@
...
+            stack_traces_count += span_dd_stack[group].size unless span_dd_stack.nil? || span_dd_stack[group].nil?
+          end
+
+          stack_traces_count >= max_collect
+        end
+
+        def collect_stack_frames
+          # caller_locations without params always returns an array but steep still thinks it can be nil
+          # So we add || [] but it will never run the second part anyway (either this or steep:ignore)
+          stack_frames = caller_locations || []
```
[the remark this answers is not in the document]
vpellan: I will rename some places where I agree it is unclear but I think there is a difference between location and frames. Locations are objects created by Ruby when calling Kernel.caller_locations. A frame is an object we create, that only selects a portion of the informations that locations are giving us.

lib/datadog/appsec/actions_handler.rb
```diff
@@ -19,7 +21,26 @@ def interrupt_execution(action_params)
         throw(Datadog::AppSec::Ext::INTERRUPT, action_params)
       end
 
-      def generate_stack(_action_params); end
+      def generate_stack(action_params)
+        if Datadog.configuration.appsec.stack_trace.enabled
```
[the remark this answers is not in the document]
vpellan: Fixed in 8bae22cdc5b9e781bc8e6dcfe53b629910a75c6a

lib/datadog/appsec/actions_handler.rb
```diff
@@ -19,7 +21,26 @@ def interrupt_execution(action_params)
...
+          context = AppSec::Context.active
```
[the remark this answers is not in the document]
vpellan: Fixed in 8bae22cdc5b9e781bc8e6dcfe53b629910a75c6a

lib/datadog/appsec/actions_handler.rb
vpellan: Discussed during a meeting, resulting in 8bae22cdc5b9e781bc8e6dcfe53b629910a75c6a

lib/datadog/appsec/actions_handler/stack_trace.rb
```diff
@@ -0,0 +1,76 @@
...
+
+          unless context.span.nil?
+            span_dd_stack = context.span.metastruct[AppSec::Ext::TAG_STACK_TRACE]
+            stack_traces_count += span_dd_stack[group].size unless span_dd_stack.nil? || span_dd_stack[group].nil?
+          end
+
+          stack_traces_count >= max_collect
+        end
+
+        def collect_stack_frames
```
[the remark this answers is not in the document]
vpellan: Fixed in 8bae22cdc5b9e781bc8e6dcfe53b629910a75c6a (Stack trace collection logic is all located in StackTraceCollection.collect)

lib/datadog/appsec/actions_handler/stack_trace/collector.rb
vpellan: Fixed in 8bae22cdc5b9e781bc8e6dcfe53b629910a75c6a (In StackTraceCollection module)

spec/datadog/appsec/actions_handler_spec.rb
```diff
@@ -97,4 +101,124 @@
...
+            trace_op_result = trace_op.metastruct.dig(stack_key, exploit_category)
+            expect(trace_op_result.size).to eq(1)
+            expect(trace_op_result.first[:id]).to eq('foo')
+          end
+        end
+      end
+    end
+
+    context 'when stack trace is enabled and context contains only span' do
+      let(:context) { OpenStruct.new(span: span_op) }
```
Strech, line 206: Please use verified double i.e (`instance_double`) here

spec/datadog/appsec/actions_handler_spec.rb
```diff
@@ -97,4 +101,124 @@
...
+    let(:span_op_metastruct) { nil }
+    let(:context) { OpenStruct.new(trace: trace_op, span: span_op) }
+    let(:stack_trace_enabled) { true }
+    let(:max_collect) { 0 }
+
+    let(:stack_key) { Datadog::AppSec::Ext::TAG_STACK_TRACE }
+    let(:exploit_category) { Datadog::AppSec::Ext::EXPLOIT_PREVENTION_EVENT_CATEGORY }
+
+    before do
+      Datadog.configure do |c|
```
Strech, line 119: FYI if you go this way - you will need to clean the settings afterwards (see other tests where we do that)

spec/datadog/appsec/actions_handler_spec.rb
```diff
@@ -97,4 +101,124 @@
...
+      allow(Datadog::AppSec::ActionsHandler::StackTraceCollection).to receive(:collect).and_return(
+        ThreadBacktraceHelper.locations_inside_nested_blocks.map.with_index do |location, index|
+          {
+            id: index,
+            text: location.to_s.encode('UTF-8'),
+            file: (location.absolute_path || location.path)&.encode('UTF-8'),
+            line: location.lineno,
+            function: location.label&.encode('UTF-8')
+          }
+        end
```
Strech, line 135: TBH i thought that helper suppose to make things easier. I think here it's either missing something or else ...

spec/datadog/appsec/actions_handler_spec.rb
```diff
@@ -97,4 +101,124 @@
       end
     end
   end
+
+  describe '.generate_stack' do
```
Strech, line 105: TBH this test suite look overly complicated and violates certain test practices and it would be easier to go over in 1:1 instead of chatting here.

Strech (changes requested):

lib/datadog/appsec/actions_handler.rb
```diff
@@ -19,7 +22,33 @@ def interrupt_execution(action_params)
...
+        # Generate stacktrace
+        utf8_stack_id = action_params['stack_id'].encode('UTF-8') if action_params['stack_id']
+        stack_frames = ActionsHandler::StackTraceCollection.collect(
+          max_depth: config.max_depth,
+          top_percent: config.max_depth_top_percent
+        )
+
+        # Add newly created stacktrace to metastruct
+        stack = context.trace.nil? ? span_stack : trace_stack
+        stack.push({ language: 'ruby', id: utf8_stack_id, frames: stack_frames })
```
Strech, line 50:
```suggestion
        (context.trace || context.span).push(language: 'ruby', id: stack_id, frames: stack_frames)
```

lib/datadog/appsec/actions_handler.rb
```diff
@@ -19,7 +22,33 @@ def interrupt_execution(action_params)
...
+
+        config = Datadog.configuration.appsec.stack_trace
+
+        # Check that the sum of stack_trace count in trace and entry_span does not exceed configuration
+        span_stack = ActionsHandler::StackTraceInMetastruct.create(context.span&.metastruct)
+        trace_stack = ActionsHandler::StackTraceInMetastruct.create(context.trace&.metastruct)
+        return if config.max_collect != 0 && span_stack.count + trace_stack.count >= config.max_collect
+
+        # Generate stacktrace
+        utf8_stack_id = action_params['stack_id'].encode('UTF-8') if action_params['stack_id']
```
Strech, line 42: This variable will be `nil` if no `params[stack_id]` is given, right? I think the whole method will not work without stack_id, hence we always have it or of not - we should do nothing and report a telemetry or generate warning or so.

Because of that that line must be adjusted to

```suggestion
        stack_id = action_params['stack_id'].encode('UTF-8')
```
_P.S it doesn't make sense to mention UTF-8 in the variable name because it's not important_

lib/datadog/appsec/actions_handler.rb
```diff
@@ -19,7 +22,33 @@ def interrupt_execution(action_params)
...
+        if context.nil? || context.trace.nil? && context.span.nil?
+          Datadog.logger.debug { 'Cannot find trace or service entry span to add stack trace' }
+          return
+        end
+
+        config = Datadog.configuration.appsec.stack_trace
+
+        # Check that the sum of stack_trace count in trace and entry_span does not exceed configuration
+        span_stack = ActionsHandler::StackTraceInMetastruct.create(context.span&.metastruct)
+        trace_stack = ActionsHandler::StackTraceInMetastruct.create(context.trace&.metastruct)
```
Strech, line 38:
```suggestion
        span_stack = StackTraceInMetastruct.create(context.span&.metastruct)
        trace_stack = StackTraceInMetastruct.create(context.trace&.metastruct)
```

lib/datadog/appsec/actions_handler/stack_trace_collection.rb
```diff
@@ -0,0 +1,37 @@
...
+
+module Datadog
+  module AppSec
+    module ActionsHandler
+      # Module that collects the stack trace into formatted hash
+      module StackTraceCollection
+        module_function
+
+        def collect(max_depth:, top_percent:)
+          locations = (caller_locations || []).reject { |location| location.to_s.include?('lib/datadog') }
```
Strech, line 11: I think it should be converted into mutating to avoid allocation

```suggestion
          locations = (caller_locations || [])
          locations.reject! { |location| location.to_s.include?('lib/datadog') }
```

lib/datadog/appsec/actions_handler/stack_trace_collection.rb
```diff
@@ -0,0 +1,37 @@
...
+
+          locations.slice!(top_limit...bottom_limit)
+          StackTraceCollection.convert(locations)
+        end
+
+        def convert(locations)
+          locations.map.with_index do |location, index|
+            {
+              id: index,
+              text: location.to_s.encode('UTF-8'),
```
Strech, line 27: This is a second time we convert location into the string, I think it makes sense to do it once at the beginning of the collection

lib/datadog/appsec/actions_handler.rb
```diff
@@ -19,7 +22,33 @@ def interrupt_execution(action_params)
...
+      def generate_stack(action_params)
+        return unless Datadog.configuration.appsec.stack_trace.enabled
+
+        context = AppSec.active_context
+        if context.nil? || context.trace.nil? && context.span.nil?
+          Datadog.logger.debug { 'Cannot find trace or service entry span to add stack trace' }
+          return
+        end
+
+        config = Datadog.configuration.appsec.stack_trace
```
y9v, line 34: this should be closer to the guard-clause on line 39

lib/datadog/appsec/actions_handler.rb
```diff
@@ -19,7 +22,33 @@ def interrupt_execution(action_params)
...
+
+        # Check that the sum of stack_trace count in trace and entry_span does not exceed configuration
+        span_stack = ActionsHandler::StackTraceInMetastruct.create(context.span&.metastruct)
```
y9v, line 37: Why are we calling span metastruct a `span_stack`? Should we rather call it `span_metastruct`? Same one line below

lib/datadog/appsec/actions_handler.rb
```diff
@@ -19,7 +22,33 @@ def interrupt_execution(action_params)
...
+        trace_stack = ActionsHandler::StackTraceInMetastruct.create(context.trace&.metastruct)
+        return if config.max_collect != 0 && span_stack.count + trace_stack.count >= config.max_collect
```
y9v, line 39: do we want to avoid stack trace collection if the stack is longer than `max_collect`, or do we want to truncate the stack?

lib/datadog/appsec/actions_handler/stack_trace_collection.rb
```diff
@@ -0,0 +1,37 @@
...
+        def collect(max_depth:, top_percent:)
+          locations = (caller_locations || []).reject { |location| location.to_s.include?('lib/datadog') }
+
+          return [] if locations.empty?
+          return StackTraceCollection.convert(locations) if max_depth.zero? || locations.size <= max_depth
+
+          top_limit = (max_depth * top_percent / 100.0).round
+          bottom_limit = locations.size - (max_depth - top_limit)
+
+          locations.slice!(top_limit...bottom_limit)
```
y9v, line 19: is `top_limit` lower than `bottom_limit`?

What are we trying to do here? Get the top n items from the array? Do we really need to calculate both limits, or can we just slice n items from one end?

y9v (changes requested):

lib/datadog/appsec/actions_handler.rb
vpellan: This does not work as we are using the push method defined in StackTraceInMetastruct

lib/datadog/appsec/actions_handler.rb
vpellan: Because StackTraceInMetastruct is specifically applying changes to '_dd.stack' field of metastruct. We give a metastruct as an input but we cannot do any changes to it other than on the stack traces

lib/datadog/appsec/actions_handler/stack_trace_collection.rb
vpellan: For example, if max_depth is 32 and top_percent is 75%, and we have more than 32 stack frames, we want to keep the 24th first stack traces, the 8th last, and drop everything in the middle. So what this method do is calculate the limits of the frames that we should drop, so we can just do a slice on the locations with a range

y9v: according to RFC, we have to take 2 slices - one from the top of the stack trace, and one from the bottom:

```
In the original RFC, some restrictions were placed on the layout of stack traces, namely:
Stack traces must contain a maximum of 32 frames.
When a stack trace is truncated, 24 frames should be from the bottom of the stack and 8 from the top.

The number of top and bottom frames on truncated stack traces can be considered a flexible requirement and it’s up to the library implementor to decide the most useful number of frames from each side of the stack for their language. In addition, since the user can configure the maximum number of frames through the use of DD_APPSEC_MAX_STACK_TRACE_DEPTH, the number of top and bottom frames for a truncated stack trace of arbitrary size must be computed dynamically based on the configuration value, with the simplest approach being a proportional one. 

Similarly, the default maximum number of frames might not be particularly useful in languages with notoriously deep stacks, for that reason, the requirement must be updated to ensure that each library can provide meaningful stack traces.

Given the above, the requirements can be amended as follows:
Stack traces must be limited to a maximum of 32 frames, or a more suitable higher value for the given language.
When a stack trace is truncated, libraries must collect some frames from both the top and bottom of the stack. As a recommendation, libraries may collect 75% of frames from the bottom of the stack and 25% from the top.
```

I think this is not what we are doing here?

y9v (changes requested):

lib/datadog/appsec/actions_handler/stack_trace_in_metastruct.rb
```diff
@@ -0,0 +1,39 @@
...
+        end
+
+        def count
+          @metastruct.dig(AppSec::Ext::TAG_STACK_TRACE, AppSec::Ext::EXPLOIT_PREVENTION_EVENT_CATEGORY)&.size || 0
+        end
+
+        def push(stack_trace)
+          @metastruct[AppSec::Ext::TAG_STACK_TRACE] ||= {}
+          @metastruct[AppSec::Ext::TAG_STACK_TRACE][AppSec::Ext::EXPLOIT_PREVENTION_EVENT_CATEGORY] ||= []
+          @metastruct[AppSec::Ext::TAG_STACK_TRACE][AppSec::Ext::EXPLOIT_PREVENTION_EVENT_CATEGORY] << stack_trace
```
y9v, line 34: Could it ever happen that we have to add more than 1 stack trace per http call?

If so, I would find it really confusing to see 2 stack traces combined into 1 stack trace (assuming we still have capacity left in the stack trace metastruct field).

vpellan:
> according to RFC, we have to take 2 slices - one from the top of the stack trace, and one from the bottom. I think this is not what we are doing here?

The slice! method will remove the part in the middle and locations will then contains the top part and the bottom part only.

y9v:
> > according to RFC, we have to take 2 slices - one from the top of the stack trace, and one from the bottom. I think this is not what we are doing here?
> 
> The slice! method will remove the part in the middle and locations will then contains the top part and the bottom part only.

sorry, my bad, you are right

PR #4271, opened 2025-01-09
[NO-TICKET] Enable typechecking more files

ivoanjo:
**What does this PR do?**

This PR enables typechecking for a lot more files that it turns out are already passing typechecks.

Many of these files we only have skeleton types that still need to be improved, so this is a small win, but it was a quite cheap one so I thought it was worth doing.

**Motivation:**

I noticed that we had at least one file that was passing typecheck, but still ignored. This sparked my curiosity, and with a bit of scripting I got a huge list of files that we could enable typechecking for already :)

**Change log entry**

None

**Additional Notes:**

N/A

**How to test the change?**

Validate that `bundle exec steep check` still passes!


sarahchen6 (approved):
nice!

marcotc (approved):

PR #4272, opened 2025-01-09
DEBUG-3182 DI railtie

p-datadog:
**What does this PR do?**

Adds a Railtie for dynamic instrumentation. The railtie delays loading of DI contrib code (currently, ActiveRecord serializer) until Rails components have been loaded.

Also adds an integration test verifying AR serialization works correctly.

**Motivation:**

Improving DI reliability 

**Change log entry**

Yes: improve loading of dynamic instrumentation components

**Additional Notes:**

**How to test the change?**

An integration test is added in this PR

### DIFF
--- a/lib/datadog/auto_instrument.rb
+++ b/lib/datadog/auto_instrument.rb
@@ -6,6 +6,11 @@
 require_relative '../datadog'
 require_relative 'tracing/contrib/auto_instrument'
 
+# DI is not loaded on Ruby 2.5 and JRuby
+if defined?(Datadog::DI::Contrib)
+  Datadog::DI::Contrib.load_now_or_later
+end
+
 Datadog::Profiling.start_if_enabled
 
 module Datadog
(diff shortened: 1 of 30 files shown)


lib/datadog/auto_instrument.rb
```diff
@@ -6,6 +6,9 @@
 require_relative '../datadog'
 require_relative 'tracing/contrib/auto_instrument'
 
+# DI is not loaded on Ruby 2.5 and JRuby
+Datadog::DI::Contrib.load_now_or_later if defined?(Datadog::DI::Contrib)
```
marcotc, line 10: Won't `Datadog::DI::Contrib.load_now_or_later` be invoked twice when this file is loaded?

I'm thinking that you would be involved here, as well as when `di.rb` is loaded. Is that correct?

p-datadog: Yes this is intended, because in between invocations there could be more third-party code loaded. For example, if a Rails app is launched with ddprofrb, the first invocation won't have Rails loaded but the second will have it.

p-datadog: Added a comment to the source code stating this.

marcotc (approved):

PR #4274, opened 2025-01-09
DEBUG-3249 remove usage of OpenStruct in DI

p-datadog:
**What does this PR do?**
Removes usage of OpenStruct in dynamic instrumentation.

Adds `ostruct` as dependency for `:test` group, since OpenStruct is used in the test suite.

Removes `ostruct` dependency in all groups as added in #4271.

**Motivation:**
Fixing the `ostruct` warning produced by rubygems

**Change log entry**
None

**Additional Notes:**
I added explicit requires of `ostruct` to tests that use it.

**How to test the change?**


lib/datadog/di/transport.rb
ivoanjo, line 1: **Important**: We're missing here the key part of triggering the warning, which is removing the `require 'ostruct'` from this class

ruby-2.5.gemfile
```diff
@@ -48,3 +48,7 @@ end
 group :dev do
   gem 'appraisal', '~> 2.4.0', require: false
 end
+
+group :test do
+  gem 'ostruct'
+end
```
ivoanjo, line 54: So in #4271 I was actually somewhat deliberate in only adding it to the Ruby 3.5 gemfile (although I completely failed to document that).

My thinking is: we only need it on Ruby 3.5, and maybe on 3.4, since all others ship their own version of ostruct.

Since adding to every variant is extra work + a lot of diff noise, IDK, consider maybe going back and removing it? (But I'll be clear this is not a blocker -- this PR can go in with this addition)

ivoanjo (changes requested):
Marked request changes as the warning is actually still emitted with the current version ;)

lib/datadog/di/transport.rb
p-datadog: Removed.

ruby-2.5.gemfile
p-datadog: I tested this and the warning about ostruct is emitted by bundler:

```
big% rbenv shell 3.4
big% be ruby test.rb
test.rb:1: warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add ostruct to your Gemfile or gemspec to silence this warning.
big% ruby test.rb 
big% 
```

Only on ruby 3.4.

Therefore I think the gem declaration should be in ruby 3.5+ where it is required and in ruby 3.4 where it will silence the warning, and does not need to be in ruby 3.3 and lower.

lib/datadog/di/transport.rb
```diff
@@ -61,9 +60,7 @@ def send_input(payload)
       attr_reader :client
 
       def send_request(desc, **options)
-        # steep:ignore:start
```
ivoanjo, line 64: Minor: Hey TIL that this was possible! May come in handy in those files where there's some bizarre error that we just can't figure out, but everything else is type checking fine

ivoanjo (approved):
:+1: LGTM Thanks for the second pass :)

lib/datadog/di/transport.rb
p-datadog: :+1:

marcotc (approved):

PR #4275, opened 2025-01-10
[NO-TICKET] Disable flaky barrier spec

ivoanjo:
**What does this PR do?**

This PR disables a spec that's been flaky for a very long time (see https://github.com/DataDog/ruby-guild/issues/191, but I suspect I saw it back in https://github.com/DataDog/ruby-guild/issues/50#issuecomment-1874210356 ) and it just failed on me again:
https://github.com/DataDog/dd-trace-rb/actions/runs/12706052002/job/35418238631?pr=4234

I'm not sure this test is getting us a lot of value, so let's stop the flaky impact until we can dedicate time to fixing it.

(Maybe @lloeki can help here as the original author of the class and tests?)

**Motivation:**

Get to zero flaky tests in our codebase!

**Change log entry**

None

**Additional Notes:**

N/A

**How to test the change?**

Validate spec is now skipped.

### DIFF
--- a/spec/datadog/core/remote/component_spec.rb
+++ b/spec/datadog/core/remote/component_spec.rb
@@ -258,8 +258,26 @@ RSpec.describe Datadog::Core::Remote::Component::Barrier do
 
     context 'with waiters' do
       it 'unblocks waiters' do
+        skip('Known flaky (assertion below sometimes fails with timeout)')
+
         waiter_thread = Thread.new(record) do |record|
           record << :one
+          # Failures:
+          #
+          #   1) Datadog::Core::Remote::Component::Barrier#lift with waiters unblocks waiters
+          #      Failure/Error: expect(barrier.wait_once).to eq :lift
+          #
+          #        expected: :lift
+          #             got: :timeout
+          #
+          #        (compared using ==)
+          #
+          #        Diff:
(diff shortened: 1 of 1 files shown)


TonyCTHsu (approved):

spec/datadog/core/remote/component_spec.rb
```diff
@@ -258,8 +258,26 @@
 
     context 'with waiters' do
       it 'unblocks waiters' do
+        skip('Known flaky (assertion below sometimes fails with timeout)')
```
TonyCTHsu, line 261: Perhaps consider only skipping in CI, using `ENV['CI']`

ivoanjo: Up for debate, but in my view:

* If someone is running this locally, either they're trying to modify this class and they should be able to spot the skipped spec in the code with the huge comment or in the rspec output
* Or they don't care about this class and having this flaky spec impact them doesn't seem particularly useful either?

ivoanjo: (I'm going to mark this as resolved to be able to merge the PR, but I'm happy to discuss this and to open a follow-up PR if we want to add this change)

PR #4279, opened 2025-01-10
Serialize span events via a dedicated field

marcotc:
**What does this PR do?**
Before this PR, tracing **span events** were serialized as a JSON string in the span tag `events`. This causes issues with size limit, as tags can only be 5000 characters. This limit is problematic when reporting multiple span events, or when reporting large data in a span events, like a stack trace.

This PR uses a new top-level span field, `span_events`, which supports the full span event representation, without sizing limitation or artificial stringification that was required in the JSON serialization from before.

Because users can already use span events through the `events` tag, and agent support for the top-level field is scheduled for a future agent version, this change check with the agent before sending span events it in the new format.

The **serialization logic is already merged** in https://github.com/DataDog/dd-trace-rb/pull/4218

**Change log entry**
Remove tracing Span Events serialization limitations.

**How to test the change?**
There are system-tests, as well as unit tests for this change.


Steepfile
```diff
@@ -162,30 +156,13 @@ target :datadog do
...
-  library 'grpc'
-  library 'delayed_job'
-  library 'opentelemetry-api'
-  library 'passenger'
-  library 'webmock'
-  library 'graphql'
-  library 'datadog-ci'
+  Dir.children('vendor/rbs').each do |vendor_gem|
+    library vendor_gem
+  end
```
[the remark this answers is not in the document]
marcotc: I think so, probably just a rebase needed.

marcotc:
@p-datadog I implemented the serialization in a previous PR: https://github.com/DataDog/dd-trace-rb/pull/4218

I should have mentioned that in this PR description, let me update it!

marcotc:
The system-tests issues happens because of the way test "skips" and force-tests-list interact: https://github.com/DataDog/system-tests/pull/3904

spec/datadog/tracing/span_spec.rb
```diff
@@ -268,7 +267,9 @@
     subject(:pretty_print) { PP.pp(span) }
 
     it 'output without errors' do
-      expect { pretty_print }.to output.to_stdout
+      Datadog::Tracing.trace('x') { |s| pp s }
```
p-datadog, line 270: You are using `pp` here but removing the require up top?

marcotc: Oops, nice catch!

TonyCTHsu (approved):
Can we refactor the method signature for `native_events_supported` ? 

Prefer keyword argument to positional argument.

lib/datadog/tracing/transport/traces.rb
```diff
@@ -95,17 +96,18 @@ def encode_one(trace)
         module Encoder
           module_function
 
-          def encode_trace(encoder, trace)
+          def encode_trace(encoder, trace, native_events_supported)
```
Strech, line 99: WDYT of named `kwarg` for the options here? It could help this readings

```ruby
@helper_encoder.join([Encoder.encode_trace(@helper_encoder, trace, true)])

# vs

@helper_encoder.join([Encoder.encode_trace(@helper_encoder, trace, native_events_supported: true)])
```

Strech (approved):
Awesome PR, I have just 1 minor thing

lib/datadog/tracing/transport/traces.rb
marcotc: Done!

PR #4283, opened 2025-01-13
DEBUG-3327 DI progress logs

p-datadog:
**What does this PR do?**
Adds logging of dynamic instrumentation operations for troubleshooting: more logging when probes are retrieved from remote config, probes being installed, probes being invoked, events being submitted to the remote worker, remote worker queue size per event type, remote worker working.

**Motivation:**
Currently when DI is not working there is no effective way to troubleshoot it since to the observer it just looks like nothing is happening. The logging added in this PR permits following DI along its steps to see how far it makes through its process.

**Change log entry**
Yes: add optional trace logging to dynamic instrumentation

**Additional Notes:**
The trace logging is conditioned on DD_TRACE_DEBUG environment variable used by the rest of dd-tlrace-rb and also by the newly added internal DI configuration variable. The config variable is intended to be (and is) checked by DI code, but I did not yet want to add another environment variable to enable DI logging specifically because in practice to debug DI we also need to verify that remote config is working, and that uses DD_TRACE_DEBUG.

This PR also adds a Datadog::DI::Logger class which is a facade over the "standard" logger adding a `trace` method. Without this the entire DI test suite would need to have expectations set on the trace output which is a  lot of expectations that usually just add overhead for no benefit.

**How to test the change?**

### DIFF
--- a/lib/datadog/di.rb
+++ b/lib/datadog/di.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_relative 'di/logger'
 require_relative 'di/base'
 require_relative 'di/error'
 require_relative 'di/code_tracker'
--- a/lib/datadog/di.rb
+++ b/lib/datadog/di.rb
@@ -58,10 +58,8 @@ if %w(1 true).include?(ENV['DD_DYNAMIC_INSTRUMENTATION_ENABLED']) # steep:ignore
   #
   # If DI is enabled programmatically, the application can (and must,
   # for line probes to work) activate tracking in an initializer.
-  if %w'1 true debug'.include?(ENV['DD_TRACE_DEBUG'])
-    # We seem to have Datadog.logger here already
-    Datadog.logger.debug("di: activating code tracking")
-  end
+  # We seem to have Datadog.logger here already
+  Datadog.logger.debug("di: activating code tracking")
   Datadog::DI.activate_tracking
 end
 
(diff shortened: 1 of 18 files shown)


lib/datadog/di.rb
```diff
@@ -58,6 +58,10 @@ def component
   #
   # If DI is enabled programmatically, the application can (and must,
   # for line probes to work) activate tracking in an initializer.
+  if %w'1 true debug'.include?(ENV['DD_TRACE_DEBUG'])
+    # We seem to have Datadog.logger here already
+    Datadog.logger.debug("di: activating code tracking")
+  end
```
ivoanjo, line 64: I'm not sure the `DD_TRACE_DEBUG` check is needed -- the logger initialization is already supposed to be looking at this setting? https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/core/configuration.rb#L279 and https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/core/configuration/components.rb#L37 should take care of it.

ivoanjo (commented):
Current draft looks reasonable, left a small note

lib/datadog/di.rb
p-datadog: I removed it.

p-datadog:
This PR now adds a Datadog::DI::Logger facade class that defines `trace` method on it which can be summarily ignored in the test suite (so that I do not need to add expectations on all of the trace output).

lib/datadog/di/configuration/settings.rb
```diff
@@ -188,6 +188,20 @@ def self.add_settings!(base)
                   o.type :bool
                   o.default false
                 end
+
+                # Enable logging of dynamic instrumentation activity.
+                # This is quite verbose.
+                option :trace_logging do |o|
```
Strech, line 194: As a question, since you want to have DI specific setting, does it makes sense to rename it to DI specific name even if internally you rely on another ENV variable. Because why `trace` should be in the name?

Strech (approved):

PR #4287, opened 2025-01-14
Move `ostruct` gem out from test group for Ruby 3.5

y9v:
Apparently `ostruct` gem is removed by `bundle clean` before the memory leak check.

**What does this PR do?**
This PR moves `ostruct` gem out from the test group.

**Motivation:**
CI failing:
https://github.com/DataDog/dd-trace-rb/actions/runs/12713392872/job/35441183658?pr=4236

**Change log entry**
None.

**Additional Notes:**
None.

**How to test the change?**
CI is enough

### DIFF
--- a/ruby-3.5.gemfile
+++ b/ruby-3.5.gemfile
@@ -27,6 +27,7 @@ gem 'json-schema', '< 3' # V3 only works with 2.5+
 gem 'memory_profiler', '~> 0.9'
 gem 'mutex_m'
 gem 'os', '~> 1.1'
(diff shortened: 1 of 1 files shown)


p-datadog (approved):

ivoanjo:
Thanks for fixing this -- I think an alternative could be to tweak the bundle install that we use to install the test group, but I think this option works well too. (Especially given nothing else is using Ruby 3.5 on our tests yet. Seems so far away, 3.5 haha ... the future)

PR #4291, opened 2025-01-15
[APPSEC-56187] Rework WAF span metrics collection

Strech:
**What does this PR do?**

Encapsulates metrics collection into `AppSec::Context` which exposes public API to call security checks. And restructure some other classes responsibilities.

**Motivation:**

The end goal is to reduce abstraction leak from instrumentation and accumulate it in a single place for the future refactoring.

⚠️ This is not the end-goal, but rather mid-state.

**Change log entry**

No.

**Additional Notes:**

In a nutshell

1. `AppSec::Processor::Context` moved into `AppSec::SecurityEngine::Runner` and lost some responsibilities it should not has.
2. A new self-sufficient representation of the WAF run introduced `AppSec::SecurityEngine::Result::{Ok, Match, Error}` with united interface and stats.
3. Security events storage moved into `Context` too to remove the knowledge leak from security engine side.
4. And `AppSec` module is now responsible for knowing is API-Security enabled or not.

**How to test the change?**

If the CI and System-Tests are passing - we are good to go.


lib/datadog/appsec/context.rb
```diff
@@ -34,18 +32,33 @@ def active
...
 
       def run_waf(persistent_data, ephemeral_data, timeout = WAF::LibDDWAF::DDWAF_RUN_TIMEOUT)
-        @waf_runner.run(persistent_data, ephemeral_data, timeout)
+        result = @waf_runner.run(persistent_data, ephemeral_data, timeout)
+
+        @mutex.synchronize do
+          @waf_metrics.timeouts += 1 if result.timeout?
+          @waf_metrics.duration_ns += result.duration_ns
+          @waf_metrics.duration_ext_ns += result.duration_ext_ns
+        end
```
y9v, line 49: what do you think, should we move the mutex inside `WAFMetrics`?

y9v (approved):

lib/datadog/appsec/context.rb
Strech: Yes, I have that in mind, I think for that we need to extract the `Metrics` tool which will do that because `WAFMetrics` is a value-object.

PR #4294, opened 2025-01-15
supported versions workflow: fix parsing and modify gem names

quinna-h:
**What does this PR do?**

- Update to only parse `gemfile.lock` files instead of also `gemfile` files
- change the `aws` contrib to use the `aws-sdk` gem 
- change the `http` contrib to use `net-http` gem
- change typos in `httpx` and `makara`: `nil` means we do not support the integration, `infinity` means we use `latest`

**Motivation:**

Addition to #4210

**Change log entry**

None

**Additional Notes:**

**How to test the change?**

Trigger the [Generate Supported Versions workflow](https://github.com/DataDog/dd-trace-rb/actions/workflows/generate-supported-versions.yml) and look at the resulting `integration_versions.md`


marcotc (approved):

p-datadog (changes requested):
We need the changelog section in the PR description filled out.

I think that none is needed but the PR also does not say how to test it.

quinna-h:
@p-datadog Done, thanks for pointing those out.

PR #4299, opened 2025-01-16
Add debug/byebug dependency to gemfiles

p-datadog:
**What does this PR do?**

Adds dependency on `debug` for Ruby 3.1+, `byebug` for Ruby 3.0 and lower

**Motivation:**

I would like to use a debugger in dd-trace-rb

**Change log entry**
None

**Additional Notes:**

This replaces the earlier byebug addition request (https://github.com/DataDog/dd-trace-rb/pull/4188)

`debug` appears to not work for Rubies below 3.1, for those I added `byebug` as in #4188.

The integration tests need the dependency to be conditional since they have a single gemfile.

rails 5 integration app broke on ruby 2.7 due to use of aliases in database.yml; presumably `psych` somehow got updated (even though we do not have it declared in gemfile), the only candidate I see to bring it in is `rdoc`. Locked `psych` to <4 to work around the alias issue.

**How to test the change?**
N/A


marcotc (approved):

TonyCTHsu:
If you do not need this gem to be added to all the gemfiles, have you consider placing it within the `dev` group from the gemfiles at root?

p-datadog:
Some of the integrations had a dependency on byebug which serves the same function - I replaced it with a dependency on debug to be consistent.

p-datadog:
I also removed `byebug` requires because they existed in a minority of applications. If we want the debugger to be always required I can do that everywhere.

marcotc (approved):

marcotc (approved):

PR #4300, opened 2025-01-16
Add AppSec::ActionHandler to unify action handling

y9v:
**What does this PR do?**
It adds `Datadog::AppSec::ActionHandler` module and unifies blocked response generation.

https://github.com/DataDog/libddwaf/blob/master/UPGRADING.md#action-semantics
https://github.com/DataDog/appsec-diagrams/blob/main/action-handling.md

**Motivation:**
We want to have a simple and unified way of handling `libddwaf` actions.

**Change log entry**
None.

**Additional Notes:**
Since Sinatra and Rails instrumentations always insert `Datadog::AppSec::Contrib::Rack::RequestMiddleware`, we don't need to catch interrupt signals anywhere other than in this Rack Middleware.

**How to test the change?**
CI and manual testing with app generator.


lib/datadog/appsec/response.rb
```diff
@@ -63,56 +42,29 @@ def graphql_response(gateway_multiplex)
...
-                         end
-
-          body = []
-          body << content(content_type)
+        def redirect_response(action_params)
+          status_code = action_params['status_code'].to_i
 
           Response.new(
-            status: options['status_code']&.to_i || 403,
-            headers: { 'Content-Type' => content_type },
```
y9v, line 91: `Content-Type` header is not needed for a redirect response

lib/datadog/appsec/contrib/graphql/appsec_trace.rb
```diff
@@ -16,16 +16,10 @@ def execute_multiplex(multiplex:)
...
 
-            multiplex_return, multiplex_response = Instrumentation.gateway.push('graphql.multiplex', gateway_multiplex) do
+            multiplex_return, = Instrumentation.gateway.push('graphql.multiplex', gateway_multiplex) do
               super
             end
 
-            # Returns an error * the number of queries so that the entire multiplex is blocked
-            if multiplex_response
-              blocked_event = multiplex_response.find { |action, _options| action == :block }
-              multiplex_return = AppSec::Response.graphql_response(gateway_multiplex) if blocked_event
```
vpellan, line 26: If I understand correctly, this PR also removes the creation of responses at framework level (e.g. GraphQL) to handle them all at Rack level ?

lib/datadog/appsec/contrib/graphql/gateway/watcher.rb
```diff
@@ -39,13 +38,15 @@ def watch_multiplex(gateway = Instrumentation.gateway)
...
                       Datadog::AppSec::Event.tag_and_keep!(context, result)
                       context.events << event
+
+                      result.actions.each do |action_type, action_params|
+                        Datadog::AppSec::ActionHandler.handle(action_type, action_params)
+                      end
                     end
 
-                    block = GraphQL::Reactive::Multiplex.publish(engine, gateway_multiplex)
+                    GraphQL::Reactive::Multiplex.publish(engine, gateway_multiplex)
```
vpellan, line 47: In the files contained in the Reactive folder of each contrib (e.g. `appsec/contrib/graphql/reactive/multiplex.rb` at the end of the `subscribe` methods, we still throw `:block` if any action is present, which is already wrong in itself. I wanted to fix it in the stacktrace PR but considering that this PR changes the way we handle actions, you may want to do it in this PR by deleting them ?

lib/datadog/appsec/contrib/graphql/gateway/watcher.rb
```diff
@@ -39,13 +38,15 @@ def watch_multiplex(gateway = Instrumentation.gateway)
 
                       Datadog::AppSec::Event.tag_and_keep!(context, result)
                       context.events << event
+
+                      result.actions.each do |action_type, action_params|
```
vpellan, line 42: So what if the actions hash looks like `{"block"=>{...}, "generate_stack"=>{...}}` ? Do we still want to generate the stacktrace ? If so, I suppose that we should do a first catch in this `each`, save the `block` state, handle all the other actions, then throw `block` again

lib/datadog/appsec/contrib/graphql/appsec_trace.rb
y9v: yes, GraphQL instrumentation should only signal with `INTERRUPT` and leave response blocking to Rack

lib/datadog/appsec/contrib/graphql/gateway/watcher.rb
y9v: This would be done in a separate PR to reduce the amount of changes here. We want to fully get rid of the Reactive Engine, since it only has a instrumentation-local scope and didn't work as intended

lib/datadog/appsec/contrib/graphql/gateway/watcher.rb
y9v: no, we don't need to throw `block`, we want to generate the stack trace and then throw `INTERRUPT`

y9v: but you are right, we don't handle the action precedence correctly now, I will think about it and fix it

y9v: done [1a5404d](https://github.com/DataDog/dd-trace-rb/pull/4300/commits/1a5404df8bcfd62d676ea3590d28861b05ded9cf)

lib/datadog/appsec/actions_handler.rb
```diff
@@ -0,0 +1,33 @@
...
+        block_request(actions_hash['block_request']) if actions_hash.key?('block_request')
+      end
+
+      def block_request(action_params)
+        throw(Datadog::AppSec::Ext::INTERRUPT, action_params)
+      end
+
+      def redirect_request(action_params)
+        throw(Datadog::AppSec::Ext::INTERRUPT, action_params)
+      end
```
Strech, line 24: I understand the intent, but in the current situation the method name is not reflecting the actual action it will lead.

If we replace it with something like (as an example) `interrupt_execution` or maybe `interrupt` it will definitely show what will happen

```ruby
      def handle(actions_hash)
        # handle actions according their precedence
        # stack and schema generation should be done before we throw an interrupt signal
        generate_stack(actions_hash['generate_stack']) if actions_hash.key?('generate_stack')
        generate_schema(actions_hash['generate_schema']) if actions_hash.key?('generate_schema')
        interrupt_execution(actions_hash['redirect_request']) if actions_hash.key?('redirect_request')
        interrupt_execution(actions_hash['block_request']) if actions_hash.key?('block_request')
      end
```

In my opinion, the internals of the `handle` method should be as explicit as possible. Suggested method name is just an example.

lib/datadog/appsec/contrib/graphql/appsec_trace.rb
```diff
@@ -16,16 +16,10 @@ def execute_multiplex(multiplex:)
 
             gateway_multiplex = Gateway::Multiplex.new(multiplex)
 
-            multiplex_return, multiplex_response = Instrumentation.gateway.push('graphql.multiplex', gateway_multiplex) do
+            multiplex_return, = Instrumentation.gateway.push('graphql.multiplex', gateway_multiplex) do
```
Strech, line 19: It's a matter of preference, but it's bugging me 😆 Maybe we can go more explicit?

```suggestion
            multiplex_return, _ = Instrumentation.gateway.push('graphql.multiplex', gateway_multiplex) do
```

lib/datadog/appsec/contrib/rack/request_body_middleware.rb
```diff
@@ -24,15 +24,15 @@ def call(env)
             # TODO: handle exceptions, except for @app.call
 
             http_response = nil
-            block_actions = catch(::Datadog::AppSec::Ext::INTERRUPT) do
+            block_action_params = catch(::Datadog::AppSec::Ext::INTERRUPT) do
```
Strech, line 27: As a suggestion, we catching interrupt here, do we want to call it `interrupt_params`? Because we might not block, but redirect.

lib/datadog/appsec/response.rb
```diff
@@ -19,100 +19,38 @@ def to_rack
...
-          multiplex_return = []
-          gateway_multiplex.queries.each do |query|
-            # This method is only called in places where GraphQL-Ruby is already required
-            query_result = ::GraphQL::Query::Result.new(
-              query: query,
-              values: JSON.parse(content('application/json'))
-            )
-            multiplex_return << query_result
-          end
+        def build(action_params, http_accept_header)
```
Strech, line 23: IMHO, we should signal with the name, that we can convert interrupt params into the response, hence, I would like to suggest naming like

1. `from_interrupt_params`
2. `from_interrupt`
3. `from_interrupt_signal`

or any other than `build` which is generic and doesn't reflect the fact that it will work solely with actions from the interrupt. In addition, do we want to convert accept headers into named kwarg?

```ruby
def from_interrupt_params(params, http_accept_header:)
```

sig/datadog/appsec/actions_handler.rbs
```diff
@@ -0,0 +1,17 @@
+module Datadog
+  module AppSec
+    module ActionsHandler
+      def handle: (::Hash[::String, ::Hash[::String, ::String]] actions_hash) -> void
```
Strech, line 4: Does it makes sense to define the `SecurityEngine::Runner::Result.action` type and use it here? If we break the contract, steep will complain, because now we can break it and never fine out

spec/datadog/appsec/actions_handler_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    end
+
+    let(:block_request_action) do
+      { 'block_request' => { 'status_code' => '403', 'type' => 'auto' } }
+    end
+
+    it 'calls generate_stack with action parameters' do
+      expect(described_class).to(
+        receive(:generate_stack).with(generate_stack_action['generate_stack']).and_call_original
+      )
```
Strech, line 26: Could this formatting improve the situation? Like that we can keep the main statement in a single line and move details on the next line

```suggestion
      expect(described_class).to receive(:generate_stack)
        .with(generate_stack_action['generate_stack'])
        .and_call_original
```

Strech (approved):
I think the overall change looks good and solid to use as an intermediate step to encapsulate the leaked logic 👍🏼

lib/datadog/appsec/contrib/graphql/appsec_trace.rb
y9v: it bugs me as well, but this is due to rubocop :( I will check if we could add disable this

lib/datadog/appsec/response.rb
y9v: this is unfortunately not only for interruption, but also for stack trace generation and other actions. How about `from_actions`?

spec/datadog/appsec/actions_handler_spec.rb
y9v: fixed [5a05cab](https://github.com/DataDog/dd-trace-rb/pull/4300/commits/5a05cabb5f260fade144f75e751bc1d3b6e0964f)

lib/datadog/appsec/response.rb
y9v: sorry, misread your comment first, `from_interrupt_params` is better

vpellan (approved):
LGTM!

lib/datadog/appsec/actions_handler.rb
y9v: fixed: [558f96f](https://github.com/DataDog/dd-trace-rb/pull/4300/commits/558f96fff49d40d5e7064f33900cc54b1174704a)

lib/datadog/appsec/contrib/rack/request_body_middleware.rb
y9v: fixed [558f96f](https://github.com/DataDog/dd-trace-rb/pull/4300/commits/558f96fff49d40d5e7064f33900cc54b1174704a)

sig/datadog/appsec/actions_handler.rbs
y9v: Used the existing `Result::actions` and `Result.action` types [15afd41](https://github.com/DataDog/dd-trace-rb/pull/4300/commits/15afd41c8ca51c4d8bd4ca6b79dbbf0b65844487)

PR #4303, opened 2025-01-20
[NO-TICKET] Fix breaking application boot due to concurrency issue in tracing

### LINKED ISSUE #2851: ddtrace crashes application on load
**Current behaviour**

ddtrace occasionally causes applications to fail on boot with error: 
```
can't add a new key into hash during iteration
/usr/local/bundle/gems/ddtrace-0.54.2/lib/ddtrace/contrib/extensions.rb:131:in `instrument'
/usr/local/bundle/gems/ddtrace-0.54.2/lib/ddtrace/contrib/rails/framework.rb:116:in `activate_action_pack!'
/usr/local/bundle/gems/ddtrace-0.54.2/lib/ddtrace/contrib/rails/framework.rb:51:in `block in setup'
/usr/local/bundle/gems/ddtrace-0.54.2/lib/ddtrace/configuration.rb:48:in `configure'
/usr/local/bundle/gems/ddtrace-0.54.2/lib/ddtrace/contrib/extensions.rb:40:in `configure'
/usr/local/bundle/gems/ddtrace-0.54.2/lib/ddtrace/contrib/rails/framework.rb:39:in `setup'
/usr/local/bundle/gems/ddtrace-0.54.2/lib/ddtrace/contrib/rails/patcher.rb:114:in `setup_tracer'
/usr/local/bundle/gems/ddtrace-0.54.2/lib/ddtrace/contrib/rails/patcher.rb:108:in `block in after_intialize'
/usr/local/bundle/gems/ddtrace-0.54.2/lib/ddtrace/utils/only_once.rb:26:in `block in run'
/usr/local/bundle/gems/ddtrace-0.54.2/lib/ddtrace/utils/only_once.rb:21:in `synchronize'
/usr/local/bundle/gems/ddtrace-0.54.2/lib/ddtrace/utils/only_once.rb:21:in `run'
/usr/local/bundle/gems/ddtrace-0.54.2/lib/ddtrace/contrib/rails/patcher.rb:105:in `after_intialize'
/usr/local/bundle/gems/ddtrace-0.54.2/lib/ddtrace/contrib/rails/patcher.rb:100:in `block in patch_after_intialize'
/usr/local/bundle/gems/activesupport-6.1.6/lib/active_support/lazy_load_hooks.rb:73:in `instance_eval'
/usr/local/bundle/gems/activesupport-6.1.6/lib/active_support/lazy_load_hooks.rb:73:in `block in execute_hook'
/usr/local/bundle/gems/activesupport-6.1.6/lib/active_support/lazy_load_hooks.rb:61:in `with_execution_control'
/usr/local/bundle/gems/activesupport-6.1.6/lib/active_support/lazy_load_hooks.rb:66:in `execute_hook'
/usr/local/bundle/gems/activesupport-6.1.6/lib/active_support/lazy_load_hooks.rb:52:in `block in run_load_hooks'
/usr/local/bundle/gems/activesupport-6.1.6/lib/active_support/lazy_load_hooks.rb:51:in `each'
/usr/local/bundle/gems/activesupport-6.1.6/lib/active_support/lazy_load_hooks.rb:51:in `run_load_hooks'
/usr/local/bundle/gems/railties-6.1.6/lib/rails/application/finisher.rb:140:in `block in <module:Finisher>'
/usr/local/bundle/gems/railties-6.1.6/lib/rails/initializable.rb:32:in `instance_exec'
/usr/local/bundle/gems/railties-6.1.6/lib/rails/initializable.rb:32:in `run'
/usr/local/bundle/gems/railties-6.1.6/lib/rails/initializable.rb:61:in `block in run_initializers'
```

This appears to be due to a concurrency issue with the `instrumented_integrations` hash. Multiple threads are accessing the hash without thread safe operations. `Datadog::Diagnostics::EnvironmentLogger.log!` iterates through `instrumented_integrations` in a separate worker thread. At the same time,  `Datadog::Contrib::Extensions#instrument` is adding keys to the same `instrumented_integrations` hash object. Current workaround is to disable ddtrace startup logs.

**Expected behaviour**

ddtrace should not cause application to crash

**Steps to reproduce**

Difficult to reproduce since it's timing related but it may be possible by artificially triggering a trace write during startup.

**Environment**

* **ddtrace version:** 0.54.2

config/initializers/datadog.rb
```rb
Datadog.configure do |c|
  c.tracer(hostname: ENV["DD_AGENT_HOST"] || "localhost",
           port: ENV["DD_TRACE_AGENT_PORT"] || 8126,
           debug: ENV["DD_TRACE_DEBUG"] == "true")

  c.tracer.sampler = Datadog::PrioritySampler.new(
    post_sampler: Datadog::Sampling::RuleSampler.new(
      [
        Datadog::Sampling::SimpleRule.new(service: c.service, sample_rate: 1.0000),
      ]
    )
  )

  c.tracer.partial_flush.enabled = true
  c.use(:aws, service_name: c.service)
  c.use(:http, split_by_domain: true)
  c.use(:httprb, split_by_domain: true)
  c.use(:grpc, service_name: c.service)
  c.use(:kafka, service_name: c.service)
  c.use(:rack, service_name: c.service, web_service_name: "#{c.service}-nginx", request_queuing: true)
  c.use(:redis, service_name: c.service)
  c.use(:rails, service_name: c.service)
  c.use(:rest_client, service_name: c.service)
  c.use(:sidekiq, service_name: c.service, client_service_name: c.service)
  c.use(:sneakers, service_name: c.service)
end
```

spec/spec_helper.rb
```rb
Datadog.configure do |c|
  c.tracer.enabled = true
  c.ci_mode.enabled = true
  c.use :rspec, operation_name: "ruby.rspec.integration"
end
```

* **Ruby version:** 3.0.5
* **Operating system:** debian:bullseye-slim
* **Relevant library versions:** Rails 6.1.6

ivoanjo:
**What does this PR do?**

This PR fixes #2851 . As explained in the original bug report, there's a concurrency issue that can be triggered by the environment logger running concurrently with activating an instrumentation.

**Motivation:**

Fix a long-standing concurrency issue.

**Change log entry**

Yes. Fix breaking application boot due to concurrency issue in tracing 

**Additional Notes:**

To fix this issue, I've made two changes:

1. Introduced a mutex to protect the `@instrumented_integrations`, thus making sure no two threads can be touching it at the same time

2. Took advantage of the fact that `#instrumented_integrations` was marked as private, and only being used by the environment logger and telemetry (e.g. read-only usage) to return a copy of the data.

This way, we can safely iterate on the data while reconfiguration is happening concurrently.

I believe the overhead of this lock is negligible, since we don't need to read this information very often.

**How to test the change?**

I wrote a reproducer to be able to see this issue easily.

The reproducer is in two parts -- `concurrent_bug_repo.rb`:

```ruby
require 'datadog'

Datadog.configure do |c|
  c.tracing.instrument(:http, split_by_domain: true)
end

$sync_with_logger = Queue.new

Thread.new do
  Datadog::Tracing::Diagnostics::EnvironmentLogger.collect_and_log!
  puts "Background thread finished!"
end

$sync_with_logger.pop

Datadog.configure do |c|
  c.tracing.instrument(:rake, split_by_domain: true)
end
```

and the following change to the environment logger:

```diff
diff --git a/lib/datadog/tracing/diagnostics/environment_logger.rb b/lib/datadog/tracing/diagnostics/environment_logger.rb
index d96cfa14c6..427861bff7 100644
--- a/lib/datadog/tracing/diagnostics/environment_logger.rb
+++ b/lib/datadog/tracing/diagnostics/environment_logger.rb
@@ -128,7 +128,15 @@ module Datadog
           end

           def collect_integrations_settings!
+            sync_once = $sync_with_logger
+
             instrumented_integrations.each_with_object({}) do |(name, integration), result|
+              if sync_once
+                sync_once << true
+                sync_once = nil
+                sleep 5
+              end
+
               integration.configuration.to_h.each do |setting, value|
                 next if setting == :tracer # Skip internal objects
```

That is, I specifically injected a `sleep` in the environment logger, and made it run on a background thread, triggering the exact same issue that had been plaguing us for years.

Running the above change without my PR will trigger the issue, and with my PR it no longer does -- since the environment logger now gets its own copy of the instrumented integrations.

### DIFF
--- a/lib/datadog/tracing/contrib/extensions.rb
+++ b/lib/datadog/tracing/contrib/extensions.rb
@@ -110,6 +110,13 @@ module Datadog
           module Settings
             InvalidIntegrationError = Class.new(StandardError)
 
+            # Used to avoid concurrency issues between registering integrations (e.g. mutation) and reporting the
+            # current integrations for logging/debugging/telemetry purposes (e.g. iteration) in the
+            # `@instrumented_integrations` hash.
+            #
+            # See https://github.com/DataDog/dd-trace-rb/issues/2851 for details on the original issue.
+            INSTRUMENTED_INTEGRATIONS_LOCK = Mutex.new
+
             def self.included(base)
               base.class_eval do
                 settings :contrib do
@@ -161,7 +168,10 @@ module Datadog
                 configuration_name = options[:describes] || :default
                 filtered_options = options.reject { |k, _v| k == :describes }
                 integration.configure(configuration_name, filtered_options, &block)
-                instrumented_integrations[integration_name] = integration
+                INSTRUMENTED_INTEGRATIONS_LOCK.synchronize do
+                  @instrumented_integrations ||= {}
+                  @instrumented_integrations[integration_name] = integration
+                end
 
                 # Add to activation list
                 integrations_pending_activation << integration
@@ -192,14 +202,16 @@ module Datadog
               @integrations_pending_activation ||= Set.new
             end
 
+            # This method is only for logging/debugging/telemetry purposes (e.g. iteration) in the
+            # `@instrumented_integrations` hash.
             # @!visibility private
             def instrumented_integrations
-              @instrumented_integrations ||= {}
+              INSTRUMENTED_INTEGRATIONS_LOCK.synchronize { (@instrumented_integrations&.dup || {}).freeze }
             end
 
             # @!visibility private
             def reset!
-              instrumented_integrations.clear
+              INSTRUMENTED_INTEGRATIONS_LOCK.synchronize { @instrumented_integrations&.clear }
               super
             end
 
--- a/spec/datadog/tracing/contrib/extensions_spec.rb
+++ b/spec/datadog/tracing/contrib/extensions_spec.rb
@@ -26,6 +26,10 @@ RSpec.describe Datadog::Tracing::Contrib::Extensions do
     before { registry.add(integration_name, integration) }
   end
 
+  before do
+    allow(Datadog.logger).to receive(:warn)
+  end
+
   context 'for' do
     describe Datadog do
       describe '#configure' do
@@ -243,6 +247,12 @@ RSpec.describe Datadog::Tracing::Contrib::Extensions do
               end
             end
           end
+        end
+
+        describe '#instrumented_integrations' do
+          subject(:instrumented_integrations) { settings.instrumented_integrations }
+
+          it { is_expected.to be_frozen }
         end
       end
     end


Strech (approved):
LGTM, but I have a question, if we dup the return of `instrumented_integrations`, do we freeze it to show an intention that this set should never be modified (even tho you can't modify the dup to affect original)

ivoanjo:
> do we freeze it to show an intention that this set should never be modified (even tho you can't modify the dup to affect original)

Yes, that's exactly the intention -- this method is documented as private and as far as I could find it it's only used in our codebase for logging.

But... if somehow I missed a spot or anyone out there has a weird patch set on top of the original method, this will flag the issue when they attempt to mutate. (Without the freeze, they'd change the copy, which wouldn't affect the original and perhaps introduce very subtle bugs)

p-datadog (approved):

PR #4307, opened 2025-01-22
[NO-TICKET] Update README section on how to add new dependency group

Strech:
**What does this PR do?**

Adds new section in the readme and remove duplicated section about how to add new test group into your setup, step by step.

**Motivation:**

I was able to do it only after the call, it makes sense to share that knowledge

**Change log entry**
No.

**Additional Notes:**

Might need some wording advice

**How to test the change?**

Follow instructions, and if it works - everything is correct then.

### DIFF
--- a/docs/DevelopmentGuide.md
+++ b/docs/DevelopmentGuide.md
@@ -44,34 +44,6 @@ New tests should be written as RSpec tests in the `spec/datadog` folder. Test fi
 
 All changes should be covered by a corresponding RSpec tests. Unit tests are preferred, and integration tests are accepted where appropriate (e.g. acceptance tests, verifying compatibility with datastores, etc) but should be kept to a minimum.
 
-**Considerations for CI**
-
-All tests should run in CI. When adding new `_spec.rb` files, you may need to add rake task to ensure your test file is run in CI.
-
- - Ensure that there is a corresponding Rake task defined in `Rakefile` under the `spec` namespace, whose pattern matches your test file. For example
-
- ```ruby
-   namespace :spec do
-     RSpec::Core::RakeTask.new(:foo) do |t, args|
-       t.pattern = "spec/datadog/tracing/contrib/foo/**/*_spec.rb"
-       t.rspec_opts = args.to_a.join(' ')
-     end
-   end
- ```
-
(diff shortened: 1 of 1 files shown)


docs/DevelopmentGuide.md
```diff
@@ -149,6 +121,88 @@ bundle exec rake dependency:lock['/app/gemfiles/ruby_3.3_stripe_*.gemfile']
 bundle exec rake dependency:lock['/app/gemfiles/ruby_3.3_stripe_latest.gemfile']
 ```
 
+**How to add new dependency group**
+
+1. Choose the Ruby runtime and group name for your tests. When defining a new group, follow the format `scope:group`.
+For example, if you want tests to run only on Ruby 3.3 for profiling, you can define this in the [`Matfixfile`](../Matrixfile).
```
ivoanjo, line 127: Typo:

```suggestion
For example, if you want tests to run only on Ruby 3.3 for profiling, you can define this in the [`Matrixfile`](../Matrixfile).
```

ivoanjo, line 127: Also I think you started using "profiling" as an example... and then maybe changed your mind and used tracing below? :eyes:

docs/DevelopmentGuide.md
```diff
@@ -149,6 +121,88 @@ bundle exec rake dependency:lock['/app/gemfiles/ruby_3.3_stripe_*.gemfile']
 bundle exec rake dependency:lock['/app/gemfiles/ruby_3.3_stripe_latest.gemfile']
 ```
 
+**How to add new dependency group**
```
ivoanjo, line 124: I wonder if we should also leave a warning to avoid creating new groups when possible... In particular, we're exploding the number of files in the repo + need to be updated + that show up on a diff (making github not able to render the diff anymore in some cases).

So something like "Consider reusing an existing group if possible, as maintaining too many groups causes trouble in CI/git"?

ivoanjo (commented):
Left a few comments!

docs/DevelopmentGuide.md
Strech: Fixed

docs/DevelopmentGuide.md
Strech: Done

y9v (approved):

ivoanjo (approved):
:+1: LGTM

TonyCTHsu (approved):

PR #4309, opened 2025-01-22
[APPSEC-56456] Make telemetry an AppSec component service

Strech:
**What does this PR do?**

Make configured AppSec telemetry a component level accessibility service.

**Motivation:**

It's a smell that inside the component we have to pass the telemetry component which is configured on AppSec component configuration/re-configuretion time. Instead it should be a component level visibility service which could be used by component services directly.

**Change log entry**

No.

**Additional Notes:**

None.

**How to test the change?**

CI is sufficient.


lib/datadog/appsec/component.rb
```diff
@@ -73,21 +73,26 @@ def create_processor(settings, telemetry)
...
       def reconfigure(ruleset:, telemetry:)
         @mutex.synchronize do
-          new = Processor.new(ruleset: ruleset, telemetry: telemetry)
+          new_processor = Processor.new(ruleset: ruleset, telemetry: telemetry)
+
+          if new_processor && new_processor.ready?
+            old_processor = @processor
+
+            @telemetry = telemetry
+            @processor = new_processor
```
y9v, line 93: I believe this should work too:
```ruby
old_processor, @processor = @processor, new_processor
```

y9v: but it's not really readable :)

y9v (approved):

lib/datadog/appsec/component.rb
Strech: I think it's not.

PR #4318, opened 2025-01-24
[NO-TICKET] Ignore docs changes in system tests workflow

Strech:
**What does this PR do?**

Ignore `docs/**` in system tests workflow.

**Motivation:**

An action item from the last guild meeting, we don't need to run whole test suite for documentation changes. This is exactly [documentation-like changes](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-excluding-paths).

**Change log entry**

No.

**Additional Notes:**

My auto-correction made a few adjustments for consistency, I decide to keep it.

**How to test the change?**

Open PR with documentation changes only, system tests should not be present. Not sure about the required checks, probably we need to configure something there too.

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -4,9 +4,22 @@ on:
   push:
     branches:
       - "**"
+    paths-ignore:
+      - ".circleci/**"
+      - ".gitlab/**"
+      - "appraisal/**"
+      - "benchmarks/**"
+      - "docs/**"
+      - "gemfiles/**"
+      - "integration/**"
+      - "sig/**"
+      - "spec/**"
+      - "suppressions/**"
+      - "tools/**"
+      - "vendor/**"
   workflow_dispatch: {}
   schedule:
-    - cron:  '00 04 * * 2-6'
+    - cron: "00 04 * * 2-6"
 
 env:
   REGISTRY: ghcr.io
@@ -33,7 +46,7 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          repository: 'DataDog/system-tests'
+          repository: "DataDog/system-tests"
           ref: ${{ env.SYSTEM_TESTS_REF }}
           persist-credentials: false
       - name: Login to Docker Hub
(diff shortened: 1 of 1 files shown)


.github/workflows/system-tests.yml
```diff
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - "**"
+    paths-ignore:
+      - "docs/**"
```
ivoanjo, line 8: This is a nice idea! TBH I think a lot more things could be added there -- `.circleci`/`.gitlab`/`appraisal`/`benchmarks`/`gemfiles`/`integration`/`sig`/`spec`/`suppressions`/`tools`/`vendor`.

None of these have any impact on released code, and thus would not impact system-tests.

Up to you if you want to do that as well -- I understand this would increase the scope of this PR a lot and you may not wanna ;)

ivoanjo (approved):
:+1: LGTM nice!

.github/workflows/system-tests.yml
```diff
@@ -4,9 +4,22 @@ on:
   push:
     branches:
       - "**"
+    paths-ignore:
+      - ".circleci/**"
+      - ".gitlab/**"
+      - "appraisal/**"
+      - "benchmarks/**"
+      - "docs/**"
+      - "gemfiles/**"
```
p-datadog, line 13: Dependencies (for example, if they are missing) affect behavior, therefore I think changes to dependencies should rerun tests.

.github/workflows/system-tests.yml
```diff
@@ -4,9 +4,22 @@ on:
...
+      - "integration/**"
+      - "sig/**"
+      - "spec/**"
+      - "suppressions/**"
+      - "tools/**"
+      - "vendor/**"
```
p-datadog, line 19: I think this should be `vendor/rbs/**`.

p-datadog (changes requested):
Requesting changes since automerge is enabled.

.github/workflows/system-tests.yml
marcotc: I will affect `dd-trace-rb` tests, but not `system-tests` testing, given system tests installs the copyof `dd-trace-rb` for a specific commit, which only depends on dependencies from https://github.com/DataDog/dd-trace-rb/blob/master/datadog.gemspec.

marcotc (approved):

.github/workflows/system-tests.yml
ivoanjo: Yup, +1 on what Marco said

.github/workflows/system-tests.yml
ivoanjo: For similar reasons as the `gemfiles` above, system-tests do not (and should not?) depend on anything in the `vendor/` folder, which is only used for other kinds of testing.

ivoanjo (approved):

PR #4320, opened 2025-01-24
[PROF-9476] Profiler cleanups extracted from managed string storage branch

ivoanjo:
**What does this PR do?**

This PR contains a bunch of cleanups that I was accumulating in the managed string storage branch (e.g. the next version of #3628).

**Motivation:**

I decided to extract these changes from my branch as they were becoming quite a lot. The managed string storage work already touches a lot of code, so this way we can review and merge these separately.

**Change log entry**

None.

**Additional Notes:**

I've broken this PR into separate commits, so it's clearer why each change is going in.

**How to test the change?**

Our existing test coverage already covers these changes.

### DIFF
--- a/benchmarks/profiler_memory_sample_serialize.rb
+++ b/benchmarks/profiler_memory_sample_serialize.rb
@@ -54,32 +54,37 @@ class ProfilerMemorySampleSerializeBenchmark
         timeline_enabled: false,
       )
     }
+    @recorder = @recorder_factory.call
+    @retained_objs = create_objects
+  end
+
+  def create_objects
+    samples_per_second = 100
+    simulate_seconds = 60
+    retained_objs = []
+
+    (samples_per_second * simulate_seconds).times do |i|
+      obj = sample_object(@recorder, i % 400)
+      retained_objs << obj if (i % @retain_every).zero?
+    end
+
+    GC.start unless @skip_end_gc
+
+    retained_objs
   end
 
   def run_benchmark
     Benchmark.ips do |x|
-      benchmark_time = VALIDATE_BENCHMARK_MODE ? { time: 0.01, warmup: 0 } : { time: 30, warmup: 2 }
+      benchmark_time = VALIDATE_BENCHMARK_MODE ? { time: 0.01, warmup: 0 } : { time: 10, warmup: 2 }
       x.config(
         **benchmark_time,
       )
 
-      x.report("sample+serialize #{ENV['CONFIG']} retain_every=#{@retain_every} heap_samples=#{@heap_samples_enabled} heap_size=#{@heap_size_enabled} heap_sample_every=#{@heap_sample_every} skip_end_gc=#{@skip_end_gc}") do
-        recorder = @recorder_factory.call
-        samples_per_second = 100
-        simulate_seconds = 60
-        retained_objs = []
-
-        (samples_per_second * simulate_seconds).times do |i|
-          obj = sample_object(recorder, i % 400)
-          retained_objs << obj if (i % @retain_every).zero?
-        end
-
-        GC.start unless @skip_end_gc
-
-        recorder.serialize
+      x.report("heap serialize #{ENV['CONFIG']} retain_every=#{@retain_every} heap_samples=#{@heap_samples_enabled} heap_size=#{@heap_size_enabled} heap_sample_every=#{@heap_sample_every} skip_end_gc=#{@skip_end_gc}") do
+        @recorder.serialize
       end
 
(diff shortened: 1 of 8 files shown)


ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -745,12 +727,17 @@ static void commit_recording(heap_recorder *heap_recorder, heap_record *heap_rec
...
-  // Update object_records with the data for this new recording
-  object_record_update_data update_data = (object_record_update_data) {
-    .heap_recorder = heap_recorder,
-    .recording = active_recording,
-  };
-  st_update(heap_recorder->object_records, active_recording->obj_id, update_object_record_entry, (st_data_t) &update_data);
+  int existing_error = st_update(heap_recorder->object_records, active_recording->obj_id, update_object_record_entry, (st_data_t) active_recording);
+  if (existing_error) {
+    object_record *existing_record = NULL;
+    st_lookup(heap_recorder->object_records, active_recording->obj_id, (st_data_t *) &existing_record);
```
AlexJF, line 733: In a previous commit you added a todo mentioning we never lookup on `object_records`. Technically this code is doing a lookup when looking for duplicates. If you actually wanted to convert `object_records` to a list this section would become O(N).

AlexJF (approved):

ext/datadog_profiling_native_extension/heap_recorder.c
ivoanjo: Good points!

So there's two things here:

1. The `st_update` itself is giving us the duplicate detection "for free", so I think it makes sense to take advantage of it in a "oops-just-in-case" kinda thing. But if we moved to an arraylist-kinda thing, I think we would not want to pay the cost of checking for duplicates.

2. My thinking behind introducing the `st_lookup` here is that, while less efficient than doing the work inside `update_object_record_entry` (as we did before) or restoring `object_record_update_data` and writing the result there, overall this situation is a "never expected to happen" -- and if it happens the next step is to log an error and stop profiling anyway. So overall, by being slightly more inefficient in the "never supposed to happen codepath" the code seems a bit shorter/simpler, which seems like a nice win?

Happy to give it a second pass if you're not convinced ;D

benchmarks/profiler_memory_sample_serialize.rb
```diff
@@ -12,7 +12,7 @@
 # This benchmark measures the performance of sampling + serializing memory profiles. It enables us to evaluate changes to
 # the profiler and/or libdatadog that may impact both individual samples, as well as samples over time.
 #
-METRIC_VALUES = { 'cpu-time' => 0, 'cpu-samples' => 0, 'wall-time' => 0, 'alloc-samples' => 1, 'timeline' => 0 }.freeze
+METRIC_VALUES = { 'cpu-time' => 0, 'cpu-samples' => 0, 'wall-time' => 0, 'alloc-samples' => 1, 'timeline' => 0, 'heap_sample' => true }.freeze
```
Strech, line 15: Just a question `alloc-samples`, but `heap_sample`, is there a reason for format change?

Strech (approved):
Hard to judge for me, just spot a small inconsistency.

But 👍🏼

ext/datadog_profiling_native_extension/heap_recorder.c
AlexJF: Yeah makes sense. I was just wondering if we'd ever want to lose the automatic oops check at any point 😅 Sounds like being blind to that may be dangerous (but we could definitely scope that to some integration test or something).

ivoanjo: Indeed. TBH my adding a note that we're not using it as a hash is more of a "call it out explicitly so someone reading through doesn't second-guess themselves when they spot it" and less of a "plan to do something about it".

In particular, I don't want to import an arraylist-kinda thing to the codebase, and libdatadog doesn't have on yet, so there isn't even something reasonable that we could compare against.

Furthermore, on a fresh head, if we know Ruby never reuses ids and always increases them, we can actually get the duplicate check somewhat for free even in an arraylist setup if we record the latest id we saw. By definition if we see ids going backwards, something is off/changed in Ruby that will mess our assumptions. (Downside: Unit testing would maybe be a bit more annoying, since we'd need to make sure we pass in objects in the correct order)

I'm still fingers-crossed that in the future we can move to https://bugs.ruby-lang.org/issues/19783 on modern Rubies and then we'd be able to stop relying on object id tricks from then on.

PR #4322, opened 2025-01-24
ci: upload gem from GitLab release stage

brettlangdon:
**What does this PR do?**
Automates Ruby gem publishing by uploading from GitLab pipeline.

**Motivation:**
Remove manual step of uploading the gem on release.

**Change log entry**

None.

**Additional Notes:**

This is modeled after the dd-trace-java and dd-trace-py processes which assume that a Git tag/GitHub Release are created first (that is the trigger to "do a release"). Rather than existing Ruby process which first builds/checks tests, then uploads the Gem, and then finally creates the GitHub Release. We can adjust this workflow, but wanted to give us a place to start, aligning with others was simplest.

Since this is done in a final release stage, it does mean that the GitLab pipeline needs to fully run and complete before the Gem will be uploaded. Any checks/tests added to GitLab will become a blocker to the Gem being uploaded.

This process also uses the built Gem that we already do in the pipeline and use in the OCI/lib-injection artifacts.

We have not added any additional checks on GitHub commit status, we can/should do that. I'd love to add to the one pipeline so it is implemented consistently for all repos.

I have not yet added the necessary RubyGem token to the process.

**How to test the change?**

### DIFF
--- a/.gitlab/release.yml
+++ b/.gitlab/release.yml
@@ -3,11 +3,9 @@ release_ruby_gem:
   image: $DOCKER_REGISTRY/images/mirror/ruby:3.2.2  
   tags: ["arch:amd64"]
   only:
-    # v2.10.0
-    # v2.10.1
-    # v2.10.0rc0
-    # v2.10.0rc5
-    - /^v[0-9]+\.[0-9]+\.[0-9]+(rc[0-9]+)?$/
+    # Version pattern from Rubygems: https://github.com/rubygems/rubygems/blob/969f32050f0dc024c411fef7cbe25b1aaed16983/lib/rubygems/version.rb#L158
+    # with a added `v` prefix.
+    - /^v[0-9]+(?>\.[0-9a-zA-Z]+)*(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/
   before_script:
     - |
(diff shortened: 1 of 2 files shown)


.gitlab/release.yml
```diff
@@ -0,0 +1,19 @@
+release_ruby_gem:
+  stage: release
+  image: $DOCKER_REGISTRY/images/mirror/ruby:3.2.2  
+  tags: ["arch:amd64"]
+  only:
+    # v2.10.0
+    # v2.10.1
+    # v2.10.0rc0
+    # v2.10.0rc5
+    - /^v[0-9]+\.[0-9]+\.[0-9]+(rc[0-9]+)?$/
```
marcotc, line 10:
```suggestion
    # Version pattern from Rubygems: https://github.com/rubygems/rubygems/blob/969f32050f0dc024c411fef7cbe25b1aaed16983/lib/rubygems/version.rb#L158
    # with a added `v` prefix.
    - /^v[0-9]+(?>\.[0-9a-zA-Z]+)*(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$/
```

marcotc: Ruby has a slightly different version pattern, allowing alphanumeric characters in all positions, as well as separating the `rc` part with a `.`.

marcotc (approved):

PR #4331, opened 2025-01-30
[PROF-9476] Reduce heap profiling overhead by using libdatadog's managed string storage 

ivoanjo:
**What does this PR do?**

This PR changes the heap profiler to use the new libdatadog managed string storage API added in https://github.com/DataDog/libdatadog/pull/725 . (And thanks to Alex for starting this work in https://github.com/DataDog/dd-trace-rb/pull/3628 )

Prior to this PR, the heap profiler was doing all of the work of keeping stack traces around for the objects being tracked. This involved keeping copies of all strings for path / function names, and also having a clever deduplication mechanism that allowed us to dedup stack traces without first copying all of the strings.

The libdatadog managed string storage allows us to simplify this work by giving us a string table we can use to store strings and trade them for long-lived ids. Then these ids can be used then adding samples to libdatadog, instead of direct pointers to the strings.

This has a bunch of advantages:
1. We were not previously deduplicating strings in the heap profiler, only whole stacks.

   Thus if we had `/path/a.rb:method_a` -> `/path/b.rb:method_b` -> ` /path/c.rb:method_c` multiple times, we would deduplicate them. But if we also had `/path/a.rb:method_a` -> `/path/b.rb:method_b`, then we'd store yet another copy of these strings. This is no longer the case when using the new API.

2. Representing stacks as arrays of ids means deduplication, hashing and comparison is much faster -- we're only comparing ids and not the contents of strings.

3. When feeding stacks again to libdatadog, it's cheaper to use ids since libdatadog doesn't need to keep re-hashing the strings to de-duplicate them; it can use the ids to do that.

This does mean that this PR is a bit... on the noisy side in terms of changes to the heap profiler. Specifically, it takes advantage of most of the simplification opportunities that managed string stable allows:

* It merges `heap_record` and `heap_stack` structures as they no longer need to exist independently, and then rips out as much code as possible based on that merge
* It simplifies stack comparison and hashing

_Update: As discussed, I've pushed a commit to set `experimental_heap_sample_rate` to 1, raising the heap sampling rate from its previous default of 10._

**Motivation:**

The key objective of this work is allowing us to reduce the memory and cpu overhead of heap profiling.

~~Earlier versions of this PR already showed reduced memory usage. I decided to open this PR already and do benchmarking in parallel, so I don't yet have updated numbers yet.~~

See below.

**Change log entry**

Yes. Reduce overhead of heap profiling

**Additional Notes:**

Currently CI is broken for this PR because it depends on https://github.com/DataDog/libdatadog/pull/844 which has not yet been merged into libdatadog.

Obviously I plan to make sure that PR is in libdatadog, and that master is using the latest libdatadog before merging this PR, but I don't think it makes sense to hold off on review until that's the case.

**How to test the change?**

The existing test coverage was already quite reasonable. I added a few more tests to make sure that our stack recording/retrieval/deduping was in good shape, as well as a benchmark for the efficiency of the interning operations.

---

**Benchmarks**:

[I wrote up my results in this doc (Datadog-internal)](https://docs.google.com/document/d/1jFgdrEoYlC1QSBzIpEz2SIUlJYF9GJOz9cUXJVEUQcY/edit?tab=t.0). Here's a quick recap of what I saw.

_High load gitlab_:

Here, we’re comparing 4 configurations:
* baseline: Unmodified application, no profiling at all
* only-profiling-heap: Application running cpu + wall-time + allocation + heap profiling, without managed string storage; using heap sampling defaults
* only-profiling-heap-mss: Same as above, but using managed string storage branch
* only-profiling-heap-mss-sr1: Same as above, but with heap profiling sampling 1:1 with allocation profiler

<b style="font-weight:normal;" id="docs-internal-guid-cb63b55c-7fff-275f-07e8-ffa6b0969da5"><div dir="ltr" style="margin-left:0pt;" align="center">
Configuration | Min memory RSS (MiB) | Max memory RSS (MiB) | Avg memory RSS (MiB) | AVG virtual memory (GiB)
-- | -- | -- | -- | --
baseline worker 0 | 893 | 908 | 900 | 1.4
baseline worker 1 | 894 | 911 | 902 | 1.4
only-profiling-heap worker 0 | 959 | 985 | 972 | 1.7
only-profiling-heap worker 1 | 959 | 980 | 970 | 1.7
only-profiling-heap-mss worker 0 | 937 | 964 | 955 | 1.6
only-profiling-heap-mss worker 1 | 934 | 962 | 951 | 1.6
only-profiling-heap-mss-sr1 worker 0 | 956 | 980 | 968 | 1.6
only-profiling-heap-mss-sr1 worker 1 | 946 | 979 | 965 | 1.6
</div></b>

image

_Rails app_:

Here, we’re comparing 5 configurations:
* baseline: Unmodified application, no profiling at all
* candidate-heap-profiling: Application running cpu + wall-time + allocation + heap profiling, without managed string storage; using heap sampling defaults
* candidate-heap-profiler-mss: Same as above, but using managed string storage branch
* candidate-heap-profiler-mss-sr1: Same as above, but with heap profiling sampling 1:1 with allocation profiler
* candidate-heap-profiling-sr1: Additional data point: what if we enabled 1:1 sampling the allocation profiler without the managed string storage?

image

image


benchmarks/profiler_string_storage_intern.rb
```diff
@@ -0,0 +1,40 @@
...
+      benchmark_time = VALIDATE_BENCHMARK_MODE ? { time: 0.01, warmup: 0 } : { time: 10, warmup: 2 }
+      x.config(
+        **benchmark_time,
+      )
+
+      x.report('intern 1000 strings') do
+        Datadog::Profiling::StackRecorder::Testing._native_benchmark_intern(@recorder, "hello, world!", 1000, false)
+      end
+
+      x.report('intern_all 1000 strings') do
```
AlexJF, line 26: Not a blocker but there may be value in benchmarking the "intern new string" path? Both of these benchmarks are really only exercising the "existing string" path?

And maybe having a realistic balance based on some statistics of profiling on typical Ruby apps (e.g. 90% new strings, 10% re-use).

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -239,19 +203,21 @@ void heap_recorder_free(heap_recorder *heap_recorder) {
...
 
   if (heap_recorder->active_recording != NULL && heap_recorder->active_recording != &SKIPPED_RECORD) {
     // If there's a partial object record, clean it up as well
-    object_record_free(heap_recorder->active_recording);
+    object_record_free(heap_recorder, heap_recorder->active_recording);
   }
 
   ruby_xfree(heap_recorder->reusable_locations);
+  ruby_xfree(heap_recorder->reusable_ids);
+  ruby_xfree(heap_recorder->reusable_char_slices);
```
AlexJF, line 220: Note to come back to before submitting: Are we freeing the string storage somewhere?

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -733,60 +687,34 @@ static void commit_recording(heap_recorder *heap_recorder, heap_record *heap_rec
...
-  heap_record_update_data update_data = (heap_record_update_data) {
-    .locations = locations,
-    .record = &heap_record,
-  };
-  st_update(heap_recorder->heap_records, (st_data_t) &lookup_key, update_heap_record_entry_with_new_allocation, (st_data_t) &update_data);
+  heap_record *new_or_existing_record = NULL; // Will be set inside update_heap_record_entry_with_new_allocation
+  bool existing = st_update(heap_recorder->heap_records, (st_data_t) stack, update_heap_record_entry_with_new_allocation, (st_data_t) &new_or_existing_record);
+  if (existing) {
+    heap_record_free(heap_recorder, stack);
+  }
```
AlexJF, line 715: Perhaps a note here or in the definition of `heap_records` that the only purpose this serves now is to reduce memory usage? Before it played the extra role of saving allocations and computations given we could index with a simpler stack-allocated key and thus not have to always create a record like we are doing here now.

Otherwise, it feels very tempting to just come here and remove `heap_records` altogether given it's not actually saving us any work? 😅

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -894,202 +799,103 @@ VALUE object_record_inspect(object_record *record) {
...
+
+  // Intern all these strings...
+  ddog_CharSlice *strings = recorder->reusable_char_slices;
+  // Put all the char slices in the same array; we'll pull them out in the same order from the ids array
   for (uint16_t i = 0; i < stack->frames_len; i++) {
     const ddog_prof_Location *location = &locations.ptr[i];
+    strings[i] = location->function.filename;
+    strings[i + stack->frames_len] = location->function.name;
+  }
+  intern_all_or_raise(recorder->string_storage, (ddog_prof_Slice_CharSlice) { .ptr = strings, .len = stack->frames_len * 2 }, recorder->reusable_ids, stack->frames_len * 2);
```
AlexJF, line 822: Nice that you managed to add the batch API as well! ❤️

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -894,202 +799,103 @@ VALUE object_record_inspect(object_record *record) {
...
+  heap_record *stack2 = (heap_record*) key2;
 
-static inline int64_t heap_record_key_entry_line(heap_record_key *key, size_t entry_i) {
-  if (key->type == HEAP_STACK) {
-    return key->heap_stack->frames[entry_i].line;
+  if (stack1->frames_len != stack2->frames_len) {
+    return ((int) stack1->frames_len) - ((int) stack2->frames_len);
   } else {
-    return key->location_slice->ptr[entry_i].line;
+    return memcmp(stack1->frames, stack2->frames, stack1->frames_len * sizeof(heap_frame));
```
AlexJF, line 860: Not sure if this is an issue but it occurred to me that low level languages sometimes do weird things like adding some padding to structs in the interest of memory alignment. Do you know if there's a risk of that happening here and if it does happen, is it guaranteed that the padding is zeroed? If it's left as "undetermined" the memcmp approach may lead to weird results?

EDIT: It seems like behaviour may be undetermined around padding so `memcmp` may be dangerous in a general sense? I suppose if we initialize the frames correctly and never change their value we can be reasonably sure that padding "corruption" won't come into play but we may want to add an explicit comment to treat frames as immutable and always zero-initialize the structs?

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -894,202 +799,103 @@ VALUE object_record_inspect(object_record *record) {
...
-  } else {
-    ddog_CharSlice filename = key->location_slice->ptr[entry_i].function.filename;
-    (*filename_ptr) = filename.ptr;
-    return filename.len;
-  }
+// The entire stack is represented by ids (name, filename) and lines (integers) so we can treat is as just
+// a big string of bytes and hash it all in one go.
+st_index_t heap_record_hash_st(st_data_t key) {
+  heap_record *stack = (heap_record*) key;
+  return st_hash(stack->frames, stack->frames_len * sizeof(heap_frame), FNV1_32A_INIT);
```
AlexJF, line 871: Similar padding issue as with `heap_record_cmp_st`.

ext/datadog_profiling_native_extension/stack_recorder.c
```diff
@@ -571,6 +589,14 @@ static VALUE _native_serialize(DDTRACE_UNUSED VALUE _self, VALUE recorder_instan
 
   ddog_prof_EncodedProfile_drop(&serialized_profile.ok);
 
+  // It's now OK to again do things that can trigger exceptions
+
+  // @ivoanjo: Maybe this could be done without the GVL being held?
+  ddog_prof_MaybeError result = ddog_prof_ManagedStringStorage_advance_gen(state->string_storage);
```
AlexJF, line 595: If the string storage itself is threadsafe then we should?

AlexJF (approved):

marcotc (approved):
> Yes. Reduce overhead of heap profiling

Don't forget to update the changelog once benchmarking results are ready. (e.g. claim both memory and cpu improvements)

benchmarks/profiler_string_storage_intern.rb
ivoanjo: TBH I added this benchmark to motivate the work on https://github.com/DataDog/libdatadog/pull/844 .

I'm not quite sure testing other mixes of strings would make a lot of difference in comparing `intern` vs `intern_all`, since in the end both call the string table in the same way, it's just that the latter avoids crossing the FFI and acquiring the write lock many times.

Or were you thinking more along the lines of turning this benchmark into a measurement of the string table performance in general? So that, e.g. if there was a regression we'd catch it?

AlexJF: Yeah without that context I had assumed the intent was to more easily catch regressions or understand evolution of performance of this part.

ext/datadog_profiling_native_extension/stack_recorder.c
ivoanjo: I left the comment as it looked slightly awkward to plumb the action to happen outside the GVL and to bring its result back in afterwards but I tried it now and it seems reasonable so outside of the GVL it goes in 1947ea0752fa3663f6fe238fb7bed77ba0ce4345 .

ext/datadog_profiling_native_extension/heap_recorder.c
ivoanjo: :+1: , see my note above

ext/datadog_profiling_native_extension/heap_recorder.c
ivoanjo: > Not sure if this is an issue but it occurred to me that low level languages sometimes do weird things like adding some padding to structs in the interest of memory alignment. Do you know if there's a risk of that happening here and if it does happen, is it guaranteed that the padding is zeroed? If it's left as "undetermined" the memcmp approach may lead to weird results?
>
> EDIT: It seems like behaviour may be [undetermined](https://stackoverflow.com/questions/52684192/is-it-guaranteed-that-the-padding-bits-of-zeroed-structure-will-be-zeroed-in-c) around padding so memcmp may be dangerous in a general sense? I suppose if we initialize the frames correctly and never change their value we can be reasonably sure that padding "corruption" won't come into play but we may want to add an explicit comment to treat frames as immutable and always zero-initialize the structs?

I had exactly the same brain wave when I was writing this. I even considered forcing packing but since it's not a standard I was on the fence and ended up not adding it.

In 3a6b8873e1f3660581f0e309ae1b00bf620b1489 I added a check that the size of the structure is the size of its elements. This is standard C11 (and `_Static_assert` has been in GCC since 2011) so we should be fine. Adding a `char` somewhere to the structure and then `+ sizeof(char)` correctly makes the `_Static_assert` trigger a compiler error.

ext/datadog_profiling_native_extension/heap_recorder.c
ivoanjo: I waited until the last moment as I didn't want to add more confusion to the overall proceedings, but once everything else was in place, this was a very easy win :)

ext/datadog_profiling_native_extension/heap_recorder.c
ivoanjo: Agreed. Added some notes about this in 7d0acaaf59d887e42dcf10287ecd749ae53fd1d9

benchmarks/profiler_string_storage_intern.rb
ivoanjo: I like your suggestion: In c29be53269ebe17db3f226b8c662200ad3f2f49a I've modified the benchmark to track performance of managed string storage more generally, rather than just comparing intern vs intern_all.

ivoanjo:
> Don't forget to update the changelog once benchmarking results are ready. (e.g. claim both memory and cpu improvements)

:+1:  Updated PR description with results!

ivoanjo:
_Update: As discussed, I've pushed a commit to set `experimental_heap_sample_rate` to 1, raising the heap sampling rate from its previous default of 10._

PR #4334, opened 2025-01-31
Add meta_struct support on traces and spans

vpellan:
**What does this PR do?**

This PR adds meta_struct support to traces and spans. meta_struct is a way to send complex structures to the backend. It is sent as a Map<String, ByteArray> with the ByteArray being the structure encoded in messagepack.

**Motivation:**

meta_struct is required for stack trace collection, which is required for exploit prevention. But this can also be used by other products.

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**


lib/datadog/tracing/metadata/meta_struct.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Tracing
+    module Metadata
+      # Adds complex structures tagging behavior through meta_struct
+      # @public_api
+      module MetaStruct
+        def set_meta_struct(meta_struct)
+          self.meta_struct.merge!(meta_struct)
```
Strech, line 10: I think the name prefix `set_` implies that any data prior to that call will be overwritten.

lib/datadog/tracing/transport/serializable_trace.rb
```diff
@@ -113,6 +113,9 @@ def to_msgpack(packer = nil)
           packer.write(span.meta)
           packer.write('metrics')
           packer.write(span.metrics)
+          packer.write('meta_struct')
+          # We encapsulate the resulting msgpack in a binary msgpack
+          packer.write(span.meta_struct.transform_values(&:to_msgpack))
```
Strech, line 118: If you want to encapsulate, then the logic of the values transformation should not be exposed, instead `to_msgpack` should be called on entire `meta_struct`

```ruby
packer.write(span.meta_struct.to_msgpack)
```

sig/datadog/tracing/span.rbs
```diff
@@ -6,13 +6,14 @@ module Datadog
       attr_accessor end_time: (Time | nil)
       attr_accessor id: Integer
       attr_accessor meta: Hash[String, String]
+      attr_accessor meta_struct: Hash[String, untyped]
```
Strech, line 9: That should be a type alias inside the `MetaStruct` rbs instead

sig/datadog/tracing/trace_operation.rbs
```diff
@@ -23,7 +23,7 @@ module Datadog
       attr_writer sampled: untyped
       attr_writer service: untyped
 
-      def initialize: (?agent_sample_rate: untyped?, ?events: untyped?, ?hostname: untyped?, ?id: untyped?, ?max_length: untyped, ?name: untyped?, ?origin: untyped?, ?parent_span_id: untyped?, ?rate_limiter_rate: untyped?, ?resource: untyped?, ?rule_sample_rate: untyped?, ?sample_rate: untyped?, ?sampled: untyped?, ?sampling_priority: untyped?, ?service: untyped?, ?tags: untyped?, ?metrics: untyped?, ?remote_parent: untyped?) -> void
+      def initialize: (?agent_sample_rate: untyped?, ?events: untyped?, ?hostname: untyped?, ?id: untyped?, ?max_length: untyped, ?name: untyped?, ?origin: untyped?, ?parent_span_id: untyped?, ?rate_limiter_rate: untyped?, ?resource: untyped?, ?rule_sample_rate: untyped?, ?sample_rate: untyped?, ?sampled: untyped?, ?sampling_priority: untyped?, ?service: untyped?, ?tags: untyped?, ?meta_struct: untyped?, ?metrics: untyped?, ?remote_parent: untyped?) -> void
```
Strech, line 26: I think for the new class you can use a type alias from `MetaStruct`

lib/datadog/tracing/metadata/meta_struct.rb
```diff
@@ -0,0 +1,29 @@
...
+        end
+
+        def get_meta_struct(key)
```
Strech, line 13: Now sure about naming, but if I call `get_meta_struct` I would expect the whole meta-struct object in return. I think we should somehow highlight that it's a part inside the meta-struct.

Maybe names like

```
get_meta_struct_value
get_meta_struct_field
```

or something way better (see ⬆️)

lib/datadog/tracing/metadata/meta_struct.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Tracing
+    module Metadata
+      # Adds complex structures tagging behavior through meta_struct
+      # @public_api
+      module MetaStruct
```
Strech, line 8: I have a question about naming, we call `Metadata` as a single word, but MetaStruct as two words. Does it make sense if we call it `Metastruct`? We can avoid long names on reading too

```ruby
get_metastruct_value("key")
```

Strech (commented):
Awesome work 👏🏼 

I think we can run a round of light improvements over the naming and interfaces, rest looks solid!

lib/datadog/tracing/metadata/meta_struct.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Tracing
+    module Metadata
+      # Adds complex structures tagging behavior through meta_struct
+      # @public_api
```
p-datadog, line 7: If this module is part of the public API I would expect at least a brief description of the methods in it added as comments, otherwise it seems difficult to use and potentially users could misuse the API (which then we would need to support, since the API is public).

p-datadog: Also, the PR says there is no change log entry needed, which is inconsistent with adding a public API.

p-datadog (changes requested):
Requesting changes due to mismatch between no change log entry claimed in PR description and the new API being marked public.

lib/datadog/tracing/metadata/meta_struct.rb
y9v: why not just `def [](key)`?

lib/datadog/tracing/metadata/meta_struct.rb
```diff
@@ -0,0 +1,29 @@
...
+      module MetaStruct
+        def set_meta_struct(meta_struct)
```
marcotc, line 9: @Strech why not `def meta_struct=` and `meta_struct(key)`?

lib/datadog/tracing/transport/serializable_trace.rb
marcotc: btw, `to_msgpack` is implicitly called by `packer.write`, if present.

lib/datadog/tracing/metadata/meta_struct.rb
marcotc: hmm, because it's merge eh... this is in interesting api

lib/datadog/tracing/transport/serializable_trace.rb
marcotc: But `packer.write(span.meta_struct.transform_values(&:to_msgpack))` is ok.

vpellan: Yes, to_msgpack is implicitly called by packer.write but we need to call it a second time, so that the values are encoded as byte array. Or else they would be maps or arrays, but the backend wants byte arrays.

vpellan: Maybe my comment is not descriptive enough. We encapsulate the values, not the whole hash, in a message pack (which is why we use transform values) which will give a String starting with something like \xDE or \xDF. Then this String will be converted to byte array because of packer.write. `span.metastruct` is just a Hash, so using `span.metastruct.to_msgpack` would give the wrong result as we only want to convert the values to message pack, not the keys. Do you think it would be preferable to use an external method, like `convert_values_to_msgpack(span.metastruct)` ?

lib/datadog/tracing/metadata/meta_struct.rb
vpellan: Sorry, this should not be public API. I will remove it.

vpellan: fixed in 740053288bd93b1263d8ce6e0996637fc4053352

lib/datadog/tracing/metadata/meta_struct.rb
vpellan: Fixed in 0066df99c67fe1654538fcb0388c7ea439d58302 (following up @marcotc comment)

sig/datadog/tracing/trace_operation.rbs
vpellan: Fixed in f36e14704124c7dff25e744abe2807fa0edb1ebe

lib/datadog/tracing/metadata/meta_struct.rb
vpellan: Fixed in 7d76a8e8cdcb7f0e82e0da7af23059be78253018

lib/datadog/tracing/metadata/meta_struct.rb
vpellan: What do you think about that ? 0066df99c67fe1654538fcb0388c7ea439d58302

marcotc (approved):

lib/datadog/tracing/metadata/meta_struct.rb
p-datadog: :+1:

lib/datadog/tracing/metadata/metastruct.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Tracing
+    module Metadata
+      # Adds complex structures tagging behavior through metastruct
+      module Metastruct
+        def metastruct=(second)
+          @metastruct = second
+        end
```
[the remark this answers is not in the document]
vpellan: It is used in TraceOperation initialisation, which itself is used by fork_clone

lib/datadog/tracing/metadata/metastruct.rb
```diff
@@ -0,0 +1,39 @@
...
+
+        def deep_merge_metastruct!(second)
```
y9v, line 12: can we add a comment to this method with couple of examples? I think it might be important to mention what happens if the data type for some key does not match existing data type.

lib/datadog/tracing/metadata/metastruct.rb
```diff
@@ -0,0 +1,39 @@
...
+            else
+              v2
+            end
+          end
+          metastruct.merge!(second.to_h, &merger) # steep:ignore BlockTypeMismatch
+        end
+
+        def get_metastruct_field(key)
+          metastruct[key]
+        end
```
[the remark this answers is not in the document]
vpellan: What do you think about that ? 65dd37b4392716e5d4a4c3f57c70e18859ca4beb

lib/datadog/tracing/metadata/metastruct.rb
vpellan: Done in 65dd37b4392716e5d4a4c3f57c70e18859ca4beb

spec/datadog/tracing/trace_operation_spec.rb
```diff
@@ -100,6 +103,10 @@
         expect(trace_op.send(:metrics)).to eq({})
       end
 
+      it do
+        expect(trace_op.send(:metastruct).to_h).to eq({})
+      end
```
Strech, line 108:
```suggestion
      it { expect(trace_op.send(:metastruct).to_h).to eq({}) }
```

Strech (approved):
Amazing move, I think it's good to go! Solid work 👏🏼 

_P.S I left few suggestions, but they are not blockers_

y9v (approved):

lib/datadog/tracing/metadata/metastruct.rb
```diff
@@ -0,0 +1,69 @@
...
+              v1.merge(v2, &merger)
+            elsif v1.is_a?(Array) && v2.is_a?(Array)
+              v1.concat(v2)
+            elsif v2.nil?
+              v1
+            else
+              v2
+            end
+          end
+          metastruct.merge!(second.to_h, &merger) # steep:ignore BlockTypeMismatch
```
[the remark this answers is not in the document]
vpellan: Yes because second is a Metastruct object too (I will change the rbs file though)

lib/datadog/tracing/metadata/metastruct.rb
```diff
@@ -0,0 +1,64 @@
...
+
+        MERGER = proc do |_, v1, v2|
+          if v1.is_a?(Hash) && v2.is_a?(Hash)
+            v1.merge(v2, &MERGER)
+          elsif v1.is_a?(Array) && v2.is_a?(Array)
+            v1.concat(v2)
+          elsif v2.nil?
+            v1
+          else
+            v2
```
p-datadog, line 20: I would do `elsif v1.nil?; v2` and in the `else` branch raise an exception that v1 and v2 are not mergeable.

lib/datadog/tracing/span_operation.rb
```diff
@@ -456,6 +463,7 @@ def build_span
           id: @id,
           meta: Core::Utils::SafeDup.frozen_or_dup(meta),
           metrics: Core::Utils::SafeDup.frozen_or_dup(metrics),
+          metastruct: @metastruct.to_h.dup,
```
p-datadog, line 466: I think I would expect `to_h` to return a new value (rather than `self` as is the case for `Hash`). I checked the diff for the `to_h` calls and it is not clear in which case the caller should `dup` the result and in which case they should not. With the current implementation it seems likely that in the future someone can forget to `dup` and mutate something unintentionally.

lib/datadog/tracing/metadata/metastruct.rb
```diff
@@ -0,0 +1,64 @@
...
+          end
+        end
+
+        def self.empty
+          new({})
+        end
+
+        def initialize(metastruct)
```
p-datadog, line 28: This is a confusing constructor. What is the type of `metastruct` here?

lib/datadog/tracing/span_operation.rb
```diff
@@ -88,6 +89,7 @@ def initialize(
 
         # Set tags if provided.
         set_tags(tags) if tags
+        @metastruct = Tracing::Metadata::Metastruct.empty
```
p-datadog, line 92: What is the purpose of initializing the metastruct fields to object instances rather than `nil` and having object instances created when they actually get something added to them?

lib/datadog/tracing/metadata/metastruct.rb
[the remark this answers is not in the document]
y9v: don't we check the type of the second object that is merged in the `MERGER` lambda function?

lib/datadog/tracing/metadata/metastruct.rb
y9v:
```ruby
if v1.is_a?(Hash) && v2.is_a?(Hash)
  ...
elsif v1.is_a?(Array) && v2.is_a?(Array)
  ...
else
  v2
end
```

spec/datadog/tracing/metadata/metastruct_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      context 'with not empty metastruct' do
+        let(:preexisting_metastruct) { { 'key' => 'value' } }
+
+        it { is_expected.to eq({ 'key' => 'value' }) }
+      end
+    end
+  end
+
+  describe '#deep_merge!' do
```
y9v, line 28: I think we should also test deep merging - now we only test merging of flat hashes

lib/datadog/tracing/metadata/metastruct.rb
y9v: I also think that passing `metastruct` as an argument when creating a `Metastruct` instance is weird. Should we call this argument `data` instead?

lib/datadog/tracing/trace_operation.rb
```diff
@@ -25,7 +25,7 @@ module Tracing
     #
     # @public_api
     class TraceOperation
-      include Metadata::Tagging
+      include Metadata
```
y9v, line 28: sorry, I will have to block because of this change - I don't think we have to include both `Metadata::Errors` and `Metadata::Analytics` here just to require 'metadata/metastruct' file?

y9v (changes requested):

lib/datadog/tracing/span_operation.rb
y9v: I changed the implementation, removed those `dup` calls and other things. Could you please review again?

lib/datadog/tracing/span.rb
```diff
@@ -185,12 +191,15 @@ def pretty_print(q)
               q.text "#{key} => #{value}"
             end
           end
-          q.group(2, 'Metrics: [', ']') do
+          q.group(2, 'Metrics: [', "]\n") do
```
Strech, line 194: ❤️

spec/datadog/tracing/metadata/metastruct_tagging_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  describe '#set_metastruct_tag' do
+    it 'sets the metastruct to a hash with given key / value pair' do
+      test_object.set_metastruct_tag(:foo, [{ some: 'value' }])
+
+      expect(test_object.get_metastruct_tag(:foo)).to eq([{ some: 'value' }])
+    end
+
+    it 'does not lose previous entries' do
+      test_object.instance_variable_set(:@metastruct, { bar: [1] })
```
Strech, line 20: I think it could be changed into a pair of `change` expectations instead

```ruby
expect { set_metastruct_tag(:bar, [1]) }.to change { get_metastruct_tag(:bar) }.from("...").to("...")

expect { set_metastruct_tag(:foo, [2]) }.not_to change { get_metastruct_tag(:bar) }.from([1]).and to change {}.from("...").to("...")

```

spec/datadog/tracing/metadata/metastruct_tagging_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  let(:test_class) do
+    Class.new { include Datadog::Tracing::Metadata::MetastructTagging }
+  end
+
+  describe '#set_metastruct_tag' do
+    it 'sets the metastruct to a hash with given key / value pair' do
+      test_object.set_metastruct_tag(:foo, [{ some: 'value' }])
+
+      expect(test_object.get_metastruct_tag(:foo)).to eq([{ some: 'value' }])
```
Strech, line 16: I think this should be verified before, hence it's better to use `change`

Strech (approved):
Awesome 👍🏼 

_P.S I left few non-blocking test improvements._

spec/datadog/tracing/metadata/metastruct_tagging_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  let(:test_class) do
+    Class.new { include Datadog::Tracing::Metadata::MetastructTagging }
+  end
+
+  describe '#set_metastruct_tag' do
+    it 'sets the metastruct to a hash with given key / value pair' do
+      expect do
+        test_object.set_metastruct_tag(:foo, [{ some: 'value' }])
+      end.to change { test_object.get_metastruct_tag(:foo) }.from(nil).to([{ some: 'value' }])
```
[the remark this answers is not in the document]
y9v: Thanks! It is indeed not as ugly. Fixed here [a70a07e](https://github.com/DataDog/dd-trace-rb/pull/4334/commits/a70a07ef7b6540b7f45792834dae1b82a7440140)

PR #4344, opened 2025-02-04
Attempt to address sucker_punch flaky test

sarahchen6:
**What does this PR do?**
Use mutexes to avoid the concurrency issue due to unsynchronized modifications of an array.

**Motivation:**
Minimize our flaky tests!

**Change log entry**
No.

**Additional Notes:**
Addresses this issue: https://github.com/DataDog/ruby-guild/issues/186

Right now the code leads to a deadlock during initialization...

**How to test the change?**
Hope to never see this bug again in CI... I test the flakiness by running the test 1000 times in a row.


spec/datadog/tracing/contrib/support/tracer_helpers.rb
ivoanjo, line 1: Minor: Is it possible to contain these changes to the sucker-punch testing? E.g. it's a bit meh to add a bunch of complexity (and to potentially hide real problems with the code) to only make one not-very-good-test happy.

sarahchen6: Good point. Will look into it!

sarahchen6: I separated out the info that sucker-punch needs so that we're not messing with the rest of the code. WDYT?

spec/datadog/tracing/contrib/support/tracer_helpers.rb
```diff
@@ -53,6 +53,15 @@ def fetch_spans(tracer = self.tracer)
...
     end
 
+    def get_spans_count(tracer = self.tracer)
+      count = 0
+      traces = fetch_traces(tracer)
+      traces.each do |trace|
+        count += trace.instance_variable_get(:@spans).length
+      end
+      count
+    end
```
ivoanjo, line 63: Oh wow. Now that I'm trying to review this and expanded this whole file do I realize that we're adding entire new instance variables to the tracer and fiddling with what it stores :eyes:

Having said that, in this situation I don't see `trace.instance_variable_get(:@spans)` being different from `trace.spans` -- did you run into any difference between them?

Also, you can use `sum` to directly get the total, no need to keep the variable yourself ;)

```
[1] pry(main)> [[1, 1, 1], [2, 2], [3, 3, 3]].sum { |it| it.length }
=> 8
```

Finally, I do think we have historically tended to accumulate common helper code that actually isn't used more than in a place or two. Since this helper is only for sucker_punch, consider moving it to that file (it should be as simple as moving the code -- yeah you can use defs inside rspec stuff, the rspec stuff is just a pretty syntax to actually defining classes and methods)

spec/datadog/tracing/contrib/sucker_punch/patcher_spec.rb
ivoanjo, line 1: > Specifically, we want to avoid calling fetch_spans which creates a span array that may be altered during test execution, leading to concurrency issues.

Actually, looking at the error, I suspect the complaint may not be about the new array that `collect(&:spans).flatten` is creating.

Because... what two threads can access that array at the same time? As far as I can see it, it's only created and returned to the caller of `fetch_spans`...?

I suspect the error is actually in this part: `traces.collect(&:spans)`. The arrays being accessed concurrently are the ones in `TraceSegment#spans` -- the rspec thread is reading them (to be able to copy them) at the same time as the suckerpunch thread is mutating them (adding new spans).

This would match with what you're doing in your fix: since you're now just accessing the count of the spans array, not iterating it, and JRuby stores the size of the array in a [regular Java instance field](https://github.com/jruby/jruby/blob/b0ecd4b31509bb41a871098410353f49c581eac4/core/src/main/java/org/jruby/RubyArray.java#L400) then we either observe the count before or after it gets modified by the background thread, and we're never in an "in-between" state as we could be if one thread was half-way through iterating the array while other was modifying it.

ivoanjo (commented):
Looks reasonable! Any reason for this to still be marked as draft? :)

spec/datadog/tracing/contrib/sucker_punch/patcher_spec.rb
sarahchen6: Yes, that's true -- apologies my comment could've been worded better! The error is in fact with `traces.collect(&:spans)` like you said when it is creating the array to-be-flattened, and my fix avoids the concurrent access by just retrieving the array length value.

spec/datadog/tracing/contrib/support/tracer_helpers.rb
sarahchen6: Ah I didn't realize I could add defs inside rspec my bad. Will do!

ivoanjo: If I'm not misremembering

```
RSpec.describe 'foo' do
  let(:val) { 'a' }
  context 'bar' do
     let(:val) { 'b' }
  end
end
```

is actually creating a class hierarchy, where the outer 'foo' is the top class and the 'bar' is a subclass. That's why the `val` is the correct value -- it's exactly like redefining a method in the subclass.

And you could even do

```
RSpec.describe 'foo' do
  let(:val) { 'a' }
  context 'bar' do
     let(:val) { super() + 'b' }
  end
end
```

to get `'ab'` as result, because well, want to guess what's happening? ;)

sarahchen6: I see! `super()` calls the top 'foo' class's `let(:val) { 'a' }`, and then appends `'b'` to get `'ab'`.

ivoanjo (approved):
:+1: LGTM. Ahaha one of those PRs that looks deceptively simple after all the work we poured into making it simple :D

PR #4345, opened 2025-02-05
Fix remote config handling for AppSec

y9v:
Before we were not setting `apply_state` to ACKNOWLEDGED on remote config content, which caused some system tests to fail.

**What does this PR do?**
This PR adds setting of `apply_state` on remote config content to `ACKNOWLEDGED` after successful processing.

**Motivation:**
Failing system tests for full denylist blocking: https://github.com/DataDog/system-tests/pull/3937

**Change log entry**
None, this is internal change. Changes via remote config were applied before too, it's just the status that wasn't set correctly.

**Additional Notes:**
None.

**How to test the change?**
CI should be enough.

### DIFF
--- a/lib/datadog/appsec/remote.rb
+++ b/lib/datadog/appsec/remote.rb
@@ -74,15 +74,12 @@ module Datadog
               case content.path.product
               when 'ASM_DD'
                 rules << parsed_content
-                content.applied
               when 'ASM_DATA'
                 data << parsed_content['rules_data'] if parsed_content['rules_data']
-                content.applied
               when 'ASM'
                 overrides << parsed_content['rules_override'] if parsed_content['rules_override']
                 exclusions << parsed_content['exclusions'] if parsed_content['exclusions']
                 custom_rules << parsed_content['custom_rules'] if parsed_content['custom_rules']
-                content.applied
               end
             end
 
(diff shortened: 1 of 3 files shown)


lib/datadog/appsec/remote.rb
```diff
@@ -74,12 +74,15 @@ def receivers(telemetry)
...
                 rules << parsed_content
+                content.applied
               when 'ASM_DATA'
                 data << parsed_content['rules_data'] if parsed_content['rules_data']
+                content.applied
               when 'ASM'
                 overrides << parsed_content['rules_override'] if parsed_content['rules_override']
                 exclusions << parsed_content['exclusions'] if parsed_content['exclusions']
                 custom_rules << parsed_content['custom_rules'] if parsed_content['custom_rules']
+                content.applied
```
Strech, line 85: I think this place is not protecting us from have content marked as applied, but have it failed later this method.

I see that as in Kafka, you ACK if you're done with incoming changes/data and you really applied it. Hence we really apply the new configuration only on line 109 with a call to `reconfigure`, hence the success of that operation should signal us about successful content use.

Strech (changes requested):
I have doubts about marking content as applied without actual use of it before hand.

lib/datadog/appsec/remote.rb
y9v: you are right, fixed that [d9bcea9](https://github.com/DataDog/dd-trace-rb/pull/4345/commits/d9bcea98ceca89e7382060d2988ea5a685fe28f1)

Strech (approved):
LGTM 👏🏼

PR #4346, opened 2025-02-05
DEBUG-3439 DEBUG-3440 remove prefixes from probe paths when path matching

p-datadog:
**What does this PR do?**

Loosens the restrictions on path matching to also try to match prefixes of probe paths to the paths that exist at runtime.

For example, suppose at runtime we have:
`/app/app/controllers/hello_controller.rb`

Previously, either exact full path or suffixes of the path were accepted:
- `/app/app/controllers/hello_controller.rb`
- `app/controllers/hello_controller.rb`
- `hello_controller.rb`
Now, paths that have extra subdirectories in front will also be accepted:
 - `local/hello_controller.rb`
 - `local/controllers/hello_controller.rb`

**Motivation:**
For projects that are stored in subdirectories in their source code repositories (e.g. monorepos), source code integration is providing a full path that has extra prefixes from the application's standpoint (and also runtime standpoint). These prefixes previously made such paths not instrumentable. With this PR, the extra prefixes will be removed to try to find a matching file.

**Change log entry**
Yes: loosen path matching in dynamic instrumentation 

**Additional Notes:**
This change can cause potentially incorrect files to match if the desired file is not yet loaded.

In order to properly implement matching accounting for these extra prefixes, much more work is needed. The ideal (in my present opinion) matching algorithm is described in a comment in this diff. This PR only implements a quick fix to get Ruby DI out to customers as soon as possible.

**How to test the change?**
Unit tests and integration test are added.

### DIFF
--- a/lib/datadog/di/code_tracker.rb
+++ b/lib/datadog/di/code_tracker.rb
@@ -155,7 +155,7 @@ module Datadog
               return inexact.first
             end
             return nil unless suffix.include?('/')
-            suffix.sub!(%r,.*/+,, '')
+            suffix.sub!(%r{.*/+}, '')
           end
         end
       end
--- a/lib/datadog/di/code_tracker.rb
+++ b/lib/datadog/di/code_tracker.rb
@@ -140,16 +140,23 @@ module Datadog
           exact = registry[suffix]
           return [suffix, exact] if exact
 
-          inexact = []
-          registry.each do |path, iseq|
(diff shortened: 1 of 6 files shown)


p-datadog:
I have now verified this PR manually against debugger-demo-ruby.

p-datadog:
I have now verified this PR manually against debugger-demo-ruby.

TonyCTHsu (approved):

sarahchen6 (approved):

PR #4355, opened 2025-02-06
Remove reactive engine from AppSec rack instrumentation

y9v:
**What does this PR do?**
It removes `Datadog::AppSec::Reactive::Engine`.

**Motivation:**
The problem with reactive engine was that it was not used the way it was intended, and therefore only added unnecessary complexity to our code.

**Change log entry**
None, this is internal change.

**Additional Notes:**
None.

**How to test the change?**
CI and manual testing with app generator.


Strech (approved):
🟢🔴🔴🔴 🚀

PR #2250, opened 2022-08-31
[SCP-184] Fix incompatible library check when there's no libruby.so

ivoanjo:
**What does this PR do?**:

Add a missing check for a `NULL` result of `dlsym` that happens when Ruby is not built with `libruby.so`.

**Motivation**:

The logic in `ddtrace_profiling_loader.c` follows Ruby's `dln.c`; see the big comment at the top of the former for why it needs to exist.

The `incompatible_library` sanity check exists to try to catch situations where the library being loaded is linked to a different `libruby.so` (or equivalent in other OSs) than the one that is doing the loading.

BUT when we implemented `incompatible_library` we missed an check that upstream Ruby does.

From https://github.com/ruby/ruby/blob/v3_1_2/dln.c#L295:

```c
static bool
dln_incompatible_library_p(void *handle)
{
    void *ex = dlsym(handle, EXTERNAL_PREFIX"ruby_xmalloc");
    void *const fp = (void *)ruby_xmalloc;
    return ex && ex != fp; // ⬅️ HERE!
}
```

Ruby actually checks that the symbol found is **not NULL** (e.g. with the `ex && ` check). This is important, because there's a valid situation in which the symbol is expected to be NULL: when Ruby is built with `--disable-shared` at compilation time.

What `--disable-shared` does is that it builds the Ruby VM logic inside the `ruby` executable itself INSTEAD OF splitting the logic between the `ruby` executable and the `libruby.so` shared library.

To properly support this configuration, we should, like upstream Ruby does, consider a library compatible if the symbol lookup returns `NULL`.

**Additional Notes**:

Here's a few ways of telling if there's a `libruby`:

* `RbConfig::CONFIG['configure_args']` will contain `--disable-shared`
* `RbConfig::CONFIG['LIBRUBY']` will be `libruby-static.a` instead of, for instance `libruby.so.3.1.1`
* Ruby will not be linked to libruby:
  ```
  # with libruby
  $ ldd `which ruby` | grep libruby
    libruby.so.3.1 => /usr/local/lib/libruby.so.3.1 (0x00007fb37e260000)

  # without libruby
  $ ldd `which ruby` | grep libruby
  (no output)
  ```

**How to test the change?**:

I couldn't find a prebuilt docker image that had Ruby compiled with `--disable-shared`.

To reproduce the issue, I downloaded the latest Ruby version, configured it with `./configure --disable-shared`, compiled and installed it, and then installed `dd-trace-rb`, and tried to start the profiler.

[...]

PR #4356, opened 2025-02-06
[PROF-11311] Remove datadog_profiling_loader

ivoanjo:
**What does this PR do?**

This PR removes the `datadog_profiling_loader` native extension, as it's no longer needed.

The profiling native loader was added in
https://github.com/DataDog/dd-trace-rb/pull/2003 .

Quoting from that PR's description:

> When Ruby loads a native extension (using `require`), it uses
> `dlopen(..., RTLD_LAZY | RTLD_GLOBAL)`
> (https://github.com/ruby/ruby/blob/67950a4c0a884bdb78d9beb4405ebf7459229b21/dln.c#L362).
>
> This means that every symbol exposed directly or indirectly by that
> native extension becomes visible to every other extension in the
> Ruby process. This can cause issues, see
> https://github.com/rubyjs/mini_racer/pull/179 .
>
> Ruby's extension loading mechanism is not configurable -- there's
> no way to tell it to use different flags when calling `dlopen`.
> To get around this, this commit introduces introduces another
> extension (profiling loader) which has only a single responsibility:
> mimic Ruby's extension loading mechanism, but when calling
> `dlopen` use a different set of flags.
>
> This idea was shamelessly stolen from @lloeki's work in
> https://github.com/rubyjs/mini_racer/pull/179, big thanks!

...and importantly ends with:

> Note that, that we know of, the profiling native extension only
> exposes one potentially problematic symbol:
> `rust_eh_personality` (coming from libddprof/libdatadog).
>
> Future versions of Rust have been patched not to expose this
> (see https://github.com/rust-lang/rust/pull/95604#issuecomment-1108563434)
> so we may want to revisit the need for this loader in the future,
> and perhaps delete it if we no longer require its services :)

And we have reached the situation predicted in that description:

1. Nowadays libdatadog no longer exposes `rust_eh_personality`

2. We have a test that validates that only expected symbols are exported by the libdatadog library (see https://github.com/DataDog/libdatadog/pull/573 ).

   Any new symbols that show up would break shipping new libdatadog versions to rubygems.org until we review them.

3. The `libdatadog_api` extension, which we've been shipping for customers since release 2.3.0 back in July 2024 has always been loaded directly without a loader without issues.

Thus, I think it's the right time to get rid of the loader.

**Motivation:**

Having the loader around is not zero cost; we've run into/caused a few issues ( https://github.com/DataDog/dd-trace-rb/pull/2250 and https://github.com/DataDog/dd-trace-rb/pull/3582 come to mind).

It also adds overhead to development: every time we need to rebuild the extensions, it's an extra extension that needs to be prepared, rebuilt, copied, etc.

I've been meaning to get rid of the loader for some time now, and this came up again this week so I decided it was time to do it.

**Change log entry**

None.

**Additional Notes:**

**Note that none if this is visible to customers -- the loader was just an additional layer we used when loading the profiler.**

In the future, we can always resurrect this approach if we figure out we need it again.

We've also discussed internally about proposing upstream to the Ruby VM to maybe add an API to do what the loader was doing, so that we wouldn't need the weird workaround.

We haven't yet decided if we're going to do that (help welcome!).

**How to test the change?**

The loader was responsible for loading the rest of profiling.

Thus, any test that uses profiling was also validating the loader and now will validate that we're doing fine without it.

TL;DR green CI is good.


AlexJF (approved):
🎉

PR #4358, opened 2025-02-10
DEBUG-3457 hack probe notifier worker starting

p-datadog:
**What does this PR do?**

Moves the call to start probe notifier worker (i.e. the background thread) to tracing Rack middleware, which also starts remote config worker.

**Motivation:**

On forking web servers like puma the probe notifier worker is currently started in the parent (supervisor) process rather than in the worker processes, causing DI to not send any events to the agent.

**Change log entry**
Yes: fix dynamic instrumentation event sending on forking web servers

**Additional Notes:**
A proper solution would, in my opinion, not invoke DI (or remote config, which is a prerequisite for DI) from tracing, however in practice all DI customers should have tracing enabled therefore the solution in this PR should work for them.

**How to test the change?**
I manually verified it against debugger-demo-ruby.
Current CI does not appear to have a configuration that reproduces debugger-demo-ruby environment (since without this PR everything works in CI but does not work in debugger-demo-ruby).

### DIFF
--- a/lib/datadog/di/component.rb
+++ b/lib/datadog/di/component.rb
@@ -84,7 +84,18 @@ module Datadog
         @probe_notifier_worker = ProbeNotifierWorker.new(settings, transport, logger, telemetry: telemetry)
         @probe_notification_builder = ProbeNotificationBuilder.new(settings, serializer)
         @probe_manager = ProbeManager.new(settings, instrumenter, probe_notification_builder, probe_notifier_worker, logger, telemetry: telemetry)
-        probe_notifier_worker.start
+
+        # If the worker is started here, it will be in the parent process
+        # of forking web servers like puma rather than in the worker processes
+        # that actually process HTTP requests, and thus DI will end up not
+        # sending any events to the agent.
+        #
+        # There is currently no "nice" way to handle this situation correctly -
+        # remote config has the same issue and the tracing Rack middleware
+        # presently starts the remote config worker.
+        #
+        # DI is hacked into that middleware the same way.
+        #probe_notifier_worker.start
       end
 
       attr_reader :settings
--- a/lib/datadog/di/component.rb
+++ b/lib/datadog/di/component.rb
@@ -95,7 +95,7 @@ module Datadog
         # presently starts the remote config worker.
         #
         # DI is hacked into that middleware the same way.
-        #probe_notifier_worker.start
+        # probe_notifier_worker.start
       end
 
       attr_reader :settings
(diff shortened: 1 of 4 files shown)


sarahchen6 (approved):

p-datadog:
The red configuration is either refusing to rerun or keeps failing, it's confusing.

lib/datadog/di/component.rb
```diff
@@ -84,7 +84,18 @@ def initialize(settings, agent_settings, logger, code_tracker: nil, telemetry: n
         @probe_notifier_worker = ProbeNotifierWorker.new(settings, transport, logger, telemetry: telemetry)
         @probe_notification_builder = ProbeNotificationBuilder.new(settings, serializer)
         @probe_manager = ProbeManager.new(settings, instrumenter, probe_notification_builder, probe_notifier_worker, logger, telemetry: telemetry)
-        probe_notifier_worker.start
```
marcotc, line 87: Is there a reason the approach we use in tracing doesn't work for DI?
In tracing, we always check if the process has forked when submitting data, and if it has, we restart the worker: https://github.com/DataDog/dd-trace-rb/blob/5319afcdba36777acf428670d65023bccee23f6b/lib/datadog/tracing/writer.rb#L110

p-datadog: I can certainly do that.

p-datadog: Done in https://github.com/DataDog/dd-trace-rb/pull/4363

lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -39,6 +39,12 @@ def call(env)
 
             boot = Datadog::Core::Remote::Tie.boot
 
+            if defined?(Datadog::DI.current_component)
+              # TODO: when would Datadog::DI be defined but
+              # Datadog::DI.current_component not be defined?
+              Datadog::DI.current_component&.probe_notifier_worker&.start
```
Strech, line 45: Also a question similar to TODO's question: Is the `current_component&.probe_notifier_worker` needs to have safe-navigation? Could it be not defined?

p-datadog: The entire DI component should be missing on Ruby 2.5 and JRuby. I am not sure (per the comment, and I did not investigate) why the DI module is defined but current_component is not defined sometimes. If current_component is defined and not nil, the rest of DI infrastructure should be loaded.

p-datadog:
This PR was replaced by https://github.com/DataDog/dd-trace-rb/pull/4363.

PR #4359, opened 2025-02-10
Fix Rack integration to tag Content-Type and Content-Length headers

### LINKED ISSUE #1659: Rack integration does not tag content-length headers from HTTP requests
I would like to have the `Content-Length` request header added as a tag for a Rack application. I'm using the following configuration:
```
Datadog.configure do |c|
  c.use :rack, request_queuing: true,
    headers: {request: ["User-Agent", "Content-Length"], response: ["Content-Type", "X-Request-ID"]},
    analytics_enabled: true
end
```
I expect to see a `http.request.headers.content_length` tag, but do not see it.

I'm using:
- Ruby 2.6.5
- Rack 2.2.3
- ddtrace 0.51.1

This happens with behind WEBrick and Unicorn.

I think this is because these headers are passed through the environment hash as `'CONTENT_LENGTH` and `CONTENT_TYPE`. Since they are not prefixed with `'HTTP_'`, the parsing code misses them.
https://github.com/DataDog/dd-trace-rb/blob/a3236ddf0cead1c2a6db57acd9e4ebefbfd45026/lib/ddtrace/contrib/rack/middlewares.rb#L256-L264
https://github.com/DataDog/dd-trace-rb/blob/a3236ddf0cead1c2a6db57acd9e4ebefbfd45026/lib/ddtrace/contrib/rack/middlewares.rb#L282-L284

ybiquitous:
**What does this PR do?**

Closes #1659

The Rack spec says:

> The environment must not contain the keys HTTP_CONTENT_TYPE or HTTP_CONTENT_LENGTH (use the versions without HTTP_).

See https://github.com/rack/rack/blob/e217a399eb116362710aac7c5b8dc691ea2189b3/SPEC.rdoc?plain=1#L119-L121

This PR fixes an issue that it's impossible to send the `Content-Type` or `Content-Length` header values as a tag.

**Motivation:**

My motivation is the same as what #1659 describes. To tag the request HTTP headers `Content-Type` and `Content-Length`, the following configuration would be necessary:

```ruby
Datadog.configure do |c|
  headers = { request: ['Content-Type', 'Content-Length'] }
  c.tracing.instrument :rack, { headers: }
```

Without this PR's bug fix, the tagging above fails, i.e., the headers are ignored.

**Change log entry**

**Additional Notes:**

`Datadog::Tracing::Contrib::Rack::Header::RequestHeaderCollection` changed by this PR is used also by used `Datadog::AppSec::Contrib::Rack::RequestMiddleware`, but I don't think the PR affect it:

https://github.com/DataDog/dd-trace-rb/blob/fba70fdc02f7a4069a74bf864c47ce8ff7b44c09/lib/datadog/appsec/contrib/rack/request_middleware.rb?plain=1#L192

**How to test the change?**

This PR adds a few test cases to the existing unit test.


marcotc (approved):
Thank you so much, @ybiquitous!

Strech (approved):

PR #4360, opened 2025-02-10
DEBUG-3472 send snapshot and status events together

p-datadog:
**What does this PR do?**

Changes probe notifier worker to send both status and snapshot events in each iteration

**Motivation:**
When a probe executes for the first time, it emits a status event (that it is emitting) and the snapshot event (execution result). Currently these events are submitted to agent in different processing iterations, which upon reflection is not good - there was one action that caused both events to happen (probe execution), therefore both events should be submitted together.

There is also an issue in debugger-demo-ruby where the snapshot events are dropped between iterations for some reason. The root cause of this issue has not yet been identified but sending both events together works around the problem.

**Change log entry**
Yes: improve dynamic instrumentation event reporting 

**Additional Notes:**

**How to test the change?**
I tested manually against debugger-demo-ruby. I don't think we have a CI configuration presently that can test this change.

### DIFF
--- a/lib/datadog/di/probe_notifier_worker.rb
+++ b/lib/datadog/di/probe_notifier_worker.rb
@@ -263,7 +263,7 @@ module Datadog
 
       def maybe_send
         rv = maybe_send_status
-        rv || maybe_send_snapshot
+        maybe_send_snapshot || rv
       end
     end
   end


sarahchen6 (approved):
If both statuses should be sent together, why not use `maybe_send_snapshot && rv`?

PR #4361, opened 2025-02-10
Retry `bundle install` flakiness for JRuby 9.2

TonyCTHsu:
**What does this PR do?**

Mitigate JRuby 9.2 flakiness for bundle install

**Change log entry**

None.

### DIFF
--- a/tasks/github.rake
+++ b/tasks/github.rake
@@ -355,7 +355,22 @@ namespace :github do
     tasks.each do |task|
       env = { 'BUNDLE_GEMFILE' => task['gemfile'] }
       cmd = 'bundle check || bundle install'
-      Bundler.with_unbundled_env { sh(env, cmd) }
+
+      # For JRuby 9.2, the `bundle install` command failed ocassionally with the NameError.
+      #
+      # Mitigate the flakiness by retrying the command up to 3 times.
+      #
+      # https://github.com/jruby/jruby/issues/7508
+      # https://github.com/jruby/jruby/issues/3656
+      retries = 0
+      begin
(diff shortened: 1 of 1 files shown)


tasks/github.rake
```diff
@@ -355,7 +355,22 @@ namespace :github do
...
+      # For JRuby 9.2, the `bundle install` command failed ocassionally with the NameError.
+      #
+      # Mitigate the flakiness by retrying the command up to 3 times.
+      #
+      # https://github.com/jruby/jruby/issues/7508
+      # https://github.com/jruby/jruby/issues/3656
+      retries = 0
+      begin
+        Bundler.with_unbundled_env { sh(env, cmd) }
+      rescue StandardError
```
p-datadog, line 368: Would be nice to log the exception for troubleshooting purposes, especially considering this applies to all configurations in my reading?

p-datadog, line 368: Would be nice to log the exception for troubleshooting purposes, especially considering this applies to all configurations in my reading?

tasks/github.rake
```diff
@@ -355,7 +355,22 @@ namespace :github do
...
+        sleep(2**retries)
+        retries += 1
+        retry if retries < 3
+        raise
+      end
```
ivoanjo, line 373: Since this is only for JRuby 9.2, I suggest maybe having an explicit test for only enabling this workaround for JRuby 9.2. Otherwise we may "hide" other issues accidentally.

ivoanjo (approved):
:+1: LGTM

p-datadog (approved):

PR #4371, opened 2025-02-11
feat(propagation): baggage support and automatic propagation

ZStriker19:
**What does this PR do?**
Following the spec here: https://www.w3.org/TR/baggage/ we implement baggage for the Ruby tracer. The otel drop in support will be added in a future PR.
**Motivation:**

**Change log entry**
* Tracing: Added support for W3C Baggage extraction and injection by default.
* Tracing: Adds api method to access and modify baggage object.

**Additional Notes:**

**How to test the change?**
Tests were added in the repo and it passes the [weblog system-tests.](https://github.com/DataDog/system-tests/pull/3929) It will not pass the parametric system-tests because of the difficulty of managing the context in server.rb, not due an issue of implementation of the actual feature.


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+* Tracing: Added support for W3C Baggage extraction and injection by default.
+* Tracing: Adds api method to access and modify baggage object.
```
marcotc, line 8: No need to add these to this file. The section "Change log entry" in the PR description will make it here during release.

lib/datadog/tracing.rb
```diff
@@ -127,6 +127,13 @@ def log_correlation
         correlation.to_log_format
       end
```
marcotc, line 129: Because this is the main method in the new public API for baggage, we should add documentation about what it does, from the user perspective. Take a look at the other methods in this class for some good examples.

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -0,0 +1,125 @@
...
+          baggage_key: BAGGAGE_KEY
+        )
+          @baggage_key = baggage_key
+          @fetcher = fetcher
+        end
+
+        def inject!(digest, data)
+          return if digest.nil? || digest.baggage.nil?
+
+          baggage_items = digest.baggage.to_a.reject { |k, v| k.nil? || v.nil? }
```
marcotc, line 35: We can skip the `to_a` and avoid creating an intermediate list:
```suggestion
          baggage_items = digest.baggage.reject { |k, v| k.nil? || v.nil? }
```

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -0,0 +1,131 @@
...
+
+          TraceDigest.new(
+            baggage: baggage,
+          )
+        end
+
+        private
+
+        # We can't use uri encode because it incorrectly encodes some characters
+        def encode_key(key)
```
marcotc, line 84: Can you document what this method is doing? I'm having a hard time understanding the full logic here.

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -0,0 +1,131 @@
...
+
+              char = [hex_str.hex].pack('C')
+              SAFE_CHARACTERS_KEY.include?(char) ? char : encoded
+            else
+              encoded
+            end
+          end
+        end
+
+        def encode_value(value)
```
marcotc, line 98: Same for this one (docs).

marcotc: Actually, are these the pretty much the same, except for `SAFE_CHARACTERS_KEY` vs `SAFE_CHARACTERS_VALUE`? If so, I recommend making them one method that receives the `safe_characters...` as parameter.

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -0,0 +1,131 @@
...
+
+              char = [hex_str.hex].pack('C')
+              SAFE_CHARACTERS_VALUE.include?(char) ? char : encoded
+            else
+              encoded
+            end
+          end
+        end
+
+        def parse_baggage_header(baggage_header)
```
marcotc, line 112: A small doc for this would be great. Not as involved as the ones above, since this method is easier to read, but just a high level doc.

marcotc (approved):

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -96,19 +103,36 @@ def encode_key(key)
         end
 
         def encode_value(value)
```
marcotc, line 105: Are you able to merge encode_key and encode_value into one method, with the parameter for SAFE_CHARACTERS_VALUE or SAFE_CHARACTERS_KEY?

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -0,0 +1,133 @@
...
+          return unless baggage
+
+          TraceDigest.new(
+            baggage: baggage,
+          )
+        end
+
+        private
+
+        # We can't use uri encode because it incorrectly encodes some characters
```
p-datadog, line 83: Which characters are these?

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -0,0 +1,133 @@
...
+        def encode_item(item, safe_characters)
+          # Strip whitespace and URL-encode the item
+          result = CGI.escape(item.strip)
+          # Replace '+' with '%20' for space encoding consistency with W3C spec
+          result = result.gsub('+', '%20')
```
p-datadog, line 88: I don't know what other differences are between CGI and URI encoding besides +/%20, but this looks like you are bringing in the `cgi` dependency only to discard its unique behavior immediately after. If you are re-encoding/fixing the encoding anyway, can the initial encoding not be done via URI?

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -0,0 +1,133 @@
...
+        #
+        # @param baggage_header [String] The W3C Baggage header string to parse
+        # @return [Hash<String, String>] A hash of decoded baggage items
+        def parse_baggage_header(baggage_header)
+          baggage = {}
+          baggages = baggage_header.split(',')
+          baggages.each do |key_value|
+            key, value = key_value.split('=', 2)
+            # If baggage is malformed, return an empty hash
+            return {} unless key && value
```
p-datadog, line 120: My reading of this line (and the next `return {}`) is that it will completely discard valid parts of the payload if there are some invalid key/value pairs. This is generally not a good idea for interoperability and scanning https://www.w3.org/TR/baggage/ I didn't see where this behavior (dropping valid data) is called for.

lib/datadog/tracing/distributed/propagation.rb
```diff
@@ -57,8 +62,7 @@ def inject!(digest, data)
           end
 
           digest = digest.to_digest if digest.respond_to?(:to_digest)
-
-          if digest.trace_id.nil?
+          if digest.trace_id.nil? && digest.baggage.nil?
             ::Datadog.logger.debug('Cannot inject distributed trace data: digest.trace_id is nil.')
```
p-datadog, line 66: The message here should be updated for the changed behavior (consulting baggage in addition to trace_id), no?

lib/datadog/tracing/distributed/propagation.rb
```diff
@@ -138,12 +142,29 @@ def extract(data)
               "Error extracting distributed trace data. Cause: #{e} Location: #{Array(e.backtrace).first}"
             )
           end
+          # Handle baggage after all other styles if present
```
p-datadog, line 145: "if present" is redundant/misleading in the comment - the code to "handle baggage", i.e. the next line, will always be executed.

lib/datadog/tracing/trace_operation.rb
```diff
@@ -101,6 +103,7 @@ def initialize(
         @trace_state = trace_state
         @trace_state_unknown_fields = trace_state_unknown_fields
         @tracer = tracer
+        @baggage = baggage || {}
```
p-datadog, line 106: The diff appears to mix having baggage be always present, like on this line, and optional like on line 338 below in the same file. Given that baggage is "always propagated if present" I think it would make more sense to not vivify it absent an explicit instruction, i.e. either an explicit write or the baggage came via propagation, to save on processing effort downstream.

lib/datadog/tracing.rb
```diff
@@ -127,6 +127,19 @@ def log_correlation
...
+      # Returns the baggage for the current trace.
+      #
+      # If there is no active trace, a new one is created.
+      #
+      # @return [Datadog::Tracing::Distributed::Baggage] The baggage for the current trace.
+      # @public_api
+      def baggage
+        # Baggage should not be dependent on there being an active trace.
+        # So we create a new TraceOperation if there isn't one.
+        active_trace = self.active_trace || tracer.continue_trace!(nil)
```
p-datadog, line 139: As someone without an in-depth knowledge of the tracing part of the library, I am wondering what other attributes of Tracing would cause the trace to be created, and also whether this method is meant to be consumed by customers or by dd-trace-rb internally.

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -0,0 +1,133 @@
...
+
+          TraceDigest.new(
+            baggage: baggage,
+          )
+        end
+
+        private
+
+        # We can't use uri encode because it incorrectly encodes some characters
+        def encode_item(item, safe_characters)
```
p-datadog, line 84: Maybe I missed it but I am not seeing a unit test for this method in the diff. Given that it performs custom escaping I would think a unit test would be appropriate with at least the special cases handled in the implementation.

lib/datadog/tracing/distributed/baggage.rb
ZStriker19: That's actually the original behavior I had but was then failing a system-test, from the RFC: _When APM SDKs are extracting baggage from propagation headers, they may encounter malformed header contents. When this occurs, the APM SDK should not try to extract any values from the entire header. In other words, APM SDKs should ignore the header and not try to extract "good" values while ignoring "bad" ones._

lib/datadog/tracing.rb
ZStriker19: Since baggage specifically is supposed to be able to be added independent of a span starting, we needed to make sure we created a traceoperation if there wasn't one already available for it. I don't know of any other attributes for which this must be true. The method is meant to be consumed by customers to access the baggage object. What would you recommend doing to make that more explicit than having `# @public_api` above it?

lib/datadog/tracing/distributed/baggage.rb
ZStriker19: We test inject and extract with special characters in baggage_spec.rb, is that not sufficient?

lib/datadog/tracing/trace_operation.rb
ZStriker19: Hmmm so keep it nil at the start? However we'll still need to check if it's empty if a customer potentially sets a baggage item, and then removes it, we'd then have an empty hash. With regards to processing effort downstream, I don't think it would make a very large difference since as soon as inject is called for baggage we exit if the baggage value is nil or empty. Let me know if I'm misunderstanding.

PR #4372, opened 2025-02-12
Run test agent check in Github Actions

TonyCTHsu:
**What does this PR do?**

Run test agent checks in GHA

**Change log entry**
None


.github/scripts/test_agent_check.rb
```diff
@@ -0,0 +1,45 @@
...
+  if summary_response.code == "200"
+    puts "APM Test Agent is running. (HTTP 200)"
+  else
+    puts "APM Test Agent is not running and was not used for testing. No checks failed."
+    exit 0
+  end
+
+  # Check for test failures
+  failures_uri = URI.parse("http://#{agent_host}:#{agent_port}/test/trace_check/failures")
+  failures_response = Net::HTTP.get_response(failures_uri)
```
Strech, line 25: Just to be sure, do we want to put some timeout settings on the request? Just in case?

TonyCTHsu: The change is actually migrate from `.circleci` with extracting environment variables. 

This has been running for quite some time, so I guess it is work quite well. Of course, we can always revisit this if we  encounter some issue.

Issue #3482, opened 2024-02-25
`rack.request` span that are stopped in middleware or before action doesn't have resource set


beauraF:
Hello everyone 👋

I'm opening a dedicated bug report following [this discussion](https://github.com/DataDog/dd-trace-rb/issues/2100#issuecomment-1603933814).

This has a significant impact on the monitoring and alerting systems implemented by our teams. Since these spans have no resources, it is not possible to infer a team from them. This broke the reporting of dashboard built by team, and also our custom code that set a service to a `rack.request` based on the resource.

I've already seen several teams remove `before_action`, by calling directly into the controller's action to workaround this behavior ; after often spending several hours before understanding this subtlety.

@delner you made a proposal in the initial discussion, I'd be happy to work on it if that's still the direction you want to go in? 

---

**Current behaviour**
`rack.request` span that are stopped in middleware or before action doesn't have resource set

**Expected behaviour**
`rack.request` span that are stopped in before action have resource set. Middleware also? 

**Environment**

* **ddtrace version:** 1.20.0
* **Ruby version:** 3.2.3

lloeki:
Hey @beauraF about this:

> you made a proposal in the initial discussion, is that's still the direction you want to go in?

are you referring to this?

> In theory, this could be done at a routing level if we instrumented the routing behavior. That would always run before any controllers, but should be able to show a route name.

beauraF:
Hello @lloeki 👋 Yes, I think so!

lloeki:
> our custom code that set a service to a `rack.request` based on the resource

That's interesting, does this mean you're overriding the `rack.request` span service name?

> In theory, this could be done at a routing level if we instrumented the routing behavior

FYI it's becoming less of a theory now: we're working on [adding routing information](https://github.com/DataDog/dd-trace-rb/pull/3849) to the `rack.request` span. This was worked on before but reverted because of issues with Rails 7.1 so we're redoing it.

What the PR does is tag the composite (because of nested apps/engines and their intermediate resolvers) route, but the same hooking process could _also_ get the *target* of each route as resolved by a route resolver, which you could peruse.

> I've already seen several teams remove before_action, by calling directly into the controller's action to workaround this behavior

I'd argue that including `before/after_action` callbacks inside the _controller_ span would be incorrect and misleading as they are not executed within the lifetime of the controller's action.

I can definitely see an argument for:

- having a span for each callback, which would allow to precisely and semantically identify which callback is responsible for whatever time is spent
[...]

Issue #2100, opened 2022-06-22
Before/after actions not included in the action_pack span


viraptor:
**Current behaviour**

Rails callbacks (before/after) are not a part of the action span.
In the `before_action` callback, active_span is:
```
SpanOperation(name:rack.request,sid:1008860920845926479,tid:1866324838581500214,pid:0)
```
But in the action call:
```
SpanOperation(name:rails.action_controller,sid:3701457545195306293,tid:1866324838581500214,pid:1008860920845926479)
```

**Expected behaviour**

active_pack spans should include the callbacks. That could be done either through a separate span or by extending the current one.

I'm raising this as a bug rather than a feature request, because otherwise we end up in situations where (for example) rack seems to own spans from `active_record`, which is not true: 
![Screen Shot 2022-06-22 at 4 08 39 pm](https://user-images.githubusercontent.com/188063/174955773-87b1c13e-ac53-405a-bc71-6f405ffc68cd.png)
Those queries are made in the `before_action` on the controller.

**Steps to reproduce**

```
class SomeController < ApplicationController
  before_action :bar

  def foo
    Datadog::Tracing.trace("this.goes.in.controller") {}
  end

  private
  
  def bar
    Datadog::Tracing.trace("this.goes.in.rack") {}
  end
end
```

**Environment**

* **ddtrace version:** 1.1.0
* **Configuration block (`Datadog.configure ...`):**
```
    c.tracing.instrument :rails, service_name: 'marketplace'
    c.tracing.instrument :rack, request_queuing: true, web_service_name: 'marketplace-rack'
```
* **Ruby version:** 2.7
* **Operating system:** linux
* **Relevant library versions:**

delner:
ActionPack instrumentation doesn't cover `before/after_action`; this is known behavior. We wanted to add coverage for this, but the nature of how it is implemented in Rails (as an ActiveSupport notification) makes it incredibly difficult. See this ancient PR: https://github.com/DataDog/dd-trace-rb/pull/333

viraptor:
So I've got an ApiController base class which knows how to name the resource correctly, but doesn't actually implement any endpoint. Only controllers inheriting from it actually have the actions and I don't want to replicate the resource naming in them.
The problem is that the `Class#method` naming in the API classes doesn't really map nicely to the requests (either being too general or too specific), so I want to use a custom name that depends on request parameters. 

I can't (as far as I know) use the span processor / filter here, because I don't have access to the parameters anymore.

[...]

PR #4375, opened 2025-02-12
Patch `ActionController::Base` and `API` instead of `Metal`

moberegger:
**What does this PR do?**

Makes a small adjustment to _where_ the patch for `rails.action_controller` is applied. Originally it `prepend`ed `ActionController::Metal`, but this PR proposes to move it earlier to `ActionController::Base` and `ActionController::API`.

**Motivation:**

In Rails, you will commonly see something like

```ruby
class MyController < ApplicationController
  # Rescue and gracefully handle error response
  rescue_from MyError, with: :render_error_response

  def index
    # ...
    fail MyError if no_good # On some condition, raise an error
  end
end
```

These are exceptions that are being handled gracefully, but Datadog will tag the `rails.action_controller` with errors. This causes APM to report errors on these controller actions, which can also cause alerts via the error tracking feature. The problem here is that this is reporting on something that is in fact being handled correctly by the code, but the tracer simply isn't aware of it.

This is happening because `datadog` is patching `ActionController::Metal`, which adds the [exception to the payload](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/action_pack/action_controller/instrumentation.rb#L117-L121) before the `ActionController` has a chance to [run the rescue handler](https://github.com/rails/rails/blob/main/actionpack/lib/action_controller/metal/rescue.rb#L26-L31).

There is a workaround, but it requires you to configure [`action_dispatch.rescue_responses` ](https://guides.rubyonrails.org/configuring.html#config-action-dispatch-rescue-responses); `datadog` will [check this](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/contrib/action_pack/utils.rb#L15) to determine whether or not this is an expected exception. But this isn't really tenable; that configuration is largely intended for Rails' `ShowExceptions` to handled raised exceptions, so you wouldn't typically configure these with exceptions that you are already handling in the controller.

By simply moving the patch, the entirety of `process_action` is traced, which gives Rails a chance to run the rescue handlers.

**Change log entry**

**Additional Notes:**

This may address a few existing issues:
- https://github.com/DataDog/dd-trace-rb/issues/3482: This one seems to be reporting that the `rack.request` resource is not correctly set when a `before_action` raises an exception. This is because with the original patch, the action callbacks ran _before_ the trace; now they will run _under_ the trace. This will allow the patch to now set the `rack.request` resource before any `before_action`s run.
- https://github.com/DataDog/dd-trace-rb/issues/2100: Reports that the action callbacks are not included in the trace (ie. they happen before the `rails.action_controller` span). This will no longer be the case. This is worth mentioning because there appears to be some differing opinions around whether or not this should be included as part of the `rails.action_controller` span. I personally think it should, because it _is_ part of the `process_action` call.

With the proposed change, it means more work is being traced. Given that the `rails.action_controller` span is intended to measure the controller action, I do think that this is more representative of what is happening in the controller.

I have chosen to also patch `ActionController::API` to support Rails' `api_mode`. Unsure if this needs to be smarter by first checking if the module exists before attempting to apply the patch. I can make that change if need be.

**How to test the change?**

I did follow the guide in https://github.com/DataDog/dd-trace-rb/blob/master/docs/DevelopmentGuide.md, and it does appear that the existing suite is still green. I may not have done it right, though. The change seems pretty superficial to me, but if there are any tests that I should add or modify, please let me know. I'm really not sure what to do here, my apologies.

I have applied this patch manually in our own API application, and everything is working as expected as far as I can tell. I don't expect "just trust me!" to be super convincing, though 😆.

### DIFF
--- a/lib/datadog/tracing/contrib/action_pack/action_controller/patcher.rb
+++ b/lib/datadog/tracing/contrib/action_pack/action_controller/patcher.rb
@@ -19,7 +19,8 @@ module Datadog
             end
 
             def patch
-              ::ActionController::Metal.prepend(ActionController::Instrumentation::Metal)
+              ::ActionController::Base.prepend(ActionController::Instrumentation::Metal)
+              ::ActionController::API.prepend(ActionController::Instrumentation::Metal)
             end
           end
         end


mscrivo:
hey @ivoanjo, @moberegger is a colleague of mine. Could we possibly get some eyes on this one in the near future? 🙇🏼

ivoanjo:
Hey @mscrivo @moberegger thanks for the ping/PR, let me go poke folks to take a look at this :)

marcotc:
👋 @moberegger thank you so much for this fix!

This change will be a breaking change, because users that for example raised errors in their `before_action`s will now see a span with error, when in the past they didn't, which can trigger Error Tracking and monitors.

Because of this, I recommend adding an option to toggle this behaviour (default to "current behaviour") at the integration level. In this case, here: https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#action-pack
The name of the option is honestly anything that makes the intent clear, one suggestion can be `c.tracing.instrument :action_pack, include_callbacks: true/false`, but I'll leave it to you what you think sounds clearer.

And in terms of testing, I suggest adding a test that raises an error during a `before_action` callback: this will behave differently whether the option is enabled or not.

Let me know if you have any concerns, or need any extra support here!

moberegger:
>  because users that for example raised errors in their before_actions will now see a span with error, when in the past they didn't, which can trigger Error Tracking and monitors.

Oof, I didn't consider this. We're not raising errors from hooks in our project so I never saw this as a possibility. 

I'll kick tires on this some more.

mgodwin:
I also find this behavior surprising and want to advocate for this PR making it in.

I wonder how many people would be bitten by the before_action errors being ignored?  If anything I would think that those should potentially be surfaced in error tracking 🤔 ?

Would it be possible to have an option to opt-in to patching ActionController::Base/API and possibly move towards deprecating the ActionController::Metal patch?

nightpool:
I've also been blocked on this—I don't think this should be considered a breaking change though, because the addition of errors is only tracking *new* data that was previously untracked.

PR #4391, opened 2025-02-14
Add Faraday instrumentation for AppSec

y9v:
**What does this PR do?**
This PR adds Faraday instrumentation to AppSec for Server-Side Request Forgery detection.

**Motivation:**
We want to detect SSRF attacks on applications using Faraday as the http client. Support for more http clients (`excon` and `rest-client`) will be added in subsequent PRs.

**Change log entry**
Yes. AppSec: Add detection of Server-Side Request Forgery attacks for Faraday http client.

**Additional Notes:**
Patcher is heavily inspired by our Tracing Faraday patcher.

**How to test the change?**
CI and manual testing with app-generator (`rails-ssrf` variant)


marcotc (approved):

lib/datadog/appsec/contrib/faraday/middleware.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Datadog
+  module AppSec
+    module Contrib
+      module Faraday
+        # Middleware for Faraday
+        class Middleware < ::Faraday::Middleware
```
Strech, line 8: I think the class name should describe "what" because now it's stating a fact which is not helpful if you want to understand what is going to happen.

And the most important, we have a check that `Middleware` is not already in the list of middlewares, hence we have non 0% chances that someone decides to call their middleware as `Middleware`

lib/datadog/appsec/contrib/faraday/middleware.rb
```diff
@@ -0,0 +1,40 @@
...
+          def call(request_env)
+            context = AppSec.active_context
+
+            if context && AppSec.rasp_enabled?
```
Strech, line 12: I think it's better with guard clause than nesting

lib/datadog/appsec/contrib/faraday/patcher.rb
```diff
@@ -0,0 +1,60 @@
...
+            require_relative 'middleware'
+            require_relative 'connection'
+
+            register_middleware!
+            add_default_middleware!
+
+            Patcher.instance_variable_set(:@patched, true)
+          end
+
+          def register_middleware!
```
Strech, line 33: I think this method is irresponsible.

lib/datadog/appsec/contrib/faraday/patcher.rb
```diff
@@ -0,0 +1,60 @@
...
+            ::Faraday::Middleware.register_middleware(datadog_appsec: Middleware)
+          end
+
+          def add_default_middleware!
```
Strech, line 37: I think this name is misleading, what does it mean `default` in this case? Default faraday middleware or default something else?

lib/datadog/appsec/contrib/faraday/connection.rb
```diff
@@ -0,0 +1,22 @@
...
+module Datadog
+  module AppSec
+    module Contrib
+      module Faraday
+        # Handles installation of our middleware if the user has *not*
+        # already explicitly configured our middleware for this correction.
+        #
+        # Wraps Faraday::Connection#initialize:
+        # https://github.com/lostisland/faraday/blob/ff9dc1d1219a1bbdba95a9a4cf5d135b97247ee2/lib/faraday/connection.rb#L62-L92
+        module Connection
```
Strech, line 12: I would like to ask for a better name. It's hard to understand what this is, should we use at least `ConnectionPatch`, but this name is also not perfect, but at least it answers the question "what is it"

lib/datadog/appsec/contrib/faraday/patcher.rb
```diff
@@ -0,0 +1,60 @@
...
+
+require_relative '../patcher'
+
+module Datadog
+  module AppSec
+    module Contrib
+      module Faraday
+        # Patcher for Faraday
+        module Patcher
+          include Datadog::AppSec::Contrib::Patcher
```
Strech, line 11: I would suggest us to stop using an empty module

spec/datadog/appsec/contrib/http/faraday_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+  end
+
+  before do
+    Datadog.configure do |c|
+      c.appsec.enabled = true
+      c.appsec.instrument :faraday
+    end
+
+    Datadog::AppSec::Context.activate(context)
```
Strech, line 29: Is it possible to replace it with stab with a mock object instead?

spec/datadog/appsec/contrib/http/faraday_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+      response = client.get('/success')
+
+      expect(response.status).to eq(200)
+      expect(response.body).to eq('OK')
+    end
+  end
+
+  context 'when there is no active context' do
+    before do
+      Datadog::AppSec::Context.deactivate
```
Strech, line 60: This is a sign of a leaking context because of the comment above ⬆️

spec/datadog/appsec/contrib/http/faraday_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+
+      client.get('/success')
+    end
+
+    it 'returns the http response' do
+      response = client.get('/success')
+
+      expect(response.status).to eq(200)
+      expect(response.body).to eq('OK')
+    end
```
Strech, line 74: IMHO response check should be a part of the general test suite because on it's own we don't care much about it, i.e let's merge em!

spec/datadog/appsec/contrib/http/faraday_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+
+    it 'calls waf with correct arguments when making a request' do
+      expect(Datadog::AppSec.active_context).to(
+        receive(:run_rasp).with(
+          Datadog::AppSec::Ext::RASP_SSRF,
+          {},
+          { 'server.io.net.url' => 'http://example.com/success' },
+          Datadog.configuration.appsec.waf_timeout
+        ).and_call_original
```
Strech, line 89: I have that feeling that it looks like an integration test, but instead of behavior testing we are doing a mix of unit-testing here, it will be good to do just a single way.

This explain why do we need to configure rules, because if we do unit-test we should not, but if we do integration-test - we do.

spec/datadog/appsec/contrib/integration/faraday_ssrf_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    spans.find { |s| s.name == 'rack.request' }
+  end
+
+  context 'when RASP check triggered for http request' do
+    before do
+      get('/ssrf', { 'url' => '169.254.169.254' }, { 'REMOTE_ADDR' => '127.0.0.1' })
+    end
+
+    it { expect(last_response).to be_forbidden }
+  end
```
Strech, line 106: Well done! The only missing piece here is a non-blocking check in order to guarantee that we are not blocking in 100% of the cases

Strech (changes requested):
Awesome to see that we have a solid ground to develop integrations. I left a few comments over naming and testing because I think we could do better and simplify even more

lib/datadog/appsec/contrib/faraday/middleware.rb
y9v: but what is the chance that someone will name their middleware `Datadog::AppSec::Contrib::Faraday::Middleware`?

lib/datadog/appsec/contrib/faraday/middleware.rb
y9v: I decided against it because with a guard clause we have to write the call `@app.call(request_env)` twice - once in the guard clause, and the second time at the end of this method. What do you think, should I still do the guard clause?

lib/datadog/appsec/contrib/faraday/patcher.rb
y9v: yes, you are right, let's call it `add_default_faraday_middleware`.
Context for this is - Faraday has a set of default middlewares that are used for all clients, and you can also add middlewares to each client separately.

spec/datadog/appsec/contrib/http/faraday_spec.rb
y9v: I think the goal of this test is to ensure that we are patching faraday correctly (different versions of the gem). I think that it is enough to test that we are calling `run_rasp` with correct arguments?

lib/datadog/appsec/contrib/faraday/patcher.rb
y9v: oh wow, thanks for the tip, I actually haven't seen that it's empty

y9v: @Strech btw, what do you say if we delete this module in a subsequent PR? (I mean I would still remove this iniclude from this PR of course)

lib/datadog/appsec/contrib/faraday/middleware.rb
y9v: yes, let's maybe call it `SSRFDetectionMiddleware`?

lib/datadog/appsec/contrib/faraday/middleware.rb
Strech: The main point - I don't need to keep in mind the context of the condition, I will focus on the main part which is - running rasp.

If I will outline it

```ruby
def call(...)
  context = ""

  if context && AppSec.rasp?
    result = ""
    if result.match?
      handle(...)
    end
  end

  @app.call(..)
end
```

versus

```ruby
def call(...)
  context = ""
  return @app.call(..) if context && AppSec.rasp?
  
  result = ""
  
  if result.match?
      handle(...)
    end
  end

  @app.call(..)
end
```

lib/datadog/appsec/contrib/faraday/patcher.rb
Strech: At first we can stop using it, yes, I have it cleaned, just need to open PR with that

spec/datadog/appsec/contrib/http/faraday_spec.rb
Strech: Yes, but why call original then? We can stop at the RASP check, right?

lib/datadog/appsec/contrib/faraday/patcher.rb
Strech: I would rephrase the verb then, we trying to configure (not add) the default faraday middlewares list, right?

spec/datadog/appsec/contrib/http/faraday_spec.rb
y9v: yes, you're right, it's already part of the integration test. And if we don't run RASP, we can also work with a context stub here.

lib/datadog/appsec/contrib/faraday/patcher.rb
y9v: yes, let's call it `configure_default_faraday_connection`

lib/datadog/appsec/contrib/faraday/middleware.rb
y9v: changed naming and refactored patched a little bit in [Improve naming in AppSec faraday patcher](https://github.com/DataDog/dd-trace-rb/pull/4391/commits/22b711546a2b85aec9faf2c07e21881bee16b672)

spec/datadog/appsec/contrib/http/faraday_spec.rb
y9v: changed specs in this commit [b7fbdd1](https://github.com/DataDog/dd-trace-rb/pull/4391/commits/b7fbdd1eb6a0e894e1e10c8315c406f42fb22d29)

spec/datadog/appsec/contrib/http/faraday_spec.rb
```diff
@@ -0,0 +1,78 @@
+# frozen_string_literal: true
+
+require 'datadog/appsec/spec_helper'
+require 'faraday'
+
+RSpec.describe 'AppSec Faraday integration' do
+  let(:waf_response_double) { double(match?: false) }
```
Strech, line 7: Should it be `instance_double` of an actual class to avoid stubbing non-existent methods?

spec/datadog/appsec/contrib/http/faraday_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+  let(:active_context) { stub_const('Datadog::AppSec::Context', double(run_rasp: waf_response_double)) }
```
Strech, line 8: I don't think it's a good idea, it allows us to craft what ever interface we want without any connection to the real code.

Instead, I would suggest this

```ruby
let(:context) { instance_double(Datadog::AppSec::Context, run_rasp: waf_response) }
let(:waf_response) { instance_double(Datadog::AppSec::SecurityEngine::Result, match?: false) }
```

_P.S I don't think that in tests it makes sense to add "something_double" to variables, because we are in tests and it should not matter until they behave the same_

spec/datadog/appsec/contrib/integration/faraday_ssrf_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+
+    stack.to_app
+  end
+
+  let(:http_service_entry_span) do
+    Datadog::Tracing::Transport::TraceFormatter.format!(trace)
+    spans.find { |s| s.name == 'rack.request' }
+  end
+
+  context 'when RASP check was positive for http request' do
```
Strech, line 100: To be honest it's confusing wording, is the positive here a good or a bad thing? Maybe we can say

> when request contains SSRF attack

lib/datadog/appsec/contrib/faraday/ssrf_detection_middleware.rb
```diff
@@ -0,0 +1,42 @@
+# rubocop:disable Naming/FileName
+# frozen_string_literal: true
+
+module Datadog
+  module AppSec
+    module Contrib
+      module Faraday
+        # AppSec SSRF detection Middleware for Faraday
+        class SSRFDetectionMiddleware < ::Faraday::Middleware
```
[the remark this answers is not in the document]
y9v: yes, I think it would be better to have the require here, since we use the class here

spec/datadog/appsec/contrib/http/faraday_spec.rb
y9v: good point, thanks!

spec/datadog/appsec/contrib/integration/faraday_ssrf_spec.rb
y9v: yes, agree

spec/datadog/appsec/contrib/http/faraday_spec.rb
y9v: totally agree

lib/datadog/appsec/contrib/faraday/ssrf_detection_middleware.rb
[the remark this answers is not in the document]
y9v: I think it is not necessary - we are checking that `Faraday` is loaded in the integration. And then we also use `::Faraday::Middleware` in the `Patcher` class when registering middleware.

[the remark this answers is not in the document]
y9v: And `::Faraday::Middleware` is part of the faraday public interface, since it is used for registering custom middlewares

spec/datadog/appsec/contrib/integration/faraday_ssrf_spec.rb
y9v: fixed all test related stuff in [0158995](https://github.com/DataDog/dd-trace-rb/pull/4391/commits/015899562ebb4f3389d035bcb59638291c29bc0a).

Thanks for the review!

PR #4395, opened 2025-02-17
Remove template with reusable workflow

TonyCTHsu:
**Motivation:**

The templating technique was working fine, however, there are some caveats. 

1. Cannot write comment directly with YAML conversion
2. Missing automation to check whether the files are out of synced.
3. Content conflict with dependabot, which does not provide option for file omission.
4. An extra layer of indirection making devs hard to change

**What does this PR do?**

Replace templating with reusable workflow for each Ruby. 

**Change log entry**

None.


Strech (approved):
Amazing 👏🏼

.github/workflows/test.yml
```diff
@@ -1,8 +1,6 @@
-# Please do NOT manually edit this file.
-# This file is generated by 'bundle exec rake template:github:test'
```
marcotc, line 2: Do we still use the `template:github:test` rake task?

PR #4403, opened 2025-02-18
APMS-14883 Add option for Mongo span resource as JSON

marcotc:
**Motivation:**
A MongoDB command is expected to be serialized as JSON, but we were doing `Hash#to_s` instead. This causes issues in the Agent and Datadog App. There's no advantage to keep using `Hash#to_s`, but it's a breaking change to update this since it affects the resource name of the `mongo.cmd` span.

**Change log entry**
Yes. Add MongoDB option to serialize Mongo command as a JSON (instead of Hash#to_s). The Datadog App expects the Mongo command to be a JSON, so all MongoDB users are highly encouraged to enable this option: `DD_TRACE_MONGO_JSON_COMMAND=1` or `c.tracing.instrument :mongo, json_command: true`. This change will modify the resource name for the span `mongo.cmd`, which can be a breaking change for monitors and dashboards that depend on this resource name.

APMS-14883


TonyCTHsu:
Lets extract `mongo` to its own dependency group, a similar approach can be find here: https://github.com/DataDog/dd-trace-rb/pull/4405

marcotc:
> Lets extract `mongo` to its own dependency group, a similar approach can be find here: #4405

I tried and it adds 100+ files to this PR, so I'll do it in a follow up.

marcotc:
> > Lets extract `mongo` to its own dependency group, a similar approach can be find here: #4405
> 
> I tried and it adds 100+ files to this PR, so I'll do it in a follow up.

Nvm, after the rebase I notice the `gemfiles` changes were already done, so I'm adding it to this PR.

spec/support/core_helpers.rb
```diff
@@ -43,6 +47,19 @@ module CoreHelpers
     def supports_value_expectations?
       false
     end
+
+    # Add support for `once`, `at_least`, and other chained matchers from `RSpec::Mocks::Matchers::Receive`
```
marcotc, line 51: This is used to allow for `expect { subject }.to log_deprecation('deprecated_option').at_least(1).time`.

aliciascott (approved):

p-datadog (approved):

PR #4418, opened 2025-02-20
Use dynamic evaluation instead of static path filter

lloeki:
**What does this PR do?**

Use dynamic evaluation instead of static path filter

**Motivation:**

- system tests and parametric tests should be required to pass
- system tests and parametric tests are useless to run on known changesets
- path filters mess with required status checks

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**


TonyCTHsu (commented):
I would like to see this change to avoid branches that are not ready for pull requests.

```
on:
  push:
    branches:
    - master
  pull_request:
    branches:
    - master
```

sarahchen6 (approved):

.github/workflows/system-tests.yml
```diff
@@ -391,7 +407,8 @@ jobs:
       - build-harness
       - build-apps
       - test
-    if: ${{ always() }}
+      - changes
+    if: ${{ needs.changes.outputs.changes == 'true' && always() }}
```
TonyCTHsu, line 411: `always()` or `!cancel()` ?

lloeki: Good point, fixed

PR #4433, opened 2025-02-24
[APPSEC-56683] Introduce automated user tracking V3

Strech:
**What does this PR do?**

TBH this PR is doing a lot, but this is the only way to switch to the V3, but to outline the main theme:

1. Overhaul to existing devise instrumentation with a new feature of tracking authenticated users
2. Update to the SDK with adjustments to the tags we set by it

**Motivation:**

This is a [latest version of the RFC](https://docs.google.com/document/d/1RT38U6dTTcB-8muiYV4-aVDCsT_XrliyakjtAPyjUpw/edit?pli=1&tab=t.0) required to satisfy ATO (Account Take Over) Protection

**Change log entry**

Yes. AppSec: Improve `devise` instrumentation to support latest Account Takeover (ATO) detection.

**Highlights**

In this release, devise instrumentation was refreshed to track authenticated users, in addition to sign-in and sign-up events. A new anonymization collection mode allows sensitive data to be anonymized.

**Additional Notes:**

1. In the integration tests I need to emulate the gem first load to be able to have reproducible and testable behavior, but this require a hack I'm not happy about.
2. Rails setup takes 150 lines, but I don't want to end up with a similar to tracing setup where I need to set some magic variables to make it work. But definitely I can rewrite it as I get the full picture. 
3. Potentially it might happen that login tags are empty, but to handle that we will need to complicate the code, and potentially we are going to rewrite our event dispatching system which will eliminate the issue by design (we could send event with data and act properly, but current implementation is limited and slightly inefficient).

**How to test the change?**

I have the closes to real integration tests and system-tests covers rest.


spec/datadog/appsec/contrib/integration/devise/authenticated_single_user_tracking_spec.rb
```diff
@@ -5,6 +5,7 @@
 require 'rack/test'
 
 require 'action_controller/railtie'
+require 'action_mailer'
```
Strech, line 8: Is it still needed?

lib/datadog/appsec/contrib/devise/data_extractor.rb
```diff
@@ -0,0 +1,79 @@
...
+  module AppSec
+    module Contrib
+      module Devise
+        # Extracts user identification data from Devise resources.
+        # Supports both regular and anonymized data extraction modes.
+        class DataExtractor
+          PRIORITY_ORDERED_ID_KEYS = [:id, 'id', :uuid, 'uuid'].freeze
+          PRIORITY_ORDERED_LOGIN_KEYS = [:email, 'email', :username, 'username', :login, 'login'].freeze
+
+          def initialize(mode)
```
y9v, line 15: what do you think, does it make to use a keyword argument here instead? To me it's not obvious that data extractor receives the mode as an argument

lib/datadog/appsec/contrib/devise/patches/signup_tracking_patch.rb
```diff
@@ -0,0 +1,69 @@
...
+
+              super do |resource|
+                context = AppSec.active_context
+
+                if context.trace.nil? || context.span.nil?
+                  Datadog.logger.debug { 'AppSec: unable to track signup events, due to missing trace or span' }
+                  next yield(resource) if block_given?
+                end
+
+                next yield(resource) if !resource.persisted? && block_given?
```
y9v, line 27: not a blocker: I think the opposite of `.persisted?` is `.new?`, in case you want to get rid of negation here

lib/datadog/appsec/contrib/devise/patches/signup_tracking_patch.rb
```diff
@@ -0,0 +1,69 @@
...
+
+                id_tag = resource.active_for_authentication? ? Ext::TAG_USR_ID : Ext::TAG_SIGNUP_USR_ID
+                context.span[id_tag] ||= id
+              end
+
+              # NOTE: We don't have a way to make one-shot receivers for events,
+              #       and because of that we will trigger an additional event even
+              #       if it was already done via the SDK
+              AppSec::Instrumentation.gateway.push(
+                'identity.set_user', AppSec::Instrumentation::Gateway::User.new(id, login)
```
y9v, line 61: Does devise always signs the user in on signup? I think there could be configurations where the signup needs to be confirmed first

lib/datadog/appsec/contrib/devise/tracking_middleware.rb
```diff
@@ -0,0 +1,91 @@
+# frozen_string_literal: true
+
+require_relative 'ext'
+require_relative '../../anonymizer'
+
+module Datadog
+  module AppSec
+    module Contrib
+      module Devise
+        # A Rack middleware capable tracking currently signed user
```
y9v, line 10: typo: "capable of tracking"

y9v (approved):
great work!

lib/datadog/appsec/contrib/devise/patches/signup_tracking_patch.rb
Strech: Yes, you are right and it's here https://github.com/DataDog/dd-trace-rb/pull/4433/files/98bab082e43cb62e216b4ca994fccf0b2dd8d44f#diff-6c4a66834380777cf90181891c226e88cddf64beb9018626617d24a8e9ff6cf4R50-R55

We will check is it ready for authentication and if yes - we assign special tags, but logically we have a successful signup with assigned ID (if available) so we run WAF (this event is ultimately just run the check)

lib/datadog/appsec/contrib/devise/patches/signup_tracking_patch.rb
Strech: Not sure how relevant for us to check that new record wasn't destroyed, but we indeed can use `new_record?`

https://github.com/rails/rails/blob/v4.2.9/activerecord/lib/active_record/persistence.rb#L83-L100

```
    # Returns true if this object hasn't been saved yet -- that is, a record
    # for the object doesn't exist in the database yet; otherwise, returns false.
    def new_record?
      sync_with_transaction_state
      @new_record
    end

    # Returns true if this object has been destroyed, otherwise returns false.
    def destroyed?
      sync_with_transaction_state
      @destroyed
    end

    # Returns true if the record is persisted, i.e. it's not a new record and it was
    # not destroyed, otherwise returns false.
    def persisted?
      !(new_record? || destroyed?)
    end
```

Strech: I decided to go with `new_record?`

lib/datadog/appsec/contrib/devise/patches/signup_tracking_patch.rb
y9v: nice!

p-datadog (approved):
I read the `lib` part of the diff but without comments I cannot follow the logic in the majority of the appsec code and verify that, for example, method implementations are reasonable according to what I would expect the methods to do (since in most cases I am not able to figure out/articulate what any of the methods is *supposed* to be doing).

lib/datadog/appsec/contrib/devise/patcher.rb
```diff
@@ -21,29 +28,35 @@ def target_version
...
-
-          def patch_authenticatable_strategy
-            ::Devise::Strategies::Authenticatable.prepend(AuthenticatablePatch)
-          end
+            ::ActiveSupport.on_load(:before_initialize) do |app|
+              GUARD_ONCE_PER_APP[app].run do
+                begin
+                  app.middleware.insert_after(Warden::Manager, TrackingMiddleware)
+                rescue RuntimeError
+                  AppSec.telemetry.error('AppSec: unable to insert Devise TrackingMiddleware')
```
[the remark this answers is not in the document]
Strech: In this specific case there is no point to have exception because it doesn't bring the value, message will be custom and type of the exception I know already. Thanks for suggestion!

lib/datadog/appsec/contrib/devise/data_extractor.rb
```diff
@@ -0,0 +1,79 @@
...
+            if object.respond_to?(:[])
+              id = object[PRIORITY_ORDERED_ID_KEYS.find { |key| object[key] }]
+              scope = find_devise_scope(object)
+
+              id = "#{scope}:#{id}" if id && scope
+              return transform(id)
+            end
+
+            id = object.id if object.respond_to?(:id)
+            id ||= object.uuid if object.respond_to?(:uuid)
```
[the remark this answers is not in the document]
Strech: Thanks for pointing out. But after the like 31 `id` will be initialized as `nil`, here is the check

```
❯❯❯ irb
irb(main):001> id = 1 if false
=> nil
irb(main):002> local_variables
=> [:id, :_]
```

lib/datadog/appsec/contrib/devise/data_extractor.rb
```diff
@@ -0,0 +1,79 @@
...
+
+          def extract_login(object)
+            return if object.nil?
+
+            if object.respond_to?(:[])
+              login = object[PRIORITY_ORDERED_LOGIN_KEYS.find { |key| object[key] }]
+              return transform(login)
+            end
+
+            login = object.email if object.respond_to?(:email)
```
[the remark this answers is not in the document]
Strech: Thanks for pointing out, but it will be initialized as local variable with `nil` value after that line (see comment above ⬆️)

PR #4435, opened 2025-02-25
Fix AppSec ActiveRecord instrumentation

y9v:
**What does this PR do?**
This PR fixes AppSec auto-instrumentation.

**Motivation:**
We were not using a `Railtie` for auto-instrumenting Rails applications. This caused problems with `ActiveRecord` adapters patching - they were not loaded at the time of patching `ActiveRecord`.

**Change log entry**
Yes. AppSec: Fix auto-instrumentation of ActiveRecord for SQL Injection detection.

**Additional Notes:**
Load hooks for different adapters were added in Rails 7.0, except for MySQL adapter, for MySQL adapter it was added in Rails 7.1.

**How to test the change?**


lib/datadog/appsec/contrib/active_record/instrumentation.rb
```diff
@@ -67,6 +67,11 @@ module ExecQueryAdapterPatch
             def exec_query(sql, *args, **rest)
               Instrumentation.detect_sql_injection(sql, adapter_name)
 
+              # Rails 5 instruduces `prepare:` keyword argument in exec_query method.
+              # Ruby 2.5 expands empty keyword arguments to empty hash,
+              # which raises ArgumentError due to incorrect number of arguments.
+              return super(sql, *args) if Gem.loaded_specs['activerecord'].version < Gem::Version.new('5.0')
```
Strech, line 73: I think if we have exec_query/3 should not we pass all 3 arguments to the super? Or shouldn't the patch be different on its own than?

lib/datadog/appsec/contrib/active_record/patcher.rb
```diff
@@ -19,27 +19,43 @@ def target_version
           end
 
           def patch
-            ActiveSupport.on_load :active_record do
+            patch_proc = lambda do |adapter_class, is_postgres: false|
```
Strech, line 22: I think it was the very first attempt to make it that way with postgres being an argument of a function. I think here is an issue which is band-aided. I think it's better to rethink and simplify patching instead of quick-fixing it, WDYT?

lib/datadog/appsec/contrib/rails/auto_instrument_railtie.rb
```diff
@@ -0,0 +1,18 @@
...
+
+module Datadog
+  module AppSec
+    module Contrib
+      module Rails
+        # Railtie to include AutoInstrumentation in rails loading
+        class AutoInstrumentRailtie < ::Rails::Railtie
+          # we want to load before config initializers so that any user supplied config
+          # in config/initializers/datadog.rb will take precedence
+          initializer 'datadog.start_appsec', before: :load_config_initializers do
```
Strech, line 11: Should we go with a narrowed scope and specific to action name instead? WDYT?

```suggestion
          initializer 'datadog.appsec.auto_instrument', before: :load_config_initializers do
```

lib/datadog/appsec/contrib/rails/auto_instrument_railtie.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+module Datadog
+  module AppSec
+    module Contrib
+      module Rails
+        # Railtie to include AutoInstrumentation in rails loading
+        class AutoInstrumentRailtie < ::Rails::Railtie
```
Strech, line 8: Do we need a require of railtie here to guarantee its presence?

Rakefile
```diff
@@ -324,6 +326,14 @@ namespace :spec do
         t.rspec_opts = args.to_a.join(' ')
       end
     end
+
+    desc '' # "Explicitly hiding from `rake -T`"
+    RSpec::Core::RakeTask.new(:integration_autoinstrument) do |t, args|
+      t.pattern = 'spec/datadog/appsec/contrib/integration_autoinstrument/**/*_spec.rb'
```
Strech, line 332: It feels like instead it should be 2 adjustments and nested folders, because it looks like band-aid

My suggestion instead is:

```suggestion
    RSpec::Core::RakeTask.new(:integration_auto_instrument) do |t, args|
      t.pattern = 'spec/datadog/appsec/contrib/integration/auto_instrument/**/*_spec.rb'
```

and an exclude to the integration like

```ruby
    RSpec::Core::RakeTask.new(:integration) do |t, args|
      t.pattern = 'spec/datadog/appsec/contrib/integration/**/*_spec.rb'
      t.exclude_pattern = 'spec/datadog/appsec/contrib/integration/auto_instrument/**/*_spec.rb,'

```

_P.S in the rest of the code we use underscore to name `auto_instrumentation`, I think it makes sense to follow._

Strech (commented):
It's cool to have auto instrumentation being fixed 👏🏼 

I think the solution needs a bit of a ❤️

lib/datadog/appsec/contrib/rails/auto_instrument_railtie.rb
y9v: it's a bit finicky - we check if railtie is loaded before requiring this file:
https://github.com/DataDog/dd-trace-rb/pull/4435/files#diff-6932a928693e0818dc0bb6586852b0453f1020945ae10f7ac655161f24c1a979R11-R12

y9v:
will split this PR in 2 - one for fixing ActiveRecord instrumentation, and one for fixing auto-instrumentation

PR #4436, opened 2025-02-25
Improve pulling images

TonyCTHsu:
**Motivation:**

Currently, the  “Initialize containers”  step is quite heavy with pulling many images, several workflow failure due to this step taking too long pulling images or waiting on the services to be ready which eventually leads to timeout.

For example: https://github.com/DataDog/dd-trace-rb/actions/runs/13510559975

The “Initialize containers” step is automatically generated by GitHub Actions, and we have no configurations or methods to reduce the time it takes.

**What does this PR do?**

1. Extract a job with the following services and run specific tasks
  - elasticsearch
  - memcached
  - mongodb
  - opensearch
  - presto

2. Since the service configuration is at job level, it is static and cannot be dynamically configured. Extracting the shared steps into composite actions.

3. Since the workflow install dependencies from network, the installation might still fail even bundler is configured with retry. Improves the reliability for downloading with manual retry.

#### Result

- Workflow time change from 15 min -> 12 min
- The `Initialize containers` step for the standard runners change from 1:40 min (around 1:30 ~ 2:00) ->  around 40s (within 1 min), while the misc runners took around 1 min

**Change log entry**

None.


.github/workflows/_unit_test.yml
```diff
@@ -60,18 +61,47 @@ jobs:
...
+  #   - mongodb
+  #   - opensearch
+  #   - presto
+  #
+  #  Benefit of this optimization:
+  #  - Unnecessary services are not started
+  #  - Reduce the overhead for pulling images
+  #  - Improve reliability for starting dependent services
+  #
+  # In the future, we could consider to extract jobs to run database dependent tasks to optimize the workflow.
```
lloeki, line 99: I was going to suggest this is a good first step but I'd like `build-test-misc` to be split into five so that each one only pulls their relevant service.

Seems like improvements are on the table as next steps, so let's roll with this PR.

tasks/github.rake
```diff
@@ -6,17 +6,19 @@ namespace :github do
   task :generate_batches do
     matrix = eval(File.read('Matrixfile')).freeze # rubocop:disable Security/Eval
 
-    # Uncomment to only runs the specified candidates
-    candidates = [
-      # 'sidekiq', # Connection refused - connect(2) for 127.0.0.1:6379 (RedisClient::CannotConnectError)
-      # 'mongodb', # Connection issue with multi db and process hang
+    # TODO: These are the execptions, find a way to describe those service dependencies in CI using a more generic mechansim.
```
lloeki, line 9: A per-group `services: [service1, service2]` key within the `Matrixfile` could be a nice spot to define such dependencies.

This could even generalise to "standard" ones and help get rid of the arbitrary split.

lloeki (approved):

tasks/github.rake
TonyCTHsu: Yup, that requires quite some changes to adopt a new data structure. So I am delaying this pain.

PR #4437, opened 2025-02-26
Fix AppSec ActiveRecord instrumentation for Rails 4 and ruby < 2.7

y9v:
**What does this PR do?**
This PR fixes AppSec ActiveRecord instrumentation for Rails version 4.

**Motivation:**
AppSec ActiveRecord instrumentation raised `ArgumentError` with Rails 4 and Ruby < 2.7.

The reason for this was the fact that Ruby versions older than 2.7 passed empty hash as a positional argument for empty hash splat.
https://rubyreferences.github.io/rubychanges/2.7.html#keyword-argument-related-changes

**Change log entry**
Yes. AppSec: Fix ActiveRecord instrumentation for Rails 4 and ruby < 2.7

**Additional Notes:**
None

**How to test the change?**
CI and manual testing.


Strech (approved):
👏🏼 Well done

PR #4440, opened 2025-02-26
DEBUG-3568 remove agent settings defaulting hack

p-datadog:
**What does this PR do?**

Removes the agent settings defaulting hack from the codebase. Agent settings should be passed down from component construction into transports in order to respect user configuration.

**Motivation:**
Continued DRY refactoring of transport code to support dynamic instrumentation

**Change log entry**
None

**Additional Notes:**
I *think* this PR does not affect customers but not entirely sure.

I would like to remove `transport_options` in a subsequent PR.

**How to test the change?**
Existing unit tests


marcotc (approved):
No way! This is an awesome clean up that always looked so ugly! 🙏

y9v (approved):

PR #4448, opened 2025-03-01
Update rubocop for ruby 2.7+

p-datadog:
**What does this PR do?**
Updates rubocop and related dependencies to newer versions.

Removes rubocop from ruby 2.6 configuration since the current versions of rubocop do not support ruby 2.6.
Removes rubocop from jruby 9.3 configuration because jruby 9.3 identifies as ruby 2.6 and hence rubocop does not support it anymore.
Removes standard from rubies that do not have rubocop, because then it would bring an old rubocop that wouldn't behave the same as the one in supported versions.
Updates rubocop to 1.71. Current version of rubocop is 1.73 but standard does not support 1.73 yet.
standard version is not explicitly specified but it gets updated to 1.45 from somewhere in the 1.20 range.
Several files edited for changes demanded by recent rubocops.
rubocop-rspec updated to most recent 2.x release, 3.0 has many new demands producing a huge diff likely to conflict with other PRs.

**Motivation:**

Noticed old version when testing https://github.com/rubocop/rubocop/issues/13933

**Change log entry**
None

**Additional Notes:**

Since rubocop, standard and related libraries need to have synchronized versions, this PR does not update rubocop to its latest version (currently 1.73, PR updates to 1.71) and e.g. rubocop-performance needs to be declared as 1.20 not 1.23.

rubocop-rspec 3.x wants to make MANY changes to our test suite, they will probably conflict with in-progress PRs.

**How to test the change?**
Existing CI


TonyCTHsu (approved):

lib/datadog/tracing/contrib/configuration/settings.rb
```diff
@@ -22,7 +22,7 @@ class Settings
           end
 
           def configure(options = {})
-            self.class.options.each do |name, _value|
```
marcotc, line 25: Love these improvements!

marcotc (approved):

y9v (approved):

PR #4459, opened 2025-03-05
[APPSEC-56614] Add Instrumentation::Gateway#pushed? method

Strech:
**What does this PR do?**

Adds a method to check wether event with specific name was already ever pushed or not

**Motivation:**

I need a way to react on one event only if another event was pushed before. To make it very straightforward I decided to allow a simple check - was an event published or not.

**Change log entry**

No. Internal change

**Additional Notes:**

The whole `Instrumentation::Gateway` is a subject to rewrite, but the core logic of "I want to know about what kind of events were pushed" remains.

**How to test the change?**

CI is good enough.


lib/datadog/appsec/instrumentation/gateway.rb
```diff
@@ -1,35 +1,23 @@
...
-        private_constant :Middleware
-
         def initialize
           @middlewares = Hash.new { |h, k| h[k] = [] }
+          @pushed_events = {}
         end
 
         def push(name, env, &block)
-          block ||= -> {}
+          @pushed_events[name] = true
```
ivoanjo, line 17: Very stupid question: Is this a possible memory leak concern? E.g. are the set of names fixed, or is it possible for them to accumulate over time?

Strech: Not stupid questions. The list of events are controlled by us and pushed by us. The "names" we push are no subject to generation (they are static) and they deliberately triggered by the instrumentation code.

Considering our control over that - it should be safe. We can restrict it by the list of known or the count if you think it makes sense to do.

_P.S Currently we are pushing through gateway_

```
identity.set_user
rails.request.action
rack.request
rack.response
rack.request.body
graphql.multiplex
sinatra.request.dispatch
sinatra.request.routed
```

y9v (approved):

p-datadog (approved):

lib/datadog/appsec/instrumentation/gateway.rb
Strech: Going to add the note and warning, thanks for advise 🤝

PR #4460, opened 2025-03-05
[PROF-11405] Graduate heap profiling from alpha to preview, second try

ivoanjo:
**What does this PR do?**

A few days ago, we merged #4331 and #4401 to master.

Starting with #4331, we used a new libdatadog feature for implementing the Ruby heap profiler. While validating these changes, we discovered this libdatadog feature had a bug (https://github.com/DataDog/libdatadog/pull/896) and thus reverted all PRs related to this heap profiler change until we could get the fix in https://github.com/DataDog/dd-trace-rb/pull/4409.

Because the bug was only in the new libdatadog feature, it was still OK to use the old approach with the buggy libdatadog (since it used an alternative, non-buggy, code path).

Now that we have the fix on the libdatadog side almost ready, I'm opening this PR to re-apply the changes from #4331 and #4401 to master.

This PR is thus almost the reverse of the revert PR in https://github.com/DataDog/dd-trace-rb/pull/4409. The only difference is that it does not yet include the changes from https://github.com/DataDog/dd-trace-rb/pull/4376 for the same reason that PR was needed originally. 

**Motivation:**

Release heap profiling feature.

**Change log entry**
Yes. Please use the one in https://github.com/DataDog/dd-trace-rb/pull/4401 .

**Additional Notes:**

There's quite a diff on this PR, but there's no differences to those earlier PRs other than adding a regression test for this issue (the regression test is a separate commit that can be reviewed by itself).

In fact, you may notice that CI is red for this PR because I'm opening it before the libdatadog fix is released. My intention is to exactly come back once the new libdatadog is released, and see this PR turn green with a simple merge from master.

**How to test the change?**

The earlier PRs already included tests; and I added a regression test for the libdatadog issue in this PR.


Strech (approved):
LGTM 👍🏼

ivoanjo:
Wohoo! So a few... days... more... than I expected, but this PR is finally green and so I'm going to press the big green button.

PR #4461, opened 2025-03-05
[PROF-11405] Revert "Temporary disable benchmark to make CI happy", second try

ivoanjo:
**What does this PR do?**

This reverts commit 569bf568b80e9615027a8d0c41a8a941deeb9981.

We had previously reverted that commit in
https://github.com/DataDog/dd-trace-rb/pull/4376 but due to a bug had to undo all the heap profiling-related changes in https://github.com/DataDog/dd-trace-rb/pull/4409 .

After https://github.com/DataDog/dd-trace-rb/pull/4460 gets merged, we can finally enable this benchmark again.

**Motivation:**

The benchmark was disabled for the reason mentioned in the comment.

**Change log entry**
None.

**Additional Notes:**

I'm going to open this PR stacked on top of
https://github.com/DataDog/dd-trace-rb/pull/4460, but this PR should only be merged after that other PR is merged to master.

**How to test the change?**

Validate this benchmark is now running in the benchmarking platform results page.

### DIFF
--- a/benchmarks/profiling_string_storage_intern.rb
+++ b/benchmarks/profiling_string_storage_intern.rb
@@ -7,11 +7,6 @@ require_relative 'benchmarks_helper'
 
 # This benchmark measures the performance of string storage-related APIs
 
-# Temporary hack to make CI happy. Our CI tries to compare benchmarks between master and branches, and if benchmarks
-# are not backwards-compatible between them (e.g. a new API was added...) it breaks. As a workaround, I'm disabling this
-# benchmark so we can merge it to master, and then I'll follow up with a micro-PR to re-enable it.
-TEMPORARY_DISABLE_BENCHMARK = true
-
 class ProfilerStringStorageIntern
   def initialize
     @recorder = Datadog::Profiling::StackRecorder.for_testing(heap_samples_enabled: true)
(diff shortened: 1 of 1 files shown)


ivoanjo:
Since the diff on this PR is super small, I've rebased it on top of current master instead of trying to do something fancier.

ivoanjo:
> A bit confused, but 👍🏼

@stretch I missed this comment before! Let me explain -- the way the benchmarks get run (or at least last I looked) is by doing the following:

* Check out this PR, run the benchmarks, get the numbers
* Check out master, **copy over benchmarks from this branch**, get the numbers

This enables some neat things...

* You can actually test master vs this branch if you're adding a new benchmark and optimizing some feature
* You're doing an apples-to-apples comparison if you change the benchmark on this branch
* etc

... **but suffers from a big assumption that's not true every time: what if the benchmark can't be made to be API-compatible with master? For instance, if you're benchmarking a new api that doesn't exist in main?**

So that's why I added the benchmark to master in https://github.com/DataDog/dd-trace-rb/pull/4460, but in disabled form. Now that the rest of the code is there, we can enable it, and the benchmark will run and not break CI.

PR #4464, opened 2025-03-06
[PROF-11475] Add extra safety checks when getting line number in profiler

ivoanjo:
**What does this PR do?**

This PR adds a few extra checks when getting the line number during stack trace gathering in the profiler.

We got one automated report of a crash in the `rb_iseq_line_no` function, and we couldn't reproduce it, so this is the first attempt at avoiding the issue -- let's see if we ever observe this crash again or not.

**Motivation:**

Improve stability of the profiler.

**Change log entry**
None.

**Additional Notes:**

See the internal PROF-11475 ticket for more details.

**How to test the change?**

Our existing test coverage already validates that stack trace gathering works fine and matches the Ruby APIs. If this condition was not correct, we'd get more lines set to 0, and our unit tests would fail. (I did some quick manual testing to check this was the case -- and indeed it is).

### DIFF
--- a/ext/datadog_profiling_native_extension/private_vm_api_access.c
+++ b/ext/datadog_profiling_native_extension/private_vm_api_access.c
@@ -312,6 +312,7 @@ VALUE thread_name_for(VALUE thread) {
 // to support our custom rb_profile_frames (see below)
 // Modifications:
 // * Support int first_lineno for Ruby 3.2.0+ (https://github.com/ruby/ruby/pull/6430)
+// * Validate iseq and pos before calling `rb_iseq_line_no` as a safety measure (see comment below for details)
 //
 // `node_id` gets used depending on Ruby VM compilation settings (USE_ISEQ_NODE_ID being defined).
 // To avoid getting false "unused argument" warnings in setups where it's not used, we need to do this weird dance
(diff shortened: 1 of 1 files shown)


ext/datadog_profiling_native_extension/private_vm_api_access.c
```diff
@@ -358,6 +359,13 @@ calc_pos(const rb_iseq_t *iseq, const VALUE *pc, int *lineno, int *node_id)
             __builtin_trap();
         }
 #endif
+
+        // In PROF-11475 we spotted a crash when calling `rb_iseq_line_no` from this method. We couldn't reproduce or
+        // figure out the root cause, but "just in case", we're validating that the iseq looks valid and that the
+        // `n` used for the position is also sane, and if they don't look good, we don't calculate the line, rather
+        // than potentially trigger any issues.
+        if (RB_UNLIKELY(!RB_TYPE_P((VALUE) iseq, T_IMEMO) || n < 0 || n > ISEQ_BODY(iseq)->iseq_size)) return 0;
```
p-datadog, line 367: I feel like this permits `n` to be one larger than `iseq_size` which does not seem right? Maybe `n >= iseq_size` would be more correct?

ivoanjo: Nicely spotted. I believe this is ok because of what's explained in line 352:

> /* use pos-1 because PC points next instruction at the beginning of instruction */

That is, it seems valid for `n` to go one beyond the number of instructions in the iseq because of this. The VM assert from line 347:

> VM_ASSERT(n <= ISEQ_BODY(iseq)->iseq_size);

also takes this into account.

p-datadog: :+1:

ivoanjo, line 367: Note: https://github.com/DataDog/dd-trace-rb/pull/4790 provides extra info on this crash

PR #4468, opened 2025-03-06
[APPSEC-56960] Add collection of request headers for identity

Strech:
**What does this PR do?**
Two things:

1. Add new event for Rack contrib named `rack.request.finish`
2. Add subscriber to the said event to set tags of [identity related headers](https://datadoghq.atlassian.net/wiki/spaces/SAAL/pages/2186870984/HTTP+header+collection) (see User event)

**Motivation:**

This will be a part of the #4433 and could be extracted without a fear of breaking anything.

**Change log entry**

No. It will be announced as of #4433 release

**Additional Notes:**

Unfortunately, we don't have time to go over the current published/subscriber model and it has very odd limitations. Hence the code is placed in the most convenient way, but not in the most right.

**How to test the change?**

CI is good enough.

### DIFF
--- a/lib/datadog/appsec/contrib/rack/gateway/watcher.rb
+++ b/lib/datadog/appsec/contrib/rack/gateway/watcher.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_relative '../ext'
 require_relative '../../../instrumentation/gateway'
 require_relative '../../../event'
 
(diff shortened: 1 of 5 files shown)


lib/datadog/appsec/contrib/rack/gateway/watcher.rb
```diff
@@ -110,6 +111,25 @@ def watch_request_body(gateway = Instrumentation.gateway)
...
+                  context = gateway_request.env[AppSec::Ext::CONTEXT_KEY]
+                  next stack.call(gateway_request.request) if context.span.nil? || !gateway.pushed?('identity.set_user')
+
+                  gateway_request.headers.each do |name, value|
+                    next unless Ext::IDENTITY_COLLECTABLE_REQUEST_HEADERS.include?(name)
+
+                    context.span["http.request.headers.#{name}"] = value
+                  end
+
+                  stack.call(gateway_request.request)
```
y9v, line 130: Do we want to call the next item in the stack with `gateway_request.request`, or just `gateway_request`? I would assume that if we have 2 watchers for the same event, they would get the same argument?

y9v (approved):

lib/datadog/appsec/contrib/rack/gateway/watcher.rb
Strech: It's the same as in [the method above](https://github.com/DataDog/dd-trace-rb/pull/4468/files#diff-77759fd9cacb661757fbe67537b41c78174137c8cda751f73d0be8746602ddb6R111). I know that it looks subpar, but it's about to change soon (I hope)

PR #4479, opened 2025-03-11
Fix APM Disablement V1 (ASM Standalone)

vpellan:
**What does this PR do?**

This PR fixes multiple bugs on APM Disablement V1 (aka ASM Standalone) when it was activated:
- When HTTP frameworks (Net::HTTP, RestClient, Excon, Faraday...) were instrumented, the sampling_priority was set to MANUAL_REJECT if tags contained _dd.p.appsec (which was used to check for upstream propagated appsec events), then not sampled later due to a change in the sampler in [this PR](https://github.com/DataDog/dd-trace-rb/pull/3956/files#diff-fb264450c82590f2bf44f09afc63b6a01bd80e4916333f7c19441240694412a1R496) This was fixed by making this check during the building step of a trace from a digest (which also means that it now works even if an HTTP framework is not instrumented)
- Distributed tracing was skipped only for Net::HTTP (in case of upstream or local security event). It now skips it on every HTTP frameworks supported, but also Sidekiq and GRPC.
- The APM Tracing disabled tag (_dd.apm.enabled=0) was not added when appsec was disabled. It is now fixed and the code related to it was moved to Tracing
- Added tests that actually tests APM disablement heartbeat traces (ASM Standalone too) and cleaned related tests

**Motivation:**

When I tried to activate SCA Standalone system-tests, I discovered that disabling AppSec and disabling APM Tracing would result in heartbeat traces not sending dd.apm.enabled=0 tag. During my investigation, I discovered more bugs (listed above).
This is also a step towards APM Disablement V2. For now, the variables are still called after appsec standalone, but this will change to APM disablement in the future and lose all reference to asm standalone, as it is in fact an APM feature.

**Change log entry**

None.

**Additional Notes:**

This PR should be followed soon by APM Disablement V2, which will replace the env var by DD_APM_TRACING_ENABLED, and every config/variable referencing asm standalone to apm disablement.

I understand that there might be some concerns about AppSec stuff being in APM territory but this is mostly a matter of naming that will be fixed when APM Disablement V2 will be released. In the V2, I will replace any instance of `Datadog.configuration.appsec.standalone.enabled` by `Datadog.configuration.tracing.apm.disabled`, `Datadog::AppSec::Ext::TAG_DISTRIBUTED_APPSEC_EVENT`  by something like `Datadog::Tracing::Ext::TAG_DISTRIBUTED_TRACE_SOURCE` (`_dd.p.ts`, which will be a bitmask). The only exception might be in the `propagate_sampling_priority?` method, in tracer.rb, where we check if AppSec is activated.

**How to test the change?**

CI and system-tests


lib/datadog/tracing/contrib/ethon/easy_patch.rb
```diff
@@ -219,6 +215,17 @@ def parse_response_headers
                 header.size != 2 ? nil : header
               end.compact.to_h
             end
+
+            # Skips distributed tracing if disabled for this instrumentation
+            # or if APM is disabled unless there is an AppSec event (from upstream distributed trace or local)
+            def should_skip_distributed_tracing?(trace)
```
TonyCTHsu, line 221: I feel like I am seeing this piece of logic repeatedly several times, perhaps it make sense to encapsulate this from an AppSec module and be reused at multiple places?

The method can be stateless or depends on configuration and the contract is basically returning a boolean.

TonyCTHsu (commented):
FYI: Currently, there is a karafka integration under review which contains distributed tracing mechanism

vpellan:
> FYI: Currently, there is a karafka integration under review which contains distributed tracing mechanism

What is the process for this kind of situation ? If I merge to master first, should I myself add a commit on the Karafka integration to add APM Disablement capabilities ? (Of course if the other way around happens I could just rebase and add it here)

lib/datadog/tracing/contrib/ethon/easy_patch.rb
vpellan: Thanks, I will investigate that, although I don't think it should be in an AppSec module, as in the V2, this will not contain any reference to AppSec (`appsec.standalone` will be replaced by `tracing.apm` and `APPSEC_EVENT` will be replaced by `TRACE_SOURCE`)

TonyCTHsu:
>What is the process for this kind of situation ?

I would suggest not to wait on it. However, it would be much welcomed to encapsulate the logic so that the pattern can be adopted easily.

lib/datadog/tracing/contrib/ethon/easy_patch.rb
vpellan: Fixed in 91f2f68de528681a94a02e45a89024f4fb8c4423

lib/datadog/tracing/tracer.rb
```diff
@@ -545,10 +550,30 @@ def skip_trace(name)
...
+      # Decide whether upstream sampling priority should be propagated, by taking into account
+      # the upstream tags and the configuration.
+      # We should always propagate if there APM is not disabled.
+      #
+      # e.g.: upstream tags containing dd.p.appsec, and appsec is enabled, return true.
+      # DEV: dd.p.appsec will be replaced by dd.p.ts in APM disablement 2.0, which will be a bitmask.
+      def propagate_sampling_priority?(upstream_tags:)
+        return true unless apm_tracing_disabled
+
+        if upstream_tags&.key?(Datadog::AppSec::Ext::TAG_DISTRIBUTED_APPSEC_EVENT)
```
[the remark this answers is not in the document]
vpellan: The type is defined in the corresponding .rbs file. Maybe this would be a great topic to bring for a guild meeting or tech sync, about how we should declare types and create some rules for the future.

lib/datadog/tracing/distributed/circuit_breaker.rb
```diff
@@ -0,0 +1,26 @@
...
+  module Tracing
+    module Distributed
+      # Helper method to decide when to skip distributed tracing
+      module CircuitBreaker
+        module_function
+
+        # Skips distributed tracing if disabled for this instrumentation
+        # or if APM is disabled unless there is an AppSec event (from upstream distributed trace or local)
+        def should_skip_distributed_tracing?(client_config: nil, datadog_config: nil, trace: nil)
+          if ::Datadog.configuration.appsec.standalone.enabled &&
```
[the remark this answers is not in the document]
vpellan: Arguments renamed in 6dd3b9518f939269afa4900c09ba16542bf465db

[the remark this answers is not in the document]
p-datadog: :1:

lib/datadog/tracing/component.rb
```diff
@@ -79,13 +79,7 @@ def build_sampler(settings)
...
         # but for features such as API Security, we need to send at least one trace per minute,
         # to keep the service alive on the backend side.
-        if settings.appsec.standalone.enabled
-          post_sampler = Tracing::Sampling::RuleSampler.new(
-            [Tracing::Sampling::SimpleRule.new(sample_rate: 1.0)],
-            rate_limiter: Datadog::Core::TokenBucket.new(1.0 / 60, 1.0),
-            default_sample_rate: 1.0 / 60
-          )
-        end
+        post_sampler = build_rate_limit_post_sampler(seconds: 60) if settings.appsec.standalone.enabled
```
y9v, line 82: I would actually leave this line as it is - it is not related to the bugfix, and it does not improve things much (in my opinion).
IMO it doesn't make sense to extract code to a function, just to make the original function smaller - in the end the code is still there, but you have to jump to find it, and this function is not really used anywhere except the original function it was extracted from.

lib/datadog/tracing/distributed/circuit_breaker.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Tracing
+    module Distributed
+      # Helper method to decide when to skip distributed tracing
+      module CircuitBreaker
```
Strech, line 7: I think this name will lead to confusion because there is specific meaning behind the "Circuit Breaker" in programming.

A few better names could be:

1. TraceSkipPolicy
2. TraceSkipStrategy
3. TraceSkipLogic

or something similar. Why is it better?

1. Instead of reading the comment above the class name, the class name itself reflects the purpose
2. The behavior patters are Stragery, Policy, Behavior, etc.

lib/datadog/tracing/distributed/circuit_breaker.rb
```diff
@@ -0,0 +1,28 @@
...
+        module_function
+
+        # Skips distributed tracing if disabled for this instrumentation
+        # or if APM is disabled unless there is an AppSec event (from upstream distributed trace or local)
+        def should_skip_distributed_tracing?(contrib_client_config: nil, contrib_datadog_config: nil, trace: nil)
```
Strech, line 12: With good class name in mind this method name could be improved and shortened to just `skip?`

lib/datadog/tracing/tracer.rb
```diff
@@ -545,10 +550,30 @@ def skip_trace(name)
...
+
+        false
+      end
+
       def profiling_enabled
         @profiling_enabled ||=
           !!(defined?(Datadog::Profiling) && Datadog::Profiling.respond_to?(:enabled?) && Datadog::Profiling.enabled?)
       end
+
+      def apm_tracing_disabled
```
Strech, line 574: Just a question, we have everywhere `_enabled` postfix, is there a strong reason to go `_disabled`? Just asking

spec/datadog/appsec/contrib/support/integration/shared_examples.rb
```diff
@@ -1,3 +1,5 @@
+require 'datadog/tracing/contrib/support/integration/shared_examples'
```
Strech, line 1: Maybe in addition we can add `frozen_string_literal`?

spec/datadog/tracing/distributed/circuit_breaker_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  subject(:circuit_breaker) { circuit_breaker_class.new }
+
+  let(:circuit_breaker_class) { Class.new { include Datadog::Tracing::Distributed::CircuitBreaker } }
+
+  describe '#should_skip_distributed_tracing?' do
+    subject(:should_skip_distributed_tracing?) do
+      circuit_breaker.send(
+        :should_skip_distributed_tracing?,
+        **{ contrib_client_config: contrib_client_config, contrib_datadog_config: contrib_datadog_config, trace: trace }
+      )
```
Strech, line 14: TBH programming in tests makes tests more fragile. I would recommend avoid it and instead use in place calls. It will allow you to avoid using `send` in addition

spec/datadog/tracing/distributed/circuit_breaker_spec.rb
```diff
@@ -0,0 +1,75 @@
+require 'datadog'
+require 'datadog/tracing/distributed/circuit_breaker'
+
+RSpec.describe Datadog::Tracing::Distributed::CircuitBreaker do
+  subject(:circuit_breaker) { circuit_breaker_class.new }
```
Strech, line 5: I think it makes sense to define subject in place instead and don't use constructs `is_expected` because with that calls in the bottom of the tests we lost understanding of what are we testing now.

Strech (changes requested):
Well done 👏🏼 

It would be amazing if we can give a better name for the current Circuit Breaker and clean tests with `frozen_string_literal` and clean test cases.

lib/datadog/tracing/component.rb
vpellan: I extracted this to be able to mock it in my tests, to replace the 60 seconds sampler with a 4 seconds one to speed up the tests. The alternative would be to mock the entire build_sampler method, do you think it's a better idea ?

lib/datadog/tracing/tracer.rb
vpellan: APM is enabled by default, and when it's enabled, we are not adding the `dd.apm.enabled=1` tag to the trace (as it is implied), but when it is disabled, we are adding the `dd.apm.enabled=0` tag. The reason why I named it apm_tracing_disabled is that it has an effect only when it is set to true (so when APM is disabled). Do you think it's better to replace it with `_enabled` ? Now I'm not sure anymore that `_disabled` is better

vpellan: Although I still think it makes sense to call that method `tag_apm_tracing_disabled!` as we're only adding that tag if apm is disabled (https://github.com/DataDog/dd-trace-rb/pull/4479/files#diff-4cc39e550d70903a0aded38fca129e91b21d344d78f83ea06eae6d04d7a5d436R200)

y9v (changes requested):

spec/datadog/tracing/distributed/circuit_breaker_spec.rb
y9v: you also don't need to use the double-splat here, you can just pass the arguments as they are

lib/datadog/tracing/distributed/circuit_breaker.rb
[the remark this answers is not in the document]
y9v: regarding argument naming: how about naming them `pin_config` and `global_config`? They are both datadog configs for contribs, so I would omit that

lib/datadog/tracing/contrib/excon/middleware.rb
```diff
@@ -30,10 +30,12 @@ def request_call(datum)
                 trace = Tracing.active_trace
                 datum[:datadog_span] = span
                 annotate!(span, datum)
-                if Datadog::AppSec::Utils::TraceOperation.appsec_standalone_reject?(trace)
-                  trace.sampling_priority = Tracing::Sampling::Ext::Priority::AUTO_REJECT
+                if Tracing.enabled? && !Tracing::Distributed::CircuitBreaker.should_skip_distributed_tracing?(
```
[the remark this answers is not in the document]
vpellan: I added it based on the existing logic. Previously, that line was `propagate!(trace, span, datum) if distributed_tracing?` and `distributed_tracing?` returns true only if Tracing is enabled. We can remove it but the check will be done way later anyway (https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/distributed/datadog.rb#L41), we may want to skip that extra logic if Tracing is disabled

lib/datadog/tracing/contrib/faraday/middleware.rb
```diff
@@ -29,10 +29,12 @@ def call(env)
 
             Tracing.trace(Ext::SPAN_REQUEST, on_error: request_options[:on_error]) do |span, trace|
               annotate!(span, env, request_options)
-              if Datadog::AppSec::Utils::TraceOperation.appsec_standalone_reject?(trace)
-                trace.sampling_priority = Tracing::Sampling::Ext::Priority::AUTO_REJECT
+              if Tracing.enabled? && !Tracing::Distributed::CircuitBreaker.should_skip_distributed_tracing?(
```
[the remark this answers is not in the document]
vpellan: Tracing.trace will still work even if Tracing is disabled and will generate a dummy trace (and span) (https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/tracer.rb#L538). In this case, it might even actually send distributed tags as trace will not be nil.

lib/datadog/tracing/distributed/circuit_breaker.rb
vpellan: Changed in f5c5f4ef1b0cabadded0a4ead8132ecd705f4e75 !

lib/datadog/tracing/distributed/circuit_breaker.rb
vpellan: Changed in f5c5f4ef1b0cabadded0a4ead8132ecd705f4e75 !

lib/datadog/tracing/tracer.rb
vpellan: Changed in 872abbcc47dcc5a879cb6f0373e3e0f2295b21e7 !

spec/datadog/appsec/contrib/support/integration/shared_examples.rb
vpellan: Fixed in 9e4eda6d94d0dd53a9fee000c82247ae31c72e58 !

spec/datadog/tracing/distributed/circuit_breaker_spec.rb
vpellan: Fixed in f151293fec70086415fdbbb65fce292fbafcfc07 !

spec/datadog/tracing/distributed/circuit_breaker_spec.rb
vpellan: Fixed in f151293fec70086415fdbbb65fce292fbafcfc07 !

lib/datadog/tracing/distributed/circuit_breaker.rb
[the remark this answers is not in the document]
vpellan: Fixed in f0b85e8547971633397c703155d2cbf57936d8d4 !

lib/datadog/tracing/component.rb
marcotc: I agree with @y9v here, that it harms the readability a this method.
At the same time, I agree with @vpellan that we should be able to test it easily.
🤷 
I'm ok keeping this change in the PR for the sake of correctness, and the fact that the method is right down here, and not in some far away file.

lib/datadog/tracing/component.rb
```diff
@@ -197,6 +191,13 @@ def build_tracer_tags(settings)
         end
       end
 
+      def build_rate_limit_post_sampler(seconds:)
```
marcotc, line 194: Can you add some very basic documentation to what the `seconds` argument means here?

lib/datadog/tracing/contrib/ethon/easy_patch.rb
```diff
@@ -110,11 +110,10 @@ def datadog_before_request(continue_from: nil)
 
               datadog_tag_request
 
-              if Datadog::AppSec::Utils::TraceOperation.appsec_standalone_reject?(datadog_trace)
-                datadog_trace.sampling_priority = Tracing::Sampling::Ext::Priority::AUTO_REJECT
-              end
-
-              if datadog_configuration[:distributed_tracing]
+              unless Tracing::Distributed::SkipPolicy.skip?(
```
marcotc, line 113: Can I persuade you to rename the `SkipPolicy.skip?` since all its uses in the PR are double negatives (e.g. `unless ... skip?` or `!Tracing::Distributed::SkipPolicy.skip?` in `lib/datadog/tracing/contrib/excon/middleware.rb`)? 

I recommend `Tracing::Distributed::Policy.enabled?`.

lib/datadog/tracing/contrib/http/circuit_breaker.rb
```diff
@@ -27,21 +27,6 @@ def internal_request?(request)
             !!(request[Datadog::Core::Transport::Ext::HTTP::HEADER_META_TRACER_VERSION] ||
               request[Datadog::Core::Transport::Ext::HTTP::HEADER_DD_INTERNAL_UNTRACED_REQUEST])
           end
-
-          def should_skip_distributed_tracing?(client_config)
```
marcotc, line 31: Great cleanup!

lib/datadog/tracing/contrib/httprb/instrumentation.rb
```diff
@@ -30,12 +30,14 @@ def perform(req, options)
                   span.service = service_name(host, request_options, client_config)
                   span.type = Tracing::Metadata::Ext::HTTP::TYPE_OUTBOUND
 
-                  if Datadog::AppSec::Utils::TraceOperation.appsec_standalone_reject?(trace)
-                    trace.sampling_priority = Tracing::Sampling::Ext::Priority::AUTO_REJECT
+                  if Tracing.enabled? && !Tracing::Distributed::SkipPolicy.skip?(
```
marcotc, line 33: Since we have the `tracer` as an argument to `skip?`, can we move the `Tracing.enabled?` check inside?.
That check is just `trace.enabled`.

lib/datadog/tracing/distributed/skip_policy.rb
```diff
@@ -0,0 +1,29 @@
...
+      # Helper method to decide when to skip distributed tracing
+      module SkipPolicy
+        module_function
+
+        # Skips distributed tracing if disabled for this instrumentation
+        # or if APM is disabled unless there is an AppSec event (from upstream distributed trace or local)
+        #
+        # pin_config is a Datadog::Core::Pin object, which gives the configuration of a single instance of an object
+        # global_config is the config for all instances of a framework
+        def skip?(pin_config: nil, global_config: nil, trace: nil)
```
marcotc, line 15: `global_config` is not actually the global config, is it? It reads to me that it is the integration config.

lib/datadog/tracing/distributed/skip_policy.rb
```diff
@@ -0,0 +1,29 @@
...
+          if ::Datadog.configuration.appsec.standalone.enabled &&
+              (trace.nil? || trace.get_tag(::Datadog::AppSec::Ext::TAG_DISTRIBUTED_APPSEC_EVENT) != '1')
```
marcotc, line 17:
```suggestion
          if (trace.nil? || trace.get_tag(::Datadog::AppSec::Ext::TAG_DISTRIBUTED_APPSEC_EVENT) != '1') &&
            ::Datadog.configuration.appsec.standalone.enabled
```
I recommend inverting the check order since the `trace` checks are much cheaper to do than the `configuration` look up.

lib/datadog/tracing/trace_operation.rb
```diff
@@ -72,6 +72,7 @@ def initialize(
         sampling_priority: nil,
         service: nil,
         profiling_enabled: nil,
+        apm_tracing_enabled: nil,
```
marcotc, line 75:
```suggestion
        tracing_enabled: nil,
```

I suggest simplifying this to `tracing_enabled`. Even `tracing_` seems redundant here tbh, but `tracing_enabled` reads well. I think the `apm_` prefix is a bit too much, since there's no ambiguity we are operating on traces here that are part of apm.

lib/datadog/tracing/trace_segment.rb
```diff
@@ -34,7 +34,8 @@ class TraceSegment
         :sampling_decision_maker,
         :sampling_priority,
         :service,
-        :profiling_enabled
+        :profiling_enabled,
+        :apm_tracing_enabled
```
marcotc, line 38: Same here.
```suggestion
        :tracing_enabled
```

lib/datadog/tracing/tracer.rb
```diff
@@ -546,10 +551,30 @@ def skip_trace(name)
...
+      end
+
       def profiling_enabled
         @profiling_enabled ||=
           !!(defined?(Datadog::Profiling) && Datadog::Profiling.respond_to?(:enabled?) && Datadog::Profiling.enabled?)
       end
+
+      def apm_tracing_enabled
+        @apm_tracing_enabled ||= !Datadog.configuration.appsec.standalone.enabled
+      end
```
marcotc, line 577: Please let me know if I'm missing something, but we already have a `self.enabled` method in this class. It disables a tracer if it is configured as disabled when initialized.

My suggestion here, but again please correct me if the requirements are different, is that we used  `Datadog.configuration.appsec.standalone.enabled` as another source for setting the value **at the call site** of `Tracer.new(enabled: ...)`, which would be `lib/datadog/tracing/component.rb`:
Something like:
```ruby
        Tracing::Tracer.new(
          enabled: settings.tracing.enabled && !settings.appsec.standalone.enabled,
          ...
        )
```

lib/datadog/tracing/tracer.rb
```diff
@@ -546,10 +551,30 @@ def skip_trace(name)
...
+      # the upstream tags and the configuration.
+      # We should always propagate if APM is not disabled.
+      #
+      # e.g.: upstream tags containing dd.p.appsec, and appsec is enabled, return true.
+      # DEV: dd.p.appsec will be replaced by dd.p.ts in APM disablement 2.0, which will be a bitmask.
+      def propagate_sampling_priority?(upstream_tags:)
+        return true if apm_tracing_enabled
+
+        if upstream_tags&.key?(Datadog::AppSec::Ext::TAG_DISTRIBUTED_APPSEC_EVENT)
+          return Datadog.configuration.appsec.enabled
```
marcotc, line 564: Can we cache the value of `Datadog.configuration.appsec.enabled` in the constructor of this class, something like `@appsec_enabled = appsec_enabled` in the constructor? Doing this look up on every created span will add overhead.

lib/datadog/tracing/tracer.rb
```diff
@@ -546,10 +551,30 @@ def skip_trace(name)
         end
       end
 
+      # Decide whether upstream sampling priority should be propagated, by taking into account
+      # the upstream tags and the configuration.
+      # We should always propagate if APM is not disabled.
```
marcotc, line 556:
```suggestion
      # We should always propagate if APM is enabled.
```

marcotc (dismissed):

y9v (approved):

lib/datadog/tracing/contrib/ethon/easy_patch.rb
vpellan: I'm not sure that I understand how changing from `skip?` to `enabled?` is going to remove double negatives. If the SkipTracePolicy is enabled, then we don't want to distribute/propagate the trace, or the other way around, we want to propagate the trace if the SkipPolicy is not enabled. so instead of having `!Tracing::Distributed::SkipPolicy.skip?` we'd have `!Tracing::Distributed::SkipTracePolicy.enabled?`. But are you considering `skip?` to be negative ? If that's the case I understand how renaming it to `enabled?` will remove double negatives, I just didn't thought of `skip?` being a negative word.

vpellan: I think I understand, you wanted to rename it to just Policy, as it would be the Policy for Distributed Tracing, not the policy to skip it? Then it would remove the negative like in your example. Maybe I could rename it to PropagationPolicy then ?

lib/datadog/tracing/tracer.rb
vpellan: So I will also address the two previous comments (about the `apm_` prefix). The goal of this fix PR is to work towards APM Disablement V2 PR (which is already ready by the way, but I will ask for reviews once the PR is merged). As stated in the first PR done in October. APM Disablement (then called ASM Standalone, which was a bad choice of name as it was not necessarily tied to ASM) is a mechanism that do not completely disable tracing and traces generated by APM, but rather limits them to 1 per minute, to:
1. Keep the service alive on the backend/UI side
2. Permit other products that uses tracing as transport (e.g. ASM) to force-send traces containing useful informations

The env var for completely disabling tracing is DD_TRACE_ENABLED, while for APM disablement, it will be DD_APM_TRACING_ENABLED. (currently DD_EXPERIMENTAL_APPSEC_STANDALONE_ENABLED). So it cannot be the same method than `enabled`, and to reflect the env var name, I called it `apm_tracing_enabled`. Do you still think we need to change that method or the name of it ?

lib/datadog/tracing/component.rb
vpellan: Done in 366b434d8f0fde32548e88ce269e2ae43d15820d !

lib/datadog/tracing/contrib/ethon/easy_patch.rb
vpellan: Renamed to PropagationPolicy.enabled? (and inverted its logic) in ab38041da4d7f9c09f17a961746755b47cecb15c !

lib/datadog/tracing/contrib/httprb/instrumentation.rb
vpellan: Fixed in 9712b3f60c30aa611493d578b99173fb1ab5f405 !
Sidekiq and GRPC did not had a guard clause for it (meaning that their propagation methods were called, but stopped at a much later point) but it does not break any test (and should actually remove a bit of overhead)
Also, for ethon and rest_client, the check will be done two times (as it was already done previously, in `perform` for ethon and in `execute` for rest_client) which is why I originally left `Tracing.enabled?` outside.
FYI trace is not a `Tracer` object but a `TraceOperation`, which does not contain enabled. It is described in the .rbs file but I believe that we should have a talk about that in a future guild meeting, like should we only describe types in .rbs files or also as comments above methods ?

lib/datadog/tracing/distributed/skip_policy.rb
vpellan: Added comment about that in 9712b3f60c30aa611493d578b99173fb1ab5f405 !

lib/datadog/tracing/distributed/skip_policy.rb
vpellan: Fixed in 2eac1e8f668717d1904b18e49b43690723546a95 !

lib/datadog/tracing/tracer.rb
vpellan: Done in c4334940b3c90ccd61420b28fb229d3d8f0df89e !

lib/datadog/tracing/tracer.rb
vpellan: Fixed in c727d6a10e1b86e5cd8491c57f071701d37c93be !

spec/datadog/tracing/contrib/support/integration/shared_examples.rb
```diff
@@ -0,0 +1,39 @@
+RSpec.shared_examples 'a trace with APM disablement tags' do |params = {}|
```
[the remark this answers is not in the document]
vpellan: Fixed in c727d6a10e1b86e5cd8491c57f071701d37c93be !

lib/datadog/tracing/contrib/ethon/easy_patch.rb
marcotc: Looks great!!!

lib/datadog/tracing/tracer.rb
marcotc: > not completely disable tracing and traces generated by APM, but rather limits them to 1 per minute

Ha! Now that you mention this, I remember the TokenBucket instance added before!

Ok, I have to reread the PR because I was reading this as a complete disablement. Which probably means others might be confused with the names too when stumbling upon the new variables and names introduced in this PR. I'll leave some concrete suggestion after another pass.

marcotc (approved):

lib/datadog/tracing/tracer.rb
```diff
@@ -546,10 +551,31 @@ def skip_trace(name)
...
       def profiling_enabled
         @profiling_enabled ||=
           !!(defined?(Datadog::Profiling) && Datadog::Profiling.respond_to?(:enabled?) && Datadog::Profiling.enabled?)
       end
+
+      def appsec_enabled
+        @appsec_enabled ||= Datadog.configuration.appsec.enabled
+      end
+
+      def apm_tracing_enabled
```
marcotc, line 576: Can you add some docs here explaining what this option does, from a Tracing point of view?

vpellan: Fixed in 839e483 !

PR #4480, opened 2025-03-11
DEBUG-3568 Remove transport API instance/spec branching

p-datadog:
**What does this PR do?**

This PR moves toward having transport API Instance and  Spec classes being API-specific. Currently, these classes are one for a component (e.g. remote/tracing/DI) but these classes combine functionality from multiple APIs and then downstream code has API-specific classes again that reference these "combined" Instance and Spec classes.

The idea is to have all of the transport-related classes be scoped to one API.

This PR removes the branching of one API -> multiple APIs -> one API again but leaves tracing instance & spec classes unchanged which makes them defined in a slightly different way from remote & DI counterparts.

**Motivation:**

Making transport code more readable.

**Change log entry**
None

**Additional Notes:**

**How to test the change?**
Existing unit tests


lib/datadog/di/transport/http/diagnostics.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_relative '../../../core/transport/http/api/spec'
```
Strech, line 3: Should this line be present to?

```
require_relative '../../../core/transport/http/api/instance'
```

lib/datadog/di/transport/http/input.rb
```diff
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_relative '../../../core/transport/http/api/spec'
```
Strech, line 3: Should this line be present to?

```
require_relative '../../../core/transport/http/api/instance'
```

Strech (approved):
👏🏼 awesome

lib/datadog/di/transport/http/diagnostics.rb
p-datadog: You are right, added.

marcotc (approved):
💯

PR #4484, opened 2025-03-12
DRY transport exceptions

p-datadog:
**What does this PR do?**

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**


marcotc (approved):

lib/datadog/core/transport/http/api/instance.rb
```diff
@@ -7,6 +7,23 @@ module HTTP
         module API
           # An API configured with adapter and routes
           class Instance
+            # Raised when an endpoint is invoked on an API that is not the
+            # of expected API class for that endpoint.
+            class EndpointNotSupportedError < StandardError
+              attr_reader :spec, :endpoint_name
+
+              def initialize(spec, endpoint_name)
```
Strech, line 15: I would like to challenge public interface here. If we raise EndpointNotSupportedError, I would expect `endpoint` to be the first argument, and the `spec` to be the second.

Or if we want to keep it this order, I would like to give it explicit naming

```ruby
raise Core::Transport::HTTP::API::Spec::EndpointNotDefinedError.new(self, name: 'info')
raise Core::Transport::HTTP::API::Spec::EndpointNotDefinedError.new(spec: self, name: 'info')
```

or any other variant with explicit initializer. The difference comes from DRY, before it the name was reflected in the class name + endpoint usage.

p-datadog: Makes sense, I flipped the argument order. Thanks for the feedback!

Strech (approved):
👍🏼 Cool

_P.S I have a feeling that we might want to mark second argument with name, but it's not a blocker_

p-datadog:
I am currently focusing on removing duplication from the transport code and making it more streamlined but if you have specific requests/suggestions I can probably sneak them in.

PR #4488, opened 2025-03-13
Fix lock deps job

TonyCTHsu:
**What does this PR do?**

Failed workflow: https://github.com/DataDog/dd-trace-rb/actions/runs/13830185897

`persist-credentials: false` was introduced in https://github.com/DataDog/dd-trace-rb/pull/4456 to improve our Github Actions security posture. However, this job actually depends on `git` operations, hence persisted credential is needed.

The fix is verified with this commit: https://github.com/DataDog/dd-trace-rb/pull/4488/commits/566e058405fa0aace98103a8331295d2dae1f7d1

**Change log entry**
None.


.github/workflows/lock-dependency.yml
```diff
@@ -91,7 +91,7 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
-          persist-credentials: false
+          persist-credentials: true
```
lloeki, line 94: Pleas add a comment explaining why (and by what) it is needed

lloeki (changes requested):

.github/workflows/lock-dependency.yml
TonyCTHsu: https://github.com/DataDog/dd-trace-rb/pull/4488/commits/41d083e1484d881e0d4660ceec3f4789dbb21ce4

.github/workflows/lock-dependency.yml
```diff
@@ -91,7 +91,7 @@ jobs:
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
         with:
-          persist-credentials: true
+          persist-credentials: true # `persist-credentials: true` is required for `git` operations
```
lloeki, line 94: not quite helpful, this is only a describing comment ("of course" it is for git operations because that thing is persisting git auth on a git clone)

I presume this is because of git-auto-commit-action, thus something like that:

```
persist-credentials: true # required for `git-autocommit-action` because XYZ
```

Where `XYZ` is actually the important bit.

It might even hint at how to do it (if it can be done) without persisting; e.g maybe the secret token needs to be passed to the action via some `with: secrets:`?

lloeki (approved):

.github/workflows/lock-dependency.yml
TonyCTHsu: https://github.com/DataDog/dd-trace-rb/pull/4491

PR #4492, opened 2025-03-13
[NO-TICKET] Experiment to replace `Profile_add` libdatadog interning API

ivoanjo:
**What does this PR do?**

This is an experiment to adopt the new APIs in https://github.com/DataDog/libdatadog/pull/917 .

**Motivation:**

Validate ongoing work in https://github.com/DataDog/libdatadog/pull/917 .

**Change log entry**

None... For now. If this is a nice win we may want to brag here ;)

**Additional Notes:**

Very early, very work in progress!

**How to test the change?**

This will not work at all in CI because it depends on the libdatadog branch I link to there.

### DIFF
--- a/ext/datadog_profiling_native_extension/stack_recorder.c
+++ b/ext/datadog_profiling_native_extension/stack_recorder.c
@@ -607,6 +607,110 @@ static VALUE ruby_time_from(ddog_Timespec ddprof_time) {
   return rb_time_timespec_new(&time, utc);
 }
 
+static ddog_prof_StringId intern_string_or_raise(locked_profile_slot locked_profile, ddog_CharSlice string) {
+  ddog_prof_StringId_Result result = ddog_prof_Profile_intern_string(&locked_profile.data->profile, string);
+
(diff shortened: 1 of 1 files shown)


ivoanjo:
At some point I did some very quick benchmarking and this PR was a win just by itself.

Re-running the benchmarking in our benchmarking platform + on a separate machine:

```
ruby 3.3.6 (2024-11-05 revision 75015d4c1f) [x86_64-linux]
profiling - stack collector master-900-2
                       107.000 i/100ms
profiling - stack collector master-900-2
                          1.072k (± 1.9%) i/s  (932.96 μs/i) -    964.284k in 899.975328s

Comparison:
profiling - stack collector intern-900-2:     1112.0 i/s
profiling - stack collector intern-900-1:     1111.8 i/s - same-ish: difference falls within error
profiling - stack collector master-900-2:     1071.9 i/s - 1.04x  slower
profiling - stack collector master-900-1:     1063.6 i/s - 1.05x  slower
```

The results are always within the margin of error.

Note that this only applies to this PR and it's incredibly naive replacement of `Profile_add` with new API. I still expect that a deeper integration where we use the interning API correctly and cache/reuse many of its results will move the needle, but that's not what this PR does.

Thus, I'm going to close this PR.

PR #4495, opened 2025-03-14
[NO-TICKET] Fix profiling flaky tests due to monotonic-to-system clock conversions

ivoanjo:
**What does this PR do?**

This PR fixes flaky test https://github.com/DataDog/ruby-guild/issues/220:

```
Failures:

  1) Datadog::Profiling::Collectors::ThreadContext#sample timeline support when timeline is enabled when thread starts Waiting for GVL records a first sample to represent the time between the previous sample and the start of Waiting for GVL
     Failure/Error:
       expect(first_sample.labels).to include(
         state: "sleeping",
         end_timestamp_ns: be_between(@time_before_gvl_waiting, @time_after_gvl_waiting),
       )

       expected {:end_timestamp_ns => 1741859359178414403, :state => "sleeping", :"thread id" => "11298 (22400)", :"thread name" => "/__w/dd-trace-rb/dd-trace-rb/spec/spec_helper.rb:253"} to include {end_timestamp_ns: (be between 1741859359178419843 and 1741859359178427868 (inclusive))}
       Diff:
       @@ -1,2 +1,4 @@
       -:end_timestamp_ns => (be between 1741859359178419843 and 1741859359178427868 (inclusive)),
       +:"thread id" => "11298 (22400)",
       +:"thread name" => "/__w/dd-trace-rb/dd-trace-rb/spec/spec_helper.rb:253",
       +:end_timestamp_ns => 1741859359178414403,
        :state => "sleeping",
     # ./spec/datadog/profiling/collectors/thread_context_spec.rb:1168:in 'block (6 levels) in <top (required)>'
     # ./spec/spec_helper.rb:238:in 'block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:123:in 'block (2 levels) in <top (required)>'
     # /usr/local/bundle/gems/webmock-3.25.0/lib/webmock/rspec.rb:39:in 'block (2 levels) in <top (required)>'

Finished in 19.02 seconds (files took 0.92371 seconds to load)
763 examples, 1 failure, 19 pending
```

by making sure our monotonic-to-system clock conversion cache does not affect these specs.

**Motivation:**

Keep profiling at zero known flaky tests!

**Change log entry**

None

**Additional Notes:**

In c8f2338b201230849c6cc255c565eed06dc5d214 we had already fixed a very similar issue.

TL;DR there's a cache in the monotonic-to-system-clock conversion code inside the profiler.

Because this conversion includes a cache, any tests that compare timestamps in profiles to `Time.now` could become flaky because of drift between clocks.

That is, because the profiler is **estimating** the system clock based on this cache, it may say that something happened a few nanos before `Time.now` when in fact it happened a few nanos after. Thus, all tests comparing timestamps to `Time.now` must be careful around resetting this cache to avoid running into this issue.

**How to test the change?**

Since this is a flaky test that depends on drift between clocks to trigger, I wasn't able to trigger it. But, other than that, the existing test coverage should be enough to cover this change.


spec/datadog/profiling/collectors/thread_context_spec.rb
```diff
@@ -131,6 +145,10 @@ def apply_delta_to_cpu_time_at_previous_sample_ns(thread, delta_ns)
       ._native_apply_delta_to_cpu_time_at_previous_sample_ns(cpu_and_wall_time_collector, thread, delta_ns)
   end
 
+  def handle_reset_request(do_reset)
```
p-datadog, line 148: I think it  would be clearer for this method to be argument-free and unconditional and lift the conditional checking to the callers.

spec/datadog/profiling/collectors/thread_context_spec.rb
```diff
@@ -1219,7 +1237,7 @@ def sample_and_check(expected_state:)
             time_before_sample = Datadog::Core::Utils::Time.as_utc_epoch_ns(Time.now)
             monotonic_time_sanity_check = Datadog::Core::Utils::Time.get_time(:nanosecond)
 
-            sample
+            sample(reset_monotonic_to_system_state: true)
```
p-datadog, line 1240: ... which would also eliminate the weird `true` argument in all of these calls.

p-datadog (approved):

spec/datadog/profiling/collectors/thread_context_spec.rb
p-datadog, line 148: I think it would be clearer to either make this method argument-free and unconditional (and lift conditional checking to callers), or rename it to something like `maybe_reset_monotonic_to_system_state` indicating its conditional nature.

ivoanjo: Fixed in 6c53e3382af53cb6351956a8bd8105147faf4cbe

PR #4496, opened 2025-03-14
Update lograge/patcher.rb so that the correct logger is checked

esposito:
**What does this PR do?**
In this PR I make sure that when the Lograge logger is present, it is checked instead of the Rails logger for showing a warning about tagged logging. 

**Motivation:**
When the lograge `keep_original_rails_log` option is used, the lograge logger cannot be found in `::Lograge::LogSubscribers::ActionController`, as the existing actioncontroller log subscriber is never removed. Therefore the warning in `lograge/patcher.rb` is shown even when lograge does use a plain logger.

**Change log entry**
Hide the warning about tagged logging when Lograge uses a plain logger and `keep_original_rails_log` is enabled.

**Additional Notes:**

**How to test the change?**
The change should be covered by the tests. If you want to test it manually, enable the Lograge `keep_original_rails_log` config option, keep the default Rails logger but use a plain logger in the lograge config. Now there should no longer be an error about tagged logging.

### DIFF
--- a/lib/datadog/tracing/contrib/lograge/patcher.rb
+++ b/lib/datadog/tracing/contrib/lograge/patcher.rb
@@ -20,10 +20,12 @@ module Datadog
 
           # patch applies our patch
           def patch
+            # First check Lograge logger directly for when keep_original_rails_log option is used
+            used_logger = ::Lograge.logger || ::Lograge::LogSubscribers::ActionController.logger
+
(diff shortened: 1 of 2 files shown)


marcotc (approved):
Awesome work!

MathijsK93:
Are there any updates on the release of this?

Strech (approved):

Strech:
Hey @MathijsK93, working on it 😉

PR #4505, opened 2025-03-17
feat(propagation): baggage support and automatic propagation

ZStriker19:
**What does this PR do?**
Following the spec here: https://www.w3.org/TR/baggage/ we implement baggage for the Ruby tracer. The otel drop in support will be added in a future PR.
**Motivation:**

**Change log entry**
* Tracing: Added support for W3C Baggage extraction and injection by default.
* Tracing: Adds api method to access and modify baggage object.

**Additional Notes:**

**How to test the change?**
Tests were added in the repo and it passes the [weblog system-tests.](https://github.com/DataDog/system-tests/pull/3929) It will not pass the parametric system-tests because of the difficulty of managing the context in server.rb, not due an issue of implementation of the actual feature.


ZStriker19:
Following up from https://github.com/DataDog/dd-trace-rb/pull/4371 which was merged with issues

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -0,0 +1,133 @@
+# frozen_string_literal: true
+
+require_relative '../metadata/ext'
+require_relative '../trace_digest'
+require_relative 'datadog_tags_codec'
+require_relative '../utils'
+require_relative 'helpers'
+require 'uri'
+require 'cgi'
```
marcotc, line 9:
```suggestion
require 'cgi'
```
This is no longer used.

lib/datadog/tracing/trace_operation.rb
```diff
@@ -101,6 +103,7 @@ def initialize(
         @trace_state = trace_state
         @trace_state_unknown_fields = trace_state_unknown_fields
         @tracer = tracer
+        @baggage = baggage || {}
```
marcotc, line 106:
```suggestion
        @baggage = baggage
```

Can you avoid initializing the baggage with an empty hash? It's a good suggestion by @p-datadog!
You'll have to initialize it then in `tracing.rb` if it's not initialized.

marcotc (approved):

PR #4507, opened 2025-03-18
Add option to override span events serialization

marcotc:
For agent-less setups, when native support is desired (e.g. the downstream agent supports native span events), this PR adds an option to override the agent decision and send native span events regardless of the environment.

**Change log entry**
Yes. Add an option to override span events serialization for agent-less environments.

**How to test the change?**
There are unit tests, and this change will allow to better coverage in parametric tests.


lib/datadog/tracing/configuration/settings.rb
```diff
@@ -249,6 +249,17 @@ def self.extended(base)
...
+              # Forces the tracer to always send span events with the native span events format
+              # regardless of the agent support. This is useful in agent-less setups.
+              #
+              # @default `DD_TRACE_NATIVE_SPAN_EVENTS` environment variable, otherwise `false`
+              # @return [Boolean,nil]
+              option :native_span_events do |o|
+                o.env Tracing::Configuration::Ext::ENV_NATIVE_SPAN_EVENTS
+                o.default nil
+                o.type :bool, nilable: true
+              end
```
ivoanjo, line 261: I think it's clear in the PR, where all of the diff shows up together what `nil` means, but when looking only at the setting definition, it's not very explicit what happens with the `nil` (and the comment for the default is actually incorrect).

TL;DR: I think it's worth updating the comment to state what happens in both `false` and `nil` situations.

(As an alternative, there's a situation in the profiler where we have this true/false/other config and [we've explicitly used `:auto`](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/core/configuration/settings.rb#L401) although it slightly complicates parsing the setting)

spec/datadog/tracing/configuration/settings_spec.rb
```diff
@@ -503,6 +503,45 @@ def propagation_style_inject
...
+
+          it { is_expected.to be true }
+        end
+
+        context 'is set to false' do
+          let(:environment) { 'false' }
+
+          it { is_expected.to be false }
+        end
+      end
```
ivoanjo, line 533: Minor: There's a helper in the `core/configuration/settings_spec.rb` that may be useful to use here. It allows this to be replaced with

```ruby
it_behaves_like 'a binary setting with', env_variable: 'DD_TRACE_NATIVE_SPAN_EVENTS', default: nil
```

spec/datadog/tracing/transport/traces_spec.rb
ivoanjo, line 1: Note to other reviewers: Review this file with `Hide whitespace` enabled on github ;)

spec/datadog/tracing/transport/traces_spec.rb
```diff
@@ -283,59 +283,113 @@
...
+
+            send_traces
+          end
+
+          it 'does not query the agent' do
+            allow(Datadog::Tracing::Transport::Traces::Chunker).to receive(:new).and_return(chunker)
+            expect_any_instance_of(Datadog::Core::Environment::AgentInfo).not_to receive(:fetch)
+
             send_traces
           end
```
ivoanjo, line 311: :+1: Nice, protects against accidentally still calling the agent

ivoanjo (approved):
:+1: LGTM

docs/GettingStarted.md
```diff
@@ -2072,6 +2072,7 @@ For example, if `tracing.sampling.default_rate` is configured by [Remote Configu
...
+| `tracing.native_span_events`                           | `DD_TRACE_NATIVE_SPAN_EVENTS`                           | `Bool`                                | `false`                      | Forces the tracer to always send span events with the native span events format regardless of the agent support. This is useful to change the serialization format in agent-less setups.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
```
maycmlee, line 2075: What do you mean by "regardless of the Agent support"?

```suggestion
| `tracing.native_span_events`                           | `DD_TRACE_NATIVE_SPAN_EVENTS`                           | `Bool`                                | `false`                      | Forces the tracer to always send span events with the native span events format, regardless of the Agent support. This is useful when you want to change the serialization format in Agent-less setups.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
```

maycmlee (commented):
A small suggestion and question.

docs/GettingStarted.md
marcotc: @maycmlee this feature is normally set to `true` or `false` depending on the agent support for the feature (we call the agent to check). But with this variable, we are overriding this behaviour, and skipping that agent check. This means we'll sent these Span Events in whatever format the environment variable configures, regardless if the agent supports it or not.

docs/GettingStarted.md
```diff
@@ -2072,6 +2072,7 @@ For example, if `tracing.sampling.default_rate` is configured by [Remote Configu
...
+| `tracing.native_span_events`                           | `DD_TRACE_NATIVE_SPAN_EVENTS`                           | `Bool`                                | `false`                      | Forces the tracer to always send span events with the native span events format, regardless of the Agent support. This is useful when you want to change the serialization format in Agent-less setups.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
```
maycmlee, line 2075:
```suggestion
| `tracing.native_span_events`                           | `DD_TRACE_NATIVE_SPAN_EVENTS`                           | `Bool`                                | `false`                      | Forces the tracer to always send span events with the native span events format, regardless of whether the Agent supports it. This is useful when you want to change the serialization format in Agent-less setups.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
```

maycmlee (approved):
@marcotc thanks for the explanation! just a small suggestion, but going to approve to not block this.

lib/datadog/tracing/configuration/settings.rb
marcotc: I added documentation, nice call out!

spec/datadog/tracing/configuration/settings_spec.rb
marcotc: Wow, I didn't know about it! I'll refactor it!

spec/datadog/tracing/transport/traces_spec.rb
marcotc: 👍

TonyCTHsu (approved):

Issue #4474, opened 2025-03-08
[FEATURE]: Use relative paths for resource names in opensearch integration


minamijoyo:
### Describe the goal of the feature

The OpenSearch resource names use relative paths, the same as Elasticsearch.

### Is your feature request related to a problem?

While migrating from Elasticsearch to OpenSearch, I noticed that the resource names in Datadog's APM are not relative paths in OpenSearch, but full paths, which include a fully qualified domain name. The problem is that no matter how short the domain name the user can specify, in VPC mode, the length of FQDN will be 60-70 characters.
While I understand the idea that different domain names should be treated as different resources, for practical purposes, the resource name is too long on the screen, and the end is not displayed, resulting in resources that cannot be distinguished.

Note that the original elasticsearch integration uses relative paths as resource names.
https://github.com/DataDog/dd-trace-rb/blob/v2.12.1/lib/datadog/tracing/contrib/elasticsearch/patcher.rb#L49-L50
https://github.com/DataDog/dd-trace-rb/blob/v2.12.1/lib/datadog/tracing/contrib/opensearch/patcher.rb#L51-L52

Besides elasticsearch, the mysql integration also could theoretically have multiple instances in a single application, but the endpoint domain name is not included in the resource name. This is not particularly problematic because multiple instances are rare. Given that, I believe the same is true for the opensearch integration. I'm unsure whether the difference between Elasticsearch integration and OpenSearch one is intentional, so I submitted this as a feature request.

sarahchen6:
Hi @minamijoyo! Thanks for raising this issue :) [This PR](https://github.com/DataDog/dd-trace-rb/pull/4509) adds an env var `DD_TRACE_OPENSEARCH_RESOURCE_PATTERN` that should allow you to set the resource pattern to be the relative URL path instead of the default absolute URL path. See [here](https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#opensearch) for more details.

The change will be released with the next `datadog` gem version (2.15.0). Thanks again!

minamijoyo:
@sarahchen6 Thank you for the fix 🎉

PR #4509, opened 2025-03-18
Update OpenSearch resource names to be relative instead of full path

sarahchen6:
**What does this PR do?**
This PR updates OpenSearch resource names to use relative paths, instead of full paths, similar to how ElasticSearch does.

**Motivation:**
Resolve customer issue migrating from ElasticSearch to OpenSearch: https://github.com/DataDog/dd-trace-rb/issues/4474

**Change log entry**
Yes. Tracing: Update OpenSearch integration to include a feature flag for resource names. By default and with the existing implementation, the resource name is set to the full URL path; however, now resource names can instead be set to use relative paths.

**Additional Notes:**

**How to test the change?**
Tested through unit tests (`bundle exec rspec spec/datadog/tracing/contrib/opensearch/patcher_spec.rb`).


p-datadog (approved):
I think this also needs a doc added to https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#opensearch. Approved with that added.

estherk15 (approved):

p-datadog (approved):

TonyCTHsu (approved):

PR #4526, opened 2025-03-19
Add Stack Trace reporting for AppSec actions

y9v:
**What does this PR do?**
This PR adds Stack Trace reporting to the Datadog Agent in case of an AppSec event with `generate_stack` action.

**Motivation:**
We are not handling `generate_stack` action yet.

**Change log entry**
Yes. AppSec: Add stack trace reporting for security events.

**Additional Notes:**
None.

**How to test the change?**
CI and local testing with app-generator (SSRF or SQLi variants).

### DIFF
--- a/lib/datadog/appsec/actions_handler.rb
+++ b/lib/datadog/appsec/actions_handler.rb
@@ -30,8 +30,8 @@ module Datadog
         active_span = AppSec.active_context&.span
         return unless active_span
 
-        event_category = Datadog::AppSec::Ext::EXPLOIT_PREVENTION_EVENT_CATEGORY
-        tag_key = Datadog::AppSec::Ext::TAG_METASTRUCT_STACK_TRACE
+        event_category = Ext::EXPLOIT_PREVENTION_EVENT_CATEGORY
+        tag_key = Ext::TAG_METASTRUCT_STACK_TRACE
 
         existing_stack_data = active_span.get_metastruct_tag(tag_key).dup || { event_category => [] }
         max_stack_traces = Datadog.configuration.appsec.stack_trace.max_stack_traces
--- a/lib/datadog/appsec/actions_handler.rb
+++ b/lib/datadog/appsec/actions_handler.rb
@@ -22,16 +22,24 @@ module Datadog
       end
 
       def generate_stack(action_params)
-        return unless Datadog.configuration.appsec.stack_trace.enabled
-
-        # TODO: check how many stack traces we already collected
-
         stack_id = action_params['stack_id']
         return unless stack_id
 
-        _backtrace = SerializableBacktrace.new(locations: Array(caller_locations), stack_id: stack_id)
+        return unless Datadog.configuration.appsec.stack_trace.enabled
 
-        # TODO: add stack frames to span metastruct
+        active_span = AppSec.active_context&.span
+        return unless active_span
+
+        event_category = Datadog::AppSec::Ext::EXPLOIT_PREVENTION_EVENT_CATEGORY
+        tag_key = Datadog::AppSec::Ext::TAG_METASTRUCT_STACK_TRACE
+
+        existing_stack_data = active_span.get_metastruct_tag(tag_key).dup || { event_category => [] }
+        return if existing_stack_data[event_category].count >= Datadog.configuration.appsec.stack_trace.max_stack_traces
+
+        backtrace = SerializableBacktrace.new(locations: Array(caller_locations), stack_id: stack_id)
+
+        existing_stack_data[event_category] << backtrace
+        active_span.set_metastruct_tag(tag_key, existing_stack_data)
       end
 
       def generate_schema(_action_params); end
--- a/lib/datadog/appsec/actions_handler.rb
+++ b/lib/datadog/appsec/actions_handler.rb
@@ -37,7 +37,6 @@ module Datadog
         return if existing_stack_data[event_category].count >= Datadog.configuration.appsec.stack_trace.max_stack_traces
 
         backtrace = SerializableBacktrace.new(locations: Array(caller_locations), stack_id: stack_id)
-
         existing_stack_data[event_category] << backtrace
         active_span.set_metastruct_tag(tag_key, existing_stack_data)
       end
--- a/lib/datadog/appsec/actions_handler.rb
+++ b/lib/datadog/appsec/actions_handler.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative 'actions_handler/serializable_backtrace'
+
 module Datadog
   module AppSec
     # this module encapsulates functions for handling actions that libddawf returns
@@ -19,7 +21,16 @@ module Datadog
         throw(Datadog::AppSec::Ext::INTERRUPT, action_params)
       end
 
-      def generate_stack(_action_params); end
+      def generate_stack(action_params)
+        # TODO: check how many stack traces we already collected
+
+        stack_id = action_params['stack_id']
+        return unless stack_id
+
+        _backtrace = SerializableBacktrace.new(locations: caller_locations, stack_id: stack_id)
+
+        # TODO: add stack frames to span metastruct
+      end
 
       def generate_schema(_action_params); end
     end
--- a/lib/datadog/appsec/actions_handler.rb
+++ b/lib/datadog/appsec/actions_handler.rb
@@ -29,7 +29,7 @@ module Datadog
         stack_id = action_params['stack_id']
         return unless stack_id
 
-        _backtrace = SerializableBacktrace.new(locations: caller_locations, stack_id: stack_id)
+        _backtrace = SerializableBacktrace.new(locations: Array(caller_locations), stack_id: stack_id)
 
         # TODO: add stack frames to span metastruct
       end
--- a/lib/datadog/appsec/actions_handler.rb
+++ b/lib/datadog/appsec/actions_handler.rb
@@ -22,6 +22,8 @@ module Datadog
       end
 
       def generate_stack(action_params)
+        return unless Datadog.configuration.appsec.stack_trace.enabled
+
         # TODO: check how many stack traces we already collected
 
         stack_id = action_params['stack_id']
--- a/lib/datadog/appsec/actions_handler.rb
+++ b/lib/datadog/appsec/actions_handler.rb
@@ -22,10 +22,10 @@ module Datadog
       end
 
       def generate_stack(action_params)
+        return unless Datadog.configuration.appsec.stack_trace.enabled
+
         stack_id = action_params['stack_id']
         return unless stack_id
-
-        return unless Datadog.configuration.appsec.stack_trace.enabled
 
         active_span = AppSec.active_context&.span
         return unless active_span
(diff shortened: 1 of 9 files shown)


lib/datadog/appsec/actions_handler.rb
```diff
@@ -19,7 +21,25 @@ def interrupt_execution(action_params)
         throw(Datadog::AppSec::Ext::INTERRUPT, action_params)
       end
 
-      def generate_stack(_action_params); end
+      def generate_stack(action_params)
+        stack_id = action_params['stack_id']
+        return unless stack_id
+
+        return unless Datadog.configuration.appsec.stack_trace.enabled
```
vpellan, line 28: Would it be better to put that guard clause first ?

Strech, line 28: WDYT if we move it to be the very first check?

lib/datadog/appsec/actions_handler.rb
```diff
@@ -19,7 +21,25 @@ def interrupt_execution(action_params)
...
+        active_span = AppSec.active_context&.span
+        return unless active_span
+
+        event_category = Datadog::AppSec::Ext::EXPLOIT_PREVENTION_EVENT_CATEGORY
+        tag_key = Datadog::AppSec::Ext::TAG_METASTRUCT_STACK_TRACE
+
+        existing_stack_data = active_span.get_metastruct_tag(tag_key).dup || { event_category => [] }
+        return if existing_stack_data[event_category].count >= Datadog.configuration.appsec.stack_trace.max_stack_traces
+
+        backtrace = SerializableBacktrace.new(locations: Array(caller_locations), stack_id: stack_id)
```
vpellan, line 39: Is `Array(caller_locations)` necessary ? `caller_locations` is already an Array

lib/datadog/appsec/actions_handler.rb
```diff
@@ -19,7 +21,25 @@ def interrupt_execution(action_params)
...
+        stack_id = action_params['stack_id']
+        return unless stack_id
+
+        return unless Datadog.configuration.appsec.stack_trace.enabled
+
+        active_span = AppSec.active_context&.span
+        return unless active_span
+
+        event_category = Datadog::AppSec::Ext::EXPLOIT_PREVENTION_EVENT_CATEGORY
+        tag_key = Datadog::AppSec::Ext::TAG_METASTRUCT_STACK_TRACE
```
Strech, line 34: We can shrink the path to just `Ext::`, maybe it will help with long names

lib/datadog/appsec/actions_handler/serializable_backtrace.rb
```diff
@@ -0,0 +1,80 @@
...
+          @serializable_locations_map = locations.each_with_object({}) do |location, map|
+            # we are dropping frames from library code without increasing frame index
+            next if location.path&.include?('lib/datadog')
+
+            frame_idx += 1
+
+            next if frame_idx >= drop_from_idx && frame_idx < drop_until_idx
+
+            map[frame_idx] = location
+          end
```
Strech, line 33: I think this should be done only when needed and we can keep initializer clean and only about settings.

lib/datadog/appsec/actions_handler/serializable_backtrace.rb
```diff
@@ -0,0 +1,80 @@
...
+            packer.write('text')
+            packer.write(location.to_s.encode('UTF-8'))
+
+            packer.write('file')
+            packer.write(location.path&.encode('UTF-8'))
+
+            packer.write('line')
+            packer.write(location.lineno)
+
+            class_name, function_name = location.label.match(/\b([\w+:{2}]*\w+)?[#|.]?\b(\w+)\z/)&.captures
```
Strech, line 66: Maybe we can move regex into the constant with clear name

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -164,6 +164,63 @@ def self.add_settings!(base)
...
+                # Minimum value is 10.
+                #
+                # Default value is 32
+                option :max_depth do |o|
+                  o.type :int
+                  o.env 'DD_APPSEC_MAX_STACK_TRACE_DEPTH'
+                  o.default 32
+
+                  o.setter do |value|
+                    value = 10 if value < 10
```
vpellan, line 189: I cannot find a reference that says that the minimum value should be 10. If it's in an RFC, can you send it in DM ? Also, it can be set to 0, which means that we accept an unlimited number of frames (sending you the RFC in DM)

lib/datadog/appsec/configuration/settings.rb
```diff
@@ -164,6 +164,63 @@ def self.add_settings!(base)
...
+                # Maximum number of stack traces to collect per span.
+                #
+                # Default value is 2
+                option :max_stack_traces do |o|
+                  o.type :int
+                  o.env 'DD_APPSEC_MAX_STACK_TRACES'
+                  o.default 2
+
+                  o.setter do |value|
+                    value = 1 if value < 1
```
vpellan, line 218: Same than 'DD_APPSEC_MAX_STACK_TRACE_DEPTH', if set to 0, it means unlimited number of stack trace

lib/datadog/appsec/actions_handler/serializable_backtrace.rb
```diff
@@ -0,0 +1,80 @@
...
+  module AppSec
+    module ActionsHandler
+      # This module serves encapsulates MessagePack serialization for caller locations.
+      #
+      # It serializes part of the stack:
+      # up to 32 frames (configurable)
+      # keeping frames from top and bottom of the stack (75% to 25%, configurable).
+      #
+      # It represents the stack trace that is added to span metastruct field.
+      class SerializableBacktrace
```
Strech, line 13: WDYT if we split interface to streamline the arguments, into something like this

```ruby
backtrace = SerializableBacktrace.new(Array(caller_locations))
backtrace.id = stack_id
# or
backtrace.set_id(stack_id)
# or
backtrace.serialized_id = stack_id
# or
backtrace.set_serialized_id(stack_id)
```

it will show that to be able to generate backtrace you need just the locations.
But to serialize it - you also need a `stack_id` to set its ID value.

spec/datadog/appsec/actions_handler/serializable_backtrace_spec.rb
```diff
@@ -0,0 +1,176 @@
+# frozen_string_literal: true
+
+require 'datadog/appsec/spec_helper'
+
+RSpec.describe Datadog::AppSec::ActionsHandler::SerializableBacktrace do
+  let(:location_struct) do
+    Struct.new(:path, :lineno, :label) do
```
Strech, line 7: Maybe we can use `Thread::Backtrace::Location` for a good reference?

spec/datadog/appsec/actions_handler/serializable_backtrace_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+
+    after do
+      Datadog.configuration.appsec.reset!
+    end
+
+    it 'correctly serializes stack attributes' do
+      result = pack_and_unpack(described_class.new(locations: [], stack_id: 'some-id'))
+
+      expect(result.fetch('id')).to eq('some-id')
+      expect(result.fetch('language')).to eq('ruby')
```
Strech, line 28: Just throwing idea (if you like it)

```suggestion
      expect(result).to include('id' => 'some-id', 'language' => 'ruby')
```

spec/datadog/appsec/actions_handler/serializable_backtrace_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+
+    it 'correctly serializes stack frames' do
+      location = location_struct.new('path/to/file.rb', 15, 'SomeModule::SomeClass#some_method')
+
+      result = pack_and_unpack(described_class.new(locations: [location], stack_id: 'some-id'))
+      frames = result.fetch('frames')
+
+      expect(frames.size).to eq(1)
+
+      aggregate_failures('frame attributes') do
```
Strech, line 39: Nice move 👍🏼

spec/datadog/appsec/actions_handler_spec.rb
```diff
@@ -97,4 +97,81 @@
...
+        active_span.set_metastruct_tag(
+          Datadog::AppSec::Ext::TAG_METASTRUCT_STACK_TRACE,
+          { 'exploit' => [1, 2] }
+        )
+      end
+
+      it 'does not add stack trace to metastruct' do
+        expect do
+          described_class.generate_stack(action_params)
+        end.not_to(change { active_span.get_metastruct_tag(Datadog::AppSec::Ext::TAG_METASTRUCT_STACK_TRACE) })
```
Strech, line 151: To avoid `end.` if you like, it could be formatted like this

```suggestion
        expect { described_class.generate_stack(action_params) }
          .not_to change { active_span.get_metastruct_tag(Datadog::AppSec::Ext::TAG_METASTRUCT_STACK_TRACE) }
```

vpellan (approved):
I think that other than the env vars defaults, it's good to go !

Strech (approved):
Well done 👍🏼 

I've left a few non-blocking suggestions

lib/datadog/appsec/actions_handler.rb
y9v: according to steep, `caller_locations` might return `nil`. `Array(nil)` returns an empty array

lib/datadog/appsec/configuration/settings.rb
y9v: good catch, thanks! I haven't noticed that in our RFCs

y9v: fixed in [4bab1cd](https://github.com/DataDog/dd-trace-rb/pull/4526/commits/4bab1cdb5813c41db770901bc2e70770d5254234)

lib/datadog/appsec/configuration/settings.rb
y9v: also fixed here [4bab1cd](https://github.com/DataDog/dd-trace-rb/pull/4526/commits/4bab1cdb5813c41db770901bc2e70770d5254234), thanks for the comment!

spec/datadog/appsec/actions_handler_spec.rb
y9v: thanks! I always have trouble formatting those expect blocks

lib/datadog/appsec/actions_handler/serializable_backtrace.rb
y9v: fixed [0d8e709](https://github.com/DataDog/dd-trace-rb/pull/4526/commits/0d8e70904322dba2fc64922f350f9c41731f0a53)

lib/datadog/appsec/actions_handler/serializable_backtrace.rb
y9v: fixed [0d8e709](https://github.com/DataDog/dd-trace-rb/pull/4526/commits/0d8e70904322dba2fc64922f350f9c41731f0a53)

spec/datadog/appsec/actions_handler/serializable_backtrace_spec.rb
y9v: done using `instance_double` in [bf5323c](https://github.com/DataDog/dd-trace-rb/pull/4526/commits/bf5323c792643880171a19cd121f8eba79b10617)

lib/datadog/appsec/actions_handler/serializable_backtrace.rb
y9v: I would actually leave `stack_id` as part of the initialize args, since it's a `SerializableBacktrace`, and we don't call `to_msgpack` method explicitly

PR #4528, opened 2025-03-19
chore(logging): enable 128bit trace id logging by default

mabdinur:
**What does this PR do?**

- Updates the default value of `DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED` to `True`.

**Motivation:**

Ensures the full 128bit trace ids are injected into logs instead of just the lower order 64bits. This ensures the logs injected into traces is consistent with the trace id shown in the Datadog UI.

**Change log entry**

Updates the default value of DD_TRACE_128_BIT_TRACEID_LOGGING_ENABLED to True, ensuring logs contain the full 128-bit trace ID for consistency with the trace ID displayed in the Datadog UI.

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/spec/support/tracer_helpers.rb
+++ b/spec/support/tracer_helpers.rb
@@ -126,7 +126,7 @@ module TracerHelpers
 
   # Ensures the given trace ID is always formatted using the 128-bit logging format
   # by wrapping Tracing::Correlation.format_trace_id(trace_id)
-  def log_injection_trace_id_128(trace_id)
+  def trace_id_128_log_injection(trace_id)
     RSpec::Mocks.with_temporary_scope do
       allow(Datadog.configuration.tracing).to receive(:trace_id_128_bit_logging_enabled).and_return(true)
       Datadog::Tracing::Correlation.format_trace_id(trace_id)
--- a/spec/support/tracer_helpers.rb
+++ b/spec/support/tracer_helpers.rb
@@ -126,7 +126,7 @@ module TracerHelpers
   end
 
   ## Wraps call to Datadog::Tracing::Correlation.format_trace_id_128 for better test readability
-  def trace_id_128_log_injection(trace_id)
+  def format_for_correlation(trace_id)
     Datadog::Tracing::Correlation.format_trace_id_128(trace_id)
   end
 
--- a/spec/support/tracer_helpers.rb
+++ b/spec/support/tracer_helpers.rb
@@ -120,7 +120,7 @@ module TracerHelpers
   end
 
   # Wraps call to Tracing::Utils::TraceId.to_low_order for better test readability
-  def low_order_trace_id
+  def low_order_trace_id(trace_id)
     Datadog::Tracing::Utils::TraceId.to_low_order(trace_id)
   end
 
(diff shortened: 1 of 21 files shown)


TonyCTHsu (dismissed):
Changing default is not trivia, how does this change going to impact current users?

mabdinur:
> Changing default is not trivia, how does this change going to impact current users?

This is a change we are rolling out across all client libraries. Ruby and PHP are the last set of libraries to receive this change. This change should not impact customers using Datadog log pipelines. Datadog log pipelines fully support correlating logs to traces using both 64bit and 128bit trace ids. This change could impact customers who export their logs to other providers. This is an edge case we are okay with.

spec/support/tracer_helpers.rb
```diff
@@ -123,6 +124,15 @@ def low_order_trace_id(trace_id)
     Datadog::Tracing::Utils::TraceId.to_low_order(trace_id)
   end
 
+  # Ensures the given trace ID is always formatted using the 128-bit logging format
+  # by wrapping Tracing::Correlation.format_trace_id(trace_id)
+  def trace_id_128_log_injection(trace_id)
```
p-datadog, line 129: I am confused as to the purpose of this method. It appears to change the trace format and then outputs the trace, by doing so, this method changes the behavior potentially being tested by the tests. Why is this approach taken in favor of, for example, some conditional logic to check either format? Or not asserting on trace contents if it is irrelevant?

mabdinur: This method should have this logic:

```ruby
if !Tracing::Utils::TraceId.to_high_order(trace_id).zero?
  Kernel.format('%032x', trace_id)
else
  Tracing::Utils::TraceId.to_low_order(trace_id).to_s
end
```

This is essentially the same logic provided by [Datadog::Tracing::Correlation.format_trace_id](https://github.com/DataDog/dd-trace-rb/blob/1fa3344bb414444bd3b2abed194334eae5aa5228/lib/datadog/tracing/correlation.rb#L96) but without the `Datadog.configuration.tracing.trace_id_128_bit_logging_enabled` condition. By mocking `Datadog.configuration.tracing.trace_id_128_bit_logging_enabled` to always return `True` I am able to get the desired functionality without duplication.

mabdinur: Right now most logging tests validate that 64bit trace ids are logged by default. With this PR I want most tests to validate that 128bit trace ids are logged by default.

To do this I mock `trace_id_128_bit_logging_enabled` configuration to always return `True`. I don't want this helper to used in test cases where `trace_id_128_bit_logging_enabled` is set to `False`. 

I can refactor this code so that we don't need the mocking.

mabdinur: I updated the method

spec/support/tracer_helpers.rb
```diff
@@ -123,6 +125,11 @@ def low_order_trace_id(trace_id)
     Datadog::Tracing::Utils::TraceId.to_low_order(trace_id)
   end
 
+  ## Wraps call to Datadog::Tracing::Correlation.format_trace_id_128 for better test readability
+  def trace_id_128_log_injection(trace_id)
```
marcotc, line 129: Can we rename this to `format_for_correlation`? I think it succinctly captures what it's trying to do.

spec/support/tracer_helpers.rb
marcotc: Thank you!

marcotc (approved):

p-datadog (approved):

PR #4543, opened 2025-03-28
Propagate current trace context to Active Record async query threads

### LINKED ISSUE #3465: Active record async queries are considered separate traces
**Current behaviour**

When leveraging Active Record's async queries functionality ([`load_async`](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-load_async), [`async_count`](https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-async_count), etc.), the queries that are executed, while noticed by ddtrace and wrapped in spans, are considered separate traces, and are therefore not easily discoverable / correlated to the overall trace (whether it be a web request, background job, or something else).

After looking briefly at the ddtrace source, my hunch is that the reason these are not all part of the same trace is that the query is executed in a different thread from the main trace, and ddtrace is not referencing the parent trace correctly.

**Expected behaviour**

Active Record async queries should be part of the overall trace.

**Steps to reproduce**

With diagnostics turned on locally, you can clearly see the trace ID attached to the async query spans is different for each query, and not the same as the overall request trace ID.

**Environment**

* **ddtrace version:** 1.19.0
* **Ruby version:** 3.3.0
* **Operating system:** macOS 14.3.1, Ubuntu 22.04
* **Relevant library versions:** Rails / Active Record 7.1.3

agrobbin:
**What does this PR do?**

The Active Record integration (for anything greater than v7.0.0) will now wrap `ConnectionPool#build_async_executor` with the already-existing `ContextCompositeExecutorService` from the concurrent-ruby tracer integration, ensuring that the current trace is continued within the thread.

Closes #3465.

**Motivation:**

When leveraging Active Record's async query functionality (`load_async`, `async_count`, etc.), the queries are executed in a different thread, via concurrent-ruby's `ThreadPoolExecutor`. This causes the span for that query to not show up in the parent trace (often a `rack.request`).

**How to test the change?**

I included tests here, but am definitely open to other approaches here!

### DIFF
--- a/lib/datadog/tracing/contrib/active_record/patcher.rb
+++ b/lib/datadog/tracing/contrib/active_record/patcher.rb
@@ -2,6 +2,7 @@
 
 require_relative '../patcher'
 require_relative 'events'
+require_relative 'async_executor/connection_pool'
 
 module Datadog
   module Tracing
@@ -19,6 +20,10 @@ module Datadog
 
           def patch
             Events.subscribe!
+
+            if Integration.version >= Gem::Version.new('7.0.0') && ConcurrentRuby::Integration.patchable?
+              ::ActiveRecord::ConnectionAdapters::ConnectionPool.prepend(AsyncExecutor::ConnectionPool)
+            end
           end
         end
       end
--- a/spec/datadog/tracing/contrib/active_record/async_spec.rb
+++ b/spec/datadog/tracing/contrib/active_record/async_spec.rb
@@ -1,0 +1,100 @@
+require 'datadog/tracing/contrib/support/spec_helper'
+require 'datadog'
+
+require 'spec/datadog/tracing/contrib/rails/support/deprecation'
+
+require 'active_record'
+
+if PlatformHelpers.jruby?
+  require 'activerecord-jdbc-adapter'
+else
+  require 'pg'
+end
+
+RSpec.describe 'ActiveRecord async query instrumentation' do
+  let(:application_record) do
+    stub_const('ApplicationRecord', Class.new(ActiveRecord::Base)).tap do |klass|
+      klass.abstract_class = true
+    end
+  end
+  let!(:widget_class) { stub_const('Widget', Class.new(application_record)) }
+
+  before do
+    skip 'Test applies only to Active Record 7.0 or higher' if ActiveRecord.version < '7.0.0'
+
+    # Reset options (that might linger from other tests)
+    Datadog.configuration.tracing[:active_record].reset!
+
+    Datadog.configure do |c|
+      c.tracing.instrument :concurrent_ruby
+      c.tracing.instrument :active_record
+      c.tracing.instrument :pg
+    end
+
+    raise_on_rails_deprecation!
+
+    ActiveRecord.async_query_executor = :global_thread_pool
+
+    # Connect the Widget database
+    root_pw = ENV.fetch('TEST_POSTGRES_ROOT_PASSWORD', 'postgres')
+    host = ENV.fetch('TEST_POSTGRES_HOST', '127.0.0.1')
+    port = ENV.fetch('TEST_POSTGRES_PORT', '5432')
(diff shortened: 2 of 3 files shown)


spec/datadog/tracing/contrib/active_record/async_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+      async_count
+    end
+
+    # Ensure the async executor thread pool is shut down, to avoid leaking threads
+    widget_class.connection_pool.async_executor.composited_executor.shutdown
+
+    expect(spans).to include(
+      an_object_having_attributes(name: 'async_query.test', trace_id: root_trace.id),
+      an_object_having_attributes(name: 'pg.exec.params', trace_id: root_trace.id),
```
agrobbin, line 90: I ran into a very odd scenario here that led me to assert against `pg.exec.params` rather than `postgres.query`.

For reasons unknown to me, even with this patch, the `sql.active_record` Active Support notification does not seem to get tracked and converted into a DD span. I imagine this is due to a similar issue with the multi-threaded nature of the async query executor, but wasn't able to track it down in short order, so thought I would get this out and see if others had an idea!

lib/datadog/tracing/contrib/active_record/patcher.rb
```diff
@@ -19,6 +20,10 @@ def target_version
 
           def patch
             Events.subscribe!
+
+            if Integration.version >= Gem::Version.new('7.0.0') && ConcurrentRuby::Integration.patchable?
```
agrobbin, line 24: I would love to hear if this is the "right" way to leverage a different DD tracing integration. There is a bit of a weird interplay between Active Record and Concurrent Ruby, so would love to hear if this is OK or if there is different preferred approach.

spec/datadog/tracing/contrib/active_record/async_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+  let!(:widget_class) { stub_const('Widget', Class.new(application_record)) }
+
+  before do
+    skip 'Test applies only to Active Record 7.0 or higher' if ActiveRecord.version < '7.0.0'
+
+    # Reset options (that might linger from other tests)
+    Datadog.configuration.tracing[:active_record].reset!
+
+    Datadog.configure do |c|
+      c.tracing.instrument :concurrent_ruby
```
agrobbin, line 29: Strictly speaking, this isn't actually necessary, but similar to the weirdness I raised in https://github.com/DataDog/dd-trace-rb/pull/4543#discussion_r2019081549, the interaction between Active Record and Concurrent Ruby is a little odd.

agrobbin:
@marcotc was just wondering if this was something you'd be able to take a look at when you get a chance. It would be great to get this into a release soon! 🤞🏻

marcotc:
@agrobbin, yes! It's pretty much what I've been doing in the last two weeks 😅: https://github.com/DataDog/dd-trace-rb/pull/4872

It required quite a major change in how we flush traces and track background threads (which is a bug that greatly affects the concurrent-ruby instrumentation). More specifically, this didn't work correctly: https://github.com/DataDog/dd-trace-rb/blob/44947b27e86b871e6028d5bece5087a145cabcb5/spec/datadog/tracing/tracer_spec.rb#L993-L1001 when `span-1` finished the whole wrapping trace would finish, and `span-2` would create a new trace as a fallback that was not associated with the `digest`.
This affects async queries because multiple queries can run in the background for a single `load_async`, depending on the DB adapter, and those would create new, unrelated traces in the background.

I still need to to fix all failing tests, and add new tests to ensure I didn't break basic span flushing.

agrobbin:
Ohh, wow, way more complicated than I thought! Looking forward to seeing all of that work land. 🎉

ivoanjo:
#4872 is still work-in-progress, but should supercede this PR, so I'm going to go ahead and close off this one! :pray:

PR #4548, opened 2025-03-31
Fix a warning related to regular expression in AppSec

y9v:
**What does this PR do?**
It fixes a following warning:
```bash
dd-trace-rb/lib/datadog/appsec/actions_handler/serializable_backtrace.rb:14: warning: character class has duplicated range: /\b([\w+:{2}]*\w+)?[#|.]?\b(\w+)\z/
```

**Motivation:**
We don't want any warnings to be triggered by our library code.

**Change log entry**
None. This change is internal.

**Additional Notes:**
None.

**How to test the change?**
```bash
ruby -W lib/datadog/appsec/actions_handler/serializable_backtrace.rb
```

PR #4550, opened 2025-03-31
Add tests verifying library is -w warning-clean

p-datadog:
**What does this PR do?**

Adds a test that no warnings (`ruby -v`) are emitted when the tracer is loaded (#4548, #4547, #4548)

**Motivation:**

Customer feedback that they don't want tracer to emit warnings

**Change log entry**
None (warnings removed by the other PRs, this PR only adds test coverage)

**Additional Notes:**

**How to test the change?**
Tests are added in this PR

### DIFF
--- a/spec/loading_spec.rb
+++ b/spec/loading_spec.rb
@@ -1,6 +1,8 @@
 require 'shellwords'
+require 'open3'
 
 REQUIRES = {
+  'datadog' => 'Datadog::Core',
   'datadog/appsec' => 'Datadog::AppSec',
   'datadog/core' => 'Datadog::Core',
   'datadog/di' => 'Datadog::DI',
@@ -34,6 +36,16 @@ RSpec.describe 'loading of products' do
       it 'loads successfully by itself' do
         rv = system("ruby -e #{Shellwords.shellescape(code)}")
         expect(rv).to be true
+      end
+
+      it 'produces no output' do
+        out, status = Open3.capture2e('ruby', '-w', stdin_data: code)
+        unless status.exitstatus == 0
+          fail("Test script failed with exit status #{status.exitstatus}:\n#{out}")
+        end
+        unless out.empty?
+          fail("Test script produced unexpected output: #{out}")
+        end
       end
     end
   end
--- a/spec/loading_spec.rb
+++ b/spec/loading_spec.rb
@@ -6,9 +6,9 @@ REQUIRES = [
   ['datadog/appsec', 'Datadog::AppSec'],
   ['datadog/core', 'Datadog::Core'],
   ['datadog/di', 'Datadog::DI',
-    lambda { RUBY_VERSION >= '2.6' && RUBY_ENGINE != 'jruby' }],
+   -> { RUBY_VERSION >= '2.6' && RUBY_ENGINE != 'jruby' }],
   ['datadog/di/preload', 'Datadog::DI::CodeTracker',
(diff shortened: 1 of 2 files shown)


Strech (changes requested):
This will be changed in #4433 with this lines https://github.com/DataDog/dd-trace-rb/pull/4433/files#diff-a2d6d6d68e8af1a8701d9cbb5c926cbae65bd3292847196f7d29979db30cf2b6R53-R57

lib/datadog/appsec/contrib/devise/patcher.rb
```diff
@@ -36,7 +36,10 @@ def patch_rememberable_strategy
             return unless ::Devise::STRATEGIES.include?(:rememberable)
 
             # Rememberable strategy is required in autoloaded Rememberable model
-            ::Devise::Models::Rememberable # rubocop:disable Lint/Void
+            # Ensure Rememberable is loaded here, we call module_eval to
+            # avoid a Ruby warning when run with -w
+            ::Devise::Models::Rememberable.module_eval do
+            end
```
y9v, line 42: I would not add this to our library code - this is purely for testing, and I think we can keep it in tests?

y9v (changes requested):

p-datadog:
I removed the appsec change.

y9v (approved):
awesome!

spec/loading_spec.rb
```diff
@@ -1,18 +1,28 @@
...
-  'datadog/di/preload' => 'Datadog::DI::CodeTracker',
-  'datadog/kit' => 'Datadog::Kit',
-  'datadog/profiling' => 'Datadog::Profiling',
-  'datadog/tracing' => 'Datadog::Tracing',
-}.freeze
+REQUIRES = [
+  ['datadog', 'Datadog::Core'],
+  ['datadog/appsec', 'Datadog::AppSec'],
+  ['datadog/core', 'Datadog::Core'],
+  ['datadog/di', 'Datadog::DI',
```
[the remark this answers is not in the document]
p-datadog: It's extra typing and I do not need it, if you insist I will make it the way you are asking.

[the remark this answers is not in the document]
y9v: no, I don't mind it staying as it is

spec/loading_spec.rb
```diff
@@ -35,6 +45,12 @@
         rv = system("ruby -e #{Shellwords.shellescape(code)}")
         expect(rv).to be true
       end
+
+      it 'produces no output' do
```
Strech, line 49: Just to make it clear what we are expecting, I would suggest slight rewording

```suggestion
      it 'produces no warnings in the output'
```

spec/loading_spec.rb
```diff
@@ -35,6 +45,12 @@
...
+        out, status = Open3.capture2e('ruby', '-w', stdin_data: code)
+        raise("Test script failed with exit status #{status.exitstatus}:\n#{out}") unless status.exitstatus == 0
```
Strech, line 51: Let's use the `Process::Status` api for the check

```suggestion
        raise("Test script failed with exit status #{status.exitstatus}:\n#{out}") unless status.success?
```

Strech (approved):
LGTM

spec/loading_spec.rb
p-datadog: I think from the customers' point of view it does not matter if dd-trace-rb produces warnings or some other output, any output that is not meaningful to them shouldn't be produced.

PR #4552, opened 2025-04-01
Add handling of string keys for `Kit:AppSec::Events`

Strech:
**What does this PR do?**

Add unified handling for string keys for AppSec kit events.

**Motivation:**

This change is motivated by customer who uses string keys (which are impossible till Ruby 2.7) to pass complex keys like `usr.login`. They wasn't handled before.

**Change log entry**

Yes. AppSec: Fix `Datadog::Kit::AppSec::Events` methods to correctly handle given string key `usr.login`.

**Additional Notes:**

**How to test the change?**


y9v (approved):
awesome!

spec/support/platform_helpers.rb
```diff
@@ -24,6 +24,11 @@ def engine_version
     Gem::Version.new(version)
   end
 
+  def engine_version_matches?(matcher_with_ruby_version)
+    operator, version = matcher_with_ruby_version.split(' ', 2).tap { |array| array.unshift('==') if array.size == 1 }
```
p-datadog, line 28: I think the space should be optional, i.e. `>=2.7` should be handled. What happens when the condition is misinterpreted by this code?

p-datadog: E.g. in the present implementation, if the condition is `>=2.7`, will the test always be executed or always skipped or something else?

Strech: Maybe it wasn't clear, but nothing will be silently skipped it will fail with error like this

```
❯❯❯ be rspec spec/datadog/kit/appsec/events_spec.rb:195
Run options: include {:focus=>true, :locations=>{"./spec/datadog/kit/appsec/events_spec.rb"=>[195]}}

Randomized with seed 41032

Datadog::Kit::AppSec::Events
  #track_login_failure
    sets additional user login data from other string keys as tags (FAILED - 1)

Failures:

  1) Datadog::Kit::AppSec::Events#track_login_failure sets additional user login data from other string keys as tags
     Failure/Error: version = Gem::Version.new("#{guard_version}.0")

     ArgumentError:
       Malformed version number string >=2.7.0
     # ./spec/support/platform_helpers.rb:32:in `ruby_version_matches?'
     # ./spec/spec_helper.rb:123:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:259:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:144:in `block (2 levels) in <top (required)>'

Finished in 0.00089 seconds (files took 0.36278 seconds to load)
1 example, 1 failure
```

do you think it's not good enough? I've added documentation to the filter with most common use cases.

p-datadog (approved):

PR #4559, opened 2025-04-04
feat: error event is now subscribable

dubloom:
**What does this PR do?**
Make `:on_error` span_operation event consistent with the others events. `:on_error` event is now subscribable. The previous `wrap_default` behavior has not been changed.

**Motivation:**
I'm currently working on reporting handled errors automatically. This new feature needs to be able to know when an error happened (for deduplication purpose). 

**Change log entry**

None.


spec/datadog/tracing/span_operation_spec.rb
```diff
@@ -66,7 +66,7 @@
 
       # on_error
       allow(callback_spy).to receive(:on_error)
-      events.on_error.wrap_default do |*args|
+      events.on_error.subscribe do |*args|
```
p-datadog, line 69: If the existing `wrap_default` behavior has not changed, does this change in the test suite effectively remove test coverage for the existing behavior?

dubloom: It does not because for instance here: https://github.com/DataDog/dd-trace-rb/blob/edcfe6a36b0ec38f03dde21f2801aef47e9697f5/spec/datadog/tracing/tracer_spec.rb#L554C59-L554C74, we are providing a block to tracer.trace which will call wrap_default. 

However, on that topic, do you know if that change is enough ? Looping in @delner as he created the test

p-datadog: I do not have much context for this PR - only what is in the description - but I would say that, if `wrap_default` usage continues to be supported, it would be better to keep the test that explicitly invokes `wrap_default` and add another test which calls `subscribe`. This should require little code in Ruby since you can use `send` and reuse the entire setup, let me know if you would like assistance with this.

For your second question are you asking if the test coverage is sufficient for the added `subscribe` method?

delner: I haven't touched this code in ages, and its been refactored a few times, so I'm not sure I'm that helpful here. I do share @p-datadog 's concern: test coverage should be explicit, around each of the public APIs. Coverage by implicit means is lost if we refactor, and it would not be obvious.

Typically we don't test private methods (because they aren't a part of the public interface and are meant to be disposable.) However, because this is a public library, and users can and WILL use anything that's publicly accessible, we privatize some methods that are meant to be stable, but that we don't want users to access (basically package-access methods.) This interface is one of those things: not for public use, meant for Datadog internal use. It should be covered/tested like a public method.

dubloom: There is not much more context than in the motivation section. I need for a new feature to subscribe to the on_error event and I saw that it was the only non conventional event so I made the change. I'll add back the test however, your arguments are very valid. Thank you for your insights !

dubloom:
@p-datadog I updated the tests, tell me if it works for you

dubloom:
After discussions, automatic report of handled errors will use another event. I'm closing this PR now

PR #4561, opened 2025-04-04
[APMAPI-1068] Process discovery

vpellan:
**What does this PR do?**

Store tracer configuration in an in-memory file using libdatadog.

**Motivation:**

Service Discovery will become the primary entry point of all products built on top of tracing libraries.

**Change log entry**

Yes. Add Service Discovery capability, enabling the tracer to reliably determine which process are instrumented using Datadog's tracer.

**Additional Notes:**

This PR cannot be merged until libdatadog 17.1 and its corresponding gem is released, and libdatadog gem is updated on dd-trace-rb.

**How to test the change?**

Parametric system-tests (cannot be force-enabled yet)

### DIFF
--- a/ext/libdatadog_api/init.c
+++ b/ext/libdatadog_api/init.c
@@ -4,15 +4,10 @@
 #include "crashtracker.h"
 #include "process_discovery.h"
 
-// Used to report Ruby VM crashes.
-// Once initialized, segfaults will be reported automatically using libdatadog.
-
 void DDTRACE_EXPORT Init_libdatadog_api(void) {
   VALUE datadog_module = rb_define_module("Datadog");
   VALUE core_module = rb_define_module_under(datadog_module, "Core");
-  VALUE crashtracking_module = rb_define_module_under(core_module, "Crashtracking");
-  VALUE process_discovery_module = rb_define_module_under(core_module, "ProcessDiscovery");
 
-  crashtracker_init(crashtracking_module);
-  process_discovery_init(process_discovery_module);
+  crashtracker_init(core_module);
+  process_discovery_init(core_module);
 }
--- a/ext/libdatadog_api/init.c
+++ b/ext/libdatadog_api/init.c
@@ -1,0 +1,18 @@
+#include <ruby.h>
+
+#include "datadog_ruby_common.h"
+#include "crashtracker.h"
+#include "process_discovery.h"
+
+// Used to report Ruby VM crashes.
+// Once initialized, segfaults will be reported automatically using libdatadog.
+
+void DDTRACE_EXPORT Init_libdatadog_api(void) {
+  VALUE datadog_module = rb_define_module("Datadog");
+  VALUE core_module = rb_define_module_under(datadog_module, "Core");
+  VALUE crashtracking_module = rb_define_module_under(core_module, "Crashtracking");
+  VALUE process_discovery_module = rb_define_module_under(core_module, "ProcessDiscovery");
+
+  crashtracker_init(crashtracking_module);
+  process_discovery_init(process_discovery_module);
+}
--- a/ext/libdatadog_api/process_discovery.c
+++ b/ext/libdatadog_api/process_discovery.c
@@ -1,3 +1,4 @@
+#include <errno.h>
 #include <ruby.h>
 #include <datadog/common.h>
 
--- a/ext/libdatadog_api/process_discovery.c
+++ b/ext/libdatadog_api/process_discovery.c
@@ -8,6 +8,13 @@
 static VALUE _native_store_tracer_metadata(int argc, VALUE *argv, DDTRACE_UNUSED VALUE _self);
 static VALUE _native_close_tracer_memfd(DDTRACE_UNUSED VALUE _self, VALUE fd);
 
+static VALUE log_error(VALUE error) {
+  VALUE datadog_module = rb_const_get(rb_cObject, rb_intern("Datadog"));
+  VALUE logger = rb_funcall(datadog_module, rb_intern("logger"), 0);
+
+  return rb_funcall(logger, rb_intern("warn"), 1, error);
+}
+
 static const rb_data_type_t tracer_memfd_type = {
   .wrap_struct_name = "Datadog::Core::ProcessDiscovery::TracerMemfd",
   .function = {
@@ -19,7 +26,8 @@ static const rb_data_type_t tracer_memfd_type = {
 
 void process_discovery_init(VALUE core_module) {
   VALUE process_discovery_class = rb_define_class_under(core_module, "ProcessDiscovery", rb_cObject);
-  rb_define_class_under(process_discovery_class, "TracerMemfd", rb_cObject);
+  VALUE tracer_memfd_class = rb_define_class_under(process_discovery_class, "TracerMemfd", rb_cObject);
+  rb_undef_alloc_func(tracer_memfd_class); // Class cannot be instantiated from Ruby
 
   rb_define_singleton_method(process_discovery_class, "_native_store_tracer_metadata", _native_store_tracer_metadata, -1);
   rb_define_singleton_method(process_discovery_class, "_native_close_tracer_memfd", _native_close_tracer_memfd, 1);
@@ -60,15 +68,13 @@ static VALUE _native_store_tracer_metadata(int argc, VALUE *argv, VALUE self) {
   );
 
   if (result.tag == DDOG_RESULT_TRACER_MEMFD_HANDLE_ERR_TRACER_MEMFD_HANDLE) {
-    rb_raise(rb_eRuntimeError, "Failed to store the tracer configuration in a memory file descriptor: %"PRIsVALUE, get_error_details_and_drop(&result.err));
+    log_error(rb_sprintf("Failed to store the tracer configuration in a memory file descriptor: %"PRIsVALUE, get_error_details_and_drop(&result.err)));
+    return Qnil;
   }
 
   // &result.ok is a ddog_TracerMemfdHandle, which is a struct only containing int fd, which is a file descriptor
   // We should just return the fd
-  int *fd = malloc(sizeof(int));
-  if (!fd) {
-    rb_raise(rb_eRuntimeError, "Failed to allocate memory for file descriptor");
-  }
+  int *fd = ruby_xmalloc(sizeof(int));
 
   *fd = result.ok.fd;
   VALUE tracer_memfd_class = rb_const_get(self, rb_intern("TracerMemfd"));
@@ -80,14 +86,21 @@ static VALUE _native_close_tracer_memfd(DDTRACE_UNUSED VALUE _self, VALUE tracer
   int *fd;
   TypedData_Get_Struct(tracer_memfd, int, &tracer_memfd_type, fd);
   if (fd == NULL) {
-    rb_raise(rb_eRuntimeError, "Failed to get the tracer configuration memory file descriptor: %s", strerror(errno));
+    log_error(rb_sprintf("Failed to get the tracer configuration memory file descriptor: %s", strerror(errno)));
+    return Qnil;
+  }
+
+  if (*fd == -1) {
+    log_error(rb_sprintf("The tracer configuration memory file descriptor has already been closed"));
+    return Qnil;
   }
 
   int close_result = close(*fd);
-  free(fd);
+  *fd = -1;
 
   if (close_result == -1) {
-    rb_raise(rb_eRuntimeError, "Failed to close the tracer configuration memory file descriptor: %s", strerror(errno));
+    log_error(rb_sprintf("Failed to close the tracer configuration memory file descriptor: %s", strerror(errno)));
+    return Qnil;
   }
 
   return Qnil;
--- a/ext/libdatadog_api/process_discovery.c
+++ b/ext/libdatadog_api/process_discovery.c
@@ -93,11 +93,6 @@ static VALUE _native_to_rb_int(DDTRACE_UNUSED VALUE _self, VALUE tracer_memfd) {
 static VALUE _native_close_tracer_memfd(DDTRACE_UNUSED VALUE _self, VALUE tracer_memfd) {
   int *fd;
   TypedData_Get_Struct(tracer_memfd, int, &tracer_memfd_type, fd);
-  if (fd == NULL) {
-    log_error(rb_sprintf("Failed to get the tracer configuration memory file descriptor: %s", strerror(errno)));
-    return Qnil;
-  }
-
   if (*fd == -1) {
     log_error(rb_sprintf("The tracer configuration memory file descriptor has already been closed"));
     return Qnil;
--- a/ext/libdatadog_api/process_discovery.c
+++ b/ext/libdatadog_api/process_discovery.c
@@ -7,14 +7,7 @@
 
 static VALUE _native_store_tracer_metadata(int argc, VALUE *argv, DDTRACE_UNUSED VALUE _self);
 static VALUE _native_to_rb_int(DDTRACE_UNUSED VALUE _self, VALUE tracer_memfd);
-static VALUE _native_close_tracer_memfd(DDTRACE_UNUSED VALUE _self, VALUE fd);
-
-static VALUE log_error(VALUE error) {
-  VALUE datadog_module = rb_const_get(rb_cObject, rb_intern("Datadog"));
-  VALUE logger = rb_funcall(datadog_module, rb_intern("logger"), 0);
-
-  return rb_funcall(logger, rb_intern("warn"), 1, error);
-}
+static VALUE _native_close_tracer_memfd(DDTRACE_UNUSED VALUE _self, VALUE tracer_memfd, VALUE logger);
 
 static const rb_data_type_t tracer_memfd_type = {
   .wrap_struct_name = "Datadog::Core::ProcessDiscovery::TracerMemfd",
@@ -32,12 +25,13 @@ void process_discovery_init(VALUE core_module) {
 
   rb_define_singleton_method(process_discovery_class, "_native_store_tracer_metadata", _native_store_tracer_metadata, -1);
   rb_define_singleton_method(process_discovery_class, "_native_to_rb_int", _native_to_rb_int, 1);
-  rb_define_singleton_method(process_discovery_class, "_native_close_tracer_memfd", _native_close_tracer_memfd, 1);
+  rb_define_singleton_method(process_discovery_class, "_native_close_tracer_memfd", _native_close_tracer_memfd, 2);
 }
 
 static VALUE _native_store_tracer_metadata(int argc, VALUE *argv, VALUE self) {
+  VALUE logger;
   VALUE options;
-  rb_scan_args(argc, argv, "0:", &options);
+  rb_scan_args(argc, argv, "1:", &logger, &options);
   if (options == Qnil) options = rb_hash_new();
 
   VALUE schema_version = rb_hash_fetch(options, ID2SYM(rb_intern("schema_version")));
@@ -70,7 +64,7 @@ static VALUE _native_store_tracer_metadata(int argc, VALUE *argv, VALUE self) {
   );
 
   if (result.tag == DDOG_RESULT_TRACER_MEMFD_HANDLE_ERR_TRACER_MEMFD_HANDLE) {
-    log_error(rb_sprintf("Failed to store the tracer configuration in a memory file descriptor: %"PRIsVALUE, get_error_details_and_drop(&result.err)));
+    rb_funcall(logger, rb_intern("debug"), 1, rb_sprintf("Failed to store the tracer configuration in a memory file descriptor: %"PRIsVALUE, get_error_details_and_drop(&result.err)));
     return Qnil;
   }
 
@@ -90,11 +84,11 @@ static VALUE _native_to_rb_int(DDTRACE_UNUSED VALUE _self, VALUE tracer_memfd) {
   return INT2NUM(*fd);
 }
 
-static VALUE _native_close_tracer_memfd(DDTRACE_UNUSED VALUE _self, VALUE tracer_memfd) {
+static VALUE _native_close_tracer_memfd(DDTRACE_UNUSED VALUE _self, VALUE tracer_memfd, VALUE logger) {
   int *fd;
   TypedData_Get_Struct(tracer_memfd, int, &tracer_memfd_type, fd);
   if (*fd == -1) {
-    log_error(rb_sprintf("The tracer configuration memory file descriptor has already been closed"));
+    rb_funcall(logger, rb_intern("debug"), 1, rb_sprintf("The tracer configuration memory file descriptor has already been closed"));
     return Qnil;
   }
 
@@ -102,7 +96,7 @@ static VALUE _native_close_tracer_memfd(DDTRACE_UNUSED VALUE _self, VALUE tracer
   *fd = -1;
 
   if (close_result == -1) {
-    log_error(rb_sprintf("Failed to close the tracer configuration memory file descriptor: %s", strerror(errno)));
+    rb_funcall(logger, rb_intern("debug"), 1, rb_sprintf("Failed to close the tracer configuration memory file descriptor: %s", strerror(errno)));
     return Qnil;
   }
 
--- a/ext/libdatadog_api/process_discovery.c
+++ b/ext/libdatadog_api/process_discovery.c
@@ -5,8 +5,8 @@
 
 static VALUE _native_store_tracer_metadata(int argc, VALUE *argv, DDTRACE_UNUSED VALUE _self);
 
-void process_discovery_init(VALUE process_discovery_module) {
-  VALUE process_discovery_class = rb_define_class_under(process_discovery_module, "Component", rb_cObject);
+void process_discovery_init(VALUE core_module) {
+  VALUE process_discovery_class = rb_define_class_under(core_module, "ProcessDiscovery", rb_cObject);
 
   rb_define_singleton_method(process_discovery_class, "_native_store_tracer_metadata", _native_store_tracer_metadata, -1);
 }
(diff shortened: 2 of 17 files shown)


ivoanjo (commented):
Left a few notes!

ext/libdatadog_api/process_discovery.c
```diff
@@ -0,0 +1,55 @@
...
+  ENFORCE_TYPE(runtime_id, T_STRING);
+  ENFORCE_TYPE(tracer_language, T_STRING);
+  ENFORCE_TYPE(tracer_version, T_STRING);
+  ENFORCE_TYPE(hostname, T_STRING);
+  ENFORCE_TYPE(service_name, T_STRING);
+  ENFORCE_TYPE(service_env, T_STRING);
+  ENFORCE_TYPE(service_version, T_STRING);
+
+  ddog_Result_TracerMemfdHandle result = ddog_store_tracer_metadata(
+    (uint8_t)schema_version,
```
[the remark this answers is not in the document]
vpellan: I actually took some time to research about NUM2UNIT and FIX2UNIT all of that to forget to add it :') I'm still wondering why should I use NUM2UNIT rather than FIX2UNIT ?

ext/libdatadog_api/init.c
```diff
@@ -0,0 +1,18 @@
...
+// Once initialized, segfaults will be reported automatically using libdatadog.
+
+void DDTRACE_EXPORT Init_libdatadog_api(void) {
+  VALUE datadog_module = rb_define_module("Datadog");
+  VALUE core_module = rb_define_module_under(datadog_module, "Core");
+  VALUE crashtracking_module = rb_define_module_under(core_module, "Crashtracking");
+  VALUE process_discovery_module = rb_define_module_under(core_module, "ProcessDiscovery");
+
+  crashtracker_init(crashtracking_module);
+  process_discovery_init(process_discovery_module);
```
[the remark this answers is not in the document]
vpellan: Thanks, moved their definintions in their corresponding modules in d663056576e8626488eebb0d1a3f7e63aa997467

ext/libdatadog_api/init.c
```diff
@@ -0,0 +1,18 @@
+#include <ruby.h>
+
+#include "datadog_ruby_common.h"
+#include "crashtracker.h"
+#include "process_discovery.h"
+
+// Used to report Ruby VM crashes.
+// Once initialized, segfaults will be reported automatically using libdatadog.
```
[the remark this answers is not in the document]
vpellan: Removed in d663056576e8626488eebb0d1a3f7e63aa997467

ext/libdatadog_api/process_discovery.c
```diff
@@ -0,0 +1,55 @@
...
+
+#include "datadog_ruby_common.h"
+
+static VALUE _native_store_tracer_metadata(int argc, VALUE *argv, DDTRACE_UNUSED VALUE _self);
+
+void process_discovery_init(VALUE process_discovery_module) {
+  VALUE process_discovery_class = rb_define_class_under(process_discovery_module, "Component", rb_cObject);
+
+  rb_define_singleton_method(process_discovery_class, "_native_store_tracer_metadata", _native_store_tracer_metadata, -1);
+}
```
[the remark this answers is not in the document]
vpellan: Thanks, I changed it to a class in d663056576e8626488eebb0d1a3f7e63aa997467

lib/datadog/core/process_discovery.rb
```diff
@@ -0,0 +1,23 @@
...
+module Datadog
+  module Core
+    # Class  used to store tracer metadata in a native file descriptor.
+    class ProcessDiscovery
+      # According to RFC, runtime_id, service_name, service_env, service_version are optional.
+      # ddcommon memfd_create exposer replace empty strings in these fields by None.
+      def self.get_metadata(settings)
+        {
+          schema_version: 1,
+          runtime_id: Core::Environment::Identity.id || '',
```
ivoanjo, line 12: Wait -- when would `id` be `nil`?

lib/datadog/core/configuration/components.rb
```diff
@@ -123,6 +124,9 @@ def initialize(settings)
           @appsec = Datadog::AppSec::Component.build_appsec_component(settings, telemetry: telemetry)
           @dynamic_instrumentation = Datadog::DI::Component.build(settings, agent_settings, @logger, telemetry: telemetry)
           @environment_logger_extra[:dynamic_instrumentation_enabled] = !!@dynamic_instrumentation
+          @process_discovery_fd = Core::ProcessDiscovery._native_store_tracer_metadata(
+            **Core::ProcessDiscovery.get_metadata(settings)
+          )
```
ivoanjo, line 129: Minor: Consider maybe adding a helper function that does these two things together -- `components.rb` is already extremely complex so any simplification of internal details that aren't relevant to components we can do is probably worth it.

ext/libdatadog_api/process_discovery.c
```diff
@@ -0,0 +1,68 @@
...
+  return INT2FIX(result.ok.fd);
+}
+
+static VALUE _native_close_tracer_memfd(DDTRACE_UNUSED VALUE _self, VALUE fd) {
+  ENFORCE_TYPE(fd, T_FIXNUM);
+
+  int close_result = close(NUM2INT(fd));
+  if (close_result == -1) {
+    rb_raise(rb_eRuntimeError, "Failed to close the tracer configuration: %s", strerror(errno));
+  }
```
ivoanjo, line 65: Hmm this is a bit primitive obsession-ish... May be worth using `TypedData_Make_Struct` to introduce an actual object to represent this?

Doing that would also open the door to improve a bit safety -- right now I can pass any integer at any time, whereas with control of the type it's possible that only correct fds can be closed (and to enforce that they can only be closed once).

ext/libdatadog_api/process_discovery.c
```diff
@@ -0,0 +1,68 @@
...
+  // We should just return the fd to test everything is working
+  return INT2FIX(result.ok.fd);
+}
+
+static VALUE _native_close_tracer_memfd(DDTRACE_UNUSED VALUE _self, VALUE fd) {
+  ENFORCE_TYPE(fd, T_FIXNUM);
+
+  int close_result = close(NUM2INT(fd));
+  if (close_result == -1) {
+    rb_raise(rb_eRuntimeError, "Failed to close the tracer configuration: %s", strerror(errno));
```
ivoanjo, line 64: This error message is quite generic, and may be visible to customers when we log it. Consider perhaps making it more specific, so it's more obvious where the issue was (e.g. we're talking about the memfd thing?)

sig/datadog/core/process_discovery.rbs
```diff
@@ -0,0 +1,20 @@
...
+        tracer_language: String,
+        tracer_version: String,
+        hostname: String,
+        service_name: String,
+        service_env: String,
+        service_version: String
+      ) -> Integer # C file descriptor
+    end
+
+    def self._native_close_tracer_memfd: (Integer fd) -> nil
```
ivoanjo, line 16: Minor:

```suggestion
    def self._native_close_tracer_memfd: (Integer fd) -> void
```

You can use `void` at the type system level to tell the typechecker that the return of this method is not supposed to be used. (E.g. there's always something, because it's Ruby, but this way you'll get screamed at if you try to use it)

ivoanjo (commented):
Left a few notes!

lib/datadog/core/configuration/components.rb
vpellan: Added in e3fa9491d0fd4190a921cf44044d37c85926561b

ext/libdatadog_api/process_discovery.c
vpellan: Fixed in e3fa9491d0fd4190a921cf44044d37c85926561b, although I used TypedData_Wrap_Struct (I couldn't get TypedData_Make_Struct to work) let me know if you think there's still improvements that can be made on that !

ext/libdatadog_api/process_discovery.c
vpellan: Changed the message a little bit to make it more specific in e3fa9491d0fd4190a921cf44044d37c85926561b

sig/datadog/core/process_discovery.rbs
vpellan: Fixed in e3fa9491d0fd4190a921cf44044d37c85926561b

ext/libdatadog_api/process_discovery.c
```diff
@@ -0,0 +1,94 @@
...
+  .function = {
+    .dfree = RUBY_DEFAULT_FREE,
+    .dsize = NULL,
+  },
+  .flags = RUBY_TYPED_FREE_IMMEDIATELY
+};
+
+void process_discovery_init(VALUE core_module) {
+  VALUE process_discovery_class = rb_define_class_under(core_module, "ProcessDiscovery", rb_cObject);
+  rb_define_class_under(process_discovery_class, "TracerMemfd", rb_cObject);
```
ivoanjo, line 22: You probably want to do `rb_undef_alloc_func(...)` on this new class as in https://github.com/DataDog/dd-trace-rb/blob/master/ext/datadog_profiling_native_extension/encoded_profile.c#L16 . This makes it so that you cannot call `TracerMemfd.new` from Ruby code, and only the native code can actually create new instances

ext/libdatadog_api/process_discovery.c
```diff
@@ -0,0 +1,94 @@
...
+  if (result.tag == DDOG_RESULT_TRACER_MEMFD_HANDLE_ERR_TRACER_MEMFD_HANDLE) {
+    rb_raise(rb_eRuntimeError, "Failed to store the tracer configuration in a memory file descriptor: %"PRIsVALUE, get_error_details_and_drop(&result.err));
+  }
+
+  // &result.ok is a ddog_TracerMemfdHandle, which is a struct only containing int fd, which is a file descriptor
+  // We should just return the fd
+  int *fd = malloc(sizeof(int));
+  if (!fd) {
+    rb_raise(rb_eRuntimeError, "Failed to allocate memory for file descriptor");
+  }
```
ivoanjo, line 71: You can allocate memory through Ruby:

```suggestion
  // &result.ok is a ddog_TracerMemfdHandle, which is a struct only containing int fd, which is a file descriptor
  // We should just return the fd
  int *fd = ruby_xmalloc(sizeof(int));
```

...this has the advantage that Ruby will check for you if memory fails to allocate AND ruby will also try to release some memory if allocation fails, so it's even nicer.

lib/datadog/core/process_discovery.rb
```diff
@@ -0,0 +1,28 @@
...
+  module Core
+    # Class used to store tracer metadata in a native file descriptor.
+    class ProcessDiscovery
+      def self.get_and_store_metadata(settings)
+        metadata = get_metadata(settings)
+        _native_store_tracer_metadata(**metadata)
+      end
+
+      # According to RFC, runtime_id, service_name, service_env, service_version are optional.
+      # ddcommon memfd_create exposer replace empty strings in these fields by None.
```
ivoanjo, line 13: Minor: The english on this one sounds a bit weird? Maybe worth running through your friendly AI for some tech writing touch-up

ivoanjo (commented):
Left a few more notes, looks very reasonable!

lib/datadog/core/process_discovery.rb
vpellan: Reworded in 9aea8f4c104c3c521568fb988422966e6bfaf9d2

ext/libdatadog_api/process_discovery.c
```diff
@@ -0,0 +1,94 @@
...
+
+  int close_result = close(*fd);
+  free(fd);
+
+  if (close_result == -1) {
+    rb_raise(rb_eRuntimeError, "Failed to close the tracer configuration memory file descriptor: %s", strerror(errno));
+  }
+
+  return Qnil;
+}
```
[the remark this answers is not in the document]
vpellan: I replaced free(fd) with an assignment to -1, and replaced all the raises by call to logger in 9aea8f4c104c3c521568fb988422966e6bfaf9d2

ext/libdatadog_api/process_discovery.c
vpellan: Added in 9aea8f4c104c3c521568fb988422966e6bfaf9d2 ! Thanks !

ext/libdatadog_api/process_discovery.c
```diff
@@ -0,0 +1,115 @@
...
+  }
+
+  // &result.ok is a ddog_TracerMemfdHandle, which is a struct only containing int fd, which is a file descriptor
+  // We should just return the fd
+  int *fd = ruby_xmalloc(sizeof(int));
+
+  *fd = result.ok.fd;
+  VALUE tracer_memfd_class = rb_const_get(self, rb_intern("TracerMemfd"));
+  VALUE tracer_memfd = TypedData_Wrap_Struct(tracer_memfd_class, &tracer_memfd_type, fd);
+  return tracer_memfd;
```
ivoanjo, line 84: (As per my comment on getting this on master) This whole part is going to need to be commented out for now -- I suggest leaving a `rb_raise(..., "TODO: Not in use yet")` behind.

ext/libdatadog_api/process_discovery.c
```diff
@@ -0,0 +1,115 @@
...
+  return INT2NUM(*fd);
+}
+
+static VALUE _native_close_tracer_memfd(DDTRACE_UNUSED VALUE _self, VALUE tracer_memfd) {
+  int *fd;
+  TypedData_Get_Struct(tracer_memfd, int, &tracer_memfd_type, fd);
+  if (fd == NULL) {
+    log_error(rb_sprintf("Failed to get the tracer configuration memory file descriptor: %s", strerror(errno)));
+    return Qnil;
+  }
```
ivoanjo, line 99: I believe this is not needed. In particular, since `ruby_malloc` never returns `NULL`, and we never set this value to `NULL` explicitly, then it's not possible for us to ever observe this being `NULL`. (Unless there's a bug elsewhere... But in that case, I'd maybe `raise` it as something being really wrong, rather than just a regular log... Also note that no `errno` is set if `fd == NULL`, since this is not a system-level error)

```suggestion
```

lib/datadog/core/process_discovery.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Core
+    # Class used to store tracer metadata in a native file descriptor.
+    class ProcessDiscovery
+      def self.get_and_store_metadata(settings, logger)
+        if (libdatadog_api_failure = Datadog::Core::LIBDATADOG_API_FAILURE)
+          logger.debug("Cannot enable crashtracking: #{libdatadog_api_failure}")
```
ivoanjo, line 9:
```suggestion
          logger.debug("Cannot enable process discovery: #{libdatadog_api_failure}")
```

lib/datadog/core/process_discovery.rb
```diff
@@ -0,0 +1,32 @@
...
+          return
+        end
+        metadata = get_metadata(settings)
+        _native_store_tracer_metadata(**metadata)
```
ivoanjo, line 13: I recommend passing in the `logger` as an argument to `_native_store_tracer_metadata`. This makes sure the correct logger is used, and not the fallback one (...logging during components init is a bit awkward right now...).

ext/libdatadog_api/process_discovery.c
```diff
@@ -0,0 +1,115 @@
...
+    char_slice_from_ruby_string(hostname),
+    char_slice_from_ruby_string(service_name),
+    char_slice_from_ruby_string(service_env),
+    char_slice_from_ruby_string(service_version)
+  );
+
+  if (result.tag == DDOG_RESULT_TRACER_MEMFD_HANDLE_ERR_TRACER_MEMFD_HANDLE) {
+    log_error(rb_sprintf("Failed to store the tracer configuration in a memory file descriptor: %"PRIsVALUE, get_error_details_and_drop(&result.err)));
+    return Qnil;
+  }
```
ivoanjo, line 75: So remember that right now this will be visible to customers, since we're logging at warn level. So... Just double checking if this is something we want to show to customers? (Same for the other messages)

spec/support/libdatadog_helpers.rb
```diff
@@ -1,12 +1,12 @@
...
 
-module CrashtrackingHelpers
+module LibdatadogHelpers
   def self.supported?
     # Only works with MRI on Linux
     if PlatformHelpers.mri? && PlatformHelpers.linux?
-      if Datadog::Core::Crashtracking::Component::LIBDATADOG_API_FAILURE
-        raise " does not seem to be available: #{Datadog::Core::Crashtracking::Component::LIBDATADOG_API_FAILURE}. " \
+      if Datadog::Core::LIBDATADOG_API_FAILURE
+        raise " does not seem to be available: #{Datadog::Core::LIBDATADOG_API_FAILURE}. " \
```
ivoanjo, line 9: Minor:

```suggestion
        raise "Libdatadog does not seem to be available: #{Datadog::Core::LIBDATADOG_API_FAILURE}. " \
```

spec/datadog/core/process_discovery_spec.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+require 'msgpack'
+
+require 'spec_helper'
+require 'datadog/core/process_discovery'
+
+RSpec.describe Datadog::Core::ProcessDiscovery do
+  describe '.get_and_store_metadata', skip: !LibdatadogHelpers.supported? do
```
ivoanjo, line 9: I suggest adding an extra test for when `LibdatadogHelpers.supported? == false`, because this may still be a configuration that our customers run into.

E.g. we don't want to accidentally break those customers where for some reason libdatadog is not working.

spec/datadog/core/process_discovery_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+    context 'when libdatadog API is available' do
+      let(:settings) do
+        instance_double(
+          'Datadog::Core::Configuration::Setting',
+          service: 'test-service',
+          env: 'TEST_ENV=true',
```
ivoanjo, line 15: This doesn't look like a very realistic env ... why is there an `=` in here? :eyes:

spec/datadog/core/process_discovery_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+            'runtime_id' => Datadog::Core::Environment::Identity.id,
+            'tracer_language' => Datadog::Core::Environment::Identity.lang,
+            'tracer_version' => Datadog::Core::Environment::Identity.gem_datadog_version_semver2,
+            'hostname' => Datadog::Core::Environment::Socket.hostname,
+            'service_name' => 'test-service',
+            'service_env' => 'TEST_ENV=true',
+            'service_version' => '1.0.0'
+          }
+        )
+      end
```
ivoanjo, line 42: Ooooh I really like this integration test with libdatadog that the correct arguments are passed and then read back. Great job :clap:

spec/datadog/core/process_discovery_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+            tracer_language: Datadog::Core::Environment::Identity.lang,
+            tracer_version: Datadog::Core::Environment::Identity.gem_datadog_version_semver2,
+            hostname: Datadog::Core::Environment::Socket.hostname,
+            service_name: 'test-service',
+            service_env: 'TEST_ENV=true',
+            service_version: '1.0.0'
+          }
+        )
+      end
+    end
```
ivoanjo, line 71: Since `get_metadata` is only needed by `get_and_store_metadata`, I suggest treating it as a private interface and not testing it at all. We already have the stronger test that involves libdatadog as well, so if somehow `get_metadata` introduces a bug, we already have test coverage for it.

We could even make the method `private` in the class.

(If you haven't seen it before -- this talk by Sandi Metz talks about what and when to test things, including internal interfaces: https://www.youtube.com/watch?v=URSWYvyc42M . I really recommend it!)

spec/datadog/core/process_discovery_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+            runtime_id: Datadog::Core::Environment::Identity.id,
+            tracer_language: Datadog::Core::Environment::Identity.lang,
+            tracer_version: Datadog::Core::Environment::Identity.gem_datadog_version_semver2,
+            hostname: Datadog::Core::Environment::Socket.hostname,
+            service_name: '',
+            service_env: '',
+            service_version: ''
+          }
+        )
+      end
```
ivoanjo, line 96: As per my suggestion above, I would move this test to work on `get_and_store_metadata` as well.

spec/datadog/core/process_discovery_spec.rb
```diff
@@ -0,0 +1,99 @@
+# frozen_string_literal: true
+
+require 'msgpack'
+
+require 'spec_helper'
+require 'datadog/core/process_discovery'
+
+RSpec.describe Datadog::Core::ProcessDiscovery do
```
ivoanjo, line 8: (As per my comment on getting this on master) This would need to become an `xdescribe` for now until we can get the libdatadog 17 upgrade.

ivoanjo (approved):
I've been thinking it's a bit sucky to keep this PR open for a long time waiting for the libdatadog release.

So here's an idea: Why not disable a few things, get this PR merged, and then open a separate PR that then re-enables the missing bits?

The latter PR I expect will be very small, and we can get this one out of the way, avoiding any need for future rebases and conflicts.

I've left two comments in places where we'll need to comment out things to make this happen. I think if we tweak/comment out those, we can get a green CI and get this PR merged in.

ext/libdatadog_api/process_discovery.c
```diff
@@ -0,0 +1,115 @@
...
+}
+
+static const rb_data_type_t tracer_memfd_type = {
+  .wrap_struct_name = "Datadog::Core::ProcessDiscovery::TracerMemfd",
+  .function = {
+    .dfree = RUBY_DEFAULT_FREE,
+    .dsize = NULL,
+  },
+  .flags = RUBY_TYPED_FREE_IMMEDIATELY
+};
```
ivoanjo, line 26: Actually I just realized there's an important detail here that should be fixed before we add this in.

Specifically, we should probably move to our own custom free function that does two things: `close(...)` if the fd is still open; and `free(...)` after that. The current approach means that if for some reason the `_native_close_tracer_memfd` doesn't get called (for instance, if there's some issue with some of the other components), then the memory will be freed (good) but the fd will be forever open (not good).

Having it in the free function will guarantee that whatever happens, the fd will also always be freed, which seems like a desirable property to have.

ivoanjo (changes requested):
Left a tiny note that came to mind!

ext/libdatadog_api/process_discovery.c
vpellan: Fixed in 52ab5d24f0ec4170edc0488eceba329a688ac5c9 !

lib/datadog/core/process_discovery.rb
vpellan: Thanks, fixed in 69d5522be4dc000089dee9f96966c0d71e6543b0 !

lib/datadog/core/process_discovery.rb
vpellan: Added in 99f5dd42ee55508a2ac857edd9078e3a6432a9fb !

ext/libdatadog_api/process_discovery.c
vpellan: Changed to debug in 99f5dd42ee55508a2ac857edd9078e3a6432a9fb !

spec/support/libdatadog_helpers.rb
vpellan: Added in 3b9b1b81a836a558a5d2a885b32f45cda2598464 !

spec/datadog/core/process_discovery_spec.rb
vpellan: Added in 3b9b1b81a836a558a5d2a885b32f45cda2598464 !

spec/datadog/core/process_discovery_spec.rb
vpellan: I got confused by thinking it was env vars, fixed in 3b9b1b81a836a558a5d2a885b32f45cda2598464 !

spec/datadog/core/process_discovery_spec.rb
vpellan: Thanks!

spec/datadog/core/process_discovery_spec.rb
vpellan: Set the method as private in 8c984cc7d6c534d1c3959a758a6a74f7e84191bd and deleted corresponding tests !

spec/datadog/core/process_discovery_spec.rb
vpellan: Moved it in 8c984cc7d6c534d1c3959a758a6a74f7e84191bd !

ext/libdatadog_api/process_discovery.c
vpellan: Added a custom free function in 022cfaa9e8ee77e585a3975a665d56363903d011 !

ext/libdatadog_api/process_discovery.c
vpellan: Temporarily disabled process discovery in a10dd5818953b7a48440fdc6b0cfd953d70abcb0 !

spec/datadog/core/process_discovery_spec.rb
vpellan: Temporarily disabled process discovery in a10dd5818953b7a48440fdc6b0cfd953d70abcb0 !

ivoanjo (approved):
:+1: Let's get this in master now! :D

spec/datadog/core/process_discovery_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+require 'spec_helper'
+require 'datadog/core/process_discovery'
+
+# TODO: Re-enable this once we have updated libdatadog to 17.1
+RSpec.xdescribe Datadog::Core::ProcessDiscovery do
+  describe '.get_and_store_metadata' do
+    context 'when libdatadog API is not available' do
+      it 'returns nil' do
+        stub_const('Datadog::Core::LIBDATADOG_API_FAILURE', LoadError.new('test'))
+        expect(described_class.get_and_store_metadata(nil, Datadog::Core::Logger.new($stdout))).to be_nil
```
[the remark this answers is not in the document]
vpellan: Fixed in e6b97029341f0b2a26e03ddedde193073326e3e3

PR #4567, opened 2025-04-08
feat: add record_exception in the datadog api

dubloom:
**What does this PR do?**
This PR adds a record_exception call in the tracer API. It mimics the behavior of the OTEL one for which you can find the documentation here: https://opentelemetry.io/docs/specs/otel/trace/exceptions/

**Motivation:**
Some customers using Sentry complained they were not capable of recording an exception. It appears that if they are using the OTEL api from our tracer, they can. However, as we want to be a competitor to Sentry, I found that it is not a good solution as it is simpler to use this capability directly from our tracer. In addition the feature is easy to emplement 

**Change log entry**
Yes. Record an exception through span events with the `record_exception`call.

**How to test the change?**
- Tests we added in the span_operation spec

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -288,6 +288,15 @@ current_span = Datadog::Tracing.active_span
 current_span.set_tag('my_tag', 'my_value') unless current_span.nil?
 ```
 
+You can record an exception in the current span. It will add the recorded exception as a span event. If you also want to set the span error tags, you should set `escaped` to true.
+```ruby
+# e.g. recording an exception in the active span
+rescue StandardError => e
+  current_span = Datadog::Tracing.active_span
+  current_span.record_exception(e, attributes={"foo":"bar"}, escaped=False)
+end
+```
+
 You can also get the current active trace using the `active_trace` method. This method will return `nil` if there is no active trace.
 
 ```ruby
--- a/lib/datadog/tracing/span_operation.rb
+++ b/lib/datadog/tracing/span_operation.rb
@@ -278,6 +278,38 @@ module Datadog
         set_error_tags(e)
       end
 
+      # Record an exception during the execution of this span by creating a span event.
+      #
+      # If escaped is True, it will also set the span error tags
+      #
+      # @return [void]
+      # @public_api
+      def record_exception(exception, attributes: nil, timestamp: nil, escaped: false)
+        exc = Core::Error.build_from(exception)
+
+        exc_type = exc.type
+        exc_msg = exc.message
+        exc_stacktrace = exc.backtrace
+
+        set_error(exc) if escaped
+
+        attrs = {
+          'exception.type' => exc_type,
+          'exception.message' => exc_msg,
+          'exception.stacktrace' => exc_stacktrace,
+          'escaped' => escaped
+        }
+
+        attrs.merge!(attributes) if attributes
+
+        span_event = SpanEvent.new(
+          :exception,
+          attributes: attrs,
+          time_unix_nano: timestamp || (Time.now.to_r * 1_000_000_000).to_i
+        )
+        @span_events << span_event
+      end
+
       # Return a string representation of the span.
       def to_s
         "SpanOperation(name:#{@name},sid:#{@id},tid:#{@trace_id},pid:#{@parent_id})"
--- a/lib/datadog/tracing/span_operation.rb
+++ b/lib/datadog/tracing/span_operation.rb
@@ -278,12 +278,19 @@ module Datadog
         set_error_tags(e)
       end
 
-      # Record an exception during the execution of this span by creating a span event.
+      # Record an exception during the execution of this span. Multiple exceptions
+      # can be recorded on a span.
+      # If escaped is true, it will set the span error tags
       #
-      # If escaped is True, it will also set the span error tags
+      # @param [Exception] exception The exception to recorded
+      # @param [optional Hash{String => String, Numeric, Boolean, Array<String, Numeric, Boolean>}]
+      #   attributes One or more key:value pairs, where the keys must be
+      #   strings and the values may be (array of) string, boolean or numeric
+      #   type.
+      # @param [optional Integer] timestamp Time when the exception was raised, in nanoseconds since epoch
+      # @param [optional Boolean] escaped Whether the exception escaped the scope of the span
       #
       # @return [void]
-      # @public_api
       def record_exception(exception, attributes: nil, timestamp: nil, escaped: false)
         exc = Core::Error.build_from(exception)
 
--- a/spec/datadog/tracing/span_operation_spec.rb
+++ b/spec/datadog/tracing/span_operation_spec.rb
@@ -995,6 +995,86 @@ RSpec.describe Datadog::Tracing::SpanOperation do
       end
     end
   end
+
+  describe '#record_exception' do
+    subject(:record_exception) do
+      begin
+        raise error
+      rescue => e
+        @error_object = Datadog::Core::Error.build_from(e)
+        span_op.record_exception(
+          e,
+          attributes: attributes,
+          timestamp: timestamp,
+          escaped: escaped
+        )
+      end
+    end
+
+    let(:error) { StandardError.new('test error') }
+    let(:attributes) { nil }
+    let(:timestamp) { nil }
+    let(:escaped) { false }
+    let(:error_object) { @error_object }
+
+    it 'creates a span event with exception details' do
+      record_exception
+
+      expect(span_op.span_events.last).to have_attributes(
+        name: :exception,
+        attributes: {
+          'exception.type' => error_object.type,
+          'exception.message' => error_object.message,
+          'exception.stacktrace' => error_object.backtrace,
+          'escaped' => false
+        }
+      )
+    end
+
+    context 'when escaped is true' do
+      let(:escaped) { true }
+
+      it 'sets error tags on the span' do
+        record_exception
+
+        expect(span_op.status).to eq(Datadog::Tracing::Metadata::Ext::Errors::STATUS)
+        expect(span_op.get_tag(Datadog::Tracing::Metadata::Ext::Errors::TAG_TYPE)).to eq(error_object.type)
+        expect(span_op.get_tag(Datadog::Tracing::Metadata::Ext::Errors::TAG_MSG)).to eq(error_object.message)
+        expect(span_op.get_tag(Datadog::Tracing::Metadata::Ext::Errors::TAG_STACK)).to eq(error_object.backtrace)
+      end
+    end
+
+    context 'with custom attributes' do
+      let(:attributes) { { custom_attr: 'value' } }
+
+      it 'merges custom attributes with exception details' do
+        record_exception
+
+        expect(span_op.span_events.last).to have_attributes(
+          name: :exception,
+          attributes: {
+            'exception.type' => error_object.type,
+            'exception.message' => error_object.message,
+            'exception.stacktrace' => error_object.backtrace,
+            'escaped' => false,
+            'custom_attr' => 'value'
+          }
+        )
+      end
+    end
+
+    context 'with custom timestamp' do
+      let(:timestamp) { 1234567890 }
+
+      it 'uses the provided timestamp' do
+        record_exception
+
+        expect(span_op.span_events.last).to have_attributes(
+          time_unix_nano: 1234567890
+        )
+      end
+    end
+  end
 end
 
 RSpec.describe Datadog::Tracing::SpanOperation::Events do


lib/datadog/tracing/span_operation.rb
```diff
@@ -278,6 +278,45 @@ def set_error(e)
...
+        set_error(exc) if escaped
+
+        attrs = {
+          'exception.type' => exc_type,
+          'exception.message' => exc_msg,
+          'exception.stacktrace' => exc_stacktrace,
+          'escaped' => escaped
+        }
+
+        attrs.merge!(attributes) if attributes
```
TonyCTHsu, line 310: Prefer an empty hash default to additional nil check.

docs/GettingStarted.md
```diff
@@ -288,6 +288,15 @@ current_span = Datadog::Tracing.active_span
 current_span.set_tag('my_tag', 'my_value') unless current_span.nil?
 ```
 
+You can record an exception in the current span. It will add the recorded exception as a span event. If you also want to set the span error tags, you should set `escaped` to true.
```
TonyCTHsu, line 291: The example code is broken due to 

1. It is invoking with positional argument, not keyword arguments
2. Does not handle correctly when `Datadog::Tracing.active_span` returns `nil`
3. `false` instead of `False` (why displaying a method default which could be ignored?)

lib/datadog/tracing/span_operation.rb
```diff
@@ -278,6 +278,45 @@ def set_error(e)
...
       end
 
+      # Record an exception during the execution of this span. Multiple exceptions
+      # can be recorded on a span.
+      # If escaped is true, it will set the span error tags
+      #
+      # @param [Exception] exception The exception to recorded
+      # @param [optional Hash{String => String, Numeric, Boolean, Array<String, Numeric, Boolean>}]
+      #   attributes One or more key:value pairs, where the keys must be
+      #   strings and the values may be (array of) string, boolean or numeric
```
TonyCTHsu, line 288: We should leverage Ruby's dynamic nature and coerce the keys into strings.

spec/datadog/tracing/span_operation_spec.rb
```diff
@@ -995,6 +995,86 @@
       end
     end
   end
+
+  describe '#record_exception' do
+    subject(:record_exception) do
+      begin
+        raise error
+      rescue => e
+        @error_object = Datadog::Core::Error.build_from(e)
```
TonyCTHsu, line 1004: This should already been handled by the method `record_exception`.

TonyCTHsu (changes requested):

lib/datadog/tracing/span_operation.rb
p-datadog: Is it not more efficient (memory allocations) to use nil rather than an empty hash?

lib/datadog/tracing/span_operation.rb
```diff
@@ -278,6 +278,45 @@ def set_error(e)
...
+      #   type.
+      # @param [optional Integer] timestamp Time when the exception was raised, in nanoseconds since epoch
```
p-datadog, line 290: Ruby has a Time type which provides nanosecond precision (https://www.rubydoc.info/stdlib/core/Time). I would expect most Ruby programs to operate on Time objects rather than the time in nanoseconds. What is the reason for using the time in nanoseconds in this method (which is a public interface) rather than Time objects?

lib/datadog/tracing/span_operation.rb
```diff
@@ -278,6 +278,45 @@ def set_error(e)
...
+      # @param [optional Boolean] escaped Whether the exception escaped the scope of the span
```
p-datadog, line 291: Where does the "escaped" terminology come from? It's the first I see this term being used to refer to unhandled exceptions.

To me the standard and significantly more understandable way of phrasing the above sentence is "Whether the exception was handled in the scope of the span".

lib/datadog/tracing/span_operation.rb
```diff
@@ -278,6 +278,45 @@ def set_error(e)
         set_error_tags(e)
       end
 
+      # Record an exception during the execution of this span. Multiple exceptions
+      # can be recorded on a span.
+      # If escaped is true, it will set the span error tags
+      #
+      # @param [Exception] exception The exception to recorded
+      # @param [optional Hash{String => String, Numeric, Boolean, Array<String, Numeric, Boolean>}]
```
p-datadog, line 286: The method does not do any type checking of the attributes, therefore I think either it needs to enforce acceptable types or it should handle all possible types (and for example call `to_s` on rely on JSON serialization, as Tony said).

There should also be test coverage for prohibited value types, if any are actually prohibited.

spec/datadog/tracing/span_operation_spec.rb
dubloom: @error_object is not used in `record_exception`. I did that because this is the only way of having the exact same stacktrace as the span_vent.

docs/GettingStarted.md
dubloom: Yes, it is an opentelemetry term. It is implemented in Python for unhandled exceptions.
TL;DR: it is not in ruby so I'll just remove it

lib/datadog/tracing/span_operation.rb
dubloom: The term was from OTEL but it was deprecated (I cannot find the doc on it) so as I say in an earlier I will remove it.

lib/datadog/tracing/span_operation.rb
```diff
@@ -278,6 +278,45 @@ def set_error(e)
...
+        exc_msg = exc.message
+        exc_stacktrace = exc.backtrace
+
+        set_error(exc) if escaped
+
+        attrs = {
+          'exception.type' => exc_type,
+          'exception.message' => exc_msg,
+          'exception.stacktrace' => exc_stacktrace,
+          'escaped' => escaped
```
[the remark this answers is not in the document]
dubloom: https://opentelemetry.io/docs/specs/otel/trace/exceptions/

This doc tells the name of the attribute ('exception.type'...). I don't think we can have symbol with a '.', can we ?

lib/datadog/tracing/span_operation.rb
```diff
@@ -278,6 +278,45 @@ def set_error(e)
...
+        }
+
+        attrs.merge!(attributes) if attributes
+
+        span_event = SpanEvent.new(
+          'exception',
+          attributes: attrs,
+          time_unix_nano: timestamp || (Time.now.to_r * 1_000_000_000).to_i
```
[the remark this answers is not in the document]
dubloom: This code was extracted from `lib/datadog/tracing/span_event.rb`. There was a problem with deduplication but if the code itself is wrong please let me know because it means the code of span_event is also wrong.

lib/datadog/tracing/span_operation.rb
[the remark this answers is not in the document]
p-datadog: Ruby permits dots in symbol names (along with a bunch of other symbols, possibly any valid string in fact).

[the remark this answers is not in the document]
p-datadog: The syntax for this is `:"foo.bar"`.

docs/GettingStarted.md
buraizu, line 291:
```suggestion
You can record an exception in the current span. It adds the recorded exception as a span event. If you also want to set the span error tags, set `escaped` to true.
```

docs/GettingStarted.md
```diff
@@ -288,6 +288,15 @@ current_span = Datadog::Tracing.active_span
...
+```ruby
+# e.g. recording an exception in the active span
```
buraizu, line 293:
```suggestion
# Example recording an exception in the active span
```

buraizu (approved):
Approving with a couple of minor doc updates requested

spec/datadog/tracing/span_operation_spec.rb
```diff
@@ -995,6 +995,86 @@
...
+          escaped: escaped
+        )
+      end
+    end
+
+    let(:error) { StandardError.new('test error') }
+    let(:attributes) { nil }
+    let(:timestamp) { nil }
+    let(:escaped) { false }
+    let(:error_object) { @error_object }
```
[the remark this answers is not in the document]
TonyCTHsu: Instance variable assignment should be avoided in tests

spec/datadog/tracing/span_operation_spec.rb
TonyCTHsu: You should be able to provide the backtrace during exception initialization with `#set_backtrace` 

```
let(:error) { StandardError.new('test error') }
```

spec/datadog/tracing/span_operation_spec.rb
```diff
@@ -995,6 +995,86 @@
...
+
+    it 'creates a span event with exception details' do
+      record_exception
+
+      expect(span_op.span_events.last).to have_attributes(
+        name: 'exception',
+        attributes: {
+          'exception.type' => error_object.type,
```
TonyCTHsu, line 1026: Instead of delegating to `error_object`, prefer the exact value assertion in test. For example :

```
"exception.type" => "StandardError",  "exception.message" => "test error", 
 ```

File: lib/datadog/tracing/span_operation.rb [hunk 13]
```diff
@@ -278,6 +278,45 @@ def set_error(e)
...
+      #
+      # @return [void]
+      def record_exception(exception, attributes: nil, timestamp: nil, escaped: false)
+        exc = Core::Error.build_from(exception)
+
+        exc_type = exc.type
+        exc_msg = exc.message
+        exc_stacktrace = exc.backtrace
+
+        set_error(exc) if escaped
```
TonyCTHsu, line 301: `set_error` has been marked as deprecated. Use `set_error_tags` instead.

docs/GettingStarted.md
```diff
@@ -282,16 +282,27 @@ end
...
 
 current_span = Datadog::Tracing.active_span
 current_span.set_tag('my_tag', 'my_value') unless current_span.nil?
 ```
 
+You can record an exception in the current span. It adds the recorded exception as a span event.
+```ruby
+# Example: recording an exception in the active span
+rescue => e
+  current_span = Datadog::Tracing.active_span
```
TonyCTHsu, line 295: Prefer safe navigator operation to nilcheck

TonyCTHsu (approved):

docs/GettingStarted.md
```diff
@@ -282,16 +282,27 @@ end
...
+  if !current_span.nil?
```
dubloom, line 296:
```suggestion
  if current_span
```

docs/GettingStarted.md
```diff
@@ -282,16 +282,27 @@ end
...
 ```
 
+You can record an exception in the current span. It adds the recorded exception as a span event.
+```ruby
+# Example: recording an exception in the active span
+rescue => e
+  current_span = Datadog::Tracing.active_span
+  if current_span
+    current_span.record_exception(e, attributes: { foo: "bar" })
+  end
```
dubloom, line 298:
```suggestion
  current_span?.record_exception(e, attributes: { foo: "bar" })
```

spec/datadog/tracing/span_operation_spec.rb
```diff
@@ -995,6 +995,81 @@
...
+        raise error
+      rescue => e
+        span_op.record_exception(
+          e,
+          attributes: {
+            array_attr: [1, 2, 3],
+            str_attr: 'value',
+            int_attr: 0,
+            bool_attr: false,
+            hash_attr: { key: 'value', nested: { foo: 'bar' } },
```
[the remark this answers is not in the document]
dubloom: Span_event only accepts str, int, bool and array. Any other types is removed.

lib/datadog/tracing/span_operation.rb
```diff
@@ -278,6 +278,34 @@ def set_error(e)
...
+      # can be recorded on a span.
+      #
+      # @param [Exception] exception The exception to record
+      # @param [optional Hash{String => String, Numeric, Boolean, Array<String, Numeric, Boolean>}]
+      #   attributes One or more key:value pairs, where the keys must be
+      #   strings and the values may be (array of) string, boolean or numeric
+      #   type.
+      #
+      # @return [void]
+      def record_exception(exception, attributes: {})
```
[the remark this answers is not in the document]
dubloom: Nice catch, thank you.

lib/datadog/tracing/span_operation.rb
```diff
@@ -278,6 +278,34 @@ def set_error(e)
...
+        exc = Core::Error.build_from(exception)
+
+        event_attributes = {
+          'exception.type' => exc.type,
+          'exception.message' => exc.message,
+          'exception.stacktrace' => exc.backtrace,
+        }
```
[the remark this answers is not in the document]
p-datadog: OK, `span_event.rbs` defines the attribute types:
```
      type attributes = Hash[String,attributeValue]
      type attributeValue = String | Integer | Float | bool | Array[String] | Array[Integer] | Array[Float] | Array[bool]
```
The `record_exception` method should be added to the type definition for SpanOperation referencing these types for `attributes`.

spec/datadog/tracing/span_operation_spec.rb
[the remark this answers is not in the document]
p-datadog: OK. This example's description is "supports non-primitive types". Removal of some of the provided values is pretty much opposite behavior of what the test claims to test. Therefore I think there needs to be another test which can be similar to the following test from span_events_spec.rb which asserts that unsupported values are removed:

```
      context 'with invalid attributes' do
        let(:attributes) do
          {
            'int' => 1,
            'invalid_arr1' => [1, 'a'],
            'invalid_arr2' => [[1]],
            'invalid_int1' => 2 << 65,
            'invalid_int2' => -2 << 65,
            'invalid_float1' => Float::NAN,
            'invalid_float2' => Float::INFINITY,
            'string' => 'bar',
          }
        end

        it 'skips invalid values' do
          expect(Datadog.logger).to receive(:warn).with(/Attribute invalid_.*/).exactly(6).times

          expect(span_event.attributes).to eq('int' => 1, 'string' => 'bar',)
        end
      end
```

Note also that the span event test above asserts a log message communicating the removed values, which should probably also be asserted in the test for record_exception.

Lastly, I would say the removal of values should be documented. The existing SpanEvent class does not document that it removes some of the attribute values.

PR #4570, opened 2025-04-09
[PROF-11524] Refactor: Avoid reading profile start_time + Remove indirections during reporting

ivoanjo:
**What does this PR do?**

This PR changes the `StackRecorder` so that instead of relying on libdatadog to track a profile's `start_time` (by passing `NULL` -- aka "use current timestamp") it now explicitly gets the current timestamp and passes it along to libdatadog.

This then means that we no longer need to read the `start_time` from libdatadog, because we're tracking it on our side.

**Motivation:**

Starting with libdatadog 17, we'll no longer be able to read the `start_time` from libdatadog, and this PR leaves us ready for that change.

**Change log entry**

None.

**Additional Notes:**

As I "was in the neighbourhood" I also did a few cleanups on the `HttpTransport` class which will also make it easier to upgrade to libdatadog 17.

**How to test the change?**

We had some existing test coverage, and I've added a bit more; I've also injected some bugs locally on purpose to validate that the test coverage is good. You can try doing the same ;)

### DIFF
--- a/benchmarks/run_all.sh
+++ b/benchmarks/run_all.sh
@@ -11,6 +11,7 @@ for file in \
   `dirname "$0"`/profiler_allocation.rb \
   `dirname "$0"`/profiler_gc.rb \
   `dirname "$0"`/profiler_hold_resume_interruptions.rb \
+  `dirname "$0"`/profiler_http_transport.rb \
   `dirname "$0"`/profiler_memory_sample_serialize.rb \
   `dirname "$0"`/profiler_sample_loop_v2.rb \
   `dirname "$0"`/profiler_sample_serialize.rb \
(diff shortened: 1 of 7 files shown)


benchmarks/run_all.sh
```diff
@@ -11,6 +11,7 @@ for file in \
   `dirname "$0"`/profiler_allocation.rb \
   `dirname "$0"`/profiler_gc.rb \
   `dirname "$0"`/profiler_hold_resume_interruptions.rb \
+  `dirname "$0"`/profiler_http_transport.rb \
```
ivoanjo, line 14: This was temporarily removed in https://github.com/DataDog/dd-trace-rb/pull/4569 and I'm re-adding it here. TL;DR whenever a benchmark uses an API that changes our current setup breaks so the benchmark needs to be disabled and then re-enabled.

Strech (approved):
LGTM

ivoanjo:
Thanks folks for the review! I've opened https://github.com/DataDog/dd-trace-rb/pull/4572 for the flaky test -- it's not related to/introduced in this PR.

PR #4571, opened 2025-04-09
Use Core::Utils::Time.now instead of Time.now

p-datadog:
**What does this PR do?**

Changes existing `Time.now` calls in `lib` to instead use `Core::Utils::Time.now`.

**Motivation:**

The time provider is overridable by users but not all code uses the time provider mechanism. I think everything under `lib` should use the time provider except for the one location where the time is retrieved at class definition.

**Change log entry**
Yes: Use time provider consistently

**Additional Notes:**

**How to test the change?**
Existing tests

### DIFF
--- a/lib/datadog/core/diagnostics/environment_logger.rb
+++ b/lib/datadog/core/diagnostics/environment_logger.rb
@@ -79,7 +79,7 @@ module Datadog
 
           # @return [String] current time in ISO8601 format
           def date
-            Time.now.utc.iso8601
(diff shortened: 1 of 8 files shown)


dubloom (approved):
Thank you !

Strech (approved):
LGTM

_P.S The only thing I might suggest is to cut the namespace since it's already in the `Core` itself. Not a blocker tho_

TonyCTHsu (approved):

PR #4578, opened 2025-04-11
Add reporting of incompatible FFI gem version to Telemetry

y9v:
**What does this PR do?**
This PR reports when an older version of `ffi` gem is used with Ruby version > 3.3 via Telemetry.

**Motivation:**
This combination of older `ffi` gem and Ruby >= 3.3 causes a known memory leak. We want to be notified about this via Telemetry.

**Change log entry**

**Additional Notes:**
None.

**How to test the change?**
CI is enough.

### DIFF
--- a/lib/datadog/appsec/component.rb
+++ b/lib/datadog/appsec/component.rb
@@ -12,7 +12,18 @@ module Datadog
       class << self
         def build_appsec_component(settings, telemetry:)
           return if !settings.respond_to?(:appsec) || !settings.appsec.enabled
-          return if incompatible_ffi_version?
+
+          if incompatible_ffi_version?
+            Datadog.logger.warn(
+              'AppSec is not supported in Ruby versions above 3.3.0 when using `ffi` versions older than 1.16.0, ' \
+              'and will be forcibly disabled due to a memory leak in `ffi`. ' \
+              'Please upgrade your `ffi` version to 1.16.0 or higher.'
+            )
+
+            telemetry.report(StandardError.new, description: 'incompatible ffi version')
+
+            return
+          end
 
           processor = create_processor(settings, telemetry)
 
@@ -31,18 +42,9 @@ module Datadog
 
         def incompatible_ffi_version?
           ffi_version = Gem.loaded_specs['ffi'] && Gem.loaded_specs['ffi'].version
-          return true unless ffi_version
+          return false unless ffi_version
 
-          return false unless Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.3') &&
-            ffi_version < Gem::Version.new('1.16.0')
-
-          Datadog.logger.warn(
-            'AppSec is not supported in Ruby versions above 3.3.0 when using `ffi` versions older than 1.16.0, ' \
-            'and will be forcibly disabled due to a memory leak in `ffi`. ' \
-            'Please upgrade your `ffi` version to 1.16.0 or higher.'
-          )
-
-          true
+          Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.3') && ffi_version < Gem::Version.new('1.16.0')
         end
 
         def create_processor(settings, telemetry)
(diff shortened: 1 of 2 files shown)


lib/datadog/appsec/component.rb
```diff
@@ -12,7 +12,18 @@ class Component
...
-          return if incompatible_ffi_version?
+
+          if incompatible_ffi_version?
+            Datadog.logger.warn(
+              'AppSec is not supported in Ruby versions above 3.3.0 when using `ffi` versions older than 1.16.0, ' \
+              'and will be forcibly disabled due to a memory leak in `ffi`. ' \
+              'Please upgrade your `ffi` version to 1.16.0 or higher.'
+            )
+
+            telemetry.report(StandardError.new, description: 'incompatible ffi version')
```
Strech, line 23: Do you think it's worth generating stack trace here and all around for what is fine to be just a log line?

spec/datadog/appsec/component_spec.rb
```diff
@@ -38,9 +39,10 @@
 
         it 'returns a Datadog::AppSec::Component instance with a nil processor and does not warn' do
```
Strech, line 40: I think it's not matching anymore?

lib/datadog/appsec/component.rb
y9v: good question - we probably don't really need to. The only reason I had in mind was to have this error in telemetry, so we could debug customer issues quicker. But the error is not really actionable on our side.
What do you think, should I close this PR?

spec/datadog/appsec/component_spec.rb
y9v: thanks! fixed [6b70f5e](https://github.com/DataDog/dd-trace-rb/pull/4578/commits/6b70f5e6489ba3b16000e79282b22047d4dd4519)

lib/datadog/appsec/component.rb
p-datadog: Telemetry is not appsec-specific, right? If so maybe description should be more specific, i.e. "incompatible ffi version - appsec disabled" ?

p-datadog: I do that with DI (`di:` prefix).

Strech: I prefix all telemetry reports with `AppSec: Something starting capital letter` and I think there is no reason to report, but rather message. We will do extract work to extract backtrace for nothing.

spec/datadog/appsec/component_spec.rb
```diff
@@ -36,11 +37,12 @@
...
         before { allow(Gem).to receive(:loaded_specs).and_return({}) }
 
-        it 'returns a Datadog::AppSec::Component instance with a nil processor and does not warn' do
+        it 'returns a Datadog::AppSec::Component with a processor instance and does not warn' do
           expect(Datadog.logger).not_to receive(:warn)
+          expect(telemetry).not_to receive(:report)
 
           component = described_class.build_appsec_component(settings, telemetry: telemetry)
-          expect(component).to be_nil
+          expect(component.processor).to be_a(Datadog::AppSec::Processor)
```
Strech, line 45: Wait a second. When we can't load FFI we can't operate WAF, and it should not initialize processor because it will crash later on. Is this intended? PR says only about telemetry, but here is the whole logic change

y9v: you are right, I changed the test without thinking much about it. I will change this back

lib/datadog/appsec/component.rb
```diff
@@ -12,7 +12,27 @@ class Component
       class << self
         def build_appsec_component(settings, telemetry:)
           return if !settings.respond_to?(:appsec) || !settings.appsec.enabled
-          return if incompatible_ffi_version?
+
+          ffi_version = Gem.loaded_specs['ffi'] && Gem.loaded_specs['ffi'].version
+          unless ffi_version
+            Datadog.logger.warn('FFI gem is not loaded, AppSec will be disabled.')
+
+            telemetry.report(StandardError.new, description: 'AppSec: ffi gem not loaded')
```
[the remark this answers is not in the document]
y9v: you are right, and I removed the unnecessary `StandardError.new`, since we can just use `telemetry.error` as it seems - it just takes a string.

[e2f6267](https://github.com/DataDog/dd-trace-rb/pull/4578/commits/e2f6267910c4e5b5880c81a0634982fb377b4d81)

PR #4597, opened 2025-04-22
Sample propagate event

ZStriker19:
**What does this PR do?**

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**


lib/datadog/tracing/tracer.rb
```diff
@@ -376,7 +376,12 @@ def bind_trace_events!(trace_op)
           event_span_op.service ||= @default_service
         end
 
+        events.trace_propagated.subscribe do |event_trace_op|
+          sample_trace(event_trace_op)
+        end
+
         events.span_finished.subscribe do |event_span, event_trace_op|
+          sample_trace(trace_op)
```
marcotc, line 384:
```suggestion
          sample_trace(trace_op) if trace_op.sampling_priority.nil?
```

lib/datadog/tracing/tracer.rb
```diff
@@ -376,7 +376,12 @@ def bind_trace_events!(trace_op)
           event_span_op.service ||= @default_service
         end
 
+        events.trace_propagated.subscribe do |event_trace_op|
+          sample_trace(event_trace_op)
+        end
+
         events.span_finished.subscribe do |event_span, event_trace_op|
+          sample_trace(trace_op) if trace_op.sampling_priority.nil?
```
marcotc, line 384:
```suggestion
          sample_trace(trace_op) if event_trace_op.sampling_priority.nil?
```

marcotc (approved):

p-datadog (changes requested):
@ZStriker19 can you please fill out the PR template?

lib/datadog/tracing/trace_operation.rb
```diff
@@ -79,7 +79,7 @@ def initialize(
         trace_state: nil,
         trace_state_unknown_fields: nil,
         remote_parent: false,
-        tracer: nil,
+        tracer: nil, # DEV-3.0: deprecated, remove in 3.0
```
p-datadog, line 82: If this argument is expected to be used by customers, I would think there should be a suitable changelog entry added to the PR description saying either what the replacement usage pattern should be or what no longer needs to be done.

Alternatively, if there is no expected customer usage of this argument, can it be completely removed now?

marcotc: It's technically a breaking change to remove it, just as a very minor technicality.
The tracer will continue to work regardless if user provides an argument for `tracer` after this change, but if we remove it the user application could start raising `unknown keyword: :tracer (ArgumentError)`.

marcotc:
> Are all of the changes in this PR adequately covered by tests? It seems to me that _some_ usage or behavior should have changed somewhere but there aren't _any_ test changes.

@p-datadog the change to the seemly untested file, `lib/datadog/tracing/tracer.rb`, is a strict refactor. We have plenty of tests in `tracer_spec.rb` for it, but they all just continue to pass.

PR #4598, opened 2025-04-23
[NO-TICKET] Minor: Make sure `datadog_ruby_common.h/.c` stay in sync

ivoanjo:
**What does this PR do?**

As per the comment we have on `datadog_ruby_common.h/.c`, to avoid having too much of a weird setup in our build, we have a copy of these files for profiling and libdatadog_api.

BUT rather than just having a comment saying "please remember to keep these up-to-date", let's add a test to make sure that happens.

**Motivation:**

As I was reviewing a PR and was about to add a comment saying "please keep these files in sync" I decided I would write some code instead of just leaving a comment.

**Change log entry**

None.

**Additional Notes:**

Suggestions are welcome on how to better solve this without having a too-weird `extconf.rb` setup.

**How to test the change?**

I've tried changing the files on only one of the folders and validated the test correctly failed.

### DIFF
--- a/spec/datadog/core/datadog_ruby_common_spec.rb
+++ b/spec/datadog/core/datadog_ruby_common_spec.rb
@@ -1,0 +1,15 @@
+RSpec.describe 'datadog_ruby_common helpers' do
+  let(:profiling_folder) { File.join(__dir__, '../../../ext/datadog_profiling_native_extension') }
+  let(:libdatadog_api_folder) { File.join(__dir__, '../../../ext/libdatadog_api') }
+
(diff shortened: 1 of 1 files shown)


spec/datadog/core/datadog_ruby_common_spec.rb
```diff
@@ -0,0 +1,15 @@
+RSpec.describe 'datadog_ruby_common helpers' do
```
Strech, line 1: Not a big deal

```suggestion
# frozen_string_literal: true

RSpec.describe 'datadog_ruby_common helpers' do
```

Strech (approved):

PR #4599, opened 2025-04-23
[Bug] declaring the missing EMPTY_HASH

nvh0412:
**What does this PR do?**

This pull request introduces a fix to the `Karafka` integration for Datadog tracing. The `EMPTY_HASH` error is happening intermittently in some of my projects due to inconsistent loaders. In some Ruby projects, we used declarative require, hence they have `EMPTY_HASH`, but some don’t, e.g Rails Railtie, which would load `karafka.rb` file before the EMPTY_HASH is available.

**Motivation:**

Despite there are some workaround solutions, e.g monkey patch our datadog gem and define the `EMPTY_HASH`, or just define it somewhere globally, but eventually, this error should be fixed asap.

**Change log entry**

Yes. Tracing: Fix Karafka contrib monitor to avoid referencing Karafka-specific constants that may raise errors due to load order. (updated by @Strech)

**How to test the change?**

### Before

Screenshot 2025-04-23 at 4 27 02 PM

### After

Screenshot 2025-04-23 at 4 53 33 PM

### DIFF
--- a/lib/datadog/tracing/contrib/karafka/monitor.rb
+++ b/lib/datadog/tracing/contrib/karafka/monitor.rb
@@ -15,9 +15,7 @@ module Datadog
             worker.processed
           ].freeze
 
-          EMPTY_HASH = {}.freeze
-
-          def instrument(event_id, payload = EMPTY_HASH, &block)
+          def instrument(event_id, payload = {}, &block)
             return super unless TRACEABLE_EVENTS.include?(event_id)
 
             Datadog::Tracing.trace(Ext::SPAN_WORKER_PROCESS) do |span|
--- a/lib/datadog/tracing/contrib/karafka/monitor.rb
+++ b/lib/datadog/tracing/contrib/karafka/monitor.rb
@@ -15,6 +15,8 @@ module Datadog
             worker.processed
           ].freeze
 
+          EMPTY_HASH = {}.freeze
+
           def instrument(event_id, payload = EMPTY_HASH, &block)
             return super unless TRACEABLE_EVENTS.include?(event_id)
 
--- a/spec/datadog/tracing/contrib/karafka/monitor_spec.rb
+++ b/spec/datadog/tracing/contrib/karafka/monitor_spec.rb
@@ -1,14 +1,11 @@
 # frozen_string_literal: true
 
 require 'datadog/tracing/contrib/support/spec_helper'
-require 'datadog/tracing/contrib/analytics_examples'
 
 require 'karafka'
 require 'datadog'
 
 RSpec.describe 'Karafka monitor' do
-  subject(:monitor) { described_class.new }
-
   before do
     Datadog.configure do |c|
       c.tracing.instrument :karafka, { distributed_tracing: true }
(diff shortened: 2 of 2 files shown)


spec/datadog/tracing/contrib/karafka/monitor_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+        )
+
+        expect(Datadog::Tracing).to have_received(:trace)
+        expect(span).to have_received(:resource=).with('Consumer#consume')
+        expect(span).to have_received(:set_tag).with(Datadog::Tracing::Contrib::Karafka::Ext::TAG_MESSAGE_COUNT, 1)
+        expect(span).to have_received(:set_tag).with(Datadog::Tracing::Contrib::Karafka::Ext::TAG_PARTITION, 1)
+        expect(span).to have_received(:set_tag).with(Datadog::Tracing::Contrib::Karafka::Ext::TAG_OFFSET, 42)
+        expect(span).to have_received(:set_tag).with(Datadog::Tracing::Contrib::Karafka::Ext::TAG_CONSUMER, 'Consumer')
+        expect(span).to have_received(:set_tag).with(Datadog::Tracing::Contrib::Ext::Messaging::TAG_DESTINATION, 'topic_name')
+        expect(span).to have_received(:set_tag).with(Datadog::Tracing::Contrib::Ext::Messaging::TAG_SYSTEM, Datadog::Tracing::Contrib::Karafka::Ext::TAG_SYSTEM)
```
Strech, line 62: This type of test is very fragile and rely on internal implementation, which makes it hard to refactor. I would suggest to go with [TraceOperation](https://github.com/DataDog/dd-trace-rb/blob/master/spec/datadog/appsec/event_spec.rb#L54)

And then [use span metadata](https://github.com/DataDog/dd-trace-rb/blob/master/spec/datadog/appsec/event_spec.rb#L165) to check the outcome

spec/datadog/tracing/contrib/karafka/monitor_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+        allow(Datadog::Tracing).to receive(:trace)
+
+        Karafka.monitor.instrument('worker.completed')
+
+        expect(Datadog::Tracing).not_to have_received(:trace)
+      end
+    end
+
+    context 'when the event is traceable' do
+      let(:span) { instance_double('Span', set_tag: nil) }
```
Strech, line 36: There is a real class behind Span and instance_double will not help here to avoid stubbing.

It's `Datadog::Tracing::SpanOperation`. But I would recommend to follow suggestion below to avoid a tight coupling with implementation.

lib/datadog/tracing/contrib/karafka/monitor.rb
```diff
@@ -15,6 +15,8 @@ module Monitor
             worker.processed
           ].freeze
 
+          EMPTY_HASH = {}.freeze
+
           def instrument(event_id, payload = EMPTY_HASH, &block)
```
Strech, line 20: I would also recommend to rename it into `EMPTY_PAYLOAD` to avoid using structure in the name to have some encapsulation of the implementation details.

```suggestion
          EMPTY_PAYLOAD = {}.freeze

          def instrument(event_id, payload = EMPTY_PAYLOAD, &block)
```

p-datadog:
I haven't checked how the name lookups work exactly to resolve the EMPTY_HASH defined in karafka, but I'm thinking we should either:

1) qualify EMPTY_HASH with the Karafka namespace it's defined in, or 
2) use a literal {}

Defining EMPTY_HASH which then aliases/shadows the karafka one I think is not a great idea, and making a constant that is used exactly once also does not strike me as being valuable.

nvh0412:
Hey @Strech @p-datadog, thanks for your feedback, I agree that a fronzen EMPTY_HASH wouldn't add much value here, so I've reverted to using a literal hash, also improved my specs. Cheers.

spec/datadog/tracing/contrib/karafka/monitor_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+RSpec.describe 'Karafka monitor' do
+  subject(:monitor) { described_class.new }
+  let(:configuration_options) { { distributed_tracing: true } }
+
+  before do
+    Datadog.configure do |c|
+      c.tracing.instrument :karafka, configuration_options
+    end
+  end
```
Strech, line 17:
```suggestion

  before do
    Datadog.configure do |c|
      c.tracing.instrument :karafka, { distributed_tracing: true }
    end
  end
```

Strech (approved):
LGTM 👍🏼 

_P.S There is a let with unnecessary primitive which makes sense to inline_

p-datadog:
@nvh0412 Can you please explain this sentence:

> In some Ruby projects, we used declarative require, hence they have EMPTY_HASH, but some don’t, e.g Rails Railtie, which would load karafka.rb file before the EMPTY_HASH is available.

It is my understanding that our integrations expect the third-party library to be loaded, therefore if karafka itself is not loaded and our integration is trying to run, that sounds like a bug that needs to be fixed.

nvh0412:
Hey @p-datadog, apologies, my previous message wasn't accurate.

The thing is, we also have a forked version of this dd-trace-rb, and we didn't encounter this error there because it was still using the previous `Monitor` implementation, before [we changed it to a `module` from a `class` with a proper parent class  in my earlier PR to this upstream repo.](https://github.com/DataDog/dd-trace-rb/pull/4147/commits/ffcef947069c1f793f3a26ccb7e928c8b0ef3fe9)

Some non-Rails Ruby projects in my company use that forked version, which led me to initially think the issue was caused by Karafka’s Railtie affecting the require order, resulting in the inconsistent error. But when I switch to use the upstream version, they've got the same issue, missing the EMPTY_HASH definition.

spec/datadog/tracing/contrib/karafka/monitor_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+    # Reset before and after each example; don't allow global state to linger.
+    Datadog.registry[:karafka].reset_configuration!
+    example.run
+    Datadog.registry[:karafka].reset_configuration!
+  end
+
+  describe '.instrument' do
+    context 'when the event is not traceable' do
+      it 'does not create a trace' do
+        allow(Datadog::Tracing).to receive(:trace)
```
p-datadog, line 28: This is in opposition to the purpose of the test (asserting that a trace is not received), therefore either this line should be removed or there needs to be a comment added explaining it.

spec/datadog/tracing/contrib/karafka/monitor_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+  around do |example|
+    # Reset before and after each example; don't allow global state to linger.
+    Datadog.registry[:karafka].reset_configuration!
+    example.run
+    Datadog.registry[:karafka].reset_configuration!
+  end
+
+  describe '.instrument' do
+    context 'when the event is not traceable' do
+      it 'does not create a trace' do
```
p-datadog, line 27: This test should also assert that `super` is called in some way.

p-datadog (changes requested):

spec/datadog/tracing/contrib/karafka/monitor_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+
+        Karafka.monitor.instrument('worker.completed')
+
+        expect(Datadog::Tracing).not_to have_received(:trace)
+      end
+    end
+
+    context 'when the event is traceable' do
+      let(:payload) { { job: job } }
+      let(:job) { instance_double('Job', class: job_class, executor: executor, messages: messages) }
```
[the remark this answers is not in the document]
nvh0412: Depending on what we want our instrument function to support, proving that the codebase works with a double like this shows that the function only cares about the interface of the Job object, not its concrete class.

However, if the Datadog team prefers to use a concrete class from the Karafka gem here, I think that’s completely fine. 🙂

nvh0412:
After reviewing this PR,  and considering that I already have a forked version with a proper implementation of the Karafka integration, there’s no urgency on our side to fix this issue.

The Datadog team was the one who modified my original implementation in a previous PR, so they should be in a good position to address it properly. I’ll go ahead and close this PR and hope the Datadog team can resolve it in a more robust way than I could. Cheers.

Strech:
Hey @nvh0412 I'm really sorry for the frustrating experience and how this was handled. We think your did a great job and we appreciate your contribution 💜 

I was already trying to run tests on that PR and I will make sure that it will make to master as soon as possible 🚀  

Meanwhile, we will discuss that situation internally and make sure to handle it better next time. I will re-open this PR and if you have any concerns or comments, just let me know.

spec/datadog/tracing/contrib/karafka/monitor_spec.rb
Strech: This was resolved via https://github.com/DataDog/dd-trace-rb/pull/4599/files#diff-b07300902cec88969c7fbc4bd689b42b089892415d9d6f38bf177c308b726fe3R28-R29

p-datadog (approved):

marcotc (approved):

PR #4603, opened 2025-04-24
[APPSEC-57303] Fingerprinting on non-attack non-business events

Strech:
**What does this PR do?**

1. Quick-Fix for AppSec rules files to make use of supplied keys before the refactoring comes
2. New abstraction - `SecurityEvent` which suppose to be used at the end of request lifecycle to tag spans
3. Fix for derivatives for non-attack non-business security events.

**Motivation:**

This work is done to complete feature [RFC-0979 Attacker Fingerprinting](https://docs.google.com/document/d/1DivOa9XsCggmZVzMI57vyxH2_EBJ0-qqIkRHm_sEvSs/edit?pli=1&tab=t.0).

**Change log entry**

No. Will be part of the bundle

**Additional Notes:**

There is still missing abstraction and not well distributed responsibilities for security events of all kind, because we explicitly check security event before adding it to the list. And that complexity of the decision could be hidden.

**How to test the change?**

CI + ST

### DIFF
--- a/.github/forced-tests-list.json
+++ b/.github/forced-tests-list.json
@@ -1,3 +1,6 @@
 {
-  
+    "APPSEC_BLOCKING": [
+        "tests/appsec/test_fingerprinting.py::Test_Fingerprinting_Endpoint_Preprocessor",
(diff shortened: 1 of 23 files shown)


y9v (approved):
awesome 🎉

PR #4604, opened 2025-04-24
feat: add automatic reporting of handled errors

dubloom:
**What does this PR do?**
The PR adds the automatic reporting of handled errors. Whenever an error is handled, it will be attached (if not re-thrown) as a span event to a `span_operation`. By adding a specific tag to the span as well, Error Tracking will be able to fingerprint the span event to report the errors in its product. 
Brief specification can be found on this [link](https://docs.google.com/document/d/13sg08mQH-6QnNT4h0DGVqbq-JF2pwWEjsCIBblUHC_w/edit?usp=sharing) 

**Motivation:**
Error Tracking released Error Tracking Standalone. This allows a user to get all the errors without using tracing. Error Tracking wants a user to be able to get handled and unhandled errors on Error Tracking without any manual instrumentation 

**Change log entry**
Yes. ErrorTracking: Add automatic reporting of handled errors. 

**How to test the change?**
- The feature is extensively tested in `spec/error_tracking/component_spec.rb`
- As the feature is using two global tracepoints, two benchmarks were added:
   - One to test the overhead of the feature alone
   - One to test the overhead of the feature in context.


lib/datadog/core/configuration/settings.rb
```diff
@@ -981,6 +981,20 @@ def initialize(*_)
           end
         end
 
+        settings :errortracking do
```
marcotc, line 984: @TonyCTHsu @p-datadog @ivoanjo @y9v, do we call this `errortracking` or `error_tracking`?

ivoanjo: The customer-facing product name is ["error tracking"](https://www.datadoghq.com/product/error-tracking/) so it seems reasonable to follow that, I think? Unless there's prior art from other datadog libraries that use this nomenclature (and in that case it may be better to align for consistency).

y9v: I would use `error_tracking` - simply because it follows the ruby convention. On the other hand, we already have `crashtracking` which does not follow it.

lib/datadog/core/errortracking/filters.rb
```diff
@@ -0,0 +1,70 @@
...
+
+module Datadog
+  module Core
+    module Errortracking
+      # The filters module is in charge of creating
+      # the filter function called in the handled_exc_tracker.
+      module Filters
+        class << self
+          def get_gem_name(file_path)
+            regex = %r{gems/([^/]+)-\d}
```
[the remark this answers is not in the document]
y9v: we don't handle a custom bundler path here, right?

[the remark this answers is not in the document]
dubloom: Start and end anchors should not be added at what we are looking for is in a middle of the path 

Additionally, custom bundler path will create a gems/ subfolder so it will work too

lib/datadog/core/errortracking/component.rb
```diff
@@ -0,0 +1,122 @@
...
+            @module_path_getter = TracePoint.new(:script_compiled) do |tp|
+              next if tp.eval_script
+
+              path = tp.instruction_sequence.path
+              next if path.nil?
+
+              @modules_to_instrument.each do |module_to_instr|
+                # The regex is looking for the name of the module with '/' before
+                # and either '/' or '.rb' after
+                add_instrumented_file(path) if path.match?(%r{/#{Regexp.escape(module_to_instr)}(?=/|\.rb)})
```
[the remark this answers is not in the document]
dubloom: adressed in 6fb24d99cd8ef0b02ebbf0a2d4636562661644a0

lib/datadog/core/errortracking/component.rb
```diff
@@ -0,0 +1,122 @@
...
+          attributes = {
+            'exception.type' => formatted_exception.type,
+            'exception.message' => formatted_exception.message,
+            'exception.stacktrace' => formatted_exception.backtrace
+          }
+          Datadog::Tracing::SpanEvent.new('exception', attributes: attributes)
+        end
+
+        def add_instrumented_file(file_path)
+          @instrumented_files[file_path] = true
```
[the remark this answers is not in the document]
dubloom: adressed in 6fb24d99cd8ef0b02ebbf0a2d4636562661644a0

Not the tracepoint is created only if `@instrumented_files`is not nil and the method is now private. Therefore, by design, it is not possible to call this method with `@instrumented_files` being `nil`

lib/datadog/core/errortracking/component.rb
```diff
@@ -0,0 +1,122 @@
...
+
+module Datadog
+  module Core
+    module Errortracking
+      # Component for error tracking.
+      #
+      # Only one instance of the Component should ever be active.
+      #
+      # The component instance records every handled exceptions from the configured scopes
+      # (user, third_party packages, specified modules of everything). T
```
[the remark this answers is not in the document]
dubloom: adressed in 6fb24d99cd8ef0b02ebbf0a2d4636562661644a0

lib/datadog/core/errortracking/collector.rb
```diff
@@ -0,0 +1,77 @@
...
+          end
+        end
+
+        def _get_span_events
+          @storage.values
+        end
+
+        def _clear_span_events
+          @storage.clear
+        end
```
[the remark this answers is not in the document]
dubloom: adressed in 6fb24d99cd8ef0b02ebbf0a2d4636562661644a0

lib/datadog/core/errortracking/collector.rb
```diff
@@ -0,0 +1,77 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Core
+    module Errortracking
```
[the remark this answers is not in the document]
dubloom: adressed in 6fb24d99cd8ef0b02ebbf0a2d4636562661644a0

lib/datadog/core/configuration/settings.rb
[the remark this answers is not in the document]
dubloom: partially addressed in 6fb24d99cd8ef0b02ebbf0a2d4636562661644a0 (i will probably make additional changes after when I'm back from PTO)

Rakefile
```diff
@@ -205,6 +205,12 @@ namespace :spec do
   end
   # rubocop:enable Style/MultilineBlockChain
 
+  desc '' # "Explicitly hiding from `rake -T`"
+  RSpec::Core::RakeTask.new(:errortracking) do |t, args|
+    t.pattern = 'spec/datadog/core/errortracking/errortracking_spec.rb'
```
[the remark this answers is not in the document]
dubloom: addressed in 6fb24d99cd8ef0b02ebbf0a2d4636562661644a0

lib/datadog/core/configuration/settings.rb
dubloom: addressed in 6fb24d99cd8ef0b02ebbf0a2d4636562661644a0

lib/datadog/core/configuration/components.rb
```diff
@@ -89,6 +98,7 @@ def build_crashtracker(settings, agent_settings, logger:)
           :telemetry,
           :tracer,
           :crashtracker,
+          :errortracker,
```
marcotc, line 101:
```suggestion
          :error_tracker,
```

lib/datadog/core/configuration/settings.rb
p-datadog: :+1: I prefer `error_tracking`

lib/datadog/core/configuration/settings.rb
```diff
@@ -981,6 +981,33 @@ def initialize(*_)
...
         end
 
+        # Datadog Error Tracking-specific configurations
+        #
+        # @public_api
+        settings :error_tracking do
+          # Enable automatic reporting of handled errors and set the scope
+          # of the errors to report: all | user | third_party
+          #
+          # @default 'DD_ERROR_TRACKING_HANDLED_ERRORS' environement variable, otherwise `nil`
```
p-datadog, line 991:
```suggestion
          # @default 'DD_ERROR_TRACKING_HANDLED_ERRORS' environment variable, otherwise `nil`
```

lib/datadog/core/configuration/settings.rb
```diff
@@ -981,6 +981,33 @@ def initialize(*_)
...
+          # @return [String, nil]
+          option :instrumentation_scope do |o|
```
p-datadog, line 993: What was the reason for not using the same name as the environment variable - `handled_errors` ?

lib/datadog/core/configuration/settings.rb
```diff
@@ -981,6 +981,33 @@ def initialize(*_)
...
+            o.default ''
+            o.env 'DD_ERROR_TRACKING_HANDLED_ERRORS'
+          end
+
+          # Enable automatic reporting of handled errors and set the module
+          # for which handled errors should be reported. List of comma separated modules
+          #
+          # @default 'DD_ERROR_TRACKING_HANDLED_ERRORS_MODULES' environment variable, otherwise `nil`
+          # @return [String, nil]
+          option :modules_to_instrument do |o|
```
p-datadog, line 1004: The environment variable is scoped to handled errors but the option name is not. Is there no way this can cause an issue later if unhandled errors need to be instrumented?

p-datadog (commented):
I think error tracking needs to be a top-level module (moved out of core). Core means functionality used by products, and my understanding is error tracking is a product.

lib/datadog/core/configuration/settings.rb
marcotc, line 993: Can you add a validation here for the 3 valid values, similar to this OpenSearch option:
https://github.com/DataDog/dd-trace-rb/blob/19100f658ce49b14a0413361ff64f62280411058/lib/datadog/tracing/contrib/opensearch/configuration/settings.rb#L54-L64

marcotc: You can then likely remove this check:
```ruby
            return if !settings.error_tracking.instrumentation_scope.empty? &&
              !['all', 'user', 'third_party'].include?(settings.error_tracking.instrumentation_scope)
```

lib/datadog/core/error_tracking/component.rb
```diff
@@ -0,0 +1,124 @@
...
+      # The component instance records every handled exceptions from the configured scopes
+      # (user, third_party packages, specified modules of everything).
+      class Component
+        attr_accessor :handled_exc_tracker,
+          :module_path_getter,
+          :tracer,
+          :modules_to_instrument,
+          :instrumented_files,
+          def self.build(settings, tracer)
+            if settings.error_tracking.instrumentation_scope.empty? && settings.error_tracking.modules_to_instrument.empty?
```
marcotc, line 22: If this is the only check, it can probably be done outside of this class, in the caller, so that this class's initializer is trivia, and only invoked when the values are valid.

lib/datadog/core/error_tracking/component.rb
```diff
@@ -0,0 +1,124 @@
...
+            'exception.message' => formatted_exception.message,
+            'exception.stacktrace' => formatted_exception.backtrace
+          }
+          Datadog::Tracing::SpanEvent.new('exception', attributes: attributes)
+        end
+
+        private
+
+        def _add_instrumented_file(file_path)
+          @instrumented_files[file_path] = true
```
marcotc, line 119: `@instrumented_files` can be a [Set](https://ruby-doc.org/stdlib-2.7.1/libdoc/set/rdoc/Set.html), so we don't have to have a boolean stored that is not used.

marcotc (commented):
Because we are adding a global TracePoint, thus adding overhead to the user application that is proportional to the user code, no Datadog code, we should add some benchmarks to ensure our solution is performant.
You can see examples in the `benchmarks` directory, like: https://github.com/DataDog/dd-trace-rb/blob/master/benchmarks/tracing_trace.rb
You probably want to create a new file, then add it here: https://github.com/DataDog/dd-trace-rb/blob/master/benchmarks/run_all.sh

lib/datadog/core/error_tracking/filters.rb
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Core
+    module ErrorTracking
+      # The filters module is in charge of creating
+      # the filter function called in the handled_exc_tracker.
+      module Filters
```
p-datadog, line 8: Suggest adding `@api private`

lib/datadog/core/error_tracking/component.rb
dubloom: addressed in ecfdce57e4880c87d13c07923e31ef91a92f64e7

lib/datadog/core/error_tracking/component.rb
dubloom: addressed in ecfdce57e4880c87d13c07923e31ef91a92f64e7

lib/datadog/core/configuration/settings.rb
dubloom: addressed in ecfdce57e4880c87d13c07923e31ef91a92f64e7

lib/datadog/core/configuration/settings.rb
dubloom: addressed in ecfdce57e4880c87d13c07923e31ef91a92f64e7

lib/datadog/core/configuration/settings.rb
dubloom: addressed in ecfdce57e4880c87d13c07923e31ef91a92f64e7

lib/datadog/core/configuration/components.rb
dubloom: addressed in ecfdce57e4880c87d13c07923e31ef91a92f64e7

lib/datadog/core/configuration/settings.rb
dubloom: addressed in ecfdce57e4880c87d13c07923e31ef91a92f64e7

lib/datadog/core/error_tracking/collector.rb
```diff
@@ -0,0 +1,77 @@
...
+            span_events = collector.get_span_events
+            if span_events
+              span_op.span_events.concat(span_events)
+              collector.clear_span_events
+            end
+          end
+        end
+
+        def initialize
+          @storage = {}
```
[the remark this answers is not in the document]
dubloom: addressed in https://github.com/DataDog/dd-trace-rb/commit/ecfdce57e4880c87d13c07923e31ef91a92f64e7

docs/GettingStarted.md
```diff
@@ -2766,6 +2766,30 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
...
+`datadog` can automatically report handled errors. The errors are attached through span events to the span in which they are handled. They are also directly reported to Error Tracking.
+
+#### Requirements
+
+- Ruby 2.6+. JRuby and TruffleRuby are not supported.
+- datadog 2.16.0+.
+
+#### Configuration
+
+To enable automatic reporting of handled errors, one of the two environement variables must be set:
```
drichards-87, line 2780:
```suggestion
To enable automatic reporting of handled errors, one of the following two environment variables must be set:
```

docs/GettingStarted.md
```diff
@@ -2766,6 +2766,30 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
...
+and gem names for which handled errors will be reported.  
+The value provided should be:
+  * A file name, for instance `main`: `main.rb` will be instrumented.
+  * A folder name, for instance `mypackage`: every ruby files in folders called `mypackage` will be instrumented.
+  * A gem name: every ruby files of this gem will be instrumented. Be careful, if you specify for instance `rails` and you have a subfolder `rails` in your project, it will also be instrumented.
+  * An absolute path (a path beginning with `/`) For example: `/app/lib/mypackage/main.rb`. If you provide just `/app/lib/mypackage`, every ruby files in the folder will be instrumented.
+  * A path relative to the current directory (a path beginning with `./`). For example, if you execute your program in `/app/`, you can provide `./lib/mypackage/main.rb`. If you provide a path like `./lib/mypackage/`, every ruby files in this folder will be instrumented.
+
+**Important**: In Ruby3.2 and before, `datadog` can only get the location of where the error was thrown and not where it is handled. This can cause the second environment variable to malfunction if the error is handled in another file than the one specified.
```
drichards-87, line 2791:
```suggestion
**Important**:  In Ruby 3.2 and earlier, `datadog` can only capture the location where the error is thrown, not where it is handled. This may cause the second environment variable to malfunction if the error is handled in a different file than the one specified.
```

docs/GettingStarted.md
```diff
@@ -2766,6 +2766,30 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
...
+
+#### Requirements
+
+- Ruby 2.6+. JRuby and TruffleRuby are not supported.
+- datadog 2.16.0+.
+
+#### Configuration
+
+To enable automatic reporting of handled errors, one of the two environement variables must be set:
+- `DD_ERROR_TRACKING_HANDLED_ERRORS` = `user|third_party|all`. Report handled errors of user code, gems or both.
```
drichards-87, line 2781:
```suggestion
- `DD_ERROR_TRACKING_HANDLED_ERRORS` = `user|third_party|all`. Report handled errors from user code, gems, or both.
```

docs/GettingStarted.md
```diff
@@ -2766,6 +2766,30 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
...
+- `DD_ERROR_TRACKING_HANDLED_ERRORS_INCLUDE` = `file, gems, path..`. A list of comma-separated paths, file names
+and gem names for which handled errors will be reported.
```
drichards-87, line 2783:
```suggestion
- `DD_ERROR_TRACKING_HANDLED_ERRORS_INCLUDE` = `file, gems, path..`. A list of comma-separated paths, file names,
and gem names for which handled errors will be reported.  
```

docs/GettingStarted.md
```diff
@@ -2766,6 +2766,30 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
...
+- Ruby 2.6+. JRuby and TruffleRuby are not supported.
+- datadog 2.16.0+.
+
+#### Configuration
+
+To enable automatic reporting of handled errors, one of the two environement variables must be set:
+- `DD_ERROR_TRACKING_HANDLED_ERRORS` = `user|third_party|all`. Report handled errors of user code, gems or both.
+- `DD_ERROR_TRACKING_HANDLED_ERRORS_INCLUDE` = `file, gems, path..`. A list of comma-separated paths, file names
+and gem names for which handled errors will be reported.  
+The value provided should be:
```
drichards-87, line 2784:
```suggestion
The value should be either:
```

docs/GettingStarted.md
```diff
@@ -2766,6 +2766,30 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
...
+  * A file name, for instance `main`: `main.rb` will be instrumented.
```
drichards-87, line 2785:
```suggestion
  * A file name, such as `main`: `main.rb` will be instrumented.
```

docs/GettingStarted.md
```diff
@@ -2766,6 +2766,30 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
...
+  * A folder name, for instance `mypackage`: every ruby files in folders called `mypackage` will be instrumented.
```
drichards-87, line 2786:
```suggestion
  * A folder name, such as `mypackage`: Every Ruby file in folders named `mypackage` will be instrumented.
```

docs/GettingStarted.md
```diff
@@ -2766,6 +2766,30 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
...
+  * A gem name: every ruby files of this gem will be instrumented. Be careful, if you specify for instance `rails` and you have a subfolder `rails` in your project, it will also be instrumented.
```
drichards-87, line 2787:
```suggestion
  * A gem name: Every Ruby file in the gem will be instrumented. Be careful, if you specify, for instance `rails` and you have a subfolder named `rails` in your project, it will also be instrumented.
```

docs/GettingStarted.md
```diff
@@ -2766,6 +2766,30 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
...
+  * An absolute path (a path beginning with `/`) For example: `/app/lib/mypackage/main.rb`. If you provide just `/app/lib/mypackage`, every ruby files in the folder will be instrumented.
```
drichards-87, line 2788:
```suggestion
  * An absolute path (a path beginning with `/`), for example, `/app/lib/mypackage/main.rb`. If you provide only `/app/lib/mypackage`, every Ruby file in that folder will be instrumented.
```

docs/GettingStarted.md
```diff
@@ -2766,6 +2766,30 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
...
+  * A path relative to the current directory (a path beginning with `./`). For example, if you execute your program in `/app/`, you can provide `./lib/mypackage/main.rb`. If you provide a path like `./lib/mypackage/`, every ruby files in this folder will be instrumented.
```
drichards-87, line 2789:
```suggestion
  * A path relative to the current directory (a path beginning with `./`). For example, if you execute your program in `/app/`, you can provide `./lib/mypackage/main.rb`. If you provide a path like `./lib/mypackage/`, every Ruby file in this folder will be instrumented.
```

drichards-87 (approved):
Added some suggestions from Docs and approved the PR.

docs/GettingStarted.md
marcotc, line 2791:
```suggestion
```

I think it's simpler to list this limitation in the description of `DD_ERROR_TRACKING_HANDLED_ERRORS_INCLUDE` itself.

docs/GettingStarted.md
marcotc, line 2783:
```suggestion
- `DD_ERROR_TRACKING_HANDLED_ERRORS_INCLUDE` = `file, gems, path..`. A list of comma-separated paths, file names,
and gem names for which handled errors will be reported.
For Ruby 3.3 or newer, the location where the error was raised and where it was rescued will be matched. For earlier versions of Ruby, only the location where the error was raised can be matched.
```

docs/GettingStarted.md
```diff
@@ -2766,6 +2766,30 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
...
+#### Requirements
+
+- Ruby 2.6+. JRuby and TruffleRuby are not supported.
+- datadog 2.16.0+.
+
+#### Configuration
+
+To enable automatic reporting of handled errors, one of the two environement variables must be set:
+- `DD_ERROR_TRACKING_HANDLED_ERRORS` = `user|third_party|all`. Report handled errors of user code, gems or both.
+- `DD_ERROR_TRACKING_HANDLED_ERRORS_INCLUDE` = `file, gems, path..`. A list of comma-separated paths, file names
```
marcotc, line 2782: Please document the default value of these environment variables, so users know at glance what to expect.

marcotc (approved):
Awesome work, I just left a few documentation suggestions!

lib/datadog/tracing/span_operation.rb
```diff
@@ -246,6 +251,10 @@ def stopped?
         !@end_time.nil?
       end
 
+      def parent?
```
p-datadog, line 254: This is also I think a confusing name - between "is this span operation a parent" and "does this span operation have a parent". Renaming the method to `root?` or `top_level?` I think would be way clearer.

p-datadog (commented):
Overall LGTM

lib/datadog.rb
```diff
@@ -10,4 +10,5 @@
 # Line probes will not work on Ruby < 2.6 because of lack of :script_compiled
 # trace point. Only load DI on supported Ruby versions.
 require_relative 'datadog/di' if RUBY_VERSION >= '2.6' && RUBY_ENGINE != 'jruby'
+require_relative 'datadog/error_tracking' if RUBY_VERSION >= '2.6' && RUBY_ENGINE != 'jruby'
```
p-datadog, line 13: I just ran into a problem with this logic (that you copied from the line just above) in https://github.com/DataDog/dd-trace-rb/pull/4647 - for reporting DI state to telemetry, we need to have at least DI configuration loaded. DI configuration also needs to be loaded if we are to permit customers to, for example, disable it on unsupported versions like Ruby 2.5.

https://github.com/DataDog/dd-trace-rb/pull/4647 will conflict with this PR unfortunately.

If this PR is meant to be tested internally for a bit rather than immediately shipped to customers, I would say that redoing how ET loading is done can be done in a follow-up PR.

lib/datadog/core/configuration/components.rb
```diff
@@ -123,6 +125,7 @@ def initialize(settings)
           @health_metrics = self.class.build_health_metrics(settings, @logger)
           @appsec = Datadog::AppSec::Component.build_appsec_component(settings, telemetry: telemetry)
           @dynamic_instrumentation = Datadog::DI::Component.build(settings, agent_settings, @logger, telemetry: telemetry)
+          @error_tracking = Datadog::ErrorTracking::Component.build(settings, tracer, logger)
```
Strech, line 128: It's better to stick with instance variable here

```suggestion
          @error_tracking = Datadog::ErrorTracking::Component.build(settings, @tracer, @logger)
```

lib/datadog/error_tracking/collector.rb
```diff
@@ -0,0 +1,84 @@
...
+          span_events = collector.get_span_events
+          span_op.span_events.concat(span_events) if span_events
+        end
+      end
+
+      def initialize
+        @span_event_per_error = {}
+      end
+
+      def add_span_event(span_op, error, span_event)
```
Strech, line 34: according to the name span arguments have priority, hence

```suggestion
      def add_span_event(span_op, span_event, error)
```

lib/datadog/error_tracking/collector.rb
```diff
@@ -0,0 +1,84 @@
...
+        # When this is the first time we add a span event for a span,
+        # we suscribe to the :after_stop event
+        if @span_event_per_error.empty?
+          events = span_op.send(:events)
```
Strech, line 38: Is this a private API?

lib/datadog/error_tracking/collector.rb
```diff
@@ -0,0 +1,84 @@
...
+          events.after_stop.subscribe(&self.class.after_stop)
+
+          # This tag is used by the Error Tracking product to report
+          # the error in Error Tracking
+          span_op.set_tag(Ext::SPAN_EVENTS_HAS_EXCEPTION, true)
+        end
+        # Set a limit to the number of span event we can store per SpanOp
+        if @span_event_per_error.key?(error) || @span_event_per_error.length < SPAN_EVENTS_LIMIT
+          @span_event_per_error[error] =
+            span_event
```
Strech, line 48:
```suggestion
          @span_event_per_error[error] = span_event
```

lib/datadog/error_tracking/collector.rb
```diff
@@ -0,0 +1,84 @@
...
+        end
+      end
+
+      if RUBY_VERSION >= '3.3'
```
Strech, line 52: Could you extract the version to the meaningful named constant please

lib/datadog/error_tracking/collector.rb
```diff
@@ -0,0 +1,84 @@
...
+      else
+        # Up to ruby3.2, we are listening to :raise event. We need to ensure
+        # that an error exiting the scope of a span is not handled in a parent span.
+        # This function will propagate the span event to the parent span. If the
+        # error is not handled in the parent span, it will be deleted by design.
+        def on_error(span_op, error)
+          return unless @span_event_per_error.key?(error)
+
+          if span_op.parent?
+            parent = span_op.send(:parent)
```
Strech, line 68: Is it a private API?

lib/datadog/error_tracking/component.rb
```diff
@@ -0,0 +1,173 @@
...
+            return false
+          end
+          if RUBY_VERSION < '2.6'
+            logger.warn(
+              "error tracking: cannot enable error tracking: Ruby 2.6+ is required, but running
+              on #{RUBY_VERSION}"
+            )
+            return false
+          end
+          true
```
Strech, line 52: I think there is not benefit to go guard-clause for such composition

```suggestion
          if RUBY_ENGINE != 'ruby'
            logger.warn("error tracking: cannot enable error tracking: MRI is required, but running on #{RUBY_ENGINE}")
            false
          elsif RUBY_VERSION < '2.6'
            logger.warn(
              "error tracking: cannot enable error tracking: Ruby 2.6+ is required, but running
              on #{RUBY_VERSION}"
           )
           false
         else
           true
         end
```

lib/datadog/error_tracking/extensions.rb
```diff
@@ -0,0 +1,16 @@
...
+
+require_relative 'configuration'
+require_relative '../core/configuration'
+
+module Datadog
+  module ErrorTracking
+    # Extends Datadog tracing with ErrorTracking features
+    module Extensions
+      # Inject Error Tracking into global objects.
+      def self.activate!
```
Strech, line 11: This method is used in a single place and IMHO it is unnecessary complexity. Do you already know that it will be done somewhere else? If not - I would remove that file and do the method body in place of usage

lib/datadog/error_tracking/filters.rb
```diff
@@ -0,0 +1,76 @@
...
+        get_gem_name(file_path) && !datadog_code?(file_path)
+      end
+
+      def file_included?(file_path, instrumented_files)
+        instrumented_files.include?(file_path)
+      end
+
+      # Generate the proc used in the tracepoint
+      def generate_filter(to_instrument_scope, handled_errors_include = nil)
+        case to_instrument_scope
```
Strech, line 48: instead of return in each `when` the last statement should go to else, and `return`-s could be eliminated

lib/datadog/tracing/span_operation.rb
```diff
@@ -140,6 +141,10 @@ def resource=(resource)
         @resource = resource.nil? ? nil : Core::Utils.utf8_encode(resource) # Allow this to be explicitly set to nil
       end
 
+      def collector
```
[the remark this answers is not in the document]
Strech: Usually we don't do prefixes in ruby, I would rather go with `collector=` method and separate reader.

.standard_todo.yml
```diff
@@ -33,12 +33,14 @@ ignore:
 - lib/*
 - lib/datadog/*
 - lib/datadog/appsec/**/**
+- lib/datadog/error_tracking/**/**
```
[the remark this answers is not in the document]
dubloom: I added that because Rubocop and standard have conflicting rules. Therefore when saving a file, rubocop will automatically make change to the code style that will eventually make the standard ci job to fail.

Strech (commented):
I think we can slightly simplify and Ruby-fy this code

.standard_todo.yml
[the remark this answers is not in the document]
y9v: I really think we should not be adding the whole folder to linter ignore list. If your editor automatically formats files when saving - you should simply turn off auto-formatting.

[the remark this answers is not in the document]
anmarchenko: standardrb also works as a formatter, no need to format using rubocop. Ruby LSP for Cursor/VSCode for example supports standardrb as a formatter

lib/datadog/error_tracking/collector.rb
dubloom: Yes

lib/datadog/error_tracking/collector.rb
dubloom: yes

y9v (changes requested):

docs/GettingStarted.md
dubloom: addressed in 6d7518177cee5d42dd37335629e76cd5a381205e

docs/GettingStarted.md
dubloom: addressed in 6d7518177cee5d42dd37335629e76cd5a381205e

docs/GettingStarted.md
dubloom: addressed in 6d7518177cee5d42dd37335629e76cd5a381205e

docs/GettingStarted.md
dubloom: addressed in 6d7518177cee5d42dd37335629e76cd5a381205e

docs/GettingStarted.md
dubloom: addressed in 6d7518177cee5d42dd37335629e76cd5a381205e

docs/GettingStarted.md
dubloom: addressed in 6d7518177cee5d42dd37335629e76cd5a381205e

docs/GettingStarted.md
dubloom: addressed in 6d7518177cee5d42dd37335629e76cd5a381205e

docs/GettingStarted.md
dubloom: addressed in 6d7518177cee5d42dd37335629e76cd5a381205e

docs/GettingStarted.md
dubloom: addressed in 6d7518177cee5d42dd37335629e76cd5a381205e

lib/datadog/error_tracking/filters.rb
```diff
@@ -0,0 +1,76 @@
...
+      def get_gem_name(file_path)
+        regex = %r{gems/([^/]+)-\d}
+        regex_match = regex.match(file_path)
+        return unless regex_match
+
+        gem_name = regex_match[1]
+
+        begin
+          Gem::Specification.find_by_name(gem_name)
+        rescue
```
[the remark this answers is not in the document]
dubloom: https://apidock.com/ruby/v1_9_3_392/Gem/Specification/find_by_name/class

An error could be risen yes (it should not but I think this is a valid safe guard)

lib/datadog/error_tracking/collector.rb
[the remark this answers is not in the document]
dubloom: The add_span_event is called within a Mutex.lock, it should fine, shouldn't it ?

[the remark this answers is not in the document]
y9v: oh yes, then it's fine

lib/datadog/error_tracking/filters.rb
```diff
@@ -0,0 +1,76 @@
...
+      def user_code?(file_path)
+        !get_gem_name(file_path)
+      end
+
+      def datadog_code?(file_path)
+        file_path.include?('lib/datadog/')
+      end
+
+      def third_party_code?(file_path)
+        get_gem_name(file_path) && !datadog_code?(file_path)
```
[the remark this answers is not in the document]
dubloom: In which case the name of the gem can be `dd-trace-rb` ?

spec/datadog/error_tracking/component_spec.rb
```diff
@@ -0,0 +1,394 @@
...
+
+  describe '.build_errortracking_component' do
+    context 'when ErrorTracking is deactivated' do
+      it 'returns nil' do
+        expect(described_class.build(settings, tracer, logger)).to be_nil
+      end
+    end
+
+    context 'when a wrong argument is passed' do
+      before { settings.error_tracking.handled_errors = 'foo' }
```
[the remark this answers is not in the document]
dubloom: I don't think so.  From my point of view, no. Error Tracking is not a critical component so it should not stop the program if not enabled because of misconfiguration.

lib/datadog/error_tracking/filters.rb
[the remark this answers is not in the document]
y9v: it had this name for version 1.x I think. But I don't think we will be backporting this feature to 1.x? Then you can simply check for `datadog`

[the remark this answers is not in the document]
dubloom: I was not planning to backport to any versions. Is this something common for Ruby Tracer ?

spec/datadog/error_tracking/component_spec.rb
```diff
@@ -0,0 +1,394 @@
...
+        rescue
+          # do nothing
+        end
+        validate_span_events
+      end
+    end
+
+    context 'when an exception is handled then raised' do
+      let(:expected_exceptions) do
+        [[]]
```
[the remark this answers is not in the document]
dubloom: Expected exception represents the span events contained in each span. The first array contains the spans and the sub-arrays the span_events.

If span1 has span_events "foo", "bar" and span2 has span_events "hello", "world".
`expected_exceptions`  = `[["foo", "bar"], ["hello", "world"]]`

spec/datadog/error_tracking/component_spec.rb
```diff
@@ -0,0 +1,394 @@
...
+
+      # Require all the test modules
+      require_relative '../error_tracking/lib1'
+      require_relative '../error_tracking/lib2'
+      require_relative '../error_tracking/sublib/sublib1'
+      require_relative '../error_tracking/sublib/sublib2'
+
+      tracer.enabled = true
+
+      allow(Gem::Specification).to receive(:find_by_name).with('mock-gem').and_return(true)
```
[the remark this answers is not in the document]
dubloom: Removing this line would make the tests fail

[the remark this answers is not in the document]
y9v: `Gem::Specification.find_by_name` actually returns a `Gem::Specification` instance, and not a boolean, so we should at least mock it properly. And why does the test actually fail if this mock is removed? Maybe the mock gem doesn't load properly?

lib/datadog/error_tracking/filters.rb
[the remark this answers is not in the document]
y9v: no, we usually don't do it, only for important bug fixes I think

lib/datadog/tracing/span_operation.rb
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/tracing/span_operation.rb
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/core/configuration/components.rb
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/error_tracking/collector.rb
```diff
@@ -0,0 +1,84 @@
...
+    # @api private
+    class Collector
+      SPAN_EVENTS_LIMIT = 100
+      LOCK = Mutex.new
+      # Proc called when the span_operation :after_stop event is published
+      def self.after_stop
+        @after_stop ||= proc do |span_op, error|
+          collector = span_op.collector
+          # if an error exited the scope of the span, we delete the corresponding SpanEvent.
+          collector.on_error(span_op, error) unless error.nil?
```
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/error_tracking/collector.rb
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/error_tracking/collector.rb
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/error_tracking/collector.rb
```diff
@@ -0,0 +1,84 @@
...
+              parent_collector = parent.collector { Collector.new }
+              parent_collector.add_span_event(parent, error, @span_event_per_error[error])
+            end
+          end
+
+          @span_event_per_error.delete(error)
+        end
+      end
+
+      def get_span_events
```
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/error_tracking/component.rb
```diff
@@ -0,0 +1,173 @@
...
+        @include_path_getter&.disable
+      end
+
+      private
+
+      # Generates a span event from the exception info.
+      #
+      # The event follows the otel semantics.
+      # https://opentelemetry.io/docs/specs/otel/trace/exceptions/
+      def _generate_span_event(exception)
```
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/error_tracking/filters.rb
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/error_tracking/collector.rb
```diff
@@ -0,0 +1,84 @@
...
+      LOCK = Mutex.new
+      # Proc called when the span_operation :after_stop event is published
+      def self.after_stop
+        @after_stop ||= proc do |span_op, error|
+          collector = span_op.collector
+          # if an error exited the scope of the span, we delete the corresponding SpanEvent.
+          collector.on_error(span_op, error) unless error.nil?
+
+          span_events = collector.get_span_events
+          span_op.span_events.concat(span_events) if span_events
```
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/error_tracking/component.rb
```diff
@@ -0,0 +1,173 @@
...
+
+          new(
+            tracer: tracer,
+            handled_errors: settings.error_tracking.handled_errors,
+            handled_errors_include: settings.error_tracking.handled_errors_include,
+          ).tap(&:start)
+        end
+
+        # Checks whether the runtime environment is supported by
+        # error tracking.
```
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/error_tracking/component.rb
```diff
@@ -0,0 +1,173 @@
...
+          true
+        end
+      end
+
+      def initialize(tracer:, handled_errors:, handled_errors_include:)
+        @tracer = tracer
+
+        # Hash containing the file path of the modules to instrument
+        @instrumented_files = Set.new unless handled_errors_include.empty?
+        @handled_errors_include = handled_errors_include
```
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/error_tracking/filters.rb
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

spec/datadog/error_tracking/component_spec.rb
```diff
@@ -0,0 +1,394 @@
...
+
+RSpec.describe Datadog::ErrorTracking::Component do
+  error_tracking_test
+
+  let(:tracer) { new_tracer(enabled: false) }
+  let(:spans) { tracer.writer.spans(:keep) }
+  let(:logger) { Logger.new($stdout) }
+  let(:settings) { Datadog::Core::Configuration::Settings.new }
+
+  def validate_span_events
```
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

spec/datadog/error_tracking/component_spec.rb
```diff
@@ -0,0 +1,394 @@
...
+      it 'creates and starts a component' do
+        component = instance_double(described_class)
+        allow(described_class).to receive(:new).and_return(component)
+        allow(component).to receive(:start).and_return(nil)
+
+        result = described_class.build(settings, tracer, logger)
+
+        expect(result).to eq(component)
+        expect(component).to have_received(:start)
+      end
```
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

spec/datadog/error_tracking/component_spec.rb
```diff
@@ -0,0 +1,394 @@
...
+      it 'capture exception details' do
+        begin
+          tracer.trace('operation') do |span|
+            @span_op = span
+            raise 'this is an exception'
+          # rubocop:disable Lint/UselessRescue
+          rescue
+            raise
+          end
+          # rubocop:enable Lint/UselessRescue
```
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

spec/datadog/error_tracking/component_spec.rb
```diff
@@ -0,0 +1,394 @@
...
+    end
+
+    before do
+      # Configure settings based on test parameters
+      settings.error_tracking.handled_errors = instrument_setting if instrument_setting
+      settings.error_tracking.handled_errors_include = handled_errors_include if handled_errors_include.any?
+
+      @errortracker = described_class.build(settings, tracer, logger)
+
+      # Require the mock gem files
```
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

spec/support/errortracking_helpers.rb
```diff
@@ -0,0 +1,77 @@
+require 'support/platform_helpers'
+
+module ErrortrackingHelpers
```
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/error_tracking/collector.rb
```diff
@@ -0,0 +1,84 @@
...
+        if @span_event_per_error.empty?
+          events = span_op.send(:events)
+          events.after_stop.subscribe(&self.class.after_stop)
+
+          # This tag is used by the Error Tracking product to report
+          # the error in Error Tracking
+          span_op.set_tag(Ext::SPAN_EVENTS_HAS_EXCEPTION, true)
+        end
+        # Set a limit to the number of span event we can store per SpanOp
+        if @span_event_per_error.key?(error) || @span_event_per_error.length < SPAN_EVENTS_LIMIT
```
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/error_tracking/component.rb
```diff
@@ -0,0 +1,173 @@
...
+          # of the files the user want to instrument. This tracepoint is used for that
+          # purpose
+          @include_path_getter = create_script_compiled_tracepoint
+        end
+      end
+
+      def create_exc_tracker_tracepoint(event)
+        TracePoint.new(event) do |tp|
+          active_span = @tracer.active_span
+          unless active_span.nil?
```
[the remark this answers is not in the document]
dubloom: addressed in e5fe33bee6f4993593a80e133f14416eee497861

lib/datadog/error_tracking/extensions.rb
dubloom: This is what is used for DI and appsec. So I think it should stay like that for consistency. What do you think ?

p-datadog: I think if this code is changed it should be changed across the entire library for all products, otherwise the inconsistency is worse (as in more confusing to understand why products are different) than an extra method that follows established convention.

spec/datadog/error_tracking/component_spec.rb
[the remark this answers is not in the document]
dubloom: addressed in 32fac1be4ed9aa416538f683fa6dfb4ae4e4f229

Steepfile
```diff
@@ -102,6 +103,8 @@ target :datadog do
   # and then complains that this type doesn't have any methods including
   # language basics like 'send' and 'raise'.
   ignore 'lib/datadog/di/probe_notifier_worker.rb'
+  ignore 'lib/datadog/error_tracking/configuration/settings.rb'
+  ignore 'lib/datadog/error_tracking/filters.rb'
```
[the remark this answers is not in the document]
dubloom: partially addressed in 32fac1be4ed9aa416538f683fa6dfb4ae4e4f229

lib/datadog.rb
dubloom: 32fac1be4ed9aa416538f683fa6dfb4ae4e4f229

Steepfile
```diff
@@ -77,6 +77,7 @@ target :datadog do
   # steep gets lost in module inclusions
   ignore 'lib/datadog/core/remote/transport/http/config.rb'
   ignore 'lib/datadog/core/remote/transport/http/negotiation.rb'
+  ignore 'lib/datadog/core/error_tracking/filters.rb'
```
p-datadog, line 80: Did you get this remedied?

docs/GettingStarted.md
```diff
@@ -2766,6 +2766,39 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
...
+- Ruby 2.6+. JRuby and TruffleRuby are not supported.
+- datadog 2.16.0+.
+
+#### Configuration
+
+You can enable automatic reporting of handled errors using the following environment variables:
+
+| Environment variable | Type | Description | Default |
+|---|---|---|---|
+| `DD_ERROR_TRACKING_HANDLED_ERRORS` | `String` | Report handled errors from user code, gems, or both. Accepted values are: `user,third_party,all`. | `nil` |
```
p-datadog, line 2784:
```suggestion
| `DD_ERROR_TRACKING_HANDLED_ERRORS` | `String` | Report handled errors from user code, third-party gems, or both. Accepted values are: `user,third_party,all`. | `nil` |
```

docs/GettingStarted.md
```diff
@@ -2766,6 +2766,39 @@ When profiling [Resque](https://github.com/resque/resque) jobs, you should set t
...
+| `DD_ERROR_TRACKING_HANDLED_ERRORS_INCLUDE` | `Array[String]` | A list of comma-separated paths, file names, and gem names for which handled errors will be reported. Possible values are specified below. <br/> For Ruby 3.3 or newer, the location where the error was raised and where it was rescued will be matched. For earlier versions of Ruby, only the location where the error was raised can be matched.  | `[]` |
+
+Alternatively, you can set error tracking parameters in code with these functions, inside a `Datadog.configure` block:
+
+| Environment variable | Type | Description | Default |
+|---|---|---|---|
+| `c.error_tracking.handled_errors` | `String` | Report handled errors from user code, gems, or both. Accepted values are: `user,third_party,all`. | `nil` |
```
p-datadog, line 2791:
```suggestion
| `c.error_tracking.handled_errors` | `String` | Report handled errors from user code, third-party gems, or both. Accepted values are: `user,third_party,all`. | `nil` |
```

lib/datadog/error_tracking/component.rb
```diff
@@ -0,0 +1,167 @@
...
+        if @instrumented_files
+          # The only thing we know about the handled errors is the path of the file
+          # in which the error was rescued. Therefore, we need to retrieve the path
+          # of the files the user want to instrument. This tracepoint is used for that
+          # purpose
+          @include_path_getter = create_script_compiled_tracepoint
+        end
+      end
+
+      def create_exc_tracker_tracepoint(event)
```
p-datadog, line 81: FYI I use `trace_point` in DI code rather than `tracepoint`, to match the `TracePoint` class name, but profiling seems to use `tracepoint`.

p-datadog (approved):

docs/GettingStarted.md
dubloom: addressed in 4ba22eef203551988c84b671db9fd1f5c75384a1

docs/GettingStarted.md
dubloom: 4ba22eef203551988c84b671db9fd1f5c75384a1

lib/datadog/error_tracking.rb
```diff
@@ -0,0 +1,18 @@
...
+
+module Datadog
+  # Namespace for Datadog ErrorTracking.
+  #
+  # @api private
+  module ErrorTracking
+    # Expose ErrorTracking to global shared objects
+    Extensions.activate!
+  end
+end
```
Strech, line 18: This module is irresponsible and should be removed

lib/datadog/error_tracking/collector.rb
```diff
@@ -0,0 +1,88 @@
...
+        # When this is the first time we add a span event for a span,
+        # we suscribe to the :after_stop event
+        if @span_event_per_error.empty?
+          events = span_op.send(:events)
+          events.after_stop.subscribe(&self.class.after_stop)
+
+          # This tag is used by the Error Tracking product to report
+          # the error in Error Tracking
+          span_op.set_tag(Ext::SPAN_EVENTS_HAS_EXCEPTION, true)
+
```
Strech, line 45:
```suggestion
```

Strech (changes requested):
I'm fine with that changes 👍🏼 

But I would like to demand the Rubocop to be enabled. First of all, I've pointed out lots of small issue which suppose to be eliminated by Rubocop. At the same time many similar issues are still present.

I'm not sure what leads to the disablement of the Rubocop, but I don't see any objective reason

ivoanjo:
> I'm not sure what leads to the disablement of the Rubocop, but I don't see any objective reason

I believe this PR is using standard, not rubocop. This is the default for any new folders -- you either opt out of standard or opt out of rubocop.

Strech (approved):
I'm approving this PR, but IMHO too many minor here-and-there corrections are not catched by standard-rb.

y9v (approved):

PR #4605, opened 2025-04-25
feat: after_stop.publish now take two arguments

dubloom:
**What does this PR do?**
When publishing the `after_stop` event, we can now provide an error as a second argument. This allows us to know just after stop if an error exited the scope of the span.

**Motivation:**
For automatic reporting of handled exceptions feature (specifications [here](https://docs.google.com/document/d/13sg08mQH-6QnNT4h0DGVqbq-JF2pwWEjsCIBblUHC_w/edit?usp=sharing)), we need to know before the `span` object is created if an error exited the scope of the span. The end goal is to not report an handled exception that has been rethrown.

My first try was to make the `:on_error` event subscribable (closed PR [here](https://github.com/DataDog/dd-trace-rb/pull/4559)). However after discussion with @p and @marcotc, we decided to not to do it this way as it was changing a public facing behavior. The proposition was to create a new `Tracing::Event`, however re-using `after_stop` is simpler and also works.

**Change log entry**
No.

**How to test the change?**
The `span_operation_spec` has been updated to check that `after_stop` receives the error when one is occuring


lib/datadog/tracing/span_operation.rb
```diff
@@ -202,7 +202,7 @@ def start(start_time = nil)
       #
       # steep:ignore:start
       # Steep issue fixed in https://github.com/soutaro/steep/pull/1467
-      def stop(stop_time = nil)
+      def stop(stop_time = nil, error = nil)
```
p-datadog, line 205: I think I would prefer a keyword argument for `error` to avoid having to pass two `nils` should we add more arguments here in the future.

p-datadog:
We reviewed other events and for example `after_finish` also takes two arguments, therefore I think it's OK to have an error argument to the `after_stop` event. We also discovered that apparently there are currently no subscribers to the `after_stop` event in the codebase.

sig/datadog/tracing/span_operation.rbs
```diff
@@ -58,7 +58,7 @@ module Datadog
 
       def start: (?untyped? start_time) -> self
 
-      def stop: (?untyped? stop_time) -> self?
+      def stop: (?untyped? stop_time, ?Exception? error) -> self?
```
Strech, line 61: Don't want to block you, but could you provide the type for `stop_time`?

Strech (commented):
Great job 👏🏼 
I agree with Oleg on kwargs instead of tossing `nil` values

Strech (approved):

sig/datadog/tracing/span_operation.rbs
dubloom: adressed in 791f59c9dd1972d5d39ef90d407bda431065436a

lib/datadog/tracing/span_operation.rb
dubloom: addressed in 791f59c9dd1972d5d39ef90d407bda431065436a

p-datadog (approved):

marcotc (approved):

PR #4592, opened 2025-04-17
[APMAPI-1116] Stable configuration

vpellan:
**What does this PR do?**

Add stable config support.

**Motivation:**

Fleet Automation will inject a configuration file onto hosts that the APM SDKs must read, apply, and report telemetry about

**Change log entry**

None.

**Additional Notes:**

Needs to wait for libdatadog 17.1 release because of `ddog_library_configurator_with_detect_process_info`, so critical code is commented out.

**How to test the change?**

System-tests, following the instructions [here](https://github.com/DataDog/system-tests/blob/main/utils/build/docker/ruby/parametric/README.md)

PR #4610, opened 2025-04-28
[NO-TICKET] Decouple environment variable from programmatic config precedence

vpellan:
**What does this PR do?**

This PR decouples environment variables and programmatic configuration precedence value by adding an environment precedence value.

**Motivation:**

in https://github.com/DataDog/dd-trace-rb/pull/4592, I need to add a new precedence value higher than environment but lower than programmatic.

**Change log entry**

Yes. Change behaviour of programmatic configurations previously overwriting environment variables. Unsetting a programmatic configuration will now restore the environment variable value.

**Additional Notes:**

I'm not sure if this is considered a breaking change, it seems like this was a bug in the first place (booting an app with an env var, a programmatic config and unsetting that programmatic config previously dismissed the env var too)

**How to test the change?**

CI.

### DIFF
--- a/sig/datadog/core/configuration/options.rbs
+++ b/sig/datadog/core/configuration/options.rbs
@@ -37,7 +37,7 @@ module Datadog
 
           private
 
-          def resolve_option: (Symbol name) -> Datadog::Core::Configuration::Option
+          def resolve_option: (Symbol name) -> Option
 
           def add_option: (untyped name) -> untyped
 
--- a/sig/datadog/core/configuration/options.rbs
+++ b/sig/datadog/core/configuration/options.rbs
@@ -37,6 +37,8 @@ module Datadog
 
           private
 
+          def resolve_option: (Symbol name) -> Datadog::Core::Configuration::Option
+
           def add_option: (untyped name) -> untyped
 
           def assert_valid_option!: (untyped name) -> untyped
--- a/spec/datadog/core/configuration/option_spec.rb
+++ b/spec/datadog/core/configuration/option_spec.rb
@@ -553,81 +553,119 @@ RSpec.describe Datadog::Core::Configuration::Option do
     # | default                    | default             | Option is reset |
     # | rc, default                | rc                  | default         |
     # | rc, default                | default             | rc              |
-    #
-    # Coding in tests is not great, but the alternative is to hardcode 2^n combinations
-    # With n the number of values in the precedence list
-    # (with fleet and customer defined stable config, this will go up to 64 combinations)
-    combinations = (0..Datadog::Core::Configuration::Option::Precedence::LIST.length).flat_map do |n|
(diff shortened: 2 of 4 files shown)


spec/datadog/core/configuration/option_spec.rb
```diff
@@ -553,30 +553,23 @@
...
-      ]
-    }.each do |name, precedences|
-      context "for #{name} set" do
+    #
+    # Coding in tests is not great, but the alternative is to hardcode 2^n combinations
+    # With n the number of values in the precedence list
+    # (with fleet and customer defined stable config, this will go up to 64 combinations)
+    combinations = (0..Datadog::Core::Configuration::Option::Precedence::LIST.length).flat_map do |n|
+      Datadog::Core::Configuration::Option::Precedence::LIST.combination(n).to_a
+    end
```
Strech, line 562: I don't think that you need to code all the possible combinations, you just need the majority, which can use transitive rule to imply that all the rest will work too, and if we found a bug - we can fix it.

The list you replace was doing exactly it.

sig/datadog/core/configuration/options.rbs
```diff
@@ -37,6 +37,8 @@ module Datadog
 
           private
 
+          def resolve_option: (Symbol name) -> Datadog::Core::Configuration::Option
```
Strech, line 40: You can cut the unnecessary namespace here.

Strech (approved):
I think it's a great change, but I do think that testing gone a bit over the edge.

marcotc (approved):

spec/datadog/core/configuration/option_spec.rb
```diff
@@ -553,81 +553,119 @@
...
-        before do
-          allow(context).to(receive(:instance_exec)) { |value, _| value }
+    # We don't need to test all precedence set combinations.
+    context 'with preset precedence' do
+      before do
+        allow(context).to(receive(:instance_exec)) { |value, _| value }
+      end
+
+      context 'when no precedence value is set and try to unset a precedence that is not set' do
+        subject!(:unset) { option.unset(Datadog::Core::Configuration::Option::Precedence::PROGRAMMATIC) }
```
Strech, line 564: Since it's not used, it should be converted into before block

```suggestion
        before { option.unset(Datadog::Core::Configuration::Option::Precedence::PROGRAMMATIC) }

```

Strech (approved):
I have a few suggestions for tests

spec/datadog/core/configuration/option_spec.rb
vpellan: Changed in [fae0b1c](https://github.com/DataDog/dd-trace-rb/pull/4610/commits/fae0b1c4d22b89202157ce8439d161da0e9b542e) !

PR #4611, opened 2025-04-28
Warn when ignoring on_error argument due to it not being a proc

p-datadog:
**What does this PR do?**

Adds a warning when customer provides `on_error` argument which is not a `Proc`. Previously the argument was silently ignored.

**Motivation:**

Better diagnostics of unexpected/incorrect usage of the library

**Change log entry**
Yes: warn when `on_error` argument is not a `Proc` (and would therefore be ignored)

**Additional Notes:**

The existing tests for on_error behavior had incorrect setup and did not actually test the path when on_error argument was not a proc.

**How to test the change?**
Repaired unit tests in this PR.

### DIFF
--- a/lib/datadog/tracing/span_operation.rb
+++ b/lib/datadog/tracing/span_operation.rb
@@ -93,8 +93,14 @@ module Datadog
         @events = events || Events.new
         @span = nil
 
-        # Subscribe :on_error event
-        @events.on_error.wrap_default(&on_error) if on_error.is_a?(Proc)
+        if on_error.nil?
+          # Nothing, default error handler is already set up.
+        elsif on_error.is_a?(Proc)
+          # Subscribe :on_error event
+          @events.on_error.wrap_default(&on_error)
+        else
+          Datadog.logger.warn("on_error argument to SpanOperation ignored because is not a Proc: #{on_error}")
+        end
 
         # Start the span with start time, if given.
         start(start_time) if start_time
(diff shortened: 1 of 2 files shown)


marcotc (approved):

totus:
https://github.com/DataDog/dd-trace-rb/issues/4670 is a direct consequence

totus:
@p-datadog, exactly. When a gem is simply being included without any specific configuration, it produces the warnings.

To be more specific, even in our test repo with a configuration like this:
```ruby
# Datadog tracing
require "datadog/auto_instrument" if ENV["DD_ENV"] == "ci"

Datadog.configure do |c|
  c.ci.enabled = ENV.fetch("DD_ENV", false) == "ci"
  c.tracing.enabled = ENV.fetch("DD_ENV", false) == "ci"
  c.service = ENV.fetch("DD_SERVICE", nil)
  c.ci.instrument :cucumber
end
```
The test execution is polluted with log entries:
```
W, [2025-05-21T03:47:15.590188 #1998]  WARN -- datadog: [datadog] on_error argument to SpanOperation ignored because is not a Proc: #<Method: Net::HTTP(Datadog::Tracing::Contrib::HTTP::Instrumentation::InstanceMethods)#annotate_span_with_error!(span, error) /home/runner/work/repo-name/repo-name/vendor/bundle/ruby/3.4.0/gems/datadog-2.16.0/lib/datadog/tracing/contrib/http/instrumentation.rb:115>
```

PR #4613, opened 2025-04-28
Avoid method redefinition warning when replacing time providers

p-datadog:
**What does this PR do?**

Undefines existing time provider methods before redefining them to avoid the following warnings:

```
/home/w/apps/dd-trace-rb/lib/datadog/core/utils/time.rb:46: warning: method redefined; discarding old get_time
/home/w/apps/dd-trace-rb/lib/datadog/core/configuration/settings.rb:725: warning: previous definition of get_time was here
```

**Motivation:**
Warning-clean library

**Change log entry**
Yes: avoid Ruby warnings when providing a custom time provider

**Additional Notes:**
The method removal is guarded in case customers remove the methods themselves to avoid the warning

**How to test the change?**
Tested manually by running the relevant examples:
```
rspec spec/datadog/core/configuration/settings_spec.rb spec/datadog/tracing/configuration/settings_spec.rb
```


marcotc (approved):

lib/datadog/core/utils/time.rb
```diff
@@ -31,6 +31,10 @@ def now
         #
         # @param block [Proc] block that returns a `Time` object representing the current wall time
         def now_provider=(block)
+          class << self
+            # Avoid method redefinition warning
+            remove_method(:now) rescue nil
+          end
           define_singleton_method(:now, &block)
```
Strech, line 38: could those 2 calls be united to avoid double cache busting?

p-datadog: Do you mean moving `define_singleton_method` into the `class << self` as `define_method`?

Strech: I'm fine with any, the goal is to do class modification just once 👍🏼

p-datadog: That did not work because apparently `class << self` does not have access to outer locals, `block` in this case.

y9v (approved):

PR #4629, opened 2025-05-07
Add a gatekeeper job for PR

cbeauchesne:
**What does this PR do?**
Add a job in the CI that runs in PR. The job will be a success if all other jobs are skipped/success, and fails otherwise

**Motivation:**

While it's possible to enforce a green CI policy using GitHub's native "required status checks" feature, doing so requires explicitly listing all job names under branch protection rules. This approach has two key drawbacks:

- It does not support optional jobs
- It introduces ongoing maintenance overhead as the job list evolves

We can also add "job that runs at the end of each workflows", and make them a requirement, but we need to do so for each workflow. This jobs achieve the same, but all-in-one by checking all jobs that are executed for a PR. 

The action used offers a `ignored-name-patterns` parameters`, but right now, dd-trace-rb CI seems to be in a very good health, so it's empty.

The two other parameters are only for the retry logic. Please not that an itinitial check is performed before the initial delay, to handle the use case where somebody re-run all-green when all other are green.

The plan is to merge this PR, wait few days to be sure that everything is fine. Then add it as a requirement.

**Change log entry**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/.github/workflows/all-green.yml
+++ b/.github/workflows/all-green.yml
@@ -16,17 +16,18 @@ permissions:
 jobs:
   all-jobs-are-green:
     runs-on: ubuntu-latest
+    timeout-minutes: 60
     steps:
       - name: Run Ensure CI Success
-        uses: DataDog/ensure-ci-success@727e7fe39ae2e1ce7ea336ec85a7369ab0731754
+        uses: DataDog/ensure-ci-success@727e7fe39ae2e1ce7ea336ec85a7369ab0731754 # v2.1.1
         with:
           # How long to wait before we start checking the job status
           # GitHub's API has a limit, so this helps avoid making too many calls
           # Note: If the job is retried, we'll check it once right away before waiting
-          initial-delay-seconds: "900"
+          initial-delay-seconds: 900
 
           # How long to wait between each check
-          max-retries: "60"
+          max-retries: 60
 
           # Sleep time between each check (default: 60 seconds)
           # polling-interval-seconds: "60"
--- a/.github/workflows/all-green.yml
+++ b/.github/workflows/all-green.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Run Ensure CI Success
-        uses: DataDog/ensure-ci-success@1517bce5564037de6a02562a6b4f119503c9781d
+        uses: DataDog/ensure-ci-success@727e7fe39ae2e1ce7ea336ec85a7369ab0731754
         with:
           initial-delay-seconds: "900"
           max-retries: "60"
--- a/.github/workflows/all-green.yml
+++ b/.github/workflows/all-green.yml
@@ -20,7 +20,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Run Ensure CI Success
-        uses: DataDog/ensure-ci-success@f40e6ffd8e60280d478b9b92209aaa30d3d56895
+        uses: DataDog/ensure-ci-success@1517bce5564037de6a02562a6b4f119503c9781d
         with:
           initial-delay-seconds: "900"
           max-retries: "60"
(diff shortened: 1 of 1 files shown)


lloeki:
Hmm, that doesn't work as a replacement for synthetic jobs?

If for some reason tomorrow, say, some or all unit tests accidentally disappeared from the runs they would be nowhere to be seen and the "all green" condition would be satisfied even though the whole thing entirely lacked unit tests.

You HAVE to list them in some way or you're only enforcing that whatever is there has passed, but not that what you expect to be there is **actually** there. Big difference.

The way I did it is list them by way of them being a `needs` dependency of a synthetic job. Done per workflow because that's easier and more local.

In addition, there are secondary jobs whose result we don't care about whether they failed or not for a PR to be mergeable. They are _not_ required status checks, on purpose.

TonyCTHsu:
Question related to initial delay, when the job being re-run, do we still need to wait for initial delay?

.github/workflows/all-green.yml
```diff
@@ -0,0 +1,26 @@
...
+  statuses: read
+
+jobs:
+  all-jobs-are-green:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run Ensure CI Success
+        uses: DataDog/ensure-ci-success@f40e6ffd8e60280d478b9b92209aaa30d3d56895
+        with:
+          initial-delay-seconds: "600"
```
TonyCTHsu, line 25: 1. The value should be string. 
2. A better value for this repository is 15 minute.
3. Rerunning the job should skip the delay.

cbeauchesne:
@TonyCTHsu with [0baf375](https://github.com/DataDog/dd-trace-rb/pull/4629/commits/0baf37583b151204a01fa7cb96de4b7184211de3), this action now : 

* perform an initial try if thee attempt is not the first one
* use your solution to tackle API limitations.

Do you mind performing a review ?

.github/workflows/all-green.yml
```diff
@@ -0,0 +1,26 @@
+name: Check Pull Request CI Status
+
+on: # yamllint disable-line rule:truthy
+  pull_request:
+    types:
```
TonyCTHsu, line 5: - The listed types are the default values to be removed. 
- Scoping to master branch would be ideal.

.github/workflows/all-green.yml
```diff
@@ -0,0 +1,26 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  checks: read
+  statuses: read
+
+jobs:
+  all-jobs-are-green:
```
TonyCTHsu, line 19: Set a 30 min timeout for the job.

.github/workflows/all-green.yml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run Ensure CI Success
+        uses: DataDog/ensure-ci-success@727e7fe39ae2e1ce7ea336ec85a7369ab0731754
+        with:
+          initial-delay-seconds: "900"
+          max-retries: "60"
```
p-datadog, line 24: Do you mind adding a comment explaining what these options are doing?

p-datadog (commented):
We usually have all jobs passing, however historically if I remember correctly we have had issues with ASAN configurations. Do you mind adding a note somewhere either in the all-green.yml file or another suitable location stating how to exclude specific configurations from the green requirement, if we need to in the future?

Otherwise this PR LGTM if I understand correctly  that it adds another job which will be red if any of the other jobs are red.

p-datadog:
Background: https://github.com/orgs/community/discussions/26733

.github/workflows/all-green.yml
cbeauchesne: done in [b686bb7](https://github.com/DataDog/dd-trace-rb/pull/4629/commits/b686bb7a9c9bb41d6357661dc2c93f6ffbcb9cfb)

.github/workflows/all-green.yml
TonyCTHsu: Set `timeout-minutes: 30` for this job

TonyCTHsu (approved):

PR #4630, opened 2025-05-07
DEBUG-3700 Replace telemetry transport with a core-based transport that supports UDS

p-datadog:
**What does this PR do?**

Replaces telemetry's simple transport by the proper "enterprise" transport that implements Unix domain sockets besides HTTP.

**Motivation:**

Shopist requires usage of UDS for everything.

We also had support escalations asking for UDS support in telemetry.

**Change log entry**
None

**Additional Notes:**
I used DI transport as a starting point. There is a fair amount of boilerplate in transport code (still) which I hope to reduce one day in the future.

Telemetry initialization logic that was in Component.build was moved to Component#initialize to centralize processing.

`http_transport` attributes within telemetry were renamed to `transport` for consistency with other transports and because UDS is now supported.

Telemetry worker will now not be created at all if telemetry is disabled. This was the easiest route to get some of the tests passing and I believe this is the correct behavior.

I feel like the `agent_settings` should be a cached attribute of `settings` rather than having consumers deal with AgentSettingsResolver (again maybe in the bright future).

**How to test the change?**

Existing CI including integration tests added in https://github.com/DataDog/dd-trace-rb/pull/4588 that verify payloads.

Integration tests were expanded to add a UDS configuration.


lib/datadog/core/telemetry/component.rb
```diff
@@ -25,89 +26,85 @@ def self.build(settings, agent_settings, logger)
...
-            enabled: @enabled,
-            heartbeat_interval_seconds: heartbeat_interval_seconds,
-            metrics_aggregation_interval_seconds: metrics_aggregation_interval_seconds,
-            emitter: Emitter.new(http_transport: http_transport),
-            metrics_manager: @metrics_manager,
-            dependency_collection: dependency_collection,
-            logger: logger,
-            shutdown_timeout: shutdown_timeout_seconds
-          )
+          @transport = if @enabled
```
Strech, line 58: I have a suggestion here, to avoid if/else for this case, WDYT about NullObject pattern? Since we have a component constructor, we can always control what will be returned. It will eliminate complexity and streamline most of the conditions (also easier to test)

lib/datadog/core/telemetry/component.rb
```diff
@@ -25,89 +26,85 @@ def self.build(settings, agent_settings, logger)
...
+                             host_prefix: 'instrumentation-telemetry-intake',
+                             url_override: settings.telemetry.agentless_url_override,
+                             url_override_source: 'c.telemetry.agentless_url_override',
+                             logger: logger,
+                           )
+                           Telemetry::Transport::HTTP.agentless_telemetry(
+                             agent_settings: agent_settings,
+                             logger: logger,
+                             # api_key should have already validated to be
+                             # not nil by +build+ method above.
```
anmarchenko, line 71: do you mean `AgentlessSettingsResolver.call` here?

lib/datadog/core/telemetry/request.rb
```diff
@@ -11,9 +11,9 @@ module Request
         class << self
           using Core::Utils::Hash::Refinement
 
-          def build_payload(event, seq_id)
+          def build_payload(event, seq_id, api_version: 'v2')
```
Strech, line 14: I think we should move `v2` into a constant then

lib/datadog/core/telemetry/transport/telemetry.rb
```diff
@@ -0,0 +1,51 @@
...
+
+          class Request < Datadog::Core::Transport::Request
+            def initialize(request_type, parcel, api_key)
+              @request_type = request_type
+              super(parcel)
+              @api_key = api_key
+            end
+
+            attr_reader :request_type
+            attr_reader :api_key
```
Strech, line 24: Can we move them on top to be consistent over the style of the file?

Strech (commented):
I have a suggestion to redo the enabled flag internals into 2 separate enabled and disabled telemetry components (or some sub-classes).

In addition I think few interfaces could be slightly changed (like other classes in this PR)

anmarchenko (approved):
I have no objections and agentless telemetry tests passed: great job @p-datadog

lib/datadog/core/telemetry/emitter.rb
```diff
@@ -11,25 +10,26 @@ module Core
...
 
         extend Core::Utils::Forking
 
-        # @param http_transport [Datadog::Core::Telemetry::Http::Transport] Transport object that can be used to send
-        #   telemetry requests via the agent
-        def initialize(http_transport:)
-          @http_transport = http_transport
+        # @param transport [Datadog::Core::Telemetry::Transport::Telemetry::Transport]
+        #   Transport object that can be used to send telemetry requests
+        def initialize(transport:, logger: Datadog.logger)
```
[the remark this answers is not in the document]
p-datadog: OK but this will just create churn in the future if other arguments are added. Telemetry uses keyword arguments everywhere which works for any set of arguments, unlike positional/keyword mix used by other components which becomes awkward as more arguments are added.

y9v (approved):
great job!

Strech (approved):

Strech:
Awesome 👏🏼

lib/datadog/core/telemetry/transport/telemetry.rb
p-datadog: Moved.

PR #4633, opened 2025-05-09
[NO-TICKET] Add logging during configuration initialization

vpellan:
**What does this PR do?**

This PR adds a default logger during configuration initialization.

**Motivation:**

During config initialization, stable config will be able to log errors related to reading or parsing config files. But when doing this, configuration isn't loaded. `logger_without_components` still rely on configuration, and will create an infinite loop.

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

CI.

### DIFF
--- a/lib/datadog/core/configuration.rb
+++ b/lib/datadog/core/configuration.rb
@@ -244,6 +244,10 @@ module Datadog
         end
       end
 
+      def configuration?
+        (defined?(@configuration) && @configuration) != nil
+      end
+
       def components?
         # This does not need to grab the COMPONENTS_READ_LOCK because it's not returning the components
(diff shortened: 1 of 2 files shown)


marcotc (approved):

ivoanjo:
Hmm I don't think this change is correct.

What other components do to be able to log during initialization is to receive the logger as an argument -- e.g. https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/core/configuration/components.rb#L96-L131 .

Creating a temporary logger makes it impossible to configure by customers (to e.g. hide messages if they don't want to see them). There were actually a few logs from the profiler that used the temporary logger, and I did get customer questions at least twice on how they can be silenced.

PR #4641, opened 2025-05-12
DEBUG-3700 Inject telemetry into metrics client

p-datadog:
**What does this PR do?**

Injects telemetry into metrics client instead of retrieving it from the components tree.

**Motivation:**

References to the components tree currently create the tree if it does not exist by default (`allow_initialization` option to `components`). This creation is problematic in the test suite (see https://github.com/DataDog/dd-trace-rb/pull/4619). Some components appear to rely on this automatic instantiation (profiling, possibly tracing), therefore removing it is non-trivial. In the meantime, metrics client can have telemetry injected into it to avoid the issue altogether (and remove what is effectively a global variable reference).

**Change log entry**
None

**Additional Notes:**

**How to test the change?**

Existing CI

### DIFF
--- a/lib/datadog/core/metrics/client.rb
+++ b/lib/datadog/core/metrics/client.rb
@@ -21,9 +21,10 @@ module Datadog
         extend Options
         extend Helpers
 
-        attr_reader :statsd, :logger
+        attr_reader :statsd, :logger, :telemetry
 
-        def initialize(logger: Datadog.logger, statsd: nil, enabled: true, **_)
+        def initialize(telemetry: nil, logger: Datadog.logger, statsd: nil, enabled: true, **_)
+          @telemetry = telemetry
           @logger = logger
           @statsd =
             if supported?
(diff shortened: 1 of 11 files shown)


lib/datadog/core/metrics/client.rb
```diff
@@ -21,9 +21,10 @@ class Client
         extend Options
         extend Helpers
 
-        attr_reader :statsd, :logger
+        attr_reader :statsd, :logger, :telemetry
 
-        def initialize(logger: Datadog.logger, statsd: nil, enabled: true, **_)
+        def initialize(telemetry: nil, logger: Datadog.logger, statsd: nil, enabled: true, **_)
```
Strech, line 26: I think I have just one question, will it be ever that telemetry is `nil` because not provided? Can we enforce its presence instead of safe-navigation?

Strech: What about NullObject pattern to avoid unnecessary checks?

p-datadog: I removed the defaulting. The concern I had was if any customer was instantiating these classes.

PR #4643, opened 2025-05-12
DEBUG-3573 Add integration test for telemetry app-heartbeat event payloads

p-datadog:
**What does this PR do?**

Adds an integration test for AppHeartbeat telemetry event with payload assertions

**Motivation:**
Debugging why telemetry is not working properly for dynamic instrumentation - the UI requires telemetry heartbeats to show correct information but these are not seen by the backend 

**Change log entry**
None

**Additional Notes:**

This PR duplicates the `mark_telemetry_started` helper from https://github.com/DataDog/dd-trace-rb/pull/4640

**How to test the change?**
PR is tests only

### DIFF
--- a/spec/support/telemetry_helpers.rb
+++ b/spec/support/telemetry_helpers.rb
@@ -1,0 +1,15 @@
+module TelemetryHelpers
+  module ClassMethods
+    def mark_telemetry_started
+      before do
+        Datadog::Core::Telemetry::Worker::TELEMETRY_STARTED_ONCE.run do
+          true
+        end
+      end
+    end
+  end
+
+  def self.included(base)
+    base.extend(ClassMethods)
+  end
+end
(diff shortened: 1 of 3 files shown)


spec/support/telemetry_helpers.rb
```diff
@@ -0,0 +1,15 @@
+module TelemetryHelpers
+  module ClassMethods
+    def mark_telemetry_started
+      before do
+        Datadog::Core::Telemetry::Worker::TELEMETRY_STARTED_ONCE.run do
```
Strech, line 5: Could it be converted into stub? Because now it mutates the global state

Strech (approved):
I like the changes, but I do have concerns over mutating "run once" constant, I would love to see it as a stub, over the mutation

spec/support/telemetry_helpers.rb
p-datadog: Added this to my backlog, I don't currently know how to do this for the run once guards.

PR #4644, opened 2025-05-13
[APPSEC-57237] Handle business logic events

Strech:
**What does this PR do?**

Adds handling of business logic events to be able to finish fingerprinting feature.

**Motivation:**

The last bits of fingerprinting is to act on business logic events that produced during sign in via auto-instrumentation or SDK.

**Change log entry**

Yes. AppSec: Add session tracking and fingerprinting support to `devise` and `rails` contribs.

**Additional Notes:**

I would like to restructure integration tests to better separate specifically contrib tests from general logic tests. Hence you would see test in the rack contrib (for now)

Second - we have an issue in libddwaf-rb which converts provided `nil` values into empty strings. Another option is to give a value `invalid` which doesn't make much sense, but it is what it is.

This is what's new in system-tests

Screenshot 2025-05-13 at 11 59 46

**How to test the change?**

CI + ST

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -22,7 +22,7 @@ permissions: {}
 env:
   REGISTRY: ghcr.io
   REPO: ghcr.io/datadog/dd-trace-rb
-  SYSTEM_TESTS_REF: appsec-57504-enable-session-tracking-and-fingerprinting # This must always be set to `main` on dd-trace-rb's master branch
+  SYSTEM_TESTS_REF: appsec-57237-ruby-enable-session-fingerprinting # This must always be set to `main` on dd-trace-rb's master branch
 
 jobs:
   changes:
--- a/lib/datadog/appsec/monitor/gateway/watcher.rb
+++ b/lib/datadog/appsec/monitor/gateway/watcher.rb
@@ -10,11 +10,17 @@ module Datadog
       module Gateway
         # Watcher for Apssec internal events
         module Watcher
+          ARBITRARY_VALUE = 'invalid'
+          EVENT_LOGIN_SUCCESS = 'users.login.success'
+          EVENT_LOGIN_FAILURE = 'users.login.failure'
+          WATCHED_LOGIN_EVENTS = [EVENT_LOGIN_SUCCESS, EVENT_LOGIN_FAILURE].freeze
+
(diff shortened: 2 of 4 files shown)


p-datadog (approved):

lib/datadog/appsec/monitor/gateway/watcher.rb
```diff
@@ -47,6 +53,30 @@ def watch_user_id(gateway = Instrumentation.gateway)
...
+            def watch_user_login(gateway = Instrumentation.gateway)
+              gateway.watch('appsec.events.user_lifecycle', :appsec) do |stack, kind|
+                context = AppSec.active_context
+
+                next stack.call(kind) unless WATCHED_LOGIN_EVENTS.include?(kind)
+
+                persistent_data = { "server.business_logic.#{kind}" => ARBITRARY_VALUE }
+                result = context.run_waf(persistent_data, {}, Datadog.configuration.appsec.waf_timeout)
+
+                if result.match? || !result.derivatives.empty?
```
y9v, line 66: can we use `result.derivatives.any?` to avoid negation?

y9v, line 66: can we use `result.derivatives.any?` here to avoid negation? Or `result.derivatives` might not be an array?

y9v (approved):

.github/workflows/system-tests.yml
```diff
@@ -22,7 +22,7 @@ permissions: {}
 env:
   REGISTRY: ghcr.io
   REPO: ghcr.io/datadog/dd-trace-rb
-  SYSTEM_TESTS_REF: appsec-57504-enable-session-tracking-and-fingerprinting # This must always be set to `main` on dd-trace-rb's master branch
+  SYSTEM_TESTS_REF: appsec-57237-ruby-enable-session-fingerprinting # This must always be set to `main` on dd-trace-rb's master branch
```
vpellan, line 25: Don't forget to change it back to main when system-tests PR is merged before merging this PR

Strech: I can't change it in system tests without merging it. Only after

Strech: The issue was - I deleted the branch (already recovered)

PR #4577, opened 2025-04-11
[PROF-11524] Upgrade libdatadog dependency to 17.1

ivoanjo:
**What does this PR do?**

Updates the native bits in the gem (profiling, crashtracking) to be able to work with the latest libdatadog 17.

I'm opening it as a draft so we can use it to test the upcoming libdatadog 17.1 release. Once that release is out and on rubygems.org, I plan to:

1. Push the actual gem dependency change on this PR
2. Do a pass on the TODO/FIXME things that are still missing

**Motivation:**

There's a number of fixes and additions in libdatadog that we want to make use of.

**Change log entry**

Yes. Upgrade libdatadog dependency to 17.1

**Additional Notes:**

With this PR, I'm able to build both crashtracking AND profiling with libdatadog master as of 11th of April. I'm also able to get a green test suite for profiling, but not for crashtracking: I've reported the issue to the folks working on that, so they're aware.

**How to test the change?**

Once everything is in place, our existing test coverage will be enough to validate these changes. Until then it's expected that CI is going to be red.

PR #4649, opened 2025-05-14
[APMAPI-1068] Enable Service Discovery

vpellan:
**What does this PR do?**

It uncomment all the code related to service discovery. 

**Motivation:**

Preparing everything so process discovery can be released within just a few clicks

**Change log entry**

Yes. Add Service Discovery capability, enabling the tracer to reliably determine which process are instrumented using Datadog's tracer.

**Additional Notes:**

Once this PR is merged, please merge this one: https://github.com/DataDog/system-tests/pull/4652

**How to test the change?**


ivoanjo:
Update: I've gone ahead and merged #4577 into this branch so it runs with libdatadog 18.1

lib/datadog/core/configuration/components.rb
```diff
@@ -211,8 +210,7 @@ def shutdown!(replacement = nil)
           telemetry.emit_closing! unless replacement&.telemetry&.enabled
           telemetry.shutdown!
 
-          # TODO: Re-enable this once we have updated libdatadog to 17.1
-          # Core::ProcessDiscovery._native_close_tracer_memfd(@process_discovery_fd, @logger) if @process_discovery_fd
+          Core::ProcessDiscovery._native_close_tracer_memfd(@process_discovery_fd, @logger) if @process_discovery_fd
```
ivoanjo, line 213: Minor: Not a blocker for this PR, but I guess it would be nice to bind a `shutdown!` to the `@process_discovery_fd`, so this would become `@process_discovery_fd&.shutdown!`, thus "leaking" less of the implementation details here.

`shutdown!` doesn't even need to be native code, it could be pure-Ruby code that then invokes the `_native_close_tracer_memfd`.

(Also bonus points inside `get_and_store_metadata` we could also store the `logger` as an instance variable of the `TracerMemfd` object, so we wouldn't need to pass it again here -- this also could be done without any changes to the native code).

spec/datadog/core/configuration/components_spec.rb
```diff
@@ -28,7 +28,12 @@
 RSpec.describe Datadog::Core::Configuration::Components do
   subject(:components) { described_class.new(settings) }
 
-  let(:logger) { instance_double(Datadog::Core::Logger) }
+  # Using a double does not give access to superclass methods like debug
```
ivoanjo, line 31:
```suggestion
```

Minor: Having to allow methods in a double to be called is something common to all doubles (it's the default rspec behavior) so I'm not sure the comment adds a lot ;)

ivoanjo (approved):
:+1: LGTM and great work driving this across the line!

lib/datadog/core/configuration/components.rb
vpellan: Opened another PR to fix this (as this one is supposed to be only about enabling the feature): https://github.com/DataDog/dd-trace-rb/pull/4755

spec/datadog/core/configuration/components_spec.rb
vpellan: Removed in this PR: https://github.com/DataDog/dd-trace-rb/pull/4755

PR #4653, opened 2025-05-14
DEBUG-3573 Send source code integration tags in telemetry and remote config

p-datadog:
**What does this PR do?**
Sends source code integration (SCI) tags in telemetry app-started event and in each remote config request.

**Motivation:**
Dynamic instrumentation UI requires SCI information in both locations to render properly.

**Change log entry**
Yes: provide source code integration data to dynamic instrumentation UI

**Additional Notes:**

**How to test the change?**

Unit tests added in this PR + system test in https://github.com/DataDog/system-tests/pull/3212


p-datadog:
System tests PR: https://github.com/DataDog/system-tests/pull/3212

lib/datadog/core/environment/git.rb
```diff
@@ -5,6 +5,7 @@
 
 module Datadog
   module Core
+    # Environment
```
y9v, line 8: do we need a comment here? This one doesn't look very useful :D

lib/datadog/core/environment/git.rb
```diff
@@ -19,6 +20,17 @@ def self.git_commit_sha
 
           @git_commit_sha = ENV[Datadog::Core::Git::Ext::ENV_COMMIT_SHA]
         end
+
+        class << self
+          private
+
+          # Resets caches in this module. Meant for internal use in the
+          # test suite only.
+          def _reset
```
y9v, line 29: I wouldn't add this method for testing purpose only. Can't we call those 2 methods directly in the test suite?

y9v (approved):

lib/datadog/core/environment/git.rb
Strech: Maybe there is a way to stub it in the tests instead? I suspect it might be tricky, but I do agree that introducing methods purely for testing purposes is not good practice.

lib/datadog/core/environment/git.rb
p-datadog: Was not my idea - rubocop demands it

lib/datadog/core/environment/git.rb
p-datadog: I would prefer not to stub in this case because then the test ends up not meaningful in my opinion. I could move the reset method in the test suite though.

p-datadog: In this case `remove_instance_variable` does not work for module_functions outside of the `self` scope so I am actually not sure how to get this code to live elsewhere.

PR #4668, opened 2025-05-19
DEBUG-3914 instance variable capture

p-datadog:
**What does this PR do?**

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/lib/datadog/di/instrumenter.rb
+++ b/lib/datadog/di/instrumenter.rb
@@ -52,6 +52,17 @@ module Datadog
     # (however, Probe instances can be replaced by OpenStruct instances
     # providing the same interface with not much effort).
     #
+    # Instrumenter (this class) is responsible for building snapshots.
+    # This is because to capture values on method entry, those values need to
+    # be duplicated or serialized into immutable values to prevent their
+    # modification by the instrumented method. Therefore this class must
+    # do at least some serialization/snapshot building and to keep the code
+    # well-encapsulated, all serialization/snapshot building should thus be
+    # initiated from this class rather than downstream code.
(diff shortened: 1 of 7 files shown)


sig/datadog/di/probe_notification_builder.rbs
```diff
@@ -16,9 +16,9 @@ module Datadog
 
       def build_errored: (Probe probe, Exception exception) -> Hash[Symbol,untyped]
 
-      def build_executed: (Probe probe, ?trace_point: TracePoint, ?rv: untyped?, ?duration: Float, ?caller_locations: Array[untyped], ?args: untyped?, ?kwargs: untyped?, ?serialized_entry_args: untyped?) -> Hash[Symbol,untyped]
+      def build_executed: (Probe probe, ?instance_vars: untyped?, ?rv: untyped?, ?duration: Float, ?caller_locations: Array[untyped], ?args: untyped?, ?kwargs: untyped?, ?serialized_entry_args: untyped?) -> Hash[Symbol,untyped]
```
Strech, line 19: You have a definition by instrumenter as `Hash[untyped, untyped]`. I would suggest to use type alias defined in Instrumenter class

Strech (approved):
Left small suggestion for type specs

sig/datadog/di/probe_notification_builder.rbs
p-datadog: I agree in principle, however I stopped writing out types because I found that the type checking is not identifying any issues in the code and updating sig files is quite a large effort.

PR #4671, opened 2025-05-20
DEBUG-3573 split telemetry event classes into own files

p-datadog:
**What does this PR do?**

Splits telemetry event classes to be in one file per class

**Motivation:**
Dynamic instrumentation requires app started or configuration change events to accurately reflect the state of dynamic instrumentation in the tracer. Currently only app started event is sent, and if dynamic instrumentation is programmatically enabled via `Datadog.configure` this is not reported to telemetry, resulting in broken DI UI.

Repairing this requires hacking the app started and configuration change events more which would be more readable if each event was defined in its own file.

**Change log entry**
None

**Additional Notes:**
I generated type definitions and steep started complaining about MessageBatch event not deriving from Base, therefore I derived it from Base in this PR.

**How to test the change?**
Existing CI


lib/datadog/core/telemetry/event/app_client_configuration_change.rb
```diff
@@ -0,0 +1,65 @@
...
+            end
+
+            res
+          end
+
+          def ==(other)
+            other.is_a?(AppClientConfigurationChange) && other.changes == @changes && other.origin == @origin
+          end
+
+          alias eql? ==
```
y9v, line 56: I think this alias is defined in the `Event::Base` class already

y9v (approved):

marcotc (approved):

lib/datadog/core/telemetry/event/app_dependencies_loaded.rb
```diff
@@ -0,0 +1,34 @@
...
+          end
+
+          private
+
+          def dependencies
+            Gem.loaded_specs.collect do |name, gem|
+              {
+                name: name,
+                version: gem.version.to_s,
+                # hash: nil,
```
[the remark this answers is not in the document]
p-datadog: :+1:

lib/datadog/core/telemetry/event/app_client_configuration_change.rb
y9v: wow, nice to know, thanks!

PR #4699, opened 2025-06-02
[APPSEC-57738] Add API Security sampler

Strech:
**What does this PR do?**

Add API security sampling logic required for [API Security RFC](https://docs.google.com/document/d/1PYoHms9PPXR8V_5_T5-KXAhoFDKQYA8mTnmS12xkGOE/edit?pli=1&tab=t.0)

**Motivation:**

[API Security RFC](https://docs.google.com/document/d/1PYoHms9PPXR8V_5_T5-KXAhoFDKQYA8mTnmS12xkGOE/edit?pli=1&tab=t.0)

**Change log entry**

Yes. AppSec: Add API Security to the `rack`, `rails`, `sinatra` and `grape` contribs.

**Additional Notes:**

I have adjusted LRU cache to give it `store` method.

**How to test the change?**

We have ST + CI to cover 99.9%


lib/datadog/appsec/api_security/sampler.rb
```diff
@@ -0,0 +1,59 @@
...
+  module AppSec
+    module APISecurity
+      # A thread-local sampler for API security based on defined delay between
+      # samples with caching capability.
+      class Sampler
+        THREAD_KEY = :datadog_appsec_api_security_sampler
+        MAX_CACHE_SIZE = 4096
+
+        class << self
+          def thread_local
```
y9v, line 18: I really like this method name

spec/datadog/appsec/api_security/route_extractor_spec.rb
```diff
@@ -0,0 +1,153 @@
...
+
+      context 'when route has nested path' do
+        before { allow(request).to receive(:path).and_return('/some/other/path') }
+
+        it { expect(described_class.route_pattern(request)).to eq('/some/other/path') }
+      end
+    end
+
+    context 'when multiple framework routes are present' do
+      context 'when Sinatra and Grape routes are present' do
```
y9v, line 122: what's the idea behind it? Always return sinatra route first?

y9v (approved):

spec/datadog/appsec/api_security/route_extractor_spec.rb
Strech: There is not way we could understand what is the end route the request going to land (from environment of the request). This is a guesstimated order, rails last and the small mountable frameworks first.

y9v: I think that's a good idea, but I would put Grape before Sinatra, just because I thought that there is a bigger chance of people mounting Grape inside Sinatra and not the other way around.

Strech: Good point, will do

PR #4715, opened 2025-06-05
Refactor system test update job

TonyCTHsu:
**What does this PR do?**

1. Ruby -> Bash
2. Sign commit 

**Motivation:**

Set example for other language adoption.

**Change log entry**
None.


.github/scripts/update_reference.sh
```diff
@@ -0,0 +1,52 @@
...
+    if cmp -s "$TARGET" "$TEMP_FILE"; then
+        echo "No references found in $TARGET"
+    else
+        # Copy the temp file back to the target
+        cp "$TEMP_FILE" "$TARGET"
+        echo "✓ Updated references in $TARGET"
+    fi
+else
+    echo "Error: Target file $TARGET does not exist"
+    rm -f "$TEMP_FILE"
```
p-datadog: You can use trap to invoke the cleanup once: https://bashcommands.com/bash-trap

PR #4716, opened 2025-06-05
Add Baggage to Span Tags

rachelyangdog:
**What does this PR do?**

This PR introduces automatic propagation of baggage key-value pairs as span tags, prefixed with baggage. The new configuration option DD_TRACE_BAGGAGE_TAG_KEYS allows customization of this feature. See the related [RFC](https://docs.google.com/document/d/1ZsgAcQHKjaBxDmhHup0zqFz07WYiOrbBksHPoXJe6u4/edit?tab=t.0#heading=h.q69qzmye8dfb) for more details.

**Motivation:**

**Change log entry**

Yes. [W3C Baggage](https://www.w3.org/TR/baggage/#example) headers can now be added to the local root span as span tags. The allowed default headers are `user.id`,`session.id`, and `account.id`.

**Additional Notes:**

**How to test the change?**


lib/datadog/tracing/distributed/propagation.rb
```diff
@@ -165,6 +166,34 @@ def propagate_baggage(data, extracted_trace_digest)
           end
         end
 
+        # Apply baggage key-value pairs as span tags
+        # respecting the DD_TRACE_BAGGAGE_TAG_KEYS configuration.
+        def apply_baggage_to_span!(baggage_hash)
+          return if baggage_hash.nil? || baggage_hash.empty?
+
+          # Fetch configuration: which keys to tag
+          raw_config = ENV[ENV_BAGGAGE_SPAN_TAGS] # is this how I get the configuration?
```
marcotc: You receive the value for `baggage_span_tags` in the `initialize` method of this class, so it doesn't have to be resolve every time. The caller then will provide the value.

lib/datadog/tracing/distributed/propagation.rb
```diff
@@ -165,6 +166,34 @@ def propagate_baggage(data, extracted_trace_digest)
...
+              Set.new(%w[user.id session.id account.id])
+            elsif raw_config.strip == ''
+              # Explicitly disabled
+              return
+            elsif raw_config.strip == '*'
+              # Wildcard: allow all keys
+              :all
+            else
+              Set.new(raw_config.split(',').map(&:strip))
+            end
```
marcotc: This kind of configuration processing should be done at the setting side, so it only needs to be done once. Take a look at `after_set` to see how you can manipulate the setting at declaration time: https://github.com/DataDog/dd-trace-rb/blob/3595efceb2168b30cfa4d914f4cd5683ef8a5a1d/lib/datadog/tracing/configuration/settings.rb#L98

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -170,6 +170,36 @@ def self.extended(base)
...
+              #
+              # Special values:
+              # * Empty string ("") disables baggage tag conversion
+              # * Wildcard ("*") converts all baggage keys to span tags
+              #
+              # @default `DD_TRACE_BAGGAGE_TAG_KEYS` environment variable, otherwise `"user.id,session.id,account.id"`
+              # @return [Array<String>, Symbol, nil] processed baggage tag keys configuration
+              option :baggage_tag_keys do |o|
+                o.env Configuration::Ext::ENV_BAGGAGE_TAG_KEYS
+                o.type :string
```
marcotc:
```suggestion
                o.type :array
```

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -170,6 +170,36 @@ def self.extended(base)
...
+                o.default 'user.id,session.id,account.id'
```
marcotc:
```suggestion
                o.default ['user.id','session.id','account.id']
```

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -170,6 +170,36 @@ def self.extended(base)
...
+                o.after_set do |value|
```
marcotc:
```suggestion
                o.after_set do |tag_keys|
```

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -125,6 +130,33 @@ def parse_baggage_header(baggage_header)
...
+          baggage_tag_keys = ::Datadog.configuration.tracing.baggage_tag_keys
+          return {} if baggage_tag_keys.empty?
+
+          # If wildcard is specified, use all baggage keys
+          baggage_tag_keys = baggage.keys if baggage_tag_keys == BAGGAGE_TAG_KEYS_MATCH_ALL
+
+          tags = {}
+
+          # Convert configured keys to span tags
+          baggage_tag_keys.each do |key|
```
marcotc:
```suggestion
          baggage_tag_keys = baggage if baggage_tag_keys == BAGGAGE_TAG_KEYS_MATCH_ALL

          tags = {}

          # Convert configured keys to span tags
          baggage_tag_keys.each do |key, _|
```

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -125,6 +130,33 @@ def parse_baggage_header(baggage_header)
...
+            next unless baggage.key?(key)
+
+            value = baggage[key]
```
marcotc: You don't need to check for the presence of the key (with `key?`) because accessing a hash is always safe in Ruby. If the key is not there, it will return `nil`, which you are checking for anyway in the next line.

```suggestion
            value = baggage[key]
```

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -125,6 +130,33 @@ def parse_baggage_header(baggage_header)
...
           baggage
         end
+
+        # Convert baggage items to trace_distributed_tags with 'baggage.' prefix
+        # based on the baggage_tag_keys configuration
+        def build_baggage_tags(baggage)
+          return {} if baggage.empty?
+
+          # Get the configuration for which baggage keys should become span tags
+          baggage_tag_keys = ::Datadog.configuration.tracing.baggage_tag_keys
```
marcotc: Let's move this to the `initializer` method instead, and cache the result:
```suggestion
          baggage_tag_keys = @baggage_tag_keys
```

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -125,6 +130,33 @@ def parse_baggage_header(baggage_header)
           end
           baggage
         end
+
+        # Convert baggage items to trace_distributed_tags with 'baggage.' prefix
+        # based on the baggage_tag_keys configuration
```
marcotc: Can you change this documentation so that it tells us the "why" for the reason we are generating span tags from baggage values?

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -125,6 +130,33 @@ def parse_baggage_header(baggage_header)
...
+          # Get the configuration for which baggage keys should become span tags
+          baggage_tag_keys = ::Datadog.configuration.tracing.baggage_tag_keys
+          return {} if baggage_tag_keys.empty?
+
+          # If wildcard is specified, use all baggage keys
+          baggage_tag_keys = baggage.keys if baggage_tag_keys == BAGGAGE_TAG_KEYS_MATCH_ALL
+
+          tags = {}
+
+          # Convert configured keys to span tags
```
marcotc: This comment can be removed, because it's simply reading the code below, and it doesn't add value not already present in the code.
```suggestion
```

marcotc (approved):

docs/GettingStarted.md
```diff
@@ -2135,6 +2135,7 @@ For example, if `tracing.sampling.default_rate` is configured by [Remote Configu
...
+| `tracing.baggage_tag_keys`                             | `DD_TRACE_BAGGAGE_TAG_KEYS`                             | `Array<String>`                       | `['user.id', 'session.id', 'account.id']` | Comma-separated list of baggage keys that should be converted to span tags. Baggage keys matching the exact, case-sensitive names in this list will be converted to span tags with the prefix "baggage.". Special values: Empty string ("") disables baggage tag conversion, Wildcard ("*") converts all baggage keys to span tags.                                                                                                                                                                                                                                                                                                   |
```
drichards-87:
```suggestion
| `tracing.baggage_tag_keys`                             | `DD_TRACE_BAGGAGE_TAG_KEYS`                             | `Array<String>`                       | `['user.id', 'session.id', 'account.id']` | Comma-separated list of baggage keys that should be converted to span tags. Baggage keys matching the exact, case-sensitive names in this list are converted to span tags with the prefix "baggage.". Special values: Empty string ("") disables baggage tag conversion, and Wildcard ("*") converts all baggage keys to span tags.                                                                                                                                                                                                                                                                                                   |
```

PR #4718, opened 2025-06-06
feat: add config DD_RUNTIME_METRICS_RUNTIME_ID_ENABLED

ZStriker19:
**What does this PR do?**
Add the additional envar DD_RUNTIME_METRICS_RUNTIME_ID_ENABLED to enable the config runtime_metrics.experimental_runtime_id_enabled as we want to move this out of being an experimental feature.

**Motivation:**
A request from Anna Yafi for her work.

**Change log entry**
- Adds the environment variable ``DD_RUNTIME_METRICS_RUNTIME_ID_ENABLED`` to enable runtime metrics for tagging runtime metrics with the current runtime ID. This is useful for tracking runtime metrics across multiple processes. Previously, this was ``DD_TRACE_EXPERIMENTAL_RUNTIME_ID_ENABLED``.

**Additional Notes:**

**How to test the change?**


jhgilbert (approved):

ivoanjo:
Definitely don't want to bike shed so I'll make it clear I'm offering feedback and I leave the decision up to you, and that disagreeing with my feedback is very ok!

Having said that, `DD_TRACE_RUNTIME_ID_ENABLED` seems a bit cryptic for what this does? E.g. perhaps `DD_TRACE_METRICS_BY_RUNTIME_ID_ENABLED` or `DD_TRACE_RUNTIME_METRICS_BY_ID_ENABLED` or something like that would be more descriptive of what this feature does?

Using github search I spotted https://github.com/DataDog/dd-trace-py/pull/13618 but didn't see any library release with `DD_TRACE_RUNTIME_ID_ENABLED` so I think now is a great opportunity to still tweak the naming, if we wish to; after releasing this it'll be a lot more awkward to change.

ZStriker19:
> Looks good! Can you please update the test unit to test for this new environment variable? 🙏

@marcotc  where were we using DD_TRACE_EXPERIMENTAL_RUNTIME_ID_ENABLED before? I checked the repo for the envar originally and couldn't find anywhere that it's actually being tested? It looks like instead we just test the config experimental_runtime_id_enabled, and since we just added an additional envar to set the config, I think it should be fine?

marcotc:
@ZStriker19 I recommend using an example that has two environment variables today, like the `debug` option. I think it's worth adding both to this PR: https://github.com/DataDog/dd-trace-rb/blob/2646e788c3406177a00db5f0c048952f53acdf22/spec/datadog/core/configuration/settings_spec.rb#L59

marcotc:
> Definitely don't want to bike shed so I'll make it clear I'm offering feedback and I leave the decision up to you, and that disagreeing with my feedback is very ok!
> 
> Having said that, `DD_TRACE_RUNTIME_ID_ENABLED` seems a bit cryptic for what this does? E.g. perhaps `DD_TRACE_METRICS_BY_RUNTIME_ID_ENABLED` or `DD_TRACE_RUNTIME_METRICS_BY_ID_ENABLED` or something like that would be more descriptive of what this feature does?
> 
> Using github search I spotted [DataDog/dd-trace-py#13618](https://github.com/DataDog/dd-trace-py/pull/13618) but didn't see any library release with `DD_TRACE_RUNTIME_ID_ENABLED` so I think now is a great opportunity to still tweak the naming, if we wish to; after releasing this it'll be a lot more awkward to change.

@ZStriker19 I couldn't find docs for `DD_TRACE_RUNTIME_ID_ENABLED`. If it's not set in stone, given we already use `DD_RUNTIME_METRICS_ENABLED` today. I suggest `DD_RUNTIME_METRICS_RUNTIME_ID_ENABLED` or the simpler `DD_RUNTIME_METRICS_RUNTIME_ID`. But regardless of the choice, I agree with @ivoanjo that `DD_TRACE_RUNTIME_ID_ENABLED` seems like it's disabling the global `runtime-id` span tag.

anna-git:
@ivoanjo @marcotc 
`DD_TRACE_RUNTIME_ID_ENABLED` cannot be changed as it's used by node js and python tracers as well. It also appears to be used by some python and nodejs organizations in metabase, other customers have been using `DD_TRACE_EXPERIMENTAL_RUNTIME_ID_ENABLED` and wouldn't understand it's changing names. 
We are also about to change the official docs mentioning runtime-id granularity to mention and explain more clearly what `DD_TRACE_RUNTIME_ID_ENABLED` does.

ivoanjo:
> DD_TRACE_RUNTIME_ID_ENABLED cannot be changed as it's used by node js and python tracers as well.

Hmm... I might've missed something? The Python support still seems to be in PR https://github.com/DataDog/dd-trace-py/pull/13618 , and [node still seems to be using experimental](https://github.com/search?q=repo%3ADataDog%2Fdd-trace-js%20DD_TRACE_EXPERIMENTAL_RUNTIME_ID_ENABLED&type=code) with no [references I could find to the non-experimental variant](https://github.com/search?q=repo%3ADataDog%2Fdd-trace-js+DD_TRACE_RUNTIME_ID_ENABLED&type=code).

This is why I mentioned that this seems to be the last good opportunity we have to change our minds on the name before it becomes really awkward to change :wink:

anna-git:
@ivoanjo sorry I meant it's used with the experimental keyword in other repositories. And I think renaming an env variable with the experimental keyword in it with a different name might confuse our customers, especially that some are already using it. 

But I understand the idea of wanting to indicate it's only about a different runtime metrics mode and not a feature _per se_. 
However, it doesn't seem that trivial as it's mapped in telemetry with "DD_TRACE_EXPERIMENTAL_RUNTIME_ID_ENABLED": "runtime_id_enabled"` which would imply renaming the telemetry key too as otherwise, it wouldn't make much sense having a key with a different telemetry name. 
As we're changing the official docs together to document the env var) , I add  @bm1549 in the loop, as he might have a different opinion on this.

bm1549:
Thanks for the ping

The telemetry name here is internal, so it can always be changed. If we don't want to, we can also remap newer env vars to older configs if we're okay with having them inconsistent. Our internal telemetry also shows that only single-digit users set the config today (it's currently undocumented) so we're good to move it over, ideally as an alias for the older name

As for the name of the env var itself, I can see the motivation for not using `DD_TRACE_RUNTIME_ID_ENABLED` here as `runtime-id` is also a part of our span data model. As such, I'd also lean towards `DD_RUNTIME_METRICS_RUNTIME_ID_ENABLED` (using the `_ENABLED` suffix to denote a boolean field, as per convention)

ivoanjo:
Thanks y'all for the input! I think the renaming would be a nice to have, but to simplify the decision I leave it to y'all in APM to decide what naming to pick and if you want to go through with the renaming or not :)

anna-git:
Ok so after discussion, it was decided to rename it to `DD_RUNTIME_METRICS_RUNTIME_ID_ENABLED`  do you think it'd be possible @ZStriker19 ? 🙇

ivoanjo:
(P.s. may be worth updating the PR name and description :P . And thanks for the patience)

marcotc (approved):

PR #4721, opened 2025-06-10
[PROF-11956] Fix profiler stopping due to bug in heap profiling serialization

ivoanjo:
**What does this PR do?**

This PR fixes a rare situation where the profiler would stop during serialization of heap profiling samples with an error message:

```
WARN -- datadog: [datadog] Profiling::Scheduler thread error. Cause: RuntimeError BUG: full_update should not be triggered during another update Location: .../lib/datadog/profiling/stack_recorder.rb:63:in `_native_serialize'
```

This issue happened due to an unexpected thread switch during heap profiling metadata operations.

**Motivation:**

Fix rare heap profiling bug, which is extra important now that we're eclaring it as preview with the next release.

**Change log entry**

Yes. Fix profiler stopping due to bug in heap profiling serialization.

**Additional Notes:**

Understanding this rare bug took some amount of staring at the code and at this assertion.

Interestingly, this bug was not caused by the big refactoring in https://github.com/DataDog/dd-trace-rb/pull/4460 ; in fact, it's very closer in spirit to https://github.com/DataDog/dd-trace-rb/pull/4240 .

(I believe the reason it showed up shortly after #4460 got merged is that this is a numbers game: the more objects that get tracked, the higher the probability of triggering this. Because #4460 raised the number of objects we track by 10x, it made this more likely to happen.)

What happened here is: We were assuming -- because we were inside the profiler's C code and holding the GVL when we called into it -- that the work inside `heap_recorder_update` behaved "atomically".

That is, we assumed that the whole of `heap_recorder_update` got to run, with no other threads running in-between, as we were holding the GVL.

That is, actually, not the case: during `heap_recorder_update` we call `ruby_ref_from_id` and `ruby_obj_memsize_of` both of which can, in rare situations, release the GVL and allow other threads to run for a while. (And currently there's little we could do about this constraint without a bigger redesign of the heap profiler.)

(This is why I claim this is similar to https://github.com/DataDog/dd-trace-rb/pull/4240 -- there's a potential switch point inside the profiler code that we didn't expect, although that one was worse as it introduced a switch point where there couldn't be one for VM correctness.)

Because we assumed the `heap_recorder_update` was atomic, we had this check inside there:

```c
if (heap_recorder->updating) {
  if (full_update) rb_raise(rb_eRuntimeError, "BUG: full_update should not be triggered during another update");
```

That is: We were saying -- if `heap_recorder_update` is atomic, there's no way a second update can start if there's an ongoing one, because by definition the previous one must've finished.

(Interestingly, there was already some comment about this being perhaps a possibility in non-full updates, although when I left that previous comment it didn't occur to me that I was already documenting a clash with the assumptions of this method...).

Having identified there can be concurrency inside `heap_recorder_update`, how is this PR fixing the issue? It turns out it's not as bad as it looks.

There's two sources of concurrency problems here:
1. There's 2+ threads trying to run `heap_recorder_update` concurrently (This caused the error we saw above)
2. There's a thread trying to add a sample at the same time a thread is trying to run `heap_recorder_update` (I suspect this could potentially corrupt the heap recorder state, and in fact I think it may be the source of https://datadoghq.atlassian.net/browse/PROF-10656 )

To fix 1.:
* For non-full updates to `heap_recorder_update` we already skip doing them (since they're optional) -- nothing to do
* There can't be two full updates at the same time as the `@no_concurrent_serialize_mutex` is preventing that -- nothing to do
* So the only remaining situation is when a non-full update started, and switched away to a thread that decided to do full update. The fix here is what we tell Ruby to yield from the current thread trying to do the full update, so that the thread doing the non-full update gets to run and finish its work.

To fix 2. we detect that an update is ongoing in `start_heap_allocation_recording` and skip any samples in that case.

As a final note to better understand this, remember that there is never any **parallelism** here, so we don't need any specific atomics or extra mutexes or whatever. At any point in time, all of the steps being discussed are running with the GVL acquired -- the problem here is that another thread can get to run at one of the specific potential switch points we have.

**How to test the change?**

This is a really good question. This issue has been happening almost daily in our validation environment since https://github.com/DataDog/dd-trace-rb/pull/4460 got merged, so we'll be able to see that profiler will no longer stop reporting data.

But triggering this with unit tests would be... really really hard. 
We'd need invasive surgery to force Ruby to switch in the exact specific situation we need between multiple threads in the middle of our C code.

So I couldn't think of a reasonable way of testing this, other than relying on our end-to-end tests. Suggestions are very welcome.

### DIFF
--- a/ext/datadog_profiling_native_extension/heap_recorder.c
+++ b/ext/datadog_profiling_native_extension/heap_recorder.c
@@ -310,6 +310,10 @@ void start_heap_allocation_recording(heap_recorder *heap_recorder, VALUE new_obj
       #else
         false
       #endif
+      // If we got really unlucky and an allocation showed up during an update (because it triggered an allocation
+      // directly OR because the GVL got released in the middle of an update), let's skip this sample as well.
+      // See notes on `heap_recorder_update` for details.
+      || heap_recorder->updating
     ) {
     heap_recorder->active_recording = &SKIPPED_RECORD;
     return;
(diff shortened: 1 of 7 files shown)


AlexJF (approved):
Ah! Nice investigation! The description and changes make sense to me. Bit ashamed of having missed that specific edge case, as you mentioned, the code was already protecting against the other combinations 😅

Let's see how things behave in CI with this!

ivoanjo:
:+1: I'll keep an eye out for the next few days to make sure this doesn't happen again. It was happening frequently enough before that we should clearly see when it's gone:

image

At any point in time we should see 3 different active `runtime-ids` in this environment, and you can see how as time goes by some profilers were stopping with the above error and we often were down to 1 (or even 0).

PR #4725, opened 2025-06-10
Update libddwaf-rb gem to version 1.24.1

y9v:
**What does this PR do?**
This PR updates `libddwaf-rb` to 1.24.1 and changes the way how the Remote Config changes are handled.

**Motivation:**
We want to report errors for separately for each Remote Config configuration.

**Change log entry**
Yes. Deprecate `config.appsec.ip_passlist`, `config.appsec.ip_denylist`, and `config.appsec.user_id_denylist`.

**Additional Notes:**
Only the ruby library supports configuration of allow- and denylists directly via library configuration. Other libraries only configure it using Remote Configuration. This is why we are deprecating it.

**How to test the change?**

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -163,9 +163,25 @@ deploy_to_reliability_env:
 #
 # We fetch the latest workflow runs from vaccine then match against the run name, polling until we find it.
 vaccine:
-  image: $DOCKER_REGISTRY/docker:20.10.13
+  image: registry.ddbuild.io/images/dd-octo-sts-ci-base:v68058725-73f34e7-2025.06-1
   tags: ["arch:amd64"]
   stage: vaccine
   needs: [create-multiarch-lib-injection-image]
-  script: |
-    .gitlab/scripts/vaccine.sh master "${CI_COMMIT_SHA}" "glci:${CI_PIPELINE_ID}"
+
+  id_tokens:
+    DDOCTOSTS_ID_TOKEN:
+      aud: dd-octo-sts
+
+  before_script:
+    # Get a token
+    - dd-octo-sts version
+    # - dd-octo-sts debug --scope DataDog/vaccine --policy dd-trace-rb.dispatch # Uncomment to debug
+    - dd-octo-sts token --scope DataDog/vaccine --policy dd-trace-rb.dispatch > token.txt
+
+  script:
+    - export GITHUB_TOKEN=$(cat token.txt)
+    - .gitlab/scripts/vaccine.sh master "${CI_COMMIT_SHA}" "glci:${CI_PIPELINE_ID}"
+
+  after_script:
+    # Revoke the token after usage
+    - dd-octo-sts revoke -t $(cat token.txt)
--- a/.gitlab/scripts/vaccine.sh
+++ b/.gitlab/scripts/vaccine.sh
@@ -70,9 +70,13 @@ log_and_run() {
     run "${cmd[@]}"
 }
 
-# obtain GitHub vaccine token from vault
+# obtain GitHub token from environment variable
 get_vaccine_token() {
-    vault kv get -field=vaccine-token kv/k8s/gitlab-runner/dd-trace-rb/github-token
+    if [[ -z "${GITHUB_TOKEN:-}" ]]; then
+        echo "Error: GITHUB_TOKEN environment variable is not set" >&2
+        exit 1
+    fi
+    echo "${GITHUB_TOKEN}"
 }
 
 # get current commit SHA


lib/datadog/appsec/configuration/settings.rb
```diff
@@ -80,16 +80,46 @@ def self.add_settings!(base)
 
               option :ip_passlist do |o|
                 o.default []
+
+                o.setter do |value|
+                  next value if value.nil? || value.empty?
+
+                  Datadog::Core.log_deprecation(disallowed_next_major: false) do
+                    'ip_passlist setting is deprecated and will be removed in the next release'
```
Strech: A small note, I think it makes sense to add a phrase what should you do if you were using those settings, like - please use rules file, etc.

lib/datadog/appsec/remote.rb
```diff
@@ -57,74 +56,35 @@ def products
...
+              next unless content
 
-            if rules.empty?
-              settings_rules = AppSec::Processor::RuleLoader.load_rules(
-                telemetry: telemetry,
-                ruleset: Datadog.configuration.appsec.ruleset
-              )
+              case change.type
+              when :insert, :update
+                AppSec.security_engine&.add_or_update_config(
```
Strech: Question, if security_engine is not in place (somehow), should we call reconfigure at the end and mark content as applied?

It feels like we might get into the situation, when there is no security engine, but we consumer changes and mark them as processed, when they are not. Or maybe I don't get something ...

lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -0,0 +1,188 @@
...
+          exclusion_data
+        ].freeze
+
+        attr_reader :waf_addresses, :ruleset_version
+
+        def initialize(appsec_settings:, telemetry:)
+          @default_ruleset = appsec_settings.ruleset
+          @default_ip_denylist = appsec_settings.ip_denylist
+          @default_user_id_denylist = appsec_settings.user_id_denylist
+          @default_ip_passlist = appsec_settings.ip_passlist
```
Strech, line 31: Should we leave a note that those parts will be deprecated, right? Or a comment why we pass AppSec settings here, because ideally we should not pass it and instead give it some sort of independent configuration options.

lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -0,0 +1,188 @@
...
+            obfuscator: {
+              key_regex: appsec_settings.obfuscator_key_regex,
+              value_regex: appsec_settings.obfuscator_value_regex
+            }
+          )
+
+          load_default_config(telemetry: telemetry, action: :init)
+
+          @waf_handle = @waf_builder.build_handle
+          @waf_addresses = @waf_handle.known_addresses
```
Strech, line 44: Does it make sense to make it lazy and populate only on demand?

lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -0,0 +1,188 @@
...
+
+          @waf_addresses = []
+          @ruleset_version = nil
+        end
+
+        def new_runner
+          SecurityEngine::Runner.new(@waf_handle.build_context)
+        end
+
+        def add_or_update_config(config:, path:)
```
Strech: Since config is the main argument, it should not be kwarg because name suggests it.

```suggestion
        def add_or_update_config(config, path:)
```

lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -0,0 +1,188 @@
...
+          @is_ruleset_update = path.include?('ASM_DD')
+
+          # default config has to be removed when adding an ASM_DD config
+          remove_config_at_path(DEFAULT_RULES_CONFIG_PATH) if @is_ruleset_update
+
+          @waf_builder.add_or_update_config(config, path: path).tap do |diagnostics|
```
Strech: Question, do we really need to open result context via block or normal assignment will work? Because we saving nothing here, it's not shorter and more expensive to do.

**[not blocker]**

```ruby
@waf_builder.add_or_update_config(config, path: path).tap do |diagnostics|
  @ruleset_version = diagnostics['ruleset_version'] if diagnostics.key?('ruleset_version')
end

# vs

diagnostics = @waf_builder.add_or_update_config(config, path: path)
@ruleset_version = diagnostics['ruleset_version'] if diagnostics.key?('ruleset_version')

diagnostics
```

lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -0,0 +1,188 @@
...
+            @ruleset_version = diagnostics['ruleset_version'] if diagnostics.key?('ruleset_version')
```
Strech: This will be same result

```suggestion
            @ruleset_version = diagnostics['ruleset_version']
```

lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -0,0 +1,188 @@
...
+          end
+        end
+
+        def report_config_errors_via_telemetry(diagnostics, action:, telemetry:)
+          raise ArgumentError, 'action must be one of TELEMETRY_ACTIONS' unless TELEMETRY_ACTIONS.include?(action)
+
+          common_tags = {
+            waf_version: Datadog::AppSec::WAF::VERSION::BASE_STRING,
+            event_rules_version: diagnostics.fetch('ruleset_version', @ruleset_version).to_s,
+            action: action.to_s
```
Strech: why not string from the beginning?

lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -0,0 +1,188 @@
...
+          config['exclusions'] ||= AppSec::Processor::RuleLoader.load_exclusions(ip_passlist: @default_ip_passlist)
+
+          @waf_builder.add_or_update_config(config, path: DEFAULT_RULES_CONFIG_PATH).tap do |diagnostics|
+            @ruleset_version = diagnostics['ruleset_version']
+
+            report_config_errors_via_telemetry(diagnostics, action: action, telemetry: telemetry)
+          end
+        end
+
+        def report_config_errors_via_telemetry(diagnostics, action:, telemetry:)
```
Strech: I think we should just use `AppSec.telemetry` here instead of tossing it. We are going to move it later to the SecurityEngine level and use it same way.

Also we could hide implementation details from name

```suggestion
        def report_configuration_diagnostics(diagnostics, action:)
```

lib/datadog/appsec/security_engine/runner.rb
```diff
@@ -56,18 +55,18 @@ def run(persistent_data, ephemeral_data, timeout = WAF::LibDDWAF::DDWAF_RUN_TIME
         end
 
         def finalize
```
Strech: As we go bang, should we add it here too?

```suggestion
        def finalize!
```

spec/datadog/appsec/component_spec.rb
```diff
@@ -3,14 +3,14 @@
 
 RSpec.describe Datadog::AppSec::Component do
   let(:telemetry) { instance_double(Datadog::Core::Telemetry::Component) }
+  let(:settings) do
+    settings = Datadog::Core::Configuration::Settings.new
+    settings.appsec.enabled = appsec_enabled
+    settings
+  end
+  let(:appsec_enabled) { true }
```
Strech: I would like to suggest here a simple case with before block instead

```ruby
let(:settings) { Datadog::Core::Configuration::Settings.new }
```

and then in places where appsec is enabled/disabled

```ruby
before { settings.appsec.enabled = true }
```

I would consider any primitive in the `let` statement as a smell.

TIP: In the next test files `tap` was used

```ruby
Datadog::Core::Configuration::Settings.new.tap do |settings|
  settings.appsec.enabled = true
end
```

spec/datadog/appsec/component_spec.rb
```diff
@@ -146,67 +127,10 @@
     context 'lock' do
       it 'makes sure to synchronize' do
```
Strech: To be honest this test doesn't make sense

spec/datadog/appsec/component_spec.rb
```diff
@@ -215,8 +139,7 @@
...
```
Strech: Same here - what are we testing here?

lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -0,0 +1,188 @@
...
+          processors
+          scanners
+          rules_override
+          rules_data
+          exclusion_data
+        ].freeze
+
+        attr_reader :waf_addresses, :ruleset_version
+
+        def initialize(appsec_settings:, telemetry:)
```
Strech, line 25: After checking other places, my suggestion is to remove `telemetry:` and just use `AppSec.telemetry`. Current use case is quite gimmicky and brings nothing but confusion.

spec/datadog/appsec/configuration/settings_spec.rb
```diff
@@ -211,6 +211,37 @@ def patcher
       end
     end
 
+    describe '#ip_passlist' do
+      subject(:ip_passlist) { settings.appsec.ip_passlist }
+
+      context 'default value' do
+        it { is_expected.to eq [] }
+      end
```
Strech:
```suggestion
      it 'returns default value' do
        expect(settings.appsec.ip_passlist).to eq []
      end
```

spec/datadog/appsec/configuration/settings_spec.rb
```diff
@@ -211,6 +211,37 @@ def patcher
...
+          expect { settings.appsec.ip_passlist = [] }.not_to log_deprecation
+        end
+
+        it 'logs a deprecation warning when set to a non-default value' do
+          expect { set_appsec_ip_passlist }.to log_deprecation(
+            'ip_passlist setting is deprecated and will be removed in the next release'
+          )
+        end
+      end
+    end
```
Strech: My suggestion is to avoid primitives in `let` statements

spec/datadog/appsec/configuration/settings_spec.rb
```diff
@@ -225,9 +256,20 @@ def patcher
...
+        it 'does not log a deprecation warning when set to default value' do
+          expect { settings.appsec.ip_denylist = [] }.not_to log_deprecation
+        end
 
-        it { expect(settings.appsec.ip_denylist).to eq(['1.1.1.1']) }
+        it 'logs a deprecation warning when set to a non-default value' do
+          expect { set_appsec_ip_denylist }.to log_deprecation(
+            'ip_denylist setting is deprecated and will be removed in the next release'
+          )
+        end
```
Strech: same here

spec/datadog/kit/identity_spec.rb
```diff
@@ -212,20 +212,30 @@
...
+      let(:settings) do
+        Datadog::Core::Configuration::Settings.new.tap do |settings|
+          settings.appsec.enabled = true
+        end
+      end
+      let(:security_engine) do
+        Datadog::AppSec::SecurityEngine::Engine.new(appsec_settings: settings.appsec, telemetry: telemetry)
+      end
+      let(:waf_runner) { security_engine.new_runner }
+      let(:appsec_span) { trace_op.build_span('root') }
```
Strech: I think it's not used

spec/datadog/kit/identity_spec.rb
```diff
@@ -212,20 +212,30 @@
...
+      let(:appsec_active_context) { nil }
+
       before do
-        allow(processor).to receive(:new_runner).and_return(instance_double(Datadog::AppSec::SecurityEngine::Runner))
+        Datadog::AppSec::Component.build_appsec_component(settings, telemetry: telemetry)
```
Strech: Do we really need all of this? I think here the usage is stubbed, not sure that it's worth it

spec/datadog/kit/appsec/events_spec.rb
```diff
@@ -9,16 +9,25 @@
...
+    end
+    let(:security_engine) do
+      Datadog::AppSec::SecurityEngine::Engine.new(appsec_settings: settings.appsec, telemetry: telemetry)
+    end
+    let(:waf_runner) { security_engine.new_runner }
+    let(:appsec_span) { trace_op.build_span('root') }
+
     before do
-      allow(processor).to receive(:new_runner).and_return(instance_double(Datadog::AppSec::SecurityEngine::Runner))
+      Datadog::AppSec::Component.build_appsec_component(settings, telemetry: telemetry)
```
Strech: Same question here, do we need all of that?

spec/datadog/appsec/security_engine/runner_spec.rb
```diff
@@ -1,25 +1,16 @@
...
 
 RSpec.describe Datadog::AppSec::SecurityEngine::Runner do
   before do
-    # NOTE: This is an intermediate step and will be removed
-    rules = Datadog::AppSec::Processor::RuleLoader.load_rules(ruleset: :recommended, telemetry: telemetry)
-    ruleset = Datadog::AppSec::Processor::RuleMerger.merge(rules: [rules], telemetry: telemetry)
-    Datadog::AppSec::Processor.new(ruleset: ruleset, telemetry: telemetry)
-
-    allow(Datadog::AppSec::WAF::Context).to receive(:new).and_return(waf_context)
+    require 'libddwaf'
```
Strech: should it be global outside of the top describe?

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -0,0 +1,597 @@
...
+  let(:telemetry) { instance_double(Datadog::Core::Telemetry::Component) }
+
+  let(:appsec_settings) do
+    settings = Datadog::Core::Configuration::Settings.new
+    settings.appsec.enabled = true
+    settings.appsec
+  end
+
+  before do
+    require 'libddwaf'
```
Strech: why in before block?

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -0,0 +1,597 @@
...
+    let(:default_ruleset) do
+      Datadog::AppSec::Processor::RuleLoader.load_rules(telemetry: telemetry, ruleset: appsec_settings.ruleset)
+    end
+
+    it 'sets waf_addresses' do
+      expect(engine.waf_addresses).to_not be_empty
+    end
+
+    it 'sets ruleset_version' do
+      expect(engine.ruleset_version).to eq(default_ruleset.dig('metadata', 'rules_version'))
```
Strech: I think we should supply hand-crafted rulset, otherwise it doesn't make much sense

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -0,0 +1,597 @@
...
+              id: 'invalid-rule-id'
+            }
+          ]
+        }
+      end
+
+      before do
+        appsec_settings.ruleset = invalid_ruleset
+
+        allow(telemetry).to receive(:inc)
```
Strech: Will be there more errors than expected?

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -0,0 +1,597 @@
...
+
+        expect { engine }.to raise_error(Datadog::AppSec::WAF::LibDDWAFError)
+      end
+    end
+  end
+
+  describe '#finalize!' do
+    it 'finalizes waf builder and waf handle' do
+      expect(engine.instance_variable_get(:@waf_builder)).to receive(:finalize!)
+      expect(engine.instance_variable_get(:@waf_handle)).to receive(:finalize!)
```
Strech: I think it's a wrong angle of testing. The main point is to test behavior, this one is testing exact implementation

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -0,0 +1,597 @@
...
+            id: 'test-custom-rule-id',
+            name: 'Test rule',
+            tags: {category: 'attack_attempt', custom: '1', type: 'custom'},
+            transformers: []
+          }
+        ]
+      }
+    end
+
+    let(:config_path) { 'datadog/603646/ASM/test-custom-rule' }
```
Strech: I think it should be inlined

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -0,0 +1,597 @@
...
+      it 'reports item-level errors through telemetry' do
+        expect(telemetry).to receive(:error).with("missing key 'conditions': [invalid-rule-one-id]")
+        expect(telemetry).to receive(:error).with("missing key 'parameters': [invalid-rule-two-id]")
+
+        engine.add_or_update_config(config: config_with_invalid_rule, path: config_path)
+      end
+    end
+
+    context 'when config loading fails with top-level error' do
+      let(:invalid_config) { '' }
```
Strech: ⬆️

Strech (commented):
Nice to see a huge improvement over existing implementation.

I left few questions about security engine, and a lot of points for tests, I think tests could be improved.

lib/datadog/appsec/security_engine/engine.rb
y9v: well, this way we can set `@ruleset_version` to `nil`, if this key is not present in diagnostics, no? And we only want to change the `@ruleset_version` if it is present in diagnostics - since this only happens when receiving the ruleset.

spec/datadog/appsec/component_spec.rb
y9v: totally agree, it was already there and I thought of dropping it, but didn't

spec/datadog/appsec/component_spec.rb
y9v: good question - I will drop it too

lib/datadog/appsec/configuration/settings.rb
y9v: [3d2abf7](https://github.com/DataDog/dd-trace-rb/pull/4725/commits/3d2abf71346cf0153af508f41d10558f0d33a4d8)

lib/datadog/appsec/remote.rb
y9v: that's a really good catch, thanks! I will add a guard-clause to the function.

y9v: fixed here: [b3db9b0](https://github.com/DataDog/dd-trace-rb/pull/4725/commits/b3db9b0f8996f04b6094bf559d9025c37bdb77d0)

lib/datadog/appsec/security_engine/engine.rb
y9v: good point, added a note [e5ef213](https://github.com/DataDog/dd-trace-rb/pull/4725/commits/e5ef213d29f17ae5e1dd3c4ff6103c15c949219f)

lib/datadog/appsec/security_engine/engine.rb
y9v: Tbh I don't know what is better - lazy method and invalidating cache on `reconfigure!`, or explicitly setting the instance variable when creating a new handle.
One argument I had for an instance variable - we will not have a slight spike in the response time of the first request that adds WAF addresses tag to a span. For that we will have a slightly slower reconfiguration.

lib/datadog/appsec/security_engine/engine.rb
y9v: agree, fixed here: [e7ee6c2](https://github.com/DataDog/dd-trace-rb/pull/4725/commits/e7ee6c232fb6e33c075c758ec17f9a58b35d3239)

lib/datadog/appsec/configuration/settings.rb
y9v: [8fb41e9](https://github.com/DataDog/dd-trace-rb/pull/4725/commits/8fb41e96bfb7aa9fb4c0d001df5ae562b2ca6fa7)

lib/datadog/appsec/security_engine/engine.rb
y9v: [1bb8ea9](https://github.com/DataDog/dd-trace-rb/pull/4725/commits/1bb8ea9a68e5eb82ab70bbb7b50359c34075423f)

lib/datadog/appsec/security_engine/engine.rb
y9v: fixed [d4ab19c](https://github.com/DataDog/dd-trace-rb/pull/4725/commits/d4ab19c5fc20d185eb34879242308ab4e9657fd4)

spec/datadog/appsec/security_engine/engine_spec.rb
y9v: yes, I agree. I thought about alternative approaches here - one would be to test if the handle was actually finalized by calling something on it and expecting an exception. WDYT?

spec/datadog/kit/appsec/events_spec.rb
y9v: it was a dumb thing - to require 'libddwaf'. I will add an explicit require.

spec/datadog/appsec/security_engine/engine_spec.rb
y9v: it is used in many places in the same `describe` block - I simply wouldn't want to repeat this string everywhere

spec/datadog/appsec/security_engine/engine_spec.rb
y9v: [c2b3f68](https://github.com/DataDog/dd-trace-rb/pull/4725/commits/c2b3f6825a3037d4a1cc38663ad86f994ff70b3a)

spec/datadog/appsec/security_engine/engine_spec.rb
y9v: why doesn't it make much sense? We are testing that we are loading the ruleset that is specified in `appsec_settings.ruleset` and set the version from it.

lib/datadog/appsec/security_engine/engine.rb
y9v: we can't use `AppSec.telemetry` before the component has fully initialized, otherwise we will get a `ThreadError: deadlock`. This is because `AppSec.telemetry` and other methods on `AppSec` are doing `components.appsec&` - which causes components to initialize again.
I would love to change this, but not in this PR.

lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -0,0 +1,194 @@
...
+            Datadog.logger.warn("Reverting to the previous configuration")
+
+            @waf_handle = old_waf_handle
+            @waf_addresses = old_waf_handle.known_addresses
+          end
+        end
+
+        private
+
+        def load_default_config(telemetry:, action:)
```
Strech: 1. I think telemetry is not the main argument here

```suggestion
        def load_default_config(action:, telemetry:)
```

2. After looking again into that method I think I know why `:action` doesn't fit here (like why we pass action which is irrelevant to the loading at all), because this method violates SRP. IMHO it should be a split of - loading as a method and reporting as a method.

spec/datadog/appsec/security_engine/engine_spec.rb
Strech: And what version is it? It doesn't make sense to test code with code what if both sides are wrong? With a control of input - crafted ruleset, we control output - expected behavior.

With current test we have a black box as ruleset, we don't know version, we assume it's correct. That's the issue I pointing out.

spec/datadog/appsec/security_engine/engine_spec.rb
Strech: **[not blocker]** (but good to align here)

I find it hard to read tests where it redefined multiple times and you have to scroll to correct parent context to find out the value)

Like this, by reading I know what invalid config implies, it's some invalid configuration, but `config_path` ... no idea.

```ruby
it 'reports errors through telemetry' do
          expect(telemetry).to receive(:error).with("bad cast, expected 'array', obtained 'string'")

          engine.add_or_update_config(invalid_config, path: config_path)
        end
```

I think it's easy to distinguish old tests from new, in old we have bunch of definitions like below, and I don't find it a pleasant to read and debug.

```ruby
let(:ip_list) { [1,2,3] }
let(:trace_name) { 'test-name' }
let(:headers) { {'a' => 'b'} }
```

lib/datadog/appsec/remote.rb
```diff
@@ -57,74 +56,34 @@ def products
...
-              custom_rules: custom_rules,
-              telemetry: telemetry
-            )
-
-            Datadog::AppSec.reconfigure(ruleset: ruleset, telemetry: telemetry)
-
-            repository.contents.each do |content|
-              content.applied if ASM_PRODUCTS.include?(content.path.product)
-            end
+            AppSec.reconfigure!
```
Strech, line 85: I find that moment confusing, above we configuring the security engine, but we don't reconfigure it, but reconfigure entire component. I find it a bit odd and it feels like a flaw of implementation.

If we update engine, we suppose to be responsible to reconfigure it, here we delegate it implicitly, (or we event don't, not sure)

spec/datadog/appsec/security_engine/engine_spec.rb
y9v: ok, I see what you mean - I will provide another ruleset in the input then

y9v: fixed [8b00844](https://github.com/DataDog/dd-trace-rb/pull/4725/commits/8b0084412102e91888795719d3d377f6e2ae6c19)

spec/datadog/appsec/security_engine/engine_spec.rb
y9v: fixed [8b00844](https://github.com/DataDog/dd-trace-rb/pull/4725/commits/8b0084412102e91888795719d3d377f6e2ae6c19)

lib/datadog/appsec/remote.rb
y9v: I left it like that because we wanted to change the way we use the mutex for reconfiguration. Agree it is very confusing now.

lib/datadog/appsec/security_engine/engine.rb
y9v: yes, you're right - moved telemetry reporting from this method in [9cd34e8](https://github.com/DataDog/dd-trace-rb/pull/4725/commits/9cd34e86aa9c5eadbbee9a66733f5c55911243e7)

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -0,0 +1,620 @@
...
+      engine.add_or_update_config(asm_dd_config, path: 'datadog/603646/ASM_DD/latest/config')
+    end
+
+    it 'finalizes old handle' do
+      expect(engine.instance_variable_get(:@waf_handle)).to receive(:finalize!)
+
+      engine.reconfigure!
+    end
+
+    it 'sets @waf_handle to a new handle' do
```
Strech: **[not blocker]**

Since we can probe the change through the `waf_addresses`, does it still make sense to look into internals?

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -0,0 +1,620 @@
...
+          expect(telemetry).to receive(:error).with("bad cast, expected 'array', obtained 'string'")
+
+          engine.add_or_update_config(invalid_config, path: 'datadog/603646/ASM_DD/latest/config')
+        end
+
+        it 'adds default config back' do
+          allow(engine.instance_variable_get(:@waf_builder)).to receive(:add_or_update_config).and_call_original
+
+          expect(engine.instance_variable_get(:@waf_builder))
+            .to receive(:add_or_update_config).with(anything, path: described_class::DEFAULT_RULES_CONFIG_PATH)
```
Strech: **[not blocker]**

Is there a behavior-way of testing that changes happen? This is very much prone to be broken during code changes/refactoring

lib/datadog/appsec/remote.rb
Strech: Maybe we can add comment about the change and it's GTG 👍🏼

Strech (approved):
I think it's in a very good shape now.

I just have few concerns about fragile specs (the one looking for internals) because they will be 100% broken on code changes

spec/datadog/appsec/security_engine/engine_spec.rb
y9v: yes, you are right, I will just drop this test

spec/datadog/appsec/security_engine/engine_spec.rb
y9v: I could check if `waf_addresses` change, but I would need to call `reconfigure!` then, so that a new WAF handle would be created.

PR #4732, opened 2025-06-13
.gitlab/one-pipeline.locked.yml : GHCR -> ECR migration

pawelchcki:
Updates .gitlab/one-pipeline.locked.yml . This version of the pipeline uses ECR instead of GHCR for tests. As a temporary fix for vaccine, dd-trace-rb is published to GHCR as well. This will break when Classic PATs are deprecated


ivoanjo (commented):
I understand this PR is automatically generated, but how can I validate that this change is correct, e.g. that an older or broken version of the file isn't being added?

TonyCTHsu:
>I understand this PR is automatically generated, but how can I validate that this change is correct, e.g. that an older or broken version of the file isn't being added?

I understand the test might just be green, but I shared similar concern. A mysterious SHA updated does not really tell much of a story. It would ideal to provide with an increment of version number and descriptive changelog.

ivoanjo:
Extra +1 on the version number: I think if there was something kind of like that, I could check "yeah we're going from v1.3.7 to v1.4.1" but with a hash I have no idea in what direction we're going.

p-datadog:
I think if a human is expected to approve this PR, it follows that a human is expected to read it, which means the PR should contain a similar description to what human-generated PRs contain. Probably filling out the PR template would be good?

PR #4736, opened 2025-06-13
Fix dry-run option 

TonyCTHsu:
**What does this PR do?**

1. Fix dry-run condition
2. Add `gh` instruction
3. Schedule everyday

**Change log entry**
None.


Strech (commented):
👍🏼 

Question: Do we have process described somewhere about those updates?

marcotc (approved):

PR #4737, opened 2025-06-13
chore(ssi): track instrumentation source

mabdinur:
**What does this PR do?**

Reports to the instrumentation telemetry platform whether an application was instrumented via SSI or manually.

**Motivation:**

Track the source of instrumentation

**Change log entry**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/lib/datadog/core/telemetry/event/app_started.rb
+++ b/lib/datadog/core/telemetry/event/app_started.rb
@@ -234,13 +234,16 @@ module Datadog
           # - `app.config`: only applies to .NET
           # - `default`: set when the user has not set any configuration for the key (defaults to a value)
           # - `unknown`: set for cases where it is difficult/not possible to determine the source of a config.
-          def conf_value(name, value, seq_id, origin = 'code')
-            {
-              name: name,
-              value: value,
-              origin: origin,
-              seq_id: seq_id,
-            }
+          def conf_value(name, value, seq_id, origin)
+            result = {name: name, value: value, origin: origin, seq_id: seq_id}
+            if origin == 'fleet_stable_config'
+              fleet_id = Core::Configuration::StableConfig.configuration.dig(:fleet, :id)
+              result[:config_id] = fleet_id if fleet_id
+            elsif origin == 'local_stable_config'
+              local_id = Core::Configuration::StableConfig.configuration.dig(:local, :id)
+              result[:config_id] = local_id if local_id
+            end
+            result
           end
 
           def to_value(value)
--- a/spec/datadog/core/telemetry/event/app_started_spec.rb
+++ b/spec/datadog/core/telemetry/event/app_started_spec.rb
@@ -186,6 +186,44 @@ RSpec.describe Datadog::Core::Telemetry::Event::AppStarted do
       end
     end
 
+    context 'with stable config' do
+      context 'with config id' do
+        before do
+          allow(Datadog::Core::Configuration::StableConfig).to receive(:configuration).and_return(
+            {
+              fleet: { id: '12345', config: { 'DD_APPSEC_ENABLED' => 'true' } },
+              local: { id: '56789', config: { 'DD_LOGS_INJECTION' => 'false' } },
+            }
+          )
+        end
+
+        it 'reports config id' do
+          expect(event.payload[:configuration]).to include(
+            { name: 'appsec.enabled', origin: 'fleet_stable_config', seq_id: id, value: true, config_id: '12345' },
+            { name: 'tracing.log_injection', origin: 'local_stable_config', seq_id: id, value: false, config_id: '56789' },
+          )
+        end
+
+        context 'without config id' do
+          before do
+            allow(Datadog::Core::Configuration::StableConfig).to receive(:configuration).and_return(
+              {
+                fleet: { config: { 'DD_APPSEC_ENABLED' => 'true' } },
+                local: { config: { 'DD_LOGS_INJECTION' => 'false' } }
+              }
+            )
+          end
+
+          it 'does not report config id' do
+            expect(event.payload[:configuration]).to include(
+              { name: 'appsec.enabled', origin: 'fleet_stable_config', seq_id: id, value: true },
+              { name: 'tracing.log_injection', origin: 'local_stable_config', seq_id: id, value: false },
+            )
+          end
+        end
+      end
+    end
+
     context 'with nil configurations' do
       before do
         Datadog.configure do |c|
--- a/.github/forced-tests-list.json
+++ b/.github/forced-tests-list.json
@@ -1,9 +1,3 @@
 {
-    "PARAMETRIC": [
-        "tests/parametric/test_config_consistency.py::Test_Stable_Config_Default::test_config_precedence",
-        "tests/parametric/test_config_consistency.py::Test_Stable_Config_Default::test_invalid_files",
-        "tests/parametric/test_config_consistency.py::Test_Stable_Config_Default::test_unknown_key_skipped",
-        "tests/parametric/test_config_consistency.py::Test_Stable_Config_Default::test_default_config",
-        "tests/parametric/test_process_discovery.py"
-    ]
+  
 }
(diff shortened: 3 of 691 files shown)


lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -104,6 +104,15 @@ def configuration
               list << conf_value(option, to_value(config.dig(*split_option)), seq_id)
             end
 
+            instrumentation_source = defined?(Datadog::SingleStepInstrument::LOADED) ? 'ssi' : 'manual'
```
mabdinur:
```suggestion
            instrumentation_source = Datadog.const_defined?(:SingleStepInstrument, false) && Datadog::SingleStepInstrument.const_defined?(:LOADED, false) ? 'ssi' : 'manual'
```

lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -104,6 +104,15 @@ def configuration
               list << conf_value(option, to_value(config.dig(*split_option)), seq_id)
             end
 
+            instrumentation_source = Datadog.const_defined?(:SingleStepInstrument, false) && Datadog::SingleStepInstrument.const_defined?(:LOADED, false) ? 'ssi' : 'manual'
```
mabdinur:
```suggestion
            instrumentation_source =
              if Datadog.const_defined?(:SingleStepInstrument, false) &&
                 Datadog::SingleStepInstrument.const_defined?(:LOADED, false)
                'ssi'
              else
                'manual'
              end
```

lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -104,6 +104,21 @@ def configuration
               list << conf_value(option, to_value(config.dig(*split_option)), seq_id)
             end
 
+            instrumentation_source =
+              if Datadog.const_defined?(:SingleStepInstrument, false) &&
+                 Datadog::SingleStepInstrument.const_defined?(:LOADED, false)
```
marcotc:
```suggestion
              if defined?(Datadog::SingleStepInstrument::LOADED)
```

spec/datadog/single_step_instrument_spec.rb
```diff
@@ -1,11 +1,29 @@
 RSpec.describe 'Single step instrument', skip: !Process.respond_to?(:fork) do
+  subject(:single_step_instrument) { load 'datadog/single_step_instrument.rb' }
+
+  before do
+    # Store the original state if needed
+    @original_single_step_instrument = defined?(Datadog::SingleStepInstrument) ? Datadog::SingleStepInstrument : nil
```
marcotc, line 6: Can you use `stub_const` instead to mock the constant?
```ruby
stub_const('Datadog::SingleStepInstrument::LOADED', true)
```
There's no need to clean it up after the test run.

lib/datadog/single_step_instrument.rb
```diff
@@ -5,6 +5,14 @@
 #
 # This file's path is private. Do not reference this file.
 #
+
+module Datadog
+  module SingleStepInstrument
+    # Flag to determine if Single Step Instrumentation was used
+    LOADED = true
```
p-datadog: Should this only be defined in the begin-rescue block below if there is no exception as rescued no line 18?

spec/datadog/single_step_instrument_spec.rb
p-datadog: My impression is the constant needs to be defined by the code under test, defining it in the test would defeat the purpose of the test.

spec/datadog/core/telemetry/event/app_started_spec.rb
```diff
@@ -32,6 +32,10 @@
       ['tracing.partial_flush.min_spans_threshold', 500],
       ['tracing.report_hostname', false],
       ['tracing.sampling.rate_limit', 100],
+      ['instrumentation_source', 'manual'],
+      ['DD_INJECT_FORCE', false],
+      ['DD_INJECTION_ENABLED', ''],
+      ['tracing.auto_instrument.enabled', false],
```
mabdinur:
```suggestion
```

lib/datadog/single_step_instrument.rb
mabdinur: Yes good catch. We only want to set this if ssi is successful

lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -162,6 +162,21 @@ def configuration
...
+              'ssi'
+            else
+              'manual'
+            end
+            inject_force = Core::Environment::VariableHelpers.env_to_bool('DD_INJECT_FORCE', false)
+            # Track ssi configurations
+            list.push(
+              conf_value('instrumentation_source', instrumentation_source, seq_id),
+              conf_value('DD_INJECT_FORCE', inject_force, seq_id),
+              conf_value('DD_INJECTION_ENABLED', ENV['DD_INJECTION_ENABLED'] || '', seq_id),
```
mabdinur:
```suggestion
              conf_value('instrumentation_source', instrumentation_source, seq_id, 'code'),
              conf_value('DD_INJECT_FORCE', inject_force, seq_id, 'code'),
              conf_value('DD_INJECTION_ENABLED', ENV['DD_INJECTION_ENABLED'] || '', seq_id, 'code'),
```

lib/datadog/single_step_instrument.rb
```diff
@@ -5,8 +5,18 @@
...
+  # This module handles conditional loading of single step auto-instrumentation,
+  # which enables Datadog tracing and profiling features when available.
+  module SingleStepInstrument
+    LOADED = false
+  end
+end
+
 begin
   require_relative 'auto_instrument'
+  Datadog::SingleStepInstrument::LOADED = true
```
marcotc: Why is `LOADED` set to `false`, then set to `true`. Do we need to know if SSI in "being initialized"?

mabdinur: I thought it was better than leaving it undefined. 

Undefined means SSI wasn't attempted 
False means SSI failed
True means SSI set up was successful 

I think all 3 states are useful to have

p-datadog: The constant reassignment produces a warning under `ruby -w`:

```
big% irb -w -Ilib -rdatadog -rdatadog/single_step_instrument
/home/w/apps/dd-trace-rb/lib/datadog/single_step_instrument.rb:19: warning: already initialized constant Datadog::SingleStepInstrument::LOADED
/home/w/apps/dd-trace-rb/lib/datadog/single_step_instrument.rb:13: warning: previous definition of LOADED was here
```

If I am remembering correctly we have had customer complaints previously when we produced warnings under `-w`.

If you wish to track the attempted-but-failed state this can be done by defining two constants - one for attempt and one for success.

p-datadog:
Is this a customer-visible change? Telemetry is not directly customer-visible but for example dynamic instrumentation UI does not work correctly without a bunch of data being sent to telemetry.

mabdinur:
> Is this a customer-visible change? Telemetry is not directly customer-visible but for example dynamic instrumentation UI does not work correctly without a bunch of data being sent to telemetry.

There is a plan to surface this config value in the Datadog UI. I believe the tracer configuration data sent to instrumentation telemetry can be queried using ddsql. But from the SDK perspective this configuration is internal to the ddtrace library.

lib/datadog/single_step_instrument.rb
marcotc: @mabdinur Oleg is right, we can't reassign constants, as it is not good practice and emits a user-visible warning.
Let's stick to one value, `true` here.

lib/datadog/single_step_instrument.rb
```diff
@@ -5,8 +5,18 @@
 #
 # This file's path is private. Do not reference this file.
 #
+
+module Datadog
+  # This module handles conditional loading of single step auto-instrumentation,
+  # which enables Datadog tracing and profiling features when available.
+  module SingleStepInstrument
+    LOADED = false
```
mabdinur:
```suggestion
```

PR #4741, opened 2025-06-16
Move AgentSettings out of AgentSettingsResolver

p-datadog:
**What does this PR do?**

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/lib/datadog/core/configuration/agent_settings.rb
+++ b/lib/datadog/core/configuration/agent_settings.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+require_relative 'ext'
+
+module Datadog
+  module Core
+    module Configuration
+      # Immutable container for the resulting settings
+      class AgentSettings
+        attr_reader :adapter, :ssl, :hostname, :port, :uds_path, :timeout_seconds
+
+        def initialize(adapter: nil, ssl: nil, hostname: nil, port: nil, uds_path: nil, timeout_seconds: nil)
+          @adapter = adapter
+          @ssl = ssl
+          @hostname = hostname
+          @port = port
+          @uds_path = uds_path
+          @timeout_seconds = timeout_seconds
+          freeze
+        end
+
+        def url
+          case adapter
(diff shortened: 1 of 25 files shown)


lib/datadog/core/configuration/agent_settings.rb
```diff
@@ -0,0 +1,52 @@
...
+          when Datadog::Core::Configuration::Ext::Agent::HTTP::ADAPTER
+            hostname = self.hostname
+            hostname = "[#{hostname}]" if IPV6_REGEXP.match?(hostname)
+            "#{ssl ? "https" : "http"}://#{hostname}:#{port}/"
+          when Datadog::Core::Configuration::Ext::Agent::UnixSocket::ADAPTER
+            "unix://#{uds_path}"
+          else
+            raise ArgumentError, "Unexpected adapter: #{adapter}"
+          end
+        end
```
ivoanjo, line 38: Minor: The code was already like this, but it occurs to me that this logic here is very weird, e.g. for everything else all the logic is done in the `AgentSettingsResolver` and the `AgentSettings` is a dumb object that only exposes the result.

...But not for this one bit? It seems a bit weird.

TL;DR would be cool if `#url` was moved back inside the `AgentSettingsResolver`, but of course not a blocker for this PR if you don't want to pile on.

(TBH if we didn't have to support older Rubies, this class would be basically a `Data.define`...)

p-datadog: I looked into this and while the other fields are determined from settings/environment/etc., `url` method is a transformation *from* AgentSettings to a more limited description (specifically, the timeout information is lost by this transformation). I don't think it is appropriate to move `url` to either of the resolvers.

ivoanjo: Hmmm half of the work the agentsettings is doing already is doing transformations so I'm not entirely convinced. :eyes: 

That said, the above was a nice to have IMO, so not a blocker for this PR.

PR #4743, opened 2025-06-17
DYNIS-22 Add git tags to dynamic instrumentation snapshot payloads

p-datadog:
**What does this PR do?**
Adds git metadata tags to dynamic instrumentation snapshot payloads 

**Motivation:**
Dynamic instrumentation/live debugger UI requires this information

**Change log entry**
Yes: Fix source code display in dynamic instrumentation/live debugger

**Additional Notes:**

This PR also DRYs existing tag generation code which was used by crash tracker and profiling.

Tracing code also sends these tags but they are not generated in the same format of a single hash, I left the tracing code unchanged.

**How to test the change?**
Unit tests added in this PR + system tests to be written

### DIFF
--- a/spec/datadog/profiling/tag_builder_spec.rb
+++ b/spec/datadog/profiling/tag_builder_spec.rb
@@ -81,10 +81,12 @@ RSpec.describe Datadog::Profiling::TagBuilder do
     describe "source code integration" do
       context "when git environment is available" do
         before do
-          allow(Datadog::Core::Environment::Git).to receive(:git_repository_url).and_return(
+          Datadog::Core::TagBuilder.reset_for_tests
+
+          expect(Datadog::Core::Environment::Git).to receive(:git_repository_url).and_return(
             "git_repository_url"
           )
-          allow(Datadog::Core::Environment::Git).to receive(:git_commit_sha).and_return("git_commit_sha")
+          expect(Datadog::Core::Environment::Git).to receive(:git_commit_sha).and_return("git_commit_sha")
         end
 
         it "includes the git repository URL and commit SHA" do
@@ -96,8 +98,10 @@ RSpec.describe Datadog::Profiling::TagBuilder do
 
       context "when git environment is not available" do
         before do
-          allow(Datadog::Core::Environment::Git).to receive(:git_repository_url).and_return(nil)
-          allow(Datadog::Core::Environment::Git).to receive(:git_commit_sha).and_return(nil)
+          Datadog::Core::TagBuilder.reset_for_tests
+
+          expect(Datadog::Core::Environment::Git).to receive(:git_repository_url).and_return(nil)
+          expect(Datadog::Core::Environment::Git).to receive(:git_commit_sha).and_return(nil)
         end
 
         it "includes the git repository URL and commit SHA" do


lib/datadog/di/probe_notification_builder.rb
```diff
@@ -158,11 +158,24 @@ def build_snapshot(probe, rv: nil, locals: nil, path: nil,
...
           timestamp: timestamp,
+          ddtags: serialized_tags,
         }
       end
 
       private
 
+      def tags
+        Core::TagBuilder.tags(settings).merge(
+          debugger_version: Core::Environment::Identity.gem_datadog_version,
```
ivoanjo, line 169: Minor: Not sure if you want this anymore? We added this for profiling as all other profilers report it, but since this is the same as library version, you may not need it at all?

lib/datadog/di/probe_notification_builder.rb
```diff
@@ -158,11 +158,24 @@ def build_snapshot(probe, rv: nil, locals: nil, path: nil,
...
             duration: duration ? duration * 1000 : 0),
           timestamp: timestamp,
+          ddtags: serialized_tags,
         }
       end
 
       private
 
+      def tags
+        Core::TagBuilder.tags(settings).merge(
```
ivoanjo, line 168: Minor: Building tags at every snapshot will probably get expensive? May be worth looking into caching these (profiling does it once a minute and crashtracking only once, as a comparison)

lib/datadog/profiling/ext.rb
ivoanjo, line 1: :+1: I need to kill off this file ;)

spec/support/tag_builder_helpers.rb
```diff
@@ -0,0 +1,8 @@
+module TagBuilderHelpers
+  # Resets caches in this module.
+  def reset_for_tests
+    remove_instance_variable(:@fixed_environment_tags) if defined?(@fixed_environment_tags)
+  end
+end
+
+Datadog::Core::TagBuilder.extend TagBuilderHelpers
```
ivoanjo, line 8: I'm... really not a fan of doing it this way. I find it really confusing for methods to "show up" on production classes that aren't really there, especially if they only show up in tests.

As an alternative, I suggest turning this into a regular method:

```
module TagBuilderHelpers
  def self.reset_for_tests
    Datadog::Core::TagBuilder.remove_instance_variable(:@fixed_environment_tags) if ...
```

This way it'll be clear in the tests that `reset_for_tests` is coming from an helper, and not something that's in the actual class.

ivoanjo (commented):
Left a few notes!

lib/datadog/di/probe_notification_builder.rb
p-datadog: I copied this from dd-trace-py, maybe it's not needed but it's easiest to be consistent with other implementations. Are you saying profiler reported the *debugger* version?

lib/datadog/di/probe_notification_builder.rb
p-datadog: Yes, I considered the overhead but 1) I did not (and do not) presently have any hard performance requirements, and 2) the full set of tags was dynamically built for each use (with this PR at least some of the tags are staying in a hash). I can try to add some more caching.

spec/support/tag_builder_helpers.rb
p-datadog: There was another complaint about this code in https://github.com/DataDog/dd-trace-rb/pull/4653. I don't have a preference as to how this is implemented other than it needs to work :) Especially with more caches as discussed above there need to be a way to reset them for tests.

lib/datadog/di/probe_notification_builder.rb
ivoanjo: I suspect we may be copy-pasting a pattern that may not not actually be needed.

At some point we had some plans of separating things into separate packages that had different versions, but that mostly didn't happen so now we're sending N different "versions" which are all the same.

If we're not really using this, we could remove it and use the `library_version` that already gets reported: https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/core/tag_builder.rb#L24 .

(But this is a minor comment, so up to you if you want to keep it or not, just pointing it out that if this was done like this as a copy-paste from profiling, you may not want to copy-paste this part)

lib/datadog/di/probe_notification_builder.rb
ivoanjo: Fair enough, although it looks to me that making message sending cheap will be key in dynamic instrumentation's low-overhead.

spec/support/tag_builder_helpers.rb
ivoanjo: I think it's fine to have resets -- they're found on most of our classes that have state but that need a bit finer control for tests as otherwise it'd be hard to simulate some cases.

But... I think changing the class/module under test for adding a reset is something I'd avoid, especially given this being Ruby we can do everything on the test code side without a lot of awkwardness :)

PR #4745, opened 2025-06-18
[NO-TICKET] Profiler: Capture native filenames in backtraces

ivoanjo:
**What does this PR do?**

This PR introduces a new feature for the Ruby profiler: showing native filenames for methods that are defined in native code (e.g. not a Ruby source file):

image

This feature is on by default, and can be disabled via code `profiling.advanced.native_filenames_enabled` or via env var `DD_PROFILING_NATIVE_FILENAMES_ENABLED`.

**Motivation:**

Consider for instance this simple example:

```ruby
require 'bigdecimal'

BigDecimal.singleton_class.prepend(
  Module.new do
    def save_rounding_mode
      super
    end
  end
)

[:example].each do
  BigDecimal.save_rounding_mode do
    puts caller
    sleep 1
  end
end
```

which Ruby will render as:

```
native-filenames-example.rb:6:in `save_rounding_mode'
native-filenames-example.rb:6:in `save_rounding_mode'
native-filenames-example.rb:12:in `block in <main>'
native-filenames-example.rb:11:in `each'
native-filenames-example.rb:11:in `<main>'
```

You'll notice that Ruby doesn't show any different file other than `native-filenames-example.rb`, even though we're calling into `Array#each` and `BigDecimal.save_rounding_mode` and those are not defined in the example file.

This happens because those methods are implemented as native code, and Ruby's default behavior is to not show the file where those methods are defined, showing instead to show the file for the caller.

This can get confusing when using the profiler, especially when source code integration is not available.

Up until now, the profiler matched how Ruby renders backtraces, including when native methods are involved. Here's how the above example looked:

image

And here's how it looks when native filenames is enabled:

image

Instead of attributing `Array#each` to `native-filenames-example.rb`, it gets attributed to the native filename where it lives: `libruby.so.3.2` (in my example). Similar for `BigDecimal.save_rounding_mode`: `bigdecimal.so`.

This means that for instance the "only my code" view now shows correctly for this example as well:

image

whereas before it would not collapse those two frames.

**Change log entry**
Yes. Capture native filenames in backtraces

**Additional Notes:**
I've also updated one of our microbenchmarks + checked in our usual test environments and the performance impact of this feature is quite small, falling within the margin of error.

**How to test the change?**
This change includes test coverage, and the above example can also be used to validate the feature is working fine.


benchmarks/profiling_sample_loop_v2.rb
```diff
@@ -14,44 +14,88 @@ class ProfilerSampleLoopBenchmark
...
+
+    Thread.new { deep_stack.call }.tap { |t| t.name = "Deep stack #{depth}" }
+  end
+
+  def run_benchmark(mode: :ruby)
+    threads = Array.new(4) { mode == :ruby ? thread_with_very_deep_stack : thread_with_very_deep_stack_and_native_frames }
+    collector = Datadog::Profiling::Collectors::ThreadContext.for_testing(recorder: @recorder)
+
+    if mode == :native
+      # TODO: REMOVE THIS AFTER MERGING PR (IT BREAKS ADDING THIS BENCHMARK IN CI)
```
AlexJF, line 51: Do you mean uncomment this after merging PR? I'm a bit confused since this should work in CI given it runs on Linux no?

benchmarks/profiling_sample_loop_v2.rb
```diff
@@ -14,44 +14,88 @@ class ProfilerSampleLoopBenchmark
...
 end
 
 puts "Current pid is #{Process.pid}"
 
 ProfilerSampleLoopBenchmark.new.instance_exec do
   create_profiler
-  4.times { thread_with_very_deep_stack }
-  run_benchmark
+  run_benchmark(mode: :ruby)
+  run_benchmark(mode: :native)
```
AlexJF, line 106: Nit: Maybe instead of raising inside `run_benchmark`, we could check the condition above and conditionally run this benchmark, logging a warning if we skipped it?

spec/datadog/profiling/collectors/stack_spec.rb
```diff
@@ -214,6 +225,89 @@ def call_sleep
...
+            end
+          end
+        end
+      end
+
+      it "matches the Ruby backtrace API" do
+        expect(gathered_stack).to eq reference_stack
+      end
+
+      context "when native filenames are enabled" do
```
AlexJF, line 245: Isn't this missing a skip if `!_native_filenames_available?`?

AlexJF (approved):

spec/datadog/profiling/collectors/stack_spec.rb
ivoanjo: It is! Nice spot, this would break on macOS. Fixed in 0247d8d82a2798c30e5e8123eb8cf65733b45f1

benchmarks/profiling_sample_loop_v2.rb
ivoanjo: Fixed in e934abc11a5260f47c75a3ce576c4f875b97eb5 to only raise when on Linux. I don't expect this to ever not be available on Linux, so I like having the exception to prove me wrong in that case, but yeah, makes sense to not block macOS/others :)

benchmarks/profiling_sample_loop_v2.rb
ivoanjo: So... our current benchmarking setup is... weird in this regard.

Specifically, it checks out this branch, runs the benchmarks, and then it copies the benchmarks to a temp folder, and then checks out master, and runs the benchmarks **from this branch** with master.

That's.... cool if you add a new benchmark for something, and then modify some existing code to make things faster, and boom, in your PR you can see the before/after right away.

But whenever the benchmark is not backwards-compatible with master (e.g. some new arguments get added, classes get changed, etc), then this completely breaks, since master doesn't really have the ability to run the benchmark without the other changes.

So yeah, fixing our benchmarks integration aside, the workaround is to just comment the backwards-incompatible changes (or the whole benchmark), get it in master first, and then in a follow-up PR it can be enabled (since by then master will be able to run the benchmark).

`>_>`

ivoanjo: Marking as resolved so I can go ahead and merge the PR, but as usual, happy to chat more about it!

PR #4750, opened 2025-06-20
[NO-TICKET] Profiler: Groundwork for sampling from inside signal handler

ivoanjo:
**What does this PR do?**

This PR contains the first part of the changes needed to allow the profiler to sample stacks from inside the signal handler: It introduces the `thread_context_collector_prepare_sample_inside_signal_handler` + `prepare_sample_thread` APIs which allow the stack for a Ruby thread to be read and stored inside our existing sampling buffer, but not actually symbolicated or recorded.

The intention here will be that inside the signal handler we'll collect the stack immediately into the buffer, but go no further.

This PR also includes a few cleanups I did as part of this groundwork.

**Motivation:**
On older Ruby versions, the profiler wasn't able to sample from inside the signal handler. This is no longer the case since https://github.com/ruby/ruby/pull/3296 and https://github.com/ruby/ruby/pull/11036 . Sampling from inside the signal handler increases profiler accuracy, and thus we want to start doing it on Rubies where it is safe to do so. (I have a few examples which I plan to share in the PR that lands the feature end-to-end).

I decided to open a PR with only the groundwork for this feature as I was starting to accumulate a bit too many changes, and I also may need to switch gears for a few days, so this leaves a lot of the work already in-tree rather than bitrotting somewhere.

**Change log entry**
None. (Will come once the feature is fully in)

**Additional Notes:**
I've opened this PR on top of https://github.com/DataDog/dd-trace-rb/pull/4745 to avoid conflicts as they both touch some of the same bits of the profiler, but otherwise they are independent.

Since I've also done a few cleanups, it may be easier to review this PR commit-by-commit.

**How to test the change?**
I've added test coverage for the changes to the `Stack` and `ThreadContext` collectors, although the end-to-end change cannot be tested yet.


ext/datadog_profiling_native_extension/collectors_thread_context.c
```diff
@@ -1459,6 +1462,26 @@ static VALUE thread_list(thread_context_collector_state *state) {
...
+// the current thread.
+//
+// Assumptions for this function are same as for `thread_context_collector_sample` except that this function is
+// expected to be called from a signal handler and to be async-signal-safe.
+//
+// Also, no allocation (Ruby or malloc) can happen.
+void thread_context_collector_prepare_sample_inside_signal_handler(VALUE self_instance) {
+  thread_context_collector_state *state;
+  if (!rb_typeddata_is_kind_of(self_instance, &thread_context_collector_typed_data)) return;
+  // This should never fail the the above check passes
```
AlexJF: `never fail IF the above`

ext/datadog_profiling_native_extension/collectors_stack.c
```diff
@@ -180,7 +173,8 @@ static VALUE _native_sample(int argc, VALUE *argv, DDTRACE_UNUSED VALUE _self) {
   int max_frames_requested = sampling_buffer_check_max_frames(NUM2INT(max_frames));
 
   ddog_prof_Location *locations = ruby_xcalloc(max_frames_requested, sizeof(ddog_prof_Location));
-  sampling_buffer *buffer = sampling_buffer_new(max_frames_requested, locations);
+  sampling_buffer buffer;
+  sampling_buffer_initialize(&buffer, max_frames_requested, locations);
```
AlexJF, line 177: Maybe the need will become more evident in later PRs but it is not clear to me in this PR what improvement this brings vs the potential for confusion as we now have something that needs `freeing` due to an internal implementation detail even though it looks to be stack managed?

ivoanjo: Good point (as usual haha!). So I did this for two reasons, both of which are a bit small so this is not strictly needed:

1. I wanted to make marking as fast as possible. By inlining this structure into the context, the `sampling_buffer_needs_marking` check is really cheap and avoids needing to follow pointers to answer the query.
2. The design on the codebase is very OO-y at times without a lot of advantage (this being C + ends up being a lot of pointers and separate allocations), and so in some cases I've been inlining these smaller pieces into bigger structures.

Having said that, especially for 1., marking is O(threads) so chasing down a few dozen-at-most-a-few-hundred-pointers is probably not a particularly noticeable change.

ext/datadog_profiling_native_extension/collectors_thread_context.c
```diff
@@ -1453,6 +1462,26 @@ static VALUE thread_list(thread_context_collector_state *state) {
...
```
ivoanjo: Thanks!

```suggestion
  // This should never fail if the above check passes
```

ext/datadog_profiling_native_extension/collectors_stack.c
ivoanjo: I'm going to go ahead and merge the PR, but let me know if you're not very convinced -- I think this is an overall small part of the change so it can be undone without any impact to the functionality.

AlexJF: I'm fine either way, was just wondering which detail I had missed. The saving of those pointer chases seems to make sense!

PR #4757, opened 2025-06-24
Update to libddwaf v1.24.1 to adopt per-config error reporting

y9v:
**What does this PR do?**
This PR updates `libddwaf-rb` to 1.24.1 and changes the way how the Remote Config changes are handled.

**Motivation:**
We want to report errors for separately for each Remote Config configuration.

**Change log entry**
Yes. Deprecate `config.appsec.ip_passlist`, `config.appsec.ip_denylist`, and `config.appsec.user_id_denylist`.

**Additional Notes:**
Only the ruby library supports configuration of allow- and denylists directly via library configuration. Other libraries only configure it using Remote Configuration. This is why we are deprecating it.

**How to test the change?**


lib/datadog/appsec/remote.rb
```diff
@@ -57,74 +56,35 @@ def products
...
+            next unless AppSec.security_engine
 
-            if rules.empty?
-              settings_rules = AppSec::Processor::RuleLoader.load_rules(
-                telemetry: telemetry,
-                ruleset: Datadog.configuration.appsec.ruleset
-              )
+            changes.each do |change|
+              content = repository[change.path]
+              next unless content || change.type == :delete
```
Strech, line 68: Will it be the same?

```suggestion
              next unless content&.type == :delete
```

Strech (approved):
Let's goooo!

lib/datadog/appsec/remote.rb
y9v: no, not quite - we want to skip if there is no content, but when the type is not `delete`, since it's ok for `delete` operation to have no content

TonyCTHsu (approved):

PR #4765, opened 2025-06-27
Add post-release workflow step to create a PR with version bump

y9v:
**What does this PR do?**
This PR adds a Github Actions workflow step to `Publish Gem` workflow to generate a PR with the next prerelease version.

This is done using existing `rake version:bump` task that increases the minor version and sets the tiny version to zero.

**Motivation:**
We want to have the next prerelease version in our `version.rb` file to simplify system-tests setup.

**Change log entry**

**Additional Notes:**
We also need to update the `lib/datadog/version.rb` file manually.

**How to test the change?**
-

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -345,3 +345,52 @@ jobs:
             --title "Update document v${GEM_VERSION}" \
           --body "This is an auto-generated PR to update documentation from [here](${JOB_URL}). Please merge (with a merge commit) when ready.\n\nTo resolve conflicts:\n\`\`\`bash\ngit merge release\ngit checkout --ours .\`\`\`." \
             --label "docs"
+
+  update-version-file:
+    if: github.ref_name == 'master'
+    name: Update version file with next prerelease version
+    runs-on: ubuntu-24.04
+    needs:
+      - verify-checks
+      - rubygems-release
+      - update-gem-version
+    permissions:
+      contents: write
+      pull-requests: write
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


y9v:
@Strech yes, there will be, and there is one more problem - we need to know the previous released version, and with this change we won't have it. So there will be another PR for fast_castle this week

.github/workflows/bump-gem-version.yml
```diff
@@ -0,0 +1,104 @@
...
+    name: Create Pull Request to update the version file
+    runs-on: ubuntu-24.04
+    needs:
+      - next-gem-version
+    permissions:
+      contents: write
+      pull-requests: write
+    env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      NEXT_VERSION: ${{ needs.next-gem-version.outputs.next_version }}.dev
```
TonyCTHsu: I would prefer to keep things simple by removing `dev`

.github/workflows/bump-gem-version.yml
```diff
@@ -0,0 +1,104 @@
+name: Bump datadog gem version
+
+on: # yamllint disable-line rule:truthy
+  repository_dispatch:
```
TonyCTHsu: `repository_dispatch` is redundant. 

We could use `workflow_call` to invoke from `publish.yml` with version already computed.

.github/workflows/bump-gem-version.yml
y9v: ok, let's remove it for now

y9v: removed in [7b47a1e](https://github.com/DataDog/dd-trace-rb/pull/4765/commits/7b47a1e693893560f5d3751152f241b9de03cc43)

PR #4766, opened 2025-06-27
DYNIS-22 Send ddtags in query string

p-datadog:
**What does this PR do?**

Changes DI transport and related code to send `ddtags` in query string instead of payload.

**Motivation:**
I used Python tracer as reference for the initial implementation and it supports sending the tags as either part of the payload or in the query string. The body sending made more sense to use. However, when writing system tests I discovered that 1) Python tracer in practice always sends the tags in query string and 2) all other tracers only support sending them in the query string as far as I can tell.

**Change log entry**
None (same changelog entry as https://github.com/DataDog/dd-trace-rb/pull/4743)

**Additional Notes:**

**How to test the change?**
Unit tests added in this PR + system tests: https://github.com/DataDog/system-tests/pull/4861

### DIFF
--- a/sig/datadog/core/tag_builder.rbs
+++ b/sig/datadog/core/tag_builder.rbs
@@ -1,13 +1,13 @@
 module Datadog
   module Core
     module TagBuilder
-      self.@fixed_environment_tags: untyped
+      self.@fixed_environment_tags: Hash[String, String]
 
-      def self.fixed_environment_tags: () -> untyped
+      def self.fixed_environment_tags: () -> Hash[String, String]
 
-      def self.tags: (untyped settings) -> untyped
+      def self.tags: (untyped settings) -> Hash[String, String]
 
-      def self.serialize_tags: (untyped tags) -> untyped
+      def self.serialize_tags: (Hash[String, String] tags) -> String
     end
   end
 end


y9v (approved):

lib/datadog/core/tag_builder.rb
```diff
@@ -44,6 +44,13 @@ def self.tags(settings)
           'version' => settings.version,
         }.compact)
       end
+
+      def self.serialize_tags(tags)
+        # DEV: Should there be some sort of escaping done here?
+        tags.map do |key, value|
+          "#{key}:#{value}"
+        end.join(',')
```
Strech, line 52:
```suggestion
        tags.map { |key, value| "#{key}:#{value}" }.join(',')
```

lib/datadog/core/transport/http/adapters/net.rb
```diff
@@ -86,6 +86,21 @@ def url
...
+              when Hash
+                path = path + '?' + URI.encode_www_form(query)
+              when String
+                path = path + '?' + query
+              when nil
+                # Nothing
+              else
+                raise ArgumentError, "Invalid type for query: #{query}"
+              end
+              path
```
Strech, line 101: You can avoid that by just returning path in `nil` case and skip assignment in other cases

```suggestion
              case query = env.query
              when Hash then path + '?' + URI.encode_www_form(query)
              when String then path + '?' + query
              when nil then path
              else raise ArgumentError, "Invalid type for query: #{query}"
              end
```

sig/datadog/core/tag_builder.rbs
```diff
@@ -6,6 +6,8 @@ module Datadog
       def self.fixed_environment_tags: () -> untyped
 
       def self.tags: (untyped settings) -> untyped
+
+      def self.serialize_tags: (untyped tags) -> untyped
```
Strech: I think we can type it, at least return statement

sig/datadog/di/transport/input.rbs
```diff
@@ -7,6 +7,11 @@ module Datadog
         end
 
         class Request < Datadog::Core::Transport::Request
+          @serialized_tags: untyped
```
Strech, line 10: I think we can type it

spec/datadog/core/remote/client_spec.rb
```diff
@@ -6,6 +6,11 @@
 require 'datadog/core/remote/client'
 
 RSpec.describe Datadog::Core::Remote::Client do
+  before(:all) do
```
Strech: This will impact all tests, not only in this file, if it's possible I would prefer to do just `before(:each)`

spec/datadog/di/probe_notifier_worker_spec.rb
```diff
@@ -5,12 +5,25 @@
 RSpec.describe Datadog::DI::ProbeNotifierWorker do
   di_test
 
+  before(:all) do
```
Strech: Same here, this goes for all tests in the run, not only this specific file

Strech (approved):
I left few suggestions and 1 major concern about testing

sig/datadog/core/tag_builder.rbs
p-datadog: I typed the core file but not DI code because  I am not getting value from type declarations.

spec/datadog/core/remote/client_spec.rb
p-datadog: Changed to before each test.

lib/datadog/core/transport/http/adapters/net.rb
p-datadog: The idea was to only change path if there is a query string to append, and to have the path modification clearly visible.

PR #4769, opened 2025-06-30
[APMAPI-1356] Add config_id in stable config telemetry

vpellan:
**What does this PR do?**

This PR adds config_id field in stable config telemetry. This also adds unit tests for stable config

**Motivation:**

We are not sending Stable config telemetry correctly

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

CI + System-tests


spec/datadog/core/configuration/stable_config_spec.rb
```diff
@@ -12,4 +14,92 @@
       end
     end
   end
+
+  describe '#configuration', skip: !LibdatadogHelpers.supported? do
+    context 'when libdatadog API is available' do
+      context 'with config_id' do
+        before do
+          FileUtils.mkdir_p('/tmp/datadog-agent/managed/datadog-agent/stable')
```
p-datadog: This should use `File.mktmpdir` (e.g. https://stackoverflow.com/questions/1139265/what-is-the-best-way-to-get-an-empty-temporary-directory-in-ruby-on-rails) rather than  hardcoding `/tmp/datadog-agent`, for at least three reasons:

1) there may already be a `/tmp/datadog-agent` on the system (for example I frequently git clone repositories into /tmp).

2) There may be a `/tmp/datadog-agent` directory created by dd-trace-rb test suite run by another user, which won't be removable.

3) Fixed paths in `/tmp` are a security vulnerability (https://duckduckgo.com/?t=h_&q=tmp+overwrite+attack&ia=web).

p-datadog (dismissed):

spec/datadog/core/configuration/stable_config_spec.rb
vpellan: Thank you ! I've replaced it in [edf5386](https://github.com/DataDog/dd-trace-rb/pull/4769/commits/edf538603e9064e8c83dc0c957e0c1451ae55afd), tell me if I missed anything else !

Rakefile
```diff
@@ -214,6 +214,16 @@ namespace :spec do
...
 
+  # rubocop:disable Style/MultilineBlockChain
+  desc '' # "Explicitly hiding from `rake -T`"
+  RSpec::Core::RakeTask.new(:stable_config) do |t, args|
+    t.pattern = 'spec/datadog/core/configuration/stable_config_spec.rb'
+    t.rspec_opts = args.to_a.join(' ')
+  end.tap do |t|
+    Rake::Task[t.name].enhance(["compile:libdatadog_api.#{RUBY_VERSION[/\d+.\d+/]}_#{RUBY_PLATFORM}"])
+  end
+  # rubocop:enable Style/MultilineBlockChain
```
ivoanjo, line 225: Not for this PR, but I'm thinking this boilerplate is adding a bit up...

So at some point (very vague, does not need to be in the short term) it may be nice to revisit a bit how we're doing this testing that requires the native extension.

ext/libdatadog_api/library_config.c
```diff
@@ -52,6 +56,9 @@ void library_config_init(VALUE core_module) {
   rb_define_alloc_func(configurator_class, _native_configurator_new);
   rb_define_method(configurator_class, "get", _native_configurator_get, 0);
 
+  rb_define_method(configurator_class, "with_local_path", _native_configurator_with_local_path, 1);
+  rb_define_method(configurator_class, "with_fleet_path", _native_configurator_with_fleet_path, 1);
```
ivoanjo: One pattern we've been using in profiling is that I recommend we use here is to create a separate submodule for the testing stuff and put it there -- https://github.com/DataDog/dd-trace-rb/blob/master/ext/datadog_profiling_native_extension/profiling.c#L71-L75 .

This makes it clear when looking at the testing code what's functionality that's expected in production vs just used to setup tests.

(The medium/long-term thinking is that we would not define `Testing` code when loaded in production, or even not compile it at all! And when we're ready to do so, we can very clearly identify such code -- it's all functions being bound inside the testing modules)

ext/datadog_profiling_native_extension/datadog_ruby_common.h
```diff
@@ -41,6 +41,12 @@ static inline ddog_CharSlice char_slice_from_ruby_string(VALUE string) {
   return char_slice;
 }
 
+static inline ddog_CStr cstr_from_ruby_string(VALUE string) {
+  ENFORCE_TYPE(string, T_STRING);
+  ddog_CStr cstr = {.ptr = RSTRING_PTR(string), .length = RSTRING_LEN(string)};
+  return cstr;
+}
+
```
ivoanjo: Minor: So I know this is a copy/paste of the thing above, but having said that, I don't think this belongs in `datadog_ruby_common.h`.

Specifically, I think the criteria here should be -- "does it get used by both extensions?". I know it's tempting to start generalizing helpers and whatnot, but the downside of that is that our common code starts to accumulate more and more things which IMHO complicates maintenance and makes it harder to spot when code can be changed/deleted/etc.

So my suggestion would be to include the helper in `library_config.c` until we start using it in more places (and if it starts being used across the libdatadog_api, but not profiling, introduce a `libdatadog_helpers.h`)

ext/libdatadog_api/library_config.c
```diff
@@ -91,22 +120,45 @@ static VALUE _native_configurator_get(VALUE self) {
...
+  if (local_config_id.ptr != NULL) {
+    VALUE local_id = local_config_id.length > 0 ? rb_utf8_str_new_cstr(local_config_id.ptr) : Qnil;
+    rb_hash_aset(local_hash, ID2SYM(rb_intern("id")), local_id);
+  }
+  rb_hash_aset(local_hash, ID2SYM(rb_intern("config")), local_config_hash);
+
+  VALUE fleet_hash = rb_hash_new();
+  if (fleet_config_id.ptr != NULL) {
+    VALUE fleet_id = fleet_config_id.length > 0 ? rb_utf8_str_new_cstr(fleet_config_id.ptr) : Qnil;
+    rb_hash_aset(fleet_hash, ID2SYM(rb_intern("id")), fleet_id);
```
ivoanjo: Minor: Since the final outcome here is to produce Ruby strings, consider making above `local_config_id`/`fleet_config_id` `VALUE`s directly. I think that'll simplify a few things, and you can initialize them to `Qnil` as well so it's easy to detect when they're not set.

spec/datadog/core/configuration/stable_config_spec.rb
```diff
@@ -12,4 +14,98 @@
       end
     end
   end
+
+  describe '#configuration', skip: !LibdatadogHelpers.supported? do
+    context 'when libdatadog API is available' do
+      let(:tmpdir) { Dir.mktmpdir }
+      context 'with config_id' do
+        before do
+          Datadog::Core::Configuration::StableConfig.instance_variable_set(:@configuration, nil)
```
ivoanjo: Minor: Not needed? Repeated from above

```suggestion
```

spec/datadog/core/configuration/stable_config_spec.rb
```diff
@@ -12,4 +14,98 @@
...
+            apm_configuration_default:
+              DD_APPSEC_ENABLED: true
+            YAML
+          )
+
+          allow(Datadog::Core::Configuration::StableConfig).to receive(:extract_configuration).and_return(
+            Datadog::Core::Configuration::StableConfig::Configurator.new.
+              with_local_path(File.join(tmpdir, 'application_monitoring.yaml')).
+              with_fleet_path(File.join(tmpdir, 'managed/datadog-agent/stable/application_monitoring.yaml')).get
+          )
```
ivoanjo: Wait, is it intended that the file structure here ends up being

```
tmpdir/application_monitoring.yaml
tmpdir/managed/datadog-agent/stable/application_monitoring.yaml
```

?

Also, since we're specifying the paths for the code to use `is/there/a/lot/of/value/still/for/using/the/very/weird/very/deep/paths?`

spec/datadog/core/configuration/stable_config_spec.rb
```diff
@@ -12,4 +14,98 @@
...
+        end
+
+        it 'returns the configuration' do
+          expect(described_class.configuration).to eq(
+            {
+              local: {id: nil, config: {"DD_LOGS_INJECTION" => "false"}},
+              fleet: {id: nil, config: {"DD_APPSEC_ENABLED" => "true"}}
+            }
+          )
+        end
```
ivoanjo: So there's different approaches to testing. Having said that, my preferred approach here is to recommend factoring out all the common bits into the describe, and then just have the individual test cases change very slightly.

Why? Consider the current tests -- they're like 40-something lines long, and so when you're trying to understand what's being tested, you read the test description, then you need to go through the whole test to mentally "diff" what changed from the above, and then you can understand the output.

If we share the setup, the specific tests can be as simple as:

```ruby
context 'with config_id' do
  let(:config_id) { 'config_id: 12345' }

  it 'returns the configuration with a nil id' do
    expect(described_class.configuration).to eq(
      {
        local: {id: 12345, config: {"DD_LOGS_INJECTION" => "false"}},
        fleet: {id: 12345, config: {"DD_APPSEC_ENABLED" => "true"}}
      }
    )
  end
end

context 'without config_id' do
  let(:config_id) { '' }

  it 'returns the configuration with a nil id' do
    expect(described_class.configuration).to eq(
      {
        local: {id: nil, config: {"DD_LOGS_INJECTION" => "false"}},
        fleet: {id: nil, config: {"DD_APPSEC_ENABLED" => "true"}}
      }
    )
  end
end
```

...which IMHO makes it really really really easy to spot what changed, and how that impacts the outcome, without having to care about the boilerplate needed for setting up the test.

lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -220,13 +220,14 @@ def agent_transport(config)
...
-              seq_id: seq_id,
-            }
+          def conf_value(name, value, seq_id, origin)
+            result = {name: name, value: value, origin: origin, seq_id: seq_id}
+            if origin == 'fleet_stable_config'
+              result[:config_id] = Core::Configuration::StableConfig.configuration.dig(:fleet, :id)
+            elsif origin == 'local_stable_config'
+              result[:config_id] = Core::Configuration::StableConfig.configuration.dig(:local, :id)
+            end
+            result
```
ivoanjo: This will send `nil` `config_id`s -- is that acceptable/expected?

(Consider perhaps answering this question with a spec :P)

ivoanjo (commented):
Left a bunch of notes! I think this is quite reasonable, and IMO should be in good shape after another pass.

spec/datadog/core/configuration/stable_config_spec.rb
vpellan: This reflects what is used by the agent for stable config (`/etc/datadog-agent/managed/datadog-agent/stable/application_monitoring.yaml`). In these tests we can change it to `tmpdir/fleet_application_monitoring.yaml` especially now that we can specify a custom path or something like that but I think keeping the paths like that also help to differentiate local from fleet config files

ext/libdatadog_api/library_config.c
vpellan: Done in 9add6f8fd9388913413b871cec681e5a28a5021f !

ext/datadog_profiling_native_extension/datadog_ruby_common.h
vpellan: Moved in e84ffbd13de268f148f7118e2f44e32bb217d064 !

lib/datadog/core/telemetry/event/app_started.rb
vpellan: I thought that it was filtered out later but it seems not. Changed in 5776520090e0a5d011d61e8f4af8c0fdddf22840 !

ext/libdatadog_api/library_config.c
vpellan: Done in 3e16e555f44563f4f3409ab5866165d5c6e9cb33 and f341b0350fac6ead96d113216e4e284efad74b67 !

spec/datadog/core/configuration/stable_config_spec.rb
vpellan: Removed in fab1b158919e4b2c1ddd380ff0f978b8c75b5f1c

spec/datadog/core/configuration/stable_config_spec.rb
vpellan: I did it a bit differently than your solution as I still believe having the whole yaml file in each context also makes it clearer than just having the config_id but I factorized the biggest part in 5fbcdf45ee0aaaf76cf64a4e0426e549a533cb30

spec/datadog/core/configuration/stable_config_spec.rb
```diff
@@ -12,4 +14,93 @@
...
+      # local config
+      File.write(
+        File.join(tmpdir, 'application_monitoring.yaml'),
+        local_config_content
+      )
+      # fleet config
+      File.write(
+        File.join(tmpdir, 'managed/datadog-agent/stable/application_monitoring.yaml'),
+        fleet_config_content
+      )
```
ivoanjo: Minor: The local config / fleet config are already written in code via the names of the lets ;)

```suggestion
      File.write(
        File.join(tmpdir, 'application_monitoring.yaml'),
        local_config_content
      )
      File.write(
        File.join(tmpdir, 'managed/datadog-agent/stable/application_monitoring.yaml'),
        fleet_config_content
      )
```

ivoanjo (approved):
:+1: LGTM

spec/datadog/core/configuration/stable_config_spec.rb
ivoanjo: I'm not entirely convinced -- in particular, to my eyes, this detour is a bit distracting because it gets me thinking "wait, is the custom path somewhat needed" and the answer is not really but it's a very small nitpick so not a blocker.

spec/datadog/core/configuration/stable_config_spec.rb
ivoanjo: Yeap, LGTM :+1:

spec/datadog/core/configuration/stable_config_spec.rb
vpellan: Fixed in [51cebd8](https://github.com/DataDog/dd-trace-rb/pull/4769/commits/51cebd88514e564df3b5ec29b25e73c09ab2f7a8)

PR #4771, opened 2025-06-30
feat(tracer): add record_exception API

dubloom:
**What does this PR do?**
Add `record_exception` in the `dd-trace-rb` API. `record_exception` will add the exception information (message, exception_type and stacktrace) in a span event attached to a span. It is possible to call multiple times `record_exception` within the lifetime of a span. 

**Motivation:**
Some customers using Sentry complained they were not capable of recording an exception. It appears that if they are using the OTEL api from our tracer, they can. However, this is not directly documented in the tracer and the API is straightforward to implement. 

This PR is a follow up of #4567. The differences are:
- This PR is based on an RFC
- The dd `record_exception` is independent from the OTEL implementation
All the comments of the previous PR were taking into account.

**Change log entry**
Yes. Add `record_exception` API to capture and attach error information to spans via span events.

**Additional Notes:**
- The function itself does not do any attributes check as it is done by the `SpanEvent` class. 
- Symbols cannot be used for attributes key as it will make the steep check fails because span event attribute keys must be string.


docs/GettingStarted.md
```diff
@@ -288,6 +288,16 @@ current_span = Datadog::Tracing.active_span
 current_span.set_tag('my_tag', 'my_value') unless current_span.nil?
 ```
 
+You can record an exception in the current span. It will add the recorded exception as a span event.
+You can record multiple exceptions during the lifetime of a span.
+```ruby
+# e.g: recording an exception in the active span
+rescue => e
+  current_span = Datadog::Tracing.active_span
+  current_span&.record_exception(e, attributes: { foo: "bar" })
```
p-datadog: The PR description and below documentation says that `foo` must be a string here?

spec/datadog/tracing/span_operation_spec.rb
```diff
@@ -999,6 +999,107 @@
...
+          e,
+          attributes: {
+            custom_attr: 'value',
+            custom_attr2: { foo: 'bar' },
+            custom_attr3: [[1]],
+            custom_attr4: [1, 'foo'],
+            custom_attr5: 2 << 65,
+            custom_attr6: -2 << 65,
+            custom_attr7: Float::NAN,
+            custom_attr8: Float::INFINITY
```
p-datadog: These also are non-string keys?

docs/GettingStarted.md
dubloom: Good catch. I tried to use symbols and run steep:check at this end. Made the changes in the code but not in tests/docs.
Thank you

dubloom: Addressed in [454c377](https://github.com/DataDog/dd-trace-rb/pull/4771/commits/454c3775ed84f9f8a195e0b4c1dae7bb92b53f14)

spec/datadog/tracing/span_operation_spec.rb
dubloom: Addressed in [454c377](https://github.com/DataDog/dd-trace-rb/pull/4771/commits/454c3775ed84f9f8a195e0b4c1dae7bb92b53f14)

lib/datadog/tracing/span_operation.rb
```diff
@@ -297,6 +297,32 @@ def set_error(e)
...
+        event_attributes = {
+          'exception.type' => exc.type,
+          'exception.message' => exc.message,
+          'exception.stacktrace' => exc.backtrace,
+        }.merge(attributes)
+
+        span_event = SpanEvent.new(
+          'exception',
+          attributes: event_attributes,
+        )
```
Strech: I think it could be a single line

lib/datadog/tracing/span_operation.rb
```diff
@@ -297,6 +297,32 @@ def set_error(e)
...
+      #
+      # @return [void]
+      def record_exception(exception, attributes: {})
+        exc = Core::Error.build_from(exception)
+
+        event_attributes = {
+          'exception.type' => exc.type,
+          'exception.message' => exc.message,
+          'exception.stacktrace' => exc.backtrace,
+        }.merge(attributes)
```
Strech: I believe you want to use `merge!` for merging into a pre-defined Hash

spec/datadog/tracing/span_operation_spec.rb
```diff
@@ -999,6 +999,107 @@
...
+  describe '#record_exception' do
+    let(:error) { StandardError.new('test error').tap { |e| e.set_backtrace(['this is a backtrace']) } }
+
+    it 'creates a span event' do
+      begin
+        raise error
+      rescue => e
+        span_op.record_exception(e)
+        span_op.record_exception(e)
+      end
```
Strech: To be honest this doesn't make sense. Since `record_exception` is not providing exception handling, we should just create exception, stuff it with what we want and give it to the method. Raising exception here do nothing

Strech (commented):
I think exception recording performance could be improved and test suite should be improved by removing exception raising.

spec/datadog/tracing/span_operation_spec.rb
dubloom: addressed in [d1a7326](https://github.com/DataDog/dd-trace-rb/pull/4771/commits/d1a7326f7058ff82bd98ff7d572a906a291ed115)

lib/datadog/tracing/span_operation.rb
dubloom: addressed in [d1a7326](https://github.com/DataDog/dd-trace-rb/pull/4771/commits/d1a7326f7058ff82bd98ff7d572a906a291ed115)

I did it that way otherwise steep will throw an error

lib/datadog/tracing/span_operation.rb
dubloom: addressed in [d1a7326](https://github.com/DataDog/dd-trace-rb/pull/4771/commits/d1a7326f7058ff82bd98ff7d572a906a291ed115)

docs/GettingStarted.md
```diff
@@ -288,6 +288,16 @@ current_span = Datadog::Tracing.active_span
 current_span.set_tag('my_tag', 'my_value') unless current_span.nil?
 ```
 
+You can record an exception in the current span. It will add the recorded exception as a span event.
```
brett0000FF:
```suggestion
You can record an exception in the current span. It adds the recorded exception as a span event.
```

brett0000FF (approved):
Thanks for the update!

Separately, we probably need to do a edit/refresh of this page in the near future. For example, it's probably easier for users to find what they are looking for if we add h4s under "Enriching traces from nested methods" for each of the different cases we cover.

docs/GettingStarted.md
dubloom: addressed in [9ec7dd2](https://github.com/DataDog/dd-trace-rb/pull/4771/commits/9ec7dd26b2949b36dbe2b42c0dd5f584eb0d72c8)

dubloom:
Here is a screenshot of a span event with its attributes containing an array value after the native format serialization change:
Screenshot 2025-07-01 at 11 09 43

Strech (approved):
Awesome 👍🏼

lib/datadog/tracing/span_operation.rb
```diff
@@ -297,6 +297,28 @@ def set_error(e)
...
+      def record_exception(exception, attributes: {})
+        exc = Core::Error.build_from(exception)
+
+        event_attributes = {
+          'exception.type' => exc.type,
+          'exception.message' => exc.message,
+          'exception.stacktrace' => exc.backtrace,
+        }
+
+        @span_events << SpanEvent.new('exception', attributes: event_attributes.merge(attributes))
```
marcotc:
```suggestion
        @span_events << SpanEvent.new('exception', attributes: event_attributes.merge!(attributes)) # steep:ignore
```

@dubloom do the `merge!` as it is much faster and does not allocate objects: https://github.com/fastruby/fast-ruby?tab=readme-ov-file#hashmerge-vs-hash-code

marcotc (changes requested):
Just the change for merge!

PR #4774, opened 2025-07-01
[WIP] Add official system-tests workflow on pushes

cbeauchesne:
**What does this PR do?**

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**


ivoanjo:
To add context to this PR: this is an experiment so we can do an apples-to-apples comparison with the existing integration.

.github/workflows/system-tests-official.yml
```diff
@@ -0,0 +1,129 @@
...
+    - cron: "00 04 * * 2-6"
+
+# Default permissions for all jobs
+permissions: {}
+
+jobs:
+  changes:
+    name: Changes
+    # temporary condition to test official workflow
+    if: github.event_name == 'push' || (github.event_name == 'pull_request' && github.head_ref == 'cbeauchesne/system-tests-official')
```
TonyCTHsu: `github.head_ref` condition should probably be removed.

.github/workflows/system-tests-official.yml
```diff
@@ -0,0 +1,129 @@
...
+      contents: read
+      packages: write
+    with:
+      library: ruby
+      binaries_artifact: dd-trace-rb
+      desired_execution_time: 300  # 5 minutes
+      scenarios: ${{ matrix.scenarios }}
+      weblogs: ${{ matrix.weblogs }}
+      skip_empty_scenarios: true
+      ref: main
```
TonyCTHsu: To be deterministic, avoid using branch. 

Update `update-system-tests.yml` to keep it up-to-date

.github/workflows/system-tests-official.yml
```diff
@@ -0,0 +1,129 @@
...
+      # A workaround to trigger the workflow for pull requests from forked repository,
+      # which does not have access to secrets.
+      #
+      # This is also useful for testing the workflow without opening a pull request.
+      - tmp/*
+  pull_request:
+    branches:
+      - master
+  workflow_dispatch: {}
+  schedule:
```
TonyCTHsu: Is it necessary to schedule it periodically?

The pull request description mentioned `this new workflow only runs on pushes on master`, however, the code does not matched the description.

TonyCTHsu (changes requested):

.github/workflows/system-tests-official.yml
cbeauchesne: I need to keep it during the review, if some comments needs changes.

Even if it does not hurt, I can remove it just before the merge, as you wish.

.github/workflows/system-tests-official.yml
cbeauchesne: Done in [59f0358](https://github.com/DataDog/dd-trace-rb/pull/4774/commits/59f0358788a9acc079672376255fe11d38e5ac05)

.github/workflows/system-tests-official.yml
cbeauchesne: I was planning to do so only if the guild chose to WF, but here we go : [4ff6166](https://github.com/DataDog/dd-trace-rb/pull/4774/commits/4ff6166c02af3c31bab0a448bd0080542198f8af)

p-datadog (changes requested):
Two questions:

1) Why does this workflow not run on pull requests, if we are attempting an "apples to apples" comparison?

2) Does the exsiting runner run "1600 weblog X scenario"?

.github/workflows/system-tests-official.yml
```diff
@@ -0,0 +1,129 @@
+name: System Tests - Official
+on: # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - master
+      # A workaround to trigger the workflow for pull requests from forked repository,
+      # which does not have access to secrets.
```
p-datadog, line 7: Currently in our library nothing useful gets triggered from forks, does this PR change that behavior? Does it run all tests successfully without having secrets? How?

.github/workflows/system-tests-official.yml
```diff
@@ -0,0 +1,129 @@
...
+      #
+      # This is also useful for testing the workflow without opening a pull request.
+      - tmp/*
```
p-datadog, line 10: There is already a pull request adding the workflow, what is the purpose of testing it without a pull request? And what is the "also" referring to, what is the other purpose of this code?

.github/workflows/system-tests-official.yml
p-datadog: What is necessary from the `github.head_ref`?

.github/workflows/system-tests-official.yml
```diff
@@ -0,0 +1,129 @@
...
+    runs-on: ubuntu-24.04
+    outputs:
+      changes: ${{ steps.changes.outputs.src }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
+      - name: Changes
+        uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3.0.2
```
p-datadog, line 33: What is the purpose of this step?

.github/workflows/system-tests-official.yml
```diff
@@ -0,0 +1,129 @@
...
+          name: dd-trace-rb  # The name must match the folder name so it extracts to binaries/dd-trace-rb on download
+          path: packaged
+
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - weblogs: ""  # All of them
+            scenarios: DEFAULT,EVERYTHING_DISABLED,APPSEC_BLOCKING_FULL_DENYLIST,APPSEC_REQUEST_BLOCKING,APPSEC_BLOCKING,GRAPHQL_APPSEC
```
p-datadog, line 76: The idea of migrating to ST workflow is to run all scenarios. Why is a list of scenarios specified here?

.github/workflows/system-tests-official.yml
TonyCTHsu: >What is necessary from the github.head_ref?

I believe this is placed in order to have the test executed from this pull request event, otherwise, the path filter job would skip those tests. So it is for development purpose.

.github/workflows/system-tests-official.yml
cbeauchesne: I copy pasted what you have in your current workflow.

.github/workflows/system-tests-official.yml
cbeauchesne: Same, copy-paste from your current workflow

.github/workflows/system-tests-official.yml
cbeauchesne: > What is necessary from the github.head_ref?

To test the workflow in this PR

> What is the purpose of this step?

Copy paste from the current workflow. IIUC, it's to run system-tests only if a certain set of file is changed

p-datadog:
And one more question: does this PR authenticate to docker hub, if so how, if not how is it not subject to rate limits?

.github/workflows/system-tests-official.yml
cbeauchesne: To have something close to the current workflow in term of coverage. 

Extending the coverage should be adressed in a separated discussion/PR.

But as I mentionned in the PR description, running eveything works fine, but in 20'

cbeauchesne:
> And one more question: does this PR authenticate to docker hub, if so how, if not how is it not subject to rate limits?

It does, the workflow inherits from your secrets, and if docker username/password are present, it do the authentification : https://github.com/DataDog/system-tests/blob/main/.github/workflows/run-end-to-end.yml#L107-L115

.github/workflows/system-tests-official.yml
ivoanjo: Afaik this is so we can compare 1:1 with the current flow, since the discussion of performance was a bit thing.

It's not very "fair" if we compare the current flow that runs a subset of the test with the official workflow running all the tests ;)

cbeauchesne:
> Why does this workflow not run on pull requests, if we are attempting an "apples to apples" comparison?

To prevent any issue on PRs, and being able to decide to use it on PR after few days of runs on `master`.

> Does the exsiting runner run "1600 weblog X scenario"?

No, it run only 122 of them (details in PR description)

.github/workflows/system-tests-official.yml
p-datadog: I disagree.

Unauthenticated rate limit, which we hit periodically, requires a certain volume of requests. Not running the full test suite and not running it on each PR reduces the request volume (probably dramatically), thus this PR could appear to be "fine" but wouldn't work in practice.

This PR also does not address its claimed purpose of having dd-trace-rb run all tests always.

.github/workflows/system-tests-official.yml
p-datadog: What are the changes to upstream ST that are necessary for this workflow?

.github/workflows/system-tests-official.yml
ivoanjo: > This PR also does not address its claimed purpose of having dd-trace-rb run all tests always.

I'm not sure that's claimed to be the purpose? From the PR notes:

> [...] to allow a comparizon with the current setup

vpellan:
Is it possible to add support for the forced-tests-list.json file ? I use it often, and at least @Strech and @marcotc have also used it. I can help if that's needed, I also did something similar for [parametric-tests.yml](https://github.com/DataDog/dd-trace-rb/commit/df9dba87dc84751053b12467f70260ec74e427df), which uses system-tests official workflow

.github/workflows/system-tests-official.yml
p-datadog: The issue with the current setup was that it did not run all tests (our CI was green and later ST CI failed). Is there another issue that this PR is attempting to address?

p-datadog:
Where would I find in the logs the confirmation that docker hub requests are authenticated?

.github/workflows/system-tests-official.yml
ivoanjo: > The issue with the current setup was that it did not run all tests (our CI was green and later ST CI failed). Is there another issue that this PR is attempting to address?

IMO This PR is attempting to break a bigger migration into smaller steps. First, provide parity with the current setup, and then improve on it.

E.g. if the current integration isn't particularly good at running all the tests (and would run into limitations, etc), it seems weird IMO to put that as the required bar for this specific PR -- although I agree with you that it's a very much wanted item.

cbeauchesne:
> Where would I find in the logs the confirmation that docker hub requests are authenticated?

https://github.com/DataDog/dd-trace-rb/actions/runs/16024726999/job/45210078369#step:8:69

.github/workflows/system-tests-official.yml
cbeauchesne: > IMO This PR is attempting to break a bigger migration into smaller steps. First, provide parity with the current setup, and then improve on it.

Yes. Furthermore, base on @lloeki comment, it's useless to run those 1600 matrix points.

But in any case, here is a run with all of them : https://github.com/DataDog/dd-trace-rb/actions/runs/16004204072

cbeauchesne:
> Is it possible to add support for the forced-tests-list.json file ? I use it often, and at least @Strech and @marcotc have also used it. I can help if that's needed, I also did something similar for [parametric-tests.yml](https://github.com/DataDog/dd-trace-rb/commit/df9dba87dc84751053b12467f70260ec74e427df), which uses system-tests official workflow


@vpellan  Yes please :) ping me if you need any help

ivoanjo:
I'm curious about:

2. ST workflow needs to cover all of the configurations that our workflow covers

I thought it did? (I might be misunderstanding some part)

p-datadog:
I did not say the right thing and updated the comment: this PR claims to run all tests that our present workflow runs but the idea for changing to ST workflow was that it would run *all* tests so that we cannot have unintentionally un-executed tests. 

So the ST workflow should not require an enumeration of tests to run as this PR is presently doing.

ivoanjo:
> [...] the idea for changing to ST workflow was that it would run all tests so that we cannot have unintentionally un-executed tests.
>
> So the ST workflow should not require an enumeration of tests to run as this PR is presently doing.

Got it, thanks for explaining. TBH I was the one that pushed for a more apples-to-apples comparison, as in our discussions so far we had a lot of (very fair) concerns about CI performance regressions with the changeover.

So that one's on me! As you see in the current PR, it's more work to match our current set than it would be to run them all.

That said, enabling _all tests_ is definitely a trade-off, and does come with a clear CI time regression (+10 minutes). So my suggestion would be:

1. Put that constraint aside for this PR
2. Use to use the next Ruby guild meeting to decide if we're happy with the trade-off here

(To be clear: I really like speedy CI, but am happy to take the increased test coverage instead)

If we decide we're happy, it seems trivial to have a follow-up to enable all the tests.

cbeauchesne:
There is also another point : @lloeki told on slack that it's useless to run the 1600 matrix point.

As it's also a waste of ressource inside system-tests CI, I will define the set of useful  matrix point  inside system-tests.

It means that if the guild chose to go with this official workflow, the definition will be way more simple - and still fast.

.github/workflows/system-tests-official.yml
cbeauchesne: @p-datadog all comments must be resolved to be able to merge, can I resolve this one ?

.github/workflows/system-tests-official.yml
cbeauchesne: Addressed in [5a8cc4f](https://github.com/DataDog/dd-trace-rb/pull/4774/commits/5a8cc4f9bb6db69e4311d460bebb15410c7fd932)

PR #4776, opened 2025-07-02
fix(mysql): prevent `db.name` from being empty

zarirhamza:
Some queries to mysql DBs have an empty database name field due to it not being a mandatory parameter. For these cases, we are setting the db.name tag to be empty and creating an empty string inferred service.

**What does this PR do?**
Prevents `db.name` tag from being empty. This is a peer tag which intake will use to set inferred service. If empty, then inferred service will appear empty. This fix prevents that. 

**Change log entry**
Yes. The mysql integration now only sets the `db.name` tag if there is a valid value

**How to test the change?**
All changes have tests. There are queries with an empty DB name and the results are now as expected


p-datadog:
This PR needs test coverage, no? Also why was the PR template largely deleted? We need the change log entry for example since this is a customer-visible change.

p-datadog (approved):

PR #4779, opened 2025-07-03
Rack: Ensure proxy span always finishes

marcotc:
**What does this PR do?**

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**


marcotc:
@Strech do you think the new changlog entry looks better?

lib/datadog/tracing/contrib/rack/trace_proxy_middleware.rb
```diff
@@ -43,7 +43,13 @@ def call(env, configuration)
             # excluding the time spent processing the request itself
             queue_span.finish
 
-            yield.tap { request_span.finish }
+            yield
+          ensure
+            # Ensure that the spans are finished even if an exception is raised.
+            # **This is very important** to prevent the trace from leaking between requests,
+            # especially because `queue_span` is normally a root span.
+            queue_span&.finish
```
ivoanjo, line 51: :eyes: Is there any issue from potentially calling `queue_span.finish` twice?

p-datadog: The `finish` method appears to do nothing if the span is already finished, such that it seems fine to call `finish` multiple times.

p-datadog (approved):

spec/datadog/tracing/contrib/rack/trace_proxy_middleware_spec.rb
```diff
@@ -5,15 +5,17 @@
 RSpec.describe Datadog::Tracing::Contrib::Rack::TraceProxyMiddleware do
   describe '#call' do
     let(:service) { 'nginx' }
+    let(:env) { double('ENV') }
```
vpellan: This might be confusing as `env` in `TraceProxyMiddleware.call(env, configuration)` isn't ENV class but just an Hash containing the request environment. Maybe replace it with something like
```suggestion
    let(:env) { double('request environment') }
```

spec/datadog/tracing/contrib/rack/trace_proxy_middleware_spec.rb
```diff
@@ -5,15 +5,17 @@
...
+
+    before do
+      allow(Datadog::Tracing::Contrib::Rack::QueueTime).to receive(:get_request_start).with(env).and_return(timestamp)
```
vpellan: Minor:
Why do we stub `get_request_start` ? We could put the time in a hash in `env` (with `HTTP_X_REQUEST_START` key) instead of making it a double, so if we change the implementation of `get_request_start` we wouldn't miss it. In its current state, that would make the test fail because `get_request_start` cannot return a result as precise as `Time.now.utc` (as it only keeps microseconds, but `Time.now.utc` is giving a result precise to the nanosecond)
I can see two ways of solving that. Either using a hardcoded time value such as `1_000_000_000` and replace `Time.now` by `Time.at` for the timestamp, or truncate the 3 last digits of `Time.now.utc`.

lib/datadog/tracing/contrib/rack/trace_proxy_middleware.rb
marcotc: It's idempotent!

spec/datadog/tracing/contrib/rack/trace_proxy_middleware_spec.rb
marcotc: You are right, it's the rack env!

spec/datadog/tracing/contrib/rack/trace_proxy_middleware_spec.rb
marcotc: I did it all!

PR #4783, opened 2025-07-07
[NO-TICKET] Minor: Fix checking for dladdr in profiling

ivoanjo:
**What does this PR do?**

This PR fixes the checks for the `dladdr`/`dladdr1` function added in https://github.com/DataDog/dd-trace-rb/pull/4745 to actually check the values of the defines.

Specifically, the `have_...` helpers in `extconf.rb` behave differently from the other defines we set up.

That is, `have_func("dladdr")` will always define `HAVE_DLADDR`, even though it might define it to `0` (not available). (Our other checks only define things when they should be used, e.g.
`$defs << "-DNO_POSTPONED_TRIGGER" if RUBY_VERSION < "3.3"`).

This means that it's not correct to use `#ifdef HAVE_DLADDR` to check if `HAVE_DLADDR` exists and is == 1, we actually need to test it separately.

In this very specific case, in practice, I suspect we won't trigger issues because our extconf.rb is

```ruby
if have_header("dlfcn.h")
  (have_struct_member("struct link_map", "l_name", "link.h") && have_func("dladdr1")) ||
    have_func("dladdr")
end
```

...and thus if the header doesn't exist, we won't run the other checks, so I think only in a situation where the header exists and doesn't have the function would there be a problem (and I'm not sure that can happen on Linux).

Anyway, I believe it's still worth fixing as this is a footgun waiting to happen, and compilation errors are always mega annoying.

**Motivation:**

Fix potential compilation error when installing profiling.

**Change log entry**

Yes. Fix potential profiler compilation issue.

**Additional Notes:**

I've actually ran into this in the past in
https://github.com/ruby/ruby/commit/7472fff7f1b5296fbfcde0e2b7411a1d87781f3f

**How to test the change?**

If CI is green, this should be good!

### DIFF
--- a/ext/datadog_profiling_native_extension/collectors_stack.c
+++ b/ext/datadog_profiling_native_extension/collectors_stack.c
@@ -4,13 +4,12 @@
 
 #include "extconf.h" // This is needed for the HAVE_DLADDR and friends below
 
-// For dladdr/dladdr1
-#if defined(HAVE_DLADDR1) || defined(HAVE_DLADDR)
+#if (defined(HAVE_DLADDR1) && HAVE_DLADDR1) || (defined(HAVE_DLADDR) && HAVE_DLADDR)
   #ifndef _GNU_SOURCE
     #define _GNU_SOURCE
   #endif
   #include <dlfcn.h>
-  #ifdef HAVE_DLADDR1
+  #if defined(HAVE_DLADDR1) && HAVE_DLADDR1
     #include <link.h>
   #endif
 #endif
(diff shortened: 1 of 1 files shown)


p-datadog (approved):
Would it be less error-prone to `undef` these defines if they are set to zero?

ivoanjo:
> Would it be less error-prone to undef these defines if they are set to zero?

I think undef would be a bit confusing TBH... If we went down that route, I think it would be preferable to perhaps ignore Ruby's defines and use our own, e.g. do something like

```ruby
if have_func("dladdr1")
  $defs << "-DUSE_DLADDR1" 
elsif have_func("dladdr")
  $defs << "-DUSE_DLADDR" 
end
```

But it's not amazingly clean either. Any specific preference?

ivoanjo:
I'm going to go ahead and merge this one -- happy to change the strategy on a follow-up once we decide what option to use.

PR #4785, opened 2025-07-08
[PROF-12098] Profiler: Allow sampling from inside the signal handler to improve accuracy

ivoanjo:
**What does this PR do?**

This PR picks up from where https://github.com/DataDog/dd-trace-rb/pull/4750 left off and finishes the new "sample from inside the signal handler" feature.

Most of the preparation work was already included in the prior PR so this PR:
* Adds a new setting `c.profiling.advanced.sighandler_sampling_enabled` / `DD_PROFILING_SIGHANDLER_SAMPLING_ENABLED` environment variable
* Uses this setting to decide when to use `thread_context_collector_prepare_sample_inside_signal_handler`
* ...and it sprinkles a few compiler barriers to make sure we get a consistent view of the state of the profiler from inside the signal handler.

**Motivation:**

On older Ruby versions, the profiler wasn't able to sample from inside the signal handler. This is no longer the case since https://github.com/ruby/ruby/pull/3296 and https://github.com/ruby/ruby/pull/11036. Sampling from inside the signal handler increases profiler accuracy, and thus we want to start doing it on Rubies where it is safe to do so.

**Change log entry**

Yes. Profiler: Allow sampling from inside the signal handler to improve accuracy

**Additional Notes:**

In its current form, the feature is disabled by default. I'm still validating it in our testing environments, and I plan to submit a follow-up PR to enable it by default on Ruby 3.2.5+/3.3.4+ once validation finishes.

**How to test the change?**

I've added test coverage on this PR. Furthermore, here's a tiny worst-case reproducer that can show the difference quite clearly:

```ruby
def empty_method # gets 100% blame
end

def slow_method
  x = "h" + "e" + "l" + "l" + "o" + "," # Line 5
  x += "w" + "o" + "r" + "l" + "d"      # Line 6
  empty_method                          # Line 7
end

10000000.times do
  slow_method
end
```

Here's how this testcase looks without sampling from the signal handler:

| |
|-|
| image |

And here's how it looks with the feature enabled:
| |
|-|
| image |

You can see how the cpu-time gets correctly attributed to lines 5 and 6 with this change.

### DIFF
--- a/ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -102,6 +102,7 @@ typedef struct {
   bool allocation_counting_enabled;
   bool gvl_profiling_enabled;
   bool skip_idle_samples_for_testing;
+  bool sighandler_sampling_enabled;
   VALUE self_instance;
   VALUE thread_context_collector_instance;
   VALUE idle_sampling_helper_instance;
@@ -358,6 +359,7 @@ static VALUE _native_new(VALUE klass) {
   state->allocation_counting_enabled = false;
   state->gvl_profiling_enabled = false;
   state->skip_idle_samples_for_testing = false;
+  state->sighandler_sampling_enabled = false;
   state->thread_context_collector_instance = Qnil;
   state->idle_sampling_helper_instance = Qnil;
   state->owner_thread = Qnil;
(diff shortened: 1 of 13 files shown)


ivoanjo:
I've added the results from validation in https://datadoghq.atlassian.net/browse/PROF-12098 (Datadog-only link, sorry!).

TL;DR I could not visually spot any obvious differences in our testing app, the difference is even smaller than I expected.

Issue #4784, opened 2025-07-08
[BUG]: undefined method 'subdomain' for an instance of Rack::Request (NoMethodError)


pcothenet:
### Tracer Version(s)

2.18.0

### Ruby Version(s)

ruby 3.4.3 (2025-04-14 revision d0b7e5b6a0) +PRISM

### Bug Report

In the past few days, we've had spikes of error requests in our application. Requests from bots (including DD Synthetics) are triggering trace errors on innocuous routes (/favicon, /wp-admin/).

```
nomethoderror: undefined method 'subdomain' for an instance of Rack::Request
/usr/local/bundle/ruby/3.4.0/gems/actionpack-8.0.2/lib/action_dispatch/journey/route.rb:153:in 'block in ActionDispatch::Journey::Route#matches?': undefined method 'subdomain' for an instance of Rack::Request (NoMethodError)
	from /usr/local/bundle/ruby/3.4.0/gems/actionpack-8.0.2/lib/action_dispatch/journey/route.rb:150:in 'Hash#each'
	from /usr/local/bundle/ruby/3.4.0/gems/actionpack-8.0.2/lib/action_dispatch/journey/route.rb:150:in 'Enumerable#all?'
	from /usr/local/bundle/ruby/3.4.0/gems/actionpack-8.0.2/lib/action_dispatch/journey/route.rb:150:in 'ActionDispatch::Journey::Route#matches?'
	from /usr/local/bundle/ruby/3.4.0/gems/actionpack-8.0.2/lib/action_dispatch/journey/router.rb:121:in 'block in ActionDispatch::Journey::Router#find_routes'
	from /usr/local/bundle/ruby/3.4.0/gems/actionpack-8.0.2/lib/action_dispatch/journey/router.rb:121:in 'Array#select!'
	from /usr/local/bundle/ruby/3.4.0/gems/actionpack-8.0.2/lib/action_dispatch/journey/router.rb:121:in 'ActionDispatch::Journey::Router#find_routes'
	from /usr/local/bundle/ruby/3.4.0/gems/datadog-2.18.0/lib/datadog/tracing/contrib/action_pack/action_dispatch/instrumentation.rb:67:in 'Datadog::Tracing::Contrib::ActionPack::ActionDispatch::Instrumentation::Journey::LazyRouter#find_routes'
[...]

PR #4793, opened 2025-07-10
Fix an error in AppSec route extractor for not-found routes in Rails 8

y9v:
**What does this PR do?**
This PR fixes an exception in AppSec RouteExtractor that happens in Rails 8.0 and above for not-found routes. The error only happens if the application has any routes with constraints, such as the subdomain.

**Motivation:**
Issue #4784 

**Change log entry**
Yes. AppSec: Fix an error in RouteExtractor for not-found routes.

**Additional Notes:**
None.

**How to test the change?**
Manual testing with app generator and CI.


Strech (approved):
LGTM 👍🏼

PR #4794, opened 2025-07-10
[NO-TICKET] Tag profiles with sequence number

ivoanjo:
**What does this PR do?**

This PR adds a new tag to reported profiles: `profile_seq`. This tag provides a sequence number for reported profiles for a given runtime-id.

**Motivation:**

This tag has already been adopted by the Go and .NET profilers, and is useful for use-cases such as:

* Focus on, or exclude application boot time (which usually has quite different profiles)
* Clearly identify when profiles are missing (because e.g. they failed to be reported and there's gaps in the sequence)

**Change log entry**

Yes. Profiling: Tag profiles with sequence number

**Additional Notes:**

This feature is quite straightforward, the only weird part is that we carry state across profiler reconfigurations (e.g. state is a singleton), and we also need to reset it after the app forks, because runtime-id changes.

**How to test the change?**

### DIFF
--- a/lib/datadog/profiling/sequence_tracker.rb
+++ b/lib/datadog/profiling/sequence_tracker.rb
@@ -1,0 +1,40 @@
+# frozen_string_literal: true
+
+require_relative '../core/utils/forking'
+
+module Datadog
+  module Profiling
+    # Used to generate the `profile_seq` tag, which effectively counts how many profiles we've attempted to report
+    # from a given runtime-id.
+    #
+    # Note that the above implies a few things:
+    # 1. The sequence number only gets incremented when we decide to report a profile and create a `Flush` for it
+    # 2. The `SequenceTracker` must live across profiler reconfigurations and resets, since no matter how many
+    #    profiler instances get created due to reconfiguration, the runtime-id is still the same, so the sequence number
+    #    should be kept and not restarted from 0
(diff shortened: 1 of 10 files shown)


lib/datadog/profiling/sequence_tracker.rb
```diff
@@ -0,0 +1,40 @@
...
+    #
+    # Note that the above implies a few things:
+    # 1. The sequence number only gets incremented when we decide to report a profile and create a `Flush` for it
+    # 2. The `SequenceTracker` must live across profiler reconfigurations and resets, since no matter how many
+    #    profiler instances get created due to reconfiguration, the runtime-id is still the same, so the sequence number
+    #    should be kept and not restarted from 0
+    # 3. The `SequenceTracker` must be reset after a fork, since the runtime-id will change, and we want to start
+    #    counting from 0 again
+    #
+    # This is why this module is implemented as a singleton that we reuse, not as an instance that we recreate.
```
ivoanjo, line 18:
```suggestion
    # This is why this module is implemented as a singleton that we reuse, not as an instance that we recreate.
    #
    # Note that this module is not thread-safe, so it's up to the callers to make sure
    # it's only used by a single thread at a time (which is what the `Profiling::Exporter`)
    # is doing.
```

p-datadog (approved):

PR #4802, opened 2025-07-14
[APPSEC-58115] Add Business Logic Events SDK v2

Strech:
**What does this PR do?**

Introduces second version of the Business Logic Events SDK.

**Motivation:**

We are pushing better SDK API that will suite to the actual data most of the time available at hand when emitting business logic events.

**Change log entry**

Yes. AppSec: Added Business Logic Events SDK v2.

**Additional Notes:**

There is a shortcut we take in order to compute service entry trace/span. In V2 you can control it, but in V2 there is no such functionality. Unfortunately, tracer doesn't support such information, even tho there is a functionality we would like to have, but it's hidden on different layer of the stack.

**How to test the change?**

CI + ST should cover 99% 😬

### DIFF
--- a/lib/datadog/kit/appsec/events/v2.rb
+++ b/lib/datadog/kit/appsec/events/v2.rb
@@ -55,8 +55,8 @@ module Datadog
 
               user_attributes = build_user_attributes(user_or_id, login)
 
-              export_tags(span, metadata, namespace: LOGIN_SUCCESS_EVENT)
-              export_tags(span, user_attributes, namespace: "#{LOGIN_SUCCESS_EVENT}.usr")
+              set_span_tags(span, metadata, namespace: LOGIN_SUCCESS_EVENT)
+              set_span_tags(span, user_attributes, namespace: "#{LOGIN_SUCCESS_EVENT}.usr")
               span.set_tag('appsec.events.users.login.success.track', 'true')
               span.set_tag('_dd.appsec.events.users.login.success.sdk', 'true')
 
@@ -110,7 +110,7 @@ module Datadog
                 raise TypeError, '`user_exists` argument must be a boolean'
               end
 
-              export_tags(span, metadata, namespace: LOGIN_FAILURE_EVENT)
+              set_span_tags(span, metadata, namespace: LOGIN_FAILURE_EVENT)
               span.set_tag('appsec.events.users.login.failure.track', 'true')
               span.set_tag('_dd.appsec.events.users.login.failure.sdk', 'true')
               span.set_tag('appsec.events.users.login.failure.usr.login', login)
@@ -161,12 +161,11 @@ module Datadog
               end
             end
 
-            def export_tags(span, source, namespace:)
-              namespace = "appsec.events.#{namespace}"
-              source.each do |key, value|
+            def set_span_tags(span, tags, namespace:)
+              tags.each do |name, value|
                 next if value.nil?
 
-                span.set_tag("#{namespace}.#{key}", value)
+                span.set_tag("appsec.events.#{namespace}.#{name}", value)
               end
             end
 
--- a/lib/datadog/kit/appsec/events/v2.rb
+++ b/lib/datadog/kit/appsec/events/v2.rb
@@ -1,0 +1,144 @@
+# frozen_string_literal: true
+
+require_relative '../../identity'
+
+module Datadog
+  module Kit
+    module AppSec
+      module Events
+        # The second version of Business Logic Events SDK
+        module V2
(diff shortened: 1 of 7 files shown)


lib/datadog/kit/appsec/events/v2.rb
```diff
@@ -0,0 +1,196 @@
...
+                raise ArgumentError, 'missing required user key `:id`' unless user_or_id.key?(:id)
+                raise TypeError, 'user key `:id` must be a String' unless user_or_id[:id].is_a?(String)
+
+                user_or_id.merge(login: login)
+              else
+                raise TypeError, '`user_or_id` argument must be either String or Hash'
+              end
+            end
+
+            def export_tags(span, source, namespace:)
```
y9v: what do you think if we rename `source` to `tags_data` or something similar?

lib/datadog/kit/appsec/events/v2.rb
```diff
@@ -0,0 +1,196 @@
...
+            def build_user_attributes(user_or_id, login)
+              raise TypeError, '`login` argument must be a String' unless login.is_a?(String)
+
+              case user_or_id
+              when nil
+                { login: login }
+              when String
+                { login: login, id: user_or_id }
+              when Hash
+                raise ArgumentError, 'missing required user key `:id`' unless user_or_id.key?(:id)
```
y9v, line 155: do we also want it to work with string keys, or we don't care?

lib/datadog/kit/appsec/events/v2.rb
```diff
@@ -0,0 +1,196 @@
...
+
+              user = ::Datadog::AppSec::Instrumentation::Gateway::User.new(nil, login)
+              ::Datadog::AppSec::Instrumentation.gateway.push('identity.set_user', user)
+            end
+
+            private
+
+            # NOTE: Current tracer implementation does not provide a way to
+            #       get the service entry span. This is a shortcut we take now.
+            def service_entry_trace
```
y9v, line 132: I would propose to rename `trace` and `span` getters and setters on `Datadog::AppSec.active_context` to `service_entry_span` and `service_entry_trace`. We are now creating the context in the service-level instrumentation, and I think it would be more clear to see which trace and span we are setting there.

Strech: I don't think it will be part of this PR, but sure we can do that.

lib/datadog/kit/appsec/events/v2.rb
Strech: Nope, the API of `Identity.set_user` require it to be Symbol, and I would enforce Symbols at places that makes sense.

lib/datadog/kit/appsec/events/v2.rb
Strech: Will look for a better name 👍🏼

lib/datadog/kit/appsec/events/v2.rb
```diff
@@ -0,0 +1,195 @@
...
+            #   )
+            #
+            # @return [void]
+            def track_user_login_success(login, user_or_id = nil, metadata = {})
+              trace = service_entry_trace
+              span = service_entry_span
+
+              if trace.nil? || span.nil?
+                return Datadog.logger.warn(
+                  'Kit::AppSec: Tracing is not enabled. Please enable tracing if you want to track events'
```
p-datadog, line 49: Could this spam customer logs too much?

lib/datadog/kit/appsec/events/v2.rb
```diff
@@ -0,0 +1,195 @@
...
+
+                span.set_tag("appsec.events.#{namespace}.#{name}", value)
+              end
+            end
+
+            # TODO: In case if we need to introduce telemetry metrics to the SDK v1
+            #       or highly increase the number of metrics, this method should be
+            #       extracted into a proper module.
+            def record_event_telemetry_metric(event)
+              telemetry = ::Datadog.send(:components)&.telemetry
```
p-datadog: Are you intending to create the full component tree here if it's not already initialized, or query the existing tree?

lib/datadog/kit/appsec/events/v2.rb
Strech: Yes, but this will happen only if customer tries to use business logic events and tracing is not enabled. Which mean we have to let customer know that it's not working

lib/datadog/kit/appsec/events/v2.rb
Strech: I think just the telemetry, not the whole tree, will take a look how you did it last time without init the whole tree

p-datadog (approved):

PR #4808, opened 2025-07-15
DEBUG-3914 serialize instance variables under "self"

p-datadog:
**What does this PR do?**

Fixes serialization of instance variables to be compliant with the snapshot RFC: the values are supposed to be under "arguments", not "locals", and serialized as an object with key "self" (or "this" for languages like Java).

**Motivation:**

Specification compliance

**Change log entry**
Yes: live debugger/dynamic instrumentation: fix serialization of instance variables

**Additional Notes:**

This PR also adds `serialized_` prefixes to variables that contain serialized values, to improve clarity of the code where some values are serialized and some are not.

Also, `throwable` is removed from method parameters as it does not make sense there - it's supposed to contain the exception that is raised by the method (this is not yet implemented in dd-trace-rb), and so is included in the return value group;  the exception cannot be raised at the point when the method begins execution.

**How to test the change?**
Unit tests updated

### DIFF
--- a/spec/datadog/di/instrumenter_spec.rb
+++ b/spec/datadog/di/instrumenter_spec.rb
@@ -611,7 +611,8 @@ RSpec.describe Datadog::DI::Instrumenter do
           expect(observed_calls.first[:duration]).to be_a(Float)
 
           expect(observed_calls.first[:serialized_entry_args]).to eq(
-            kwarg: {type: 'Integer', value: '42'}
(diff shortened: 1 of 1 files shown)


p-datadog:
@Strech The serializer does not actually store `self`. The write to `self` is of serialized data which are primitive Ruby objects (strings/integers/etc.) and there are no references to the actual `self` retained. Serializer itself doesn't store the arguments being serialized either.

PR #4811, opened 2025-07-17
[PROF-12141] Upgrade to libdatadog 19.1

ivoanjo:
**What does this PR do?**

This PR bumps the libdatadog version to 19.1 --
https://github.com/DataDog/libdatadog/pull/1146 .

**Motivation:**

Make sure we're up-to-date with the latest libdatadog goodies. In particular, this version brings faster crashtracking reporting (which is nice because otherwise it could time out and not report data, or only report partial data).

**Change log entry**

Yes. Upgrade libdatadog dependency to 19.1.

**Additional Notes:**

N/A

**How to test the change?**

We have test coverage for features that depend on libdatadog, so green CI is good!

### DIFF
--- a/datadog.gemspec
+++ b/datadog.gemspec
@@ -69,7 +69,7 @@ Gem::Specification.new do |spec|
 
   # When updating the version here, please also update the version in `libdatadog_extconf_helpers.rb`
   # (and yes we have a test for it)
-  spec.add_dependency 'libdatadog', '~> 18.1.0.1.0'
+  spec.add_dependency 'libdatadog', '~> 19.1.0.1.0'
 
   # Will no longer be a default gem on Ruby 3.5, see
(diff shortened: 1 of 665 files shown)


ivoanjo:
Interesting, crashtracking symbolication seems to be having a few hiccups, looking into it.

ivoanjo:
I've bisected the regression to https://github.com/DataDog/libdatadog/pull/1046 : it seems to no longer properly symbolicate the stack on older Rubies (probably not due to Ruby, but because of older Linux base image/compiler).

I've reported this to the crashtracking folks; since we're in no rush to upgrade to 19.1 let's wait until they've been able to look into this.

Strech (approved):

ivoanjo:
Fix is in https://github.com/DataDog/libdatadog/pull/1154 . I'll close this PR for now, let's wait until the next libdatadog version to upgrade.

PR #4813, opened 2025-07-17
[PROF-12143] Report GC tuning environment variables with profiles

ivoanjo:
**What does this PR do?**

This PR tweaks the information that gets attached to every profile to include any GC tuning settings that were present.

This information can be seen in the "System Info" section of the single profile view.

**Motivation:**

As more GC tuning settings get added to Ruby, it's useful for performance reasons to know if a given app is operating with the defaults or something has been tweaked.

See <https://mailchi.mp/railsspeed/stop-tuning-rubys-gc> for an interesting discussion of this.

We want to make it easier to spot these GC tuning settings (maybe even automatically in the future?).

**Change log entry**

Yes. Report GC tuning environment variables with profiles.

**Additional Notes:**

Note that only settings that have been changed from the default are sent, as it doesn't seem relevant to send a bunch of empty settings.

**How to test the change?**

I've added test coverage; here's how it looks in staging:

image

### DIFF
--- a/spec/datadog/profiling/collectors/info_spec.rb
+++ b/spec/datadog/profiling/collectors/info_spec.rb
@@ -105,6 +105,28 @@ RSpec.describe Datadog::Profiling::Collectors::Info do
         end
       end
     end
+
+    describe "gc tuning reporting" do
+      context "when no gc tuning env vars are set" do
+        it "reports an empty hash" do
+          expect(info.fetch(:runtime).fetch(:gc_tuning)).to eq({})
+        end
+      end
+
+      context "when some gc tuning env vars are set" do
+        around do |example|
+          ClimateControl.modify("RUBY_GC_HEAP_FREE_SLOTS" => "12345") do
+            example.run
+          end
+        end
+
+        it "reports the gc tuning env vars" do
+          expect(info.fetch(:runtime).fetch(:gc_tuning)).to eq({
+            RUBY_GC_HEAP_FREE_SLOTS: "12345"
+          })
+        end
+      end
+    end
   end
 end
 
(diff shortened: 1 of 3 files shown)


p-datadog (approved):

spec/datadog/profiling/collectors/info_spec.rb
```diff
@@ -105,6 +105,28 @@
         end
       end
     end
+
+    describe "gc tuning reporting" do
+      context "when no gc tuning env vars are set" do
+        it "reports an empty hash" do
+          expect(info.fetch(:runtime).fetch(:gc_tuning)).to eq({})
```
Strech, line 112: Not much, but shorter


```suggestion
          expect(info.dig(:runtime, :gc_tuning)).to eq({})
```

spec/datadog/profiling/collectors/info_spec.rb
```diff
@@ -105,6 +105,28 @@
...
+
+      context "when some gc tuning env vars are set" do
+        around do |example|
+          ClimateControl.modify("RUBY_GC_HEAP_FREE_SLOTS" => "12345") do
+            example.run
+          end
+        end
+
+        it "reports the gc tuning env vars" do
+          expect(info.fetch(:runtime).fetch(:gc_tuning)).to eq({
```
Strech, line 124:
```suggestion
          expect(info.dig(:runtime, :gc_tuning)).to eq({
```

Strech (approved):
LGTM 👍🏼

spec/datadog/profiling/collectors/info_spec.rb
ivoanjo: Actually I did do it like this on purpose -- I wanted a `KeyError: key not found: :gc_tuning (KeyError)` (or similar) if it failed, rather than a `nil does not eq {}`. Small but... why not ;)

Strech: Fair enough 👍🏼

PR #4814, opened 2025-07-18
[NO-TICKET] Fix emitting duplicate warnings on agent configuration mismatch

ivoanjo:
**What does this PR do?**

When conflicting agent settings are provided, we emit a warning about it:

```
$ DD_AGENT_HOST=hello DD_TRACE_AGENT_URL=http://world:12345 bundle exec ddprofrb exec ruby -e "sleep 1"
W, [2025-07-18T09:27:22.611475 #80060]  WARN -- datadog: [datadog] Configuration mismatch: values differ between DD_TRACE_AGENT_URL environment variable ("world") and DD_AGENT_HOST environment variable ("hello"). Using "world" and ignoring other configuration.
I, [2025-07-18T09:27:22.620557 #80060]  INFO -- datadog: [datadog] DATADOG CONFIGURATION - CORE - {"date":"2025-07-18T08:27:22Z","os_name":"x86_64-pc-linux","version":"2.19.0","lang":"ruby","lang_version":"3.2.5","env":null,"service":"-e","dd_version":null,"debug":false,"tags":null,"runtime_metrics_enabled":false,"vm":"ruby-3.2.5","health_metrics_enabled":false,"profiling_enabled":false,"dynamic_instrumentation_enabled":false}
W, [2025-07-18T09:27:22.621957 #80060]  WARN -- datadog: [datadog] Configuration mismatch: values differ between DD_TRACE_AGENT_URL environment variable ("world") and DD_AGENT_HOST environment variable ("hello"). Using "world" and ignoring other configuration.
W, [2025-07-18T09:27:23.622573 #80060]  WARN -- datadog: [datadog] Configuration mismatch: values differ between DD_TRACE_AGENT_URL environment variable ("world") and DD_AGENT_HOST environment variable ("hello"). Using "world" and ignoring other configuration.
```

Unfortunately, as you can see above, the warning gets repeated 3 times.

This PR fixes this issue by ensuring that the
`AgentSettingsResolver`, responsible for emitting the logs, only gets called once during library setup.

**Change log entry**

Yes. Fix emitting duplicate warnings on agent configuration mismatch 

**Motivation:**

Repeated warnings are confusing, noisy and don't look very good.

**Additional Notes:**

N/A

**How to test the change?**

I've updated our test coverage (even fixed a spec that wasn't doing anything) + the command above can be used to validate that only one warning gets emitted now.


lib/datadog/core/telemetry/component.rb
```diff
@@ -96,6 +98,8 @@ def initialize( # standard:disable Metrics/MethodLength
             logger: logger,
             shutdown_timeout: settings.telemetry.shutdown_timeout_seconds,
           )
+
+          @agent_settings = agent_settings
```
p-datadog, line 102: Would it be safer to do this assignment prior to the telemetry creation code, e.g. after line 53?

spec/datadog/core/telemetry/event/synth_app_client_configuration_change_spec.rb
```diff
@@ -3,10 +3,10 @@
...
-  let(:id) { double('seq_id') }
-  let(:event) { event_class.new }
-
-  subject(:payload) { event.payload }
+RSpec.describe Datadog::Core::Telemetry::Event::SynthAppClientConfigurationChange do
+  subject(:event) { described_class.new(agent_settings: instance_double(Datadog::Core::Configuration::AgentSettings, adapter: Datadog::Core::Configuration::Ext::Agent::HTTP::ADAPTER)) }
 
+  it 'contains only the configuration' do
+    expect(event.payload.keys).to eq([:configuration])
+  end
```
p-datadog, line 11: Merci

lib/datadog/core/telemetry/component.rb
ivoanjo: Hmm... I don't think it makes a difference... From what I see, `start` needs `@worker` and `@worker` is the assignment above this one... :thinking: 

Or am I missing something? :eyes:

Strech: ~I think Oleg means that it's an instance variable and if we are about to assign it, does it make sense to move assignment outside of the if condition?~

If we don't need that variable for non-enabled telemetry, it's GTG 👍🏼 (sorry I misread the padding)

Strech (approved):
👍🏼

lib/datadog/core/telemetry/component.rb
ivoanjo: > If we don't need that variable for non-enabled telemetry, it's GTG 👍🏼 (sorry I misread the padding)

Yeah we don't!

---

I discussed this with Oleg over PM, and the TL;DR is that he was thinking that since we have an assignment to `telemetry_specific_agent_settings`, if we saved the attribute before we did the assignment, we wouldn't need the rename and we could assign to `agent_settings` still.

That's indeed possible, yet we agreed that since there are two "kinds" of agent settings: 1. What agent settings the library is configured with; and 2. What agent settings *telemetry* is configured with, the current implementation works nicely as it has different names for both (even though they're both instances of agent settings).

PR #4815, opened 2025-07-18
fix(benchmarks): remove http stubs

dubloom:
**What does this PR do?**
Replace stubbed HTTP call by function call in error tracking API benchmarks

**Motivation:**
The goal of this benchmarks is to test the feature against a simulation of HTTP requests. Before I was using WebMock to fake an HTTP request. However this led to some flakiness:
https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-rb/-/jobs/1033687618

As mocking the request is not mandatory, I replaced it with a function call to remove the flakiness

**Change log entry**
No.

### DIFF
--- a/benchmarks/error_tracking_api.rb
+++ b/benchmarks/error_tracking_api.rb
@@ -6,8 +6,7 @@ return unless __FILE__ == $PROGRAM_NAME || VALIDATE_BENCHMARK_MODE
 require_relative 'benchmarks_helper'
 require 'datadog'
 require 'benchmark'
-require 'net/http'
-require 'webmock'
+
 class ErrorTrackingApiBenchmark
   module NoopWriter
     def write(trace)
@@ -24,16 +23,13 @@ class ErrorTrackingApiBenchmark
 
   def initialize
     ::Datadog::Tracing::Writer.prepend(NoopWriter)
+  end
 
-    # Enable WebMock but allow connections to the Datadog agent
-    WebMock.disable_net_connect!(allow: ['testagent:9126', 'localhost:9126', '127.0.0.1:9126'])
-
-    WebMock.stub_request(:any, 'http://example.com/test').to_return do
-      # Sleep for 50ms to simulate network latency
-      sleep(0.05)
-
-      { status: 200, body: 'OK' }
-    end
+  def simulate_http_request
+    # Simulate some work that might be done in a real application
+    # Sleep for 50ms to simulate network latency or processing time
+    sleep(0.05)
+    { status: 200, body: 'OK' }
   end
 
   def benchmark_with_http_request_simulation_no_error_tracking
@@ -42,7 +38,7 @@ class ErrorTrackingApiBenchmark
 
       x.report('without error tracking with http') do
         Datadog::Tracing.trace('http.request') do
-          Net::HTTP.get_response(URI('http://example.com/test'))
+          simulate_http_request
           begin
             raise 'Test error'
           rescue
@@ -66,7 +62,7 @@ class ErrorTrackingApiBenchmark
 
       x.report('error tracking with http - all') do
         Datadog::Tracing.trace('http.request') do
-          Net::HTTP.get_response(URI('http://example.com/test'))
+          simulate_http_request
           begin
             raise 'Test error'
           rescue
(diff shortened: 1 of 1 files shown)


benchmarks/error_tracking_api.rb
```diff
@@ -24,16 +23,13 @@ def benchmark_time(time: 20, warmup: 2)
...
-      sleep(0.05)
-
-      { status: 200, body: 'OK' }
-    end
+  def simulate_http_request
+    # Simulate some work that might be done in a real application
+    # Sleep for 50ms to simulate network latency or processing time
+    sleep(0.05)
+    { status: 200, body: 'OK' }
   end
```
ivoanjo, line 33: Hey! Thanks for fixing this but... I'm left a bit confused -- this is a benchmark that has sleeps in it? :eyes:. What are we measuring here exactly?

Even the `benchmark_time` method already states:
> _@param [Integer] time in seconds. The default is 20 seconds.Because benchmark-ips tries to run one iteration per 100ms, this means we'll have around 200 samples (give or take a small margin of error)._

If we already know the result of this benchmark... is this adding value?

dubloom: My goal with this benchmark was to mesure the overhead of the feature is in a situation which is more realistic than just a try/catch alone. If you feel it does not make sense I'll not be against deleting it
I think the param comment is leftover from other benchmarks when I created it.

ivoanjo: > If you feel it does not make sense I'll not be against deleting it

The thing is, if I disable the sleep, here's the results:

```
ivo.anjo@rubyshade:~/datadog/second-dd-trace-rb$ be ruby benchmarks/error_tracking_api.rb 
Current pid is 464096
ruby 3.2.5 (2024-07-26 revision 31d0f1a2e7) [x86_64-linux]
error - without error tracking with http
                    I, [2025-07-18T14:06:28.336962 #464120]  INFO -- datadog: [datadog] DATADOG CONFIGURATION - CORE - {"date":"2025-07-18T13:06:28Z","os_name":"x86_64-pc-linux","version":"2.19.0","lang":"ruby","lang_version":"3.2.5","env":null,"service":"error_tracking_api","dd_version":null,"debug":false,"tags":null,"runtime_metrics_enabled":false,"vm":"ruby-3.2.5","health_metrics_enabled":false,"profiling_enabled":false,"dynamic_instrumentation_enabled":false}
     1.000 i/100ms
error - without error tracking with http
                         19.845 (± 0.0%) i/s   (50.39 ms/i) -    397.000 in  20.005458s
ivo.anjo@rubyshade:~/datadog/second-dd-trace-rb$ be ruby benchmarks/error_tracking_api.rb 
Current pid is 464962
ruby 3.2.5 (2024-07-26 revision 31d0f1a2e7) [x86_64-linux]
error - without error tracking with http - without sleep
                    I, [2025-07-18T14:06:57.454495 #464985]  INFO -- datadog: [datadog] DATADOG CONFIGURATION - CORE - {"date":"2025-07-18T13:06:57Z","os_name":"x86_64-pc-linux","version":"2.19.0","lang":"ruby","lang_version":"3.2.5","env":null,"service":"error_tracking_api","dd_version":null,"debug":false,"tags":null,"runtime_metrics_enabled":false,"vm":"ruby-3.2.5","health_metrics_enabled":false,"profiling_enabled":false,"dynamic_instrumentation_enabled":false}
     1.573k i/100ms
error - without error tracking with http - without sleep
                         16.416k (± 4.7%) i/s   (60.92 μs/i) -    328.757k in  20.071172s

Comparison:
error - without error tracking with http - without sleep:    16415.9 i/s
error - without error tracking with http:       19.8 i/s - 827.22x  slower
```

It looks to me we're putting an ant together with an elephant on the scale, and then looking at the scale and checking if there's any differences based on the ant skipping breakfast or having a really nice continental breakfast :eyes: 

So my suggestion would be to either delete or fix the benchmark. As-is I don't think there's any value in the data we're getting, unless we accidentally do the equivalent of pointing a sci-fi giant growth ray at the ant.

ivoanjo:
Superceded by https://github.com/DataDog/dd-trace-rb/pull/4816

PR #4819, opened 2025-07-21
[PROF-11898] Switch profiler stack truncation strategy and improve sampling performance

ivoanjo:
**What does this PR do?**

This PR changes how we do stack sampling and truncation in the profiler by:

1. Removing the count of frames truncated from the reported stack. This allows us to get rid of the `stack_depth_for` which actually had an off-by-one bug (at least -- there may have been other bugs hiding in there because it calculated the number of frames differently from `rb_profile_frames`)
2. Moving the truncated frames marker to the leaf frame, mirroring the behavior in the Python profiler
3. Reverse the direction of stack sampling to improve cache hit performance

**Motivation:**

I started this PR to fix the off-by-one in the "frames omitted frame" and to align the Ruby stack truncation behavior with Python, and realized that the new truncation logic could also be taken advantage of to improve sampling performance.

The amount of work we save when we can reuse a cached frame is still somewhat small, but I expect that, when we move to the new interning APIs in libdatadog, we'll benefit even more since e.g. we'll be able to reuse ids or entire stacks.

**Change log entry**

Yes. Switch profiler stack truncation strategy and improve sampling performance

**Additional Notes:**

Up until now, truncated stacks looked like this:

> image-20250717-144640

This behavior has been unchanged for years, and it could cause some confusion as it prevents stacks from merging (e.g. by having a different "N frames omitted" value).

In fact, Ruby was alone among the our profilers in having this behavior.

After this PR, truncation looks like this (when caused by the library):

> image

Truncating the leaf and not the root allows us to sample in a way that keeps our buffer stable and reusable across calls. I've added a new benchmark to show this, and on my test machine I see a speed-up of around 8%:

```
ruby 3.4.4 (2025-05-14 revision a38531fd3f) +PRISM [x86_64-linux]
tmp - stack collector (varying depth) master-2
                       131.000 i/100ms
tmp - stack collector (varying depth) master-2
                          1.324k (± 1.4%) i/s  (755.22 μs/i) -     13.362k in  10.093502s

Comparison:
tmp - stack collector (varying depth) reverse-stack-2:     1431.4 i/s
tmp - stack collector (varying depth) reverse-stack-1:     1407.3 i/s - same-ish: difference falls within error
tmp - stack collector (varying depth) master-2:     1324.1 i/s - 1.08x  slower
tmp - stack collector (varying depth) master-1:     1319.6 i/s - 1.08x  slower
```

**How to test the change?**

I've updated our test coverage to validate the new behavior. Here's also a tiny Ruby script I used when I wanted to see this in the UX:

```ruby
def thread_with_very_deep_stack(depth: 500)
  deep_stack = proc do |n|
    if n > 0
      deep_stack.call(n - 1)
    else
      sleep
    end
  end

  Thread.new { deep_stack.call(depth) }.tap { |t| t.name = "Deep stack #{depth}" }
end

thread = thread_with_very_deep_stack()
sleep 1
```

### DIFF
--- a/benchmarks/profiling_sample_loop_v2.rb
+++ b/benchmarks/profiling_sample_loop_v2.rb
@@ -109,7 +109,7 @@ class ProfilerSampleLoopBenchmark
         end
       end
 
-      x.save! "#{File.basename(__FILE__)}-results.json" unless VALIDATE_BENCHMARK_MODE
+      x.save! "#{File.basename(__FILE__)}-#{mode}-results.json" unless VALIDATE_BENCHMARK_MODE
       x.compare!
     end
 
@@ -132,7 +132,7 @@ class ProfilerSampleLoopBenchmark
         add_extra_frame_and_sample(collector) # This makes the stack change
       end
 
-      x.save! "#{File.basename(__FILE__)}-results.json" unless VALIDATE_BENCHMARK_MODE
+      x.save! "#{File.basename(__FILE__)}-varying-depth-results.json" unless VALIDATE_BENCHMARK_MODE
       x.compare!
     end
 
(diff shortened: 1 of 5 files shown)


ivoanjo:
(Force-pushed after rebasing on top of latest master to bring unrelated fixes for things that were breaking CI)

PR #4822, opened 2025-07-22
Add span events documentation

marcotc:
**What does this PR do?**

Adds documentation for how to add SpanEvents to APM spans. This feature has exited for a bit, but was undocumented.

**Change log entry**
No

### DIFF
--- a/docs/GettingStarted.md
+++ b/docs/GettingStarted.md
@@ -306,6 +306,23 @@ You can also get the current active trace using the `active_trace` method. This 
 current_trace = Datadog::Tracing.active_trace
 ```
 
+### Adding span events
+
+Span events are timestamped annotations attached to a span, useful for recording structured logs, exceptions, or custom events during a span's lifetime. You can create and add custom span events to the current active span as follows:
+
+```ruby
+event = Datadog::Tracing::SpanEvent.new(
+ "custom.event.name",
+ attributes: { "key1" => "value1", "key2" => 123 }
+)
+
+span.span_events << event
+```
+
+You can add multiple events to a span. Each event **must** include a name, an optional set of attributes, and an optional timestamp (in nanoseconds). If no timestamp is provided, the current time is used.
+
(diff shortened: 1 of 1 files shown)


soulcutter:
Thanks for this one!

buraizu (approved):

ivoanjo (approved):
:+1: LGTM -- I've retriggered CI to get over the flaky failures

docs/GettingStarted.md
```diff
@@ -306,6 +306,23 @@ You can also get the current active trace using the `active_trace` method. This
...
+```ruby
+event = Datadog::Tracing::SpanEvent.new(
+ "custom.event.name",
+ attributes: { "key1" => "value1", "key2" => 123 }
+)
+
+span.span_events << event
+```
+
+You can add multiple events to a span. Each event **must** include a name, an optional set of attributes, and an optional timestamp (in nanoseconds). If no timestamp is provided, the current time is used.
```
p-datadog, line 322: Would it be useful for customers to include a timestamp in the example?

soulcutter: The behavior of that object is [likely to change](https://github.com/DataDog/dd-trace-rb/blob/6aa7549f8c9632dea7d9e7e667b07721af6425c9/lib/datadog/tracing/span_event.rb#L22), and top-level documentation like this is likely to accidentally diverge. Something like this might be more evergreen:

> Each event must be a well-formed instance of [`::Datadog::Tracing::SpanEvent`](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/tracing/span_event.rb).

No matter what, this documentation is an improvement over it being entirely missing. 🍻

marcotc: I think the timestamp usage is niche, and given how easy it is to copy/paste this snippet, I wouldn't like unnecessary timestamp parameters being created in the wild when not required, just to cause headaches.

PR #4823, opened 2025-07-23
[NO-TICKET] Enable standard-rb on lib folder

vpellan:
**What does this PR do?**

This PR enables standard-rb on the lib folder, and fixes all the offenses it has detected

**Motivation:**

Config inversion will use standard-rb by adding a custom cop. We do not want to miss any file in them, and this is a step toward migrating to standard-rb.

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

CI.


lib/datadog/kit/identity.rb
```diff
@@ -102,6 +99,9 @@ def set_trace_and_span_context(method, trace = nil, span = nil)
 
           yield(trace, span)
         end
+        # rubocop:enable Metrics/AbcSize
+        # rubocop:enable Metrics/PerceivedComplexity
+        # rubocop:enable Metrics/CyclomaticComplexity
```
y9v, line 110: I think this change was not indended?

y9v: usually it's a bad practice to disable some rules for more than one method

lib/datadog/opentelemetry/api/context.rb
```diff
@@ -50,7 +50,7 @@ def value(key)
           @trace.otel_value(key)
         end
 
-        alias [] value
+        alias_method :[], :value
```
y9v: `alias` and `alias_method` have slightly different behaviour - are we ok with this change?

lib/datadog/opentelemetry/sdk/trace/span.rb
```diff
@@ -39,7 +39,7 @@ def record_exception(exception, attributes: nil)
         end
 
         # `alias` performed to match {OpenTelemetry::SDK::Trace::Span} aliasing upstream
-        alias []= set_attribute
+        alias_method :[]=, :set_attribute
```
y9v: same here - and the comment is not updated

lib/datadog/opentelemetry/sdk/trace/span.rb
```diff
@@ -161,15 +160,17 @@ def override_datadog_values(span, key, value)
...
-              value.casecmp('true') == 0 ? 1 : 0
+              (value.casecmp('true') == 0) ? 1 : 0
             )
           end
 
           span.set_tag('http.status_code', value) if key == 'http.response.status_code'
         end
 
+        # rubocop:enable Metrics/CyclomaticComplexity,Metrics/PerceivedComplexity
+
```
y9v: again - I don't think this change is intended

lib/datadog/tracing/contrib/faraday/middleware.rb
```diff
@@ -96,11 +95,13 @@ def handle_response(span, env, options)
             span.set_tags(
               Datadog.configuration.tracing.header_tags.response_tags(env[:response_headers])
             )
-          rescue StandardError => e
+          rescue => e
             Datadog.logger.error(e.message)
             Datadog::Core::Telemetry::Logger.report(e)
           end
 
+          # rubocop:enable Metrics/AbcSize
```
y9v, line 104: this needs to be reverted

lib/datadog/tracing/contrib/propagation/sql_comment/comment.rb
```diff
@@ -15,7 +15,7 @@ def initialize(hash)
 
             def to_s
               @string ||= begin
-                ret = String.new
+                ret = +''
```
y9v, line 18: I don't agree with this change = `String.new` is better readable than cryptic `+''`

lib/datadog/tracing/contrib/rack/middlewares.rb
```diff
@@ -269,19 +265,24 @@ def parse_url(env, original_env)
...
+              request_uri.delete_prefix(base_url)
+            end
 
             base_url + fullpath
           end
 
+          # rubocop:enable Metrics/AbcSize
+          # rubocop:enable Metrics/CyclomaticComplexity
+          # rubocop:enable Metrics/PerceivedComplexity
+          # rubocop:enable Metrics/MethodLength
```
y9v: again - this change has to be reverted

lib/datadog/tracing/distributed/datadog.rb
```diff
@@ -65,7 +65,7 @@ def extract(data)
           # DEV: To be valid we need to have a trace id and a parent id
           #      or when it is a synthetics trace, just the trace id.
           # DEV: `Fetcher#id` will not return 0
-          return unless (trace_id && parent_id) || (origin && trace_id)
+          return if trace_id.nil? || parent_id.nil? && origin.nil?
```
y9v, line 67: not sure this is equivalent

lib/datadog/tracing/distributed/trace_context.rb
```diff
@@ -148,7 +148,7 @@ def build_trace_flags(digest)
 
         # @see https://www.w3.org/TR/trace-context/#tracestate-header
         def build_tracestate(digest)
-          tracestate = String.new('dd=')
+          tracestate = +'dd='
```
y9v, line 151: again - I wouldn't do that

lib/datadog/tracing/distributed/trace_context.rb
```diff
@@ -350,7 +354,7 @@ def extract_datadog_fields(dd_tracestate)
               tags ||= {}
               tags["#{Tracing::Metadata::Ext::Distributed::TAGS_PREFIX}#{key}"] = value
             else
-              unknown_fields ||= String.new
+              unknown_fields ||= +''
```
y9v, line 359: same here - can we disable this rule?

lib/datadog/tracing/metadata/tagging.rb
```diff
@@ -80,10 +80,10 @@ def clear_tag(key)
         end
 
         # Convenient interface for setting a single tag.
-        alias []= set_tag
+        alias_method :[]=, :set_tag
 
         # Convenient interface for getting a single tag.
-        alias [] get_tag
+        alias_method :[], :get_tag
```
y9v, line 86: again - `alias` and `alias_method` are not equal

lib/datadog/tracing/span.rb
```diff
@@ -113,7 +113,7 @@ def started?
       def stopped?
         !@end_time.nil?
       end
-      alias :finished? :stopped?
+      alias_method :finished?, :stopped?
```
y9v, line 116: same here

lib/datadog/tracing/workers.rb
```diff
@@ -107,7 +107,7 @@ def enqueue_trace(trace)
 
         private
 
-        alias flush_data callback_traces
+        alias_method :flush_data, :callback_traces
```
y9v, line 110: again `alias` changed to `alias_method`

lib/datadog/opentelemetry/api/context.rb
p-datadog: My impression is that the behavior of `alias` (to the extent that it differs from `alias_method`) is actually not wanted, do you disagree?

lib/datadog/tracing/distributed/datadog.rb
p-datadog: Seems fine to me? trace_id was checked twice.

p-datadog: Though  I do not appreciate formatting tools altering logic. I would disable whatever rule rubocop or standard would be complaining about here if I wanted to preserve the logic as written.

tasks/lint.rake
```diff
@@ -0,0 +1,39 @@
...
+
+  # standard-rb does not enable the Style/FrozenStringLiteralComment cop.
+  # As we will still support Rubies < 3.4 for years, we need to check that all .rb files in lib folder start with frozen_string_literal: true.
+  desc 'Check that all .rb files in lib folder start with frozen_string_literal: true'
+  task :frozen_string_literal do
+    files_without_magic_comment = []
+
+    Find.find('lib') do |path|
+      # Skip vendor folders
+      Find.prune if path.include?('/vendor')
```
p-datadog:
```suggestion
      Find.prune if File.basename(path) == 'vendor'
```

y9v (approved):

lib/datadog/opentelemetry/api/context.rb
vpellan: After discussing it with the ruby guild, we are ok with this change (and unless we disable it inline, we cannot use alias with standard-rb in this case)

lib/datadog/kit/identity.rb
vpellan: Reverted in [d19bbda](https://github.com/DataDog/dd-trace-rb/pull/4823/commits/d19bbda1c8e78ce4812850c447524ec9e1791fd9)

lib/datadog/tracing/contrib/propagation/sql_comment/comment.rb
vpellan: After discussing in the guild about that we agreed to stick to standard-rb rules

lib/datadog/tracing/distributed/datadog.rb
vpellan: The truth table for `!((A && B) || (A && C))` and `!A || !B && !C` is the same. The rule says to not use mixed logical operators in an `unless`. The comment is not updated (we now accept any origin, https://github.com/DataDog/dd-trace-rb/commit/d2a183f709cf975564834d20651ac99625c99d15) so I will update it, and this should make the code clearer (we return if no trace_id, or if trace_id but no parent_id and no origin)

lib/datadog/opentelemetry/sdk/trace/span.rb
vpellan: The OTel spec passes so the behaviour we expect is not changed in that case

vpellan: In this case, as we want to match the upstream definition and behaviour, we should use alias. I will change it back to alias.

lib/datadog/opentelemetry/api/context.rb
vpellan: In this case, as we want to match the upstream definition and behaviour, we should use alias. I will change it back to alias and add a comment.

TonyCTHsu (approved):

lib/datadog/opentelemetry/api/context.rb
vpellan: Reverted in [1e2dec4](https://github.com/DataDog/dd-trace-rb/pull/4823/commits/1e2dec457ecbbeb10f6cf869d678c088e8c8db5c)

lib/datadog/opentelemetry/sdk/trace/span.rb
vpellan: Reverted in [1e2dec4](https://github.com/DataDog/dd-trace-rb/pull/4823/commits/1e2dec457ecbbeb10f6cf869d678c088e8c8db5c)

lib/datadog/tracing/distributed/datadog.rb
vpellan: Comment updated in [1e2dec4](https://github.com/DataDog/dd-trace-rb/pull/4823/commits/1e2dec457ecbbeb10f6cf869d678c088e8c8db5c)

ivoanjo:
This is great, thank you! :bow:

PR #4826, opened 2025-07-24
[NO-TICKET] Fix stable_config test

vpellan:
**What does this PR do?**

This PR adds stable config to the Matrixfile, and fix a test that was checking Datadog.logger instead of config_init_logger.

**Motivation:**

I forgot to add it stable_config to the Matrixfile and when I did, I saw a test that was flaky because it checked the wrong method call.

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

CI.


ivoanjo (approved):
:+1: LGTM

p-datadog (approved):

PR #4831, opened 2025-07-28
[NO-TICKET] Minor: Add more details to libdatadog build error message

ivoanjo:
**What does this PR do?**

This PR adds more details to the warning message that's printed when something is wrong with libdatadog.

Here's two examples of how they'll show up with this change:

> WARN: Skipping build of libdatadog_api (issue setting up `libdatadog` gem: There was an error loading libdatadog: RuntimeError hello!). Some functionality will not be available.

or

> WARN: Skipping build of libdatadog_api (issue setting up `libdatadog` gem: The `libdatadog` gem installed on your system is missing binaries for your platform variant. Your platform: `x86_64-linux`; available binaries: `lib`, `LICENSE`, `NOTICE`, `include`, `LICENSE-3rdparty.yml`, `bin`, `cmake`). Some functionality will not be available.

**Motivation:**

In practice, Ruby hides these messages by default so I find these are more useful for new team members: I've seen two different people running into them when running `bundle exec rake clean compile` recently.

**Change log entry**

Yes. Add more details to libdatadog build error message

**Additional Notes:**

N/A

**How to test the change?**

This change includes test coverage. You can also trigger the issue manualy, by, e.g. misconfiguring `LIBDATADOG_VENDOR_OVERRIDE` env variable to point at a wrong or non-existing folder.

### DIFF
--- a/ext/libdatadog_api/extconf.rb
+++ b/ext/libdatadog_api/extconf.rb
@@ -26,7 +26,9 @@ if ENV['DD_NO_EXTENSION'].to_s.strip.downcase == 'true'
 end
 skip_building_extension!('current Ruby VM is not supported') if RUBY_ENGINE != 'ruby'
 skip_building_extension!('Microsoft Windows is not supported') if Gem.win_platform?
-skip_building_extension!('issue setting up `libdatadog` gem') if Datadog::LibdatadogExtconfHelpers.libdatadog_issue?
+
+libdatadog_issue = Datadog::LibdatadogExtconfHelpers.libdatadog_issue?
+skip_building_extension!("issue setting up `libdatadog` gem: #{libdatadog_issue}") if libdatadog_issue
 
 require 'mkmf'
 
--- a/ext/libdatadog_extconf_helpers.rb
+++ b/ext/libdatadog_extconf_helpers.rb
@@ -5,6 +5,8 @@ require 'pathname'
 
 module Datadog
   # Contains a bunch of shared helpers that get used during building of extensions that link to libdatadog
+  #
+  # Note: Specs for this file currently live in `spec/datadog/profiling/native_extension_helpers_spec.rb`.
   module LibdatadogExtconfHelpers
     # Used to make sure the correct gem version gets loaded, as extconf.rb does not get run with "bundle exec" and thus
     # may see multiple libdatadog versions. See https://github.com/DataDog/dd-trace-rb/pull/2531 for the horror story.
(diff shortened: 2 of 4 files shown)


ext/libdatadog_api/extconf.rb
```diff
@@ -26,7 +26,9 @@ def skip_building_extension!(reason)
 end
 skip_building_extension!('current Ruby VM is not supported') if RUBY_ENGINE != 'ruby'
 skip_building_extension!('Microsoft Windows is not supported') if Gem.win_platform?
-skip_building_extension!('issue setting up `libdatadog` gem') if Datadog::LibdatadogExtconfHelpers.libdatadog_issue?
+
+libdatadog_issue = Datadog::LibdatadogExtconfHelpers.libdatadog_issue?
```
p-datadog: The question mark implies a boolean return value, I would add a `libdatadog_issue` method with the behavior proposed in this PR and if necessary for backward compatibility make `libdatadog_issue?` double-negate that method to obtain a boolean. If BC is not needed the `libdatadog_issue?` can be renamed to `libdatadog_issue`.

ivoanjo: No compatibility needed for this. I'll update the PR with this change.

ivoanjo: Fixed in c18576be4d71b4d9d4d81c0a0f318b134ca62939 :D

ext/libdatadog_extconf_helpers.rb
```diff
@@ -122,9 +124,16 @@ def self.try_loading_libdatadog
...
 
-    def self.libdatadog_issue?
-      try_loading_libdatadog { |_exception| return true }
-      Libdatadog.pkgconfig_folder.nil?
+    def self.libdatadog_issue
+      try_loading_libdatadog do |exception|
+        return "There was an error loading `libdatadog`: #{exception.class} #{exception.message}"
+      end
+
+      unless Libdatadog.pkgconfig_folder
```
y9v, line 133: not a blocker:

Just by looking at the function name I would never guess that it tries to load `libdatadog`. Does it make sense to remove this function and have more verbose code in `extconf.rb`?
```ruby
Datadog::LibdatadogExtconfHelpers.try_loading_libdatadog do |exception|
  skip_building_extension!("issue setting up `libdatadog` gem: There was an error loading `libdatadog`: #{exception.class} #{exception.message}")
end

unless Libdatadog.pkgconfig_folder
  skip_building_extension!(
    "The `libdatadog` gem installed on your system is missing binaries for your platform variant. " \
      "Your platform: " \
      "`#{Libdatadog.current_platform}`; available binaries: `#{Libdatadog.available_binaries.join("`, `")}`")
end
```

p-datadog: I can agree with this, there is only one consumer of `libdatadog_issue` that I see making me wonder why the helper exists at all.

ivoanjo: Reasonable feedback folks! I've pushed 99cc522af358a76eaa73b52533f0f95fbf96a9f2. with a rename of the method to `load_libdatadog_or_get_issue` and a comment to explain why it exists here.

The reason it exists here is that the `extconf.rb` is extremely awkward to test given how it depends on Ruby mkmf's helper.

So in general it's best to avoid having any too-clever logic on the `extconf.rb` as it won't be able to be properly tested.

ivoanjo: Marking as resolved so I can go ahead and merge the PR, but as usual, do let me know if you have further feedback and I'm very happy to open a follow-up PR addressing it.

PR #4833, opened 2025-07-30
Change the mutex around WAF reconfiguration

y9v:
**What does this PR do?**
This PR changes the way we use mutexes in AppSec.

It removes the mutex around `AppSec::Component#shutdown!`, and makes `AppSec::Component#shutdown!` a no-op.

A mutex around WAF Handle change is also simplified - now we only have a mutex for WAF Handle reference counting, and the WAF Handle and Context creation is moved outside of the mutex.

**Motivation:**
We don't want webserver threads to wait for calls to libddwaf.

**Change log entry**

**Additional Notes:**
None.

**How to test the change?**
CI and manual testing.


lib/datadog/appsec/security_engine/runner.rb
```diff
@@ -9,9 +9,10 @@ module SecurityEngine
       class Runner
         SUCCESSFUL_EXECUTION_CODES = [:ok, :match].freeze
 
-        def initialize(waf_context)
+        def initialize(waf_handle)
           @mutex = Mutex.new
-          @waf_context = waf_context
+          @waf_handle = waf_handle
+          @waf_context = @waf_handle.build_context
```
Strech: If we know that util runner is done waf handle will exist, we can defer that call to a method with memoization.

lib/datadog/appsec/security_engine/runner.rb
```diff
@@ -56,6 +57,7 @@ def run(persistent_data, ephemeral_data, timeout = WAF::LibDDWAF::DDWAF_RUN_TIME
 
         def finalize!
           @waf_context.finalize!
+          AppSec.security_engine.release_waf_handle(@waf_handle)
```
Strech: This will violate separation of concerns. What if waf_handle will be not handle but registry itself?

With that in mind we can acquire and decommission handles?

y9v: addressed in [761c125](https://github.com/DataDog/dd-trace-rb/pull/4833/commits/761c125cf9cb514600144decd744a24e69e265c0)

lib/datadog/appsec/security_engine/runner.rb
y9v: done in [761c125](https://github.com/DataDog/dd-trace-rb/pull/4833/commits/761c125cf9cb514600144decd744a24e69e265c0)

lib/datadog/appsec/security_engine/runner.rb
```diff
@@ -54,14 +55,19 @@ def run(persistent_data, ephemeral_data, timeout = WAF::LibDDWAF::DDWAF_RUN_TIME
...
         end
 
+        def waf_context
+          @waf_context ||= @waf_handle.build_context
+        end
+
         def finalize!
-          @waf_context.finalize!
+          @waf_context&.finalize!
+          @handle_ref.release(@waf_handle)
```
Strech, line 65: Just to be sure, do we want to put it into `ensure` to prevent leaking on errors here? (and I guess we need a test for it)

y9v: yes, that's a good idea - will put it into `ensure` and test it

y9v: fixed here: [d9b3e5e](https://github.com/DataDog/dd-trace-rb/pull/4833/commits/d9b3e5eba74dc126d6b163f78ba913ae20c1e9a5)

The test is ugly though - I don't see any other way to test this other than to test that the `ThreadSafeRef` instance receives a call to release the handle

spec/datadog/appsec/security_engine/runner_spec.rb
```diff
@@ -178,4 +178,16 @@
...
+
+  describe '#finalize!' do
+    it 'releases the waf handle when an error occurs' do
+      # waf_context is lazily initialized, so we need to ensure it is called
+      runner.waf_context
+
+      expect(waf_context).to receive(:finalize!).and_raise(StandardError)
+      expect(thread_safe_ref).to receive(:release).with(waf_handle)
+
+      runner.finalize! rescue StandardError
```
Strech: you can still test is as

```suggestion
expect { runner.finalize! }.to raise(StandardError)
```

Strech (approved):

spec/datadog/appsec/security_engine/runner_spec.rb
```diff
@@ -171,4 +178,19 @@
       end
     end
   end
+
+  describe '#finalize!' do
+    it 'releases the waf handle when an error occurs' do
+      # waf_context is lazily initialized, so we need to ensure it is called
+      runner.waf_context
+
+      expect(waf_context).to receive(:finalize!).and_raise(StandardError)
```
Strech: This one is better to convert into `allow`

spec/datadog/appsec/security_engine/runner_spec.rb
```diff
@@ -171,4 +178,19 @@
...
+      expect(thread_safe_ref).to receive(:release).with(waf_handle)
+
+      begin
+        runner.finalize!
+      rescue
+      end
```
Strech: This one is better to be normal expectation of exception

y9v: agree, fixed [4475813](https://github.com/DataDog/dd-trace-rb/pull/4833/commits/44758138273d01f5d743d74b2e071c15d197b9c8)

PR #4836, opened 2025-08-05
[NO-TICKET] Tweak benchmark validation to avoid flakiness

ivoanjo:
**What does this PR do?**

This PR tweaks the profiling benchmark validation to avoid flakiness in CI such as https://github.com/DataDog/dd-trace-rb/actions/runs/16683065404/job/47226121608#step:5:16961 :

```
Failures:

  1) Profiling benchmarks profiling_sample_loop_v2 runs without raising errors
     Failure/Error: raise("Wait time exhausted! Requested: #{spec}, waited: #{actual}")

     RuntimeError:
       Wait time exhausted! Requested: 10 seconds, waited: 10.01 seconds, 100 attempts with backoff 0.1
```

It does this by:
* Raising the timeout from 10 to 15 seconds
* Tweaking the workload on the benchmark in validation mode to finish faster

**Motivation:**

Avoid flakiness in the benchmark validation code.

**Change log entry**

None.

**Additional Notes:**

I've also included an improvement to the error generated so that it includes the process stdout/stderr and we don't have to go around hunting for it.

**How to test the change?**

Green CI is good!


ivoanjo:
I'll be out for a few days so if this looks good I suggest merging it without waiting for me to avoid any more CI flakiness around this test.

spec/support/synchronization_helpers.rb
```diff
@@ -35,14 +35,11 @@ def expect_in_fork(fork_expectations: nil, timeout_seconds: 10)
...
-    rescue => e
+    rescue
       stdout ||= File.read(fork_stdout.path)
       stderr ||= File.read(fork_stderr.path)
 
-      puts stdout
-      warn stderr
-
-      raise e
+      raise "Failure or timeout during `expect_in_fork`, STDOUT: `#{stdout}`, STDERR: `#{stderr}`"
```
p-datadog: I would keep the original exception here for diagnostics/troubleshooting?

marcotc: Done!

marcotc (approved):

PR #4837, opened 2025-08-05
Only retrieve code from telemetry response objects that have it

p-datadog:
**What does this PR do?**

Adds a `respond_to?` check to telemetry logging code to only retrieve `code` from response objects that respond to this method (the InternalErrorResponse does not)

Also, this PR fixes the positive logging to only happen for successful responses (and not for responses that were obtained from network but not with 2xx code).

**Motivation:**
Warnings being emitted in telemetry code

**Change log entry**
None

**Additional Notes:**

**How to test the change?**

Unit tests added


lib/datadog/core/telemetry/emitter.rb
```diff
@@ -31,7 +31,17 @@ def request(event)
...
-          logger.debug { "Telemetry sent for event `#{event.type}` (response code: #{res.code})" }
+          if res.ok?
+            logger.debug { "Telemetry sent for event `#{event.type}`" }
+          else
+            diag = if res.respond_to?(:code)
+              "#{res}: code #{res.code}"
+            else
+              # InternalErrorResponse does not have a code
+              res.to_s
+            end
```
Strech: WDYT to unify that via `res.to_s` for both cases, as we still want to know what's going on.

spec/datadog/core/telemetry/emitter_spec.rb
```diff
@@ -8,7 +8,11 @@
   subject(:emitter) { described_class.new(transport, logger: logger) }
   let(:logger) { logger_allowing_debug }
   let(:transport) { double(Datadog::Core::Telemetry::Transport::HTTP::Client) }
-  let(:response) { double(Datadog::Core::Transport::HTTP::Adapters::Net::Response) }
+  let(:response) do
+    double(Datadog::Core::Transport::HTTP::Adapters::Net::Response).tap do |response|
+      allow(response).to receive(:ok?).and_return(response_ok)
+    end
```
Strech: I think it's better to split it into `instance_double` first and then put the allow into before block right below.

It's considered bad practice to use `double` and `tap` to stub method, it also could be achieved via

```ruby
instance_double(Datadog::Core::Transport::HTTP::Adapters::Net::Response, ok?: response_ok)
```

p-datadog: I forget about that pattern but it's the one I want in this case, thanks.

lib/datadog/core/telemetry/emitter.rb
p-datadog: I moved the code to provide code in the inspect method.

Strech (approved):
👍🏼

lib/datadog/core/transport/response.rb
```diff
@@ -34,7 +34,10 @@ def internal_error?
         end
 
         def inspect
-          "#{self.class} ok?:#{ok?} unsupported?:#{unsupported?}, " \
+          maybe_code = if respond_to?(:code)
+            " code:#{code}," # steep:ignore
+          end
+          "#{self.class} ok?:#{ok?},#{maybe_code} unsupported?:#{unsupported?}, " \
```
vpellan, line 40: Would it be possible to move that part to classes that actually defines `code` ? Right now I can see two, `Core::Transport::HTTP::Adapters::Net::Response` and `Core::Transport::HTTP::Adapters::Test::Response` which already adds `code` to `inspect` (so in this case we'd have it twice). Or maybe add a `code` method to that module that also returns nil ?

p-datadog: @vpellan I thought I started working on this but I can't find the code now. Given that this PR fixes a bug I will merge it and I made a note to myself to follow up on your comment and further improve inspection here.

PR #4839, opened 2025-08-06
Add appsec.waf.requests telemetry

y9v:
**What does this PR do?**
This PR adds reporting of `appsec.waf.requests` telemetry metric with following tags:
 - waf_version
 - event_rules_version
 - rule_triggered
 - request_blocked
 - waf_error
 - waf_timeout
 - rate_limited

**Motivation:**
We want to get telemetry metrics for WAF.

**Change log entry**
None. This is internal change.

**Additional Notes:**
The `input_truncated` metric will be added in a separate PR, since it requires changes to `libddwaf-rb`.

**How to test the change?**
CI and manual testing.

### DIFF
--- a/lib/datadog/appsec/security_engine/engine.rb
+++ b/lib/datadog/appsec/security_engine/engine.rb
@@ -60,7 +60,7 @@ module Datadog
           remove_config_at_path(DEFAULT_RULES_CONFIG_PATH) if @is_ruleset_update
 
           diagnostics = @waf_builder.add_or_update_config(config, path: path)
-          @builder_ruleset_version = diagnostics['ruleset_version'] if diagnostics.key?('ruleset_version')
+          @reconfigured_ruleset_version = diagnostics['ruleset_version'] if diagnostics.key?('ruleset_version')
           report_configuration_diagnostics(diagnostics, action: 'update', telemetry: AppSec.telemetry)
 
           # we need to load default config if diagnostics contains top-level error for rules or processors
@@ -69,7 +69,7 @@ module Datadog
               diagnostics.dig('rules', 'error') ||
               diagnostics.dig('processors', 'errors'))
             diagnostics = load_default_config(telemetry: AppSec.telemetry)
-            @builder_ruleset_version = diagnostics['ruleset_version']
+            @reconfigured_ruleset_version = diagnostics['ruleset_version']
             report_configuration_diagnostics(diagnostics, action: 'update', telemetry: AppSec.telemetry)
           end
 
@@ -86,7 +86,7 @@ module Datadog
 
           if result && path != DEFAULT_RULES_CONFIG_PATH && path.include?('ASM_DD')
             diagnostics = load_default_config(telemetry: AppSec.telemetry)
-            @builder_ruleset_version = diagnostics['ruleset_version']
+            @reconfigured_ruleset_version = diagnostics['ruleset_version']
             report_configuration_diagnostics(diagnostics, action: 'update', telemetry: AppSec.telemetry)
           end
 
@@ -100,7 +100,7 @@ module Datadog
 
         def reconfigure!
           new_waf_handle = @waf_builder.build_handle
-          @ruleset_version = @builder_ruleset_version
+          @ruleset_version = @reconfigured_ruleset_version
 
           @handle_ref.current = new_waf_handle
         rescue WAF::Error => e
--- a/sig/datadog/appsec/security_engine/engine.rbs
+++ b/sig/datadog/appsec/security_engine/engine.rbs
@@ -8,6 +8,8 @@ module Datadog
 
         @ruleset_version: ::String
 
+        @reconfigured_ruleset_version: ::String?
+
         DEFAULT_RULES_CONFIG_PATH: ::String
         TELEMETRY_ACTIONS: ::Array[::String]
         DIAGNOSTICS_CONFIG_KEYS: ::Array[::String]


lib/datadog/appsec/context.rb
```diff
@@ -36,7 +37,9 @@ def initialize(trace, span, waf_runner)
         @span = span
         @events = []
         @waf_runner = waf_runner
+
         @metrics = Metrics::Collector.new
+        @request_milestones = RequestMilestones.new
```
Strech: I don't think that "request milestones" reflect the purpose, it's same metrics. Based on the description

> Data about the milestones that happened during request processing

Judging by the description of the class it's request processing stats. The issue with current name - it creates an expression of something more when it's actually just stats.

lib/datadog/appsec/context.rb
```diff
@@ -10,7 +11,7 @@ class Context
       ActiveContextError = Class.new(StandardError)
 
       # TODO: add delegators for active trace span
-      attr_reader :trace, :span, :events
+      attr_reader :trace, :span, :events, :request_milestones
```
Strech: I think we should not expose it to outside directly, but rather expose API to operate with it. Same as we don't expose WAF runner, but expose API to use it. It will allow to extend it if needed without changing the caller side.

Or we should give it a better name to emphasis on what it is actually is.

spec/datadog/appsec/metrics/telemetry_spec.rb
```diff
@@ -93,4 +93,139 @@
...
+        expect(telemetry).to receive(:inc).with(
+          'specsec', 'waf.requests', 1, tags: {
+            waf_version: '1.42.99',
+            rule_triggered: 'false',
+            request_blocked: 'false',
+            rate_limited: 'false',
+            waf_error: 'true',
+            waf_timeout: 'false'
+          }
+        )
```
Strech: I think to make emphasis on what we actually testing here it makes sense to decompose expectation and limit it to the actual check of error tag only.

Yes we probably want to have a single test to check for all tags, but all the similarly looking tests of specific tags, should just check those tags only, like

```ruby
expect(telemetry).to receive(:inc).with(
  'specsec', 'waf.requests', 1, tags: hash_including(waf_error: 'true')
)
```

sig/datadog/appsec/request_milestones.rbs
```diff
@@ -0,0 +1,10 @@
+module Datadog
+  module AppSec
+    class RequestMilestones
+      attr_accessor request_blocked: bool
```
Strech: I think those name would change too because now we are already in the context of the request, and yet those metrics should be prefixed. If they will be guarded by `Context` class there will be no need to prefix them as the class name implies it.

lib/datadog/appsec/metrics/telemetry.rb
```diff
@@ -17,6 +17,19 @@ def report_rasp(type, result)
...
+            waf_version: Datadog::AppSec::WAF::VERSION::BASE_STRING,
+            rule_triggered: (!waf_metrics.matches.zero?).to_s,
+            request_blocked: request_milestones.request_blocked.to_s,
+            rate_limited: request_milestones.request_rate_limited.to_s,
+            waf_error: (!waf_metrics.errors.zero?).to_s,
+            waf_timeout: (!waf_metrics.timeouts.zero?).to_s
+          }
+
+          AppSec.telemetry.inc(Ext::TELEMETRY_METRICS_NAMESPACE, 'waf.requests', 1, tags: tags)
+        end
```
Strech: I think this is a wrong move. An existing telemetry is sent on every call, where this method is more exporter of metrics via telemetry that is suppose to be called once. And IMHO it should be reflected via module name and the method name.

I don't think that this placement is correct. Technically we have here RaspTelemetry which is about rasp. In existing Context you can see that I have `record_rasp` and then `export_rasp_metrics`. Same could be done here - recording and exporting. And yes, exporting is done via Telemetry, not span Metrics functionality

Strech (commented):
Draft looks promising, but the composition of the responsibilities should be changed. 

I think if we are about to expose stats it should be done clear and straightforward. The feel of the existing placement of the methods is a context of use, but not the context of responsibilities.

lib/datadog/appsec/context.rb
y9v: I think you are right, "milestones" is not a good name here. But I think that "metrics" or "stats" also wouldn't work - these are more flags than metrics.

y9v:
@Strech I changed the approach - I think it should be more straightforward now

lib/datadog/appsec/metrics/telemetry_exporter.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+module Datadog
+  module AppSec
+    module Metrics
+      # A class responsible for exporting WAF request metrics via Telemetry.
+      module TelemetryExporter
+        module_function
+
+        def export_waf_request_metrics(waf_metrics, context)
```
Strech: I think a method prefix is enough to know that metrics is about the WAF

```suggestion
        def export_waf_request_metrics(metrics, context)
```

lib/datadog/appsec/context.rb
```diff
@@ -55,6 +55,22 @@ def run_rasp(type, persistent_data, ephemeral_data, timeout = WAF::LibDDWAF::DDW
         result
       end
 
+      def mark_as_interrupted!
+        @interrupted = true
+      end
+
+      def interrupted?
+        @interrupted == true
```
Strech: I think it's better to have default value as `false` in the initializer instead. Undefined instance variables make it harder to follow the code logic.

lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -129,7 +133,7 @@ def load_default_config(telemetry:)
           config['exclusions'] ||= AppSec::Processor::RuleLoader.load_exclusions(ip_passlist: @default_ip_passlist)
 
           diagnostics = @waf_builder.add_or_update_config(config, path: DEFAULT_RULES_CONFIG_PATH)
-          @ruleset_version = diagnostics['ruleset_version']
+          @builder_ruleset_version = diagnostics['ruleset_version']
```
Strech: This is very unobvious move. In the code above there is no clue how the `@builder_ruleset_version` was set, even tho we return result from which we can retrieve that data and now it become significant to spot as the issue.

lib/datadog/appsec/security_engine/runner.rb
```diff
@@ -9,10 +9,14 @@ module SecurityEngine
       class Runner
         SUCCESSFUL_EXECUTION_CODES = [:ok, :match].freeze
 
-        def initialize(handle_ref)
+        attr_reader :ruleset_version, :waf_addresses
+
+        def initialize(handle_ref:, ruleset_version:, waf_addresses:)
```
Strech: If addresses could be retrieved via `waf_handle`, why don't we do it like in engine?

```
@waf_addresses = waf_handle.known_addresses
```

The problem of converting normal arguments to the kwagrs is that we will loose the sight of what is the main argument and what is an additional (potentially optional).

As we could extract addresses from handle, we just need to know the version of the ruleset used and it's an additional argument. So in this case we can reduce interface to


```suggestion
        def initialize(handle_ref, ruleset_version:)
```

lib/datadog/appsec/context.rb
```diff
@@ -55,6 +55,22 @@ def run_rasp(type, persistent_data, ephemeral_data, timeout = WAF::LibDDWAF::DDW
...
+      end
+
+      def ruleset_version
```
Strech: If I will read that sentence `"Context ruleset version"` I would ask myself - what kind of ruleset version that context is talking about? I think we should prefix encapsulated methods we expose to make it clear what we are talking about.

Same goes for waf addresses, I do think we might give it a better name, like `waf runner supported addresses` or so. yes it's long, but it's clear

sig/datadog/appsec/security_engine/runner.rbs
```diff
@@ -12,11 +12,19 @@ module Datadog
 
         @waf_context: WAF::Context
 
+        @ruleset_version: ::String
+
+        @waf_addresses: ::Array[::String]
```
Strech: I think we should define it in a better place (responsible for addresses) and reference it here. Otherwise the origin might change, but type check will not fail

spec/datadog/appsec/integration/waf_requests_telemetry_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      c.tracing.instrument :http
+
+      c.appsec.enabled = true
+      c.appsec.instrument :rack
+
+      c.appsec.waf_timeout = 10_000_000 # in us
+      c.appsec.ip_passlist = []
+      c.appsec.ip_denylist = []
+      c.appsec.user_id_denylist = []
+      c.appsec.ruleset = :recommended
```
Strech: I think the issue with that - unpredictability. The default changes and test start failing, when it should not.

spec/datadog/appsec/integration/waf_requests_telemetry_spec.rb
```diff
@@ -0,0 +1,90 @@
...
+      map '/waf' do
+        run ->(_env) { [200, {'Content-Type' => 'text/html'}, ['OK']] }
+      end
+    end
+
+    stack.to_app
+  end
+
+  describe 'appsec.waf.requests telemetry' do
+    subject(:response) { last_response }
```
Strech: Is it used anywhere?

Strech (changes requested):
Nice job!

I would like to sort out the `Security Engine` / `Runner` / `Context` responsibilities and interfaces. It's clear that they toss around necessary data, but it's not a solid interface nor clear responsibilities.

lib/datadog/appsec/metrics/telemetry_exporter.rb
y9v: fixed in [f4de7ed](https://github.com/DataDog/dd-trace-rb/pull/4839/commits/f4de7edb068f62f5413a2678b9197d6063015c30)

lib/datadog/appsec/context.rb
y9v: fixed in [f4de7ed](https://github.com/DataDog/dd-trace-rb/pull/4839/commits/f4de7edb068f62f5413a2678b9197d6063015c30)

lib/datadog/appsec/security_engine/runner.rb
y9v: fixed in [f4de7ed](https://github.com/DataDog/dd-trace-rb/pull/4839/commits/f4de7edb068f62f5413a2678b9197d6063015c30)

sig/datadog/appsec/security_engine/runner.rbs
y9v: fixed in [f4de7ed](https://github.com/DataDog/dd-trace-rb/pull/4839/commits/f4de7edb068f62f5413a2678b9197d6063015c30)

lib/datadog/appsec/security_engine/engine.rb
y9v: I improved it a little in [43d830a](https://github.com/DataDog/dd-trace-rb/pull/4839/commits/43d830a36bea160e5d7e58eb770f0bdd5511dcfd). But we still have to keep the ruleset version that is in the builder, but not in the handle (since it wasn't reconfigured yet).

One option would be to do it in `WAF::HandleBuilder` in `libddwaf-rb`, what do you think?

spec/datadog/appsec/integration/waf_requests_telemetry_spec.rb
y9v: fixed [e7a12fe](https://github.com/DataDog/dd-trace-rb/pull/4839/commits/e7a12fe04cab674ff7ed61d2a6c9d2fe57d81bb1)

spec/datadog/appsec/integration/waf_requests_telemetry_spec.rb
y9v: fixed [e7a12fe](https://github.com/DataDog/dd-trace-rb/pull/4839/commits/e7a12fe04cab674ff7ed61d2a6c9d2fe57d81bb1)

lib/datadog/appsec/context.rb
y9v: renamed to `waf_runner_ruleset_version` and `waf_runner_known_addresses` (I like "supported" addresses too, and it makes sense, but we already have "known" addresses in libddwaf, so I think we rather use the same term, or rename it everywhere)

[10379c3](https://github.com/DataDog/dd-trace-rb/pull/4839/commits/10379c302c1e4f83313489616884e3ad8d4f269c)

lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -64,7 +60,7 @@ def add_or_update_config(config, path:)
           remove_config_at_path(DEFAULT_RULES_CONFIG_PATH) if @is_ruleset_update
 
           diagnostics = @waf_builder.add_or_update_config(config, path: path)
-          @ruleset_version = diagnostics['ruleset_version'] if diagnostics.key?('ruleset_version')
+          @builder_ruleset_version = diagnostics['ruleset_version'] if diagnostics.key?('ruleset_version')
```
Strech: I think `builder_` prefix doesn't make it clear what is it, as I understand from the code (actually well improved now) it's a version we get after reconfiguration or update.

My suggestion is to name with "what" instead of "how" in mind. Current name reflecting how it will be used (in the builder), when we actually should say what is it, instead (suggestions: `reconfigured_ruleset_version`, `updated_ruleset_version`, `new_ruleset_version`, etc.)

lib/datadog/appsec/context.rb
Strech:
```suggestion
        @interrupted
```

Strech (approved):
Well done 👍🏼 

_P.S I left 2 non-blocking suggestions_

lib/datadog/appsec/security_engine/engine.rb
y9v: fixed [f99771e](https://github.com/DataDog/dd-trace-rb/pull/4839/commits/f99771e2781867325138b4182a61c5f74bb95eea)

PR #4841, opened 2025-08-06
[NO-TICKET] Send forced tests to official system-tests workflow

cbeauchesne:
**What does this PR do?**

Use the new input of system-tests workflow.

I also added a small TODO list in the workflow file: I've realized we'll need to do several minor tasks when we'll switch to this workflow, and it'll be easy to forget one of them.

**Motivation:**

Feature that exists in the custom workflow

**Change log entry**

**Additional Notes:**

**How to test the change?**

The CI here forced `test_standard_tags.py` to be executed. As this file contains test that do not pass, the CI fails as expected : https://github.com/DataDog/dd-trace-rb/actions/runs/16784113445/job/47530527960?pr=4841

### DIFF
--- a/.github/forced-tests-list.cfg
+++ b/.github/forced-tests-list.cfg
@@ -10,5 +10,3 @@
 
 # Example :
 # tests/parametric/test_config_consistency.py::Test_Stable_Config_Default::test_config_precedence
-
-tests/test_standard_tags.py
--- a/.github/workflows/system-tests-official.yml
+++ b/.github/workflows/system-tests-official.yml
@@ -16,7 +16,6 @@ on: # yamllint disable-line rule:truthy
       #
       # This is also useful for testing the workflow without opening a pull request.
       - tmp/*
-      - cbeauchesne/system-tests-forced-tests
   # pull_request:  # To be uncommented if the guild chooses this workflow
   #   branches:
   #     - master


p-datadog:
https://github.com/DataDog/dd-trace-rb/blob/master/docs/ForcingSystemTests.md specifies the tests under scenarios and this PR does not, why is that?

cbeauchesne:
> https://github.com/DataDog/dd-trace-rb/blob/master/docs/ForcingSystemTests.md specifies the tests under scenarios and this PR does not, why is that?

Because it's not needed: if a test is specified in the `-F` parameter, but does not belong to the scenario currently executed, it's ignored.

The reason for that is surface contact between repos : everything that exists on two separated repos makes any change on it complicated and prone to errors. The current definition store three differents informations in common with two repos : 

1. test names
2. scenario names
3. link between scenario and tests

The proposal here remove `2.` and `3.` => if we change a scenario name, or if we move a test from one scenario to another, it won't break your CI.

p-datadog:
Sounds like this PR offers different behavior from what we currently have:

Currently, a test added to forced test list will always run.

This PR proposes to run the tests in the forced test list only in some situations (when scenario somehow is executed, and when this would be is unspecified).

cbeauchesne:
Thanks for the comment. Just to clarify: the assumption that

> "a test added to forced test list will always run"

…is not accurate. There’s no mechanism in system-tests ensuring that forced tests always run unconditionally.

Both the existing workflow and this PR use the `-F` option, and the logic inside system-tests that determines which tests are executed is the same in both cases.

PR #4843, opened 2025-08-07
Adding baggage propagation telemetry metrics

rachelyangdog:
**What does this PR do?**

- Adding telemetry metrics for successful header injections and extractions for baggage
- adding telemetry metrics for malformed baggage header extractions, and when max bytes or max item limits in baggage are exceeded
- for more information: [baggage telemetry RFC](https://docs.google.com/document/d/1N-dux1_n1vVu391Z5IDT_zWUwZp-MPqD7k7bhND8SK8/edit?tab=t.0)

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/spec/datadog/tracing/distributed/baggage_spec.rb
+++ b/spec/datadog/tracing/distributed/baggage_spec.rb
@@ -106,7 +106,7 @@ RSpec.shared_examples 'Baggage distributed format' do
         end
 
         it 'logs a warning and stops injecting excess items' do
-          expect(Datadog.logger).to receive(:warn).with('Baggage item limit exceeded, dropping excess items')
+          expect(Datadog.logger).to receive(:warn).with('Baggage item limit (64) exceeded, dropping excess items')
           inject!
           expect(data['baggage'].split(',').size).to eq(Datadog::Tracing::Distributed::Baggage::DD_TRACE_BAGGAGE_MAX_ITEMS)
         end
@@ -121,7 +121,7 @@ RSpec.shared_examples 'Baggage distributed format' do
         end
 
         it 'logs a warning and stops injecting excess items' do
-          expect(Datadog.logger).to receive(:warn).with('Baggage header size exceeded, dropping excess items')
+          expect(Datadog.logger).to receive(:warn).with('Baggage header size (8192) exceeded, dropping excess items')
           inject!
           expect(data['baggage']).to eq('key1=value1')
         end


lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -157,6 +177,17 @@ def build_baggage_tags(baggage)
 
           tags
         end
+
+        # Record telemetry metrics for baggage operations
```
rachelyangdog, line 188: I'm not sure how I'm supposed to send telemetry metrics..

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -157,6 +183,32 @@ def build_baggage_tags(baggage)
...
+            telemetry&.inc('instrumentation_telemetry_data.tracers', metric_name, value, tags: tags)
+          rescue => e
+            # Silently handle telemetry errors to avoid affecting main functionality
+            ::Datadog.logger.debug("Failed to record baggage telemetry metric: #{e.message}")
+          end
+        end
+
+        def truncate_baggage_items_if_needed(baggage_items)
+          if baggage_items.size > DD_TRACE_BAGGAGE_MAX_ITEMS
+            ::Datadog.logger.warn('Baggage item limit exceeded, dropping excess items')
```
p-datadog: I see that we already have a bunch of identical calls to `logger.warn`, but I want to confirm that:

1) these warnings can be acted on by the customer (or, if not, are still useful to the customer by telling them about dropped data)

2) they are sufficiently informative (we have the maximum number of items available in the code immediately above, but it is not included in the warning message)

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -157,6 +183,32 @@ def build_baggage_tags(baggage)
...
         end
+
+        # Record telemetry metrics for baggage operations
+        def record_telemetry_metric(metric_name, value, tags)
+          begin
+            telemetry = ::Datadog.send(:components).telemetry
+            telemetry&.inc('instrumentation_telemetry_data.tracers', metric_name, value, tags: tags)
+          rescue => e
+            # Silently handle telemetry errors to avoid affecting main functionality
+            ::Datadog.logger.debug("Failed to record baggage telemetry metric: #{e.message}")
```
p-datadog: I understand the reasoning behind adding this rescue but there are other calls to `telemetry.inc` in the library and they are not similarly protected, meaning `telemetry.inc` must not fail in a way that would be rescued here (if it did, this would cause problems for other components like appsec).

Therefore, I think this rescue is not needed.

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -157,6 +183,32 @@ def build_baggage_tags(baggage)
 
           tags
         end
+
+        # Record telemetry metrics for baggage operations
+        def record_telemetry_metric(metric_name, value, tags)
+          begin
+            telemetry = ::Datadog.send(:components).telemetry
+            telemetry&.inc('instrumentation_telemetry_data.tracers', metric_name, value, tags: tags)
```
p-datadog: We agreed that `telemetry` will always be present. DI code treats it as optional but the consensus was that we would expect telemetry object to always be available if component tree is initialized. (Hence the `&` should not be necessary here.)

lib/datadog/tracing/distributed/baggage.rb
rachelyangdog: I initially added this as a helper function but realized it's not needed so I deleted it! Thank you

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -78,6 +92,11 @@ def extract(data)
           # Convert selected baggage items to span tags based on configuration
           baggage_tags = build_baggage_tags(baggage)
 
+          # Record telemetry for successful extraction only if baggage is not empty
+          if !baggage.empty?
```
Strech:
```suggestion
          unless baggage.empty?
```

Strech (approved):
LGTM 👍🏼

p-datadog:
Flaky test reported in https://github.com/DataDog/ruby-guild/issues/251

PR #4844, opened 2025-08-07
Remove Datadog::Tracing::Component include from Core::Configuration::Components

p-datadog:
**What does this PR do?**

Removes the include of Tracing::Component into Core::Configuration::Components. Consequently, changes callers of `build_tracer` etc. to invoke that method on `Tracing::Component`.

Conceptually, the tracer initialization was included into core configuration. Now the tracer initialization is scoped to tracing part of the library.

The unit tests for `build_tracer` and friends have gotten moved from `core` to `tracing`, where they arguably belong.

**Motivation:**

Investigation into injecting telemetry into tracing (for https://github.com/DataDog/dd-trace-rb/pull/4843)

**Change log entry**
None

**Additional Notes:**

**How to test the change?**

Existing unit tests


lib/datadog/core/configuration/components.rb
```diff
@@ -129,7 +127,16 @@ def initialize(settings)
...
 
-          self.class.configure_tracing(settings)
+          # Configure non-privileged components.
+          Datadog::Tracing::Contrib::Component.configure(settings)
+        end
+
+        # Hot-swaps with a new sampler.
+        # This operation acquires the Components lock to ensure
+        # there is no concurrent modification of the sampler.
+        def reconfigure_live_sampler(settings = Datadog.configuration)
```
Strech: If I understood the method right, it's better to call `reconfigure_sampler_live` otherwise it will be a live sampler which is confusing.

But I would even go further and remove `_live_` part because it doesn't highlight much. Better to do then `lock_and_reconfigure_sampler(settings)`

Strech (approved):
I have 1 non-blocking suggestion. Otherwise 👍🏼

lib/datadog/core/configuration/components.rb
```diff
@@ -129,7 +127,16 @@ def initialize(settings)
           @error_tracking = Datadog::ErrorTracking::Component.build(settings, @tracer, @logger)
           @environment_logger_extra[:dynamic_instrumentation_enabled] = !!@dynamic_instrumentation
 
-          self.class.configure_tracing(settings)
+          # Configure non-privileged components.
+          Datadog::Tracing::Contrib::Component.configure(settings)
```
vpellan, line 131: Would it make sense to move this in `Datadog::Tracing::Component.build_tracer` ?

lib/datadog/core/configuration/components.rb
p-datadog: I renamed to `reconfigure_sampler`, locking ought to be an implementation aspect?

lib/datadog/core/configuration/components.rb
p-datadog: From my reading of the code, this line should actually be in `startup!` method because (if I understood correctly) it reconfigures the integration patches to other libraries for the currently active tracer, and the tracer becomes currently active when it starts up not when it is built. If you agree I can make this change in a follow-up PR.

lib/datadog/core/configuration/components.rb
Strech: Yes, locking could be encapsulated (as implementation detail) 👍🏼

lib/datadog/core/configuration/components.rb
vpellan: You're right it makes more sense to put it in `startup!`

PR #4845, opened 2025-08-08
[NO-TICKET] Add typing for `Core::Configuration::Options`, `OptionDefinition` and `Option`

vpellan:
**What does this PR do?**

This PR adds typing for `Core::Configuration::Options`, `OptionDefinition` and `Option`

**Motivation:**

I added a bit of typing on these files in my Config Inversion PR and wanted to make a separate PR for that. I thought why not complete these files. It was harder than I thought and most likely not entirely correct. But this covers our current usage of them, and in the future, we can change them if needed

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

CI should be green


lib/datadog/core/configuration/option.rb
```diff
@@ -172,38 +174,38 @@ def default_precedence?
...
-            end
-
-            values.compact!
-            values.each.with_object({}) do |v, hash|
               pair = v.split(':', 2)
-              hash[pair[0]] = pair[1]
+              result[pair[0]] = pair[1]
             end
+
+            result
```
vpellan: `values` is first evaluated as an Array of String by Steep. `v.empty? ? nil : v` will cause a type error as we replace some values by nil. `each_with_object` also raise a typing error as it will believe that `result` is of `Configuration::Option::env_value` type (as it is `coerce_env_variable` return type) and will say that it does not have method `[]=`

lib/datadog/core/configuration/option.rb
```diff
@@ -172,38 +174,38 @@ def default_precedence?
...
               v.gsub!(/\A[\s,]*|[\s,]*\Z/, '')
+              next if v.empty?
 
-              v.empty? ? nil : v
+              result << v
             end
 
-            values.compact!
-            values
+            result
```
vpellan: Same problem as above

lib/datadog/core/configuration/option.rb
```diff
@@ -329,18 +331,23 @@ def get_value_and_resolved_env_from(env_vars, source: 'environment variable')
...
+            # Steep thinks that Array([String]) will return an array of arrays.
+            envs = definition.env.is_a?(Array) ? definition.env : [definition.env]
+
+            envs.each do |env|
+              env_var = env_vars[env]
+              next if env_var.nil?
 
               resolved_env = env
-              value = coerce_env_variable(env_vars[env])
+              value = coerce_env_variable(env_var)
```
vpellan: Steep considers that
```
next if env_vars[env].nil?

env_vars[env]
```
can still return `nil`, and because `coerce_env_variable` is not accepting `nil`, it will raise a typing error. IIRC, it is because another thread can modify `env_vars[env]` value in between these two instructions

p-datadog (approved):

lib/datadog/core/configuration/option.rb
```diff
@@ -172,38 +174,38 @@ def default_precedence?
         private
 
         def coerce_env_variable(value)
-          return context_exec(value, &@definition.env_parser) if @definition.env_parser
+          return context_exec(value, &definition.env_parser) if definition.env_parser
```
delner: Why convert to attr_reader from instance variable?

Minor, but whatever we go with, I think line 179 and any other reference to definition should match.

lib/datadog/core/configuration/option.rb
```diff
@@ -172,38 +174,38 @@ def default_precedence?
...
 
           case @definition.type
           when :hash
             values = value.split(',') # By default we only want to support comma separated strings
 
-            values.map! do |v|
+            result = {}
+            values.each do |v|
```
delner: Looks like an optimization... fewer loops. 👍

lib/datadog/core/configuration/option.rb
delner: Looking at this again, I think you're right to use the attr_reader. Let's update line 179 too.

lib/datadog/core/configuration/option.rb
delner: > because another thread can modify `env_vars[env]` value in between these two instructions

How do you reason?

lib/datadog/core/configuration/option_definition.rb
```diff
@@ -22,7 +22,7 @@ class OptionDefinition
           :type,
           :type_options
 
-        def initialize(name, meta = {}, &block)
+        def initialize(name, meta, &block)
```
delner, line 25: Why remove the default here? I suppose one probably should provide options in most cases, but this could be a breaking change?

delner (approved):
A question and some minor comments, but otherwise looks good.

lib/datadog/core/configuration/option.rb
```diff
@@ -172,38 +174,38 @@ def default_precedence?
...
           when :int
             Integer(value, 10)
           when :float
             Float(value)
           when :array
             values = value.split(',')
 
-            values.map! do |v|
+            result = []
+            values.each do |v|
```
Strech: A question: Why not `each_with_object` which is more idiomatic here (and other places)?

lib/datadog/core/configuration/option.rb
```diff
@@ -28,7 +28,8 @@ module Precedence
             def <=>(other)
               return nil unless other.is_a?(Value)
 
-              numeric <=> other.numeric
+              # Steep does not handle well Structs (https://github.com/ruby/rbs/blob/master/docs/data_and_struct.md)
+              numeric <=> other.numeric # steep:ignore NoMethod
```
Strech: I've typed structs, not sure what is the issue here? If you define methods or include the module it should work as structs just class definitions

lib/datadog/core/configuration/option.rb
```diff
@@ -329,18 +331,23 @@ def get_value_and_resolved_env_from(env_vars, source: 'environment variable')
           resolved_env = nil
 
           if definition.env
-            Array(definition.env).each do |env|
-              next if env_vars[env].nil?
+            # Steep thinks that Array([String]) will return an array of arrays.
+            envs = definition.env.is_a?(Array) ? definition.env : [definition.env]
```
Strech: I don't like idea of changing code because some tool is uncapable of handling basics.

It would be good to see the output, because I know that `Array(Hash)` will cause problems, but definitely not `Array([String])`

lib/datadog/core/configuration/option.rb
vpellan: I wrote a small example of what I believe is the reason why Steep thinks that `env_vars[env]` might still be nil (I'm sure I've read this in a Github issue a while ago but cannot find it anymore)
```
h = {key: 1}

thread1 = Thread.new do
  puts 'should not be printed' if h[:key] == nil

  sleep 0
  puts 'key is now nil' if h[:key] == nil
end

thread2 = Thread.new do
  h[:key] = nil
end

thread1.join
thread2.join
```
`sleep 0` is to simulate the GVL changing the thread that may happen between these two instructions. This will only print `key is now nil`. In our case, Steep thinks that `env_vars[env]` can be changed to nil by another thread after the guard clause, so even with the guard clause, it will still type it to `Value?` instead of `Value`. But if we create a copy of `env_vars[env]`, it cannot be changed to `nil` by another.

lib/datadog/core/configuration/option_definition.rb
vpellan: `OptionDefinition.new` is only used by `Builder.to_definition`, which always provides `meta`. The default value may actually create bugs. In this `initialize` method, `@type_options` will be set to `nil` if we use the default value of `meta` (while `type_options` is always set to `{}` by the Builder). [This condition](https://github.com/DataDog/dd-trace-rb/pull/4845/files#diff-967e3a5897f6948c0c094485c27ee3b500bf458d2654af77553c51e133a9cd0dL229) will then cause a `NoMethodError` because it will try to use `[]` on `nil`

lib/datadog/core/configuration/option.rb
vpellan: Updated in 660fc42783ba228702d4c8b30b15b50cc9dad7a0

lib/datadog/core/configuration/option.rb
vpellan: Reverted in 3af0c1e0910723f0863a6a5bb4e0b3bd27da6ce4

lib/datadog/core/configuration/option.rb
vpellan: During our call, we concluded that Struct should not be used here. It is more complicated that a simple struct, as it includes Comparable and defines a method. We changed it to a class in 3af0c1e0910723f0863a6a5bb4e0b3bd27da6ce4

lib/datadog/core/configuration/option.rb
vpellan: Fixed in 3af0c1e0910723f0863a6a5bb4e0b3bd27da6ce4 (I still needed to add a variable so that I could annotate it, but this will be removed once config inversion is merged)

lib/datadog/core/configuration/option.rb
```diff
@@ -22,12 +22,22 @@ module Precedence
...
+            def initialize(numeric, name, origin)
+              @numeric = numeric
+              @name = name
+              @origin = origin
+            end
+
             def <=>(other)
               return nil unless other.is_a?(Value)
 
+              # Steep does not handle well Structs (https://github.com/ruby/rbs/blob/master/docs/data_and_struct.md)
```
Strech: I guess it's obsolete now

lib/datadog/core/configuration/option.rb
```diff
@@ -159,7 +169,8 @@ def reset
 
         def default_value
           if definition.default.instance_of?(Proc)
-            context_eval(&definition.default)
+            # Steep does not handle well Procs (https://github.com/ruby/rbs/issues/736)
+            context_eval(&definition.default) # steep:ignore BlockTypeMismatch
```
Strech: Could it be fixed with annotation now?

lib/datadog/core/configuration/option.rb
Strech: I guess it could stay as is in favor of Steep annotation instead of code change in favor of the tool?

lib/datadog/core/configuration/option.rb
```diff
@@ -212,7 +217,7 @@ def coerce_env_variable(value)
             value
           else
             raise InvalidDefinitionError,
-              "The option #{@definition.name} is using an unsupported type option for env coercion `#{@definition.type}`"
+              "The option #{definition.name} is using an unsupported type option for env coercion `#{definition.type}`"
```
Strech: IMHO there is no need to use reader, 99% cases readers is for public interface and I don't see any benefit of that change

lib/datadog/core/configuration/option.rb
```diff
@@ -329,18 +334,27 @@ def get_value_and_resolved_env_from(env_vars, source: 'environment variable')
           resolved_env = nil
 
           if definition.env
-            Array(definition.env).each do |env|
-              next if env_vars[env].nil?
+            # Steep thinks that Array([String]) will return an array of arrays.
+            # We create a new variable and annotate it to avoid the error.
+            # As aliases will be removed from the DSL with config inversion,
+            # this will be removed in the future as `definition.env` will be a `String?`.
```
Strech: Obsolete?

lib/datadog/core/configuration/option.rb
```diff
@@ -329,18 +334,27 @@ def get_value_and_resolved_env_from(env_vars, source: 'environment variable')
...
+            #
+            # @type var envs: Array[String]
+            envs = Array(definition.env)
+            envs.each do |env|
+              env_var = env_vars[env]
```
Strech: to be honest naming is killing me here, if I read next lines I loose focus because all things look alike. It makes sense to simplify naming - less short names, less prefixes and more sense

lib/datadog/core/configuration/options.rb
```diff
@@ -55,7 +57,9 @@ def define_helpers(helpers)
             helpers.each do |name, block|
               next unless block.is_a?(Proc)
 
-              define_method(name, &block)
+              # Steep doesn't understand relation between Method and Proc.
+              # https://github.com/ruby/rbs/issues/736
+              define_method(name, &block) # steep:ignore BlockTypeMismatch
```
Strech: Could annotation resolve this?

lib/datadog/core/configuration/options.rb
```diff
@@ -125,7 +129,7 @@ def assert_valid_option!(name)
           end
         end
 
-        InvalidOptionError = Class.new(StandardError)
+        class InvalidOptionError < StandardError; end
```
Strech: This is a wrong direction, no need to open class to be empty and previous version was correct

sig/datadog/core/configuration/option.rbs
```diff
@@ -19,40 +23,60 @@ module Datadog
...
-            attr_reader numeric: int
-            attr_reader name: Symbol
-            attr_reader origin: String
+            def initialize: (Integer numeric, Symbol name, String origin) -> void
+            def <=>: (untyped other) -> Integer?
           end
         end
 
-        def initialize: (Configuration::OptionDefinition definition, Configuration::Settings context) -> void
+        # context is a class that includes `Configuration::Options::InstanceMethods`.
```
Strech: I think it should be interface definition instead of comment

sig/datadog/core/configuration/option.rbs
```diff
@@ -19,40 +23,60 @@ module Datadog
...
-        def set: (untyped value, ?precedence: Precedence::Value) -> void
+        def set: (Object? value, ?precedence: Precedence::Value, ?resolved_env: String?) -> Object?
 
-        def get: () -> untyped
+        def get: () -> Object?
 
         def reset: () -> void
 
-        def default_value: () -> untyped
+        def default_value: () -> (Configuration::OptionDefinition::option_type | Configuration::Options::GenericSettingsClass)
```
Strech: Already presented namespace could be skipped

sig/datadog/core/configuration/option.rbs
```diff
@@ -19,40 +23,60 @@ module Datadog
...
-        def validate: (untyped type, untyped value) -> untyped
+        def validate: (Symbol? type, Object? value) -> bool
 
-        def context_exec: (*untyped args) { () -> untyped } -> untyped
+        def internal_set: (Object? value, Precedence::Value precedence, String? resolved_env) -> Object?
 
-        def context_eval: () { () -> untyped } -> untyped
+        # `context_exec/eval` can take any block (and args) and just apply them to the context using `instance_exec/eval`.
+        # Which makes it impossible to type. Even if we know the type of the blocks that uses `context_exec`,
+        # in the case of `after_set`, the block returns void, but `setter`, `resetter` and `env_parser` return a value.
```
Strech: I think it's just a type alias instead of comment then

Strech (changes requested):
I think it's a right direction to move in some areas, but there are still lots of ignore and the code that could be improved or returned to a previous state.

I would like to suggest either to narrow down the scope of changes and take out pieces of the PR and do as a separate less bulky changes or continue iteration

lib/datadog/core/configuration/options.rb
p-datadog: `        class InvalidOptionError < StandardError; end` is how classes are normally defined, no?

Strech: No, for cases when you don't need to open class for extra definitions like initialize and methods, you can use open-less definition.

vpellan: In that case I will also correct https://github.com/DataDog/dd-trace-rb/pull/4845/files#diff-7e0fbcca4a616a5c1dae88b8c33fd4f534021639f9370b0c9496281ec25c9ae1R47

lib/datadog/core/configuration/option.rb
vpellan: I could not find any annotation for that purpose, but adding an intermediate value `env_parser` set to `@definition.env_parser` then using this value instead of `definition.env_parser` does not change the code logic and also works. Changed in c0e0991180e95d5d920288b28f482dcb40ebe89a

lib/datadog/core/configuration/option.rb
vpellan: Removed in c0e0991180e95d5d920288b28f482dcb40ebe89a

lib/datadog/core/configuration/option.rb
vpellan: Not with annotation but typing context_eval and context_exec with type variable fixed it (c0e0991180e95d5d920288b28f482dcb40ebe89a)

lib/datadog/core/configuration/option.rb
vpellan: Reverted it in c0e0991180e95d5d920288b28f482dcb40ebe89a

lib/datadog/core/configuration/option.rb
vpellan: Removed in c0e0991180e95d5d920288b28f482dcb40ebe89a

lib/datadog/core/configuration/option.rb
vpellan: Changed naming in c0e0991180e95d5d920288b28f482dcb40ebe89a

lib/datadog/core/configuration/options.rb
vpellan: Not with annotation but actually defining the block instead of defining it as a Proc fixed it (c0e0991180e95d5d920288b28f482dcb40ebe89a)

lib/datadog/core/configuration/options.rb
vpellan: Reverted in c0e0991180e95d5d920288b28f482dcb40ebe89a

sig/datadog/core/configuration/option.rbs
vpellan: I created Interface in c0e0991180e95d5d920288b28f482dcb40ebe89a but did not used it in that case. context needs to be a class so we can use methods like instance_of? or nil?. Instead, I typed `self` to `GenericSettingsClass` using annotation [here](https://github.com/DataDog/dd-trace-rb/pull/4845/commits/c0e0991180e95d5d920288b28f482dcb40ebe89a#diff-9c7b169fea70e91e80f1011cd63de7c5fe9ae0ca0a89f14198e9d0405c599b1bR121)

sig/datadog/core/configuration/option.rbs
vpellan: Removed in c0e0991180e95d5d920288b28f482dcb40ebe89a

sig/datadog/core/configuration/option.rbs
vpellan: Used Type variable instead c0e0991180e95d5d920288b28f482dcb40ebe89a

lib/datadog/core/configuration/option.rb
BridgeAR: I could imagine this is an issue with steep itself. I don't know enough about thread safety in Ruby to be certain about it though.
Suggestion: the change itself won't hurt one way or the other, so changing it should be fine. On top of that, we should check for issues in steep and if none exists, we can open a new one. That way it can be addressed in the right spot.

lib/datadog/core/configuration/option.rb
vpellan: I will open an issue on https://github.com/ruby/rbs about that

lib/datadog/core/configuration/option.rb
y9v: @vpellan I don't think the example is similar to the issue you have - in the example you have a global variable `h` that could be changed by other thread. In the code `env_vars` is the method argument, a local variable. How could it be changed from another thread?

lib/datadog/core/configuration/option.rb
```diff
@@ -172,20 +181,17 @@ def default_precedence?
...
+            values.each_with_object({}) do |v, hash| # $ Hash[String, String]
               v.gsub!(/\A[\s,]*|[\s,]*\Z/, '')
+              next if v.empty?
 
-              v.empty? ? nil : v
-            end
-
-            values.compact!
-            values.each.with_object({}) do |v, hash|
               pair = v.split(':', 2)
```
marcotc, line 195: Great work improving performance here!

sig/datadog/core/configuration/option.rbs
```diff
@@ -2,15 +2,19 @@ module Datadog
   module Core
     module Configuration
       class Option
-        @context: Configuration::Settings
+        type env_value = Hash[String, String] | Array[String] | String | Integer | Float | bool | nil
+        type env_vars = Hash[String, String] | ::RBS::Unnamed::ENVClass
```
marcotc, line 6: `::RBS::Unnamed::ENVClass` TIL

lib/datadog/core/configuration/option.rb
vpellan: I'm not sure what triggers the error in that case and why creating an intermediate variable fixes it but this was the closest issue I could think of. I will try to reproduce that error in a separate project and open an issue on steep repository.

sig/datadog/core/configuration/option.rbs
vpellan: https://github.com/ruby/rbs/blob/607dc0573276048ba41f07fe160df102898fe717/core/env.rbs It seems that it's undocumented though

PR #4861, opened 2025-08-19
DI Expression Language initial implementationt

p-datadog:
**What does this PR do?**

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**


lib/datadog/di/el/compiler.rb
```diff
@@ -0,0 +1,160 @@
...
+        def compile_partial(ast)
+          case ast
+          when Hash
+            if ast.length != 1
+              raise DI::Error::InvalidExpression, "Expected hash of length 1: #{ast}"
+            end
+            op, target = ast.first
+            case op
+            when 'ref'
+              unless String === target
```
Strech, line 57: If you want, you can also go this way

```suggestion
              unless target.is_a?(String)
```

lib/datadog/di/el/compiler.rb
```diff
@@ -0,0 +1,160 @@
...
+                raise NotImplementedError
+              else
+                # Ruby technically allows all kinds of symbols in variable
+                # names, for example spaces and many characters.
+                # Start out with strict validation to avoid possible
+                # surprises and need to escape.
+                unless target =~ %r{\A(@?)([a-zA-Z0-9_]+)\z}
+                  raise DI::Error::BadVariableName, "Bad variable name: #{target}"
+                end
+                method_name = (($1 == '@') ? 'iref' : 'ref')
```
Strech, line 84: What about `@@cvar`?

lib/datadog/di/el/compiler.rb
```diff
@@ -0,0 +1,160 @@
...
+          if Hash === target && target.length == 1 && target.keys.first == 'ref' &&
+              String === (value = target.values.first)
+            escape(value)
+          else
+            '(expression)'
+          end
+        end
+
+        def escape(needle)
+          needle.gsub("\\") { "\\\\" }.gsub('"') { "\\\"" }.gsub('#') { "\\#" }
```
Strech, line 155: I think this is an expensive escaping. Will shellscape work or at least we can avoid using blocks

```ruby
needle.gsub("\\", "\\\\")
```

lib/datadog/di/el/compiler.rb
```diff
@@ -0,0 +1,160 @@
...
+              compiled_targets = target.map do |item|
+                "(#{compile_partial(item)})"
+              end
+              compiled_op = MULTI_ARG_METHODS[op]
+              "(#{compiled_targets.join(" #{compiled_op} ")})"
+            when 'substring'
+              unless Array === target && target.length == 3
+                raise DI::Error::InvalidExpression, "Improper #{op} syntax"
+              end
+              "#{op}(#{target.map { |target| "(#{compile_partial(target)})" }.join(", ")})"
```
watson: Tiny nit: Consider not re-using the `target` variable name to avoid confusion with the other `target` variable

```suggestion
              "#{op}(#{target.map { |arg| "(#{compile_partial(arg)})" }.join(", ")})"
```

lib/datadog/di/instrumenter.rb
```diff
@@ -310,9 +310,16 @@ def hook_line(probe, &block)
...
             )
-              if rate_limiter.nil? || rate_limiter.allow?
+              continue = if condition = probe.condition
+                locals = Instrumenter.get_local_variables(tp)
+                context = EL::Context.new(locals: locals, target: tp.self)
+                condition.satisfied?(context)
+              else
+                true
+              end
+              if continue and rate_limiter.nil? || rate_limiter.allow? # standard:disable Style/AndOr
```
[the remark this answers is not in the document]
p-datadog: and/or are used intentionally to remove the need for parentheses.

docs/DynamicInstrumentation.md
```diff
@@ -0,0 +1,40 @@
...
+rather than an error, as in some other languages' implementations of
+expression language.
+- `getmember` looks up instance variables, not attributes (which look
+like method calls, and could be implemented via methods).
+DI avoids running application code as much as possible for safety and
+performance reasons.
+- Since expression language evaluation is done in Ruby, it is not
+possible to guarantee that it will not invoke application code
+(since all operations in Ruby can be redefined at runtime, even for example
+addition of numbers).
```
watson, line 40: Isn't it possible, at runtime, to check if a given property access is going to execute userland code?

lib/datadog/di/el/compiler.rb
```diff
@@ -0,0 +1,160 @@
...
+          when String
+            "\"#{escape(ast)}\""
+          when Array
+            # Arrays are commonly used as arguments of operators/methods,
+            # but there are no arrays at the top level in the syntax that
+            # we currently understand. Provide a helpful error message in case
+            # syntax is expanded in the future.
+            raise DI::Error::InvalidExpression, "Array is not valid at its location, do you need to upgrade dd-trace-rb? #{ast}"
+          else
+            raise DI::Error::InvalidExpression, "Unknown type in AST: #{ast}"
```
watson, line 137: I assume these exceptions are caught elsewhere for both logging purposes and to basically invalidate the probe, and that they don't interfere with the running of the instrumented application?

lib/datadog/di/instrumenter.rb
watson, line 320: Wouldn't it be better to move the `rate_limiter.allow?` check up before the probe condition is evaluated? I assume evaluating the probe condition is much more expensive than checking if we're being rate limited, in which case the condition doesn't matter, right?

docs/DynamicInstrumentation.md
p-datadog: Yes but it requires doing the work in C. Currently DI is implemented entirely in Ruby. Implementing safety-critical parts in C (which includes serialization, besides expression language evaluation) would alleviate many if not all concerns but is not currently in scope.

lib/datadog/di/el/compiler.rb
p-datadog: There are existing exception rescues that catch both compilation and evaluation exceptions and prevent them from affecting customer application.

I reviewed the relevant code again and for compilation exceptions, I was missing probe status reporting which I have now added. I also added integration tests for both types of exceptions.

For evaluation exceptions, they should be reported via `evaluationErrors` which is not part of this PR. For now those exceptions get caught but are not properly reported.

lib/datadog/di/instrumenter.rb
p-datadog: We cannot check rate limiter before condition evaluation to guarantee probe execution when condition is met infrequently.

However I think the code can be optimized to skip expression evaluation if rate limit is not met (i.e. after the first time the probe is invoked). I looked at what would be involved in that and to keep the size of this PR from growing even more I would like to implement this optimization in a separate PR. This is going to require some changes to the rate limiter because presently the check is part of the code which also marks the rate limiter as having been "used".

lib/datadog/di/el/context.rb
```diff
@@ -0,0 +1,31 @@
...
+          @target = target
+        end
+
+        attr_reader :locals
+        attr_reader :target
+
+        def fetch(var_name)
+          # TODO this should be a bad reference?
+          return nil unless locals
+          locals[var_name.to_sym]
```
Strech, line 22: Could this lead to symbols explosion? (even tho now they are collectable, but still)

lib/datadog/di/el/evaluator.rb
```diff
@@ -0,0 +1,167 @@
...
+            var.length
+          else
+            raise DI::Error::ExpressionEvaluationError, "Unsupported type for length: #{var.class}: #{var_name}"
+          end
+        end
+
+        def is_empty(var, var_name)
+          if var.nil? || Numeric === var
+            false
+          elsif [Array, String].include?(var.class)
```
Strech: This will always create a new array, maybe simple `or` condition is good enough?

```suggestion
          elsif var.is_a?(Array) || var.is_a?(String)
```

lib/datadog/di/el/evaluator.rb
```diff
@@ -0,0 +1,167 @@
...
+              block.call([key, value], key, value)
+            end.to_h
+          else
+            raise DI::Error::ExpressionEvaluationError, "Bad collection type for filter: #{collection.class}"
+          end
+        end
+
+        def instanceof(object, cls_name)
+          cls = object.class
+          loop do
```
Strech, line 155: Could this loop potentially endless? Does it make sense to put guard to the superclass above which you will not move, like `BasicObject`?

lib/datadog/di/error.rb
```diff
@@ -48,6 +48,31 @@ class ProbePreviouslyFailed < Error
...
+      # leaving some choices to the language implementation in the tracers.
+      # It is therefore possible that some technically valid expressions are
+      # prohibited by our implementation.
+      #
+      # It is also possible that the sanitizers/validators prohibit some
+      # esoteric constructs that are technically valid in Ruby,
+      # for example if instance variable name rules are relaxed to allow
+      # arbitrary characters in them as permitted in method names.
+      class InvalidExpression < Error
+      end
```
Strech, line 65: If you don't plan to open class for come definitions, you can save some instructions

```suggestion
      InvalidExpression = Class.new(Error)
```

lib/datadog/di/probe_builder.rb
```diff
@@ -39,6 +40,7 @@ module ProbeBuilder
           max_capture_depth: config["capture"]&.[]("maxReferenceDepth"),
           max_capture_attribute_count: config["capture"]&.[]("maxFieldCount"),
           rate_limit: config["sampling"]&.[]("snapshotsPerSecond"),
+          condition: (cond = config.dig('when', 'json')) && EL::Compiler.new.compile(cond), # steep:ignore
```
Strech, line 43: instead of ignore, you can do type assertion with `#: <type>`

lib/datadog/di/el/compiler.rb
p-datadog: I don't actually because the === version is implemented entirely by Ruby runtime and its behavior is fixed, whereas `is_a?` can be overridden by customer code.

lib/datadog/di/el/compiler.rb
p-datadog: Class variables are not supported at the moment. They have no equivalent in other languages and most likely won't be supported until there is customer demand for them.

lib/datadog/di/el/compiler.rb
p-datadog: 1) 

`needle.gsub("\\", "\\\\")` is a no-op, as we recently learned

2)

In my test the block version is faster:

```
irb(main):004> Benchmark.realtime { 1_000_000.times { "abc\\def".gsub("\\") { "\\\\" }}}
=> 0.615793973000109
irb(main):005> Benchmark.realtime { 1_000_000.times { "abc\\def".gsub("\\", "\\\\") }}
=> 0.7263220919999185
```

Probably because the non-block version evaluates/expands stuff in the replacement string.

lib/datadog/di/el/context.rb
p-datadog: Explosion - unlikely unless the customer tries really hard (and then they would be DOSing their own application...).

The locals come from the runtime with symbol keys, not converting the variable name to a symbol here means I would have to convert *all* of the local names to strings.

lib/datadog/di/el/evaluator.rb
p-datadog: Repaired.

lib/datadog/di/el/compiler.rb
Strech: Wow, you're right for the ruby > 3.3 it is definitely true!

```
ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [arm64-darwin23]
              String   239.000 i/100ms
               Block   292.000 i/100ms
              String      2.416k (± 1.5%) i/s  (413.83 μs/i) -     12.189k in   5.045389s
               Block      2.854k (± 2.7%) i/s  (350.38 μs/i) -     14.308k in   5.017227s

Comparison:
               Block:     2854.0 i/s
              String:     2416.5 i/s - 1.18x  slower

```

lib/datadog/di/el/evaluator.rb
p-datadog:
```
irb(main):001> BasicObject.superclass
=> nil
```
I suspect if this behavior were to change, there will be all kinds of problems with all kinds of existing Ruby code.

lib/datadog/di/el/compiler.rb
Strech: For the no-op this is the way, but block is faster, so doesn't make much sense

```ruby
"abc\\def".gsub("\\", "\\\\\\")
```

lib/datadog/di/error.rb
p-datadog: I don't understand what the savings are supposed to be with the second version but to me the second version is used for run-time class definition, in this case I am defining the class lexically and Ruby has a syntax for that which is the one I am using.

lib/datadog/di/el/compiler.rb
Strech:
```
irb(main):023* class A
irb(main):024*   def ===(klass)
irb(main):025*     false
irb(main):026*   end
irb(main):027> end
=> :===
irb(main):028> A === A
=> false
irb(main):029> B = Class.new
=> B
irb(main):030> B == B
=> true
```

lib/datadog/di/probe_builder.rb
p-datadog: I haven't gotten value from type checking to justify the time expenditure on writing even the "basic" types, much less trying to figure out how to deal with complex situations (and steep bugs/limitations). There is some typed code in DI but most of the values are untyped therefore dealing with non-trivial cases is not worthwhile.

p-datadog:
> WDYT about some faster parsing with crafted parser, or it's irrelevant for that? ruby/racc or any other?

Other languages generate bytecode directly. I looked into doing so for Ruby briefly but hadn't made meaningful progress. I intend to revisit this.

Because we already receive the expressions in an AST, compilers in the traditional sense probably are not the best tool to use.

lib/datadog/di/el/compiler.rb
p-datadog: Right. === is still better because it requires customers to patch core classes which is supposed to be less common than overriding is_a?. But, fair point, and it means the type check should be done in C to prevent customer overrides from influencing it, which means optimizing it is also not very valuable.

Strech: Ok 👍🏼

lib/datadog/di/el/compiler.rb
```diff
@@ -0,0 +1,160 @@
...
+                # Start out with strict validation to avoid possible
+                # surprises and need to escape.
+                unless target =~ %r{\A(@?)([a-zA-Z0-9_]+)\z}
+                  raise DI::Error::BadVariableName, "Bad variable name: #{target}"
+                end
+                method_name = (($1 == '@') ? 'iref' : 'ref')
+                "#{method_name}('#{target}')"
+              end
+            when *SINGLE_ARG_METHODS
+              method_name = op.gsub(/[A-Z]/) { |m| "_#{m.downcase}" }
```
y9v, line 88: can't we just call `downcase` on the string, since it will anyway only downcase letters? Or am I missing something?

p-datadog: The code adds underscores between words.

Strech (approved):
🎉 Well done!

lib/datadog/di/el/evaluator.rb
Strech: Yep, just want to be sure it will not loop forever 👍🏼

p-datadog:
One optimization I already made for a future PR is defining the evaluation method during expression "compilation" rather than evaluating the string at runtime as this PR is doing. This optimization produces a speed-up of around 15% for the evaluation.

Then I looked at the generated code using RubyVM::InstructionSequence#disasm. It's actually very lightweight, pretty much just an invocation of the methods in Evaluator to perform the operations/access variables. I'm certain there are further optimizations that can be made but I expect they would be in the micro-optimization category and I suspect changing to racc or any other parser won't be beneficial.

For example, for regular expression matching, the regular expression should be compiled during the "compilation" phase of EL and not during evaluation as this PR is doing.

I see the primary benefit of generating bytecode directly being safety - we wouldn't be calling `eval` on something that is derived from user input - rather than performance.

watson (approved):
Great work!

PR #4862, opened 2025-08-19
Generate metrics for GraphQL operation execution

marcotc:
**What does this PR do?**
This PR adds the span responsible for generating service page metrics to the `graphql.execute` span, which represents a single GraphQL operation execution.
This allows for toggling the APM service service page between, for example, `rack.request` and `graphql.execute`.
The default today is to prioritize `graphql.execute` as the primary operation, but this wasn't working for Ruby because there were no metrics to populate the the service page with. This PR fixes that.

The service page will look something like this, with GraphQL operations in the `Resources` tab (instead of `POST /graphql`):

Screenshot 2025-08-19 at 3 43 21 PM

This PR also a changes in the resource name for the `graphql.execute`, from just the operation name (`simpleQuery`) to operation type + operation name (`query simpleQuery`). This aligns closely with the desired grouping pattern for GraphQL operations, and prevents conflicts when a mutation and query have the same name.
This is a non-breaking change because the `graphql.execute` span could not be monitored on, since there were no metrics generated for it (we won't be able to change it this easily anymore 🙂).

**Change log entry**
Yes. * Added: Generate metrics for GraphQL operation execution. * Changed: The `graphql.execute` span resource now includes the operation type.


lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -50,8 +50,12 @@ def execute_query(*args, query:, **kwargs)
             trace(
               proc { super },
               'execute',
-              query.selected_operation_name,
+              operation_resource(query.selected_operation),
               lambda { |span|
+                # Ensure this span can be aggregated by in the Datadog App, and generates RED metrics.
+                span.set_tag(Tracing::Metadata::Ext::TAG_KIND, Tracing::Metadata::Ext::SpanKind::TAG_SERVER)
+                span.set_tag('span.kind', 'server')
```
p-datadog: Are these two lines doing the same thing?

Strech (approved):
Awesome 👏🏼 

P.S still need to eliminate extra tag kind setter

vpellan:
LGTM, beside the duplicate tag !

lib/datadog/tracing/contrib/graphql/unified_trace.rb
marcotc: they are!!! nice catch!

Issue #4852, opened 2025-08-15
[BUG]: Constant resolution issue with rails runner and datadog 2.19.0


jonswar:
### Tracer Version(s)

2.19.0

### Ruby Version(s)

ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [arm64-darwin22]

### Bug Report

After upgrading to Datadog 2.19.0, there's an issue with constant resolution when using `rails runner` to execute Ruby files. The issue occurs when a script tries to reference constants defined within a module namespace that were loaded during Rails initialization.

The issue appears to be related the patching of `Kernel.load`. If I comment out [lines 78-110 of lib/datadog/tracing/contrib/rails/runner.rb](https://github.com/DataDog/dd-trace-rb/blob/0804cf3e153528fe774c4f61703e99f1d6813f07/lib/datadog/tracing/contrib/rails/runner.rb#L78-L110), the bug goes away.

## Reproduction Steps

1. Create a new Rails application:
```bash
rails new bug --minimal
```

2. Add datadog 2.18.0 to the Gemfile:
```ruby
gem 'datadog', '2.18.0'
```

3. Create a file `config/initializers/datadog.rb`:
```ruby
require 'datadog/tracing'

Datadog.configure do |c|
  c.tracing.instrument :rails
end
```

4. Create a file `config/initializers/example_module.rb`:
```ruby
module ExampleNamespace
  module Jobs
  end
end
```

5. Create a file `lib/tasks/example_task.rb`:
```ruby
module ExampleNamespace
  puts Jobs.name
end
```

6. Run the task, it should print "ExampleNamespace::Jobs":
```bash
bundle exec rails runner lib/tasks/example_task.rb
```

7. Upgrade to datadog 2.19.0 in Gemfile
```ruby
gem 'datadog', '2.18.0'
```

8. Run the task, it should fail:
```bash
bundle exec rails runner lib/tasks/example_task.rb
```

### Configuration Block

See above

### Error Logs

N/A

### Operating System

Darwin Kernel Version 24.5.0 (but also repro'd on Linux)

ivoanjo:
Hey! Thanks for the great report. If you have it around still, it would be great to have a backtrace of when the issue happens :)

For @marcotc: This does seem to be a recent regression since https://github.com/DataDog/dd-trace-rb/pull/4681 was where we introduced this?

jonswar:
Sure.

```
% bundle exec rails runner lib/tasks/example_task.rb
[...]

PR #4863, opened 2025-08-20
[APMS-16842] Fix Rails Runner constants defined within a module namespace wrapped in anonymous module

vpellan:
**What does this PR do?**

This PR fixes a bug where Rails Runner constants that were defined within a module namespace were wrapped in an anonymous module.

In Rails runner source code, the `perform` method calls `Kernel.load` with a single argument, the `filename`. But there's a second argument, `wrap`, that is defaulted to `false` in Ruby source code. So Rails calls `Kernel.load` with `wrap` default value, `false`. In our code, we changed the default value to `true`. When `wrap` is set to `true`, the loaded script will be executed under an anonymous module. The reason why this value has been changed to `true` is unknown and setting it back to `false` fixes the issue.

**Motivation:**

https://github.com/DataDog/dd-trace-rb/issues/4852

**Change log entry**

None.

**Additional Notes:**

`self.prepend` is never called as it contains a typo (we need to name it `self.prepended`)
2.19.0 fix still instrumented Rails code because we defined a `Kernel` module, and Rails code was using this module instead of the top module when calling `Kernel.load`. In this case, `super` was calling `Module.load` instead of `Kernel.load` (which was still working because `Module` has `Kernel` in its ancestors). I've renamed `Kernel` to `InstrumentedKernel`  and fixed the typo so that `self.prepended` is executed

**How to test the change?**

I added a test that reproduces the ticket scenario


lib/datadog/tracing/contrib/rails/runner.rb
```diff
@@ -69,14 +69,14 @@ def eval(*args)
             end
           end
 
-          def self.prepend(base)
-            base.const_set(:Kernel, Kernel)
+          def self.prepended(base)
+            base.const_set(:Kernel, InstrumentedKernel)
           end
```
vpellan, line 74: What would be better between this, and deleting `self.prepended` and keep the `Kernel` name for the module

lib/datadog/tracing/contrib/rails/runner.rb
```diff
@@ -105,7 +105,7 @@ def self.load(file, wrap = true)
                 end
                 Contrib::Analytics.set_rate!(span, Datadog.configuration.tracing[:rails])
 
-                super
+                Kernel.load(file, wrap)
```
y9v: I would not change `super` to `Kernel.load` - there might be other modules prepended to `Kernel`, and we probably wouldn't want to skip them

vpellan: Reverted in [1d09b09](https://github.com/DataDog/dd-trace-rb/pull/4863/commits/1d09b0989b650705a15ad2fdc08ecf4cd9aeb5a1)

y9v (approved):
:+1:

lib/datadog/tracing/contrib/rails/runner.rb
```diff
@@ -69,14 +69,14 @@ def eval(*args)
...
+          def self.prepended(base)
+            base.const_set(:Kernel, InstrumentedKernel)
           end
 
           # Instruments the `Kernel.load` method, but only for the
           # `Rails::Command::RunnerCommand` class.
-          module Kernel
-            def self.load(file, wrap = true)
+          module InstrumentedKernel
+            def self.load(file, wrap = false)
```
p-datadog: Why not use `self.load(*args)` which permits the signature to change (by new Ruby versions) and makes no assumption on the correct default value?

lib/datadog/tracing/contrib/rails/runner.rb
p-datadog: Is this the call you are attempting to override?
```
railties/lib/rails/commands/runner/runner_command.rb:            Kernel.load expanded_file_path
```

vpellan: Yes exactly, line 44. This is also what the previous fix did

lib/datadog/tracing/contrib/rails/runner.rb
vpellan: You're right, thanks ! Changed in [c86cf57](https://github.com/DataDog/dd-trace-rb/pull/4863/commits/c86cf578ebc88eb64da5cd21fe819b231e96e831)

p-datadog (approved):

lib/datadog/tracing/contrib/rails/runner.rb
p-datadog: :+1: I do not have an opinion on your question, either way is good with me.

marcotc (approved):

PR #4864, opened 2025-08-21
[AIDM-6] Surface GraphQL errors in Error Tracking

marcotc:
**What does this PR do?**
Adds [Datadog Error Tracking](https://docs.datadoghq.com/error_tracking/) support for GraphQL, using with OpenTelemetry semantics. This allows rich reporting and monitoring on GraphQL execution errors.

Screenshot 2025-08-21 at 4 39 21 PM

**Change log entry**
Yes. Add GraphQL error tracking with OpenTelemetry semantics

**Additional Notes:**
- Uses the [`exception` event name and standard OpenTelemetry attributes](https://opentelemetry.io/docs/specs/semconv/exceptions/exceptions-spans/#exception-event) when enabled
- Backward compatible, disabled by default


docs/GettingStarted.md
```diff
@@ -890,12 +890,12 @@ To activate your integration, use the `Datadog.configure` method:
 # Inside Rails initializer or equivalent
 # For graphql >= v2.2
 Datadog.configure do |c|
-  c.tracing.instrument :graphql, with_unified_tracer: true, **options
+  c.tracing.instrument :graphql, error_tracking: true, with_unified_tracer: true, **options
```
vpellan, line 893: I believe this should not be added here, as the list of options below is already describing it, unless we plan to enable it by default in the future (which is why we have `with_unified_tracer: true` here.) Same for graphql < 2.2

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -224,21 +227,43 @@ def add_query_error_events(span, errors)
...
+            else
+              @event_name = Ext::EVENT_QUERY_ERROR
+              @message_key = 'message'
+              @type_key = 'type'
+              @stacktrace_key = 'stacktrace'
+              @locations_key = 'locations'
+              @path_key = 'path'
+              @extensions_key = 'extensions.'
+            end
+          end
```
vpellan: I'm not a fan of doing this in a separate method as I don't think it improves code readability, as you have to look for this method to understand what instance variables initialize sets. I think we can either extract it in initialize (as it is part of it) or put this method right after initialize so we don't have to search for it to understand what initialize does.

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -13,6 +13,16 @@ module GraphQL
         module UnifiedTrace
           def initialize(*args, **kwargs)
             @has_prepare_span = respond_to?(:prepare_span)
+
+            # Cache configuration values to avoid repeated lookups
+            config = Datadog.configuration.tracing[:graphql]
+            @service_name = config[:service_name]
+            @analytics_enabled = config[:analytics_enabled]
+            @analytics_sample_rate = config[:analytics_sample_rate]
+            @error_extensions_config = config[:error_extensions]
```
vpellan, line 24: Great improvement ! Less access to hash and Datadog.configuration ! :)

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -224,21 +227,43 @@ def add_query_error_events(span, errors)
...
+                  @locations_key => serialize_error_locations(graphql_error['locations']),
+                  @path_key => graphql_error['path'],
                 )
               )
             end
           end
 
+          # Do this once, to avoid rechecking configuration on every request
+          # and make the error handling code more readable.
+          def load_error_event_attributes(error_tracking)
```
vpellan: We should also add the type signature of that method to `sig/datadog/tracing/contrib/graphql/unified_trace.rbs`, as well as the new instance variables

vpellan (approved):
LGTM ! (except the type signature as Steep 1.10 will yell at you for undeclared method definition, although the file is currently ignored anyway)

docs/GettingStarted.md
marcotc: Yes, it's a much better default experience, thus why it's here. Let me clarify that.

marcotc: Ok, so it does say recommend in the docs.

lib/datadog/tracing/contrib/graphql/unified_trace.rb
marcotc: 😄

lib/datadog/tracing/contrib/graphql/unified_trace.rb
marcotc: Ok, I moved it right next to `initialize`.

lib/datadog/tracing/contrib/graphql/unified_trace.rb
marcotc: Sounds good!

PR #4870, opened 2025-08-27
AppSec: Fix `undefined method request_method=`

marcotc:
**What does this PR do?**

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/lib/datadog/appsec/api_security/route_extractor.rb
+++ b/lib/datadog/appsec/api_security/route_extractor.rb
@@ -52,8 +52,9 @@ module Datadog
           elsif request.env.key?(RAILS_ROUTE_KEY)
             request.env[RAILS_ROUTE_KEY].delete_suffix(RAILS_FORMAT_SUFFIX)
           elsif request.env.key?(RAILS_ROUTES_KEY) && !request.env.fetch(RAILS_PATH_PARAMS_KEY, {}).empty?
-            pattern = request.env[RAILS_ROUTES_KEY].router
-              .recognize(request) { |route, _| break route.path.spec.to_s }
+            route_set = request.env[RAILS_ROUTES_KEY]
+            rails_request = route_set.request_class.new(request.env)
+            pattern = route_set.router.recognize(rails_request) { |route, _| break route.path.spec.to_s }
 
             # NOTE: If rails is unable to recognize request it returns empty Array
             pattern = nil if pattern&.empty?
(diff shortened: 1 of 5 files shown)


lib/datadog/appsec/api_security/route_extractor.rb
```diff
@@ -52,8 +52,9 @@ def self.route_pattern(request)
           elsif request.env.key?(RAILS_ROUTE_KEY)
             request.env[RAILS_ROUTE_KEY].delete_suffix(RAILS_FORMAT_SUFFIX)
           elsif request.env.key?(RAILS_ROUTES_KEY) && !request.env.fetch(RAILS_PATH_PARAMS_KEY, {}).empty?
-            pattern = request.env[RAILS_ROUTES_KEY].router
-              .recognize(request) { |route, _| break route.path.spec.to_s }
+            route_set = request.env[RAILS_ROUTES_KEY]
+            rails_request = route_set.request_class.new(request.env)
```
Strech, line 56: Hey @marcotc, thanks for taking care of that. We don't have `grape` contrib, but there was a plan to start supporting it as main API framework.

The init of the router will generate new instance with unprocessed env, could you please explain what is the actual issue here of using stored router?

marcotc: The main issue is that `recognize(request)` expects a wrapped ActionPack request, not a raw Rack request, and it throws the error `undefined method request_method=` when running.

Thankfully this is easy to reproduce in tests (which are being added).

Tbh, I only really have to fix the Rails issue today, but I noticed that the Grape and Sinatra paths were not tested in integration tests (which the the only way to trigger the Rails failure today), so I started adding them in this PR as well.

Still WIP though!

Strech: 🤔 Hmmm, we have tested it with Rails and we decide to scope out Sinatra to system tests, but both working. In which case it do you get that error?

If it happens solely in tests like an old "integration" test we poses, than I would challenge the quality of the test.

We also have a test app where I run it across different Rails app versions.

Strech: BTW, [here are integrations tests](https://github.com/DataDog/dd-trace-rb/blob/master/spec/datadog/appsec/integration/api_security/schema_extraction_spec.rb) that will use route extraction (but not on all Rails versions)

If you don't mind could you share the error case you are fixing? I would love to help on this.

Strech: I think I get the issue now, problem starts not because sampling window, but because the `HEAD` request. It will take a [path into the router](https://github.com/rails/rails/blob/v6.1.7.10/actionpack/lib/action_dispatch/journey/router.rb#L115-L119) which will assign the method into it to be able to find it across the available routes.

Rails will try to [modify the request](https://github.com/rails/rails/blob/v6.1.7.10/actionpack/lib/action_dispatch/journey/router.rb#L134-L145) in order to find it.

As a suggestion, to avoid this expensive path, could we wrap the request only for this specific scenario of HEAD request then?

PR #4872, opened 2025-08-28
Add support for ActiveRecord async queries

### LINKED ISSUE #3465: Active record async queries are considered separate traces

marcotc:
Fixes #3465 

Async query execution, introduced in [Rails 7.0.0](https://github.com/rails/rails/releases/tag/v7.0.0#:~:text=ActiveRecord%3A%3ARelation%23load_async) and expanded in [Rails 7.1.0](https://guides.rubyonrails.org/7_1_release_notes.html#active-record-api-for-general-async-queries), now correctly parents the async spans to the context where the query resolution happens.
This context is the place where the results from the async query were consumed. For example: `async_relation = Model.load_async` schedules the async query, but the query span is only attached to the active trace at `relation.load` time). This only affects parenting, as the execution time of the async span is always accurate.

The active trace, started in a thread, will now see spans from a background async thread attached to it.

**Change log entry**
Yes. Add support for ActiveRecord async queries

**Additional Notes:**

**How to test the change?**


ivoanjo:
Hey, is this one still relevant?

PR #4874, opened 2025-08-29
Add WaterDrop integration (i.e.: trace messages produced by Karafka)

### LINKED ISSUE #4868: [BUG]: Karafka integration doesn't inject trace digest to produced messages
### Tracer Version(s)

2.19.0

### Ruby Version(s)

3.4.1

### Relevent Library and Version(s)

karafka 2.4.18, 2.5.0

### Bug Report

As Karafka is now supported by the Datadog tracer with a `distributed_tracing` configuration, I'd expect produced messages to automatically include the current trace's digest, so the current trace is automatically continued when a Karafka consumer consumes it.

Note that it is possible to add a simple Waterdrop middleware to inject the headers (as I did test myself):
```ruby
module KarafkaDdMiddleware
  def self.call(message)
    dd_trace = Datadog::Tracing.active_trace
    message[:headers] ||= {}
    Datadog::Tracing::Contrib::Karafka.inject(dd_trace, message[:headers]) if dd_trace
    message
  end
end

Karafka.producer.middleware.append(KarafkaDdMiddleware)
```

### Reproduction Code

initialize project:
```bash
rails new testing_karafka_datadog
bundle add datadog karafka
```

Datadog config:
```ruby
Datadog.configure do |c|
  c.tracing.instrument :karafka, distributed_tracing: true, enabled: true
end
```

Karafka config and consumer (barely changed after running `karafka install`):
```ruby
# karafka.rb (without any manually added middleware to inject the trace on the message)
class KarafkaApp < Karafka::App
  setup do |config|
    config.kafka = { 'bootstrap.servers': '127.0.0.1:9092' }
    config.client_id = 'YOUR_APP_NAME'

    config.group_id = 'YOUR_APP_NAME_consumer'
    config.consumer_persistence = !Rails.env.development?
  end

  routes.draw do
    topic :example do
      consumer ExampleConsumer
    end
  end
end

# app/consumers/example_consumer.rb
class ExampleConsumer < ApplicationConsumer
  def consume
    messages.each do |message|
      Datadog::Tracing.trace("consumer") do
        puts "[#{self.class.name}] current trace_id: #{Datadog::Tracing.active_trace&.id || 'not set'}"
        puts "[#{self.class.name}] current span_id: #{Datadog::Tracing.active_span&.id || 'not set'}"
        puts message.payload
      end
    end
  end
end
```

Producing a message (with karafka server open in another terminal):
```ruby
# in rails console
Datadog::Tracing.trace("producer") do
  puts "[producing] current trace_id: #{Datadog::Tracing.active_trace&.id || 'not set'}"
  puts "[producing] current span_id: #{Datadog::Tracing.active_span&.id || 'not set'}"

  Karafka.producer.produce_sync(topic: 'example', payload: { 'ping' => 'pong' }.to_json)
end
```

Outputs (note that the 'producing' trace is not continued in the consumer):
```
[producing] current trace_id: 139135027431419632075626282499067703165
[producing] current span_id: 1760224571734279276
[ExampleConsumer - before #each] current trace_id: 139135027431419632072840599894897151586
[ExampleConsumer - before #each] current span_id: 1385937625620920418
[ExampleConsumer - inside #each] current trace_id: 139135027431419632072840599894897151586
[ExampleConsumer - inside #each] current span_id: 2562426227694958412
{"ping" => "pong"}
[ExampleConsumer - after #each] current trace_id: 139135027431419632072840599894897151586
[ExampleConsumer - after #each] current span_id: 1385937625620920418
```

### Configuration Block

_No response_

### Error Logs

_No response_

### Operating System

_No response_

### How does Datadog help you?

_No response_

Drowze:
**What does this PR do?**

Adds support for WaterDrop (i.e. Karafka producers). Fixes #4868

**Motivation:**
We use Datadog and Karafka. For a long time we've been using a custom patch to have some distributed tracing support, but since @nvh0412 got the ball rolling by [adding traces for Karafka consumers](https://github.com/DataDog/dd-trace-rb/pull/4147), I've decided to continue the work there and also add proper tracing support for Karafka producers.
Hope we can get an ever-improving good support for Karafka in Datadog! 😄 

_(Thanks @mensfeld for the amazing Kafka framework!)_

**Change log entry**

**Additional Notes:**
Although WaterDrop is part of the Karafka framework, it can be used in isolation (which is handy if the application doesn't need any message consumer).

**How to test the change?**
See:
* [Karafka's getting started](https://karafka.io/docs/Getting-Started/) (`Karafka.producer` is a WaterDrop instance)
* [WaterDrop's getting started](https://karafka.io/docs/WaterDrop-Getting-Started/) (for using WaterDrop without Karafka)


docs/GettingStarted.md
```diff
@@ -1158,10 +1158,32 @@ end
...
+| --------------------- | -------------------------- | ------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------- |
+| `enabled`             | `DD_TRACE_KARAFKA_ENABLED` | `Bool` | Specifies whether the integration should create spans.                                                                                                                       | `true`  |
+| `distributed_tracing` |                            | `Bool` | Enables [distributed tracing](#distributed-tracing) (when iterating through the kafka messages, each of the messages' traces will be resumed for the duration of the block). | `false` |
+
+### WaterDrop
+
+The WaterDrop integration provides tracing of the `waterdrop` gem (a dependency of `karafka`, but also can be used standalone).
+
+This integration is automatically activated with the Karafka framework. If you're not using Karafka (e.g.: your app _produce_ messages for a different app _consume_), enable it through `Datadog.configure`:
```
estherk15: I wasn't sure if this is what you meant of _produce_ is a feature of Karafka.

```suggestion
This integration activates automatically with the Karafka framework. If your application doesn’t use Karafka (for example, if it only produces messages consumed by another app), enable it manually with `Datadog.configure`:
```

estherk15 (approved):

docs/GettingStarted.md
Drowze: You got it right. And your version reads much better than mine, thanks for the suggestion 😄  (just committed it)

lib/datadog/tracing/contrib/waterdrop/monitor.rb
```diff
@@ -0,0 +1,73 @@
...
+
+            Datadog::Tracing.trace(Ext::SPAN_PRODUCER) do |span, trace|
+              action = nil
+              trace_digest = trace.to_digest
+
+              if payload.key?(:messages)
+                action = event_id.sub('messages.produced', 'produce_many')
+
+                topics = payload[:messages].map { |m| m[:topic] }.uniq
+                span.set_tag(Contrib::Ext::Messaging::TAG_DESTINATION, topics)
```
Drowze: A bit unsure about `TAG_DESTINATION`... this tag can either be an array of kafka _topics_ (when `produce_many_...` is called) or a simple string (when generating a single message at a time)

Same goes for the `TAG_PARTITION` - can be either an array or a number, depending on whether we call batch produce methods or not

lib/datadog/tracing/contrib/waterdrop/monitor.rb
```diff
@@ -0,0 +1,73 @@
...
+              super
+            end
+          end
+
+          private
+
+          def inject(trace_digest, message)
+            message[:headers] ||= {}
+            WaterDrop.inject(trace_digest, message[:headers])
+          end
```
Drowze: I am a bit on the fence about using a custom the custom monitor to inject the trace context 🤔 
WaterDrop producers _do_ have a `middleware` interface... They just don't have a way to automatically add some middleware to all new instances.

I suppose an alternative approach might be patching the `WaterDrop::Producer#setup` method, so a middleware is automatically added (if it wasn't already) right after configuring a producer. E.g.:
```ruby
module Datadog
  module Tracing
    module Contrib
      module WaterDrop
        class Middleware
          def self.call(message)
            message[:headers] ||= {}
            WaterDrop.inject(Tracing.active_trace&.to_digest, message[:headers])
            message
          end
        end

        module ProducerPatch
          def setup(&block)
            super.tap do
              current_middlewares = middleware.instance_variable_get(:@steps)
              if current_middlewares.none? { |middleware| middleware == Middleware }
                middleware.append(Middleware)
              end
            end
          end
        end

        module Patcher
          def patch
            # ...
            ::WaterDrop::Producer.prepend(ProducerPatch)
          end
        end
      end
    end
  end
end
```

lib/datadog/tracing/contrib/waterdrop/patcher.rb
```diff
@@ -0,0 +1,30 @@
...
+          module_function
+
+          def target_version
+            Integration.version
+          end
+
+          def patch
+            require_relative 'monitor'
+
+            ::WaterDrop::Instrumentation::Monitor.prepend(Monitor)
```
Drowze: One thing to keep in mind: the WaterDrop monitor is configurable (just like Karafka's). That means that this would bypass our monitor patch:

```ruby
::WaterDrop::Producer.new do |c|
  c.kafka = { "bootstrap.servers": "127.0.0.1:9092" }
  c.monitor = ActiveSupport::Notifications
end
```

As this is also the case in Karafka, the karafka integration suffers from the same problem (see [here](https://karafka.io/docs/Monitoring-and-Logging/#opentelemetry))

Perhaps one approach that we can take is patching the `WaterDrop::Producer#setup` method, so we replace the monitor right after configuring a producer (or the `Karafka::App`) (and also we could trigger a warning if the user had changed the monitor from the default one)

marcotc:
Thank you, @Drowze! We'll take a look in the next few days.

mensfeld:
I think we oculd tackle this by adding a process wide monitor that would react on builder events.

mensfeld:
Yes I will expose porper instrumentation events on the WaterDrop layer for you to use.

Something like:

```ruby

WaterDrop.instrumentation.subscribe('producer.created') do |event|
  event.producer.middleware...
end
```

mensfeld:
> They just don't have a way to automatically add some middleware to all new instances.

tadam https://github.com/karafka/waterdrop/pull/680

Drowze:
Thanks a lot for this change @mensfeld 😄 I'll make use of this new notification asap!

In order to support versions prior to your PR though I suppose we'll end up with something like this (as we've discussed already - but just to keep Datadog folks on the same page):
```ruby
class Integration
  MINIMUM_CLASS_LEVEL_INSTRUMENTATION_VERSION = Gem::Version.new('2.8.8')

  def self.supports_class_level_instrumentation?
    version >= MINIMUM_CLASS_LEVEL_INSTRUMENTATION_VERSION
  end
end

module Patcher
  def patch
    if Integration.supports_class_level_instrumentation?
      require_relative 'middleware'
      ::WaterDrop.instrumentation.subscribe('producer.configured') do |event|
        producer = event[:producer]
        producer.config.middleware.append(Middleware)
      end
    else
      require_relative 'monitor'
      ::WaterDrop::Instrumentation::Monitor.prepend(Monitor)
    end
  end
end
```

By the way I suppose we can do something similar for the Karafka integration (but I'll leave this for separate PR):
```ruby
def patch
  ::Karafka.monitor.subscribe("app.initialized") do
    if Karafka::App.config.monitor.is_a?(Module) # e.g.: ActiveSupport::Notifications or any other class/module
      Karafka::App.config.monitor.prepend(Monitor)
    else # e.g.: a monitor instance from karafka-core
      Karafka::App.config.monitor.singleton_class.prepend(Monitor)
    end
  end
end
```

mensfeld:
Np. I should craft the releases with this in the upcoming days. Need to make sure docs are aligned before that.

marcotc:
👋 @Drowze, with the super fast improvements that @mensfeld did (thank you so much!), do your use cases need to instrument older versions of WaterDrop?
In other words, is it reasonable to assume that upgrading WaterDrop is not a meaningful hurdle? I see that WaterDrop v2 has been around since 2021, so I'd assume the vast majority of users are on the v2 series, thus the upgrade would be non-breaking.

I'm trying to save your time, and keep only the nice new instrumentation, using the newly exposed API, if possible.

Drowze:
Hey @marcotc! Thanks for reaching :)

While I agree it'd be nice to just stick with the new API (which was not even released yet), I'm not sure about breaking compatibility with all past Waterdrop versions 😕 

One thing to keep in mind: the Karafka ecosystem does **not** follow strict semantic versioning. Instead it usually employs breaking change in minor versions and reserve major versions major architectural changes. For context, see the [Karafka docs on Versions Lifecycle and EOL](https://karafka.io/docs/Versions-Lifecycle-and-EOL/)

Personally I don't see a good reason for not supporting any Waterdrop version before @mensfeld's changes.... I imagine other Karafka users might get confused and frustrated when e.g.: they try to use dd-trace-rb's `karafka` integration just to find out that consumers tracing works but producer tracing doesn't because they are using a month old gem versions.
Also I'd like to note that dd-trace-rb supports `karafka` v2.3.0+, which was released on Jan/2024, so supporting only the latest `waterdrop` version is a bit of weird decision when we consider _"Karafka"_ as a framework.

While I'd prefer to keep the support of karafka/waterdrop versions more or less concise, I understand if Datadog team would prefer to only support those versions with a more supported API to instrument on. Please let me know your thoughts and I'll update the PR 😃 !


That being said... Out of curiosity I just compiled the waterdrop requirement on all Karafka releases (see below). Nothing too interesting there, but might be a good reference.


<details>
<summary>Waterdrop requirement on all karafka git tags</summary>

```fish
# script used for compiling this list (using fish)
for karafka_version in (git tag | sort -V)
    echo "$karafka_version" | string match -rq '(?<major>\d+)\.(?<minor>\d+)\.(?<revision>\d+)(?:\.(?<prerelease>[0-9a-zA-Z]+))?'
    set -l waterdrop_version (git show "$karafka_version:karafka.gemspec" | grep 'waterdrop' | xargs)
    echo "$karafka_version | $waterdrop_version"
end
```
```
0.5.0.3 | waterdrop=spec.add_dependency waterdrop, ~> 0.3.2.4
1.0.1 | waterdrop=spec.add_dependency waterdrop, ~> 0.4
1.2.11 | waterdrop=spec.add_dependency waterdrop, ~> 1.2.4
1.2.12 | waterdrop=spec.add_dependency waterdrop, ~> 1.2.4
1.2.13 | waterdrop=spec.add_dependency waterdrop, ~> 1.2.4
v1.0.1 | waterdrop=spec.add_dependency waterdrop, ~> 1.4.0
v1.3.0 | waterdrop=spec.add_dependency waterdrop, ~> 1.3.0
v1.3.1 | waterdrop=spec.add_dependency waterdrop, ~> 1.3.0
v1.4.0 | waterdrop=spec.add_dependency waterdrop, ~> 1.4.0
v1.4.1 | waterdrop=spec.add_dependency waterdrop, ~> 1.4.0
v1.4.2 | waterdrop=spec.add_dependency waterdrop, ~> 1.4.0
v1.4.3 | waterdrop=spec.add_dependency waterdrop, ~> 1.4.0
v1.4.4 | waterdrop=spec.add_dependency waterdrop, ~> 1.4.0
v1.4.5 | waterdrop=spec.add_dependency waterdrop, ~> 1.4.0
v1.4.6 | waterdrop=spec.add_dependency waterdrop, ~> 1.4.0
v1.4.7 | waterdrop=spec.add_dependency waterdrop, ~> 1.4.0
v1.4.8 | waterdrop=spec.add_dependency waterdrop, ~> 1.4.0
v1.4.9 | waterdrop=spec.add_dependency waterdrop, ~> 1.4.0
v1.4.10 | waterdrop=spec.add_dependency waterdrop, ~> 1.4
v1.4.11 | waterdrop=spec.add_dependency waterdrop, ~> 1.4
v1.4.12 | waterdrop=spec.add_dependency waterdrop, ~> 1.4
v1.4.13 | waterdrop=spec.add_dependency waterdrop, ~> 1.4
v1.4.14 | waterdrop=spec.add_dependency waterdrop, ~> 1.4
v2.0.0 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.0.alpha1 | waterdrop=spec.add_dependency waterdrop, >= 2.1.0, < 3.0.0
v2.0.0.alpha2 | waterdrop=spec.add_dependency waterdrop, >= 2.2.0, < 3.0.0
v2.0.0.alpha3 | waterdrop=spec.add_dependency waterdrop, >= 2.2.0, < 3.0.0
v2.0.0.alpha4 | waterdrop=spec.add_dependency waterdrop, >= 2.2.0, < 3.0.0
v2.0.0.alpha5 | waterdrop=spec.add_dependency waterdrop, >= 2.3.0, < 3.0.0
v2.0.0.alpha6 | waterdrop=spec.add_dependency waterdrop, >= 2.3.0, < 3.0.0
v2.0.0.beta1 | waterdrop=spec.add_dependency waterdrop, >= 2.3.0, < 3.0.0
v2.0.0.beta2 | waterdrop=spec.add_dependency waterdrop, ~> 1.4
v2.0.0.beta3 | waterdrop=spec.add_dependency waterdrop, >= 2.3.0, < 3.0.0
v2.0.0.beta4 | waterdrop=spec.add_dependency waterdrop, >= 2.3.1, < 3.0.0
v2.0.0.beta5 | waterdrop=spec.add_dependency waterdrop, >= 2.3.1, < 3.0.0
v2.0.0.rc1 | waterdrop=spec.add_dependency waterdrop, >= 2.3.1, < 3.0.0
v2.0.0.rc2 | waterdrop=spec.add_dependency waterdrop, >= 2.3.3, < 3.0.0
v2.0.0.rc3 | waterdrop=spec.add_dependency waterdrop, >= 2.3.3, < 3.0.0
v2.0.0.rc4 | waterdrop=spec.add_dependency waterdrop, >= 2.4.0, < 3.0.0
v2.0.0.rc5 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.1 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.2 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.3 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.4 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.5 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.6 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.7 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.8 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.9 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.10 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.11 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.12 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.13 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.14 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.15 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.16 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.17 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.18 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.19 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.20 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.21 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.22 | waterdrop=spec.add_dependency waterdrop, >= 2.4.1, < 3.0.0
v2.0.23 | waterdrop=spec.add_dependency waterdrop, >= 2.4.3, < 3.0.0
v2.0.24 | waterdrop=spec.add_dependency waterdrop, >= 2.4.7, < 3.0.0
v2.0.25 | waterdrop=spec.add_dependency waterdrop, >= 2.4.7, < 3.0.0
v2.0.26 | waterdrop=spec.add_dependency waterdrop, >= 2.4.7, < 3.0.0
v2.0.27 | waterdrop=spec.add_dependency waterdrop, >= 2.4.9, < 3.0.0
v2.0.28 | waterdrop=spec.add_dependency waterdrop, >= 2.4.9, < 3.0.0
v2.0.29 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.0.30 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.0.31 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.0.32 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.0.33 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.0.34 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.0.35 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.0.36 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.0.37 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.0.38 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.0.39 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.0.40 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.0.41 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.1.0 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.1.1 | waterdrop=spec.add_dependency waterdrop, >= 2.4.10, < 3.0.0
v2.1.2 | waterdrop=spec.add_dependency waterdrop, >= 2.5.3, < 3.0.0
v2.1.3 | waterdrop=spec.add_dependency waterdrop, >= 2.5.3, < 3.0.0
v2.1.4 | waterdrop=spec.add_dependency waterdrop, >= 2.5.3, < 3.0.0
v2.1.5 | waterdrop=spec.add_dependency waterdrop, >= 2.6.1, < 3.0.0
v2.1.5.beta1 | waterdrop=spec.add_dependency waterdrop, >= 2.6.1.beta1, < 3.0.0
v2.1.6 | waterdrop=spec.add_dependency waterdrop, >= 2.6.2, < 3.0.0
v2.1.7 | waterdrop=spec.add_dependency waterdrop, >= 2.6.2, < 3.0.0
v2.1.8 | waterdrop=spec.add_dependency waterdrop, >= 2.6.2, < 3.0.0
v2.1.9 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.1.10 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.1.11 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.1.12 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.1.13 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.2.0 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.2.1 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.2.2 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.2.3 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.2.4 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.2.5 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.2.6 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.2.7 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.2.8 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.2.8.beta1 | waterdrop=spec.add_dependency waterdrop, >= 2.6.6, < 3.0.0
v2.2.9 | waterdrop=spec.add_dependency waterdrop, >= 2.6.10, < 3.0.0
v2.2.10 | waterdrop=spec.add_dependency waterdrop, >= 2.6.10, < 3.0.0
v2.2.11 | waterdrop=spec.add_dependency waterdrop, >= 2.6.10, < 3.0.0
v2.2.12 | waterdrop=spec.add_dependency waterdrop, >= 2.6.11, < 3.0.0
v2.2.13 | waterdrop=spec.add_dependency waterdrop, >= 2.6.11, < 3.0.0
v2.2.14 | waterdrop=spec.add_dependency waterdrop, >= 2.6.11, < 3.0.0
v2.3.0 | waterdrop=spec.add_dependency waterdrop, >= 2.6.12, < 3.0.0
v2.3.0.alpha1 | waterdrop=spec.add_dependency waterdrop, >= 2.6.12, < 3.0.0
v2.3.0.alpha2 | waterdrop=spec.add_dependency waterdrop, >= 2.6.12, < 3.0.0
v2.3.0.rc1 | waterdrop=spec.add_dependency waterdrop, >= 2.6.12, < 3.0.0
v2.3.1 | waterdrop=spec.add_dependency waterdrop, >= 2.6.12, < 3.0.0
v2.3.2 | waterdrop=spec.add_dependency waterdrop, >= 2.6.12, < 3.0.0
v2.3.3 | waterdrop=spec.add_dependency waterdrop, >= 2.6.12, < 3.0.0
v2.3.4 | waterdrop=spec.add_dependency waterdrop, >= 2.6.12, < 3.0.0
v2.4.0 | waterdrop=spec.add_dependency waterdrop, >= 2.7.0, < 3.0.0
v2.4.0.beta1 | waterdrop=spec.add_dependency waterdrop, >= 2.7.0.rc1, < 3.0.0
v2.4.0.beta2 | waterdrop=spec.add_dependency waterdrop, >= 2.7.0.rc1, < 3.0.0
v2.4.0.rc1 | waterdrop=spec.add_dependency waterdrop, >= 2.7.0.rc2, < 3.0.0
v2.4.1 | waterdrop=spec.add_dependency waterdrop, >= 2.7.0, < 3.0.0
v2.4.2 | waterdrop=spec.add_dependency waterdrop, >= 2.7.0, < 3.0.0
v2.4.3 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.4 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.5 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.6 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.7 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.8 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.9 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.10 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.11 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.12 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.13 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.14 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.15 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.16 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.17 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.4.18 | waterdrop=spec.add_dependency waterdrop, >= 2.7.3, < 3.0.0
v2.5.0 | waterdrop=spec.add_dependency waterdrop, >= 2.8.3, < 3.0.0
v2.5.0.beta1 | waterdrop=spec.add_dependency waterdrop, >= 2.8.3, < 3.0.0
v2.5.0.beta2 | waterdrop=spec.add_dependency waterdrop, >= 2.8.3, < 3.0.0
v2.5.0.rc1 | waterdrop=spec.add_dependency waterdrop, >= 2.8.3, < 3.0.0
v2.5.0.rc2 | waterdrop=spec.add_dependency waterdrop, >= 2.8.3, < 3.0.0
v2.5.1.beta1 | waterdrop=spec.add_dependency waterdrop, >= 2.8.3, < 3.0.0
```
</details>

mensfeld:
> While I agree it'd be nice to just stick with the new API (which was not even released yet), I'm not sure about breaking compatibility with all past Waterdrop versions 

This change (if not used) is backwards compatible. WaterDrop had no breaking changes for years aside of one super edge case in transactions and one deprecated setting early 2024.

> In other words, is it reasonable to assume that upgrading WaterDrop is not a meaningful hurdle?

Yes. While Karafka **can** be complex, WaterDrop upgrades are not.

I bump dependency of waterdrop version in karafka framework mostly to support weird edge cases and sync APIs. If you are not using Karafka, WaterDrop standalone is in my opinion low effort tool with extremely low (so far) maintenance burden.

New version with the introduced changes to support this will be released at most next week as I am traveling at the moment.

mensfeld:
@Drowze I will release rc1 for you today of waterdrop for testing.

vpellan:
[Launching CI](https://github.com/DataDog/dd-trace-rb/pull/4994)

appraisal/ruby-3.0.rb
```diff
@@ -97,12 +97,10 @@
 build_coverage_matrix('rest-client')
 build_coverage_matrix('mongo', min: '2.1.0')
 build_coverage_matrix('dalli', [2])
+build_coverage_matrix('karafka', min: '2.3.0')
+build_coverage_matrix('waterdrop', min: '2.8.8.rc1')
```
vpellan: `waterdrop` 2.8.8.rc1 specifies that it only supports Ruby 3.1 and above, which makes the CI fails for this file.

lib/datadog/tracing/contrib/waterdrop/ext.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Tracing
+    module Contrib
+      module WaterDrop
+        module Ext
+          ENV_ENABLED = 'DD_TRACE_WATERDROP_ENABLED'
```
vpellan, line 8: Please add this environment variable in `supported-configurations.json` (with `"version": ["A"]`), then run `bundle exec rake local_config_map:generate` or else the environment variable will not be used by the tracer

appraisal/ruby-3.0.rb
```diff
@@ -97,12 +97,10 @@
 build_coverage_matrix('rest-client')
 build_coverage_matrix('mongo', min: '2.1.0')
 build_coverage_matrix('dalli', [2])
+build_coverage_matrix('karafka', min: '2.3.0')
```
vpellan, line 100: `karafka` 2.5.1 requires Ruby 3.2 or above, so we cannot have a 'latest' appraisal, only the min one. Looking at the karafka changelog, it seems that 2.5.1 was not the first version to not support Ruby 3.0 through transitive dependencies, although the gemspec said 3.0 was compatible, which is probably why there was a `karafka-min` appraise lower.

This should still be okay for Ruby 3.1 as long as we don't update the lockfiles, we can do that on our end.

appraisal/ruby-3.1.rb
```diff
@@ -98,6 +98,7 @@
 build_coverage_matrix('mongo', min: '2.1.0')
 build_coverage_matrix('dalli', [2])
 build_coverage_matrix('karafka', min: '2.3.0')
+build_coverage_matrix('waterdrop', min: '2.8.8.rc1')
```
vpellan, line 101: Please run `bundle exec ruby appraisal/generate.rb && bundle exec rake dependency:lock['/app/gemfiles/ruby_3.*_waterdrop_*.gemfile']` for each supported Rubies so that the gemfiles for waterdrop are generated and used in the CI.

Matrixfile
```diff
@@ -126,6 +126,7 @@
     'karafka-latest' => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby',
     # karafka 2.0.41 contains broken dependency of karafka-core 2.2.4, which depends on Ruby (>= 2.6.0)
     'karafka-min'    => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby',
+    'waterdrop'      => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby'
```
vpellan: Using the `build_coverage_matrix` results in two appraisals: `waterdrop-min` and `waterdrop-latest`. Also, the resulting appraisals will not contain karafka, so these should be in their own hash or it will try to run karafka tests and fail. You'd also need to add it to the Rakefile in the list of integrations
```suggestion
    'karafka-min'    => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby'
  },
  'waterdrop' => {
    'waterdrop-min'      => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ❌ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby',
    'waterdrop-latest' => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ❌ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby'
```

vpellan (commented):
Hello ! Thank you very much for your contribution ! I gave it a first pass and left a bunch of suggestions, feel free to pick them up but if you prefer, we can handle them !

*Not in comments: to fix lint errors, you can run `bundle exec rake standard:fix`
To generate prototype signature files, you can use `bundle exec rake rbs:prototype`
If at all possible, please try to change `untyped` items to the correct types! See the following guide: docs/StaticTypingGuide.md*

ericfirth:
Eric from the Data Streams Monitoring team here, I am going to begin working on getting this PR to pass CI. We (the DSM team) would love to support the full karafka ecosystem with DSM and this PR would be helpful for that, in that it would be silly to support Waterdrop only for DSM purposes and not traces as well

mensfeld:
@ericfirth ping me in case you need any help (Karafka ecosystem author here)

Matrixfile
```diff
@@ -127,6 +127,11 @@
     # karafka 2.0.41 contains broken dependency of karafka-core 2.2.4, which depends on Ruby (>= 2.6.0)
     'karafka-min'    => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby',
   },
+  'waterdrop' => {
+    # WaterDrop 2.8.8+ requires Ruby >= 3.1.0 (JRuby 9.4+ only, not 9.2/9.3)
+    'waterdrop-latest' => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ❌ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby-9.4+',
+    'waterdrop-min'    => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ❌ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby-9.4+'
```
vpellan: I think you don't need to do that. There should be an automatic mapping between JRuby versions and Ruby version. Because Ruby 2.5 and 2.6 are disabled JRuby 9.2 and 9.3 will not run.
See https://github.com/DataDog/dd-trace-rb/blob/e4ed4400eb7dd156c0dd2fad222714cd9944d589/tasks/runtime_matcher.rb#L7

vpellan:
```suggestion
    'waterdrop-latest' => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ❌ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby',
    'waterdrop-min'    => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ❌ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby'
```

Steepfile
```diff
@@ -330,9 +330,12 @@ target :datadog do
   ignore 'lib/datadog/tracing/contrib/karafka/configuration/settings.rb'
   ignore 'lib/datadog/tracing/contrib/karafka/distributed/propagation.rb'
   ignore 'lib/datadog/tracing/contrib/karafka/ext.rb'
+  ignore 'lib/datadog/tracing/contrib/karafka/framework.rb'
   ignore 'lib/datadog/tracing/contrib/karafka/integration.rb'
   ignore 'lib/datadog/tracing/contrib/karafka/monitor.rb'
   ignore 'lib/datadog/tracing/contrib/karafka/patcher.rb'
+  ignore 'lib/datadog/tracing/contrib/waterdrop.rb'
+  ignore 'lib/datadog/tracing/contrib/waterdrop/'
```
vpellan: We should not add any files here, this is the reason we extracted individual files from paths. I believe it is better to see untyped signatures (although it would be even better to have no untyped at all in the .sig files) than ignoring entire files or folders. You seem to have taken a bit of time to type some of your additions, thank you very much ! It would be better to not ignore these improvements !
```suggestion
ignore 'lib/datadog/tracing/contrib/karafka/integration.rb'
ignore 'lib/datadog/tracing/contrib/karafka/monitor.rb'
ignore 'lib/datadog/tracing/contrib/karafka/patcher.rb'
```

vpellan (approved):
LGTM! (Left two suggestions, I think the Matrixfile one should be applied before merging)

Matrixfile
ericfirth: fixed

Steepfile
ericfirth: ok. i added the only files that weren't able to be type checked because of metaprogramming and typed the rest

PR #4877, opened 2025-09-03
[APMAPI-1578] Update runtime-id in memfd file after forks

vpellan:
**What does this PR do?**

This PR updates runtime-id in memfd file after a fork

**Motivation:**

Currently the runtime-id is still set to the main process, which is wrong.

**Change log entry**

Yes. Fix Service Discovery capabilities on forked processes.

**Additional Notes:**

**How to test the change?**

This PR includes a test ran by the CI.


lib/datadog/core/process_discovery.rb
```diff
@@ -31,6 +39,19 @@ def self.get_and_store_metadata(settings, logger)
           service_version: settings.version || ''
         }
       end
+
+      private_class_method def self.apply_at_fork_patch
```
Strech, line 43: I think it's better to name it in natural order `apply_patch_at_fork`

lib/datadog/core/process_discovery.rb
```diff
@@ -31,6 +39,19 @@ def self.get_and_store_metadata(settings, logger)
...
+        Utils::AtForkMonkeyPatch.apply!
+        Utils::AtForkMonkeyPatch.at_fork(:child) do
+          next if (components = Datadog.send(:components, allow_initialization: false)).nil?
+
+          metadata = get_metadata(Datadog.configuration)
+          logger = components.logger
+          memfd = _native_store_tracer_metadata(logger, **metadata)
+          memfd.logger = logger if memfd
+          components.instance_variable_set(:@process_discovery_fd, memfd)
```
Strech: Very small suggestion over formatting and spacing


```suggestion
          logger = components.logger
          metadata = get_metadata(Datadog.configuration)
          
          memfd = _native_store_tracer_metadata(logger, **metadata)
          memfd.logger = logger if memfd

          components.instance_variable_set(:@process_discovery_fd, memfd)
```

spec/datadog/core/process_discovery_spec.rb
```diff
@@ -85,4 +85,34 @@
...
+    end
+
+    it 'updates the process discovery file descriptor' do
+      expect_in_fork do
+        native_fd = Datadog.send(:components, allow_initialization: false).process_discovery_fd
+        fd = described_class._native_to_rb_int(native_fd)
+        buffer = IO.new(fd)
+        buffer.rewind
+        raw_content = buffer.read
+        content = MessagePack.unpack(raw_content)
```
Strech:
```suggestion
        content = MessagePack.unpack(buffer.read)
```

vpellan: Thanks, applied in [c35dacc](https://github.com/DataDog/dd-trace-rb/pull/4877/commits/c35dacca52a909c3fcb29bbaac804fca04cd7229)

lib/datadog/core/process_discovery.rb
vpellan: Applied in [c35dacc](https://github.com/DataDog/dd-trace-rb/pull/4877/commits/c35dacca52a909c3fcb29bbaac804fca04cd7229) !

lib/datadog/core/process_discovery.rb
```diff
@@ -31,6 +41,21 @@ def self.get_and_store_metadata(settings, logger)
...
+          next if (components = Datadog.send(:components, allow_initialization: false)).nil?
+
+          metadata = get_metadata(Datadog.configuration)
+          logger = components.logger
+
+          memfd = _native_store_tracer_metadata(logger, **metadata)
+          memfd.logger = logger if memfd
+
+          components.instance_variable_set(:@process_discovery_fd, memfd)
+        end
```
ivoanjo: Should we somehow close the previous one? (Are we "leaking" the previous memfd?)

ivoanjo: (Replying to myself -- I think we may not be leaking per se, but on a quick glance the old one will possibly remain alive until GC runs which may not be so soon so I'm not sure if that's the correct semantics? -- May be worth checking)

Strech: Good point

vpellan: Changed ProcessDiscovery to a singleton that will always close the fd before creating a new one: 995d2cf31b7c0a4a4ceb715e363d68359c6e1f55 (I also moved it to startup! instead of initialize as it was causing issues with replace_components! in [9f8326b](https://github.com/DataDog/dd-trace-rb/pull/4877/commits/9f8326bd8d3fa635243ba94595bb0372ea1d4449) )

lib/datadog/core/configuration/components.rb
```diff
@@ -155,6 +154,8 @@ def startup!(settings, old_state: nil)
             remote&.start
           end
 
+          Core::ProcessDiscovery.get_and_store_metadata(settings, @logger)
+
           Core::Diagnostics::EnvironmentLogger.collect_and_log!(@environment_logger_extra)
```
ivoanjo: So I think I understand why this was moved here away from the `initialize`.

I suspect it's because the order of execution of code in **reconfiguration** is
`initialize` -> `shutdown!` -> `startup!`; that any a global/singleton set up we do in `initialize` would be undone by `shutdown!`.

Did I get that right? If so, I think we should be more deliberate about documenting this. E.g. probably both here as well as in `components_spec` we should leave a comment explaining that because this is a singleton, it needs to be started after shutdown runs otherwise it would be undone.

(Otherwise this doesn't look like a deliberate decision and someone might mistakenly move it to `initialize` and update the unit test to match while missing why that is not correct)

lib/datadog/core/process_discovery.rb
```diff
@@ -2,34 +2,62 @@
 
 require 'datadog/core/process_discovery/tracer_memfd'
 
+require_relative 'utils/at_fork_monkey_patch'
+require_relative 'utils/only_once'
+
 module Datadog
   module Core
     # Class used to store tracer metadata in a native file descriptor.
     class ProcessDiscovery
```
ivoanjo: Minor: Since we never create instances of `ProcessDiscovery`, this should probably become a `module`

```suggestion
    module ProcessDiscovery
```

lib/datadog/core/process_discovery.rb
```diff
@@ -2,34 +2,62 @@
...
-      def self.get_and_store_metadata(settings, logger)
-        if (libdatadog_api_failure = Datadog::Core::LIBDATADOG_API_FAILURE)
-          logger.debug("Cannot enable process discovery: #{libdatadog_api_failure}")
-          return
+      ONLY_ONCE = Core::Utils::OnlyOnce.new
+
+      class << self
+        def get_and_store_metadata(settings, logger)
```
ivoanjo: Minor: With the evolution of the code, this method has become a bit misnamed (doesn't get anything, and even "store" is a bit of a weird way of putting it). Perhaps just call it `publish`?

lib/datadog/core/process_discovery.rb
```diff
@@ -2,34 +2,62 @@
...
+        def apply_at_fork_patch
+          Utils::AtForkMonkeyPatch.apply!
+          Utils::AtForkMonkeyPatch.at_fork(:child) do
+            next if (components = Datadog.send(:components, allow_initialization: false)).nil?
+
+            logger = components.logger
+            get_and_store_metadata(Datadog.configuration, logger)
+          end
+        end
       end
```
ivoanjo: A bit like my comment on the `components.rb`, the code is correct but I don't think other than looking at the git commit history, that it betrays _intent_ -- that is neither this code nor the tests scream "THIS IS BEING DONE BECAUSE THE RUNTIME-ID CHANGES AFTER A FORK".

It's important for the code to scream these things when they're not obvious as otherwise folks need to try to reverse-engineer why something was left the way it was, and often it's hard to know why.

As a comparison, observe how the `crashtracking/component.rb` is doing it: https://github.com/DataDog/dd-trace-rb/blob/e74332805306eccf16b2f95f7eb0e47f1acc2309/lib/datadog/core/crashtracking/component.rb#L71-L75

and then the specs reference back to this "refreshing of tags" again:

https://github.com/DataDog/dd-trace-rb/blob/e74332805306eccf16b2f95f7eb0e47f1acc2309/spec/datadog/core/crashtracking/component_spec.rb#L152-L163

lib/datadog/core/process_discovery.rb
```diff
@@ -2,34 +2,62 @@
...
+          }
+        end
+
+        def apply_at_fork_patch
+          Utils::AtForkMonkeyPatch.apply!
+          Utils::AtForkMonkeyPatch.at_fork(:child) do
+            next if (components = Datadog.send(:components, allow_initialization: false)).nil?
+
+            logger = components.logger
+            get_and_store_metadata(Datadog.configuration, logger)
```
ivoanjo: Minor: `Datadog.logger` already does the `components(allow_initialization: false)` internally, so I believe this can be simplified to `get_and_store_metadata(Datadog.configuration, Datadog.logger)`.

spec/datadog/core/process_discovery_spec.rb
```diff
@@ -26,14 +32,14 @@
...
+          native_fd = described_class.instance_variable_get(:@file_descriptor)
 
           # Extract content from created memfd
           fd = described_class._native_to_rb_int(native_fd)
-          buffer = IO.new(fd)
+          buffer = IO.new(fd, autoclose: false)
           buffer.rewind
-          raw_content = buffer.read
-          content = MessagePack.unpack(raw_content)
+          content = MessagePack.unpack(buffer.read)
```
ivoanjo: Minor: The topic of "how much to duplicate or not" in specs is an interesting one ;) 

Having said that, I do think this is the kind of thing that would benefit from being extracted, as it adds boilerplate for all testcases, it's always the same, and it even has nontrivial bits (e.g. using `autoclose: false`).

By "extract", it can be as simple as `def extract_content` inside this describe (that's allowed) or `let(:content) { ... }`.

spec/datadog/core/process_discovery_spec.rb
```diff
@@ -26,14 +32,14 @@
...
 
         it 'stores metadata successfully' do
-          native_fd = described_class.get_and_store_metadata(settings, Datadog::Core::Logger.new(StringIO.new))
+          described_class.get_and_store_metadata(settings, Datadog::Core::Logger.new(StringIO.new))
+          native_fd = described_class.instance_variable_get(:@file_descriptor)
 
           # Extract content from created memfd
           fd = described_class._native_to_rb_int(native_fd)
-          buffer = IO.new(fd)
+          buffer = IO.new(fd, autoclose: false)
```
ivoanjo: Since the `autoclose: false` is both:
* **really important to not make our test-suite flaky**

as well as

* **extremely non-obvious**

it's one more thing that's useful to leave behind a comment saying why we're doing this, e.g. `# This MUST be autoclose: false because our code handles closing separately and if Ruby also tries to close the wrong file descriptor can be closed, potentially creating flaky specs and other bad things"

spec/datadog/core/process_discovery_spec.rb
```diff
@@ -85,4 +91,34 @@
...
+          {
+            'schema_version' => 1,
+            'runtime_id' => Datadog::Core::Environment::Identity.id,
+            'tracer_language' => Datadog::Core::Environment::Identity.lang,
+            'tracer_version' => Datadog::Core::Environment::Identity.gem_datadog_version_semver2,
+            'hostname' => Datadog::Core::Environment::Socket.hostname,
+            'service_name' => 'test-service'
+          }
+        )
+      end
```
ivoanjo: This spec isn't very good: It passes if `apply_at_fork_patch` + the onlyonce get removed.

That's because it's testing the same codepath as above "hey if I call things manually they get called" but the codepath we need to check is "if there's a fork, do things get called automatically AND with the updated values.

Testing the atfork stuff is somewhat annoying. See https://github.com/DataDog/dd-trace-rb/pull/4204 for one example of how I tested it for crashtracking.

ivoanjo (commented):
Left a few notes! In particular a big common undercurrent is "we're doing a bunch of things for very specific reasons, and it's very important to leave behind a trail so that they stick out like a sore thumb".

These are the kinds of comments that are really really important, much more than the 

```
# Do something
do_something()
```

That AI likes to add ;)

spec/datadog/core/process_discovery_spec.rb
vpellan: Added a check for ProcessDiscovery.publish in c8bc30dd90d1d0fc63717c39295ec73e85a83966

spec/datadog/core/process_discovery_spec.rb
vpellan: Thanks, I added a comment in c8bc30dd90d1d0fc63717c39295ec73e85a83966

spec/datadog/core/process_discovery_spec.rb
vpellan: I've put it in let(:content) as its also used in the fork part in c8bc30dd90d1d0fc63717c39295ec73e85a83966

lib/datadog/core/process_discovery.rb
vpellan: Thanks, removed in c8bc30dd90d1d0fc63717c39295ec73e85a83966

lib/datadog/core/process_discovery.rb
vpellan: I added a comment in c8bc30dd90d1d0fc63717c39295ec73e85a83966

lib/datadog/core/process_discovery.rb
vpellan: Renamed to publish in c8bc30dd90d1d0fc63717c39295ec73e85a83966 !

lib/datadog/core/process_discovery.rb
vpellan: Thanks, changed to a module in c8bc30dd90d1d0fc63717c39295ec73e85a83966

lib/datadog/core/configuration/components.rb
vpellan: Yes it was because of that exact reason. I added a comment in c8bc30dd90d1d0fc63717c39295ec73e85a83966

lib/datadog/core/process_discovery.rb
```diff
@@ -2,34 +2,58 @@
...
-          tracer_version: Core::Environment::Identity.gem_datadog_version_semver2,
-          hostname: Core::Environment::Socket.hostname,
-          service_name: settings.service || '',
-          service_env: settings.env || '',
-          service_version: settings.version || ''
-        }
+        def shutdown!
+          @file_descriptor&.shutdown!
+          @file_descriptor = nil
+        end
```
ivoanjo: Tbh after the latest changes in the code, I think we could simplify a bit the whole "save the logger code for later" and just take `Datadog.logger` as an argument on `shutdown!` (not a blocker for this PR though!).

spec/datadog/core/process_discovery_spec.rb
```diff
@@ -85,4 +90,30 @@
...
```
ivoanjo: Minor: Maybe throw in an extra test that the runtime-id changes...? Arguably this is testing a bit runtime-id behavior, not this class but... I think it does document a bit more our intentions. (It's also ok to keep as-is, hence it being minor)

```suggestion
    it 'updates the process discovery file descriptor' do
      allow(described_class).to receive(:publish).and_call_original
      
      parent_runtime_id = Datadog::Core::Environment::Identity.id

      expect_in_fork do
        expect(described_class).to have_received(:publish)
        expect(content).to eq(
          {
            'schema_version' => 1,
            'runtime_id' => Datadog::Core::Environment::Identity.id,
            'tracer_language' => Datadog::Core::Environment::Identity.lang,
            'tracer_version' => Datadog::Core::Environment::Identity.gem_datadog_version_semver2,
            'hostname' => Datadog::Core::Environment::Socket.hostname,
            'service_name' => 'test-service'
          }
        )
        expect(context.fetch('runtime_id')).to_not eq(parent_runtime_id)
      end
```

vpellan: Added in [c7e7082](https://github.com/DataDog/dd-trace-rb/pull/4877/commits/c7e7082a809e2e21979bc804b28b0bf67969d61b)

lib/datadog/core/process_discovery.rb
vpellan: Added in [c7e7082](https://github.com/DataDog/dd-trace-rb/pull/4877/commits/c7e7082a809e2e21979bc804b28b0bf67969d61b) !

PR #4884, opened 2025-09-09
Dynamic Instrumentation: fix gsub usage in evaluate_template

p-datadog:
**What does this PR do?**

Changes `gsub` usage to block form from string second argument, which performs unwanted interpolation.

**Motivation:**

`gsub` with string arguments still runs them through the regular expression engine and interpolates backslashes and some special characters in the replacement string. The desired behavior was a simple string replacement.

**Change log entry**
Yes: DI: fix incorrect template expression evaluation in some cases

**Additional Notes:**

The existing code did not grant access to anything that shouldn't have been accessible (i.e. the #{} syntax). It would replace double backslashes with single backslashes, escapes like `\n` with the corresponding literals like the newline, and include parts of the string into the string via `\\``. 

Additionally, the UI does not currently permit actually specifying message templates for customers.

**How to test the change?**
Unit tests added

### DIFF
--- a/spec/datadog/di/probe_notification_builder_spec.rb
+++ b/spec/datadog/di/probe_notification_builder_spec.rb
@@ -326,4 +326,23 @@ RSpec.describe Datadog::DI::ProbeNotificationBuilder do
       end
     end
   end
+
+  describe '#evaluate_template' do
+    context 'when there are variables to be substituted' do
+      let(:template) { "{@hello} {@world}" }
+      let(:vars) do
+        {
+          'hello' => 'test',
+          # We need double backslash to check for proper sub/gsub usage.
+          'world' => "\"'\\\\a\#{value}",
+        }
+      end
+
+      let(:expected) { "test \"'\\\\a\#{value}" }
+
+      it 'substitutes correctly' do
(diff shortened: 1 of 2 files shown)


spec/datadog/di/probe_notification_builder_spec.rb
```diff
@@ -326,4 +326,23 @@
...
   end
+
+  describe '#evaluate_template' do
+    context 'when there are variables to be substituted' do
+      let(:template) { "{@hello} {@world}" }
+      let(:vars) do
+        {
+          hello: 'test',
+          # We need double backslash to check for proper sub/gsub usage.
+          world: "\"'\\\\a\#{value}",
```
Strech: Maybe this would help with some extra unneeded backslashes?

```suggestion
          world: %("'\\\\a\#{value}),
```

spec/datadog/di/probe_notification_builder_spec.rb
```diff
@@ -326,4 +326,23 @@
...
+        }
+      end
+
+      let(:expected) { "test \"'\\\\a\#{value}" }
```
Strech: Same could be done here

```suggestion
      let(:expected) { %(test "'\\\\a\#{value}) }
```

PR #4888, opened 2025-09-11
Update version for env vars containing OTEL aliases

vpellan:
**What does this PR do?**

Change config versions to version B for DD config that have OTEL aliases

**Motivation:**

As we use some OTEL env vars as aliases of existing DD env vars, we needed to create a new version of these configs on the Feature Parity Dashboard, which is version 'B', as other tracers consider them two different env vars with different behaviour.

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**


y9v:
@vpellan just curious: why are we versioning those config variables, and why are we using letters for it?

vpellan:
@y9v A config can have multiple implementations, so this is a way to tell the FPD that we use a specific implementation. We use letters because using semver could be confusing, and we could confuse them with the tracer version. In that case I changed them to another implementation that have OTEL_ env vars as aliases. Other tracers don't do that and implement these OTEL_ env vars as separate config. Adding them as aliases to the "A" version is wrong for other tracers.

PR #4889, opened 2025-09-11
DI: add exception serialization tests

p-datadog:
**What does this PR do?**

Adds unit tests for serializing exception classes.

**Motivation:**
Exception capture would include the exceptions into dynamic instrumentation payloads, yet there was no existing test coverage for how they would be serialized.

**Change log entry**
None

**Additional Notes:**
For safety reasons we do NOT serialize exception messages - I don't see how to do so while guaranteeing that we do not call customer code. It is expected that custom exceptions may have their `message` method overwritten.

Unfortunately this means only the exception class would be available in dynamic instrumentation along with any instance variables set by the custom exception classes.

**How to test the change?**

### DIFF
--- a/spec/datadog/di/serializer_spec.rb
+++ b/spec/datadog/di/serializer_spec.rb
@@ -105,20 +105,26 @@ RSpec.describe Datadog::DI::Serializer do
       # the lack of serialization of their messages (because we cannot do
       # so safely - guaranteeing not to invoke customer code).
       {name: 'Exception instance', input: IOError.new('test error'),
-      expected: {type: 'IOError', fields: {}}},
+       expected: {type: 'IOError', fields: {}}},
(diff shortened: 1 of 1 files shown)


spec/datadog/di/serializer_spec.rb
```diff
@@ -75,6 +100,31 @@ class DISerializerCustomExceptionTestClass; end
        expected: {type: 'Date', value: '2020-01-02'}},
       {name: 'DateTime value', input: DateTime.new(2020, 1, 2, 3, 4, 5),
        expected: {type: 'DateTime', value: '2020-01-02T03:04:05+00:00'}},
+
+      # Exception classes do not have a dedicated serialier, but document
```
ivoanjo: Minor:
```suggestion
      # Exception classes do not have a dedicated serializer, but document
```

p-datadog:
Thanks for this, I made a note to add a C extension as a follow-up project.

PR #4895, opened 2025-09-16
[NO-TICKET] Add typing stats in PR comment

vpellan:
**What does this PR do?**

It adds typing stats in comments of PRs

**Motivation:**

We're using Steep but the migration started two years ago and kinda stalled. This will show contributors what the current state of typing in dd-trace-rb is, and the plan for the future is to track the progress of typing, e.g. in a Datadog dashboard.

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

This should add a comment on this PR

### DIFF
--- a/spec/datadog/di/serializer_spec.rb
+++ b/spec/datadog/di/serializer_spec.rb
@@ -30,7 +30,32 @@ end
 # Should have no instance variables
 class DISerializerSpecTestClass; end
 
-class DISerializerCustomExceptionTestClass; end
+class DISerializerCustomExceptionTestClass < StandardError; end
+
+class DISerializerExceptionWithFieldsTestClass < StandardError
+  def initialize(message)
+    super
+    @test_field = 'bar'
+  end
+end
+
+class DISerializerExceptionWithMessageFieldTestClass < StandardError
+  def initialize(message)
+    super
+    @message = 'bar'
+  end
+end
+
+class DISerializerExceptionWithMessageRaiseTestClass < StandardError
+  def initialize(message)
+    super
+    @message = 'bar'
+  end
+
+  def message
+    raise 'uh oh'
+  end
+end
 
 RSpec.describe Datadog::DI::Serializer do
   di_test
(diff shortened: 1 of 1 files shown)


vpellan:
(sorry I missclicked on close)

Strech:
👏🏼 awesome work! I have some concerns about partial non/partial typing with complex types

```
@store: Hash[untyped, untyped]
@data: Array[untyped]
```

so this is oftentimes impossible to type without defining interface (maybe) even then it will be to tedious work ...

vpellan:
> I have some concerns about partial non/partial typing with complex types

As we can add comments on .rbs files, maybe we could define an inline comment like `# untyped:accept` or something similar to basically ignore that line ?

Strech:
Also [Prism](https://github.com/ruby/prism) is used by steep itself, don't we want to use it too instead of regexp?

lloeki:
Nitpick: would it be possible to have the lists of files be in code fences?

e.g instead of:

- lib/datadog/appsec.rb
- lib/datadog/appsec/component.rb
- lib/datadog/appsec/configuration/settings.rb
- ...

have:

```lib/datadog/appsec.rb
lib/datadog/appsec/component.rb
lib/datadog/appsec/configuration/settings.rb
...
```

or (but I like it less):

- `lib/datadog/appsec.rb`
- `lib/datadog/appsec/component.rb`
- `lib/datadog/appsec/configuration/settings.rb`

or maybe even:

- [`lib/datadog/appsec.rb`](https://github.com/DataDog/dd-trace-rb/blob/6940a05913ece5df9cc4138dbdb2f5ca6a47bae0/lib/datadog/appsec.rb)
- [`lib/datadog/appsec/component.rb`](https://github.com/DataDog/dd-trace-rb/blob/6940a05913ece5df9cc4138dbdb2f5ca6a47bae0/lib/datadog/appsec/component.rb)
- [`lib/datadog/appsec/configuration/settings.rb`](https://github.com/DataDog/dd-trace-rb/blob/6940a05913ece5df9cc4138dbdb2f5ca6a47bae0/lib/datadog/appsec/configuration/settings.rb)

vpellan:
> or maybe even:
> 
> [lib/datadog/appsec.rb](https://github.com/DataDog/dd-trace-rb/blob/6940a05913ece5df9cc4138dbdb2f5ca6a47bae0/lib/datadog/appsec.rb)
[lib/datadog/appsec/component.rb](https://github.com/DataDog/dd-trace-rb/blob/6940a05913ece5df9cc4138dbdb2f5ca6a47bae0/lib/datadog/appsec/component.rb)
[lib/datadog/appsec/configuration/settings.rb](https://github.com/DataDog/dd-trace-rb/blob/6940a05913ece5df9cc4138dbdb2f5ca6a47bae0/lib/datadog/appsec/configuration/settings.rb)

I tried to add links but this would make the comment go over the 65k characters limit. I will try to add them in code blocks !

lloeki:
> so this is oftentimes impossible to type without defining interface (maybe) even then it will be to tedious work ...

I understand,  it's tedious but it's also the best signal we have. Maybe not a full interface but an opportunity to type alias?

> As we can add comments on .rbs files, maybe we could define an inline comment like # untyped:accept or something similar to basically ignore that line ?

I think that's one of the ways Sorbet addressed that... to very detrimental effect: basically by ignoring untyped it silences the untypedness and by way of transitivity makes it viral in a kind of broken window effect.

I don't have a proper solution :/ Maybe we can just try as is and see how it turns out to be in practice?

vpellan:
> Also [Prism](https://github.com/ruby/prism?rgh-link-date=2025-09-17T12%3A31%3A19Z) is used by steep itself, don't we want to use it too instead of regexp?

As discussed privately, work is currently ongoing to use RBS AST instead of Regexp. This is the first iteration of that feature and it will later contains use an AST, and run itself on the diff instead of the whole codebase.

PR #4899, opened 2025-09-17
Define a method for DI expression language evaluation

p-datadog:
**What does this PR do?**

Initial expression language implementation translated the expression AST to Ruby code as a string, and evaluated that string via `eval` each time the expression needed to be evaluated.

This PR moves the Ruby code into a method that is defined during probe installation, such that during expression evaluation (when customer code runs) there is no need to parse the code every time.

**Motivation:**

Performance improvement for expression language evaluation.

**Change log entry**
None: this is a customer-visible change in the sense that performance of our product improves, however expression language hasn't yet made it into a dd-trace-rb release and therefore there is no reference point for customers to compare performance to, yet.

**Additional Notes:**
I took the expression language integration tests and modified them to evaluate each expression 10,000 times and add up the runtime across all tests. With the change in this PR the runtime improved by 15%.

This PR removes the `compiled_expr` assertions from expression language integration tests. I haven't deleted the corresponding entries from YAML files yet in case they would be useful later, but most likely the compiled form will be opaque from now on.

**How to test the change?**
Existing unit tests


lib/datadog/di/el/expression.rb
```diff
@@ -11,18 +11,23 @@ def initialize(compiled_expr)
...
-        def ==(other)
-          other.is_a?(self.class) &&
-            compiled_expr == other.compiled_expr
+          cls = Class.new(Evaluator)
+          cls.class_exec do
+            eval(<<-RUBY, Object.new.send(:binding), __FILE__, __LINE__ + 1) # standard:disable Security/Eval
+              def evaluate(context)
+                @context = context
+                #{compiled_expr}
+              end
```
Strech, line 21: Just to be sure, is this expression already sanitized and prepared? Thinking about potential injection or anything non-eval safe here.

In addition the question - is eval really needed? Could it be replaced with `define_method`?

```ruby
cls = Class.new(Evaluator)
cls.define_method(:evaluate) do |context|
  @context = context
  compiled_expr
end
```

Strech (approved):
I have a suggestion to replace the `eval` with `define_method`

lib/datadog/di/el/expression.rb
p-datadog: The sanitization is done in the first PR (https://github.com/DataDog/dd-trace-rb/pull/4861).

I translated the "compiled" form to DSL/`define_method` style as an experiment. Unfortunately it makes the evaluation 65% slower. The DSL style produces more Procs and while they can possibly be reduced/eliminated, that would require a LOT more code to be written.

The best approach probably is to generate the VM instructions directly, as other languages are doing, but this is also a larger undertaking (that I do plan on attempting again).

PR #4900, opened 2025-09-17
DI: Fix line instrumentation when target line raises an exception

p-datadog:
**What does this PR do?**

Makes it so that when an exception is raised on a line which is instrumented via  a line probe, line execution is only reported once.

**Motivation:**

To permit instrumenting `end` lines of methods, DI installs trace points for `line` and `end` events. I learned that when a line raises an exception, this causes the trace point to be invoked twice: one time before the line executes with the `line` event, and again after the exception is raised with the `end` event (since technically the method exits - prematurely - due to the exception).

This PR adds a flag to probe to track whether a line trace point has already been invoked with a `line` event. If this happened, we would ignore a subsequent invocation with a different event type (such as `end`).

**Change log entry**
Yes: correctly produce a single log event when line probe is instrumenting a line that raises an exception

**Additional Notes:**

**How to test the change?**
Tests are added

### DIFF
--- a/spec/datadog/di/hook_line_load.rb
+++ b/spec/datadog/di/hook_line_load.rb
@@ -30,7 +30,7 @@ class HookLineIvarLoadTestClass
   def test_exception
     local = 42
     raise TestException, 'Intentional exception'       # Line 32
-    local               # Line 33
+    local               # Line 33 # standard:ignore Lint/UnreachableCode
   end
 end
 
--- a/spec/datadog/di/hook_line_load.rb
+++ b/spec/datadog/di/hook_line_load.rb
@@ -30,7 +30,7 @@ class HookLineIvarLoadTestClass
   def test_exception
     local = 42
     raise TestException, 'Intentional exception'       # Line 32
-    local               # Line 33 # standard:ignore Lint/UnreachableCode
+    local               # Line 33 # standard:disable Lint/UnreachableCode
   end
 end
 
--- a/spec/datadog/di/hook_line_load.rb
+++ b/spec/datadog/di/hook_line_load.rb
@@ -1,9 +1,9 @@
 # Comment line - not executable
 
 class HookLineLoadTestClass
-  def test_method       # Line 2
-    42                  # Line 3
-  end
+  def test_method       # Line 4
+    42                  # Line 5
+  end                   # Line 6
 
   def test_method_with_local
     local = 42 # standard:disable Style/RedundantAssignment
@@ -16,6 +16,9 @@ class HookLineLoadTestClass
 end
 
 class HookLineIvarLoadTestClass
+  class TestException < StandardError
+  end
+
   def initialize
     @ivar = 42
   end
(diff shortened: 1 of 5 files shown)


spec/datadog/di/hook_line_load.rb
```diff
@@ -16,15 +16,24 @@ def test_method_with_arg(arg)
...
+    raise TestException, 'Intentional exception'       # Line 32
+    local               # Line 33 # standard:disable Lint/UnreachableCode
+  end
 end
 
-unless (actual = File.read(__FILE__).count("\n")) == 30
-  raise "Wrong number of lines in hook_line_load.rb: actual #{actual}, expected 30"
+unless (actual = File.read(__FILE__).count("\n")) == 39
+  raise "Wrong number of lines in hook_line_load.rb: actual #{actual}, expected 39"
 end
```
marcotc, line 39:
```suggestion
actual = IO.readlines(__FILE__).size
raise "Wrong number of lines in hook_line_load.rb: actual #{actual}, expected #{__LINE__}" if actual != __LINE__
```
Just a minor suggestion, feel free to not adapt if it doesn't make sense to you.

spec/datadog/di/instrumenter_spec.rb
```diff
@@ -1208,6 +1288,39 @@
...
+        load File.join(File.dirname(__FILE__), 'hook_line_load.rb')
+      end
+
+      let(:probe) do
+        Datadog::DI::Probe.new(file: 'hook_line_load.rb', line_no: 32,
+          id: 1, type: :log, rate_limit: rate_limit)
+      end
+
+      let(:payload) do
+        expect_any_instance_of(TracePoint).to receive(:enable).and_call_original
```
marcotc, line 1304: I noticed that all the tests introduced in this PR pass even if `expect_any_instance_of(TracePoint).to receive(:enable).and_call_original`. Is asserting that `TracePoint.enable` is called, with the original implementation important here? In other words, is this assertion not already capture by the other assertions, like `observed_calls.length`?

For context, I normally see `expect_any_instance_of(TracePoint).to receive(:enable).and_call_original` being used to restore the original behaviour after mocking a function. I'm just wondering why it is important here.

spec/datadog/di/hook_line_load.rb
p-datadog: The idea here is to prevent unintentional modification of the file, since tests set probes on specific line numbers. I think your suggestion makes this logic a no-op as no matter how many lines are in the file, the assertion will pass.

spec/datadog/di/instrumenter_spec.rb
p-datadog: Some of the other tests assert on arguments to `enable`. This is to verify that the trace point is targeted, or not.

When there are no argument assertions, the idea is to verify that the instrumentation that produces observed calls was installed by the test in question. `observed_calls` should be test-specific, therefore in theory the assertions that don't require specific arguments to `enable` shouldn't affect the test, but I sleep better at night knowing each test verifies that instrumentation is running for that test.

spec/datadog/di/hook_line_load.rb
marcotc: Makes sense!

PR #4902, opened 2025-09-22
[APMAPI-1607] Upgrade libdatadog dependency to version 21.0

vpellan:
Check out the
https://github.com/DataDog/dd-trace-rb/blob/master/docs/DevelopmentGuide.md
for guidance on how to set up your development environment,
run the test suite, write new integrations, and more.
-->

**What does this PR do?**

Updates the native bits in the gem (profiling, crashtracking) to be able to work with the latest libdatadog 18.

**Motivation:**

DSM needs libdatadog 21, and I had to update the native bits of stable config

**Change log entry**

Yes. Upgrade libdatadog dependency to 21.0

**Additional Notes:**

(I still need to run the CI task to update the lock files)

**How to test the change?**

CI


p-datadog:
> Updates the native bits in the gem (profiling, crashtracking) to be able to work with the latest libdatadog 18.


Did you mean libdatadog 21?

vpellan:
> Did you mean libdatadog 21?

Yes I didn't caught that, thanks !

.gitlab/install_datadog_deps.rb
```diff
@@ -125,25 +124,114 @@
...
+  *cached_gems,
+  versioned_path.join("gems/ffi-#{ffi_version}/ext"),
+  libdatadog_musl,
+  # Remove additional libdatadog files to reduce package size
+  versioned_path.join("gems/libdatadog-#{libdatadog_version}-#{RUBY_PLATFORM}/vendor/libdatadog-#{libdatadog_version.split('.').take(3).join('.')}/#{RUBY_PLATFORM}-musl"),
+  versioned_path.join("gems/libdatadog-#{libdatadog_version}-#{RUBY_PLATFORM}/vendor/libdatadog-#{libdatadog_version.split('.').take(3).join('.')}/#{RUBY_PLATFORM}-gnu"),
+  # Remove debug symbols and unnecessary files from libdatadog
+  versioned_path.join("gems/libdatadog-#{libdatadog_version}-#{RUBY_PLATFORM}/vendor/libdatadog-#{libdatadog_version.split('.').take(3).join('.')}/include"),
+  versioned_path.join("gems/libdatadog-#{libdatadog_version}-#{RUBY_PLATFORM}/vendor/libdatadog-#{libdatadog_version.split('.').take(3).join('.')}/src"),
+]
```
ivoanjo: I don't quite understand this change: looking at the gemfile contents, right now there's no `/src` directory and there's no `-gnu` directory either... :thinking: 

Also, the `-musl` directory is already removed by `libdatadog_musl` :eyes:.

Do you mind double-checking if this change is really what's intended?

.gitlab/install_datadog_deps.rb
```diff
@@ -125,25 +124,114 @@
...
+  if File.exist?(file_path)
+    puts "Removing: #{file_path}"
+    FileUtils.rm_r(file_path, verbose: true)
+  end
+end
+
+# Remove object files and static libraries using glob patterns
+libdatadog_base = versioned_path.join("gems/libdatadog-#{libdatadog_version}-#{RUBY_PLATFORM}/vendor/libdatadog-#{libdatadog_version.split('.').take(3).join('.')}")
+Dir.glob(libdatadog_base.join('*.o')).each { |f| FileUtils.rm_f(f) if File.exist?(f) }
+Dir.glob(libdatadog_base.join('*.a')).each { |f| FileUtils.rm_f(f) if File.exist?(f) }
```
ivoanjo: I.... don't think this is correct either: the libdatadog release packages should definitely not include `.o` or `.a` files. ...? :thinking:

.gitlab/install_datadog_deps.rb
```diff
@@ -125,25 +124,114 @@
...
+  if File.exist?(dir_path)
+    puts "Removing: #{dir_path}"
+    FileUtils.rm_rf(dir_path)
+  end
+end
+
+# Remove any remaining object files and static libraries
+Dir.glob(versioned_path.join('**/*.o')).each { |f| FileUtils.rm_f(f) if File.exist?(f) }
+Dir.glob(versioned_path.join('**/*.a')).each { |f| FileUtils.rm_f(f) if File.exist?(f) }
+Dir.glob(versioned_path.join('**/*.la')).each { |f| FileUtils.rm_f(f) if File.exist?(f) }
```
ivoanjo: I... is this AI? This doesn't make any sense either. None of these folders are in the released gems from rubygems.org. What am I missing?

.gitlab/install_datadog_deps.rb
```diff
@@ -125,25 +124,114 @@
...
   # Symlink those directories to be utilized by Ruby compiled with shared libraries
   FileUtils.ln_sf Gem.extension_api_version, ruby_api_version
 end
+
+# Strip debug symbols from shared libraries to reduce package size
+shared_libraries = Dir.glob(versioned_path.join('**/*.so'))
+shared_libraries.each do |lib|
+  puts "Stripping debug symbols from #{lib}"
+  system("strip --strip-debug #{lib}") if File.exist?(lib)
+end
```
ivoanjo: Doing this makes it harder to debug issues with the crashtracker; is this really required? If possible, I think it's desirable to keep the currently-available information.

.gitlab/install_datadog_deps.rb
```diff
@@ -125,25 +124,114 @@
...
+Dir.glob(versioned_path.join('**/*.zip')).each { |f| FileUtils.rm_f(f) if File.exist?(f) }
+Dir.glob(versioned_path.join('**/*.tar')).each { |f| FileUtils.rm_f(f) if File.exist?(f) }
+Dir.glob(versioned_path.join('**/*.bz2')).each { |f| FileUtils.rm_f(f) if File.exist?(f) }
+Dir.glob(versioned_path.join('**/*.xz')).each { |f| FileUtils.rm_f(f) if File.exist?(f) }
+
+# Remove any remaining documentation and metadata files
+Dir.glob(versioned_path.join('**/*.md')).each { |f| FileUtils.rm_f(f) if File.exist?(f) }
+Dir.glob(versioned_path.join('**/*.txt')).each { |f| FileUtils.rm_f(f) if File.exist?(f) && !f.include?('requirements') }
+Dir.glob(versioned_path.join('**/*.yml')).each { |f| FileUtils.rm_f(f) if File.exist?(f) }
+Dir.glob(versioned_path.join('**/*.yaml')).each { |f| FileUtils.rm_f(f) if File.exist?(f) }
```
ivoanjo: My notes above apply here as well. This doesn't look correct.

datadog.gemspec
```diff
@@ -69,7 +69,7 @@ Gem::Specification.new do |spec|
 
   # When updating the version here, please also update the version in `libdatadog_extconf_helpers.rb`
   # (and yes we have a test for it)
-  spec.add_dependency 'libdatadog', '~> 18.1.0.1.0'
+  spec.add_dependency 'libdatadog', '~> 21.0.0.1.0'
```
ivoanjo: I've spotted that the [`21.0.0.1.0` release on rubygems](https://rubygems.org/gems/libdatadog) is almost 10x as big as it should be due to some changes in the libdatadog packages.

We'll need to spin a `21.0.0.2.0` (or v22, if that shows up faster) before we can release this to customers. (That said, we that shouldn't stop us from getting this PR in good shape and green, because it means later we only really need to bump the version).

ext/libdatadog_api/library_config.c
```diff
@@ -98,27 +99,33 @@ static VALUE _native_configurator_get(VALUE self) {
   ddog_Configurator *configurator;
   TypedData_Get_Struct(self, ddog_Configurator, &configurator_typed_data, configurator);
 
-  ddog_Result_VecLibraryConfig configurator_result = ddog_library_configurator_get(configurator);
+  // Wrapping config_logged_result into a Ruby object enables the Ruby GC to manage its memory
+  // We need to allocate memory for config_logged_result because once it is out of scope, it will be freed (at the end of this function)
+  // So we cannot reference it with &config_logged_result
+  // We are doing this in case one of the ruby API raises an exception before the end of this function,
+  // so the allocated memory will still be freed
+  ddog_LibraryConfigLoggedResult *configurator_logged_result = (ddog_LibraryConfigLoggedResult *)ruby_xmalloc(sizeof(ddog_LibraryConfigLoggedResult));
```
ivoanjo: Minor: As more of a consistency thing, I suggest always using calloc (zero memory) instead of malloc (memory can contain garbage) unless we have a really good performance-related reason not to.

It's true that in its current state the code doesn't need this -- the result will be replaced in the next line, but in the future we may refactor this and trigger the footgun...

```suggestion
  ddog_LibraryConfigLoggedResult *configurator_logged_result = ruby_xcalloc(1, sizeof(ddog_LibraryConfigLoggedResult));
```

ext/libdatadog_api/library_config.c
```diff
@@ -98,27 +99,33 @@ static VALUE _native_configurator_get(VALUE self) {
...
+  *configurator_logged_result = ddog_library_configurator_get(configurator);
+  VALUE config_logged_result_rb = TypedData_Wrap_Struct(config_logged_result_class, &config_logged_result_typed_data, configurator_logged_result);
 
-  if (configurator_result.tag == DDOG_RESULT_VEC_LIBRARY_CONFIG_ERR_VEC_LIBRARY_CONFIG) {
-    ddog_Error err = configurator_result.err;
+  if (configurator_logged_result->tag == DDOG_LIBRARY_CONFIG_LOGGED_RESULT_ERR) {
+    ddog_Error err = configurator_logged_result->err;
```
ivoanjo, line 112: I'm curious -- why this change to move the `TypedData_Wrap_Struct` up? I believe the previous structure was correct as well so I'm not sure what this change is trying to prevent that wasn't prevented before.

lib/datadog/core/configuration.rb
```diff
@@ -172,8 +172,10 @@ def logger
         end
       end
 
-      def config_init_logger
-        configuration? ? logger : logger_without_configuration
+      # During stable config initialization, we also have to check for DD_TRACE_DEBUG in local and fleet config.
+      # We pass the debug source value
+      def config_init_logger(debug_source_value = DATADOG_ENV[Ext::Diagnostics::ENV_DEBUG_ENABLED])
+        configuration? ? logger : logger_without_configuration(debug_source_value)
       end
```
ivoanjo: Hmmm... I think having this default may not work as intended, or at least it's a bit of a footgun.

Specifically, since the value of `debug_source_value` gets cached inside `@temp_config_logger`, if this method gets called with the default of only looking at `DD_TRACE_DEBUG`, it really doesn't matter if later it gets called with any other argument (or vice-versa).

I wonder if perhaps a simpler option would be, since the logs get stored inside `Datadog::Core::Configuration::StableConfig`, perhaps delaying printing them so that we can use the regular `Datadog.logger`, which itself should (I believe) already respect the stable config results...?

spec/datadog/core/configuration/stable_config_spec.rb
```diff
@@ -90,12 +95,84 @@
...
 
         it 'returns the configuration' do
-          expect(described_class.configuration).to eq(
+          expect(described_class.configuration).to include(
             {
               local: {config: {"DD_LOGS_INJECTION" => "false"}},
               fleet: {config: {"DD_APPSEC_ENABLED" => "true"}}
             }
           )
+          expect(described_class.configuration[:logs]).to be_a(String)
```
ivoanjo: Minor: As more of a rspec tip, these two can actually be combined, e.g.

```ruby
          expect(described_class.configuration).to include(
            {
              local: {config: {"DD_LOGS_INJECTION" => "false"}},
              fleet: {config: {"DD_APPSEC_ENABLED" => "true"}},
              logs: be_a(String),
            }
          )
```

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -213,19 +213,15 @@
...
-            library_name: 'dd-trace-rb',
-            library_version: Datadog::VERSION::STRING,
-            family: 'ruby',
-            tags: ['tag1:value1', 'tag2:value2', 'language:ruby-testing-123', 'service:ruby-testing-123'],
-          )
-          expect(crash_report_message[:files][:'/proc/self/maps']).to_not be_empty
-          expect(crash_report_message[:os_info]).to_not be_empty
+          expect(crash_report_message[:version]).to eq('1.0')
+          expect(crash_report_message[:kind]).to eq('Crash ping')
+          expect(crash_report_message[:crash_uuid]).to_not be_nil
```
ivoanjo: Wait -- why was this changed? Specifically, I think this test was made weaker:

* Now it doesn't check that a specific function is included in the stack trace (so the test won't fail if symbolication is incorrect, which as been in the past)
* We no longer check the size of the stack trace either
* We don't check the tags are propagated correctly

I think it's ok to add more things to the test, but I don't understand why that was made by removing other things... Did the format change again so as not to have those elements?

ivoanjo (changes requested):
This PR had a bunch of things that surprised me...

I think the changes on `install_datadog_deps.rb` might've been trying to fix the issue I pointed out with the upstream libdatadog v21 package that's on rubygems.org, but that's the wrong place to fix them -- we need to respin that package, otherwise only SSI customers get the "fix", and all other customers don't.

I suspect that once we fix the libdatadog release, most (all?) changes in `install_datadog_deps.rb` can be reverted.

Because [libdatadog v21.0.0.1.0 on rubygems](https://rubygems.org/gems/libdatadog/versions/21.0.0.1.0) is in bad shape, I'm marking this PR as "Request changes" as we mustn't merge it until the libdatadog release gets fixed. Otherwise our non-SSI customers would find the next dd-trace-rb release would suddenly be 10x as big as the previous one because of libdatadog.

There's a few more things I don't quite understand why they were changed, so I've left a few questions about them.

.gitlab/install_datadog_deps.rb
ericfirth: i think these can go away. i added this change because the size was too big and that was happening because of the tar including the tar issue you found (https://dd.slack.com/archives/C01TCF143GB/p1759335750454959)

p-datadog:
Are all of the changes in this PR required to be in one PR? https://github.com/DataDog/dd-trace-rb/pull/4902/commits/515f6256abb1a08cdae60c162c0a19780a305d0d for example looks to be independent from "updating libdatadog dependency".

p-datadog:
This PR also looks to have been incorrectly rebased.

ericfirth:
sorry about screwing up this branch. I removed all my changes and fixed the bad rebase

spec/datadog/core/crashtracking/component_spec.rb
ericfirth: i reverted this. but the format of crash tracking on libdatadog v21 is significantly different

ivoanjo: > the format of crash tracking on libdatadog v21 is significantly different

Hmmm.... It's not the first time the Ruby crashtracking specs catch regressions... I'll ask about this in #crashtracking on our slack.

ivoanjo: From discussion in the channel, it's probably https://github.com/DataDog/libdatadog/pull/1209 -- now there's supposed to be 2 reports to the webserver, and our asserts should run on the second report, not the first

gyuheon0h: The crashtracker now sends two messages, a `crash_ping` which is an almost empty telemetry message on intial crash detection, and a `crash_report` which is the original message that was being sent, which has not changed at all.

gyuheon0h: `crash_ping` should arrive first, but you can ID by checking the tag `is_crash_ping:true` for `crash_pings` and `is_crash:true` for the original `crash_report`s

ext/libdatadog_api/library_config.c
vpellan: `ddog_library_config_drop` now takes a `ddog_LibraryConfigLoggedResult` as argument. I think it was changed so that you can drop the result of `ddog_library_configurator_get` without looking if it contains the result or an error. We don't use that capability here but maybe other tracers does.

ext/libdatadog_api/library_config.c
vpellan: Applied in 722bca490f92b1a6b2b895a25a56da6d461efde6

lib/datadog/core/configuration.rb
vpellan: I moved the logging in Components initialization in 722bca490f92b1a6b2b895a25a56da6d461efde6

spec/datadog/core/configuration/stable_config_spec.rb
vpellan: Applied in 722bca490f92b1a6b2b895a25a56da6d461efde6

spec/datadog/core/configuration/stable_config_spec.rb
```diff
@@ -90,14 +103,77 @@
...
+              apm_configuration_default:
+                DD_TRACE_DEBUG: true
+            YAML
+          end
+
+          it 'prints debug logs' do
+            # Datadog.logger is not reset between tests, so we need to build a new logger
+            expect { described_class.log_result(Datadog::Core::Configuration::Components.build_logger(Datadog.configuration)) }.to output(/Reading stable configuration from files/).to_stdout
+          end
+        end
```
ivoanjo: Minor: Since we are effectively dependency injecting the logger, this could be simplified to not actually use a logger, e.g. something along the lines of:

```ruby
logger = double('test logger')
expect(logger).to receive(:debug).with(/Reading stable configuration from files/)

described_class.log_result(logger)
```

E.g. rather than building a bunch of things and using a lot of fancy `output`/`to_stdout` helpers, instead check that the right method gets called on the object that's passed in. This would avoid worrying about any state in the log or where output is going.

spec/datadog/core/configuration/stable_config_spec.rb
```diff
@@ -90,14 +103,77 @@
...
               local: {config: {"DD_LOGS_INJECTION" => "false"}},
-              fleet: {config: {"DD_APPSEC_ENABLED" => "true"}}
+              fleet: {config: {"DD_APPSEC_ENABLED" => "true"}},
+              logs: be_a(String)
             }
           )
         end
       end
+
+      context 'with DD_TRACE_DEBUG set during configuration initialization' do
```
ivoanjo: Minor: Should this section perhaps be under a `describe '#log_result'`? `log_result` seems to be the method under test here :eyes:

lib/datadog/core/process_discovery.rb
```diff
@@ -37,14 +37,16 @@ def shutdown!
...
             tracer_language: Core::Environment::Identity.lang,
             tracer_version: Core::Environment::Identity.gem_datadog_version_semver2,
             hostname: Core::Environment::Socket.hostname,
             service_name: settings.service || '',
             service_env: settings.env || '',
-            service_version: settings.version || ''
+            service_version: settings.version || '',
+            # TODO: Implement process tags and container id
+            process_tags: '',
+            container_id: ''
```
ivoanjo, line 49: What's the plan for this TODO? Curious if it it's still expected to be filled in before merging this PR or if this is for later (and why have it as a blank TODO vs not adding it until we need it)

ext/libdatadog_api/process_discovery.c
```diff
@@ -42,34 +42,40 @@ static VALUE _native_store_tracer_metadata(int argc, VALUE *argv, VALUE self) {
...
+  ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_TRACER_VERSION, StringValueCStr(tracer_version));
+  ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_HOSTNAME, StringValueCStr(hostname));
+  ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_SERVICE_NAME, StringValueCStr(service_name));
+  ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_SERVICE_ENV, StringValueCStr(service_env));
+  ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_SERVICE_VERSION, StringValueCStr(service_version));
+  ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_PROCESS_TAGS, StringValueCStr(process_tags));
+  ddog_tracer_metadata_set(builder, DDOG_METADATA_KIND_CONTAINER_ID, StringValueCStr(container_id));
+
+  ddog_Result_TracerMemfdHandle result = ddog_tracer_metadata_store(builder);
+  ddog_tracer_metadata_free(builder);
```
ivoanjo, line 78: Does `_store(...)` really not consume the `builder`? (I've seen some libdatadog APIs where that happens, may be worth doublechecking if that's the case here)

lib/datadog/core/process_discovery.rb
vpellan: This will be done in a future PR as this is a new feature (we'll have to add an env var to make it work, the default is false which returns empty process_tags and container_id just like here), and it can work without it for now. I tried to limit the changes of this PR to the ones that were required to make this update work (mostly fixing breaking changes, although i did add stable config logs, but it was a small enough change I believe)

ext/libdatadog_api/process_discovery.c
vpellan: It doesn't look like it ? https://github.com/DataDog/libdatadog/blob/d9b1999d7c5d4645c4752c3f3ffe248d0e526c6c/datadog-library-config-ffi/src/tracer_metadata.rs#L119

spec/datadog/core/configuration/stable_config_spec.rb
vpellan: Updated in [5023f8b](https://github.com/DataDog/dd-trace-rb/pull/4902/commits/5023f8b779cf13006c0dc02ea52d2479452d8988)

spec/datadog/core/configuration/stable_config_spec.rb
vpellan: Now that I look at them, I feel like these tests aren't doing much. My goal was to ensure that setting `DD_TRACE_DEBUG` in the config files would set the logger level accordingly. But this has three steps, check that `Datadog.configuration` is updated accordingly, which can be done in `#configuration` previous tests, check that `build_logger` or any new logger would have its level set accordingly which is already done in `components_spec.rb`, and check that we call logger.debug which only needs to be done once. I've simplified them in [5023f8b](https://github.com/DataDog/dd-trace-rb/pull/4902/commits/5023f8b779cf13006c0dc02ea52d2479452d8988)

ext/libdatadog_api/process_discovery.c
ivoanjo: Slightly weird design choice but sure lol xD thanks for checking ;)

vpellan: I've checked with @paullegranddc and @dmehala and it does not consume it

PR #4904, opened 2025-09-22
Add app-endpoints-loaded Telemetry event

y9v:
**What does this PR do?**
This PR adds a Telemetry event with type `app-endpoints`.

**Motivation:**
We need to report application endpoints when the app starts for AppSec API Security feature.

**Change log entry**

**Additional Notes:**
We need to send endpoints in batches to reduce memory allocation on the client. For this reason, `is_first` attribute is being set in the payload - when set to `false`, the sent data is appended to the data that was sent previously.

**How to test the change?**
CI

### DIFF
--- a/lib/datadog/core/telemetry/component.rb
+++ b/lib/datadog/core/telemetry/component.rb
@@ -19,6 +19,8 @@ module Datadog
       #
       # @api private
       class Component
+        ENDPOINT_COLLECTION_MESSAGE_LIMIT = 300
+
         attr_reader :enabled, :logger, :transport, :worker
 
         include Core::Utils::Forking
@@ -166,7 +168,7 @@ module Datadog
         end
 
         # Report application endpoints
-        def app_endpoints_loaded(endpoints, page_size: 300)
+        def app_endpoints_loaded(endpoints, page_size: ENDPOINT_COLLECTION_MESSAGE_LIMIT)
           return if !@enabled || forked?
 
           endpoints.each_slice(page_size).with_index do |endpoints_slice, i|
--- a/sig/datadog/core/telemetry/component.rbs
+++ b/sig/datadog/core/telemetry/component.rbs
@@ -2,6 +2,8 @@ module Datadog
   module Core
     module Telemetry
       class Component
+        ENDPOINT_COLLECTION_MESSAGE_LIMIT: Integer
+
         @enabled: bool
         @log_collection_enabled: bool
         @stopped: bool
@@ -32,7 +34,7 @@ module Datadog
 
         def integrations_change!: () -> void
 
-        def app_endpoints_loaded: (Array[Event::AppEndpointsLoaded::endpoint] endpoints, page_size: Integer) -> void
+        def app_endpoints_loaded: (Event::AppEndpointsLoaded::endpoints endpoints, page_size: Integer) -> void
 
         def inc: (String namespace, String metric_name, Datadog::Core::Telemetry::Metric::input_value value, ?tags: Datadog::Core::Telemetry::Metric::tags_input, ?common: bool) -> void
 
--- a/spec/datadog/core/telemetry/component_spec.rb
+++ b/spec/datadog/core/telemetry/component_spec.rb
@@ -310,31 +310,26 @@ RSpec.describe Datadog::Core::Telemetry::Component do
       let(:enabled) { true }
 
       it 'enqueues event with all endpoints when they are within page size limit' do
-        expect(Datadog::Core::Telemetry::Event::AppEndpointsLoaded).to receive(:new).with(
-          endpoints, is_first: true
-        ).and_call_original
+        expect(Datadog::Core::Telemetry::Event::AppEndpointsLoaded).to receive(:new)
+          .with(endpoints, is_first: true).and_call_original
 
         telemetry.app_endpoints_loaded(endpoints)
 
-        expect(worker).to have_received(:enqueue).with(
-          an_instance_of(Datadog::Core::Telemetry::Event::AppEndpointsLoaded)
-        ).once
+        expect(worker).to have_received(:enqueue)
+          .with(an_instance_of(Datadog::Core::Telemetry::Event::AppEndpointsLoaded)).once
       end
 
       it 'enqueues event with paginated endpoints when they are not within page size limit' do
-        expect(Datadog::Core::Telemetry::Event::AppEndpointsLoaded).to receive(:new).with(
-          endpoints[0, 1], is_first: true
-        ).and_call_original
+        expect(Datadog::Core::Telemetry::Event::AppEndpointsLoaded).to receive(:new)
+          .with(endpoints[0, 1], is_first: true).and_call_original
 
-        expect(Datadog::Core::Telemetry::Event::AppEndpointsLoaded).to receive(:new).with(
-          endpoints[1, 1], is_first: false
-        ).and_call_original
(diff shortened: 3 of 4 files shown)


lib/datadog/core/telemetry/component.rb
```diff
@@ -165,6 +165,16 @@ def client_configuration_change!(changes)
           @worker.enqueue(Event::AppClientConfigurationChange.new(changes, 'remote_config'))
         end
 
+        # Report application endpoints
+        #
+        # `is_first_event` needs to be set to `false` when endpoint data that is being sent
+        # should be appended in the backend to endpoint data that we already have sent
+        def app_endpoints_loaded(serialized_endpoints, is_first_event:)
```
p-datadog: Are you doing the batching in appsec then? I would think this is a telemetry implementation aspect?

y9v: The main reason for batching is to reduce the memory footprint, so I have a lazy enumerator in appsec that generates batches of endpoints and emits telemetry events. It is way easier and less risky than changing telemetry to pass this lazy enumerator through all the telemetry classes (or pass pages of endpoints).

lib/datadog/core/telemetry/component.rb
```diff
@@ -165,6 +165,16 @@ def client_configuration_change!(changes)
           @worker.enqueue(Event::AppClientConfigurationChange.new(changes, 'remote_config'))
         end
 
+        # Report application endpoints
+        #
+        # `is_first_event` needs to be set to `false` when endpoint data that is being sent
+        # should be appended in the backend to endpoint data that we already have sent
```
Strech: Sorry, but I have hard time understanding that comment.

lib/datadog/core/telemetry/component.rb
Strech: I would like to suggest few things:

1. Outline that batching is not happening here, and we should care about payload. More awesome will be to enforce it with type system, but it's just a crazy idea
2. I've checked other telemetry files and we use word `initial` to outline first, maybe we can use it here too (instead of `is_first_event`)? We already operating on events level, so we can drop "event" word

lib/datadog/core/telemetry/event/app_endpoints_loaded.rb
```diff
@@ -0,0 +1,30 @@
...
+
+require_relative 'base'
+
+module Datadog
+  module Core
+    module Telemetry
+      module Event
+        # Telemetry event class for sending 'app-endpoints' payload
+        class AppEndpointsLoaded < Base
+          def initialize(serialized_endpoints, is_first_event:)
```
Strech: Do you think we need that `serialized` prefix? We still don't know how they are serialized, so that doesn't help us

sig/datadog/core/telemetry/component.rbs
```diff
@@ -32,6 +32,8 @@ module Datadog
 
         def integrations_change!: () -> void
 
+        def app_endpoints_loaded: (Array[Datadog::Core::Telemetry::Event::AppEndpointsLoaded::serialized_endpoint], is_first_event: bool) -> void
```
Strech:
```suggestion
        def app_endpoints_loaded: (Array[Event::AppEndpointsLoaded::serialized_endpoint], is_first_event: bool) -> void
```

lib/datadog/core/telemetry/component.rb
y9v: rephrased in [820683b](https://github.com/DataDog/dd-trace-rb/pull/4904/commits/820683ba2c064286b26836199464223b8743db58) to:

```ruby
# `is_initial` must be set to `false`` when the endpoint data being sent
# should be merged with previously sent data on the backend,
# rather than treated as a new initial dataset.
```

lib/datadog/core/telemetry/event/app_endpoints_loaded.rb
y9v: Removed the prefix in [900c8f4](https://github.com/DataDog/dd-trace-rb/pull/4904/commits/900c8f4933679860a1828e0b61d213facea00a51). I think you are right - if we have stricter types for the endpoint hash, we don't need this prefix.

y9v:
@Strech tbh I wouldn't check the payload size in the `Telemetry#app_endpoints_loaded` method.

Reasons for that:
- as far as I understand the main aim is to reduce the memory footprint in the client, not the payload size for the server (as json payload needs to be parsed in order for elements to be counted)
- this makes the method more complicated then it should, because the payload size should be configurable, and I wouldn't spread this between the serializer and telemetry method

lib/datadog/core/telemetry/component.rb
p-datadog: I wouldn't change "first" to "initial" for two reasons:

1) "initial" event is sent when the application starts. The lifecycle is different from "first" event as part of the endpoints
2) The endpoints RFC uses "is_first" terminology, I think it would be confusing if we used "intial" to refer to the same thing"

y9v: I moved pagination to `Telemetry#app_endpoints_loaded` to avoid confusion and mistakes. I guess we can still pass a lazy enumerable to it.

y9v: Renamed `initial` back to `first` - since in our case we will really only ever use this argument for pagination.

[7b9fa40](https://github.com/DataDog/dd-trace-rb/pull/4904/commits/7b9fa40e040a64746c5a289d48f072e5eaef2f7e)

sig/datadog/core/telemetry/component.rbs
```diff
@@ -32,6 +32,8 @@ module Datadog
 
         def integrations_change!: () -> void
 
+        def app_endpoints_loaded: (Array[Event::AppEndpointsLoaded::endpoint] endpoints, page_size: Integer) -> void
```
Strech: It's better to define `Event::AppEndpointsLoaded::endpoints` to have a better control over it and simplify both RBS files

lib/datadog/core/telemetry/component.rb
```diff
@@ -165,6 +165,15 @@ def client_configuration_change!(changes)
           @worker.enqueue(Event::AppClientConfigurationChange.new(changes, 'remote_config'))
         end
 
+        # Report application endpoints
+        def app_endpoints_loaded(endpoints, page_size: 300)
```
Strech: Does it makes sense to make it a constant somewhere?

spec/datadog/core/telemetry/component_spec.rb
```diff
@@ -272,6 +272,73 @@
...
+
+        expect(worker).to have_received(:enqueue).with(
+          an_instance_of(Datadog::Core::Telemetry::Event::AppEndpointsLoaded)
+        ).once
+      end
+
+      it 'enqueues event with paginated endpoints when they are not within page size limit' do
+        expect(Datadog::Core::Telemetry::Event::AppEndpointsLoaded).to receive(:new).with(
+          endpoints[0, 1], is_first: true
+        ).and_call_original
```
Strech: Maybe this fits better (1 line less)?

```suggestion
        expect(Datadog::Core::Telemetry::Event::AppEndpointsLoaded).to receive(:new)
          .with(endpoints[0, 1], is_first: true).and_call_original
```

y9v: yes, looks better this way [5377458](https://github.com/DataDog/dd-trace-rb/pull/4904/commits/537745862a77dd8b6ccf1a8f617c281593936037)

PR #4908, opened 2025-09-23
DI: evaluate message template in probes

p-datadog:
**What does this PR do?**

Adds message template evaluation to dynamic instrumentation.

Probe specifications (received via remote config) contain the textual message template entered by the user and the parsed AST. The client library is supposed to evaluate the AST to build the message from the message template.

Initial DI implementation took the textual template and performed basic substitution on it (e.g. method duration). This PR implements proper AST evaluation using the expression language code recently merged in previous PRs.

**Motivation:**

DI feature parity with other languages

**Change log entry**
Yes: dynamic instrumentation: message templates are now evaluated when creating log messages for instrumented code

**Additional Notes:**

**How to test the change?**
Unit/integration tests added

### DIFF
--- a/spec/datadog/di/integration/probe_notification_builder_spec.rb
+++ b/spec/datadog/di/integration/probe_notification_builder_spec.rb
@@ -43,7 +43,8 @@ RSpec.describe Datadog::DI::ProbeNotificationBuilder do
       let(:probe) do
         Datadog::DI::Probe.new(
           id: '123', type: :log, file: 'X', line_no: 1,
-          capture_snapshot: true)
+          capture_snapshot: true
+        )
       end
 
       context 'with snapshot' do
@@ -132,12 +133,12 @@ RSpec.describe Datadog::DI::ProbeNotificationBuilder do
       context 'with template segments' do
         let(:probe_spec) do
           {id: '11', name: 'bar', type: 'LOG_PROBE', where: {
-            typeName: 'Foo', methodName: 'bar'},
-            segments: [
-              {str: 'hello'},
-              {json: {ref: 'bar'}},
-            ],
-          }
+                                                       typeName: 'Foo', methodName: 'bar'
+                                                     },
+           segments: [
+             {str: 'hello'},
+             {json: {ref: 'bar'}},
+           ],}
         end
 
         let(:probe) do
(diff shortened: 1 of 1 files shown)


lib/datadog/di/probe_builder.rb
```diff
@@ -45,6 +52,24 @@ module ProbeBuilder
       rescue KeyError => exc
         raise ArgumentError, "Malformed remote configuration entry for probe: #{exc.class}: #{exc}: #{config}"
       end
+
+      def build_template_segments(segments)
+        segments&.map do |segment|
+          if Hash === segment
```
y9v, line 58: how about changing this to a guard clause?

```ruby
next unless segment.is_a?(Hash)
```

lib/datadog/di/probe_notification_builder.rb
```diff
@@ -169,12 +171,17 @@ def format_caller_locations(caller_locations)
...
       end
 
-      def evaluate_template(template, context)
-        message = template.dup
-        context.locals&.each do |key, value|
-          message.gsub!("{@#{key}}") { value.to_s }
-        end
-        message
+      def evaluate_template(template_segments, context)
+        template_segments.map do |segment|
```
y9v, line 175: you can avoid array allocation and build a string right away with something like this:

```ruby
template_segments.each_with_object('') do |segment, result|
  ...
end
```

lib/datadog/di/probe_builder.rb
p-datadog: I have TODO comments in the other branches and I plan on writing code there, this version doesn't allow for the comments or the code.

lib/datadog/di/probe_notification_builder.rb
p-datadog:
```
require 'benchmark/ips'

src=%w(the quick brown fox jumps over the lazy dog)


Benchmark.ips do |x|
x.report("map & join") { 1_000_000.times {

src.map { |x| x.upcase }.join('')

} }
x.report("string concat") { 1_000_000.times {

src.each_with_object('') do |segment, result|
    result += segment.upcase
end


 } }
x.compare!
end





big% ruby ruby-perf/string_join.rb
ruby 3.3.4 (2024-07-09 revision be1089c8ec) [x86_64-linux]
          map & join     1.000 i/100ms
       string concat     1.000 i/100ms
          map & join      0.868 (± 0.0%) i/s     (1.15 s/i) -      5.000 in   5.757662s
       string concat      0.687 (± 0.0%) i/s     (1.46 s/i) -      4.000 in   5.825828s

Comparison:
          map & join:        0.9 i/s
       string concat:        0.7 i/s - 1.26x  slower


```

It is faster to map & join.

PR #4909, opened 2025-09-23
DI: Condition evaluation for method probes

p-datadog:
**What does this PR do?**

Implements condition evaluation for method probes: if condition for the probe was set in UI, the instrumentation will only be called if the condition is met.

**Motivation:**
Feature parity with other languages.

**Change log entry**
Yes: dynamic instrumentation: support conditions for method probes

**Additional Notes:**

**How to test the change?**
Tests added

### DIFF
--- a/lib/datadog/di/instrumenter.rb
+++ b/lib/datadog/di/instrumenter.rb
@@ -118,7 +118,26 @@ module Datadog
 
         mod = Module.new do
           define_method(method_name) do |*args, **kwargs, &target_block| # steep:ignore
-            if rate_limiter.nil? || rate_limiter.allow?
+            continue = true
+            if condition = probe.condition
+              begin
+                # This context will be recreated later, unlike for line probes.
+                context = EL::Context.new(
+                  locals: serializer.combine_args(args, kwargs, self),
+                  target_self: self,
+                  probe: probe, settings: settings, serializer: serializer,
+                  caller_locations: caller_locations,
+                )
+                continue = condition.satisfied?(context)
+              rescue
+                raise if settings.dynamic_instrumentation.internal.propagate_all_exceptions
+
+                # TODO log / report via telemetry?
+                continue = false
+              end
+            end
+
+            if continue and rate_limiter.nil? || rate_limiter.allow?
               # Arguments may be mutated by the method, therefore
               # they need to be serialized prior to method invocation.
               serialized_entry_args = if probe.capture_snapshot?
--- a/lib/datadog/di/instrumenter.rb
+++ b/lib/datadog/di/instrumenter.rb
@@ -3,6 +3,7 @@
 require_relative '../core/utils/time'
 
 # rubocop:disable Lint/AssignmentInCondition
+# rubocop:disable Style/AndOr
 
 module Datadog
   module DI
@@ -516,3 +517,4 @@ module Datadog
 end
 
 # rubocop:enable Lint/AssignmentInCondition
+# rubocop:enable Style/AndOr
--- a/lib/datadog/di/serializer.rb
+++ b/lib/datadog/di/serializer.rb
@@ -79,6 +79,16 @@ module Datadog
       attr_reader :redactor
       attr_reader :telemetry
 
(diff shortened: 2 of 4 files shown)


lib/datadog/di/instrumenter.rb
```diff
@@ -118,7 +119,26 @@ def hook_method(probe, &block)
...
+                continue = condition.satisfied?(context)
+              rescue
+                raise if settings.dynamic_instrumentation.internal.propagate_all_exceptions
+
+                # TODO log / report via telemetry?
+                continue = false
+              end
+            end
+
+            if continue and rate_limiter.nil? || rate_limiter.allow?
```
Strech, line 141: Is that necessary to use `and` here? Maybe round brackets could resolve that?

lib/datadog/di/serializer.rb
```diff
@@ -79,6 +79,16 @@ def initialize(settings, redactor, telemetry: nil)
...
       attr_reader :telemetry
 
+      def combine_args(args, kwargs, target_self)
+        counter = 0
+        args.each_with_object({}) do |value, c|
+          counter += 1
+          # Conversion to symbol is needed here to put args ahead of
+          # kwargs when they are merged below.
+          c[:"arg#{counter}"] = value
+        end.update(kwargs).update(self: target_self)
```
Strech: We can squeeze a bit of performance if we replace last `update` with `[]=`

```
irb(main):017* Benchmark.ips do |x|
irb(main):018*   x.report("update") { 1_000_000.times { Hash.new.update(a: 1) } }
irb(main):019*   x.report("[]=") { 1_000_000.times { Hash.new[:a] = 1 } }
irb(main):020*   x.compare!
irb(main):021> end
ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [arm64-darwin23]
              update     1.000 i/100ms
                 []=     1.000 i/100ms
              update      7.266 (± 0.0%) i/s  (137.62 ms/i) -     37.000 in   5.092963s
                 []=     12.192 (± 0.0%) i/s   (82.02 ms/i) -     61.000 in   5.003209s

Comparison:
                 []=:       12.2 i/s
              update:        7.3 i/s - 1.68x  slower
```

And also we can use `merge!` instead of `update` as we anyway get the new hash after `each_with_object`, but the code will not be chained as now

Strech (approved):
I have few performance suggestions

lib/datadog/di/instrumenter.rb
p-datadog: I use `and` intentionally to avoid typing parentheses (less pinky finger use).

Strech: Understand, what I find suboptimal is that we alter so heavy the rubocop via inline modifications, that I start asking myself - why do we have it in first place if we don't follow it? WDYT?

p-datadog: I do not need a style enforcer in Ruby, my code is already written the way I want it to look :)

lib/datadog/di/serializer.rb
p-datadog: Yes but:

```

hash = {a: 1, b: 2, c: 3, d: 4, e: 5}

Benchmark.ips do |x|
x.report("update") { 1_000_000.times { Hash.new.update(hash) } }
x.report("[]=") { 1_000_000.times {
h={}
hash.each do |k, v|
h[k]=v
end

 } }
x.compare!
end


ruby 3.3.4 (2024-07-09 revision be1089c8ec) [x86_64-linux]
              update     1.000 i/100ms
                 []=     1.000 i/100ms
              update      4.016 (± 0.0%) i/s  (249.01 ms/i) -     21.000 in   5.229777s
                 []=      2.699 (± 0.0%) i/s  (370.50 ms/i) -     14.000 in   5.187968s

Comparison:
              update:        4.0 i/s
                 []=:        2.7 i/s - 1.49x  slower
```

p-datadog: :+1:

PR #4913, opened 2025-09-24
DI: report evaluation errors

p-datadog:
**What does this PR do?**

Reports expression evaluation errors in the payload when building the message from template segments

**Motivation:**
Implementation of expression language for DI

**Change log entry**
Yes: dynamic instrumentation: report evaluation errors to the UI

**Additional Notes:**

**How to test the change?**

Tests are included


lib/datadog/di/probe_notification_builder.rb
```diff
@@ -181,7 +184,11 @@ def evaluate_template(template_segments, context)
           else
             raise ArgumentError, "Invalid template segment type: #{segment}"
           end
+        rescue => exc
+          evaluation_errors << "#{exc.class}: #{exc}"
+          '[evaluation error]'
         end.join('')
```
Strech: BTW, you can just use `join` it will be the same effect

Strech (approved):
👍🏼

lib/datadog/di/probe_notification_builder.rb
p-datadog: :+1:

PR #4916, opened 2025-09-26
move DD log correlation patch for Active Job to wrap full execution

agrobbin:
**What does this PR do?**
Instead of using `around_perform`, we now do the same thing that [Active Job itself](https://github.com/rails/rails/blob/v8.0.3/activejob/lib/active_job/logging.rb#L31-L33) does to tag logs with `[ActiveJob]`, the job class name, and job ID, which is wrapping `perform_now`.

**Motivation:**
The current Active Job integration injects a log correlation tag into the logger via `around_perform`, which is too late to pick up the following logs:

```
[ActiveJob] [FooJob] [540c4718-6668-4f3c-9411-ca52a4cffd15] Performing FooJob (Job ID: 540c4718-6668-4f3c-9411-ca52a4cffd15) from Sidekiq(default) enqueued at 2025-09-26T02:17:44.160363327Z with arguments: #<GlobalID:0x00007f0fae88cb50 @uri=#<URI::GID gid://foo-rails/Foo/019983cf-8422-72d5-85b7-f3dc09a4887e>>
[ActiveJob] [FooJob] [540c4718-6668-4f3c-9411-ca52a4cffd15] Performed FooJob (Job ID: 540c4718-6668-4f3c-9411-ca52a4cffd15) from Sidekiq(default) in 990.29ms
```

**Change log entry**

**Additional Notes:**

**How to test the change?**
I looked for an existing test to exercise this, but was unable to find one, and didn't see any tests specifically related to Active Job itself.

### DIFF
--- a/lib/datadog/tracing/contrib/active_job/log_injection.rb
+++ b/lib/datadog/tracing/contrib/active_job/log_injection.rb
@@ -6,15 +6,11 @@ module Datadog
       module ActiveJob
         # Active Job log injection wrapped around job execution
         module LogInjection
-          def self.included(base)
-            base.class_eval do
-              around_perform do |_, block|
-                if Datadog.configuration.tracing.log_injection && logger.respond_to?(:tagged)
-                  logger.tagged(Tracing.log_correlation, &block)
-                else
-                  block.call
-                end
-              end
+          def perform_now
+            if Datadog.configuration.tracing.log_injection && logger.respond_to?(:tagged)
+              logger.tagged(Tracing.log_correlation) { super }
+            else
+              super
             end
           end
         end


vpellan:
Hey @agrobbin  ! Thanks for your contribution ! Just a heads up to tell you that I will look at this pull request next week as soon as possible !

vpellan:
Hey @agrobbin ! Unfortunately this change does not work with Rails < 6.0 (which we still support). It seems that `perform_now` is called before we start a trace, meaning that the trace_id and the span_id on the logs will be set to 0.
You can test it by using the `ruby_2.7_rails5_postgres_sidekiq` gemfile and running `rspec spec/datadog/tracing/contrib/rails/rails_active_job_spec.rb:211` (you don't actually need sidekiq for this test, it's just that parts of this spec that requires it)
Is this something you want to investigate ?

agrobbin:
@vpellan interesting! I looked at the source and it seems the logging of Active Job itself was reimplemented in Rails 6+, so it makes sense that this wouldn't work quite the same in those versions.

I've introduced branching to only wrap the full execution in Rails 6+, which should ensure that test passes again!

lib/datadog/tracing/contrib/active_job/patcher.rb
```diff
@@ -26,7 +26,11 @@ def patch
 
           def inject_log_correlation
             ::ActiveSupport.on_load(:active_job) do
-              include LogInjection
+              if target_version < Gem::Version.new('6.0.0')
```
vpellan: `target_version` cannot be accessed from ActiveSupport
```suggestion
              if ::ActiveJob.gem_version < Gem::Version.new('6.0.0')
```

vpellan:
Hey @agrobbin  ! Thanks for the fix ! I left a small comment but except that, LGTM ! We need to run these changes on our latest CI before merging, would you agree to give us the authorization to merge master on your branch ? So we don't have to bother you each time we need to run the CI. Thanks !

agrobbin:
Updated! I also rebased off of the latest `master`. If things change on `master` before this can be merged, feel free to merge it into this branch (I've checked the "allow edits ... by maintainers" box).

lib/datadog/tracing/contrib/active_job/log_injection.rb
```diff
@@ -6,17 +6,31 @@ module Contrib
...
         module LogInjection
-          def self.included(base)
-            base.class_eval do
-              around_perform do |_, block|
-                if Datadog.configuration.tracing.log_injection && logger.respond_to?(:tagged)
-                  logger.tagged(Tracing.log_correlation, &block)
-                else
-                  block.call
+          # Active Job 4 / 5 don't execute `perform_now` at the right point, so we do best effort log correlation tagging
+          module ActiveJob4
```
y9v: One minor suggestion:

What do you think, does it make sense to avoid the duplicated "ActiveJob" in the class name `ActiveJob::LogInjection::ActionJobX` and name the module by what it does? I'm thinking of something like `ActiveJob::LogInjection::AroundPerformPatch` and `ActiveJob::LogInjection::TaggedPatch`?

agrobbin: @y9v sounds good to me! Updated.

vpellan:
Thank you very much for your contribution, this will be available in the next release !

PR #4918, opened 2025-09-26
[APPSEC-59194] Implement trace tagging logic

Strech:
**What does this PR do?**

Changes the logic of trace keeping (altering sampling priority) based on rules configuration instead of hardcoded values.

**Motivation:**

This is a part of RFC for JWT analysis and trace tagging.

**Change log entry**

Yes. AppSec: WAF rules can now enforce traces to be always kept or sampled.

**Additional Notes:**

The `Event` class is a subject of the refactoring that will happen next.

Highlight:

Every WAF rule might want to utilize two new features, a `keep` flag, that controls whether the trace is always reported or sampled and `attributes` field that allows an arbitrary tags to be reported.

```
{
    "output": {
        "keep": (true | false),
        "event": (true | false),
        "attributes": { 
        }
    }
}
```

For the full list of changes, please refer to the `libddawf` [release notes](https://github.com/DataDog/libddwaf/blob/master/CHANGELOG.md#v1250-unstable).

**How to test the change?**

CI + [ST](https://github.com/DataDog/system-tests/pull/5352)

### DIFF
--- a/spec/datadog/appsec/integration/jwt_analysis_spec.rb
+++ b/spec/datadog/appsec/integration/jwt_analysis_spec.rb
@@ -57,7 +57,7 @@ RSpec.describe 'Blocking with deny and pass list configuration' do
           'transformers' => [],
           'output' => {
             'event' => false,
-            'keep' => false,
+            'keep' => true,
             'attributes' => {'_dd.appsec.api.jwt.no_expiry' => {'value' => 1}}
           }
         }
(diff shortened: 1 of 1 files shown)


lib/datadog/appsec/contrib/faraday/ssrf_detection_middleware.rb
```diff
@@ -21,7 +22,8 @@ def call(request_env)
             result = context.run_rasp(Ext::RASP_SSRF, {}, ephemeral_data, Datadog.configuration.appsec.waf_timeout)
 
             if result.match?
-              AppSec::Event.tag_and_keep!(context, result)
+              AppSec::Event.tag(context, result)
+              TraceKeeper.keep!(context.trace) if result.keep?
```
y9v, line 26: I really like this separation

PR #4919, opened 2025-09-26
Add endpoint collection for Rails routes

y9v:
**What does this PR do?**
This PR adds API Security Endpoint Collection feature for Rails framework. The feature is disabled by default for now.

**Motivation:**
We want to know about all existing routes in an instrumented application, not only routes, which get traffic.

**Change log entry**
Yes. AppSec: Add endpoint collection for Rails application for API Security.

**Additional Notes:**
None.

**How to test the change?**
CI and manual testing with generated apps.

### DIFF
--- a/lib/datadog/appsec/contrib/rails/patcher.rb
+++ b/lib/datadog/appsec/contrib/rails/patcher.rb
@@ -134,6 +134,8 @@ module Datadog
           end
 
           def subscribe_to_routes_loaded
+            return unless Datadog.configuration.appsec.api_security.endpoint_collection.enabled
+
             ::ActiveSupport.on_load(:after_routes_loaded) do |app|
               AppSec.telemetry.app_endpoints_loaded(
                 APISecurity::EndpointCollection::RailsRoutesSerializer.new(app.routes.routes).to_enum
--- a/lib/datadog/appsec/contrib/rails/patcher.rb
+++ b/lib/datadog/appsec/contrib/rails/patcher.rb
@@ -10,6 +10,7 @@ require_relative 'gateway/watcher'
 require_relative 'gateway/request'
 require_relative 'patches/render_to_body_patch'
 require_relative 'patches/process_action_patch'
+require_relative '../../api_security/endpoint_collection/rails_routes_serializer'
 
 require_relative '../../../tracing/contrib/rack/middlewares'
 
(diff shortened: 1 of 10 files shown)


lib/datadog/appsec/contrib/rails/patcher.rb
```diff
@@ -131,6 +133,16 @@ def patch_action_controller
             end
           end
 
+          def subscribe_to_routes_loaded
+            return unless Datadog.configuration.appsec.api_security.endpoint_collection.enabled
```
p-datadog, line 153: Should this be checked in the `after_routes_loaded` callback? If the user reconfigures dd-trace-rb to enable endpoint collection would that take effect?

lib/datadog/appsec/contrib/rails/patcher.rb
```diff
@@ -131,6 +133,16 @@ def patch_action_controller
...
+
+            ::ActiveSupport.on_load(:after_routes_loaded) do |app|
+              AppSec.telemetry.app_endpoints_loaded(
```
p-datadog, line 156: I missed this method in https://github.com/DataDog/dd-trace-rb/pull/4904 but reading it now :+1:

p-datadog (approved):

lib/datadog/appsec/contrib/rails/patcher.rb
y9v: If the user reconfigures `dd-trace-rb`, we still would not be able to collect routes, since `after_routes_loaded` callback is only fired once after the routes are loaded. So I think it shouldn't make a difference for now.

spec/datadog/appsec/integration/contrib/rails/endpoint_collection_spec.rb
```diff
@@ -0,0 +1,138 @@
...
+
+      c.remote.enabled = false
+    end
+
+    app.initialize!
+
+    app.routes.draw do
+      resources :products
+
+      match '/search', to: 'search#index', via: :all
```
Strech, line 67: Does it make sense to add move common route definitions, including

1. get/post/put/patch
2. namespaces (and conditions, to be sure)
3. resources + only (too see that reduction works)

y9v: good idea, added more specs here: [5ca236f](https://github.com/DataDog/dd-trace-rb/pull/4919/commits/5ca236f47d6f36d25d502830fbc18649e3d497c6)

PR #4921, opened 2025-09-29
Fix CI after image update

lloeki:
**What does this PR do?**

- Adjust GHA bundle cache
- Fix erros showing up in CI

**Motivation:**

Cache key does not take into account dimensions that can matter when images change beyond a ruby engine's "major.minor".

**Change log entry**

None

**Additional Notes:**

**Jira**
- [LANGPLAT-853](https://datadoghq.atlassian.net/browse/LANGPLAT-853)
- [LANGPLAT-854](https://datadoghq.atlassian.net/browse/LANGPLAT-854)
- [LANGPLAT-856](https://datadoghq.atlassian.net/browse/LANGPLAT-856)
- [LANGPLAT-857](https://datadoghq.atlassian.net/browse/LANGPLAT-857)
- [LANGPLAT-858](https://datadoghq.atlassian.net/browse/LANGPLAT-858)

**How to test the change?**

CI should pass

[LANGPLAT-858]: https://datadoghq.atlassian.net/browse/LANGPLAT-858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[LANGPLAT-853]: https://datadoghq.atlassian.net/browse/LANGPLAT-853?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4920, opened 2025-09-29
[NO-TICKET] Update steep to 1.10

vpellan:
**What does this PR do?**

This updates Steep to 1.10 and fixes newly introduces checks (such as missing declaration)

**Motivation:**

Working on typing analysis project

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

PR #4924, opened 2025-09-29
Add Ruby 3.5 preview support

sarahchen6:
**What does this PR do?**
Adds Ruby 3.5 preview support to CI and integrations.

**Motivation:**
Start testing Ruby 3.5!

**Change log entry**
No change log entry.

**Additional Notes:**
Tests that are skipped and need further investigation:
* Grape integration
* GRPC integration (disabled for Ruby 3.4 as well)
* Sequel integration
* Sinatra integration
* 1 test for `ratio_in_yjit` metric
* 2 tests in `Datadog::Profiling::Collectors::CpuAndWallTimeWorker`
* 1 test in `Datadog::Tracing::Contrib::Shoryuken::Tracer`

These all lead with `TODO: Ruby 3.5 - `...

The skipped tests are documented on the Ruby Guild board: https://github.com/orgs/DataDog/projects/47/views/10?query=sort%3Aupdated-desc+is%3Aopen&sliceBy%5Bvalue%5D=lang+support

**How to test the change?**
Green CI.

### DIFF
--- a/gemfiles/ruby_3.5_activesupport.gemfile.lock
+++ b/gemfiles/ruby_3.5_activesupport.gemfile.lock
@@ -4,7 +4,7 @@ PATH
     datadog (2.22.0)
       datadog-ruby_core_source (~> 3.4, >= 3.4.1)
       libdatadog (~> 18.1.0.1.0)
-      libddwaf (~> 1.25.1.0.1)
+      libddwaf (~> 1.25.1.1.0)
       logger
       msgpack
 
@@ -69,9 +69,6 @@ GEM
     crass (1.0.6)
     datadog-ruby_core_source (3.4.1)
     date (3.4.1)
-    debug (1.11.0)
-      irb (~> 1.10)
-      reline (>= 0.3.8)
     diff-lcs (1.6.2)
     digest-crc (0.7.0)
       rake (>= 12.0.0, < 14.0.0)
@@ -99,6 +96,7 @@ GEM
     erubi (1.13.1)
     extlz4 (0.3.5)
     ffi (1.17.2)
+    fiddle (1.1.8)
     google-protobuf (3.25.8)
     grape (2.4.0)
       activesupport (>= 6.1)
@@ -121,9 +119,12 @@ GEM
     king_konf (1.0.1)
     libdatadog (18.1.0.1.0)
     libdatadog (18.1.0.1.0-aarch64-linux)
-    libddwaf (1.25.1.0.1)
+    libdatadog (18.1.0.1.0-x86_64-linux)
+    libddwaf (1.25.1.1.0)
       ffi (~> 1.0)
-    libddwaf (1.25.1.0.1-aarch64-linux)
+    libddwaf (1.25.1.1.0-aarch64-linux)
+      ffi (~> 1.0)
+    libddwaf (1.25.1.1.0-x86_64-linux)
       ffi (~> 1.0)
     logger (1.7.0)
     lograge (0.14.0)
@@ -148,6 +149,7 @@ GEM
       mini_portile2 (~> 2.8.2)
       racc (~> 1.4)
     os (1.1.4)
+    ostruct (0.6.3)
     pp (0.6.2)
       prettyprint
     prettyprint (0.2.0)
@@ -195,6 +197,12 @@ GEM
       mini_portile2 (~> 2.6)
       rake (> 12)
     rdkafka (0.23.1-aarch64-linux-gnu)
+      ffi (~> 1.15)
+      json (> 2.0)
+      logger
+      mini_portile2 (~> 2.6)
+      rake (> 12)
+    rdkafka (0.23.1-x86_64-linux-gnu)
       ffi (~> 1.15)
       json (> 2.0)
       logger
@@ -252,28 +260,33 @@ GEM
 PLATFORMS
   aarch64-linux
   ruby
+  x86_64-linux
 
 DEPENDENCIES
   actionpack
   actionview
   active_model_serializers (>= 0.10.0)
   activesupport (~> 7)
+  base64
   benchmark
   benchmark-ips (~> 2.8)
   benchmark-memory (< 0.2)
+  bigdecimal
   byebug
   climate_control (~> 1.2.0)
   concurrent-ruby
   datadog!
-  debug
   dogstatsd-ruby (>= 3.3.0, != 5.1.0, != 5.0.1, != 5.0.0)
   extlz4 (~> 0.3, >= 0.3.3)
+  fiddle
   google-protobuf (~> 3.0, != 3.7.1, != 3.7.0)
   grape
   json-schema (< 3)
   lograge
   memory_profiler (~> 0.9)
+  mutex_m
   os (~> 1.1)
+  ostruct
   pry
   racecar (>= 0.3.5)
   rake (>= 10.5)
@@ -286,7 +299,7 @@ DEPENDENCIES
   simplecov (~> 0.22.0)
   warning (~> 1)
   webmock (>= 3.10.0)
-  webrick (>= 1.7.0)
+  webrick (>= 1.8.2)
 
 BUNDLED WITH
    2.7.2
--- a/gemfiles/ruby_3.5_aws.gemfile.lock
+++ b/gemfiles/ruby_3.5_aws.gemfile.lock
@@ -1700,9 +1700,12 @@ GEM
       addressable (>= 2.4)
     libdatadog (18.1.0.1.0)
     libdatadog (18.1.0.1.0-aarch64-linux)
+    libdatadog (18.1.0.1.0-x86_64-linux)
     libddwaf (1.25.1.1.0)
       ffi (~> 1.0)
     libddwaf (1.25.1.1.0-aarch64-linux)
+      ffi (~> 1.0)
+    libddwaf (1.25.1.1.0-x86_64-linux)
       ffi (~> 1.0)
     logger (1.7.0)
     memory_profiler (0.9.14)
@@ -1759,6 +1762,7 @@ GEM
 PLATFORMS
   aarch64-linux
   ruby
+  x86_64-linux
 
 DEPENDENCIES
   aws-sdk
--- a/gemfiles/ruby_3.5_contrib.gemfile.lock
+++ b/gemfiles/ruby_3.5_contrib.gemfile.lock
@@ -4,7 +4,7 @@ PATH
     datadog (2.22.0)
       datadog-ruby_core_source (~> 3.4, >= 3.4.1)
       libdatadog (~> 18.1.0.1.0)
-      libddwaf (~> 1.25.1.0.1)
+      libddwaf (~> 1.25.1.1.0)
       logger
       msgpack
 
@@ -32,31 +32,25 @@ GEM
       bigdecimal
       rexml
     datadog-ruby_core_source (3.4.1)
-    date (3.4.1)
-    debug (1.11.0)
-      irb (~> 1.10)
-      reline (>= 0.3.8)
     diff-lcs (1.6.2)
     docile (1.4.1)
     dogstatsd-ruby (5.7.1)
-    erb (5.0.2)
     extlz4 (0.3.5)
     ffi (1.17.2)
+    fiddle (1.1.8)
     google-protobuf (3.25.8)
     hashdiff (1.2.1)
-    io-console (0.8.1)
-    irb (1.15.2)
-      pp (>= 0.6.0)
-      rdoc (>= 4.0.0)
-      reline (>= 0.4.2)
     json (2.15.0)
     json-schema (2.8.1)
       addressable (>= 2.4)
     libdatadog (18.1.0.1.0)
     libdatadog (18.1.0.1.0-aarch64-linux)
-    libddwaf (1.25.1.0.1)
+    libdatadog (18.1.0.1.0-x86_64-linux)
+    libddwaf (1.25.1.1.0)
       ffi (~> 1.0)
-    libddwaf (1.25.1.0.1-aarch64-linux)
+    libddwaf (1.25.1.1.0-aarch64-linux)
+      ffi (~> 1.0)
+    libddwaf (1.25.1.1.0-x86_64-linux)
       ffi (~> 1.0)
     logger (1.7.0)
     memory_profiler (0.9.14)
@@ -66,16 +60,12 @@ GEM
     multi_json (1.17.0)
     mustermann (3.0.4)
       ruby2_keywords (~> 0.0.1)
+    mutex_m (0.3.0)
     os (1.1.4)
-    pp (0.6.2)
-      prettyprint
-    prettyprint (0.2.0)
+    ostruct (0.6.3)
     pry (0.15.2)
       coderay (~> 1.1)
       method_source (~> 1.0)
-    psych (5.2.6)
-      date
-      stringio
     public_suffix (6.0.2)
     que (2.4.1)
     rack (3.2.1)
@@ -94,17 +84,12 @@ GEM
     rake-compiler (1.3.0)
       rake
     rbtree (0.4.6)
-    rdoc (6.14.2)
-      erb
-      psych (>= 4.0.0)
     redis (5.4.1)
       redis-client (>= 0.22.0)
     redis-client (0.26.1)
       connection_pool
     redis-namespace (1.11.0)
       redis (>= 4)
-    reline (0.6.2)
-      io-console (~> 0.5)
     resque (2.7.0)
       mono_logger (~> 1)
       multi_json (~> 1.0)
@@ -167,7 +152,6 @@ GEM
     sorted_set (1.0.3)
       rbtree
       set (~> 1.0)
-    stringio (3.1.7)
     sucker_punch (3.2.0)
       concurrent-ruby (~> 1.0)
     thor (1.4.0)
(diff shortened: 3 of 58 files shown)


ivoanjo:
Really excited for this! Ping me if anything's needed from the profiling side :)

sarahchen6:
@ivoanjo Thanks! Will do 😄

p-datadog:
Due to https://github.com/DataDog/dd-trace-rb/pull/4920 landing I think this PR should refer to steep 1.10

.gitlab-ci.yml
```diff
@@ -35,7 +35,7 @@ default:
   parallel:
     matrix:
       # ADD NEW RUBIES HERE
-      - RUBY_VERSION: ["3.4", "3.3", "3.2", "3.1", "3.0", "2.7", "2.6"]
+      - RUBY_VERSION: ["3.5", "3.4", "3.3", "3.2", "3.1", "3.0", "2.7", "2.6"]
```
lloeki, line 38: IIRC this is OK as it only builds the base Ruby image that will be used downstream

.gitlab-ci.yml
```diff
@@ -71,7 +71,7 @@ promote-image:
...
```
lloeki, line 74: IIRC this is OK as it only promotes the base Ruby image that will be used downstream

.gitlab-ci.yml
```diff
@@ -109,7 +109,8 @@ install-dependencies:
   tags: ["arch:$ARCH"]
   parallel:
     matrix:
-      - RUBY_VERSION: ["3.4", "3.3", "3.2", "3.1", "3.0", "2.7", "2.6"]
+      # ADD NEW RUBIES HERE
+      - RUBY_VERSION: ["3.5", "3.4", "3.3", "3.2", "3.1", "3.0", "2.7", "2.6"]
```
lloeki: ... but IIRC this is problematic as it will cause the SSI OCI package to contain 3.5 and thus implicitly add tentative support for 3.5, as well as make the image size go up.

docker-compose.yml
```diff
@@ -159,6 +159,24 @@ services:
       - extension-build-tmp:/app/tmp
       - bundle-3.4:/usr/local/bundle
       - "ddagent_var_run:${TEST_DDAGENT_VAR_RUN}"
+
```
lloeki: Nitpick: inconsistent extra like

docker-compose.yml
```diff
@@ -159,6 +159,24 @@ services:
...
+  tracer-3.5:
+    image: ghcr.io/datadog/images-rb/engines/ruby:3.5
```
lloeki: Out of scope for this PR but we should probably start to explicitly reference e.g `x.y-gnu-gcc` throughout as that's what we actually require.

Indeed `x.y` is currently aliased to that but is supposed to be only the runtime without compiler; see [images-rb's README](https://github.com/DataDog/images-rb?tab=readme-ov-file#ruby-engines).

appraisal/ruby-3.5.rb
```diff
@@ -0,0 +1,248 @@
...
+end
+
+appraise 'rails61-trilogy' do
+  gem 'rails', '~> 6.1.0'
+  gem 'trilogy'
+  gem 'activerecord-trilogy-adapter'
+  gem 'sprockets', '< 4'
+  gem 'lograge', '~> 0.11'
+  gem 'net-smtp'
+end
```
lloeki: As discussed, Rails 6 isn't compatible with Ruby 3.5, so let's remove this whole thing.

spec/datadog/core/runtime/metrics_spec.rb
```diff
@@ -251,6 +251,8 @@
         context 'with YJIT enabled and RubyVM::YJIT.stats_enabled? true' do
           before do
             skip('Test only runs on Ruby >= 3.3') if RUBY_VERSION < '3.3.'
+            # TODO: Remove this once Ruby 3.5 is released
+            skip('ratio_in_yjit metric not available in Ruby 3.5.0-preview1') if RUBY_VERSION.start_with?('3.5.0preview1')
```
lloeki: Probably you wanted `RUBY_DESCRIPTION =~ /3\.5\.0preview\d+/`

Indeed:

```
$ docker run --rm -it  ghcr.io/datadog/images-rb/engines/ruby:3.5
irb(main):001> RUBY_VERSION
=> "3.5.0"
irb(main):002> RUBY_DESCRIPTION
=> "ruby 3.5.0preview1 (2025-04-18 master d06ec25be4) +PRISM [aarch64-linux]"
```

Matrixfile
```diff
@@ -134,9 +134,11 @@
   'lograge' => {
     'activesupport' => '✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby'
   },
+  # MongoDB tests excluded for Ruby 3.5 due to bson gem native extension compilation failures
+  # with the Ruby 3.5.0-preview1 image. TODO: Ruby 3.5 - Re-enable after investigation.
```
lloeki: I think it's OK with a recent `bson` gem, but minimum `mongodb` version might need to be raised.

flake.lock
```diff
@@ -1,76 +0,0 @@
-{
```
lloeki, line 1: This one should not be removed, it's Nix, not Bundler.

integration/images/ruby/3.5/Dockerfile
```diff
@@ -0,0 +1,54 @@
+FROM ruby:3.5.0-preview1
```
lloeki: Wondering why those don't use images-rb as a base

.gitlab-ci.yml
sarahchen6: Ah I see. Will remove, thanks!

appraisal/ruby-3.5.rb
sarahchen6: Sounds good

spec/datadog/core/runtime/metrics_spec.rb
sarahchen6: Oh cool, thanks!

Matrixfile
sarahchen6: Turns out `mongodb` doesn't support Ruby 3.4+, so I think we should ❌ them both. Ref: https://github.com/mongodb/mongo-ruby-driver#mongodb-ruby-driver

integration/images/ruby/3.5/Dockerfile
sarahchen6: Oh good point. I was copying the format of the other versions' Dockerfiles, but I'll update this.

sarahchen6: Followed up with #4940 to change all base images to use `images-rb` 🙂

spec/datadog/tracing/contrib/shoryuken/tracer_spec.rb
```diff
@@ -183,6 +183,9 @@ def perform(sqs_msg, body)
       # end
 
       it do
+        # Test does not work correctly on Ruby 3.5.0-preview1.
+        # TODO: Ruby 3.5 - Re-enable after investigation.
+        skip 'Skipped on Ruby 3.5.0-preview1' if RUBY_VERSION.start_with?('3.5')
         expect { perform_async }.to_not raise_error
```
marcotc, line 187: Given these skips are strictly scope to the newly introduced Ruby 3.5, I'm more than happy to these changes hit the road!

.gitlab/Dockerfile-3.5
```diff
@@ -0,0 +1 @@
+FROM ghcr.io/datadog/images-rb/engines/ruby:3.5-centos-gcc
```
marcotc, line 1: Do we not want to version it to `preview1`, making it explicit (and maybe ensuring the correct image version around the time of a version bump) @lloeki? (here and in our regular CI)

docs/Compatibility.md
```diff
@@ -9,7 +9,8 @@ The Ruby Datadog Trace library is open source. See the [dd-trace-rb][1] GitHub r
 
 | Type  | Documentation              | Version   | Support type              | Gem version support |
 |-------|----------------------------|-----------|---------------------------|---------------------|
-| MRI   | https://www.ruby-lang.org/ | 3.4       | [latest](#support-latest) | Latest              |
+| MRI   | https://www.ruby-lang.org/ | 3.5       | [latest](#support-latest) | Latest              |
```
ivoanjo: I don't think this part should be changed yet. Similar to https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/version.rb#L24 which also needs to be changed, but not yet.

Specifically between now and December 25th (assuming 3.5 follows the usual release schedule) there may be something that changes that breaks dd-trace-rb. It's unlikely, but I don't think it makes sense to claim support for 3.5 when we've not actually validated that the library works with the final 3.5.0 release -- we've only checked with the snapshots.

TL;DR I think we should only bump this note either very shortly before, or after the upstream 3.5.0 is out.

ivoanjo (commented):
I've left a note in slack, but summing it here too:

* This is great
* We should probably not claim support for 3.5 until very shortly before the 3.5.0 release (so we can actually validate with that version)
* Let's merge this in and work separately on the skips

docs/Compatibility.md
sarahchen6: Agreed! I removed the documentation update.

hestonhoffman (approved):

.gitlab-ci.yml
```diff
@@ -109,6 +109,7 @@ install-dependencies:
   tags: ["arch:$ARCH"]
   parallel:
     matrix:
+      # ADD NEW RUBIES HERE
       - RUBY_VERSION: ["3.4", "3.3", "3.2", "3.1", "3.0", "2.7", "2.6"]
```
ivoanjo, line 114: Are we missing a 3.5 here? :eyes:
If it's on purpose, maybe leave a `# TODO: Not adding 3.5 yet because...` so we don't forget?

spec/datadog/tracing/contrib/shoryuken/tracer_spec.rb
```diff
@@ -183,6 +183,9 @@ def perform(sqs_msg, body)
       # end
 
       it do
+        # Test does not work correctly on Ruby 3.5.0-preview1.
+        # TODO: Ruby 3.5 - Re-enable after investigation.
+        skip 'Skipped on Ruby 3.5.0-preview1' if RUBY_VERSION.start_with?('3.5')
```
p-datadog:
```suggestion
        skip 'Skipped on Ruby 3.5.0-preview1' if RUBY_VERSION =~ /\A3\.5.*preview/
```
What do you think about this pattern to get these tests running on 3.5 final automatically?

p-datadog: Also, would `pending` work instead of `skip` to tell us if the test fixes itself?

appraisal/ruby-3.5.rb
```diff
@@ -0,0 +1,195 @@
...
+  gem 'rails', '~> 8.0.0'
+  gem 'trilogy'
+  gem 'sprockets', '< 4'
+  gem 'lograge', '~> 0.11'
+  gem 'net-smtp'
+end
+
+appraise 'rails-old-redis' do
+  # All dependencies except Redis < 4 are not important, they are just required to run Rails tests.
+  gem 'redis', '< 4'
```
p-datadog, line 57: It's not clear from the diff what the "not old" redis configurations are - is it the resque-redis ones below? I think this configuration (old redis) can just be removed.

appraisal/ruby-3.5.rb
```diff
@@ -0,0 +1,195 @@
...
+  gem 'typhoeus'
+end
+
+build_coverage_matrix('stripe', 7..12, min: '5.15.0')
+build_coverage_matrix('opensearch', [2], gem: 'opensearch-ruby')
+build_coverage_matrix('elasticsearch', [7])
+build_coverage_matrix('faraday')
+build_coverage_matrix('excon')
+build_coverage_matrix('rest-client')
+build_coverage_matrix('mongo', min: '2.1.0')
```
p-datadog: For Ruby 3.4 mongo min is now 2.20 after https://github.com/DataDog/dd-trace-rb/pull/4921, does 3.5 really work with mongo 2.1?

gemfiles/ruby_3.5_mongo_min.gemfile
```diff
@@ -0,0 +1,43 @@
...
+gem "rake-compiler", "~> 1.1", ">= 1.1.1"
+gem "rspec", "~> 3.13"
+gem "rspec-collection_matchers", "~> 1.1"
+gem "rspec_junit_formatter", ">= 0.5.1"
+gem "rspec-wait", "~> 0"
+gem "simplecov", "~> 0.22.0"
+gem "warning", "~> 1"
+gem "webmock", ">= 3.10.0"
+gem "webrick", ">= 1.8.2"
+gem "mongo", "= 2.1.0"
```
p-datadog: I see this here but there is no lock file for this gemfile?

spec/datadog/tracing/contrib/shoryuken/tracer_spec.rb
ivoanjo: Uuuh I like like this -- rather than "we need to remember", actually break CI on a non-preview ruby! This is a great idea +1!

.gitlab-ci.yml
sarahchen6: Yes this was on purpose due to https://github.com/DataDog/dd-trace-rb/pull/4924#discussion_r2401778678 (just unresolved it temporarily for visibility), but good idea - I'll add a comment!

spec/datadog/tracing/contrib/shoryuken/tracer_spec.rb
sarahchen6: Cool TIL about `pending`! But turns out this test was failing because it needed the `ostruct` gem (thanks @ivoanjo  https://github.com/DataDog/ruby-guild/issues/258), so I'll un-skip this entirely. 

I applied this logic to the other non-Matrixfile skipped tests though. Since `RUBY_VERSION` is `3.5.0`, I check to see `if RUBY_DESCRIPTION.include?('preview')`

appraisal/ruby-3.5.rb
sarahchen6: Sounds good, removed. I also am unsure what the "old redis" is for but kept it because of its presence in the previous Ruby versions.

appraisal/ruby-3.5.rb
sarahchen6: It seems like `mongo` is not compatible with Ruby 3.4+ (https://github.com/mongodb/mongo-ruby-driver#mongodb-ruby-driver), so I'll remove this line.

gemfiles/ruby_3.5_mongo_min.gemfile
sarahchen6: Ah yes I had run `appraisal/generate.rb` then removed `mongo` which (if I understand correctly) caused the gemfile to be created but not the lockfile... I'll remove this gemfile anyway because we're not testing Ruby 3.5 with `mongo` ([mongodb/mongo-ruby-driver#mongodb-ruby-driver](https://github.com/mongodb/mongo-ruby-driver?rgh-link-date=2025-10-07T18%3A32%3A31Z#mongodb-ruby-driver)).

docker-compose.yml
sarahchen6: Added here :) https://github.com/DataDog/dd-trace-rb/pull/4959 (and to this PR)

appraisal/ruby-3.5.rb
sarahchen6: Ah JK - `rails_old_redis` is tested here: https://github.com/DataDog/dd-trace-rb/blob/master/Matrixfile#L260-L262

p-datadog (approved):

gemfiles/ruby_3.5_mongo_min.gemfile
```diff
@@ -0,0 +1,43 @@
...
+gem "rake-compiler", "~> 1.1", ">= 1.1.1"
+gem "rspec", "~> 3.13"
+gem "rspec-collection_matchers", "~> 1.1"
+gem "rspec_junit_formatter", ">= 0.5.1"
+gem "rspec-wait", "~> 0"
+gem "simplecov", "~> 0.22.0"
+gem "warning", "~> 1"
+gem "webmock", ">= 3.10.0"
+gem "webrick", ">= 1.8.2"
+gem "mongo", "= 2.20.0"
```
p-datadog, line 33: There is still no lock file for this gemfile, I feel like this should be flagged somehow by tooling. Is this gemfile used?

sarahchen6: I had assumed that the `Lock Dependency` workflow would create the lockfile on its own, given that the gemfile is used: https://github.com/DataDog/dd-trace-rb/pull/4924/files#diff-fd714a2c217a623cc12e25983cda125917a7cf711126a53daee1fc94b6365192R141. But from #4928, there is no `Lock Dependency` automation, assuming because I didn't manually generate the lockfiles, so it seems like I need to generate the lockfile myself for the automation to then update. Will do so.

p-datadog: :+1:

marcotc: The GH action we use to create commits does not pick up new files today, it's a known issue: https://github.com/planetscale/ghcommit-action/issues/43

We can fix that by staging the files before committing.

.gitlab/Dockerfile-3.5
lloeki: I understand that you'd somehow wnat to mark a difference between "that's a preview" and "that's not a preview", but TBH I dont' think it's helpful to go down to such precision: functionally if you say `:3.5` there then you want 3.5.anything, preview or not; otherwise you'd say `:3.5.0-preview1` just as you'd say `:3.4.6` (but that would exclude `:3.4.7`).

Also what I've noticed in practice is that we forget to remove `-preview1` in some areas and then it lingers til we realise where we forgot to update it.

This way, as soon as upstream 3.5 is out we update+publish the image on images-rb and it'll be picked up right away†, minimising ceremonial code churn.

† `docker buildx build` will check the registry, if only to look for cache info, which will make it pull the proper up-to-date layers; therefore CI will always be up to date, and locally as soon as you do `buildx build` or pull.

lloeki: I can add alias tags down to `major.minor.patch(-extra)` for each image in images-rb though, for a consumer to decide what they want to have a contract on.

lloeki: Opened: https://github.com/DataDog/images-rb/issues/67

PR #4926, opened 2025-10-01
DI: serializer for log message output

p-datadog:
**What does this PR do?**

Adds a serializer for values when they are written into log messages.

**Motivation:**
Log messages must be human-readable and they are a lot shorter than the full snapshots, therefore they require a more compact serialization format.

**Change log entry**
None (this changes customer-visible behavior but expression language has not yet been released to customers)

**Additional Notes:**

**How to test the change?**

Tests are included


lib/datadog/di/serializer.rb
```diff
@@ -155,7 +158,8 @@ def serialize_value(value, name: nil,
             end
           end
 
-          serialized = {type: class_name(cls)}
+          # steep assuming invalid types
+          serialized = {}.update(type: class_name(cls))
```
marcotc: Hmm, this code is now worse, in order to please steep. Can we add a steep ignore instead?

spec/datadog/di/integration/instrumentation_spec.rb
```diff
@@ -572,7 +572,8 @@ def run_test
                 expect(status).to match(expected_emitting_payload)
               end
               expect(component.probe_notifier_worker).to receive(:add_snapshot) do |snapshot|
-                expect(snapshot.fetch(:message)).to eq 'hello '
+                # TODO should we serialize nil as empty string?
```
marcotc, line 575: an explicit "nil" string, as it's implemented now seems ok, given it looks like the pattern we are implementing is more similar to pp than print.

marcotc (approved):

lib/datadog/di/serializer.rb
p-datadog: :+1: 
https://github.com/soutaro/steep/issues/1860

PR #4928, opened 2025-10-02
Add JRuby 10.0 support

sarahchen6:
**What does this PR do?**
Adds JRuby 10.0 support to CI.

**Motivation:**
JRuby 10.0 is out!

**Change log entry**
No change log entry.

**Additional Notes:**

**How to test the change?**
Green CI.


docker-compose.yml
```diff
@@ -208,6 +208,23 @@ services:
       - .:/app
       - bundle-jruby-9.4:/usr/local/bundle
       - "ddagent_var_run:${TEST_DDAGENT_VAR_RUN}"
+
```
lloeki: Nitpick: inconsistent extra like

spec/datadog/appsec/contrib/rack/integration_test_spec.rb
```diff
@@ -503,9 +503,10 @@
               expect(spans).to have(1).items
               expect(span).to have_tag('_dd.rc.boot.time')
               expect(span.get_tag('_dd.rc.boot.time')).to be_a Float
-              expect(span).to have_tag('_dd.rc.boot.timeout')
-              expect(span.get_tag('_dd.rc.boot.timeout')).to eq 'true'
-              expect(span).to_not have_tag('_dd.rc.boot.ready')
+              # TODO: JRuby 10.0 - Remove this skip after investigation.
+              expect(span).to have_tag('_dd.rc.boot.timeout') unless RUBY_PLATFORM == 'java' && RUBY_ENGINE_VERSION.start_with?('10.0')
```
marcotc: Take a look at `PlatformHelpers`, and see if helps matching the affected JRuby version.

marcotc: Tbh, you can even define a `PlatformHelpers.jruby_10?` there, since we are using it quite a lot in this PR. (if it turns out we actually fix most if the skips, then disregard this suggestion)

sarahchen6: Sounds good! I'll add a comment to delete it too

Matrixfile
```diff
@@ -137,8 +137,10 @@
     'activesupport' => '✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby'
   },
   'mongodb' => {
-    'mongo-latest' => '✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby',
-    'mongo-min'    => '✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby',
+    # MongoDB does not support JRuby 10.0, which fails on an error related the bson_ruby NativeService for latest versions.
+    # https://github.com/mongodb/mongo-ruby-driver#mongodb-ruby-driver
+    'mongo-latest' => '✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby / ❌ jruby 10.0',
+    'mongo-min'    => '✅ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ✅ jruby / ❌ jruby 10.0',
```
sarahchen6: I'm stuck on how to properly skip JRuby 10.0 for `mongodb` tests (see also the `tasks/runtime_matcher.rb` file), or if that's even the right way to go. JRuby 10.0 fails with the latest version 2.21.3 with an error related to:
```
Failure/Error: Unable to find org.bson_ruby.NativeService.basicLoad(NativeService.java to read failed line

Java::JavaLang::NoSuchMethodError:
  'org.jruby.RubyModule org.jruby.Ruby.fastGetModule(java.lang.String)'...
```
Given the lack of support, I propose to skip these tests, but I'm not sure how best to skip them. What am I missing?

ivoanjo: This sounds like a bug in the jruby version of the bson gem; did you spot which version we're using?

(We could adjust the dsl to support skipping jruby versions, but it may be easier to skip on all jruby versions temporarily rather than doing that...?)

vpellan: Would it be possible to use PlatformHelpers.jruby_100? in the specs themselves ?

ivoanjo: Or that, although I'm not the hugest of fans of skipping an entire set of tests permanently in that way, because it goes in the "it's easy to forget, and it _looks like_ they may be enabled on a quick glance" bin.

sarahchen6: We're using the latest version 5.1.1 for `bson`!

How about I skip JRuby 10.0 in the specs themselves and leave a comment in the `Matrixfile` that JRuby 10.0 is being skipped? This way, we're not skipping all JRuby mongodb tests while also maintaining visibility in the `Matrixfile`.

ivoanjo: Yeah I think that seems an OK compromise

p-datadog:
:+1:

PR #4930, opened 2025-10-02
Run each AppSec integration test in a separate process

y9v:
**What does this PR do?**
This PR changes `rake test:appsec:integration` so that the task runs each test file as a separate process.

**Motivation:**
Rails has a lot of global state, which is quite hard to reset between tests that create different minimal Rails applications.

**Change log entry**

**Additional Notes:**
None.

**How to test the change?**
CI


y9v:
Closing this, since we have a ticket to work on that

PR #4936, opened 2025-10-03
Enable standard for most of the remaining files in the codebase

p-datadog:
**What does this PR do?**

Switches various misc remaining files to standard from rubocop.

**Motivation:**
Continuing migration from rubocop to standard.

**Change log entry**
None

**Additional Notes:**

I will have a separate PR for integration apps since those changes could be not desired: https://github.com/DataDog/dd-trace-rb/pull/4937

**How to test the change?**
Existing tests


benchmarks/support/di_snapshot_target.rb
```diff
@@ -1,3 +1,13 @@
+# Dynamic instrumentation target files must not be modified by
+# formatting tools.
+# standard:disable all
+
+# padding
+# padding
+# padding
+# padding
+# padding
+
```
sarahchen6, line 10: instead of the padding, would it be better to change the line probe number to something like `# Line 24` (and the other relevant lines such as in `di_snapshot.rb`)?

p-datadog: The idea is to add the padding so that the tests do not need to be changed every time there is a change in the target files.

PR #4952, opened 2025-10-07
Remove ruby version check on http request spec

RicardoTrindade:
Cleaning up http_request_spec ruby version checks since CI runs against 2.5+.
Nothing else changed with the PR, all tests should continue to pass. There is still one other place or two that still do these checks that I'd like to cleanup in another PR

PR #4954, opened 2025-10-07
Cleanup instrumentation specs with ruby version check

RicardoTrindade:
Similar to https://github.com/DataDog/dd-trace-rb/pull/4952
Removes Ruby 2.3 version check specs associated with aws instrumentation. Since the gem only supports Ruby 2.5+ I believe these specs can be removed.


p-datadog:
Hi @RicardoTrindade , thank you for the PRs, we will handle CI and merging from here.

vpellan:
[CI run](https://github.com/DataDog/dd-trace-rb/pull/4974)

PR #4958, opened 2025-10-07
RC mutex unlock fix via ensure

p-datadog:
**What does this PR do?**

Alternate implementation of https://github.com/DataDog/dd-trace-rb/pull/4957 that does not use synchronize.

This PR moves the begin-ensure block below the lock call, such that unlocking is only attempted if locking succeeds.

**Motivation:**

Flaky CI 

**Change log entry**

**Additional Notes:**

**How to test the change?**


Strech:
Duplicate #4957

PR #4960, opened 2025-10-08
Add route serializers for Grape and Sinatra apps

y9v:
**What does this PR do?**
This PR adds endpoint collection for Grape and Sinatra applications that are embedded in a Rails app.

**Motivation:**
We want to report routes for embedded Grape APIs and embedded Sinatra apps within Rails.

**Change log entry**

**Additional Notes:**
I had to add `grape` and `sinatra` gems to `rails-app` appraisal for this.

APPSEC-59518

**How to test the change?**
CI and app generator (generic rails app)

### DIFF
--- a/lib/datadog/appsec/contrib/rails/patcher.rb
+++ b/lib/datadog/appsec/contrib/rails/patcher.rb
@@ -154,7 +154,7 @@ module Datadog
 
             GUARD_ROUTES_REPORTING_ONCE_PER_APP[::Rails.application].run do
               AppSec.telemetry.app_endpoints_loaded(
-                APISecurity::EndpointCollection::RailsCollector.new(routes).build_enum
+                APISecurity::EndpointCollection::RailsCollector.new(routes).to_enum
               )
             end
           rescue => e
--- a/lib/datadog/appsec/api_security/endpoint_collection/rails_collector.rb
+++ b/lib/datadog/appsec/api_security/endpoint_collection/rails_collector.rb
@@ -15,7 +15,7 @@ module Datadog
             @routes = routes
           end
 
-          def build_enum
+          def to_enum
             Enumerator.new do |yielder|
               @routes.each do |route|
                 if route.dispatcher?
--- a/sig/datadog/appsec/api_security/endpoint_collection/rails_collector.rbs
+++ b/sig/datadog/appsec/api_security/endpoint_collection/rails_collector.rbs
@@ -7,7 +7,7 @@ module Datadog
 
           def initialize: (Array[_RailsRoute] routes) -> void
 
-          def build_enum: () -> Enumerator[Core::Telemetry::Event::AppEndpointsLoaded::endpoint]
+          def to_enum: () -> Enumerator[Core::Telemetry::Event::AppEndpointsLoaded::endpoint]
 
           private
 
--- a/spec/datadog/appsec/api_security/endpoint_collection/rails_collector_spec.rb
+++ b/spec/datadog/appsec/api_security/endpoint_collection/rails_collector_spec.rb
@@ -4,9 +4,9 @@ require 'spec_helper'
 require 'datadog/appsec/api_security/endpoint_collection/rails_collector'
 
 RSpec.describe Datadog::AppSec::APISecurity::EndpointCollection::RailsCollector do
-  describe '#build_enum' do
+  describe '#to_enum' do
     it 'returns an Enumerator' do
-      expect(described_class.new([]).build_enum).to be_a(Enumerator)
+      expect(described_class.new([]).to_enum).to be_a(Enumerator)
     end
 
     it 'serializes Rails dispatcher routes' do
@@ -18,7 +18,7 @@ RSpec.describe Datadog::AppSec::APISecurity::EndpointCollection::RailsCollector 
       expect(Datadog::AppSec::APISecurity::EndpointCollection::RailsRouteSerializer)
         .to receive(:serialize).and_call_original
 
-      described_class.new([route]).build_enum.first
+      described_class.new([route]).to_enum.first
     end
 
     # Grape and Sinatra routes are tested in endpoint collection integration test,


lib/datadog/appsec/api_security/endpoint_collection/rails_routes_serializer.rb
```diff
@@ -15,9 +18,21 @@ def initialize(routes)
           def to_enum
             Enumerator.new do |yielder|
               @routes.each do |route|
-                next unless route.dispatcher?
+                if route.dispatcher?
+                  yielder.yield serialize_route(route)
+                elsif grape_app?(route.app.rack_app)
```
Strech: I think responsibility-wise the knowledge that you dig inside the route specific path should be encapsulated in route identification method.

lib/datadog/appsec/api_security/endpoint_collection/rails_routes_serializer.rb
```diff
@@ -15,9 +18,21 @@ def initialize(routes)
...
+                  serializer = GrapeRoutesSerializer.new(route.app.rack_app.routes, path_prefix: route.path.spec.to_s)
```
Strech: If you look from the high-level perspective, you may notice that it could be more beneficial to give entire route to the serializer, instead of digging into specific fields.

As a second point to that - class name is "Route Serializer" which implies that a single route will be serialized, but the arguments are not a single route

lib/datadog/appsec/api_security/endpoint_collection/rails_routes_serializer.rb
```diff
@@ -15,9 +18,21 @@ def initialize(routes)
...
+
+                  serializer.to_enum.each do |serialized_route|
+                    yielder.yield serialized_route
+                  end
+                elsif sinatra_app?(route.app.rack_app)
+                  serializer = SinatraRoutesSerializer.new(route.app.rack_app.routes, path_prefix: route.path.spec.to_s)
```
Strech: ⬆️

lib/datadog/appsec/api_security/endpoint_collection/rails_routes_serializer.rb
```diff
@@ -36,6 +51,18 @@ def serialize_route(route)
...
+            return false unless defined?(::Grape::API)
+
+            rack_app.is_a?(Class) && rack_app < ::Grape::API
+          end
+
+          def sinatra_app?(rack_app)
+            return false unless defined?(::Sinatra::Base)
+
+            rack_app.is_a?(Class) && rack_app < ::Sinatra::Base
+          end
```
Strech: I think that knowledge does not belong to the Rails route serializer, but to each specific serializer instead

lib/datadog/appsec/api_security/endpoint_collection/sinatra_routes_serializer.rb
```diff
@@ -0,0 +1,43 @@
...
+            @path_prefix = path_prefix
+          end
+
+          def to_enum
+            Enumerator.new do |yielder|
+              @routes.each do |method, routes|
+                next if method == 'HEAD'
+
+                routes.each do |route, _, _|
+                  yielder.yield serialize_route(route, method)
```
Strech: Just to point out - the order of arguments above is method - routes, and the order here is route - method.

sig/datadog/appsec/api_security/endpoint_collection/grape_routes_serializer.rbs
```diff
@@ -0,0 +1,28 @@
+module Datadog
+  module AppSec
+    module APISecurity
+      module EndpointCollection
+        interface _GrapeRoute
+          def request_method: () -> String
```
Strech: Just in case, the point of `::String` is to use top-level definition in case someone made a mistake of namespacing the String under modules

sig/datadog/appsec/api_security/endpoint_collection/grape_routes_serializer.rbs
```diff
@@ -0,0 +1,28 @@
+module Datadog
+  module AppSec
+    module APISecurity
+      module EndpointCollection
+        interface _GrapeRoute
```
Strech: Should'n it be defined inside the grape route serializer class and without prefix? Or in a separate module? I think it's a confusing place to define out of context interface.

lib/datadog/appsec/api_security/endpoint_collection/rails_routes_serializer.rb
y9v: Yes, but it has a downside - `GrapeRoutesSerializer` and `SinatraRoutesSerializer` can now be used for endpoint collection of stand-alone sinatra apps, and grape routes that are embedded somewhere else, not in rails.

Class name is actually "Routes Serializer", plural.

lib/datadog/appsec/api_security/endpoint_collection/rails_routes_serializer.rb
y9v: I don't agree - it's specific to how rails mounts rack apps, especially if we pass a whole route to this method

lib/datadog/appsec/api_security/endpoint_collection/sinatra_routes_serializer.rb
y9v: the order of arguments above comes from how sinatra stores routing information - it's a hash with the request method as key, and routes for this method as an array.

The `serialize_route` takes route as a first argument because of it's name - I think it would be weirder to have a method named `serialize_route` that takes a method as a first argument, and route as a second?

lib/datadog/appsec/api_security/endpoint_collection/rails_routes_serializer.rb
y9v: but considering that you proposed to move the route identification method to `GrapeRoutesSerializer`, it implies that the grape routes serializer need to know about the structure of Rails routes?

lib/datadog/appsec/api_security/endpoint_collection/rails_routes_serializer.rb
Strech: Ok, can you point the rails-specific parts here?

lib/datadog/appsec/api_security/endpoint_collection/rails_routes_serializer.rb
Strech: True, I didn't notice that it's a routes (plural) collector, that is not related to a specific route, my bad.

lib/datadog/appsec/api_security/endpoint_collection/sinatra_routes_serializer.rb
Strech: Does it makes sense to make method an `kwarg` then? Not sure, but at least positioning is less confusing, but maybe it just bike-shedding

```
serialize_route(route, method: method)
```

lib/datadog/appsec/api_security/endpoint_collection/rails_routes_serializer.rb
Strech: Let me rephrase - I think that serializer of rails, should do serialization and has no knowledge about the Grape or Sinatra and also it should work with a single route.

But because it's working with an array of routes, you have to put that logic in there. Will it be possible to correct serialization and classes to work with a single route (non-rails specific) that could probe it and decide which serializer should be applied?

sig/datadog/appsec/api_security/endpoint_collection/grape_routes_serializer.rbs
y9v: moved those interfaces to `EndpointCollection` module in e207805fb3e539dd9f2f4e4d534137a8ff10618e. The prefix `_` [has to be there unfortunately](https://github.com/ruby/rbs/blob/11bddb4fea521138917e1667c09f7f20f7f2bcc7/docs/syntax.md?plain=1#L29).

lib/datadog/appsec/api_security/endpoint_collection/sinatra_routes_serializer.rb
y9v: reorganized those modules as we discussed here: e207805fb3e539dd9f2f4e4d534137a8ff10618e

lib/datadog/appsec/contrib/rails/patcher.rb
```diff
@@ -154,7 +154,7 @@ def report_routes_via_telemetry(routes)
 
             GUARD_ROUTES_REPORTING_ONCE_PER_APP[::Rails.application].run do
               AppSec.telemetry.app_endpoints_loaded(
-                APISecurity::EndpointCollection::RailsRoutesSerializer.new(routes).to_enum
+                APISecurity::EndpointCollection::RailsCollector.new(routes).build_enum
```
Strech: I think `to_enum` was a good conventional name

PR #4961, opened 2025-10-08
Report actual exception when reporting ThreadErrors

p-datadog:
**What does this PR do?**

Reports actual exception when reporting "deadlock" during configuration processing.

**Motivation:**
In the investigation that led to the fix in https://github.com/DataDog/dd-trace-rb/pull/4957, we discovered that the error was mutex not locked and not a deadlock. This could have been surfaced by the message directly, potentially leading to the issue being fixed sooner and saving us some aggravation due to flaky CI.

**Change log entry**

**Additional Notes:**

**How to test the change?**

Existing tests

### DIFF
--- a/lib/datadog/core/configuration.rb
+++ b/lib/datadog/core/configuration.rb
@@ -238,7 +238,7 @@ module Datadog
           yield write_components
         rescue ThreadError => e
           logger_without_components.error(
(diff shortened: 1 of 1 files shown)


y9v:
Thanks for fixing that!

PR #4962, opened 2025-10-08
Add new gemfiles to lock-dependency commit

marcotc:
This issue was noticed recently (https://github.com/DataDog/dd-trace-rb/pull/4924#discussion_r2414455430) and it caused by a known limitation of [planetscale/ghcommit-action](https://github.com/planetscale/ghcommit-action/issues/43).

Here you can see where our job failed to pick up the newly added gemfile.lock: https://github.com/DataDog/dd-trace-rb/actions/runs/18348438937/job/52262026349#step:7:298

**Change log entry**
No

### DIFF
--- a/.github/workflows/lock-dependency.yml
+++ b/.github/workflows/lock-dependency.yml
@@ -99,6 +99,8 @@ jobs:
           path: gemfiles
           pattern: lock-dependency-${{ github.run_id }}-*
           merge-multiple: true
+      - name: Stage changes # Since ghcommit-action does not pick up unstaged files
+        run: git add gemfiles
       - run: git diff --color=always
       - name: Commit changes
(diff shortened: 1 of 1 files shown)


.github/workflows/lock-dependency.yml
```diff
@@ -99,6 +99,8 @@ jobs:
           path: gemfiles
           pattern: lock-dependency-${{ github.run_id }}-*
           merge-multiple: true
+      - name: Stage changes # Since ghcommit-action does not pick up unstaged files
```
Strech, line 102: I think it should run after `git diff`, otherwise you need to adjust diff to show staged changes

p-datadog: `git diff` will not show staged files, one solution should be `git add -u` after `git add gemfiles` and `git diff --cached`.

Strech: I think a simple order change would do a thing

PR #4966, opened 2025-10-09
Add a guard for missing `AppSec.telemetry` in Endpoint Collection

y9v:
**What does this PR do?**
This PR fixes an exception that might happen when `AppSec.telemetry` is set to `nil`, and we are trying to report Rails routes. As an example, this might happen when the default ruleset file is empty or broken.

**Motivation:**
Failing system tests for `APPSEC_CORRUPTED_RULES` and `APPSEC_MISSING_RULES` scenarios:
https://github.com/DataDog/system-tests-dashboard/actions/runs/18364392471/job/52314703288#step:46:86

**Change log entry**

**Additional Notes:**
APPSEC-59579

**How to test the change?**
CI

### DIFF
--- a/lib/datadog/appsec/contrib/rails/patcher.rb
+++ b/lib/datadog/appsec/contrib/rails/patcher.rb
@@ -149,9 +149,7 @@ module Datadog
             # We do not support Rails 4.x for Endpoint Collection,
             # mainly because the Route#verb was a Regexp before Rails 5.0
             return if target_version < Gem::Version.new('5.0')
-
             return unless Datadog.configuration.appsec.api_security.endpoint_collection.enabled
-
             return unless AppSec.telemetry
 
             GUARD_ROUTES_REPORTING_ONCE_PER_APP[::Rails.application].run do


.github/workflows/system-tests.yml
```diff
@@ -81,7 +81,7 @@ jobs:
         include:
           - weblogs: ""  # All of them
             scenarios: DEFAULT,EVERYTHING_DISABLED,APPSEC_BLOCKING_FULL_DENYLIST,APPSEC_BLOCKING,GRAPHQL_APPSEC,PARAMETRIC
-          - weblogs: rack,rails70
+          - weblogs: rack,rails72
```
y9v, line 84: this was the reason why system-tests did not fail for `APPSEC_CORRUPTED_RULES` and `APPSEC_MISSING_RULES` scenarios

lib/datadog/appsec/contrib/rails/patcher.rb
```diff
@@ -152,13 +152,15 @@ def report_routes_via_telemetry(routes)
 
             return unless Datadog.configuration.appsec.api_security.endpoint_collection.enabled
 
+            return unless AppSec.telemetry
```
Strech, line 153: Small suggestion, could we tie it up, there is no need for extra space between to guard-clauses

y9v: yes, good point

PR #4969, opened 2025-10-10
Add support of Forwarded header for remote IP detection

y9v:
**What does this PR do?**
This PR adds support for `Forwarded` header for remote IP detection.

**Motivation:**
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Forwarded

**Change log entry**
Yes. Tracing: Add `Forwarded` header to the list of headers used for remote IP detection.

**Additional Notes:**
None.

**How to test the change?**
CI

### DIFF
--- a/lib/datadog/core/utils/network.rb
+++ b/lib/datadog/core/utils/network.rb
@@ -73,11 +73,8 @@ module Datadog
 
               next unless value
 
-              ips = if name == 'forwarded'
-                extract_ips_from_forwarded_header(value)
-              else
-                value.split(',')
-              end
+              ips = value.split(',')
+              ips = process_forwarded_header_values(ips) if name == 'forwarded'
 
               ips.each do |ip|
                 parsed_ip = ip_to_ipaddr(ip.strip)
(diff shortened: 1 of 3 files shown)


lib/datadog/core/utils/network.rb
```diff
@@ -72,7 +73,16 @@ def ip_header(headers, ip_headers_to_check)
 
               next unless value
 
-              ips = value.split(',')
+              ips = if name == 'forwarded'
+                value.downcase.split(';').each_with_object([]) do |tuple_str, acc|
+                  next unless tuple_str.start_with?('for=')
+
+                  acc << tuple_str.delete_prefix('for=').strip
```
Strech: I think it makes sense to use bang variants of those functions

* delete_prefix -> delete_prefix!
* strip -> strip!

lib/datadog/core/utils/network.rb
```diff
@@ -72,7 +73,16 @@ def ip_header(headers, ip_headers_to_check)
 
               next unless value
 
-              ips = value.split(',')
+              ips = if name == 'forwarded'
```
Strech: WDYT of extracting forwarded parsing in a function, but keep generic still here (with a comma separation)?

lib/datadog/core/utils/network.rb
y9v: yeah, I will; but those variants both come with a nasty catch in their return value:

> Removes leading and trailing whitespace from str. **Returns nil if str was not altered.**

> Deletes leading prefix from str, **returning nil if no change was made.**

Strech: Yes, but we still can change code

```ruby
tuple_str.delete_prefix!('for=')
tuple_str.strip!

acc.push(tuple_str)
```

lib/datadog/core/utils/network.rb
y9v: good point, extracted in [4250f54](https://github.com/DataDog/dd-trace-rb/pull/4969/commits/4250f548a8db3639910a86577d5c407ab7f35843)

lib/datadog/core/utils/network.rb
y9v: fixed! [4250f54](https://github.com/DataDog/dd-trace-rb/pull/4969/commits/4250f548a8db3639910a86577d5c407ab7f35843)

PR #4976, opened 2025-10-14
[APMAPI-1618] Only show diff in typing stats

vpellan:
**What does this PR do?**

It only show the diff between base branch and current branch for typing stats

**Motivation:**

It will add more value to the typing stats comment

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

When it will be ready, it should display a comment saying there is no changes. You'll be able to see the changes through the edit history of that comment


Strech (approved):
👏🏼 Well done!

PR #4978, opened 2025-10-14
Add a note about which code is instrumentable by dynamic instrumentation

p-datadog:
**What does this PR do?**

Add a note about which code is instrumentable by dynamic instrumentation to documentation.

**Motivation:**
Clarify how to use DI products for our customers.

**Change log entry**
None

**Additional Notes:**

**How to test the change?**

N/A

### DIFF
--- a/docs/DynamicInstrumentation.md
+++ b/docs/DynamicInstrumentation.md
@@ -1,5 +1,34 @@
 # Dynamic Instrumentation
 
+## Instrumentable Code
+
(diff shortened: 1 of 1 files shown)


p-datadog:
@ivoanjo I am not sure yet. There is way more to describe than the existing doc. For now it's meant to be more of a support resource that I could point to should someone ask a question. I think the language/contents might also need to be different for "official" type documentation.

PR #4979, opened 2025-10-14
DI: notify about failed probe condition evaluation, with rate limiting

p-datadog:
**What does this PR do?**
Adds notifications about failed condition evaluation for line and method probes. The notifications are subject to a fixed rate limit of one per second.

**Motivation:**
Feature parity with other languages

**Change log entry**
None (expression language is not yet released to customers)

**Additional Notes:**

**How to test the change?**

Tests are added

### DIFF
--- a/benchmarks/di_instrument.rb
+++ b/benchmarks/di_instrument.rb
@@ -212,10 +212,10 @@ class DIInstrumentBenchmark
     calls = 0
     probe = Datadog::DI::Probe.new(id: 1, type: :log,
       file: targeted_file, line_no: targeted_line + 1)
-    if defined?(Datadog::DI::ProcResponder)
-      rv = instrumenter.hook_line(probe, responder)
+    rv = if defined?(Datadog::DI::ProcResponder)
+      instrumenter.hook_line(probe, responder)
     else
-      rv = instrumenter.hook_line(probe, &executed_proc)
+      instrumenter.hook_line(probe, &executed_proc)
     end
     unless rv
       raise "Line probe (targeted) was not successfully installed"


lib/datadog/di/instrumenter.rb
```diff
@@ -130,10 +126,34 @@ def hook_method(probe, &block)
...
+                  # the probe notifier builder requires a context.
+                  begin
+                    responder.probe_condition_evaluation_failed_callback(context, exc)
+                  rescue
+                    raise if settings.dynamic_instrumentation.internal.propagate_all_exceptions
+
+                    # TODO log / report via telemetry?
+                  end
+                else
+                  _ = 42 # stop standard from wrecking this code
```
y9v, line 149: maybe disabling standard via a comment is better?

lib/datadog/di/proc_responder.rb
```diff
@@ -0,0 +1,32 @@
...
+  module DI
+    # An adapter to convert procs to responders.
+    #
+    # Used in test suite and benchmarks.
+    #
+    # @api private
+    class ProcResponder
+      def initialize(executed_proc, failed_proc = nil)
+        @executed_proc = executed_proc
+        @failed_proc = failed_proc
```
y9v, line 13: maybe `proc_on_failure` would be a more clear name?

lib/datadog/di/instrumenter.rb
p-datadog: It's pretty insistent, do you have a suggestion?

lib/datadog/di/proc_responder.rb
p-datadog: I don't know, "on failure" is now inconsistent with "executed" and it's more letters to type?

PR #4984, opened 2025-10-16
[PROF-12743] Runtime stack collection callback registration

gyuheon0h:
**What does this PR do?**
We want to collect runtime frames for the crashtracker. This approach follows how we access the call stack for profiling. 

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**

Run a ruby program with the crashtracker initialized. Runtime stacks should be visible in the `experimental` section of the crash report.

You can also run this test:
`"Ruby and C method runtime stack capture"`

and see the outputted runtime stacktrace
```
{
  "format": "Datadog Runtime Callback 1.0",
  "frames": [
    {
      "file": "<C extension>",
      "function": "free"
    },
    {
      "file": "test.rb",
      "function": "final_crash_point",
      "line": 248
    },
    {
      "file": "<C extension>",
      "function": "times"
    },
    {
      "file": "test.rb",
      "function": "final_crash_point",
      "line": 247
    },
.....
    {
      "file": "<C extension>",
      "function": "fork"
    },
    {
      "file": "test.rb",
      "function": "<main>",
      "line": 267
    }
  ]
}
```

### DIFF
--- a/ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
+++ b/ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
@@ -4,6 +4,9 @@
 // plumbing required to safely poke at internal structures. Sharing that setup
 // avoids duplicating another native extension with the same (fragile) access
 // patterns, and keeps the overall install/build surface area smaller.
+//
+// This also means that this functionality is depend on the profiling native extension
+// being available and built
 #include "extconf.h"
 
 #if defined(__linux__)
@@ -152,14 +155,7 @@ static void ruby_runtime_stack_callback(
     return;
   }
 
-  bool truncated = frame_count >= RUNTIME_STACK_MAX_FRAMES;
-
   for (int i = frame_count - 1; i >= 0; i--) {
-    if (truncated && i == 0) {
-      emit_placeholder_frame(emit_frame, "<truncated frames>");
-      return;
-    }
-
     frame_info *info = &runtime_stack_buffer[i];
 
     if (info->is_ruby_frame) {
@@ -205,6 +201,10 @@ static void ruby_runtime_stack_callback(
 
       emit_frame(&frame);
     }
+  }
+
+  if (frame_count == RUNTIME_STACK_MAX_FRAMES) {
+    emit_placeholder_frame(emit_frame, "<truncated frames>");
   }
 }
 
--- a/spec/datadog/core/crashtracking/component_spec.rb
+++ b/spec/datadog/core/crashtracking/component_spec.rb
@@ -321,6 +321,10 @@ RSpec.describe Datadog::Core::Crashtracking::Component, skip: !LibdatadogHelpers
       describe 'Ruby and C method runtime stack capture' do
         let(:runtime_stack) { crash_report_experimental[:runtime_stack] }
 
+        before do
+          raise 'This spec requires profiling (native extension not available)' unless Datadog::Profiling.supported?
+        end
+
         it 'captures both Ruby and C method frames in mixed stacks' do
           expect_in_fork(fork_expectations: fork_expectations, timeout_seconds: 15) do
             crash_stack_helper_class = Class.new do
@@ -367,13 +371,13 @@ RSpec.describe Datadog::Core::Crashtracking::Component, skip: !LibdatadogHelpers
 
           expect(frames).to include(
             hash_including(
-              function: 'gsub'
+              function: 'each'
             )
           )
 
           expect(frames).to include(
             hash_including(
-              function: 'each'
+              function: 'gsub'
             )
           )
         end


gyuheon0h:
* **#4984** <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-rb/4984?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/dd-trace-rb/4984?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

gyuheon0h:
Paper trail

[ ] Checking unreasonable string sizes
[ ] Checking that string pointers point to valid strings
[ ] Checking that control frames are valid
[ ] Checking that iseq is valid, and the instruction size is not unreasonable
[ ] Validating that pointers are readable using mincore
[ ] Checking for recursive frames
[ ] Strings can have different representations, take that into account
[ ] Ruby apps commonly have *very deep stacks*. [We default in the profiler to collecting 400](https://datadoghq.atlassian.net/wiki/spaces/PROF/pages/5307007274/Profiling+Libraries+Frame+Truncation+Behaviors#Ruby) and we've seen GitHub go close to 600, handle this
[ ] Pay attention to structure keeping the bytecode-to-line mapping

ext/libdatadog_api/extconf.rb
```diff
@@ -42,6 +42,9 @@ def skip_building_extension!(reason)
 #   (https://github.com/msgpack/msgpack-ruby/blob/18ce08f6d612fe973843c366ac9a0b74c4e50599/ext/msgpack/extconf.rb#L8)
 append_cflags '-std=gnu99'
```
gyuheon0h, line 44: Taken from `ext/datadog_profiling_native_extension/extconf.rb`, so that we can access private Ruby headers

.gitignore
```diff
@@ -57,3 +57,5 @@ lib/*.so
 
 # Benchmark result files
 /benchmarks/*.json
+
```
gyuheon0h: Will be removed when this repo is updated to use libdatadog v24; just for lower friction when local dev rn

ext/libdatadog_api/crashtracker.c
```diff
@@ -1,7 +1,49 @@
-#include <ruby.h>
-#include <datadog/crashtracker.h>
+#include "extconf.h"
```
gyuheon0h: Taken from `ext/datadog_profiling_native_extension/private_vm_api_access.c` so that we can access internal Ruby structures

gyuheon0h:
@codex review

gyuheon0h:
@codex review

lib/datadog/core/crashtracking/component.rb
```diff
@@ -57,6 +57,7 @@ def start
           Utils::AtForkMonkeyPatch.apply!
 
           start_or_update_on_fork(action: :start, tags: tags)
+          register_runtime_stack_callback
```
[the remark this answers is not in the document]
gyuheon0h: Not true, if a callback is already registered, it should replace it

ext/libdatadog_api/datadog_runtime_stack.c
```diff
@@ -0,0 +1,406 @@
...
+            }
+
+            // Get the class/module path
+            if (RB_TYPE_P(actual_owner, T_CLASS) || RB_TYPE_P(actual_owner, T_MODULE)) {
+              owner_name = rb_class_path(actual_owner);
+            }
+
+            // Fallback to rb_class_name if rb_class_path fails
+            if (owner_name == Qnil) {
+              owner_name = rb_class_name(actual_owner);
```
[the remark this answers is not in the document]
gyuheon0h: Papertrail: We are already in a crashing process and agreed with Ivo that this best effort is okay.

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -326,7 +431,7 @@ def build_crashtracker(**options)
     described_class.new(
       agent_base_url: options[:agent_base_url] || 'http://localhost:6006',
       tags: options[:tags] ||
-        {'tag1' => 'value1', 'tag2' => 'value2', 'language' => testing_string, 'service' => testing_string},
+        { 'tag1' => 'value1', 'tag2' => 'value2', 'language' => testing_string, 'service' => testing_string },
```
Strech: I think it's obsolete change

Strech (commented):
Can't comment much at the code, LGTM, just a nano-note 😅

spec/datadog/core/crashtracking/component_spec.rb
gyuheon0h: A lint CI job broke, so I ran the command it told me to and it made this change heh

gyuheon0h:
@codex review

lib/datadog/core/crashtracking/component.rb
```diff
@@ -81,8 +82,29 @@ def stop
...
         private
 
+        def register_runtime_stack_callback
+          # Always use frame-based callback since that's the only type we support
+          success = self.class._native_register_runtime_stack_callback
+
+          unless success
+            error_message = 'Failed to register runtime stack callback: registration returned false (may be already registered or unsupported)'
+            logger.error(error_message)
+            raise StandardError, error_message
```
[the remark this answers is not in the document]
gyuheon0h: Comment was wrong; ignore

.gitignore
ivoanjo: Tip: You can actually add this to `.gitignore_global` on your home folder to avoid having to add this in the PR OR keep this inside of `tmp/`.

It's not amazing to keep temporary "will be removed later" files in PRs unless it's really clear that they are (e.g. by having a really big comment saying "THIS SHOULD BE REMOVED BEFORE MERGING, IT'S ONLY FOR TESTING" as otherwise it's easy for these things to get lost and go in, adding unneeded cruft to the repo.

Rakefile
```diff
@@ -229,7 +229,12 @@ namespace :spec do
     t.pattern = CORE_WITH_LIBDATADOG_API.join(', ')
     t.rspec_opts = args.to_a.join(' ')
   end.tap do |t|
-    Rake::Task[t.name].enhance(["compile:libdatadog_api.#{RUBY_VERSION[/\d+.\d+/]}_#{RUBY_PLATFORM}"])
+    Rake::Task[t.name].enhance(
+      [
+        "compile:libdatadog_api.#{RUBY_VERSION[/\d+.\d+/]}_#{RUBY_PLATFORM}",
+        "compile:datadog_profiling_native_extension.#{RUBY_VERSION}_#{RUBY_PLATFORM}"
```
ivoanjo, line 238: To someone coming after we merge this PR, this looks like it can be a bit confusing -- "why is core with libdatadog depending on profiling again?" so I recommend adding here a comment explaining why this is needed.

lib/datadog/core.rb
```diff
@@ -20,6 +20,14 @@ module Core
...
         e.message
       end
+
+    RUNTIME_STACKS_FAILURE =
+      begin
+        require 'datadog/runtime_stacks'
+        nil
+      rescue LoadError => e
+        e.message
+      end
```
ivoanjo: Since this is only used/needed for the crashtracking component I suggest moving this to that file -- no need to "publish" something at the top-level of the library, since this is effectively an implementation detail of the crashtracker

(same for `datadog/runtime_stacks.rb`)

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,411 @@
...
+    "_native_register_runtime_stack_callback",
+    _native_register_runtime_stack_callback,
+    0
+  );
+  rb_define_singleton_method(
+    runtime_stacks_class,
+    "_native_is_runtime_callback_registered",
+    _native_is_runtime_callback_registered,
+    0
+  );
```
ivoanjo: Similar to my suggestion on `core.rb`, I suggest scoping this inside crashtracking, and not top-level on the datadog module, since this is an implementation detail of the crashtracker that no other part of the system needs to know or care about.

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,411 @@
+#include "extconf.h"
```
ivoanjo, line 10: So we both know the reason why this file lives inside the profiling extension, or what it even does but... since most of our discussion so far was over private slack, nobody else will know.

Thus, I think this deserves a big comment explaining:

* What this is
* Why it ended up living here
* How it doesn't interact with profiling
* Why does it mirror somethings in private_vm_api_access but not others

Also "runtime_stacks" is not very obvious -- may I suggest `crashtracking_runtime_stacks.c` or something among those lines that makes it very clear that this is for crashtracking

lib/datadog/core/crashtracking/component.rb
```diff
@@ -57,6 +57,9 @@ def start
           Utils::AtForkMonkeyPatch.apply!
 
           start_or_update_on_fork(action: :start, tags: tags)
+          if DATADOG_ENV['DD_CRASHTRACKER_EMIT_RUNTIME_STACKS'] == 'true'
```
ivoanjo: What's the intention behind this setting? E.g. is this something that customers will be able to toggle on an off, or just something for temporary debugging?

Depending on the answer, we'll probably need a few adjustments...

lib/datadog/core/crashtracking/component.rb
```diff
@@ -81,8 +84,42 @@ def stop
...
 
+        def runtime_callback_registered?
+          return false if Datadog::Core::RUNTIME_STACKS_FAILURE
+          return false unless Datadog.const_defined?(:RuntimeStacks, false)
+
+          Datadog::RuntimeStacks._native_is_runtime_callback_registered
+        rescue => e
+          logger.debug("Runtime stack callback status check not available: #{e.message}")
+          false
+        end
```
ivoanjo: If this exists only for testing, I recommend moving to the testing code directly (e.g. the less code we declare "is for production", the better)

lib/datadog/core/crashtracking/component.rb
```diff
@@ -81,8 +84,42 @@ def stop
...
+
+          unless success
+            error_message = 'Failed to register runtime stack callback: registration returned false'
+            logger.error(error_message)
+            raise StandardError, error_message
+          end
+        rescue => e
+          logger.error("Failed to register runtime stack callback: #{e.message}")
+          raise
+        end
```
ivoanjo: This pattern is a bit weird -- if `success == false` we log an error, then turn it into an exception, then immediately log it again, at error level.

Something logged at error level will show up quite noisly in customer's logs so for something like crashtracking where if it doesn't work the impact is very low, I'm not sure "alarming" customers in this way is worth it...

lib/datadog/core/crashtracking/component.rb
```diff
@@ -57,6 +57,9 @@ def start
           Utils::AtForkMonkeyPatch.apply!
 
           start_or_update_on_fork(action: :start, tags: tags)
+          if DATADOG_ENV['DD_CRASHTRACKER_EMIT_RUNTIME_STACKS'] == 'true'
+            register_runtime_stack_callback
```
ivoanjo: Minor: Do we maybe want to register the callback _before_ we start the crashtracker?

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -318,6 +351,105 @@
...
+                {a: 1, b: 2}.each do |key, value|
+                  if key == :a
+                    Fiddle.free(42)
+                  end
+                end
+              end
+            end
+
+            top_level_ruby_method
+          RUBY
```
ivoanjo: So I understand that yeah when we use `expect_in_fork` we'll have rspec frames on the stack. But why is that a problem? E.g. the profiler specs happily assert on rspec being on the stack with no issue.

In particular, doing it this way is a bit of extra work so I think simplifying can give us the same level of coverage without as much complexity on the test.

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,411 @@
...
+      }
+
+      if (e == EAGAIN && retries-- > 0) {
+          continue;
+      }
+
+      // Unknown errno, we assume mapped to avoid cascading faults in crash path
+      return true;
+  }
+}
```
ivoanjo, line 86: So this part at all doesn't seem to be Ruby-specific... Would it make sense for this to live inside libdatadog? 👀

ext/datadog_profiling_native_extension/runtime_stacks.c
```diff
@@ -0,0 +1,411 @@
...
+#include "private_vm_api_access.h"
+#include <sys/mman.h>
+#include <unistd.h>
+#include <errno.h>
+#include <string.h>
+
+// This was renamed in Ruby 3.2
+#if !defined(ccan_list_for_each) && defined(list_for_each)
+  #define ccan_list_for_each list_for_each
+#endif
```
ivoanjo: I don't think this is used anywhere ;)

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,411 @@
...
+  for (long i = 0; i < len && i < 128; i++) {
+    unsigned char c = (unsigned char)ptr[i];
+
+    if (c == 0 && i < len - 1) return false;
+
+    // Control characters (except tab, newline, return) is sus
+    if (c < 0x20 && c != 0x09 && c != 0x0A && c != 0x0D) return false;
+
+    // High bytes
+    if (c >= 0xF8) return false;
```
ivoanjo: Hmmm this is slightly sus... or AI...?

E.g. I'm pretty sure < 0x20 also includes... 0x09 and 0x0A and 0x0D...?

Also, why 0xF8? I'm guessing it may be something with utf-8 string validation but I think the code needs a bit of commenting to help me understand why these validations are correct

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,411 @@
...
+
+static bool is_reasonable_string_size(VALUE str) {
+  if (str == Qnil) return false;
+  if (!RB_TYPE_P(str, T_STRING)) return false;
+
+  // Check if the heap object pointed to by str is readable
+  if (!is_pointer_readable((const void *)str, sizeof(struct RBasic))) return false;
+
+  // For strings, we need to check the full RString structure
+  if (!is_pointer_readable(RSTRING(str), sizeof(struct RString))) return false;
```
ivoanjo: Is this AI...? This seems a redundant check...?

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,411 @@
...
+  if ((void*)cfp < stack_start || (void*)cfp >= stack_end) {
+    return false;
+  }
+
+  if (!is_pointer_readable(cfp, sizeof(rb_control_frame_t))) {
+    return false;
+  }
+
+  return true;
+}
```
ivoanjo: We're validating some pointers, but not others... Why?

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,411 @@
...
+
+  // Validate iseq size
+  if (iseq->body->iseq_size > 100000) return false; // > 100K instructions, suspicious
+
+  return true;
+}
+
+static void ruby_runtime_stack_callback(
+  void (*emit_frame)(const ddog_crasht_RuntimeStackFrame*)
+) {
```
ivoanjo, line 133: So this function is very complex, because it also does a very complex thing. My question is -- what happens if we get something wrong here? That is, we already have a bunch of safeguards, but what can actually happen if we miss a spot in pratice? Is there a safety net for us, or not?

ext/datadog_profiling_native_extension/runtime_stacks.c
```diff
@@ -0,0 +1,411 @@
...
+  if (current_thread == Qnil) return;
+
+  static const rb_data_type_t *thread_data_type = NULL;
+  if (thread_data_type == NULL) {
+    thread_data_type = RTYPEDDATA_TYPE(current_thread);
+    if (!thread_data_type) return;
+  }
+
+  rb_thread_t *th = (rb_thread_t *) rb_check_typeddata(current_thread, thread_data_type);
+  if (!th) return;
```
ivoanjo: This check doesn't make a lot of sense -- on the profiling side we actually get this info initially from a "known good" thread, and then we check that other threads we get match this info.

But in this version, we're just checking if the current thread matches itself... which it always will. So this doesn't actually check anything, that I can see?

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,411 @@
...
+        continue;
+      }
+
+      VALUE name = rb_iseq_base_label(iseq);
+      const char *function_name = "<unknown>";
+      if (name != Qnil) {
+        function_name = safe_string_ptr(name);
+      }
+
+      VALUE filename = rb_iseq_path(iseq);
```
ivoanjo: So this slightly makes me nervous, and comes back to my question of "do we have a safety net for this or not". We're asking the VM to get something for us and thus we have no control over valid pointers and the VM doing sane things if something is wrong...

ivoanjo (commented):
Gave it a big pass! I think in general this seems reasonable -- most of my thoughts/worries are around `ruby_runtime_stack_callback` and making sure we get it into shape.

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
gyuheon0h: This might be due to my unfamiliarity with Ruby internals, but the first check is intended to ensure the VALUE actually points to a live struct RBasic. Without the first check, even touching RSTRING(str) would dereference garbage and crash.The second check validates the string-specific payload (struct RString) since Ruby string’s header extends RBasic with extra fields.

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
gyuheon0h: Yeah, 0x09 and 0x0A and 0x0D are included < 0x20 because we are manually whitelisting them, but thinking of it again, chars like tab, newline, return wont exist so we prob don't need this check at all; should just block all control chars. 

I used 0xF8 because utf8 never legally uses bytes 0xF8–0xFF; but I will improve this check and add comments to make this clear; probably reject anything >= 0xF5.

Anyways, ill implement a more comprehensive, safe string checker

lib/datadog/core/crashtracking/component.rb
gyuheon0h: Its not necessary per se but I would like a kill switch just in case something goes wrong. Just noticed that the convention is to use `DD_CRASHTRACKING_*` prefix, updated to do so

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
ivoanjo: Ah... In fact `RSTRING(...)` is just a cast so that's why I was confused:

```c
/**
 * Convenient casting macro.
 *
 * @param   obj  An object, which is in fact an ::RString.
 * @return  The passed object casted to ::RString.
 */
#define RSTRING(obj)            RBIMPL_CAST((struct RString *)(obj))
```

but TBH it's probably best to avoid as much as possible Ruby macros and other calls here to avoid relying on too many assumptions of what they do.

That is -- this should probably be a `is_pointer_readable((const void *)str, sizeof(struct RString))` and thus avoid the reliance on Ruby headers other than to know "what's the size of it".

lib/datadog/core/crashtracking/component.rb
ivoanjo: Hmm... I'm a bit split on this one.

If we keep it, I do think we probably need to add it to the settings system as just another setting -- it's weird/inconsistent if we have it only via environment variable, as often our customers configure things via the settings system and not via environment.

(Environment also has the weird property of being automatically inherited by other called processes, which the settings system doesn't do).

For the "in case something goes wrong" I'm entirely sure it's worth it, TBH.

While we've sometimes added toggles for newer things for this purpose, again for me I see this whole thing as an extension of crashtracking, which is almost an implementation detail -- it's just a slightly "noisy" implementation detail because of what it needs to do.

But, if something does go wrong, we are able to disable crashtracking and/or even stop the extensions from building, so there's already at least two levers that we can use to turn this off (albeit slightly indirectly). So... I'm not sure we need to have "turn this very small detail off" for the "just in case" use-case.

gyuheon0h: >it's weird/inconsistent if we have it only via environment variable, as often our customers configure things via the settings system and not via environment.

Yup this makes sense. I did something similar thing with Python atm, but I need to wait until post code freeze to add to config registry IIRC

>So... I'm not sure we need to have "turn this very small detail off" for the "just in case" use-case.

This also makes sense. On the Python side, we went forward with having a flag (To be soon a part of system settings), but I see your point; adding a "just in case" flag for something that can already be guarded feels unnecessary and noisy. I'd like to align across runtimes however, so I will get back to you on this.

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
gyuheon0h: Spinning back on this, we don't need this validation. We will emit whatever string is present, and libdatadog handles safe utf8 conversion.

lib/datadog/core/crashtracking/component.rb
gyuheon0h: I'll go ahead with no flag for Ruby!

gyuheon0h:
@codex review

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
gyuheon0h: Updated to be cleaner, with more guards, but at a high level, the purpose of this helper is to  to prove the control frame itself lives inside the current thread’s stack slab and that the `rb_control_frame_t` structure is readable. Once we pass that check, every field we actually dereference later is gated separately:

ext/datadog_profiling_native_extension/datadog_ruby_common.h
```diff
@@ -41,6 +41,13 @@ static inline ddog_CharSlice char_slice_from_ruby_string(VALUE string) {
   return char_slice;
 }
 
+static inline ddog_CharSlice char_slice_from_cstr(const char *cstr) {
```
ivoanjo: Since this is only used by `crashtracking_runtime_stacks.c`, I recommend moving it there and not adding it to the common utils.

In particular, the objective of the common utils is _code sharing_ but we need to be careful for it to not become the "misc stuff drawer" that many people have at home ;)

So once this becomes used elsewhere, we can move it here, but until then, I'd suggest avoiding.

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.h
ivoanjo, line 1: Minor: Since only the `init` prototype is needed, I'd suggest adding the prototype directly to `profiling.c` and thus avoiding the need for _yet one more file_.

(Because we're not using subfolders in the extension, avoiding too many small files is kinda nice?)

lib/datadog/core/crashtracking/component.rb
```diff
@@ -9,6 +9,14 @@
 module Datadog
   module Core
     module Crashtracking
+      RUNTIME_STACKS_FAILURE =
+        begin
+          require_relative 'crashtracking_runtime_stacks'
+          nil
+        rescue LoadError => e
+          e.message
+        end
```
ivoanjo: I'm not the biggest fan of this... Specifically -- it creates a new situation in the application where the profiling native extension is loaded (including many of its classes), but the ruby-level code of profiling isn't loaded.

The `profiling/load_native_extension` file was meant to be used by profiling, not for things outside of profiling.

My suggestion would be to instead use `Datadog::Profiling.supported?` / `Datadog::Profiling.unsupported_reason`. These are the public APIs that profiling exposes to the outside, and that let profiling still maintain control of loading of the extension.

lib/datadog/core/crashtracking/component.rb
```diff
@@ -83,6 +92,19 @@ def stop
...
+
+          # Even if runtime stacks callback registration fails, we shouldn't block crashtracker
+          success = Crashtracking::RuntimeStacks._native_register_runtime_stack_callback
+
+          return if success
+
+          logger.debug('Failed to register runtime stack callback: registration returned false')
+        rescue => e
+          logger.debug("Failed to register runtime stack callback: #{e.message}")
+        end
```
ivoanjo: Minor: I think the `RuntimeStacks` class doesn't particularly add a lot at this point -- maybe just hook the native methods onto `Crashtracking::Component`?

Steepfile
```diff
@@ -110,6 +110,7 @@ target :datadog do
   ignore 'lib/datadog/core/configuration/ext.rb'
   ignore 'lib/datadog/core/configuration/settings.rb'
   ignore 'lib/datadog/core/contrib/rails/utils.rb'
+  ignore 'lib/datadog/core/crashtracking/component.rb'
```
ivoanjo: I'm curious, was keeping the types here too annoying? The functions we're using seem reasonably simple 👀

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -318,6 +375,82 @@
...
+      it 'returns false and logs debug when native method raises exception' do
+        crashtracker = build_crashtracker(logger: logger)
+        error = StandardError.new('Native error')
+
+        expect(Datadog::Core::Crashtracking::RuntimeStacks).to receive(:_native_is_runtime_callback_registered).and_raise(error)
+        expect(logger).to receive(:debug).with('Runtime stack callback status check not available: Native error')
+
+        expect(crashtracker.runtime_callback_registered?).to be false
+      end
     end
```
ivoanjo: I don't quite understand the need for `runtime_callback_registered?`... 

It's being added for this file only, and then we're testing it here but... we're using it nowhere? It seems like we could get rid of it instead?

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -198,25 +255,25 @@
       let(:parsed_request) { JSON.parse(request.body, symbolize_names: true) }
       let(:crash_report) { parsed_request.fetch(:payload).first }
       let(:crash_report_message) { JSON.parse(crash_report.fetch(:message), symbolize_names: true) }
+      let(:crash_report_experimental) { crash_report_message.fetch(:experimental) }
+      let(:log_messages) { crash_report_message.fetch(:log_messages) }
```
ivoanjo: Minor: This is not actually used anywhere? (Maybe remnant of earlier version?)

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -318,6 +375,82 @@
...
+
+          frames = runtime_stack[:frames]
+          # Check that the crashing function is captured
+          expect(frames).to include(
+            hash_including(
+              function: 'free'
+            )
+          )
+
+          puts frames.to_json
```
ivoanjo: Remember to clean this up ;)

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -318,6 +375,82 @@
...
+
+      describe 'Ruby and C method runtime stack capture' do
+        let(:runtime_stack) { crash_report_experimental[:runtime_stack] }
+
+        it 'captures both Ruby and C method frames in mixed stacks' do
+          expect_in_fork(fork_expectations: fork_expectations, timeout_seconds: 15) do
+            crash_tracker = build_crashtracker(agent_base_url: agent_base_url)
+            crash_tracker.start
+
+            ComponentSpecCrashStackHelper.new.top_level_ruby_method
```
ivoanjo: Minor: Since the `ComponentSpecCrashStackHelper` is only needed here, I suggest inlining it here. That is, keep it as-is (that's fine), just move all of the class definition to inside `expect_in_fork`, and then call it. Ruby doesn't mind, I promise ;)

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -1,9 +1,43 @@
...
 
+class ComponentSpecCrashStackHelper
+  def top_level_ruby_method
+    ruby_method_with_c_calls
+  end
+
+  def ruby_method_with_c_calls
+    'hello world'.gsub("world") do |_match|
+      {a: 1, b: 2}.each do |key, _value|
+        Fiddle.free(42) if key == :a
```
ivoanjo: Minor: Small simplification ;)

```suggestion
        Fiddle.free(42)
```

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,503 @@
...
+  VALUE core_module = rb_define_module_under(datadog_module, "Core");
+  VALUE crashtracking_module = rb_define_module_under(core_module, "Crashtracking");
+  VALUE runtime_stacks_class = rb_define_class_under(crashtracking_module, "RuntimeStacks", rb_cObject);
+
+  rb_define_singleton_method(
+    runtime_stacks_class,
+    "_native_register_runtime_stack_callback",
+    _native_register_runtime_stack_callback,
+    0
+  );
```
ivoanjo: So meta-question that occurred to me about this. Why is `crashtracking_runtime_stacks_init` separate from `_native_register_runtime_stack_callback`?

That is, do these need to live as separate things, or could we wire this up so that loading this file sets up the runtime_stack_callback automatically and then the Ruby code doesn't even have to worry about it?

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
ivoanjo: Hey I don't think this one got answered? (It was marked as resolved?)

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,503 @@
...
+  emit_frame(&frame);
+}
+
+// Walk the Ruby control-frame stack for the crashing thread, emit frames in order, and fall back
+// to placeholder markers when corruption is detected so that the crash report still completes.
+// We intentionally keep using Ruby helper APIs (rb_vm_frame_method_entry, rb_iseq_line_no, etc.)
+// here: crashtracker runs only once per crash, the collector tolerates callback faults, and the
+// worst outcome of a bad helper call is a missing runtime stack. We choose this because we prefer readable,
+// idiomatic integration with VM helpers instead of reimplementing a big chunk of Ruby internal APIs.
+static void ruby_runtime_stack_callback(
```
ivoanjo, line 131: Because we have a safety net and can rely on some VM behavior... At this point, since this method is an inlined version of `rb_profile_frames` or its cousin `ddtrace_rb_profile_frames`, do you see any issue with using those directly here?

Since that's what Ruby ends up kinda doing for its own crash reporter, it seems that would save us from having to maintain here a copy of all of this thorny logic... 🤔

lib/datadog/core/crashtracking/component.rb
gyuheon0h: Good idea!

spec/datadog/core/crashtracking/component_spec.rb
gyuheon0h: Hmmm I guess the original intention was to verify registration; but (1) we should be confident that registration implementation is sound which means (2) we don't need this

spec/datadog/core/crashtracking/component_spec.rb
gyuheon0h: `crash_report_experimental` is used; but thanks for the catch with log_messages

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
gyuheon0h: If we include runtime callback registration into the init function, then runtime callback registration will be triggered every time the profiling extension is loaded. If we/customer disables crashtracking, then runtime callback registration will still happen which is not necessarily atomic. WDTY?

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
gyuheon0h: Oops, sorry, meant to reply to this. Yeah, this can be something added to `libdatadog`. I would prefer to keep this in this PR, then separately include this functionality in `libdatadog` and update this when we make a new `libdatadog` release and bump the version within `dd-trace-rb`. 

Basically, I'd rather not have this PR wait on making a new libdatadog release and bumping the version here.

gyuheon0h:
>Hey I don't think this one got answered? (It was marked as resolved?)

Oops sorry, I responded

lib/datadog/core/crashtracking/component.rb
gyuheon0h: Yeah makes sense. I can't call that at the top level to set RUNTIME_STACKS_FAILURE since doing a top level load will pull profiling.rb (and thus core) while core is still loading and cause circular require issues. I just removed it and call the public profiling API when necessary, lazily.

I have to update a guard in [telemetry spec](https://github.com/DataDog/dd-trace-rb/blob/4bae22823e981ee45fdfaa58a4244cb6777d0ed1/spec/datadog/core/telemetry/integration/telemetry_spec.rb#L86-L94) because `expect` -> `to_receive` is a strict check for one method call and now, we are calling `Datadog::Profiling.unsupported_reason` more than once

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
gyuheon0h: I’d stick with the current emit-as-you-walk approach because `ddtrace_rb_profile_frames`/`rb_profile_frames` Even with crashtracker’s safety nets, that means:
1. We would have to allocate a sizeable buffer up front within the signal handler (or reuse one) just to turn around and iterate it; extra work and extra memory in the crash path.
2. Spending more time in the signal handler to fill the buffer and then emit, instead of emitting frames immediately as we discover them.
3. We won't be able to do partial frame emission

Even though we do have safety nets, I still want to increase the probability of a successful runtime stack collection

gyuheon0h: >extra memory in the crash path

Thought about this, and we can just preallocate a buffer on registration; not within crash context, but still would be extra work within the signal handler

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
ivoanjo: > If we include runtime callback registration into the init function, then runtime callback registration will be triggered every time the profiling extension is loaded. 

Indeed! I think that's fine? E.g. since exactly we still have the big lever which is enabling/disabling crashtracking, it seems both useful (in terms of less code) and harmless (in terms of I can't think of a strong reason to avoid it) to always set it up here.

> If we/customer disables crashtracking, then runtime callback registration will still happen which is not necessarily atomic. WDTY?

Can you clarify what you mean here be "not necessarily atomic"? E.g., what's the worst thing that happen if crashtracking is disabled? :thinking:

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
ivoanjo: Fair enough! Do you mind leaving a comment in the code (TODO: We plan to move this to libdatadog....?). 

(At least that way, even if we don't do it yet, we leave a hint for future generations/PRs :D )

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
ivoanjo: So here's why I'm kinda concerned about this:

We have inlined here a lot of very specific logic from the VM, with only very light validation of it. If you look at the stack collector spec, it has almost 1000 lines of tests exercising our version of `ddtrace_rb_profile_frames` and comparing it very strictly to what Ruby is outputting.

This version imports things that `ddtrace_rb_profile_frames` doesn't, and additionally inlines/refactors a lot of it, and doesn't have a lot of test coverage so it's extremely hard to a) validate that it's complete and matches what the VM does; and b) that it remains correct, especially in corner cases.

For instance consider this: if we give this file and we ask a 3rd colleague to make sure it's correct (not just "looks correct"). What would they need to do? They would need to effectively find all of the Ruby logic in the VM that went into making this, and then examine if we inlined it correctly, and well, that it will be fine from Ruby 2.5 to 4.0, and then consider the simplifications and other small tweaks we're doing.

(And the profiler's code paths get executed hundreds of times per second, so there's a bit of a "if there's a crash or a bug, it'll show up somewhere, due to sheer bruteforcing of how often this code runs". Vs this code that will be executed rarely, in comparison).

So TBH I know we've invested in this thus far, but I'm not sure optimizing for "signal handler does more work" vs maintainability is the right thing here.

lib/datadog/core/crashtracking/component.rb
```diff
@@ -99,6 +116,20 @@ def start_or_update_on_fork(action:, tags:)
...
           logger.error("Failed to #{action} crash tracking: #{e.message}")
         end
+
+        def runtime_stacks_supported?
+          @runtime_stacks_failure ||= begin
+            require 'datadog/profiling'
+            Datadog::Profiling.supported? ? nil : "profiling not supported: #{Datadog::Profiling.unsupported_reason}"
+          rescue => e
+            e.message
+          end
```
ivoanjo: Minor: the `rescue` bit is not needed -- the profiler already handles failures that may happen in those methods

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
gyuheon0h: By "not atomic" I meant that registering a callback function here is a crashtracking feature, and logically it feels like it should only happen when crashtracking is on. 

However, there should be no negative side effects of a runtime callback being registered even if crashtracking is turned off; it'll just never be triggered. 

Sounds better to just always init, so that Ruby does not have to worry about it

lib/datadog/core/crashtracking/component.rb
gyuheon0h: Yeah thanks for this catch. This whole functionality is not needed if we just register on profiling extension init though. Updated to not have Ruby code deal with anything runtime callback related

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
gyuheon0h: Hmmm okay yeah; that makes sense. In the worst case, we can examine metrics around how well this performs and can revist. Updated to use the profiling api!

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
ivoanjo: :+1: I agree it's slightly weird at first to register if it's not going to be used (e.g. it maybe deserves a comment for posterity in the code ;D )

lib/datadog/core/crashtracking/component.rb
ivoanjo: Simplification enables more simplification! ;D

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,261 @@
...
+  void (*emit_frame)(const ddog_crasht_RuntimeStackFrame*)
+) {
+  // Grab the Ruby thread we crashed on; crashtracker only runs once.
+  VALUE current_thread = rb_thread_current();
+  if (current_thread == Qnil) return;
+
+  if (crashtracker_thread_data_type == NULL) return;
+
+  rb_thread_t *th = (rb_thread_t *) rb_check_typeddata(current_thread, crashtracker_thread_data_type);
```
ivoanjo: This is not quite correct, actually. Specifically -- if `current_thread` is not a thread, this will attempt to raise a ruby-level exception which is not got at all here.

I actually was curious to see what happened so I did this to make it unhappy:

```diff
@@ -65,6 +65,8 @@ static frame_info runtime_stack_buffer[512];
 
 long syscall(long number, ...);
 
+static VALUE array;
+
 // align down to power of two
 static inline uintptr_t align_down(uintptr_t x, uintptr_t align) {
   return x & ~(align - 1u);
@@ -175,7 +177,7 @@ static void ruby_runtime_stack_callback(
 
   if (crashtracker_thread_data_type == NULL) return;
 
-  rb_thread_t *th = (rb_thread_t *) rb_check_typeddata(current_thread, crashtracker_thread_data_type);
+  rb_thread_t *th = (rb_thread_t *) rb_check_typeddata(array, crashtracker_thread_data_type);
   if (!th || !is_pointer_readable(th, sizeof(*th))) return;
 
   // Use the profiling helper to gather frames into our static buffer.
@@ -245,6 +247,9 @@ static void ruby_runtime_stack_callback(
 }
 
 void crashtracking_runtime_stacks_init(void) {
+  rb_global_variable(&array);
+  array = rb_ary_new();
+
   if (crashtracker_thread_data_type == NULL) {
     VALUE current_thread = rb_thread_current();
     if (current_thread == Qnil) return;
@@ -258,4 +263,3 @@ void crashtracking_runtime_stacks_init(void) {
   // Register immediately so Ruby doesn't need to manage this explicitly.
   ddog_crasht_register_runtime_frame_callback(ruby_runtime_stack_callback);
 }
-
```

and it seems like in practice the crashtracker times out.

TL;DR I suggest maybe introducing a `is_thread(VALUE)` to `private_vm_api_access.c` (refactoring the code from `thread_struct_from_object` to be shared maybe?) that uses the `rb_typeddata_is_kind_of` instead and can be used here.

Why add to `private_vm_api_access.c` instead of here? I think we're very close to not needing any private headers in this file, and at this point we might as well take the advantages of code sharing with the profiler.

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,261 @@
...
+static const rb_data_type_t *crashtracker_thread_data_type = NULL;
+
+static void ruby_runtime_stack_callback(
+  void (*emit_frame)(const ddog_crasht_RuntimeStackFrame*)
+);
+
+// Use a fixed, preallocated buffer for crash-time runtime stacks to avoid
+// heap allocation in the signal/crash path.
+static const int RUNTIME_STACK_MAX_FRAMES = 512;
+static frame_info runtime_stack_buffer[512];
```
ivoanjo: Minor: I recommend making `RUNTIME_STACK_MAX_FRAMES` a `#define` and thus we can use it to initialize the `runtime_stack_buffer` instead of repeating the 512.

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
ivoanjo: Minor: I recommend actually calling it out as a "TODO" so it jumps even more out

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,261 @@
...
+
+  long len = RSTRING_LEN(str);
+  const char *ptr = RSTRING_PTR(str);
+
+  if (!ptr) return "<null>";
+
+  if (!is_pointer_readable(ptr, len > 0 ? len : 1)) return "<unreadable>";
+
+  return ptr;
+}
```
ivoanjo, line 125: I suggest combining all three of these functions into one. Specifically:

* We can unify the `strlen` and `RSTRING_LEN` things -- all 3 functions want to do things with the length
* `is_reasonable_string_size` has comments that mention the work that `safe_string_ptr` does
* There's repeated `Qnil` and `NULL` checks

And even some things are out of order -- `RB_TYPE_P(str, T_STRING)` actually reads the header before `if (!is_pointer_readable((const void *)str, sizeof(struct RBasic))) return false;` so if the string is not readable we've already done something wrong.

Having one unified function I think would solve these duplicated sharp edges.

(I know `char_slice_from_cstr` is also being used for a few literals -- those can use `DDOG_CHARSLICE_C` instead)

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,261 @@
...
+  // Number of pages spanned
+  size_t pages = 1u + (last != start);
+  if (pages > 2u) pages = 2u;
+
+  unsigned char vec[2];
+
+  int retries = 5;
+  for (;;) {
+      size_t len = pages * (size_t)page_size;
+      long rc = syscall(SYS_MINCORE, (void*)start, len, vec);
```
ivoanjo, line 68: It occurred to me that these defines/`syscall` are Linux-only, and will break on macOS.

Although we don't officially support macOS for the profiler, there's colleagues (@\AlexJF, @\marcotc that do develop on macOS) and so we try to make sure we still compile and generally don't break there.

TL;DR: We should ifdef-out this whole file using a `#if defined(__linux__)` or something similar.

(I'm actually not sure if the crashtracker works on macOS, I don't think it does, so as long as `bundle exec rake compile` does not fail that should be _good enough_)

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,261 @@
...
+
+  // Use the profiling helper to gather frames into our static buffer.
+  int frame_count = ddtrace_rb_profile_frames(
+    current_thread,
+    0,
+    RUNTIME_STACK_MAX_FRAMES,
+    runtime_stack_buffer
+  );
+
+  if (frame_count <= 0) return;
```
ivoanjo: I wonder if we should add a placeholder frame here, e.g. "<No frames>" or something, so we'll know if we're missing the data because of this branch.

Also, perhaps we could also copy the logic on `collectors_stack.c` for the truncated placeholder (copied here for reference -- it's basically a size check + replace the last frame)

```c
  if (captured_frames == (long) buffer->max_frames) {
    add_truncated_frames_placeholder(buffer);
  }

static void add_truncated_frames_placeholder(sampling_buffer* buffer) {
  buffer->locations[0] = (ddog_prof_Location) {
    .mapping = {.filename = DDOG_CHARSLICE_C(""), .build_id = DDOG_CHARSLICE_C(""), .build_id_id = {}},
    .function = {.name = DDOG_CHARSLICE_C("Truncated Frames"), .filename = DDOG_CHARSLICE_C(""), .filename_id = {}},
    .line = 0,
  };
}
```

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,261 @@
...
+
+  for (int i = frame_count - 1; i >= 0; i--) {
+    frame_info *info = &runtime_stack_buffer[i];
+
+    if (info->is_ruby_frame) {
+      const rb_iseq_t *iseq = (const rb_iseq_t *)info->as.ruby_frame.iseq;
+      const char *function_name = "<unknown>";
+      const char *file_name = "<unknown>";
+
+      if (iseq && is_pointer_readable(iseq, sizeof(rb_iseq_t))) {
```
ivoanjo: Together with my note on getting rid of `rb_thread_t` above, it looks to me that knowing about `rb_iseq_t` is the last thing on this file that needs the internal headers.

But here we actually only need to know about `rb_iseq_t` for the `sizeof` check. Thus, my suggestion is -- add a `size_of_rb_iseq_t()` function to `private_vm_api_access.c` and use it here, so that we can rid this file of having to care about the internal headers.

E.g. this would become something like

```suggestion
      const void *iseq = (const void *)info->as.ruby_frame.iseq;
      const char *function_name = "<unknown>";
      const char *file_name = "<unknown>";

      if (iseq && is_pointer_readable(iseq, size_of_rb_iseq_t())) {
```

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
gyuheon0h: Awesome ,thanks for this catch. 

I'm not sure we can have it as a helper in `private_vm_api_access.c` if that means that helper has to fetch ruby thread type by using `rb_thread_current()`, since that means that code will be triggered post crash, so we will be basically comparing the types of two `rb_thread_current()` calls post crash, which is meaningless. 

I think a nicer approach will be to just `rb_typeddata_is_kind_of` within the callback, as such

```
static void ruby_runtime_stack_callback(
  void (*emit_frame)(const ddog_crasht_RuntimeStackFrame*)
) {
  // Grab the Ruby thread we crashed on; crashtracker only runs once.
  VALUE current_thread = rb_thread_current();
  if (current_thread == Qnil) return;

  if (crashtracker_thread_data_type == NULL) return;

  if (!rb_typeddata_is_kind_of(current_thread, crashtracker_thread_data_type)) {
    emit_placeholder_frame(emit_frame, "<runtime stack not found>");
    return;
  }
  
  ........
  ```
  
This way, we are comparing the type of ruby thread pre crash (which we know is valid) to the type of ruby thread post crash (what we want to validate)

I also did the same validation you did with passing in a garbage value, and it does not hang this way ^_^

File: ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c (hunk 42 above)
Comment (gyuheon0h) [reply ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c]: Great idea! I need to also make helpers in `private_vm_api_access.c` for `rb_iseq_base_label` and `rb_iseq_path`, since those are private

File: ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c [hunk 43]
```diff
@@ -0,0 +1,235 @@
...
+// TODO: This function is not necessarily Ruby specific. This will be moved to
+// `libdatadog` in the future as a shared utility function.
+static inline bool is_pointer_readable(const void *ptr, size_t size) {
+  if (!ptr || size == 0) return false;
+
+  uintptr_t page_size = (uintptr_t)sysconf(_SC_PAGESIZE);
+  // fallback for weird value; 0 or not a power of two
+  if (page_size == 0 || (page_size & (page_size - 1u))) {
+      page_size = 4096;
+  }
```
ivoanjo: Minor: It occurred to me that we should probably initialize this in `init` similarly as we do for `crashtracker_thread_data_type`. It seems wasteful to keep looking this up on every call to `is_pointer_readable`.

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,235 @@
...
+
+static ddog_CharSlice safe_string_value(VALUE str) {
+  if (str == Qnil) return DDOG_CHARSLICE_C("<nil>");
+
+  // Validate object header readability before touching it
+  if (!is_pointer_readable((const void *)str, sizeof(struct RBasic))) return DDOG_CHARSLICE_C("<corrupted>");
+  if (!RB_TYPE_P(str, T_STRING)) return DDOG_CHARSLICE_C("<not_string>");
+
+  // Validate payload readability
+  if (!is_pointer_readable((const void *)str, sizeof(struct RString))) return DDOG_CHARSLICE_C("<corrupted>");
```
ivoanjo: Minor: I believe the two `is_pointer_readable` checks here can be combined as:

* Any Ruby object should be at least `sizeof(struct RString))`
* The "happy path" is that the page **is readable** so we speed up that situation, rather than paying the 2-check 
* Not 100% sure but I don't think Ruby objects will ever be half-half in a given memory page, so I think in practice after rounding we'd always be doing exactly the same `mincore` call twice

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,235 @@
...
+
+  long len = RSTRING_LEN(str);
+  if (len < 0 || len > 1024) return DDOG_CHARSLICE_C("<corrupted>");
```
ivoanjo: Minor:

```suggestion
  if (len < 0 || len > 1024) return DDOG_CHARSLICE_C("<length_over_limit>");
```

Since there's many things that can lead to `<corrupted>`, but "length over limit" I can see happening in some setups with very deep paths, I suggest lightly distinguishing them just in case we need to figure out why something is not showing up properly.

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,235 @@
...
+
+      emit_frame(&frame);
+    }
+  }
+}
+
+void crashtracking_runtime_stacks_init(void) {
+  if (crashtracker_thread_data_type == NULL) {
+    VALUE current_thread = rb_thread_current();
+    if (current_thread == Qnil) return;
```
ivoanjo: Minor: I suggest raising an exception here -- since this is a "should never happen", it's nice to effectively assert that the "should never happen" is actually true. Then one of two things happen: we never see this ever again; or we see it and we learn why our "should never happen" isn't actually true. Win/win ;)

ivoanjo (approved):
👍 LGTM. This looks great! I left a small number of very small tweaks, but overall this looks good to go in.

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
ivoanjo: > I'm not sure we can have it as a helper in private_vm_api_access.c if that means that helper has to fetch ruby thread type by using rb_thread_current()

Yup, you're right -- I think it currently always gets initialized as it gets used in one of the extension self-tests but if we'd gone that way we'd probably want to make it more explicit.

This to say, I'm happy with the current approach you suggested, it's very reasonable as well.

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,239 @@
...
+  if (frame_count <= 0) {
+    emit_placeholder_frame(emit_frame, "<runtime stack not found>");
+    return;
+  }
+
+  bool truncated = frame_count >= RUNTIME_STACK_MAX_FRAMES;
+
+  for (int i = frame_count - 1; i >= 0; i--) {
+    if (truncated && i == 0) {
+      emit_placeholder_frame(emit_frame, "<truncated frames>");
```
p-datadog:
```suggestion
      emit_placeholder_frame(emit_frame, "<truncated frame>");
```
I think this should be singular?

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -318,6 +317,67 @@
...
+
+          expect(frames).to include(
+            hash_including(
+              function: 'gsub'
+            )
+          )
+
+          expect(frames).to include(
+            hash_including(
+              function: 'each'
```
p-datadog: Should there be another frame for `free`? If so shouldn't that also be asserted?

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
gyuheon0h: thoughts on frame(s)?

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -318,6 +317,67 @@
           end
         end
       end
+
+      describe 'Ruby and C method runtime stack capture' do
+        let(:runtime_stack) { crash_report_experimental[:runtime_stack] }
+
+        it 'captures both Ruby and C method frames in mixed stacks' do
```
ivoanjo, line 328: Btw it occurred to me that if someone is running this spec manually during local development, they may not realize they need to have built the profiler as well, and this will be a confusing failure.

What do you think of maybe adding a `raise 'This spec requires profiling' unless Datadog::Profiling.supported?` or something like that to cause an explicit clear failure instead?

spec/datadog/core/crashtracking/component_spec.rb
gyuheon0h: There is on line 348

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
gyuheon0h: Or I can probably explicitly say how many frames were truncated

gyuheon0h: Nevermind! The profiling stack unwind API only returns the number of frames walked, given some max limit. If we hit the limit, we should let the customers know that some frame(s) were truncated, 

How profiling does this
[truncation](https://github.com/DataDog/dd-trace-rb/blob/f2b0f0a20be616a2d5345b53909eaf0046d3ceb7/ext/datadog_profiling_native_extension/collectors_stack.c#L386-L391)
[truncation message](https://github.com/DataDog/dd-trace-rb/blob/f2b0f0a20be616a2d5345b53909eaf0046d3ceb7/ext/datadog_profiling_native_extension/collectors_stack.c#L528-L534)

spec/datadog/core/crashtracking/component_spec.rb
p-datadog: :+1: missed it

spec/datadog/core/crashtracking/component_spec.rb
```diff
@@ -318,6 +317,73 @@
...
+
+            crash_tracker = build_crashtracker(agent_base_url: agent_base_url)
+            crash_tracker.start
+
+            crash_stack_helper_class.new.top_level_ruby_method
+          end
+
+          frames = runtime_stack[:frames]
+
+          puts frames.to_json
```
p-datadog: debug print

ext/datadog_profiling_native_extension/crashtracking_runtime_stacks.c
```diff
@@ -0,0 +1,236 @@
...
+    const rb_data_type_t *thread_data_type = RTYPEDDATA_TYPE(current_thread);
+    if (!thread_data_type) {
+      rb_raise(rb_eRuntimeError, "crashtracking_runtime_stacks_init: RTYPEDDATA_TYPE returned NULL");
+    }
+
+    crashtracker_thread_data_type = thread_data_type;
+  }
+
+  // Register immediately so Ruby doesn't need to manage this explicitly.
+  ddog_crasht_register_runtime_frame_callback(ruby_runtime_stack_callback);
```
p-datadog, line 232: My understanding is  this is a crucial line in the PR  - if profiling extension loads, it registers this callback with libdatadog and then the functionality in this PR will be available. But there is no dependency from libdatadog to profiling.

If this is accurate, it would be great to have this information somewhere for our own future use. I'm not sure what the best place for it would be though - maybe a top-level comment in this file, or a note in https://github.com/DataDog/dd-trace-rb/blob/master/docs/DevelopmentGuide.md?

PR #4988, opened 2025-10-17
Fix `http.route` tag for rack apps mounted in Rails

y9v:
**What does this PR do?**
This PR removes the behavior that used the request path as a fallback when computing the route for Rack applications mounted inside Rails.

**Motivation:**
The current approach to setting the http.route tag for mounted Rack applications causes issues for the API Catalog. Specifically, when a mounted Rack application includes parameters in its request path, it can lead to an unbounded number of reported routes.

This issue can occur in the following cases:
- Mounted applications such as the Sidekiq web UI, which is a Rack app with a simple [router](https://github.com/sidekiq/sidekiq/blob/main/lib/sidekiq/web/router.rb).
- Mounted Sinatra and Grape applications, when the Datadog tracer is configured to instrument Rails but not Sinatra or Grape directly.

Before the fix route for an embedded Rack app would look like `/path/for/mounted/app/users/1`, after the fix it would be `/path/for/mounted/app`.

**Change log entry**
Yes. Fix `http.route` tag for Rack applications mounted inside Rails.

**Additional Notes:**
APPSEC-59663

**How to test the change?**
CI and app generator.


Strech (approved):
🔥 Amazing! Well done!

PR #4992, opened 2025-10-20
[PROF-12763] Tweak profiler signal handler tests to avoid flakiness

ivoanjo:
**What does this PR do?**

This PR tweaks the "sample from the signal handler" profiler specs to run with GC disabled.

To avoid accidentally running out of memory in `loop_until`, we additionally changed two things:
* Use `Process.clock_gettime` (that returns a zero-allocation float) instead of `Time.now` (that returns a new `Time` instance)
* Add a `check_condition_every_seconds` to avoid calling the condition checking code too often, as it can do allocations

**Motivation:**

This change is an attempt at avoiding the flakiness seen in <https://github.com/DataDog/dd-trace-rb/actions/runs/18560146732/job/52906922299>:

```
Failures:

  1) Datadog::Profiling::Collectors::CpuAndWallTimeWorker#start sampling from signal handler when signal handler sampling is disabled does not prepare samples in the signal handler
     Failure/Error: expect(sample_count).to be_within(20).percent_of(signal_handler_enqueued_sample)
       expected 15 to be within 20% of 20
     # ./spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb:1053:in 'block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:272:in 'block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:154:in 'block (2 levels) in <top (required)>'
     # /usr/local/bundle/gems/webmock-3.24.0/lib/webmock/rspec.rb:39:in 'block (2 levels) in <top (required)>'
     # ./spec/support/execute_in_fork.rb:32:in 'ForkableExample#run'

Finished in 27.13 seconds (files took 1.14 seconds to load)
787 examples, 1 failure, 19 pending

Failed examples:

rspec ./spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb:1068 # Datadog::Profiling::Collectors::CpuAndWallTimeWorker#start sampling from signal handler when signal handler sampling is disabled does not prepare samples in the signal handler
```

The assertion that failed checks that the number of times the profiler actually got to sample is close to the number of times it tried to sample.

Unfortunately, I could not reproduce the issue locally but I suspect that what made this condition flaky is GC: the profiler is not able to sample during GC and thus if a long-enough-running GC happens at the wrong time, this condition could fail.

I am suspicious that GC is the source also because `loop_until` caused lots of allocations previously.

Hopefully this was the source of the flakiness; if not, I'll come up with more ideas next time ;)

**Change log entry**

None.

**Additional Notes:**

Fixes https://github.com/DataDog/ruby-guild/issues/272

**How to test the change?**

If CI is green we're good! The logic on the test itself hasn't changed.

### DIFF
--- a/spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
+++ b/spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
@@ -1043,11 +1043,23 @@ RSpec.describe Datadog::Profiling::Collectors::CpuAndWallTimeWorker do
 
         skip_if_signal_handler_sampling_not_supported
 
-        start
+        GC.start
 
-        loop_until { cpu_and_wall_time_worker.stats.fetch(:signal_handler_enqueued_sample) >= 20 }
+        begin
+          # The expectations below compare number of enqueued samples with actual samples. To avoid flakiness from a
+          # slow GC in the middle of the test causing less samples than expected to be taken (since we only sample after
+          # GC ends), we disable GC during the test setup.
+          # (Note for future changes: Be careful with what you do with GC disabled!)
+          GC.disable
 
-        cpu_and_wall_time_worker.stop
+          start
+
+          loop_until(check_condition_every_seconds: 0.01) { cpu_and_wall_time_worker.stats.fetch(:signal_handler_enqueued_sample) >= 20 }
+
+          cpu_and_wall_time_worker.stop
+        ensure
+          GC.enable
+        end
 
         expect(sample_count).to be > 0
         expect(sample_count).to be_within(20).percent_of(signal_handler_enqueued_sample)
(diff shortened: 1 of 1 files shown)


AlexJF (approved):

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
```diff
@@ -1648,10 +1660,21 @@ def build_thread_context_collector(recorder, **options)
     )
   end
 
-  def loop_until(timeout_seconds: 5)
-    deadline = Time.now + timeout_seconds
+  def loop_until(timeout_seconds: 5, check_condition_every_seconds: 0)
+    started_at = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_second)
```
marcotc, line 1664: Very nice catch to use the monotonic clock, just in case!

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
```diff
@@ -1043,11 +1043,23 @@
...
-        start
+        GC.start
 
-        loop_until { cpu_and_wall_time_worker.stats.fetch(:signal_handler_enqueued_sample) >= 20 }
+        begin
+          # The expectations below compare number of enqueued samples with actual samples. To avoid flakiness from a
+          # slow GC in the middle of the test causing less samples than expected to be taken (since we only sample after
+          # GC ends), we disable GC during the test setup.
+          # (Note for future changes: Be careful with what you do with GC disabled!)
+          GC.disable
```
marcotc, line 1053: We can create a block for `with_gc_diasabled`, if we think this is a pattern going forward, and requires the explicit syntax to ensure we know that "what's in this block can be dangerous".

But that's just a note, no real action needed here.

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
ivoanjo: I didn't mention here, but I did check that if the condition never became true for a long time, the app wouldn't blow up due to too much memory. That was the main hiccup on this change -- took me a while to figure out how I could disable GC to avoid it blocking the profiler sampling but still keep the test semantics :)

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
ivoanjo: Yeap, good call! I'll be lazy for now but for sure I'll keep an eye and add that helper if this starts showing up more :)

PR #5004, opened 2025-10-28
[NO-TICKET] Extract all .rb files from folders in Steepfile

vpellan:
**What does this PR do?**

This PR details all the currently existing ignored .rb files in the Steepfile instead of specifying folders

**Motivation:**

This will ensure that new files added in these folders later are NOT ignored unless explicitly added to the list (and hopefully enforce typing on them)

**Change log entry**

None.

**Additional Notes:**

Newly added methods to ignored files will still be ignored. A way to fix that would actually be to remove any ignored file (but this would drop the status of typing to ~18%)

**How to test the change?**


y9v (approved):
really good idea, thank you!

ivoanjo (approved):
:+1: I like it! Also it makes our `Steepfile` look worse which gives us a motivator to continue improving this ;)

PR #5006, opened 2025-10-29
Add a fallback to inferred route for AppSec Api security sampler

y9v:
**What does this PR do?**
This PR changes `AppSec::APISecuity::RouteExtractor` to infer the route if it cannot be extracted, instead of falling back to request path. This route is then used for AppSec API security sampler.

**Motivation:**
We want to have more accurate sampling for requests to barebones Rack applications, or Rack-based frameworks that we don't instrument directly.

**Change log entry**

**Additional Notes:**
APPSEC-59868

**How to test the change?**
CI.

### DIFF
--- a/lib/datadog/tracing/contrib/rack/route_inference.rb
+++ b/lib/datadog/tracing/contrib/rack/route_inference.rb
@@ -27,8 +27,8 @@ module Datadog
           def infer(path)
             segments = path.delete_prefix('/').split('/', MAX_NUMBER_OF_SEGMENTS + 1).first(MAX_NUMBER_OF_SEGMENTS)
 
-            segments.map! do |segment|
-              next if segment.empty? # steep:ignore
+            segments.map! do |segment| #: Array[String?]
+              next if segment.empty?
 
               case segment
               when INT_PARAM_REGEX then '{param:int}'
@@ -40,7 +40,7 @@ module Datadog
               end
             end
 
-            segments.compact!
+            segments.compact! #: Array[String]
 
             "/#{segments.join("/")}"
           rescue


spec/datadog/appsec/api_security/route_extractor_spec.rb
```diff
@@ -130,8 +136,8 @@
           let(:action_dispatch_request) { double('ActionDispatch::Request', env: {}, script_name: '', path: '/users/1') }
 
           it 'uses action dispatch request for route recognition' do
-            expect(router).to receive(:recognize).with(action_dispatch_request).and_return('/users/1')
-            expect(described_class.route_pattern(request)).to eq('/users/1')
+            expect(router).to receive(:recognize).with(action_dispatch_request).and_return('/users/:id(.:format)')
```
y9v, line 140: this change is not quite related to this PR, but this is a more accurate approximation of what the router might return

lib/datadog/appsec/api_security/route_extractor.rb
```diff
@@ -66,7 +68,9 @@ def self.route_pattern(request)
             #       to generic request path
             (pattern || request.path).delete_suffix(RAILS_FORMAT_SUFFIX)
           else
-            request.path
+            Tracing::Contrib::Rack::RouteFromPathInference.infer(request.path).tap do |inferred_route|
+              request.env[Tracing::Contrib::Rack::Ext::DATADOG_INFERRED_ROUTE] = inferred_route
+            end
```
Strech: This seems to be a general pattern of use (in tracing too), read it or compute it and write it.

Does it makes sense to adjust the way it's working or wrap it up to hide that knowledge (as it breaks SRP and the AppSec method definition).

We can also hide the knowledge of "how" from the module name.

```ruby
Tracing::Contrib::Rack::RouteInference.infer(env)
# or
Tracing::Contrib::Rack::RouteInference.infer_from(env)
```

then we don't know what's in the black-box and we can use same in the tracer part

y9v: that's a good idea, extracted it in [d17fc75](https://github.com/DataDog/dd-trace-rb/pull/5006/commits/d17fc75c9fb95ea01a9360fbc727c2edb43070a7)

p-datadog (approved):

lib/datadog/tracing/contrib/rack/route_inference.rb
```diff
@@ -15,13 +15,20 @@ module RouteFromPathInference
...
+            request_env[DATADOG_INFERRED_ROUTE_ENV_KEY] ||=
+              infer(request_env['SCRIPT_NAME'].to_s + request_env['PATH_INFO'].to_s)
+          end
+
           def infer(path)
             segments = path.delete_prefix('/').split('/', MAX_NUMBER_OF_SEGMENTS + 1).first(MAX_NUMBER_OF_SEGMENTS)
 
             segments.map! do |segment|
-              next if segment.empty?
+              next if segment.empty? # steep:ignore
```
Strech: I think it's possible to type segment via steep annotations

y9v: you are right, I forgot about new inline annotations. Fixed in [f400f71](https://github.com/DataDog/dd-trace-rb/pull/5006/commits/f400f71a033aa0c63a382cb90a0937d3fcfecd21)

PR #5007, opened 2025-10-29
[FFL-1273] bindings for ffe in openfeature provider

sameerank:
**What does this PR do?**

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/spec/datadog/open_feature/component_spec.rb
+++ b/spec/datadog/open_feature/component_spec.rb
@@ -78,24 +78,14 @@ RSpec.describe Datadog::OpenFeature::Component do
   end
 
   describe '#shutdown!' do
-<<<<<<< HEAD
-=======
-    subject(:component) { described_class.new(settings, agent_settings, logger: logger, telemetry: telemetry) }
-
->>>>>>> 0f5c9c2e83 (Create Reporter for exposures reporting)
     before do
       settings.open_feature.enabled = true
       settings.remote.enabled = true
     end
 
-<<<<<<< HEAD
     subject(:component) { described_class.new(settings, agent_settings, logger: logger, telemetry: telemetry) }
 
     it 'flushes worker and stops it' do
-=======
-    it 'flushes reporter cache and stops worker' do
-      expect(reporter).to receive(:clear)
->>>>>>> 0f5c9c2e83 (Create Reporter for exposures reporting)
       expect(worker).to receive(:flush)
       expect(worker).to receive(:stop).with(true)
 


ext/libdatadog_api/extconf.rb
sameerank, line 1: These changes are just for local development. I will revert all the changes in this file after `libdatadog` is published with `datadog-ffe-ffi`

ivoanjo: I'm curious, did [`LIBDATADOG_VENDOR_OVERRIDE`](https://github.com/DataDog/dd-trace-rb/blob/master/ext/LIBDATADOG_DEVELOPMENT.md) not work for you? Usually that's how we've been developing/testing against libdatadog changes.

setup_ffe.sh
```diff
@@ -0,0 +1,244 @@
...
+#endif // __cplusplus
+
+// Stub function declarations - can be extended as needed
+
+#ifdef __cplusplus
+}  // extern "C"
+#endif  // __cplusplus
+
+#endif  /* DDOG_PROFILING_H */
+EOF
```
ivoanjo: Tip: You don't actually need to build profiling -- `DD_PROFILING_NO_EXTENSION=true` skips trying it to build at all.

setup_ffe.sh
```diff
@@ -0,0 +1,244 @@
...
+echo "Generating Makefile..."
+ruby extconf.rb
+echo "Compiling extension..."
+make
+echo "Installing extension..."
+cp libdatadog_api.*.bundle ../../lib/
+echo "Cleaning up build artifacts..."
+make clean
+rm -f Makefile
+cd ../..
```
ivoanjo: Minor: We have [rake-compiler](https://github.com/rake-compiler/rake-compiler) as a dependency, you can use `bundle exec rake compile:libdatadog_api.NAME_OF_THE_PLATFORM` to build the extension (You can use `bundle exec rake -T` to see all targets and which platform name you should use)

setup_ffe.sh
sameerank, line 1: To be deleted after `libdatadog` is published with `datadog-ffe-ffi`, at which point we no longer need to build and install it locally

ext/libdatadog_api/extconf.rb
sameerank: Oh I hadn't tried it because I couldn't find the `build-profiling-ffi.sh` script in step 5 and assumed the next step was related

https://github.com/DataDog/dd-trace-rb/blob/7520057545ec4787837b84005d7c27dbd120ad11/ext/LIBDATADOG_DEVELOPMENT.md#L55-L56

I tried it now though and it's working! I can revert this file back 🙌

setup_ffe.sh
sameerank: Thanks for the tip! ~Updated in https://github.com/DataDog/dd-trace-rb/pull/5007/commits/e974e4e8734100a0e1929e90ad93952313550e67~

`bundle exec rake -T` is pretty neat! So `rake` can generate the correctly formatted platform name itself https://github.com/DataDog/dd-trace-rb/pull/5007/commits/0eae0ce9f277457363e91135379c4ab35ae1d031

setup_ffe.sh
sameerank: Incorporated this into the script though I'm not sure that I'm using it right.

Most of the profiling dependencies are in `ext/datadog_profiling_native_extension` which I'm understand that env variable lets us skip, but there's still one profiling dependency in `ext/libdatadog_api`

https://github.com/DataDog/dd-trace-rb/blob/f10f1df8e9e0d9a9e0683864897f9a57dfc088ed/ext/libdatadog_api/datadog_ruby_common.h#L8

So I still ended up needing to generate the `profiling.h` file

ext/libdatadog_api/extconf.rb
ivoanjo: Aaaah right. It's been replaced [by the builder crate](https://github.com/datadog/libdatadog?tab=readme-ov-file#builder-crate). I'll whip up a small PR to update the docs.

setup_ffe.sh
ivoanjo: Aaah oops that's actually an oversight from when `datadog_ruby_common.h` was extracted out of profiling: this can be replaced with `#include <datadog/common.h>`. I'll open a PR to fix this

setup_ffe.sh
ivoanjo: I've never looked into it, but we could probably get rake to generate an alias that doesn't include the platform name as the task suffix -- the suffix comes from `rake-compiler`.

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,212 @@
...
+    ddog_ffe_evaluation_context_drop(context);
+  }
+  ruby_xfree(ptr);
+}
+
+
+
+static VALUE evaluation_context_initialize_with_attributes(VALUE self, VALUE targeting_key, VALUE attributes_hash) {
+  Check_Type(targeting_key, T_STRING);
+  Check_Type(attributes_hash, T_HASH);
```
ivoanjo, line 138: Minor: I recommend using our own `ENFORCE_TYPE` instead of `Check_Type` -- they behave similarly, but `ENFORCE_TYPE` gets you a much better error exception message (you can easily see it working, by e.g. on purpose switching to a wrong type here and seeing the error you get)

lib/datadog/open_feature/binding.rb
```diff
@@ -0,0 +1,125 @@
...
+        :flag_metadata,
+        keyword_init: true
+      )
+
+      # Evaluation class that wraps the native binding methods
+      class Evaluator
+        def initialize(ufc_json)
+          unless Binding.supported?
+            raise(ArgumentError, "Feature Flags are not supported: #{Datadog::Core::LIBDATADOG_API_FAILURE}")
+          end
```
ivoanjo: Since this gets repeated a lot, consider perhaps extracting it? (E.g. into a `Binding.ensure_supported!` or something like that?)

ivoanjo:
I've opened https://github.com/DataDog/dd-trace-rb/pull/5018 with some small fixes as I mentioned in the comments.

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,430 @@
...
+  return TypedData_Wrap_Struct(klass, &configuration_typed_data, config);
+}
+
+static void configuration_free(void *ptr) {
+  ddog_ffe_Handle_Configuration *config = (ddog_ffe_Handle_Configuration *) ptr;
+  if (config && *config) {
+    ddog_ffe_configuration_drop(config);
+  }
+  ruby_xfree(ptr);
+}
```
dd-oleksii, line 85: nit: not entirely sure about this but handle *is* a pointer and it fits to data field (`void *`), so I think we can avoid allocations here, call `rb_undef_alloc_func(configuration_class)`, and call `TypedData_Wrap_Struct` in `configuration_initialize`.

```suggestion
static void configuration_free(void *ptr) {
  ddog_ffe_configuration_drop(&(ddog_ffe_Handle_Configuration) ptr);
}
```

Similarly, for all other handle types

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,430 @@
...
+    *config, 
+    RSTRING_PTR(flag_key), 
+    DDOG_FFE_EXPECTED_FLAG_TYPE_STRING,  // Default to string type
+    *context
+  );
+
+  // Check if resolution_details is NULL (no assignment returned)
+  if (resolution_details_out == NULL) {
+    return Qnil;
+  }
```
dd-oleksii, line 259: nit: `ddog_ffe_get_assignment` always returns a result, so this check is not necessary

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,430 @@
...
+  resolution_details_class = rb_const_get(resolution_details_class, rb_intern("ResolutionDetails"));
+  
+  VALUE resolution_details_obj = resolution_details_alloc(resolution_details_class);
+  
+  ddog_ffe_Handle_ResolutionDetails *resolution_details_ptr;
+  TypedData_Get_Struct(resolution_details_obj, ddog_ffe_Handle_ResolutionDetails, &resolution_details_typed_data, resolution_details_ptr);
+  
+  *resolution_details_ptr = resolution_details_out;
+
+  return resolution_details_obj;
```
dd-oleksii, line 272: nit re using handles directly without extra allocation, this would simplify this section:
```suggestion
  return TypedData_Wrap_Struct(resolution_details_class, &resolution_details_typed_data, resolution_details_out);
```

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,430 @@
...
+  VALUE resolution_details_class = rb_define_class_under(binding_module, "ResolutionDetails", rb_cObject);
+  rb_define_alloc_func(resolution_details_class, resolution_details_alloc);
+
+  rb_define_method(resolution_details_class, "value", resolution_details_get_value, 0);
+  rb_define_method(resolution_details_class, "reason", resolution_details_get_reason, 0);
+  rb_define_method(resolution_details_class, "error_code", resolution_details_get_error_code, 0);
+  rb_define_method(resolution_details_class, "error_message", resolution_details_get_error_message, 0);
+  rb_define_method(resolution_details_class, "variant", resolution_details_get_variant, 0);
+  rb_define_method(resolution_details_class, "allocation_key", resolution_details_get_allocation_key, 0);
+  rb_define_method(resolution_details_class, "do_log", resolution_details_get_do_log, 0);
```
dd-oleksii, line 56: major: need to add `flag_metadata` and `extra_logging` methods

lib/datadog/open_feature/binding/native_evaluator.rb
```diff
@@ -0,0 +1,199 @@
...
+            # All fields are nil - treat as evaluation failure
+            ResolutionDetails.new(
+              value: default_value,
+              variant: nil,
+              error_code: :flag_not_found,
+              error_message: "Native evaluation returned empty result",
+              reason: :error,
+              allocation_key: nil,
+              do_log: nil
+            )
```
dd-oleksii, line 73: major: this is an internal error, not `not found` case. We should report it as such

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,430 @@
...
+      return ID2SYM(rb_intern("targeting_key_missing"));
+    case DDOG_FFE_ERROR_CODE_INVALID_CONTEXT:
+      return ID2SYM(rb_intern("invalid_context"));
+    case DDOG_FFE_ERROR_CODE_PROVIDER_NOT_READY:
+      return ID2SYM(rb_intern("provider_not_ready"));
+    case DDOG_FFE_ERROR_CODE_GENERAL:
+      return ID2SYM(rb_intern("general"));
+    default:
+      return Qnil;
+  }
```
dd-oleksii, line 362: minor: "general" is a safer default
```suggestion
  switch (error_code) {
    case DDOG_FFE_ERROR_CODE_OK:
      return Qnil;
    case DDOG_FFE_ERROR_CODE_TYPE_MISMATCH:
      return ID2SYM(rb_intern("type_mismatch"));
    case DDOG_FFE_ERROR_CODE_PARSE_ERROR:
      return ID2SYM(rb_intern("parse_error"));
    case DDOG_FFE_ERROR_CODE_FLAG_NOT_FOUND:
      return ID2SYM(rb_intern("flag_not_found"));
    case DDOG_FFE_ERROR_CODE_TARGETING_KEY_MISSING:
      return ID2SYM(rb_intern("targeting_key_missing"));
    case DDOG_FFE_ERROR_CODE_INVALID_CONTEXT:
      return ID2SYM(rb_intern("invalid_context"));
    case DDOG_FFE_ERROR_CODE_PROVIDER_NOT_READY:
      return ID2SYM(rb_intern("provider_not_ready"));
    case DDOG_FFE_ERROR_CODE_GENERAL:
      return ID2SYM(rb_intern("general"));
    default:
      return ID2SYM(rb_intern("general"));
  }
```

lib/datadog/open_feature/binding/native_evaluator.rb
```diff
@@ -0,0 +1,199 @@
...
+            # Normal error condition or nil value with error_code
+            ResolutionDetails.new(
+              value: default_value,
+              variant: result.variant,
+              error_code: result.error_code || :flag_not_found,
+              error_message: result.error_message || "Flag evaluation failed",
+              reason: result.reason || :error,
+              allocation_key: result.allocation_key,
+              do_log: result.do_log
+            )
```
dd-oleksii, line 84: major 🐛: this incorrectly squashes non-error evaluation that resulted in no value (e.g., flag disabled or default allocation null, or null json flags) into error.
```suggestion
          elsif result.variant.nil?
            # Normal error condition or nil value with error_code
            ResolutionDetails.new(
              value: default_value,
              variant: result.variant,
              error_code: result.error_code,
              error_message: result.error_message,
              reason: result.reason,
              allocation_key: result.allocation_key,
              do_log: result.do_log
            )
```

lib/datadog/open_feature/binding/native_evaluator.rb
```diff
@@ -0,0 +1,199 @@
...
+            default_value = expected_type
+            expected_type = nil
+          end
+          
+          # Validate input parameters
+          raise TypeError, "flag_key must be a String" unless flag_key.is_a?(String)
+          
+          # First try to handle evaluation using Ruby-based logic
+          ruby_result = try_ruby_evaluation(flag_key, evaluation_context, default_value)
+          return ruby_result if ruby_result
```
dd-oleksii, line 42: major: native evaluation should be tried first?

sameerank: Ah yeah this is just some mess from moving code around while I debug. The evaluation code in Ruby shouldn't be used at all (and probably should be deleted) after we get the binding working

sameerank:
Accidental merge. Re-drafting here: https://github.com/DataDog/dd-trace-rb/pull/5034

PR #4901, opened 2025-09-22
DSM Implementation for Ruby with Karafka instrumented

ericfirth:
- **Bootstrap file for binding ddsketch for Ruby**
- **Class creation/freeing**
- **Bind more ddsketch methods**
- **Implement encode**
- **Move inside `Core`**
- **Add spec and type signatures**
- **Refactor wiring up of specs using libdatadog_api**
- **Wire up testing for ddsketch**
- **Import unmodified ddsketch.proto**
- **Minor test simplification**
- **Bootstrap support for decoding sketches for testing**
- **Add simple test that ddsketch protobuf can be decoded**
- **Improve libdatadog development docs**
- **Fix error handling and add test coverage**
- **Initial monkey patching of the Ruby Kafka to no-op**
- **Add setting for if data streams is enabled**
- **Karafka consumes are instrumented but as a no-op**
- **Can decode the pathway context**
- **Implement core Data Streams Monitoring (DSM) for Karafka**
- **Implement DDSketch integration for Data Streams Monitoring**
- **Implement all remaining processor TODOs and fix test expectations**
- **Perfect Karafka DSM integration testing**
- **Final Karafka DSM integration test fixes**
- **Fix libdatadog_api for libdatadog v21 API compatibility**
- **Use available libdatadog version 19.1.0.1.0 instead of non-existent 21.0.0.0.0**
- **Fix DSM processor transport to use proper HTTP client for /v0.1/pipeline_stats endpoint**
- **Fix DSM pathway context and agent transport**
- **Add comprehensive debug logging to DSM processor**
- **Fix DSM loop detection bug and add debug logging**
- **Add comprehensive debug logging for DSM aggregation**
- **Add detailed bucket time and agent payload logging**
- **Fix encoding issue in DSM payload sending and clean up excessive debug logs**
- **Add debug logging for EdgeTags direction to investigate edgeDirection issue**
- **Fix missing type:kafka tag in Karafka DSM checkpoint - matches Python/Node.js implementation**
- **Add consumer group tag to Karafka DSM checkpoint - matches Node.js implementation pattern**
- **Clean up DSM Ruby implementation for release**

**What does this PR do?**

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**

PR #5008, opened 2025-10-29
Streamline ddsketch loading, remove circular dependencies

p-datadog:
**What does this PR do?**
Changes how ddsketch / libdatadog integration is handled:

Before, there was a `lib/datadog/core/ddsketch.rb` file which contained a class-level `supported?` method. Having this file meant that you could always require `datadog/core/ddsketch`, but you wouldn't always get a functional ddsketch out of that require.

The proposed change is to delete this file completely. Loading `datadog/core` will attempt to bring in ddsketch (and this is already behavior in master), therefore no additional require is needed for ddsketch.

To check for ddsketch presence, the `supported?` method was moved up to `Datadog::Core` module (up from `Datadog::Core::DDSketch`). Meaning:

1) if ddsketch is not supported, `Datadog::Core::DDSketch` will be not defined at all.
2) one does not need to require ddsketch to see if it's supported.

**Motivation:**
Removing circular dependency: ddsketch -> core -> configuration -> ddsketch

**Change log entry**
None

**Additional Notes:**

**How to test the change?**
Existing CI

### DIFF
--- a/lib/datadog/core/ddsketch.rb
+++ b/lib/datadog/core/ddsketch.rb
@@ -2,18 +2,16 @@
 
 module Datadog
   module Core
-    # Used to access ddsketch APIs.
+    # Datadog::Core::DDSketch is defined completely in the native extension.
+    # Do not define it here (for example, as an empty class) because we don't
+    # want to be able to create instances of the empty stub class if the
+    # native extension is missing or failed to load.
     #
-    # This class is not empty; all of this class is implemented as native code.
+    # Use Core.ddsketch_supported? to determine if DDSketch class exists and
+    # is usable.
     #
-    # @api private
-    class DDSketch # rubocop:disable Lint/EmptyClass
-      # This constructor is replaced by the C extension and will only be
-      # used if the C extension failed to build or was not loaded,
-      # in which case DDSketch is not available.
-      def initialize
-        raise ArgumentError, "DDSketch is not supported: #{Datadog::Core::LIBDATADOG_API_FAILURE}"
-      end
-    end
+    # See https://github.com/datadog/dd-trace-rb/pull/5008 and
+    # https://github.com/DataDog/dd-trace-rb/pull/4901 for the background on
+    # dependency issues with DDSketch.
   end
 end


p-datadog:
CI is failing due to missing type declaration, I'll add it if there is consensus that this change is the way to go.

lib/datadog/core.rb
```diff
@@ -20,6 +20,10 @@ module Core
       rescue LoadError => e
         e.message
       end
+
+    module_function def ddsketch_supported?
+      LIBDATADOG_API_FAILURE.nil?
+    end
```
ivoanjo: Minor: I believe we should move away from using `module_function` in general in dd-trace-rb, because in most cases we don't need the behavior of "call on self + call on included".

Since the intention here is to create a `Datadog::Core.ddsketch_supported?`, I don't think we need to create the "module inclusion" version.

```suggestion
    def self.ddsketch_supported?
      LIBDATADOG_API_FAILURE.nil?
    end
```

p-datadog:
The technical circular dependency was removed in https://github.com/DataDog/dd-trace-rb/pull/4901/commits/43e1f23b0c0e74dea8cc248ec962827f901b65b9#diff-6566ed78aec0d0209e0f3b7776e7db57d4694f438d12d3b6b9312f1209bfd994L3, specifically the require of 'datadog/core' from 'datadog/core/ddsketch'. However, 'datadog/core/ddsketch' still references Datadog::Core, but now without the require being present.

So, technically, the circular dependency is there, but not in 'require' form at the moment.

ivoanjo:
Oh, right, I understand. Indeed `Datadog::Core::DDSketch.supported?` would break unless core was loaded.

p-datadog:
I tried to put back in the empty class but this causes issues because it's not supposed to be creatable when the C extension is missing, and trying to fix that took me back to the current state of things before the changes in this PR.

I put in the ddsketch.rb so that type checking rake task is happy, and that file only contains a comment about the DDSketch class, but no definition of the class itself.

lib/datadog/core/ddsketch.rb
```diff
@@ -2,18 +2,16 @@
...
-      def initialize
-        unless self.class.supported?
-          raise(ArgumentError, "DDSketch is not supported: #{Datadog::Core::LIBDATADOG_API_FAILURE}")
-        end
-      end
-    end
+    # Datadog::Core::DDSketch is defined completely in the native extension.
+    # Do not define it here (for example, as an empty class) because we don't
+    # want to be able to create instances of the empty stub class if the
+    # native extension is missing or failed to load.
```
ivoanjo, line 8: It's not a blocker for this PR if you don't agree with me, but I'm not a fan of this pattern of working **really** hard not to define the class.

In particular, because I've seen that in profiling, even when classes start as pure-native, they often evolve small Ruby-level helpers that are 200% easier to write in Ruby code instead of writing some gnarly C code "just to avoid adding a Ruby file".

So I think living with the empty definition here is better than just keeping a comment; also because it lets tools such as LSPs know where a class is defined (and then you can read the comment attached to the class that tells you "the rest is in C").

What profiling does in exchange is that the profiling classes are only required _if the native extension is loaded successfully_ and I would suggest adopting the pattern here. Specifically: define the class in this file but make it so it won't exist if the extension isn't loaded. This happens not because the class is not defined in Ruby code but because the Ruby code doesn't get required unless the extension is up and running.

Strech: I would like to add that `core.rb` now has knowledge of `ddsketch` when maybe should not due to its nature of being "general" class/module.

p-datadog:
CI default configuration does not have libdatadog_api extension, check into using that for the test that ddsketch is missing

PR #5011, opened 2025-10-29
Add Datadog-Send-Real-Http-Status header and retry on 429 responses

mhlidd:
## Summary
- always send the Datadog-Send-Real-Http-Status header on trace requests
- retain trace chunks for each payload so 429 responses can be retried
- add retry scheduling/backoff logic for 429 responses and update specs

## Testing
- bundle exec rspec spec/datadog/tracing/writer_spec.rb spec/datadog/tracing/transport/http/integration_spec.rb

------
https://chatgpt.com/codex/tasks/task_b_69025a97a96c83328556eb7f8b43bc7d

### DIFF
--- a/lib/datadog/core/transport/http.rb
+++ b/lib/datadog/core/transport/http.rb
@@ -46,6 +46,7 @@ module Datadog
         def default_headers
           {
             Datadog::Core::Transport::Ext::HTTP::HEADER_CLIENT_COMPUTED_TOP_LEVEL => '1',
+            Datadog::Core::Transport::Ext::HTTP::HEADER_SEND_REAL_HTTP_STATUS => '1',
             Datadog::Core::Transport::Ext::HTTP::HEADER_META_LANG =>
               Datadog::Core::Environment::Ext::LANG,
             Datadog::Core::Transport::Ext::HTTP::HEADER_META_LANG_VERSION =>
--- a/lib/datadog/tracing/transport/http/traces.rb
+++ b/lib/datadog/tracing/transport/http/traces.rb
@@ -24,6 +24,7 @@ module Datadog
               super(http_response)
               @service_rates = options.fetch(:service_rates, nil)
               @trace_count = options.fetch(:trace_count, 0)
+              @traces = options.fetch(:traces, [])
             end
           end
 
@@ -98,7 +99,7 @@ module Datadog
                 http_response = super
 
                 # Process the response
-                response_options = {trace_count: env.request.parcel.trace_count}.tap do |options|
+                response_options = {trace_count: env.request.parcel.trace_count, traces: env.request.parcel.traces}.tap do |options|
                   # Parse service rates, if configured to do so.
                   if service_rates? && !http_response.payload.to_s.empty?
                     body = JSON.parse(http_response.payload)
--- a/lib/datadog/tracing/transport/traces.rb
+++ b/lib/datadog/tracing/transport/traces.rb
@@ -14,11 +14,12 @@ module Datadog
         class EncodedParcel
           include Datadog::Core::Transport::Parcel
 
-          attr_reader :trace_count
+          attr_reader :trace_count, :traces
 
-          def initialize(data, trace_count)
+          def initialize(data, trace_count, traces: [])
             super(data)
             @trace_count = trace_count
+            @traces = traces
           end
 
           def count
@@ -32,7 +33,7 @@ module Datadog
 
         # Traces response
         module Response
-          attr_reader :service_rates, :trace_count
+          attr_reader :service_rates, :trace_count, :traces
         end
 
         # Traces chunker
@@ -67,19 +68,22 @@ module Datadog
           def encode_in_chunks(traces)
             encoded_traces = if traces.respond_to?(:filter_map)
               # DEV Supported since Ruby 2.7, saves an intermediate object creation
-              traces.filter_map { |t| encode_one(t) }
+              traces.filter_map { |t| encode_pair(t) }
             else
-              traces.map { |t| encode_one(t) }.reject(&:nil?)
+              traces.map { |t| encode_pair(t) }.reject(&:nil?)
             end
 
             Datadog::Core::Chunker.chunk_by_size(encoded_traces, max_size).map do |chunk|
-              [encoder.join(chunk), chunk.size]
+              encoded_chunk = chunk.map(&:encoded)
+              trace_chunk = chunk.map(&:trace)
+
+              [encoder.join(encoded_chunk), trace_chunk.size, trace_chunk]
             end
           end
 
           private
 
-          def encode_one(trace)
+          def encode_pair(trace)
             encoded = Encoder.encode_trace(
               encoder,
               trace,
@@ -95,7 +99,13 @@ module Datadog
               return nil
             end
 
-            encoded
+            EncodedTrace.new(encoded, trace)
+          end
+
+          EncodedTrace = Struct.new(:encoded, :trace) do
+            def size
+              encoded.size
+            end
           end
         end
 
@@ -123,6 +133,8 @@ module Datadog
         # This class initializes the HTTP client, breaks down large
         # batches of traces into smaller chunks and handles
         # API version downgrade handshake.
+        TOO_MANY_REQUESTS_STATUS = 429
+
         class Transport
           attr_reader :client, :apis, :default_api, :current_api_id, :logger
 
@@ -142,30 +154,21 @@ module Datadog
               native_events_supported: native_events_supported?
             )
 
-            responses = chunker.encode_in_chunks(traces.lazy).map do |encoded_traces, trace_count|
-              request = Request.new(EncodedParcel.new(encoded_traces, trace_count))
+            responses = []
 
-              client.send_traces_payload(request).tap do |response|
-                if downgrade?(response)
-                  downgrade!
-                  return send_traces(traces)
-                end
+            chunker.encode_in_chunks(traces.lazy).each do |encoded_traces, trace_count, trace_chunk|
+              request = Request.new(EncodedParcel.new(encoded_traces, trace_count, traces: trace_chunk))
+
+              response = client.send_traces_payload(request)
+              if downgrade?(response)
+                downgrade!
+                return send_traces(traces)
               end
-            end
 
-            # Force resolution of lazy enumerator.
-            #
-            # The "correct" method to call here would be `#force`,
-            # as this method was created to force the eager loading
-            # of a lazy enumerator.
-            #
-            # Unfortunately, JRuby < 9.2.9.0 erroneously eagerly loads
-            # the lazy Enumerator during intermediate steps.
-            # This forces us to use `#to_a`, as this method works for both
-            # lazy and regular Enumerators.
-            # Using `#to_a` can mask the fact that we expect a lazy
-            # Enumerator.
-            responses = responses.to_a
+              responses << response
+
+              break if response.code.to_i == TOO_MANY_REQUESTS_STATUS
+            end
 
             Datadog.health_metrics.transport_chunked(responses.size)
 
--- a/lib/datadog/tracing/workers.rb
+++ b/lib/datadog/tracing/workers.rb
@@ -52,7 +52,8 @@ module Datadog
           begin
             traces = @trace_buffer.pop
             traces = Pipeline.process!(traces)
-            @trace_task.call(traces, @transport) unless @trace_task.nil? || traces.empty?
+            result = @trace_task.call(traces, @transport) unless @trace_task.nil? || traces.empty?
+            handle_send_result(result, traces)
           rescue => e
             # ensures that the thread will not die because of an exception.
             # TODO[manu]: findout the reason and reschedule the send if it's not
@@ -108,6 +109,20 @@ module Datadog
         private
 
         alias_method :flush_data, :callback_traces
+
+        def handle_send_result(result, original_traces)
+          case result
+          when Datadog::Tracing::Writer::SendResult
+            if result.too_many_requests?
+              @trace_buffer.unshift(*result.retry_traces)
+              false
+            else
+              result.success?
+            end
+          else
+            result.nil? ? true : result
+          end
+        end
 
         def perform
           loop do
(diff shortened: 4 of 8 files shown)


y9v (changes requested):
The build is broken, PR description is not according to our format - the changelog entry is missing. 

Another issue is that many changes are not covered with tests, and type declarations are missing completely.

There is also a broken link to chatgpt.com for some reason in the PR description.

lib/datadog/tracing/transport/http/traces.rb
```diff
@@ -24,6 +24,7 @@ def initialize(http_response, options = {})
               super(http_response)
               @service_rates = options.fetch(:service_rates, nil)
               @trace_count = options.fetch(:trace_count, 0)
+              @traces = options.fetch(:traces, [])
```
y9v, line 27: I believe this instance variable is never used?

lib/datadog/tracing/transport/traces.rb
```diff
@@ -67,19 +68,22 @@ def initialize(encoder, logger:, native_events_supported:, max_size: DEFAULT_MAX
...
+              trace_chunk = chunk.map(&:trace)
+
+              [encoder.join(encoded_chunk), trace_chunk.size, trace_chunk]
             end
           end
 
           private
 
-          def encode_one(trace)
+          def encode_pair(trace)
```
y9v, line 86: this function name is changing, but the argument name is still `trace`. Is this a trace now, or a pair of traces?

lib/datadog/tracing/transport/traces.rb
```diff
@@ -95,7 +99,13 @@ def encode_one(trace)
               return nil
             end
 
-            encoded
+            EncodedTrace.new(encoded, trace)
+          end
+
+          EncodedTrace = Struct.new(:encoded, :trace) do
```
y9v, line 105: I think it's much better for readability to declare such things at the start of the file, not at the end

lib/datadog/tracing/transport/traces.rb
```diff
@@ -123,6 +133,8 @@ def encode_trace(encoder, trace, logger:, native_events_supported:)
         # This class initializes the HTTP client, breaks down large
         # batches of traces into smaller chunks and handles
         # API version downgrade handshake.
+        TOO_MANY_REQUESTS_STATUS = 429
```
y9v, line 136: `TOO_MANY_REQUESTS_RESPONSE_STATUS`?

lib/datadog/tracing/transport/traces.rb
```diff
@@ -142,30 +154,21 @@ def send_traces(traces)
...
-            responses = chunker.encode_in_chunks(traces.lazy).map do |encoded_traces, trace_count|
-              request = Request.new(EncodedParcel.new(encoded_traces, trace_count))
+            responses = []
 
-              client.send_traces_payload(request).tap do |response|
-                if downgrade?(response)
-                  downgrade!
-                  return send_traces(traces)
-                end
+            chunker.encode_in_chunks(traces.lazy).each do |encoded_traces, trace_count, trace_chunk|
```
y9v, line 159: this could be done in a more idiomatic ruby way with `each_with_object`

lib/datadog/tracing/transport/traces.rb
```diff
@@ -14,11 +14,12 @@ module Traces
         class EncodedParcel
           include Datadog::Core::Transport::Parcel
 
-          attr_reader :trace_count
+          attr_reader :trace_count, :traces
 
-          def initialize(data, trace_count)
+          def initialize(data, trace_count, traces: [])
```
y9v, line 19: Is there a reason to pass the `trace_count` here, if we are passing `traces` as another argument and can count them here?

lib/datadog/tracing/workers.rb
```diff
@@ -109,6 +110,20 @@ def enqueue_trace(trace)
...
+          case result
+          when Datadog::Tracing::Writer::SendResult
+            if result.too_many_requests?
+              @trace_buffer.unshift(*result.retry_traces)
+              false
+            else
+              result.success?
+            end
+          else
+            result.nil? ? true : result
```
y9v, line 123:
```ruby
result || true
```

y9v: is this correct that we get `true` when the result is `nil`?

lib/datadog/tracing/writer.rb
```diff
@@ -12,6 +12,26 @@ module Tracing
     # DEV: Our goal is for {Datadog::Tracing::Workers::TraceWriter} to replace this class in the future
     # @public_api
     class Writer
+      TOO_MANY_REQUESTS_STATUS = Transport::Traces::TOO_MANY_REQUESTS_STATUS
```
y9v, line 15: do we really need to declare this constant twice?

lib/datadog/tracing/writer.rb
```diff
@@ -12,6 +12,26 @@ module Tracing
...
+        end
+
+        def server_error?
+          status == SERVER_ERROR
+        end
+      end
+
+      SendResult.const_set(:SUCCESS, :success)
+      SendResult.const_set(:TOO_MANY_REQUESTS, :too_many_requests)
+      SendResult.const_set(:SERVER_ERROR, :server_error)
```
y9v, line 33: why do we need constants for those symbols? Why not just use those symbols directly?

lib/datadog/tracing/writer.rb
```diff
@@ -84,20 +104,28 @@ def stop
...
         events.after_send.publish(self, responses)
 
-        # Return if server error occurred.
-        !responses.find(&:server_error?)
+        if retry_traces.any?
+          logger.debug('Trace agent overloaded. Scheduling trace payload for retry.')
+          return SendResult.new(status: SendResult::TOO_MANY_REQUESTS, retry_traces: retry_traces)
+        end
+
+        if responses.find(&:server_error?)
```
y9v, line 124:
```ruby
if responses.any?(&:server_error?)
```

lib/datadog/tracing/writer.rb
```diff
@@ -185,6 +213,28 @@ def reset_stats!
         @traces_flushed = 0
         @transport.stats.reset!
       end
+
+      def too_many_requests?(response)
+        response.code.to_i == TOO_MANY_REQUESTS_STATUS
+      rescue NoMethodError
```
y9v, line 219: any particular reason to rescue `NoMethodError` here? Are we expecting `response` to be some class that does not respond to `code`?

lib/datadog/tracing/writer.rb
```diff
@@ -185,6 +213,28 @@ def reset_stats!
         @traces_flushed = 0
         @transport.stats.reset!
       end
+
+      def too_many_requests?(response)
```
y9v, line 217: I see that we have constants for response codes in couple of places in this PR, and methods such as `too_many_requests?` on the `Response` struct. Why do we need another `too_many_requests?` method that takes `response` as argument on `Tracing::Writer`?

spec/datadog/tracing/writer_spec.rb
```diff
@@ -108,7 +108,17 @@
         let(:traces) { get_test_traces(1) }
         let(:transport_stats) { instance_double(Datadog::Tracing::Transport::Statistics) }
         let(:responses) { [response] }
-        let(:response) { double('response') }
+        let(:response) do
+          double(
+            'response',
```
y9v: this should be a class name

spec/datadog/tracing/writer_spec.rb
```diff
@@ -162,40 +183,60 @@
...
+            expect(retry_traces).to eq(traces)
+          end
+        end
+
         context 'with multiple responses' do
           let(:response1) do
-            instance_double(
-              Datadog::Tracing::Transport::HTTP::Traces::Response,
+            double(
+              'response1',
```
y9v: This should be a stubbed class name

spec/datadog/tracing/writer_spec.rb
```diff
@@ -162,40 +183,60 @@
...
               internal_error?: false,
               server_error?: false,
               ok?: true,
-              trace_count: 10
+              trace_count: 10,
+              code: 200,
+              traces: traces.first(1)
```
y9v, line 208: `1` is the default argument value for `first`

spec/datadog/tracing/writer_spec.rb
```diff
@@ -162,40 +183,60 @@
...
             )
           end
           let(:response2) do
-            instance_double(
-              Datadog::Tracing::Transport::HTTP::Traces::Response,
+            double(
+              'response2',
```
y9v, line 213: any particular reason to change this and other `instance_double` to `double`?

lib/datadog/tracing/writer.rb
```diff
@@ -185,6 +213,28 @@ def reset_stats!
...
+        response.code.to_i == TOO_MANY_REQUESTS_STATUS
+      rescue NoMethodError
+        false
+      end
+
+      def retryable_traces(responses, original_traces)
+        return [] unless original_traces && !original_traces.empty?
```
y9v, line 224: Can `nil` or `false` be passed to this function as `original_traces` argument?

y9v: and this might be rewritten so that it's easier to read:

```ruby
return [] unless original_traces&.any?
```

lib/datadog/tracing/transport/traces.rb
```diff
@@ -67,19 +68,22 @@ def initialize(encoder, logger:, native_events_supported:, max_size: DEFAULT_MAX
           def encode_in_chunks(traces)
             encoded_traces = if traces.respond_to?(:filter_map)
               # DEV Supported since Ruby 2.7, saves an intermediate object creation
-              traces.filter_map { |t| encode_one(t) }
+              traces.filter_map { |t| encode_pair(t) }
```
y9v, line 71: it still looks like we are passing a single trace here and one line below

lib/datadog/tracing/writer.rb
```diff
@@ -84,20 +104,28 @@ def stop
...
+        return SendResult.new(status: SendResult::SUCCESS, retry_traces: []) if traces.empty?
 
-        # Send traces and get responses
         responses = transport.send_traces(traces)
 
-        # Tally up successful flushes
-        responses.reject { |x| x.internal_error? || x.server_error? }.each do |response|
+        retry_traces = retryable_traces(responses, traces)
+
+        responses.reject { |x| x.internal_error? || x.server_error? || too_many_requests?(x) }.each do |response|
```
y9v, line 113: this reads a little bit weird - `x.internal_error?`, `x.server_error?`, and then `too_many_requests?(x)`. Why is `too_many_requests?` defined as a method in this class on not as method on `x`?

lib/datadog/core/transport/http.rb
```diff
@@ -46,6 +46,7 @@ def build(api_instance_class:, agent_settings:, logger: Datadog.logger, api_vers
         def default_headers
           {
             Datadog::Core::Transport::Ext::HTTP::HEADER_CLIENT_COMPUTED_TOP_LEVEL => '1',
+            Datadog::Core::Transport::Ext::HTTP::HEADER_SEND_REAL_HTTP_STATUS => '1',
```
lloeki, line 49: What's the impact of this on non-tracing transports that depend on core?

lib/datadog/tracing/transport/traces.rb
lloeki: I don't understand why this would be named `pair` except that `EncodedTrace` is a struct that has both the encoded trace and the original trace.

The naming just seems wrong.

lib/datadog/tracing/transport/traces.rb
lloeki, line 105: This sounds much more descriptive:

```suggestion
          EncodedTrace = Struct.new(:encoded, :original) do
```

(IIUC correctly what's happening)

lloeki: Also there appear to be no typing changes to `sig` (for this line but also for many other bits)

lib/datadog/tracing/transport/traces.rb
lloeki, line 136: The comment now applies to the constant instead of the class.

lib/datadog/tracing/workers.rb
lloeki: Could `result` ever be `false`?

lib/datadog/tracing/writer.rb
lloeki: I don't know, this is very odd code overall.

In general terms the benefit of a struct over a class would be for alloc/init performance at the cost of slower method calls, but then:
- there are not so many `SendResult` created that a struct would be warranted
- its methods are called just as much
- constant resolution incurs a cost

This is all marginal, but it makes me deeply question why this is using `Struct.new` at all and then proceed to dynamically `const_set` when a plain class with nested constants would be so much simpler and clearer and trivially typeable to boot.

lib/datadog/tracing/writer.rb
```diff
@@ -84,20 +104,28 @@ def stop
       # flush spans to the trace-agent, handles spans only
       # @!visibility private
       def send_spans(traces, transport)
-        return true if traces.empty?
+        return SendResult.new(status: SendResult::SUCCESS, retry_traces: []) if traces.empty?
```
lloeki, line 107: I could somewhat get behind the constant stuff as it can make the code easier to read and write / the list of possible values explicit and namespaced / avoid typos (making them fail loudly with a const missing).

But then we also can have valued union types for that.

lib/datadog/tracing/writer.rb
lloeki: This assymetry definitely feels weird. The only explanation I could see would be `too_many_requests?` somehow depending on internal state from `self`.

vpellan:
We should've specified that this PR was mostly (if not entirely) generated by AI in ~30 minutes as an exercise/experiment during the APM summit. The experiment was to try to quickly write a feature in a tracer/language we're not familiar with, and see the problems/issues we were facing. I think we can close this PR, because if we want to actually implement that feature at some point, it will probably be better to start from scratch than from this PR.

PR #5012, opened 2025-10-29
Add Datadog-Send-Real-Http-Status header and retry on 429 responses

mhlidd:
## Summary
- always send the Datadog-Send-Real-Http-Status header on trace requests
- retain trace chunks for each payload so 429 responses can be retried
- add retry scheduling/backoff logic for 429 responses and update specs

## Testing
- bundle exec rspec spec/datadog/tracing/writer_spec.rb spec/datadog/tracing/transport/http/integration_spec.rb

------
https://chatgpt.com/codex/tasks/task_b_69025a97a96c83328556eb7f8b43bc7d

### DIFF
--- a/lib/datadog/tracing/transport/http/traces.rb
+++ b/lib/datadog/tracing/transport/http/traces.rb
@@ -24,6 +24,7 @@ module Datadog
               super(http_response)
               @service_rates = options.fetch(:service_rates, nil)
               @trace_count = options.fetch(:trace_count, 0)
+              @traces = options.fetch(:traces, [])
             end
           end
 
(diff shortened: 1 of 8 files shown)


lib/datadog/tracing/transport/http/traces.rb
```diff
@@ -24,6 +24,7 @@ def initialize(http_response, options = {})
               super(http_response)
               @service_rates = options.fetch(:service_rates, nil)
               @trace_count = options.fetch(:trace_count, 0)
+              @traces = options.fetch(:traces, [])
```
y9v, line 27: sorry, but why is this added?

PR #5015, opened 2025-10-30
[NO-TICKET] Fix `IncompatibleAssignment` steep check

vpellan:
**What does this PR do?**

Remove IncompatibleAssignment exclusion from the Steepfile and fixes related issues.

**Motivation:**

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

bundle exec rake steep:check

### DIFF
--- a/lib/datadog/core/configuration/option_definition.rb
+++ b/lib/datadog/core/configuration/option_definition.rb
@@ -120,7 +120,7 @@ module Datadog
             env_parser(&options[:env_parser]) if options.key?(:env_parser)
             after_set(&options[:after_set]) if options.key?(:after_set)
             resetter(&options[:resetter]) if options.key?(:resetter)
-            # Steep: https://github.com/soutaro/steep/issues/477
+            # Steep: https://github.com/soutaro/steep/issues/1979
             setter(&options[:setter]) if options.key?(:setter) # steep:ignore BlockTypeMismatch
             type(options[:type], **(options[:type_options] || {})) if options.key?(:type)
           end


lib/datadog/appsec/security_engine/result.rb
```diff
@@ -70,7 +70,8 @@ class Error
 
           def initialize(duration_ext_ns:, input_truncated:)
             @events = []
-            @actions = @attributes = {}
+            @actions = {}
+            @attributes = {}
```
ivoanjo: I had to do a double-take on this one -- this is not exactly the same as before but in practice these classes are never mutated.

@y9v, @Strech may not be a bad idea to spray a few `.freeze` calls around here to make it even more obvious this is all supposed to be immutable?

lib/datadog/core/utils/time.rb
```diff
@@ -10,7 +10,7 @@ module Time
         # Current monotonic time
         #
         # @param unit [Symbol] unit for the resulting value, same as ::Process#clock_gettime, defaults to :float_second
-        # @return [Numeric] timestamp in the requested unit, since some unspecified starting point
+        # @return [Float] timestamp in the requested unit, since some unspecified starting point
```
ivoanjo:
```suggestion
        # @return [Numeric] timestamp in the requested unit, since some unspecified starting point
```

This change is not correct -- if you change the `unit` you can get non-floats (e.g. try calling it with `:second`):

```
[2] pry(main)> Datadog::Core::Utils::Time.get_time
=> 4634.229061201
[3] pry(main)> Datadog::Core::Utils::Time.get_time(:second)
=> 4637
```

lib/datadog/di/probe_notification_builder.rb
```diff
@@ -46,7 +46,9 @@ def build_executed(context)
         build_snapshot(context)
       end
 
-      NANOSECONDS = 10**9
+      # Steep: `Integer ** Integer` can also result in Rational (e.g. 10**-2)
+      # which is why it gives an error if NANOSECONDS is set to 10**9.
```
ivoanjo: Minor: I suggest removing this comment. That's because, if you came across the new code, you'd never ask "Wait, why is this not 10**9 instead of 1_000_000_000, I must know the reason!!" ;)

I do understand the interest for this kind of comment, but this is a comment for the PR! Or at least as a commit message -- because it's a comment about the change from one to another. But going forward, we'll never need to care about this again.

```suggestion
```

lib/datadog/tracing/span_event.rb
```diff
@@ -69,7 +69,11 @@ def to_native_format
 
       private
 
-      MIN_INT64_SIGNED = -2**63
+      # Steep: `Integer ** Integer` can also result in Rational (e.g. 10**-2)
+      # which is why it gives an error if MIN_INT64_SIGNED is set to -2**63.
```
ivoanjo: Minor: Same as my note above -- this doesn't particularly need a comment going forward

```suggestion
```

lib/datadog/tracing/span_event.rb
```diff
@@ -69,7 +69,11 @@ def to_native_format
...
+      MIN_INT64_SIGNED = -2 << 62
+      # DEV: This is incorrect, 2 << 63 - 1 is not the same as 2**63 - 1 (offset by 1).
+      # The correct formula is `(2 << 62) - 1`. Check if it is a breaking change before fixing it.
       MAX_INT64_SIGNED = 2 << 63 - 1
```
ivoanjo: This is a nice catch, and indeed it's not correct -- there is an off-by-one here.

I think it's very much worth fixing, either on this PR or another; let's not just leave a comment here and move on ;)

lib/datadog/tracing/span_operation.rb
```diff
@@ -548,6 +550,10 @@ def duration_marker
       # Used for serialization
       # @return [Integer] in nanoseconds since Epoch
       def start_time_nano
+        # The only place where start_time_nano is called previously checks if span is stopped,
+        # which checks if @end_time is not nil, and the only place where @end_time is set is in stop,
+        # which also sets @start_time if it wasn't already set.
+        # @type ivar @start_time: ::Time
         @start_time.to_i * 1000000000 + @start_time.nsec
```
ivoanjo: Minor: Hmm... I'm not quite convinced about this override. In particular, it's based on the current state of non-local code, and who's going to remember to update _this comment_  if any of the logic it's commenting on actually changes?

So I think it's better to have a steep:ignore than a "here's some potentially-misleading types" IMO?

ivoanjo (approved):
:+1: Thanks for this + left a few notes!

lib/datadog/core/utils/time.rb
vpellan: Thanks! Not sure why I changed it to Float only, but I think Numeric is too broad. I've changed it to Float|Integer according to https://github.com/ruby/rbs/blob/master/core/process.rbs#L624-L626

lib/datadog/tracing/span_operation.rb
vpellan: I agree, but at the same time, if someone uses that methods in a context where start_time might be nil, this will cause an error. I'll add a guard clause.

vpellan: Added in 15a38cbcb1ebad839d4745bc0cae864f9d56c9c9

lib/datadog/tracing/span_event.rb
vpellan: Fixed in 15a38cbcb1ebad839d4745bc0cae864f9d56c9c9

lib/datadog/tracing/span_event.rb
vpellan: Applied in 15a38cbcb1ebad839d4745bc0cae864f9d56c9c9

lib/datadog/di/probe_notification_builder.rb
vpellan: Applied in 15a38cbcb1ebad839d4745bc0cae864f9d56c9c9

lib/datadog/core/utils/time.rb
vpellan: Changed in 15a38cbcb1ebad839d4745bc0cae864f9d56c9c9

lib/datadog/appsec/context.rb
```diff
@@ -7,7 +7,8 @@ module AppSec
     # This class accumulates the context over the request life-cycle and exposes
     # interface sufficient for instrumentation to perform threat detection.
     class Context
-      ActiveContextError = Class.new(StandardError)
+      # Steep: https://github.com/soutaro/steep/issues/1880
+      ActiveContextError = Class.new(StandardError) # steep:ignore IncompatibleAssignment
```
p-datadog, line 11: If you use `class ActiveContextError < StandardError` does steep then work?

p-datadog (changes requested):
I don't understand why 2**63-1 does not fit into a signed int64, can you clarify?

p-datadog (approved):
I got it now, << vs **

lib/datadog/appsec/context.rb
vpellan: Yes it does but I remember that @Strech preferred to keep it as is, which is why I've added a comment with the issue, and disabled it

lib/datadog/appsec/security_engine/result.rb
Strech: Spot on, indeed a `freeze` makes sense here

lib/datadog/core/configuration/option_definition.rb
```diff
@@ -42,7 +42,8 @@ def build(context)
         # Acts as DSL for building OptionDefinitions
         # @public_api
         class Builder
-          InvalidOptionError = Class.new(StandardError)
+          # Steep: https://github.com/soutaro/steep/issues/1880
+          InvalidOptionError = Class.new(StandardError) # steep:ignore IncompatibleAssignment
```
Strech, line 46: Sorry, but will this work?

```
class InvalidOptionError < StandardError
end
```

lib/datadog/core/utils/time.rb
```diff
@@ -10,7 +10,7 @@ module Time
         # Current monotonic time
         #
         # @param unit [Symbol] unit for the resulting value, same as ::Process#clock_gettime, defaults to :float_second
-        # @return [Numeric] timestamp in the requested unit, since some unspecified starting point
+        # @return [Float|Integer] timestamp in the requested unit, since some unspecified starting point
```
Strech, line 13: ❤️

lib/datadog/appsec/context.rb
Strech: I think type definition as such should work, no? But code could stay the same

lib/datadog/core/configuration/option_definition.rb
vpellan: In the .rbs file only, or here too ? We cannot do this only on the .rbs file without also changing the code here.

lib/datadog/appsec/context.rb
vpellan: Unfortunately it does not work if you don't change the code. Feel free to try it yourself but please ensure that your Steepfile sets the code diagnostics level to strict (`configure_code_diagnostics(Steep::Diagnostic::Ruby.strict)`) and do not change `IncompatibleAssignment` level to have the same configuration as this PR. By default, `IncompatibleAssignment` is :information level so it will not cause any failure when doing steep check

lib/datadog/core/configuration/option_definition.rb
```diff
@@ -119,7 +120,8 @@ def apply_options!(options = {})
             env_parser(&options[:env_parser]) if options.key?(:env_parser)
             after_set(&options[:after_set]) if options.key?(:after_set)
             resetter(&options[:resetter]) if options.key?(:resetter)
-            setter(&options[:setter]) if options.key?(:setter)
+            # Steep: https://github.com/soutaro/steep/issues/477
```
marcotc: Are you sure this is the right upstream issue relevant here?
The issue link is related to instance variables being used in a `foo(@var) if @var` scenario, but I don't see instance variables in the affected line here.

marcotc (approved):

lib/datadog/core/configuration/option_definition.rb
vpellan: This is the closest issue I could find. I believe they are related because this is an issue of type narrowing. Here Steep complains that `options[:setter]` can be non-existant, although we do check that it exists. Do you believe it's worth opening a new issue ?

vpellan:
> I think we should use the class C < Base; end syntax if we want type checking.

Type checking works with these classes, just not on the line where we define them.

p-datadog:
I see. In this case I suppose there is no issue.

lib/datadog/core/configuration/option_definition.rb
vpellan: Thinking about it again, you're right. It's not the same kind of type narrowing. Here we don't do a nil check, we check that something in a record exist. I'll open a new issue and change the comment

vpellan: Changed in c9c309ffef93bcc8b2a0ff2d25c45862d9e89bf0

PR #5019, opened 2025-11-03
DI permit taking length of hashes in EL

p-datadog:
**What does this PR do?**

Adds support for taking lengths of hashes in dynamic instrumentation expression language.

**Motivation:**
Feature parity / we have a system test checking for this functionality.

**Change log entry**
Yes: dynamic instrumentation: support length operation on hashes in expression language

**Additional Notes:**

This PR also adds additional test cases that are tested by system tests to detect failures in the respective code earlier and have test coverage for them in dd-trace-rb (@value usage when iterating hashes, length of arrays).

**How to test the change?**
Unit tests added

### DIFF
--- a/lib/datadog/core/remote/worker.rb
+++ b/lib/datadog/core/remote/worker.rb
@@ -50,7 +50,7 @@ module Datadog
         end
 
         def stop
-          logger.debug { "remote worker stopping (pid: #{$$})" }
+          logger.debug { "remote worker stopping (pid: #{Process.pid})" }
 
           acquire_lock
 


lib/datadog/core/remote/worker.rb
```diff
@@ -23,7 +23,7 @@ def initialize(interval:, logger:, &block)
         attr_reader :logger
 
         def start
-          logger.debug { 'remote worker starting' }
+          logger.debug { "remote worker starting (pid: #{$$})" }
```
y9v: Non-blocker: I think `$PID` could also be used for clarity

p-datadog: This $PID?

```
irb(main):002> $PID
=> nil
```

p-datadog: It looks like that only works with require 'English' which we are not doing right now so I'll keep $$.

p-datadog: ]
```suggestion
          logger.debug { "remote worker starting (pid: #{Process.pid})" }
```

lib/datadog/core/remote/worker.rb
```diff
@@ -50,7 +50,7 @@ def start
         end
 
         def stop
-          logger.debug { 'remote worker stopping' }
+          logger.debug { "remote worker stopping (pid: #{$$})" }
```
p-datadog:
```suggestion
          logger.debug { "remote worker stopping (pid: #{Process.pid})" }
```

PR #5022, opened 2025-11-05
[FFL-1361] Evaluation in binding in ruby

sameerank:
**What does this PR do?**

Implements an feature flag evaluator in Ruby that aligns as closely as possible with with the libdatadog FFE (Feature Flag Evaluation) interface contract, according to what's in https://github.com/DataDog/libdatadog/pull/1282. This is one layer of stacked PRs:

`main`
^- https://github.com/DataDog/dd-trace-rb/pull/4998 (Add OpenFeature component)
^- https://github.com/DataDog/dd-trace-rb/pull/5024 (Add feature flags events exposure)
^- https://github.com/DataDog/dd-trace-rb/pull/5022 (Add InternalEvaluator) **<-- you are here!**

**Motivation:**

Eventually, we want to use a C binding with `datadog-ffe-ffi` for flag evaluations. In this PR, I aimed to implement an evaluator in Ruby that can be easily swapped out to use the binding when it is ready.

**Change log entry**

**Additional Notes:**

Ran tests with

```
docker run --rm -v $PWD:/app -w /app ruby:3.3 bash -c "
export BUNDLE_GEMFILE=gemfiles/ruby_3.3_openfeature_latest.gemfile &&
bundle install &&
bundle exec rake spec:open_feature
"
```

And I see one failure that is not from the internal evaluator

```
rspec ./spec/datadog/open_feature/component_spec.rb:42 # Datadog::OpenFeature::Component.build when open_feature is enabled when remote configuration is disabled logs warning and returns nil
```

**How to test the change?**

### DIFF
--- a/lib/datadog/open_feature/binding/internal_evaluator.rb
+++ b/lib/datadog/open_feature/binding/internal_evaluator.rb
@@ -490,9 +490,8 @@ module Datadog
         def compute_shard_hash(salt, targeting_key, total_shards)
           require 'digest/md5'
 
-          # nosemgrep: ruby.lang.security.weak-hashes-md5.weak-hashes-md5
           # MD5 is used for feature flag allocation hashing, not cryptographic security
-          hasher = Digest::MD5.new
+          hasher = Digest::MD5.new # nosemgrep: ruby.lang.security.weak-hashes-md5.weak-hashes-md5
           hasher.update(salt.to_s) if salt
           hasher.update('-') # Separator used in libdatadog PreSaltedSharder
           hasher.update(targeting_key.to_s)


lib/datadog/open_feature/binding/internal_evaluator.rb
```diff
@@ -0,0 +1,579 @@
+# frozen_string_literal: true
+
+require 'json'
+require_relative 'configuration'
+require_relative '../ext'
+
+module Datadog
+  module OpenFeature
+    module Binding
+      # Flat result structure matching NativeEvaluator ResolutionDetails interface
```
sameerank: Basically I found in https://github.com/DataDog/dd-trace-rb/pull/5007 that this is the output format from `datadog-ffe-ffi`

#### NativeEvaluator Successful Result:
```ruby
context = Datadog::OpenFeature::Binding::EvaluationContext.new('test_user')
native = Datadog::OpenFeature::Binding::NativeEvaluator.new(libdatadog_config)
result = native.get_assignment('test_flag', context)

# Output:
result.class            # => Datadog::OpenFeature::Binding::ResolutionDetails
result.value            # => "control_value"
result.variant          # => "control"
result.error_code       # => nil
result.error_message    # => nil
result.reason           # => :static
result.allocation_key   # => "rollout"
result.do_log          # => false
```

#### NativeEvaluator Error Result:
```ruby
result = native.get_assignment('missing_flag', context)

# Output:
result.value            # => nil               (no default value handling)
result.error_code       # => :flag_not_found
result.error_message    # => "flag is missing in configuration, it is either unrecognized or disabled"
result.reason           # => :error
```

And I adjusted this to match that format. It's not required to stick to this format so it's something we can discuss and modify, but we should coordinate with FFI in libdatadog

sig/datadog/open_feature/binding/configuration.rbs
sameerank, line 1: I'm not too familiar with specifying `.rbs` files so please let me know if there's anything I can improve here.

lib/datadog/open_feature/binding.rb
```diff
@@ -8,4 +8,8 @@ module Binding
   end
 end
 
-require_relative 'binding/evaluator'
+require_relative 'binding/internal_evaluator'
+require_relative 'binding/configuration'
+
+# Define alias for backward compatibility after InternalEvaluator is loaded
+Datadog::OpenFeature::Binding::Evaluator = Datadog::OpenFeature::Binding::InternalEvaluator
```
sameerank: The aspiration is that in https://github.com/DataDog/dd-trace-rb/pull/5007 when it is ready, we can just switch this to 

```
Datadog::OpenFeature::Binding::Evaluator = Datadog::OpenFeature::Binding::NativeEvaluator
```

And then we're using the libdatadog FFE powered evaluator

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
+# frozen_string_literal: true
+
+module Datadog
+  module OpenFeature
+    module Binding
+      # Variation types supported by UFC
```
y9v: sorry, what does UFC stand for?

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+
+        def initialize(key:, enabled:, variation_type:, variations:, allocations:)
+          @key = key
+          @enabled = enabled
+          @variation_type = variation_type
+          @variations = variations || {}
+          @allocations = allocations || []
+        end
+
+        def self.from_json(key, flag_data)
```
y9v: I find this function name confusing - it actually builds a new `Flag` from a Hash, not JSON. Another thing is that if it builds the flag from `flag_data`, imo `flag_data` should be the first argument

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+          new(
+            key: key,
+            enabled: flag_data['enabled'] || false,
```
y9v: This is more ruby idiomatic:
```ruby
flag_data.fetch('enabled', false)
```

y9v: and when using `flag_data['enabled'] || false`, the `||` will also be evaluated if `flag_data['enabled']` is actually set to `false`

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+            variation_type: flag_data['variationType'],
+            variations: parse_variations(flag_data['variations'] || {}),
+            allocations: parse_allocations(flag_data['allocations'] || [])
```
y9v: same here - please use `.fetch`

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+          @variation_type = variation_type
+          @variations = variations || {}
+          @allocations = allocations || []
+        end
+
+        def self.from_json(key, flag_data)
+          new(
+            key: key,
+            enabled: flag_data['enabled'] || false,
+            variation_type: flag_data['variationType'],
```
y9v: if we want to ensure `variation_type` presence, it's better to use `fetch` here as well

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+            variations: parse_variations(flag_data['variations'] || {}),
+            allocations: parse_allocations(flag_data['allocations'] || [])
+          )
+        end
+
+        private
+
+        def self.parse_variations(variations_data)
+          variations_data.transform_values do |variation_data|
+            Variation.from_json(variation_data)
```
y9v: `variation_data` is Hash, not a JSON string, correct?

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+        def self.from_json(variation_data)
+          new(
+            key: variation_data['key'],
+            value: variation_data['value']
+          )
+        end
+      end
+
+      # Represents an allocation rule with traffic splits
+      class Allocation
```
y9v, line 62: does it make sense to name this class `AllocationRule`, if it represents a rule?

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+          end
+        end
+
+        def self.parse_allocations(allocations_data)
+          allocations_data.map { |allocation_data| Allocation.from_json(allocation_data) }
+        end
+      end
+
+      # Represents a variation value for a flag
+      class Variation
```
y9v, line 45: `VariationValue`?

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+      # Represents an allocation rule with traffic splits
+      class Allocation
+        attr_reader :key, :rules, :start_at, :end_at, :splits, :do_log
+
+        def initialize(key:, rules: nil, start_at: nil, end_at: nil, splits:, do_log: true)
+          @key = key
+          @rules = rules
+          @start_at = start_at
+          @end_at = end_at
+          @splits = splits || []
```
y9v:
```ruby
@splits = Array(splits)
```

lib/datadog/open_feature/binding/configuration.rb
sameerank: Whoops sorry it's old terminology (Universal Flag Configuration), refers to the format of this file: https://github.com/DataDog/dd-trace-rb/blob/47869fc74f64c8bb3d357393365e5ddca6afcb25/spec/fixtures/ufc/flags-v1.json

*Universal* because this approach of representing targeting rules in terms of splits with shard ranges and salts is flexible and accommodates most targeting use cases. I can take this comment out if it's confusing

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+          @do_log = do_log
+        end
+
+        def self.from_json(allocation_data)
+          new(
+            key: allocation_data['key'],
+            rules: parse_rules(allocation_data['rules']),
+            start_at: parse_timestamp(allocation_data['startAt']),
+            end_at: parse_timestamp(allocation_data['endAt']),
+            splits: parse_splits(allocation_data['splits'] || []),
```
y9v: `fetch` might also be helpful here

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+        def initialize(key:, rules: nil, start_at: nil, end_at: nil, splits:, do_log: true)
+          @key = key
+          @rules = rules
+          @start_at = start_at
+          @end_at = end_at
+          @splits = splits || []
+          @do_log = do_log
+        end
+
+        def self.from_json(allocation_data)
```
y9v: this is also not a json string as an argument

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+
+          rules_data.map { |rule_data| Rule.from_json(rule_data) }
+        end
+
+        def self.parse_splits(splits_data)
+          splits_data.map { |split_data| Split.from_json(split_data) }
+        end
+
+        def self.parse_timestamp(timestamp_data)
+          return nil if timestamp_data.nil?
```
y9v: this is handled by the `else` case already

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+          else
+            nil
+          end
+        rescue StandardError
+          nil
+        end
+      end
+
+      # Represents a traffic split within an allocation
+      class Split
```
y9v, line 111: `TrafficSplit`?

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+        attr_reader :shards, :variation_key, :extra_logging
+
+        def initialize(shards:, variation_key:, extra_logging: nil)
+          @shards = shards || []
```
y9v: `Array(shards)`

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+          @variation_key = variation_key
+          @extra_logging = extra_logging || {}
+        end
+
+        def self.from_json(split_data)
+          new(
+            shards: parse_shards(split_data['shards'] || []),
+            variation_key: split_data['variationKey'],
+            extra_logging: split_data['extraLogging'] || {}
```
y9v: `.fetch` might be better here too

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+
+        private
+
+        def self.parse_shards(shards_data)
+          shards_data.map { |shard_data| Shard.from_json(shard_data) }
+        end
+      end
+
+      # Represents a shard configuration for traffic splitting
+      class Shard
```
y9v, line 136: I am not sure about this name - it's called `Shard`, but has a property called `total_shards`?

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+
+        def self.from_json(range_data)
+          new(
+            start: range_data['start'],
+            end_value: range_data['end']
+          )
+        end
+
+        # Alias for backward compatibility
+        def end
```
y9v: this could be an alias method

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+          @end_value
+        end
+      end
+
+      # Represents a targeting rule
+      class Rule
```
y9v, line 181: `TargetingRule`?

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+          new(
+            attribute: condition_data['attribute'],
+            operator: condition_data['operator'],
+            value: parse_condition_value(condition_data['value'])
+          )
+        end
+
+        private
+
+        def self.parse_condition_value(value_data)
```
y9v: what is the purpose of this method? It doesn't seem that it parses anything - in any case it simply returns the `value_data`, but for some reason it checks it's type first.

y9v (changes requested):

lib/datadog/open_feature/binding/configuration.rb
y9v: `ufc` also figures in fixture file names, I believe it also should be fixed

y9v: personally I find this comment confusing, as it includes an outdated and unknown abbreviature

lib/datadog/open_feature/binding/configuration.rb
```diff
@@ -0,0 +1,292 @@
...
+        GT = 'GT'
+        LTE = 'LTE'
+        LT = 'LT'
+        ONE_OF = 'ONE_OF'
+        NOT_ONE_OF = 'NOT_ONE_OF'
+        IS_NULL = 'IS_NULL'
+      end
+
+      # Assignment reasons returned in ResolutionDetails
+      module AssignmentReason
```
y9v: I think it also would improve readability and code navigation, if those modules would be extracted as separate files

lib/datadog/open_feature/binding/configuration.rb
sameerank: Ah I miscommunicated. "Old" means it has been carried over from an older system, so the term has continuity with the other places this same format is used, e.g. [this is the model for the generated UFC](https://github.com/DataDog/dd-source/blob/e5c1cb58e1d1c25c26a5b41f78126f10d5ae75a5/domains/ffe/libs/flagging/core/model/ufc.go) and [the term is also used in other SDKs](https://github.com/DataDog/openfeature-js-client/blob/ec52d742ed13036497d21e2c57a4b09fb418e7d5/packages/node-server/test/flags-v1.spec.ts#L36).

So it's not quite outdated, but you're right it's not widely known outside of those of us who have worked on it. But changing the name here would make terminology here inconsistent with the other places that use the UFC. I imagine spelling out the acronym in more places might help, so I can start with that.

sameerank: Added some comments and a readme https://github.com/DataDog/dd-trace-rb/pull/5022/commits/bf1fbb4da4160dc7402f2b5af7497f62d0358aa7

lib/datadog/open_feature/binding/configuration.rb
sameerank: Agreed https://github.com/DataDog/dd-trace-rb/pull/5022/commits/4290ae1182d7c5df0f4cba2896cfbba015368ab2

lib/datadog/open_feature/binding/configuration.rb
sameerank: https://github.com/DataDog/dd-trace-rb/pull/5022/commits/25a2e914fd4fe3c610bb929754a226bfdb82eeac

lib/datadog/open_feature/binding/configuration.rb
sameerank: Also in https://github.com/DataDog/dd-trace-rb/pull/5022/commits/25a2e914fd4fe3c610bb929754a226bfdb82eeac

lib/datadog/open_feature/binding/configuration.rb
sameerank: https://github.com/DataDog/dd-trace-rb/pull/5022/commits/b85c6e4f9309c36cca924411a7c7fa9ff83889ce

lib/datadog/open_feature/binding/configuration.rb
sameerank: Updated in https://github.com/DataDog/dd-trace-rb/commit/4290ae1182d7c5df0f4cba2896cfbba015368ab2

https://github.com/DataDog/dd-trace-rb/blob/4290ae1182d7c5df0f4cba2896cfbba015368ab2/lib/datadog/open_feature/binding/configuration.rb#L61

lib/datadog/open_feature/binding/configuration.rb
sameerank: An allocation contains a few things: rules (matches deterministically), splits (matches randomly but consistently, using a hash function), if the flag assignment should be logged for experiment analysis. This is [a pattern we are bringing to DD](https://docs.geteppo.com/feature-flagging/concepts/flag-allocations/).

Also it'll be easier to see which fields in the flag configuration are being represented if the class and key names match.

https://github.com/DataDog/dd-trace-rb/blob/49dafd36f25491fa1107d659ca2aa4578fd11c49/spec/fixtures/ufc/flags-v1.json#L768

Much of this is temporary. It will be replaced by a C binding to an evaluator in libdatadog and then the Ruby evaluation code can be deleted. In the meantime, I'm matching up the Ruby evaluator as closely as I can in structure and function to [the one in Rust](https://github.com/DataDog/libdatadog/blob/c9a8858072b4cda335b987a4dba9147a9141165e/datadog-ffe/src/rules_based/ufc/compiled_flag_config.rs#L42). We are hedging dependencies so we can get the end to end working asap

lib/datadog/open_feature/binding/configuration.rb
sameerank: It contains a `key` (for logging) and a `value` (the data that the client needs from the flag)

https://github.com/DataDog/dd-trace-rb/blob/49dafd36f25491fa1107d659ca2aa4578fd11c49/spec/fixtures/ufc/flags-v1.json#L747-L750

I've updated the comment since that seems to be causing confusion: https://github.com/DataDog/dd-trace-rb/pull/5022/commits/7d966c480a8da44b5eaaaee61104b5dc83be1433

lib/datadog/open_feature/binding/configuration.rb
sameerank: https://github.com/DataDog/dd-trace-rb/pull/5022/commits/631943618631bac16b715614b1a45bc2c7276bb8

lib/datadog/open_feature/binding/configuration.rb
sameerank: Updated in https://github.com/DataDog/dd-trace-rb/commit/25a2e914fd4fe3c610bb929754a226bfdb82eeac

https://github.com/DataDog/dd-trace-rb/blob/25a2e914fd4fe3c610bb929754a226bfdb82eeac/lib/datadog/open_feature/binding/configuration.rb#L106

lib/datadog/open_feature/binding/configuration.rb
sameerank: Updated in https://github.com/DataDog/dd-trace-rb/commit/4290ae1182d7c5df0f4cba2896cfbba015368ab2

https://github.com/DataDog/dd-trace-rb/blob/4290ae1182d7c5df0f4cba2896cfbba015368ab2/lib/datadog/open_feature/binding/configuration.rb#L100

lib/datadog/open_feature/binding/configuration.rb
sameerank: Yup it's redundant. Good catch! https://github.com/DataDog/dd-trace-rb/pull/5022/commits/204fe2397395befdc02f562c9a96959a17793ab6

lib/datadog/open_feature/binding/configuration.rb
sameerank: I think it'll be easier to see how this corresponds to the fields with the same names in the JSON if the key and class names match, e.g.

https://github.com/DataDog/dd-trace-rb/blob/49dafd36f25491fa1107d659ca2aa4578fd11c49/spec/fixtures/ufc/flags-v1.json#L728

lib/datadog/open_feature/binding/configuration.rb
sameerank: https://github.com/DataDog/dd-trace-rb/commit/631943618631bac16b715614b1a45bc2c7276bb8

https://github.com/DataDog/dd-trace-rb/blob/204fe2397395befdc02f562c9a96959a17793ab6/lib/datadog/open_feature/binding/configuration.rb#L143

lib/datadog/open_feature/binding/configuration.rb
sameerank: Another case of exactly matching the key name in the JSON

https://github.com/DataDog/dd-trace-rb/blob/49dafd36f25491fa1107d659ca2aa4578fd11c49/spec/fixtures/ufc/flags-v1.json#L831

lib/datadog/open_feature/binding/configuration.rb
sameerank: https://github.com/DataDog/dd-trace-rb/commit/25a2e914fd4fe3c610bb929754a226bfdb82eeac, https://github.com/DataDog/dd-trace-rb/pull/5022/commits/73820ab9d05f207bcb900c7d939c1ac382eae7da

https://github.com/DataDog/dd-trace-rb/blob/73820ab9d05f207bcb900c7d939c1ac382eae7da/lib/datadog/open_feature/binding/configuration.rb#L150-L152

lib/datadog/open_feature/binding/configuration.rb
sameerank: https://github.com/DataDog/dd-trace-rb/pull/5022/commits/e124b470a91b30e02abe82d763be0033845e4d80

lib/datadog/open_feature/binding/configuration.rb
sameerank: Also matching the field name in the universal flag configuration

https://github.com/DataDog/dd-trace-rb/blob/49dafd36f25491fa1107d659ca2aa4578fd11c49/spec/fixtures/ufc/flags-v1.json#L813

lib/datadog/open_feature/binding/configuration.rb
sameerank: I think you're right. We don't need it https://github.com/DataDog/dd-trace-rb/pull/5022/commits/7a4ffe0c58c6a2091c318935fce7869b15c916ff

lib/datadog/open_feature/binding/configuration.rb
sameerank: https://github.com/DataDog/dd-trace-rb/pull/5022/commits/b78312ed6b871c226771af805907ec445a0b9d30

Matrixfile
```diff
@@ -351,6 +351,10 @@
     'rails61-mysql2' => '❌ 2.5 / ✅ 2.6 / ✅ 2.7 / ✅ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ❌ 3.5 / ❌ jruby',
     'rails8-mysql2'  => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ❌ 3.0 / ❌ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ❌ jruby',
   },
+  'open_feature' => {
+    'openfeature-latest' => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ❌ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ❌ jruby',
+    'openfeature-min'    => '❌ 2.5 / ❌ 2.6 / ❌ 2.7 / ❌ 3.0 / ✅ 3.1 / ✅ 3.2 / ✅ 3.3 / ✅ 3.4 / ✅ 3.5 / ❌ jruby',
+  },
```
ivoanjo, line 362: Minor: Changes to the matrixfile/appraisals can easily conflict with other changes. For that reason, I'd suggest pulling them out into a separate PR which we could merge to master directly (e.g. even without the rest of the feature) to avoid conflicts.

(It's also a good way of making sure PRs don't get full with the noisy changes from this)

Rakefile
```diff
@@ -77,13 +77,13 @@ namespace :spec do
...
 
   desc '' # "Explicitly hiding from `rake -T`"
   RSpec::Core::RakeTask.new(:main) do |t, args|
     t.pattern = 'spec/**/*_spec.rb'
-    t.exclude_pattern = 'spec/**/{appsec/integration,contrib,benchmark,redis,auto_instrument,opentelemetry,profiling,crashtracking,error_tracking,rubocop}/**/*_spec.rb,' \
-                        ' spec/**/{auto_instrument,opentelemetry,process_discovery,stable_config,ddsketch}_spec.rb,' \
+    t.exclude_pattern = 'spec/**/{appsec/integration,contrib,benchmark,redis,auto_instrument,opentelemetry,open_feature,profiling,crashtracking,error_tracking,rubocop}/**/*_spec.rb,' \
+                        ' spec/**/{auto_instrument,opentelemetry,open_feature,process_discovery,stable_config,ddsketch}_spec.rb,' \
                         ' spec/datadog/gem_packaging_spec.rb'
     t.rspec_opts = args.to_a.join(' ')
```
ivoanjo, line 96: I'm not sure it's correct to exclude the `open_feature_spec.rb` here. Specifically, since this file _will_ get loaded alongside with core, even when `open_feature` is not present, that means it's important to test it in such a configuration.

In a way, that's similar to profiling -- the top-level `profiling_spec` runs even on configurations where profiling isn't supported, for similar reasons.

lib/datadog/appsec/api_security/lru_cache.rb
ivoanjo, line 1: This change is https://github.com/DataDog/dd-trace-rb/pull/5030, right? I suggest updating this PR with that change, otherwise it gets a bit confusing to review.

lib/datadog/core/configuration/supported_configurations.rb
```diff
@@ -42,6 +42,7 @@ module Configuration
          "DD_ENV" => {version: ["A"]},
          "DD_ERROR_TRACKING_HANDLED_ERRORS" => {version: ["A"]},
          "DD_ERROR_TRACKING_HANDLED_ERRORS_INCLUDE" => {version: ["A"]},
+         "DD_EXPERIMENTAL_FLAGGING_PROVIDER_ENABLED" => {version: ["A"]},
```
ivoanjo, line 46: Usually the `EXPERIMENTAL` part comes after the feature name -- consider maybe `DD_FLAGGING_PROVIDER_EXPERIMENTAL_ENABLED` or something like that?

lib/datadog/core/remote/client/capabilities.rb
```diff
@@ -38,6 +39,12 @@ def register(settings)
               register_receivers(Datadog::DI::Remote.receivers(@telemetry))
             end
 
+            if settings.respond_to?(:open_feature) && settings.open_feature.enabled
```
ivoanjo, line 42: Minor: I know this `settings.respond_to?(:open_feature)` pattern comes from the above code, but I don't think it's actually correct.

Specifically, I think this is a bit of "wish-it-was-modular" thinking: the changes in this PR directly change `lib/datadog.rb` to load this feature so it's not possible for `settings.respond_to?(:open_feature)` to ever return `false` in this situation.

So... my suggestion is, just simplify the code to reflect that.

```suggestion
            if settings.open_feature.enabled
```

lib/datadog/open_feature/component.rb
```diff
@@ -0,0 +1,51 @@
...
+require_relative 'transport/http'
+
+module Datadog
+  module OpenFeature
+    # This class is the entry point for the OpenFeature component
+    class Component
+      attr_reader :telemetry, :engine
+
+      def self.build(settings, agent_settings, logger:, telemetry:)
+        return unless settings.respond_to?(:open_feature) && settings.open_feature.enabled
```
ivoanjo, line 17: Similar note as in `remote/client/capabilities.rb` -- I suggest removing the `settings.respond_to?(:open_feature) &&`.

lib/datadog/open_feature/component.rb
```diff
@@ -0,0 +1,51 @@
...
+
+        unless settings.respond_to?(:remote) && settings.remote.enabled
+          logger.warn('OpenFeature: Could not be enabled without Remote Configuration Management available')
```
ivoanjo: Since this is a user-visible message, I suggest making it a bit more user-friendly. For instance, explaining how to enable remote configuration (or at least pointing at the docs where that's explained).

IMO we should think of these this way -- "What's the next thing that a customer will want to do after reading this message?" and try to help with that task.

lib/datadog/open_feature/exposures/context.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Datadog
+  module OpenFeature
+    module Exposures
+      module Context
+        module_function
```
ivoanjo: Minor: Why does this need to be a `module_function` again? :eyes:

lib/datadog/open_feature/exposures/deduplicator.rb
```diff
@@ -0,0 +1,36 @@
...
+
+        def duplicate?(event)
+          cache_key = digest(event.flag_key, event.targeting_key)
+          cache_digest = digest(event.allocation_key, event.variation_key)
+
+          stored = @cache[cache_key]
+          return true if stored == cache_digest
+
+          @mutex.synchronize { @cache.store(cache_key, cache_digest) }
+          false
```
ivoanjo: I don't think this is correct, thread synchronization-wise. The `LRUCache` mutates itself even during lookups, so if one thread is looking up the key in line 21 and another is trying to store a new key in line 24, they may interleave and bad things can happen.

Thus, the `@mutex` really needs to go around the whole interaction with `@cache`, not just calls to `store`.

lib/datadog/open_feature/binding/configuration.rb
y9v: Tbh I think if you are adding an abstraction over this data to avoid working with a big Hash directly, it makes no sense to duplicate key names 1:1 to class names. The class names should make sense on their own, since it helps with overall code readability. But I guess this is your decision in the end, especially if this code is temporary as you mentioned.

lib/datadog/open_feature/extensions.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module Datadog
+  module OpenFeature
+    module Extensions
+      def self.activate!
+        Core::Configuration::Settings.extend(Configuration::Settings)
+      end
+    end
+  end
+end
```
ivoanjo, line 16: I know we've used this pattern in other places in the codebase, but I suggest just inlining this where it's used -- we're basically carrying a whole new file with 14 lines of code to add this indirection.

lib/datadog/open_feature/transport/http.rb
ivoanjo, line 1: This is a lot of copy-paste to reuse the current transport code... I know there's already a lot going on in this PR, but I would gently suggest considering just writing a new transport from scratch rather than copying multiple really confusingly-written files.

ivoanjo (commented):
I've been summoned by you asking a review of profiling-rb... although I'm not sure why? Were you looking for an extra review?
I've given it a pass, but I stopped short at some of the more deeper domain classes.

Ok so this is a lot code!. The PR description states

> Eventually, we want to use a C binding with datadog-ffe-ffi for flag evaluations. In this PR, I aimed to implement an evaluator in Ruby that can be easily swapped out to use the binding when it is ready.

and then you reply at some point with

> Much of this is temporary. It will be replaced by a C binding to an evaluator in libdatadog and then the Ruby evaluation code can be deleted.

So this again gets me thinking: What's the plan here again? Are we actually planning on ever shipping the pure-Ruby version to customers? Or is this just for internal testing?

Because, again, this being a lot of code, the answers to those questions are very important in determining how much time we should spend on making sure this PR is solid. Do my doubts here make sense?

lib/datadog/open_feature/exposures/deduplicator.rb
Strech: Sorry that this branch contains a lots of code that already have changed. I think the intent wasn't to open full PR, but rather just binding and to ease the review have 3 PR's with smaller context.

I will look up on the corrections and apply some of them tho. Sorry that it went out of hands 😞

lib/datadog/core/configuration/supported_configurations.rb
Strech: I didn't pick the name, it's in align with other libs 😥

ivoanjo: If we're still on time to change that, may be worth it; if not, well, it's experimental anyway :sob:

lib/datadog/open_feature/exposures/context.rb
Strech: This is outdated code.

lib/datadog/open_feature/extensions.rb
Strech: Thanks for the feedback

lib/datadog/open_feature/transport/http.rb
Strech: 👋🏼 I'm who wrote copy/paste transport. To be honest I wasn't feeling comfortable to go with self-written due to the limitations of my time on this project. At the same time I don't see much a different approach in most components, all of them use `Core` to write the transport, as did I.

I don't mind a better transport if I could, but that's not the case. Do you think it's a blocker?

Strech:
@ivoanjo @y9v I had a plan to split the work to component and then transport and binding as a separate PRs, but by the accident you got pinged with a full blast PR.

I would apply certain suggestions to my dummy binding PR and open it for review, in addition I will extract certain pieces already suggested. 

Sorry that you get though this and thanks for lots of suggestions! ❤️

lib/datadog/open_feature/transport/http.rb
ivoanjo: (We've chatted privately in more detail, but just recording it here that the answer is not at all ;D)

sameerank:
Thanks for the reviews @ivoanjo @y9v ! I didn't know opening a PR onto another stacked branch to consolidate our work would trigger a wider request for reviews. I will convert this PR back to draft so I can rebase this work on the base branch again

lib/datadog/open_feature/binding/internal_evaluator.rb
```diff
@@ -0,0 +1,564 @@
...
+              "flag is missing in configuration, it is either unrecognized or disabled")
+          end
+
+          # Return success result with nil value if flag is disabled - matches libdatadog behavior
+          unless flag.enabled
+            return ResolutionDetails.new(
+              value: nil,
+              variant: nil,
+              error_code: :Ok,  # :Ok indicates success (matches libdatadog ErrorCode::Ok)
+              error_message: '',  # Empty string for Ok cases (matches libdatadog FFI)
```
sameerank: Even though, a disabled flag code includes a message in the errors file (`"flag is disabled"`)

https://github.com/DataDog/libdatadog/blob/3d9e641016f3a6c05bfdb1786c4b1a84342d4bbf/datadog-ffe/src/rules_based/error.rs#L39-L41

It is ignored in this line: `let error_message = value.as_ref().err().map(|err| err.to_string());` The `.err` means we don't use non-error messages and a disabled flag is considered ok. At this point the error message is `None`

But then, this line in the FFI converts it to an empty string: `None => BorrowedStr::empty(),`

https://github.com/DataDog/libdatadog/blob/3d9e641016f3a6c05bfdb1786c4b1a84342d4bbf/datadog-ffe-ffi/src/assignment.rs#L436

lib/datadog/open_feature/binding/assignment_reason.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module Datadog
+  module OpenFeature
+    module Binding
+      # Assignment reasons returned in ResolutionDetails
+      module AssignmentReason
+        TARGETING_MATCH = 'TARGETING_MATCH'
+        SPLIT = 'SPLIT'
+        STATIC = 'STATIC'
+      end
```
sameerank, line 13: All caps due to `#[serde(rename_all = "SCREAMING_SNAKE_CASE")]`

https://github.com/DataDog/libdatadog/blob/b78a66bcf487afbe46c4e409834cecbc882ef01d/datadog-ffe/src/rules_based/ufc/assignment.rs#L13-L21

sameerank: https://github.com/DataDog/libdatadog/blob/3d9e641016f3a6c05bfdb1786c4b1a84342d4bbf/datadog-ffe-ffi/src/assignment.rs#L144

```
pub enum Reason {
    Static,
    Default,
    TargetingMatch,
    Split,
    Disabled,
    Error,
}
```

https://github.com/DataDog/libdatadog/blob/3d9e641016f3a6c05bfdb1786c4b1a84342d4bbf/datadog-ffe-ffi/src/assignment.rs#L373-L391

```
impl From<&ResolutionDetails> for Reason {
    fn from(value: &ResolutionDetails) -> Self {
        match value.as_ref() {
            Ok(assignment) => assignment.reason.into(),
            Err(EvaluationError::FlagDisabled) => Reason::Disabled,
            Err(EvaluationError::DefaultAllocationNull) => Reason::Default,
            Err(_) => Reason::Error,
        }
    }
}
impl From<AssignmentReason> for Reason {
    fn from(value: AssignmentReason) -> Self {
        match value {
            AssignmentReason::TargetingMatch => Reason::TargetingMatch,
            AssignmentReason::Split => Reason::Split,
            AssignmentReason::Static => Reason::Static,
        }
    }
}
```

lib/datadog/open_feature/ext.rb
```diff
@@ -3,11 +3,24 @@
...
     module Ext
+      # OpenFeature error codes  
       ERROR = 'ERROR'
       INITIALIZING = 'INITIALIZING'
       UNKNOWN_TYPE = 'UNKNOWN_TYPE'
       PROVIDER_FATAL = 'PROVIDER_FATAL'
       PROVIDER_NOT_READY = 'PROVIDER_NOT_READY'
+      TYPE_MISMATCH = 'TYPE_MISMATCH'
+      PARSE_ERROR = 'PARSE_ERROR'
+      FLAG_NOT_FOUND = 'FLAG_NOT_FOUND'
```
sameerank: There are also in

https://github.com/DataDog/libdatadog/blob/3d9e641016f3a6c05bfdb1786c4b1a84342d4bbf/datadog-ffe-ffi/src/assignment.rs#L133-L142

lib/datadog/open_feature/ext.rb
```diff
@@ -3,11 +3,24 @@
...
+      
+      # Rust EvaluationError enum values (internal error codes)
```
sameerank: https://github.com/DataDog/libdatadog/blob/3d9e641016f3a6c05bfdb1786c4b1a84342d4bbf/datadog-ffe/src/rules_based/error.rs#L16

sig/datadog/open_feature/binding/resolution_details.rbs
```diff
@@ -8,29 +8,29 @@ module Datadog
 
         attr_accessor variant: ::String?
 
-        attr_accessor error_code: ::String?
+        attr_accessor error_code: ::Symbol?
```
sameerank: `libdatadog` returns an `ErrorCode` enum so a symbol felt a little more appropriate, but I can change it back to a string if that's preferred

lib/datadog/open_feature/evaluation_engine.rb
```diff
@@ -0,0 +1,77 @@
...
+      # PLEASE DELETE when Datadog:OpenFeature:Provider is updated to pass strings
+      # Map symbol types to strings for internal evaluator
+      TYPE_MAP = {
+        boolean: 'boolean',
+        string: 'string',
+        integer: 'integer',
+        number: 'float',
+        float: 'float',
+        object: 'object'
+      }.freeze
```
sameerank: This is here to minimize the footprint of the changes to the `.get_assignment` API. Doing this the right way (in `lib/datadog/open_feature/provider.rb`) would involve touching more test and rbs files which might make a rebase even trickier. This **should be deleted** after a rebase

Or if it's easier to just copy the `InternalEvaluator` files into a fresh branch and hook them up to the `EvaluationEngine` that's fine too

lib/datadog/open_feature/evaluation_engine.rb
```diff
@@ -22,7 +23,7 @@ def initialize(reporter, telemetry:, logger:)
 
       def fetch_value(flag_key:, default_value:, expected_type:, evaluation_context: nil)
         unless ALLOWED_TYPES.include?(expected_type)
-          message = "unknown type #{expected_type.inspect}, allowed types #{ALLOWED_TYPES.join(", ")}"
+          message = "unknown type #{expected_type.inspect}, allowed types #{ALLOWED_TYPES.join(', ')}"
```
sameerank: A rubocop autofix `bundle exec rubocop -a lib/datadog/open_feature/evaluation_engine.rb`

sameerank:
Closing this because we're moving forward with the libdatadog FFI implementation

PR #5023, opened 2025-11-05
Move profiling TagBuilder comment back to profiling

p-datadog:
**What does this PR do?**
Moves the comment about keeping track of profiling-related tags back to profiling TagBuilder.

**Motivation:**
In https://github.com/DataDog/dd-trace-rb/commit/54a7359a2c8504ca9248673a5209c1ac631b7c0e / https://github.com/DataDog/dd-trace-rb/pull/4743 I DRYed the TagBuilder class. As part of that work I moved the comment referencing the tag spreadsheet from profiling to core. This spreadsheet referenced some non-profiling-specific tags, which is I think why I thought it would be appropriate in the core TagBuilder where most of the tags mentioned are, but the spreadsheet is in fact meant for profiling tags.

This PR moves the comment back to profiling TagBuilder.

Note that I don't know if we have a similar spreadsheet/reference for the core (non-product-specific) tags.

**Change log entry**
None

**Additional Notes:**

**How to test the change?**

No tests needed

### DIFF
--- a/lib/datadog/profiling/tag_builder.rb
+++ b/lib/datadog/profiling/tag_builder.rb
@@ -1,24 +1,48 @@
 # frozen_string_literal: true
 
-require_relative "../core/tag_builder"
-require_relative "../core/utils"
+require_relative '../core/tag_builder'
+require_relative '../core/utils'
+
+require 'set'
 
 module Datadog
   module Profiling
     # Builds a hash of default plus user tags to be included in a profile
     #
-    # When changing or adding profiling-related tags, make sure they are
-    # kept in sync with
-    # https://docs.google.com/spreadsheets/d/1LOGMf4c4Avbtn36uZ2SWvhIGKRPLM1BoWkUP4JYj7hA/
-    # (Datadog internal link).
-    #
     # @api private
     module TagBuilder
       include Datadog::Profiling::Ext::Transport::HTTP # Tag name constants
 
+      # When changing or adding profiling-related tags, make sure they are
+      # kept in sync with
+      # https://docs.google.com/spreadsheets/d/1LOGMf4c4Avbtn36uZ2SWvhIGKRPLM1BoWkUP4JYj7hA/
+      # (Datadog internal link).
+      #
+      # For consistency between the different profilers, every tag should be
+      # vetted before it gets reported with a profile, as otherwise it's too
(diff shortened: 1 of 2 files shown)


lib/datadog/profiling/tag_builder.rb
```diff
@@ -6,6 +6,13 @@
 module Datadog
   module Profiling
     # Builds a hash of default plus user tags to be included in a profile
+    #
+    # When changing or adding profiling-related tags, make sure they are
+    # kept in sync with
+    # https://docs.google.com/spreadsheets/d/1LOGMf4c4Avbtn36uZ2SWvhIGKRPLM1BoWkUP4JYj7hA/
+    # (Datadog internal link).
+    #
+    # @api private
```
p-datadog, line 12: Added also this declaration for good measure.

lib/datadog/profiling/tag_builder.rb
```diff
@@ -1,31 +1,57 @@
...
-        # Other metadata
         profile_seq:,
         profiler_version: Core::Environment::Identity.gem_datadog_version
       )
         hash = Core::TagBuilder.tags(settings).merge(
           FORM_FIELD_TAG_PROFILER_VERSION => profiler_version,
           'profile_seq' => profile_seq.to_s,
         )
+        user_tag_keys = settings.tags.keys
+        hash.keep_if { |tag| user_tag_keys.include?(tag) || ALLOWED_TAGS.include?(tag) }
```
p-datadog, line 54: @ivoanjo Should this warn on dropping tags?

ivoanjo: Yeah... I considered it, but I don't think logging is appropriate for this (nobody will ever see it), and I couldn't think of a better way that didn't amount to "stop using the Core::TagBuilder and go back to a fully profiler-specific one". 

Any suggestions?

p-datadog: I suppose the test should detect this situation so it's probably fine.

ivoanjo:
This is good from my side, but I don't like doing a self- :+1: since I contributed to the PR ;) 

Hopefully someone else in the team can give a quick pass and ack that we're both sane.

Strech (approved):

y9v (approved):

PR #5025, opened 2025-11-06
fix(graphql): handle nil value for locations in unified_trace

codella:
**What does this PR do?**

It then adds a pre-condition check in `serialize_error_locations` to handle the case when the provided argument `locations` is `nil`, and makes`serialize_error_locations` a class method for testability.

**Motivation:**

The logic in `serialize_error_locations‎` assumes that `locations` is always present in the Hash returned by `::GraphQL::Error#to_h`, but in reality is only present when if the error can be associated to an AST node ([ref](https://github.com/rmosolgo/graphql-ruby/blob/0afa241775e5a113863766cce126214dee093464/lib/graphql/execution_error.rb#L36)). This results in occasional `NoMethodError`s being raised by applications using this gem.

**Change log entry**

Yes. Tracing: Fix NoMethodError in graphql integration when error has no locations. (Added by @ivoanjo)

**Additional Notes:**

Stack Trace:

```
NoMethodError: undefined method `map' for nil (NoMethodError)

            locations.map do |location|
                     ^^^^
    from datadog/tracing/contrib/graphql/unified_trace.rb:264:in 'serialize_error_locations'
    from datadog/tracing/contrib/graphql/unified_trace.rb:246:in 'block in add_query_error_events'
    from datadog/tracing/contrib/graphql/unified_trace.rb:214:in 'each'
    from datadog/tracing/contrib/graphql/unified_trace.rb:214:in 'add_query_error_events'
    from datadog/tracing/contrib/graphql/unified_trace.rb:70:in 'block in execute_query'
    from datadog/tracing/contrib/graphql/unified_trace.rb:182:in 'block in trace'
    from datadog/tracing/trace_operation.rb:243:in 'block in measure'
    from datadog/tracing/span_operation.rb:167:in 'measure'
    from datadog/tracing/trace_operation.rb:243:in 'measure'
    from datadog/tracing/tracer.rb:425:in 'start_span'
    from datadog/tracing/tracer.rb:166:in 'block in trace'
    from datadog/tracing/context.rb:45:in 'activate!'
    from datadog/tracing/tracer.rb:165:in 'trace'
    from datadog/tracing.rb:29:in 'trace'
    from datadog/tracing/contrib/graphql/unified_trace.rb:161:in 'trace'
    from datadog/tracing/contrib/graphql/unified_trace.rb:50:in 'execute_query'
    from graphql/execution/interpreter.rb:70:in 'block (3 levels) in run_all'
    from graphql/dataloader/null_dataloader.rb:19:in 'append_job'
    from graphql/execution/interpreter.rb:58:in 'block (2 levels) in run_all'
    from graphql/execution/interpreter.rb:54:in 'each'
    from graphql/execution/interpreter.rb:54:in 'each_with_index'
    from graphql/execution/interpreter.rb:54:in 'block in run_all'
    from graphql/tracing/trace.rb:40:in 'execute_multiplex'
    from datadog/tracing/contrib/graphql/unified_trace.rb:44:in 'block in execute_multiplex'
    from datadog/tracing/contrib/graphql/unified_trace.rb:180:in 'block in trace'
    from datadog/tracing/trace_operation.rb:243:in 'block in measure'
    from datadog/tracing/span_operation.rb:167:in 'measure'
    from datadog/tracing/trace_operation.rb:243:in 'measure'
    from datadog/tracing/tracer.rb:425:in 'start_span'
    from datadog/tracing/tracer.rb:166:in 'block in trace'
    from datadog/tracing/context.rb:45:in 'activate!'
    from datadog/tracing/tracer.rb:165:in 'trace'
    from datadog/tracing.rb:29:in 'trace'
    from datadog/tracing/contrib/graphql/unified_trace.rb:161:in 'trace'
    from datadog/tracing/contrib/graphql/unified_trace.rb:44:in 'execute_multiplex'
    from graphql/execution/interpreter.rb:38:in 'run_all'
    from graphql/schema.rb:1492:in 'multiplex'
    from graphql/schema.rb:1468:in 'execute'
    from app/controllers/graphql_controller.rb:19:in 'execute'
```

**How to test the change?**

I was expecting `bundle exec rake test:graphql` to cover the code changes made in `contrib/graphql/unified_trace.rb`, but it looks like that's not the case. Any suggestions on how to proceed or how to add the necessary code coverage are very welcome!

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -77,7 +77,7 @@ jobs:
   test:
     needs:
       - build
-    uses: DataDog/system-tests/.github/workflows/system-tests.yml@7cc82692d3d02b17f20d6d0fcbee09de21f27ad4  # Automated: This reference is automatically updated.
+    uses: DataDog/system-tests/.github/workflows/system-tests.yml@2eb26336ed43c2db1b7f90fff17ce4f5fc36a12b  # Automated: This reference is automatically updated.
     secrets:
       TEST_OPTIMIZATION_API_KEY: ${{ secrets.DD_API_KEY }}  # key used to pushed test results to test optim
       DD_API_KEY: ${{ secrets.DD_API_KEY }}  # key used in tests runs
@@ -91,7 +91,7 @@ jobs:
       desired_execution_time: 300  # 5 minutes
       scenarios_groups: tracer_release
       skip_empty_scenarios: true
-      ref: 7cc82692d3d02b17f20d6d0fcbee09de21f27ad4  # Automated: This reference is automatically updated.
+      ref: 2eb26336ed43c2db1b7f90fff17ce4f5fc36a12b  # Automated: This reference is automatically updated.
       force_execute: ${{ needs.build.outputs.forced_tests }}
       parametric_job_count: 8
       push_to_test_optimization: true
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -146,7 +146,7 @@ requirements_json_test:
 
 configure_system_tests:
   variables:
-    SYSTEM_TESTS_REF: 7cc82692d3d02b17f20d6d0fcbee09de21f27ad4 # Automated: This reference is automatically updated.
+    SYSTEM_TESTS_REF: 2eb26336ed43c2db1b7f90fff17ce4f5fc36a12b # Automated: This reference is automatically updated.
     SYSTEM_TESTS_SCENARIOS_GROUPS: "simple_onboarding,lib-injection"
 
 save_versions:


ivoanjo:
Hey @codella, thanks for spotting this and sending a fix our way!

If you happen to have the stack trace of one of those `NoMethodError`s, it would be cool for us to look into reproducing and see if there's any more situations we might be missing. If not -- that's cool too!

codella:
> Hey @codella, thanks for spotting this and sending a fix our way!
> 
> If you happen to have the stack trace of one of those `NoMethodError`s, it would be cool for us to look into reproducing and see if there's any more situations we might be missing. If not -- that's cool too!

Hey! Thanks for being so snappy! 😲 I updated the description with the stack trace 🚀

codella:
@ivoanjo maybe you can help me. I am running the GraphQL tests with `bundle exec rake test:graphql`, but it doesn't look like they are covering `serialize_error_locations` 🤷🏽 I even temporarily raised an exception inside `serialize_error_locations` to see some test fail, but all I got was 100% green tests 😬

ivoanjo:
Ah, I think you want `bundle exec rake test:graphql_unified_trace_patcher`. I was able to see a failing test suite when I broke `serialize_error_locations` with this.

I believe we have them separate because this functionality depends on modern graphql gem versions, yet we still want to support (and test with) the old ones.

codella:
@ivoanjo I've been spending some time trying to add a test in `unified_trace_patcher_spec.rb` / `test_schema_examples.rb` to replicate the situation where `GraphQL::ExecutionError` doesn't have the `ast_node` attribute set, but I couldn't manage it.

Do you have any pointers, or alternative ways to write a test for it? I'm a bit puzzled that I couldn't get a test to replicate the issue, given that `GraphQL::ExecutionError` allows `ast_node` to be `nil` and my app can raise `GraphQL::ExecutionError` without setting `ast_node` 🤷🏽‍♂️

ivoanjo:
Good question... I spent a few minutes looking into the graphql gem and our tests and how we may trigger the issue to then test the fix, and I also couldn't trigger the error. TBH I think the fact that this is proving hard to test via that route is suggesting to me that it's not the right way to go, given what we want to test here.

Specifically, we want to test that _some specific query in some situation doesn't break_. But if this case is possible (you've definitely seen it in production) but awkward to trigger (we've both not had success at this), I'm thinking even if we find a way to trigger the situation _on the current gem version_, what if in a future version, the error actually is slightly different, and maybe doesn't trigger in the same way, the dd-trace-rb test may start passing for the wrong reasons (doesn't break -- but perhaps not because the code isn't testing for nil).

TL;DR My suggestion here is moving away from integration testing and into a simpler unit test: one that directly tests `serialize_error_locations` or `add_query_error_events` which mock data. I think for a "doesn't break on `nil`" case, that should be both fine and more resistant to future changes in the graphql gem.

codella:
@ivoanjo alright, I extracted both `add_query_error_events` and `serialize_error_locations` in their own module `ErrorSpanEvent`. This way I could easily test the change. I hope it's not too invasive. If you have any suggestions, let me know!

ivoanjo:
> This way I could easily test the change. I hope it's not too invasive. If you have any suggestions, let me know!

Hmmm, to be honest I'm not the biggest fan of how it looks in this version because the `ErrorSpanEvent` module _appears_ separate from `UnifiedTrace` at first glance but depends extremely deeply on it -- e.g. it needs so many things from `UnifiedTrace` that I'd expect many changes to `UnifiedTrace` will probably be reflected here. :thinking: 

What do you think of going back to the previous state, and making `serialize_error_locations` a `self` method on `UnifiedTrace`? That way we could directly test it without having to extract it (I hesitate on extracting only that method, as it seems too small to live alone). :eyes: 

Another useful thing to adjust is `unified_trace.rbs` (the type declarations) -- if we change them to 

```ruby
def serialize_error_locations: (Array[Hash[String, Integer]]? locations) -> Array[String]

# or in my suggested version

def self.serialize_error_locations: (Array[Hash[String, Integer]]? locations) -> Array[String]
```

then the typechecker also starts complaining if the code doesn't deal with `nil`.

WDYT? :smile:

codella:
@ivoanjo super legit suggestion! I implemented it as you said--let me know if I missed anything!

codella:
Thank you @ivoanjo for your help getting this right!

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -285,7 +285,11 @@ def add_query_error_events(span, errors)
           #   ]
           # is serialized as:
           #   ["3:10", "7:8"]
-          def serialize_error_locations(locations)
+          def self.serialize_error_locations(locations)
+            # locations are only provided by the `graphql` library when the error can
+            # be associated to a particular point in the query.
+            return [] if locations.nil?
```
p-datadog: @codella Have you considered returning nil here? Is the value simply being sent to datadog backend in the trace payload? I don't know if it's valuable but keeping nil for lack of locations may be more clearly indicating that there isn't a location?

ivoanjo: Tbh I'm not sure such a distinction is very valuable... :thinking:

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -268,28 +288,13 @@ def add_query_error_events(span, errors)
                   @type_key => parsed_error.type,
                   @stacktrace_key => parsed_error.backtrace,
                   @message_key => graphql_error['message'],
-                  @locations_key => serialize_error_locations(graphql_error['locations']),
+                  @locations_key =>
+                    Datadog::Tracing::Contrib::GraphQL::UnifiedTrace.serialize_error_locations(graphql_error['locations']),
```
Strech, line 292: Would the shorter module resolution work here?

```suggestion
                  @locations_key => UnifiedTrace.serialize_error_locations(graphql_error['locations']),
```

sig/datadog/tracing/contrib/graphql/unified_trace.rbs
```diff
@@ -99,7 +99,7 @@ module Datadog
 
           def add_query_error_events: (SpanOperation span, Array[GraphQL::Error] errors) -> void
 
-          def serialize_error_locations: (Array[Hash[String, Integer]] locations) -> Array[String]
+          def self.serialize_error_locations: (Array[Hash[String, Integer]]? locations) -> Array[String]
```
Strech:
```suggestion
          def self.serialize_error_locations: (::Array[::Hash[::String, ::Integer]]? locations) -> ::Array[::String]
```

lib/datadog/tracing/contrib/graphql/unified_trace.rb
ivoanjo: Since the module gets included elsewhere, I added the full module path to make sure there's no ambiguity.

lib/datadog/tracing/contrib/graphql/unified_trace.rb
```diff
@@ -176,6 +176,26 @@ def platform_resolve_type_key(type, *args, **kwargs)
...
+          #    {"line" => 7, "column" => 8},
+          #   ]
+          # is serialized as:
+          #   ["3:10", "7:8"]
+          def self.serialize_error_locations(locations)
+            # locations are only provided by the `graphql` library when the error can
+            # be associated to a particular point in the query.
+            return [] if locations.nil?
+
+            locations.map do |location|
```
p-datadog, line 194:
```suggestion
            locations&.map do |location|
```
If it's all the same to you, I suggest this

ivoanjo: You're clearly concerned with having a `nil` and not a `[]` here... :thinking: 

Semantically I think the `[]` is clearer -- it's the empty list of locations rather than `nil` which I think is the catch-all more ambiguous for what happened...

Do you mind sharing more of why you think strongly we should go with `nil` here?

p-datadog: I do not have a strong opinion at all actually, I thought nil made more sense.

vpellan (approved):
LGTM!

lib/datadog/tracing/contrib/graphql/unified_trace.rb
ivoanjo: (I'm going to mark this as resolved to merge the PR, let me know if you're still not convinced, happy to open a PR to tweak this after)

ivoanjo:
Thanks everyone for the help!

lib/datadog/tracing/contrib/graphql/unified_trace.rb
codella: @p-datadog hey! I have considered returning `nil` here, too. What made me return an empty array at the end is that we always return an `Array`, this way. That's the only reasoning I had 😊

PR #5030, opened 2025-11-07
[NO-TICKET] Extract AppSec LRUCache implementation into Core::Utils

Strech:
**What does this PR do?**

Extract LRU Cache implementation (non thread-safe) into `Core::Utils`

**Motivation:**

This non thread-safe implementation will be used in another upcoming component.

**Change log entry**

None.

**Additional Notes:**

This code was already reviewed once by the guild-members, but please have another look. I've renamed `store(...)` method into conventional `[]=`, but have to suppress the warning, because we need to return value of the assignment as normal Hash would do.

**How to test the change?**

CI is good enough


lib/datadog/core/utils/lru_cache.rb
```diff
@@ -30,26 +30,13 @@ def [](key)
           end
         end
 
-        def store(key, value)
-          return @store[key] = value if @store.delete(key)
+        def []=(key, value)
+          return @store[key] = value if @store.delete(key) # standard:disable Lint/ReturnInVoidContext
```
ivoanjo: Actually the `Lint/ReturnInVoidContext` is correct here. Specifically, this `return` does not `return` the result from `@store[key] = value`, it **always unconditionally** returns `value`.

Maybe this clarifies what I mean:

```ruby
[10] pry(main)> class Lettuce
[10] pry(main)*   def []=(key, value)
[10] pry(main)*     puts "Got called with #{key} #{value}"
[10] pry(main)*     42
[10] pry(main)*   end  
[10] pry(main)* end  
=> :[]=
[11] pry(main)> l = Lettuce.new
=> #<Lettuce:0x00007a6618b9b718>
[12] pry(main)> l[10] = 1
Got called with 10 1
=> 1
```

So it's not like we have a bug here -- we wanted to return `value` anyway, but the cop is correct that using `return`  makes it look like we could change the code to return something else, and that's not actually true!

I think in practice it may be easier to slightly rewrite the method to avoid the `return`:

```ruby
unless @store.delete(key)
  # NOTE: evict the oldest entry if store reached the maximum allowed size
  @store.shift if @store.size >= @max_size
end

@store[key] = value
```

...or something like it?

ivoanjo (approved):
:+1: I scratched my head for a while on that `ReturnInVoidContext` ;)

lib/datadog/core/utils/lru_cache.rb
Strech: Yes, we can (even tho I think we shouldn't write different code because some linters doesn't understand it). But it doesn't hurt to change it. Will try then. Thanks 🤗

Strech: I've changed it into old good if/else to avoid altering behavior. Should be good now with steep

y9v (approved):

PR #5031, opened 2025-11-07
Add DSM Integration to the Waterdrop instrumentation

ericfirth:
This is one commit on top of the Waterdrop PR (here: https://github.com/DataDog/dd-trace-rb/pull/4874) to add DSM to the Waterdrop instrumentation

**What does this PR do?**
This PR adds DSM instrumentation to Waterdrop ensuring that if DSM is enabled and they use the Karafka ecosystem, a customer can have DSM through their architecture

**Motivation:**
Adding DSM to Ruby and all design partners use Waterdrop

**Change log entry**

Yes. Waterdrop producers are compatible with Data Streams Monitoring (DSM)

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/lib/datadog/tracing/contrib/waterdrop/patcher.rb
+++ b/lib/datadog/tracing/contrib/waterdrop/patcher.rb
@@ -28,6 +28,15 @@ module Datadog
 
               included_middlewares = producer.middleware.instance_variable_get(:@steps)
               producer.middleware.append(Middleware) unless included_middlewares.include?(Middleware)
+
+              if Datadog.configuration.data_streams.enabled
+                producer.monitor.subscribe('message.acknowledged') do |ack_event|
+                  if Datadog::DataStreams.enabled?
+                    payload = ack_event.payload
+                    Datadog::DataStreams.track_kafka_produce(payload[:topic], payload[:partition], payload[:offset])
+                  end
+                end
+              end
             end
           end
         end
--- a/spec/datadog/tracing/contrib/karafka/monitor_spec.rb
+++ b/spec/datadog/tracing/contrib/karafka/monitor_spec.rb
@@ -2,10 +2,7 @@
 
 require 'datadog/tracing/contrib/support/spec_helper'
 
-# FFI::Function background native thread
-ThreadHelpers.with_leaky_thread_creation(:rdkafka) do
-  require 'karafka'
-end
+require 'karafka'
 require 'datadog'
 
 RSpec.describe 'Karafka monitor' do
--- a/lib/datadog/tracing/contrib/waterdrop/middleware.rb
+++ b/lib/datadog/tracing/contrib/waterdrop/middleware.rb
@@ -19,6 +19,17 @@ module Datadog
                 WaterDrop.inject(trace_op.to_digest, message[:headers] ||= {})
               end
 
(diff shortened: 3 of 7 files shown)


spec/datadog/tracing/contrib/karafka/monitor_spec.rb
```diff
@@ -2,7 +2,10 @@
 
 require 'datadog/tracing/contrib/support/spec_helper'
 
-require 'karafka'
+# FFI::Function background native thread
+ThreadHelpers.with_leaky_thread_creation(:rdkafka) do
```
p-datadog: Instead of this I think the exclusion should be added to the spec helper (if needed), as per https://github.com/DataDog/dd-trace-rb/pull/5032.

lib/datadog/tracing/contrib/waterdrop/patcher.rb
```diff
@@ -28,6 +28,13 @@ def patch
 
               included_middlewares = producer.middleware.instance_variable_get(:@steps)
               producer.middleware.append(Middleware) unless included_middlewares.include?(Middleware)
+
+              producer.monitor.subscribe('message.acknowledged') do |ack_event|
+                if Datadog::DataStreams.enabled?
```
Strech: Since there is no one click activation in Ruby, does it make sense to put condition before subscription? That way we will reduce the footprint

Strech (approved):
I left a question about the footprint in case we don't have data streams activated.

_P.S Please fill in the PR template_

lib/datadog/tracing/contrib/waterdrop/patcher.rb
ericfirth: The reason this isn't before the subscription is that `Datadog::DataStreams.enabled?` isn't just checking if the setting for data streams is enabled, but that a processor has been initialized (thus making sure there isn't a huge influx of error logs), which depending on loading order, I was worried might not happen before the subscription is logged and then this just silently doesn't work.

In order to reduce the footprint, I made it so the subscription only happens based on the configuration variable, but the callback still requires the processor

lib/datadog/tracing/contrib/waterdrop/patcher.rb
```diff
@@ -28,6 +28,15 @@ def patch
...
               producer.middleware.append(Middleware) unless included_middlewares.include?(Middleware)
+
+              if Datadog.configuration.data_streams.enabled
+                producer.monitor.subscribe('message.acknowledged') do |ack_event|
+                  if Datadog::DataStreams.enabled?
+                    payload = ack_event.payload
+                    Datadog::DataStreams.track_kafka_produce(payload[:topic], payload[:partition], payload[:offset])
+                  end
+                end
+              end
```
vpellan, line 39: `Datadog::DataStreams.enabled?` can only be true if `Datadog.configuration.data_streams.enabled` is true, so would it be possible to do something like this instead ?

```suggestion
              if Datadog::DataStreams.enabled?
                producer.monitor.subscribe('message.acknowledged') do |ack_event|
                  payload = ack_event.payload
                  Datadog::DataStreams.track_kafka_produce(payload[:topic], payload[:partition], payload[:offset])
                end
              end
```

PR #5032, opened 2025-11-07
Fix thread leak in DSM specs, clean up FFI thread leak checks

p-datadog:
**What does this PR do?**

1) closes DSM Processor after the test to avoid leaking background threads created by the processor
2) adds an exclusion for FFI threads to the thread leak detector
3) removes existing exclusion of detector in a test that triggered FFI threads

**Motivation:**
Fix https://github.com/DataDog/ruby-guild/issues/274

**Change log entry**
None

**Additional Notes:**

**How to test the change?**

Ran changed files locally to verify no output from thread leak detector

### DIFF
--- a/spec/datadog/tracing/contrib/httprb/instrumentation_spec.rb
+++ b/spec/datadog/tracing/contrib/httprb/instrumentation_spec.rb
@@ -40,11 +40,6 @@ RSpec.describe Datadog::Tracing::Contrib::Httprb::Instrumentation do
     Datadog.configure do |c|
       c.tracing.instrument :httprb, configuration_options
     end
-
-    # LibFFI native thread
-    allow(::HTTP::Connection).to receive(:new).and_wrap_original do |method, *args, &block|
-      method.call(*args, &block)
-    end
   end
 
   around do |example|


spec/datadog/tracing/contrib/httprb/instrumentation_spec.rb
```diff
@@ -43,9 +43,7 @@
 
     # LibFFI native thread
     allow(::HTTP::Connection).to receive(:new).and_wrap_original do |method, *args, &block|
-      ThreadHelpers.with_leaky_thread_creation(:ffi) do
-        method.call(*args, &block)
-      end
+      method.call(*args, &block)
     end
```
ivoanjo: This block only existed to apply the `ThreadHelpers`, so it should be removed. 

Leaving it behind I think is quite confusing because at least for me it's one of those situations where I'd go "Wait... Is this doing anything? Why is it here? Am I missing something?" :wink: 

P.s.: I think https://github.com/DataDog/dd-trace-rb/blob/master/spec/datadog/tracing/contrib/ethon/support/thread_helpers.rb can be removed, similarly to this one.

spec/datadog/data_streams/processor_spec.rb
```diff
@@ -22,16 +22,25 @@
...
   end
 
   describe '#initialize' do
-    it 'sets up periodic worker with custom interval' do
-      processor = described_class.new(interval: 5.0, logger: logger, settings: settings, agent_settings: agent_settings)
-      expect(processor.loop_base_interval).to eq(5.0)
+    context 'when custom interval is provided' do
+      let(:processor) do
+        described_class.new(interval: 5.0, logger: logger, settings: settings, agent_settings: agent_settings)
+      end
```
ivoanjo, line 40: Really minor: Consider introducing a `let(:interval)` that would make this even clearer/cleaner imo ;)

spec/datadog/data_streams/processor_spec.rb
```diff
@@ -22,16 +22,25 @@
   let(:agent_settings) { Datadog::Core::Configuration::AgentSettings.new(adapter: :test, hostname: 'localhost', port: 9999) }
   let(:processor) { described_class.new(interval: 10.0, logger: logger, settings: settings, agent_settings: agent_settings) }
 
+  after do
+    processor.stop(true)
+  end
```
ivoanjo, line 28: Minor: If the background thread shutdown takes too long, `stop` returns _without_ waiting for `join` to finish, and thus the thread could potentially still remain around for a little bit more, triggering the leak detection. (I use the same pattern in [the profiler specs](https://github.com/DataDog/dd-trace-rb/blob/6eff05b1f2ddb654189660724867fabe9adacdfe/spec/datadog/profiling/scheduler_spec.rb#L41))

```suggestion
  after do
    processor.stop(true)
    processor.join
  end
```

ericfirth (approved):

spec/datadog/tracing/contrib/httprb/instrumentation_spec.rb
p-datadog: That has a bunch of callers, I'll do it in a separate PR.

spec/datadog/data_streams/processor_spec.rb
p-datadog: I'd rather not because now it's changing the tests for unrelated reasons. I agree that there's a bit of copy pasting going on.

PR #5039, opened 2025-11-11
Update macos images for Github Actions to use macos-15

y9v:
**What does this PR do?**
This PR updates macos runners to version 15.

**Motivation:**
The macOS-13 images are being deprecated and jobs that use them are getting automatically cancelled:
https://github.com/DataDog/dd-trace-rb/actions/runs/19269676399/job/55098464682?pr=5038

**Change log entry**

**Additional Notes:**
None.

**How to test the change?**
CI.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -18,9 +18,6 @@ jobs:
       matrix:
         platform:
           - os: darwin
-            cpu: x86_64
-            base: macos-15-intel # always x86_64-darwin
-          - os: darwin
             cpu: arm64
             base: macos-15 # always arm64-darwin
           - os: linux


.github/workflows/check.yml
```diff
@@ -142,7 +142,7 @@ jobs:
         with:
           persist-credentials: false
       - name: Run actionlint
-        uses: docker://rhysd/actionlint:1.7.7
+        uses: docker://rhysd/actionlint:1.7.8
```
y9v, line 145: this is required in this PR, since version 1.7.7 of actionlint does not know `macos15-intel` runners and fails

.github/workflows/test-macos.yaml
```diff
@@ -17,18 +17,22 @@ jobs:
...
           - '3.4'
           - '3.5'
           # ADD NEW RUBIES HERE
+        include:
+          - os: macos-15-intel
+            ruby: '2.5'
+          - os: macos-15-intel
+            ruby: '2.6'
+          - os: macos-15-intel
+            ruby: '2.7'
```
ivoanjo: TBH I'd go for removing this -- we still list macOS in https://github.com/DataDog/dd-trace-rb/blob/master/docs/Compatibility.md as "Dev environments only" and I don't think it's very common to run macOS 15 on Intel (and we can always re-add this if we find a use-case)

```suggestion
```

.github/workflows/nix.yml
```diff
@@ -19,10 +19,10 @@ jobs:
         platform:
           - os: darwin
             cpu: x86_64
-            base: macos-13 # always x86_64-darwin
+            base: macos-15-intel # always x86_64-darwin
```
ivoanjo: Minor: Same note as below, I'd go for removing macOS on Intel vs keeping it.

ivoanjo (approved):
:+1: Thanks for fixing this one up!

.github/workflows/nix.yml
y9v: yes, let'go for it

PR #5033, opened 2025-11-07
feat: add process tags to traces

wantsui:
**What does this PR do?**

The goal of AIDM-253 is to add process tags to the trace payloads.

After this gets merged, the next step is to add it for the other products.

To run the tests in docker
```
docker compose run --rm tracer-3.3 /bin/bash
```

Main tests:
```
bundle install --gemfile /app/gemfiles/ruby_3.3_rails8.gemfile
bundle exec rake compile
bundle exec rake test:core_with_rails
TEST_DATADOG_INTEGRATION=1 BUNDLE_GEMFILE=/app/gemfiles/ruby_3.3_rails8.gemfile bundle exec rspec spec/datadog/core/environment/process_spec.rb


bundle exec rspec spec/datadog/tracing/transport/trace_formatter_spec.rb
bundle exec rspec spec/datadog/core/tag_normalizer_spec.rb
bundle exec rspec spec/datadog/core/configuration/settings_spec.rb
```

**Motivation:**

We're trying to add process tags to various payloads so they can be used in different use cases.

Note I still want to try adding server type but I'll have to tackle that in a separate PR.

**Change log entry**

Yes. Adds process tags to trace payloads with the new environment variable: `DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED`. 

**Additional Notes:**

**How to test the change?**

PR #5041, opened 2025-11-11
swap out the existing headers normalization logic with the tag normalizer

wantsui:
**What does this PR do?**

Uses the new tag normalization logic that was copied from the Trace Agent for headers.

**Motivation:**

While working on https://github.com/DataDog/dd-trace-rb/pull/5033, it was noticed that the current normalization logic in dd-trace-rb is missing some of the logic from https://github.com/DataDog/datadog-agent/blob/45799c842bbd216bcda208737f9f11cade6fdd95/pkg/trace/traceutil/normalize.go#L131 , specifically around:
- character limit truncation after 200 characters
- consecutive `_` are merged
- if the first letter is not a letter, look for the next first letter since tags have to start with a letter

**Change log entry**

feat: swap out the existing headers normalization logic with the tag normalizer

**Additional Notes:**

I am not sure if we consider this to be a bug or a missing feature, but note that with this new logic, some of the headers results are going to be empty strings ie: `""`, which can be seen in the tests below.

**How to test the change?**

I've been using:
```
bundle exec rspec spec/datadog/tracing/metadata/ext_spec.rb
```


spec/datadog/tracing/metadata/ext_spec.rb
```diff
@@ -26,15 +26,16 @@
...
 
     context 'with unsupported characters' do
       let(:tag) { "\\ \t!@?" }
 
       it 'replaces each with an underscore' do
-        is_expected.to eq('_' * tag.size)
+        # turns into an empty string because in a tag, the first character must be a letter
+        is_expected.to eq('')
       end
     end
```
ivoanjo: When changing the logic -- please update the test description as well! It'd be quite confusing to have a test saying "it replaces each with an underscore" that actually is not true :P

wantsui: oh! good catch, I'll use a better description for this!

wantsui: I updated the description but open to feedback on the wording !

ivoanjo: Seems reasonable! :)

wantsui: Thanks! Just a heads up I'm still changing the normalization logic in https://github.com/DataDog/dd-trace-rb/pull/5033/, so I'll circle back to this PR after the process tags PR to see if I need to update these test assertions.

spec/datadog/tracing/metadata/ext_spec.rb
```diff
@@ -25,16 +25,24 @@
...
+      it 'normalizes them to an empty string' do
+        is_expected.to eq("")
       end
     end
 
-    context 'with unsupported characters' do
+    context 'with a mix of unsupported and supported characters' do
+      let(:tag) { "\\ \t!@?a" }
+
+      it 'the first letter to show up is the supported character' do
```
Strech, line 36: We try to make out of this line a real statement, readable by us, now that reads

```
it the first letter to show up is the supported character
```

to be honest I don't get what it said. Check out some other tests, you will find something like

```
it 'returns only supported characters'
```

or something similar

wantsui:
Closing this until I'm ready to revisit.

Issue #5040, opened 2025-11-11
[BUG]: Rails upgrade to 8.1 causes `undefined method 'path_parameters'`


renchap:
### Tracer Version(s)

2.22.0

### Ruby Version(s)

ruby 3.4.7 (2025-10-08 revision 7a5688e2a2) +PRISM [arm64-darwin25]

### Relevent Library and Version(s)

Rails 8.1.1

### Bug Report

When updating our app from Rails 8.0.4 to Rails 8.1.1, all web requests started to fail with the following error:

```
NoMethodError: undefined method 'path_parameters' for an instance of Rack::Request (NoMethodError)
```

When disabling the AppSec Rails integration (commenting `c.appsec.instrument :rails`), the error does not occur.

I am not sure if this comes from an interaction with another library, but the commit introducing this is simply bumping the Rails gems.

### Configuration Block

```
Datadog.configure do |c|
  enabled = !Rails.env.local?

  c.env = Rails.env

  c.tracing.enabled = enabled
  c.remote.enabled = enabled

  c.service = 'notos'
  c.tags = { 'app' => 'notos' }
  c.version = Notos::VERSION

  c.tracing.log_injection = !Rails.env.local?
  c.diagnostics.startup_logs.enabled = !Rails.env.local?

  c.tracing.instrument :rack, { request_queuing: true }
  c.tracing.instrument :rails
  c.tracing.instrument :action_cable
  c.tracing.instrument :active_job
  c.tracing.instrument :redis
  c.tracing.instrument :http, { split_by_domain: true }
  c.tracing.instrument :faraday, { split_by_domain: true }
  c.tracing.instrument :httprb, { split_by_domain: true }

  c.appsec.enabled = enabled
  c.appsec.instrument :rails

  # c.diagnostics.debug = true
end
```

### Error Logs

There are no Datadog errors, but a Ruby exception

```
NoMethodError: undefined method 'path_parameters' for an instance of Rack::Request (NoMethodError)
        req_params  = req.path_parameters
                         ^^^^^^^^^^^^^^^^
    from action_dispatch/journey/router.rb:44:in 'recognize'
    from datadog/appsec/api_security/route_extractor.rb:60:in 'Datadog::AppSec::APISecurity::RouteExtractor.route_pattern'
    from datadog/appsec/api_security/sampler.rb:47:in 'sample?'
    from datadog/appsec/api_security.rb:14:in 'Datadog::AppSec::APISecurity.sample?'
[...]

PR #5042, opened 2025-11-12
Fix APISecurity::RouteExtractor for Rails 8.1.1+

y9v:
**What does this PR do?**

This PR:
- fixes route extraction for Rails 8.1.1+
- adds a `rescue` to `APISecurity::RouteExtractor.route_pattern` with telemetry error reporting
- changes the slow path for Rails to always wrap the rack request to router request class
- adds explicit string conversion for route pattern value to `APISecurity::Sampler`

**Motivation:**
https://github.com/DataDog/dd-trace-rb/issues/5040

Rails 8.1.1 removes the `action_dispatch.route_uri_pattern` from `request.env` and replaces it with `action_dispatch.route`:
https://github.com/rails/rails/commit/6fd05c92b4820eeb537d5213812ad2f76e92b533

**Change log entry**
Yes. AppSec: Fix route extraction error for Rails 8.1.1+.

**Additional Notes:**
None.

**How to test the change?**
CI and manual testing.

### DIFF
--- a/lib/datadog/appsec/api_security/route_extractor.rb
+++ b/lib/datadog/appsec/api_security/route_extractor.rb
@@ -64,6 +64,9 @@ module Datadog
             #       it does not have all the methods that ActionDispatch::Request has.
             #       Before trying to use the router to recognize the route, we need to
             #       create a new ActionDispatch::Request from the request env
+            #
+            # NOTE: Rails mutates HEAD request by changing the method to GET
+            #       and uses it for route recognition to check if the route is defined
             request = request.env[RAILS_ROUTES_KEY].request_class.new(request.env)
 
             pattern = request.env[RAILS_ROUTES_KEY].router
@@ -78,8 +81,8 @@ module Datadog
           else
             Tracing::Contrib::Rack::RouteInference.read_or_infer(request.env)
           end
-        rescue
-          AppSec.telemetry&.error('AppSec: Could not extract route pattern for APISecurity sampler')
+        rescue => e
+          AppSec.telemetry&.report(e, description: 'AppSec: Could not extract route pattern')
 
           nil
         end
--- a/spec/datadog/appsec/api_security/route_extractor_spec.rb
+++ b/spec/datadog/appsec/api_security/route_extractor_spec.rb
@@ -219,7 +219,8 @@ RSpec.describe Datadog::AppSec::APISecurity::RouteExtractor do
         it { expect(described_class.route_pattern(request)).to be_nil }
 
         it 'reports the error via telemetry' do
-          expect(telemetry).to receive(:error).with('AppSec: Could not extract route pattern for APISecurity sampler')
+          expect(telemetry).to receive(:report)
+            .with(an_instance_of(StandardError), description: 'AppSec: Could not extract route pattern')
 
           described_class.route_pattern(request)
         end


lib/datadog/appsec/api_security/route_extractor.rb
```diff
@@ -52,11 +56,15 @@ def self.route_pattern(request)
             pattern = request.env[SINATRA_ROUTE_KEY].split(SINATRA_ROUTE_SEPARATOR, 2)[1]
             "#{request.script_name}#{pattern}"
           elsif request.env.key?(RAILS_ROUTE_KEY)
-            request.env[RAILS_ROUTE_KEY].delete_suffix(RAILS_FORMAT_SUFFIX)
+            request.env[RAILS_ROUTE_KEY].path.spec.to_s.delete_suffix(RAILS_FORMAT_SUFFIX)
+          elsif request.env.key?(RAILS_ROUTE_URI_PATTERN_KEY)
+            request.env[RAILS_ROUTE_URI_PATTERN_KEY].delete_suffix(RAILS_FORMAT_SUFFIX)
           elsif request.env.key?(RAILS_ROUTES_KEY) && !request.env.fetch(RAILS_PATH_PARAMS_KEY, {}).empty?
-            # NOTE: Rails mutate HEAD request in order to understand that route is supported.
-            #       It will assing GET request method and run the route recognition.
```
Strech, line 58: I think it's still valid note about mutation, maybe we can rephrase it to keep the concern of mutation

lib/datadog/appsec/api_security/route_extractor.rb
```diff
@@ -70,6 +78,10 @@ def self.route_pattern(request)
           else
             Tracing::Contrib::Rack::RouteInference.read_or_infer(request.env)
           end
+        rescue
+          AppSec.telemetry&.error('AppSec: Could not extract route pattern for APISecurity sampler')
```
Strech: I think it makes sense to use `telemetry.report` instead and don't mention the place of the usage like sampling (caller might by any)

Question: It's not a good idea to have potential exception in exception handling, but maybe we can fallback to infering the route?

lib/datadog/appsec/api_security/sampler.rb
```diff
@@ -45,7 +45,9 @@ def sample?(request, response)
           return true if @sample_delay_seconds.zero?
           return false if response.status == 404
 
-          key = Zlib.crc32("#{request.request_method}#{RouteExtractor.route_pattern(request)}#{response.status}")
+          route_pattern = RouteExtractor.route_pattern(request).to_s
```
Strech, line 48: I think it's better to always return something instead, otherwise we might have a bug and just not sampling at all.

lib/datadog/appsec/api_security/route_extractor.rb
y9v: yes, you are right, I will try to add this info back

lib/datadog/appsec/api_security/route_extractor.rb
y9v: I guess we could do that, considering that we are also rescuing exceptions in `RouteInference.infer` (what bothers me is that this could change)

p-datadog: I looked at the codebase and it seems appsec uses `telemetry.error` while other components use `telemetry.report`, with `error` not taking the exception as argument, I'm curious what the reasoning is for omitting the exception?

lib/datadog/appsec/api_security/route_extractor.rb
```diff
@@ -70,6 +78,10 @@ def self.route_pattern(request)
           else
             Tracing::Contrib::Rack::RouteInference.read_or_infer(request.env)
           end
+        rescue
```
p-datadog: Should this rescue be somewhere higher up the call stack to catch other possible issues that would affect customer applications?

lib/datadog/appsec/api_security/route_extractor.rb
y9v: As far as I was aware, we should not report exceptions via telemetry, since these might be exceptions from customers code base, containing PII (far fetched, but this is what I heard).

@TonyCTHsu could you please confirm this?

TonyCTHsu: I would redirect to the follow the [guide](https://github.com/DataDog/dd-trace-rb/blob/e0ab5b7e332e7f4faa80e312e75a9af7e743d491/lib/datadog/core/telemetry/logging.rb#L10-L22) for discussion

lib/datadog/appsec/api_security/route_extractor.rb
y9v: added back in [a693370](https://github.com/DataDog/dd-trace-rb/pull/5042/commits/a6933708ed1b7a5342a1dc227c597e6c3e6f3e20)

lib/datadog/appsec/api_security/route_extractor.rb
y9v: changed to `telemetry.report` [a693370](https://github.com/DataDog/dd-trace-rb/pull/5042/commits/a6933708ed1b7a5342a1dc227c597e6c3e6f3e20)

lib/datadog/appsec/api_security/sampler.rb
y9v: if the `route_pattern` is `nil`, we will use "#{request.request_method}#{nil}#{response.status}" as the key - so it should be fine?

Strech: Yes, as it will be interpolated next, you can just skip this change

Strech (approved):
Thanks for taking care 🤗

lib/datadog/appsec/api_security/route_extractor.rb
y9v: Higher up the call stack we have `APISecurity::Sampler`, which samples based on several things, one of which is the `#{request_method}#{request_path}#{route_pattern}" string. So it still works ok if we rescue in `RouteExtractor` and have empty route pattern - sampling will be less endpoint specific, but will still work.

If I move this up to `APISecurity::Sampler`, we will have to add some sampling logic for such error case.

So I would prefer to keep it here for now.

PR #5048, opened 2025-11-14
[APMAPI-1643] Change supported-configurations.json to v2 format

vpellan:
**What does this PR do?**

This PR change the supported-configuration.json file to use the v2 format, and change the related rake task. It also updates the CI checks.

**Motivation:**

v2 format will enable us to verify the default value, the type and the programmatic path against what is registered on the FPD.

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**


supported-configurations.json
```diff
@@ -1,984 +1,2502 @@
...
+      }
+    ],
+    "DD_PROFILING_EXPERIMENTAL_HEAP_ENABLED": [
+      {
+        "version": "A",
+        "type": "boolean",
+        "propertyKeys": ["advanced.experimental_heap_enabled"],
+        "default": "false"
+      }
+    ],
```
ivoanjo, line 393: Gave an ultra-fast pass and noticed that `propertyKeys` is not the full path -- e.g. `profiling.collection.enabled` vs `collection.enabled` or `profiling.advanced....` vs `advanced...`. Is that intended?

vpellan: I generated that with the config DSL output, and just extracted the name from OptionDefinition, as it was a very long task otherwise. For now we don't do any check on it, this is just an indicative field, I can remove it if it creates confusion as it does not add any value for now.

vpellan: This is why the path is incomplete: https://github.com/DataDog/dd-trace-rb/blob/vpellan/config-inversion-v2-backfill-json/lib/datadog/core/configuration/options.rb#L28

lib/datadog/core/configuration/config_helper.rb
```diff
@@ -57,7 +57,7 @@ def get_environment_variable(name, default_value = nil, source_env: @source_env)
           # Until we've correctly implemented support for datadog-ci-rb, we disable config inversion if ci is enabled.
           if !defined?(::Datadog::CI) &&
               (name.start_with?('DD_', 'OTEL_') || @alias_to_canonical[name]) &&
-              !@supported_configurations[name]
+              !@supported_configurations.include?(name)
```
p-datadog, line 60: Is `@supported_configurations` still a hash? If yes I would find `key?` to be much easier to read than `include?` even if Hash defines `include?` (since the usage suggests to the reader that the container is a hash not an array).

rubocop/custom_cops/env_string_validation_cop.rb
```diff
@@ -28,7 +28,7 @@ class EnvStringValidationCop < RuboCop::Cop::Base
 
     # Configuration for allowed environment variable names
     # This list should be populated with the allowed environment variable names
-    ALLOWED_ENV_VARS = Datadog::Core::Configuration::SUPPORTED_CONFIGURATIONS.keys
+    ALLOWED_ENV_VARS = Datadog::Core::Configuration::SUPPORTED_CONFIGURATIONS
```
p-datadog, line 31: The doc still says that supported configurations is a map and line 30 says allowed env vars is a list, I am confused how a map is assigned to a list here.

supported-configurations.json
ivoanjo: I'd go with removing anything we're not using right now and don't yet have plans for ;)

lib/datadog/core/configuration/config_helper.rb
vpellan: It is an Set now thanks to your feedback on Slack!

rubocop/custom_cops/env_string_validation_cop.rb
vpellan: Could you point out to the doc that still says `SUPPORTED_CONFIGURATIONS` is a map ? in the json file, supportedConfiguration value is still a json map. It is in the generated .rb file that it isn't one anymore but you're not really supposed to interact with this file.

supported-configurations.json
vpellan: Removed in c61058a4ab5ca1205e782027171dcce096742ec8

rubocop/custom_cops/env_string_validation_cop.rb
vpellan: I'll close this comment, if there's a concern with the documentation I'll open a new PR for this.

lib/datadog/core/configuration/supported_configurations.rb
```diff
@@ -1,333 +1,341 @@
...
 
+require 'set'
+
 # This file is auto-generated from `supported-configurations.json` by `rake local_config_map:generate`.
 # Do not change manually! Please refer to `docs/AccessEnvironmentVariables.md` for more information.
 
 module Datadog
   module Core
     module Configuration
       SUPPORTED_CONFIGURATIONS =
```
p-datadog, line 11: My proposed  change was to rename this to `SUPPORTED_CONFIGURATION_NAMES`.

PR #5053, opened 2025-11-17
Update injector to v1.1

lloeki:
**What does this PR do?**

- Update Ruby injector to latest version
- Adjust OCI package contents to match

**Motivation:**

Add various fixes + deployment mode

**Change log entry**

- SSI: add support for Bundler development mode
- SSI: report UI-oriented injection results
- SSI: guard against Bundler global `force_ruby_platform`
- SSI: guard against Bundler 4.0 and Bundler 2.7 in 4.0 mode
- SSI: guard against Ruby 3.5+

**Additional Notes:**

System tests are being updated to cover more of this.

**How to test the change?**

CI should pass

### DIFF
--- a/.gitlab/install_datadog_deps.rb
+++ b/.gitlab/install_datadog_deps.rb
@@ -96,6 +96,9 @@ env = {
 ].each do |gem|
   version = gem_version_mapping.delete(gem)
 
+  # This hardcoded dependency list actually depends on the lockfile contents,
+  # which is per-version. The better solution is of course to refactor this to
+  # not hardcode the list and use solely what was locked.
   next if version.nil?
 
   gem_install_cmd = "gem install #{gem} "\


lloeki:
Odd behaviour at the [`create-multi-arch-lib-injection-image` stage](https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-rb/-/jobs/1244730046#L174):

```
$ echo "TAGS are $TAGS_AS_ARGUMENT"
TAGS are --tag registry.ddbuild.io/ssi/dd-lib-ruby-init:8119bb9fc46d2b7c8fa7440144ef0ab4c33a570d --tag registry.ddbuild.io/ssi/dd-lib-ruby-init:glci83149026 --tag registry.ddbuild.io/ssi/dd-lib-ruby-init:glci83149026-g8119bb9fc46d2b7c8fa7440144ef0ab4c33a570d 
$ datadog-package decompose --single-manifest-image packaging/*linux*amd64*.tar scripts/lib-injection/decomposed-amd64
Using local package package at packaging/datadog-apm-library-ruby-2.23.0.dev.bedb6bb.glci1244730020.g8119bb9f-linux-amd64.tar
2025/11/19 14:52:02 could not extract layer: could not create symlink: symlink injector.rb scripts/lib-injection/decomposed-amd64/contents/auto_inject.rb: no such file or directory
```

The odd part is that if I examine the artifacts from the [previous stage](https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-rb/-/jobs/1244730036) they contain perfectly valid symlinks as expected, are the same as before, and have been working just fine til now:

```
$ tar xf datadog-apm-library-ruby-2.23.0.dev.bedb6bb.glci1244730020.g8119bb9f-linux-amd64.tar
$ tar tvzf datadog-apm-library-ruby-2.23.0.dev.bedb6bb.glci1244730020.g8119bb9f-linux-amd64/blobs/sha256/80abb142a762d320db497f3ee946859166151287e14bc8bfe3658ee2bf7e1d3c | head -3
lrwxrwxrwx  0 root   1000        0 Nov 19 15:48 auto_inject.rb -> injector.rb
lrwxrwxrwx  0 root   1000        0 Nov 19 15:48 host_inject.rb -> injector.rb
-rw-r--r--  0 root   1000     2628 Nov 19 15:48 injector.rb
```

So far no idea what's going on with that `datadog-package decompose` command failing, except `no such file or directory` seems to suggest a path component of `scripts/lib-injection/decomposed-amd64/contents` isn't there.

.gitlab/install_datadog_deps.rb
```diff
@@ -91,6 +96,8 @@
 ].each do |gem|
   version = gem_version_mapping.delete(gem)
 
+  next if version.nil?
```
p-datadog, line 102: A comment here would be helpful to understand why the version wouldn't be returned/found.

p-datadog (approved):

.gitlab/install_datadog_deps.rb
lloeki: Good point. I'll add the comment.

Immediate answer though: this hardcoded dependency list actually depends on the lockfile contents, which is per-version.

The better solution is of course to refactor this to _not_ hardcode the list.

PR #5054, opened 2025-11-17
FFL-1273 Native feature flags evaluation

dd-oleksii:
**What does this PR do?**

Implement OpenFeature evaluator using bindings to native libdatadog evaluator.

**Change log entry**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/ext/libdatadog_api/feature_flags.c
+++ b/ext/libdatadog_api/feature_flags.c
@@ -195,9 +195,11 @@ static int evaluation_context_foreach_callback(VALUE key, VALUE value, VALUE arg
     return ST_CONTINUE;
   }
 
-  // Ensure we don't exceed capacity (should not happen if capacity = hash size)
+  // Ensure we don't exceed capacity.
   if (builder->attr_count >= builder->attr_capacity) {
-    return ST_STOP;
+    // This should never happen because evaluation_context_from_hash()
+    // pre-allocates attr_capacity equal to iterated Hash size.
+    rb_raise(feature_flags_error_class, "Internal: Attribute count exceeded capacity");
   }
 
   ddog_ffe_AttributePair *attr = &builder->attrs[builder->attr_count];
@@ -349,8 +351,10 @@ static VALUE resolution_details_get_raw_value(VALUE self) {
     case DDOG_FFE_VARIANT_VALUE_OBJECT:
       return str_from_borrow(value.string);
     case DDOG_FFE_VARIANT_VALUE_NONE:
-    default:
       return Qnil;
+    default:
+      // This should never happen as we checked for all possible tag values.
+      rb_raise(feature_flags_error_class, "Internal: Unexpected ResolutionDetails value tag");
   }
 }
 
@@ -380,8 +384,10 @@ static VALUE resolution_details_get_flag_type(VALUE self) {
     case DDOG_FFE_VARIANT_VALUE_OBJECT:
       return ID2SYM(id_object);
     case DDOG_FFE_VARIANT_VALUE_NONE:
-    default:
       return Qnil;
+    default:
+      // This should never happen as we checked for all possible tag values.
+      rb_raise(feature_flags_error_class, "Internal: Unexpected ResolutionDetails value tag");
   }
 }
 
@@ -411,7 +417,7 @@ static VALUE resolution_details_get_variant(VALUE self) {
 static VALUE resolution_details_get_allocation_key(VALUE self) {
   ddog_ffe_Handle_ResolutionDetails resolution_details =
     (ddog_ffe_Handle_ResolutionDetails)rb_check_typeddata(self, &resolution_details_typed_data);
-  struct ddog_ffe_BorrowedStr allocation_key =ddog_ffe_assignment_get_allocation_key(resolution_details);
+  struct ddog_ffe_BorrowedStr allocation_key = ddog_ffe_assignment_get_allocation_key(resolution_details);
   return str_from_borrow(allocation_key);
 }
 


ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,366 @@
...
+  // ResolutionDetails class
+  VALUE resolution_details_class = rb_define_class_under(binding_module, "ResolutionDetails", rb_cObject);
+  rb_undef_alloc_func(resolution_details_class);
+  rb_define_method(resolution_details_class, "value", resolution_details_get_value, 0);
+  rb_define_method(resolution_details_class, "reason", resolution_details_get_reason, 0);
+  rb_define_method(resolution_details_class, "error_code", resolution_details_get_error_code, 0);
+  rb_define_method(resolution_details_class, "error_message", resolution_details_get_error_message, 0);
+  rb_define_method(resolution_details_class, "variant", resolution_details_get_variant, 0);
+  rb_define_method(resolution_details_class, "allocation_key", resolution_details_get_allocation_key, 0);
+  rb_define_method(resolution_details_class, "log?", resolution_details_get_do_log, 0);
```
Strech, line 107: I think there should be `error?` method too

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,366 @@
...
+
+  if (!resolution_details) {
+    return Qnil;
+  }
+
+  struct ddog_ffe_VariantValue value = ddog_ffe_assignment_get_value(resolution_details);
+
+  switch (value.tag) {
+    case DDOG_FFE_VARIANT_VALUE_NONE:
+      return Qnil;
```
Strech: I think it makes sense to leave a comment on why do we return `Qnil` here and what this special case is

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,366 @@
...
+};
+
+static void evaluation_context_free(void *ptr) {
+  ddog_ffe_Handle_EvaluationContext context = (ddog_ffe_Handle_EvaluationContext) ptr;
+  if (context) {
+    ddog_ffe_evaluation_context_drop(&context);
+  }
+}
+
+static VALUE evaluation_context_initialize_with_attributes(VALUE klass, VALUE targeting_key, VALUE attributes_hash) {
```
Strech: In OpenFeature SDK evaluation context is a single-hash containing all keys. Will it be possible to avoid allocations and not extract the targeting key?

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,366 @@
...
+
+void feature_flags_init(VALUE core_module) {
+  // Always define the Binding module - it will reuse existing if it exists
+  VALUE binding_module = rb_define_module_under(core_module, "FeatureFlags");
+
+  // Configuration class
+  VALUE configuration_class = rb_define_class_under(binding_module, "Configuration", rb_cObject);
+  rb_undef_alloc_func(configuration_class);
+  rb_define_singleton_method(configuration_class, "new", configuration_new, 1);
+  rb_define_method(configuration_class, "get_assignment", configuration_get_assignment, 2);
```
Strech: I find it weird to have code like this

```ruby
def reconfigure!(configuration)
  @evaluator = Datadog::Core::FeatureFlags::Configuration.new(configuration)
end
```

because configuration implies a static thing unable to provide me anything except it's values

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,366 @@
...
+  ddog_ffe_Handle_ResolutionDetails resolution_details = (ddog_ffe_Handle_ResolutionDetails)rb_check_typeddata(self, &resolution_details_typed_data);
+
+  if (!resolution_details) {
+    return Qnil;
+  }
+
+  // Use the new FFI function to get the error code
+  enum ddog_ffe_ErrorCode error_code = ddog_ffe_assignment_get_error_code(resolution_details);
+
+  switch (error_code) {
```
Strech, line 470: To avoid double mappings, I would suggest to stick with [OpenFeature SDK](https://github.com/open-feature/ruby-sdk/blob/main/lib/open_feature/sdk/provider/error_code.rb) and return string in capital case. (it's [standard across all libs](https://pkg.go.dev/github.com/open-feature/go-sdk/openfeature#ErrorCode))

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,366 @@
...
+    case DDOG_FFE_REASON_DEFAULT:
+      return ID2SYM(rb_intern("default"));
+    case DDOG_FFE_REASON_TARGETING_MATCH:
+      return ID2SYM(rb_intern("targeting_match"));
+    case DDOG_FFE_REASON_SPLIT:
+      return ID2SYM(rb_intern("split"));
+    case DDOG_FFE_REASON_DISABLED:
+      return ID2SYM(rb_intern("disabled"));
+    case DDOG_FFE_REASON_ERROR:
+      return ID2SYM(rb_intern("error"));
```
Strech: Same issue as with error code, let's stick to OpenFeature SDK (see [java-sdk](https://github.com/open-feature/java-sdk/blob/main/src/main/java/dev/openfeature/sdk/Reason.java#L6-L14))

ext/libdatadog_api/feature_flags.c
dd-oleksii: This code is weird because `Configuration` is not supposed to be called an "Evaluator". This one is not weird:
```ruby
def reconfigure!(configuration)
  @configuration = Datadog::Core::FeatureFlags::Configuration.new(configuration)
end
```

ext/libdatadog_api/feature_flags.c
dd-oleksii: Yep, will return strings in the expected format

ext/libdatadog_api/feature_flags.c
Strech: My comment was about `get_assignment` call from configuration, it's not semantic at all and it will be questioned on review, configuration is not suppose to find a value based on some computations, because it's called a configuration. It's settings, values, nothing more.

There is no good reason to copy Rust names in a C-ext. **It's not a blocker**, but C-ext should hide some unnecessary details of Rust implementation, that doesn't make much sense for us in Ruby, not it's just 1:1 copy.

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,359 @@
...
+
+  const long max_attr_count = RHASH_SIZE(hash);
+  ddog_ffe_AttributePair *const attrs =
+      ruby_xcalloc(max_attr_count, sizeof(struct ddog_ffe_AttributePair));
+
+  // Convert hash to attribute pairs
+  const VALUE keys = rb_funcall(hash, rb_intern("keys"), 0);
+  // Number of currently filled attributes.
+  long attr_count = 0;
+  for (long i = 0; i < max_attr_count; i++) {
```
Strech: Could `rb_hash_foreach()` used instead for direct iteration? That's an unnecessary allocation

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,359 @@
...
+      continue;
+    }
+
+    attrs[attr_count].name = RSTRING_PTR(key);
+
+    // Set the value based on its Ruby type
+    switch (TYPE(value)) {
+      case T_STRING:
+        attrs[attr_count].value.tag = DDOG_FFE_ATTRIBUTE_VALUE_STRING;
+        attrs[attr_count].value.string = RSTRING_PTR(value);
```
Strech: Could it be the case that pointer doesn't outlive the in between calls (time between data preparation and context creation/evaluation) and will be GC-ed?

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,359 @@
...
+        break;
+    }
+
+    attr_count += 1;
+  }
+
+  const ddog_ffe_Handle_EvaluationContext context = ddog_ffe_evaluation_context_new(
+    targeting_key,
+    attrs,
+    max_attr_count
```
Strech: This must be `attr_count` because we skip the targeting key

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,359 @@
...
+  const long max_attr_count = RHASH_SIZE(hash);
+  ddog_ffe_AttributePair *const attrs =
+      ruby_xcalloc(max_attr_count, sizeof(struct ddog_ffe_AttributePair));
+
+  // Convert hash to attribute pairs
+  const VALUE keys = rb_funcall(hash, rb_intern("keys"), 0);
+  // Number of currently filled attributes.
+  long attr_count = 0;
+  for (long i = 0; i < max_attr_count; i++) {
+    const VALUE key = rb_ary_entry(keys, i);
```
Strech: This could return `Qnil` for out of bound index, does it make sense to put a guard clause here?

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,359 @@
...
+        attrs[attr_count].value.boolean = true;
+        break;
+      case T_FALSE:
+        attrs[attr_count].value.tag = DDOG_FFE_ATTRIBUTE_VALUE_BOOLEAN;
+        attrs[attr_count].value.boolean = false;
+        break;
+      default:
+        // Default to string representation
+        attrs[attr_count].value.tag = DDOG_FFE_ATTRIBUTE_VALUE_STRING;
+        attrs[attr_count].value.string = RSTRING_PTR(rb_funcall(value, rb_intern("to_s"), 0));
```
Strech: Will this reference live long enough? It's an immediate raw pointer without keeping the reference to a created string

ext/libdatadog_api/feature_flags.c
dd-oleksii: What do you mean by unnecessary allocation? If you mean allocation of `attrs`, `rb_hash_foreach()` won't remove it

ext/libdatadog_api/feature_flags.c
dd-oleksii: `i` is bound by `RHASH_SIZE(hash)` so it shouldn't be out of bound. We also have `Check_Type(key, T_STRING)` a couple of lines below.

ext/libdatadog_api/feature_flags.c
dd-oleksii: No, GC doesn't run while we're in C and haven't released GVL. `ddog_ffe_evaluation_context_new` also copies all the necessary data, so it does not refer to Ruby-owned strings

ext/libdatadog_api/feature_flags.c
dd-oleksii: Yes. The pointer is pointing into Ruby heap which won't get collected until we exit from C extension

ext/libdatadog_api/feature_flags.c
dd-oleksii: Added

ext/libdatadog_api/feature_flags.c
dd-oleksii: Done. `EvaluationContext` is now internal and get_assignment method accepts hash directly

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,382 @@
...
+  //
+  // struct ddog_ffe_ArrayMap_BorrowedStr metadata =
+  //   ddog_ffe_assignnment_get_flag_metadata(resolution_details);
+  //
+  // for (size_t i = 0; i < metadata.count; i++) {
+  //   ddog_ffe_KeyValue_BorrowedStr kv = metadata.elements[i];
+  //   VALUE key = str_from_borrow(kv.key);
+  //   VALUE value = str_from_borrow(kv.value);
+  //   rb_hash_aset(hash, key, value);
+  // }
```
dd-oleksii: For reviewers: this bug should not stop us from merging and releasing feature flags. We don't use `flag_metadata` internally and we don't make any promises to the user. (We have actually recently decided to remove all attributes from `flag_metadata`, so we don't have users accidentally relying on what is unstable)

ext/libdatadog_api/feature_flags.c
dd-oleksii: I'm OK renaming this to something else if there's a strong push but the current argument is debatable and more of a personal taste thing.

<details>
<summary>Warning: personal opinions</summary>

The "configuration" name does not imply that it's simple, static, or transparent.

Having configuration encapsulated, handle itself and return feature flag values is pretty much in spirit of OOP. This process is not computationally heavy and it can be said that configuration _is_ just providing its values.

The calling code doesn't look weird either:
```ruby
native_details = configuration.get_assignment(flag_key, expected_type.to_sym, context)
```

You shouldn't really know how much processing is going on under the hood. Some SDKs have a simpler configuration format where it's just a map from flag key to resolution details, so `get_assignment` is just a couple of lines.

The only issue I see is that there's too many different kinds of "configurations" floating around, so the name may be ambiguous in some context. This one I'm not sure how to solve as there's not many good name candidates.

"Evaluator" is a second best in this case but it's focusing on just one thing that you can do with configuration and is ignoring future things that we might do (e.g., querying for the number of flags or flag keys, storing it, etc.). It's also does not remove ambiguity as we already have `EvaluationEngine` and `NativeEvaluator`, so `CoreEvaluator` would only add confusion to the mix. Nouning is also a code smell and is a plague of software engineering 😢

</details>

lib/datadog/open_feature/native_evaluator.rb
```diff
@@ -0,0 +1,74 @@
...
+module Datadog
+  module OpenFeature
+    # Evaluation using native extension
+    class NativeEvaluator
+      def initialize(configuration)
+        @libdatadog_api_failure = Datadog::Core::LIBDATADOG_API_FAILURE
+        if configuration && !@libdatadog_api_failure
+          @configuration = Datadog::Core::FeatureFlags::Configuration.new(configuration)
+        end
+      end
```
Strech: This will be done at component level and we would not enable component if libdatadog is not loaded. https://github.com/DataDog/dd-trace-rb/pull/5047/files#diff-49253ad22909b3e9d6d6ccbde589b945d565e50dd33078addf02dd12c3069c06R35-R40

I think we can assume that it's loaded and working and will not be called otherwise.

lib/datadog/open_feature/native_evaluator.rb
```diff
@@ -0,0 +1,74 @@
...
+          variant: variant,
+          allocation_key: native_details.allocation_key,
+          reason: native_details.reason,
+          error?: native_details.error?,
+          error_code: native_details.error_code,
+          error_message: native_details.error_message,
+          log?: native_details.log?,
+          flag_metadata: native_details.flag_metadata,
+          extra_logging: {},
+        )
```
Strech: Can't we use `native_details` for that? Otherwise it's a double conversion Native -> Component -> OpenFeature. Here the result will be passed to the Provider and provider will convert it into OpenFeature SDK details.

lib/datadog/open_feature/native_evaluator.rb
```diff
@@ -0,0 +1,74 @@
...
+        end
+      end
+
+      def get_assignment(flag_key, default_value, context, expected_type)
+        return build_fatal_error(default_value) if @libdatadog_api_failure
+
+        configuration = @configuration
+        return build_not_ready_error(default_value) if configuration.nil?
+
+        native_details = configuration.get_assignment(flag_key, expected_type.to_sym, context)
```
Strech: It was converted from symbols originally to the strings because you've said that it should be a string and now we converting it to the symbol, why? I can pass a symbol from the `EvaluationEngine` level instead and avoid those allocations, but why has it changed?

> Please keep expected flag types aligned with OF: "string" | "integer" | "float" | "boolean" | "object"

[Slack](https://dd.slack.com/archives/C0959U2JP2N/p1762965416657949?thread_ts=1762948575.383089&cid=C0959U2JP2N)

sig/datadog/core/feature_flags.rbs
```diff
@@ -0,0 +1,58 @@
+module Datadog
+  module Core
+    module FeatureFlags
+      class Error < StandardError
+      end
+
+      class Configuration
+        def self.new: (String json_str) -> Configuration
+
+        # Get assignment for a feature flag
```
Strech: This documentation is expected in ruby files, not RBS. Maybe we can find a better place because no one is going to look here for the documentation

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,382 @@
...
+// the caller is responsible for the returned EvaluationContext and
+// must free it when no longer in use.
+static ddog_ffe_Handle_EvaluationContext evaluation_context_from_hash(VALUE hash) {
+  ENFORCE_TYPE(hash, T_HASH);
+
+  const char *targeting_key = NULL;
+  const long max_attr_count = RHASH_SIZE(hash);
+  ddog_ffe_AttributePair *const attrs = ruby_xcalloc(max_attr_count, sizeof(struct ddog_ffe_AttributePair));
+
+  const VALUE keys = rb_funcall(hash, rb_intern_const("keys"), 0);
```
Strech: This will allocate array, instead we should iterate the hash without an extra allocation

Strech:
@codex review

lib/datadog/open_feature/native_evaluator.rb
```diff
@@ -0,0 +1,74 @@
...
+
+        variant = native_details.variant
+        value = native_details.value
+        if expected_type == 'object' && value.is_a?(String)
+          # JSON flags return value as string. We need to parse it.
+          value = JSON.parse(value)
```
Strech: Is this possible to move to the C-ext instead? Then we can just pass the original native_details further and not allocate extra object in between.

When it comes to the `variant` for convenience (if it will simplify the code) we can handle it on evaluation engine level

lib/datadog/open_feature/evaluation_engine.rb
```diff
@@ -17,7 +17,7 @@ def initialize(reporter, telemetry:, logger:)
         @telemetry = telemetry
         @logger = logger
 
-        @evaluator = NoopEvaluator.new(nil)
+        @evaluator = NativeEvaluator.new(nil)
```
Strech: This should stay as `NoopEvaluator` as we create `NativeEvaluator` only when configuration comes. It will offload `NativeEvaluator` evaluator from handling unnecessary edge-cases which we can scope out early at a higher-level.

sig/datadog/core/feature_flags.rbs
Strech: What about [this place instead](https://github.com/DataDog/dd-trace-rb/pull/5054/files/8e202f64c25c0a04d747f7020eb72eec69713d1b?file-filters%5B%5D=.rb&show-viewed-files=true#diff-212d08b38cbfae2aeebc01a1f47b0c1f55dc6df7724e92993ff2884083ea3250)?

lib/datadog/open_feature/evaluation_engine.rb
Strech: But if you want we can get rid of `NoopEvaluator`, but maybe adjust an error message like it was in no-op.

lib/datadog/open_feature/native_evaluator.rb
dd-oleksii: Let's do symbols throughout then. Symbols are easier/faster to compare.

Sorry for the back-and-forth. In that thread I was correcting the names and wasn't paying too much attention to the types — I wasn't working on the C extension so didn't have a preference

lib/datadog/open_feature/native_evaluator.rb
dd-oleksii: I can move the call to `JSON.parse` inside the C extension and I think that makes sense.

One caveat is that this is setting up a trap: every call to `result.value` parses a JSON, which may be slow. I surveyed the code and it looks like `value` is only called once (in provider), so we should be fine. But I feel unease leaving that trap charged

wdyt?

sig/datadog/core/feature_flags.rbs
dd-oleksii: Did a bit of searching and it looks like we should place this documentation in C files.

lib/datadog/open_feature/evaluation_engine.rb
dd-oleksii: I have restored NoopEvaluator usage.

However, I tweaked its return value:
- reason: ERROR (instead of INITIALIZING. That's more standard reason and should better match other SDK)
- error_code: PROVIDER_NOT_READY
- error_message: Waiting for flags configuration (removed "universal" as that's an internal name)

lib/datadog/open_feature/native_evaluator.rb
dd-oleksii: oh, another caveat is that it's delaying parsing until later, so `JSON::ParserError` won't be raised until Provider tries to convert it (which is a tad too late 😢)

Let's keep this part as is for now

ext/libdatadog_api/feature_flags.c
dd-oleksii: done

lib/datadog/open_feature/native_evaluator.rb
dd-oleksii: UPD: I have updated the code to use symbols throughout

sig/datadog/core/feature_flags.rbs
dd-oleksii: Moved to C files

ext/libdatadog_api/feature_flags.c
dd-oleksii: Fixed

Strech:
@codex find critical bugs and edge cases in C code

lib/datadog/open_feature/native_evaluator.rb
```diff
@@ -0,0 +1,49 @@
...
+        native_details = @configuration.get_assignment(flag_key, expected_type, context)
+
+        variant = native_details.variant
+        value = native_details.value
+        if expected_type == :object && value.is_a?(String)
+          # JSON flags return value as string. We need to parse it.
+          value = JSON.parse(value)
+        elsif variant.nil?
+          value = default_value
+        end
```
Strech:
```ruby
if variant.nil?
  native_details.value = default_value
end
```

lib/datadog/open_feature/native_evaluator.rb
Strech: We can parse JSON just once and keep the value. Should be possible in C (via Ruby API) and in Ruby via instance variables. Let's try the fastest now and keep it as TODO

dd-oleksii: UPD: moved to Core ResolutionDetails after discussion

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,534 @@
...
+  VALUE feature_flags_module = rb_define_module_under(core_module, "FeatureFlags");
+
+  feature_flags_error_class = rb_define_class_under(feature_flags_module, "Error", rb_eStandardError);
+
+  VALUE configuration_class = rb_define_class_under(feature_flags_module, "Configuration", rb_cObject);
+  rb_undef_alloc_func(configuration_class);
+  rb_define_singleton_method(configuration_class, "new", configuration_new, 1);
+  rb_define_method(configuration_class, "get_assignment", configuration_get_assignment, 3);
+
+  resolution_details_class = rb_define_class_under(feature_flags_module, "ResolutionDetails", rb_cObject);
```
ivoanjo, line 98: When storing Ruby objects (in these cases classes: `feature_flags_error_class` and `resolution_details_class`) somewhere, we need to let the Ruby GC know about it, otherwise bad things can happen -- it's unlikely that Ruby will decide to GC these classes BUT it's not unlikely that `GC.compact` will move them somewhere else on the heap (and then really bad things could happen if you assume the object is still where it was).

TL;DR you probably want to call `rb_global_variable(&...)` here (so that Ruby will scan and update those references) OR `rb_gc_register_mark_object(...)` to tell the GC to effectively make them immortal and immovable.

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,534 @@
...
+static inline VALUE str_from_borrow(ddog_ffe_BorrowedStr str) {
+  if (str.ptr == NULL) {
+    return Qnil;
+  }
+
+  return rb_str_new((const char *)str.ptr, str.len);
+}
+
+void feature_flags_init(VALUE core_module) {
+  // Always define the FeatureFlags module - it will reuse existing one if it exists.
```
ivoanjo: Very minor: We use a similar pattern in many places in the gem, so I'm not sure the comment is particularly needed?

```suggestion
```

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,534 @@
...
+  rb_undef_alloc_func(resolution_details_class);
+  rb_define_method(resolution_details_class, "raw_value", resolution_details_get_raw_value, 0);
+  rb_define_method(resolution_details_class, "flag_type", resolution_details_get_flag_type, 0);
+  rb_define_method(resolution_details_class, "variant", resolution_details_get_variant, 0);
+  rb_define_method(resolution_details_class, "allocation_key", resolution_details_get_allocation_key, 0);
+  rb_define_method(resolution_details_class, "reason", resolution_details_get_reason, 0);
+  rb_define_method(resolution_details_class, "error?", resolution_details_is_error, 0);
+  rb_define_method(resolution_details_class, "error_code", resolution_details_get_error_code, 0);
+  rb_define_method(resolution_details_class, "error_message", resolution_details_get_error_message, 0);
+  rb_define_method(resolution_details_class, "log?", resolution_details_get_do_log, 0);
```
ivoanjo: Minor: It's kinda unexpected that `error?` => `is_error` but `log?` => `get_do_log`.

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,534 @@
...
+// SAFETY: The returned borrowed string points directly to Ruby's
+// internal string buffer. This is safe as long as the GVL is held
+// preventing garbage collection. (It is held automatically when C
+// extension is called.)
+static inline ddog_ffe_BorrowedStr borrow_str(VALUE str) {
+  ENFORCE_TYPE(str, T_STRING);
+  return (ddog_ffe_BorrowedStr){
+    .ptr = (const uint8_t*)RSTRING_PTR(str),
+    .len = RSTRING_LEN(str)
+  };
```
ivoanjo, line 73: I couldn't help that notice that

```c
typedef struct ddog_ffe_BorrowedStr {
  /**
   * May be NULL if `len` is `0`.
   */
  const uint8_t *ptr;
  uintptr_t len;
} ddog_ffe_BorrowedStr;
```

is basically the same as `ddog_ByteSlice`

```c
typedef struct ddog_Slice_U8 {
  /**
   * Should be non-null and suitably aligned for the underlying type. It is
   * allowed but not recommended for the pointer to be null when the len is
   * zero.
   */
  const uint8_t *ptr;
  /**
   * The number of elements (not bytes) that `.ptr` points to. Must be less
   * than or equal to [isize::MAX].
   */
  uintptr_t len;
} ddog_Slice_U8;

typedef struct ddog_Slice_U8 ddog_ByteSlice;
```

and very similar to 

```c
typedef struct ddog_Slice_CChar {
  /**
   * Should be non-null and suitably aligned for the underlying type. It is
   * allowed but not recommended for the pointer to be null when the len is
   * zero.
   */
  const char *ptr;
  /**
   * The number of elements (not bytes) that `.ptr` points to. Must be less
   * than or equal to [isize::MAX].
   */
  uintptr_t len;
} ddog_Slice_CChar;

typedef struct ddog_Slice_CChar ddog_CharSlice;
```

I guess feedback for the folks working on the libdatadog side -- consider using more of the existing types, as for instance we already have a bunch of helpers such as `char_slice_from_ruby_string` in `datadog_ruby_common.h` that can be reused for these existing types (e.g. avoids a lot of new boilerplate!)

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,534 @@
...
+
+// Create a new Ruby string from borrowed string. Returns nil if the
+// borrowed string pointer is NULL.
+static inline VALUE str_from_borrow(ddog_ffe_BorrowedStr str) {
+  if (str.ptr == NULL) {
+    return Qnil;
+  }
+
+  return rb_str_new((const char *)str.ptr, str.len);
+}
```
ivoanjo, line 84: Similar to my note above, we have a `ruby_string_from_vec_u8` that we could reuse if the types matched ;)

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,546 @@
...
+
+static void configuration_free(void *ptr) {
+  ddog_ffe_Handle_Configuration config = (ddog_ffe_Handle_Configuration)ptr;
+  ddog_ffe_configuration_drop(&config);
+}
+
+static ddog_ffe_ExpectedFlagType expected_type_from_value(VALUE expected_type) {
+  ENFORCE_TYPE(expected_type, T_SYMBOL);
+
+  const ID id = rb_sym2id(expected_type);
```
ivoanjo, line 146: I guess in this case it may be ok, but I'll share this detail which might come in handy.

TL;DR: Since Ruby 2.2 symbols [can be garbage collected](https://rubykaigi.org/2014/presentation/S-NarihiroNakamura/). But that's a it of an oversimplification -- there's actually two kinds of symbols: ones that can and ones that can't be garbage collected.

Doing `rb_sym2id` turns a symbol into one that can't be garbage collected. Since we're calling it based on arbitrary input here, any symbol here that doesn't match will life forever and be effectively leaked. So... use `sym2id` carefully.

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,546 @@
...
+    .hash = hash,
+    .targeting_key = NULL,
+    .attrs = ruby_xcalloc(RHASH_SIZE(hash), sizeof(ddog_ffe_AttributePair)),
+    .attr_count = 0,
+    .attr_capacity = RHASH_SIZE(hash)
+  };
+
+  return (ddog_ffe_Handle_EvaluationContext)rb_ensure(
+    protected_context_build, (VALUE)&builder, protected_context_cleanup, (VALUE)&builder);
+}
```
ivoanjo, line 286: There's a very dangerous mixing of types going on here!!!

In particular, if there's an exception, `protected_context_cleanup` runs and `rb_ensure` returns `Qnil` as if it was a `ddog_ffe_Handle_EvaluationContext` which would be an instant segfault.

My suggestion is changing things up a bit -- make `protected_context_build` not return anything and just set up the `builder`, and then only return a marker (like `Qtrue`) to indicate success.

Then here in this function you can go `if (rb_ensure(...) == Qtrue) return ddog_ffe_evaluation_context_new(...)`.

(In some cases casting things to `VALUE` is somewhat unavoidable but... it's a very very very sharp tool...)

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,546 @@
...
+  // Extract targeting_key separately if present
+  if (strcmp(name, "targeting_key") == 0 && TYPE(value) == T_STRING) {
+    builder->targeting_key = RSTRING_PTR(value);
+    return ST_CONTINUE;
+  }
+
+  // Skip nil values
+  if (TYPE(value) == T_NIL) {
+    return ST_CONTINUE;
+  }
```
ivoanjo, line 196: Minor:

```suggestion
  if (value == Qnil) {
    return ST_CONTINUE;
  }
```

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,546 @@
...
+    default:
+      // Default to string representation
+      attr->value.tag = DDOG_FFE_ATTRIBUTE_VALUE_STRING;
+      // SAFETY: RSTRING_PTR returns pointer to the Ruby heap, which
+      // lives long enough for ddog_ffe_evaluation_context_new to copy the underlying
+      // string. It will not get garbage collected until we exit C
+      // extension or release GVL.
+      attr->value.string = RSTRING_PTR(rb_funcall(value, rb_intern("to_s"), 0));
+      break;
+  }
```
ivoanjo, line 232: I don't believe this comment is correct. How will the Ruby GC know to keep this string alive? Between calls to `to_s` anything can happen -- Ruby may decide to switch away to another thread and do millions of allocations, GC cycles, compaction, etc.

Once we introduce a "thread may switch away here" point in the code, all bets on "nothing gets moved from under us" are off... :/

I'm not even sure the other `RSTRING_PTR`s used (e.g. for the targeting key) are safe from Ruby changing things as e.g. Ruby has a few tricks to share storage between strings and can decide to undo them. 

To avoid this my suggestion is to do less in native code. Write a Ruby version of the code that turns everything into one of the expected basic types, and only then go into the native code to do the final conversion. (Or just do a single upcall somewhere in the native code into that helper method, and then come back to C). This way here you can go "it's an expected type or I just raise and stop everything" and not need to worry about what's going to happen if you call into the VM.

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,546 @@
...
+  struct evaluation_context_builder *builder = (struct evaluation_context_builder *)arg;
+
+  ENFORCE_TYPE(key, T_STRING);
+  const char *name = RSTRING_PTR(key);
+
+  // Extract targeting_key separately if present
+  if (strcmp(name, "targeting_key") == 0 && TYPE(value) == T_STRING) {
+    builder->targeting_key = RSTRING_PTR(value);
+    return ST_CONTINUE;
+  }
```
ivoanjo: Minor: Should we somehow raise or complain if the `targeting_key` isn't the expected type?

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,546 @@
...
+ * call-seq:
+ *   resolution_details.variant() -> String or nil
+ *
+ * Get the variant identifier.
+ *
+ * @return [String, nil] The variant identifier or nil
+ */
+static VALUE resolution_details_get_variant(VALUE self) {
+  ddog_ffe_Handle_ResolutionDetails resolution_details =
+    (ddog_ffe_Handle_ResolutionDetails)rb_check_typeddata(self, &resolution_details_typed_data);
```
ivoanjo, line 404: Minor: It's kinda weird that in half the methods we break this into 2 lines and others it's in one line?

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,546 @@
...
+ * Check if the resolution resulted in an error.
+ *
+ * @return [Boolean] True if there was an error
+ */
+static VALUE resolution_details_is_error(VALUE self) {
+  ddog_ffe_Handle_ResolutionDetails resolution_details =
+    (ddog_ffe_Handle_ResolutionDetails)rb_check_typeddata(self, &resolution_details_typed_data);
+  enum ddog_ffe_Reason reason = ddog_ffe_assignment_get_reason(resolution_details);
+  return reason == DDOG_FFE_REASON_ERROR ? Qtrue : Qfalse;
+}
```
ivoanjo: Minor: Maybe turn this into Ruby code? The less C code the better ;)

ext/libdatadog_api/feature_flags.c
ivoanjo: So will this forever return an empty hash? Or is it the intention to still enable this at some point?

ivoanjo (commented):
Hey! @Strech asked me to do a pass on the native code, hope this helps! 🙇

ext/libdatadog_api/feature_flags.c
dd-oleksii: `is_log` doesn't make sense here and the underlying field is called "doLog" in configuration and virtually everywhere else

ext/libdatadog_api/feature_flags.c
dd-oleksii: Yeah, makes sense. I will take a look in next releases of libdatadog. They do have different semantics in how they handle NULL pointer though, so they are not exactly compatible

ext/libdatadog_api/feature_flags.c
dd-oleksii: Similarly to previous comment, they handle NULL differently but will take a look at how to consolidate them

ext/libdatadog_api/feature_flags.c
dd-oleksii: Yeah, I'm aware of that and we don't intern/symbolize arbitrary input. `expected_type` is tightly controlled on Ruby side so only known values are passed in

ext/libdatadog_api/feature_flags.c
dd-oleksii: oh, that's a good catch! I misread the doc and was assuming rb_ensure is re-throwing exception if it occurred. It looks like it doesn't, so will rework this handling (it's probably simpler with rb_protect)

ext/libdatadog_api/feature_flags.c
dd-oleksii: It may not switch or do anything with Ruby state as we're holding GVL. This is a basic assumption and otherwise you couldn't do _anything_ in C as Ruby might have decided to relocate things between any two calls.

By anything I mean that even this basic call does not work (first line from doc):
```c
result = rb_funcall(obj, rb_intern("=="), 1, Qnil);
```
because call to `rb_intern` might trigger GC and relocate obj.

ext/libdatadog_api/feature_flags.c
dd-oleksii: Do we use an automatic code formatter for C?

ext/libdatadog_api/feature_flags.c
dd-oleksii: We will enable it at some point
1. We will uncomment the code here once the bug is fixed in libdatadog.
2. The libdatadog version will return empty metadata for now but this will change in the future as we decide what data we want to expose to users.

ext/libdatadog_api/feature_flags.c
dd-oleksii: We will try to proceed and evaluate the flag without a targeting key. If targeting key turns out to be required the underlying evaluator will return an error. I'll add a comment here

ext/libdatadog_api/feature_flags.c
dd-oleksii: Updated this piece to use rb_protect and much less VALUE-casting! Thanks again for the find

ext/libdatadog_api/feature_flags.c
dd-oleksii: Used `rb_global_variable()`! thanks for the catch

ext/libdatadog_api/feature_flags.c
dd-oleksii: I have fixed it manually for now 🙌

ext/libdatadog_api/feature_flags.c
dd-oleksii: Yeah, I just recently found that I can mix Ruby and C on the same class, so moving 😎

ext/libdatadog_api/feature_flags.c
ivoanjo: > It may not switch or do anything with Ruby state as we're holding GVL.

Unfortunately that's not correct. Specifically -- yes, you're holding the GVL when you came into C code, but once you use `rb_funcall` you've dispatched back to the Ruby VM and anything can happen: you can lose the GVL, GC can happen, fiber switches, exceptions, fork(), etc.

Hence my suggestion of rearranging the code so that you're back to the basic types, and then the assumption of "we don't switch away" is back on the table again.

ext/libdatadog_api/feature_flags.c
ivoanjo: We've historically been lazy in setting it up but if you'd find it useful, please do send a PR with it!

(I've not been a fan of most formatters I've seen being used so I was never too motivated to look into this)

ext/libdatadog_api/feature_flags.c
ivoanjo: (Note that yes, Ruby will restore the GVL to you once `rb_funcall` returns so effectively when this code is executing you do have the GVL back; it's just that you can no longer assume the ruby state "hasn't moved" once you introduce the ability for a thread switch to happen. [Ask me how I know...](https://github.com/DataDog/dd-trace-rb/pull/4240) -- btw the `unsafe_api_calls_check.h` here can be used if you want to add test coverage for this to the code)

dd-oleksii: I have to say that Ruby is super-unfriendly to C extensions 😭 _and_ very lacking in documentation about these things.

It looks like we're good as long as we don't call into any Ruby code, so I have removed `to_s` call here — it was not necessary here and more of an attempt to be nice and process things that we don't have to. I also surveyed that we have no other rb_funcall anywhere in the extension.

(Moving part of this code to Ruby is possible but doesn't seem to provide any benefit as we still have to iterate the structure and copy it in the way C expects)

dd-oleksii: Also added a warning note on `borrow_str`

ext/libdatadog_api/feature_flags.c
ivoanjo: Up to you, but perhaps it may be worth reflecting some of these details in the comment in the code :)

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,544 @@
...
+static ID id_any;
+static ID id_integer;
+static ID id_float;
+
+// SAFETY: The returned borrowed string points directly to Ruby's
+// internal string buffer.
+//
+// This is safe as long as the GVL is held preventing garbage
+// collection. It is held automatically when C extension is called.)
+// Note that calling into any Ruby cade (rb_funcall, or even
```
Strech:
```suggestion
// collection. It is held automatically when C extension is called.
// Note that calling into any Ruby code (rb_funcall, or even
```

ext/libdatadog_api/feature_flags.c
dd-oleksii: Expanded the note and opened [FFL-1450](https://datadoghq.atlassian.net/browse/FFL-1450) to track it 🫡

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,548 @@
...
+  }
+
+  return rb_str_new((const char *)str.ptr, str.len);
+}
+
+void feature_flags_init(VALUE core_module) {
+  VALUE feature_flags_module = rb_define_module_under(core_module, "FeatureFlags");
+
+  feature_flags_error_class = rb_define_class_under(feature_flags_module, "Error", rb_eStandardError);
+  rb_gc_register_address(&feature_flags_error_class);
```
ivoanjo, line 89: Note that the docs for `rb_gc_register_address` say that this function should be called [before assigning to the variables](https://github.com/ruby/ruby/blob/5f5da2c293766b98e543b3215b1af7ad6950977d/include/ruby/internal/gc.h#L393). (Applies to the other call below too)

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,548 @@
...
+
+  // Skip nil values
+  if (value == Qnil) {
+    return ST_CONTINUE;
+  }
+
+  // Ensure we don't exceed capacity (should not happen if capacity = hash size)
+  if (builder->attr_count >= builder->attr_capacity) {
+    return ST_STOP;
+  }
```
ivoanjo: Minor: Since this should not happen -- maybe we could raise instead, rather than silently ignoring?

ivoanjo (approved):
I'm a bit short on time today so I gave a pass only on the `feature_flags.c` file -- that part LGTM (after a final suggested fix)

ext/libdatadog_api/feature_flags.c
dd-oleksii: 🙇

ext/libdatadog_api/feature_flags.c
dd-oleksii: Ideally, this would be a debug assert. Do we have something like that? I've found `DD_DEBUG` which is only used once

ivoanjo: Usually we turn these into exceptions -- the fact that "something unexpected went wrong" came from native code vs Ruby code isn't particularly relevant in most situations (e.g. this happening is not very different from an `ENFORCE_TYPE` failing)

lib/datadog/core/feature_flags.rb
```diff
@@ -0,0 +1,56 @@
...
+          return @value if defined?(@value)
+
+          # NOTE: Raw value method call doesn't support memoization right now
+          value = raw_value
+
+          # NOTE: Lazy parsing of the JSON is a temporary solution and will be
+          #       moved into C extension
+          @value = json?(value) ? JSON.parse(value) : value
+        rescue JSON::ParserError => e
+          raise Error, "Failed to parse JSON value: #{e.message}"
```
y9v: do we want to raise this error, or logging and reporting telemetry would be enough?

lib/datadog/open_feature/native_evaluator.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+require_relative '../core/feature_flags'
+
+module Datadog
+  module OpenFeature
+    # This class is an interface of evaluation logic using native extension
+    class NativeEvaluator
+      def initialize(configuration)
+        @configuration = Core::FeatureFlags::Configuration.new(configuration)
```
y9v, line 13: this is a bit weird - why are we creating configuration from configuration? I guess the argument here is some configuration hash? Can we try to find a better name for it?

lib/datadog/open_feature/native_evaluator.rb
```diff
@@ -0,0 +1,24 @@
...
+      end
+
+      def get_assignment(flag_key, default_value, context, expected_type)
```
y9v: 4 positional arguments is a bit too much, can we use keyword args here?

Something like:
```ruby
def assignment_for_key(key, default_value:, context:, expected_type:)
end
```

lib/datadog/core/feature_flags.rb
p-datadog:
```suggestion
          raise Error, "Failed to parse JSON value: #{e.class}: #{e}"
```

lib/datadog/open_feature/native_evaluator.rb
dd-oleksii: The input configuration is a json string. Perhaps `configuration_json` is a more fitting name

lib/datadog/open_feature/provider.rb
```diff
@@ -119,6 +119,13 @@ def evaluate(flag_key, default_value:, expected_type:, evaluation_context:)
           reason: result.reason,
           flag_metadata: result.flag_metadata
         )
+      rescue => e
+        ::OpenFeature::SDK::Provider::ResolutionDetails.new(
+          value: default_value,
+          error_code: Ext::PROVIDER_FATAL,
```
dd-oleksii: `PROVIDER_FATAL` means that provider entered irrecoverable error state. This should be `PARSE_ERROR` for json exception, or `GENERAL` for everything else instead.

lib/datadog/core/feature_flags.rb
dd-oleksii: We raise this error here because there's little ResolutionDetails can do about it. It is rescued in `Provider`, which wraps it in error `ResolutionDetails` as OpenFeature requires from us

lib/datadog/core/feature_flags.rb
```diff
@@ -0,0 +1,56 @@
...
+        # Check if the resolution resulted in an error
+        #
+        # @return [Boolean] True if there was an error
+        def error?
+          reason == 'ERROR'
+        end
+
+        private
+
+        def json?(value)
```
p-datadog, line 55: Looking at the code in  this class, it seems like this method does not in fact check or verify that the argument is valid JSON, therefore it shouldn't be called `json?` but could be called, perhaps, `string?`.

lib/datadog/core/feature_flags.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
+require 'json'
+
+module Datadog
+  module Core
+    # Feature flags evaluation using libdatadog
+    # The classes in this module are defined as C extensions in ext/libdatadog_api/feature_flags.c
+    module FeatureFlags
```
p-datadog, line 11: Is this module meant to be used by customers? If not, please add `@api private` annotation and check the other classes and annotate them as well.

lib/datadog/open_feature/evaluation_engine.rb
```diff
@@ -38,14 +39,18 @@ def fetch_value(flag_key:, default_value:, expected_type:, evaluation_context: n
         @telemetry.report(e, description: 'OpenFeature: Failed to fetch flag value')
 
         ResolutionDetails.build_error(
-          value: default_value, error_code: Ext::PROVIDER_FATAL, error_message: e.message
+          value: default_value, error_code: Ext::GENERAL, error_message: e.message
         )
       end
 
       def reconfigure!(configuration)
```
p-datadog, line 51: Can this method have a comment that says what type `configuration` is supposed to be?

lib/datadog/open_feature/evaluation_engine.rb
```diff
@@ -38,14 +39,18 @@ def fetch_value(flag_key:, default_value:, expected_type:, evaluation_context: n
...
+        if configuration.nil?
+          @logger.debug('OpenFeature: Removing configuration')
 
-        @evaluator = NoopEvaluator.new(configuration)
+          return @evaluator = NoopEvaluator.new(configuration)
+        end
+
+        @evaluator = NativeEvaluator.new(configuration)
       rescue => e
         message = 'OpenFeature: Failed to reconfigure, reverting to the previous configuration'
```
p-datadog, line 60: Can you please change the next line below to use `#{e.class}: #{e}` pattern to report both the class of the exception and its message, and check the rest of the PR for where exceptions are reported to use the same pattern.

lib/datadog/open_feature/native_evaluator.rb
p-datadog: `configuration_json` is clearer than `configuration`, however given that it is common for programmers to say "JSON" when they mean a data structure rather than its serialized textual form, `configuration_string` I think would be a better choice in this case.

Could you also add a comment stating where this configuration string is expected to come from?

lib/datadog/open_feature/native_evaluator.rb
p-datadog: :+1: Can you please change to keyword arguments?

p-datadog: And also can you please add a comment stating what the method does?

lib/datadog/open_feature/provider.rb
```diff
@@ -119,6 +119,13 @@ def evaluate(flag_key, default_value:, expected_type:, evaluation_context:)
           reason: result.reason,
           flag_metadata: result.flag_metadata
         )
+      rescue => e
+        ::OpenFeature::SDK::Provider::ResolutionDetails.new(
+          value: default_value,
+          error_code: Ext::GENERAL,
+          error_message: e.message,
```
p-datadog: e.class & e

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,548 @@
...
+
+#include "datadog_ruby_common.h"
+
+#include <stdint.h>
+#include <string.h>
+#include <stdlib.h>
+
+// Forward declarations
+static VALUE configuration_new(VALUE klass, VALUE json_str);
+static void configuration_free(void *ptr);
```
p-datadog, line 15: Is this function taking a configuration as a parameter? If yes `configuration` would be a better name for the argument, `void *` already indicates it's a pointer.

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,548 @@
...
+  }
+
+  return rb_str_new((const char *)str.ptr, str.len);
+}
+
+void feature_flags_init(VALUE core_module) {
+  VALUE feature_flags_module = rb_define_module_under(core_module, "FeatureFlags");
+
+  rb_gc_register_address(&feature_flags_error_class);
+  feature_flags_error_class = rb_define_class_under(feature_flags_module, "Error", rb_eStandardError);
```
p-datadog, line 90: This class is also defined in Ruby code, what is the purpose of having two definitions?

dd-oleksii:
@p-datadog I'll work on addressing the comments but could you please clarify what comments are actually blocking the merge?

lib/datadog/core/feature_flags.rb
dd-oleksii: It kind of does. `flag_type == :object` means that the raw_value is expected to be JSON-serialized string. `is_a?(String)` check is secondary for extra protection (and imo could be dropped)

ext/libdatadog_api/feature_flags.c
p-datadog: The comment says "should not happen if capacity = hash size". Capacity and hash size of what? And are they expected to be the same?

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,548 @@
...
+      return LONG2NUM(value.integer);
+    case DDOG_FFE_VARIANT_VALUE_FLOAT:
+      return rb_float_new(value.float_);
+    case DDOG_FFE_VARIANT_VALUE_BOOLEAN:
+      return value.boolean ? Qtrue : Qfalse;
+    case DDOG_FFE_VARIANT_VALUE_OBJECT:
+      return str_from_borrow(value.string);
+    case DDOG_FFE_VARIANT_VALUE_NONE:
+    default:
+      return Qnil;
```
p-datadog, line 354: Should the default case raise an exception instead of being treated as None?

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,548 @@
...
+      attr->value.boolean = true;
+      break;
+    case T_FALSE:
+      attr->name = name;
+      attr->value.tag = DDOG_FFE_ATTRIBUTE_VALUE_BOOLEAN;
+      attr->value.boolean = false;
+      break;
+    default:
+      // Skip unsupported attribute types.
+      return ST_CONTINUE;
```
p-datadog, line 231: Is there an exception guard around the entire feature? If so, would it make sense to raise an exception here instead of silently ignoring unhandled input?

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,548 @@
...
+      return ID2SYM(id_integer);
+    case DDOG_FFE_VARIANT_VALUE_FLOAT:
+      return ID2SYM(id_float);
+    case DDOG_FFE_VARIANT_VALUE_BOOLEAN:
+      return ID2SYM(id_boolean);
+    case DDOG_FFE_VARIANT_VALUE_OBJECT:
+      return ID2SYM(id_object);
+    case DDOG_FFE_VARIANT_VALUE_NONE:
+    default:
+      return Qnil;
```
p-datadog, line 387: Ignoring unhandled input here also

ext/libdatadog_api/feature_flags.c
```diff
@@ -0,0 +1,548 @@
...
+  return str_from_borrow(allocation_key);
+}
+
+/*
+ * call-seq:
+ *   resolution_details.reason() -> String
+ *
+ * Get the reason for the resolution.
+ *
+ * @return [String] One of: "STATIC", "DEFAULT", "TARGETING_MATCH", "SPLIT", "DISABLED", "ERROR", "UNKNOWN"
```
p-datadog, line 430: If these are returned to Ruby code, what is the reasoning behind the values being uppercased strings rather than symbols? Ruby generally would use symbols in similar situations in my opinion.

lib/datadog/core/feature_flags.rb
p-datadog: I see, do you mind adding this as a code comment for future readers?

ext/libdatadog_api/feature_flags.c
dd-oleksii: Yes, there is a guard. No, it doesn't make sense because we can continue and produce results in most cases.

ext/libdatadog_api/feature_flags.c
dd-oleksii: These are dictated by OpenFeature and we are complying with their interface. See also: https://github.com/DataDog/dd-trace-rb/pull/5054#discussion_r2536498227

ext/libdatadog_api/feature_flags.c
dd-oleksii: @p-datadog this function is a callback for `rb_hash_foreach` and is iterating over a Hash. The comment is referring to the size of that Hash that is being iterated on and capacity is `attr_capacity` (which is a capacity of `builder->attrs`). They are in fact the same (see [line 261](https://github.com/DataDog/dd-trace-rb/pull/5054/files/868878989a9603b32d7a5c89f3bbfd82b9ac1acd#diff-ba6a5ff386a0d50c34100cce034a05440488f4aca7b921aa2ba2b61940d02353R261)):
```c
  struct evaluation_context_builder builder = {
    .hash = hash,
    .targeting_key = NULL,
    .attrs = ruby_xcalloc(RHASH_SIZE(hash), sizeof(ddog_ffe_AttributePair)),
    .attr_count = 0,
    .attr_capacity = RHASH_SIZE(hash)
  };
```

lib/datadog/open_feature/native_evaluator.rb
Strech: We are not going to add the data type to the name as it brakes some basics of the encapsulation, today it's a string, tomorrow it's a Hash.

At the same time we might need to rename it into a better name to reflect better what is it. Let us thing, and come up with a better version

lib/datadog/open_feature/evaluation_engine.rb
Strech: We can't include message in telemetry as it could expose the private information, but the error class for sure!

lib/datadog/open_feature/native_evaluator.rb
Strech: Makes sense!

lib/datadog/open_feature/native_evaluator.rb
Strech: I think we are going to resolve it with just comment on what is expected. It doesn't matter what configuration is and it's a subject to change (probably).

I tried configuration_payload or so, but it's bells and whistles. I would rather go with comment

ext/libdatadog_api/feature_flags.c
dd-oleksii: Yes and no. `ptr` is a `data` portion of [`RTypedData`](https://docs.ruby-lang.org/capi/en/master/d6/d45/struct_r_typed_data.html). It happens to be `ddog_ffe_Handle_Configuration` right now, so it's technically a configuration but not the one exposed to Ruby.

Note that this function is not exposed to Ruby, it's a part of integration with GC

ext/libdatadog_api/feature_flags.c
dd-oleksii: We need one here so we can refer to it later. I _think_ we can remove the one defined in Ruby, although it would make the code confusing.

cc @Strech for your opinion

ivoanjo: (For what it's worth, I like keeping the definitions on the Ruby side too, as a lot of tools gets confusing with the Ruby -> C transition and so it's clearer to have something on the Ruby side even though it gets used from C.)

(Instead of defining a class, it's possible to read from the constant instead, but the amount of code is similar so I'd keep it as-is)

PR #5056, opened 2025-11-18
Add top-level type `any` as semantic version of `untyped`

Strech:
**What does this PR do?**

Add a Datadog top-level semantic type `any`.

**Motivation:**

The type `any` is going to represent a scenario when the narrowing is impossible to do or when it doesn't make sense to do. And type `untyped` should represent an actual untyped type (no time, not done, etc)

**Change log entry**

No.

**Additional Notes:**

We can get more value out of `untyped` checks once we start moving to `any` in case when it's really any type.

**How to test the change?**

CI

### DIFF
--- a/sig/datadog.rbs
+++ b/sig/datadog.rbs
@@ -1,4 +1,26 @@
 module Datadog
+  # This type is a semantic version of `untyped` when the accepted or returned
+  # value is literally anything. And `untyped` should be used as a signal that
+  # the typing is not finished yet.
+  #
+  # WARNING: This type should be used sparingly and only when there are no distinct
+  #          types that can be used to represent the value.
+  #
+  # Examples:
+  #
+  #   when Hash values are impossible to narrow down to a specific type
+  #
+  #   ```rbs
+  #   def my_function: (::Hash[::String, any] payload) -> void
+  #   ```
+  #
+  #   or when an arbitrary Array with uncontrolled values is needed
+  #
+  #   ```rbs
+  #   def my_function: (::Array[any] payload) -> void
+  #   ```
+  type any = untyped
+
   extend Core::Extensions
   extend Core::Configuration
 


lloeki:
Very cool.

AIUI both are _logically_ equivalent type-wise, they are _semantically_ not:

* `untyped`: "I did not have time to define the proper type for this, so I accept everything and the name serves as a marker to revisit later."
* `any`: "I did have time to thoroughly think about what this type should be and it is definitely intended to accept _every possible type_, so I am communicating that this is _intendend as the proper type here_."

It's [a bit like](https://github.com/ruby/rbs/blob/master/docs/syntax.md#void-boolish-or-top):

> `void`, `boolish`, or `top`?
> 
> They are all equivalent for the type system; they are all top type.
> 
> `void` tells developers a hint that the value should not be used. `boolish` implies the value is used as a truth value. `top` is anything else.

For context `any` was a thing but was renamed `untyped`:

```
sig/datadog/open_feature/exposures/worker.rbs:48:23: [error] Cannot find type `any`
│ Diagnostic ID: RBS::UnknownTypeName
│
└         def perform: (*any) -> void
```

Maybe we can make a case upstream to have both?

vpellan (approved):

sig/datadog.rbs
```diff
@@ -1,4 +1,26 @@
 module Datadog
+  # This type is a semantic version of `untyped` when the accepted or returned
+  # value is literally anything. And `untyped` should be used as a signal that
+  # the typing is not finished yet.
+  #
+  # WARNING: This type should be used sparingly and only when there are no distinct
+  #          types that can be used to represent the value.
```
lloeki, line 7: Many a time it's indeed possible to use generics, and possibly generics with a union type, which one can conveniently shorten with an alias type.

But in some cases it's definitely `any`.

lloeki (approved):
LGTM

sig/datadog.rbs
Strech: Let me try to propose that change to the upstream too.

marcotc: I agree with @Strech.
I feel like any large enough project that uses RBS would have this same issue. We can figure out together how communicate this need upstream.

PR #5066, opened 2025-11-19
Telemetry: default OnlyOnceSuccessful limit to nil

p-datadog:
**What does this PR do?**

Changes default limit of OnlyOnceSuccessful from 0 to nil.

**Motivation:**
We have nil to signify absence of value.

**Change log entry**
None

**Additional Notes:**

**How to test the change?**
Existing CI

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -77,7 +77,7 @@ jobs:
   test:
     needs:
       - build
-    uses: DataDog/system-tests/.github/workflows/system-tests.yml@7cc82692d3d02b17f20d6d0fcbee09de21f27ad4  # Automated: This reference is automatically updated.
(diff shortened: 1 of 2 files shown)


lib/datadog/core/utils/only_once_successful.rb
```diff
@@ -81,6 +76,15 @@ def reset_ran_once_state_for_tests
             @retries = 0
           end
         end
+
+        private
```
ivoanjo: This exposes `limited?` and `reset_ran_once_state_for_tests` which seems like an accident...?

p-datadog: That is actually part of a future PR, I'll fix it here.

p-datadog:
It's not currently bothering me so I will leave it alone for now...

PR #5067, opened 2025-11-19
Telemetry: inject dependencies into telemetry app started event

p-datadog:
**What does this PR do?**
Injects component tree into telemetry AppStarted event constructor, instead of various methods of the event referencing various global variables.

**Motivation:**
Make telemetry code easier to read/reason about for the upcoming change to enable telemetry in child processes

**Change log entry**
None

**Additional Notes:**
The event does not hold a reference to the component tree - the tree is used only while constructor is running to build the various payloads that will be sent out later.

The unit tests now perform many references to `Datadog.send(:components)` which looks odd (unit tests reference global state) but this was existing behavior - telemetry was reading global state including in unit tests.

**How to test the change?**
Existing CI


lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -18,27 +22,28 @@ def type
...
           private
 
-          def products
+          def products(components)
             # @type var products: Hash[Symbol, Hash[Symbol, Hash[Symbol, String | Integer] | bool | nil]]
             products = {
               appsec: {
-                enabled: Datadog::AppSec.enabled?,
+                # TODO take appsec status out of component tree?
+                enabled: components.settings.appsec.enabled,
```
p-datadog, line 40: @y9v @Strech I can change this to read the state from component tree rather than configuration, like the other components, if you agree?

lib/datadog/profiling/profiler.rb
```diff
@@ -17,6 +17,10 @@ def initialize(worker:, scheduler:)
         @scheduler = scheduler
       end
 
+      def enabled?
+        !!scheduler&.running?
```
ivoanjo: Minor: If the `Profiler` exists, then a `Scheduler` exists -- the previous code had the `&.` when calling running just because the profiler could be `nil`, not because the scheduler could.

```suggestion
        !!scheduler.running?
```

lib/datadog/core/configuration/components.rb
```diff
@@ -120,7 +121,7 @@ def initialize(settings)
           # This agent_settings is intended for use within Core. If you require
           # agent_settings within a product outside of core you should extend
           # the Core resolver from within your product/component's namespace.
-          agent_settings = AgentSettingsResolver.call(settings, logger: @logger)
+          @agent_settings = AgentSettingsResolver.call(settings, logger: @logger)
```
ivoanjo, line 124: TBH I hesitated a while for this one at first... It's weird to think of the agent settings "as a component", but in practice what's happening here is more akin to treating components as a "DI container", closer to something like [dry-container](https://dry-rb.org/gems/dry-container/0.11/).

So looking at it in that direction, maybe it's not a bad direction to evolve in... So yeah looks reasonable after thinking about it a bit.

lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -18,27 +22,28 @@ def type
...
               },
               profiler: {
-                enabled: Datadog::Profiling.enabled?,
+                enabled: !!components.profiler&.enabled?,
               },
```
ivoanjo, line 44: I guess this was already like this before this PR from an earlier change... but is it me or does this check get made **before** the profiler actually is started in the components start?

E.g. it looks like profiler will always be disabled for this event? This was probably introduced when we started sending this event from components and not on first trace...

lib/datadog/profiling/profiler.rb
```diff
@@ -17,6 +17,10 @@ def initialize(worker:, scheduler:)
         @scheduler = scheduler
       end
 
+      def enabled?
+        !!scheduler.running?
```
p-datadog:
```suggestion
        scheduler.running?
```

lib/datadog/core/telemetry/component.rb
```diff
@@ -109,13 +109,17 @@ def disable!
           @worker&.enabled = false
         end
 
-        def start(initial_event_is_change = false)
+        def start(initial_event_is_change = false, components:)
```
p-datadog, line 112: This is ugly and I think I want keyword arguments for both, I can change that here or leave as is to keep PR smaller.

lib/datadog/core/telemetry/event/app_started.rb
p-datadog: I am investigating, but you seem to be correct - I enabled profiling and the app-started payload claims profiling is disabled.

lib/datadog/core/telemetry/event/app_started.rb
Strech: ~Yes, thanks a lot!~

Strech: ~I think it's fine now as we don't have 1-click activation, but once we will do it, we need to remember changing that.~

lib/datadog/core/telemetry/event/app_started.rb
ivoanjo: It occurred to me that we could fix this by changing `Profiling.enabled?` to be `!profiler.nil?`. The logic in [Profiling::Component.build_profiler_component](https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/profiling/component.rb#L13) only returns a `Profiler` if everything is in good shape.

I'm thinking we don't really need the complexity of the distinction of "is the profiler enabled AND ready AND started" (which is what we implicitly have right now) and "is the profiler enabled AND ready" is enough for this use-case.

lib/datadog/core/telemetry/event/app_started.rb
Strech: After the discussion with the team AppSec guild, the answer was - it should be an actual state (that we really was able to activate it) because the configuration state will be included in the initial telemetry event anyway.

So the answer is nope, please keep it as it was.

lib/datadog/core/telemetry/event/app_started.rb
ivoanjo: (Since I don't think this issue started by any change in this PR, I don't think it needs to be fixed in this PR -- we can fix separately)

p-datadog:
Agree, all of the references to the component tree are not great. I haven't looked deeply into what it would take to eliminate them -my thinking is it might be tricky because telemetry is pretty much collecting the state of everything else in dd-trace-rb, thus either the test becomes extremely mocky or we have to construct a lot of variables in the tests.

lib/datadog/core/telemetry/event/app_started.rb
p-datadog: This is fixed in https://github.com/DataDog/dd-trace-rb/pull/5075 by changing the `enabled?` logic as you suggested.

PR #5072, opened 2025-11-19
add process tags to profiling

wantsui:
**What does this PR do?**

Implements AIDM-254, which is to add process tags to the profiling payloads when the experimental DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED is added. Note I don't think this should be in the changelog until we are more ready with this feature.

Requires https://github.com/DataDog/libdatadog/pull/1359 

**Motivation:**

AIDM-254

**Change log entry**

No.

**Additional Notes:**

When reviewing this, keep in mind that we need the implementations across tracers to be consistent:
Java PR: https://github.com/DataDog/dd-trace-java/pull/8780
Python PR: https://github.com/DataDog/dd-trace-py/pull/15226
.NET PR: https://github.com/DataDog/dd-trace-dotnet/pull/7715

This PR is dependent on
- https://github.com/DataDog/dd-trace-rb/pull/5033/ is merged.
- https://github.com/DataDog/libdatadog/pull/1359

Open Question(s):
1. Is there a concept of a "root" payload in dd-trace-rb for profiling?

**How to test the change?**

I ran the ` integration/apps/ruby/app/fibonacci.rb` script against a real Datadog Agent and can see the output of the tags in the profiles:

image

### DIFF
--- a/spec/datadog/profiling/http_transport_spec.rb
+++ b/spec/datadog/profiling/http_transport_spec.rb
@@ -369,6 +369,7 @@ RSpec.describe Datadog::Profiling::HttpTransport do
         boundary = request["content-type"][%r{^multipart/form-data; boundary=(.+)}, 1]
         body = WEBrick::HTTPUtils.parse_form_data(StringIO.new(request.body), boundary)
         event_data = JSON.parse(body.fetch("event"))
+
         expect(event_data["process_tags"]).to eq(process_tags)
       end
     end


ivoanjo:
Now that #5033 has landed, this probably needs a rebase/merge from master :eyes:

wantsui:
I'm closing this one until I get more updates on libdatadog. I'll reopen once I'm ready to work on it again.

wantsui:
Note: The new libdatadog version with the profiling payload change is out: https://github.com/DataDog/libdatadog/releases/tag/v25.0.0, however I don't see it in the gems yet: https://rubygems.org/gems/libdatadog. 

(I'm asking around to figure out steps are needed here).

ivoanjo:
@wantsui the magic is in https://github.com/DataDog/libdatadog/tree/main/ruby#releasing-a-new-version-to-rubygemsorg ! Lmk if you need any help :)

benchmarks/profiling_http_transport.rb
```diff
@@ -57,6 +57,7 @@ def flush
       code_provenance_file_name: 'example_code_provenance_file_name.json',
       code_provenance_data: '',
       tags_as_array: [],
+      process_tags: nil,
```
wantsui: note: this is not enough because it turns out one of the benchmarks failed because it needs the old payload format: https://gitlab.ddbuild.io/DataDog/apm-reliability/dd-trace-rb/-/jobs/1354472003

wantsui: resolved in 725d2f72f3985ea4cce1836ba6aed9eb2120433a, hopefully

benchmarks/profiling_http_transport.rb
```diff
@@ -59,7 +59,12 @@ def flush
       tags_as_array: [],
       internal_metadata: {no_signals_workaround_enabled: false},
       info_json: JSON.generate({profiler: {benchmarking: true}}),
-    )
+    }
+
+    flush_params = Datadog::Profiling::Flush.instance_method(:initialize).parameters
+    params[:process_tags] = nil if flush_params.any? { |type, name| name == :process_tags }
+
+    Datadog::Profiling::Flush.new(**params)
```
ivoanjo, line 69: Ahaha sorry you ran into this issue and it's a damn creative solution xD

I think the only missing bit is documenting WTH is going on here for the future 

```suggestion
    # This was added as a workaround to allow https://github.com/DataDog/dd-trace-rb/pull/5072 to be merged
    # Once that PR is merged, this can be cleaned up :)
    flush_params = Datadog::Profiling::Flush.instance_method(:initialize).parameters
    params[:process_tags] = nil if flush_params.any? { |type, name| name == :process_tags }

    Datadog::Profiling::Flush.new(**params)
```

lib/datadog/core/process_discovery.rb
```diff
@@ -44,7 +44,7 @@ def get_metadata(settings)
             service_name: settings.service || '',
             service_env: settings.env || '',
             service_version: settings.version || '',
-            # TODO: Implement process tags and container id
+            # TODO: Implement process tags and container id and determine how it changes with this PR
             process_tags: '',
```
ivoanjo, line 48: This looks to still be TODO? Is there something to be done for this PR?

lib/datadog/profiling/exporter.rb
```diff
@@ -70,6 +70,12 @@ def flush
 
         uncompressed_code_provenance = code_provenance_collector.refresh.generate_json if code_provenance_collector
 
+        process_tags = nil
+        if Datadog.configuration.experimental_propagate_process_tags_enabled
+          value = Datadog::Core::Environment::Process.serialized
+          process_tags = value unless value.nil? || value.empty?
```
ivoanjo: Minor: I was reminded here of a book called [Confident Ruby](https://pragprog.com/titles/agcr/confident-ruby/), which I very much recommend!

In particular, this code is not very confident -- is value going to be nil? is it empty? is it a old discarded bubblegum wrapper?

TL;DR is it a problem to pass an empty string to libdatadog? We have quite a lot of boilerplate code in this PR that boils down to trying to get this nil across, and that code would be much simplified if we just said "it's always a string, but can be empty".

(And by "minor" I mean, I'm leaving this feedback in case you agree and want to try improving the code, but it's not a blocker if you/we have better things to do :D)

lib/datadog/profiling/tag_builder.rb
```diff
@@ -2,6 +2,7 @@
 
 require_relative '../core/tag_builder'
 require_relative '../core/utils'
+require_relative '../core/environment/ext'
```
ivoanjo: What's up with this change? Is it still needed?

spec/datadog/profiling/flush_spec.rb
```diff
@@ -9,6 +9,7 @@
     let(:code_provenance_file_name) { "the_code_provenance_file_name.json" }
     let(:code_provenance_data) { "the_code_provenance_data" }
     let(:tags_as_array) { [%w[tag_a value_a], %w[tag_b value_b]] }
+    let(:process_tags) { [%w[process_tag_a value_a], %w[process_tag_b value_b]] }
```
ivoanjo: Wait, isn't `process_tags` supposed to be a string?

ivoanjo:
Also I don't usually suggest this, but looking at how small the current PR is AND how noisy the development history is -- consider squash merging this one? (If you don't have permissions, ping me and I can do it once you're ready)

benchmarks/profiling_http_transport.rb
```diff
@@ -59,7 +59,12 @@ def flush
       tags_as_array: [],
       internal_metadata: {no_signals_workaround_enabled: false},
       info_json: JSON.generate({profiler: {benchmarking: true}}),
-    )
+    }
+
+    flush_params = Datadog::Profiling::Flush.instance_method(:initialize).parameters
+    params[:process_tags] = nil if flush_params.any? { |type, name| name == :process_tags }
```
Strech:
```suggestion
    params[:process_tags] = nil if flush_params.any? { |_, name| name == :process_tags }
```

spec/datadog/profiling/http_transport_spec.rb
```diff
@@ -356,6 +358,21 @@
...
+
+      it "includes the process tags in the payload" do
+        success = http_transport.export(flush)
+
+        expect(success).to be true
+
+        boundary = request["content-type"][%r{^multipart/form-data; boundary=(.+)}, 1]
+        body = WEBrick::HTTPUtils.parse_form_data(StringIO.new(request.body), boundary)
+        event_data = JSON.parse(body.fetch("event"))
+        expect(event_data["process_tags"]).to eq(process_tags)
```
Strech, line 373:
```suggestion
        event_data = JSON.parse(body.fetch("event"))

        expect(event_data["process_tags"]).to eq(process_tags)
```

lib/datadog/core/process_discovery.rb
wantsui: Thanks for catching that, that was a reminder to myself to review the process tags as related to profiling and I forgot to take it out. I think the only part of this comment that I should actually keep is the container id one since I have made no changes on that.
```suggestion
            # TODO: Implement container id
```

wantsui: Actually given that I was able to get this to appear on profiles without changing this part, it looks like Process Discovery is a completely different thing?
I reverted the comment: https://github.com/DataDog/dd-trace-rb/pull/5072/commits/76056ebea9a07af5410a68aa80d7e1e154bfad3d but kept it as a TODO because if there are any changes, I will do it in a separate PR.

lib/datadog/profiling/tag_builder.rb
wantsui: removed in https://github.com/DataDog/dd-trace-rb/pull/5072/commits/d4f22ee4817de1e6a323794e1d94b2946a945fe0#diff-ab318643a3f2e3ba9e3056d5d9acf8045bdff65bd055ed49de4786b34d42e320!

lib/datadog/profiling/exporter.rb
wantsui: I really appreciate that you pointed this out because it led me to dig into exactly what the libdatadog change was for this. Turns out that it will accept an empty string ("") so there was no need for me to send the nil value.

https://github.com/DataDog/libdatadog/blob/51c8cb4eebe9fe245fc739033b7a3494311e520f/libdd-profiling/src/cxx.rs#L167-L168

```
        /// * `process_tags` - Process-level tags as comma-separated string (e.g.,
        ///   "runtime:native,profiler_version:1.0") Pass empty string "" if not needed
```


As a result, I refactored this in a060d9326bfb3c555b84c4cd2ed67bdcb4719a5e and d4f22ee4817de1e6a323794e1d94b2946a945fe0

spec/datadog/profiling/flush_spec.rb
wantsui: I refactored this in a060d9326bfb3c555b84c4cd2ed67bdcb4719a5e and d4f22ee4817de1e6a323794e1d94b2946a945fe0 to be a string, so this is now addressed :)

lib/datadog/core/process_discovery.rb
ivoanjo: Process discovery is indeed unrelated to sending profiles and traces, AFAIK; it's about exposing information about the process in a different way, for an external reader (usually the datadog agent or something like that) to find it.

So I'm guessing we'll probably want to include the `process_tags` here BUT I think that's an entirely different PR.

wantsui: I am tracking this in AIDM-279 and will resolve this comment!

PR #5073, opened 2025-11-20
[APMAPI-1774] Fix stable config segfault during error handling

vpellan:
**What does this PR do?**

This fixes a segfault that happens when stable config returns an error. he error is free twice

**Motivation:**

https://github.com/DataDog/dd-trace-rb/actions/runs/19506133836/job/55832894119?pr=5062

**Change log entry**

Yes. Stable config: Fix potential segfault during configuration file parsing

**Additional Notes:**

**How to test the change?**


ext/libdatadog_api/library_config.c
```diff
@@ -167,6 +148,5 @@ static VALUE _native_configurator_get(VALUE self) {
   rb_hash_aset(result, ID2SYM(rb_intern("local")), local_hash);
   rb_hash_aset(result, ID2SYM(rb_intern("fleet")), fleet_hash);
 
-  RB_GC_GUARD(config_logged_result_rb);
```
ivoanjo, line 170: There's a memory leak being flagged because we're not doing what `config_logged_result_free` did before -- the `ddog_library_config_drop` and the `free`.

Having said that, eliminating the Ruby object means we don't actually need the `ruby_xcalloc`, we can do stack allocation instead:

```ruby
 ddog_LibraryConfigLoggedResult configurator_logged_result = ddog_library_configurator_get(configurator);
```

(we'll still need the `ddog_library_config_drop` call)

ext/libdatadog_api/library_config.c
```diff
@@ -116,7 +98,6 @@ static VALUE _native_configurator_get(VALUE self) {
     } else {
       log_warning_without_config(message);
     }
-    RB_GC_GUARD(config_logged_result_rb);
     return rb_hash_new();
   }
```
ivoanjo, line 115: I think this change makes sense BUT I think we should document clearly the assumptions that make it make sense.

Specifically, we're assuming it's OK to hold on to libdatadog-allocated memory until the end of the method (where we'll need to clear it, see my comment below) **because it's highly unlikely any of the work we're doing will raise exceptions and cause the memory to leak**.

We should document the start and end of "don't throw exceptions here" so that in the future this assumption is respected, as otherwise in the next PR someone may innocently add a random call that would break the assumptions.

Minor: Additionally, there's one more thing we _optionally_ could do to make the code tighter, which is to do as much work/allocations as possible before actually calling into libdatadog. This would reduce the "critical region" where we're worrying about exceptions.

Specifically, `local_config_hash`, `fleet_config_hash`, `local_hash`, `fleet_hash`, `result` and a lot of their calls can happen before we actually hold on to libdatadog memory. If there's an error and that effort was useless... well... that's a shame, but errors are supposed to be rare so it's not a problem.

ext/libdatadog_api/library_config.c
vpellan: I reverted to using the ruby GC, but moved TypedData_Wrap_Struct after the error handling, so it will be easier to maintain later

ext/libdatadog_api/library_config.c
```diff
@@ -116,10 +115,11 @@ static VALUE _native_configurator_get(VALUE self) {
     } else {
       log_warning_without_config(message);
     }
-    RB_GC_GUARD(config_logged_result_rb);
     return rb_hash_new();
```
ivoanjo: We're actually leaking the memory from the `ruby_xcalloc` above in this branch.
My suggestion is to move the xcalloc together with the `TypedData_Wrap_Struct`.

Something like:

```c
ddog_LibraryConfigLoggedResult *configurator_logged_result = NULL;

// Introduce a scope so that the stack-allocated `result` does not get used after `configurator_logged_result` is initialized 
{
  // (No malloc at first...)
  ddog_LibraryConfigLoggedResult result = ddog_library_configurator_get(configurator);


  if (configurator_logged_result->tag == DDOG_LIBRARY_CONFIG_LOGGED_RESULT_ERR) {
    ...
  }
  
   // (Now we malloc and copy the result into the malloc)
  ddog_LibraryConfigLoggedResult *configurator_logged_result = ruby_xcalloc(1, sizeof(ddog_LibraryConfigLoggedResult));
  *configurator_logged_result = result;
}
```

vpellan: Something like this ? [8e7b502](https://github.com/DataDog/dd-trace-rb/pull/5073/commits/8e7b502e78240147d949a5a0ddcff0362acd13c2)

ext/libdatadog_api/library_config.c
```diff
@@ -99,27 +99,29 @@ static VALUE _native_configurator_get(VALUE self) {
...
   }
 
+  // Wrapping config_logged_result into a Ruby object enables the Ruby GC to manage its memory
+  // We need to allocate memory for config_logged_result because once it is out of scope, it will be freed (at the end of this function)
+  // So we cannot reference it with &config_logged_result
```
ivoanjo: Minor: Minor simplification -- I don't think this part is still relevant/needed

```suggestion
```

PR #5074, opened 2025-11-20
Telemetry: send events in forked children

p-datadog:
**What does this PR do?**
Changes telemetry to send events in forked children

**Motivation:**

**Change log entry**
Yes: fix Live Debugger / Dynamic Instrumentation UI for forking web servers

**Additional Notes:**
Telemetry has some special logic to deal with app-started/configuration-change events due to the component tree being created multiple times by dd-trace-rb, but system tests expecting a single set of events.

This PR further augments that logic to go back from configuration-change events to app-started in the forked children, which are reported as brand new processes.

**How to test the change?**

Tests added

### DIFF
--- a/docs/TelemetryDevelopment.md
+++ b/docs/TelemetryDevelopment.md
@@ -1,0 +1,66 @@
+# Telemetry Development
+
+## Telemetry Presence
+
+`dd-trace-rb` is written to assume that the telemetry component is always
+present. If telemetry is disabled, the component is still created but does
+nothing.
+
+Most components call methods on `telemetry` unconditionally. There are two
+exceptons: DI and Data Streams are written to assume that `telemetry` may be nil.
+However, this assumption is not necessary and these components may be
+changed in the future to assume that `telemetry` is always present.
+
+## Event Submission Prior To Start
+
+Telemetry is unique among other components in that it permits events to be
+submitted to it prior to its worker starting. This is done so that errors
+during `Datadog.configure` processing can be reported via telemetry, because
+the errors can be produced prior to telemetry worker starting. The telemetry
+component keeps the events and sends them after the worker starts.
+
+## Initial Event
+
+`dd-trace-rb` can be initialized multiple times during application boot.
+For example, if customers follow our documentation and require
+`datadog/auto_instrument`, and call `Datadog.configure`, they would get
+`Datadog.configure` invoked two times total (the first time by `auto_instrument`)
+and thus telemetry instance would be created twice. This happens in the
+applications used with system tests.
+
+System tests, on the other hand, require that there is only one `app-started`
+event emitted, because they think the application is launched once.
+To deal with this we have a hack in the telemetry code to send an
+`app-client-configuration-change` event instead of the second `app-started`
+event. This is implemented via the `SynthAppClientConfigurationChange` class.
+
+## Fork Handling
+
+We must send telemetry data from forked children.
+
+Telemetry started out as a diagnostic tool used during application boot,
+but is now used for reporting application liveness (and settings/state)
+throughout the application lifetime. Live Debugger / Dynamic Instrumentation,
+for example, require ongoing `app-heartbeat` events emitted via telemetry
+to provide a working UI to customers.
+
+It is somewhat common for customers to preload the application in the parent
+web server process and process requests from children. This means telemetry
+is initialized from the parent process, and it must emit events in the
+forked children.
+
+We use the standard worker `after_fork` handler to recreated the worker
+thread in forked children. However, there are two caveats to keep in mind
+which are specific to telemetry:
+
+1. Due to telemetry permitting event submission prior to its start, it is
+not sufficient to simply reset the state from the worker's `perform` method,
+as is done in other components. We must only reset the state when we are
+in the forked child, otherwise we'll trash any events submitted to telemetry
+prior to its worker starting.
+
+2. The child process is a brand new application as far as the backend/UI is
+concerned, having a new runtime ID, and therefore the initial event in the
+forked child must always be `app-started`. Since we track the initial event
+in the telemetry component, this event must be changed to `app-started` in
+forked children regardless of what it was in the parent.


lib/datadog/core/utils/only_once.rb
```diff
@@ -31,9 +31,7 @@ def ran?
           @mutex.synchronize { @ran_once }
         end
 
-        private
-
-        def reset_ran_once_state_for_tests
+        def reset
           @mutex.synchronize { @ran_once = false }
         end
```
ivoanjo, line 38: I don't think this is the right way to go -- I think this is bending the `OnlyOnce` in a really weird direction generally only to support a specific use-case of one of its users.

Rather than resetting the only once, consider creating a new instance?

p-datadog: I reverted this and create a new instance as you requested.

lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -11,6 +11,12 @@ class AppStarted < Base
           def initialize(components:)
             # To not hold a reference to the component tree, generate
             # the event payload here in the constructor.
+            #
+            # Important: do not store data that contains (or is derived from)
+            # the runtime_id oor sequence numbers.
```
marcotc:
```suggestion
            # the runtime_id or sequence numbers.
```

lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -11,6 +11,12 @@ class AppStarted < Base
...
+            # This event is reused when a process forks, but in the
+            # child process the runtime_id would be different and sequence
+            # number would obviously also be different.
```
marcotc:
```suggestion
            # number is reset.
```

lib/datadog/core/telemetry/component.rb
```diff
@@ -110,7 +119,7 @@ def disable!
         end
 
         def start(initial_event_is_change = false, components:)
-          return if !@enabled
+          return unless enabled?
 
           initial_event = if initial_event_is_change
```
marcotc, line 124: I was reviewing the changes in `SynthAppClientConfigurationChange` and I was wondering: do you think it would make sense to check for forking in this class (`Component`) so that we can change this `if` condition to:
```suggestion
          initial_event = if initial_event_is_change && !@forked
```
This would mean we make no changes to `SynthAppClientConfigurationChange`. The changes in `SynthAppClientConfigurationChange`, the "mode switching", were hard to follow for me, since I wasn't sure why it would need to pretend to be a different event sometimes. Turns out it's because of of this `Component#start` method. I was thinking that we could keep all the logic around `AppStarted` vs `SynthAppClientConfigurationChange` here.

(Btw, in terms of detecting a fork: it either be done here, or borrowed from `worker.forked?`, but that's probably a weird dependency)

lib/datadog/core/telemetry/worker.rb
```diff
@@ -240,6 +260,28 @@ def disable_on_not_found!(response)
...
+        # The closing event will be (or should be) sent by the worker
+        # in the parent process.
+        # Also, discard any accumulated events since they will be sent by
+        # the parent.
+        def after_fork
+          # If telemetry is disabled, we still reset the state to avoid
+          # having wrong state. It is possible that in the future telemetry
+          # will be re-enabled after errors.
+          buffer.clear
+          @initial_event_once = Utils::OnlyOnceSuccessful.new(APP_STARTED_EVENT_RETRIES)
```
ivoanjo: Consider moving this to a separate method, that is shared with `initialize`. With such big classes/with complex logic, it's easy to change something and then completely forget to change the `after_fork` to match.

If `initialize` and `after_fork` share the logic, than it's easier to both remember that they both exist (you can even leave a comment in that method with a reminder) and keep the logic in-sync.

As a bonus -- you can maybe even recreate the buffer there and thus not need to add `clear` either ;)

lib/datadog/core/telemetry/worker.rb
```diff
@@ -240,6 +260,28 @@ def disable_on_not_found!(response)
...
+          # In the child process, we get a new runtime_id.
+          # As such we need to send AppStarted event.
+          # In the parent process, the event may have been the
+          # SynthAppClientConfigurationChange instead of AppStarted,
+          # and in that case we need to convert it to the "regular"
+          # AppStarted event.
+          if @initial_event.is_a?(Event::SynthAppClientConfigurationChange)
+            @initial_event.reset! # steep:ignore
```
Strech: Why to ignore? I think it could be fixed with annotation instead

sig/datadog/core/telemetry/event/synth_app_client_configuration_change.rbs
```diff
@@ -3,9 +3,13 @@ module Datadog
     module Telemetry
       module Event
         class SynthAppClientConfigurationChange < AppStarted
-          def type: () -> "app-client-configuration-change"
+          def type: -> ("app-client-configuration-change" | "app-started")
```
Strech: TBH I think it should be unspecific instead, because the point is to give type, not the value

```suggestion
          def type: -> ::String
```

lib/datadog/core/telemetry/worker.rb
p-datadog: I have made the suggested change.

lib/datadog/core/telemetry/component.rb
p-datadog: While I expect SynthAppClientConfigurationChange to be used in the vast majority of cases (customers who follow our instructions to require auto_instrument and then call Datadog.configure), it won't be used if customers are manually loading minimal instrumentation or do not configure the tracer programmatically. Therefore I don't think this check can be only in SynthAppClientConfigurationChange.

I do recognize the complexity of initial event + fork handling and I'm thinking to write some docs explaining the situation.

Per our discussion, I will also make a follow-up PR that uses the at_fork monkey patch instead of the after_fork handling that requires the worker to poll process id (since there is the extra gotcha having to do with events being submitted to telemetry prior to the worker starting).

p-datadog:
I added  a new documentation file that contains all of the info I think is relevant to telemetry development (initial event handling, fork handling, event submission prior to worker start).

PR #5075, opened 2025-11-20
Telemetry: fix reporting of profiling and appsec state

p-datadog:
**What does this PR do?**
Fixes telemetry reporting of profiling and appsec enabled/disabled state (in app-started events).

Profiler was reported as not running in the app-started event regardless of actual state, due to the way enablement check was implemented - it looked at whether profiler was started, not just enabled, and profiler was started after telemetry by the component tree.

This PR changes the check to consider whether the Profiler instance exists, which should indicate whether  profiling is actually enabled and working.

AppSec was reported based on its configuration, not actual state. If AppSec was requested to be enabled but failed prerequisites (Ruby version/FFI version) or failed to initialize, it would still be reported as enabled.

We have the configuration for each component sent separately which contains the requested enabled/disabled values, therefore the product field should have the actual state.

I took the liberty to change `AppSec.enabled?` to read component state also.

Other, smaller changes while I was looking at the code:

Telemetry is now shut down last, to permit all other code to send diagnostics during shutdown. Previously process discovery was shut down last, it didn't send anything via telemetry that I saw, but I moved it ahead of telemetry.

**Motivation:**
Code review in https://github.com/DataDog/dd-trace-rb/pull/5067

**Change log entry**
None

**Additional Notes:**

AppSec does not store the settings into its component, therefore all configuration is not necessarily correct (during reconfiguration, existing components will report/use new configuration rather than "their" configuration). This most notably makes `rasp_enabled?` code weird in this PR, but applies to all of AppSec configuration from a quick look.

AppSec initialization block did not report the exception it rescued, I changed it to do so (the tests I wrote didn't report any reason why AppSec failed to initialize without this change).

Only profiling reports the reason why it wasn't enabled, DI and AppSec do not.

**How to test the change?**
Integration tests added to verify both configured enabled/disabled state and actual enabled/disabled state for all products reported to telemetry (AppSec, DI, profiling).

### DIFF
--- a/sig/datadog/appsec/component.rbs
+++ b/sig/datadog/appsec/component.rbs
@@ -3,8 +3,6 @@ module Datadog
     class Component
       @security_engine: SecurityEngine::Engine?
 
-      @telemetry: Core::Telemetry::Component
-
       def self.build_appsec_component: (Core::Configuration::Settings settings, telemetry: Core::Telemetry::Component) -> Component?
 
       private
@@ -15,9 +13,7 @@ module Datadog
 
       attr_reader security_engine: AppSec::SecurityEngine::Engine?
 
-      attr_reader telemetry: Core::Telemetry::Component
-
-      def initialize: (security_engine: AppSec::SecurityEngine::Engine, telemetry: Core::Telemetry::Component) -> void
+      def initialize: (security_engine: AppSec::SecurityEngine::Engine) -> void
 
       def reconfigure!: () -> void
 


p-datadog:
So this PR is on top of https://github.com/DataDog/dd-trace-rb/pull/5074 and most of the diff actually was (and was reviewed) in that PR.

lib/datadog/appsec/component.rb
```diff
@@ -46,8 +46,8 @@ def build_appsec_component(settings, telemetry:)
 
           security_engine = SecurityEngine::Engine.new(appsec_settings: settings.appsec, telemetry: telemetry)
           new(security_engine: security_engine, telemetry: telemetry)
-        rescue
-          Datadog.logger.warn('AppSec is disabled, see logged errors above')
+        rescue => exc
+          Datadog.logger.warn("AppSec is disabled: #{exc.class}: #{exc}; there may be additional logged errors above")
```
Strech:
```suggestion
        rescue => e
          Datadog.logger.warn("AppSec is disabled: #{e.class}: #{e}; there may be additional logged errors above")
```

spec/datadog/core/telemetry/integration/telemetry_spec.rb
```diff
@@ -520,4 +520,405 @@
...
+        )
+        expect(payload.dig('payload', 'products')).to include(
+          'appsec' => {'enabled' => true},
+        )
+
+        assert_remaining_events
+      end
+    end
+
+    context 'when appsec is requested to be enabled but fails initialization' do
```
Strech, line 776: I think for completeness we need to add a test that appsec is disabled, technically same change were done to profiling

Strech (approved):
LGTM, I've left a note on disabled state of the component tests missing

spec/datadog/core/telemetry/integration/telemetry_spec.rb
p-datadog: I did not quite understand this comment because there are no tests for any product where that product is explicitly disabled. But I added them for all 3 products.

p-datadog:
@codex review

p-datadog:
Changes since last approval by @Strech : claude found missing telemetry reporting in appsec rescue (fixed) and repaired broken integration tests due to asserting on incorrect field name in payload (config var-style di.enabled vs env var-style DD_DYNAMIC_INSTRUMENTATION_ENABLED). Tests were also rewritten for clarity to help understand this fix.

lib/datadog/profiling.rb
```diff
@@ -61,8 +61,7 @@ def self.allocation_count
     end
 
     def self.enabled?
-      profiler = Datadog.send(:components).profiler
-      !!profiler&.enabled?
+      !!Datadog.send(:components).profiler
```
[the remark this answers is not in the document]
p-datadog: Checked the implementation — it looks fine as is. The `products.profiler.enabled` field in the app-started event is intended to capture initialization state at startup — whether profiling was configured and successfully initialized — rather than whether the scheduler thread happens to be running at that moment. The previous `scheduler.running?` check was reading runtime state at a point where the scheduler hadn't been started yet, so it always returned `false` regardless of configuration.

On the downstream caller: `Tracer#profiling_enabled` memoizes with `||=`, so it evaluates once and caches — runtime state changes after initialization don't affect it in either the old or new implementation.

Runtime failures post-startup would be reported through error telemetry rather than through the app-started snapshot.

lib/datadog/appsec.rb
```diff
@@ -11,25 +11,21 @@ module Datadog
...
 
-      # NOTE:  This is a temporary workaround for type checking.
-      #
-      #        We want to move from possible nil-component to the disabled-component
-      #        on an initialization error. Technically, telemetry will be never
-      #        used if AppSec was not able to initialize, so it's safe to assume
-      #        that telemetry will never be used and will be nil at the same time.
       def telemetry
-        components.appsec&.telemetry || components.telemetry
+        components.telemetry
```
y9v, line 28: this is a good change, but I think it also makes sense to remove `@telemetry` from AppSec component, and the attr reader for it:
https://github.com/DataDog/dd-trace-rb/blob/539fead22bea6868b5af821e4f5415066ccde232/lib/datadog/appsec/component.rb#L78

spec/datadog/core/telemetry/integration/telemetry_spec.rb
```diff
@@ -523,4 +523,302 @@
...
+      end
+      expect(integrations_batch).not_to be_nil
+    end
+
+    # Configuration names use env var names (DD_PROFILING_ENABLED, not
+    # profiling.enabled) because AppStarted#option_telemetry_name prefers
+    # option.definition.env over the setting path when an env var is defined.
+    shared_examples 'reports requested configuration and actual product state' do |product_key:, configuration:, actual_state:|
+      requested = configuration[:value]
+      running = actual_state[:enabled] || actual_state['enabled']
```
eregon:
```suggestion
      running = actual_state.fetch('enabled')
```
It doesn't seem used with a symbol, and better fail early if passing an incorrect expectation.
EDIT: It could even be a simple boolean, but some cases check a message in addition, that could be a separate Hash key.

spec/datadog/appsec/component_spec.rb
```diff
@@ -48,7 +48,7 @@
       it 'returns nil when security engine fails to instantiate' do
         settings.appsec.ruleset = {}
 
-        expect(telemetry).to receive(:report)
+        expect(telemetry).to receive(:report).twice
```
vpellan: Maybe add a comment (or even a test) to give information about what the reports are ? I suppose it's the one in `require_libddwaf` + the one you added in `build_appsec_component`. And tht makes me wonder if we really want to report it twice, or only the one in `build_appsec_component` is enough?

spec/datadog/core/telemetry/integration/telemetry_spec.rb
```diff
@@ -523,4 +523,302 @@
...
+
+    let(:component) { Datadog.send(:components).telemetry }
+
+    # Helper function to not copy/paste all of these settings for each
+    # test case, and also to not call Datadog.configure twice for each test.
+    def common_configuration(c)
+      c.agent.port = http_server_port
+      c.telemetry.enabled = true
+      c.telemetry.metrics_aggregation_interval_seconds = 1
+    end
```
vpellan: Could it be possible to do the opposite ? E.g. have the `Datadog.configure` block with these settings in the top level `before` block, with a call to a method defined in a `let` in examples? So you'd have a single `Datadog.configure` block?

vpellan (approved):
LGTM for Core/Tracing, left two non-blocking comments

lib/datadog/appsec/component.rb
p-datadog: Good catch — already applied in a previous push (renamed to `e` to match the rest of the codebase).

lib/datadog/appsec.rb
p-datadog: You're right — removed `@telemetry`, the `attr_reader`, and the `telemetry:` keyword from `initialize` / `new`. All runtime code already goes through `AppSec.telemetry` → `components.telemetry`, so the component instance never needs its own reference.

spec/datadog/core/telemetry/integration/telemetry_spec.rb
p-datadog: Good suggestion — switched to `actual_state.fetch('enabled')`. The hash always uses string keys here, so the symbol fallback was unnecessary and `fetch` will flag incorrect expectations early.

spec/datadog/core/telemetry/integration/telemetry_spec.rb
p-datadog: Good call — refactored in 2fd3892c14. Single `Datadog.configure` in a top-level `before`, with `let(:product_mock_setup)` and `let(:product_configuration)` overridden per context. RSpec's lazy `let` resolution means the inner context's overrides evaluate when the outer `before` calls them, so mocks are set up before `Datadog.configure` runs.

spec/datadog/appsec/component_spec.rb
p-datadog: Indeed, for this case the reports are almost identical.  Removed the outer one in 2c6022a396 with a TODO to reconsider where telemetry reporting should live.

p-datadog:
@codex review

PR #4985, opened 2025-10-16
Remove pii_safe to remove dynamic messages from logs backend

bouwkast:
**What does this PR do?**
Removes the `pii_safe` parameter introduced in PR #4728 to ensure telemetry error logs comply with constant message requirements. We are not supposed to be sending dynamic message content to the logs intake for various reasons (that is what we got agreement to send, we have deduplication logic to limit number of logs, and we don't want any risk whatsoever of sending up PII).

**Motivation:**

Logs have been blocked from being ingested for many months now with exclusion filters.
We'd like to get logs back in ASAP as we want to start proactively resolving errors in `dd-trace-` libraries that are surfaced from these logs as they go into our Error Tracking product.

**Change log entry**

None.

**Additional Notes:**

**I am _unfamiliar_ with `dd-trace-rb` and I asked Claude Code to help with the revert of `pii_safe`, please review this PR as such.**

This is based off of my document for my findings for Ruby here https://docs.google.com/document/d/16SFeOp7yxzT6dM1O0TKtWkgDsiq6SeV8P_MAYboCdOI/edit?tab=t.o3m3y2eqq0gx

There are a few instances of string interpolation used, I've looked into the flow for the values that they'd be and I think they map to a small subset of known strings from what I can tell, so I am fine overlooking them, but there is risk that we interpolate other future messages:

- https://github.com/DataDog/dd-trace-rb/blob/c5e3981cf80a8936195a247232243c250f1d8ecc/lib/datadog/di/probe_notifier_worker.rb#L277
- https://github.com/DataDog/dd-trace-rb/blob/c5e3981cf80a8936195a247232243c250f1d8ecc/lib/datadog/tracing/distributed/propagation.rb#L83
- https://github.com/DataDog/dd-trace-rb/blob/c5e3981cf80a8936195a247232243c250f1d8ecc/lib/datadog/tracing/contrib/patcher.rb#L55

- `probe_notifier` appears to be two possibilities: `probe status` and `snapshot`
- `propagation` appears to be values such as ` Datadog::Tracing::Distributed::Propagation::HTTP` (our propagators)
- `patcher` appears to be our integration patchers within `dd-trace-rb` so something like `Datadog::Tracing::Contrib::Rails::Patcher` -> there are many of these but seems reasonable IMO

**How to test the change?**

🤷 
I removed / altered tests associated with this parameter so if the current test suite passes then hopefully good.

**Jira**

[...]

PR #5076, opened 2025-11-21
Telemetry-safe error reporting for native extensions

marcotc:
_**(I'll squash the commits on reviews finish)**_

**What does this PR do?**

This PR ensure errors raised from native `ext/` code have enough context to be valuable in telemetry.

**Motivation:**

Because error information sent to telemetry cannot have arbitrary, dynamic data, we must ensure that we are only sending values that are known at gem build time.

Because our fallback is to only report the exception class, this PR adds a `telemetry_message` to native exceptions, so that their error information is not completely lost.

**Change log entry**
 Yes. Telemetry: Added static error reporting for native extensions.

**Additional Notes:**

There's a follow up PR to reduce code duplication for the `datadog_ruby_common.c/h` files: https://github.com/DataDog/dd-trace-rb/pull/5088

**How to test the change?**
Easiest and fastest test it: `bundle exec rake clean compile && bundle exec rake spec:profiling`
There are more products supported by `libdatadog`, but at that point, just run CI :)

### DIFF
--- a/ext/datadog_profiling_native_extension/ruby_helpers.c
+++ b/ext/datadog_profiling_native_extension/ruby_helpers.c
@@ -12,8 +12,6 @@ static ID _id2ref_id = Qnil;
 static ID inspect_id = Qnil;
 static ID to_s_id = Qnil;
 
-static ID telemetry_message_id = Qnil;
-
 void ruby_helpers_init(void) {
   rb_global_variable(&module_object_space);
 
@@ -21,8 +19,6 @@ void ruby_helpers_init(void) {
   _id2ref_id = rb_intern("_id2ref");
   inspect_id = rb_intern("inspect");
   to_s_id = rb_intern("to_s");
-
-  telemetry_message_id = rb_intern("@telemetry_message");
 }
 
 // Internal helper for raising pre-formatted syserr exceptions


ext/datadog_profiling_native_extension/private_vm_api_access.c
```diff
@@ -46,6 +46,17 @@
...
+extern VALUE eNativeArgumentError;
+extern VALUE eNativeTypeError;
+// Declare exception class globals from ruby_helpers.h
+// (We can't include ruby_helpers.h here as it pulls in public Ruby headers that conflict with private VM headers)
+#define raise_error(native_exception_class, fmt, ...) \
+  _raise_error(native_exception_class, "" fmt, ##__VA_ARGS__)
+NORETURN(
+  void _raise_error(VALUE native_exception_class, const char *fmt, ...)
+  __attribute__ ((format (printf, 2, 3)));
+);
```
ivoanjo: Yeah this file is kinda... always a pain. TBH maybe it's easy to just not change anything here -- we'll miss exceptions from this file, but they're rare-ish so I'm not sure we care.

Either that or perhaps having a very simple `extern raise_native_error(char *);` that gets exported elsewhere and that this class uses directly (so this class doesn't get to pick the specific exception class or template stuff -- always get a native error + a static string)"?

ext/datadog_profiling_native_extension/profiling.c
```diff
@@ -54,6 +54,13 @@ void DDTRACE_EXPORT Init_datadog_profiling_native_extension(void) {
...
   rb_funcall(native_extension_module, rb_intern("private_class_method"), 1, ID2SYM(rb_intern("native_working?")));
+  // Exception classes defined in Ruby, in the `Datadog::Profiling` namespace
+  eNativeRuntimeError = rb_const_get(profiling_module, rb_intern("NativeRuntimeError"));
+  rb_global_variable(&eNativeRuntimeError);
+  eNativeArgumentError = rb_const_get(profiling_module, rb_intern("NativeArgumentError"));
+  rb_global_variable(&eNativeArgumentError);
+  eNativeTypeError = rb_const_get(profiling_module, rb_intern("NativeTypeError"));
+  rb_global_variable(&eNativeTypeError);
 
   ruby_helpers_init();
```
ivoanjo: Defining these should maybe move to `ruby_helpers_init`?

ext/datadog_profiling_native_extension/setup_signal_handler.c
```diff
@@ -38,7 +38,7 @@ static void install_sigprof_signal_handler_internal(
   sigemptyset(&signal_handler_config.sa_mask);
 
   if (sigaction(SIGPROF, &signal_handler_config, &existing_signal_handler_config) != 0) {
-    rb_exc_raise(rb_syserr_new_str(errno, rb_sprintf("Could not install profiling signal handler (%s)", handler_pretty_name)));
+    raise_syserr(errno, true, "Could not install profiling signal handler", __FILE__, __LINE__, handler_pretty_name);
   }
```
ivoanjo, line 42: Hmmm passing in `handler_pretty_name` as the function name argument to `raise_syserr` is... a bit weird. Is it a problem to template this string?

If needed, I'm thinking we can turn this into an enum -- we always know which `handler_pretty_name`s are allowed, since it's all in the code anyway...

ext/datadog_profiling_native_extension/setup_signal_handler.c
```diff
@@ -71,11 +71,7 @@ static void install_sigprof_signal_handler_internal(
       );
```
ivoanjo, line 71: :arrow_up: The `rb_exc_raise` on line 59 one possibly needs to change to match the others :eyes:

ext/libdatadog_api/crashtracker.c
ivoanjo, line 1: I'm guessing some of the shared error code may need to move to `datadog_ruby_common.h`/`.c` so it can be used here too...

ext/datadog_profiling_native_extension/ruby_helpers.c
```diff
@@ -19,35 +24,78 @@ void ruby_helpers_init(void) {
...
+// Raises a NativeError exception with seperate telemetry-safe and detailed messages.
+void _raise_native_error(VALUE native_exception_class, const char *detailed_message, VALUE static_message) {
+  #ifdef DD_DEBUG
+    if (native_exception_class != eNativeRuntimeError &&
+        native_exception_class != eNativeArgumentError &&
+        native_exception_class != eNativeTypeError) {
+        rb_raise(rb_eRuntimeError, "[ddtrace] BUG: _raise_native_error called with an exception that might not support two error messages. "
+            "Must be one of eNativeRuntimeError, eNativeArgumentError, or eNativeTypeError, was: %s", rb_class2name(native_exception_class));
+    }
+  #endif
```
ivoanjo: TBH I don't think this check needs to be protected by `DD_DEBUG` -- anyway something is already wrong, and we're about to throw an exception, so throwing a different exception about a second bug seems fine too -- we just need to fix one before we see the next one.

ext/datadog_profiling_native_extension/ruby_helpers.c
```diff
@@ -19,35 +24,78 @@ void ruby_helpers_init(void) {
   _id2ref_id = rb_intern("_id2ref");
   inspect_id = rb_intern("inspect");
   to_s_id = rb_intern("to_s");
+  new_id = rb_intern("new");
+}
+// Raises a NativeError exception with seperate telemetry-safe and detailed messages.
+void _raise_native_error(VALUE native_exception_class, const char *detailed_message, VALUE static_message) {
```
ivoanjo: Minor: Every call that is see to `_raise_native_error` seems to be of a shape `_raise_native_error(..., rb_str_new_cstr(something))` so perhaps that last argument could be a `char *` anyway?

ext/datadog_profiling_native_extension/ruby_helpers.c
```diff
@@ -19,35 +24,78 @@ void ruby_helpers_init(void) {
...
+
+  _raise_native_error(
+    args->exception_class,
+    args->exception_message,
+    rb_str_new_cstr(args->telemetry_message)
+  );
 }
 
-void grab_gvl_and_raise(VALUE exception_class, const char *format_string, ...) {
+void _grab_gvl_and_raise(VALUE native_exception_class, const char *format_string, ...) {
```
ivoanjo: So in terms of naming, I guess the `_` prefix here is used to indicate this is somehow internal... I think? I'm usually a sucker for more explicit names, so perhaps this + similar functions could be called `helper_...` or `private_...` or  `internal_...` so they pop out more as "oh this should only be used in specific places".

Also, I realize I've not been particularly consistent at marking such things clearly so far lol :D

(Something something complaint about visibility of things in C being annoying...)

ext/datadog_profiling_native_extension/ruby_helpers.c
```diff
@@ -19,35 +24,78 @@ void ruby_helpers_init(void) {
...
-      "grab_gvl_and_raise called by thread holding the global VM lock. exception_message: '%s'",
-      args.exception_message
+    // Render telemetry message without formatted arguments, only the exception class.
+    char telemetry_message[MAX_RAISE_MESSAGE_SIZE];
+    snprintf(
+      telemetry_message,
+      MAX_RAISE_MESSAGE_SIZE,
+      "grab_gvl_and_raise called by thread holding the global VM lock: %%s (%s)",
+      rb_class2name(native_exception_class)
     );
```
ivoanjo, line 85: TBH the important part here is the `format_string`, the exception class is not particularly interesting :eyes:

ext/datadog_profiling_native_extension/ruby_helpers.c
```diff
@@ -75,11 +125,28 @@ void grab_gvl_and_raise_syserr(int syserr_errno, const char *format_string, ...)
...
+
+    // Then add the original exception message, to create the complete new message.
+    char exception_message[MAX_RAISE_MESSAGE_SIZE];
+    snprintf(exception_message, MAX_RAISE_MESSAGE_SIZE, telemetry_message, args.exception_message);
+
+    _raise_native_error(
+      eNativeRuntimeError,
+      exception_message,
+      rb_str_new_cstr(telemetry_message)
     );
```
ivoanjo: I'm thinking it'd be nice to keep the original `format_string` as the exception message for telemetry: this way we'd be able to see what it was, rather than seeing a %%s and then needing to go hunting

sig/datadog/profiling.rbs
ivoanjo, line 1: I guess this is slightly outdated vs what's there in profiling.rb right now?

spec/datadog/profiling/collectors/code_provenance_spec.rb
ivoanjo, line 1: Oops, I guess my tests slightly broke on macOS, thanks for fixing this! :bow:

ivoanjo (commented):
Thanks for looking into this! Did a pass :)

ext/datadog_profiling_native_extension/setup_signal_handler.c
marcotc: Yeah, this was like a the last call site I had to "fix" and I think I just tried to fit it in the existing methods.
I'll see I can try the enum so I don't bastardize it.

ext/datadog_profiling_native_extension/ruby_helpers.c
marcotc: Very reasonable! I had it as a rb_bug before changing it to raise much later, and I agree that we can just raise it.

ext/datadog_profiling_native_extension/ruby_helpers.c
marcotc: Thank you, I'll try that.
I kept changing it from VALUE to char* during development, and I think I missed a last pass.

ext/datadog_profiling_native_extension/ruby_helpers.c
marcotc: I'll do that! I noticed that when working with our "`_` methods for testing only"

ext/datadog_profiling_native_extension/ruby_helpers.c
marcotc: Why not both 😄? You are right, let me add that.

ext/datadog_profiling_native_extension/ruby_helpers.c
marcotc: Makes sense!

sig/datadog/profiling.rbs
marcotc: I didn't do all the janitorial work 😛. I'll keep this comment open as a reminder.

ext/datadog_profiling_native_extension/profiling.c
marcotc: 👍

ext/datadog_profiling_native_extension/private_vm_api_access.c
marcotc: I'll leave these unchanged for the time being.

ext/datadog_profiling_native_extension/setup_signal_handler.c
marcotc: 👍

ext/datadog_profiling_native_extension/setup_signal_handler.c
marcotc: I added a strict compile time + runtime check for the list of handler names.

ext/libdatadog_api/crashtracker.c
marcotc: 👍

ext/datadog_profiling_native_extension/ruby_helpers.c
```diff
@@ -11,43 +11,141 @@ static VALUE module_object_space = Qnil;
 static ID _id2ref_id = Qnil;
 static ID inspect_id = Qnil;
 static ID to_s_id = Qnil;
+static ID new_id = 0;
+static ID telemetry_message_id = Qnil;
+
+void ruby_helpers_init(VALUE profiling_module) {
```
ivoanjo: I think the `profiling_module` argument is a leftover of a previous version -- I don't see it being used? 👀

ext/datadog_profiling_native_extension/datadog_ruby_common.c
```diff
@@ -78,3 +92,16 @@ ddog_Vec_Tag convert_tags(VALUE tags_as_array) {
...
+  VALUE core_module = rb_const_get(datadog_module, rb_intern("Core"));
+  VALUE native_module = rb_const_get(core_module, rb_intern("Native"));
+
+  rb_global_variable(&eNativeRuntimeError);
+  rb_global_variable(&eNativeArgumentError);
+  rb_global_variable(&eNativeTypeError);
+
+  eNativeRuntimeError = rb_const_get(native_module, rb_intern("RuntimeError"));
+  eNativeArgumentError = rb_const_get(native_module, rb_intern("ArgumentError"));
+  eNativeTypeError = rb_const_get(native_module, rb_intern("TypeError"));
```
ivoanjo: Minor: Maybe sprinkle `ENFORCE_TYPE(..., T_MODULE/T_CLASS)` or the similar after every read from Ruby -- this way we get a nice error message if something is `nil` rather than later getting a confusing `NoMethodError` or similar

ext/datadog_profiling_native_extension/ruby_helpers.c
```diff
@@ -11,43 +11,141 @@ static VALUE module_object_space = Qnil;
...
+  VALUE datadog_module = rb_const_get(rb_cObject, rb_intern("Datadog"));
+  datadog_ruby_common_init(datadog_module);
```
ivoanjo: Minor: I had suggested adding the init here but on second thought it may be cleaner to call it directly in profiling.c -- that way `profiling.c` calls all of the inits rather than having inits that call other inits. 🤔

ext/datadog_profiling_native_extension/ruby_helpers.c
```diff
@@ -58,29 +156,61 @@ void grab_gvl_and_raise(VALUE exception_class, const char *format_string, ...) {
...
       syserr_errno,
       args.exception_message
     );
+
+    private_raise_native_error(
+      eNativeRuntimeError,
+      exception_message,
+      telemetry_message
+    );
   }
```
ivoanjo, line 95: Minor: Previously this part was duplicated between `grab_gvl_and_raise_syserr` and `grab_gvl_and_raise` but there wasn't a lot of logic... Now that there's a few more lines I think it may be worth extracting this into a separate method?

ext/datadog_profiling_native_extension/setup_signal_handler.h
```diff
@@ -2,9 +2,12 @@
 
 #include <signal.h>
 #include "datadog_ruby_common.h"
+#include "ruby_helpers.h"
 
 void empty_signal_handler(DDTRACE_UNUSED int _signal, DDTRACE_UNUSED siginfo_t *_info, DDTRACE_UNUSED void *_ucontext);
-void install_sigprof_signal_handler(void (*signal_handler_function)(int, siginfo_t *, void *), const char *handler_pretty_name);
+#define install_sigprof_signal_handler(signal_handler_function) \
+  private_install_sigprof_signal_handler(signal_handler_function, SIGNAL_HANDLER_NAME(signal_handler_function))
+void private_install_sigprof_signal_handler(void (*signal_handler_function)(int, siginfo_t *, void *), signal_handler_name_t handler_pretty_name);
```
ivoanjo: The magic is cool but... To be honest in this case I'd go for explicitly passing in the `signal_handler_name_t` entry (even if it's a bit verbose) over having the fancier version (C being already confusing as it is...)

ext/datadog_profiling_native_extension/ruby_helpers.h
```diff
@@ -67,6 +103,25 @@ NORETURN(void raise_syserr(
...
+
+static const char *const SIGNAL_HANDLER_NAMES[SIGNAL_HANDLER_NAME_COUNT] = {
+  [SIGNAL_HANDLER_NAME_NULL] = "", // Unused
+  [SIGNAL_HANDLER_NAME_handle_sampling_signal] = "handle_sampling_signal",
+  [SIGNAL_HANDLER_NAME_testing_signal_handler] = "testing_signal_handler",
+  [SIGNAL_HANDLER_NAME_holding_the_gvl_signal_handler] = "holding_the_gvl_signal_handler",
+  [SIGNAL_HANDLER_NAME_empty_signal_handler] = "empty_signal_handler"
+};
+
+#define SIGNAL_HANDLER_NAME(var) SIGNAL_HANDLER_NAME_##var
```
ivoanjo: This should probably live in `setup_signal_handler.h, since it exists only for those APIs.

ext/datadog_profiling_native_extension/setup_signal_handler.c
```diff
@@ -96,7 +101,13 @@ static inline void toggle_sigprof_signal_handler_for_current_thread(int action)
...
   sigaddset(&signals_to_toggle, SIGPROF);
   int error = pthread_sigmask(action, &signals_to_toggle, NULL);
-  if (error) rb_exc_raise(rb_syserr_new_str(error, rb_sprintf("Unexpected failure in pthread_sigmask, action=%d", action)));
+
+  if (error) {
+    // Ensure we know statically the possible values for `action`.
+    int telemetry_safe_action = (action >= SIG_BLOCK && action <= SIG_UNBLOCK) ? action : -1;
+
+    raise_telemetry_safe_syserr(error, "Unexpected failure in pthread_sigmask, action=%d", telemetry_safe_action);
+  }
```
ivoanjo, line 105: Minor: Since action is an argument for `pthread_sigmask`... Do we really need this extra validation? I don't see how this ever could have user-level data...

ext/libdatadog_api/crashtracker.c
```diff
@@ -12,24 +14,36 @@ static bool first_init = true;
...
+  rb_define_singleton_method(crashtracker_class,
+                             "_native_start_or_update_on_fork",
+                             _native_start_or_update_on_fork, -1);
+  rb_define_singleton_method(crashtracker_class, "_native_stop", _native_stop,
+                             0);
 }
 
-static VALUE _native_start_or_update_on_fork(int argc, VALUE *argv, DDTRACE_UNUSED VALUE _self) {
+static VALUE _native_start_or_update_on_fork(int argc, VALUE *argv,
+                                             DDTRACE_UNUSED VALUE _self) {
```
ivoanjo: Minor: Since this PR already has quite a few changes, I'd suggest reducing diff noise from reformatting as much as possible -- we can reformat on a separate "brainless-just-approve-it" PR :)

ext/libdatadog_api/helpers.h
```diff
@@ -0,0 +1,25 @@
...
+
+// Raises a Ruby error for the error result.
+// The error message in `result.err` is appended to the provided `message`.
+//
+// @param[in] message (const char *) The error message
+// @param[in] result (struct { ddog_Error res; ... }) Any type of result
+#define raise_lib_error(message, result)                                        \
+  do {                                                                         \
+    raise_error(eNativeRuntimeError, message ": %" PRIsVALUE, get_error_details_and_drop(&result.err)); \
+  } while (0)
```
ivoanjo, line 27: This is cool! 👍

lib/datadog/core/telemetry/logging.rb
```diff
@@ -65,6 +71,20 @@ def error(description)
 
           log!(event)
         end
+
+        private
+
+        # A constant string representing the exception
+        def message_for_telemety(exception)
```
ivoanjo: Minor: Typo -- `telemety`

spec/datadog/core/telemetry/logging_spec.rb
```diff
@@ -82,6 +82,126 @@ def log!(_event)
         end
       end
     end
+    context 'with an Erro error carrying telemetry message' do
```
ivoanjo: Minor: "Erro error"? 🤣 

```suggestion
    context 'with an Errno error carrying telemetry message' do
```

spec/datadog/core/telemetry/logging_spec.rb
```diff
@@ -82,6 +82,126 @@ def log!(_event)
...
+      context 'with description' do
+        let(:description) { 'Operation failed' }
+
+        it 'includes both description and telemetry message' do
+          expect(component).to receive(:log!).with(instance_of(Datadog::Core::Telemetry::Event::Log)) do |event|
+            expect(event.payload).to include(
+              logs: [{message: 'Errno::ENOENT: Operation failed (Safe message)', level: 'ERROR', count: 1,
+                      stack_trace: a_string_including('REDACTED')}]
+            )
+          end
```
ivoanjo, line 119: Wait, should the assertion include the `description`?

spec/datadog/core/telemetry/logging_spec.rb
```diff
@@ -82,6 +82,126 @@ def log!(_event)
...
+
+          report
+        end
+      end
+    end
+    context 'with NativeError' do
+      before do
+        skip unless defined?(Datadog::Core::Native::Error)
+      end
```
ivoanjo: Minor: Since these classes are quite simple, I'd suggest always loading them (and assuming they are loaded), rather than having to worry about loading them only sometimes... 👀

spec/datadog/core/telemetry/logging_spec.rb
```diff
@@ -82,6 +82,126 @@ def log!(_event)
...
+      it 'includes the telemetry-safe message in telemetry' do
+        expect(component).to receive(:log!).with(instance_of(Datadog::Core::Telemetry::Event::Log)) do |event|
+          expect(event.payload).to include(
+            logs: [{message: 'Datadog::Core::Native::RuntimeError: (This is a safe profiler error)', level: 'ERROR', count: 1,
+                    stack_trace: a_string_including('REDACTED')}]
+          )
+        end
+        begin
+          raise Datadog::Core::Native::RuntimeError.new('This is a safe profiler error', 'This is a safe profiler error')
```
ivoanjo: Should we use different messages here to make sure the right one gets used?

spec/datadog/core/telemetry/logging_spec.rb
```diff
@@ -82,6 +82,126 @@ def log!(_event)
...
+        end
+      end
+      context 'with description' do
+        it 'includes both description and telemetry message' do
+          expect(component).to receive(:log!).with(instance_of(Datadog::Core::Telemetry::Event::Log)) do |event|
+            expect(event.payload).to include(
+              logs: [{message: 'Datadog::Core::Native::RuntimeError: Profiler failed to start (Failed to initialize native extension)', level: 'ERROR', count: 1,
+                      stack_trace: a_string_including('REDACTED')}]
+            )
+          end
```
ivoanjo, line 150: Should we assert on the `description`? (Similar to above?)

spec/datadog/core/telemetry/logging_spec.rb
```diff
@@ -82,6 +82,126 @@ def log!(_event)
...
+        it 'omits the dynamic exception message from telemetry' do
+          expect(component).to receive(:log!).with(instance_of(Datadog::Core::Telemetry::Event::Log)) do |event|
+            expect(event.payload).to include(
+              logs: [{message: 'Datadog::Core::Native::RuntimeError', level: 'ERROR', count: 1,
+                      stack_trace: a_string_including('REDACTED')}]
+            )
+            expect(event.payload[:logs].map { |log| log[:message] }).not_to include(/Failed to initialize.*0x[0-9a-f]+/)
+          end
+          begin
+            raise Datadog::Core::Native::RuntimeError, 'Failed to initialize string storage: Error at address 0xdeadbeef'
```
ivoanjo: Minor: I'd suggest simplifying the message and the assertion -- that regexp in particular seems a bit too complex for what we're testing

spec/datadog/core/telemetry/logging_spec.rb
```diff
@@ -82,6 +82,126 @@ def log!(_event)
...
+          expect(component).to receive(:log!).with(instance_of(Datadog::Core::Telemetry::Event::Log)) do |event|
+            expect(event.payload).to include(
+              logs: [{message: 'Datadog::Core::Native::RuntimeError: (Static message)', level: 'ERROR', count: 1,
+                      stack_trace: a_string_including('REDACTED')}]
+            )
+            expect(event.payload[:logs].map { |log| log[:message] }).not_to include('Dynamic info 0xabc123')
+          end
+          begin
+            raise Datadog::Core::Native::RuntimeError.new('Dynamic info 0xabc123', 'Static message')
+          rescue => e
```
ivoanjo: This seems to be testing the same as `it 'includes the telemetry-safe message in telemetry'` or am I missing something? 🤔 

I guess `description` is missing in `report`... maybe?

spec/datadog/profiling/spec_helper.rb
ivoanjo, line 1: Don't forget to discard this change ;)

spec/support/core_helpers.rb
```diff
@@ -62,6 +62,95 @@ def supports_value_expectations?
     end
   end
 
+  RSpec::Matchers.define :raise_native_error do |expected_class, expected_message = nil, expected_telemetry_message = nil, &block|
```
ivoanjo: Hmmmmm I'm not entirely convinced this helper is worth the complexity it introduces... E.g. we already specify which class is expected when raising (the native class variant) so this seems to somewhat duplicate the testing on the helpers (e.g. the ones in `telemetry/logging_spec.rb`) and I'm not sure we're getting a lot out of that?

Consider maybe removing? 🤔

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
```diff
@@ -199,7 +200,9 @@
 
         exception = try_wait_until(backoff: 0.01) { cpu_and_wall_time_worker.send(:failure_exception) }
 
-        expect(exception.message).to include "pre-existing SIGPROF"
+        expect(exception.message).to include "(empty_signal_handler): There's a pre-existing SIGPROF"
+        # The handler name and signal name are statically check as safe for telemetry
+        expect(exception.telemetry_message).to include "(empty_signal_handler): There's a pre-existing SIGPROF"
       end
```
ivoanjo, line 203: If possible, I'd like to avoid the additional testing of the telemetry message -- the `telemetry/logging_spec.rb` should be the one responsible for testing that.

(This applies to other specs below)

spec/datadog/profiling/collectors/thread_context_spec.rb
```diff
@@ -1968,6 +1974,18 @@ def sample_and_check(expected_state:)
...
+      before do
+        skip "GVL profiling timeline crashes on macOS" if PlatformHelpers.mac?
+      end
+      it "raises a telemetry-instrumented error" do
+        expect { sample_after_gvl_running(t1) }.to raise_error(Datadog::Core::Native::RuntimeError) do |error|
+          expect(error.message).to include("GVL profiling requires timeline to be enabled")
+          expect(error.telemetry_message).to eq("GVL profiling requires timeline to be enabled")
+        end
+      end
+    end
```
ivoanjo, line 1979: Wait, but if GVL profiling is going to raise, why does it still crash on macOS... 👀 

(Not a blocker, but if you could share the Ruby stack info when the crash happens I'm really curious what's going on here)

ivoanjo (commented):
Gave it another pass!

ext/libdatadog_api/crashtracker.c
marcotc: Argg, I kept fighting my editor, because it kept reformatting the file. Sorry this one slipped, I'll do another pass to check for unnecessary changes like this.

ext/libdatadog_api/helpers.h
marcotc: 🙇

ext/datadog_profiling_native_extension/ruby_helpers.c
marcotc: Done!

spec/datadog/core/telemetry/logging_spec.rb
marcotc: Yes, the `description` is provided by us in the method call to `Telemetry#report`.
I'll document the method accordingly.

The messages provided to `description` were assessed as all safe before we started the work on https://github.com/DataDog/dd-trace-rb/pull/4985.

spec/datadog/core/telemetry/logging_spec.rb
marcotc: I've changed it to assert that the dynamic exception message is not present. I'm assuming that's what you meant, since the description is already being matched in `message: 'Datadog::Core::Native::RuntimeError: Profiler failed to start (Failed to initialize native extension)'` (the description is `Profiler failed to start`).

spec/datadog/core/telemetry/logging_spec.rb
marcotc: I simplified pretty much all of them.

spec/datadog/core/telemetry/logging_spec.rb
marcotc: You are right!

spec/datadog/profiling/collectors/cpu_and_wall_time_worker_spec.rb
marcotc: 👍

spec/datadog/profiling/collectors/thread_context_spec.rb
marcotc: I'm looking into this one closer: I want to make sure it's not a real crash.

It's tripping our "[BUG] Datadog Ruby profiler detected callback nested inside sample" error.

ext/datadog_profiling_native_extension/ruby_helpers.c
marcotc: The `datadog_ruby_common_init` initializers will also have to be called from libdatadog's init method, thus why I created the shared `datadog_ruby_common_init` method.

Do you think we want to duplicate the logic in `datadog_ruby_common_init` in both the profiler and libdatadog? There's no difference in the code: it will be copy/paste (and must be kept in sync).

ext/libdatadog_api/crashtracker.c
ivoanjo: To be fair I've been super-lazy in adding a formatter since well... I'm usually writing most of the C myself and so it's easy to keep it consistent and matching what I like in that case ;) 

So I'm not against adding a formatter as "Ivo writes all the C" is becoming less and less the case ;)

spec/datadog/core/telemetry/logging_spec.rb
ivoanjo: Ah, right! I think I misread part of this test >_>

spec/datadog/core/telemetry/logging_spec.rb
ivoanjo: Looks good!

ext/datadog_profiling_native_extension/ruby_helpers.c
ivoanjo: Ah, wait, no!

I think `datadog_ruby_common_init` makes sense! What I mean is, I'm thinking we should "flatten" the `_init` calls -- e.g. call them all from `profiling.c` (profiling extension) / `init.c` (libdatadog extension) rather than doing as we're doing here `profiling.c -> ruby_helpers_init -> datadog_ruby_common_init`.

My thinking is -- initialization of the native extension is a very sensitive time (e.g. if we want to raise exceptions, we need to make sure exceptions are initialized first, etc) so having an explicit clear order for what's happening and when will hopefully help us make sure things are kept in a reasonable order vs it being a bit more abstracted away.

spec/datadog/profiling/collectors/thread_context_spec.rb
ivoanjo: Thanks Marco for sharing this in slack! I've copy-pasted my answer here as we may want to refer to it in the future.

There's a few of this situations where the profiler gets called from a context where we don't want to lose the GVL in the middle of what we're doing. Sometimes we can't (not safe for the VM), others it simplifies a lot of the logic -- it means some part of the profiler runs "atomically" -- we don't have to consider "what if we lose the GVL here and some other part of the profiler tries to trigger again, etc."

In https://github.com/DataDog/dd-trace-rb/pull/4240 I've added the `debug_enter_unsafe_context` which is used on our tests to detect when this might happen -- I've sprinkled it on APIs where we want to be able to assume that this is the case.

By changing the exception raising code to call the Ruby-level constructor, this PR actually introduces one such "can lose the GVL here" step where previously there was none -- before all the exception raising code was C and it ran beginning to end without there being possibility of thread switches. Going into Ruby code for running `Datadog::Core::Native::Error#initialize` means the VM can decide to go "oh yeah this is a good time to release the GVL and give it to another thread!".

Yes this is slightly brittle in that nobody promised us which APIs can release the GVL or not, but hey 1) that's why we have a detector for it rather than just praying 2) welcome to the profiling C extension ;)

(The way `debug_enter_unsafe_context` works is that it sets up a postponed job and then checks if it ran or not before the end of the unsafe section. If it does run, it means Ruby checked for interruptions, and Ruby checks for interruptions as part of releasing/acquiring the GVL).

I had not thought of the ramifications of that part of the change here...

That said, in this specific situation in this specific test raising and introducing a GVL change point would probably not be an issue. But I think in the general case for all of the profiler it's probably nice if we keep this property of "raising an exception doesn't introduce a GVL change-point".

To do that, my suggestion is removing the `rb_funcall` when initializing the exception and instead doing it like we're doing for syserr in `ruby_helpers.c`: allocate the object using C code and then use `rb_ivar_set` to add the telemetry message. Looking at Ruby's `rb_raise` implementation, we should be able to use `rb_exc_new_str` to still create an instance of our class, then set the ivar, then raising.

ext/datadog_profiling_native_extension/ruby_helpers.c
marcotc: Ahh, that makes sense! Will do!

spec/datadog/profiling/collectors/thread_context_spec.rb
marcotc: OK, so this new test triggers our `check_for_unsafe_api_calls` check even on master, running on Linux (Ruby 3.4):
```ruby
    context "when timeline is disabled" do
      let(:timeline_enabled) { false }
      it "raises a telemetry-instrumented error" do
        expect { sample_after_gvl_running(t1) }.to raise_error(RuntimeError, "GVL profiling requires timeline to be enabled")
      end
    end
```
```
/usr/local/lib/libruby.so.3.4(rb_bug) /usr/src/ruby/error.c:1115
/app/lib/datadog_profiling_native_extension.3.4.7_aarch64-linux.so(check_for_unsafe_api_calls+0x24) [0xffff8e6c7604] ../../../../ext/datadog_profiling_native_extension/unsafe_api_calls_check.c:44
/app/lib/datadog_profiling_native_extension.3.4.7_aarch64-linux.so(unsafe_api_calls_check_init) (null):0
/usr/local/lib/libruby.so.3.4(rb_postponed_job_flush+0x160) [0xffffac33f720] /usr/src/ruby/vm_trace.c:1852
/usr/local/lib/libruby.so.3.4(rb_threadptr_execute_interrupts+0x2d4) [0xffffac2c9700] /usr/src/ruby/thread.c:2483
/usr/local/lib/libruby.so.3.4(rb_vm_check_ints+0xc) [0xffffac32812c] /usr/src/ruby/vm_core.h:2132
/usr/local/lib/libruby.so.3.4(rb_vm_check_ints) /usr/src/ruby/vm_core.h:2123
/usr/local/lib/libruby.so.3.4(vm_pop_frame) /usr/src/ruby/vm_insnhelper.c:456
/usr/local/lib/libruby.so.3.4(rb_vm_pop_frame) /usr/src/ruby/vm_insnhelper.c:465
/usr/local/lib/libruby.so.3.4(vm_call0_cfunc_with_frame) /usr/src/ruby/vm_eval.c:167
/usr/local/lib/libruby.so.3.4(vm_call0_cfunc) /usr/src/ruby/vm_eval.c:178
/usr/local/lib/libruby.so.3.4(vm_call0_body) /usr/src/ruby/vm_eval.c:229
/usr/local/lib/libruby.so.3.4(rb_call0+0x2e8) [0xffffac32b6c8] /usr/src/ruby/vm_eval.c:101
/usr/local/lib/libruby.so.3.4(rb_class_new_instance_kw+0xec) [0xffffac1cf27c] /usr/src/ruby/object.c:2191
/usr/local/lib/libruby.so.3.4(rb_exc_new_str+0x44) [0xffffac1244ec] /usr/src/ruby/error.c:1485
/usr/local/lib/libruby.so.3.4(rb_raise+0x8c) [0xffffac12571c] /usr/src/ruby/error.c:3768
/app/lib/datadog_profiling_native_extension.3.4.7_aarch64-linux.so(thread_context_collector_sample_after_gvl_running+0x18) [0xffff8e6bf85c] ../../../../ext/datadog_profiling_native_extension/collectors_thread_context.c:1964
```
From roughly [this Ruby call](https://github.com/ruby/ruby/blob/a8b49ab48703dfb8862ca28a443da0e764d692c6/eval.c#L1874).

My thinking so far is two fold:
1. The behaviour has not changed with this PR.
2. Should we call `debug_leave_unsafe_context()` right before raising an error? We know we'll do something that can acquire the GVL, might as well declare that? And the current task is exiting as well, due to the error.

marcotc: What do you think of this solution: https://github.com/DataDog/dd-trace-rb/pull/5076/commits/58cf67e46d06db5f2bad7ca0588bbec4a213f1ed

ext/datadog_profiling_native_extension/setup_signal_handler.h
```diff
@@ -3,8 +3,14 @@
...
 
+typedef struct {
+  void (*function)(int, siginfo_t *, void *);
+  const char *name;
+} signal_handler_t;
+
 void empty_signal_handler(DDTRACE_UNUSED int _signal, DDTRACE_UNUSED siginfo_t *_info, DDTRACE_UNUSED void *_ucontext);
-void install_sigprof_signal_handler(void (*signal_handler_function)(int, siginfo_t *, void *), const char *handler_pretty_name);
+
+void install_sigprof_signal_handler(const signal_handler_t *signal_handler);
```
ivoanjo: This looks like a weird end-state of a few experiments -- e.g. at this point we're not back to the original code, but we're functionally equivalent to it, so I would suggest going back to the original signature as I think they're basically the same?

ext/datadog_profiling_native_extension/collectors_thread_context.c
```diff
@@ -2131,18 +2132,22 @@ static uint64_t otel_span_id_to_uint(VALUE otel_span_id) {
...
   }
 
-  static VALUE _native_sample_after_gvl_running(DDTRACE_UNUSED VALUE self, VALUE collector_instance, VALUE thread) {
+  static VALUE _native_sample_after_gvl_running(DDTRACE_UNUSED VALUE self, VALUE collector_instance, VALUE thread, VALUE allow_exception) {
     ENFORCE_THREAD(thread);
+    ENFORCE_BOOLEAN(allow_exception);
 
-    debug_enter_unsafe_context();
+    printf("ThreadContextCollector#_native_sample_after_gvl_running called\n");
+    printf("allow_exception: %d\n", allow_exception == Qtrue ? 1 : 0);
```
ivoanjo: Minor: Don't forget to remove these ;)

ext/datadog_profiling_native_extension/datadog_ruby_common.h
```diff
@@ -32,6 +35,14 @@
 
 NORETURN(void raise_unexpected_type(VALUE value, const char *value_name, const char *type_name, const char *file, int line, const char* function_name));
 
+// Helper to raise errors with formatted messages
+NORETURN(void raise_error(VALUE error_class, const char *fmt, ...)) __attribute__ ((format (gnu_printf, 2, 3)));
+
+// Exception classes defined in Ruby, in the `Datadog::Core` namespace.
+extern VALUE eNativeRuntimeError;
+extern VALUE eNativeArgumentError;
+extern VALUE eNativeTypeError;
```
ivoanjo: Minor: I wonder if we should call them `eDatadogRuntimeError` vs native?

ext/datadog_profiling_native_extension/profiling.c
```diff
@@ -55,6 +57,7 @@ void DDTRACE_EXPORT Init_datadog_profiling_native_extension(void) {
   rb_define_singleton_method(native_extension_module, "native_working?", native_working_p, 0);
   rb_funcall(native_extension_module, rb_intern("private_class_method"), 1, ID2SYM(rb_intern("native_working?")));
 
+  datadog_ruby_common_init(datadog_module);
```
ivoanjo: ❗ I spotted that `native_working_p` may raise exceptions so we need to initialize the helpers before calling into it.

(TBH I hesitated in perhaps moving the call to `native_working_p` after **all** initialization, but I kinda like this property of "fusing off" the native gem behavior if the health check fails) 

I tried to suggest a change but the new github UX was buggy so here's the block without "suggestion" lol:

```c
VALUE datadog_module = rb_define_module("Datadog");

// MUST be called before all other initialization
datadog_ruby_common_init(datadog_module);

VALUE profiling_module = rb_define_module_under(datadog_module, "Profiling");
VALUE native_extension_module = rb_define_module_under(profiling_module, "NativeExtension");

rb_define_singleton_method(native_extension_module, "native_working?", native_working_p, 0);
rb_funcall(native_extension_module, rb_intern("private_class_method"), 1, ID2SYM(rb_intern("native_working?")));

```

ext/datadog_profiling_native_extension/ruby_helpers.c
```diff
@@ -12,82 +12,167 @@ static ID _id2ref_id = Qnil;
...
+        rb_sprintf(fmt, rb_class2name(native_exception_class) ?: "(Unknown)")
+      );
+      rb_ivar_set(exception, telemetry_message_id, rb_str_new_cstr(fmt));
+      rb_exc_raise(exception);
+  }
+
+  VALUE exception = rb_exc_new_cstr(native_exception_class, detailed_message);
+  rb_ivar_set(exception, telemetry_message_id, rb_str_new_cstr(static_message));
+  rb_exc_raise(exception);
+}
```
ivoanjo: Minor: Looks to me we can share a bit more code between both branches?

E.g....

```
VALUE exception = 
   (known exception...) ?
       rb_exc_new_cstr(....) :
       rb_exc_new_str(...)
rb_ivar_set
rb_exc_raise
```

ext/datadog_profiling_native_extension/ruby_helpers.c
```diff
@@ -12,82 +12,167 @@ static ID _id2ref_id = Qnil;
...
+  private_raise_native_error(native_exception_class, telemetry_safe_message, telemetry_safe_message);
 }
 
-typedef struct {
-  int syserr_errno;
-  char exception_message[MAX_RAISE_MESSAGE_SIZE];
-} syserr_raise_args;
+// The message must be statically bound and checked.
+void raise_telemetry_safe_syserr(int syserr_errno, const char *telemetry_safe_format, ...) {
+  FORMAT_VA_ERROR_MESSAGE(telemetry_safe_message, telemetry_safe_format);
```
ivoanjo: TBH These two seem to exist only for setup_signal_handler.c so I'm not sure they're worth the extra complexity.

At this point... I think we should just give up on having that info in telemetry. (This would allow inlining private_raise_syserr into trigger_raise as well which seems a nice simplification too)

ext/datadog_profiling_native_extension/ruby_helpers.c
```diff
@@ -12,82 +12,167 @@ static ID _id2ref_id = Qnil;
...
 
-  rb_bug("[ddtrace] Unexpected: Reached the end of grab_gvl_and_raise_syserr while raising '%s'\n", args.exception_message);
+  rb_bug("[ddtrace] Unexpected: Reached the end of grab_gvl_and_raise while raising '%s'\n", args.exception_message);
 }
 
+typedef struct {
+  int syserr_errno;
+  char exception_message[MAX_RAISE_MESSAGE_SIZE];
+  char telemetry_message[MAX_RAISE_MESSAGE_SIZE];
+} syserr_raise_args;
```
ivoanjo: This doesn't seem to be needed anymore?

ext/datadog_profiling_native_extension/ruby_helpers.c
```diff
@@ -12,82 +12,167 @@ static ID _id2ref_id = Qnil;
...
+    args.exception_class = Qnil;
+    args.syserr_errno = syserr_errno;
+
+    snprintf(short_error_type, MAX_RAISE_MESSAGE_SIZE, "Errno %d", syserr_errno);
+  } else {
+    args.exception_class = native_exception_class;
+    args.syserr_errno = 0;
 
-  args.syserr_errno = syserr_errno;
+    snprintf(short_error_type, MAX_RAISE_MESSAGE_SIZE, "%s", rb_class2name(native_exception_class) ?: "(Unknown)");
```
ivoanjo: I don't think this `rb_class2name` is safe here -- we're executing without the GVL.

We could test all 3 known exception classes with `==` and have a hardcoded `char *` for them, BUT to be honest this is a corner case of a corner case of a corner case so I'd suggest simplifying instead and removing `short_error_type`.

ext/datadog_profiling_native_extension/ruby_helpers.h
```diff
@@ -39,15 +40,47 @@ static inline int check_if_pending_exception(void) {
 
 #define VALUE_COUNT(array) (sizeof(array) / sizeof(VALUE))
 
+// Raises an exception of the specified class with the formatted string as its message.
+// This macro ensures that the literay string is sent for telemetry, while the formatted
```
ivoanjo: Minor:

```suggestion
// This macro ensures that the literal string is sent for telemetry, while the formatted
```

ext/datadog_profiling_native_extension/setup_signal_handler.c
```diff
@@ -96,7 +92,9 @@ static inline void toggle_sigprof_signal_handler_for_current_thread(int action)
   sigemptyset(&signals_to_toggle);
   sigaddset(&signals_to_toggle, SIGPROF);
   int error = pthread_sigmask(action, &signals_to_toggle, NULL);
-  if (error) rb_exc_raise(rb_syserr_new_str(error, rb_sprintf("Unexpected failure in pthread_sigmask, action=%d", action)));
+
+  // `action` is telemetry-safe as it is only provided the caller methods in this file.
+  if (error) raise_telemetry_safe_syserr(error, "Unexpected failure in pthread_sigmask, action=%d", action);
```
ivoanjo: Minor: Rather than using `raise_telemetry_safe_syserr` for this one, we could have an if of the action (I think it's fine to assume it's always going to be SIG_BLOCK or SIG_UNBLOCK), and have one literal error message for each.

lib/datadog/core/native/error.rb
```diff
@@ -0,0 +1,32 @@
...
+
+module Datadog
+  module Core
+    module Native
+      # Base error type for exceptions raised by our native extensions.
+      # These errors have both the original error message and a telemetry-safe message.
+      # The telemetry-safe message is statically defined and does not possess dynamic data.
+      #
+      # IMPORTANT: Native errors must not call Ruby methods during creation, to avoid
+      # releasing the GVL when raising errors in unsafe contexts (see `debug_enter_unsafe_context`).
```
ivoanjo: I guess we can remove this, since we weren't successful in preventing Ruby code from running in exceptions :/

lib/datadog/core/telemetry/logging.rb
```diff
@@ -65,6 +74,20 @@ def error(description)
...
+        def message_for_telemetry(exception)
+          if exception.respond_to?(:telemetry_message)
+            exception.telemetry_message
+          # For the 150+ `Errno` error classes, we set the telemetry message as an instance variable.
+          # We don't want to create/modify such a large number of classes for a small feature.
+          # Note: Errno classes are subclasses of SystemCallError.
+          elsif exception.is_a?(SystemCallError) && exception.instance_variable_defined?(:@telemetry_message)
+            exception.instance_variable_get(:@telemetry_message)
+          end
+        end
```
ivoanjo, line 85: Minor: I wonder if we should just get rid of the attr_acessor and simplify the code to always use the instance_variable_defined + instance_variable_get...

lib/datadog/core/native.rb
ivoanjo, line 1: I'm not sure this is very useful? E.g. declaring `Native` in `core/native/error` creates it anyway.

(E.g. we already have so many files and so many complexity overall in the project that I think anything we can reduce is a win)

spec/datadog/profiling/collectors/code_provenance_spec.rb
```diff
@@ -34,7 +36,10 @@
           version: Datadog::VERSION::STRING,
           paths: contain_exactly(
             start_with("/"),
-            include("extensions").and(include(RUBY_PLATFORM)),
+            satisfy do |path|
+              path.include?("extensions") &&
+                (path.include?(platform_fragment) || path.include?(hyphenated_platform_fragment))
+            end,
```
ivoanjo, line 44: Minor: I wonder if this could be extracted into a `let(:expected_platform)` that would allow this code to go back to `include("extensions").and(include(expected_platform))`?

spec/datadog/profiling/collectors/stack_spec.rb
```diff
@@ -825,30 +825,18 @@ def dummy_template.#{method_name}(ready_queue)
...
-      it "returns true" do
-        expect(described_class._native_filenames_available?).to be true
-      end
-    end
-
-    context "on non-linux", if: !PlatformHelpers.linux? do
-      it "returns false" do
-        expect(described_class._native_filenames_available?).to be false
-      end
+    it "returns true" do
```
ivoanjo: Minor: Let's leave here a hint about why we assume this is true

```suggestion
    it "returns true on linux and macOS" do
```

spec/datadog/profiling/collectors/thread_context_spec.rb
ivoanjo: Marking as resolved as we actually couldn't find any way to raise without introducing a switch point (my claims above were incorrect), but with `allow_exception` this works fine now.

ext/datadog_profiling_native_extension/datadog_ruby_common.h
marcotc: Good idea!

ext/datadog_profiling_native_extension/ruby_helpers.c
marcotc: Leaving this comment open, as a reminder for our discussion regarding simplifying these methods.

ivoanjo:
The failure in the benchmarks...

```
+ bundle exec ruby benchmarks/profiling_sample_gvl.rb
Current pid is 3296
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [x86_64-linux]
profiling - gvl benchmark samples
benchmarks/profiling_sample_gvl.rb:64:in `_native_sample_after_gvl_running': wrong number of arguments (given 3, expected 2) (ArgumentError)
        Datadog::Profiling::Collectors::ThreadContext::Testing._native_sample_after_gvl_running(@collector, @target_thread, false)
                                                                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
	from benchmarks/profiling_sample_gvl.rb:64:in `block (2 levels) in run_benchmark'
	from (eval at /usr/local/bundle/gems/benchmark-ips-2.14.0/lib/benchmark/ips/job/entry.rb:63):6:in `call_times'
	from /usr/local/bundle/gems/benchmark-ips-2.14.0/lib/benchmark/ips/job.rb:285:in `block in run_warmup'
	from /usr/local/bundle/gems/benchmark-ips-2.14.0/lib/benchmark/ips/job.rb:268:in `each'
	from /usr/local/bundle/gems/benchmark-ips-2.14.0/lib/benchmark/ips/job.rb:268:in `run_warmup'
	from /usr/local/bundle/gems/benchmark-ips-2.14.0/lib/benchmark/ips/job.rb:253:in `block in run'
```

is because the "use the test code from this branch with master" doesn't work if master isn't API-compatible with this branch.

TBH what I've done in the past is just disable the benchmark and re-enable it later, but it's kinda meh workaround. Suggestions welcome :sweat_smile:

ext/datadog_profiling_native_extension/datadog_ruby_common.c
```diff
@@ -78,3 +89,8 @@ ddog_Vec_Tag convert_tags(VALUE tags_as_array) {
 
   return tags;
 }
+
+void datadog_ruby_common_init(VALUE datadog_module) {
+  // No longer needed - using Ruby's built-in exception classes
+  (void)datadog_module;
+}
```
ivoanjo: I'd suggest removing this + its callsites; it's a bit of complexity (including comments in "profiling.c" saying this needs to be called first) and I'm not sure it's worth leaving around.

ext/datadog_profiling_native_extension/datadog_ruby_common.c
```diff
@@ -18,6 +21,14 @@ void raise_unexpected_type(VALUE value, const char *value_name, const char *type
   );
 }
 
+void raise_error(VALUE error_class, const char *fmt, ...) {
+  va_list args;
+  va_start(args, fmt);
+  VALUE message = rb_vsprintf(fmt, args);
+  va_end(args);
+  rb_raise(error_class, "%"PRIsVALUE, message);
+}
```
ivoanjo: Wait, I'm confused by this... Is this a leftover from an earlier version? 

That is, I thought the correct version of `raise_error` was the `#define raise_error` we have in `ruby_helpers.c` (the one that grabs the format string separately, etc); this one seems to just be a weird direct passthrough to `rb_raise`.

Am I missing something? 🤔 👀

spec/datadog/profiling/native_extension_spec.rb
```diff
@@ -12,26 +12,90 @@
...
+      it "raises a TypeError" do
+        expect { raise_native_type_error }.to raise_error(::TypeError) do |error|
+          expect(error.message).to eq("type error test")
+          expect(error.instance_variable_get(:@telemetry_message)).to eq("type error test")
+        end
+      end
+
+      it "is an instance of TypeError" do
+        expect { raise_native_type_error }.to raise_exception(TypeError)
       end
```
ivoanjo: Minor: The "is an instance of" tests can be removed, since the `raise_error(...)` validations in the "it raises ..." testcases already include that.

ext/datadog_profiling_native_extension/ruby_helpers.h
```diff
@@ -39,26 +40,49 @@ static inline int check_if_pending_exception(void) {
...
 #define ENFORCE_SUCCESS_HELPER(expression, have_gvl) \
-  { int result_syserr_errno = expression; if (RB_UNLIKELY(result_syserr_errno)) raise_syserr(result_syserr_errno, have_gvl, ADD_QUOTES(expression), __FILE__, __LINE__, __func__); }
+  { int result_syserr_errno = expression; if (RB_UNLIKELY(result_syserr_errno)) raise_enforce_syserr(result_syserr_errno, have_gvl, ADD_QUOTES(expression), __FILE__, __LINE__, __func__); }
 
 #define RUBY_NUM_OR_NIL(val, condition, conv) ((val condition) ? conv(val) : Qnil)
 #define RUBY_AVG_OR_NIL(total, count) ((count == 0) ? Qnil : DBL2NUM(((double) total) / count))
 
 // Called by ENFORCE_SUCCESS_HELPER; should not be used directly
-NORETURN(void raise_syserr(
+NORETURN(void raise_enforce_syserr(
```
ivoanjo: Minor: To match the `private` naming we used above, maybe these should become `private_raise_syserr`?

ext/datadog_profiling_native_extension/ruby_helpers.c
```diff
@@ -12,83 +12,130 @@ static ID _id2ref_id = Qnil;
...
-}
 
-void grab_gvl_and_raise(VALUE exception_class, const char *format_string, ...) {
-  raise_args args;
 
-  args.exception_class = exception_class;
+// Raises an exception with separate telemetry-safe and detailed messages.
+// Make sure to *not* invoke Ruby code as this function can run in unsafe contexts.
+// NOTE: Raising the exception acquires the GVL (unsafe), but it also aborts the current execution flow.
+// @see debug_enter_unsafe_context
```
ivoanjo:
```suggestion
// NOTE: Raising an exception always invokes Ruby code so it requires the GVL and is not compatible with "debug_enter_unsafe_context".
// @see debug_enter_unsafe_context
```

ext/datadog_profiling_native_extension/datadog_ruby_common.c
marcotc: Nice catch!
This is left over from before.

I delete it and fixed the references, which created a small cascade of method movements.

ext/datadog_profiling_native_extension/datadog_ruby_common.c
marcotc: This method should actually be initializing the `ID telemetry_message_id`.
The reason it's not is because I inadvertently reverted some of the usages of `raise_error` in the libdatadog folder.

Now that I added them back, we need to keep these init methods.

ext/datadog_profiling_native_extension/ruby_helpers.h
marcotc: The name `private_raise_syserr` already exists 😅, but I renamed it to  `private_raise_enforce_syserr`.

ext/datadog_profiling_native_extension/ruby_helpers.c
marcotc: Applied!

spec/datadog/profiling/native_extension_spec.rb
marcotc: Thank you! They are totally redundant!

ivoanjo:
> @[marcotc](https://github.com/marcotc) marcotc requested a review from [ivoanjo](https://github.com/ivoanjo) [2 weeks ago](https://github.com/DataDog/dd-trace-rb/pull/5076#event-21735052670)

Btw this one is still on my radar, catching up on a pile of things this week! :sweat_smile:

PR #5082, opened 2025-11-24
Disable Ruby 3.5 preview1 testing

lloeki:
**What does this PR do?**

Disable Ruby 3.5 preview1 testing

**Motivation:**

Crashes with simplecov are simply too frequent.

**Change log entry**

None

**Additional Notes:**

Testing will be attempted to be reintroduced at 4.0 preview2 level.

**How to test the change?**

CI should not have 3.5


ivoanjo (approved):
👍 LGTM

PR #5086, opened 2025-11-24
DEBUG-3558 DI: chunk snapshot payloads

p-datadog:
**What does this PR do?**
Implements chunking of dynamic instrumentation snapshot payloads to stay under the 10 MB intake limit.

**Motivation:**
Without chunking, if a probe (or multiple probes) produce many events, all of the events will be dropped by the intake.

**Change log entry**
Yes: dynamic instrumentation: fix sending large quantities of snapshots

**Additional Notes:**
There is also "snapshot pruning" which is supposed to be done to reduce the size of individual snapshot events below the 1 MB limit for a "log line". This is not in scope for this PR (or Ruby Live Debugger GA).

This PR ensures that if many events are generated by Live Debugger, they all get sent to the backend, but does not help with huge individual snapshots getting dropped.

This PR does however add diagnostics when snapshots are dropped due to excessive size.

**How to test the change?**
Unit tests added

### DIFF
--- a/lib/datadog/core/utils/array.rb
+++ b/lib/datadog/core/utils/array.rb
@@ -9,10 +9,14 @@ module Datadog
           if array.respond_to?(:filter_map)
             # DEV Supported since Ruby 2.7, saves an intermediate object creation
             array.filter_map(&block)
-          else
+          elsif array.is_a?(Enumerator::Lazy)
             # You would think that .compact would work here, but it does not:
             # the result of .map could be an Enumerator::Lazy instance which
             # does not implement #compact on Ruby 2.5/2.6.
+            array.map(&block).reject do |item|
+              item.nil?
+            end
+          else
             array.each_with_object([]) do |item, memo|
               new_item = block.call(item)
               memo.push(new_item) unless new_item.nil?


lib/datadog/di/transport/input.rb
```diff
@@ -24,6 +29,16 @@ def initialize(parcel, serialized_tags)
...
 
+          # The maximum chunk size that intake permits is 10 MB.
+          #
+          # Two bytes are for the [ and ] of JSON array syntax.
+          MAX_CHUNK_SIZE = 10 * 1024 * 1024 - 2
+
+          # Try to send smaller payloads to avoid large network requests.
+          # If a payload is larger than default chunk size but is under the
+          # max chunk size, it will still get sent out.
+          DEFAULT_CHUNK_SIZE = 2 * 1024 * 1024
```
Strech, line 49: Maybe we could add the size resolution here via `DEFAULT_CHUNK_SIZE_BYTES`?

lib/datadog/di/transport/input.rb
```diff
@@ -24,6 +29,16 @@ def initialize(parcel, serialized_tags)
         class Transport
           attr_reader :client, :apis, :default_api, :current_api_id, :logger
 
+          # The maximum chunk size that intake permits is 10 MB.
+          #
+          # Two bytes are for the [ and ] of JSON array syntax.
+          MAX_CHUNK_SIZE = 10 * 1024 * 1024 - 2
```
Strech: Maybe we could add the size resolution here via `MAX_CHUNK_SIZE_BYTES`?

lib/datadog/di/transport/input.rb
```diff
@@ -36,9 +51,39 @@ def current_api
...
+              encoder.encode(snapshot)
+            end
+
+            Datadog::Core::Chunker.chunk_by_size(
+              encoded_snapshots, DEFAULT_CHUNK_SIZE,
+            ).each do |chunk|
+              if chunk.length == 1 && chunk.first.length > MAX_CHUNK_SIZE
+                # Drop the chunk.
+                # TODO report via telemetry metric?
+                logger.debug { "di: dropping too big snapshot" }
```
Strech: Not sure, but could it be more than 1 chunk going beyond max size? Or what is this scenario we are special handling?

sig/datadog/di/transport/input.rbs
```diff
@@ -30,12 +30,16 @@ module Datadog
           attr_reader current_api_id: untyped
 
           attr_reader logger: untyped
+          MAX_CHUNK_SIZE: untyped
+          DEFAULT_CHUNK_SIZE: untyped
```
Strech:
```suggestion
          MAX_CHUNK_SIZE: ::Integer
          DEFAULT_CHUNK_SIZE: ::Integer
```

lib/datadog/di/transport/input.rb
p-datadog: The sizes are in bytes unless specified otherwise. This is a convention used in many places in dd-trace-rb.

p-datadog:
I got the limits wrong in the initial implementation of this PR, this has now been corrected.

lib/datadog/di/transport/input.rb
Strech: Good to know

lib/datadog/core/utils/array.rb
```diff
@@ -0,0 +1,24 @@
...
+          if array.respond_to?(:filter_map)
+            # DEV Supported since Ruby 2.7, saves an intermediate object creation
+            array.filter_map(&block)
+          else
+            # You would think that .compact would work here, but it does not:
+            # the result of .map could be an Enumerator::Lazy instance which
+            # does not implement #compact on Ruby 2.5/2.6.
+            array.map(&block).reject do |item|
+              item.nil?
+            end
```
Strech, line 18: This could be done in a single iteration with `each_with_object` (or we can shadow the `item` and reduce variables)

```suggestion
            array.each_with_object([]) do |item, memo|
              new_item = block.call(item)
              memo.push(new_item) unless new_item.nil?
            end
```

sig/datadog/core/utils/array.rbs
```diff
@@ -0,0 +1,9 @@
+module Datadog
+  module Core
+    module Utils
+      module Array
+        def self.filter_map: (::Array[any] array) { (any) -> any } -> ::Array[any]
```
Strech, line 5: to be honest I ❤️ it

lib/datadog/di/transport/input.rb
p-datadog: The behavior changed a bit after I adjusted the code for the correct limits.

There are two limits that are relevant: size of any one snapshot - 1 MB and the size of the batch - 5 MB.

Given these limits, the batch (with chunking) will never exceed its max size since it is always legal to send a batch of one snapshot and the snapshot is limited to 1 MB.

For individual snapshots, yes, multiple snapshots could exceed their size, and each one will be logged. The logging however is at debug level which means the customers won't see any of it (normally). If this logging becomes an issue in the future I can add some sort of throttling but for now I think it's not necessary to worry about too much log output.

lib/datadog/core/utils/array.rb
Strech: @p-datadog I think you have reverted suggestion, but if you want to cover lazy evaluation, you can add a simple check and keep the single iteration/single allocation branch


```ruby
elsif array.is_a?(Enumerator::Lazy)
  # your example for lazy enumerator
else
  # my suggestion each_with_object(...)
end
```

p-datadog: I made this change.

PR #5093, opened 2025-11-25
Fix flaky data_streams bucket aggregation test

ericfirth:
Fixes race condition in 'aggregates multiple checkpoints into DDSketch histograms' test.

Root Cause:
The test captured 'now = Time.now.to_f' at the start, but the checkpoints
created immediately after used their own Core::Utils::Time.now internally.
These timestamps could differ by milliseconds, causing checkpoints to be
bucketed into a different time window than the test expected.

When the test calculated the expected bucket:
  bucket_time_ns = now_ns - (now_ns % processor.bucket_size_ns)

And looked it up:
  bucket = processor.buckets[bucket_time_ns]

The bucket would be nil if the checkpoints used a slightly different
timestamp and landed in an adjacent bucket.

Solution:
Use Core::Utils::Time.now_provider to inject a fixed timestamp
(Time.utc(2000, 1, 1, 0, 0, 0)) for the entire test. This ensures
all checkpoints use the exact same deterministic timestamp,
eliminating the race condition completely.

The test logic remains unchanged - we're just controlling the time
source to make it deterministic rather than dependent on wall clock
timing.

Related to incident #46145
Seed that reproduced the flake: 61335

**What does this PR do?**

**Motivation:**

**Change log entry**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/spec/datadog/data_streams/processor_spec.rb
+++ b/spec/datadog/data_streams/processor_spec.rb
@@ -190,6 +190,7 @@ RSpec.describe Datadog::DataStreams::Processor do
         allow(Datadog::Core::Utils::Time).to receive(:now).and_return(frozen_time)
         allow(Datadog::Tracing).to receive(:active_span).and_return(nil)
 
+        # Stop background worker to prevent it from flushing buckets during manual inspection
         processor.stop(true)
 
         processor.set_produce_checkpoint(type: 'kafka', destination: 'topicA', manual_checkpoint: false)


p-datadog:
:+1: on the general approach but it looks like the tests in question are still intermittently failing with this PR.

Would be nice to have the time fixing code extracted into a helper but I can do that either in this PR or in a follow-up PR.

spec/datadog/data_streams/processor_spec.rb
```diff
@@ -186,47 +186,54 @@
...
-        now = Time.now.to_f
+        fixed_time = Time.utc(2000, 1, 1, 0, 0, 0)
+        allow(Datadog::Tracing).to receive(:active_span).and_return(nil)
 
-        # Create multiple checkpoints with the same tags to aggregate
-        processor.set_produce_checkpoint(type: 'kafka', destination: 'topicA', manual_checkpoint: false)
-        processor.set_produce_checkpoint(type: 'kafka', destination: 'topicA', manual_checkpoint: false)
-        processor.set_produce_checkpoint(type: 'kafka', destination: 'topicA', manual_checkpoint: false)
+        begin
+          Datadog::Core::Utils::Time.now_provider = -> { fixed_time }
```
Strech: Maybe a regular stub is better? Then you don't need to worry with ensure block

```ruby
allow(Datadog::Core::Utils::Time).to receive(:now_provider).and_return(-> { fixed_time })
```

p-datadog: :+1: if this works it looks like  a great way to do it

ericfirth:
> 👍 on the general approach but it looks like the tests in question are still intermittently failing with this PR.
> 
> Would be nice to have the time fixing code extracted into a helper but I can do that either in this PR or in a follow-up PR.

@p-datadog 
Perhaps something like
```ruby
with_frozen_time do
  # ... test stuff
end

# with a time set in case thats important
with_frozen_time(Time.utc(2000, 1, 1, 0, 0, 0)) do
   # ... test stuff
end
```

spec/datadog/data_streams/processor_spec.rb
```diff
@@ -186,47 +186,46 @@
...
+          processor.set_produce_checkpoint(type: 'kafka', destination: 'topicA', manual_checkpoint: false)
 
-        # Access internal buckets to verify aggregation
-        expect(processor.buckets).not_to be_empty
+          processor.send(:process_events)
 
-        # Find the bucket for this time window
-        now_ns = (now * 1e9).to_i
-        bucket_time_ns = now_ns - (now_ns % processor.bucket_size_ns)
+          now_ns = (Time.utc(2000, 1, 1, 0, 0, 0).to_f * 1e9).to_i
```
y9v: I find this time here a bit confusing - it's not obvious that this is the default value for frozen time. I agree with @Strech - regular stub is easier to read

spec/datadog/data_streams/processor_spec.rb
```diff
@@ -186,47 +186,46 @@
...
-        expect(stats[:edge_latency].encode).not_to be_empty
-        expect(stats[:full_pathway_latency].encode).to be_a(String)
-        expect(stats[:full_pathway_latency].encode).not_to be_empty
+          expect(stats[:edge_latency].count).to eq(3)
+          expect(stats[:full_pathway_latency].count).to eq(3)
+
+          expect(stats[:edge_latency].encode).to be_a(String)
+          expect(stats[:edge_latency].encode).not_to be_empty
+          expect(stats[:full_pathway_latency].encode).to be_a(String)
+          expect(stats[:full_pathway_latency].encode).not_to be_empty
```
y9v, line 230: using `aggregate_failures` here might improve test output in case of failure

spec/datadog/data_streams/processor_spec.rb
```diff
@@ -186,47 +186,48 @@
 
     describe 'internal bucket aggregation' do
       it 'aggregates multiple checkpoints into DDSketch histograms' do
-        now = Time.now.to_f
+        frozen_time = Time.utc(2000, 1, 1, 0, 0, 0)
+        allow(Datadog::Core::Utils::Time).to receive(:now).and_return(frozen_time)
+        allow(Datadog::Tracing).to receive(:active_span).and_return(nil)
+
+        processor.stop(true)
```
p-datadog, line 194: I am confused by this line - is the processor leaked from another test? If it's made for this test and not used then is it possible to not make it at all?

ericfirth: The processor is created in a before block on all tests (and initialization includes starting the background worker) and there is a after block that runs after each tests to stop it. But for this test alone, we need to stop it before creating the checkpoints so as to prevent the worker from draining the buckets before we inspect them in the tests.

There were actually 2 reasons this test flaked and that was the other one. I added a comment to explain the reasoning of that line

p-datadog (approved):

PR #5095, opened 2025-11-25
Transports: DRY HTTP client code

p-datadog:
**What does this PR do?**

Extracts `Core::Transport::HTTP::Client` from the various `HTTP::Client`s that are in individual components' transports.

The `HTTP::Client`s  change from modules to classes derived from the core Client.

Most of the clients contained identical code with exception of tracing `Traces` one - this has extra logic for tracking and sending statistics. I added callbacks to the core client to be invoked after successful and failed requests, and the `Traces` class uses them to track the statistics.

**Motivation:**
Making our codebase more legible

**Change log entry**
None

**Additional Notes:**
The code drops the default of `logger` to `Datadog.logger` for the `Traces` class since it now derives from the core one and I didn't put the default on the core one.

I also have a typing issue with `api` instances that I would love some help with.

**How to test the change?**

### DIFF
--- a/spec/datadog/tracing/transport/http/client_spec.rb
+++ b/spec/datadog/tracing/transport/http/client_spec.rb
@@ -14,7 +14,7 @@ RSpec.describe Datadog::Tracing::Transport::HTTP::Client do
   end
 
   describe '#send_request' do
-    subject(:send_request) { client.send_request(request, &block) }
+    subject(:send_request) { client.send(:send_request, request, &block) }
 
     let(:request) { instance_double(Datadog::Core::Transport::Request) }
     let(:response_class) { stub_const('TestResponse', Class.new { include Datadog::Core::Transport::HTTP::Response }) }
@@ -44,6 +44,9 @@ RSpec.describe Datadog::Tracing::Transport::HTTP::Client do
           allow(response).to receive(:not_found?).and_return(false)
           allow(response).to receive(:unsupported?).and_return(false)
 
+          # This test is very similar to the core client test;
+          # one difference is the expectation here on
+          # update_stats_from_response!.
           expect(client).to receive(:update_stats_from_response!)
             .with(response)
         end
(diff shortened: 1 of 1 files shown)


lib/datadog/core/telemetry/transport/http/telemetry.rb
```diff
@@ -3,19 +3,20 @@
...
         module HTTP
           module Telemetry
-            module Client
+            class Client < Core::Transport::HTTP::Client
               def send_telemetry_payload(request)
-                send_request(request) do |api, env| # steep:ignore
-                  api.send_telemetry(env)
+                send_request(request) do |api, env|
+                  # TODO how to make api have the derived type for steep?
+                  api.send_telemetry(env) # steep:ignore
```
vpellan, line 19: Best I can think of is to add a method definition for send_telemetry in the telemetry.rbs file. This will create a `Ruby::MethodDefinitionMissing` diagnostic on the Client class but it is a hint level diagnostic by default so we don't really care. I cannot make it work with inline annotations.

We could also rename all the `send_**` methods to `send_payload` or something like that, which would also exists in `Core::Transport::HTTP::API::Instance`

p-datadog: Thanks! Continuing to work on transport code, the next-next PR will remove all clients other than the core one, therefore maybe this issue will go away on its own.

lib/datadog/core/transport/http/client.rb
```diff
@@ -0,0 +1,69 @@
...
+          end
+
+          # Callback that is invoked if a request failed with an exception.
+          #
+          # Override in subclasses.
+          def on_exception(exception)
+            message = build_exception_message(exception)
+
+            logger.debug(message)
+          end
```
ivoanjo, line 59: Minor: If the intention is for these to be overridden, maybe consider leaving a `raise` here so we don't forget to override? (That, or change the initializer to check if `self` responds_to the expected methods and them leave only comments behind?)

PR #5096, opened 2025-11-26
Test suite: add tests for lack of output when loading core only

p-datadog:
**What does this PR do?**

**Motivation:**

We already have a test in master that datadog/core can be required successfully, but we do not have a test asserting lack of error output. Right now core can be required but there are exceptions being raised if calling `Datadog.configure` with only core actually loaded.

The idea of this PR was to add test coverage and clean up the code issue(s) that the tests would uncover, but the code issue is the fact that our Core is also holding Components which in turn depend on all of the products, and this issue must be addressed to have a clean architecture.

**Change log entry**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/lib/datadog/core/telemetry/event/app_started.rb
+++ b/lib/datadog/core/telemetry/event/app_started.rb
@@ -52,7 +52,7 @@ module Datadog
             products = {
               appsec: {
                 # TODO take appsec status out of component tree?
-                enabled: components.settings.appsec.enabled,
+                enabled: components.settings.respond_to?(:appsec) && components.settings.appsec.enabled,
               },
               profiler: {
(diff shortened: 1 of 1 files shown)


p-datadog:
I tried to fix the tests here and this is leading to a bunch of `respond_to?` checks in telemetry app started event implementation for the various product parts of the settings, where this is not an issue in "production" because all of our products are loaded all the time in practice.

I thought about this and my current plan is to propose lifting Components and Settings out of core to top level (https://github.com/p-datadog/datadog-docs/tree/master/components-refactor) to clean up the conceptual confusion, at which point the workarounds in the code at implementation level should not be necessary any more.

PR #5099, opened 2025-11-27
Transports: DRY HTTP Transport & Client classes

p-datadog:
**What does this PR do?**

Removes duplicated code from Transport classes and deletes most of the HTTP::Client classes.

This PR continues on the path started in https://github.com/DataDog/dd-trace-rb/pull/5095.

Most of the HTTP::Client classes were functionally identical, except for the Tracing::Traces one which had additional logic for API version downgrades.  The difference between the Client classes was in `send_XXX_payload` methods which were largely identical.  They have been replaced by a single `send_request` method with the XXX as the argument. Then, all of the HTTP::Client classes have been removed except for the Tracing::Traces one which derives from the core HTTP::Client.

This PR moves  also moves the downgrading methods to the core class because they will be used by DI in the next PR. Accordingly, some unit tests have been moved to core. Not all because I think there needs to be additional work to have mock APIs, since existing unit tests under Tracing::Traces utilize the APIs in Traces.

Finally, a base Transport class in core has been created with the API management code that was identical across all of the Transport classes in the various components. The per-component Transport classes remain since they still have unique methods but these now all derive from the core base class.

**Motivation:**
Making transport layer more legible to implement API downgrades for DI

**Change log entry**
None

**Additional Notes:**

**How to test the change?**

Existing tests

### DIFF
--- a/lib/datadog/core/transport/http/client.rb
+++ b/lib/datadog/core/transport/http/client.rb
@@ -18,12 +18,20 @@ module Datadog
             @logger = logger
           end
 
-          def send_request(action, request)
+          def send_request(action, request, &block)
+            send_request_impl(request) do |api, env|
+              api.public_send("send_#{action}", env)
+            end
+          end
+
+          private
+
+          def send_request_impl(request, &block)
             # Build request into env
             env = build_env(request)
 
             # Get responses from API
-            api.public_send("send_#{action}", env).tap do |response|
+            yield(api, env).tap do |response|
               on_response(response)
             end
           rescue => exception


lib/datadog/core/transport/http/client.rb
```diff
@@ -18,9 +18,15 @@ def initialize(api, logger:)
...
+          def send_request(action, request, &block)
+            send_request_impl(request) do |api, env|
+              api.public_send("send_#{action}", env)
+            end
+          end
+
           private
 
-          def send_request(request, &block)
+          def send_request_impl(request, &block)
```
p-datadog: I would love a better name for this method but can't think of one at the moment.

lib/datadog/core/transport/transport.rb
```diff
@@ -0,0 +1,90 @@
...
+        def initialize(version)
+          super
+
+          @version = version
+        end
+
+        def message
+          "No downgrade from transport API version #{version} is available!"
+        end
+      end
```
ivoanjo, line 37: So I realize this already partially existed before.

Having said that -- these two situations only happen when we have a bug in our code that can't be recovered from, and we don't particularly have a need to access the version or to recover from the individual errors.

So my suggestion is -- let's get rid of these and replace them with `raise ArgumentError, "Unknown API version: #{api_id}"` (or similar) below. I think that saves us some code and we can always introduce the more specific classes if we ever need to do anything with them.

lib/datadog/core/transport/transport.rb
```diff
@@ -0,0 +1,90 @@
...
+
+        class << self
+          # The HTTP client class to use for requests, derived from
+          # Core::Transport::HTTP::Client.
+          #
+          # Important: this attribute is NOT inherited by derived classes -
+          # it must be set by every Transport class that wants to have a
+          # non-default HTTP::Client instance.
+          attr_accessor :http_client_class
+        end
```
ivoanjo, line 51: This... is a kinda weird pattern. May I suggest making it an argument that gets passed to the constructor, or a regular method that needs to be overwritten by subclasses?

lib/datadog/core/transport/transport.rb
```diff
@@ -0,0 +1,90 @@
...
+
+        def message
+          "No downgrade from transport API version #{version} is available!"
+        end
+      end
+
+      # Base class for transports.
+      class Transport
+        attr_reader :client, :apis, :default_api, :current_api_id, :logger
```
ivoanjo, line 41: Minor: I think some (all?) of these can be made `private`

lib/datadog/core/transport/http/client.rb
ivoanjo: I think this PR changed every `send_request` to not pass in a block. Being that the case, why not merge these two and make `send_request` directly trigger the `public_send`?

spec/datadog/core/transport/transport_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+        v2: api_v2,
+        v1: api_v1
+      ].with_fallbacks(v2: :v1)
+    end
+
+    let(:api_v1) { instance_double(Datadog::Core::Transport::HTTP::API::Instance, 'v1', encoder: encoder_v1) }
+    let(:api_v2) { instance_double(Datadog::Core::Transport::HTTP::API::Instance, 'v2', encoder: encoder_v2) }
+    let(:encoder_v1) { instance_double(Datadog::Core::Encoding::Encoder, 'v1', content_type: 'text/plain') }
+    let(:encoder_v2) { instance_double(Datadog::Core::Encoding::Encoder, 'v2', content_type: 'text/csv') }
+  end
```
ivoanjo, line 27: Minor: Is it me or does every entry in this spec include this shared_context? E.g. I think we can simplify it

spec/datadog/core/transport/transport_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+
+  describe '#initialize' do
+    include_context 'APIs with fallbacks'
+
+    it { is_expected.to have_attributes(apis: apis, current_api_id: current_api_id) }
+  end
```
ivoanjo, line 33: Minor: Maybe remove this? This test is trivial and this behavior ends up being exercised separately by the other tests

lib/datadog/core/transport/http/client.rb
p-datadog: I did this change and tests majorly broke because they do a lot of mocking. I do like the change, I suppose need to spend more time rewriting the tests.

lib/datadog/core/transport/transport.rb
p-datadog: I can do this, I don't think I'm strongly attached to the existing exceptions.

lib/datadog/core/transport/transport.rb
p-datadog: My thinking in general was to make the transports more declarative but I can try a version with an instance method.

spec/datadog/core/transport/transport_spec.rb
p-datadog: Probably due to copy-pasting from the tracing test.

lib/datadog/core/transport/http/client.rb
ivoanjo: Yeah... somehow not surprised. This is why I mentioned

> TBH I suspect in the long run it may be easier to start from scratch and rebuilding the support for the required feature-set VS slowly untangling the current mess back into a good design step by step.

I believe this design is at this point so weird that "chipping" at it is incredibly hard.

lib/datadog/core/transport/transport.rb
ivoanjo: So I think declarative sometimes works really well...

In our case, I think it's made it really hard to reason and fix things with the current transport design. My thinking is that at this point we should strip out as much as possible of it until we get to a better state. Then, once we're happier with the transport design, we can reintroduce the declarative features as a nicer DSL for underpinnings we're happy with, and it won't affect with the overall design we're trying to fix.

spec/datadog/core/transport/transport_spec.rb
ivoanjo: (I think we overuse `shared_context`s sometimes, so I'm always on the look out for simplifying them)

lib/datadog/core/transport/http/client.rb
```diff
@@ -18,14 +18,12 @@ def initialize(api, logger:)
...
-          private
-
-          def send_request(request, &block)
+          def send_request(action, request)
             # Build request into env
             env = build_env(request)
 
             # Get responses from API
-            yield(api, env).tap do |response|
+            api.public_send("send_#{action}", env).tap do |response|
```
vpellan: Would it be possible to rename all the `API::Instance.send_telemetry` or `send_stats` to `send_request` ? That way we would not need `public_send` here and just call `api.send_request(env)`

spec/datadog/core/transport/http/client_spec.rb
```diff
@@ -11,8 +11,8 @@
     it { is_expected.to have_attributes(api: api) }
   end
 
-  describe '#send_request' do
-    subject(:send_request) { client.send(:send_request, request, &block) }
+  describe '#send_request_impl' do
```
vpellan, line 14: I suppose all these `send_request_impl` changes were related to a previous change that has not been updated. It makes the CI fails but except that I think PR looks good

p-datadog:
I will be addressing the comments in the future PRs, thank you for the feedback.

PR #5110, opened 2025-12-04
feat(telemetry): Add process tags to instrumentation telemetry

wantsui:
**What does this PR do?**

Adds process tags to instrumentation telemetry for https://datadoghq.atlassian.net/browse/AIDM-255 

**Motivation:**

https://datadoghq.atlassian.net/browse/AIDM-255

Follows the work done in https://github.com/DataDog/dd-trace-rb/pull/5033

The Python implementation that I followed can be found here: https://github.com/DataDog/dd-trace-py/pull/15227 
**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

```
docker compose run --rm tracer-3.3 /bin/bash
bundle exec rspec spec/datadog/core/telemetry/integration/telemetry_spec.rb
```

### DIFF
--- a/spec/datadog/di/integration/probe_notification_builder_spec.rb
+++ b/spec/datadog/di/integration/probe_notification_builder_spec.rb
@@ -23,6 +23,7 @@ RSpec.describe Datadog::DI::ProbeNotificationBuilder do
         allow(settings).to receive(:env).and_return('fake env')
         allow(settings).to receive(:version).and_return('fake version')
         allow(settings).to receive(:tags).and_return({})
+        allow(settings).to receive(:experimental_propagate_process_tags_enabled).and_return(false)
       end
     end
 


lib/datadog/core/telemetry/request.rb
```diff
@@ -64,6 +66,15 @@ def host
...
             }
           end
+
+          def tag_process_tags!(hash)
+            return unless Datadog.configuration.experimental_propagate_process_tags_enabled
+
+            process_tags = Core::Environment::Process.serialized
+            return if process_tags.empty?
+
+            hash[:process_tags] = process_tags
```
marcotc: Looks like it should be inside `application` (`hash[:application][:process_tags]`): https://github.com/DataDog/dd-trace-java/pull/8837/files#diff-181302dc1caf7e99d96755e0405c0b445baf107ad04f6f3d3649d20494f097b4R177

lib/datadog/core/telemetry/request.rb
```diff
@@ -64,6 +69,15 @@ def host
               kernel_version: Core::Environment::Platform.kernel_version
             }
           end
+
+          def tag_process_tags!(app)
+            return unless Datadog.configuration.experimental_propagate_process_tags_enabled
```
p-datadog: The caller has `config = Datadog.configuration`, do you mind passing config as parameter into this method instead of performing another global variable read here?

p-datadog: Or you could move the conditional to the caller to avoid passing config here.

wantsui: Addressed in https://github.com/DataDog/dd-trace-rb/pull/5110/commits/2848b63de8bf4755766834ad1fa69283d8748911! It also triggered a steep error so I fixed that too :)

lib/datadog/core/telemetry/request.rb
wantsui: Addressed in b08f00b344597d7c98f11e8050734bd3d4d23195

PR #5111, opened 2025-12-05
DI: add a C extension

p-datadog:
**What does this PR do?**
Adds two C extension functions to libdatadog_api for Dynamic Instrumentation:

1) `all_iseqs`: returns `RubyVM::InstructionSequence` objects for loaded code, needed for instrumenting third-party code with line probes. Not yet wired into instrumentation — that happens in #5496.
2) `exception_message`: returns the raw argument passed to the Exception constructor, bypassing any Ruby-level `message` method overrides.

Integrates `exception_message` into `ProbeNotificationBuilder` so method probe snapshots now populate the `throwable` field (type, message, stacktrace).

The C extension is now a hard requirement for DI — `environment_supported?` checks for it, and the new code calls it directly with no fallback.

DI specs under `spec/datadog/di/` are moved from `spec:main` to the new `spec:di:di_with_ext` task, which compiles the extension before running tests. The task is added to the Matrixfile so CI runs them.

**Motivation:**
- Line probes require iseq objects that the VM does not expose to Ruby code; the C extension provides access.
- DI must never invoke customer code; `exception_message` accesses the raw C-level message, bypassing overridden `message` methods.
- The C extension always compiles on MRI (the only platform DI supports), so there is no need for fallback code paths. Making it a hard requirement simplifies the codebase and makes failures explicit.

**Change log entry**
Yes. Dynamic Instrumentation: exception messages are now safely reported in method probe snapshots. The libdatadog_api C extension is now required for Dynamic Instrumentation.

**Additional Notes:**
The C code lives in the existing libdatadog_api extension (does not use libdatadog itself — this avoids creating another extension).

An error boundary is added around `probe_executed_callback` in method probes so DI errors never propagate to or replace customer exceptions.

Non-string exception constructor arguments are reported as `<REDACTED: not a string value>` rather than calling `.to_s` (which could invoke customer code). Nil arguments pass through as nil.

Follow-up work:
- Populate CodeTracker iseq registry with `all_iseqs`/`file_iseqs` to enable line probes on code loaded before DI starts (#5496)
- Use `trace_points` field of iseqs to detect line probes targeting non-executable lines
- Report ranges of executable code to SymDB

**How to test the change?**
Unit tests added for C extension functions (`all_iseqs`, `exception_message`). Unit tests for `serialize_throwable` cover: exception present/absent, overridden message, non-string arg, nil arg. Integration test verifies full path from method probe exception capture through to snapshot with populated throwable. Manual verification screenshot included below.

Screenshot of a manual test with a non-string Exception argument:

### DIFF
--- a/sig/datadog/di/probe_notification_builder.rbs
+++ b/sig/datadog/di/probe_notification_builder.rbs
@@ -3,6 +3,7 @@ module Datadog
     class ProbeNotificationBuilder
       NANOSECONDS: Integer
       MILLISECONDS: Integer
+      BACKTRACE_FRAME_PATTERN: Regexp
 
       @serializer: Serializer
 


ivoanjo:
> The C code is in libdatadog_api extension, however it does not use libdatadog. This was done to not create another extension.

I think this is the right decision -- managing multiple extensions adds complexity AND makes installation slower.

ext/libdatadog_api/ddsketch.h
ivoanjo, line 1: TBH I'm not sure this is very valuable -- I think having the `init` header inlined in  ìnit.c` is fine if there's no C APIs being exposed. (It saves us from a bit of an explosion of files)

(Same note for `di.h`)

ext/libdatadog_api/ruby_helpers.c
```diff
@@ -0,0 +1,22 @@
+#include <stdbool.h>
+#include "ruby_internal.h"
+#include "ruby_helpers.h"
+
+inline int ddtrace_imemo_type(VALUE imemo) {
+  // This mask is the same between Ruby 2.5 and 3.3-preview3. Furthermore, the intention of this method is to be used
+  // to call `rb_imemo_name` which correctly handles invalid numbers so even if the mask changes in the future, at most
+  // we'll get incorrect results (and never a VM crash)
+  return (RBASIC(imemo)->flags >> FL_USHIFT) & IMEMO_MASK;
+}
```
ivoanjo: Minor: Do you maybe want to move this to `datadog_ruby_common.h` (making it static inline) so we'd share it between both extensions?

ext/libdatadog_api/ruby_internal.h
ivoanjo, line 1: Since these are not shared among all files, I'd suggest defining the prototypes in the files that need them (organization in C is hard enough, so I think there's value in avoiding an explosion of files, if we can)

ext/libdatadog_api/ruby_helpers.c
```diff
@@ -0,0 +1,22 @@
...
+inline bool ddtrace_imemo_iseq_p(VALUE v) {
+    if (rb_objspace_internal_object_p(v)) {
+        if (RB_TYPE_P(v, T_IMEMO)) {
+            if (ddtrace_imemo_type(v) == IMEMO_TYPE_ISEQ) {
+                return true;
+            }
+        }
+    }
+    return false;
+}
```
ivoanjo: Minor: I'd suggest maybe moving this to `di.c`, since this is only used there?

ext/libdatadog_api/ruby_helpers.c
```diff
@@ -0,0 +1,22 @@
...
+// Returns whether the argument is an IMEMO of type ISEQ.
+inline bool ddtrace_imemo_iseq_p(VALUE v) {
+    if (rb_objspace_internal_object_p(v)) {
+        if (RB_TYPE_P(v, T_IMEMO)) {
+            if (ddtrace_imemo_type(v) == IMEMO_TYPE_ISEQ) {
+                return true;
+            }
+        }
+    }
+    return false;
```
ivoanjo: Minor:

```
is (this)
  a really (awkward way)
    of (writing)
      return rb_objspace_internal_object_p(v) && RB_TYPE_P(v, T_IMEMO) && ddtrace_imemo_type(v) == IMEMO_TYPE_ISEQ;
```

ext/libdatadog_api/di.c
```diff
@@ -0,0 +1,123 @@
+#include <ruby.h>
+#include <stdbool.h>
+#include "ruby_internal.h"
+#include "ruby_helpers.h"
+
+#ifndef DDTRACE_UNUSED
+#define DDTRACE_UNUSED  __attribute__((unused))
+#endif
```
ivoanjo: This is already defined in `datadog_ruby_common.h` btw ;)

ext/libdatadog_api/di.c
```diff
@@ -0,0 +1,123 @@
...
+  VALUE v = (VALUE)vstart;
+  for (; v != (VALUE)vend; v += stride) {
+    if (ddtrace_imemo_iseq_p(v)) {
+      VALUE iseq = rb_iseqw_new((void *) v);
+      rb_ary_push(array, iseq);
+    }
+  }
+
+  return 0;
+}
```
ivoanjo: Maybe it's worth mentioning here that we're doing the same as https://github.com/ruby/debug/blob/master/ext/debug/iseq_collector.c although we did arrive at it independently in the beginning (but when there's a reference for what we're doing that's maintained by ruby core, I think that's a very relevant reference to keep an eye on!)

ext/libdatadog_api/di.c
```diff
@@ -0,0 +1,123 @@
...
+      def self.bar; define_method(:dynamic) { 42 } end
+    end
+
+After +bar+ executes, it will have defined the method +dynamic+ which
+exists on the same lines of code as +bar+.
+
+This situation is considered an error in DI/LD products - a probe
+specification must resolve to exactly one code location, because the UI
+only has provision to how one code location as the instrumentation target.
+Thus, in this case, instrumentation should produce an error.
```
ivoanjo: This is relevant info... and yet I'm not sure it belongs here attached to this method. E.g. this is more on the consumer side "how to deal with the data we get from all_iseqs" than "this is about all_iseqs". Maybe move it elsewhere...?

ext/libdatadog_api/di.c
```diff
@@ -0,0 +1,123 @@
...
+*/
+static VALUE all_iseqs(DDTRACE_UNUSED VALUE _self) {
+  struct ddtrace_di_os_each_struct oes;
+
+  oes.array = rb_ary_new();
+  rb_objspace_each_objects(ddtrace_di_os_obj_of_i, &oes);
+  RB_GC_GUARD(oes.array);
+  return oes.array;
```
ivoanjo: Minor:

1. It's possible to pass the array directly into `rb_objspace_each_objects` as a pointer (without the extra struct).
2. The `RB_GC_GUARD` is not needed since the compiler MUST keep the array around on the stack -- it's going to return it!

p-datadog:
Thanks for confirming — glad the approach makes sense!

ext/libdatadog_api/ddsketch.h
p-datadog: Agreed, forward declarations in `init.c` are cleaner when there's no shared API. Removed both headers and inlined the declarations:
- **ext/libdatadog_api/init.c** — forward declarations for `ddsketch_init` and `di_init`
- Deleted `ddsketch.h` and `di.h`

ext/libdatadog_api/ruby_helpers.c
p-datadog: Makes sense — moved `ddtrace_imemo_type` to `datadog_ruby_common.h` as `static inline` along with the `IMEMO_MASK` constant. The profiling extension can pick it up when the copies are next synced.

ext/libdatadog_api/ruby_internal.h
p-datadog: Yep, much tidier that way. Moved the prototypes directly into `di.c` and deleted `ruby_internal.h`.

ext/libdatadog_api/ruby_helpers.c
p-datadog: Done — moved `ddtrace_imemo_iseq_p` into `di.c` as a static function. With `ddtrace_imemo_type` now in `datadog_ruby_common.h`, the whole `ruby_helpers.c`/`ruby_helpers.h` pair is gone.

ext/libdatadog_api/ruby_helpers.c
p-datadog: Ha, fair enough. Flattened to a single return statement.

ext/libdatadog_api/di.c
p-datadog: Ah right, missed that! Removed the duplicate and switched to including `datadog_ruby_common.h` instead.

ext/libdatadog_api/di.c
p-datadog: Definitely worth having that reference. Added a comment pointing to ruby/debug's `iseq_collector.c`.

ext/libdatadog_api/di.c
p-datadog: Right, this reads more like design notes for the consumer than documentation for the C function itself. Trimmed the `all_iseqs` comment down to what the function returns and moved the consumer-side notes into `lib/datadog/di.rb` near `file_iseqs`.

ext/libdatadog_api/di.c
p-datadog: Nice spot on both counts. Removed the struct wrapper — passing `&array` directly now — and dropped the `RB_GC_GUARD`.

Strech:
@codex Review for bugs and edge-cases

lib/datadog/di/probe_notification_builder.rb
```diff
@@ -153,6 +153,68 @@ def build_status(probe, message:, status:, exception: nil)
...
+      #
+      # Ruby backtrace format: "/path/file.rb:42:in `method_name'"
+      #
+      # @param backtrace [Array<String>, nil] from Exception#backtrace
+      # @return [Array<Hash>, nil]
+      def format_backtrace(backtrace)
+        return [] if backtrace.nil?
+
+        backtrace.map do |frame|
+          if frame =~ /\A(.+):(\d+):in\s+[`'](.+)'\z/
```
Strech: Could you extract it in a constant that name helps us to understand the purpose easier?

p-datadog: Makes sense — the regex is hard to parse at a glance. Extracted to `BACKTRACE_FRAME_PATTERN`.

Fixed in:
- **lib/datadog/di/probe_notification_builder.rb:61** — added `BACKTRACE_FRAME_PATTERN` constant
- **lib/datadog/di/probe_notification_builder.rb:214** — used constant in `format_backtrace`

lib/datadog/di/probe_notification_builder.rb
```diff
@@ -153,6 +153,68 @@ def build_status(probe, message:, status:, exception: nil)
...
+        else
+          # Non-string constructor argument — return a redacted placeholder
+          # rather than calling .to_s which could be customer code.
+          # The exception class is already reported via the :type field.
+          '<REDACTED: not a string value>'
+        end
+        {
+          type: exception.class.name,
+          message: message,
+          stacktrace: format_backtrace(exception.backtrace),
```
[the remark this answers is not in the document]
p-datadog: Yep, good catch on the inconsistency. Added a `DI.exception_backtrace` C extension function in follow-up PR #5521 that reads the raw `bt` ivar directly — same pattern as `exception_message`. No Ruby method dispatch involved.

PR #5124, opened 2025-12-09
[NO-TICKET] Fix `FallbackAny` steep check

vpellan:
**What does this PR do?**

RemoveFallbackAny exclusion from the Steepfile and fixes related issues.

**Motivation:**

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

bundle exec rake steep:check

### DIFF
--- a/lib/datadog/profiling/collectors/info.rb
+++ b/lib/datadog/profiling/collectors/info.rb
@@ -15,14 +15,7 @@ module Datadog
       # gathering and easily support more flexible/dynamic info collection in the future.
       class Info
         def initialize(settings)
-          # Steep: This one is weird, we set it to nil here but it is always set
-          # to type `profiler_info` in the collect_profiler_info method.
-          # We type it as `profiler_info` in the .rbs file, but set it to `profiler_info?` here
-          # to satisfy the type checker.
-          # Because we type it as `profiler_info` in the .rbs file, the type checker will
-          # complain that the branch in the collect_profiler_info method is unreachable.
-          # But it is just :information level, so we can ignore it.
-          @profiler_info = nil # @type ivar @profiler_info: profiler_info?
+          @profiler_info = nil
 
           # Steep: https://github.com/soutaro/steep/issues/363
           @info = { # steep:ignore IncompatibleAssignment
@@ -105,7 +98,7 @@ module Datadog
         end
 
         def collect_profiler_info(settings)
-          unless @profiler_info
+          @profiler_info ||= begin
             lib_datadog_gem = ::Gem.loaded_specs["libdatadog"]
 
             libdatadog_version =
@@ -117,13 +110,12 @@ module Datadog
                 "#{Libdatadog::VERSION}-(unknown)"
               end
 
-            @profiler_info = {
+            {
               version: Datadog::Core::Environment::Identity.gem_datadog_version,
               libdatadog: libdatadog_version,
               settings: collect_settings_recursively(settings.profiling),
             }.freeze
           end
-          @profiler_info
         end
 
         # The settings/option model isn't directly serializable because
--- a/sig/datadog/profiling/collectors/info.rbs
+++ b/sig/datadog/profiling/collectors/info.rbs
@@ -3,7 +3,7 @@ module Datadog
     module Collectors
       class Info
         @info: info
-        @profiler_info: profiler_info
+        @profiler_info: profiler_info?
         @gc_tuning_info: ::Hash[::Symbol, ::String]
 
         def initialize: (untyped settings) -> void


lib/datadog/tracing/span_operation.rb
```diff
@@ -159,13 +159,12 @@ def measure
...
-            # block is application code that we don't want to hinder.
-            # * We don't yield during a fatal error, as the application is likely trying to
-            #   end its execution (either due to a system error or graceful shutdown).
-            return_value = yield(self) unless e && !e.is_a?(StandardError)
           end
+          # We should yield to the provided block when possible, as this
+          # block is application code that we don't want to hinder.
+          # * We don't yield during a fatal error, as the application is likely trying to
+          #   end its execution (either due to a system error or graceful shutdown).
+          return_value = yield(self) unless e && !e.is_a?(StandardError)
```
marcotc: Do you know why this change was made?
It might be an ok change (there's a subtle different if `logger.debug { "Failed to start span: #{e}" }` above raises an error), but I'm wondering what broke here to cause this code change.

lib/datadog/tracing/span_operation.rb
```diff
@@ -466,7 +466,7 @@ def publish(*args)
               @handler.call(*args)
             rescue => e
               logger.debug do
-                "Error in on_error handler '#{@default}': #{e.class}: #{e} at #{Array(e.backtrace).first}"
+                "Error in on_error handler '#{@handler}': #{e.class}: #{e} at #{Array(e.backtrace).first}"
```
marcotc, line 474: Nice catch!

lib/datadog/core/utils/sequence.rb
```diff
@@ -12,7 +12,8 @@ def initialize(seed = 0, &block)
         end
 
         def next
-          next_item = @next_item ? @next_item.call(@current) : @current
+          # Steep: https://github.com/soutaro/steep/issues/477
+          next_item = @next_item ? @next_item.call(@current) : @current # steep:ignore NoMethod
```
Strech: Why not annotation with interface instead?

lib/datadog/di/instrumenter.rb
```diff
@@ -114,7 +114,9 @@ def hook_method(probe, responder)
         settings = self.settings
 
         mod = Module.new do
-          define_method(method_name) do |*args, **kwargs, &target_block| # steep:ignore
+          define_method(method_name) do |*args, **kwargs, &target_block| # steep:ignore NoMethod
```
Strech, line 117: What about type annotations?

sig/datadog/core/telemetry/event/app_started.rbs
```diff
@@ -3,12 +3,16 @@ module Datadog
     module Telemetry
       module Event
         class AppStarted < Base
+          @configuration: Array[Hash[Symbol, untyped]]
```
Strech: This should be defined outside as a type alias instead. Same applies to next 2 instance variables

sig/datadog/core/vendor/multipart-post/multipart/post/composite_read_io.rbs
```diff
@@ -4,6 +4,8 @@ module Datadog
       module Multipart
         module Post
           class CompositeReadIO
+            @ios: ::Array[untyped]
```
Strech: Is it really untyped if we have right next and `@io` definition from parts.rbs?

lib/datadog/core/utils/sequence.rb
vpellan: Here the problem is type narrowing (more details on the issue linked). I can remove the `# steep:ignore` and replace it by `# @type ivar @next_item: ^(::Integer) -> untyped` but it will create another diagnostic, saying that the second part of that ternary operator will be unreachable. This diagnostic is at `:information` level so it's not going to create an error, do you think it's preferable ?

lib/datadog/di/instrumenter.rb
vpellan: Here the problem is define_method. Steep gives that error:
```
Type `::Datadog::DI::Instrumenter` does not have method `define_method` (Ruby::NoMethod)
```
I just narrowed the steep:ignore to NoMethod so we can still catch other errors

vpellan: And I used type annotation below to fix an error that was linked to **kwargs

lib/datadog/core/utils/sequence.rb
vpellan: Thinking about it again, I think it's better to have an `UnreachableBranch` diagnostic, as a general case, if the method creating a `NoMethod` diagnostic is actually removed from the class, we will not be able to catch it. I'll replace as much `steep:ignore` as I can with that

lib/datadog/tracing/span_operation.rb
vpellan: I've reverted it and added a steep:ignore but it is caused by these bugs (they seem related): https://github.com/soutaro/steep/issues/919
https://github.com/soutaro/steep/issues/1971

lib/datadog/core/utils/sequence.rb
vpellan: Applied in 4be18ce420156488af903e4e5d18f8fa64e54835. I've tried to remove as much steep:ignore as possible in cf6491eae1322da09ce664effc2f7d5e9703fbbc. It could be its own PR If you feel like it bloats this PR too much

sig/datadog/core/telemetry/event/app_started.rbs
vpellan: Applied in 4be18ce420156488af903e4e5d18f8fa64e54835

sig/datadog/core/vendor/multipart-post/multipart/post/composite_read_io.rbs
vpellan: Added the type in 4be18ce420156488af903e4e5d18f8fa64e54835

lib/datadog/profiling/collectors/info.rb
```diff
@@ -15,7 +15,15 @@ module Collectors
...
         def initialize(settings)
-          @profiler_info = nil
+          # Steep: This one is weird, we set it to nil here but it is always set
+          # to type `profiler_info` in the collect_profiler_info method.
+          # We type it as `profiler_info` in the .rbs file, but set it to `profiler_info?` here
+          # to satisfy the type checker.
+          # Because we type it as `profiler_info` in the .rbs file, the type checker will
+          # complain that the branch in the collect_profiler_info method is unreachable.
+          # But it is just :information level, so we can ignore it.
+          @profiler_info = nil # @type ivar @profiler_info: profiler_info?
```
ivoanjo: Minor: I'm OK with doing it this way although we could change `collect_profiler_info` to use a `||= begin ... ` to match the other methods and allow us to remove this.

(I believe this initialization was only added because the `unless @profiler_info` check would generate a warning in some Ruby versions, but using `||= begin` will not generate the warning so it's a feasible replacement as well).

sig/datadog/appsec/actions_handler/serializable_backtrace.rbs
```diff
@@ -2,10 +2,14 @@ module Datadog
   module AppSec
     module ActionsHandler
       class SerializableBacktrace
+        include ::MessagePack::_Extension
+
         CLASS_AND_FUNCTION_NAME_REGEX: ::Regexp
 
+        @stack_id: String
```
Strech:
```suggestion
        @stack_id: ::String
```

sig/datadog/appsec/actions_handler/serializable_backtrace.rbs
```diff
@@ -2,10 +2,14 @@ module Datadog
...
+        @locations: ::Array[::Thread::Backtrace::Location]
+
         def initialize: (locations: ::Array[::Thread::Backtrace::Location], stack_id: String) -> void
```
Strech:
```suggestion
        def initialize: (locations: ::Array[::Thread::Backtrace::Location], stack_id: ::String) -> void
```

sig/datadog/appsec/context.rbs
```diff
@@ -15,7 +15,7 @@ module Datadog
 
       @interrupted: bool
 
-      ActiveContextError: ::StandardError
+      ActiveContextError: StandardError
```
Strech, line 18: I think we want to have a top-level class here

sig/datadog/core/configuration/config_helper.rbs
```diff
@@ -6,14 +6,13 @@ module Datadog
 
         @source_env: env_hash
         @raise_on_unknown_env_var: bool
-        @supported_configurations: Array[String]
+        @supported_configurations: Set[String]
```
Strech, line 9:
```suggestion
        @supported_configurations: ::Set[::String]
```

sig/datadog/core/configuration/config_helper.rbs
```diff
@@ -6,14 +6,13 @@ module Datadog
...
         @aliases: Hash[String, Array[String]]
-        @deprecations: Hash[String, String]
         @alias_to_canonical: Hash[String, String]
 
         def initialize: (
           ?source_env: env_hash,
-          ?supported_configurations: Array[String],
+          ?supported_configurations: Set[String],
```
Strech, line 15:
```suggestion
          ?supported_configurations: ::Set[::String],
```

sig/datadog/core/configuration/deprecations.rbs
```diff
@@ -4,11 +4,11 @@ module Datadog
       module Deprecations
         LOG_DEPRECATIONS_ONLY_ONCE: Datadog::Core::Utils::OnlyOnce
 
-        def self.log_deprecations_from_all_sources: (Datadog::Core::Logger logger, ?deprecations: Hash[String, String], ?alias_to_canonical: Hash[String, String]) -> void
+        def self.log_deprecations_from_all_sources: (Datadog::Core::Logger logger, ?deprecations: Set[String], ?alias_to_canonical: Hash[String, String]) -> void
```
Strech, line 7:
```suggestion
        def self.log_deprecations_from_all_sources: (Datadog::Core::Logger logger, ?deprecations: ::Set[::String], ?alias_to_canonical: ::Hash[::String, ::String]) -> void
```

sig/datadog/core/configuration/deprecations.rbs
```diff
@@ -4,11 +4,11 @@ module Datadog
...
 
         private
 
-        def self.log_deprecated_environment_variables: (Datadog::Core::Logger logger, ConfigHelper::env_hash source_env, String source_name, Hash[String, String] deprecations, Hash[String, String] alias_to_canonical) -> void
+        def self.log_deprecated_environment_variables: (Datadog::Core::Logger logger, ConfigHelper::env_hash source_env, String source_name, Set[String] deprecations, Hash[String, String] alias_to_canonical) -> void
```
Strech, line 11:
```suggestion
        def self.log_deprecated_environment_variables: (Datadog::Core::Logger logger, ConfigHelper::env_hash source_env, ::String source_name, ::Set[::String] deprecations, ::Hash[::String, ::String] alias_to_canonical) -> void
```

lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -54,17 +54,17 @@ def new_runner
         end
 
         def add_or_update_config(config, path:)
-          @is_ruleset_update = path.include?('ASM_DD')
+          is_ruleset_update = path.include?('ASM_DD')
```
marcotc, line 57: This seems like a minor code issue that was surfaced by type checking! 👍  🔥

lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -48,7 +48,7 @@ def app_started?
           private
 
           def products(components)
-            # @type var products: Hash[Symbol, Hash[Symbol, Hash[Symbol, String | Integer] | bool | nil]]
+            # @type var products: telemetry_products
```
marcotc, line 51: ❤️ for the type reuse

lib/datadog/core/semaphore.rb
```diff
@@ -20,10 +20,9 @@ def signal
 
       def wait(timeout = nil)
         wake_lock.synchronize do
-          # steep specifies that the second argument to wait is of type
-          # ::Time::_Timeout which for some reason is not Numeric and is not
-          # castable from Numeric.
-          wake.wait(wake_lock, timeout) # steep:ignore
+          # Reason why _Timeout is not Numeric is because Numeric does not implement `*`
+          # Which is required for timeout. But all core classes inheriting from Numeric implement `*`.
```
marcotc: I think we can remove this comment from `lib`, since we are just respecting the interface of [`ConditionVariable#wait`](https://github.com/ruby/rbs/blob/e7982676710c3c57527402a6d36b7b344f8caa74/core/thread.rbs#L1493):
```suggestion
```

We can add a comment to the rbs files if you think it's necessary, but since we are just asking for the same time as the upstream method, I'm not sure if it's necessary.

lib/datadog/profiling/collectors/code_provenance.rb
```diff
@@ -161,7 +161,7 @@ def initialize(kind:, name:, version:, path:, extra_paths:)
           end
 
           def to_json(arg = nil)
-            # Steep: https://github.com/ruby/rbs/pull/2691
+            # Steep: https://github.com/ruby/rbs/pull/2691 (remove after Ruby 4.0 release)
```
marcotc:
```suggestion
            # Steep: https://github.com/ruby/rbs/pull/2691 (remove after RBS 4.0 release)
```

lib/datadog/tracing/distributed/baggage.rb
```diff
@@ -171,11 +171,14 @@ def build_baggage_tags(baggage)
           return {} if baggage_tag_keys.empty?
 
           # If wildcard is specified, use all baggage keys
-          baggage_tag_keys = baggage if baggage_tag_keys == BAGGAGE_TAG_KEYS_MATCH_ALL
+          baggage_tag_keys = baggage.keys if baggage_tag_keys == BAGGAGE_TAG_KEYS_MATCH_ALL
```
marcotc: This change creates an intermediate copy of `keys` unnecessarily, as `each` doesn't need to create a copy to iterate over the keys. I remember doing this on purpose 😅 

I recommend keeping the code as it was, and getting steep to cooperate or unfortunately ignore the type error:
```suggestion
          baggage_tag_keys = baggage if baggage_tag_keys == BAGGAGE_TAG_KEYS_MATCH_ALL
```

sig/datadog/appsec/actions_handler/serializable_backtrace.rbs
vpellan: Thanks! I'll change it on my changes, but I think it is preferable to do it globally on another PR

lib/datadog/core/semaphore.rb
vpellan: removed in f03a22d33e2291fb6d34d3b3cf016a8d0c3f0dcf

lib/datadog/profiling/collectors/code_provenance.rb
vpellan: Thanks ! I assumed that RBS 4.0 would be released at the same time as Ruby 4.0... Fixed in f03a22d33e2291fb6d34d3b3cf016a8d0c3f0dcf

lib/datadog/tracing/distributed/baggage.rb
vpellan: I added a steep:ignore here: f03a22d33e2291fb6d34d3b3cf016a8d0c3f0dcf

sig/datadog/appsec/actions_handler/serializable_backtrace.rbs
vpellan: Changed in 3c1c2b733b71cd65323233fdd9e80133d0356286

lib/datadog/profiling/collectors/code_provenance.rb
marcotc: I thought so too! I guess it's hard to have all the ecosystem in place by Christmas 😅.

marcotc: I thought so too! I guess it's hard to have all the ecosystem in place by Christmas 😅.

lib/datadog/profiling/collectors/info.rb
vpellan: Thanks for the tip! Changed in 8a2cb5db8e90eeac0fe8d823ecf490d04ad08727 !

PR #5126, opened 2025-12-09
RC: Add process tags to remote configuration payloads

wantsui:
**What does this PR do?**

Adds process tags to the dynamic instrumentation payloads, similar to these two PRs:
- https://github.com/DataDog/dd-trace-py/pull/15298
- https://github.com/DataDog/dd-trace-java/pull/8705

**Note** In both of those the remote configuration payload required the process tags to be an array so there's now a new method in Process.rb to do this.

**Motivation:**

For https://datadoghq.atlassian.net/browse/AIDM-257 

**Change log entry**

None.

**Additional Notes:**

Note: This PR should not be merged until https://github.com/DataDog/dd-trace-rb/pull/5110 gets merged

**How to test the change?**

I've been running:

```
docker compose run --rm tracer-3.3 /bin/bash
bundle exec rspec spec/datadog/core/remote/client_spec.rb
bundle exec rspec spec/datadog/core/environment/process_spec.rb
```

### DIFF
--- a/lib/datadog/core/environment/process.rb
+++ b/lib/datadog/core/environment/process.rb
@@ -15,13 +15,13 @@ module Datadog
         def self.serialized
           return @serialized if defined?(@serialized)
 
-          @serialized = tags_array.join(',').freeze
+          @serialized = tags.join(',').freeze
         end
 
         # This method returns an array in the format ["k1:v1","k2:v2","k3:v3"]
         # @return [Array<String>] array of normalized key:value pairs
-        def self.tags_array
-          return @tags_array if defined?(@tags_array)
+        def self.tags
+          return @tags if defined?(@tags)
           tags = []
 
           workdir = TagNormalizer.normalize_process_value(entrypoint_workdir.to_s)
@@ -35,7 +35,7 @@ module Datadog
 
           tags << "#{Environment::Ext::TAG_ENTRYPOINT_TYPE}:#{TagNormalizer.normalize(entrypoint_type, remove_digit_start_char: false)}"
 
-          @tags_array = tags.freeze
+          @tags = tags.freeze
         end
 
         # Returns the last segment of the working directory of the process
--- a/spec/datadog/core/environment/process_spec.rb
+++ b/spec/datadog/core/environment/process_spec.rb
@@ -121,19 +121,19 @@ RSpec.describe Datadog::Core::Environment::Process do
       end
     end
   end
-  describe '::tags_array' do
-    subject(:tags_array) { described_class.tags_array }
+  describe '::tags' do
+    subject(:tags) { described_class.tags }
 
     it { is_expected.to be_a_kind_of(Array) }
 
     it 'is an array of strings' do
-      expect(tags_array).to all(be_a(String))
+      expect(tags).to all(be_a(String))
     end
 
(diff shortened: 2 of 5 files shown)


lib/datadog/core/environment/process.rb
```diff
@@ -30,6 +30,26 @@ def self.serialized
           @serialized = tags.join(',').freeze
```
p-datadog: Can you change `serialized` to call the new method instead of duplicating all of the code please?

marcotc (approved):
Looks good, with @p-datadog's suggestion implemented.

lib/datadog/core/environment/process.rb
wantsui: Yes! I refactored in 4989577c1d12c9edba9eab55561e991b9e239b60 and changed some of the tests to account for that. Let me know if this change looks ok!

spec/datadog/core/environment/process_spec.rb
```diff
@@ -147,4 +148,107 @@ def reset_serialized!
...
+        allow(Dir).to receive(:pwd).and_return(pwd)
+        allow(File).to receive(:expand_path).and_call_original
+        allow(File).to receive(:expand_path).with('.').and_return('/app')
+        reset_memoized_variables!
+      end
+
+      after do
+        reset_memoized_variables!
+      end
+    end
```
marcotc: This shared context is the same as https://github.com/DataDog/dd-trace-rb/blob/18e2b12ebbf4dbcbc590e3739aedfd2c91ae26de/spec/datadog/core/environment/process_spec.rb#L14

and equivalent to: https://github.com/DataDog/dd-trace-rb/blob/18e2b12ebbf4dbcbc590e3739aedfd2c91ae26de/spec/datadog/core/remote/client_spec.rb#L45-L66
I suggest we share this context :)
The shared context is kind of complicated, so I think duplicating it will cause trouble in the future, in case of drift.

I recommend moving it to the top-level of this file, declaring it as `RSpec.shared_context 'with mocked process environment'`.

I can then be used in this file, and imported in `spec/datadog/core/remote/client_spec.rb`.

p-datadog: +1 except I think the shared context should go into `support` somewhere.

wantsui: Done in 38c6d02153684b16c676a5470d2010022d7c3f34 - the refactor using the support folder passes locally so I'm just waiting to see if there's some CI check I haven't thought of!

lib/datadog/core/environment/process.rb
```diff
@@ -14,6 +14,14 @@ module Process
         # @return [String] comma-separated normalized key:value pairs
         def self.serialized
           return @serialized if defined?(@serialized)
+
+          @serialized = tags_array.join(',').freeze
+        end
+
+        # This method returns an array in the format ["k1:v1","k2:v2","k3:v3"]
+        # @return [Array<String>] array of normalized key:value pairs
+        def self.tags_array
```
Strech: I think it's better to avoid using structures as a part of the name. It makes code fragile to refactoring. In addition in ruby we would just say - `tags` implying that it's a list of tags.

Strech (approved):

PR #5138, opened 2025-12-12
chore(otel): fix telemetry collection for otel metrics

mabdinur:
**What does this PR do?**

Ensures configurations used in OpenTelemetry Metrics SDK are reported to the instrumentation telemetry platform. Updates configuration fallback logic to properly detect when metrics-specific configs use default values and fall back to general OTLP exporter configs per OpenTelemetry spec.

**Motivation:**

Emit diagnostic data about the state of the OpenTelemetry Components.

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

Regression was caught by system tests. This update to system tests will be merged after this PR: https://github.com/DataDog/system-tests/pull/5811

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -77,7 +77,7 @@ jobs:
   test:
     needs:
       - build
-    uses: DataDog/system-tests/.github/workflows/system-tests.yml@43fcd80e6531f07b928b467e72cc89e9752042db  # Automated: This reference is automatically updated.
+    uses: DataDog/system-tests/.github/workflows/system-tests.yml@91b7434cee90262199f503010b27754aa9e38a29  # Automated: This reference is automatically updated.
     secrets:
       TEST_OPTIMIZATION_API_KEY: ${{ secrets.DD_API_KEY }}  # key used to pushed test results to test optim
       DD_API_KEY: ${{ secrets.DD_API_KEY }}  # key used in tests runs
@@ -91,7 +91,7 @@ jobs:
       desired_execution_time: 300  # 5 minutes
       scenarios_groups: tracer_release
       skip_empty_scenarios: true
-      ref: 43fcd80e6531f07b928b467e72cc89e9752042db  # Automated: This reference is automatically updated.
+      ref: 91b7434cee90262199f503010b27754aa9e38a29  # Automated: This reference is automatically updated.
       force_execute: ${{ needs.build.outputs.forced_tests }}
       parametric_job_count: 8
       push_to_test_optimization: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -146,7 +146,7 @@ requirements_json_test:
 
 configure_system_tests:
   variables:
-    SYSTEM_TESTS_REF: 43fcd80e6531f07b928b467e72cc89e9752042db # Automated: This reference is automatically updated.
+    SYSTEM_TESTS_REF: 91b7434cee90262199f503010b27754aa9e38a29 # Automated: This reference is automatically updated.
     SYSTEM_TESTS_SCENARIOS_GROUPS: "simple_onboarding,lib-injection,docker_ssi"
 
 save_versions:
--- a/vendor/rbs/opentelemetry-api/0/opentelemetry-api.rbs
+++ b/vendor/rbs/opentelemetry-api/0/opentelemetry-api.rbs
@@ -1,6 +1,6 @@
 module OpenTelemetry
   class Context
-    def self.current: () -> self
+    def self.current: () -> instance
     def ensure_trace: () -> Datadog::Tracing::TraceOperation?
     def set_values: (Hash[Symbol, Hash[String, String] | String]) -> self
   end


lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -159,6 +159,23 @@ def configuration(settings, agent_settings)
               get_telemetry_origin(settings, 'tracing.contrib.peer_service_mapping')
             )
 
+            # OpenTelemetry configuration options (using environment variable names)
+            list.push(
+              conf_value('OTEL_EXPORTER_OTLP_ENDPOINT', settings.opentelemetry.exporter.endpoint, seq_id, get_telemetry_origin(settings, 'opentelemetry.exporter.endpoint')),
+              conf_value('OTEL_EXPORTER_OTLP_HEADERS', DATADOG_ENV['OTEL_EXPORTER_OTLP_HEADERS'] || '', seq_id, get_telemetry_origin(settings, 'opentelemetry.exporter.headers')),
```
y9v: you can use `DATADOG_ENV.fetch('OTEL_EXPORTER_OTLP_HEADERS', '')` to provide a fallback value

lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -159,6 +159,23 @@ def configuration(settings, agent_settings)
...
+              conf_value('OTEL_EXPORTER_OTLP_PROTOCOL', settings.opentelemetry.exporter.protocol, seq_id, get_telemetry_origin(settings, 'opentelemetry.exporter.protocol')),
+              conf_value('OTEL_EXPORTER_OTLP_TIMEOUT', settings.opentelemetry.exporter.timeout_millis, seq_id, get_telemetry_origin(settings, 'opentelemetry.exporter.timeout_millis')),
+              conf_value('DD_METRICS_OTEL_ENABLED', settings.opentelemetry.metrics.enabled, seq_id, get_telemetry_origin(settings, 'opentelemetry.metrics.enabled')),
+              conf_value('OTEL_METRICS_EXPORTER', settings.opentelemetry.metrics.exporter, seq_id, get_telemetry_origin(settings, 'opentelemetry.metrics.exporter')),
+              conf_value('OTEL_EXPORTER_OTLP_METRICS_ENDPOINT', settings.opentelemetry.metrics.endpoint, seq_id, get_telemetry_origin(settings, 'opentelemetry.metrics.endpoint')),
+              conf_value('OTEL_EXPORTER_OTLP_METRICS_HEADERS', DATADOG_ENV['OTEL_EXPORTER_OTLP_METRICS_HEADERS'] || '', seq_id, get_telemetry_origin(settings, 'opentelemetry.metrics.headers')),
```
y9v: same here - `fetch` can be used

lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -159,6 +159,23 @@ def configuration(settings, agent_settings)
               get_telemetry_origin(settings, 'tracing.contrib.peer_service_mapping')
             )
 
+            # OpenTelemetry configuration options (using environment variable names)
+            list.push(
+              conf_value('OTEL_EXPORTER_OTLP_ENDPOINT', settings.opentelemetry.exporter.endpoint, seq_id, get_telemetry_origin(settings, 'opentelemetry.exporter.endpoint')),
```
y9v, line 166: is `settings.opentelemetry` always defined? And do we want to send all those configuration values if the customer is not using opentelemetry?

lib/datadog/opentelemetry/metrics.rb
```diff
@@ -105,6 +100,15 @@ def configure_otlp_exporter(provider)
       rescue LoadError => e
         @logger.warn("Could not load OTLP metrics exporter:  #{e.class}: #{e}")
       end
+
+      # Returns metrics config value if explicitly set, otherwise falls back to exporter config or computed default value.
+      def get_metrics_config_with_fallback(metrics_config, exporter_config, option_name, computed_default = nil)
```
y9v: it might be better to use keyword arguments here

lib/datadog/opentelemetry/metrics.rb
```diff
@@ -105,6 +100,15 @@ def configure_otlp_exporter(provider)
...
+        if metrics_config.using_default?(option_name)
+          exporter_config.send(option_name) || computed_default
```
y9v: it's a good practice to use `public_send` instead of `send`, unless you want to call private methods

y9v (changes requested):
This PR fixes a bug, but there is not a single test here, which verifies that the bug is fixed now. Would it be possible to add tests?

mabdinur:
> This PR fixes a bug, but there is not a single test here, which verifies that the bug is fixed now. Would it be possible to add tests?

I linked the the system tests for this change in the PR description. I can duplicate these tests in dd-trace-rb but I think that reduces the value of having shared tests. I prefer not writing the same test cases multiple times but I can make this change if it is required to merge this PR

y9v:
> > This PR fixes a bug, but there is not a single test here, which verifies that the bug is fixed now. Would it be possible to add tests?
> 
> I linked the the system tests for this change in the PR description. I can duplicate these tests in dd-trace-rb but I think that reduces the value of having shared tests. I prefer not writing the same test cases multiple times but I can make this change if it is required to merge this PR

Sorry, I missed the link to system-tests PR. It's up to you, but I would suggest to add at least one test here - simply because it makes it easier to spot a mistake early on in the future. If I am working on the library code, I run the tests we have, but not necessarily system tests (which are more like a contract).

lib/datadog/core/telemetry/event/app_started.rb
mabdinur: Yup settings.opentelemetry.* configurations should always be defined. These configurations are initialized when the core configuration object is created: https://github.com/DataDog/dd-trace-rb/blob/a827d0fc70c7edaa63740f7b093ebee6ac19f642/lib/datadog/core/configuration/settings.rb#L1046.

If a customer is not using otel then these env vars won't be set and will be filtered out here: https://github.com/DataDog/dd-trace-rb/blob/4641d8b42b7d7d6d4ad3ab99f58a57a87459f27e/lib/datadog/core/telemetry/event/app_started.rb#L236-L237.

lib/datadog/core/telemetry/event/app_started.rb
mabdinur: I'll just remove the default value. It's no longer needed

mabdinur:
```suggestion
              conf_value('OTEL_EXPORTER_OTLP_METRICS_HEADERS', DATADOG_ENV['OTEL_EXPORTER_OTLP_METRICS_HEADERS'], seq_id, get_telemetry_origin(settings, 'opentelemetry.metrics.headers')),
```

lib/datadog/core/telemetry/event/app_started.rb
y9v: great, makes sense, thanks for the explanation

lib/datadog/core/telemetry/event/app_started.rb
mabdinur:
```suggestion
              conf_value('OTEL_EXPORTER_OTLP_HEADERS', DATADOG_ENV['OTEL_EXPORTER_OTLP_HEADERS'], seq_id, get_telemetry_origin(settings, 'opentelemetry.exporter.headers')),
```

lib/datadog/opentelemetry/metrics.rb
mabdinur: Done, I also simplified the arguments we only need option_name, computed_default.

mabdinur:
> > > This PR fixes a bug, but there is not a single test here, which verifies that the bug is fixed now. Would it be possible to add tests?
> 
> > 
> 
> > I linked the the system tests for this change in the PR description. I can duplicate these tests in dd-trace-rb but I think that reduces the value of having shared tests. I prefer not writing the same test cases multiple times but I can make this change if it is required to merge this PR
> 
> 
> 
> Sorry, I missed the link to system-tests PR. It's up to you, but I would suggest to add at least one test here - simply because it makes it easier to spot a mistake early on in the future. If I am working on the library code, I run the tests we have, but not necessarily system tests (which are more like a contract).

Yup, I can do that. This change only impacts instrumentation telemetry. I'll add a test cases that captures all supported configs and the success metrics

lib/datadog/core/telemetry/event/app_started.rb
```diff
@@ -159,6 +159,23 @@ def configuration(settings, agent_settings)
...
+              conf_value('OTEL_EXPORTER_OTLP_HEADERS', DATADOG_ENV['OTEL_EXPORTER_OTLP_HEADERS'], seq_id, 'env_var'),
```
p-datadog: We do have a setting that reads this environment variable and the value goes through a parser that performs validation and may drop some of the data. Since we send values from our configuration, reading user input here seems to be inconsistent. Is it really a lot of work to serialize a hash of our settings here? We have for example what looks to be very similar logic above on line 153.

Same comment for line 171 below.

lib/datadog/opentelemetry/sdk/metrics_exporter.rb
```diff
@@ -12,7 +12,7 @@ class MetricsExporter < ::OpenTelemetry::Exporter::OTLP::Metrics::MetricsExporte
 
         def initialize(endpoint:, timeout:, headers:, protocol:)
           super(endpoint: endpoint, timeout: timeout, headers: headers)
-          @telemetry_tags = {'protocol' => protocol, 'encoding' => 'protobuf'}
+          @telemetry_tags = {'protocol' => "http", 'encoding' => 'protobuf'}
```
p-datadog: The value of `protocol` is now ignored. Should this validate that `protocol` is `http` and raise/warn if a different protocol was requested? Is this customer-callable code? If not, should the signature be changed to no longer take `protocol`?

lib/datadog/opentelemetry/configuration/settings.rb
```diff
@@ -140,13 +140,13 @@ def self.add_settings!(base)
...
                   o.env 'OTEL_EXPORTER_OTLP_METRICS_TIMEOUT'
-                  o.default nil
+                  o.default 10000
                 end
 
                 option :protocol do |o|
                   o.type :string, nilable: true
                   o.env 'OTEL_EXPORTER_OTLP_METRICS_PROTOCOL'
-                  o.default nil
+                  o.default "http/protobuf"
```
p-datadog, line 149: Are these changes really not customer-visible? What would happen if customers explicitly set the value of these options to `nil`, which is presently (i.e. in our most recent releases) allowed?

p-datadog:
@mabdinur Can you please adjust the changelog section of the PR to be compliant with the requirements stated in the first comment above?

lib/datadog/core/telemetry/event/app_started.rb
mabdinur: No, this is just to be compatible with existing system tests. I will revert this change and just update the test.

lib/datadog/opentelemetry/sdk/metrics_exporter.rb
mabdinur: protocol is always http (grpc export is not yet supported in the opentelemetry sdk). I will update this class up to make this explicit and add a comment to where the exporter is initialized.

lib/datadog/opentelemetry/configuration/settings.rb
mabdinur: `normalize_protocol` setter actually rejects all values that are not `http/protobuf`. It logs a warning stating that only  `http/protobuf` is supported. Even if the value is manually set nil we will always default to this value. 

In a future PR we will add support for `http/json` and `grpc` but this isn't a priority for this quarter.

supported-configurations.json
vpellan: You should create a version B with default to `http/protobuf` on the FPD
```suggestion
"version": "B",
"type": "string",
"default": "http/protobuf"
```

supported-configurations.json
```diff
@@ -2183,9 +2183,9 @@
     ],
     "OTEL_EXPORTER_OTLP_HEADERS": [
       {
-        "version": "A",
+        "version": "B",
         "type": "map",
-        "default": ""
+        "default": null
```
vpellan: Why this change ? Looking at the settings.rb file, `OTEL_EXPORTER_OTLP_HEADERS` default is an empty hash, not nil, so `""` default seems correct.

sig/datadog/opentelemetry/sdk/metrics_exporter.rbs
```diff
@@ -10,7 +10,11 @@ module Datadog
 
         METRIC_EXPORT_FAILURES: "otel.metrics_export_failures"
 
-        def initialize: (endpoint: ::String, timeout: ::Float, headers: ::Hash[::String, any], protocol: ::String) -> void
+        TELEMETRY_NAMESPACE: 'tracers'
+
+        TELEMETRY_TAGS: {'protocol' => "http", 'encoding' => 'protobuf'}
+
+        def initialize: (endpoint: ::String, timeout: ::Float, headers: ::Hash[::String, any]) -> void
```
vpellan: This isn't defined anymore ?

mabdinur: Nope. This method has the same signature as the parent. I'll remove it

mabdinur:
```suggestion
```

supported-configurations.json
mabdinur: I mixed this up with OTEL_EXPORTER_OTLP_METRICS_HEADERS, I will revert this change

PR #5142, opened 2025-12-14
Transports: remove Spec classes, non-core Instance classes

p-datadog:
**What does this PR do?**

Continues the cleanup of transports by:
/ Removing all Spec classes
/ Removing all non-core Instance classes
/ Removing the two associated exception classes
/ Removing `send_request_impl` (as well as the layer of indirection that was going through Spec)
/ Removing `api_instance_class` since there is now only one Instance class in the codebase
/ Renames some variables to match their class types: api/instance/endpoint

**Motivation:**

Continuing streamlining of transport code

**Change log entry**
None

**Additional Notes:**

OpenFeature transport was made different in https://github.com/DataDog/dd-trace-rb/pull/5047. I didn't change it other than to keep the code working. As such it retains Spec and Instance classes.

This PR addresses several of the comments in https://github.com/DataDog/dd-trace-rb/pull/5099, in particular @vpellan 's comment about renaming Spec send_XXX methods to send_request started me on most of the code deletion in this PR.

**How to test the change?**
Existing tests


marcotc (approved):
Beautiful clean up!

lib/datadog/open_feature/transport.rb
```diff
@@ -45,7 +48,6 @@ def send_exposures(env)
 
         def self.build(agent_settings:, logger:)
           Core::Transport::HTTP.build(
-            api_instance_class: HTTP::Instance,
```
p-datadog, line 48: This breaks openfeature because the transport logic is in the Instance class defined above but now there is no way to pass that to the core transport code to actually use.

PR #5145, opened 2025-12-15
Remove rbs file from `lib/datadog/opentelemetry/api`

y9v:
**What does this PR do?**
It removes a `.rbs` file from `lib` folder - as this causes an RDoc error when installing the gem.

```
arsing documentation for datadog-2.23.0
Before reporting this, could you check that the file you're documenting
has proper syntax:

  ~/.rbenv/versions/3.2.8/bin/ruby -c lib/datadog/opentelemetry/api/baggage.rbs

RDoc is not a full Ruby parser and will fail when fed invalid ruby programs.

The internal error was:

        (ArgumentError) wrong number of arguments (given 0, expected 1)

ERROR:  While executing gem ... (ArgumentError)
    wrong number of arguments (ArgumentError)
...
```

**Motivation:**
We need to fix this.

**Change log entry**
Yes. Fix RDoc error when installing the `datadog` gem.

**Additional Notes:**
None.

**How to test the change?**
Manually.


y9v:
@ivoanjo good point, but I will do it in a separate PR - there are steep errors when using the file from `lib` folder, and they actually point at real problems with the code

PR #5151, opened 2025-12-15
Transports: clean up types

p-datadog:
**What does this PR do?**

Cleans up and improves typing of transport code.

It also repairs some outdated (incorrect now) type definitions that went out of sync with code after the changes in https://github.com/DataDog/dd-trace-rb/pull/5142.

It also moves `encoding` to the base `Endpoint` class. I think conceptually this is correct - if we support encoding specification in an endpoint in one transport, it should be a core feature available to (used by?) all transports.

**Motivation:**

Making type changes in https://github.com/DataDog/dd-trace-rb/pull/5150  I noticed the transport type definitions were outdated.

**Change log entry**
None

**Additional Notes:**
I disabled steep in a couple of places because I don't have bandwidth to research how to make it work there. 

**How to test the change?**
Existing CI

### DIFF
--- a/sig/datadog/open_feature/transport.rbs
+++ b/sig/datadog/open_feature/transport.rbs
@@ -11,18 +11,12 @@ module Datadog
         class Spec
           @endpoint: Core::Transport::HTTP::API::Endpoint
 
-          attr_accessor endpoint: Core::Transport::HTTP::API::Endpoint
-
           def initialize: () -> void
 
           def call: (Core::Transport::HTTP::Env) { (Core::Transport::HTTP::Env) -> Core::Transport::Response } -> Core::Transport::Response
         end
 
-        class Instance < Core::Transport::HTTP::API::Instance
-          def send_exposures: (Core::Transport::HTTP::Env) -> Core::Transport::Response
-        end
-
-        @api: HTTP::Instance
+        @api: Core::Transport::HTTP::API::Instance
 
         @logger: Core::Logger
 


lib/datadog/tracing/transport/traces.rb
```diff
@@ -165,7 +165,7 @@ def send_traces(traces)
           end
 
           def stats
-            @client.stats
+            client.stats
```
Strech, line 168: To be honest, I think this is unnecessary change, and I personally consider reader methods a public interface for outside and `@`-methods to be used internally

p-datadog: steep refuses to work with the instance variable, the type of `client` changes for this transport because there is a tracing-specific Client class. It's either this or steep-ignoring this line.

Strech: Doesn't that mean that we should define client as union of types it can take shape of?

p-datadog: I have to research union types and shapes to do so. So far when I encountered what look like union types the diagnostics from steep have been not very helpful. I definitely do not have time to research this and make work prior to the next release.

p-datadog: The other consideration here is  I currently still have motivation to continue refactoring transport code and the typing contortions may not be needed long term because perhaps there should only be a single Client class. https://github.com/DataDog/dd-trace-rb/pull/5142 for example removed the need to type `api_instance_class` parameters because the respective code was deleted.

Strech: Maybe a quick suggestion, just as a quick try or fail, if it will not work, fine to change it, we can fix it later

```
@client: Type1 | Type2 | Type3
```

p-datadog: steep is not permitting a change of the instance variable type. I currently set `client` method to be the tracing client in the tracing transport. Where `@client` may be changed is in the core transport, but then we'd have a core type referencing a tracing type which is at least not great, maybe it's debatable whether that is worse than changing from instance variable to attribute method.

I checked the differences between tracing client and core client and it's collecting counts of failed responses and emitting them via telemetry. An argument can be made that this functionality should be applied to all transports, perhaps always/automatically. Certainly sounds like a useful feature for DI/LD. If this is done, the code can be moved to the core client and then there won't be any need for tracing to have a custom client.

Strech: I'm going to resolve it to unblock you

p-datadog:
This now also removes the open feature Instance class from the rbs file - the class was removed in https://github.com/DataDog/dd-trace-rb/pull/5158.

marcotc (approved):

PR #5153, opened 2025-12-15
DI: switch duration calculation to using Process.clock_gettime directly

p-datadog:
**What does this PR do?**

Tries to fix the flaky test for reported method-under-test duration in DI: https://github.com/DataDog/ruby-guild/issues/278

This version uses `Process.clock_gettime` directly because DI shouldn't be calling the customer-overridden time provider. Maybe there is some leaking time provider overrides that make this test fail. It is suspicious how the reported duration is very large and also does not seem to change the overall test runtime (see comments in the guild issue).

**Motivation:**

Flaky tests

**Change log entry**
Yes: dynamic instrumentation: do not use customer-provided time provider when calculating method duration

**Additional Notes:**

**How to test the change?**
Test modified to be more strict, let's see if this works


ivoanjo (approved):
👍 I'd be very surprised if this fixes the flakiness issue, but it's also harmless to try so this is a "why not?" for me

Strech (approved):

p-datadog:
If the reported runtime of 26/40 seconds is accurate and not just some sort of test state leakage from overridden time provider elsewhere, I don't know how this is even possible and what to do about it. I could set the constraints on execution time that it's more than zero and less than infinity, but now I am wondering if there are really trivial Ruby methods that all of a sudden take 1000x longer to execute for no apparent reason.

PR #5159, opened 2025-12-16
DEBUG-4548 Telemetry: send events in forked children + telemetry metrics reset after fork

p-datadog:
Worker race fix (described in https://github.com/DataDog/ruby-guild/issues/279 and originally reported as performance regression) has been extracted into https://github.com/DataDog/dd-trace-rb/pull/5176.

**What does this PR do?**

Restores https://github.com/DataDog/dd-trace-rb/pull/5074 and implements telemetry metrics reset after fork, which was causing wrong metric counts in datadog-ci end to end tests.

The metrics reset is done via the "at fork monkey patch". Existing worker race fix (in #5176) used the worker after-fork treatment - this PR consolidates after fork logic into the monkey patch handler.

**Motivation:**

**Change log entry**
Yes: fix Live Debugger / Dynamic Instrumentation UI for forking web servers

**Additional Notes:**

The original PR and the race fix are in separate commits for ease of review.

This PR now uses "at fork monkey patch" to reset the metrics manager after fork. The added tests caused existing tests to start failing due to shared global state; this was fixed in https://github.com/DataDog/dd-trace-rb/pull/5175.

We still need to mark the worker as "restarting after fork", but the restart is triggered from the monkey patch callback and not from enqueue call like before.

**How to test the change?**
New integration tests added. They have been aggravatingly flaky in CI with the previous implementation that used both worker after fork logic and at fork monkey patch, and they *seem* much less flaky with the current implementation that does everything from the monkey patch.

Additionally I manually tested against @anmarchenko 's reproducer.

### DIFF
--- a/lib/datadog/core/telemetry/worker.rb
+++ b/lib/datadog/core/telemetry/worker.rb
@@ -266,7 +266,7 @@ module Datadog
             # 2) don't want to keep a permanent reference to in any case.
             # Therefore we have this +reset!+ method that changes the
             # event type while keeping the payload.
-            @initial_event.reset! # steep:ignore
+            @initial_event.reset! # steep:ignore NoMethod
           end
 
           if enabled? && !worker.nil?


p-datadog:
Worker race fix has been extracted in https://github.com/DataDog/dd-trace-rb/pull/5176, this PR will be rebased on top of https://github.com/DataDog/dd-trace-rb/pull/5176.

spec/support/platform_helpers.rb
```diff
@@ -85,5 +85,21 @@ def skip_any_instance_on_buggy_jruby
...
+      end
+    end
+
+    def forking_platform_only
+      if PlatformHelpers.jruby?
+        before(:all) do
+          skip "Test requires fork to be implemented, JRuby does not"
+        end
+      end
+    end
```
Strech, line 103: Does it make sense to replace with this helper type instead? https://github.com/DataDog/dd-trace-rb/blob/8b3619d5086188ff6b6c4764d36e9287c537daff/spec/spec_helper.rb#L115-L142

_P.S maybe we can add it to `before(:all)`_

lib/datadog/core/telemetry/worker.rb
```diff
@@ -220,6 +238,45 @@ def disable_on_not_found!(response)
...
+          # AppStarted event.
+          if @initial_event.is_a?(Event::SynthAppClientConfigurationChange)
+            # It would be great to just replace the initial event in
+            # +initialize_state+ method. Unfortunately this event requires
+            # the entire component tree to build its payload, which we
+            # 1) don't currently have in telemetry and
+            # 2) don't want to keep a permanent reference to in any case.
+            # Therefore we have this +reset!+ method that changes the
+            # event type while keeping the payload.
+            @initial_event.reset! # steep:ignore
```
vpellan: Could you please narrow the `steep:ignore` to the actual diagnostic that was causing an error, and if possible add a comment or a link to the RBS or Steep issue ? This might be a type narrowing issue here

sig/datadog/core/telemetry/event/synth_app_client_configuration_change.rbs
```diff
@@ -3,9 +3,13 @@ module Datadog
     module Telemetry
       module Event
         class SynthAppClientConfigurationChange < AppStarted
-          def type: () -> "app-client-configuration-change"
+          def type: -> ("app-client-configuration-change" | "app-started")
```
vpellan, line 6:
```suggestion
          def type: () -> ("app-client-configuration-change" | "app-started")
```

sig/datadog/core/utils/only_once_successful.rbs
```diff
@@ -8,15 +8,15 @@ module Datadog
...
-        def failed?: () -> bool
+        def failed?: -> bool
 
         private
 
-        def check_limit!: () -> void
+        def check_limit!: -> void
 
-        def limited?: () -> bool
+        def limited?: -> bool
```
vpellan, line 19: Why these changes?

spec/datadog/core/telemetry/metrics_integration_spec.rb
```diff
@@ -0,0 +1,190 @@
...
+  end
+
+  let(:response) do
+    double(Datadog::Core::Transport::HTTP::Response,
+      ok?: true,)
+  end
+
+  let(:events) { [] }
+
+  def assert_events(events)
```
vpellan, line 77: You should be able to use a `shared_context` instead and use it with `include_context`

spec/datadog/core/telemetry/worker_forking_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+      before do
+        # Reduce interval between event submissions in worker
+        # to make the test run faster.
+        expect(component.worker).to receive(:loop_wait_time).at_least(:once).and_return(1)
+      end
+
+      # Behavior in the child should be the same regardless of what
+      # was sent in the parent, because the child is a new application
+      # (process) from the backend's perspective.
+      def fork_and_assert
```
vpellan, line 171: This could probably also be in a `shared_context` block

sig/datadog/core/utils/only_once_successful.rbs
p-datadog: We omit empty parentheses in the method definitions so I actually am curious why they are written in the sig files.

lib/datadog/core/telemetry/worker.rb
p-datadog: I added the specific complaint. For linking to steep issues, I don't know which exact issue it is and none of my own functionality bug reports have been addressed (https://github.com/soutaro/steep/issues?q=sort%3Aupdated-desc%20is%3Aissue%20is%3Aopen%20author%3Ap-datadog) so frankly I don't see the point of identifying the exact steep bug anyway.

spec/support/platform_helpers.rb
p-datadog: I could do it but let's discuss with the guild so that we are all on the same page with the preferred style (I added to the agenda).

p-datadog:
The tests I added are still flaking in CI. The most recent failure was on ruby 2.5 and I have seen previous failures on 2.5 also. I spent ~3 weeks trying to figure out the root cause and the problem is that once I start adding diagnostics the failures disappear. So, after 3 weeks, I am skipping the tests on 2.5 and if they fail on anything below 3.0 I will skip all of those versions also.

lib/datadog/core/telemetry/worker.rb
vpellan: Soutaro seems busy with rbs-inline but I hope he will continue development of steep and look at those issues

PR #4838, opened 2025-08-05
DEBUG-4090 DI: probe file support

p-datadog:
**What does this PR do?**

Adds ability to instrument the application by specifying probes in a JSON file.

This functionality has been implemented in the Java tracer for a while now.

**Motivation:**
Feature parity across languages

**Change log entry**
Yes: dynamic instrumentation/live debugger: support `DD_DYNAMIC_INSTRUMENTATION_PROBE_FILE` environment variable for specifying probes in a file

**Additional Notes:**

**How to test the change?**
Unit and integration tests added

PR #5160, opened 2025-12-17
DEBUG-3499 RC: remove StringIO layer 

p-datadog:
**What does this PR do?**

Removes wrapping of content with StringIO instance in remote config.

As a side effect, I think these changes also remove a copy of all of the content being done for JSON parsing and another copy being made for sha1 calculation.

After fixing test failures, I now also remove ReadError since the RC content now cannot fail to read. The code copy-pasted in each of the remotes that read out of the stringio also became a trivial string reference.

**Motivation:**
Reduce the amount of code in our library and make remote config code easier to reason about.

**Change log entry**
None

**Additional Notes:**

The only place where the stringio is used appears to be the sha1 digest, but the digest can equally well take an arbitrarily large string as an argument and even works correctly for string lengths under 100 MB (it starts to disagree with openssl's idea of how sha1 should be calculated for strings in the range of half a gigabyte).

Additionally, the StringIO instance is rewound after being read, this code is also deleted.

**How to test the change?**

Existing CI

### DIFF
--- a/lib/datadog/appsec/remote.rb
+++ b/lib/datadog/appsec/remote.rb
@@ -7,8 +7,6 @@ module Datadog
   module AppSec
     # Remote
     module Remote
-      class ReadError < StandardError; end
-
       class NoRulesError < StandardError; end
 
       class << self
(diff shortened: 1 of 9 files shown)


p-datadog:
Interesting. The entire DI is built on top of RC and we implemented our own mechanism for reading the configuration from files - https://github.com/DataDog/dd-trace-rb/pull/4838.

PR #5161, opened 2025-12-17
DI: set test duration upper bound to 1000 seconds

p-datadog:
**What does this PR do?**

Makes DI duration assertion upper bound Big Enough For Everyone (TM), including our CI. This PR now requires that Ruby figures out whether the length of the array of defined methods is less than 42 in less than 1000 seconds.

**Motivation:**
The test in question failed again in https://github.com/DataDog/dd-trace-rb/actions/runs/20288698696/job/58268221593?pr=5159

**Change log entry**
None

**Additional Notes:**

See https://github.com/DataDog/ruby-guild/issues/278 for the history of me trying to retain this check in minimally useful form.

**How to test the change?**
Existing CI

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -77,7 +77,7 @@ jobs:
   test:
     needs:
       - build
-    uses: DataDog/system-tests/.github/workflows/system-tests.yml@72f8bf07e673066fc952bd7c4d1ab37ead0bccd1  # Automated: This reference is automatically updated.
+    uses: DataDog/system-tests/.github/workflows/system-tests.yml@0032fc1c262ec50dc0858d106d038858a266865c  # Automated: This reference is automatically updated.
     secrets:
       TEST_OPTIMIZATION_API_KEY: ${{ secrets.DD_API_KEY }}  # key used to pushed test results to test optim
       DD_API_KEY: ${{ secrets.DD_API_KEY }}  # key used in tests runs
@@ -91,7 +91,7 @@ jobs:
       desired_execution_time: 300  # 5 minutes
       scenarios_groups: tracer_release
       skip_empty_scenarios: true
-      ref: 72f8bf07e673066fc952bd7c4d1ab37ead0bccd1  # Automated: This reference is automatically updated.
+      ref: 0032fc1c262ec50dc0858d106d038858a266865c  # Automated: This reference is automatically updated.
       force_execute: ${{ needs.build.outputs.forced_tests }}
       parametric_job_count: 8
       push_to_test_optimization: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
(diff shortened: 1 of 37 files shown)


spec/datadog/di/integration/instrumentation_spec.rb
```diff
@@ -595,10 +595,12 @@ def run_test
...
               # instead of using our helper which could invoke customer code
               # and also be mocked.
-              # Go back to requiring the runtime to be under one second.
               # The reported duration in local test runs is about 0.03 seconds.
               expect(value).to be > 0
-              expect(value).to be < 1
+              # Set upper bound at 1000 seconds... should be safe given the
+              # highest value seen so far was 40 seconds (for a method that
+              # compares length of an array with an integer).
+              expect(value).to be < 1000
```
ivoanjo, line 603: I wonder if we should change the approach here. That is, what about the spec code measuring the duration itself and then asserting that the duration is <= + within a delta of `value`?

Also -- shouldn't the `expect(value).to be < 1` be removed? The comments are a bit confusing too, this seems like a mix of two versions

p-datadog: The comparison with 1 is no longer in the test.

>That is, what about the spec code measuring the duration itself and then asserting that the duration is <= + within a delta of value?

I am not hopeful that this will fix it.

I also don't like the idea of doing math in the test, I prefer the tests to be more straightforward.

What I would prefer to do instead is either use https://github.com/NoRedInk/rspec-retry with a fixed number of attempts (say, 3) or implement a custom retrier which would demand that, let's say, out of 5 executions, 4 are less than 1 second in duration.

ivoanjo: I think we're mixing two things up here: performance and correctness. My suggestion would be to focus the test on correctness and separately add benchmarking.

Our fancy custom benchmarking setups exist exactly to make sure benchmarks run in a controlled environment where "oops 10s performance hiccup" doesn't happen. I'm not sure it's worth trying to bend CI to our will rather than doing that?

(As an example, the profiling tests that do use timing are testing correctness -- profiler takes x samples and sleeps between them within y period -- rather than "how fast is sampling going")

marcotc: To paraphrase @ivoanjo's suggestion, if we are checking for a valid `@duration`, all we care about is that it's a number, and greater than 0.

If we want to check that `@duration` captures the accurate timing, then we can do something like `def long_test_method; sleep 0.01; 42;end` and assert that `@duration > 0.01` and probably `@duration < 1`. I think this is a separate test from "duration is a number and greater than zero".

On a related note, maybe the line `Object.methods.length` might be adding extra variance to the test timing, maybe something related to VM internals like constant caching. I suggest using `sleep(0)` if you simply just want `@duration` to never be zero.

p-datadog: If the current version of the test fails CI I will remove the upper bound completely.

PR #5162, opened 2025-12-17
DEBUG-3499 RC: inject dependencies

p-datadog:
**What does this PR do?**

Injects dependencies into remote config:
/ `logger` arguments are no longer defaulted (the default was added by me for safety due to lack of understanding of how our codebase works, given that we can change RC code it's not necessary)
/ `settings` are now injected instead of being read globally

**Motivation:**
Making RC more understandable

**Change log entry**
None

**Additional Notes:**

**How to test the change?**

Existing CI


ivoanjo (approved):
👍 LGTM

p-datadog:
This PR broke datadog-ci because apparently the latter does not pass in the logger explicitly. I reviewed the diff in this PR and don't see a test change with respect to the logger parameter now being passed in, leading me to think that the datadog-ci usage is not tested in our repo.

PR #5163, opened 2025-12-17
Transports: truncate http response bodies in inspections

p-datadog:
**What does this PR do?**

Truncates HTTP response bodies when inspecting HTTP responses

**Motivation:**
Investigation of https://github.com/DataDog/ruby-guild/issues/280.

Currently the full request body is dumped into the log. The request body can be several MB long which makes the entire log unreadable.

**Change log entry**
None

**Additional Notes:**

**How to test the change?**

Tests added


lib/datadog/core/transport/response.rb
```diff
@@ -37,6 +37,15 @@ def inspect
           maybe_code = if respond_to?(:code)
             " code:#{code}," # steep:ignore
           end
+          payload = self.payload
+          # Truncation thresholds are arbitrary but we need to truncate the
+          # payload here because outputting multi-MB request body to the
+          # log is not useful.
+          #
+          # Note that payload can be nil here.
+          if payload && payload.length > 2000
```
p-datadog: I had the limit originally set to 5000 but I think that was too large. Reduced to 2000 and I have paused the investigation in https://github.com/DataDog/ruby-guild/issues/280 - I may change the limit further downwards in the future.

ivoanjo (approved):
👍 LGTM

PR #5164, opened 2025-12-18
Transports: remove api_version

p-datadog:
**What does this PR do?**

Removes `api_version` from transports.

There are actually four different meanings of "api version";

1. The one that this PR is removing. I think this concept dates back to before we had the current API::Map and Endpoint classes, and this is how different endpoints were selected. In master this api version is never passed in explicitly and is always defaulted to nil. As far as I can tell it doesn't do anything.
2. Telemetry has a parameter called "api version". This is actually sent in the payloads, and is presently set to v2 in telemetry itself as a default. External to telemetry callers never provide a different api version. Given our transport architecture I don't think it would work to change the endpoint by just providing a different version externally to a component like telemetry.
3. Tracing has a parameter called "api version". I think it's conceptually the same thing as the first type of "api version" and may be the precursor to "api version" being in core component code. This one is in the public API, therefore I retained the `transport_options` argument in one place and added notes that it's now ignored. The transport code in master already does not use `transport_options` or the api version provided therein, as far as I can tell.
4. There are one or two places that refer to our existing endpoints as "api versions". I think this conceptually goes with the first usage again but now "api version" is an Endpoint/Instance(?) object instead of a string, I am very confused what the actual type of value is supposed to be passed in. Luckily as far as I can tell thus parameter is never actually used so it doesn't matter.

**Motivation:**
Continuing cleanup of transport code

**Change log entry**
None

**Additional Notes:**

**How to test the change?**

Existing CI

### DIFF
--- a/spec/datadog/tracing/sync_writer_spec.rb
+++ b/spec/datadog/tracing/sync_writer_spec.rb
@@ -39,6 +39,19 @@ RSpec.describe Datadog::Tracing::SyncWriter do
 
       it { is_expected.to have_attributes(transport: transport) }
     end
+
+    context 'when transport options include headers' do
+      let(:agent_settings) { instance_double(Datadog::Core::Configuration::AgentSettings) }
+      let(:options) { {agent_settings: agent_settings, logger: logger, transport_options: transport_options} }
+
+      let(:transport_options) { {headers: {foo: 'bar'}} }
+
+      it 'passes the headers into transport' do
+        expect(sync_writer.transport.apis.length).to eq 2
+        expect(sync_writer.transport.apis['v0.4'].headers).to include(foo: 'bar')
+        expect(sync_writer.transport.apis['v0.3'].headers).to include(foo: 'bar')
+      end
+    end
   end
 
   describe '#write' do
--- a/spec/datadog/tracing/workers/trace_writer_spec.rb
+++ b/spec/datadog/tracing/workers/trace_writer_spec.rb
@@ -32,15 +32,27 @@ RSpec.describe Datadog::Tracing::Workers::TraceWriter do
     context 'given :transport_options' do
       let(:options) { {transport_options: transport_options} }
 
-      let(:transport_options) { {api_version: 42} }
+      let(:transport_options) { {headers: {foo: 'bar'}} }
 
       before do
         expect(Datadog::Tracing::Transport::HTTP).to receive(:default)
-          .with(transport_options.merge(agent_settings: test_agent_settings, logger: Datadog.logger))
+          .with(transport_options.merge(agent_settings: test_agent_settings, logger: logger))
           .and_return(transport)
       end
 
       it { expect(writer.transport).to be transport }
+    end
+
+    context 'when transport options include headers' do
+      let(:options) { {transport_options: transport_options} }
+
+      let(:transport_options) { {headers: {foo: 'bar'}} }
+
+      it 'passes the headers into transport' do
+        expect(writer.transport.apis.length).to eq 2
+        expect(writer.transport.apis['v0.4'].headers).to include(foo: 'bar')
+        expect(writer.transport.apis['v0.3'].headers).to include(foo: 'bar')
+      end
     end
 
     context 'given :agent_settings' do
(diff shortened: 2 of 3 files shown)


lib/datadog/tracing/workers/trace_writer.rb
```diff
@@ -21,15 +21,19 @@ class TraceWriter < Core::Worker
           :transport,
           :agent_settings
 
+        # This method used to have the +:transport_options+ option, but
+        # it now does nothing and is ignored (but can still be passed in).
+        #
+        # DEV-3.0: change to keyword arguments?
+        #
```
vpellan: What about the headers in transport_options ? If they're unused can't we also remove them ?

spec/datadog/tracing/component_spec.rb
```diff
@@ -358,8 +358,8 @@
             context 'and :async' do
               context 'is set' do
                 let(:writer) { Datadog::Tracing::Writer.new(agent_settings: test_agent_settings) }
-                let(:writer_options) { {transport_options: :bar} }
-                let(:writer_options_test_mode) { {transport_options: :baz} }
+                let(:writer_options) { {foo: :bar} }
+                let(:writer_options_test_mode) { {foo: :baz} }
```
vpellan, line 362: if `transport_options` do nothing now, there's no reason to change that test? It would still test the backward compatibility?

p-datadog: The reason I changed it is that "transport_options" imply that some sort of transport options can be configured, but this is not the case - I actually don't know what options are valid, my feeling is none at all. I changed to "foo" to indicate that the actual option is not meaningful.

lib/datadog/tracing/workers/trace_writer.rb
p-datadog: I found one caller that uses them - DSM transport: https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/data_streams/transport/http.rb#L23-L26.  Thus the headers argument is not trivially removable but we should probably standardize on one mechanism of passing them in, perhaps integrated with the encoding mechanism.

p-datadog:
Following up on discussion with @marco I have done a manual test passing headers via `transport_options` and this seems to work in master, despite having no test coverage in our library:
```
as=Datadog::Core::Configuration::AgentSettingsResolver.call(Datadog.configuration)
writer=Datadog::Tracing::Writer.new(agent_settings:as,transport_options:{headers:{foo: 'bar'}})
writer.transport
```
```
#<Datadog::Tracing::Transport::Traces::Transport:0x00007afc1b36b1d0
 @apis=
  {"v0.4"=>
    #<Datadog::Core::Transport::HTTP::API::Instance:0x00007afc1bd2a7e8
     @adapter=
      #<Datadog::Core::Transport::HTTP::Adapters::Net:0x00007afc1b36b310
       @hostname="127.0.0.1",
       @port=8126,
       @ssl=false,
       @timeout=30>,
     @endpoint=
      #<Datadog::Tracing::Transport::HTTP::Traces::API::Endpoint:0x00007afc1b36b2c0
       @encoder=Datadog::Core::Encoding::MsgpackEncoder,
       @path="/v0.4/traces",
       @service_rates=true,
       @verb=:post>,
     @headers=
      {"Datadog-Client-Computed-Top-Level"=>"1",
       "Datadog-Meta-Lang"=>"ruby",
       "Datadog-Meta-Lang-Version"=>"3.3.4",
       "Datadog-Meta-Lang-Interpreter"=>"ruby-x86_64-linux",
       "Datadog-Meta-Lang-Interpreter-Vendor"=>"ruby",
       "Datadog-Meta-Tracer-Version"=>"2.24.0",
       :foo=>"bar"}>,
   "v0.3"=>
    #<Datadog::Core::Transport::HTTP::API::Instance:0x00007afc1bd2a798
     @adapter=
      #<Datadog::Core::Transport::HTTP::Adapters::Net:0x00007afc1b36b310
       @hostname="127.0.0.1",
       @port=8126,
       @ssl=false,
       @timeout=30>,
     @endpoint=
      #<Datadog::Tracing::Transport::HTTP::Traces::API::Endpoint:0x00007afc1b36b220
       @encoder=Datadog::Core::Encoding::MsgpackEncoder,
:
```

I will reinstate `transport_options` and add a test passing `headers` through it.

p-datadog:
We identified https://github.com/DataDog/datadog-lambda-rb/blob/main/lib/datadog/lambda.rb as a consumer of our tracing Writer classes. This consumer does not pass `transport_options` at all and should thus not be affected by the changes in this PR.

PR #5167, opened 2025-12-18
DEBUG-3499 RC: add diagnostics for invalid values

p-datadog:
**What does this PR do?**

Adds parameter validation to some methods in remote config.

**Motivation:**
When developing tests, such as in https://github.com/DataDog/dd-trace-rb/pull/5165, the changes in this PR helped reduce the time to understand where my data is wrong.

**Change log entry**
None

**Additional Notes:**

I originally had the contents of this PR in https://github.com/DataDog/dd-trace-rb/pull/5165 but, since it technically changes RC, I made it into a separate PR that can be merged after the release we are pushing for at the moment.

**How to test the change?**

Existing CI

### DIFF
--- a/lib/datadog/core/remote/configuration/content.rb
+++ b/lib/datadog/core/remote/configuration/content.rb
@@ -30,6 +30,9 @@ module Datadog
               # when developing tests, for example.
               raise ArgumentError, 'data must not be nil'
             end
+            unless String === data
+              raise ArgumentError, "Invalid type for data: #{data.class}: expected String"
+            end
 
             @path = path
             @data = data


lib/datadog/core/remote/configuration/digest.rb
```diff
@@ -26,6 +26,19 @@ class InvalidHashTypeError < StandardError; end
 
           class << self
             def hexdigest(type, data)
+              unless String === data
```
vpellan, line 29: Is using `===` over `is_a?` just a matter of preference ? Also, Is it possible to do this check in `lib/datadog/core/remote/configuration/content.rb`? Or can we have a Content instance with data that isn't a string ?

vpellan (approved):
LGTM

lib/datadog/core/remote/configuration/digest.rb
p-datadog: I actually used to hate `===` because its behavior is not changeable (the receiver is `String`, short of patching core classes `===` always works the same) and it inverts the order of operands conceptually, but now I use it mostly because I don't need to use the shift key to type it. I can change to `is_a?` if you prefer.

For the other two questions, I can probably add more validation and I don't think other types *should* be happening in our codebase.

marcotc: IMO, leave the `===` vs `is_a?` discussion to the linter.

p-datadog: I had a non-nil check in Content, added a requirement for data to be String there.

PR #5168, opened 2025-12-18
DEBUG-3499 DI: fix accounting when intrumenting upon class definition, add instr…

p-datadog:
…umentation leak detector

**What does this PR do?**

Repairs a method instrumentation leak: if the probe was for a not-yet-loaded class, and the class was later loaded, the accounting was not updated to reflect that the probe has been installed. Subsequently, this probe would not be removed by the probe manager (it could be removed directly via Instrumenter but this is not a path that would be taken in production).

Additionally, the PR adds an instrumentation leak detector: DI now tracks the number of outstanding instrumentations (by type - line/method) and verifies that after each test it's zero. This leak detector identified the leak that is described above.

**Motivation:**
Ruby DI GA

**Change log entry**
Yes: DI/LD: fix instrumentation leak for method probes when the probe references a class that is loaded after the probe is received by the application

**Additional Notes:**

**How to test the change?**

Unit tests added

### DIFF
--- a/sig/datadog/di.rbs
+++ b/sig/datadog/di.rbs
@@ -1,17 +1,20 @@
 module Datadog
   module DI
+    type instrumentation_kind = :line | :method
     INSTRUMENTED_COUNTERS_LOCK: Mutex
 
     def self.enabled?: () -> bool
+
     def self.component: () -> Component
-    def self.instrumented_count: (?Symbol? kind) -> Integer
 
-    def self.instrumented_count_inc: (Symbol kind) -> void
+    def self.instrumented_count: (?instrumentation_kind? kind) -> Integer
 
-    def self.instrumented_count_dec: (Symbol kind) -> void
+    def self.instrumented_count_inc: (instrumentation_kind kind) -> void
+
+    def self.instrumented_count_dec: (instrumentation_kind kind) -> void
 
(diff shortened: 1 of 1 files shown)


sig/datadog/di.rbs
```diff
@@ -1,7 +1,17 @@
...
+    def self.component: () -> Component
+    def self.instrumented_count: (?Symbol? kind) -> Integer
+
+    def self.instrumented_count_inc: (Symbol kind) -> void
+
+    def self.instrumented_count_dec: (Symbol kind) -> void
+
+    private
+
+    def self.validate_kind!: (Symbol kind) -> void
```
marcotc:
```suggestion
    type instrumentation_kind = :line | :method
    INSTRUMENTED_COUNTERS_LOCK: Mutex

    def self.enabled?: () -> bool

    def self.component: () -> Component

    def self.instrumented_count: (?instrumentation_kind? kind) -> Integer

    def self.instrumented_count_inc: (instrumentation_kind kind) -> void

    def self.instrumented_count_dec: (instrumentation_kind kind) -> void

    private

    def self.validate_kind!: (instrumentation_kind kind) -> void
```

I think this might work to improve checking the argument value.

PR #5173, opened 2025-12-19
[NO-TICKET] Fix `UnknownInstanceVariable` steep check

vpellan:
**What does this PR do?**

Remove `UnknownInstanceVariable` exclusion from the Steepfile and fixes related issues.
(Also re-enables `UnknownRecordKey` which had no failures anyway)

**Motivation:**

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

bundle exec rake steep:check

### DIFF
--- a/lib/datadog/core/utils/safe_dup.rb
+++ b/lib/datadog/core/utils/safe_dup.rb
@@ -5,8 +5,6 @@ module Datadog
     module Utils
       # Helper methods for safer dup
       module SafeDup
-        # String#+@ was introduced in Ruby 2.3
-        #
         # Steep: https://github.com/soutaro/steep/issues/2001
         def self.frozen_or_dup(v) # steep:ignore MethodBodyTypeMismatch
           # For the case of a String we use the methods +@ and -@.
--- a/sig/datadog/di/probe.rbs
+++ b/sig/datadog/di/probe.rbs
@@ -54,7 +54,7 @@ module Datadog
       attr_reader template_segments: Array[untyped]?
       attr_reader rate_limiter: Datadog::Core::RateLimiter
 
-      attr_reader condition_evaluation_failed_rate_limiter: Datadog::Core::TokenBucket?
+      attr_reader condition_evaluation_failed_rate_limiter: Datadog::Core::RateLimiter?
 
       def capture_snapshot?: () -> bool
       def line?: () -> bool


lib/datadog/tracing/span.rb
```diff
@@ -80,7 +80,7 @@ def initialize(
 
         @meta = meta || {}
         @metrics = metrics || {}
-        @metastruct = metastruct || {}
+        @metastruct = metastruct || Metastruct.new
```
vpellan, line 83: `Span` includes `Metadata` which includes `MetastructTagging`, which already defines `@metastruct` and some helper methods. These methods are not used on `Span` but on `SpanOperation` only which is why we did not caught that before. However that also means that when creating a `Span` from a `SpanOperation` `metastruct` argument is a `Metastruct` object, not a Hash.

We cannot remove the attr_reader above because we use it in `SerializableTrace`, and the attr_reader will override the private `metastruct` method in `MetastructTagging`, meaning that we have to set `@metastruct` to something here, so we cannot replace it by
```ruby
@metastruct = metastruct if metastruct
```
or maybe if we do that, we can change the visibility of `metastruct` method from `MetastructTagging` to public?

sig/datadog/di/probe.rbs
```diff
@@ -54,6 +54,8 @@ module Datadog
       attr_reader template_segments: Array[untyped]?
       attr_reader rate_limiter: Datadog::Core::RateLimiter
 
+      attr_reader condition_evaluation_failed_rate_limiter: Datadog::Core::TokenBucket?
```
p-datadog:
```suggestion
      attr_reader condition_evaluation_failed_rate_limiter: Datadog::Core::RateLimiter?
```
I think.

lib/datadog/core/utils/safe_dup.rb
```diff
@@ -6,7 +6,9 @@ module Utils
       # Helper methods for safer dup
       module SafeDup
         # String#+@ was introduced in Ruby 2.3
```
marcotc: Can I bug you to delete this line too, since we missed it in an earlier [cleanup](https://github.com/DataDog/dd-trace-rb/commit/00ee0a367c8bcc9562a706fffeeb5da0bda353b7#diff-87a62985245401118726f2db8a6ce408ef2d92980f5751e27ff30baff170a07fR8)?
```suggestion
```

sig/datadog/di/probe.rbs
marcotc: I agree that typing this with the interface `Core::RateLimiter`, instead of the concrete class `TokenBucket` seems like the right intention, since the `Probe` doesn't really care about the implementation of rate limiting.

vpellan: Changed in 3e72003ff27051f8a08d70cb1c750395316bcf76!

lib/datadog/core/utils/safe_dup.rb
vpellan: Removed in 3e72003ff27051f8a08d70cb1c750395316bcf76

PR #5175, opened 2025-12-19
DEBUG-4548 Fix global state dependency in crashtracking and process discovery tests

p-datadog:
**What does this PR do?**

Adds state resets to crash tracking and process discovery tests

**Motivation:**
Existing tests are failing with the changes in https://github.com/DataDog/dd-trace-rb/pull/5159

**Change log entry**
None

**Additional Notes:**
We already have tests in master for after fork money patcher that reset the list of registered after fork handlers (such as the ones set by crash tracking and process discovery). Why these have not ever failed is a mystery to me. Maybe this has to do with how tests are grouped in CI.

Regardless, in https://github.com/DataDog/dd-trace-rb/pull/5159 I now use the monkey patcher for telemetry and this is causing test failures because crash tracking and process discovery assume that their after fork handlers have been installed somehow (and they usually get installed by the library itself when it gets loaded, or coincidentally by other tests in my local test runs) but the resets in telemetry and/or the monkey patch unit tests themselves clear out those handlers.

This PR adds state resets to the existing tests.

**How to test the change?**
Existing CI + https://github.com/DataDog/dd-trace-rb/pull/5159

### DIFF
--- a/spec/datadog/core/process_discovery_spec.rb
+++ b/spec/datadog/core/process_discovery_spec.rb
@@ -115,8 +115,7 @@ RSpec.describe Datadog::Core::ProcessDiscovery do
       double(Datadog::Core::Configuration::Settings,
         service: 'dummy',
         env: 'dummy',
-        version: 'dummy',
-      )
+        version: 'dummy',)
     end
 
     it 'updates the process discovery file descriptor' do


spec/datadog/core/process_discovery_spec.rb
```diff
@@ -94,9 +94,38 @@
...
+        version: 'dummy',)
     end
 
     it 'updates the process discovery file descriptor' do
+      # The fork handler is installed by +publish+.
+      # If +publish+ is not called, this test is relying on the handler
+      # being installed by a different test and not cleared out by yet another
+      # test. This does not hold in general because we have multiple tests
+      # that reset after fork handlers.
+      described_class.publish(dummy_settings)
```
p-datadog, line 127: I would prefer to call some sort of init method for Process Discovery instead of submitting a payload to publish, but currently process discovery does not appear to have any initialization I can invoke.

PR #5176, opened 2025-12-19
DEBUG-4548 Core: fix worker shutdown race 

p-datadog:
**What does this PR do?**

Fixes a race condition in worker start/stop code described in https://github.com/DataDog/ruby-guild/issues/279.

Additionally, the changes in this PR started failing open feature worker specs which were not waiting for the work submitted to the worker to be processed before asserting on the outcome. I moved the telemetry flushing code to core worker modules to make it available to open feature also.

This was originally part of https://github.com/DataDog/dd-trace-rb/pull/5159 but I extracted the changes into their own PR after discovering the worker tests disabled on JRuby.

**Motivation:**

The original implementation of this change in https://github.com/DataDog/dd-trace-rb/pull/5074 uncovered an existing race condition in worker start/stop code which this PR is fixing.

**Change log entry**
Yes: Core: improve worker shutdown reliability

**Additional Notes:**

I found three tests that are marked skipped on JRuby due to flakiness that seem to be directly affected by the worker start/stop race (one of the tests literally only calls start & stop on a worker). 

**How to test the change?**

Unit tests added.

I also found 3 skipped tests on JRuby that I believe are skipped because of the bug fixed in this PR, those tests are unskipped.

For the tests unskipped by this PR, I ran the entire CI with this PR many times in https://github.com/DataDog/dd-trace-rb/pull/5171 while working on further fixes (forking-related) and those tests have not re-flaked.

### DIFF
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -77,7 +77,7 @@ jobs:
   test:
     needs:
       - build
-    uses: DataDog/system-tests/.github/workflows/system-tests.yml@0032fc1c262ec50dc0858d106d038858a266865c  # Automated: This reference is automatically updated.
+    uses: DataDog/system-tests/.github/workflows/system-tests.yml@91b7434cee90262199f503010b27754aa9e38a29  # Automated: This reference is automatically updated.
     secrets:
       TEST_OPTIMIZATION_API_KEY: ${{ secrets.DD_API_KEY }}  # key used to pushed test results to test optim
       DD_API_KEY: ${{ secrets.DD_API_KEY }}  # key used in tests runs
@@ -91,7 +91,7 @@ jobs:
       desired_execution_time: 300  # 5 minutes
       scenarios_groups: tracer_release
       skip_empty_scenarios: true
-      ref: 0032fc1c262ec50dc0858d106d038858a266865c  # Automated: This reference is automatically updated.
+      ref: 91b7434cee90262199f503010b27754aa9e38a29  # Automated: This reference is automatically updated.
       force_execute: ${{ needs.build.outputs.forced_tests }}
       parametric_job_count: 8
       push_to_test_optimization: ${{ github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository }}
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -146,7 +146,7 @@ requirements_json_test:
 
 configure_system_tests:
   variables:
-    SYSTEM_TESTS_REF: 0032fc1c262ec50dc0858d106d038858a266865c # Automated: This reference is automatically updated.
+    SYSTEM_TESTS_REF: 91b7434cee90262199f503010b27754aa9e38a29 # Automated: This reference is automatically updated.
     SYSTEM_TESTS_SCENARIOS_GROUPS: "simple_onboarding,lib-injection,docker_ssi"
 
 save_versions:
--- a/benchmarks/di_snapshot.rb
+++ b/benchmarks/di_snapshot.rb
@@ -103,9 +103,7 @@ class DISnapshotBenchmark
       x.compare!
     end
 
-    # DI does not provide an API to remove a specific probe because
-    # this functionality is currently not needed by the product.
-    probe_manager.remove_other_probes([])
+    probe_manager.clear_hooks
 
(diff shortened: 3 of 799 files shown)


p-datadog:
JRuby 9.2 failures:
```
     NoMethodError:
       undefined method `join' for #<Thread:0x6f4ac7a8>
       Did you mean?  JSON
     # ./lib/datadog/tracing/workers.rb:97:in `join'
     # ./lib/datadog/tracing/workers.rb:91:in `stop'
     # ./spec/datadog/tracing/workers_spec.rb:24:in `block in <main>'
     # ./spec/spec_helper.rb:275:in `block in <main>'
     # ./spec/spec_helper.rb:155:in `block in <main>'
     # /usr/local/bundle/gems/webmock-3.25.0/lib/webmock/rspec.rb:39:in `block in <main>'
     # /usr/local/bundle/gems/rspec-wait-0.0.10/lib/rspec/wait.rb:47:in `block in <main>'
```

If more recent jrubies are passing I will skip 9.2 only.

lib/datadog/core/workers/queue.rb
```diff
@@ -34,10 +50,93 @@ def work_pending?
...
+        # there was no pending or in progress work.
+        #
+        # Flushing can time out because there is a constant stream of work
+        # submitted at the same or higher rate than it is processed.
+        # Flushing can also fail if the worker thread is not running -
+        # this method will not flush from the calling thread.
+        def flush(timeout: nil)
+          # Default timeout is 5 seconds.
+          # Specific workers can override it to be more or less
+          timeout ||= 5
```
vpellan, line 67:
```suggestion
        def flush(timeout: 5)
```

sig/datadog/core/telemetry/worker.rbs
```diff
@@ -39,7 +39,7 @@ module Datadog
 
         def dequeue: () -> Array[Event::Base]
 
-	def flush: () -> (true | false)
+	def flush: (?timeout: Numeric?) -> bool
```
vpellan, line 42: Will we ever use it with something other than Integer or Float ? If not we can narrow it. Same in queue.rbs

lib/datadog/core/workers/queue.rb
p-datadog: I have the argument defaulted to nil for consistency with https://github.com/DataDog/dd-trace-rb/pull/5159/files#diff-9b0739bf5b667dd145cd7a817c33f0ad00e5818bc579f57b78fd29b221cb9e52R127 which needs a longer timeout.

If the default is a literal like 5 there is the risk of that signature being copy-pasted elsewhere at which point it might end up overriding intermediate timeouts, like we have with the `work_pending?` method in workers.

sig/datadog/core/telemetry/worker.rbs
p-datadog: Conceptually, the value is a number. What would be the harm if someone passed Rational here? On the other hand, it takes non-zero effort to ascertain that Integer and Float cover all possible values (our steep runs are quite long....)

p-datadog:
Looks like CI is not always getting enough iterations in jruby
```


Failures:

  1) Datadog::Core::Workers::Polling when included into a worker #stop when the worker has just been started when run several times stops the worker on each iteration
     Failure/Error: expect(expected_state_met).to be true

       expected true
            got false
     # ./spec/datadog/core/workers/polling_spec.rb:202:in 'block in <main>'
     # ./spec/spec_helper.rb:277:in 'block in <main>'
     # ./spec/spec_helper.rb:155:in 'block in <main>'
     # /usr/local/bundle/gems/webmock-3.26.1/lib/webmock/rspec.rb:39:in 'block in <main>'
     # /usr/local/bundle/gems/rspec-wait-0.0.10/lib/rspec/wait.rb:47:in 'block in <main>'

Finished in 3 minutes 56.4 seconds (files took 9.61 seconds to load)
5839 examples, 1 failure, 677 pending

Failed examples:

rspec ./spec/datadog/core/workers/polling_spec.rb:162 # Datadog::Core::Workers::Polling when included into a worker #stop when the worker has just been started when run several times stops the worker on each iteration

```

p-datadog:
The file descriptor test is also failing: https://github.com/DataDog/dd-trace-rb/actions/runs/20491809166/job/58885607186?pr=5178

The diagnostics needs to be improved - at least include what the `readlink` call resolved to, and also probably should only show the new entries (or new entries separately from the existing ones).

PR #5178, opened 2025-12-20
Fix ruby warnings when accessing undefined instance variables

p-datadog:
**What does this PR do?**

Adds `defined(@foo)` guards when accessing instance variables that are initialized lazily

**Motivation:**
Remove `ruby -w` warnings I am seeing in local test runs

**Change log entry**
None

**Additional Notes:**

**How to test the change?**

Existing CI

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -148,6 +148,7 @@ configure_system_tests:
   variables:
     SYSTEM_TESTS_REF: 91b7434cee90262199f503010b27754aa9e38a29 # Automated: This reference is automatically updated.
     SYSTEM_TESTS_SCENARIOS_GROUPS: "simple_onboarding,lib-injection,docker_ssi"
+    SYSTEM_TEST_BUILD_ATTEMPTS: 3
 
 save_versions:
   image: $DOCKER_REGISTRY/images/mirror/ruby:3.2.2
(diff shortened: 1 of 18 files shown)


lib/datadog/tracing/contrib/extensions.rb
```diff
@@ -220,12 +220,20 @@ def integrations_pending_activation
             # `@instrumented_integrations` hash.
             # @!visibility private
             def instrumented_integrations
-              INSTRUMENTED_INTEGRATIONS_LOCK.synchronize { (@instrumented_integrations&.dup || {}).freeze }
+              INSTRUMENTED_INTEGRATIONS_LOCK.synchronize do
+                if defined?(@instrumented_integrations)
+                  @instrumented_integrations&.dup
+                end || {}
+              end.freeze
```
vpellan: The `freeze` was freezing the resulting hash of `@instrumented_integrations&.dup || {}` but now it freezes the Mutex

marcotc: It freezes return of `synchronize do ... end`, which is still correct.

BUT, it freezes it outside of the critical session.
It's unlikely to make a difference here, since `instrumented_integrations` always returns a new Hash instance, but might as well leave it as it was before.
```suggestion
              INSTRUMENTED_INTEGRATIONS_LOCK.synchronize do
                (if defined?(@instrumented_integrations)
                  @instrumented_integrations&.dup
                end || {}).freeze
              end
```

marcotc:
I feel like we need a Cop for this.

vpellan:
> I feel like we need a Cop for this.

Steep should be able to catch that but this file is ignored.

lib/datadog/tracing/contrib/extensions.rb
p-datadog: Applied manually

PR #5188, opened 2025-12-29
Ignore "leaked" pipe file descriptors in JRuby, improve diagnostics 

p-datadog:
**What does this PR do?**

Updates the test checking for leaked file descriptors to exclude pipes under JRuby.

**Motivation:**
The test is flaking in CI - diagnostics added in this PR revealed that there are pipe FDs showing as leaked under JRuby.

**Change log entry**
None

**Additional Notes:**
This test was skipped completely on JRuby prior to https://github.com/DataDog/dd-trace-rb/pull/5176 - with this PR there is still more test coverage for file descriptor leaks compared to what we had previously.

Also, MRI retains the full file descriptor leakage diagnostics, as before.

**How to test the change?**

Existing tests

### DIFF
--- a/spec/datadog/integration_spec.rb
+++ b/spec/datadog/integration_spec.rb
@@ -79,13 +79,13 @@ RSpec.describe 'Datadog integration' do
       # Ignore the pipes too.
       #
       # standard:disable Lint/ConstantDefinitionInBlock:
-      IGNORE_JRUBY_FDS_REGEXP = %r,
+      IGNORE_JRUBY_FDS_REGEXP = %r{
         \A (
           anon_inode:\[eventpoll\] |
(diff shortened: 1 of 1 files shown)


spec/datadog/profiling/collectors/stack_spec.rb
```diff
@@ -809,6 +809,9 @@ def dummy_template.#{method_name}(ready_queue)
     end
 
     after do
+      # Ensure we are not leaking file descriptors
+      ready_pipe.map(&:close)
```
ivoanjo, line 813: This one is not strictly needed unless the test is failing (since both pipes are closed in the `before`), but it's harmless to have so why not ;)

Strech (approved):

PR #5198, opened 2026-01-06
[NO-TICKET] Re-enable ASAN testing for profiling

ivoanjo:
**What does this PR do?**

Some months ago, in https://github.com/DataDog/dd-trace-rb/pull/4667, we disabled ASAN testing due to flakiness.

As per https://github.com/ruby/setup-ruby/issues/850 and with the release of Ruby 4, we can now move asan testing to use Ruby 4, so I thought I'd give it a try again and see if the flakiness is gone.

~~(I'm opening this as a draft to check first if everything is OK in the new validation)~~

**Update**:
* It's working fine, marking as non-draft
* This PR is on top of #5197 because otherwise this step would fail not because of ASAN but because of the specs that are fixed by that PR. It's otherwise completely unrelated.

**Motivation:**

ASAN (or AddressSanitizer) is a tool to detect memory issues, including memory leaks, in native libraries. See <https://research.google/pubs/addresssanitizer-a-fast-address-sanity-checker/> for details.

**Change log entry**

None.

**Additional Notes:**

If this flakes again, we'll disable it again.

**How to test the change?**

Validate that the new step runs and is green.


ivoanjo:
Update: Now that #5197 is ready, this is good to be merged! But, since this has caused some flakiness in the past, I'll hold off on merging it until the next gem release is out, just in case.

ivoanjo:
2.24.0 is out, so I'm going to go ahead and merge this in!

PR #5201, opened 2026-01-06
[PROF-13510] Heap profiling for ruby 4.x

r1viollet:
**What does this PR do?**

This PR reimplements part of the heap profiler to support Ruby 4 (which was
disabled in https://github.com/DataDog/dd-trace-rb/pull/5148 ).

**Motivation:**

Support heap profiling on Ruby 4!

**Change log entry**

Yes. Add support for heap profiling on Ruby 4.

*For the highlights*:

```markdown
# Heap profiling can now be used on Ruby 4

We've now added support for heap profiling on Ruby 4.0 .
[Check the docs](https://docs.datadoghq.com/profiler/enabling/ruby/?tab=environmentvariables#configuration) for how to enable it.
```

**Additional Notes:**

The Datadog Ruby heap profiler tracks live heap objects by storing their `object_id` when they're allocated, then later using `ObjectSpace._id2ref` to check if those objects are still alive. This mechanism is **currently incompatible with Ruby 4.x**.

##### Key Components

1. **`collectors_cpu_and_wall_time_worker.c`** - Main sampling coordinator
2. **`heap_recorder.c`** - Tracks live heap objects using object IDs

###### Allocation Flow (Before Fix)

```
on_newobj_event()  →  start_heap_allocation_recording()  →  end_heap_allocation_recording()
                              ↓
                      rb_obj_id(new_obj)  ← PROBLEM HERE
```

###### Liveness Check Flow

```
heap_recorder_update()  →  st_object_record_update()  →  ruby_ref_from_id()
                                                              ↓
                                                      ObjectSpace._id2ref(obj_id)
```

###### What Changed

Ruby 4.x changed how `object_id` works internally. The key issue:

- `on_newobj_event` is called during object allocation (object is in "in-between state")
- Calling `rb_obj_id()` during this event **mutates the object** (assigns an ID)
- This mutation is **not safe** during the allocation tracepoint in Ruby 4.x
- Reference: [Ruby Issue #21710](https://bugs.ruby-lang.org/issues/21710)

##### Implemented Solution: Deferred Object ID Recording

We defer the `rb_obj_id()` call to after the allocation tracepoint completes using `rb_postponed_job_trigger`.

###### Allocation Flow (After Fix - Ruby 4+)

```
on_newobj_event()
    ↓
start_heap_allocation_recording()
    - Store VALUE in heap_recorder->active_deferred_object
    - Store metadata in heap_recorder->active_deferred_object_data
    ↓
end_heap_allocation_recording()
    - Move to pending_recordings[] array
    ↓
rb_postponed_job_trigger()
    ↓
finalize_heap_allocation_from_postponed_job()  ← Runs outside allocation tracepoint
    - heap_recorder_finalize_pending_recordings()
    - Call rb_obj_id() safely
    - Commit object_record to heap_record
```

⚠️ These changes are AI assisted and will require careful review & analysis of performance impacts.

**How to test the change?**

The usual test coverage uses the new code path on Ruby 4.

We've also added additional coverage in https://github.com/DataDog/prof-correctness/pull/89 + internally in the reliability environment.


r1viollet:
Reminder that to measure this, we should do ON/OFF. Some of the cost will be in the VM itself.

ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -89,6 +90,9 @@ unsigned int MAX_ALLOC_WEIGHT = 10000;
   static rb_postponed_job_handle_t sample_from_postponed_job_handle;
   static rb_postponed_job_handle_t after_gc_from_postponed_job_handle;
   static rb_postponed_job_handle_t after_gvl_running_from_postponed_job_handle;
+  #ifdef DEFERRED_HEAP_ALLOCATION_RECORDING
+    static rb_postponed_job_handle_t finalize_heap_allocation_from_postponed_job_handle;
+  #endif
```
ivoanjo: Minor: More as a style thing, in general I've avoided overly-ifdefing things out when they're harmless.

That is, the cost of a leftover extra field on Rubies that don't need it is so small that I usually prefer the advantage of less code and easier to reason due to less ifdef branching. (Same for most spots in this file)

ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -1404,6 +1423,27 @@ static VALUE _native_resume_signals(DDTRACE_UNUSED VALUE self) {
...
+
+  if (!ddtrace_rb_ractor_main_p()) {
+    return;
+  }
+
+  // Get the heap_recorder from the thread_context_collector
+  heap_recorder *recorder = thread_context_collector_get_heap_recorder(state->thread_context_collector_instance);
+  if (recorder == NULL) return;
+
+  heap_recorder_finalize_pending_recordings(recorder);
```
ivoanjo: This is flipped from how the rest of the codebase does it. That is, usually the flow goes cpu_and_wall_time_worker -> thread_context -> stack_recorder -> heap_recorder  and the worker doesn't really know **what** goes on deeper in the sampling code, it's only the driver for the stimulus.

In practice what this allows is testing the operations separately from the stimulus. That is, if the thread_context (or later parts of the code) does all of the work based on the stimulus, we have a really nice "seam" to break down testing logic into two really easy to test parts: 1) Test stimulus fires in a real app with real Ruby; 2) Test without background threads, concurrency, weird complexity just the response to the stimulus.

TL;DR This should ideally be reduced to a single thread_context_collector_finish_heap_sample call or somesuch?

ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -1222,6 +1235,12 @@ static void on_newobj_event(DDTRACE_UNUSED VALUE unused1, DDTRACE_UNUSED void *u
   state->stats.allocation_sampled++;
 
   during_sample_exit(state);
+
+  #ifdef DEFERRED_HEAP_ALLOCATION_RECORDING
+  // On Ruby 4+, we need to trigger a postponed job to finalize the heap allocation recording.
+  // During on_newobj_event, we can't safely call rb_obj_id(), so we defer it until the event completes.
+  rb_postponed_job_trigger(finalize_heap_allocation_from_postponed_job_handle);
+  #endif
```
ivoanjo: It occurs to me that doing this _always_ is unneeded overhead. As it is right now:

1) If N objects get allocated, we do this N times even though Ruby will already call us after we asked it the first time
2) If we don't heap sample an object we still trigger the postponed job
3) We're doing this on Ruby 4 even if heap profiling is disabled

My suggestion would be to have `thread_context_collector_sample_allocation` return a flag that tells Ruby if the postponed job should be done or not. (And `sample_allocation` would get it as a new return from `track_object`, for instance); a similar pattern happens for `thread_context_collector_on_gc_finish` and `thread_context_collector_on_gvl_running` for instance.

ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -16,6 +16,7 @@
 #include "private_vm_api_access.h"
 #include "setup_signal_handler.h"
 #include "time_helpers.h"
+#include "heap_recorder.h"
```
ivoanjo: Minor: This is actually not needed (leftover from previous attempt?)

```suggestion
```

ext/datadog_profiling_native_extension/extconf.rb
```diff
@@ -144,6 +144,11 @@ def skip_building_extension!(reason)
 # On Ruby 4, we can't ask the object_id from IMEMOs (https://github.com/ruby/ruby/pull/13347)
 $defs << "-DNO_IMEMO_OBJECT_ID" unless RUBY_VERSION < "4"
 
+# On Ruby 4, we need to defer calling rb_obj_id during heap allocation recording
+# because it's not safe to mutate objects during the newobj tracepoint
+# (see https://bugs.ruby-lang.org/issues/21710)
+$defs << "-DDEFERRED_HEAP_ALLOCATION_RECORDING" unless RUBY_VERSION < "4"
```
ivoanjo: Minor: Usually the way we lay down these flags is to slightly flip them -- that is, consider that the thing modern Rubies do is the default, and define `NO_...` flags to avoid the behavior on older Rubies.

No strong feelings about it, more of a consistency thing with the below.

lib/datadog/profiling/component.rb
ivoanjo, line 1: Do we maybe want to introduce an experimental flag to enable this new codepath during testing? I remember we said no, then maybe (based on prof-correctness still showing some work to be done) so adding here a note so we decide ;)

spec/datadog/profiling/component_spec.rb
```diff
@@ -335,13 +335,16 @@
...
+
+            it "initializes StackRecorder with heap sampling support" do
               expect(Datadog::Profiling::StackRecorder).to receive(:new)
-                .with(hash_including(heap_samples_enabled: false, heap_size_enabled: false))
+                .with(hash_including(heap_samples_enabled: true, heap_size_enabled: true))
                 .and_call_original
 
-              expect(logger).to receive(:warn).with(/Datadog Ruby heap profiler is currently incompatible with Ruby 4/)
-
               build_profiler_component
```
ivoanjo: If we do add an extra setting to toggle the new code, this testcase is still worth keeping. On the other hand, if we don't, then this test-case has become redundant: since Ruby 4 is no longer special, the existing tests below already cover this situation. (The test was added exactly because there was an exception for Ruby 4)

ext/datadog_profiling_native_extension/heap_recorder.h
```diff
@@ -123,6 +123,20 @@ int end_heap_allocation_recording_with_rb_protect(heap_recorder *heap_recorder,
...
+// Finalize any pending heap allocation recordings by getting their object IDs.
+// This should be called via a postponed job, after the on_newobj_event has completed.
+// Returns true if there were pending recordings that were finalized.
+bool heap_recorder_finalize_pending_recordings(heap_recorder *heap_recorder);
+
+// Check if there are any pending recordings waiting to be finalized.
+bool heap_recorder_has_pending_recordings(heap_recorder *heap_recorder);
+
+// Mark pending recordings to prevent GC from collecting the objects
+// while they're waiting to be finalized.
```
ivoanjo: Minor: Given we're talking about GC and objects all the times, the word "finalized" can be a bit confusing ;)

```suggestion
// while they're waiting for the recording to be finalized.
```

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -79,6 +79,18 @@ static void object_record_free(heap_recorder*, object_record*);
...
+// A pending recording is used to defer the object_id call on Ruby 4+
+// where calling rb_obj_id during on_newobj_event is unsafe.
+typedef struct {
+  VALUE object_ref;
+  heap_record *heap_record;
+  live_object_data object_data;
+} pending_recording;
+
+#define MAX_PENDING_RECORDINGS 64
+#endif
```
ivoanjo: Minor: The same note about not overly if-def'ing things from the cpu and wall collector I believe applies here as well

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -399,6 +461,77 @@ void heap_recorder_update_young_objects(heap_recorder *heap_recorder) {
...
+void heap_recorder_mark_pending_recordings(heap_recorder *heap_recorder) {
+  if (heap_recorder == NULL) {
+    return;
+  }
+
+  for (uint i = 0; i < heap_recorder->pending_recordings_count; i++) {
+    VALUE obj = heap_recorder->pending_recordings[i].object_ref;
+    if (obj != Qnil) {
+      rb_gc_mark(obj);
+    }
```
ivoanjo: The object can't ever be `Qnil` (or if it is, we have a bug...), since it was set from an allocation and `Qnil` is not a heap-allocated object. (It's a tagged pointer)

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -399,6 +461,77 @@ void heap_recorder_update_young_objects(heap_recorder *heap_recorder) {
...
+  }
+
+  // Also mark the active deferred object if it's set
+  if (heap_recorder->active_deferred_object != Qnil) {
+    rb_gc_mark(heap_recorder->active_deferred_object);
+  }
```
ivoanjo: Minor: This can can indeed trivially be `Qnil`; but btw it's OK to mark `Qnil`, so maybe remove the branch anyway? Less code ;)

ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
```diff
@@ -1404,6 +1423,27 @@ static VALUE _native_resume_signals(DDTRACE_UNUSED VALUE self) {
...
+  cpu_and_wall_time_worker_state *state = active_sampler_instance_state;
+
+  if (state == NULL) return;
+
+  if (!ddtrace_rb_ractor_main_p()) {
+    return;
+  }
+
+  // Get the heap_recorder from the thread_context_collector
+  heap_recorder *recorder = thread_context_collector_get_heap_recorder(state->thread_context_collector_instance);
```
ivoanjo: Ah, on a second pass, there's two things missing here that can create a bit of a sharp edge:

a. We're missing `during_sample_enter` and `during_sample_exit` -- these set a flag that allows us to avoid nested operations inside the profiler. E.g. some sharp edges along the line of "the profiler is sampling something else -> it calls some VM api that causes the VM to check for interruptions -> the VM decides now it's a really nice time to flush heap things -> our current state may not be in a consistent sate" (or reverse -- maybe this is the function that started first, and it triggers an allocation, and the flip situation happens)

b. (Minor) We're missing the `discrete_dynamic_sampler_before_sample` and `discrete_dynamic_sampler_after_sample` calls to update the dynamic sampling rate mechanism. In practice, this means that work done inside this function isn't accounted as being profiler overhead. TBH what we're doing right now isn't a lot but... yeah maybe at least leave a comment saying "This is not being accounted for the dynamic sampling rate update, and it's ok because the amount of work we do in this case is very small"

Both things happen in e.g. `on_newobj_event` so doing the same here should be enough.

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -314,13 +341,29 @@ void start_heap_allocation_recording(heap_recorder *heap_recorder, VALUE new_obj
...
+  #ifdef DEFERRED_HEAP_ALLOCATION_RECORDING
+  // On Ruby 4+, we can't call rb_obj_id during on_newobj_event as it mutates the object.
+  // Instead, we store the VALUE reference and will get the object_id later via a postponed job.
+  // active_deferred_object != Qnil indicates we're in deferred mode.
+  heap_recorder->active_deferred_object = new_obj;
+  heap_recorder->active_deferred_object_data = (live_object_data) {
+    .weight = weight * heap_recorder->sample_rate,
+    .class = intern_or_raise(heap_recorder->string_storage, alloc_class),
+    .alloc_gen = rb_gc_count(),
+  };
```
ivoanjo: Minor: Maybe this could be initialized before the #ifdef on the stack, since both branches need the same `live_object_data`

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -367,6 +411,24 @@ static VALUE end_heap_allocation_recording(VALUE protect_args) {
...
+    if (heap_recorder->pending_recordings_count < MAX_PENDING_RECORDINGS) {
+      heap_record *heap_record = get_or_create_heap_record(heap_recorder, locations);
+      heap_record->num_tracked_objects++; // Pre-increment since we're going to commit this later
+
+      pending_recording *pending = &heap_recorder->pending_recordings[heap_recorder->pending_recordings_count++];
+      pending->object_ref = heap_recorder->active_deferred_object;
+      pending->heap_record = heap_record;
+      pending->object_data = heap_recorder->active_deferred_object_data;
+    }
+    heap_recorder->active_deferred_object = Qnil;
```
ivoanjo, line 454: Minor: Maybe clean `active_deferred_object_data` as well? Just to avoid any confusion with leftover state.

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -367,6 +411,24 @@ static VALUE end_heap_allocation_recording(VALUE protect_args) {
   heap_recorder *heap_recorder = args->heap_recorder;
   ddog_prof_Slice_Location locations = args->locations;
 
+  #ifdef DEFERRED_HEAP_ALLOCATION_RECORDING
+  // On Ruby 4+, active_deferred_object != Qnil indicates deferred mode
+  if (heap_recorder->active_deferred_object != Qnil) {
+    // Store the recording in the pending list for later finalization
+    if (heap_recorder->pending_recordings_count < MAX_PENDING_RECORDINGS) {
```
ivoanjo: This check isn't needed -- in particular the invariant is that `active_deferred_object` only gets set when `heap_recorder->pending_recordings_count < MAX_PENDING_RECORDINGS` and thus it's not possible (unless there's a bug, and in that case we should instead raise an error) for this if to ever be false.

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -399,6 +461,77 @@ void heap_recorder_update_young_objects(heap_recorder *heap_recorder) {
   heap_recorder_update(heap_recorder, /* full_update: */ false);
 }
 
+#ifdef DEFERRED_HEAP_ALLOCATION_RECORDING
+bool heap_recorder_has_pending_recordings(heap_recorder *heap_recorder) {
+  if (heap_recorder == NULL) {
+    return false;
+  }
+  return heap_recorder->pending_recordings_count > 0;
+}
```
ivoanjo: Minor: Since this is only for debugging, I wonder if we should just return the count instead of a boolean. Same cost for the logic, and a bit easier to debug from the Ruby side.

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -399,6 +461,77 @@ void heap_recorder_update_young_objects(heap_recorder *heap_recorder) {
...
+
+    VALUE obj = pending->object_ref;
+    VALUE ruby_obj_id = rb_obj_id(obj);
+    if (!FIXNUM_P(ruby_obj_id)) {
+      // Bignum object id - not supported, skip this recording
+      pending->heap_record->num_tracked_objects--;
+      cleanup_heap_record_if_unused(heap_recorder, pending->heap_record);
+      unintern_or_raise(heap_recorder, pending->object_data.class);
+      continue;
+    }
```
ivoanjo: This can be simplified because this is a fatal issue -- Ruby attributes object ids from a sequential counter, and thus once we see one, we might as well stop profiling. (And thus we don't even need to clean up stuff -- the heap_recorder_free should be enough).

The only thing we need to do is to let the cpu_and_wall_time_worker know that something bad happened, via either an exception or a flag, so it knows to stop all operations.

(In practice this is a should never happen so while we could handle it more graciously and cleverly it's not needed and we just need to make it as simple as possible while remaining correct just in case this assumption proves incorrect)

The exception raise is also incomplete as we're not handling it in the cpu_and_wall_time_worker and thus if that `unintern_or_raise` does raise an exception the exception will leak onto user code.

I know, this is _really_ fiddly... Suggestions very welcome >_>

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -399,6 +461,77 @@ void heap_recorder_update_young_objects(heap_recorder *heap_recorder) {
...
+
+    // Commit the recording
+    int existing_error = st_update(heap_recorder->object_records, record->obj_id, update_object_record_entry, (st_data_t) record);
+    if (existing_error) {
+      // Duplicate object_id - this shouldn't happen normally, but handle it gracefully
+      pending->heap_record->num_tracked_objects--;
+      cleanup_heap_record_if_unused(heap_recorder, pending->heap_record);
+      object_record_free(heap_recorder, record);
+    }
+  }
```
ivoanjo, line 495: This looks like an inlined version of `commit_recording`, possibly with the difference that the `num_tracked_objects` has been pre-incremented.

I think since this logic is quite fiddly, it's probably worth trying to unify them -- possibly having a flag that says "I've already incremented the object!"; or just have callers do the increment.

(For instance -- I _just_ noticed that the new code doesn't check for overflow on `num_tracked_objects` and this is the kind of "oops" that can easily creep in if the logic gets too duplicated)

ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
ivoanjo: Edit: Actually I wonder if we can even get recursion where finalize gets called -> goes into the vm -> vm calls finalize again (e.g. if there's an allocation) or something. `during_sample_enter`/`during_sample_exit` would also protect against that.

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -699,6 +701,47 @@ def introduce_distinct_stacktraces(i, obj)
...
+              test_object = Object.new
+              track_object_without_finalize(test_object)
+            end
+
+            expect(has_pending_recordings?).to be true
+
+            described_class::Testing._native_finalize_pending_heap_recordings(stack_recorder)
+
+            expect(has_pending_recordings?).to be false
+          end
```
ivoanjo: Minor: I'm not sure these tests add a lot... Specifically, because they don't assert on any results, it's easy for them to pass while not doing the correct thing.

Furthermore, we have existing coverage where we already check if the objects we intend to sample/track do get sampled/track.

So... I'd be inclined to remove these tests (and the pending_heap_recordings? helper maybe as well?).

ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
r1viollet: reworking this atm, let's see if new version still has this defect

ext/datadog_profiling_native_extension/heap_recorder.c
r1viollet: I'd rather have the clear split here tbh

ext/datadog_profiling_native_extension/heap_recorder.c
r1viollet: I tried something with a:
    delayed_error(state, "Heap profiling: bignum object id detected. Heap profiling cannot continue.");

lib/datadog/profiling/component.rb
r1viollet: Makes sense, we will want to remove it soon, though we do want to give it a spin on relenv

r1viollet:
I took into account major comments.
Next step perf tests.
Then we can do style fixups.

ext/datadog_profiling_native_extension/heap_recorder.c
r1viollet: makes sense

ext/datadog_profiling_native_extension/heap_recorder.c
r1viollet: tried a slight refactoring, still fiddly

ext/datadog_profiling_native_extension/heap_recorder.c
r1viollet: feels weird to do so ^^

ext/datadog_profiling_native_extension/extconf.rb
r1viollet: I aligned on the other convention there was in that file.

ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
r1viollet: I'll let you reconsider new behaviour which has other issues :-).

ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
r1viollet: refactored.

spec/datadog/profiling/stack_recorder_spec.rb
```diff
@@ -699,6 +701,47 @@ def introduce_distinct_stacktraces(i, obj)
...
+          it "clears pending recordings after finalization" do
+            skip "Only applies to Ruby 4+ with deferred heap allocation recording" if RUBY_VERSION < "4"
+
+            test_object = Object.new
+
+            track_object_without_finalize(test_object)
+
+            expect(has_pending_recordings?).to be true
+
+            described_class::Testing._native_finalize_pending_heap_recordings(stack_recorder)
```
r1viollet, line 721: The test shows that you need these bits which is not ideal (but I still like that the test forces me to do this).

spec/datadog/profiling/stack_recorder_spec.rb
r1viollet: things slightly changed. I'll let you refer to comment above (in this file)

ext/datadog_profiling_native_extension/heap_recorder.c
r1viollet: As this is style, I'll leave it to the end.

ivoanjo:
Quick side-by-side results from Ruby 3.3 vs 4.0 in our Ruby on Rails test app:

> image

ivoanjo:
TODO: We're still investigating a few benchmarks / validation results, but this should be ready for review.

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -316,25 +341,47 @@ void start_heap_allocation_recording(heap_recorder *heap_recorder, VALUE new_obj
...
+    // The intuition here is: Usually we ask for an `after_allocation` callback only when the buffer is about to go
+    // from empty -> non-empty, because this is going to be mapped onto a postponed job, so after it gets queued once
+    // it doesn't seem worth it to keep spamming requests.
+    // Yet, if for some reason the postponed job doesn't flush the pending list (or if e.g. it ran with `during_sample == true ` and thus
+    // was skipped) we need to have some mechanism to recover -- and so if the buffer starts accumulating too much we
+    // start always requesting the callback to happen so that we eventually flush the buffer.
+    bool needs_after_allocation =
+      heap_recorder->pending_recordings_count == 0 || heap_recorder->pending_recordings_count >= MAX_PENDING_RECORDINGS / 2;
+
+    return needs_after_allocation;
```
AlexJF: Nit: I think this logic is tied with the `#ifdef` above (which is why the `return true` in line 352 mentions this code block. We could promote `needs_after_allocation` as a top-level variable with default value `false`, move this logic to the first `#ifdef` and leave this second `#ifdef` to focus on the capturing of the deferred object data.

I feel doing this makes it harder to forget to update things if our logic for `needs_after_allocation` changes in the future?

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -130,6 +140,15 @@ struct heap_recorder {
   // Data for a heap recording that was started but not yet ended
   object_record *active_recording;
 
+  // Pending recordings that need to be finalized after on_newobj_event completes.
+  // On Ruby 4+, we can't call rb_obj_id during the newobj event, so we store the
+  // VALUE reference here and finalize it via a postponed job.
+  pending_recording pending_recordings[MAX_PENDING_RECORDINGS];
+  // Temporary storage for the recording in progress, used between start and end
+  VALUE active_deferred_object;
+  live_object_data active_deferred_object_data;
```
AlexJF, line 149: Minor: This duplication of "responsibility/intent" between `active_recording` and `active_deferred` feels weird and error-prone (e.g. I'm not sure if rate-based skipping in line 343 plays well with your `#ifdef` handling logic on the end - after all, sounds possible for the end to not see a valid `active_deferred_object` reference in this case. Could we maybe just have `active_recording` be a `flexible_object_record ` which would be a `object_record` with either a `long obj_id;` or a `VALUE object_ref;`?

ext/datadog_profiling_native_extension/heap_recorder.c
```diff
@@ -399,6 +464,48 @@ void heap_recorder_update_young_objects(heap_recorder *heap_recorder) {
...
+    pending_recording *pending = &heap_recorder->pending_recordings[i];
+
+    // This is the step we couldn't do during the original sample call -- we're now expected to be called in a context
+    // where it's finally safe to call this
+    long obj_id = obj_id_or_fail(pending->object_ref);
+
+    // Create the object record now that we have the object_id
+    object_record *record = object_record_new(obj_id, pending->heap_record, pending->object_data);
+
+    commit_recording(heap_recorder, pending->heap_record, record);
```
r1viollet, line 494: 💭 This can raise, while object_record_new is not freed from above. Though I'm not sure duplicate IDs can happen.

r1viollet: Looks like the raise code path turn off everything, so this is fine.

ext/datadog_profiling_native_extension/heap_recorder.c
ivoanjo: Yup, makes sense, changed in e77f801cdc3b32b4cb4090199a7d280d18caa581

ext/datadog_profiling_native_extension/heap_recorder.c
ivoanjo: Yes, this is a very good point... and connecting with your other point...

> Also, if the deferred mechanism is also compatible with Ruby < 4, could it make sense to unify codepaths and have all be deferred? Or does this incur a perf penalty?

I'm definitely thinking of this being the next step. Sometimes there's situations where we can't escape from the separate codepaths in the profiler, but as much as possible for the ones where it's feasible to do it, we should!

This brings me back to...

> Could we maybe just have active_recording be a flexible_object_record  which would be a object_record with either a long obj_id; or a VALUE object_ref;?

My thinking is, on the path to using the new approach for all Rubies, it would be nice for testing if we could toggle between both as a setting/env var, rather than requiring compilation.

Such a setting should be straightforward -- effectively replace the `#ifdef USE_DEFERRED_HEAP_ALLOCATION_RECORDING` with a regular `if (use_new_codepath)`.

...but not if we unify `active_recording` with `active_deferred_object`. So that's why I'm hesitating on that one a bit.

(I do agree the in-between state is a potential footgun and I hope not to come back to this comment in a few months and say oops we missed a spot...)

> e.g. I'm not sure if rate-based skipping in line 343 plays well with your #ifdef handling logic on the end - after all, sounds possible for the end to not see a valid active_deferred_object reference in this case

We are -- on line 353 we respect the "setting the `active_recording`" for skipping, even on the new codepath. And in line 397, which is used by both codepaths, we reset the `active_recording`.

There was, however, a weird corner case I spotted after your call out -- we still had a check in `end_heap_allocation_recording` that made it look like we were checking one branch but not the other. But that could never happen -- whenever `active_recording == &SKIPPED_RECORD` we always reset it in `end_heap_allocation_recording_with_rb_protect` so the code inside `end_heap_allocation_recording` would never trigger in practice. So, as with any "will never trigger in practice", I've turned it into an exception in f06a73105b0dea7921bd1e28e1ec5d507ab3e431 to avoid future confusion.

ivoanjo: (I'm going to mark this as resolved to go ahead and merge the PR, but happy to iterate on this if you're not very convinced, as usual!)

ivoanjo:
I think this is good shape! Pressing the nice green button!

PR #5206, opened 2026-01-07
[APPSEC-60097] Add more data collection for downstream requests

Strech:
**What does this PR do?**

Add the first part of the [API10 RFC](https://docs.google.com/document/d/1DIGuCl1rkhx5swmGxKO7Je8Y4zvaobXBlgbm6C89yzU/edit?pli=1&tab=t.0). Where we supply more data to the WAF on downstream requests, more stats and metrics. Ideally we should process request/response body and track redirects.

1. Add new WAF addresses <------------- you are here
2. Add request/response body processing
3. Add best-effort redirects tracking

**Motivation:**

I want to start with small adjustments for RASP methods and metrics and see where it will lead without doing a refactoring, we are missing some other metrics and telemetry and before changing it would be good to see the full picture.

**Change log entry**

Yes. AppSec: Add analysis of the downstream requests.

**Additional Notes:**

First of all this is not a perfect implementation, we already notice an accumulation of the look a like methods and modules. And that will be addresses in rework of the event engine we using.

Second, this PR fixes few overlooks and enable (partially) ignore typespecs. With typespecs I decide to go as low as I can to avoid PR bloating and prepare a ticket to clean the rest.

**How to test the change?**

CI + [ST](https://github.com/DataDog/system-tests/pull/5955)

### DIFF
--- a/lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
+++ b/lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
@@ -18,7 +18,7 @@ module Datadog
             ephemeral_data = {
               'server.io.net.url' => url,
               'server.io.net.request.method' => method.to_s.upcase,
-              'server.io.net.request.headers' => lowercase_request_headers
+              'server.io.net.request.headers' => normalize_request_headers
             }
 
             result = context.run_rasp(Ext::RASP_SSRF, {}, ephemeral_data, timeout, phase: Ext::RASP_REQUEST_PHASE)
@@ -28,7 +28,7 @@ module Datadog
 
             ephemeral_data = {
               'server.io.net.response.status' => response.code.to_s,
-              'server.io.net.response.headers' => lowercase_response_headers(response)
+              'server.io.net.response.headers' => normalize_response_headers(response)
             }
 
             result = context.run_rasp(Ext::RASP_SSRF, {}, ephemeral_data, timeout, phase: Ext::RASP_RESPONSE_PHASE)
@@ -39,13 +39,16 @@ module Datadog
 
           private
 
-          def lowercase_request_headers
+          # NOTE: Starting version 2.1.0 headers are already normalized via internal
+          #       variable `@processed_headers_lowercase`. In case it's available,
+          #       we use it to avoid unnecessary transformation.
+          def normalize_request_headers
             return @processed_headers_lowercase if defined?(@processed_headers_lowercase)
 
             processed_headers.transform_keys(&:downcase)
           end
 
-          def lowercase_response_headers(response)
+          def normalize_response_headers(response)
             # NOTE: Headers values are always an `Array` in `Net::HTTPResponse`,
             #       but we want to avoid accidents and will wrap them in no-op
             #       `Array` call just in case of a breaking change in the future


Strech:
@codex find critical bugs and corner cases in this PR's code

lib/datadog/appsec/contrib/faraday/ssrf_detection_middleware.rb
```diff
@@ -10,33 +10,49 @@ module Contrib
...
+            result = context.run_rasp(Ext::RASP_SSRF, {}, ephemeral_data, timeout, phase: Ext::RASP_REQUEST_PHASE)
+            handle(result) if result.match?
+
+            response = @app.call(env)
 
             ephemeral_data = {
-              'server.io.net.url' => request_env.url.to_s
+              'server.io.net.response.status' => response.status.to_s,
+              'server.io.net.response.headers' => response.headers.transform_keys(&:downcase)
             }
```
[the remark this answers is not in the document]
Strech: 👍🏼

lib/datadog/appsec/contrib/excon/ssrf_detection_middleware.rb
```diff
@@ -13,27 +13,62 @@ module Excon
         # AppSec Middleware for Excon
         class SSRFDetectionMiddleware < ::Excon::Middleware::Base
           def request_call(data)
-            return super unless AppSec.rasp_enabled? && AppSec.active_context
+            context = AppSec.active_context
+            return super unless context && AppSec.rasp_enabled?
+
+            timeout = Datadog.configuration.appsec.waf_timeout
```
y9v, line 19: not related to this PR, but an idea for future - how about we remove the `timeout` argument from `#run_rasp` and `#run_waf`, since it's always read from the configuration?

lib/datadog/appsec/contrib/excon/ssrf_detection_middleware.rb
```diff
@@ -13,27 +13,62 @@ module Excon
...
+          def request_url(data)
+            klass = (data[:scheme] == 'https') ? URI::HTTPS : URI::HTTP
+            klass.build(host: data[:host], path: data[:path], query: data[:query]).to_s
+          end
+
+          def normalize_headers(headers)
+            return {} if headers.nil? || headers.empty?
+
+            headers.each_with_object({}) do |(key, value), memo|
+              memo[key.downcase] = value.is_a?(Array) ? value.join(', ') : value
```
y9v, line 59: nit: do we want a space after a comma here when joining?

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
```diff
@@ -11,29 +11,59 @@ module RestClient
...
+          def lowercase_request_headers
+            return @processed_headers_lowercase if defined?(@processed_headers_lowercase)
 
-            if result.match?
-              AppSec::Event.tag(context, result)
-              TraceKeeper.keep!(context.trace) if result.keep?
+            processed_headers.transform_keys(&:downcase)
+          end
+
+          def lowercase_response_headers(response)
```
y9v: it's not very clear why this method is called `lowercase_response_headers` - it's not downcasing header keys, just joining values

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
```diff
@@ -11,29 +11,59 @@ module RestClient
...
 
-            ephemeral_data = {'server.io.net.url' => url}
-            result = context.run_rasp(Ext::RASP_SSRF, {}, ephemeral_data, Datadog.configuration.appsec.waf_timeout)
+            response
+          end
+
+          private
+
+          def lowercase_request_headers
+            return @processed_headers_lowercase if defined?(@processed_headers_lowercase)
```
y9v, line 46: where is `@processed_headers_lowercase` defined? In the instrumented class? I think it might make sense to add a comment here.

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
```diff
@@ -11,29 +11,59 @@ module RestClient
...
+            handle(result, context: context) if result.match?
 
-            ephemeral_data = {'server.io.net.url' => url}
-            result = context.run_rasp(Ext::RASP_SSRF, {}, ephemeral_data, Datadog.configuration.appsec.waf_timeout)
+            response
+          end
+
+          private
+
+          def lowercase_request_headers
```
y9v: not sure about using `lowercase`, since ruby always uses `downcase` - but don't mind leaving it as it is

y9v:
nice work! Thanks for fixing the url for excon instrumentation!

lib/datadog/appsec/contrib/excon/ssrf_detection_middleware.rb
Strech: Yes, that's how other libs are doing that

lib/datadog/appsec/contrib/excon/ssrf_detection_middleware.rb
Strech: Yes, I think we should do that because it became a pattern everywhere, makes sense to hide it! If you don't mind let's do it separate PR

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
Strech: As humans we say upper-case and lower-case, but downcase is a Ruby specific implementation details. It's better to stick to general form.

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
Strech: Makes sense - it's a framework specific, which is actually only on latest version, will leave the comment 👍🏼

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
Strech: My bad, will update the name

lib/datadog/appsec/contrib/excon/ssrf_detection_middleware.rb
y9v: yes, of course, it's definitely not part of this PR

y9v (approved):

PR #5208, opened 2026-01-07
Add DBM back propagation with container tags and process tags

wantsui:
**What does this PR do?**

The goal of this is to add container tags support to DBM, which also includes the process tags: https://docs.google.com/document/d/15GtNOKGBCt6Dc-HsDNnMmCdZwhewFQx8yUlI9in5n3M/edit?tab=t.0#heading=h.fyt7z8ehztcm.

To achieve this, the following changes were made
1. The existing FNV hash logic was refactored out so it could be reused for this update
2. AgentInfo was updated to compute a hash if the Datadog Agent returns a `Datadog-Container-Tags-Hash` header. The hash combines process_tags and container_tags_hash.
3. When agent_info is received, compute a new hash if the container information is new
4. Injects `ddsh='...'` as a SQL comment
5. Injects `_dd.propagated_hash` as a span "meta"

**Motivation:**

The Python PR does this for DBM and DSM in one go: 
- https://github.com/DataDog/dd-trace-py/pull/15293

The Java PR does this in two:
- https://github.com/DataDog/dd-trace-java/pull/9156
- https://github.com/DataDog/dd-trace-java/pull/9282

For Ruby, I am adding all the functionality needed for the back propagation for DBM first then will do DSM in a later PR.

**Change log entry**

No.

**Additional Notes:**

I am not sure if this requires a changelog entry.

**How to test the change?**

```
bundle exec rspec \
spec/datadog/core/utils/fnv_spec.rb \
spec/datadog/core/environment/agent_info_spec.rb \
spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
```

I also sent data from this branch to our backend and under query samples with a default number in place of the propagation hash to represent both the process tags and containers, ie: https://app.datadoghq.com/databases/samples

The result is that:

1) the db comment has ddsh added:
```
"/*dde='dev',ddps='dbm-sql-comment-demo',ddpv='0.1',ddh='postgres',dddb='postgres',ddsh='1234567890',dddbs='pg',traceparent='<some string>'*/"
```
2) the "span" gets `"_dd.propagated_hash": 1234567890` under `meta`

### DIFF
--- a/lib/datadog/core/environment/agent_info.rb
+++ b/lib/datadog/core/environment/agent_info.rb
@@ -128,9 +128,6 @@ module Datadog
             # new_container_tags_value (non nil) over @container_tags_checksum (string?) to avoid steep errors
             data = process_tags + new_container_tags_value
             @propagation_checksum = Core::Utils::FNV.fnv1_64(data)
-          elsif !defined?(@propagation_checksum)
-            # Cache nil to avoid repeated fetch attempts when agent doesn't provide the header
-            @propagation_checksum = nil
           end
         end
       end


lib/datadog/core/utils/fnv.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Core
+    module Utils
+      module FNV
+        # FNV-1a 64-bit hash function.
```
marcotc:
```suggestion
        # FNV-1a 64-bit hash function.
        # @see https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function#FNV-1a_hash Algorithm
        # @see https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function#FNV_hash_parameters Prime and Offset Basis
```

lib/datadog/core/environment/base_hash.rb
```diff
@@ -0,0 +1,57 @@
...
+
+require_relative '../utils/fnv'
+require_relative 'process'
+
+module Datadog
+  module Core
+    module Environment
+      # Creates a base hash that combines process and container tags
+      # This is currently needed for DBM for back propagation.
+      module BaseHash
```
marcotc: I'm not sure if this module should exist, since it's a mix of computation and caching for a hash that is directly calculated from AgentInfo.

And because AgentInfo is tied to the lifecycle of Components: https://github.com/DataDog/dd-trace-rb/blob/6db1463ef9441bdd581efdafcb88e7da2dbbf56d/lib/datadog/core/configuration/components.rb#L127
This hash should also be recreated when components are recreated.

I see two options here:
1. Store the hash in a new Component, if we think it deserves its own top-level entity ([these are all the components today](https://github.com/DataDog/dd-trace-rb/blob/6db1463ef9441bdd581efdafcb88e7da2dbbf56d/lib/datadog/core/configuration/components.rb#L96-L112)).
2. Store it in another component. The best, but not perfect, candidate today is storing it in the AgentInfo.

I think that for simplicity, we can store it in the AgentInfo today.
This way it is recalculated this hash when the AgentInfo gets new data from the agent, and the hash cache is automatically busted on AgentInfo recreation (which can mean that the agent host/port has changed).

lib/datadog/core/environment/agent_info.rb
```diff
@@ -51,27 +53,43 @@ module Environment
...
         # @return [Datadog::Core::Remote::Transport::HTTP::Negotiation::Response] the response from the agent
         # @return [nil] if an error occurred while fetching the information
         def fetch
           res = @client.send_info
           return unless res.ok?
 
+          extract_container_tags_hash(res)
+
+          # if there are container tags, compute the base hash
+          Environment::BaseHash.compute(container_tags_hash) if container_tags_hash
```
marcotc: `#fetch` is not guaranteed to be called before `lib/datadog/tracing/contrib/propagation/sql_comment.rb` runs.

Today, the only user of `AgentInfo`, `Transport`, always calls `#fetch` when it reads its value: https://github.com/DataDog/dd-trace-rb/blob/6db1463ef9441bdd581efdafcb88e7da2dbbf56d/lib/datadog/tracing/transport/traces.rb#L186

This is a less of a problem for `Transport` since it does not run on the application critical path; it runs on the background trace flushing thread, which is naturally rate limited, running at most once per second.

I recommend adding a new method to AgentInfo called `def propagation_hash`, which either returns the cached value or fetches if not cached. There's no need to protect this in a mutex, since the races here would be rate and the agent request is a very cheap endpoint; a mutex on every read is also expensive.

We should still compute the propagation hash on direct calls to `fetch`, since that can happen if `Transport` is called before any SqlComment code is called.

lib/datadog/tracing/contrib/propagation/sql_comment/ext.rb
```diff
@@ -20,6 +20,9 @@ module Ext
             # The value should be `true` when `full` mode
             TAG_DBM_TRACE_INJECTED = '_dd.dbm_trace_injected'
 
+            # Used for the propagated hash on span metric
```
marcotc:
```suggestion
            # Hash of the agent's container tags and this process' tags
```

lib/datadog/core/environment/agent_info.rb
```diff
@@ -51,27 +53,43 @@ module Environment
...
           other.is_a?(self.class) && other.agent_settings == agent_settings
         end
+
+        private
+
+        def extract_container_tags_hash(res)
+          return unless res.respond_to?(:headers)
+
+          header_value = res.headers[Core::Transport::Ext::HTTP::HEADER_CONTAINER_TAGS_HASH]
+          @container_tags_hash = header_value if header_value && !header_value.empty?
```
Strech: This is a bad practice because it's an implicit assignment and side-effect. Better to return value and assign in at the place of an actual call

lib/datadog/core/environment/base_hash.rb
Strech: I also think that name doesn't reflect the use-case. 

Agree with Marco that this class is doing 2 things, and composition always wins (meaning it's better to have smaller composable classes, than 1 class that is hard to reuse).

Hence logically, it makes sense to separate storage (caching) from computation, and maybe we can find a better storage place, but computation could be extracted.

lib/datadog/core/utils/fnv.rb
```diff
@@ -0,0 +1,24 @@
...
```
Strech: I think it's important to mention that it's a "non-cryptographic hash function"

sig/datadog/core/environment/base_hash.rbs
Strech, line 1: Formatting here went off

spec/datadog/core/environment/agent_info_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  describe '#container_tags_hash' do
+    context 'when the header is missing' do
+      before do
+        allow(response).to receive(:headers).and_return({})
+      end
+      it 'returns nil' do
+        agent_info.fetch
+        expect(agent_info.container_tags_hash).to be nil
+      end
```
Strech, line 102:
```suggestion
      before { allow(response).to receive(:headers).and_return({}) }
      
      it 'returns nil' do
        agent_info.fetch
        expect(agent_info.container_tags_hash).to be nil
      end
```

spec/datadog/core/environment/agent_info_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      it 'does not compute the base hash' do
+        expect(Datadog::Core::Environment::BaseHash).not_to receive(:compute)
+        agent_info.fetch
+      end
+    end
+    context 'when the header is present' do
```
Strech, line 117:
```suggestion
    end

    context 'when the header is present' do
```

Strech:
Formatting on RBS files is off

lib/datadog/core/environment/agent_info.rb
```diff
@@ -51,27 +53,43 @@ module Environment
       #
       # @see https://github.com/DataDog/datadog-agent/blob/f07df0a3c1fca0c83b5a15f553bd994091b0c8ac/pkg/trace/api/info.go#L20
       class AgentInfo
-        attr_reader :agent_settings, :logger
+        attr_reader :agent_settings, :logger, :container_tags_hash
```
p-datadog: I think the `container_tags_hash` should be a separate line with a comment saying when (or how) to initialize it, since it's not guaranteed to return  a non-nil value.

lib/datadog/core/environment/base_hash.rb
```diff
@@ -0,0 +1,57 @@
...
+        @current_hash = nil
+        @current_container_hash = nil
+
+        class << self
+          # Only compute a container hash if the agent returns a new Datadog-Container-Tags-Hash header
+          #
+          # @param container_tags_hash [String] hash received by the agent header
+          # @return [Integer, nil] the new FNV hash value or nil
+          def compute(container_tags_hash)
+            return current if container_tags_hash.nil? || container_tags_hash.empty?
```
p-datadog: I find this logic brittle, why not have one method which computes the hash which would reject invalid input (e.g. nil) and another method which returns the current value?

p-datadog: As far as I see the return value of `compute` is not used in this PR, if there aren't other callers planned, I would simply not claim a particular return value for this method (and return nil here instead of `current`).

lib/datadog/core/environment/base_hash.rb
```diff
@@ -0,0 +1,57 @@
...
+        @mutex = Mutex.new
+        @current_hash = nil
+        @current_container_hash = nil
+
+        class << self
+          # Only compute a container hash if the agent returns a new Datadog-Container-Tags-Hash header
+          #
+          # @param container_tags_hash [String] hash received by the agent header
+          # @return [Integer, nil] the new FNV hash value or nil
+          def compute(container_tags_hash)
```
p-datadog: Rename to `update` ?

lib/datadog/core/environment/base_hash.rb
wantsui: Thanks for the feedback! I ended up refactoring it into Agent Info in 5bfe76cd4db078f12292a6701b03f56aabfdb918 . Let me know if you see any possible issues with this refactor.

lib/datadog/core/environment/agent_info.rb
wantsui: Done in https://github.com/DataDog/dd-trace-rb/pull/5208/commits/5bfe76cd4db078f12292a6701b03f56aabfdb918!

lib/datadog/core/environment/base_hash.rb
wantsui: I completely removed all the basehash logic in https://github.com/DataDog/dd-trace-rb/pull/5208/commits/5bfe76cd4db078f12292a6701b03f56aabfdb918, so I'll resolve this comment. The new version of this is under `update_propagation_hash` in the agent_info.rb file.

lib/datadog/core/environment/base_hash.rb
wantsui: Done in https://github.com/DataDog/dd-trace-rb/pull/5208/commits/5bfe76cd4db078f12292a6701b03f56aabfdb918, though now it's in agent_info as `update_propagation_hash` :)

sig/datadog/core/environment/base_hash.rbs
wantsui: Resolving since I removed this file in https://github.com/DataDog/dd-trace-rb/pull/5208/commits/5bfe76cd4db078f12292a6701b03f56aabfdb918!

lib/datadog/core/utils/fnv.rb
wantsui: Done in https://github.com/DataDog/dd-trace-rb/pull/5208/commits/5bfe76cd4db078f12292a6701b03f56aabfdb918!

lib/datadog/core/environment/agent_info.rb
wantsui: Thanks for pointing this out, I remove this line in https://github.com/DataDog/dd-trace-rb/pull/5208/commits/5bfe76cd4db078f12292a6701b03f56aabfdb918, let me know if this is an improvement!

lib/datadog/core/environment/agent_info.rb
```diff
@@ -52,26 +55,66 @@ module Environment
...
           res = @client.send_info
           return unless res.ok?
 
+          new_container_tags_hash = extract_container_tags_hash(res)
+
+          # if there are new container tags, regenerate the hash
+          if new_container_tags_hash && new_container_tags_hash != @container_tags_hash
+            @container_tags_hash = new_container_tags_hash
+            update_propagation_hash
+          end
```
marcotc: I think we can hide all this inside a private method, to clean up `fetch`:
```suggestion
          update_propagation_hash(res)
```
(Feel free to change the private methods as you see fit)

lib/datadog/core/utils/fnv.rb
```diff
@@ -0,0 +1,24 @@
...
+module Datadog
+  module Core
+    module Utils
+      module FNV
+        # FNV-1a 64-bit non-cryptographic hash function.
+        # @see https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function#FNV-1a_hash Algorithm
+        # @see https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function#FNV_hash_parameters Prime and Offset Basis
+        def self.fnv1_64(data)
+          fnv_offset_basis = 14695981039346656037
+          fnv_prime = 1099511628211
```
marcotc: These look like they are better represented as class constants.

In this specific case, this has a practical performance impact since 14695981039346656037 and 0xFFFFFFFFFFFFFFFF are too large to fit in a VALUE, thus a new object is created for them in the heap every time, if they stay as local variables:
```
require 'objspace'; ObjectSpace.memsize_of(14695981039346656037)
# => 40
```

For reference, 1099511628211 fits in a VALUE, thus is fully allocated in the execution stack:
```ruby
require 'objspace'; ObjectSpace.memsize_of(1099511628211)
# => 0
```

wantsui: Let me know if 55577f818aa623b8b11778a64747693112dd3317 addresses this correctly!

lib/datadog/core/environment/agent_info.rb
wantsui: Refactored in 55577f818aa623b8b11778a64747693112dd3317, let me know if this is what you mean!

lib/datadog/core/environment/agent_info.rb
```diff
@@ -52,26 +55,64 @@ module Environment
...
         attr_reader :agent_settings, :logger
+        # Container tags hash originally set to nil, but gets populated from #fetch when available
+        attr_reader :container_tags_hash
 
         def initialize(agent_settings, logger: Datadog.logger)
           @agent_settings = agent_settings
           @logger = logger
           @client = Remote::Transport::HTTP.root(agent_settings: agent_settings, logger: logger)
+          @container_tags_hash = nil
+          @propagation_hash = nil
```
y9v: I think we do not need to initialize this variable here?

lib/datadog/core/environment/agent_info.rb
```diff
@@ -52,26 +55,64 @@ module Environment
...
+        end
+
+        private
+
+        def extract_container_tags_hash(res)
+          return unless res.respond_to?(:headers)
+
+          header_value = res.headers[Core::Transport::Ext::HTTP::HEADER_CONTAINER_TAGS_HASH]
+          header_value if header_value && !header_value.empty?
+        end
```
y9v: I think it might be better for readability to not extract this code into a separate function, and keep it in `process_container_tags_hash`

lib/datadog/core/environment/agent_info.rb
```diff
@@ -52,26 +55,64 @@ module Environment
...
+
+        def process_container_tags_hash(res)
+          new_container_tags_hash = extract_container_tags_hash(res)
+
+          # if there are new container tags from the agent, regenerate the hash
+          if new_container_tags_hash && new_container_tags_hash != @container_tags_hash
+            @container_tags_hash = new_container_tags_hash
+
+            process_tags = Process.serialized
+            data = process_tags + new_container_tags_hash
```
y9v: I don't understand, how is `+` working on a hash here?

```ruby
{a: 1} + {b: 2}
(irb):1:in '<main>': undefined method '+' for an instance of Hash (NoMethodError)
```

lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -14,6 +14,10 @@ module SqlComment
           def self.annotate!(span_op, mode)
             return unless mode.enabled?
 
+            # Add base hash to the span metric
+            base_hash = Datadog.send(:components).agent_info.propagation_hash
```
y9v: is `agent_info` always present here?

lib/datadog/core/environment/agent_info.rb
```diff
@@ -52,26 +55,64 @@ module Environment
...
 
         def ==(other)
           other.is_a?(self.class) && other.agent_settings == agent_settings
         end
+
+        # Returns the propagation hash from the Agent if not previously cached
+        # @return [Integer, nil] the FNV hash based on the container and process tags or nil
+        def propagation_hash
+          return @propagation_hash if @propagation_hash
+          fetch if @container_tags_hash.nil?
```
Strech: This is an implicit assignment, I've pointed that out in previous review and it should be return value assignment instead, i.e

```
return @propagation_hash if @propagation_hash

@propagation_hash = func_call()
```

lib/datadog/tracing/contrib/propagation/sql_comment.rb
marcotc: Yes, it's a component in Core, and cannot be skipped.

lib/datadog/core/environment/agent_info.rb
wantsui: Thanks for pointing out how confusing this is, I'll rename the variable since both of these are just string values being added :)

lib/datadog/core/utils/fnv.rb
```diff
@@ -0,0 +1,26 @@
...
+
+        FNV_OFFSET_BASIS = 14695981039346656037
+        FNV_PRIME = 1099511628211
+        FNV_64_BIT_MASK = 0xFFFFFFFFFFFFFFFF
+
+        def self.fnv1_64(data)
+          hash_value = FNV_OFFSET_BASIS
+          data.each_byte do |byte|
+            hash_value ^= byte
+            hash_value = (hash_value * FNV_PRIME) & FNV_64_BIT_MASK
```
wantsui, line 19: Note that Python is doing something different when doing this hash (the reverse): https://github.com/DataDog/dd-trace-py/blob/5a3e90aab8f6535519e9a68f3198ee853c16e12e/ddtrace/internal/utils/fnv.py#L30-L31

```
    for byte in data:
        hval = (hval * fnv_prime) % fnv_size
        hval = hval ^ _get_byte(byte)
    return hval
```
cc @vandonr @marcotc @dubloom

wantsui: If we need all the hashes to be the same then Ruby actually needs to follow this operation also for back propagation 🤔

lib/datadog/core/environment/agent_info.rb
wantsui: Removed in https://github.com/DataDog/dd-trace-rb/pull/5208/commits/21d49716e4f3816f5ef5ec99ef62668e9782b827

lib/datadog/core/environment/agent_info.rb
wantsui: I'm still refactoring a bit but I got rid of extract_container_tags_hash: https://github.com/DataDog/dd-trace-rb/pull/5208/commits/21d49716e4f3816f5ef5ec99ef62668e9782b827

lib/datadog/core/environment/agent_info.rb
wantsui: I renamed the variables in this commit https://github.com/DataDog/dd-trace-rb/pull/5208/commits/21d49716e4f3816f5ef5ec99ef62668e9782b827, to better show that the tags_hash is really just a string at this point.

lib/datadog/core/environment/agent_info.rb
wantsui: Refactored in https://github.com/DataDog/dd-trace-rb/pull/5208/commits/21d49716e4f3816f5ef5ec99ef62668e9782b827

lib/datadog/core/environment/agent_info.rb
```diff
@@ -60,18 +65,50 @@ def initialize(agent_settings, logger: Datadog.logger)
...
+          container_tags_hash = @container_tags_hash
+          return nil unless container_tags_hash
+
+          process_tags = Process.serialized
+          data = process_tags + container_tags_hash
+          @propagation_hash = Core::Utils::FNV.fnv1_64(data)
```
wantsui: @vandonr - Ruby will hash the process tags + container tags hash so the `ddsh` will be another hash, which is what @dubloom did for Python. Let me know if this look right to you for the approach :)

lib/datadog/core/environment/agent_info.rb
y9v: in 21d4971 I see that only `new` was removed:

```ruby
data = process_tags + container_tags_hash
```

Could we name it `container_tags_hash_str` or something similar?

lib/datadog/core/environment/agent_info.rb
```diff
@@ -52,6 +55,8 @@ module Environment
       # @see https://github.com/DataDog/datadog-agent/blob/f07df0a3c1fca0c83b5a15f553bd994091b0c8ac/pkg/trace/api/info.go#L20
       class AgentInfo
         attr_reader :agent_settings, :logger
+        # Container tags hash originally set to nil, but gets populated from #fetch when available
+        attr_reader :container_tags_hash
```
Strech: [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) has a big disadvantage for refactoring, we are exposing internal structure via name and each change to the structure will lead to the change of all the places of use.

lib/datadog/core/utils/fnv.rb
wantsui: Resolving this since after asking around, @vandonr mentioned that the hashes don't need to match across languages.

lib/datadog/core/environment/agent_info.rb
marcotc: Hash here means "hash digest", not a Ruby Hash, so it represents the nature of the data.

But this is the second time in this PR that the name "hash" has caused some confusion, maybe we should change it for clarity?

Some options (including current state):
1. hash: can be confused with a Ruby [Hash](https://docs.ruby-lang.org/en/4.0/Hash.html).
2. digest: can be confused with a [TraceDigest](https://github.com/DataDog/dd-trace-rb/blob/fc10410daf1e5bef23dd5cfa436bf954bad5ae28/lib/datadog/tracing/trace_digest.rb#L9C11-L9C22)?
3. hash code?
4. hash digest?
5. checksum?

Strech: Thanks Marco, yep, this is another reason to change `_hash` into something more specific. I like `_checksum` because it doesn't need a hash prefix

y9v: I also like `_checksum`

p-datadog: +1 for "checksum"

lib/datadog/core/environment/agent_info.rb
```diff
@@ -60,18 +65,50 @@ def initialize(agent_settings, logger: Datadog.logger)
...
+          update_container_tags_hash(res)
+
           res
         end
 
         def ==(other)
           other.is_a?(self.class) && other.agent_settings == agent_settings
         end
+
+        # Returns the propagation hash from the Agent if not previously cached
```
p-datadog:
```suggestion
        # Returns the propagation hash from the Agent.
```
My reading of the diff is that the (hash->)checksum from agent is always returned. Whether the value is cached is an implementation detail that is probably not relevant to the callers.

lib/datadog/core/environment/agent_info.rb
```diff
@@ -60,18 +65,50 @@ def initialize(agent_settings, logger: Datadog.logger)
...
+        # @return [Integer, nil] the FNV hash based on the container and process tags or nil
+        def propagation_hash
+          return @propagation_hash if @propagation_hash
```
p-datadog: I think you can do

```suggestion
          return @propagation_hash if defined?(@propagation_hash)
```

And from the next line onward the code should only run if this is the first execution.

lib/datadog/core/environment/agent_info.rb
```diff
@@ -60,18 +65,50 @@ def initialize(agent_settings, logger: Datadog.logger)
...
+          fetch if @container_tags_hash.nil?
+          container_tags_hash = @container_tags_hash
+          return nil unless container_tags_hash
```
p-datadog: `@container_tags_hash` appears to be written to but never read outside of this method. If this is correct, I keeping it as a local (not instance variable) so that the only state is stored in `@propagation_hash` (and I assume renamed hash -> checksum).

lib/datadog/core/environment/agent_info.rb
```diff
@@ -60,18 +65,50 @@ def initialize(agent_settings, logger: Datadog.logger)
...
+
+          process_tags = Process.serialized
+          data = process_tags + container_tags_hash
+          @propagation_hash = Core::Utils::FNV.fnv1_64(data)
+        end
+
+        private
+
+        def update_container_tags_hash(res)
+          return unless res.respond_to?(:headers)
```
p-datadog: What is `res` permitted to be then? This condition is allowing arbitrary objects like strings, numbers, any random input to be fed to `update_container_tags_hash` with no errors or diagnostics (but probably not with intended results).

I find it is much better to only call methods with valid input, and have the methods fail on invalid input, instead of ignoring invalid input in the method implementations. The fail-on-invalid-input is generally how most of Ruby code works (standard library, third-party libraries, etc.).

lib/datadog/core/environment/agent_info.rb
```diff
@@ -60,18 +65,50 @@ def initialize(agent_settings, logger: Datadog.logger)
...
+
+          header_value = res.headers[Core::Transport::Ext::HTTP::HEADER_CONTAINER_TAGS_HASH]
+          new_container_tags_value = header_value if header_value && !header_value.empty?
+
+          # if there are new container tags from the agent,
+          # set the hash to nil so it gets recomputed the next time the hash string is created
+          if new_container_tags_value && new_container_tags_value != @container_tags_hash
+            @container_tags_hash = new_container_tags_value
```
p-datadog: I read the diff and I did not understand the interaction of container_tags_hash write here with the other method. Can this be explicitly described somewhere?

lib/datadog/core/environment/agent_info.rb
```diff
@@ -52,6 +55,8 @@ module Environment
       # @see https://github.com/DataDog/datadog-agent/blob/f07df0a3c1fca0c83b5a15f553bd994091b0c8ac/pkg/trace/api/info.go#L20
       class AgentInfo
         attr_reader :agent_settings, :logger
+        # Container tags hash originally set to nil, but gets populated from #fetch when available
```
p-datadog: Can this comment be expanded to be understandable to someone who is less familiar with this functionality, specifically can it be rewritten to reference higher level concepts - `fetch` is an low-level implementation method. In ordinary use, are container tags always available? Sometimes available? What are the conditions when they become available? Never available unless an action is taken? What is that action?

lib/datadog/core/environment/agent_info.rb
wantsui: Resolving this because it turns out the FNV encoding doesn't have to be the same across tracers but I'll leave a comment inside the code stating that we just need the checksum to be the same on the span tag and also the DBM data.

lib/datadog/core/environment/agent_info.rb
wantsui: Renaming to container_tags_checksum, thanks all! See: c7ef20de753ddbae3a769cf968b26ba41055de4f

lib/datadog/core/environment/agent_info.rb
marcotc: I suggest making it private, since it's use only for testing.

lib/datadog/core/environment/agent_info.rb
marcotc: The `res` will always be a  response here, so we can avoid this defensive check altogether.

lib/datadog/core/environment/agent_info.rb
```diff
@@ -59,19 +62,61 @@ def initialize(agent_settings, logger: Datadog.logger)
...
+
+        # Datadog::Core::Environment::Container extracts the container id if possible and sends them to the Trace Agent via the header Datadog-Container-ID
+        # The Trace Agent takes the container id and looks for matching container tags to compute a SHA256 checksum via the response header DATADOG-CONTAINER-TAGS-HASH
+        # https://github.com/DataDog/datadog-agent/blob/c923da011c8e51c35c0d05b6b10d016521915e7d/pkg/trace/api/info.go#L203-L227
+        # When deciding whether the propagation checksum should be updated, we need to be aware of some concerns
+        #     - It is possible that older Trace Agents may not have this specific header
+        #     - It is possible that we don't have access to the value if the Trace Agent is temporarily down. In these cases, we need to check for the value again on the next call to the info endpoint
+        #     - If we have access to the value, we need to check if it changed from the previous value.
+        def update_container_tags(res)
+          header_value = res.headers[Core::Transport::Ext::HTTP::HEADER_CONTAINER_TAGS_HASH]
```
wantsui, line 120: I removed the early return and it turned out it surfaced an error with the HTTP Negotiation Response where it was missing headers
```
ces') Error during traces flush: dropped 1 items. Cause: undefined method `headers' for an instance of Datadog::Core::Remote::Transport::HTTP::Negotiation::Response Location: /app/lib/datadog/core/environment/agent_info.rb:111:in `update_container_tags'
```

wantsui: Fixed with a232339ea1967edfac28d31c5cf9a40e3eca637b where I added the missing code to lib/datadog/core/transport/http/response.rb

lib/datadog/core/environment/agent_info.rb
wantsui: Let me know if the new comments in 1b73deda11f4412ba08f6c2f5147a43f471a4ed9 helps! I can clarify anything that's unclear.

wantsui: To answer it in this comment as well, the container ids are extracted from the cgroup folder, and not availabe in cgroupv2 at the moment, but I think this is a comment I can add to the containers code rather than in the agent info code. I can do this in a separate PR though.

Once a tracer sends the container id to the Trace Agent, the Trace Agent uses this ID to find the container tags it is aware of and computes a sha256 hash of them to send back to the tracers.

wantsui: In the past I've seen the Trace Agent fail to return the container tags correctly if:
- The tracer fails to send the container id in the first place (which can happen in newer Docker versions). I believe there's work related to origin detection to handle this.
-  The Trace Agent runs into permission/setups issues related to the Docker or whatever container integration is in use that allows it to grab container tags in the first place

lib/datadog/core/environment/agent_info.rb
```diff
@@ -59,19 +62,62 @@ def initialize(agent_settings, logger: Datadog.logger)
...
           other.is_a?(self.class) && other.agent_settings == agent_settings
         end
+
+        # Returns the propagation hash from the Agent if not previously cached
+        # Currently called/used by the DBM code to inject the propagation hash into the SQL comment
+        # @return [Integer, nil] the FNV hash based on the container and process tags or nil
+        def propagation_hash
+          # Can't use defined?(@propagation_hash) here because it will return true if @propagation_hash is nil
+          return @propagation_hash if @propagation_hash
+          fetch if @container_tags_checksum.nil?
```
Strech: Sorry if I pointed it our 3rd time, but this diff shows me that this is still an implicit assignment. Is it Github diff issue, or it's still that moment I've pointed out?

lib/datadog/core/environment/agent_info.rb
```diff
@@ -59,19 +62,62 @@ def initialize(agent_settings, logger: Datadog.logger)
...
+          container_tags_checksum = @container_tags_checksum
+          return nil unless container_tags_checksum
```
Strech:
```suggestion
          return unless @container_tags_checksum
```

lib/datadog/core/environment/agent_info.rb
```diff
@@ -59,19 +62,62 @@ def initialize(agent_settings, logger: Datadog.logger)
...
+          data = process_tags + container_tags_checksum
+          @propagation_hash = Core::Utils::FNV.fnv1_64(data)
+        end
+
+        private
+
+        # Trace Agent 7.69.0+ provides a SHA256 checksum in the response header DATADOG-CONTAINER-TAGS-HASH based on the container id computed in Datadog::Core::Environment::Container
+        # During calls to the Trace Agent, this checksum is cached but invalidated if a new value is returned
+        # The resulting propagation_hash uses the container_tags_checksum
+        # https://github.com/DataDog/datadog-agent/pull/38515
```
marcotc, line 105: I was thinking of adding some information on why this value exist, in addition to how it's calculated, something like:

```
        # This is a short checksum that uniquely identifies this process, its container environment, and 
        # the Datadog agent it connects to.
        #
        # It is used to correlate multiple signals emitted by this process among themselves (e.g. traces to
        # sql queries to pub/sub events).
        # Because some of these signals have strict size restrictions, we cannot use complete propagation
        # methods (e.g. a W3C Trace Context), so short checksum is calculated, then later correlated by the
        # Datadog App.
        #
        # This checksum only has to be internally consistent: the same value must be used by every signal
        # emitted by this process+container+agent combinations). It is not required that this checksum is
        # consistent with other SDKs.
```

lib/datadog/core/environment/agent_info.rb
wantsui: Sorry, I'm still refactoring some changes around based on the other comments so I haven't fully addressed this yet. It's still on my todo!

lib/datadog/core/environment/agent_info.rb
wantsui: I removed the defensive check in https://github.com/DataDog/dd-trace-rb/pull/5208/commits/1b73deda11f4412ba08f6c2f5147a43f471a4ed9, but then it surfaced up an actual error in transport/http.rb in that "headers" was never defined which is now fixed in: a232339ea1967edfac28d31c5cf9a40e3eca637b. 

I'll mark this as resolved but feel free to reopen if it isn't!

lib/datadog/core/environment/agent_info.rb
wantsui: Added in 84e119a482bdb918f4ec85b85b0a7a77a976bd88

lib/datadog/core/environment/agent_info.rb
```diff
@@ -59,19 +62,74 @@ def initialize(agent_settings, logger: Datadog.logger)
...
         def ==(other)
           other.is_a?(self.class) && other.agent_settings == agent_settings
         end
+
+        # Returns the propagation hash from the Agent.
+        # Currently called/used by the DBM code to inject the propagation hash into the SQL comment
+        # @return [Integer, nil] the FNV hash based on the container and process tags or nil
+        def propagation_hash
+          # Can't use defined?(@propagation_hash) here because it will return true if @propagation_hash is nil
+          return @propagation_hash if @propagation_hash
```
marcotc:
```suggestion
          return @propagation_hash if defined?(@propagation_hash) && @propagation_hash
```

lib/datadog/core/environment/agent_info.rb
```diff
@@ -59,19 +62,73 @@ def initialize(agent_settings, logger: Datadog.logger)
...
+        #     - It is possible that older Trace Agents may not have this specific header
+        #     - It is possible that we don't have access to the value if the Trace Agent is temporarily down. In these cases, we need to check for the value again on the next call to the info endpoint
+        #     - If we have access to the value, we need to check if it changed from the previous value.
+        def update_container_tags(res)
+          header_value = res.headers[Core::Transport::Ext::HTTP::HEADER_CONTAINER_TAGS_HASH]
+          new_container_tags_value = header_value if header_value && !header_value.empty?
+
+          # if the Trace Agent returns a new value for the checksum, invalidate the cached propagation hash
+          if new_container_tags_value && new_container_tags_value != @container_tags_checksum
+            @container_tags_checksum = new_container_tags_value
```
marcotc, line 125: If the agent doesn't return a valid `new_container_tags_value`, then we'll not populate `@container_tags_checksum`.
This means that we'll always call `fetch` in `fetch if @container_tags_checksum.nil?` in `def propagation_hash`, so all callers of `propagation_hash` will make an extra HTTP request to the agent every time, as nothing is cached.

I recommend caching after **any** agent response to avoid that.
The value of `@container_tags_checksum` will be corrected by other callers of `fetch` (the tracer calls it once per second, while flushing spans).

wantsui: Addressed in the latest refactor: 9e4c28852c454214248c6ebfc4722cf98541c928

lib/datadog/core/environment/agent_info.rb
wantsui: Update: I synced with @marcotc on the best way to handle the implicit assignment since my main obstacles were:
1. We need a way to read the value of the propagation_checksum (renamed from propagation_hash) in any part of the code like DBM sql comments and DSM (my next PR).
2. We need a way to save/write the values of propagation_checksum and container_tags_checksum.

I did a refactor where I moved most of the logic into `update_container_tags` so now this is responsible for updating item 2 above whenever we get the headers.

However, there is a race condition that can happen if sql_commenter tries to get the propagation_checksum **_before_** we even make a call to the agent, so if this happens and we don't have any value here yet, there's a new call to `fetch` to trigger the agent_info call again.


See: 9e4c28852c454214248c6ebfc4722cf98541c928

If there's a better way to tackle this, I'm open to suggestions!

lib/datadog/core/environment/agent_info.rb
p-datadog: :+1: The comments added to the code help greatly and can you please add the info in these PR comments (when tags won't be returned) also to the code comments.

lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -14,6 +14,10 @@ module SqlComment
           def self.annotate!(span_op, mode)
             return unless mode.enabled?
 
+            # Add base hash to the span tag
+            base_hash = Datadog.send(:components).agent_info.propagation_checksum
```
p-datadog, line 18: This adds a call to the agent. Is this code going to be running on the main thread (in customer applications)? If yes I think a blocking network call is not something that can be done here.

wantsui: I'll review this more tomorrow, I just want to address this comment:
> / If fetch returns an invalid response, it will be called again. This seems to me that in the ideal world, fetch calls should be throttled (e.g. once every 1 minute or 5 minutes). What is the maximum realistic call frequency? Is multiple times per second possible?

This is a good idea, I'm now thinking of adding a "maximum" number of retries approach for the fetch call so we don't do more than some number, ie: maybe 5(?) or maybe make it request based. In theory if the tracer is failing to fetch this information from the Trace Agent within a reasonable amount of time or attempts, I'm in favor of just stopping it given that it may mean any of the following:
1. the tracer is actually not able to grab the container id
2. trace agent can't grab container tags
What do you think @marcotc ?

marcotc: A blocking call can only happen once, and only if this code here is reached before other parts of the code initializes `agent_info`.

If we remove this call, we'll send SQL queries that cannot be correlated correctly (they will look different from all the remaining queries in the application's lifecycle).

This is the same issue with remote configuration: we don't want to make a synchronous network call on startup, but at some point, we need that info to make a decision. For remote config, we call `Tie.boot` where we know we can't go on without the correct config: https://github.com/DataDog/dd-trace-rb/blob/e700369fef6302d420a299c0cfe1e119c211c62c/lib/datadog/tracing/contrib/rack/middlewares.rb#L41

I think we should leave this here, and make sure that it only ever makes a network call if `agent_info.propagation_checksum` or `agent_info.fetch` were never called. Once called, their result must be cached (even a failed result in this case). I believe this is already the case with the implementation today, but I'll double-check.

marcotc: Negative caching was done in: https://github.com/DataDog/dd-trace-rb/commit/9e4c28852c454214248c6ebfc4722cf98541c928 ✅

marcotc:
@p-datadog 
> / The caching logic is somewhat complex but I do not see test coverage for it in the diff. Specifically that `fetch` is only called once, that if `fetch` returns a response without propagation info in it it won't be called again, and possibly that if `fetch` returns an invalid response it _will_ be called again.

@wantsui I think there are missing tests, since `agent_info.rb` has many different code paths now, and each should be exercised at least one. Take a look at the `if/unless` clauses and object comparisons (e.g `new_container_tags_value != @container_tags_checksum`) and make sure we are testing each code path.

> / If `fetch` returns an invalid response, it will be called again. This seems to me that in the ideal world, `fetch` calls should be throttled (e.g. once every 1 minute or 5 minutes). What is the maximum realistic call frequency? Is multiple times per second possible?

Fetch is unconditionally called once per second in the background thread that flushes traces: https://github.com/DataDog/dd-trace-rb/blob/e700369fef6302d420a299c0cfe1e119c211c62c/lib/datadog/tracing/transport/traces.rb#L186

The whole background processing can be skip if there are no spans created within that second, but for a "healthy" application, it will be constantly called.

That background call is the one we are depending on to catch agent updates.

We can improve that logic (skip frequent checks), but in that case, we really don't want to send traces the agent without knowing its updated capabilities, since that case is related to error reporting, and we don't want to miss reporting application errors.

> / This PR does not add locking around the `fetch` calls that I can see, therefore it permits multiple of them to be in flight at the same time. I think this is not affecting correctness but I am curious if it is considered acceptable for performance. If yes, a note to this effect in the source code would be helpful.

So, we could lock on `fetch`, but I suggest @wantsui not to lock because:
1. The race for "fetching" can only happen once, since any successful calls to `propagation_checksum` will return early with a cached value, even on failure.
2. The result is the same, since they are only directly writing to internal instance variables.
3. I'm not sure if the complexity is worth it (added testing; low but non-zero chance of deadlocks).

But I'm not really "firm" on this one, if you have an alternative argument.

lib/datadog/core/environment/agent_info.rb
marcotc: @Strech, I wasn't able to help want completely make the code "pure functional", so that no side-effects (aka the internal assignment) would be removed.
Since this class has an asymmetric API (for setting the data vs reading it), I don't really have good ideas on how to better structure it.
If you do, I'd love to hear suggestions on how to get rid of this undesirable behaviour.

lib/datadog/tracing/contrib/propagation/sql_comment.rb
wantsui: I just added two more sets of changes to add more tests: [Add more tests to cover fetch concerns, retries, and other logic arising from agent_info](https://github.com/DataDog/dd-trace-rb/pull/5208/commits/bddb96698bbeb4b8dc581d42443fb40b2a7a4316) and [Remove the fetch call if nil because there is no way to address the agent timeout without refactoring a lot. Update tests.](https://github.com/DataDog/dd-trace-rb/pull/5208/commits/f7270b3d0f6b71d392a4f1be262dcd4a9af1794a) to improve the behavior, though it came at the cost of removing the fetch call in propagation_checksum completely.

Let me know if this looks better from a performance and testing perspective :)

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
```diff
@@ -37,6 +43,39 @@
...
+      it 'sets the propagated hash on the span tag as string' do
+        described_class.annotate!(span_op, propagation_mode)
+        expect(span_op.get_tag('_dd.propagated_hash')).to eq('1234567890')
+      end
+
+      it 'sets as tag (meta) not as metric' do
+        described_class.annotate!(span_op, propagation_mode)
+
+        expect(span_op.get_tag('_dd.propagated_hash')).to eq('1234567890')
+      end
```
marcotc: There are the same test.

lib/datadog/core/environment/agent_info.rb
Strech: I think here I was mistaken, after the last explanation, I get on how does it work and I think it's good 👍🏼 

I will resolve it. Thanks for walking me through

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
wantsui: removed: https://github.com/DataDog/dd-trace-rb/pull/5208/commits/e7890a6c86197f412cb044356712f11401da2bfe

lib/datadog/core/environment/agent_info.rb
wantsui: Done in https://github.com/DataDog/dd-trace-rb/pull/5208/commits/1e0814e8c329ab7e1e920307d06444aadc321fcc, let me know if more comments are needed, right now the comments are
```
        # When deciding whether the propagation checksum should be updated, we need to be aware of some concerns:
        #     - The tracer fails to send the container id in the first place
        #     - It is possible that older Trace Agents may not have this specific header
        #     - It is possible that we don't have access to the value if the Trace Agent is temporarily down. In these cases, we need to check for the value again on the next call to the info endpoint
        #     - If we have access to the value, we need to check if it changed from the previous value.
        #     - The Trace Agent runs into a permissions/setup issue.
```

wantsui: I'm marking this as resolved, feel free to comment on the latest iteration if there are still clarifying comments that I can add!

lib/datadog/core/environment/agent_info.rb
wantsui: A lot of refactoring took place, `        def update_container_tags(res)` is private now.

lib/datadog/core/environment/agent_info.rb
```diff
@@ -59,19 +62,77 @@ def initialize(agent_settings, logger: Datadog.logger)
...
+          update_container_tags(res)
+
           res
         end
 
         def ==(other)
           other.is_a?(self.class) && other.agent_settings == agent_settings
         end
+
+        # Returns the propagation checksum, with part of the datafrom the Agent.
```
p-datadog:
```suggestion
        # Returns the propagation checksum, with part of the data from the Agent.
```
I assume this is what was intended?

lib/datadog/core/environment/agent_info.rb
```diff
@@ -59,19 +62,77 @@ def initialize(agent_settings, logger: Datadog.logger)
...
+          if new_container_tags_value && new_container_tags_value != @container_tags_checksum
+            @container_tags_checksum = new_container_tags_value
+
+            process_tags = Process.serialized
+            # new_container_tags_value (non nil) over @container_tags_checksum (string?) to avoid steep errors
+            data = process_tags + new_container_tags_value
+            @propagation_checksum = Core::Utils::FNV.fnv1_64(data)
+          elsif !defined?(@propagation_checksum)
+            # Cache nil to avoid repeated fetch attempts when agent doesn't provide the header
+            @propagation_checksum = nil
```
p-datadog: My understanding is this line is no longer needed since the fetch was removed?

wantsui: removed in 4f35dbc85aba0750ee76b452eca3f1b6dba0778b 
cc @marcotc for awareness

lib/datadog/core/environment/agent_info.rb
wantsui: In the latest commit, this is now renamed to "new_container_tags_value" since I got other feedback about not adding the type into the variable name:
```
  process_tags = Process.serialized
            # new_container_tags_value (non nil) over @container_tags_checksum (string?) to avoid steep errors
            data = process_tags + new_container_tags_value
            @propagation_checksum = Core::Utils::FNV.fnv1_64(data)
```

PR #5209, opened 2026-01-07
Add process and container tags support for DSM

wantsui:
**What does this PR do?**

The goal of this is to add container+process tags support to DSM here: https://docs.google.com/document/d/15GtNOKGBCt6Dc-HsDNnMmCdZwhewFQx8yUlI9in5n3M/edit?tab=t.0#heading=h.fyt7z8ehztcm.

Some backend info about the data streams monitoring in general: 

1. APM spans are tagged with `pathway.hash`
2. There's also a DSM stats payload, where a hash is always sent under 'stats'. 

When this process tags setting is enabled:
1. The tracer sends a container id (if available) to the Trace Agent
2. The Trace Agent looks for container tags and sends back a hash called `Datadog-Container-Tags-Hash`
3. When a DSM related APM span is created (think of the kafka-esque libraries), we update the `pathway.hash` to include the propagation_checksum.
4. The DSM stats pipeline sends stats along with various tags. **What's new:** we're now also sending a new field called `ProcessTags` here, but it will take some additional work for this information to be leveraged. (Right now I cannot find it anywhere)

This PR is based off the DBM changes in https://github.com/DataDog/dd-trace-rb/pull/5208, except with modifcations for DSM.

**Motivation:**

For AIDM-259

The Python PR does this for DBM and DSM in one go:
- https://github.com/DataDog/dd-trace-py/pull/15293

The Java PR does this in two:
- https://github.com/DataDog/dd-trace-java/pull/9156
- https://github.com/DataDog/dd-trace-java/pull/9282

The Node.js PR:
- https://github.com/DataDog/dd-trace-js/pull/7212

**Change log entry**

No.

**Additional Notes:**

No.

**How to test the change?**

I tested this two ways:

1. First I made sure the tests ran locally
```
bundle exec rspec spec/datadog/data_streams/processor_spec.rb
```
2. I had a script that sent data to DSM and APM with and without this mode and verified that the metric tag called "hash" changed accordingly. See AIDM-259 for screenshots of the working behavior in the UI.

### DIFF
--- a/spec/datadog/data_streams/processor_spec.rb
+++ b/spec/datadog/data_streams/processor_spec.rb
@@ -338,20 +338,16 @@ RSpec.describe Datadog::DataStreams::Processor do
   describe '#compute_pathway_hash' do
     subject(:pathway_hash) { processor.send(:compute_pathway_hash, 0, ['type:kafka']) }
 
-    shared_examples 'a valid pathway hash' do
-      it 'computes a valid pathway hash' do
-        expect(pathway_hash).to be_a(Integer)
-        expect(pathway_hash).to be > 0
-      end
-    end
-
     context 'when process tags are enabled' do
       let(:settings) { double('Settings', service: 'test-service', env: 'test', experimental_propagate_process_tags_enabled: true) }
 
       context 'and agent info has been fetched' do
         let(:agent_info) { instance_double(Datadog::Core::Environment::AgentInfo, propagation_checksum: 1234567890) }
 
-        include_examples 'a valid pathway hash'
+        it 'computes a valid pathway hash' do
+          expect(pathway_hash).to be_a(Integer)
+          expect(pathway_hash).to be > 0
+        end
 
         it 'includes the propagation checksum in the pathway hash' do
           expect do
@@ -363,7 +359,10 @@ RSpec.describe Datadog::DataStreams::Processor do
       context 'and agent info has not been fetched yet' do
         let(:agent_info) { instance_double(Datadog::Core::Environment::AgentInfo, propagation_checksum: nil) }
 
-        include_examples 'a valid pathway hash'
+        it 'computes a valid pathway hash' do
+          expect(pathway_hash).to be_a(Integer)
+          expect(pathway_hash).to be > 0
+        end
       end
     end
 
@@ -373,7 +372,10 @@ RSpec.describe Datadog::DataStreams::Processor do
       context 'and agent info has been fetched' do
         let(:agent_info) { instance_double(Datadog::Core::Environment::AgentInfo, propagation_checksum: 1234567890) }
 
-        include_examples 'a valid pathway hash'
+        it 'computes a valid pathway hash' do
+          expect(pathway_hash).to be_a(Integer)
+          expect(pathway_hash).to be > 0
+        end
 
         it 'does not include the propagation checksum in the pathway hash' do
           expect do
@@ -385,7 +387,10 @@ RSpec.describe Datadog::DataStreams::Processor do
       context 'and agent info has not been fetched yet' do
         let(:agent_info) { instance_double(Datadog::Core::Environment::AgentInfo, propagation_checksum: nil) }
 
-        include_examples 'a valid pathway hash'
+        it 'computes a valid pathway hash' do
+          expect(pathway_hash).to be_a(Integer)
+          expect(pathway_hash).to be > 0
+        end
       end
     end
   end


spec/datadog/data_streams/processor_spec.rb
```diff
@@ -333,4 +336,57 @@
       end
     end
   end
+
+  describe '#compute_pathway_hash with base hash' do
```
marcotc: Because `#compute_pathway_hash` does not cache its result, I don't think we should test that two successive calls to it, when we change the inputs, have different results since that's natural for non-caching methods.

I think we only need two tests: one with a `nil` base_hash and another with a valid base_hash.

lib/datadog/data_streams/processor.rb
```diff
@@ -365,6 +366,13 @@ def compute_pathway_hash(current_hash, tags)
         env = @settings.env || 'none'
 
         bytes = service.bytes + env.bytes
+
+        # If there is a base hash, add it to DSM back propagation
+        base_hash = Core::Environment::BaseHash.current
+        if base_hash
+          bytes += [base_hash].pack('Q<').bytes
+        end
```
marcotc: The PR looks good overall!
One question I have is: how do I know that `base_hash` goes right here, in the byte stream; after `env` and before `tags`?

wantsui: I'm just following the python version which does this: https://github.com/DataDog/dd-trace-py/blob/8b739fe0837a22cab76116050e8b7e4b45407c6c/ddtrace/internal/process_tags/__init__.py#L92, do you think I need to add better tests in this PR to show that the bytes are comprised in the expected order? (https://github.com/DataDog/dd-trace-py/pull/15293)

spec/datadog/data_streams/processor_spec.rb
wantsui: I removed the extra tests from https://github.com/DataDog/dd-trace-rb/pull/5209/commits/231fd9cc9c0591b39da8edc6d85e1329c739c119

lib/datadog/data_streams/processor.rb
```diff
@@ -316,6 +316,16 @@ def flush_stats
...
           }
 
+          if @settings.experimental_propagate_process_tags_enabled && payload
+            process_tags = Core::Environment::Process.serialized
+            # Datadog Agent expects an array of tag strings, not a single serialized string
+            # Otherwise it fails at "Error decoding request msgp: attempted to decode type \"str\" with method for \"array\" at ProcessTags"}
+            # Split "key1:val1,key2:val2" into ["key1:val1", "key2:val2"]
+            unless process_tags.empty?
+              payload['ProcessTags'] = process_tags.split(',')
+            end
```
marcotc: If I recall correctly, that's what `Core::Environment::Process.tags` is, isn't it?
```suggestion
            payload['ProcessTags'] = Core::Environment::Process.tags
```

And `Core::Environment::Process.tags` is never empty.

lib/datadog/data_streams/processor.rb
```diff
@@ -365,6 +375,17 @@ def compute_pathway_hash(current_hash, tags)
...
 
         bytes = service.bytes + env.bytes
+
+        # If there is a propagation checksum, add it to DSM back propagation
+        # Follows Python: https://github.com/DataDog/dd-trace-py/blob/8b739fe0837a22cab76116050e8b7e4b45407c6c/ddtrace/internal/datastreams/processor.py#L423
+        # bytes order is service + env + process tags + container tags
+        if @settings.experimental_propagate_process_tags_enabled
+          propagation_checksum = Datadog.send(:components).agent_info.propagation_checksum
+          if propagation_checksum
+            bytes += [propagation_checksum].pack('Q<').bytes
```
marcotc: `propagation_checksum` is added to `bytes` after `env.bytes` and before `tag.bytes`.
Does this matter? Or is it similar to the DBM case, where it just needs to be internally consistent?

I think that either: it doesn't matter, or it does matter for a specific reason. In either way, we should have some mention of it in `def compute_pathway_hash`.

lib/datadog/data_streams/processor.rb
```diff
@@ -316,6 +316,16 @@ def flush_stats
             'Hostname' => hostname
           }
 
+          if @settings.experimental_propagate_process_tags_enabled && payload
```
p-datadog: payload is always not nil here

lib/datadog/data_streams/processor.rb
wantsui: After doing some research, it looks like the behavior is inconsistent between Java and Python.

This Ruby PR is based on the [Python version](https://github.com/DataDog/dd-trace-py/blob/02f5d7783cf7666ddad9d583fadffd7be164666e/ddtrace/internal/datastreams/processor.py#L423), which does `service + env + process tags hash`

Meanwhile the [Java version](https://github.com/DataDog/dd-trace-java/blob/d10055d2a1d3e41671ab944e97192919d3797f11/internal-api/src/main/java/datadog/trace/api/BaseHash.java#L38-L53) is different, which does: `service + env + primary tag + process tags + container tags`

If consistency is important, we can change it later but right now I don't see it cc @vandonr .

I'll put a note in the comments that Java does it differently than Python (which is what Ruby is following).

lib/datadog/data_streams/processor.rb
wantsui: I tried removing this to be only the setting in the condition:

```
if @settings.experimental_propagate_process_tags_enabled
```

But then steep complained :(

Instead I added a comment in my last commit to explain that payload cannot be nil but the check is for steep: 152693bb24c3af0b1fcb88adc760b3d6340e339c.

```
          # payload cannot be nil, but the check here is for steep
```

Feel free to suggest better wording!

marcotc: The type checker is incorrect, so it's giving us bad advice 🙂
Let's do what we know is correct, and put a typecheck exception comment in this line, like [`# steep:ignore SteepErrorNameHere`](https://github.com/soutaro/steep/blob/8521be6192b88d5c58514783d7f58585c5545497/manual/ignore.md?plain=1#L19)

lib/datadog/data_streams/processor.rb
marcotc: There's not much we can do right here, in this method, with this information (`# Current bytes order follows Python: https://github.c...`), so I think we need @vandonr to let us know if only internal consistency is important (same application instance), or cross-application too.

spec/datadog/data_streams/processor_spec.rb
```diff
@@ -333,4 +336,61 @@
...
+
+        it 'includes the propagation checksum in the pathway hash' do
+          expect do
+            allow(agent_info).to receive(:propagation_checksum).and_return(nil)
+          end.to change { processor.send(:compute_pathway_hash, 0, ['type:kafka']) }.from(0x1ca26aa13235b3b7).to(0x42fc8338f4ba11f9)
+        end
+      end
+
+      context 'and agent info has not been fetched yet' do
+        # We must always return a valid pathway hash when agent info has not been fetched yet
```
marcotc: I don't think we need this comment, since the specs read `agent info has not been fetched yet, a valid pathway hash`. I think it already reads as "a valid pathway should be computed event when the agent info is not yet fetched"

lib/datadog/data_streams/processor.rb
vandonr: I think only internal consistency is important, for instance, in dotnet we don't have primary tag either, there is a comment that it's not supported in dotnet and _voilà_

vandonr: I'm actually sure. The only product I'm aware of where we wanted cross-tracer consistency in hashes was schema tracking, I spent a lot of time solving little inconsistencies between just java and dotnet, only to hear that the product was actually paused and in limbo :|

lib/datadog/data_streams/processor.rb
```diff
@@ -316,6 +316,10 @@ def flush_stats
             'Hostname' => hostname
           }
 
+          if @settings.experimental_propagate_process_tags_enabled
+            payload['ProcessTags'] = Core::Environment::Process.tags # steep:ignore NoMethod
```
vpellan: I think this shouldn't need a steep:ignore ? The `tags` method seems to be defined in Core::Environment::Process

spec/datadog/data_streams/processor_spec.rb
wantsui: removed in c1bed06664b8513a12a0a8b93ad7e59d174e175b :)

lib/datadog/data_streams/processor.rb
wantsui: addressed in c1bed06664b8513a12a0a8b93ad7e59d174e175b

lib/datadog/data_streams/processor.rb
wantsui: When I remove the `steep:ignore`, it throws:

```
lib/datadog/data_streams/processor.rb:320:35: [error] Type `(::Hash[::String, untyped] | nil)` does not have method `[]=`
│
└             payload['ProcessTags'] = Core::Environment::Process.tags
                                     ~

          +------------------------------------------------------------------------------+
```

For now I'll just apply @marcotc 's suggestion of ignoring it :)

lib/datadog/data_streams/processor.rb
wantsui: How would you like this reflected in the code - as a comment that the standardization needs to be consistent within the same library?

vandonr: if you want, but I think this should be the default, what would need a comment is if consistency was needed across repositories

lib/datadog/data_streams/processor.rb
p-datadog: There is a nil assignment to payload on line 304 to work around another steep deficiency. So now we have a workaround for the workaround, plus two places which are conditioned on payload but should in actuality always execute, great...

lib/datadog/data_streams/processor.rb
p-datadog: I made https://github.com/DataDog/dd-trace-rb/pull/5346 to reduce contortions in this method made for steep. Honestly some sort of bulk ignore of hash type checking would probably not hide any actual problems reported by steep, since it doesn't seem to handle hashes in a useful way at all?

lib/datadog/data_streams/processor.rb
wantsui: I tried to shorten the comment here a5f206851d9e3ad17adf705da0f522f4d37baf45 , feel free to suggest alternatives.

lib/datadog/data_streams/processor.rb
```diff
@@ -365,6 +369,16 @@ def compute_pathway_hash(current_hash, tags)
         env = @settings.env || 'none'
 
         bytes = service.bytes + env.bytes
+
+        # Note: The order of the bytes needs to be consistent within a tracer.
+        # The other tracers don't have consistency (Java adds a primary tag, Python does not).
+        if @settings.experimental_propagate_process_tags_enabled
+          propagation_checksum = Datadog.send(:components).agent_info.propagation_checksum
```
p-datadog: I am sad at a global reference but probably dependency injecting agent info is best done in a separate PR.

lib/datadog/data_streams/processor.rb
marcotc: I think I addressed both the typing issue, and [@ivoanjo's comment](https://github.com/DataDog/dd-trace-rb/pull/5346/changes#r2793513497) in this commit I pushed: https://github.com/DataDog/dd-trace-rb/pull/5209/changes/975d30edbff5ddbda5feac0dd4c087d793885666

lib/datadog/data_streams/processor.rb
marcotc: We can address this!
@wantsui add the agent_info to `Datadog::DataStreams::Processor.new` in `lib/datadog/core/configuration/components.rb`. The agent info is created right in components.sb, and `Processor` respects component lifecycle!

lib/datadog/data_streams/processor.rb
p-datadog: :+1:

lib/datadog/data_streams/processor.rb
wantsui: Addressed in f1f662415bc748d9df1cd3b7a952c3378465042a and c1218a0a8d6fbcab155b4f0a26af81cdcb5696cd

spec/datadog/data_streams/processor_spec.rb
```diff
@@ -333,4 +334,59 @@
...
+
+  describe '#compute_pathway_hash' do
+    subject(:pathway_hash) { processor.send(:compute_pathway_hash, 0, ['type:kafka']) }
+
+    shared_examples 'a valid pathway hash' do
+      it 'computes a valid pathway hash' do
+        expect(pathway_hash).to be_a(Integer)
+        expect(pathway_hash).to be > 0
+      end
+    end
```
Strech: I don't think there is a big value brought by this extraction. Shared tests are in general cumbersome to track and it's better to just copy-parse this simple check.

lib/datadog/data_streams/processor.rb
```diff
@@ -301,28 +303,30 @@ def decode_pathway_b64(encoded_ctx)
...
 
-        # Send to agent outside mutex to avoid blocking customer code if agent is slow/hung
-        send_stats_to_agent(payload) if payload
+        payload = {
+          'Service' => @settings.service,
+          'TracerVersion' => Datadog::VERSION::STRING,
+          'Lang' => 'ruby',
+          'Stats' => stats_buckets,
+          'Hostname' => hostname
+        } # : ::Hash[::String, (String | ::Array[::String])]
```
Strech: For some consistency

```suggestion
        } # : ::Hash[::String, (::String | ::Array[::String])]
```

also, ideally, this should become as a type alias and be referred instead of defined, i.e

```
type payload = ::Hash[::String, (::String | ::Array[::String])]

# : DataStreams::Processor::payload
```

otherwise, you can easily de-synchronize the type definitions between method and callers and will have to go in each place and modify it

sig/datadog/data_streams/processor.rbs
```diff
@@ -68,7 +70,7 @@ module Datadog
         def record_checkpoint_stats: (hash: ::Integer, parent_hash: ::Integer, edge_latency_sec: ::Float, full_pathway_latency_sec: ::Float, payload_size: ::Integer, tags: ::Array[::String], timestamp_sec: ::Float) -> void
         def record_consumer_stats: (topic: ::String, partition: ::Integer, offset: ::Integer, timestamp: ::Time) -> void
         def aggregate_consumer_stats_by_partition: (::String topic, ::Integer partition, ::Integer offset, ::Time timestamp) -> void
-        def send_stats_to_agent: (::Hash[::String, untyped] payload) -> void
+        def send_stats_to_agent: (::Hash[::String, (String | ::Array[::String])] payload) -> void
```
Strech:
```suggestion
        def send_stats_to_agent: (::Hash[::String, (::String | ::Array[::String])] payload) -> void
```

spec/datadog/data_streams/processor_spec.rb
wantsui: I brought it back in f3dc62e1ef373930f9dec58160c9e558d9f72294

PR #5222, opened 2026-01-10
Fix NoMethodError when empty? is called on Integer in Runner#run

terra-namibia:
**What does this PR do?**

Fix `NoMethodError: undefined method 'empty?' for an instance of Integer` in `Datadog::AppSec::SecurityEngine::Runner#run`.

The `reject!` block calls `v.empty?` on all non-boolean values, but `Integer` and `Float` do not respond to `empty?`. This PR adds a `respond_to?(:empty?)` check before calling `empty?`.

**Motivation:**

When a Rails API returns a numeric JSON response (e.g., `render json: count`), the response body is captured as an Integer in `ephemeral_data`. The current implementation assumes all values respond to `empty?`, causing:

**Change log entry**

- AppSec: Fix `NoMethodError` when filtering data containing Integer or Float values in `Runner#run`

**Additional Notes:**

- The fix uses `respond_to?(:empty?)` pattern which is safe for any object type
- Applied to both `persistent_data` (line 33) and `ephemeral_data` (line 40)
- Also handles other types without `empty?` method (e.g., `Symbol`)

**How to test the change?**

1. **Automated tests added:**
- bundle exec rspec spec/datadog/appsec/security_engine/runner_spec.rb
   - `it 'does not remove keys with integer values'`
   - `it 'does not remove keys with float values'`

2. **Manual reproduction:**
   ```ruby
   # In a Rails controller
   def count
     render json: User.count  # Returns Integer
   end

### DIFF
--- a/lib/datadog/appsec/security_engine/runner.rb
+++ b/lib/datadog/appsec/security_engine/runner.rb
@@ -27,7 +27,7 @@ module Datadog
           persistent_data.reject! do |_, v|
             next false if v.is_a?(TrueClass) || v.is_a?(FalseClass)
 
-            v.nil? || v.empty?
+            v.nil? || (v.respond_to?(:empty?) && v.empty?)
           end
 
           ephemeral_data.reject! do |_, v|


Strech:
Hey @terra-namibia thanks for your contribution and heads up. I'm going to run CI for you and reconfirm the issue.

lib/datadog/appsec/security_engine/runner.rb
```diff
@@ -33,7 +33,7 @@ def run(persistent_data, ephemeral_data, timeout = WAF::LibDDWAF::DDWAF_RUN_TIME
           ephemeral_data.reject! do |_, v|
             next false if v.is_a?(TrueClass) || v.is_a?(FalseClass)
 
-            v.nil? || v.empty?
+            v.nil? || (v.respond_to?(:empty?) && v.empty?)
```
Strech, line 36: There is a `persistent_data` right above it and I think it needs the same fix. Do you mind updating it too? Or I can push it for you.

terra-namibia: Done! I've applied the same fix to `persistent_data` as well. Thanks for catching that!

Strech (approved):

Strech:
@terra-namibia thanks again for the fix. Can you tell us how do you stumble upon that issue?

terra-namibia:
@Strech I detected this error via my error monitoring tool.

Strech:
Can you share the time when you spot it, I'm trying to adjust our internal alerts, and see how did we miss it. Thanks in advance!

terra-namibia:
Sure! Here are the details:

- **First detected**: Jan 07, 2026 03:36:48 UTC
- **dd-trace-rb version**: 2.23.0

10+ occurrences were observed within 5 minutes after the first detection.

y9v:
@terra-namibia thank you for your contribution. Your fix was released with v2.25.0, please feel free to try it out!

Issue #1389, opened 2021-03-01
Tracing inside Fibers


rmosolgo:
👋  Hi! I'm coming here by way of a bug report in graphql-ruby: https://github.com/rmosolgo/graphql-ruby/issues/3366

A recent GraphQL-Ruby release included a Fiber-based batch loading plugin: it takes little "jobs" during GraphQL execution, then uses a bunch of Fibers to run the jobs concurrently. (Hopefully, it'll leverage the Ruby 3 scheduler soon so that I/O in those fibers happens in parallel, but that's not done yet.) 

A user noticed that code run inside Fibers doesn't show up in DataDog APM, and that it's likely due to the use of `Thread.current` for storing DataDog state: 

https://github.com/DataDog/dd-trace-rb/blob/079f18f410f590252030578aa240576f3fc4c6e5/lib/ddtrace/opentracer/thread_local_scope_manager.rb#L35-L37

(Inside a new Fiber, `Thread.current` is empty.) 

I'm sorry to say I don't have an idea for how to address this, but I thought I'd open an issue in case y'all have any suggestions. This certainly seems related to #1136 , since both are about using Fibers. 

Does anyone have a suggestion for how this issue might be addressed?

delner:
Thanks for the heads up @rmosolgo . Indeed `ddtrace` does depend on `Thread.current` to store trace context; when this isn't available, or changes, then active traces will not cross execution contexts unless the previous context is loaded & accessed in the new context.

We expect this to be case with `Ractor` and `Fiber` alike, so we're actively evaluating 1) what the trace behavior should be, and 2) how such behavior should be implemented. Sorry we don't have something for this right now, but feedback like this really does help us contextualize the right solution, so this is very helpful. If you or anyone else has suggestions, we'll happily consider that as well.

Once I round up with the team I expect I'll have more to share.

AlexRiedler:
Is there any update you can provide around support for Ractor's + Fibers in Ruby ? Seems like a pretty vital feature for the future of Ruby development, particularly performance.

delner:
@AlexRiedler We spent a bit of time looking into this. Last I remember (some time ago), Ractors themselves have been a bit unstable, not receiving widespread use, and there are some significant challenges in loading constants/sharing data across Ractor boundaries. It's still something I'm interested in adding in the future, but isn't currently prioritized. We're open to changing that if there's significant community interest in this feature, though!

[...]

PR #5226, opened 2026-01-13
Add Thread-bound tracing context

### LINKED ISSUE #5117: [BUG]: current trace is missing when in a new Fiber (e.g. `Enumerator.new`)
### Tracer Version(s)

2.22.0

### Ruby Version(s)

ruby 3.4.7p58

### Relevent Library and Version(s)

openai

### Bug Report

This library uses `Thread.current` for storing / accessing the current trace / span, which works in 99% of cases. However, the one place this doesn't work as expected is when fibers are used in very specific ways, such as with [`Enumerator.new` and external iteration](https://ruby-doc.org/3.4.1/Enumerator.html#class-Enumerator-label-External+Iteration). Unluckily for us, OpenAI's internal request execution logic uses exactly that ([source](https://github.com/openai/openai-ruby/blob/v0.37.0/lib/openai/internal/transport/pooled_net_requester.rb#L115-L190)).

Because of that, we cannot rely on `Thread.current` to provide us with the current trace, since that store is actually "Fiber-local", not "Thread-local", and it does not inherit to "child" fibers. "Fiber storage variables" on the other hand, are inherited and are designed to handle an Enumerator-created fiber.

In our case, this was noticed specifically with the OpenAI Ruby SDK, but seems to be an issue in general whenever logic that is traced is executed within an `Enumerator.new` block.

A minimal reproduction that illustrates the issue:

```ruby
Thread.current[:foo] = 123

e = Enumerator.new do |y|
  y << Thread.current[:foo]
end

e.next
# => nil
```

We've worked around this by patching the OpenAI request transport class to use a "Fiber storage variable":

```ruby
module PooledNetPatcherPatch
  DATADOG_TRACE_DIGEST_FIBER_KEY = :_openai_datadog_trace_digest

  def execute(...)
    Fiber[DATADOG_TRACE_DIGEST_FIBER_KEY] = Tracing.active_trace&.to_digest

    super
  ensure
    Fiber[DATADOG_TRACE_DIGEST_FIBER_KEY] = nil
  end

  private

  def with_pool(...)
    Tracing.continue_trace!(Fiber[DATADOG_TRACE_DIGEST_FIBER_KEY])

    super
  end
end

OpenAI::Internal::Transport::PooledNetRequester.prepend(PooledNetPatcherPatch)
```

I wasn't really sure if this was something that could be easily fixed, but it seemed worth raising to at least start a discussion and share our current solution!

### Reproduction Code

_No response_

### Configuration Block

_No response_

### Error Logs

_No response_

### Operating System

_No response_

### How does Datadog help you?

_No response_

marcotc:
Closes: #5117
Addresses: #1389

**What does this PR do?**
- Introduces the `tracing.context_scope` (`DD_TRACE_CONTEXT_SCOPE`) configuration to choose how in-process trace context is stored:
  - `fiber_isolated` (default): each Fiber has its own trace.
  - `thread` (recommended): a single trace is shared by every Fiber running on the same Thread.

**Motivation:**

Fiber-based context management was introduced [very early in this library's history](https://github.com/DataDog/dd-trace-rb/commit/69cb4c1d47), and it was not intentional: it was a byproduct of Ruby's `Thread#[]` API looking like the correct way to handle Thread-local storage.

Changing this management strategy is risky, so we didn't change it unless it was strictly required.

As Fibers are being used more every day in Ruby, it's now a good time to address this old issue.

**Change log entry**
Yes. Add Thread-bound tracing context, configured by `tracing.context_scope` (`DD_TRACE_CONTEXT_SCOPE`).

**Additional Notes:**
- Invalid values fall back to `'fiber_isolated'`, with a warning.
- `context(key)` remains for now, but it's unused and complex. It is marked as deprecated.

**How to test the change?**
- `bundle exec rspec spec/datadog/tracing/context_provider_spec.rb`
- `bundle exec rspec spec/datadog/tracing/component_spec.rb`
- `bundle exec rspec spec/datadog/tracing/configuration/settings_spec.rb`


lib/datadog/tracing/configuration/settings.rb
```diff
@@ -128,6 +128,33 @@ def self.extended(base)
                 o.type :bool
               end
 
+              # Configures how a trace is propagated within the Ruby process.
+              # This changes the trace continuity when spans are created in different Fibers or Threads.
+              #
+              # The available options, and how active trace is bound:
+              # * `'thread'` (recommended): Bound to `Thread.current`. The active trace is the same for all Fiber execution in a Thread. This will be the default behavior in a future major release.
+              # * `'fiber_isolated'` (default): Bound to `Fiber.current`.
```
ivoanjo, line 153: Minor: Maybe `thread`/`fiber` would be enough here? I understand why, but having a suffix only for fiber looks a bit off? 👀

lib/datadog/tracing/context_provider.rb
```diff
@@ -5,77 +5,149 @@
...
-    #
-    # @see https://ruby-doc.org/core-3.1.2/Thread.html#method-i-5B-5D Thread attributes are fiber-local
+    # A concurrency-safe repository for storing and retrieving the current trace context.
     class DefaultContextProvider
-      # Initializes the default context provider with a fiber-bound context.
-      def initialize
-        @context = FiberLocalContext.new
+      # @param scope [ContextScope] the scope to use for context storage.
+      #   The scope defines the lifecycle of the context.
+      def initialize(scope: FiberIsolatedScope.new)
```
ivoanjo, line 12: Minor: Since this is expected to always follow the setting now, consider maybe removing the default? In particular, I find that having defaults that are not what we should be using in production makes it easier to go "oops we didn't pass in the argument in this one case and the code still ran fine and just ignored what it was supposed to do".

lib/datadog/tracing/context_provider.rb
```diff
@@ -5,77 +5,149 @@
...
+    #
+    # @see https://ruby-doc.org/core-3.1.2/Thread.html#method-i-thread_variable_get
+    class ThreadScope < ContextScope
+      def current=(context)
+        Thread.current.thread_variable_set(@key, context)
+      end
+
+      protected
+
+      def get_current
```
Strech, line 119: I think in Ruby conventions we don't use `get_` prefixed names. In this case API would be

```ruby
ThreadScope.current = "..."
ThreadScope.current # => "..."
ThreadScope.current_for_thread(thread) # => "..."
```

But I would like to point out that this methods (also below) violate SRP and I think maybe it's better to name them Rails-like `find_or_initialize` style

lib/datadog/tracing/context_provider.rb
```diff
@@ -5,77 +5,149 @@
...
     #
-    # @see https://ruby-doc.org/core-3.1.2/Thread.html#method-i-5B-5D Thread attributes are fiber-local
-    class FiberLocalContext
-      # To support multiple tracers simultaneously, each {Datadog::Tracing::FiberLocalContext}
-      # instance has its own fiber-local variable.
+    # @abstract
+    class ContextScope
       def initialize
-        @key = :"datadog_context_#{FiberLocalContext.next_instance_id}"
+        @key = :"datadog_context_#{self.class.next_instance_id}"
```
ivoanjo, line 51: Hmm, since you're poking at this, I'll confess this has been bothering me for a long time.

There's two things here: 1) symbol keys that are used as thread-local keys get turned from dynamic GC'able symbols into symbols that live forever...

And 2) since the context object gets set on a thread and never gets unset, the context object itself also lives forever...

That together means that 

```ruby
require 'datadog'
Datadog::Tracing::ThreadScope.new while true
```

...this is enough to blow up an app very quickly.

Of course in practice we don't recreate configuration (and thus the tracer)... too many times... e.g. somewhere around < 100 times on a Rails app?

Given that in pratice dd-trace-rb gets used as a global singleton, perhaps it's finally worth getting rid of the "It provides unique instance ID generation to ensure multiple scope instances do not conflict with each other." capability and do away with this pitfall completely?

(There's a few other minor benefits, such as getting rid of the sequence generator below and the mutex; and even simplifying how the profiler peeks at the context -- right now it needs to retrieve this key to know what's the current context it should be looking at)

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -128,6 +128,33 @@ def self.extended(base)
...
+              #
+              # DEV: `Fiber#storage` might be required, depending on how Ruby usage of Fibers evolves.
+              # DEV: * `:fiber_storage`: Bound to `Fiber.storage`. The active trace is inherited by child Fibers, unless storage is [explicitly not inherited on Fiber creation](https://docs.ruby-lang.org/en/master/Fiber.html#method-c-new).
+              #
+              # @default `DD_TRACE_CONTEXT_SCOPE` environment variable, otherwise `'fiber_isolated'`
+              # DEV-3.0: Change default to `'thread'`, as Fiber execution is normally part of a single high-level operation.
```
ivoanjo, line 159: So I've gone back and forth on this one: first I thought "changing the default is a bit too much, what if people run a bunch of unrelated stuff in fibers?"; then I went like "well but if you have your puma/unicorn/sidekiq and then multiplex some request work into multiple fibers this probably matches what you'd expect"; and then I went back to "wait but won't this screw falcon and other no-threads-all-fibers-workloads?".

So... yeah I'm not sure at this point -- perhaps the best default is not a fixed one, but picking based on the environment? But that's extra complexity as well.

lib/datadog/tracing/context_provider.rb
```diff
@@ -5,77 +5,149 @@
...
+
+      def get_current_for(thread)
+        context = thread.thread_variable_get(@key)
+        return context if context
+
+        context = Context.new
+        thread.thread_variable_set(@key, context)
+        context
+      end
+    end
```
ivoanjo, line 131: So... this setting needs a slight adjustment in the profiler ;)

Specifically the `Collectors::ThreadContext`/`collectors_thread_context.c` in `trace_identifiers_for` was previously reading the `Thread.current[@key]` to attach to profiler samples, and thus will miss this new location. (Specs are in `thread_context_spec.rb`, including for this capability)

I took a peek and the "meh" part is that there's no public C API for `thread_variable_get` (there is for `Thread.current[]`) but it shouldn't be very hard to work around that: a) I see these are "just" a hash that's stored in an invisible instance variable `locals` (not `@locals` -- that's why it's invisible to Ruby code) and those can be read from C; b) we could always set the instance on the thread object directly; c) we could call a method on the profiler here/have the profiler monkey patch it; d) we could arrange to get this info when it's safe to call into Ruby code and cache the result. I'm leaning towards either a or d...?

THAAAAT said: Are you interested in poking at the profiler update as well, or shall I go ahead and open a follow-up PR to this one to do that update?

spec/datadog/tracing/component_spec.rb
```diff
@@ -525,6 +527,36 @@
...
+          it 'builds tracer with FiberIsolatedScope-based context provider' do
+            tracer = build_tracer
+            provider = tracer.instance_variable_get(:@provider)
+
+            expect(provider).to be_a(Datadog::Tracing::DefaultContextProvider)
+            expect(provider.instance_variable_get(:@context)).to be_a(Datadog::Tracing::FiberIsolatedScope)
+          ensure
+            tracer.shutdown!
+          end
+        end
```
ivoanjo: Minor: Maybe worth adding a spec here that shows that the correct default is used when the setting is invalid? 

We explicitly do allow the setting to be invalid and then correct for it in the code -- which looks to me was very intentional BUT I'm not sure this intention will be very clear when looking a the tracing component in isolation and not within the context of the full PR.

spec/datadog/tracing/context_provider_spec.rb
```diff
@@ -149,29 +245,143 @@ def fiber_contexts
...
+        is_expected.to be_a_kind_of(Datadog::Tracing::Context)
+        expect(current).to_not be(fiber_isolated_scope.current)
+      end
+    end
+  end
+
+  describe '#current=' do
+    subject(:set_current) { fiber_isolated_scope.current = context }
+
+    let(:context) { double }
```
ivoanjo, line 269: Minor: I like always naming these objects, in case there's a bug and you are poking at the test and see them going around :)

```suggestion
    let(:context) { double('Dummy context') }
```

ivoanjo (approved):
Looks good! I'm curious if I can convince you that we should move away from the unique id, as I believe it brings more trouble than benefits, and I left a note on how the profiler needs a few tweaks to match, but overall it LGTM.

buraizu (approved):

lib/datadog/tracing/configuration/settings.rb
```diff
@@ -128,6 +128,33 @@ def self.extended(base)
...
+              # @return [String]
+              option :context_scope do |o|
+                o.env Tracing::Configuration::Ext::ENV_CONTEXT_SCOPE
+                o.default 'fiber_isolated'
+                o.type :string
+                o.after_set do |value|
+                  unless ['fiber_isolated', 'thread'].include?(value)
```
mabdinur: Should the configuration value be case sensitive?

mabdinur (approved):
LGTM from an api standpoint

p-datadog:
Question  about option names: why is fiber isolated and thread not? What does "isolated" mean in this context?

lib/datadog/tracing/configuration/settings.rb
p-datadog: I do not understand why one is isolated and the other is not, could you explain?

lib/datadog/tracing/context_provider.rb
p-datadog: I suppose the question here is whether DefaultContextProvider is part of public API of our library - if yes, the default needs to stay (and I would think it should be explicitly documented here as being for backwards compatibility), otherwise the default should be be removed if the scope is always passed in by our own code.

lib/datadog/tracing/context_provider.rb
```diff
@@ -5,77 +5,149 @@
...
-    # @see https://ruby-doc.org/core-3.1.2/Thread.html#method-i-5B-5D Thread attributes are fiber-local
+    # A concurrency-safe repository for storing and retrieving the current trace context.
     class DefaultContextProvider
-      # Initializes the default context provider with a fiber-bound context.
-      def initialize
-        @context = FiberLocalContext.new
+      # @param scope [ContextScope] the scope to use for context storage.
+      #   The scope defines the lifecycle of the context.
+      def initialize(scope: FiberIsolatedScope.new)
+        @context = scope
```
p-datadog, line 13: What is the difference between 'context' and 'scope'? Are they the same thing?

lib/datadog/tracing/context_provider.rb
```diff
@@ -5,77 +5,149 @@
...
       end
 
-      # Sets the current context.
-      def context=(ctx)
-        @context.local = ctx
+      # Initialize or override the active context.
+      # @param context [Context] the trace context to store in the active scope
```
p-datadog, line 17: This line says that context is stored in a scope, therefore implying that context and scope are not the same thing?

sig/datadog/tracing/context_provider.rbs
```diff
@@ -1,23 +1,50 @@
+interface _StorageLike
+  def []: (Symbol) -> Datadog::Tracing::Context?
+  def []=: (Symbol, Datadog::Tracing::Context?) -> Datadog::Tracing::Context?
+end
+
 module Datadog
   module Tracing
     class DefaultContextProvider
-      @context: FiberLocalContext
+      @context: ContextScope
```
p-datadog, line 9: The `@context` is now a scope, not  a context? Why not rename to `@scope` then?

p-datadog (commented):
Maybe I missed it but I am not seeing integration tests verifying scopedness to threads or fibers, do they exist in this PR?

p-datadog:
This PR needs the env var added to the config registry for CI to go green.

eregon:
Not sure if applicable here, but there is also "Fiber storage" i.e. `Fiber[:foo] = 42` and `Fiber#storage` which gets inherited between Fibers.
And might be better to establish relations between Fibers, unsure.
(Actually I see https://github.com/DataDog/dd-trace-rb/issues/5117 also mentioned that)

Any reason to prefer thread-local variables?

My concern is there might be multiple Fibers on a single thread doing very different things, potentially handling different requests with Falcon. Fiber storage variables might help with that, though I'm unsure about the details, e.g. how to avoid inheriting when unwanted (update: I think with `Fiber.new(storage: nil)`).

marcotc:
After reviewing all the feedback and use cases, I think that better solution is to have a Fiber-inherited and independent execution context, instead of Thread-local (or the existing Fiber-local).
Making context independent per-Fiber allows for concurrent or cooperative execution without concern on how they will interweave in memory: each Fiber's spans would be separately flushed, then organically stitched together as they'd share the same `trace_id`.

PR #5240, opened 2026-01-15
Fix reporting of multi-method routes for Endpoint Collection

y9v:
**What does this PR do?**
This PR fixes reporting of routes that accept several methods, for example:

```ruby
match '/multi-method-route', to: 'multi_method#index', via: %i[get post]
```

For such routes, Rails sets `verb` to `GET|POST` - by concatenating all accepted methods with `|`.

**Motivation:**
We have a system-tests failure for a freshly added route:
https://github.com/DataDog/system-tests/actions/runs/21032732488/job/60473172774?pr=5955

**Change log entry**
Yes. AppSec: Fix reporting of Rails routes that accept multiple request methods.

**Additional Notes:**
APPSEC-60637

**How to test the change?**
CI.

### DIFF
--- a/spec/datadog/appsec/api_security/endpoint_collection/rails_collector_spec.rb
+++ b/spec/datadog/appsec/api_security/endpoint_collection/rails_collector_spec.rb
@@ -16,7 +16,7 @@ RSpec.describe Datadog::AppSec::APISecurity::EndpointCollection::RailsCollector 
       )
 
       expect(Datadog::AppSec::APISecurity::EndpointCollection::RailsRouteSerializer)
-        .to receive(:serialize).and_call_original
+        .to receive(:serialize).with(route).once.and_call_original
 
       described_class.new([route]).to_enum.first
     end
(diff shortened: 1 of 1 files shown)


spec/datadog/appsec/api_security/endpoint_collection/rails_collector_spec.rb
```diff
@@ -21,6 +21,18 @@
...
 
+    it 'serializes Rails dispatcher routes that support multiple methods' do
+      route = instance_double(
+        'ActionDispatch::Journey::Route', dispatcher?: true, verb: 'GET|POST',
+        path: instance_double('ActionDispatch::Journey::Path::Pattern', spec: '/events')
+      )
+
+      expect(Datadog::AppSec::APISecurity::EndpointCollection::RailsRouteSerializer)
+        .to receive(:serialize).twice.and_call_original
+
```
Strech, line 35: I think we should check arguments, otherwise what are we testing here? Or the next check should be adjusted, because we don't confirm that we will have what we expect. What is tested is that something will happen 2 times

Strech (approved):
I leave a non-blocking suggestion on test improvements. Otherwise, well done! Clean and clear

Strech (approved):
I leave a non-blocking suggestion on test improvements. Otherwise, well done! Clean and clear

spec/datadog/appsec/api_security/endpoint_collection/rails_collector_spec.rb
y9v: good point, extended the test in [49af521](https://github.com/DataDog/dd-trace-rb/pull/5240/commits/49af5210455ee9248896994f794097ab7015d80b)

PR #5241, opened 2026-01-15
[APMAPI-1616] Use AST traversal for typing stats instead of Regexp

vpellan:
**What does this PR do?**

Replace Regexp in typing stats by AST traversal.

**Motivation:**

Prevent any false positives/negatives. This actually found 30 false positives and 46 false negatives.
We will then be able to add the Module path of a function and use it for comparing instead of the line, to prevent creating a comment when a line is just moved around.

**Change log entry**

None.

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/.github/scripts/typing_stats_compute.rb
+++ b/.github/scripts/typing_stats_compute.rb
@@ -178,6 +178,11 @@ signature_paths.each do |sig_path|
   filtered_declarations = ast_traversal(declarations)
 
   filtered_declarations[:methods].each do |method|
+    # Skip definitions with last comment line being `untyped:accept`
+    if method.comment&.string&.end_with?("untyped:accept\n")
+      typed_methods_size += 1
+      next
+    end
     # Overloading can be done inline so we cannot point to the specific overloaded method definition.
     # That's why we combine the status of all overloads to get the overall status of the method.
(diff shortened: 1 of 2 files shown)


marcotc (approved):

.github/scripts/typing_stats_compute.rb
```diff
@@ -61,6 +55,113 @@
...
+      ast_traversal(declaration.members, result)
+    when ::RBS::AST::Declarations::TypeAlias,
+      ::RBS::AST::Declarations::Constant,
+      ::RBS::AST::Declarations::Global,
+      ::RBS::AST::Members::Var,
+      ::RBS::AST::Members::Attribute
+      result[:others] << declaration
+    # Only this one does not have a type field
+    when ::RBS::AST::Members::MethodDefinition
+      result[:methods] << declaration
```
y9v, line 75: do we need `else` here?

vpellan: It shouldn't be needed, what would fall there are`public`/`private` keywords, `include`/`extend`/`prepend`, and `alias` keyword (the one for method aliases, not type aliases) and we don't care about them in our case as we want to check if something is explicitly `untyped` or not

Issue #5101, opened 2025-11-29
[BUG]: Subprocess failing with an unhandled SIGPROF


louissobel:
### Tracer Version(s)

2.14.0

### Ruby Version(s)

ruby 3.3.6

### Bug Report

We are running `dd-trace-rb` `2.14.0` on Ruby `3.3.6`. One service is a Resque worker that occasionally uses the Subprocess gem (1.5.7) to fork / exec a subprocess.

We are seeing this subprocess intermittently exit with an unhandled SIGPROF: `terminated by signal 27`. This issue feels similar to https://github.com/DataDog/dd-trace-rb/issues/2955 . 

Given the SIGPROF is unhandled, I think it's more likely that it's being delivered to the post-exec new process (which has its handlers reset to SIG_DFL) than that it's going to the post-fork, pre-exec Ruby process.

One way that could happen is from a race condition between `pthread_kill` and Ruby's `exec`. It's possible that the SIGPROF from https://github.com/DataDog/dd-trace-rb/blob/v2.14.0/ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c#L677 is sent to the thread that's about to perform an `exec`, but is pending until after the `exec` completes and the new process image takes over.

In the meantime, we're going to disable the Signal-based profiling to see if this stops the errors.

### Reproduction

A basic reproduction script of the `pthread_kill` / `exec` timing issue scenario is included. It uses FFI to run a loop that repeatedly uses `pthread_kill` to Signal a Ruby thread that is about to `exec`. The Ruby code has a `TRAP` handler installed. The ruby code prints something like:

```
TRAP received
TRAP received
TRAP received
TRAP received
Trace/BPT trap: 5
```

The core dump / CrashReport produced confirms that  the `Trace/BPT trap: 5` comes from the **`echo`** binary, rather than from Ruby. This shows that it's _possible_ that a `pthread_kill` initiated signal from a pre-exec thread be delivered to a post-exec (and post Signal-handler reset) process image. This is in-line with the Posix specification that `exec` preserves pending signals (https://pubs.opengroup.org/onlinepubs/007904875/functions/exec.html), but nice to see it actually happen.

### Stub C Library
Compiled with `gcc  -shared -o repro.so -fPIC repro.c`

```c
#include <unistd.h>
#include <pthread.h>

void signal_loop(unsigned long tid, int signal) {
  while(1) {
    int result = pthread_kill((pthread_t)tid, signal);
[...]

PR #5246, opened 2026-01-16
[PROF-13475] Fix profiler interrupting new processes during `exec`

ivoanjo:
**What does this PR do?**

This PR adds an on-by-default (on Ruby 2.7+) monkey patch to `Kernel#exec` whenever the profiler is enabled, so as to fix #5105 (without requiring the "no signals" workaround).

This feature can be disabled via the `DD_PROFILING_EXEC_WORKAROUND_ENABLED` environment variable or the `profiling.advanced.exec_workaround_enabled` setting.

**Motivation:**

The root cause of the issue is that the profiler gathers data by sending `SIGPROF` unix signals to Ruby application threads.

When using `Kernel#exec` on Linux, it can happen that a signal sent before calling `exec` arrives after the new process is running, causing it to fail with the `Profiling timer expired` error message.

For this to happen, it needs very specific timing, so that's why it took a while for us to chase this down and understand what was going on.

**Change log entry**

Yes. Fix profiler interrupting new processes with the message `Profiling timer expired` during `exec`

**Additional Notes:**

To fix this issue, we build atop PR the capability added in #5239 to stop the profiler quickly, and in this PR we add the monkey patch that stops the profiler before going into `exec`.

**How to test the change?**

I've added test coverage for the change itself, and it's possible when running with `DD_TRACE_DEBUG=true` to observe that the profiler stops (and logs that it stops) before `exec` is called.

To reproduce the exact issue from #5101 I used the `repro.so` library from the bug description with a slightly modified version of the reproducer:

```ruby
require 'ffi'
require 'datadog'

module Repro
  extend FFI::Library
  ffi_lib 'repro.so'

  attach_function :signal_loop, %i[ulong int], :void, blocking: true
  attach_function :get_tid, [], :ulong
end

deadline = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_second) + 0.001

while (now = Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_second)) < deadline
  # Burn cpu on purpose
end

exec(%w[echo echo], 'hello', {})
```

as well as this awful hack to make the profiler sample so fast (this slows down the app massively) that almost always on my machine I'm able to trigger the issue:

```diff
diff --git a/ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c b/ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
index e0631aa4dc..ccdb3b7a79 100644
--- a/ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
+++ b/ext/datadog_profiling_native_extension/collectors_cpu_and_wall_time_worker.c
@@ -663,7 +663,9 @@ static void handle_sampling_signal(DDTRACE_UNUSED int _signal, DDTRACE_UNUSED si
 static void *run_sampling_trigger_loop(void *state_ptr) {
   cpu_and_wall_time_worker_state *state = (cpu_and_wall_time_worker_state *) state_ptr;

   uint64_t minimum_time_between_signals = MILLIS_AS_NS(10);
+
+  sleep_for(MILLIS_AS_NS(10));

   while (atomic_load(&state->should_run)) {
     state->stats.trigger_sample_attempts++;
@@ -708,15 +710,15 @@ static void *run_sampling_trigger_loop(void *state_ptr) {
       }
     }

-    sleep_for(minimum_time_between_signals);
+    // sleep_for(minimum_time_between_signals);

     // The dynamic sampling rate module keeps track of how long samples are taking, and in here we extend our sleep time
     // to take that into account.
     // Note that we deliberately should NOT combine this sleep_for with the one above because the result of
     // `dynamic_sampling_rate_get_sleep` may have changed while the above sleep was ongoing.
-    uint64_t extra_sleep =
+    /*uint64_t extra_sleep =
       dynamic_sampling_rate_get_sleep(&state->cpu_dynamic_sampling_rate, monotonic_wall_time_now_ns(DO_NOT_RAISE_ON_FAILURE));
-    if (state->dynamic_sampling_rate_enabled && extra_sleep > 0) sleep_for(extra_sleep);
+    if (state->dynamic_sampling_rate_enabled && extra_sleep > 0) sleep_for(extra_sleep);*/
   }

   return NULL; // Unused
```

### DIFF
--- a/spec/datadog/profiling/component_spec.rb
+++ b/spec/datadog/profiling/component_spec.rb
@@ -145,13 +145,9 @@ RSpec.describe Datadog::Profiling::Component do
           build_profiler_component
         end
 
-        context "when gc_enabled is true" do
+        context "when gc_enabled is true", ruby: '>= 2.7' do
           before do
             settings.profiling.advanced.gc_enabled = true
-
-            # Since RUBY_VERSION is used to test if the ExecMonkeyPatch should be applied, mocking it below on an old
-            # Ruby would cause it to wrongly be triggered, so we avoid running these specs there.
-            skip "Behavior does not apply to current Ruby version" if RUBY_VERSION < "2.7"
 
             stub_const("RUBY_VERSION", testing_version)
           end
@@ -319,18 +315,13 @@ RSpec.describe Datadog::Profiling::Component do
           end
         end
 
-        context "when heap profiling is enabled" do
+        context "when heap profiling is enabled", ruby: '>= 2.7' do
           # Universally supported ruby version for allocation profiling by default
           let(:testing_version) { "3.3.0" }
 
           before do
             settings.profiling.advanced.experimental_heap_enabled = true
             settings.profiling.advanced.gc_enabled = false # Disable this to avoid any additional warnings coming from it
-
-            # Since RUBY_VERSION is used to test if the ExecMonkeyPatch should be applied, mocking it below on an old
-            # Ruby would cause it to wrongly be triggered, so we avoid running these specs there.
-            skip "Behavior does not apply to current Ruby version" if RUBY_VERSION < "2.7"
-
             stub_const("RUBY_VERSION", testing_version)
           end
 
@@ -686,13 +677,11 @@ RSpec.describe Datadog::Profiling::Component do
           end
         end
 
-        context "when can_apply_exec_monkey_patch? is true" do
+        context "when can_apply_exec_monkey_patch? is true", ruby: '>= 2.7' do
           let(:exec_monkey_patch) { class_double(exec_monkey_patch_name, apply!: true) }
 
           before do
             allow(described_class).to receive(:can_apply_exec_monkey_patch?).and_return(true)
-
-            skip "Behavior does not apply to current Ruby version" if RUBY_VERSION < "2.7"
 
             require "datadog/profiling/ext/exec_monkey_patch" # Make sure it's loaded, so we can mock it cleanly
 
--- a/spec/datadog/profiling/ext/exec_monkey_patch_spec.rb
+++ b/spec/datadog/profiling/ext/exec_monkey_patch_spec.rb
@@ -2,11 +2,10 @@
 
 require "datadog/profiling/spec_helper"
 
-RSpec.describe "Datadog::Profiling::Ext::ExecMonkeyPatch" do
+RSpec.describe "Datadog::Profiling::Ext::ExecMonkeyPatch", ruby: '>= 2.7' do
   let(:described_class) { Datadog::Profiling::Ext::ExecMonkeyPatch }
 
   subject! do
-    skip("This monkey patch is not available on Ruby 2.6 or below") if RUBY_VERSION < "2.7"
     skip("Profiling is not supported on JRuby") if PlatformHelpers.jruby?
 
     require "datadog/profiling/ext/exec_monkey_patch"


lib/datadog/core/configuration/settings.rb
```diff
@@ -436,6 +436,23 @@ def initialize(*_)
...
+            # When using `Kernel#exec` on Linux, it can happen that a signal sent before calling `exec` arrives after
+            # the new process is running, causing it to fail with the `Profiling timer expired` error message.
+            # To avoid this, the profiler installs a monkey patch on `Kernel#exec` to stop profiling before actually
+            # calling `exec`.
+            # This monkey patch is available for Ruby 2.7+; let us know if you need it on earlier Rubies.
+            # For more details see https://github.com/DataDog/dd-trace-rb/issues/5101 .
+            #
+            # @default `DD_PROFILING_EXEC_WORKAROUND_ENABLED` environment variable as a boolean,
+            # otherwise `true`
+            option :exec_workaround_enabled do |o|
```
marcotc: Why not use a "positive" (direct? descriptive?) name like
```suggestion
            option :shutdown_on_exec do |o|
```
since it's a sensible behaviour (not really a workaround, is it?), and the defaults is `true`, which I'm in agreement with. It's also probably an option that won't be needed if this change is working correctly for a while.

ivoanjo:
Re: https://github.com/DataDog/dd-trace-rb/pull/5247#issuecomment-3762215848

> Just a general thought: have you considered a library constructor, using [__attribute__((constructor))](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-constructor-function-attribute), like [this example](https://github.com/namhyung/uftrace/blob/a7b5dd2f6120366e27107266370f523c4e771ee0/libmcount/mcount.c#L2258)? It ultimately sets up signal handlers in [parse_sigspec](https://github.com/namhyung/uftrace/blob/a7b5dd2f6120366e27107266370f523c4e771ee0/libmcount/mcount.c#L352). My reading is that constructors run before a signal arrives in the new process... but now that I think of it, the new process might be unrelated and have nothing to do with dd-trace-rb, so I guess it won't solve this problem.

The problem is indeed the latter -- the new process is _completely_ unrelated and it's kinda very bizarre that the signal _can_ arrive in this manner and it's one of those  "I wouldn't believe it if I couldn't reproduce it" situations...

If it was another Ruby process running dd-trace-rb, we could instead mask off `SIGPROF` signals and then re-enable them again when the profiler starts (the signal mask is also inherited by child processes) which would be a lot more straightforward (and it's roughly what we already do in https://github.com/DataDog/dd-trace-rb/blob/master/lib/datadog/profiling/ext/dir_monkey_patches.rb ).

lib/datadog/core/configuration/settings.rb
ivoanjo: That's a good point!

I guess I was thinking here of this in practice being similar to the `DirMonkeyPatches` -- we're monkey patching core as a only way of working around a weird system issue.

But this naming I guess does over-index on _how_ we do this and not what/why. That is, if on Ruby 4.1 there was a `at_exec` feature in the VM, we would no longer _need_ a monkey patch, but we would still need to shutdown. And your suggested name would still make sense for that future, whereas "exec_workaround" would sound increasingly weird.

TL;DR I like your idea! I'll need to update the feature parity dashboard and whatnot, but I'll do such changes before merging it.

Strech: I find "workaround" too ambiguous and uncertain, but of course we don't want to expose too much details in the name. Agree with Marco on something that highlight what behavior will be enabled, instead of general wording

sig/datadog/profiling/ext/exec_monkey_patch.rbs
```diff
@@ -0,0 +1,15 @@
+module Datadog
+  module Profiling
+    module Ext
+      module ExecMonkeyPatch
+        def self.apply!: () -> true
+      end
+
+      module ExecMonkeyPatch
+        module ObjectMonkeyPatch
+          def exec: (?) -> untyped
```
Strech: Not sure, but shouldn't we mark it as `void`?

spec/datadog/profiling/component_spec.rb
```diff
@@ -147,6 +148,11 @@
         context "when gc_enabled is true" do
           before do
             settings.profiling.advanced.gc_enabled = true
+
+            # Since RUBY_VERSION is used to test if the ExecMonkeyPatch should be applied, mocking it below on an old
+            # Ruby would cause it to wrongly be triggered, so we avoid running these specs there.
+            skip "Behavior does not apply to current Ruby version" if RUBY_VERSION < "2.7"
```
Strech: Just in case we have a helper https://github.com/DataDog/dd-trace-rb/blob/master/spec/spec_helper.rb#L134-L136

lib/datadog/core/configuration/settings.rb
ivoanjo: Renamed in e121503f356b5ac43086012b1dd978b8f02bbaa5

sig/datadog/profiling/ext/exec_monkey_patch.rbs
ivoanjo: Yup, fixed in da7ed2c1118089a591f2d6b4980082ffdcf24077

spec/datadog/profiling/component_spec.rb
ivoanjo: Aaah I had forgotten it existed. Updated in 00cc5a1b2ce0c1be5443d432744836f400bcce0b .`component_spec.rb` has a lot more places where we could use this helper but I focused only on cleaning up the changes in this PR to avoid a really big random diff. (I'll either migrate over time or send a big PR later, it's a small thing so no rush IMO)

PR #5247, opened 2026-01-16
[NO-TICKET] Fix rare race in profiler causing flaky spec on Ruby 2.7

ivoanjo:
**What does this PR do?**

While working on another PR, I was running the profiler test suite locally on Ruby 2.7 and spotted that
`./spec/datadog/profiling/collectors/stack_spec.rb:872` was sometimes flaky:

```
Datadog::Profiling::Collectors::Stack
  when sampling a thread with empty locations
    gathers a one-element stack with a "In native code" placeholder (FAILED - 1)

Failures:

  1) Datadog::Profiling::Collectors::Stack when sampling a thread with empty locations gathers a one-element stack with a "In native code" placeholder
     Failure/Error: expect(gathered_stack).to contain_exactly(have_attributes(base_label: "", path: "In native code", lineno: 0))

       expected collection contained:  [(have attributes {:base_label => "", :lineno => 0, :path => "In native code"})]
       actual collection contained:    []
       the missing elements were:      [(have attributes {:base_label => "", :lineno => 0, :path => "In native code"})]
     # ./spec/datadog/profiling/collectors/stack_spec.rb:873:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:279:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:157:in `block (2 levels) in <top (required)>'
     # ruby-2.7.8/gems/webmock-3.26.1/lib/webmock/rspec.rb:39:in `block (2 levels) in <top (required)>'
     # ruby-2.7.8/gems/rspec-wait-0.0.10/lib/rspec/wait.rb:47:in `block (2 levels) in <top (required)>'
     # ./spec/support/execute_in_fork.rb:32:in `run'

Finished in 0.03459 seconds (files took 0.3921 seconds to load)
1 example, 1 failure

Failed examples:

rspec ./spec/datadog/profiling/collectors/stack_spec.rb:872 # Datadog::Profiling::Collectors::Stack when sampling a thread with empty locations gathers a one-element stack with a "In native code" placeholder
```

The root of the issue lies in the `thread_with_empty_locations` might not be completely created by the time we sample it.

In fact, if I added a
`wait_for { thread_with_empty_locations.status }.to eq "sleep"` to the test that failed, the flakiness went away.

Debugging it, in practice what happened in the failed runs is that the profiler observed `ec->vm_stack == NULL` and thus went with `return 0` instead of `PLACEHOLDER_STACK_IN_NATIVE_CODE`.

This PR fixes the issue by adding a check for `cfp == NULL` similar to the one we already had below for `end_cfp == NULL`.

**Motivation:**

In practice a real profile has thousands of samples so skipping one while the thread is being created is not a real issue, but the flakiness is, so I decided to fix it anyway.

**Change log entry**

Yes. Minor: Fix rare race in profiler causing flaky spec on Ruby 2.7

**Additional Notes:**

N/A

**How to test the change?**

I used
`rspec_n -s ./spec/datadog/profiling/collectors/stack_spec.rb:872 1000` to trigger the issue on `ruby 2.7.8p225 (2023-03-30 revision 1f4d455848) [x86_64-linux]`and usually would get a failure within 1-5 tries.

After I changed this logic, I could no longer reproduce the failure.

### DIFF
--- a/.github/actions/bundle-cache/action.yml
+++ b/.github/actions/bundle-cache/action.yml
@@ -13,18 +13,17 @@ outputs:
 runs:
   using: "composite"
   steps:
-    - id: ruby-alias
+    - name: Generate unique Ruby environment key
+      id: ruby-env-id
       shell: bash
       run: |
-        engine=$(ruby -e "puts RUBY_ENGINE")
-        alias=$(ruby -e "puts RUBY_ENGINE_VERSION")
-
-        echo "ruby-alias=$engine-$alias" >> "$GITHUB_OUTPUT"
+        id=$(ruby -e "puts RUBY_DESCRIPTION") # Includes patchlevel, os, and arch
+        echo "id=$id" >> "$GITHUB_OUTPUT"
     - name: Generate lockfile
       id: lockfile
       shell: bash
       env:
-        LOCKFILE: lockfile-${{ steps.ruby-alias.outputs.ruby-alias }}
+        LOCKFILE: lockfile-${{ steps.ruby-env-id.outputs.id }}
       run: |
         bundle lock
         echo "lockfile=$LOCKFILE" >> "$GITHUB_OUTPUT"
@@ -37,7 +36,7 @@ runs:
       uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
       id: restore-cache
       with:
-        key: bundle-${{ runner.os }}-${{ runner.arch }}-${{ steps.ruby-alias.outputs.ruby-alias }}-${{ hashFiles('*.lock') }}
+        key: bundle-${{ steps.ruby-env-id.outputs.id }}-${{ hashFiles('*.lock') }}
         path: "/usr/local/bundle"
     - if: steps.restore-cache.outputs.cache-hit != 'true'
       shell: bash


marcotc:
Just a general thought: have you considered a library constructor, using [`__attribute__((constructor))`](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-constructor-function-attribute), like [this example](https://github.com/namhyung/uftrace/blob/a7b5dd2f6120366e27107266370f523c4e771ee0/libmcount/mcount.c#L2258)? It ultimately sets up signal handlers in [parse_sigspec](https://github.com/namhyung/uftrace/blob/a7b5dd2f6120366e27107266370f523c4e771ee0/libmcount/mcount.c#L352). My reading is that constructors run before a signal arrives in the new process... but now that I think of it, the new process might be unrelated and have nothing to do with dd-trace-rb, so I guess it won't solve this problem.

(Btw, the current fix looks good!)

ivoanjo:
@marcotc I believe your comment is related to the other PR you just reviewed (thanks btw!) -- I'll reply there https://github.com/DataDog/dd-trace-rb/pull/5246 .

PR #5252, opened 2026-01-19
[APPSEC-60626] Fix type spec for vendor/gem.rbs

Strech:
**What does this PR do?**

Fix type definition of `Gem::BasicSpecification#version` method

**Motivation:**

This is the way

**Change log entry**

No.

**Additional Notes:**

None.

**How to test the change?**

CI

### DIFF
--- a/sig/datadog/profiling/collectors/code_provenance.rbs
+++ b/sig/datadog/profiling/collectors/code_provenance.rbs
@@ -2,38 +2,54 @@ module Datadog
   module Profiling
     module Collectors
       class CodeProvenance
-        def initialize: (?standard_library_path: String, ?ruby_native_filename: String?) -> void
-        def refresh: (?loaded_files: Array[String], ?loaded_specs: Array[Gem::BasicSpecification]) -> self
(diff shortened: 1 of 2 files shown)


sig/datadog/profiling/collectors/code_provenance.rbs
```diff
@@ -2,38 +2,56 @@ module Datadog
...
+
+          @kind: ::String
+
+          @name: ::String
+
+          @version: ::String
+
+          @paths: ::Array[::String]
+
+          def initialize: (kind: ::String, name: ::String, version: version, path: ::String, extra_paths: ::Array[::String?]) -> void
```
ivoanjo: Minor: It's kinda weird to define the `version` separately and only use it once?

ivoanjo (approved):
👍 LGTM thanks for the update

sig/datadog/profiling/collectors/code_provenance.rbs
Strech: Yeah, will drop it, I thought it was something, but then nope, just once, and I forget to inline it. Will do 👍🏼

PR #5255, opened 2026-01-19
Fix typing for AppSec::Component

y9v:
**What does this PR do?**
This PR fixes types for `AppSec::Component`, and removes it from steep ignored files.

**Motivation:**
We don't want any AppSec files to be ignored by our steep configuration.

**Change log entry**
None. Internal change.

**Additional Notes:**
APPSEC-60626

**How to test the change?**
CI

### DIFF
--- a/sig/datadog/core/configuration/settings.rbs
+++ b/sig/datadog/core/configuration/settings.rbs
@@ -33,27 +33,27 @@ module Datadog
 
           def enabled=: (bool) -> void
 
-          def instrument: (Symbol) -> void
+          def instrument: (::Symbol) -> void
 
-          def trace_rate_limit: () -> Integer
+          def trace_rate_limit: () -> ::Integer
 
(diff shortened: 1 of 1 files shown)


sig/datadog/core/configuration/settings.rbs
```diff
@@ -33,6 +33,8 @@ module Datadog
 
           def enabled=: (bool) -> void
 
+          def instrument: (Symbol) -> void
```
Strech:
```suggestion
          def instrument: (::Symbol) -> void
```

y9v: good point, thanks! Will do the same for other core types for `_AppSec` interface here [5631706](https://github.com/DataDog/dd-trace-rb/pull/5255/commits/56317068b89b3fa5d35ac980064ed9a7764d631f)

PR #5257, opened 2026-01-19
CI: Cache gemfiles

marcotc:
**What does this PR do?**

Adds one downloaded-package cache per MRI version for the appraisal Gemfiles selected by `Matrixfile`. Unit-test child jobs restore an exact cache or the newest same-runtime fallback before installing their assigned appraisal bundles.

Default-branch jobs populate and save exact cache generations from committed lockfiles. Pull request jobs are restore-only, so dependency changes reuse the previous generation and download only missing packages without creating branch-specific generations. The existing installed base-bundle cache does not change.

Package manifests share the same matrix selection as test batching, subtract packages supplied by the base bundle, select the runner platform, and retain the Ruby 2.5 Git-sourced Rails revision. Ordinary gem archives download directly from their locked RubyGems sources; Bundler runs only when a Git source must be packaged.

**Motivation:**

Matrix child jobs currently download the same appraisal packages repeatedly. Caching package archives once per Ruby version reduces duplicate network work without storing every installed appraisal bundle or serializing matrix installation.

**Change log entry**

None. This changes CI dependency caching only.

**Additional Notes:**

Superseded immutable generations rely on GitHub's inactivity and quota-based eviction. Package-cache writes require both an explicit caller opt-in and the default branch ref.

Hosted Ruby 4.0 production-manifest measurement produced 760 files: 126,930,432 bytes before cache compression and 124,069,015 bytes in GitHub cache storage. Cold population took 4 seconds, save took 6 seconds, and an exact restore took 1.573 seconds. The isolated measurement cache was deleted afterward.

**How to test the change?**

- `bundle exec rspec spec/tasks/github_matrix_spec.rb spec/tasks/package_cache_manifest_spec.rb spec/tasks/package_cache_archive_spec.rb` — 15 examples, 0 failures.
- Focused RuboCop, `yamllint --strict`, `actionlint`, `git diff --check`, and focused `act` validation pass.
- Focused `act` scenarios cover cold save, warm exact restore, absent cache, fallback restore, repeated missing delta, and rolling replacement.
- Production manifests populated and validated locally for Ruby 2.5 through 4.0. Ruby 2.5 Git-sourced Rails and a representative Ruby 2.6 appraisal installed from local package archives without package downloads.
- Hosted fixture and production Ruby 4.0 cold/warm runs passed under isolated `bundle-packages-test-v1` keys; all test keys were deleted after measurement.


ivoanjo:
Hey, this is kinda out-of-date -- is this something we still want?

marcotc:
I just checked today, and I still want to add the caching I started in this PR, even after all recent changes (the caching WIP in this PR has not yet been addressed).

PR #5258, opened 2026-01-20
[APPSEC-60626] Return rest_client to be checked by Steep

Strech:
**What does this PR do?**

Returns `rest_client` contrib under the Steep check

**Motivation:**

We want to eliminate all ignored AppSec files

**Change log entry**

No.

**Additional Notes:**

RBS and Steep can't comprehend mutation method like `Hash#transform_values!` because that method type spec is defined as `def transform_values!: () { (V) -> V } -> self` which locks the type of the `V` and Steep assumed that result will have the same type before and after the method call. Because of that I have to ignore one function 😞 

**How to test the change?**
CI


lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
```diff
@@ -48,12 +48,17 @@ def normalize_request_headers
             processed_headers.transform_keys(&:downcase)
           end
 
+          # NOTE: Headers values are always an `Array` in `Net::HTTPResponse`,
+          #       but we want to avoid accidents and will wrap them in no-op
+          #       `Array` call just in case of a breaking change in the future
+          #
+          # FIXME: Steep has issues with `transform_values!` modifying the original
+          #        type and it failed with "Cannot allow block body" error
```
vpellan, line 56: This is really weird because for `transform_values` it [has the correct type signature](https://github.com/ruby/rbs/blob/5771976d7d0dd6be13e42d2d791d5ccbfee68bda/core/hash.rbs#L1981-L1982) and works... Why is `transform_values!` typed differently ? That makes no sense...
Could you open an issue on the rbs repo for [these lines](https://github.com/ruby/rbs/blob/5771976d7d0dd6be13e42d2d791d5ccbfee68bda/core/hash.rbs#L2003-L2004), or a PR to fix it upstream, and add the link here to keep track of it ? If you don't have the time I can do it!

EDIT: It was actually working previously but was changed in https://github.com/ruby/rbs/commit/aae95840712b2b5b4f7833a52b380a3b8d4d8398
Maybe this type signature could work ?
```
def transform_values!: () -> Enumerator[V, self]
                       | () { (V) -> V } -> self
                       | [A] () { (V) -> A } -> Hash[K, A]
```
(Haven't tested it locally yet)
I'd suggest to still open an issue

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
```diff
@@ -48,12 +48,17 @@ def normalize_request_headers
...
+          # steep:ignore:start
           def normalize_response_headers(response)
-            # NOTE: Headers values are always an `Array` in `Net::HTTPResponse`,
-            #       but we want to avoid accidents and will wrap them in no-op
-            #       `Array` call just in case of a breaking change in the future
             response.net_http_res.to_hash.transform_values! { |value| Array(value).join(', ') }
```
marcotc: Do you think we are able to override the stdlib definition of `transform_values!: () { (V) -> V } -> self` to fix this type issue? No worries if not.

Strech: I don't think so, and the reason for that - we don't have deterministic way to define the resulting `V`.

Technically we can go with `V | untyped`, but that's actually just `untyped` ... and I'm not sure what's the best here, to allow anything (no type checking for untyped) or to lock the type.

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
```diff
@@ -48,12 +48,17 @@ def normalize_request_headers
...
           end
 
+          # NOTE: Headers values are always an `Array` in `Net::HTTPResponse`,
+          #       but we want to avoid accidents and will wrap them in no-op
+          #       `Array` call just in case of a breaking change in the future
+          #
+          # FIXME: Steep has issues with `transform_values!` modifying the original
+          #        type and it failed with "Cannot allow block body" error
+          # steep:ignore:start
           def normalize_response_headers(response)
```
vpellan: I think we should narrow the type errors as much as possible
```suggestion
def normalize_response_headers(response) # steep:ignore MethodBodyTypeMismatch
```

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
```diff
@@ -48,12 +48,17 @@ def normalize_request_headers
...
-            # NOTE: Headers values are always an `Array` in `Net::HTTPResponse`,
-            #       but we want to avoid accidents and will wrap them in no-op
-            #       `Array` call just in case of a breaking change in the future
             response.net_http_res.to_hash.transform_values! { |value| Array(value).join(', ') }
           end
+          # steep:ignore:end
```
vpellan:
```suggestion
response.net_http_res.to_hash.transform_values! { |value| Array(value).join(', ') } # steep:ignore BlockBodyTypeMismatch
end
```

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
vpellan: We're not able to override type signatures provided by RBS unfortunately... This will result in early steep failure with RBS::DuplicatedDeclaration error. I think this is a feature missing from Steep ?

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
Strech: There is an interesting commit message

> It's not possible to represent side-effects and the receiver type changing, however some type checkers may choose to perform weak updates based on the types passed to these methods and in these situations its useful to still use the original type variables and not method-bound ones.

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
Strech: Nice, thanks!

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
Strech: ~I'm going to open a PR with fix, I think both could be same as `transform_values` then~ Issue is better

lloeki: >  Why is transform_values! typed differently ? That makes no sense...

To me it does! The definition is not just a freestanding `def transform_values!: () { (V) -> V } -> self`

It's:

```
class Hash[unchecked out K, unchecked out V] < Object
  include Enumerable[[ K, V ]]

  def transform_values!: () -> Enumerator[V, self]
                       | () { (V) -> V } -> self
```

Indeed since:
* the generic type `V` comes from the very definition of the type of your Hash (`Hash[K, V]`) 
* `transform_values!` mutates the hash

Ergo `transform_values!` _must not_ change the values type, otherwise it would violate the Hash's own typing!

And indeed, if you want to change the value type, you gotta return a new hash (of a different `Hash[K, V2]` type) with the non-mutating `transform_values` method:

```
  def transform_values: () -> Enumerator[V, Hash[K, untyped]]
                      | [A] () { (V) -> A } -> Hash[K, A]
```

That, or the original hash needs a wider `V | V2` type (but then typing may not be able to enforce that it's consistently `V` across all values (or `V2` afterwards), which is probably not what one wants.

Strech:
@vpellan @lloeki I've opened upstream question https://github.com/ruby/rbs/issues/2819

lib/datadog/appsec/contrib/rest_client/request_ssrf_detection_patch.rb
Strech: https://github.com/ruby/rbs/issues/2819

PR #5264, opened 2026-01-21
run CI at least once a day, including weekends

cbeauchesne:
**What does this PR do?**
Run CI once a day during weekends

**Motivation:**

> Datadog monitors are not able to skip saturday/sunday. Without those runs  it'll trigger false positives for weekends, and could also lead to missed errors. 

And also

> there was even the case of an update to some dependency (doesn't matter which) that happened on a Saturday
and that broke 2.5 CI because reasons (doesn't matter)
> but since schedule runs only on Friday late evening + Sunday late evening this was only pinpointed to a "something changed during these 48h" instead of a more direct "THAT DAY EXACTLY".

**Change log entry**

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -14,6 +14,7 @@ on: # yamllint disable-line rule:truthy
       - master
   schedule:
     - cron: 0 */4 * * 1-5
+    - cron: 0 3 * * 0,6
 
 # Default permissions for all jobs
 permissions: {}
--- a/.github/workflows/system-tests.yml
+++ b/.github/workflows/system-tests.yml
@@ -14,7 +14,7 @@ on: # yamllint disable-line rule:truthy
       - master
   workflow_dispatch: {}
   schedule:
-    - cron: "00 04 * * 2-6"
+    - cron: "00 04 * * 0-6"
 
 # Default permissions for all jobs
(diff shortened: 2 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -14,6 +14,7 @@ on: # yamllint disable-line rule:truthy
       - master
   schedule:
     - cron: 0 */4 * * 1-5
+    - cron: 0 3 * * 0,6
```
ivoanjo, line 17: I'd go as far as just making always the same -- if we ever want to reduce CI pressure, there's other ways?

```suggestion
    - cron: 0 */4 * * 0-6
```

ivoanjo (approved):
👍 This looks reasonable to me!

.github/workflows/test.yml
cbeauchesne: It would run the CI every 4h during WE. For the purpose of the monitor, it's not necessary. But let me know if you'd like to have this anyway

ivoanjo: That's fine, let's leave it as-is then

p-datadog:
We are in a release, please wait until release finishes to merge

PR #5275, opened 2026-01-23
Fix AppSec Sinatra contrib types

y9v:
**What does this PR do?**
This PR removes `lib/appsec/contrib/sinatra` from steep ignored files and fixes types.

**Motivation:**
We want all `AppSec` to be properly typechecked.

**Change log entry**

**Additional Notes:**
APPSEC-60626

**How to test the change?**
CI

### DIFF
--- a/vendor/rbs/rack/0/rack.rbs
+++ b/vendor/rbs/rack/0/rack.rbs
@@ -1,12 +1,12 @@
 module Rack
   class Request
-    def initialize: (::Hash[::String, any] env) -> void
+    def initialize: (::Hash[::String, untyped] env) -> void
 
-    def env: () -> ::Hash[::String, any]
+    def env: () -> ::Hash[::String, untyped]
 
-    def params: () -> ::Hash[::String, any]
+    def params: () -> ::Hash[::String, untyped]
 
-    def body: () -> any
+    def body: () -> untyped
 
     def get?: () -> bool
 
--- a/vendor/rbs/sinatra/0/sinatra.rbs
+++ b/vendor/rbs/sinatra/0/sinatra.rbs
@@ -4,7 +4,7 @@ module Sinatra
   end
 
   class Request
-    def env: () -> rack_env
+    def env: () -> ::Hash[::String, untyped]
   end
 
   class Base


sig/datadog/appsec/instrumentation/gateway.rbs
```diff
@@ -17,7 +17,10 @@ module Datadog
 
         def push: (::String name, argument env) ?{ () -> final_call_result } -> stack_result
 
-        def watch: (::String name, ::Symbol key) { (stack next, argument env) -> stack_result } -> void
+        def watch: ("sinatra.request.dispatch", ::Symbol key) { (stack, ::Datadog::AppSec::Contrib::Sinatra::Gateway::Request) -> stack_result } -> void
+                 | ("sinatra.request.routed", ::Symbol key) { (stack, [::Datadog::AppSec::Contrib::Sinatra::Gateway::Request, ::Datadog::AppSec::Contrib::Sinatra::Gateway::RouteParams]) -> stack_result } -> void
+                 | ("sinatra.response.body.json", ::Symbol key) { (stack, Gateway::DataContainer) -> stack_result } -> void
+                 | (::String name, ::Symbol key) { (stack, argument) -> stack_result } -> void
```
y9v: @Strech what do you think of this approach?

lib/datadog/appsec/contrib/sinatra/gateway/watcher.rb
```diff
@@ -23,7 +23,7 @@ def watch
 
               def watch_request_dispatch(gateway = Instrumentation.gateway)
                 gateway.watch('sinatra.request.dispatch', :appsec) do |stack, gateway_request|
-                  context = gateway_request.env[AppSec::Ext::CONTEXT_KEY]
+                  context = gateway_request.env[AppSec::Ext::CONTEXT_KEY] # : Context
```
marcotc, line 26: Sorry if I'm out of the loop: is this the "new" syntax for type overrides? `# : ...`

lib/datadog/appsec/contrib/sinatra/gateway/watcher.rb
```diff
@@ -49,8 +49,9 @@ def watch_request_dispatch(gateway = Instrumentation.gateway)
               end
 
               def watch_request_routed(gateway = Instrumentation.gateway)
-                gateway.watch('sinatra.request.routed', :appsec) do |stack, (gateway_request, gateway_route_params)|
-                  context = gateway_request.env[AppSec::Ext::CONTEXT_KEY]
+                gateway.watch('sinatra.request.routed', :appsec) do |stack, args|
+                  gateway_request, gateway_route_params = args # : [Gateway::Request, Gateway::RouteParams]
```
marcotc, line 53: It's annoying that you have to do this, just to add a type annotation. 😢

sig/datadog/appsec/instrumentation/gateway.rbs
marcotc: I guess our "dynamically-typed" language loses its charm in RBS signatures.

But regarding this change, I don't see how it could be made much better, it looks ok I think.

The only thing we could theoretically do is move each argument type (or signature) to the integration package: e.g. declare a type for `[::Datadog::AppSec::Contrib::Sinatra::Gateway::Request, ::Datadog::AppSec::Contrib::Sinatra::Gateway::RouteParams]`, or maybe even a type for the whole `("sinatra.request.routed", ::Symbol key) { (stack, [::Datadog::AppSec::Contrib::Sinatra::Gateway::Request, ::Datadog::AppSec::Contrib::Sinatra::Gateway::RouteParams]) -> stack_result } -> void` in the Sinatra contrib RBS namespace).
Maybe that looks better... maybe not 🤷

lib/datadog/appsec/contrib/sinatra/gateway/watcher.rb
Strech: It's type assertion `#: ...` [no space in OG example, I think with space is better tho](https://github.com/soutaro/steep/blob/master/manual/annotations.md#type-assertion)`

sig/datadog/appsec/contrib/rack/gateway/request.rbs
```diff
@@ -4,33 +4,35 @@ module Datadog
       module Rack
         module Gateway
           class Request < Instrumentation::Gateway::Argument
-            attr_reader env: untyped
+            @env: ::Hash[::String, top]
```
Strech: Does it make sense to use `any` here instead of top?

sig/datadog/appsec/contrib/sinatra/gateway/watcher.rbs
```diff
@@ -4,11 +4,13 @@ module Datadog
       module Sinatra
         module Gateway
           module Watcher
-            def self.watch: () -> untyped
+            def self.watch: () -> void
 
-            def self.watch_request_dispatch: (?Datadog::AppSec::Instrumentation::Gateway gateway) -> untyped
+            def self.watch_request_dispatch: (?Datadog::AppSec::Instrumentation::Gateway gateway) -> void
```
Strech: We can reduce namespacing here

```suggestion
            def self.watch_request_dispatch: (?AppSec::Instrumentation::Gateway gateway) -> void
```

sig/datadog/appsec/contrib/sinatra/patcher.rbs
```diff
@@ -2,6 +2,10 @@ module Datadog
   module AppSec
     module Contrib
       module Sinatra
+        interface _SinatraRequest
```
Strech: I think that should go into Sinatra stub instead, as other Sinatra defined interfaces

sig/datadog/appsec/instrumentation/gateway.rbs
Strech: There is no value in that kind of typing as it creates fragile signature - as soon as you change what you send, sigs with specific names like `sinatra.request.dispatch` will be broken

lib/datadog/appsec/contrib/sinatra/gateway/watcher.rb
y9v: yes, it's super ugly tbh, too bad we can't do type annotations for block arguments

sig/datadog/appsec/instrumentation/gateway.rbs
y9v: well, this is what typing is for, isn't it - we have watchers that watch for certain keys, and they operate with an assumption that the argument will be of a certain type?

What do you think we should do instead? Use type annotations locally in `.watch`?

Strech: > well, this is what typing is for

No, typing is not about exact values that belongs to the logic of the code, that's about defining type boundaries of the input.

> we have watchers that watch for certain keys, and they operate with an assumption that the argument will be of a certain type?

Yes we do, and you said it yourself - they wait for key of certain type, the value of the key doesn't matter for type checking because it doesn't belong to the type scope, it's a logic of the code.

y9v: removed type checking for `watch` args: [435b0c6](https://github.com/DataDog/dd-trace-rb/pull/5275/commits/435b0c631e3c5dcd01cd6575416ec2771f4671fd)

sig/datadog/appsec/contrib/sinatra/patcher.rbs
y9v: agree, fixed [f36cf95](https://github.com/DataDog/dd-trace-rb/pull/5275/commits/f36cf95796e5c676dd4357c01789f2b39705ea81)

vendor/rbs/rack/0/rack.rbs
```diff
@@ -0,0 +1,17 @@
+module Rack
+  class Request
+    def initialize: (::Hash[::String, any] env) -> void
+
+    def env: () -> ::Hash[::String, any]
+
+    def params: () -> ::Hash[::String, any]
```
Strech: I don't think we can use any in the stub because it belongs to the Datadog, here let's just use `untyped`

```suggestion
    def initialize: (::Hash[::String, untyped] env) -> void

    def env: () -> ::Hash[::String, untyped]

    def params: () -> ::Hash[::String, untyped]
```

vendor/rbs/sinatra/0/sinatra.rbs
```diff
@@ -3,6 +3,13 @@ module Sinatra
     def initialize: (::Array[::String] body, ::Integer status, ::Hash[::String, ::String] headers) -> void
   end
 
+  class Request
+    def env: () -> rack_env
```
Strech: The stub suppose to be self-sufficient and not depend from external types. Hence it's better to define `rack_env` here

y9v: yes, good point. Fixed in [506d366](https://github.com/DataDog/dd-trace-rb/pull/5275/commits/506d3660cc262b507a2c2f50c3b96a258a88fd6a)

vendor/rbs/rack/0/rack.rbs
y9v: 👍🏼 [506d366](https://github.com/DataDog/dd-trace-rb/pull/5275/commits/506d3660cc262b507a2c2f50c3b96a258a88fd6a)

PR #5281, opened 2026-01-26
[APPSEC-60450] Extract Knuth sampler into Core

Strech:
**What does this PR do?**

I'm extracting Knuth factor sampler into the Core to allow AppSec reuse it for different counting (based on internal counter) and share it with Tracer sampler.

**Motivation:**

Reduce code duplication and fill in the gaps in Core with some generics we are missing.

**Change log entry**

No. This is an internal change

**Additional Notes:**

In AppSec we want to use as input increasing counter and golden ratio, when Tracer wants to use trace ID and different factor. So composition works quite well.

In addition I've improved RateSampler typespecs ❤️ 

**How to test the change?**

CI

### DIFF
--- a/spec/datadog/core/knuth_sampler_spec.rb
+++ b/spec/datadog/core/knuth_sampler_spec.rb
@@ -77,7 +77,7 @@ RSpec.describe Datadog::Core::KnuthSampler do
       end
 
       it 'samples approximately 50% of sequential inputs' do
-        sampled_count = (0...1000).count { |i| sampler.sample?(i) }
+        sampled_count = (0...1000).sum { |i| sampler.sample?(i) ? 1 : 0 }
         expect(sampled_count).to be_within(100).of(500)
       end
     end
@@ -104,7 +104,7 @@ RSpec.describe Datadog::Core::KnuthSampler do
       subject(:sampler) { described_class.new(0.1) }
 
       it 'samples ~10% of 1000 inputs' do
-        sampled_count = (0...1000).count { |i| sampler.sample?(i) }
+        sampled_count = (0...1000).sum { |i| sampler.sample?(i) ? 1 : 0 }
         expect(sampled_count).to be_within(15).of(100)
       end
     end
@@ -113,7 +113,7 @@ RSpec.describe Datadog::Core::KnuthSampler do
       subject(:sampler) { described_class.new(0.25) }
 
       it 'samples ~25% of 1000 inputs' do
-        sampled_count = (0...1000).count { |i| sampler.sample?(i) }
+        sampled_count = (0...1000).sum { |i| sampler.sample?(i) ? 1 : 0 }
         expect(sampled_count).to be_within(38).of(250)
       end
     end
@@ -122,7 +122,7 @@ RSpec.describe Datadog::Core::KnuthSampler do
       subject(:sampler) { described_class.new(0.9) }
 
       it 'samples ~90% of 1000 inputs' do
-        sampled_count = (0...1000).count { |i| sampler.sample?(i) }
+        sampled_count = (0...1000).sum { |i| sampler.sample?(i) ? 1 : 0 }
         expect(sampled_count).to be_within(135).of(900)
       end
     end


lib/datadog/core/knuth_sampler.rb
```diff
@@ -0,0 +1,64 @@
...
+    #
+    # @see https://en.wikipedia.org/wiki/Hash_function#Multiplicative_hashing
+    class KnuthSampler
+      # Maximum unsigned 64-bit integer for uniform distribution across 64-bit input space.
+      UINT64_MAX = (1 << 64) - 1
+      UINT64_MODULO = 1 << 64
+
+      # Golden ratio constant for optimal distribution.
+      # @see https://en.wikipedia.org/wiki/Hash_function#Fibonacci_hashing
+      DEFAULT_KNUTH_FACTOR = 11400714819323198485
```
mabdinur, line 22: In ddtrace-py we set the knuth factor to: [11400714819323199488](https://github.com/DataDog/dd-trace-py/blame/3600f3b0383e98dd56255a264127f5dcb65fb6b2/ddtrace/appsec/_asm_request_context.py#L149). This looks like a rounding error in dd-trace-py.

lib/datadog/core/knuth_sampler.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      # Sets the rate and recalculates the sampling threshold.
+      #
+      # @param rate [Float] Sampling rate between +0.0+ and +1.0+ (inclusive).
+      #   Invalid values fall back to +1.0+ (sample everything).
+      def rate=(rate)
+        rate = rate.to_f
+
+        unless rate >= 0.0 && rate <= 1.0
+          Datadog.logger.warn('Sample rate is not between 0.0 and 1.0, falling back to 1.0')
```
mabdinur: nit: Can we log the invalid sample rate. It will make the code easier to debug

lib/datadog/tracing/sampling/rate_sampler.rb
```diff
@@ -1,54 +1,43 @@
...
           super()
-
-          unless sample_rate >= 0.0 && sample_rate <= 1.0
-            Datadog.logger.warn('sample rate is not between 0 and 1, falling back to 1')
-            sample_rate = 1.0
-          end
-
-          self.sample_rate = sample_rate
-
+          @sampler = Core::KnuthSampler.new(sample_rate, knuth_factor: KNUTH_FACTOR)
```
mabdinur, line 21: Can we make the sampler attribute private. This will reduce the size of our public api

lib/datadog/core/knuth_sampler.rb
Strech: Yes, I've notice that too, and [here is discussion](https://dd.slack.com/archives/C01TTSAKVGW/p1769438990719929).

lib/datadog/core/knuth_sampler.rb
Strech: Good idea 👍🏼

lib/datadog/tracing/sampling/rate_sampler.rb
Strech: How do you mean? Not sure that I get which attribute you are referring too.

lib/datadog/core/knuth_sampler.rb
```diff
@@ -0,0 +1,64 @@
...
+    # Deterministic sampler using Knuth multiplicative hash algorithm.
+    #
+    # This sampler provides consistent sampling decisions based on an input value,
+    # ensuring the same input always produces the same sampling decision for a given rate.
+    #
+    # The algorithm multiplies the input by a large prime (Knuth factor), takes modulo
+    # to constrain to a fixed range, and compares against a threshold derived from the sample rate.
+    #
+    # @see https://en.wikipedia.org/wiki/Hash_function#Multiplicative_hashing
+    class KnuthSampler
```
p-datadog, line 15: Is this class meant to be used by customers? Do you mind adding `@api private` or `@api public` to indicate this?

lib/datadog/tracing/sampling/rate_sampler.rb
```diff
@@ -1,54 +1,43 @@
...
 
         def sample_rate(*_)
-          @sample_rate
+          @sampler.rate
         end
 
         def sample_rate=(sample_rate)
-          @sample_rate = sample_rate
-          @sampling_id_threshold = sample_rate * Tracing::Utils::EXTERNAL_MAX_ID
+          @sampler.rate = sample_rate
```
p-datadog: `@sampler` can be recreated here instead of being mutated, which I think is generally the pattern we use throughout the library. Is there a particular reason for adding the mutation to the new class?

lib/datadog/core/knuth_sampler.rb
Strech: It's private, will add 👍🏼

lib/datadog/tracing/sampling/rate_sampler.rb
Strech: Right, I did inherit the mutation from the original Tracer implementation. But I think you right, we can get rid of it as that's the only place to change it and it could be done via recreation indeed.

I will remove mutation and change code here 👍🏼

lib/datadog/tracing/sampling/rate_sampler.rb
Strech: I'm going to resolve it because nothing extra is exposed, but let me know if you think otherwise 🙏🏼

p-datadog:
:+1: merci

PR #5286, opened 2026-01-28
(DO-NOT-MERGE) Configuration descriptions

vpellan:
**What does this PR do?**

Write a description for all configs

**Motivation:**

Config registry will be used to generate public doc, so it is important to have a description for all of them.
The short description will be used by the 'Service Config' -> 'SDK & Agent Configuration' panel on the UI so this is also important that all of them are completed.

**Change log entry**

None

**Additional Notes:**

**How to test the change?**

### DIFF
--- a/config-descriptions.json
+++ b/config-descriptions.json
@@ -6,42 +6,48 @@
       "description": "Enables AI Guard so the library can make AI Guard API calls; disabling it skips those calls without removing the library.",
       "existingPublicDescription": null,
       "shortDescription": "Enable AI Guard so the library can make AI Guard API calls",
-      "internal": false
+      "internal": false,
+      "alreadyDocumentedOnFpd": false
     },
     "DD_AI_GUARD_ENDPOINT": {
       "programmaticKey": "ai_guard.endpoint",
       "description": "URL of the AI Guard API endpoint that this library will send requests to.",
       "existingPublicDescription": null,
       "shortDescription": "URL of the AI Guard API endpoint that this library will send requests to",
-      "internal": false
+      "internal": false,
+      "alreadyDocumentedOnFpd": true
     },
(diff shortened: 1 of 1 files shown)


config-descriptions.json
```diff
@@ -0,0 +1,2705 @@
+{
+  "rules": "description < 1000 chars, shortDescription < 100 chars",
+  "ai_guard": {
+    "DD_AI_GUARD_ENABLED": {
+      "programmaticKey": "ai_guard.enabled",
+      "description": "Enables AI Guard so the library can make AI Guard API calls; disabling it skips those calls without removing the library.",
```
y9v, line 6: should we add some information on how those calls are skipped when AI Guard is disabled? What it does is that AI Guard SDK will never raise (when configured to raise on ABORT or DENY), or always will return `ALLOW` evaluation (when configured not to raise)

config-descriptions.json
```diff
@@ -0,0 +1,2705 @@
...
+      "existingPublicDescription": null,
+      "shortDescription": "Enable AI Guard so the library can make AI Guard API calls",
```
y9v, line 8: "Enable calls to AI Guard API via the SDK"?

config-descriptions.json
```diff
@@ -0,0 +1,2705 @@
...
+      "internal": false,
+      "alreadyDocumentedOnFpd": false
+    },
+    "DD_AI_GUARD_ENDPOINT": {
+      "programmaticKey": "ai_guard.endpoint",
+      "description": "URL of the AI Guard API endpoint that this library will send requests to.",
+      "existingPublicDescription": null,
+      "shortDescription": "URL of the AI Guard API endpoint that this library will send requests to",
```
y9v, line 16: "URL to AI Guard API endpoint"?

PR #5287, opened 2026-01-28
Fix flaky AppSec Rack integration spec

y9v:
**What does this PR do?**
This PR changes the way how RC worker is accessed on `remote` component. If this won't fix the actual flakyness, it will at least make the error message more clear.

**Motivation:**
This test is flaky:
[Merge pull request #5251 from DataDog/appsec-60626-return-excon-and-r… · DataDog/dd-trace-rb@409b199](https://github.com/DataDog/dd-trace-rb/actions/runs/21139472197/job/60789597575)

**Change log entry**
None. Internal change.

**Additional Notes:**
APMLP-933

**How to test the change?**
CI


y9v:
@strech no, this is not what the error is - the error is that `call` is being called on a nil object - on component remote worker (as I assume after investigation):
```
Rack integration tests for an application with remote configuration enabled not responding has boot tags
     Failure/Error: expect(span).to have_tag('_dd.rc.boot.time')

     NoMethodError:
       undefined method `call' for nil:NilClass
       Did you mean?  caller
```

Issue #1197, opened 2020-10-02
Easy and unobtrusive method tracing


norbertnytko:
Tracking a lot of method calls can be cumbersome in complex systems. Imagine repeating over and over the same piece of code for dozens of classes and methods.

```ruby
# models/foo.rb
class Foo
  SPAN_NAME = "method.call"
  def bar
    Datadog.tracer.trace("method.call", resource: "#{self.class}.#{__method__}") do
      execute_awesome_logic
    end
  end
end
```
It gives control, but in exchange blurs what's really important here - business logic.

I'd like to propose two alternative approaches.

Well know based on mixin.
```ruby
# models/foo.rb
class Foo
  include Datadog::MethodTracer
  trace_method :bar

  def bar
    execute_awesome_logic
  end
end
```

Or a bit more sophisticated based on [IoC](https://martinfowler.com/bliki/InversionOfControl.html) principles and inspired by Aspect-oriented programming.
```ruby
# foo.rb
class Foo
  def bar
    execute_awesome_logic
  end
end
```
```ruby
# initializers/method_tracer.rb

# instance based
method_tracer = Datadog::MethodTracer.new
method_tracer.add_method Foo, :bar

# or class based
Datadog::MethodTracer.add_method Foo, :bar
```

Already have PoC for those two, let me know if it makes sense to include in gem.

norbertnytko:
@marcotc @delner I'd appreciate your feedback.

marcotc:
👋 @norbertnytko, thank you for bringing this up!
We've had an implementation before of the first approach you suggested (https://github.com/DataDog/dd-trace-rb/pull/670) and I think it makes in the tracer. Given community feedback from many different users, I don't see a reason to withhold such feature.

I want to make sure to respect your time, so I'm gathering internal feedback about this feature before I tell for sure that you should work towards preparing a PR with you proposed changes. I'll let you know soon!

norbertnytko:
@marcotc Bump https://github.com/DataDog/dd-trace-rb/pull/1215

lewhit:
Any updates on a timeline for getting this feature out?

ivoanjo:
Thanks for the patience, folks! Our amazing colleague @lloeki is actually working on a new approach to method tracing to support these and other use cases we’re running into. We’ve closed off #1348 for now, and should have something to show in its place pretty soon :slightly_smiling_face:

No pressure, @lloeki . None at all 👀

viraptor:
[...]

PR #670, opened 2019-01-15
Adds method tracing helper

ronan-mch:
This PR adds a new module to `ddtrace`. `MethodTracing` allows developers to quickly add tracing to their classes without muddying up their code with trace blocks. 

## Example usage

```ruby
class FizzBuzz

  extend Datadog::MethodTracing
  trace_methods self, :fizz, :buzz

  ...
end
```

PR #5294, opened 2026-01-30
Add simple method tracing API

lloeki:
**What does this PR do?**

Add a simple method tracing API

**Motivation:**

A long overdue and oft-requested feature #1197 

**Change log entry**

Add simple method tracing API

**Additional Notes:**

- ~This uses [Graft](https://github.com/DataDog/graft-rb)~
- ~The intent is to bootstrap usage and kickstart implementation of an instrumentation API~

Related existing work:

- [attributes (.Net)](https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/dotnet/dd-api/#instrument-methods-through-attributes), [decorators (Python)](https://datadoghq.dev/dd-trace-api-py/pdocs/ddtrace_api.html#Tracer.wrap), [annotations (Java)](https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/java/dd-api#trace-annotations), [wrapping (Node)](https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/nodejs/dd-api/?tab=wrapper#creating-spans), [annotations (PHP)](https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/php/dd-api?tab=currentspan#annotations): Ruby has no attributes, decorators, or annotations, instead the idiomatic way is a DSL-ish method performing either AMC or prepend, which is what gets implemented here.
- [Instrument methods through configuration](https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/dotnet/dd-api/#instrument-methods-through-configuration) via `DD_TRACE_METHODS` (.Net, PHP, Java), [system properties (Java)](https://docs.datadoghq.com/tracing/trace_collection/custom_instrumentation/java/dd-api#datadog-trace-methods): this needs to transform a (list of) `String` into a constant + method; such an operation is carried out by Graft ([`HookPoint.parse`](https://github.com/DataDog/graft-rb/blob/53eee0053480184e22dcbf24de36ecfc457a8144/lib/graft/hook_point.rb#L28-L37), [`HookPoint.resolve_module`](https://github.com/DataDog/graft-rb/blob/53eee0053480184e22dcbf24de36ecfc457a8144/lib/graft/hook_point.rb#L51-L57)); the constants and methods can only be resolved if they are defined before configuration happens; this significantly increases complexity and is out of scope for this PR.

Things that these language define or allow setting:
- resource name (.Net, Java, Python, Node, PHP)
- (operation) name (.Net, Java, Python (as `name`), Node (as `name`), PHP (as `name`))
- no parent (Java) (creates new trace)
- service (Python, Node, PHP)
- tags (PHP)
- span type (Python, Node, PHP)

**How to test the change?**

- CI
- play with it

### DIFF
--- a/spec/datadog/kit/tracing/method_tracer_spec.rb
+++ b/spec/datadog/kit/tracing/method_tracer_spec.rb
@@ -194,16 +194,6 @@ RSpec.describe Datadog::Kit::Tracing::MethodTracer do
         expect(spans.count { |s| s.name == 'custom_dynamic_span' }).to eq(1)
       end
 
-      it 'does not trace dynamic method outside of a trace context' do
-        Datadog::Kit::Tracing::MethodTracer.trace_method(Dummy, :dynamic_method, dynamic: true)
-
-        result = dummy.dynamic_method
-
-        expect(dummy.called?).to be true
-        expect(result).to eq(42)
-        expect(spans.count { |s| s.name == 'Dummy#dynamic_method' }).to eq(0)
-      end
-
       it 'raises NoMethodError when dynamic method is not handled by method_missing' do
         Datadog::Kit::Tracing::MethodTracer.trace_method(Dummy, :unhandled_method, dynamic: true)
 
@@ -662,21 +652,6 @@ RSpec.describe Datadog::Kit::Tracing::MethodTracer do
   end
 
   describe '#trace_method' do
-    context 'outside of a trace context' do
-      it 'does not trace' do
-        dummy_class.instance_eval do
-          extend Datadog::Kit::Tracing::MethodTracer
-
-          trace_method :foo
-        end
-
-        dummy.foo
-
-        expect(dummy.called).to eq(1)
-        expect(spans.count { |s| s.name == 'Dummy#foo' }).to eq(0)
-      end
-    end
-
     context 'with dynamic: true for method_missing-handled methods' do
       let(:dummy_class) do
         Class.new do


lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,48 @@
...
+      module MethodTracer
+        class << self
+          # Trace a method by class and name
+          def trace_method(klass, method_name, span_name = nil)
+            raise ArgumentError, 'class must respond to :name' unless klass.respond_to?(:name)
+
+            hook_point = "#{klass.name}##{method_name}"
+            span_name ||= hook_point
+
+            hook_module = Module.new do
```
lloeki, line 67: Wondering if the module should define `self.name`, `self.to_s`, `self.inspect`, so that it's friendly in `ancestors` instead of a generic `#<Module: 0x12ca34fe>`

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,48 @@
...
+              define_method(method_name) { |*args, &block| ::Datadog::Tracing.trace(span_name) { super(*args, &block) } }
+              ruby2_keywords(method_name) if respond_to?(:ruby2_keywords)
+            end
+
+            klass.prepend(hook_module)
```
lloeki: Note that this makes it fundamentally incompatible with any method that has been AMC'd (alias method chain, whether it's Rails or manually).

Expect infinite loops otherwise.

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+module Datadog
+  module Kit
+    module Tracing
```
lloeki, line 5: This new API is in `Datadog::Kit::Tracing` instead of `Datadog::Tracing` to avoid addition of more public API endpoints outside of `Kit`.

The intent being to have all the public API reside in `Datadog::Kit`, and none anywhere else.

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,48 @@
...
+      #
+      # Or directly:
+      #
+      #     Datadog::Kit::Tracing::MethodTracer.trace_method(MyClass, :foo, 'my_span_name')
+      #
+      # Span name is optional and defaults to class#method
+      module MethodTracer
+        class << self
+          # Trace a method by class and name
+          def trace_method(klass, method_name, span_name = nil)
```
lloeki: Should this NOOP outside of a trace context instead of creating a fresh new trace?

lloeki:
Strange CI failures, apparently unrelated, like:

```
Datadog::Core::Metrics::Client
  is expected to have attributes {statsd: #<Double "statsd">}
  #count
    when #statsd is nil
      is expected not to have received count(:foo, kind of Numeric, {tags: ["language:ruby", "language-interpreter:ruby-x86_64-linux", "language-version:4.0.1", "tracer-version:2.28.0"]}) 0 times
    when #statsd is a Datadog::Statsd
      which raises an error
        behaves like logs an error without raising
          is expected not to raise Exception
      and given a block
        that yields args
          is expected to have received count(:foo, kind of Numeric, {tags: ["language:ruby", "language-interpreter:ruby-x86_64-linux", "language-version:4.0.1", "tracer-version:2.28.0"]}) 1 time
        that does not yield args
          behaves like logs an error without raising
            is expected not to raise Exception
      and given no options
        is expected to have received count(:foo, kind of Numeric, {tags: ["language:ruby", "language-interpreter:ruby-x86_64-linux", "language-version:4.0.1", "tracer-version:2.28.0"]}) 1 time
      and given options
        that are empty
          is expected to have received count(:foo, kind of Numeric, {tags: ["language:ruby", "language-interpreter:ruby-x86_64-linux", "language-version:4.0.1", "tracer-version:2.28.0"]}) 1 time
        that contain :tags
          is expected to have received count(:foo, kind of Numeric, {tags: ["language:ruby", "language-interpreter:ruby-x86_64-linux", "language-version:4.0.1", "tracer-version:2.28.0", "foo", "bar"]}) 1 time
          which are frozen
            is expected to have received count(:foo, kind of Numeric, {tags: ["language:ruby", "language-interpreter:ruby-x86_64-linux", "language-version:4.0.1", "tracer-version:2.28.0", "foo", "bar"]}) 1 time
        that are frozen
          is expected to have received count(:foo, kind of Numeric, {tags: ["language:ruby", "language-interpreter:ruby-x86_64-linux", "language-version:4.0.1", "tracer-version:2.28.0"]}) 1 time
  #time
    when #statsd is a Datadog::Statsd
      and given no options
        is expected to have received distribution(:foo, kind of Numeric, {tags: ["language:ruby", "language-interpreter:ruby-x86_64-linux", "language-version:4.0.1", "tracer-version:2.28.0"]}) 1 time
      and given a block
        is expected to yield control
        which raises an error
          is expected to raise #<RuntimeError: RuntimeError>
      which raises an error
        is expected not to raise Exception
      and given options
        that are frozen
          is expected to have received distribution(:foo, kind of Numeric, {tags: ["language:ruby", "language-interpreter:ruby-x86_64-linux", "language-version:4.0.1", "tracer-version:2.28.0"]}) 1 time
        that contain :tags
          is expected to have received distribution(:foo, kind of Numeric, {tags: ["language:ruby", "language-interpreter:ruby-x86_64-linux", "language-version:4.0.1", "tracer-version:2.28.0", "foo", "bar"]}) 1 time
          which are frozen
            is expected to have received distribution(:foo, kind of Numeric, {tags: ["language:ruby", "language-interpreter:ruby-x86_64-linux", "language-version:4.0.1", "tracer-version:2.28.0", "foo", "bar"]}) 1 time
        that are empty
          is expected to have received distribution(:foo, kind of Numeric, {tags: ["language:ruby", "language-interpreter:ruby-x86_64-linux", "language-version:4.0.1", "tracer-version:2.28.0"]}) 1 time
    when #statsd is nil
      is expected not to have received distribution(:foo, kind of Numeric, {tags: ["language:ruby", "language-interpreter:ruby-x86_64-linux", "language-version:4.0.1", "tracer-version:2.28.0"]}) 0 times
  #configure
    given options including
      :statsd
        is expected to change `metrics.statsd` from #<Double "statsd"> to #<InstanceDouble(Datadog::Statsd) (anonymous)>
      :enabled
        as true
          is expected to equal true
        as false
          is expected to equal false
  #default_statsd_client
    is expected to equal #<InstanceDouble(Datadog::Statsd) (anonymous)> (FAILED - 116)
    with Datadog::Statsd not loaded
      loads Datadog::Statsd library (FAILED - 117)
  #send_stats?
    when disabled
      example at ./spec/datadog/core/metrics/client_spec.rb:352 (FAILED - 118)
    when enabled
      and Statsd
        is initialized
          example at ./spec/datadog/core/metrics/client_spec.rb:362 (FAILED - 119)
        is nil
          example at ./spec/datadog/core/metrics/client_spec.rb:368 (FAILED - 120)
  #gauge
    when #statsd is nil
      example at ./spec/datadog/core/metrics/client_spec.rb:543 (FAILED - 121)
    when #statsd is a Datadog::Statsd
      which raises an error
        behaves like logs an error without raising
          example at ./spec/datadog/core/metrics/client_spec.rb:19 (FAILED - 122)
      and given no options
        example at ./spec/datadog/core/metrics/client_spec.rb:568 (FAILED - 123)
      and given a block
        that yields args
          example at ./spec/datadog/core/metrics/client_spec.rb:561 (FAILED - 124)
        that does not yield args
          behaves like logs an error without raising
            example at ./spec/datadog/core/metrics/client_spec.rb:19 (FAILED - 125)
      and given options
        that contain :tags
          example at ./spec/datadog/core/metrics/client_spec.rb:590 (FAILED - 126)
          which are frozen
            example at ./spec/datadog/core/metrics/client_spec.rb:595 (FAILED - 127)
        that are empty
          example at ./spec/datadog/core/metrics/client_spec.rb:577 (FAILED - 128)
        that are frozen
          example at ./spec/datadog/core/metrics/client_spec.rb:583 (FAILED - 129)
```

and summaries:

```
  1046) Tracer integration tests rule sampler with rule matching span with matching tags behaves like flushed trace 
        Got 2 failures:
        Shared Example Group: "flushed trace" called from ./spec/datadog/tracing/integration_spec.rb:353

        1046.1) Failure/Error: unused_statsd.each(&:close)
                  #<InstanceDouble(Datadog::Statsd) (anonymous)> was originally created in one example but has leaked into another example and can no longer be used. rspec-mocks' doubles are designed to only last for one example, and you need to create a new one in each example you wish to use it for.
```

lloeki:
Steep hits a:

```
2026-02-03 11:10:05.984: FATAL: [Steep 1.10.0] [typecheck:typecheck@1] [background] [#typecheck_source(path=lib/datadog/core/logger.rb)] [#type_check_file(lib/datadog/core/logger.rb@datadog)] [synthesize:(3:1)] [synthesize:(5:1)] [synthesize:(6:3)] [synthesize:(11:5)] [synthesize:(12:7)] [synthesize:(20:7)] [synthesize:(21:9)] [synthesize:(29:9)] [synthesize:(30:11)] [synthesize:(31:13)]
Unexpected error: #<NoMethodError: undefined method `selector' for an instance of Parser::Source::Map::Keyword>
```

EDIT: https://github.com/soutaro/steep/issues/2015

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Added a comment

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,64 @@
...
+      # Traced methods are only traced if already within a trace (i.e they do
+      # not create traces by themselves).
+      #
+      # Note: this uses Module#Prepend, so do not use on methods that have been
+      # alias method chained or you risk an infinite recusion crash.
+      module MethodTracer
+        class << self
+          # Trace an instance method by module and name
+          def trace_method(mod, method_name, span_name = nil)
+            raise ArgumentError, 'module name is nil' if mod.name.nil? && span_name.nil?
```
Strech: I think potentially name could be called on wrong object, WDYT if we guard it with `respond_to?`

```
irb(main):001* def trace_method(mod, method_name, span_name = nil)
irb(main):002*   raise ArgumentError, 'module name is nil' if mod.name.nil? && span_name.nil?
irb(main):003*   puts "OK"
irb(main):004> end
=> :trace_method
irb(main):005> trace_method("Some", 1, 2)
(irb):2:in `trace_method': undefined method `name' for an instance of String (NoMethodError)

  raise ArgumentError, 'module name is nil' if mod.name.nil? && span_name.nil?
```

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,64 @@
...
+
+            hook_point = "#{mod.name}##{method_name}"
+            span_name ||= hook_point
+
+            args = (RUBY_VERSION >= '2.7.') ? '...' : '*args, &block'
+
+            hook_module = Module.new do
+              # `args` is static, `method_name` is validated by the `method_defined?` check
+              # thus this `eval` is safe
+              eval(<<-RUBY, nil, __FILE__, __LINE__ + 1) # standard:disable Security/Eval
```
Strech, line 82: Question: Do you think it will be hard to go with `define_method` instead?

lloeki: It's not hard (see https://github.com/DataDog/dd-trace-rb/pull/5294/commits/f6cbbaa04939c73e37e43520d2c4086845a51fd4), but [it's slower than `...`](https://github.com/ruby/forwardable/commit/b606c3bf0a131bf2ebc14306a3d3ae2e6838fc4b) (which can't be used with a proc) + depends on `ruby2_keywords` (and its bugs across versions).

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Well, we're doing runtime typing now? ;)

TBH I don't think it's really useful (especially since it's documented in types) but I'm not too hung up on it, so if you feel strongly that it makes for a better devex because of the currently lousy error in that case I can add it :shrug:.

lloeki: Done

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,74 @@
...
+              def #{method_name}(#{args})
+                return super(#{args}) unless ::Datadog::Tracing.active_trace
+
+                ::Datadog::Tracing.trace('#{span_name}') { super(#{args}) }
+              end
+              RUBY
+            end
+
+            mod.prepend(hook_module)
```
lloeki, line 94: This makes non-public methods suddenly public...

lib/datadog/kit/tracing/method_tracer.rb
Strech: Gotcha, resolving this one 👍🏼

lib/datadog/kit/tracing/method_tracer.rb
p-datadog: Should these have `@api public` then?

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,74 @@
...
+      #       def foo; 'hello'; end
+      #
+      #       trace_method :foo, 'my_span_name'
+      #     end
+      #
+      # Or directly:
+      #
+      #     Datadog::Kit::Tracing::MethodTracer.trace_method(MyClass, :foo, 'my_span_name')
+      #
+      # Span name is optional and defaults to class#method
```
p-datadog: Does `class` in lowercase mean the class name will be lowercased?

p-datadog: Doesn't seem to be the case based on tests?

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,74 @@
...
+            args = (RUBY_VERSION >= '2.7.') ? '...' : '*args, &block'
+
+            hook_module = Module.new do
+              # `args` is static, `method_name` is validated by the `method_defined?` check
+              # thus this `eval` is safe
+              eval(<<-RUBY, nil, __FILE__, __LINE__ + 1) # standard:disable Security/Eval
+              def #{method_name}(#{args})
+                return super(#{args}) unless ::Datadog::Tracing.active_trace
+
+                ::Datadog::Tracing.trace('#{span_name}') { super(#{args}) }
```
p-datadog: `span_name` here is user input and should be escaped or sanitized, quotes in it will make this eval not compile

p-datadog:
I support `method_missing` handlers in DI instrumentation. This PR does not explicitly give reasons for rejecting them. I think this is a common use case and if it's not supported probably the reason should be written down for the customers who attempt it.

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Good catch, typo!

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Good catch as well

lloeki:
> I support method_missing handlers in DI instrumentation. This PR does not explicitly give reasons for rejecting them. I think this is a common use case and if it's not supported probably the reason should be written down for the customers who attempt it.

@p-datadog I agree this is interesting to support.

The reason why it wasn't is that the initial use case described was this DSL:

```
class Foo
  extend Datadog::Kit::Tracing::MethodTracer

  def foo = 42
  trace_method :foo
end
```

A corresponding usage with `method_missing` would be:

```
class Foo
  extend Datadog::Kit::Tracing::MethodTracer

  def method_missing(meth, *args, &block)
    case meth
    when :foo then 42
    else super
    end
  end

  trace_method :foo
  trace_method :bar
end
```

Since these methods are dynamic (virtual? is that a better term?), they "exist" only at runtime, and can only be checked for via `respond_to_missing?`, so even then they can't be checked for before an instance exists.

Therefore they can't be checked for existence at the class level, which is where the DSL operates.

With the current implementation that enforces the check, both `trace_method :foo` (handled by `method_missing`) and `trace_method :bar` (entirely unhandled) would raise an exception.

I am considering an implementation where the user can simply relax that existence (+visibility) check, since the check simply can't be done at the class level. Visibility doesn't matter anyway as `method_missing` is for public methods: one could theoretically implement private- an protected-like virtual methods but they would really need to go out of their way examining the caller bindings and raising appropriately but I feel that's overkill to address). In that case the hook module's  `super` will simply forward the message up the ancestor chain where the method does not exist and so fall back to invoke `method_missing`.

With that check relaxed, both `trace_method :foo` (handled by `method_missing`) and `trace_method :bar` (entirely unhandled) would stop raising an exception.

A side effect is that a `bar` method would start existing, and when called, traced, but then raise the standard `NoMethodError` exception. The behaviour is thus correct, although a) there's an extra span being generated and b) `respond_to?(:bar)` is `true`. But then there's also user error at the start, it's just that we can't really guard against it; at least the span error field will contain the exact location of the error and help pinpoint the problem.

An alternative considered was to hook on `method_missing` instead and check for the `meth` value, but then things like ActiveRecord's magic that dynamically _defines_ methods upon first invocation of `method_missing` would make it stop working since subsequent invocations would be on a defined method (and thus forever skip our `method_missing`).

So... the _other_ reason why is that it's more complex to handle and I focused on laying down the basics ;)

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Fixed.

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Strangely, from a `grep -R '@api' lib`, there seems to be no `@api public` around, only some `@api private` 🤔 

How odd, I could swear there were some.

lloeki: Duh, it's `@public_api`. Now since there are a few `@api private` I wonder which one we should use.

lloeki: Oh, wait: https://msp-greg.github.io/yard/YARD/CLI/Yardoc.html#:~:text=@api%20public

Have we been using this wrong the whole time?

lloeki:
Dug into steep with:

```console
$ find lib -name '*.rb' | while read -r f; do echo -n "#### $f"; out=$(steep check "$f" 2>&1); rc=$?; if [[ $rc -ne 0 || "$out" == *FATAL* ]]; then echo -e "\rFAIL $f"; echo "$f" >> steep.failed; else echo -e "\r  OK $f"; fi; done
-----8<----
$ cat steep.failed 
lib/datadog/kit/appsec/events/v2.rb
lib/datadog/tracing/contrib/rack/route_inference.rb
lib/datadog/core/logger.rb
```

Steep failures:

```console
$ steep check lib/datadog/core/logger.rb; echo "=> exit: $?"
2026-02-04 15:29:23.361: FATAL: [Steep 1.10.0] [typecheck:typecheck@6] [background] [#typecheck_source(path=lib/datadog/core/logger.rb)] [#type_check_file(lib/datadog/core/logger.rb@datadog)] [synthesize:(3:1)] [synthesize:(5:1)] [synthesize:(6:3)] [synthesize:(11:5)] [synthesize:(12:7)] [synthesize:(20:7)] [synthesize:(21:9)] [synthesize:(29:9)] [synthesize:(30:11)] [synthesize:(31:13)]
Unexpected error: #<NoMethodError: undefined method 'selector' for an instance of Parser::Source::Map::Keyword>
-----8<----
=> exit: 0
```

```console
$ steep check lib/datadog/tracing/contrib/rack/route_inference.rb; echo "=> exit: $?"
2026-02-04 15:27:49.863: FATAL: [Steep 1.10.0] [typecheck:typecheck@6] [background] [#typecheck_source(path=lib/datadog/tracing/contrib/rack/route_inference.rb)] [#type_check_file(lib/datadog/tracing/contrib/rack/route_inference.rb@datadog)] [synthesize:(3:1)] [synthesize:(4:3)] [synthesize:(5:5)] [synthesize:(6:7)] [synthesize:(9:9)] [synthesize:(10:11)] [synthesize:(27:11)] [synthesize:(28:13)] [synthesize:(28:13)] [synthesize:(30:13)]
Unexpected error: #<ArgumentError: wrong number of arguments (given 0, expected 3)>
-----8<----
=> exit: 0
```

```console
$ steep check lib/datadog/kit/appsec/events/v2.rb; echo "=> exit: $?"
2026-02-04 15:31:18.321: FATAL: [Steep 1.10.0] [typecheck:typecheck@7] [background] [#typecheck_source(path=lib/datadog/kit/appsec/events/v2.rb)] [#type_check_file(lib/datadog/kit/appsec/events/v2.rb@datadog)] [synthesize:(3:1)] [synthesize:(6:1)] [synthesize:(7:3)] [synthesize:(8:5)] [synthesize:(9:7)] [synthesize:(11:9)] [synthesize:(12:11)] [synthesize:(22:11)] [synthesize:(44:13)] [synthesize:(44:13)] [synthesize:(45:15)] [synthesize:(48:15)] [synthesize:(48:18)] [synthesize:(48:18)]
Unexpected error: #<RuntimeError: Unknown name for build_instance: Tracing::TraceSegment>
-----8<----
=> exit: 1
```

The latter has these type failures:

```
lib/datadog/kit/appsec/events/v2.rb:133:14: [error] The method cannot return a value of type `(::Datadog::Tracing::TraceSegment | nil)` because declared as type `(Tracing::TraceSegment | Tracing::TraceOperation | nil)`
│   (::Datadog::Tracing::TraceSegment | nil) <: (Tracing::TraceSegment | Tracing::TraceOperation | nil)
│     ::Datadog::Tracing::TraceSegment <: (Tracing::TraceSegment | Tracing::TraceOperation | nil)
│       ::Datadog::Tracing::TraceSegment <: Tracing::TraceSegment
│         ::Object <: Tracing::TraceSegment
│           ::BasicObject <: Tracing::TraceSegment
│
│ Diagnostic ID: Ruby::ReturnTypeMismatch
│
└               return Datadog::Tracing.active_trace unless Datadog::AppSec.active_context
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

lib/datadog/kit/appsec/events/v2.rb:132:16: [error] Cannot allow method body have type `(::Datadog::Tracing::TraceOperation | nil)` because declared as type `(Tracing::TraceSegment | Tracing::TraceOperation | nil)`
│   (::Datadog::Tracing::TraceOperation | nil) <: (Tracing::TraceSegment | Tracing::TraceOperation | nil)
│     ::Datadog::Tracing::TraceOperation <: (Tracing::TraceSegment | Tracing::TraceOperation | nil)
│       ::Datadog::Tracing::TraceOperation <: Tracing::TraceSegment
│         ::Object <: Tracing::TraceSegment
│           ::BasicObject <: Tracing::TraceSegment
│
│ Diagnostic ID: Ruby::MethodBodyTypeMismatch
│
└             def service_entry_trace
                  ~~~~~~~~~~~~~~~~~~~

lib/datadog/kit/appsec/events/v2.rb:141:14: [error] The method cannot return a value of type `(::Datadog::Tracing::SpanOperation | nil)` because declared as type `(Tracing::SpanOperation | nil)`
│   (::Datadog::Tracing::SpanOperation | nil) <: (Tracing::SpanOperation | nil)
│     ::Datadog::Tracing::SpanOperation <: (Tracing::SpanOperation | nil)
│       ::Datadog::Tracing::SpanOperation <: Tracing::SpanOperation
│         ::Object <: Tracing::SpanOperation
│           ::BasicObject <: Tracing::SpanOperation
│
│ Diagnostic ID: Ruby::ReturnTypeMismatch
│
└               return Datadog::Tracing.active_span unless Datadog::AppSec.active_context
                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

lib/datadog/kit/appsec/events/v2.rb:140:16: [error] Cannot allow method body have type `(::Datadog::Tracing::SpanOperation | nil)` because declared as type `(Tracing::SpanOperation | nil)`
│   (::Datadog::Tracing::SpanOperation | nil) <: (Tracing::SpanOperation | nil)
│     ::Datadog::Tracing::SpanOperation <: (Tracing::SpanOperation | nil)
│       ::Datadog::Tracing::SpanOperation <: Tracing::SpanOperation
│         ::Object <: Tracing::SpanOperation
│           ::BasicObject <: Tracing::SpanOperation
│
│ Diagnostic ID: Ruby::MethodBodyTypeMismatch
│
└             def service_entry_span
                  ~~~~~~~~~~~~~~~~~~

Detected 4 problems from 1 file
```

lib/datadog/kit/tracing/method_tracer.rb
ivoanjo: Note that we have some of our own yard magic that's using the `public_api` directly: https://github.com/DataDog/dd-trace-rb/blob/master/yard/extensions.rb

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,106 @@
...
+      #       extend Datadog::Kit::Tracing::MethodTracer
+      #
+      #       def foo; 'hello'; end
+      #
+      #       trace_method :foo, 'my_span_name'
+      #     end
+      #
+      # Or directly:
+      #
+      #     Datadog::Kit::Tracing::MethodTracer.trace_method(MyClass, :foo, 'my_span_name')
```
ivoanjo: Spotted this was probably changed in the code but not in the comment ;)

```suggestion
      #       trace_method :foo, span_name: 'optional_span_name'
      #     end
      #
      # Or directly:
      #
      #     Datadog::Kit::Tracing::MethodTracer.trace_method(MyClass, :foo, span_name: 'optional_span_name')
```

lib/datadog/kit/tracing/method_tracer.rb
ivoanjo: Minor: (Comment is outdated now?)

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,106 @@
...
+              raise NoMethodError, "undefined method #{method_name.inspect} for class #{mod}"
+            end
+
+            hook_point = "#{mod.name}##{method_name}"
+            custom_span_name = span_name
+            span_name ||= hook_point
+
+            unless span_name.is_a?(String)
+              raise ArgumentError, 'span name is not a String'
+            end
```
ivoanjo, line 63: Minor: A bit of Ruby duck typing + freezing would be perhaps in order here?

```suggestion
            span_name = span_name.to_str
```

spec/datadog/kit/tracing/method_tracer_spec.rb
```diff
@@ -0,0 +1,784 @@
...
+
+      it 'does not trace dynamic method outside of a trace context' do
+        Datadog::Kit::Tracing::MethodTracer.trace_method(Dummy, :dynamic_method, dynamic: true)
+
+        result = dummy.dynamic_method
+
+        expect(dummy.called?).to be true
+        expect(result).to eq(42)
+        expect(spans.count { |s| s.name == 'Dummy#dynamic_method' }).to eq(0)
+      end
```
ivoanjo: Minor: Since the "trace or not" codepath is always the same, I'd probably avoid repeating this test (we already cover this above). Same for other places below where we repeat this.

(In particular, I think repeating tests somewhat dilutes the important parts of things being tested as the longer the test the easier it is to "glaze over" and miss important details. The same would apply to repeating the naming tests, since the naming doesn't depend on the other details.)

lloeki:
```
$ steep check lib/datadog/core/logger.rb
Unexpected error: #<NoMethodError: undefined method 'selector' for an instance of Parser::Source::Map::Keyword>
```

Traced down to [this line](https://github.com/soutaro/steep/blob/v1.10.0/lib/steep/diagnostic/ruby.rb#L120)

```
      class InsufficientPositionalArguments < Base
        attr_reader :node
        attr_reader :params

        def initialize(node:, params:)
          send = case node.type
                 when :send, :csend
                   node
                 when :block, :numblock
                   node.children[0]
                 end

          loc = if send
                  send.loc.selector.with(end_pos: send.loc.expression.end_pos) # steep:ignore NoMethod
                else
                  node.loc.expression
                end

          super(node: node, location: loc)
          @params = params
        end

        def header_line
          "More positional arguments are required"
        end
      end
```

`node`:

```
s(:block,
  s(:zsuper),
  s(:args),
  s(:dstr,
    s(:str, "["),
    s(:begin,
      s(:send,
        s(:self), :progname)),
    s(:str, "] "),
    s(:begin,
      s(:lvar, :where)),
    s(:begin,
      s(:yield))))
```

`params`:

```
#<Steep::Interface::Function::Params:0x0000000129f13f18 @positional_params=#<Steep::Interface::Function::Params::PositionalParams:0x0000000129f13f68 @head=#<Steep::Interface::Function::Params::PositionalParams::Required:0x0000000129f13f90 @type=#<Steep::AST::Types::Any:0x0000000125554e90 @fvs=Set[]>>, @tail=#<Steep::Interface::Function::Params::PositionalParams:0x0000000129f14008 @head=#<Steep::Interface::Function::Params::PositionalParams::Optional:0x0000000129f14030 @type=#<Steep::AST::Types::Any:0x0000000125554e90 @fvs=Set[]>>, @tail=#<Steep::Interface::Function::Params::PositionalParams:0x0000000129f14058 @head=#<Steep::Interface::Function::Params::PositionalParams::Optional:0x0000000129f14080 @type=#<Steep::AST::Types::Any:0x0000000125554e90 @fvs=Set[]>>, @tail=nil>>>, @keyword_params=#<Steep::Interface::Function::Params::KeywordParams:0x0000000129f13f40 @requireds={}, @optionals={}, @rest=nil>>

```

`send`:

```
#<Parser::Source::Map::Keyword:0x0000000129a53668 @keyword=#<Parser::Source::Range lib/datadog/core/logger.rb 853...858>, @begin=nil, @end=nil, @expression=#<Parser::Source::Range lib/datadog/core/logger.rb 853...858>, @node=s(:zsuper)>
```

So the problem is `InsufficientPositionalArguments` on a `:zsuper`.

[Lots of `super` in there](https://github.com/DataDog/dd-trace-rb/blob/02463ef440c1bf778e9baf3d57171ba1d0bdd3f1/lib/datadog/core/logger.rb#L29-L39), but only one without positional args... and voilà:

```diff
diff --git a/lib/datadog/core/logger.rb b/lib/datadog/core/logger.rb
index c551660f98..14b2b704ad 100644
--- a/lib/datadog/core/logger.rb
+++ b/lib/datadog/core/logger.rb
@@ -28,7 +28,7 @@ module Datadog
 
         if message.nil?
           if block
-            super do
+            super(severity) do
               "[#{self.progname}] #{where}#{yield}"
             end
           else
```

Will report details properly upstream.

ivoanjo (approved):
👍 LGTM!

A few follow-up notes:

* I think it's worth copying in the PR description the usage example
* This API looks quite straightforward and easy to support, I don't expect any issues with adding it
* Thanks for including notes on how other Datadog SDKs already do it -- and agree that in general they all do something idiomatic to each language so I think the approach in this PR matches how we've been doing

lloeki:
```
$ steep check lib/datadog/tracing/contrib/rack/route_inference.rb
Unexpected error: #<ArgumentError: wrong number of arguments (given 0, expected 3)>
```

Traced down to:
- [this `type` def](https://github.com/soutaro/steep/blob/v1.10.0/lib/steep/ast/node/type_assertion.rb#L19):

    ```ruby
    class TypeAssertion
      def type(context, subtyping, type_vars)
    ```

- [this `type` call](https://github.com/soutaro/steep/blob/v1.10.0/lib/steep/type_inference/block_params.rb#L129):

    ```ruby
        node.children.each do |arg|
          case
          when arg.type == :mlhs
    ```

- [this `from_node` call](https://github.com/soutaro/steep/blob/v1.10.0/lib/steep/type_construction.rb#L3382):

    ```ruby
        constr = synthesize_children(node, skips: skips)
        if block_params
          # @type var node: Parser::AST::Node & Parser::AST::_BlockNode
          block_annotations = source.annotations(block: node, factory: checker.factory, context: nesting)

          constr.type_block_without_hint(
            node: node,
            block_params: TypeInference::BlockParams.from_node(block_params, annotations: block_annotations),
            block_annotations: block_annotations,
            block_body: block_body
          )
        end
    ```

So... it's choking around some block.

There's only [one block](https://github.com/DataDog/dd-trace-rb/blob/6c49ab130fa64b7f8bd2a20ca0f426ffa5178975/lib/datadog/tracing/contrib/rack/route_inference.rb#L27-L48):

```ruby
          def infer(path)
            segments = path.delete_prefix('/').split('/', MAX_NUMBER_OF_SEGMENTS + 1).first(MAX_NUMBER_OF_SEGMENTS)

            segments.map! do |segment| #: Array[String?]
              next if segment.empty?

              case segment
              when INT_PARAM_REGEX then '{param:int}'
              when INT_ID_PARAM_REGEX then '{param:int_id}'
              when HEX_PARAM_REGEX then '{param:hex}'
              when HEX_ID_PARAM_REGEX then '{param:hex_id}'
              when STRING_PARAM_REGEX then '{param:str}'
              else segment
              end
            end

            segments.compact! #: Array[String]

            "/#{segments.join("/")}"
          rescue
            nil
          end
```

Removing the annotation resolves the crash:

```
$ bundle exec steep check lib/datadog/tracing/contrib/rack/route_inference.rb
# Type checking files:

F

lib/datadog/tracing/contrib/rack/route_inference.rb:31:14: [error] Cannot break with a value of type `nil` because type `::String` is assumed
│   nil <: ::String
│
│ Diagnostic ID: Ruby::BreakTypeMismatch
│
└               next if segment.empty?
                ~~~~

```

The annotation on the block:
- crashes steep (will open issue upstream)
- is inconsistent with the one below

Annotations are there because `segments` is successively `Array[String]` then `Array[String?] then `Array[String]` in an effort to save one array allocation, at the cost of traversing twice.

So by iterating once in a new instance and types are then correct:

```
diff --git a/lib/datadog/tracing/contrib/rack/route_inference.rb b/lib/datadog/tracing/contrib/rack/route_inference.rb
index 65f40ff4c7..5f6869a938 100644
--- a/lib/datadog/tracing/contrib/rack/route_inference.rb
+++ b/lib/datadog/tracing/contrib/rack/route_inference.rb
@@ -27,10 +27,10 @@ module Datadog
           def infer(path)
             segments = path.delete_prefix('/').split('/', MAX_NUMBER_OF_SEGMENTS + 1).first(MAX_NUMBER_OF_SEGMENTS)
 
-            segments.map! do |segment| #: Array[String?]
+            segments.each_with_object([]) do |segment, a|
               next if segment.empty?
 
-              case segment
+              a << case segment
               when INT_PARAM_REGEX then '{param:int}'
               when INT_ID_PARAM_REGEX then '{param:int_id}'
               when HEX_PARAM_REGEX then '{param:hex}'
@@ -40,8 +40,6 @@ module Datadog
               end
             end
 
-            segments.compact! #: Array[String]
-
             "/#{segments.join("/")}"
           rescue
             nil
```

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Wow, that is... utterly confusing.

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Good catch, will address

lib/datadog/kit/tracing/method_tracer.rb
lloeki: It is, I thought I marked it resolved.

lib/datadog/kit/tracing/method_tracer.rb
lloeki: I don't think "be liberal in what you accept" is the good path here: most probably a non-`String` (even if `String`able) is a usage mistake, so I'm veering on the safer side.

- If people desire passing a non-string as a string here they can be explicit about the conversion at the call site
- We can always relax if we hear complaints, but the other way around is trickier as it is a breaking change

spec/datadog/kit/tracing/method_tracer_spec.rb
lloeki: Agreed.

That said, I kind of kept the repetition and regularity intentionally to get a feel of how/whether a shared example abstracting a way common behaviour would be useful.

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,106 @@
...
+      # Trace methods easily
+      #
+      # Use as DSL module:
+      #
+      #     class MyClass
+      #       extend Datadog::Kit::Tracing::MethodTracer
+      #
+      #       def foo; 'hello'; end
+      #
+      #       trace_method :foo, 'my_span_name'
```
y9v: this doesn't need to be placed after the `def foo` declaration, right?

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,106 @@
...
+      #
+      # Span name is optional and defaults to 'Class#method'
+      #
+      # Traced methods are only traced if already within a trace (i.e they do
+      # not create traces by themselves).
+      #
+      # Dynamic methods (e.g via `method_missing`) can be traced via `dynamic: true`
+      # by relaxing method existence sanity checks.
+      #
+      # Note: this uses Module#Prepend, so do not use on methods that have been
```
y9v: `Note:` should be probably upcased to be properly identified as a tag?

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,106 @@
...
+          # Trace an instance method by module and name
+          #
+          # @param mod [Module] module or class containing the method to trace
+          # @param method_name [Symbol] name of the method to trace
+          # @param span_name [String, nil] optional span name (defaults to "Module#method")
+          # @param dynamic [Boolean] if true, skip method existence check (for method_missing-handled methods)
+          # @return [void]
+          def trace_method(mod, method_name, span_name: nil, dynamic: false)
+            raise ArgumentError, 'mod is not a module' unless mod.is_a?(Module)
+            raise ArgumentError, 'module name is nil' if mod.name.nil? && span_name.nil?
```
y9v: might be too much for an error message, but should we also provide some info that `span_name` might be passed to fix this when module name is nil?

lib/datadog/tracing/contrib/rack/route_inference.rb
```diff
@@ -40,8 +40,6 @@ def infer(path)
               end
             end
 
-            segments.compact! #: Array[String]
-
```
y9v, line 44: why are those changes in this PR? It's totally unrelated, or am I missing something?

lib/datadog/tracing/contrib/rack/route_inference.rb
```diff
@@ -27,10 +27,10 @@ def read_or_infer(request_env)
           def infer(path)
             segments = path.delete_prefix('/').split('/', MAX_NUMBER_OF_SEGMENTS + 1).first(MAX_NUMBER_OF_SEGMENTS)
 
-            segments.map! do |segment| #: Array[String?]
+            segments.each_with_object([]) do |segment, a|
```
y9v: this will create a new array here - and `map!` was used to reduce allocations

lib/datadog/kit/tracing/method_tracer.rb
lloeki: It does because the method has to exist (sanity check + making it public vs private, etc); same as `private :foo` & al.

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Ah that was not supposed to be a tag, just a callout for the user

lib/datadog/kit/tracing/method_tracer.rb
lloeki: drat that was on my TODO list and I forgot about addressing that

lib/datadog/tracing/contrib/rack/route_inference.rb
lloeki: unrelated, but steep was failing, which blocked the PR, so I went on to fix typing

lib/datadog/tracing/contrib/rack/route_inference.rb
lloeki: See commit message https://github.com/DataDog/dd-trace-rb/pull/5294/commits/cd374327f0136e1cea4ba01957c7ea234dde6f20: it's a tradeoff between allocating and traversing twice (`map!` + `compact!`)

If you're THAT concerned about allocations then look at the line before this one ;) and then probably the whole thing should be rewritten to parse and emit segments in one go.

ivoanjo:
I don't consider that blocking feedback, but I do think it's pretty weird to include all the extra unrelated fixes/changes to code in this PR, my suggestion is to extract them to a separate PR.

lib/datadog/kit/tracing/method_tracer.rb
y9v: I would mention this in the documentation - for me this wouldn't be obvious

lib/datadog/tracing/contrib/rack/route_inference.rb
lloeki: Seems like the change is broken anyway 🤦 

```
  6) Datadog::Tracing::Contrib::Rack::RouteInference.infer removes empty segments from path
     Failure/Error: expect(described_class.infer('/foo//1/bar')).to eq('/foo/{param:int}/bar')

       expected: "/foo/{param:int}/bar"
            got: "/foo//1/bar"

       (compared using ==)
     # ./spec/datadog/tracing/contrib/rack/route_inference_spec.rb:75:in `block (3 levels) in <top (required)>'
     # ./spec/datadog/tracing/contrib/support/tracer_helpers.rb:96:in `block (2 levels) in <module:TracerHelpers>'
     # ./spec/spec_helper.rb:277:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:155:in `block (2 levels) in <top (required)>'
     # /usr/local/bundle/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
     # /usr/local/bundle/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block (2 levels) in <top (required)>'
     # ./spec/support/execute_in_fork.rb:32:in `run'

  7) Datadog::Tracing::Contrib::Rack::RouteInference.infer replaces params containing at least 6 hexadecimal digits
     Failure/Error: expect(described_class.infer('/foo/bar/FFF111')).to eq('/foo/bar/{param:hex}')

       expected: "/foo/bar/{param:hex}"
            got: "/foo/bar/FFF111"

       (compared using ==)
     # ./spec/datadog/tracing/contrib/rack/route_inference_spec.rb:58:in `block (3 levels) in <top (required)>'
     # ./spec/datadog/tracing/contrib/support/tracer_helpers.rb:96:in `block (2 levels) in <module:TracerHelpers>'
     # ./spec/spec_helper.rb:277:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:155:in `block (2 levels) in <top (required)>'
     # /usr/local/bundle/gems/webmock-3.13.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <top (required)>'
     # /usr/local/bundle/gems/rspec-wait-0.0.9/lib/rspec/wait.rb:46:in `block (2 levels) in <top (required)>'
     # ./spec/support/execute_in_fork.rb:32:in `run'
```

lloeki:
> I don't consider that blocking feedback, but I do think it's pretty weird to include all the extra unrelated fixes/changes to code in this PR, my suggestion is to extract them to a separate PR.

The reason is in essence that these code changes are because steep crashes internally, recovers, but also stops validating some files because the crash had them fail; it just so happens that one of these crashes was triggered by a code change in this PR.

I will:
- add Steepfile ignores so that steep does not crash internally anymore (which is somewhat identical to the behaviour with the internal crash: the file is inconsistently type-checked, if at all)
- extract the unrelated fixes to another PR, in which the ignores will thus be able to be un-ignored.

lib/datadog/tracing/contrib/rack/route_inference.rb
y9v: thank you for attempting to fix this anyway! We had many files ignored in steep and I haven't noticed that the file I am adding is also being ignored

y9v (approved):
thank you for this PR! This will be a really nice feature

lib/datadog/tracing/contrib/rack/route_inference.rb
lloeki: I don't think you missed anything.

It's not so much that it is manually ignored (in `Steepfile`, which it is not... well, I added it because of it crashing) but that. the internal crash in steep while parsing the annotation causes steep to kinda ignore (some? all? of) the file's contents.

lloeki:
- Opened: https://github.com/DataDog/dd-trace-rb/pull/5319
- Reverted changes and simply ignore the problematic files

lib/datadog/tracing/contrib/rack/route_inference.rb
lloeki: Reverted and opened https://github.com/DataDog/dd-trace-rb/pull/5319

lib/datadog/kit/tracing/method_tracer.rb
lloeki: I feel like it sort of is because it behaves exactly like `public`/`private`/`protected` keywords:

```
irb(main):001* class Foo
irb(main):002*   private :foo
irb(main):003*   def foo; end
irb(main):004> end
(irb):2:in 'Module#private': undefined method 'foo' for class 'Foo' (NameError)
	from (irb):2:in '<class:Foo>'
	from (irb):1:in '<main>'
	from <internal:kernel>:168:in 'Kernel#loop'
	from /Users/loic.nageleisen/Source/github.com/DataDog/dd-trace-rb/vendor/bundle/ruby/3.4.0/gems/irb-1.16.0/exe/irb:9:in '<top (required)>'
	from /Users/loic.nageleisen/Source/github.com/DataDog/dd-trace-rb/vendor/bundle/ruby/3.4.0/bin/irb:25:in 'Kernel#load'
	from /Users/loic.nageleisen/Source/github.com/DataDog/dd-trace-rb/vendor/bundle/ruby/3.4.0/bin/irb:25:in '<main>'
```

... or any other method decorator pattern really (because these could - and very much mandatorily used to - be implemented via alias method chaining, which needs the method defined at "DSL call" time). I mean that's how things work in Ruby since there's no first-party Python-like decorator/C#-like attribute/Java-like annotation (or rather, dynamism is such a first-party construct of Ruby that these first-party thingies are entirely unneeded).

Also the provided example hints at the order.

**Anyway** I was going to add something like "needs to be after `def` because existence is checked for except for dynamic methods" when I realised there's this already:

```
      # Dynamic methods (e.g via `method_missing`) can be traced via `dynamic: true`
      # by relaxing method existence sanity checks.
```

Which does mention existence checks; in a sense putting `trace_method` before `def` is relying on it being dynamically defined later on (that "later" being immediately after, but it's still later in a runtime sense).

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Added.

lib/datadog/kit/tracing/method_tracer.rb
lloeki: done

lib/datadog/kit/tracing/method_tracer.rb
lloeki: done

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Improved doc to clarify

lib/datadog/kit/tracing/method_tracer.rb
lloeki: implemented

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,110 @@
...
+      #       extend Datadog::Kit::Tracing::MethodTracer
+      #
+      #       def foo; 'hello'; end
+      #
+      #       trace_method :foo, span_name: 'optional_span_name'
+      #     end
+      #
+      # Or directly:
+      #
+      #     Datadog::Kit::Tracing::MethodTracer.trace_method(MyClass, :foo, span_name: 'optional_span_name')
```
mabdinur, line 20: Should we update this method to support setting resource names and tags? I can see this being the next feature request

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,110 @@
...
+                name || "#<Datadog::Tracing::Kit::MethodTracer(#{method_name.inspect}#{suffix})>"
+              end
+
+              # this `eval` is safe:
+              # - `args` is static
+              # - `method_name` is validated by the `method_defined?` check
+              # - `span_name` is validated to be a String and inspected to be quoted
+              eval(<<-RUBY, nil, __FILE__, __LINE__ + 1) # standard:disable Security/Eval
+              def #{method_name}(#{args})
+                return super(#{args}) unless ::Datadog::Tracing.active_trace
```
mabdinur, line 84: The proposed Ruby behavior differs from dd-trace-py. What’s the motivation for this? In Python, decorators create a span even when there’s no active trace. Would letting the method tracer start a trace break existing instrumentation?

mabdinur (approved):
left some questions, overall LGTM thanks for picking this up 😄

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Yes, I was kind of punting on this and going minimalist given that the list of existing things across languages differs quite a lot (see PR description).

Since this might generate a longer discussion, I would like to propose merging this minimal PR as is and getting the ball rolling to extend functionality in a second one.

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Safety and defensiveness.

It's very easy and quite common for Ruby scripts to include bits from other places that belong to full-fledged apps. This would then generate an swarm of spurious, unattached, confusing spans out of nowhere at best, or a swarm of logs that a script can't send spans to a nonexistent agent at worst.

Plus it's easier to relax this behaviour later than to tighten it up while people started depending on it.

Note about an idea: this functionality could be gated by an argument kind of like the "no parent" argument Java has. This way generating such traces outside of the usual app scope is an voluntary choice explicitly made by the user.

But maybe see https://github.com/DataDog/dd-trace-rb/pull/5294#discussion_r2771217574 regarding discussing these features.

spec/datadog/kit/tracing/method_tracer_spec.rb
lloeki: Removed

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Tracking: https://github.com/DataDog/dd-trace-rb/issues/5323

lib/datadog/kit/tracing/method_tracer.rb
lloeki: Tracking: https://github.com/DataDog/dd-trace-rb/issues/5324

lib/datadog/kit/tracing/method_tracer.rb
```diff
@@ -0,0 +1,110 @@
...
+      # Trace methods easily
+      #
+      # Use as DSL module:
+      #
+      #     class MyClass
+      #       extend Datadog::Kit::Tracing::MethodTracer
+      #
+      #       def foo; 'hello'; end
+      #
+      #       trace_method :foo, span_name: 'optional_span_name'
```
bquorning, line 15: It would be nice to also see an example of how to add tracing to a class method.

lloeki: Fair point.

Given that:

- "class methods" don't exist! They are merely methods of instances of the class `Class`
- but we wouldn't want a given method to be hooked for _every single class_!

... one would [straight up hook onto that instance's singleton](https://github.com/DataDog/dd-trace-rb/blob/47df115b5c3175d38850d0b1411d78626336da4a/spec/datadog/kit/tracing/method_tracer_spec.rb#L103-L134), i.e `MyClass.singleton_class`:

```ruby
class MyClass
  class << self
    def foo; 'hello'; end
  end
end

Datadog::Kit::Tracing::MethodTracer.trace_method(MyClass.singleton_class, :foo, span_name: 'MyClass.foo')
```

Alternatively, this works too:

```ruby
class MyClass
  class << self
    extend Datadog::Kit::Tracing::MethodTracer

    def foo; 'hello'; end
    trace_method :foo, span_name: 'MyClass.foo'
  end
end
```

Since this singleton class is a dynamic class, `span_name` must be provided.

A convenience DSL-ish class-context `trace_class_method` (or maybe better: `trace_singleton_class_method` could be added:

```ruby
def trace_singleton_class_method(method_name, span_name: nil, dynamic: false)
  span_name ||= "#{name}.#{method_name}" if respond_to?(:name)

  Datadog::Kit::Tracing::MethodTracer.trace_method(self.singleton_class, :foo, span_name: span_name, dynamic: false)
```

I ~will open~ opened an [issue to track that](https://github.com/DataDog/dd-trace-rb/issues/5333).

bquorning: > A convenience DSL-ish class-context `trace_class_method` (or maybe better: `trace_singleton_class_method` could be added

👍🏼 Yes, I was looking for something along those lines. (It also matches the API that we suggested way back in #670).

PR #5293, opened 2026-01-29
Revert "Revert "Add DBM back propagation with container tags and process tags""

wantsui:
Reverts DataDog/dd-trace-rb#5290 to bring back the original: https://github.com/DataDog/dd-trace-rb/pull/5208

Do not merge this PR unless the below PR can be merged:
- https://github.com/DataDog/dd-trace-rb/pull/5296

#### FROM THE ORIGINAL PR DESCRIPTION

**What does this PR do?**

The goal of this is to add container tags support to DBM, which also includes the process tags: https://docs.google.com/document/d/15GtNOKGBCt6Dc-HsDNnMmCdZwhewFQx8yUlI9in5n3M/edit?tab=t.0#heading=h.fyt7z8ehztcm.

To achieve this, the following changes were made
1. The existing FNV hash logic was refactored out so it could be reused for this update
2. AgentInfo was updated to compute a hash if the Datadog Agent returns a `Datadog-Container-Tags-Hash` header. The hash combines process_tags and container_tags_hash.
3. When agent_info is received, compute a new hash if the container information is new
4. Injects `ddsh='...'` as a SQL comment
5. Injects `_dd.propagated_hash` as a span "meta"

**Motivation:**

The Python PR does this for DBM and DSM in one go: 
- https://github.com/DataDog/dd-trace-py/pull/15293

The Java PR does this in two:
- https://github.com/DataDog/dd-trace-java/pull/9156
- https://github.com/DataDog/dd-trace-java/pull/9282

For Ruby, I am adding all the functionality needed for the back propagation for DBM first then will do DSM in a later PR.

**Change log entry**

No.

**Additional Notes:**

I am not sure if this requires a changelog entry.

**How to test the change?**

```
bundle exec rspec \
spec/datadog/core/utils/fnv_spec.rb \
spec/datadog/core/environment/agent_info_spec.rb \
spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
```

I also sent data from this branch to our backend and under query samples with a default number in place of the propagation hash to represent both the process tags and containers, ie: https://app.datadoghq.com/databases/samples

The result is that:

1) the db comment has ddsh added:
```
"/*dde='dev',ddps='dbm-sql-comment-demo',ddpv='0.1',ddh='postgres',dddb='postgres',ddsh='1234567890',dddbs='pg',traceparent='<some string>'*/"
```
2) the "span" gets `"_dd.propagated_hash": 1234567890` under `meta`

PR #5296, opened 2026-01-30
fix: add controls over the back propagation setting for DBM

wantsui:
**What does this PR do?**

For AIDM-258.

https://github.com/DataDog/dd-trace-rb/pull/5208 had the issue of not having enough controls over when the injection was applied. We need a way for users to control when the SQL comment gets injected or not, and to add the propagation_checksum even if the only thing available is the process tags.

This PR applies the controls in the below manner:

| DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED | DD_DBM_INJECT_SQL_BASEHASH | APM Span has _dd.propagated_hash | APM Span has _dd.tags.process | DBM Query Sample has ddsh |
|:----------------------------------------------:|:--------------------------:|:--------------------------------:|:-----------------------------:|:-------------------------:|
| TRUE                                           | TRUE                       | Yes                              | Yes                           | Yes                       |
| TRUE                                           | FALSE                      | No                              | Yes                            | No                        |
| FALSE                                          | FALSE                      | No                               | No                            | No                        |
| FALSE                                          | TRUE                       | No                               | No                            | No                        |

In order to achieve this:
1. the logic in agent_info had to be refactored to compute the propagated hash even if there is no container tag available.
2. DD_DBM_INJECT_SQL_BASEHASH was added to the config options for the three dbm related integrations today.
3. Tests were adjusted to be able test the new behavior

**Motivation:**

While reviewing https://github.com/DataDog/dd-trace-js/pull/7212, I realized I did it wrong for Ruby.

**Change log entry**

No.

**Additional Notes:**

Only merge this PR if https://github.com/DataDog/dd-trace-rb/pull/5293 can be merged first.

**How to test the change?**

See AIDM-258  for thorough screenshots of the combinations of this feature.

Note: DD_DBM_INJECT_SQL_BASEHASH is confusing as a setting name, but it's already been implemented in other tracers so it'll be more work to change this now. I added a todo in the comments to evaluate changing this in the future.

Note 2: A future todo is to write simplified tests following rspec standards better following https://github.com/DataDog/dd-trace-rb/pull/5296#discussion_r2774915911.

Note 3: Regarding the comment about having this in the changelog, since the feature is still requires backend/frontend type work, I'm going to remove it from the changelog until we're in a state where we can enable this by default to avoid causing confusion about these settings that we don't want folks to use yet.


lib/datadog/core/environment/agent_info.rb
```diff
@@ -116,17 +119,23 @@ def ==(other)
...
-          if new_container_tags_value && new_container_tags_value != @container_tags_checksum
-            @container_tags_checksum = new_container_tags_value
+          # If there was no previous propagation_checksum, then we should calculate the checksum by checking the agent and getting process info
+          if @propagation_checksum.nil? || (new_container_tags_value && new_container_tags_value != @container_tags_checksum)
+            @container_tags_checksum = new_container_tags_value if new_container_tags_value
 
             process_tags = Process.serialized
-            # new_container_tags_value (non nil) over @container_tags_checksum (string?) to avoid steep errors
-            data = process_tags + new_container_tags_value
+            data = process_tags
```
vandonr: why not `data = Process.serialized` directly ? It looks clear enough to me, using a variable just for a single usage on the following line looks wrong to me

vandonr:
nit: PR description is wrong, values for the columns `APM Span has _dd.propagated_hash` and `APM Span has _dd.tags.process` are swapped

lib/datadog/core/environment/agent_info.rb
wantsui: addressed in 6eba1d21fb456d76e5d9ed0c25bbc4569dbf77c9! Now it's three lines (but I can't do anything about the steep error so it had to be split):
```
            data = Process.serialized
            # Add container tags if available (helps Steep with type narrowing)
            container_tags = @container_tags_checksum
            data += container_tags if container_tags
```

y9v:
I think it would make sense to add a changelog entry for this?

lib/datadog/core/environment/agent_info.rb
```diff
@@ -116,17 +119,22 @@ def ==(other)
...
+
           header_value = res.headers[Core::Transport::Ext::HTTP::HEADER_CONTAINER_TAGS_HASH]
           new_container_tags_value = header_value if header_value && !header_value.empty?
 
           # if the Trace Agent returns a new value for the checksum, calculate and cache the propagation checksum
-          if new_container_tags_value && new_container_tags_value != @container_tags_checksum
-            @container_tags_checksum = new_container_tags_value
+          # If there was no previous propagation_checksum, then we should calculate the checksum by checking the agent and getting process info
+          if @propagation_checksum.nil? || (new_container_tags_value && new_container_tags_value != @container_tags_checksum)
+            @container_tags_checksum = new_container_tags_value if new_container_tags_value
```
Strech: There is already check on non-nil value of the `new_container_tags_value` right above that line

```suggestion
            @container_tags_checksum = new_container_tags_value
```

spec/datadog/core/environment/agent_info_spec.rb
```diff
@@ -56,153 +56,220 @@
...
-        allow(Datadog::Core::Environment::Process).to receive(:serialized).and_call_original
+          allow(response).to receive(:headers).and_return({'Datadog-Container-Tags-Hash' => 'test'})
+          agent_info.fetch
 
-        agent_info.fetch
-        first_checksum = agent_info.propagation_checksum
+          # When container tags are available, the new checksum has the container tags
+          new_value = agent_info.propagation_checksum
+          expect(new_value).not_to be_nil
+          expect(new_value).to be_a(Integer)
```
Strech: They are mutually exclusive, I would recommend to check what you expect, otherwise, you also need to check that it's not string and no float and not etc, etc if you do "what I'm not expecting" checks

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
```diff
@@ -44,28 +45,90 @@
...
+
+      it 'does not set the propagated hash (_dd.propagated_hash) span tag' do
         described_class.annotate!(span_op, propagation_mode)
-        expect(span_op.get_tag('_dd.propagated_hash')).to eq('1234567890')
+        expect(span_op.get_tag('_dd.propagated_hash')).to be_nil
       end
     end
 
-    context 'when the base hash is not present' do
+    context 'when DD_DBM_INJECT_SQL_BASEHASH=true and DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED=true but propagation_checksum returns nil' do
```
Strech: To be honest the whole point is to describe it in human readable form, because rest we can pick up from the well written test (if that's the case). If you feel that you need to express that here - maybe it makes sense to rewrite the test

lib/datadog/core/environment/agent_info.rb
```diff
@@ -105,6 +99,15 @@ def ==(other)
         # https://github.com/DataDog/datadog-agent/pull/38515
         attr_reader :container_tags_checksum
 
+        # Computes the propagation checksum (base hash) from process tags and optionally container tags when it changes
+        # Controlled by DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED
+        # This is needed in traces (dsm and dbm related spans), DBM, and DSM.
+        # The base hash is used for correlation across signals (traces, DBM, data streams, etc.) in environments
+        # where full context propagation isn't feasible due to size restrictions.
```
marcotc: I think these two lines should be moved to the public interface, for visibility of someone looking `propagation_checksum`, and why it exists.

lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -14,9 +14,18 @@ module SqlComment
...
-            # Add base hash to the span tag
-            base_hash = Datadog.send(:components).agent_info.propagation_checksum
-            span_op.set_tag(Ext::TAG_PROPAGATED_HASH, base_hash.to_s) if base_hash
+            config = Datadog.configuration
+
+            # Add base hash to the span tag if both flags are enabled:
+            # 1. DD_DBM_INJECT_SQL_BASEHASH set globally or via the per db integration inject_sql_basehash flag
+            # 2. DD_PROPAGATE_PROCESS_TAGS_ENABLED set globally
+            # This check matches the behavior in Python
+            # https://github.com/DataDog/dd-trace-py/blob/5e94be5c97b42060e5800e35d8fa41472fb8c569/ddtrace/propagation/_database_monitoring.py#L89
```
marcotc: I'm not sure if we need the bullet points 1 and 2, since the code:
```ruby
if mode.inject_sql_basehash? && config.experimental_propagate_process_tags_enabled
```
is pretty much that, and quite legible.

I think that the comment should tells us things that we wouldn't learn by simply reading the code.
Which for this case would be: what the whole `if` block is doing (e.g. "Add DBM propagation hash when enabled and available"); other developer comments (e.g. "This matches the behavior in the Python tracer:").

lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -14,9 +14,18 @@ module SqlComment
...
+            # TODO: Check if we actually want inject_sql_basehash to control a span tag since the name only implies sql comments
```
marcotc: Regarding `# TODO: Check if we actually want...`, is it reasonable to do this before we merge the PR, or would this be an uncertainty we must live with for a bit? We can go either way, but it's good to know which one we are dealing with here.

lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -40,8 +49,15 @@ def self.prepend_comment(sql, span_op, trace_op, mode)
               Ext::KEY_PEER_SERVICE => peer_service,
             }
 
-            base_hash = Datadog.send(:components).agent_info.propagation_checksum
-            tags[Ext::KEY_BASE_HASH] = base_hash.to_s if base_hash
+            # Add base hash to SQL comment if both flags are enabled:
+            # 1. DD_DBM_INJECT_SQL_BASEHASH set globally or via the per db integration inject_sql_basehash flag
+            # 2. DD_PROPAGATE_PROCESS_TAGS_ENABLED set globally
+            # This matches the behavior in Python:
+            # # https://github.com/DataDog/dd-trace-py/blob/5e94be5c97b42060e5800e35d8fa41472fb8c569/ddtrace/propagation/_database_monitoring.py#L139
```
marcotc: Same as the comment in `self.annotate!`

lib/datadog/tracing/contrib/propagation/sql_comment.rb
wantsui: I ended up posting a PR description table to capture the current combination and effects that @vandonr reviewed so I can remove this TODO now! It just seemed odd to me to call it "inject_sql_basehash" but it was also adding data to the span tag, but the alternative of having another env var to control the span tag is out of scope. I am under the impression that this is also temporary.

I am going through all the comments one by one and will remove this TODO :)

spec/datadog/core/environment/agent_info_spec.rb
```diff
@@ -56,153 +56,220 @@
       end
     end
 
-    context 'when fetch has populated the value' do
+    context 'when DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED is false' do
```
marcotc: When possible, let's use names that interpret meaning of this `context`, so that devs can reason about it from a high-level:
```suggestion
    context 'when process tags propagation is disabled' do
```
We can then ask questions like, both today and in the future: "Did we set the correct flags to enable this propagation?", "Can we remove the conditional checks?".

spec/datadog/core/environment/agent_info_spec.rb
```diff
@@ -56,153 +56,220 @@
...
+          allow(response).to receive(:headers).and_return({})
+          agent_info.fetch
 
-        expected_checksum = Datadog::Core::Utils::FNV.fnv1_64(process_tags + 'value')
-        expect(new_value).to eq(expected_checksum)
-      end
+          # First call has no headers yet, so it only checksums the process tags
+          first_checksum = agent_info.propagation_checksum
+          expect(first_checksum).not_to be_nil
+          expect(first_checksum).to eq(Datadog::Core::Utils::FNV.fnv1_64(process_tags))
```
marcotc: `FNV.fnv1_64` can never return `nil`:
```suggestion
          expect(first_checksum).to eq(Datadog::Core::Utils::FNV.fnv1_64(process_tags))
```

spec/datadog/core/environment/agent_info_spec.rb
```diff
@@ -56,153 +56,220 @@
...
+          # When container tags are available, the new checksum has the container tags
+          new_value = agent_info.propagation_checksum
+          expect(new_value).not_to be_nil
+          expect(new_value).to be_a(Integer)
+          expect(new_value).not_to eq(first_checksum)
 
-        agent_info.fetch
-        second_checksum = agent_info.propagation_checksum
+          expected_checksum = Datadog::Core::Utils::FNV.fnv1_64(process_tags + 'test')
+          expect(new_value).to eq(expected_checksum)
```
marcotc:
```suggestion
          expect(new_value).not_to eq(first_checksum)
          expect(new_value).to eq(Datadog::Core::Utils::FNV.fnv1_64(process_tags + 'test'))
```

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
```diff
@@ -44,28 +45,90 @@
...
       let(:mode) { 'service' }
+      let(:inject_sql_basehash) { true }
 
       before do
         allow(Datadog.send(:components)).to receive(:agent_info).and_return(agent_info)
         allow(agent_info).to receive(:propagation_checksum).and_return(1234567890)
       end
 
-      it 'sets the propagated hash on the span tag as string' do
+      context 'and DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED=true' do
```
marcotc, line 57: Same here regarding using high-level context names.

lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -40,8 +49,15 @@ def self.prepend_comment(sql, span_op, trace_op, mode)
...
+            if mode.inject_sql_basehash? && config.experimental_propagate_process_tags_enabled
+              base_hash = Datadog.send(:components).agent_info.propagation_checksum
+              tags[Ext::KEY_BASE_HASH] = base_hash.to_s if base_hash
```
marcotc: I noticed we change the nomenclature between `hash` and `base_hash` when using the checksum here in the SQL propagator, and I was wondering it if it's actually a "base" hash, or just a hash, and I found this in the spec: `The suggested key name for this value is ‘ddsh’ (dd service hash).`

I don't see a reference for the nomenclature "base" in the spec.
Unless I'm missing something, I think we should remove the base nomenclature from the references here.
And if I'm mistaken, I probably want to what what "base" is here :)

spec/datadog/core/environment/agent_info_spec.rb
wantsui: That's fair, I simplified the test to only check `expect(result).to eq(Datadog::Core::Utils::FNV.fnv1_64(process_tags))` now in 778cde064554563eb7d3bf4ccf7e027ce93332d6 and removed the other expectations since that's my main goal.

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
wantsui: I refactored in https://github.com/DataDog/dd-trace-rb/commit/778cde064554563eb7d3bf4ccf7e027ce93332d6 to remove descriptions that relied heavily on environment variable names. Let me know if this looks better now!

lib/datadog/core/environment/agent_info.rb
wantsui: Refactored in https://github.com/DataDog/dd-trace-rb/commit/778cde064554563eb7d3bf4ccf7e027ce93332d6

lib/datadog/tracing/contrib/propagation/sql_comment.rb
wantsui: I removed the bullets in https://github.com/DataDog/dd-trace-rb/commit/778cde064554563eb7d3bf4ccf7e027ce93332d6 :)

lib/datadog/tracing/contrib/propagation/sql_comment.rb
wantsui: Removed the TODOs in https://github.com/DataDog/dd-trace-rb/commit/778cde064554563eb7d3bf4ccf7e027ce93332d6

lib/datadog/tracing/contrib/propagation/sql_comment.rb
wantsui: Refactored in https://github.com/DataDog/dd-trace-rb/commit/778cde064554563eb7d3bf4ccf7e027ce93332d6

spec/datadog/core/environment/agent_info_spec.rb
wantsui: I couldn't accept this change in the GH UI for some reason but I replicated this change in https://github.com/DataDog/dd-trace-rb/commit/778cde064554563eb7d3bf4ccf7e027ce93332d6

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
wantsui: Thanks for the feedback. See https://github.com/DataDog/dd-trace-rb/commit/778cde064554563eb7d3bf4ccf7e027ce93332d6, which now goes back to more meaningful human names.

lib/datadog/tracing/contrib/propagation/sql_comment.rb
wantsui: I see what you mean, in https://github.com/DataDog/dd-trace-rb/commit/778cde064554563eb7d3bf4ccf7e027ce93332d6, this has been renamed to `propagation_hash` if that's any better?

lib/datadog/tracing/contrib/propagation/sql_comment.rb
vandonr: no the `inject_sql_basehash` is not temporary, it's here to stay. Thinking that it also controls whether a tag is added to the span is true but the wrong way to frame it. Having the tag on the span is a necessary consequence of having the hash in the query. The logical viewpoint is more :
`inject_sql_basehash` -> controls if base hash is inserted in query
`if hash injected in query` -> then we need to write it in the span to be able to reconciliate the entities in the backend.
Idk if it's clear, and it's really purely a philosophical difference,  but it may make you feel more at peace with that setting "controlling 2 things" ^^

spec/datadog/core/environment/agent_info_spec.rb
Strech: This is also an anti-pattern here, you are expected to know the result, because what if in your computations mistake, then your test is going to verify that computed with mistake value is equal to mistakenly computed value. Hence the check should be explicit without computations used in the method itself.

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
Strech: Well, you have changed description - that is good, but the test doesn't explain it - this is still bad. But I would say if no one else objecting - I'm fine (but personally, I would rewrite spec)

lib/datadog/core/environment/agent_info.rb
```diff
@@ -78,9 +78,11 @@ def ==(other)
           other.is_a?(self.class) && other.agent_settings == agent_settings
         end
 
-        # Returns the propagation checksum, with part of the data from the Agent.
+        # Returns the propagation checksum, comprising of process tags and optionally container tags (from the Trace Agent)
         # Currently called/used by the DBM code to inject the propagation checksum into the SQL comment.
         #
+        # This hash is used for correlation across signals (traces, DBM, data streams, etc.) in environments
```
p-datadog:
```suggestion
        # This checksum is used for correlation across signals (traces, DBM, data streams, etc.) in environments
```
I assume you are just referring to the thing we are calling "checksum" and not a digest?

lib/datadog/core/environment/agent_info.rb
```diff
@@ -105,6 +101,13 @@ def ==(other)
         # https://github.com/DataDog/datadog-agent/pull/38515
         attr_reader :container_tags_checksum
 
+        # Computes the propagation checksum from process tags and optionally container tags when it changes
```
p-datadog:
```suggestion
        # Computes the propagation checksum from process tags and optionally container tags when process tags change
```
It is not clear what "it" is referring to, if it's "process tags" I made the suggested change

lib/datadog/core/environment/agent_info.rb
```diff
@@ -105,6 +101,13 @@ def ==(other)
...
+        # Controlled by DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED
+        # This is needed in traces (dsm and dbm related spans), DBM, and DSM.
+        #
+        # Only runs when  is true.
```
p-datadog: This line is confusing - what is true and what is running? Do you mean the method is invoked in some circumstances only or the method only does meaningful work in some circumstances only (but it would still "run", technically)

lib/datadog/core/environment/agent_info.rb
```diff
@@ -116,17 +119,22 @@ def ==(other)
...
+        def update_propagation_checksum(res)
+          return unless Datadog.configuration.experimental_propagate_process_tags_enabled
+
           header_value = res.headers[Core::Transport::Ext::HTTP::HEADER_CONTAINER_TAGS_HASH]
           new_container_tags_value = header_value if header_value && !header_value.empty?
 
           # if the Trace Agent returns a new value for the checksum, calculate and cache the propagation checksum
-          if new_container_tags_value && new_container_tags_value != @container_tags_checksum
+          # If there was no previous propagation_checksum, then we should calculate the checksum by checking the agent and getting process info
+          if @propagation_checksum.nil? || (new_container_tags_value && new_container_tags_value != @container_tags_checksum)
```
p-datadog: The "value" to "checksum" comparison is weird - is the "value" itself a "checksum"? If so can it be called "checksum"?

lib/datadog/core/environment/agent_info.rb
```diff
@@ -116,17 +119,22 @@ def ==(other)
...
             @container_tags_checksum = new_container_tags_value
 
-            process_tags = Process.serialized
-            # new_container_tags_value (non nil) over @container_tags_checksum (string?) to avoid steep errors
-            data = process_tags + new_container_tags_value
+            data = Process.serialized
```
p-datadog: "data" is very non-descriptive. This PR does not change the `serialized` method, therefore, does `serialized` still return process tags?

lib/datadog/core/environment/agent_info.rb
```diff
@@ -116,17 +119,22 @@ def ==(other)
...
+            # Add container tags if available (helps Steep with type narrowing)
+            container_tags = @container_tags_checksum
+            data += container_tags if container_tags
```
p-datadog: Based on this line I can suggest renaming "data" to "container_and_process_tags" or "all_tags" or something along these lines.

lib/datadog/core/environment/agent_info.rb
```diff
@@ -116,17 +119,22 @@ def ==(other)
...
+          # If there was no previous propagation_checksum, then we should calculate the checksum by checking the agent and getting process info
+          if @propagation_checksum.nil? || (new_container_tags_value && new_container_tags_value != @container_tags_checksum)
             @container_tags_checksum = new_container_tags_value
 
-            process_tags = Process.serialized
-            # new_container_tags_value (non nil) over @container_tags_checksum (string?) to avoid steep errors
-            data = process_tags + new_container_tags_value
+            data = Process.serialized
+            # Add container tags if available (helps Steep with type narrowing)
+            container_tags = @container_tags_checksum
```
p-datadog: This is also a weird line to read because a checksum (scalar value, integer(? I think) in this case) is very different from tags (array of strings). What is the actual value here?

p-datadog (commented):
Is "basehash" actually a technical term or just a weird spelling of "base hash"?

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
```diff
@@ -44,28 +45,90 @@
...
+          Datadog.configure do |c|
+            c.experimental_propagate_process_tags_enabled = false
+          end
+          example.run
+          without_warnings { Datadog.configuration.reset! }
+        end
+
+        it 'does not set the propagated hash (_dd.propagated_hash) span tag' do
+          described_class.annotate!(span_op, propagation_mode)
+          expect(span_op.get_tag('_dd.propagated_hash')).to be_nil
```
marcotc, line 86: `_dd.propagated_hash` won't be fully nil, since Process.serialized can never return `nil`.

lib/datadog/core/environment/agent_info.rb
marcotc: This comparison of a `*_value` variable vs a `*_checksum` variable make it sound like we are comparing two different things: `new_container_tags_value != @container_tags_checksum`

lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -14,9 +14,15 @@ module SqlComment
...
-            base_hash = Datadog.send(:components).agent_info.propagation_checksum
-            span_op.set_tag(Ext::TAG_PROPAGATED_HASH, base_hash.to_s) if base_hash
+            config = Datadog.configuration
+
+            # Add DBM propagation hash when enabled and available.
+            # This matches the behavior in the Python tracer:
+            # https://github.com/DataDog/dd-trace-py/blob/5e94be5c97b42060e5800e35d8fa41472fb8c569/ddtrace/propagation/_database_monitoring.py#L89
+            if mode.inject_sql_basehash? && config.experimental_propagate_process_tags_enabled
+              propagation_hash = Datadog.send(:components).agent_info.propagation_checksum
+              span_op.set_tag(Ext::TAG_PROPAGATED_HASH, propagation_hash.to_s) if propagation_hash
```
marcotc:
```suggestion
              checksum = Datadog.send(:components).agent_info.propagation_checksum
              span_op.set_tag(Ext::TAG_PROPAGATED_HASH, checksum.to_s) if checksum
```

lib/datadog/tracing/contrib/propagation/sql_comment.rb
```diff
@@ -40,8 +46,13 @@ def self.prepend_comment(sql, span_op, trace_op, mode)
...
             }
 
-            base_hash = Datadog.send(:components).agent_info.propagation_checksum
-            tags[Ext::KEY_BASE_HASH] = base_hash.to_s if base_hash
+            # Add DBM propagation hash to SQL comment when enabled and available.
+            # This matches the behavior in the Python tracer:
+            # https://github.com/DataDog/dd-trace-py/blob/5e94be5c97b42060e5800e35d8fa41472fb8c569/ddtrace/propagation/_database_monitoring.py#L139
+            if mode.inject_sql_basehash? && config.experimental_propagate_process_tags_enabled
+              propagation_hash = Datadog.send(:components).agent_info.propagation_checksum
+              tags[Ext::KEY_BASE_HASH] = propagation_hash.to_s if propagation_hash
```
marcotc:
```suggestion
              checksum = Datadog.send(:components).agent_info.propagation_checksum
              tags[Ext::KEY_BASE_HASH] = checksum.to_s if checksum
```

lib/datadog/core/environment/agent_info.rb
wantsui: I refactored this section in ea1b503ec7a93cfe28403d45714c41b95786d1c4 to `checksum_input` since it goes into the calculation if that's any better:

```
if @propagation_checksum.nil? || (new_container_tags_checksum && new_container_tags_checksum != @container_tags_checksum)
            @container_tags_checksum = new_container_tags_checksum

            checksum_input = Process.serialized
            # Use local variable for Steep type narrowing (helps Steep with type narrowing)
            container_checksum = @container_tags_checksum
            checksum_input += container_checksum if container_checksum

            @propagation_checksum = Core::Utils::FNV.fnv1_64(checksum_input)
          end
```

lib/datadog/core/environment/agent_info.rb
wantsui: Thanks for the feedback, I changed up the variable names in ea1b503ec7a93cfe28403d45714c41b95786d1c4. Feel free to suggest better names!

spec/datadog/core/environment/agent_info_spec.rb
wantsui: Thanks, I went over this with @marcotc today and I ended up with mocks of some of the behavior to focus on what is being tested in b859616fc122b6fc816cdff14220fc6a508ef806 .
Let me know if this addresses your concerns better!

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
wantsui: Fixed in 00b618eb6d2078394362b74d1cfd3f5802acfd90 because you're right this hash will always return something (falls back to process tags if there are no container tags).

lib/datadog/core/environment/agent_info.rb
wantsui: In https://github.com/DataDog/dd-trace-rb/pull/5296/changes/ea1b503ec7a93cfe28403d45714c41b95786d1c4, I refactored to
```
        # Setting DD_EXPERIMENTAL_PROPAGATE_PROCESS_TAGS_ENABLED to true enables the following behavior:
        #  - Computes a propagation checksum from process tags and optionally container tags when tags change
        #  - This is needed in traces (dsm and dbm related spans), DBM, and DSM.
```


Open to suggestions :)

p-datadog (approved):

wantsui:
> Is "basehash" actually a technical term or just a weird spelling of "base hash"?

Following up on @p-datadog 's question - I added a TODO in the code and noted in the PR description noting that we should look into changing **DD_DBM_INJECT_SQL_BASEHASH** to a better name in the future across the tracers.

For now, it's out of scope for this PR since it's already been implemented in the other tracers :)

lib/datadog/core/environment/agent_info.rb
wantsui: resolved in https://github.com/DataDog/dd-trace-rb/commit/ea1b503ec7a93cfe28403d45714c41b95786d1c4

lib/datadog/core/environment/agent_info.rb
wantsui: Addressed/refactored in https://github.com/DataDog/dd-trace-rb/commit/ea1b503ec7a93cfe28403d45714c41b95786d1c4

spec/datadog/core/environment/agent_info_spec.rb
```diff
@@ -56,153 +56,201 @@
...
-      it 'returns the same cached value on subsequent calls' do
-        first_result = agent_info.propagation_checksum
-        second_result = agent_info.propagation_checksum
+      it 'returns nil when container tags are not present' do
+        allow(response).to receive(:headers).and_return({})
+        agent_info.fetch
 
-        expect(first_result).to eq(second_result)
-        expect(first_result).to be_a(Integer)
+        expect(agent_info.propagation_checksum).to be_nil
```
Strech: This should be written with a transition check, like this:

```ruby
expect { agent_info.fetch }.not_to change { agent_info.propagation_checksum }.from(nil)
```

spec/datadog/core/environment/agent_info_spec.rb
```diff
@@ -56,153 +56,201 @@
...
-        result = agent_info.propagation_checksum
-        expect(result).to be_a(Integer)
-        expect(result).to eq(Datadog::Core::Utils::FNV.fnv1_64('process:tagstest'))
+        expect(agent_info.propagation_checksum).to be_nil
       end
 
-      it 'returns the same cached value on subsequent calls' do
-        first_result = agent_info.propagation_checksum
-        second_result = agent_info.propagation_checksum
+      it 'returns nil when container tags are not present' do
```
Strech: I think here is better to describe what is going on instead, and maybe define a context

```ruby
context 'when container tags are not present' do
  before { allow(response).to receive(:headers).and_return({}) }

  it 'does not set propagation checksum' do
    # ...
  end
end
```

wantsui: Fixed in c9052d1e21796101aa5ee65bbcdd4e615b3d04eb

spec/datadog/core/environment/agent_info_spec.rb
wantsui: Addressed in c9052d1e21796101aa5ee65bbcdd4e615b3d04eb

lib/datadog/tracing/contrib/propagation/sql_comment.rb
marcotc: Ok, i suggest extracting `mode.inject_sql_basehash? && config.experimental_propagate_process_tags_enabled` into a method, say `inject_hash?`, so that the if now becomes `if inject_hash?(mode, config)` and it's clear that it's the same condition in all places in this file.

wantsui: Done in 78efb35713edef4ac0990a7585cbdb61acc01cc0

spec/datadog/core/environment/agent_info_spec.rb
```diff
@@ -56,153 +56,202 @@
...
-        expect(agent_info.propagation_checksum).to be_nil
-      end
+        it 'returns nil before fetch' do
+          expect(agent_info.propagation_checksum).to be_nil
+        end
 
-      it 'returns nil after failed fetch' do
-        agent_info.fetch
-        expect(agent_info.propagation_checksum).to be_nil
+        it 'returns nil after failed fetch' do
```
Strech, line 119:
```
it { expect { agent.fetch }.not_to { agent_info.propagation_checksum }.from(nil) }
```

wantsui: Thank for walking through this with me today! I'll mark this as resolved for now, and left a comment in the PR description to improve the tests in the future. I agree with you that the tests can be improved (and I'll try to improve them as a side test as an exercise)!

spec/datadog/tracing/contrib/propagation/sql_comment_spec.rb
wantsui: Thanks for the feedback. I'll revisit improving the tests in the future!

wantsui:
Note: Regarding the comment about having this in the changelog, since the feature is still requires backend/frontend type work, I'm going to remove it from the changelog until we're in a state where we can enable this by default to avoid causing confusion about these settings that we don't want folks to use yet.

PR #5300, opened 2026-02-02
Add `waf.init` and `waf.updates` telemetry for AppSec

y9v:
**What does this PR do?**
This PR adds reporting of `waf.init` and `waf.updates` telemetry metrics for AppSec product.

According to the RFC, `waf.init` is reported when AppSec is enabled; and `waf.updates` is reported each time when WAF ruleset was updated, including startup.

**Motivation:**
We are still missing some important telemetry metrics for AppSec.

**Change log entry**
None. This is internal change.

**Additional Notes:**
APPSEC-60123

**How to test the change?**
CI and manual testing.

### DIFF
--- a/spec/datadog/appsec/security_engine/engine_spec.rb
+++ b/spec/datadog/appsec/security_engine/engine_spec.rb
@@ -68,7 +68,7 @@ RSpec.describe Datadog::AppSec::SecurityEngine::Engine do
             event_rules_version: '1.0.0',
             success: 'true'
           }
-        ).once
+        )
 
         engine
       end
@@ -110,7 +110,7 @@ RSpec.describe Datadog::AppSec::SecurityEngine::Engine do
             event_rules_version: '',
             success: 'false'
           }
-        ).once
+        )
 
         expect { engine }.to raise_error(Datadog::AppSec::WAF::LibDDWAFError)
       end
@@ -174,7 +174,7 @@ RSpec.describe Datadog::AppSec::SecurityEngine::Engine do
             event_rules_version: '',
             success: 'false'
           }
-        ).once
+        )
 
         expect { engine }.to raise_error(Datadog::AppSec::WAF::LibDDWAFError)
       end
@@ -682,7 +682,7 @@ RSpec.describe Datadog::AppSec::SecurityEngine::Engine do
           event_rules_version: '1.0.0',
           success: 'true'
         }
-      ).once
+      )
 
       engine.reconfigure!
     end
@@ -730,7 +730,7 @@ RSpec.describe Datadog::AppSec::SecurityEngine::Engine do
             event_rules_version: '2.0.0',
             success: 'false'
           }
-        ).once
+        )
 
         engine.reconfigure!
       end


lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -40,12 +40,30 @@ def initialize(appsec_settings:, telemetry:)
...
           @handle_ref = ThreadSafeRef.new(@waf_builder.build_handle)
+
+          telemetry.inc(
+            Ext::TELEMETRY_METRICS_NAMESPACE, 'waf.init', 1,
+            tags: {
+              waf_version: Datadog::AppSec::WAF::VERSION::BASE_STRING,
+              event_rules_version: @ruleset_version.to_s,
+              success: 'true'
+            }
+          )
```
Strech: Feels like a lot of copy-paste? What about method extraction? I think it totally makes sense

spec/datadog/appsec/component_spec.rb
```diff
@@ -47,6 +48,7 @@
       it 'returns nil when security engine fails to instantiate' do
         settings.appsec.ruleset = {}
 
+        allow(telemetry).to receive(:inc)
```
Strech: Shouldn't we assert something here?

spec/datadog/appsec/context_spec.rb
```diff
@@ -160,6 +161,7 @@
 
     context 'when a run was a failure' do
       before do
+        allow(telemetry).to receive(:inc)
```
Strech: Isn't top-level allow not enough?

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -15,26 +15,105 @@
...
       before do
         appsec_settings.ruleset = {}
+
+        allow(Datadog.logger).to receive(:error)
+        allow(telemetry).to receive(:report)
+        allow(telemetry).to receive(:inc)
       end
 
-      it 'reports error though telemetry, prints an error log message and re-raises' do
+      it 'reports error through telemetry' do
```
Strech, line 86: Sorry, but what's a point of splitting?

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -71,11 +150,34 @@
...
       end
 
-      it 'reports errors through telemetry' do
+      it 'reports error through telemetry' do
         expect(telemetry).to receive(:error).with("missing key 'conditions': [invalid-rule-id]")
 
         expect { engine }.to raise_error(Datadog::AppSec::WAF::LibDDWAFError)
       end
+
+      it 'prints an error log message' do
```
Strech, line 159: Maybe it's better to merge?

Strech (commented):
I would give it another round of refactoring, but direction is good 👍🏼

lib/datadog/appsec/security_engine/engine.rb
y9v: :+1: extracted to a function [9564810](https://github.com/DataDog/dd-trace-rb/pull/5300/commits/9564810c6582987f560199d0922e74324c02ede4)

lib/datadog/appsec/security_engine/engine.rb
```diff
@@ -103,17 +107,39 @@ def reconfigure!
...
+            'updates',
+            success: false,
+            ruleset_version: @reconfigured_ruleset_version,
+            telemetry: AppSec.telemetry
+          )
         end
 
         private
 
+        def increment_waf_metric(metric_name, success:, ruleset_version:, telemetry:)
```
Strech: I think `increment_waf_metric` exposes too much on what is going on. Could it be just `increment_metric` or even without `increment`, because with `increment` we expose that something will be added somewhere, but is it really important for the caller?

Also it's a bit weird that we pass the telemetry to the method that suppose to encapsulate it.

y9v: renamed to `#metric`. Regarding passing telemetry - unfortunately we have to, since in `#initialize` method we cannot use `AppSec.telemetry`, and need to use the telemetry instance passed to `#initialize` as an argument.
This is one of the reasons I didn't extract those calls into a function the first time.

spec/datadog/appsec/component_spec.rb
y9v: good catch - we don't need this actually. Will remove it

lib/datadog/appsec/security_engine/engine.rb
y9v: [b46924f](https://github.com/DataDog/dd-trace-rb/pull/5300/commits/b46924f458ebfdb08800f60c3ed60d04067e157a)

spec/datadog/appsec/context_spec.rb
y9v: [a1af9bd](https://github.com/DataDog/dd-trace-rb/pull/5300/commits/a1af9bdb687d862cf924bf2271e2c661d0698e90)

spec/datadog/appsec/component_spec.rb
y9v: [a1af9bd](https://github.com/DataDog/dd-trace-rb/pull/5300/commits/a1af9bdb687d862cf924bf2271e2c661d0698e90)

spec/datadog/appsec/security_engine/engine_spec.rb
y9v: well, I am adding a third expectation here, which would make this test something like "it reports error through telemetry, reports waf.init metric, prints an error log message and re-raises".  I thought that having more atomic tests here would be better for readability

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -15,26 +15,105 @@
...
+      end
+
+      it 'reports waf.init metric once with correct tags' do
+        expect(telemetry).to receive(:inc).with(
+          Datadog::AppSec::Ext::TELEMETRY_METRICS_NAMESPACE, 'waf.init', 1, tags: {
+            waf_version: Datadog::AppSec::WAF::VERSION::BASE_STRING,
+            event_rules_version: '1.0.0',
+            success: 'true'
+          }
+        ).once
```
Strech:
```suggestion
        )
```

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -15,26 +15,105 @@
...
+          tags: hash_including(success: 'true')
+        )
+
+        expect(telemetry).to receive(:inc).with(
+          Datadog::AppSec::Ext::TELEMETRY_METRICS_NAMESPACE, 'waf.init', 1, tags: {
+            waf_version: Datadog::AppSec::WAF::VERSION::BASE_STRING,
+            event_rules_version: '',
+            success: 'false'
+          }
+        ).once
```
Strech:
```suggestion
        )
```

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -71,11 +150,34 @@
...
```
Strech:
```suggestion
        )
```

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -572,12 +674,25 @@
...
 
+    it 'reports waf.updates metric with success: true' do
+      expect(Datadog::AppSec.telemetry).to receive(:inc).with(
+        Datadog::AppSec::Ext::TELEMETRY_METRICS_NAMESPACE, 'waf.updates', 1,
+        tags: {
+          waf_version: Datadog::AppSec::WAF::VERSION::BASE_STRING,
+          event_rules_version: '1.0.0',
+          success: 'true'
+        }
+      ).once
```
Strech:
```suggestion
      )
```

spec/datadog/appsec/security_engine/engine_spec.rb
```diff
@@ -593,14 +708,32 @@
...
+        )
+
+        expect(Datadog::AppSec.telemetry).to receive(:inc).with(
+          Datadog::AppSec::Ext::TELEMETRY_METRICS_NAMESPACE, 'waf.updates', 1,
+          tags: {
+            waf_version: Datadog::AppSec::WAF::VERSION::BASE_STRING,
+            event_rules_version: '2.0.0',
+            success: 'false'
+          }
+        ).once
```
Strech:
```suggestion
        )
```

PR #5302, opened 2026-02-02
[APPSEC-60450] Introduce AppSec::Context state

Strech:
**What does this PR do?**

This is an intermediate step to an architecture where we are able to accumulate during the request/response lifecycle data within a context and share it between contribs and allow delayed usage (no immediately use it, but via event/hook).

**Motivation:**

I need to be able to have some stats for the duration of the request/response and we also want to detach data accumulation from data usage. In contribs we immediately call WAF, that prevents us from instrumenting different areas and access more desired data, instead we instrument where we have processed and complete data.

Instead we want to accumulate it and be able to use at any moment and share it within request lifetime.

In the upcoming PR that state will be used to keep some thread-local counters.

**Change log entry**

No.

**Additional Notes:**

I've also kicked few files from `Steepignore` to boost out typing. Just a low-hanging fruits.

**How to test the change?**

CI

### DIFF
--- a/lib/datadog/appsec/context.rb
+++ b/lib/datadog/appsec/context.rb
@@ -22,6 +22,9 @@ module Datadog
 
       # Shared mutable storage for counters, flags, and data accumulated during
       # the request's lifecycle.
+      #
+      # NOTE: This attribute is a subject to change, but in a current form
+      #       it's a `Hash`-like structure.
       attr_reader :state
 
       class << self


lib/datadog/appsec/context.rb
```diff
@@ -4,14 +4,25 @@
...
       # Steep: https://github.com/soutaro/steep/issues/1880
       ActiveContextError = Class.new(StandardError) # steep:ignore IncompatibleAssignment
 
       # TODO: add delegators for active trace span
-      attr_reader :trace, :span, :events
+      attr_reader :trace, :span
+
+      # Shared mutable storage for counters, flags, and data accumulated during
+      # the request's lifecycle.
+      attr_reader :state
```
p-datadog, line 28: Would be helpful to indicate the data type (looks to be a hash based on constructor).

p-datadog (approved):

lib/datadog/appsec/context.rb
Strech: This is a subject to change, at the beginning we are satisfied with PORO Hash, but it might change once we start refactoring our events system

Strech: I will add a note about that (also RBS highlights that)

y9v (approved):

